PR #37, opened 2013-11-02
Add forked to indicator to query

shayel:
Shows where a query was forked to an from (my first time writing both django and angular: apologies if I'm not doing it right)

* Add "parent_id" field to show where a query was cloned from
* Add parent and forked queries to query (fiddle) screen


arikfr:
I was keeping this pull request open as a reminder to implement this functionality, but realized that I won't do it any time soon, and it started to become ridiculous. So with heavy heart, I'm closing this...

Sorry, @shayel.

PR #214, opened 2014-05-14
Adding map and heatmap visualizations

christophervalles:
Worth saying that you need two columns on the resultset for this to work one should be named "lat" and another "lon"


arikfr:
Closing in favor of #466. Apologies for leaving it floating for so long.

PR #347, opened 2015-01-12
Query params for filters

barnash:
I made it possible to register filters values in the query string - for dashboard view.
For example:
http://localhost:9001/dashboard/mydashboard?filter1=4&filter2=5

### DIFF
--- a/rd_ui/app/scripts/controllers/dashboard.js
+++ b/rd_ui/app/scripts/controllers/dashboard.js
@@ -1,5 +1,5 @@
 (function() {
-  var DashboardCtrl = function($scope, Events, Widget, $routeParams, $http, $timeout, $q, Dashboard) {
+  var DashboardCtrl = function($scope, Events, Widget, $routeParams, $location, $http, $timeout, $q, Dashboard) {
     $scope.refreshEnabled = false;
     $scope.refreshRate = 60;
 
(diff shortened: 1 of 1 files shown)


rd_ui/app/scripts/controllers/dashboard.js
```diff
@@ -32,22 +32,23 @@
                 // TODO: first object should be a copy, otherwise one of the chart filters behaves different than the others.
                 filters[filter.name] = filter;
                 filters[filter.name].originFilters = [];
+                if (_.has($location.search(), filter.friendlyName)) {
```
arikfr: How about we use `filter.name`? It has more chances to be more URL friendly (in most cases it's the table column name -> no capital letters and no spaces).

barnash:
@arikfr Yeah, you are right, I thought that it will just be easy to create the url since the friendly name is right there when you want to create he deep link. Ofc name makes much more sense. Changed.

arikfr:
Merged. @barnash congrats on your first contribution :-)

arikfr:
@barnash two things I overlooked:
1. Following this change, dashboard filters are enabled for all dashboards regardless of the feature flag. This is fixed in: 16f0413af8e15153586f55028548b284f400f3c7
2. `filter.name` is in the format of "filter-name::filter", while I thought it's only "filter-name"... This is no so friendly, so I might change it in the future to make it easier.

barnash:
Cool, sounds good.

Issue #307, opened 2014-10-15
Browser notifications no longer work


arikfr:
Looks like `window.webkitNotification` is now `window.Notification` in newer versions of Chrome/Firefox.

PR #356, opened 2015-01-21
Bug #307 Fix: Notifications stopped working

alexanderlz:
Updated to use the newer Notification object (as described in MDN)

### DIFF
--- a/rd_ui/app/scripts/services/notifications.js
+++ b/rd_ui/app/scripts/services/notifications.js
@@ -32,7 +32,6 @@
 
             //using the 'tag' to avoid showing duplicate notifications
             var notification = new Notification(title, {'tag': title+content, 'body': content});
-            lastNotification = notification;
             notification.onclick = function () {
                 window.focus();
                 this.cancel();
--- a/rd_ui/app/scripts/services/notifications.js
+++ b/rd_ui/app/scripts/services/notifications.js
@@ -1,10 +1,9 @@
 (function () {
     var notifications = function (Events) {
         var notificationService = {};
-        var lastNotification = null;
 
         notificationService.isSupported = function () {
-            if (window.webkitNotifications) {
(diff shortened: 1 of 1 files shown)


joeysim:
Where's the like button?

rd_ui/app/scripts/services/notifications.js
```diff
@@ -27,23 +30,13 @@
                 return;
             }
 
-            if (document.webkitVisibilityState && document.webkitVisibilityState == 'visible') {
```
arikfr: This is here to make sure we don't show notifications when the page is in focus. Is there a reason you removed it?

rd_ui/app/scripts/services/notifications.js
```diff
@@ -27,23 +30,13 @@
...
-            }
-
-            if (lastNotification) {
-                lastNotification.cancel();
-            }
-
-            var notification = window.webkitNotifications.createNotification('', title, content);
-            lastNotification = notification;
+            //using the 'tag' to avoid showing duplicate notifications
+            var notification = new Notification(title, {'tag': title+content, 'body': content});
```
arikfr: Why aren't you using the title anymore?

rd_ui/app/scripts/services/notifications.js
alexanderlz: To emphasize the effect of notifying!
the truth is I removed it while testing to always see the notifications, and decided it's not that bad to keep it that way.
wdyt?

rd_ui/app/scripts/services/notifications.js
alexanderlz: We are.
the title is the first parameter passed to the constructor

arikfr: :facepalm:

Issue #303, opened 2014-10-13
"Show Total" duplicates values on multiple runs


arikfr:
To Reproduce:
* Take any graph
* Select show total.
* Select show total again.
* Notice how the numbers multiply (it seems like it's summing the total together with the series).
* Repeat until infinity and profit. ;-)

PR #357, opened 2015-01-24
Bug #303: 'Show Total' duplicates values on multiple runs

alexanderlz:
Prevents "totalling" more than once.

### DIFF
--- a/rd_ui/app/scripts/ng_highchart.js
+++ b/rd_ui/app/scripts/ng_highchart.js
@@ -143,11 +143,13 @@
                   }
                 });
 
-                if (!hasTotalsAlready) this.addSeries({
-                  data: _.values(data),
-                  type: 'line',
-                  name: 'Total'
-                }, false)
+                if (!hasTotalsAlready) {
+                    this.addSeries({
+                      data: _.values(data),
+                      type: 'line',
+                      name: 'Total'
+                    }, false)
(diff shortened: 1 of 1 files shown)


rd_ui/app/scripts/ng_highchart.js
```diff
@@ -126,6 +126,12 @@
             {
               text: 'Show Total',
               onclick: function () {
+                var hasTotalsAlready = _.some(this.series, function (s) {
+                    return s.name == 'Total';
+                })
+                if (hasTotalsAlready){
+                    return;
```
arikfr: Instead of just returning, keep the flow the same without recalculating:

```javascript
var totalsSeries = _.find(this.series, function(s) { return s.name == 'Total'; });
var data = {};

_.each(this.series, function (s) {
    s.setVisible(false, false);
    if (totalsSeries === undefined) {
        _.each(s.data, function (p) {
            data[p.x] = data[p.x] || {'x': p.x, 'y': 0};
            data[p.x].y = data[p.x].y + p.y;
        });
    }
});

if (totalsSeries === undefined) {
    this.addSeries({
        data: _.values(data),
        type: 'line',
        name: 'Total'
    }, false)
} else {
    totalsSeries.setVisible(true, false);
}

this.redraw();
```

PR #363, opened 2015-01-27
Feature: support for query parameters

arikfr:
This is initial work on support for query parameters. Currently it works by detecting mustache style parameters in the query text, and merging them with values from the URL query string.

No UI yet, you have to use the query string.

image

### DIFF
--- a/rd_ui/app/index.html
+++ b/rd_ui/app/index.html
@@ -126,6 +126,7 @@
 <script src="/scripts/ui-bootstrap-tpls-0.5.0.min.js"></script>
 <script src="/bower_components/bucky/bucky.js"></script>
 <script src="/bower_components/pace/pace.js"></script>
+<script src="/bower_components/mustache/mustache.js"></script>
 <!-- endbuild -->
 
 <!-- build:js({.tmp,app}) /scripts/scripts.js -->
--- a/rd_ui/app/scripts/controllers/dashboard.js
+++ b/rd_ui/app/scripts/controllers/dashboard.js
@@ -129,7 +129,8 @@
 
       $scope.query = $scope.widget.getQuery();
       var parameters = Query.collectParamsFromQueryString($location, $scope.query);
-      $scope.queryResult = $scope.query.getQueryResult(undefined, parameters);
+      var maxAge = $location.search()['maxAge'];
+      $scope.queryResult = $scope.query.getQueryResult(maxAge, parameters);
       $scope.nextUpdateTime = moment(new Date(($scope.query.updated_at + $scope.query.ttl + $scope.query.runtime + 300) * 1000)).fromNow();
 
       $scope.type = 'visualization';
--- a/rd_ui/app/scripts/controllers/dashboard.js
+++ b/rd_ui/app/scripts/controllers/dashboard.js
@@ -16,7 +16,7 @@
             var w = new Widget(widget);
 
             if (w.visualization) {
-              promises.push(w.getQuery().getQueryResultPromise());
+              promises.push(w.getQuery().getQueryResult().toPromise());
             }
 
             return w;
@@ -104,7 +104,7 @@
     };
   };
 
-  var WidgetCtrl = function($scope, Events, Query) {
+  var WidgetCtrl = function($scope, $location, Events, Query) {
     $scope.deleteWidget = function() {
       if (!confirm('Are you sure you want to remove "' + $scope.widget.getName() + '" from the dashboard?')) {
         return;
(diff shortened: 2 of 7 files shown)


arikfr:
@sidazhang @jeremi FYI, this is going to be released soon (you were asking about this in the past).

sidazhang:
@arikfr This is awesome. I am very excited to try this out.

And all these commercial vendors dares to charge $1000+ a month?

arikfr:
@sidazhang let me know if you need guidance on how to upgrade from your current re:dash version.

gabrielcrowdtilt:
Ooohhhh...I'm on v0.5.0+b681 and just demo-ed this. Can it be done with dashboards with charts too?

arikfr:
@gabrielcrowdtilt yes. works the same with dashboard/charts.

maflorvega:
Hi Arik!. I tried to add a table/chart using a query with this solution but the table/chart added doesn't show any results. I guess this is because  the dashboard url doesn't have the value parameters needed to executed successfully the query associated with the table/chart. How can I do it correctly?

arikfr:
@maflorvega as you figured, you need to pass the parameter(s). You do this the same as you do in the query view: pass it from the URL with the prefix `p_`. So for example, if the param name is "param" you would pass "p_param=...".

maflorvega:
First of all, thanks for the quick response!. Do you mean that I must to add  the prefixes to the dashboard url? For example for the dashboard "test" and the param name "id", the final url must be "http://xx.xx.x.xx/dashboard/test/source?p_id=23"?. I tried this solution and the result was 404 error. I feel like I am missing something....

maflorvega:
Thanks Arik!!!

matiascarazzo:
Arik , do you know if there is any way to set a default value to the parameters?
Maybe use some angular code in a text box? or even in the query itself?

Regards

arikfr:
@matiascarazzo not at the moment, but I do want to add this option. Improving the usability of the parameters feature is the next thing I'm going to work on. Next week I'm on vacation, but hopefully I get to work on it the week after when I'm back.

afterlastangel:
@arikfr what about using it with api/queries/.../results.json ? I'm trying to build dynamic api to use with my Dashboard UI.

arikfr:
@afterlastangel I did some work to prepare support for it, and maybe in the next release it will be possible.

afterlastangel:
@arikfr  what about widget/virtualization with parameters? I'm going create a chart with date range as parameter, so we need a way add the date range picker. If you didn't have a planing for it, then I could try with the widget then make a pull request.

arikfr:
Widgets/visualizations already support parameters, see: demo.redash.io/queries/146/source?p_action=create for example. What's missing is UI and that's planned for one of the next releases.

afterlastangel:
@arikfr  thank you. I just question about the Widgets UI for parameters, if you didn't implement it yet, I could try to create some for my own using.

Bitcoyne:
@arikfr I'm having trouble passing parameters into a column with a string or date data type. 

where url param: 
?p_account=uk

Error: 
`column "uk" does not exist in table`


When using single quotes in the url param, the url is then encoded: 
p_account=%27uk%27

causing a new error: 
`Error running query: syntax error at or near "uk" LINE 5: and account like &#39;uk&#39; ^`

Any workaround on this?

rohanpd:
@bitcoyne in your query you could wrap your parameter in quotes and percentages as follows:

`WHERE account = '%{{account}}%'`

ChiragKParmar:
@arikfr I created a user with 'read only' permission on the datasource, but that user was not able to get result by passing query string. is this a known issue?

arikfr:
@ChiragKParmar currently it's the intended behavior for two reasons:
1. With current implementation, we just run a new query generated of the requested one + parameters.
2. Even when we overcome this (by using the backend support for parameters) there is still the risk of SQL injection to run any query.

Until #2 properly addressed, we can't allow read only users use parameters. :(

arctica:
Hi @arikfr 

Any update on the GUI part for parameters? It's the last piece in the puzzle for us. Three things would be especially useful:

- A datepicker (actually two, for ranges)
- A dropdown select box which could be filled by another query
- Default values

sks-369:
Any update on the GUI part of parameters?

arikfr:
It's been implemented and released.

ryanmerolle:
@arikfr 

Is the query parameters dropdown select box for available for testing in any particular version yet?

Thanks for all the hard work!

Issue #168, opened 2014-04-09
Visualization: save as image


amirnissim:
currently people take a screenshot to export visualizations to docs and presentations.
"export/save as image" feature will make it easier for them.

bonus points: copy to clipboard

PR #368, opened 2015-01-31
Issue #168. Visualization: save as image.

alexanderlz:
Using https://github.com/gabelerner/canvg library to render html5 canvas from Highcharts svg,
then - downloading as png.


arikfr:
Use bower to add the canvg dependency -- `bower install --save gabelerner/canvg`.

PR #370, opened 2015-02-01
add ISO datetime to filename when saving chart as image

alexanderlz:


rd_ui/app/scripts/ng_highchart.js
```diff
@@ -158,15 +158,15 @@
               text: 'Save Image',
               onclick: function () {
                 var canvas = document.createElement('canvas'); 
-                window.canvg(canvas, this.getSVG())
+                window.canvg(canvas, this.getSVG());
                 var href = canvas.toDataURL('image/png');
                 var a = document.createElement('a');
                 a.href = href;
-                var filenameSuffix = '';
+                var filenameSuffix = new Date().toISOString();
```
arikfr: `toISOString` includes characters like :, which might be a problem on some file systems. Add `.replace(/:/g,'_').replace('Z', '')` to make it nicer/safer.

PR #374, opened 2015-02-18
Support unicode queries in search API

akariv:


akariv:
@arikfr now with modified tests.

arikfr:
Thanks, @akariv .

PR #377, opened 2015-02-20
'Download Dataset' fix - error in case of big numeric values

olgakogan:
removed redundant handling of large numbers when generating a csv file (causes ValueError: timestamp out of range)


arikfr:
Thanks, @olgakogan. It was a remnant from days when we were serializing datetime objects as Unix timestamp and indeed not needed anymore...

PR #390, opened 2015-03-16
Update bootstrap.sh

fedex1:
# modified by @fedex1 3/15/2015 seems to be the latest version at this point in time.


fedex1:
arikfr merged... deleting to clean up.

PR #393, opened 2015-03-19
Fix: allow Unicode and other special chars in column names

arikfr:
Stopped using Angular's $parse and just accessing the property directly.


arikfr:
@akariv FYI, once you upgrade, you can use Hebrew characters in column names.

PR #398, opened 2015-03-23
Fix: make the data_sources' name unique

lenguyenthedat:
Data_sources' name should be unique since it should be uniquely identified from the user interface, as well as when `sudo -u redash bin/run ./manage.py ds delete` is called.

Ref: https://github.com/EverythingMe/redash/issues/277

### DIFF
--- a/migrations/0007_make_ds_name_unique.py
+++ b/migrations/0007_make_ds_name_unique.py
@@ -1,0 +1,12 @@
+from redash.models import db
+
+if __name__ == '__main__':
+    db.connect_db()
+
+    with db.database.transaction():
+        # Make sure all data sources names are unique.
+        db.database.execute_sql("""UPDATE data_sources SET name = name || ' ' || id;""")
+        # Add unique constraint on data_sources.name.
+        db.database.execute_sql("ALTER TABLE data_sources ADD CONSTRAINT unique_name UNIQUE (name);")
+
+    db.close_db(None)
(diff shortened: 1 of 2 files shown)


arikfr:
This will only apply the change to new installations that use the `database create_tables` command. For existing ones, you need to add a migration. To save you the trouble, here it is (you might need to verify the syntax):

```python
# migrations/0007_make_ds_name_unique.py
from redash.models import db

if __name__ == '__main__':
    db.connect_db()

    with db.database.transaction():
        #  Make sure all data sources names are unique. Fix those that aren't.
        db.database.execute_sql("""UPDATE data_sources
SET name = new_names.name
FROM (
    SELECT id, name || ' ' || rank as name
    FROM (SELECT id, name, rank() OVER (PARTITION BY name ORDER BY created_at DESC) FROM data_sources) ds WHERE rank > 1
) AS new_names
WHERE data_sources.id = new_names.id;
""")
        # Add unique constraint on data_sources.name.
        db.database.execute_sql("ALTER TABLE data_sources ADD CONSTRAINT unique_name UNIQUE (name);")

    db.close_db(None)
```

lenguyenthedat:
Got it, however, that migration code doesn't guarantee `unique` for the new name though, for example, if I have: `test`, `test`, `test 2`

It will rename  the second one to `test 2`, which is already used.

arikfr:
Nice catch. Applying `name` + `id` only to the duplicates looks reasonable to me. 

I think that having unique constraint is a good move regardless. Also changing `DELETE` and other operations to use `id` is a good move too.

arikfr:
Merged manually in https://github.com/EverythingMe/redash/commit/e25b86b10daf57e833530d78a8c073a0aded0f51.

Thanks!

PR #402, opened 2015-04-06
New release process.

arikfr:
From now on, instead of creating a new release for every merge into master, it will update a single RC release with the name of the current version (for example: "v0.6 - RC").

Promoting an RC release to become final release, will be a manual process of editing the latest RC release.

Once production release is created the next build will create a new RC release.


gabrielcrowdtilt:
Congrats on your hard work, excited for this new release.

PR #404, opened 2015-04-13
Improvement: make Python datasource to use the RestrictedPython sandbox

erans:
Python datasource that uses RestrictedPython. Only modules listed in "allowedImportModules" (command separated) will be allowed to be imported and the code assume they are installed on the server running the actual code.

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -15,6 +15,7 @@ __all__ = [
     'TYPE_STRING',
     'TYPE_DATE',
     'TYPE_FLOAT',
+    'SUPPORTED_COLUMN_TYPES',
     'register',
     'get_query_runner',
     'import_query_runners'
@@ -28,6 +29,14 @@ TYPE_STRING = 'string'
 TYPE_DATETIME = 'datetime'
 TYPE_DATE = 'date'
 
+SUPPORTED_COLUMN_TYPES = {
+    TYPE_INTEGER : None,
+    TYPE_FLOAT : None,
+    TYPE_BOOLEAN : None,
+    TYPE_STRING : None,
+    TYPE_DATETIME : None,
+    TYPE_DATE : None
+}
 
 class BaseQueryRunner(object):
     def __init__(self, configuration):
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -29,14 +29,14 @@ TYPE_STRING = 'string'
 TYPE_DATETIME = 'datetime'
 TYPE_DATE = 'date'
 
-SUPPORTED_COLUMN_TYPES = {
-    TYPE_INTEGER : None,
-    TYPE_FLOAT : None,
-    TYPE_BOOLEAN : None,
-    TYPE_STRING : None,
-    TYPE_DATETIME : None,
-    TYPE_DATE : None
-}
+SUPPORTED_COLUMN_TYPES = set([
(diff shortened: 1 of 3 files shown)


redash/query_runner/__init__.py
```diff
@@ -28,6 +29,14 @@
 TYPE_DATETIME = 'datetime'
 TYPE_DATE = 'date'
 
+SUPPORTED_COLUMN_TYPES = {
```
arikfr: Why `dict` and not `set`?

redash/query_runner/python.py
```diff
@@ -1,9 +1,44 @@
...
+import importlib
+
+logger = logging.getLogger(__name__)
+
+try:
+    from RestrictedPython import compile_restricted
+    from RestrictedPython.Guards import safe_builtins
+
+    enabled = True
+except ImportError:
```
arikfr: No need for this - it's in `requirements.txt`.

redash/query_runner/python.py
```diff
@@ -19,10 +54,12 @@ def get_query_result(query_id):
 
     return json.loads(query.latest_query_data.data)
 
-
-def execute_query(data_source_name, query):
+def execute_query(data_source_name_or_id, query):
     try:
-        data_source = models.DataSource.get(models.DataSource.name==data_source_name)
+        if type(data_source_name) == int:
+            data_source = models.DataSource.get(models.DataSource.id==data_source_name_or_id)
```
arikfr: you can use `DataSource.get_by_id`.

redash/query_runner/python.py
```diff
@@ -45,24 +102,63 @@ def configuration_schema(cls):
...
+    @classmethod
     def annotate_query(cls):
         return False
 
     def __init__(self, configuration_json):
+        global ALLOWED_MODULES
+
         super(Python, self).__init__(configuration_json)
 
+        if "allowedImportModules" in self.configuration and self.configuration["allowedImportModules"]:
```
arikfr: `if self.configuration.get('allowedImportModules', None):`

redash/query_runner/__init__.py
erans: these are strings, AFAIK, using a dict means the "in" operator will use a hash search instead of matching the strings. In this case I think a set is the same as a list (maybe a bit better because it might be sorted).

redash/query_runner/python.py
erans: OK. Will remove it - unless you want to remove it from the requirements file instead?

redash/query_runner/python.py
erans: Will update the code

redash/query_runner/python.py
erans: Will update the code

redash/query_runner/python.py
arikfr: No, let's keep it in the requirements. Feels harmless enough.

erans: OK. Will remove the try..except

redash/query_runner/__init__.py
arikfr: Interesting point, but does it really matter? Considering we're talking about 6 elements and rarely used function? I think the set variant will be easier to understand.

erans: You are the BDFL :-) will change it to a set.

PR #408, opened 2015-04-19
Feature: additional metadata in query annotation (username, query id, queue name)

alexanderlz:
Added username/query id to "annotated_query".
added search_by_hash method to Query model.


arikfr:
I've added some more changes: f4be83b06f146d149091d591ab8a8be45dad624f (use query id from UI instead of lookup, annotate scheduled queries, and add queue name) and merged.

PR #410, opened 2015-04-22
Feature: sort by Y values charts that have a single value per series

stanhu:
We have a lot of bar charts that should be sorted by y values. This pull request adds a "sortY" option.

screen shot 2015-04-21 at 10 24 02 pm

### DIFF
--- a/rd_ui/app/scripts/ng_highchart.js
+++ b/rd_ui/app/scripts/ng_highchart.js
@@ -325,34 +325,21 @@
                 series = seriesCopy;
               }
 
-              if (chartOptions['sortY'] === true || chartOptions['sortY'] === undefined) {
-                var sortable = true;
+              // If this is a chart that has just one row for multiple columns, sort
+              // by the Y values. For example:
+              //
+              // A  | B  | C
+              // 20 | 30 | 15
+              //
+              // Will be sorted:
+              // C  | A  | B
+              // 15 | 20 | 30
+              var sortable = _.every(series, function(s) { return s.data.length == 1 });
 
-                for (var i = 0; i < series.length; i++) {
-                  if (series[i].data.length != 1) {
-                    sortable = false;
-                    break;
-                  }
-                }
-
-                if (sortable) {
-                  var sortFunction = function(s1, s2) {
-                    if (s1.data.length == 1 && s2.data.length == 1) {
-                      var a = s1.data[0].y;
-                      var b = s2.data[0].y;
-
-                      if (a < b) {
-                        return -1;
-                      } else if (a == b) {
-                        return 0;
-                      } else {
-                        return 1;
-                      }
-                    }
-                  }
-
-                  series.sort(sortFunction);
-                }
+              if (sortable) {
+                series = _.sortBy(series, function (s) {
+                  return s.data[0].y
+                });
               }
 
               if (!('xAxis' in chartOptions && 'type' in chartOptions['xAxis'])) {
--- a/rd_ui/app/scripts/ng_highchart.js
+++ b/rd_ui/app/scripts/ng_highchart.js
@@ -325,6 +325,36 @@
                 series = seriesCopy;
               }
 
+              if (chartOptions['sortY'] === true || chartOptions['sortY'] === undefined) {
+                var sortable = true;
+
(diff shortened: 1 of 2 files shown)


rd_ui/app/scripts/ng_highchart.js
```diff
@@ -325,6 +325,36 @@
                 series = seriesCopy;
               }
 
+              if (chartOptions['sortY'] === true || chartOptions['sortY'] === undefined) {
```
arikfr: `chartOptions['sortY'] === undefined` means that "old" queries without this setting will default to enabling this, right? I prefer that we don't do this, to avoid confusion and unexpected results for existing charts/users.

rd_ui/app/scripts/ng_highchart.js
```diff
@@ -325,6 +325,36 @@
...
+                var sortable = true;
+
+                for (var i = 0; i < series.length; i++) {
+                  if (series[i].data.length != 1) {
```
arikfr: Isn't `series[i].data.length` the array of points in the chart? In that case, why only sort if length is 1? Am I missing something?

arikfr: Btw, you can use [Underscore.js](http://underscorejs.org/) here:

```javascript
var sortable = _.find(series, function(s) { return s.data.length != 1 });
```

rd_ui/app/scripts/ng_highchart.js
```diff
@@ -325,6 +325,36 @@
...
+                        return -1;
+                      } else if (a == b) {
+                        return 0;
+                      } else {
+                        return 1;
+                      }
+                    }
+                  }
+
+                  series.sort(sortFunction);
```
arikfr: This can benefit from Underscore.js too:

```javascript
_.each(series, function(s) {
  s.data = _.sortBy(s.data, 'y');
});
```

Although the logic might need to be more complex in case of multiple series (is this what the check above supposed to prevent?).

rd_ui/app/scripts/ng_highchart.js
stanhu: Right, that makes sense.

rd_ui/app/scripts/ng_highchart.js
stanhu: The goal of the sort-by-y functionality is to reorder different series in a bar chart. This only makes sense if there is one data point per series. Does that make sense, or are there cases which I haven't considered?

arikfr:
Btw, until this merged, if you uncheck the "Sort X values" option of the chart, it will preserve the order the results returned from the database, so you can already sort by y value if you change the query.

See this as an example:
http://demo.redash.io/queries/232/source#add

Just play with the "Sort X values" option.

rd_ui/app/scripts/ng_highchart.js
arikfr: So the data looks like this? One row, with multiple columns?

X  | Y1 | Y2
------------- | ------------- | -------------
Whatever  | 1 | 2

stanhu: Yes. Is there a better way of doing this?

stanhu:
Right, the sort by X functionality doesn't do what we need here. As mentioned above, we have a table that looks like:

Test | Development |  Production
---------- | --------- | ---------- | 
 100 | 200 | 300

rd_ui/app/scripts/ng_highchart.js
arikfr: Probably not...

But, now that I better understand what you did here, I think that we can keep this logic enabled for everyone and we don't need the toggle. 

If you think we should keep the toggle, then we need to find different wording for it as "Sort Y values" is confusing (or implement proper sorting of Y values for all scenarios...).

Also the sorting code becomes:
```javascript
series = _.sortBy(series, function(s) { return s.data[0].y });
```

And detection:
```javascript
var sortable = _.every(series, function(s) { returns s.data.length == 1 });
```

stanhu: Great suggestions. I forgot about Underscore; it's been a while since I've used it. I've updated and retested.

Issue #418, opened 2015-05-07
chart name is not shown on the dashboard


runa:
When you add a chart to a dashboard the *chart name* is not shown.

This is specially problematic when you add 2 different charts from the same query: you don't have the name from each and when both are 'similar' (ie, 2 pie charts) it's pretty confusing.

arikfr:
![image](https://cloud.githubusercontent.com/assets/71468/7523311/9d023aec-f503-11e4-9c22-459787710129.png)

Something like makes sense? ("User's Actions" is the query name and "Very Interesting Pie" is the chart name)

runa:
yep!

arikfr:
great. follow #421 for status.

PR #421, opened 2015-05-07
Feature: show visualization name next to query name (#418)

arikfr:


arikfr:
@erans iirc, you asked for this too.

PR #432, opened 2015-05-15
Allow undefined max_age parameter in query_results endpoint

stanhu:
An Error 500 would be returned by the endpoint if you attempt to
pass a query parameter to the dashboard since `maxAge` is undefined in the JavaScript code. This line of code was being executed:

https://github.com/EverythingMe/redash/blob/74a5121be232a0a6002928c38c7290298370c7e0/rd_ui/app/scripts/services/resources.js#L454

### DIFF
--- a/redash/controllers.py
+++ b/redash/controllers.py
@@ -450,7 +450,7 @@ api.add_resource(VisualizationAPI, '/api/visualizations/<visualization_id>', end
 class QueryResultListAPI(BaseResource):
     @require_permission('execute_query')
     def post(self):
-        params = request.json
+        params = request.get_json(force=True)
 
         if settings.FEATURE_TABLES_PERMISSIONS:
             metadata = utils.SQLMetaData(params['query'])
@@ -476,7 +476,7 @@ class QueryResultListAPI(BaseResource):
             activity=params['query']
         ).save()
 
-        max_age = int(params['max_age'])
+        max_age = int(params.get('max_age', -1))
 
         if max_age == 0:
             query_result = None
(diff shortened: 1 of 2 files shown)


redash/controllers.py
```diff
@@ -450,7 +450,7 @@ def delete(self, visualization_id):
 class QueryResultListAPI(BaseResource):
     @require_permission('execute_query')
     def post(self):
-        params = request.json
```
arikfr: Why can't we keep using `request.json`?

stanhu: I tried to keep it, but the test case seems to serialize it into a format that can't be parsed by `request.json`:

https://github.com/EverythingMe/redash/blob/master/tests/test_controllers.py#L33

Is it missing a header of some sort? I didn't have time to research. I just changed it to be consistent with the rest of the code in the controllers.

arikfr: Got it. I think there is a way to tell Flask to return json regardless of
the content type header. But keep it as is, I'll fix the test helper and
update it later.

stanhu: Looks like I'm right:

http://stackoverflow.com/questions/20001229/how-to-get-posted-json-in-flask

I can fix this now if you'd like.

stanhu: Updated! Using `get_json(force=True)` is not much shorter than `json.loads(request.data)`. :)

arikfr: :-) true that.

I wonder if once I get to fix the test helper, we should keep the `get_json(force=True)` (it's used in other places too, I think) or not ...

arikfr:
Thanks!

PR #433, opened 2015-05-15
Make it obvious that the query link is clickable.

stanhu:
Underline when hovering over the link and add glyphicon

Before:

image

After:

image


stanhu:
Hmm, I think I need to take into account the `view_query` and `edit_query` permissions.

arikfr:
The anchor feels a bit too "bold", how about we wrap it with `<small>`? The result will be:

image

stanhu:
Great idea. Updated.

arikfr:
Merged. Thanks!

PR #442, opened 2015-06-03
Fix: when the server has non UTC timezone, timestamps were wrong

arikfr:


redash/tasks.py
```diff
@@ -272,7 +271,7 @@ def execute_query(self, query, data_source_id, metadata):
     redis_connection.delete(QueryTask._job_lock_id(query_hash, data_source.id))
 
     if not error:
-        query_result = models.QueryResult.store_result(data_source.id, query_hash, query, data, run_time, datetime.datetime.utcnow())
+        query_result = models.QueryResult.store_result(data_source.id, query_hash, query, data, run_time, utils.utcnow())
```
arikfr: @runa this is the actual fix for the timezone issue. Until I get to properly test this and merge it in, if you want you can patch your install by editing `/opt/redash/current/tasks.py` and changing on this line `datetime.datetime.utcnow()` to: `datetime.datetime.now(pytz.utc)` and adding `import pytz` at the top of the file.

PR #454, opened 2015-06-15
Fix SSL Setup document which does not work correctly

hakobera:
Current sample nginx config on http://redash.io/deployment/ssl.html does not work correctly.

- When I access to `https://redash_server/`, it redirect to `http://redash_server/login`. Because `X-Forwarded-Proto` header is not set correctly.
- I think it's easy to understand that dividing `server` settings for `http` and `https` and avoid to use `if` statement (see: http://wiki.nginx.org/IfIsEvil)


arikfr:
Thanks!

PR #455, opened 2015-06-16
added Mongo JSON serializer to correctly serialize ObjectId + datetime.datetime serialization

erans:


redash/utils.py
```diff
@@ -95,9 +95,9 @@ def default(self, o):
         if isinstance(o, decimal.Decimal):
             return float(o)
 
-        if isinstance(o, datetime.date):
+        if isinstance(o, datetime.date) or isinstance(o, datetime.datetime):
```
arikfr: Why is this needed?

```python
>>> isinstance(datetime.datetime.now(), datetime.date)
True
```

erans: Mongo (and other runners) sometimes return a datetime.datetime object so I thought instead of having it in the mongodb json serializer I'll add it here.

datetime.date != datetime.datetime :-)

arikfr: From what I understand `datetime.datetime` inherits `datetime.date`, and hence why `isinstance(datetime.datetime.now(), datetime.date) == True`. You see different behavior?

erans: Yet without this line the deserializer yells it cant deserialize
datetime.datetime.

That's why I added it.

erans: OK. My bad. It accidentally called the original class. I'll remove it.

PR #458, opened 2015-06-24
Initial and very early support for ElasticSearch query runner

erans:
Initial very early release of an ElasticSearch query runner. It only support Lucene style queries (single line, similar to what Kibana uses but without aggregations).


arikfr:
Thanks :)

PR #459, opened 2015-06-25
Fixed an error in case a query doesn't have last_modified_date

olgakogan:
This used to cause the queries api call to return 500


arikfr:
Thanks.

PR #460, opened 2015-06-26
Added SAML support

mark9white:
I've added SAML support as an alternative to Google Apps authentication, eg for OneLogin or Okta.

It adds the REDASH_SAML_METADATA_URL config value which needs to point to the SAML provider metadata url, eg https://app.onelogin.com/saml/metadata/<ID>

And an optional REDASH_SAML_CALLBACK_SERVER_NAME which contains the server name of the redash server for the callbacks from the SAML provider (eg demo.redash.io)

On the SAML provider side, example configuration for OneLogin is:
SAML Consumer URL: http://demo.redash.io/saml/login
SAML Audience: http://demo.redash.io/saml/callback
SAML Recipient: http://demo.redash.io/saml/callback

with parameters FirstName and LastName, both configured to be included in the SAML assertion.


arikfr:

PR #466, opened 2015-07-01
Feature: Map visualization (using Leaflet)

### LINKED ISSUE #39: Map display for query results
Please add a map display for query results containing geo data.
Geodata types can be - 
* lat/long
* Address

BrunoSalerno:
@arikfr I open the PR.
This is a map visualization that uses leaflet. Right now only supports point features. You can select the coordinates columns, and draw the points using markers or colored circles. 
Each feature has a popup that shows all the 'columns' of the row.

btw: I didn't know what to do with the size of the main div inside the template `map.html`

(Closes #39)

### DIFF
--- a/rd_ui/app/scripts/visualizations/map.js
+++ b/rd_ui/app/scripts/visualizations/map.js
@@ -10,7 +10,7 @@
       '</map-renderer>';
 
     var editTemplate = '<map-editor></map-editor>';
-    var defaultOptions = {};
+    var defaultOptions = { 'height': 500 };
 
     VisualizationProvider.registerVisualization({
       type: 'MAP',
@@ -27,6 +27,18 @@
       restrict: 'E',
       templateUrl: '/views/visualizations/map.html',
       link: function($scope, elm, attrs) {
+
+        var setBounds = function(){
+          var b = $scope.visualization.options.bounds;
+
+          if(b){
+            $scope.map.fitBounds([[b._southWest.lat, b._southWest.lng],[b._northEast.lat, b._northEast.lng]]);
+          } else if ($scope.features.length > 0){
+            var group= new L.featureGroup($scope.features);
+            $scope.map.fitBounds(group.getBounds());
+          }
+        };
+
         $scope.$watch('[queryResult && queryResult.getData(), visualization.options.draw,visualization.options.latColName,'+
             'visualization.options.lonColName,visualization.options.classify,visualization.options.classify]',
           function() {
@@ -99,16 +111,7 @@
               }).join('');
             };
 
-            var setBounds = function(){
-              var b = $scope.visualization.options.bounds;
 
-              if(b){
-                $scope.map.fitBounds([[b._southWest.lat, b._southWest.lng],[b._northEast.lat, b._northEast.lng]]);
-              } else if ($scope.features.length > 0){
-                var group= new L.featureGroup($scope.features);
-                $scope.map.fitBounds(group.getBounds());
-              }
-            };
 
             function getBounds(e) {
               $scope.visualization.options.bounds = $scope.map.getBounds();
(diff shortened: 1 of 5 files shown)


rd_ui/app/views/visualizations/map.html
```diff
@@ -0,0 +1 @@
+<div style='margin:20px;height:500px;width:500px;'></div>
```
arikfr: Can it inherit width from the containing element? So the user can control the width like he can for the other widgets.

As for height, maybe change this to be a visualization property (with some default value)?

rd_ui/app/scripts/visualizations/map.js
```diff
@@ -0,0 +1,207 @@
...
+                $scope.visualization.options.bounds = $scope.map.getBounds();
+              });
+
+              var group = new L.featureGroup($scope.features);
+              var b = $scope.visualization.options.bounds;
+              $scope.map.fitBounds(b ?
+               [[b._southWest.lat, b._southWest.lng],[b._northEast.lat, b._northEast.lng]]: group.getBounds());
+
+              // We redraw the map if it was loaded in a hidden tab
+              if ($('a[href="#'+$scope.visualization.id+'"]').length > 0) {
```
arikfr: :| I'll add support for signaling the visualization if it's visible, to avoid this.
(no need to do anything on your end)

arikfr:
This is really great. Let's get the width/height sorted, and merge :)

BrunoSalerno:
@arikfr great! I'm working on it

BrunoSalerno:
@arikfr done. I couldn't use `VisualizationProvider.defaultOptions`. Is it alright this way or there is a specific way to use `defaultOptions`?

arikfr:
@BrunoSalerno what do you mean you couldn't use them?

rd_ui/app/scripts/visualizations/map.js
```diff
@@ -0,0 +1,223 @@
...
+                  $scope.visualization.options.classification.push({value: queryData[row][classify], color: null});
+                }
+              }
+
+              $.each($scope.visualization.options.classification, function (i, c) {
+                c.color = color(parseInt((i / $scope.visualization.options.classification.length) * 100));
+              });
+
+              if (!$scope.map) {
+                $(elm[0].children[0]).replaceWith($("<div style='margin:1%;height:"+$scope.visualization.options.height+"px;width:95%;'></div>"));
```
arikfr: You don't need jQuery here, you can just change the template to:
```html
<div style='margin:1%;width:95%;height:{{visualization.options.height}}px;'></div>
```

BrunoSalerno:
@arikfr I set `defaultOptions` in `VisualizationProvider`, so I expect them to be set in `scop.visualization.options`, but they are not. Am I missing something?

rd_ui/app/scripts/visualizations/map.js
arikfr: I've played around with your branch a little, and noticed that when height is being changed this way (dynamically), it's better to have the map in an inner div. So the template becomes:

```html
<div style='margin:1%;width:95%;height:{{visualization.options.height}}px;'>
    <div style="width:100%; height:100%;"></div>
</div>
```

And line 138 changes to: `$scope.map = L.map(elm[0].children[0].children[0]);`.

And you need to add watch for height:
```javascript
              $scope.$watch('visualization.options.height', function() {

                  $scope.map.invalidateSize(false);
                  setBounds();

              });
```

rd_ui/app/views/visualizations/map_editor.html
```diff
@@ -0,0 +1,55 @@
+<div class="form-horizontal">
+    <div class="form-group">
+        <label class="col-lg-2">Map height (px)</label>
+        <div class="col-sm-4">
+            <input class="form-control" type="text" ng-model = "visualization.options.height" />
```
arikfr: Better to have here `type="number"` (then it's being saved as number, and you can control the value with arrows).

arikfr:
@BrunoSalerno re. defaultOptions, looks like it works only for the default visualization type. I'll fix it an update your code to use it.

BrunoSalerno:
@arikfr nice. I'm pushing the changes and adding a FIXME in the editor directive (in the height part)

arikfr:
:+1:

PR #472, opened 2015-07-03
map visualization: options fixed

BrunoSalerno:
@arikfr hi ;). There were some problems with the options (for example, classify categories weren't visible before saving). This PR fixes that.

### DIFF
--- a/rd_ui/app/scripts/visualizations/map.js
+++ b/rd_ui/app/scripts/visualizations/map.js
@@ -12,7 +12,9 @@
     var editTemplate = '<map-editor></map-editor>';
     var defaultOptions = {
       'height': 500,
-      'draw': 'Marker'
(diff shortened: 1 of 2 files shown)


rd_ui/app/scripts/visualizations/map.js
```diff
@@ -12,7 +12,9 @@
     var editTemplate = '<map-editor></map-editor>';
     var defaultOptions = {
       'height': 500,
-      'draw': 'Marker'
+      'draw': 'Marker',
+      'draw_options' : ['Marker','Color'],
```
arikfr: It's worth moving `draw_options` to the `$scope` as well, as there is no point in saving it along with every map visualization.

BrunoSalerno:
@arikfr I agree, done ;)

PR #473, opened 2015-07-04
Bump up version of site to 0.6.2

hakobera:
Current version is `0.6.2` but official site says it is `0.5.0`.

In addition, I think it is necessary include in page title,because user always want latest version document.
If you want remain `version` string, it should update at every release.


arikfr:
You're 100% right. Thanks for pointing this out and fixing it.

PR #475, opened 2015-07-05
Feature: Support InfluxDB v0.9+

hakobera:
This pull request experimental support for [InfluxDB](https://influxdb.com/) but only v0.9+. Because InfluxDB has no backward compatibility for older than 0.8.

## Config

- Type: influxdb
- Options:
  - Url (url) (mandatory)

## Add Data Source command

```
bin/run ./manage.py ds new -n example -t influxdb -o '{"url":"influxdb://root:root@localhost:8086/example"}'
```

### DIFF
--- a/redash/query_runner/influx_db.py
+++ b/redash/query_runner/influx_db.py
@@ -1,0 +1,83 @@
+import json
+import logging
+
+from redash.utils import JSONEncoder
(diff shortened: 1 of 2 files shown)


arikfr:
This is great and much appreciated! I wanted to add InfluxDB support for a long time...

PR #479, opened 2015-07-12
Feature: improved error reporting and new log information support for the Python query runner

erans:
- Improved error handling to show the real error (including indentation and syntax errors - it should show the right row number)
- Added support for "print" statements which becomes log messages show in a single query window (where you edit). It's on by default. To remove it simply called 'disable_print_log()'

### DIFF
--- a/rd_ui/app/views/query.html
+++ b/rd_ui/app/views/query.html
@@ -192,6 +192,16 @@
             </div>
             <div class="alert alert-danger" ng-show="queryResult.getError()">Error running query: <strong>{{queryResult.getError()}}</strong></div>
 
+            <div class="row log-container" ng-show="showLog">
+                <span ng-show="showLog">Log Information:</span>
+                <table>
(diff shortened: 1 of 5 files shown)


rd_ui/app/views/query.html
```diff
@@ -192,6 +192,16 @@ <h4 class="modal-title">Query Archive</h4>
             </div>
             <div class="alert alert-danger" ng-show="queryResult.getError()">Error running query: <strong>{{queryResult.getError()}}</strong></div>
 
+            <div class="row log-container" ng-show="showLog">
```
arikfr: Does it take space (due to the `margin-bottom` style) when ng-show=false?

erans: I checked and it basically hides without taking any height, so the margin doesn't affect anything.

arikfr: :+1:

PR #481, opened 2015-07-12
Query API CORS Support

erans:
Controlled from an environment variable named "REDASH_QUERIES_RESULT_CORS". Can contain a comma separated list of domains.

### DIFF
--- a/redash/controllers.py
+++ b/redash/controllers.py
@@ -14,11 +14,11 @@ import logging
 from flask import render_template, send_from_directory, make_response, request, jsonify, redirect, \
     session, url_for, current_app
 from flask.ext.restful import Resource, abort
-from flask_login import current_user, login_user, logout_user
+from flask_login import current_user, login_user, logout_user, login_required
 import sqlparse
 
-from redash import redis_connection, statsd_client, models, settings, utils, __version__
-from redash.wsgi import app, auth, api
+from redash import statsd_client, models, settings, utils
+from redash.wsgi import app, api
 from redash.tasks import QueryTask, record_event
 from redash.cache import headers as cache_headers
 from redash.permissions import require_permission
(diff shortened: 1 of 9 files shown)


arikfr:
It includes your Python runner changes even though I merged it already. Can you rebase, just to make sure we're not overriding anything?

redash/controllers.py
```diff
@@ -488,6 +488,22 @@ def csv_response(query_result):
         headers.update(cache_headers)
         return make_response(s.getvalue(), 200, headers)
 
+    @staticmethod
+    def add_access_control_allow_origin_header(headers={}):
```
arikfr: I'm not sure about Flask's life cycle, but if controllers are reused, the default headers will be reused, aren't they?

redash/controllers.py
```diff
@@ -488,6 +488,22 @@ def csv_response(query_result):
...
+            if origin in settings.QUERIES_RESULT_CORS:
+                headers['Access-Control-Allow-Origin'] = origin
+                headers['Access-Control-Allow-Credentials'] = 'true'
+                if request.method == 'OPTIONS':
+                    headers['Access-Control-Request-Method'] = 'GET, POST, PUT'
+                    headers['Access-Control-Allow-Headers'] = 'Content-Type'
+
+    @require_permission('view_query')
+    def options(self, query_id=None, query_result_id=None, filetype='json'):
+        self.add_access_control_allow_origin_header()
```
arikfr: If you don't pass anything, it just sets the local headers variables. How is it working then?

redash/controllers.py
```diff
@@ -517,9 +533,15 @@ def get(self, query_id=None, query_result_id=None, filetype='json'):
 
                 record_event.delay(event)
 
+            headers = {}
+
+            if settings.QUERIES_RESULT_CORS and len(settings.QUERIES_RESULT_CORS) > 0:
```
arikfr: `settings.QUERIES_RESULTS_CORS` won't be truthy if len not > 0...

redash/controllers.py
erans: they should be. I think the generate make_response adds the headers you send on top of the default headers.

redash/controllers.py
erans: true. It was from a previous version. I'll remove the default.

redash/controllers.py
erans: yeah, I think I can remove that and keep the len.

PR #483, opened 2015-07-13
Feature: Updated more fine grained support for CORS in QueryResultAPI

erans:


redash/settings.py
```diff
@@ -81,7 +81,13 @@ def parse_boolean(str):
 CLIENT_SIDE_METRICS = parse_boolean(os.environ.get("REDASH_CLIENT_SIDE_METRICS", "false"))
 ANALYTICS = os.environ.get("REDASH_ANALYTICS", "")
 
-QUERIES_RESULT_CORS = set_from_string(os.environ.get("REDASH_QUERIES_RESULT_CORS", ""))
+# CORS settings for the Query Result API (and possbily future external APIs).
+# In most cases all you need to do is set REDASH_CORS_ACCESS_CONTROL_ALLOW_ORIGIN
+# to the calling domain (or domains in a comma separated list).
+ACCESS_CONTROL_ALLOW_ORIGIN = set_from_string(os.environ.get("REDASH_CORS_ACCESS_CONTROL_ALLOW_ORIGIN", ""))
+ACCESS_CONTROL_ALLOW_CREDENTIALS = parse_boolean(os.environ.get("REDASH_CORS_ACCESS_CONTROL_ALLOW_CREDENTIALS", "false"))
```
arikfr: why `parse_boolean` here and convert to string above? just keep it string.

erans: Because I need to verify its a valid value. I prefer to do it while we read the configuration and not when I'm setting it

PR #499, opened 2015-07-19
Feature: Support Impala as DataSource

alexanderlz:
initial support for Cloudera Impala.
http://www.cloudera.com/content/cloudera/en/products-and-services/cdh/impala.html

CDH version 5.2 and above.

python connector via https://github.com/cloudera/impyla


redash/query_runner/impala_ds.py
```diff
@@ -0,0 +1,137 @@
...
+import sys
+
+from redash.query_runner import *
+from redash.utils import JSONEncoder
+
+logger = logging.getLogger(__name__)
+
+try:
+    from impala.dbapi import connect
+    from dateutil import parser
```
arikfr: no need to have dateutil in a try/except block.

redash/query_runner/impala_ds.py
```diff
@@ -0,0 +1,137 @@
...
+                "database": {
+                    "type": "string"
+                },
+                "host": {
+                    "type": "string"
+                },
+                "port": {
+                    "type": "number"
+                },
+                "host": {
```
arikfr: host and ldap_user appear twice.

redash/query_runner/impala_ds.py
```diff
@@ -0,0 +1,137 @@
...
+    def run_query(self, query):
+        #connection = connect(host=self.configuration['host'],
+        #                     port=(self.configuration['port'] if 'port' in self.configuration else None),
+        #                     protocol=(self.configuration['protocol'] if 'protocol' in self.configuration else None),
+        #                     timeout=(self.configuration['timeout'] if 'timeout' in self.configuration else None),
+        #                     database=(self.configuration['database'] if 'database' in self.configuration else None),
+        #                     use_ldap=(self.configuration['use_ldap'] if 'use_ldap' in self.configuration else None),
+        #                     ldap_user=(self.configuration['ldap_user'] if 'ldap_user' in self.configuration else None),
+        #                     ldap_password=(self.configuration['ldap_password'] if 'ldap_password' in self.configuration else None),)
+        connection = connect(**self.configuration)
```
arikfr: connect might fail too, so better have it in the try block too.

redash/query_runner/impala_ds.py
```diff
@@ -0,0 +1,137 @@
...
+
+            data = {'columns': columns, 'rows': rows}
+            json_data = json.dumps(data, cls=JSONEncoder)
+            error = None
+            cursor.close()
+        except KeyboardInterrupt:
+            connection.cancel()
+            error = "Query cancelled by user."
+            json_data = None
+        except Exception as e:
```
arikfr: What kind of exception Impala will throw in case of syntax error? Might want to return it as error instead of rethrowing.

requirements.txt
```diff
@@ -30,3 +30,5 @@ RestrictedPython==3.6.0
 wtf-peewee==0.2.3
 pysaml2==2.4.0
 pycrypto==2.6.1
+impyla==0.10.0
+gspread==0.2.5
```
arikfr: please don't add them to the requirements.

redash/query_runner/impala_ds.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __init__(self, configuration_json):
+        super(Impala, self).__init__(configuration_json)
+
+    def run_q(self, query):
+        results, error = self.run_query(query)
+
+        if error is not None:
+            raise Exception("Failed getting schema.")
+        return json.loads(results)['rows']
```
arikfr: Just add to run_query another parameter (json_result) with default of True, and when it's False don't do json.dumps and return data as is. I wanted to implement something like this for all connectors. Maybe even do the json.dumps in a layer above.

If you still prefer to keep this method, give it a more describing name and make it "protected".

redash/query_runner/impala_ds.py
```diff
@@ -0,0 +1,165 @@
...
+            schemas_query = """
+                show schemas;
+            """
+
+            tables_query = """
+                show tables in %s;
+            """
+
+            columns_query = """
+                show column stats %s;
```
arikfr: Not that it makes much difference, but why not `columns_query = "show column stats %s;"`? (applies to all three)

alexanderlz: show column stats - works on specific table
http://www.cloudera.com/content/cloudera/en/documentation/cloudera-impala/latest/topics/impala_show.html#show_tables_unique_1

arikfr: I meant why the """ instead of regular string?

redash/query_runner/impala_ds.py
alexanderlz: for now I'll keep it the same (protecting + different name)
will implement your suggestion in next pull req :-)

PR #501, opened 2015-07-20
All pg queries with no results

johnkearney:
Queries like creating a table, dropping a table, unloading data, etc all return an error message: "Error running query: 'NoneType' object is not iterable". I think this would be clearer as: "Error running query: Query completed but it returned no data."


arikfr:
Thanks, @johnkearney! 

And thanks for cleaning up the env files.

PR #503, opened 2015-07-22
Fix: if you change the result object, python runner wouldn't return any results

arikfr:


redash/query_runner/python.py
```diff
@@ -187,9 +184,9 @@ def run_query(self, query):
             #       One option is to use ETA with Celery + timeouts on workers
             #       And replacement of worker process every X requests handled.
 
-            exec(code) in restricted_globals, script_locals
+            exec(code) in restricted_globals, self._script_locals
 
-            json_data = json.dumps(self._result)
```
arikfr: @erans FYI.

The problem was here: if you set a different object to `result` in the script, then `self._result` will no longer reference to it.

erans: I had to do it due to scoping issue and the support for the print statement

erans: Perhaps I can add a statement to override it.

arikfr: Actually it's better to accumulate the log statements in another variable, and then put them in result["log"], or they might overriden -- I'll do it in this branch.

erans: OK but in that case it may override anything that was set in result["log"]

arikfr: I could merge if it exists already, but doesn't feel like it worth the hassle. People shouldn't abuse this :)

See fix here: https://github.com/EverythingMe/redash/commit/07b88d0b53189cf777c252029d361ad00c05fe9e
(removed the need for WeakRef too)

erans: OK. Looks good.

PR #510, opened 2015-07-27
for aws based and password less logins to boxes

rghose:


KensoDev:
@rghose you can also add `ssh-add ~/.pem/your-key-name.pem` and then run the command.
That would be another good documentation addition IMHO

arikfr:
Thanks!

Issue #450, opened 2015-06-10
docker support


saritasa:
It would be nice to have Dockerfile in repo, so we can build a container automatically

arikfr:
It is in the backlog. Can't commit to ETA, but I will comment here when it's available.

KensoDev:
@arikfr @saritasa I plan on working on it next week.

tjwudi:
Is anybody still working on this? I am currently investigating to add a Dockerfile to redash too. I would like to follow up if anybody have made a working version ;)

KensoDev:
@tjwudi I was planning on this but got drawn to other project internally.

I don't have a working version yet, but it's definitely something I am planning on doing since we are dockerizing a lot of the infrastructure right now and redash is becoming a major part of our admin interfaces.

tjwudi:
@KensoDev thanks for your info. Then I will start to work on it. Once I got it working I will send in a PR.

arikfr:
I'm adding some thoughts on how to add Docker support for re:dash, in case someone would like to pick up the glove :)

Basically the whole setup process is "documented" in the [bootstrap script](https://github.com/EverythingMe/redash/blob/master/setup/bootstrap.sh), but the Docker setup should only take inspiration from it rather reuse it entirely.

Basically what I would do is:
1. Use PostgreSQL official container.
2. Use Redis official container.
3. Use Nginx official container.
4. Create "base" re:dash container, which will be used for the workers and web server.
5. Move the [database bootstrap](https://github.com/EverythingMe/redash/blob/master/setup/bootstrap.sh#L143-L151) part into a script in the `bin` folder.

Then tie everything together with Docker compose and some setup script to create the database once everything is up. Basic re:dash setup should have the following containers: PostgreSQL, Redis, Nginx, re:dash web server, re:dash "master" celery worker (with --beat option) & one additional re:dash celery worker.

I'm quite new to Docker, so if someone has other suggestions, I'll be happy to hear.

tjwudi:
@arikfr Oh you use nginx in a separate container? Is there any reason not use nginx in the same container with redash?

KensoDev:
@tjwudi That's a Docker best practice, but you need to understand what is *your* purpose of doing this.
If the purpose it to make redash easily distributed across machines, uses @arikfr suggestion.

This way, you can compose the workload easier and not worry about processes running inside Docker, each Docker will be a process that you can monitor using upstart or anything like that.

arikfr:
[...]

PR #512, opened 2015-07-27
Move datasource requirements from bootstrap to own requirements file

johnkearney:
It would be nicer for us if the requirements for new datasources were separated from bootstrapping of an instance. Would you mind having a change like this?


arikfr:
Totally makes sense and having the versions defined is a good move.

I'm curious now -- why is it making a difference for you?

johnkearney:
Thanks! (There are a few more requests like this still to come.)
In short this is useful to us (at Swrve) because we have our own deployment scripts that replicate your bootstrap.sh and related files but configure the redash for our needs - so being able to separate out the data from the code makes our deployment files more maintainable.

arikfr:
Nice. I guess that once #450 (Docker support) is implemented, it will make your life easier...

PR #515, opened 2015-07-30
Fix: change wording on login screen "username or email" -> "email"

moyomot:
Username authentication are not allowed.
So, I modified authentication display.


arikfr:
The reason it's "username or email" is due to the fact that sometimes the email column contains a username, like the default "admin" user. But it's not the first time I get this feedback, so I'm merging.

Thanks!

moyomot:
:+1:

PR #527, opened 2015-08-09
Fix: support for filters in BigQuery.

arikfr:
BigQuery doesn't support :: in column names, so using __ to split instead of ::.


arikfr:
@rm420 FYI. Will be part of the next release (v0.7.1).

rm420:
@arikfr Thanks a bunch!

PR #529, opened 2015-08-10
Enables multi-filtering for big-query.

rm420:
This fixes multi-filter for Google BigQuery. Before the fix, BigQuery had an issue with the hyphen in "multi-filter". now both "multiFilter" and "multi-filter" can be used.

@arikfr  - I didn't get a chance yet to test this on another datasource, but it most certainly works with bq

### DIFF
--- a/rd_ui/app/scripts/services/resources.js
+++ b/rd_ui/app/scripts/services/resources.js
@@ -224,11 +224,7 @@
             seriesName = String(value);
           }
 
-          if (type == 'multiFilter') {
-            seriesName = String(value);
-          }
-
-          if (type == 'multi-filter') {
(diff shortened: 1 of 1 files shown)


rd_ui/app/scripts/services/resources.js
```diff
@@ -224,6 +224,10 @@
             seriesName = String(value);
           }
 
+          if (type == 'multiFilter') {
```
arikfr: better to add this to the existing if (`type == '...' || type == '...'`).

arikfr:
Looking good, I'll give it a try w/ Postgres tomorrow.

arikfr:
Thanks!

PR #551, opened 2015-09-09
Adding hidden options for widgets.

jmvasquez:
Adding hidden options for widgets. That allows save javascript, html, into other code functionality without display the empty widget.


rd_ui/app/scripts/controllers/dashboard.js
```diff
@@ -125,6 +125,12 @@
   };
 
   var WidgetCtrl = function($scope, $location, Events, Query) {
+    $scope.disableWidget = {'visibility': 'visible'};
+    $scope.checkWidth = function (width) {
+     if (width == 0)
+        $scope.disableWidget = {'visibility': 'hidden'};
```
arikfr: Why aren't you setting this in the controller, instead of in the `checkWidth` function?

rd_ui/app/views/dashboard.html
```diff
@@ -58,7 +58,7 @@ <h3 class="panel-title">
                 </div>
             </div>
 
-            <div class="panel panel-default rd-widget-textbox" ng-if="type=='textbox'" ng-mouseenter="showControls = true" ng-mouseleave="showControls = false">
+            <div class="panel panel-default rd-widget-textbox" ng-style="widget.disableWidget" ng-if="type=='textbox'" ng-mouseenter="showControls = true" ng-mouseleave="showControls = false">
```
arikfr: I've just realized that this can be simplified by using the `ng-hide` directive like this: `<div class="..." ... ng-hide="widget.width == 0" ...>`. In which case you won't need the update to the filter nor the code in the controller.

jmvasquez: When I tried to do it, the widget was not present (instead hidden) in the page, so the functionality was not the same. That was my previous experience.

arikfr: In what way it wasn't the same?

jmvasquez: Using ng-hide, the code inserted into the widget was not present in the page, so the functionality fails. For example, you created a widget with a javascript inside as text. The text shuold be present in the page but not displayed in the widget. Well, with ng-hide, the text was hidden to, NOT present in the page, so the functionality fails. If you have any question, please let me know. Thanks.

arikfr: Are you sure it was ng-hide and not ng-if?
ng-hide just applies `display:none` to the element, while ng-if doesn't render it. In theory, there shouldn't be difference between what you did and ng-hide, except that in your case it might take height on the page but still will be invisible.

jmvasquez: I´m sure, I used ng-hide. Maybe I had a mistake. But I'm sure that I used ng-hide.

arikfr:
Merged. I've changed it a bit before merging: hidden size is only available for text boxes.

PR #556, opened 2015-09-10
Feature: options for Cohort visualization

nathanlubchenco:
This is an attempt to make the cohort visualization be not just for daily, but also for weekly or monthly cohorts.  Does this seem like a good approach?


arikfr:
Making cohorts configurable is a welcomed change. I suggest that the time
interval will be a select input instead of text input (I can show you an
example if needed), and you can override the existing directive instead of
creating a new one.

Thanks!

nathanlubchenco:
Great, thanks for the feedback, a select input does seem far superior.

rd_ui/app/scripts/visualizations/cohort.js
```diff
@@ -49,13 +56,13 @@
                             container: container,
                             cohort: data,
                             title: null,
-                            timeInterval: 'daily',
+                            timeInterval: $scope.timeInterval,
```
arikfr: Shouldn't it be `$scope.visualization.options.timeInterval`?

rd_ui/app/views/visualizations/cohort_editor.html
```diff
@@ -1,15 +1,14 @@
...
-    <label class="control-label">Bucket Total Value Column</label>
-    <select ng-model="total_column" ng-options="value as key for (key, value) in columns" class="form-control"></select>
-    <label class="control-label">Day Number Column</label>
-    <select ng-model="value_column" ng-options="value as key for (key, value) in columns" class="form-control"></select>
-    <label class="control-label">Day Value Column</label>
-    <select ng-model="day_column" ng-options="value as key for (key, value) in columns" class="form-control"></select>
-</div>
\ No newline at end of file
+    <label class="control-label">Time Interval</label>
+    <select class="form-control" ng-model="cohortOptions.timeInterval">
```
arikfr: Shouldn't it be `visualization.options.timeInterval`?

rd_ui/app/scripts/visualizations/cohort.js
```diff
@@ -49,13 +56,13 @@
                             container: container,
```
arikfr: Also, because there might be existing cohort visualizations without these options, you should initialize the options with default values in the renderer.

rd_ui/app/scripts/visualizations/cohort.js
```diff
@@ -2,6 +2,13 @@
...
 
     cohortVisualization.config(['VisualizationProvider', function(VisualizationProvider) {
+
+      var editTemplate = '<cohort-editor></cohort-editor>';
+        var defaultOptions = {
+          timeInterval: 'daily',
+          timeLabel: 'day'
+        };
+
         VisualizationProvider.registerVisualization({
```
arikfr: You need to pass the editTemplate to the `VisualizationProvider.registerVisualization` call as `editorTemplate` value.

nathanlubchenco:
Thanks for the continued patience and help.   Is this the correct way to add the default values?

nathanlubchenco:
still not quite working, working on resolving, i'll keep you posted

arikfr:
The way you set default values is correct, but will only work for new cohort visualizations created. To apply them for existing visualizations you need to use them also in the renderer -- on line 29, in the link function of `cohortRenderer`, you can add:

```javascript
$scope.visualization.options.timeInterval = $scope.visualization.options.timeInterval || 'daily';
$scope.visualization.options.timeLabel = $scope.visualization.options.timeLabel || 'day';
```

Another option is to define `defaultOptions` in higher scope, and then reuse it in the link function, then you can just do: `$scope.visualization.options = _.extend({}, defaultOptions, $scope.visualization.options);`.

I should probably somehow abstract the default values concept into the base of the visualization renderers, so we don't have to redo this in every visualization. But for now...

arikfr:
Btw, you can set the `timeLabel` value based on the value of `timeInterval` because they are always linked.

Something like:

```javascript
var timeLabels = {'daily': 'Day', 'weekly': 'Week', 'monthly': 'Month'};
var timeLabel = timeLabels[$scope.visualization.options.timeInterval];
```

arikfr:
@nathanlubchenco looks like you've finished -- can I merge this?

nathanlubchenco:
Didn't end up having time to test a deployed version, but if it looks good
to you, then yes please.

arikfr:
Merged! Thanks!

nathanlubchenco:
Thank you, appreciate your help with everything.

PR #575, opened 2015-09-28
Feature: new ElasticSearch datasource, and rename previous one to Kibana

liorr:
Kept the existing ElasticSearch query implementation and added support for basic auth.

Added a new V2 ElasticSearch query to support for full blown ElasticSearch Search API (https://www.elastic.co/guide/en/elasticsearch/reference/current/search.html)

TBD: Support for Aggregations

### DIFF
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -2,6 +2,7 @@ import json
 import logging
 import sys
 import urllib
+from requests.auth import HTTPBasicAuth
 
 from redash.query_runner import *
 
@@ -23,7 +24,13 @@ ELASTICSEARCH_TYPES_MAPPING = {
     "boolean" : TYPE_BOOLEAN,
     "string" : TYPE_STRING,
     "date" : TYPE_DATE,
+    "object" : TYPE_STRING,
     # "geo_point" TODO: Need to split to 2 fields somehow
+}
+
+ELASTICSEARCH_BUILTIN_FIELDS_MAPPING = {
+    "_id" : "Id",
+    "_score" : "Score"
 }
 
 PYTHON_TYPES_MAPPING = {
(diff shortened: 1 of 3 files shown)


liorr:
@arikfr can you please review it.

I want to add support for Aggregations (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html) but it will require the query to return more than one result set is that supported? 

cc @ekampf

arikfr:
Thanks! I will review this soon. 

As for aggregation and multiple result sets - this is not supported. But maybe you can transform the multiple result sets into a single set by adding another column(s) that indicate the "group"?

ekampf:
@arikfr maybe its worth renaming existing ES provider to be called "Kibana" as it does Kibana specific logic (like continuously fetching all results with a cursor - which can be too large\time consuming for a Search oriented query)

arikfr:
@ekampf that's a good idea! Will make things clearer. Just need to override the `type` class method , to return still return "elasticsearch" for backward compatibility (another option is to add a migration that updates the type name for existing data sources).

arikfr:
Sorry for not reviewing yet -- have some deadline I need to meet on another project + leaving on vacation tomorrow until Tuesday. I will get back to it once I'm back.

redash/query_runner/elasticsearch.py
```diff
@@ -121,11 +86,24 @@ def __init__(self, configuration_json):
...
-    def get_mappings(self, url):
+            logger.setLevel(logging.DEBUG)
+
+        self.server_url = self.configuration["server"]
+        if self.server_url[-1] == "/":
+            self.server_url = self.server_url[:-1]
+
+        basic_auth_user = self.configuration["basic_auth_user"]
+        basic_auth_password = self.configuration["basic_auth_password"]
+        self.auth = HTTPBasicAuth(basic_auth_user, basic_auth_password) if basic_auth_user and basic_auth_password  else None
```
arikfr: The following will be more readable:
```python
self.auth = None
if basic_auth_user and basic_auth_password:
    self.auth = HTTPBasicAuth(basic_auth_user, basic_auth_password)
```

redash/query_runner/elasticsearch.py
```diff
@@ -153,32 +138,85 @@ def parse_results(self, mappings, result_fields, raw_result, result_columns, res
...
+
+# {
+#     "index" : "logstash-2015.04.*",
+#     "query" : "type:events AND eventName:UserUpgrade AND channel:selfserve",
+#     "fields" : ["@timestamp", "userId", "channel", "utm_source", "utm_medium", "utm_campaign", "utm_content"],
+#     "size" : 250,
+#     "sort" : "@timestamp:asc"
+# }
+#
+#
```
arikfr: No need to keep this, I moved it to the docs: http://docs.redash.io/en/latest/usage/elasticsearch_querying.html

redash/query_runner/elasticsearch.py
```diff
@@ -252,4 +285,85 @@ def run_query(self, query):
...
+#       "match": {
+#           "user" : "kimchy"
+#       }
+#     },
+#     "fields" : ["@timestamp", "tweet", "user"],
+#     "size" : 15,
+#     "sort" : "@timestamp:asc"
+# }
+#
+#
```
arikfr: Worth moving this to the doc page: http://docs.redash.io/en/latest/usage/elasticsearch_querying.html

redash/query_runner/elasticsearch.py
```diff
@@ -252,4 +285,85 @@ def run_query(self, query):
...
+
+            url = "{0}/{1}/_search".format(self.server_url, index_name)
+            mapping_url = "{0}/{1}/_mapping".format(self.server_url, index_name)
+
+            mappings = self._get_mappings(mapping_url)
+
+            logger.debug(json.dumps(mappings, indent=4))
+
+            logger.debug("Using URL: {0}".format(url))
+            logger.debug("Using Query: {0}".format(json.dumps(query_params)))
```
arikfr: Using `.format` method here means that even if the logging level is higher than debug, the string will be created. Use the old format: `logger.debug("Using Query: %s", query_params)` (I've also removed the `json.dumps` call, because otherwise it will get serialized every time regardless of logging level).

arikfr:
It looks good, aside from few small things. I suggest we do what @ekampf suggested (rename old one to Kibana) and it can be merged.

@erans wdyt about the rename?

erans:
Go for it. I'm fine with that.

arikfr:
@liorr ping.

liorr:
Sorry but I am sick, I'll get back to it as soon as I feel a bit better.

arikfr:
Sure thing, feel better!

ekampf:
@arikfr ping? :)

arikfr:
@ekampf sorry about that... I finally getting back to re:dash. On the good side: I'm working on making re:dash my full time job :)

ekampf:
@arikfr good luck!

PR #576, opened 2015-09-29
Fix: support for duplicate columns for MySQL query runner

### LINKED ISSUE #571: Columns with same name (on different tables) are messing up the results
If we have two columns with the same name in the same query the result is overridden. In this query I also found another bug: the `COUNT` and the `GROUP_CONCAT` does not show results if they don't have an alias.

```sql
SELECT s.name,
       t.name,
       count(tg.id),
       group_concat(tg.taggable_id
                    ORDER BY tg.taggable_id ASC separator ', ')
FROM taggings tg,
     tags t,
     questions q,
     subgroups s
WHERE t.id = tg.tag_id
  AND tg.taggable_id = q.id
  AND tg.taggable_type = 'Question'
  AND q.year >= 2009
  AND s.id = q.subgroup_id
GROUP BY s.name,
         t.id
```

Result:
screen shot 2015-09-22 at 2 07 53 pm

Should be:
screen shot 2015-09-22 at 2 08 52 pm

Adding an alias to all columns fixes it.

joaofraga:
This Pull Request fixes #571 issues with duplicated column names:

> If we have two columns with the same name in the same query the result is overridden. In this query I also found another bug: the `COUNT` and the `GROUP_CONCAT` does not show results if they don't have an alias.

This is my very first python code. All comments are welcome.

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -79,7 +79,9 @@ class BaseQueryRunner(object):
                 duplicates_counter += 1
 
             column_names.append(column_name)
-            new_columns.append([column_name, col[1]])
+            new_columns.append({'name': column_name,
+                                'friendly_name': column_name,
+                                'type': col[1]})
 
         return new_columns
 
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -75,7 +75,7 @@ class BaseQueryRunner(object):
         for col in columns:
             column_name = col[0]
             if column_name in column_names:
-                column_name += " " + str(duplicates_counter)
+                column_name = "{}_{}".format(column_name, duplicates_counter)
                 duplicates_counter += 1
 
             column_names.append(column_name)
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -75,7 +75,7 @@ class BaseQueryRunner(object):
         for col in columns:
             column_name = col[0]
             if column_name in column_names:
-                column_name = "{}_{}".format(column_name, duplicates_counter)
+                column_name = "{}{}".format(column_name, duplicates_counter)
                 duplicates_counter += 1
 
             column_names.append(column_name)
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -67,6 +67,22 @@ class BaseQueryRunner(object):
     def run_query(self, query):
         raise NotImplementedError()
 
+    def fetch_columns(self, columns):
+        column_names = []
+        duplicates_counter = 1
+        new_columns = []
+
+        for col in columns:
+            column_name = col[0]
+            if column_name in column_names:
+                column_name += " " + str(duplicates_counter)
+                duplicates_counter += 1
+
+            column_names.append(column_name)
+            new_columns.append([column_name, col[1]])
+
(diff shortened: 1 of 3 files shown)


joaofraga:
Should I implement this method in another query runner object?

redash/query_runner/__init__.py
```diff
@@ -67,6 +67,22 @@ def configuration_schema(cls):
...
 
+    def fetch_columns(self, columns):
+        column_names = []
+        duplicates_counter = 1
+        new_columns = []
+
+        for col in columns:
+            column_name = col[0]
+            if column_name in column_names:
+                column_name += " " + str(duplicates_counter)
```
arikfr: The column names are used as field names in the object passed to the client, so it's better to use an underscore.

arikfr: Also you can use string formatting: `column_name = "{}_{}".format(column_name, duplicates_counter)`.

redash/query_runner/pg.py
```diff
@@ -121,29 +121,14 @@ def run_query(self, query):
...
+                columns_data = self.fetch_columns([(i.name, types_map.get(i.type_code, None)) for i in cursor.description])
 
-                    columns.append({
-                        'name': column_name,
-                        'friendly_name': column_name,
-                        'type': types_map.get(column.type_code, None)
-                    })
+                columns = [{'name': col[0],
+                            'friendly_name': col[0],
+                            'type': col[1]} for col in columns_data]
```
arikfr: Maybe change `fetch_columns` method to return an array in this format? It will save the need to make another loop at the array.

arikfr:
Thank you @joaofraga for your effort on this! In general it's looking good, but I left some comments with small improvements.

While this can be applied to most of the other query runners, I prefer to test it before pushing to the release. So I'll gradually apply them to the other query runners, once I have the chance to test it.

redash/query_runner/pg.py
joaofraga: Sure, will do this.

redash/query_runner/__init__.py
joaofraga: @arikfr I prefer to use table_name.column_name in this duplicated issues, what do you think about it?  Can I get the table name of the given column in some place? Also how can I debug in console what I'm working with? I couldn't did that with `print`.

joaofraga:
I've made the improvements that you suggested. I searched in the others query runners to apply that but my limited knowledge (practically non-existent :joy:) of Python didn't allow me to do much.

redash/query_runner/__init__.py
```diff
@@ -67,6 +67,24 @@ def configuration_schema(cls):
...
 
+    def fetch_columns(self, columns):
+        column_names = []
+        duplicates_counter = 1
+        new_columns = []
+
+        for col in columns:
+            column_name = col[0]
+            if column_name in column_names:
+                column_name = "{}_{}".format(column_name, duplicates_counter)
```
arikfr: I've just realized that we should keep the old format ("{}{}", without the underscore). Because it will change names for existing queries, which might result in existing visualizations breaking for users.

joaofraga: Ok @arikfr I'll do it.

arikfr:
Merged. Thanks!

PR #579, opened 2015-09-29
Include materialized views in postgres schema

robotblake:
Tested against PostgreSQL 9.4, currently running in production.

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -87,7 +87,22 @@ class PostgreSQL(BaseQueryRunner):
         query = """
         SELECT table_schema, table_name, column_name
         FROM information_schema.columns
-        WHERE table_schema NOT IN ('pg_catalog', 'information_schema');
+        WHERE table_schema NOT IN ('pg_catalog', 'information_schema')
+        UNION
+        SELECT
+          pg_namespace.nspname table_schema,
+          pg_class.relname table_name,
+          pg_attribute.attname column_name
+        FROM pg_catalog.pg_class
+        JOIN pg_catalog.pg_namespace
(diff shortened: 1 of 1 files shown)


arikfr:
I will make sure it works with Redshift, and then merge. 

Thanks!

redash/query_runner/pg.py
```diff
@@ -87,7 +87,22 @@ def get_schema(self):
...
+          pg_class.relname table_name,
+          pg_attribute.attname column_name
+        FROM pg_catalog.pg_class
+        JOIN pg_catalog.pg_namespace
+          ON pg_namespace.oid = pg_class.relnamespace
+         AND pg_namespace.nspname NOT IN ('pg_catalog', 'information_schema')
+        JOIN pg_catalog.pg_attribute
+          ON pg_attribute.attrelid = (
+            pg_namespace.nspname || '.' || pg_class.relname
+          )::regclass
```
arikfr: When running this query on Redshift (which supposed to act as Postgres v8.4, iirc), I get the following error:

```
ERROR:  cannot cast type text to regclass
```

An easy way to handle this (instead of trying to figure which versions of Postgres support this), is to run two queries -- one is the original, and the other is this new one. If it fails, just log it and continue. If it worked, merge its results with the previous one (merge the results in Python instead of in Postgres).

WDYT?

arikfr:
@robotblake ping.

redash/query_runner/pg.py
robotblake: Sorry I missed this, sounds good to me!

arikfr:
Please re-open after update. Thanks.

Issue #117, opened 2014-03-03
TV mode


ranbena:
Have a fullscreen button on dashboards (would be cool to use the [fullscreen API](https://developer.mozilla.org/en-US/docs/Web/Guide/API/DOM/Using_full_screen_mode))

Perhaps use reveal.js to navigate between dashboard items, but if it's not supported by AngularJS you can always [do it yourself](http://desandro.github.io/3dtransforms/docs/carousel.html).

shenyubao:
:+1:

TonyStarkBy:
So what about fullscreen? There is a cool library - https://github.com/sindresorhus/screenfull.js
Only one small issue in safari: keyboard is not working in fullscreen mode, but it's not need :)

P. S. https://sindresorhus.com/screenfull.js you can test it

rohanpd:
This issue is solved now isn't it? by #748

arikfr:
@rohanpd not really, because if you have more widgets than fits the screen it's not really helping. Proper TV mode will probably have some auto scrolling support.

PR #586, opened 2015-10-05
Dashboard column change

matthew-sochor:


arikfr:
I'll have to see how it looks before merging.

Btw, the comment about changing bootstrap.css from your first commit still relevant?

matthew-sochor:
No, the next commit with container-fluid fixed that issue.  I'll edit the first commit message.  

I think a more elegant solution might be to have the ability to toggle between a wide format and the current tall and skinny format. I really wanted the full width/3 columns because we display some dashboards on TVs and the huge margins are a real drawback.

rd_ui/app/scripts/filters.js
```diff
@@ -49,9 +49,12 @@ angular.module('redash.filters', []).
...
     return function (widgetWidth) {
       if (widgetWidth == 1) {
-        return 6;
+        return 4;
+      } else if (widgetWidth == 2) {
+        return 8;
+      } else {
+        return 12;
       }
-      return 12;
```
arikfr: While you have support for full width widgets in code, there is no UI for it.

arikfr: Also, looks like the gridster preview wasn't updated for the new dimensions: 

image

arikfr:
I think that until we do some redesign of the whole interface, we should hide this behind a toggle (can be done with ng-class, and adding a toggle to the dashboard options view). Otherwise it just doesn't look good:

image

(@ranbena, when will you implement the TV Mode? #117 :))

arikfr:
I apologize but I can't merge this at current state. But expect improvements to the grid soon enough...

PR #588, opened 2015-10-05
Docker deployment support

tjwudi:
Add docker support (see #450).

Couple things to note:

- I removed imagemin task in grunt since 
   - It is obsolete and cannot be downloaded anymore (at least on my side). It can block building so I removed it.
   - It is useless, I mean, in re:dash
- Example docker-compose.yaml is added.
- Made service listen on `0.0.0.0` instead of only `127.0.0.1`. So docker containers can accept incoming traffic then.


Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM debian:wheezy
```
arikfr: Is there a reason to use Debian and not Ubuntu as the base image?

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# Build frontend sources
+WORKDIR /opt/redash/current/rd_ui
+USER redash
+RUN mkdir -p /opt/redash/.npm/node_modules
+RUN npm config set prefix /opt/redash/.npm/node_modules
+RUN npm install -g bower
+RUN npm install -g grunt-cli
+RUN npm install
+RUN /opt/redash/.npm/node_modules/bin/bower install
+RUN /opt/redash/.npm/node_modules/bin/grunt build
```
arikfr: The whole Node.js stack isn't needed in production (the release tarballs already include "compiled" assets), and shouldn't be part of the base image. We can create another image for development, that derives from the base one and installs the Node.js stack.

setup/bootstrap_docker.sh
```diff
@@ -0,0 +1,85 @@
...
+apt-get update
+apt-get install -y python-pip python-dev curl build-essential pwgen libffi-dev sudo git-core
+pip install -U setuptools
+
+# redash user
+useradd --system --comment " " --create-home redash
+
+# PostgreSQL
+apt-get update
+apt-get -y install libpq-dev postgresql
```
arikfr: Isn't there a `postgresql-client` package we can use instead of `postgresql`?

setup/bootstrap_docker.sh
```diff
@@ -0,0 +1,85 @@
...
+
+# Redis client
+mkdir -p /tmp/redash && cd $_
+curl -L https://github.com/antirez/redis/archive/3.0.4.tar.gz -o redis.tar.gz
+mkdir redis
+tar -zxvf redis.tar.gz -C redis --strip-components 1
+cd redis
+make redis-cli
+cp src/redis-cli /usr/local/bin
+rm -rf /tmp/redis
```
arikfr: Are you sure we need redis installed, to install the Redis Python package?

Because otherwise we don't need the Redis client on this image.

setup/docker_init_migration.sh
```diff
@@ -0,0 +1,15 @@
...
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0005_add_updated_at.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0006_queries_last_edit_by.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0007_add_schedule_to_queries.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0008_make_ds_name_unique.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0009_add_api_key_to_user.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0010_create_alerts.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0010_allow_deleting_datasources.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0011_migrate_bigquery_to_json.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0012_add_list_users_permission.py
+sudo -u postgres PYTHONPATH=. bin/run python migrations/0013_update_counter_options.py
```
arikfr: In what case we need to run `docker_init_migrations.sh`? Assuming the user runs `docker_init_postgres.sh` it will create the database according to the latest db structure, and no migrations will be needed.

setup/files/supervisord.conf
```diff
@@ -5,13 +5,13 @@ pidfile=/opt/redash/supervisord/supervisord.pid
 directory=/opt/redash/current
 
 [inet_http_server]
-port = 127.0.0.1:9001
+port = 0.0.0.0:9001
```
arikfr: Considering that you're using a separate file for docker anyway, better to revert the changes to this file.

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# Reset working directory and user for future `docker attach` sessions
+WORKDIR /opt/redash/current
+USER root
+
+CMD service redash_supervisord start
```
arikfr: Ideally, it will be better replace supervisord with running multiple docker images, but we can keep it for now. But let's not run it daemonized -- you can run supervisord in foreground mode (with the `-n` argument).

docker-compose-example.yaml
```diff
@@ -0,0 +1,17 @@
...
+  build: .
+  ports:
+    - "5000:5000"
+    - "9001:9001"
+  links:
+    - redis
+    - postgres
+  stdin_open: true
+redis:
+  image: redis
```
arikfr: Let's pin versions both of Redis & Postgres to avoid compatibility issues.

arikfr:
Thanks, @tjwudi ! This is much appreciated (and long awaited).

I've left some comments.

setup/docker_init_migration.sh
tjwudi: Yeah you are right. I didn't know that this script is actually covered already in `docker_init_postgres.sh`.

setup/files/supervisord.conf
tjwudi: Well, I think it might be (generally) a better practice to bind to `0.0.0.0`. I mean, there is no reason not to do so :) Do you have further thinking on this?

arikfr: It's better from security perspective. The Python Web server is more
vulnerable than nginx that proxies it.

But anyway better not to have the supervisord control port exposed because
iirc it has no authentication.

tjwudi: Makes sense! Learning many new things here. Will update then. :)

Dockerfile
tjwudi: Yeah, that's what I did to `supervisord_docker.conf` by setting `nodeamon=true`

arikfr: Missed that, but anyway no need to run it as service/with init. Just run
the supervisor command.

tjwudi:
Also, @arikfr is it a good practice to use db containers?  Since files modified will be discarded when container stops, then all data will be lost if we do not commit the container regularly.

tjwudi:
Updated!

arikfr:
@tjwudi that's what data volumes are for: https://docs.docker.com/userguide/dockervolumes/

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+MAINTAINER Di Wu <diwu@yelp.com>
+
+COPY . /opt/redash/current/
+
+WORKDIR /opt/redash/current
+
+ENV FILES_BASE_URL /opt/redash/current/setup/files/
+
+# Base packages
+RUN apt-get dist-upgrade
```
arikfr: This line is probably not needed. I had some issues with AWS Ubuntu images, and added it. But normally everything should work fine without it.

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+RUN apt-get update
+RUN apt-get install -y python-pip python-dev curl build-essential pwgen libffi-dev sudo git-core
```
arikfr: Contrary to what you would expect, it's a best practice to merge multiple RUN commands into one multi command. So in this case it will become:

```
RUN apt-get update \
         apt-get install -y ...
```

This to avoid bloat of multiple layers that won't be really used.

So it's hard to point what commands you should group together, and it's left to the Docker file creator to decide, but I would take all the apt-get commands from this file and group them into one RUN command.

docker-compose-example.yaml
```diff
@@ -0,0 +1,17 @@
...
+  links:
+    - redis
+    - postgres
+  stdin_open: true
+redis:
+  image: redis:3.0
+  ports:
+    - "6379:6379"
+postgres:
+  image: postgres:9.4
```
arikfr: Currently the bootstrap script installs Postgres 9.3 and Redis 2.8.17. Let's keep using those for now to maintain parity between environments. If we end up needing a feature from a newer version in the future, we will be able to migrate easily using Docker. Hopefully by then many users will migrate to Docker based setup.

Dockerfile
tjwudi: Which line is it? GitHub didn't point out which one exactly

arikfr: `RUN apt-get dist-upgrade`

tjwudi:
@arikfr then probably add data volumes in **docker-compose-example.yaml**?

arikfr:
yep.

tjwudi:
Note: I think it would be better to update docs too in this PR. But given that the image should be pushed to docker hub by EverythingMe, or @arikfr specifically, I would suggest we add docs after we've published the image to docker hub. I mean, I will create a separate branch for docs update.

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Get supervisord startup script
+RUN cp $FILES_BASE_URL"supervisord_docker.conf" /opt/redash/supervisord/supervisord.conf
+
+RUN cp $FILES_BASE_URL"redash_supervisord_init" /etc/init.d/redash_supervisord
+
+# Fix permissions
+RUN chown -R redash /opt/redash
+
+CMD bash /etc/init.d/redash_supervisord start
```
arikfr: You don't need to use init script - just invoke `supervisord`:

```
CMD ['supervisord', '-c', '/opt/redash/supervisord/supervisord.conf']
```

arikfr: Ok, instead of commenting on separate lines, I reviewed the whole file and here's what I suggest. Note that this is only a suggestion.

```Dockerfile
FROM ubuntu:trusty
MAINTAINER Di Wu <diwu@yelp.com>

# Ubuntu packages
RUN apt-get update && \
    apt-get install -y python-pip python-dev curl build-essential pwgen libffi-dev sudo git-core && \
    # Postgres client
    apt-get -y install libpq-dev postgresql-client && \
    # Additional packages required for data sources:
    apt-get install -y libssl-dev libmysqlclient-dev

ENV REDASH_STATIC_ASSETS_PATH="../rd_ui/dist/" \
    REDASH_LOG_LEVEL="INFO" \
    REDASH_REDIS_URL=redis://redis:6379/0 \
    REDASH_DATABASE_URL="postgresql://redash@postgres/redash"

# Get latest version tarball:
# We install it directly into /opt/redash/current rather than symlink, because in 
# a Docker environment we will replace the re:dash image, rather than upgrade it.
RUN wget $(python latest_release_url.py) -O redash.tar.gz && \
    mkdir -p /opt/redash/current && \
    tar -C /opt/redash/current -xvf redash.tar.gz && \
    rm redash.tar.gz

# Install Python dependencies
WORKDIR /opt/redash/current
RUN pip install -U setuptools && \
    pip install -r requirements.txt && \
    # pip requirements for all data source types
    pip install -r requirements_all_ds.txt

# Setup supervisord
RUN mkdir -p /opt/redash/supervisord && \
    mkdir -p /opt/redash/logs && \
    pip install supervisor==3.1.2 && \
    cp /opt/redash/current/setup/files/supervisord_docker.conf /opt/redash/supervisord/supervisord.conf

# Get latest version tarball
CMD ["supervisord -c /opt/redash/supervisord/supervisord.conf"]
```

Notes:
1. I grouped all apt-get commands from across the file.
2. I removed user creation: we don't need the postgres user, it exists in the postgres image. Also we don't really need the redash user, as it less needed in Docker environment and will only complicate things (we will run things as root; we can revisit this later on).
3. Instead of `.env` file, let's use ENV directive and `env_file`/`env` option in docker-compose. We need to remember to pass REDASH_COOKIE_SECRET in docker-compose. Also we need to update the supervisord config to not use bin/run.
4. latest_release_url.py is a script that will get the url of the latest tarball file (see below).
5. I'm not sure how the Docker postgres image handles authentication, so the database URL might need to be updated. Also we need to update the `docker_init_postgres.sh` script to run from the context of the docker host, and run some commands with the redash image and some with the postgres image.
6. I suggest to move the Dockerfile and its related scripts/files into `setup/docker`.

```python
# latest_release_url.py
import urllib2
import json

latest = json.load(urllib2.urlopen("https://api.github.com/repos/EverythingMe/redash/releases/latest"))

print latest['assets'][0]['browser_download_url']
```

arikfr: Correction to note #5: we might be able to run the init_postgres script from the redash image, because it already has postgres client, therefore maybe no need to change the script.

arikfr:
@tjwudi I can create the image before merging, if you want to write the docs already. Or we can do it in a separate branch - up to you.

lenguyenthedat:
Thanks for doing this! :+1:

arikfr:
Just noticed: we need to add nginx image too (I assume they have an official one).

arthurpsmith:
Has somebody got this working? I don't quite see how the docker_init_postgres.sh will run. I've actually got a docker image of re:dash running that is quite similar to this, but I had to run the postgres database initialization in a script at startup - i.e. the startup command is:
CMD /opt/redash_start.sh
- also this was a bit tricky with authentication since having postgres running on a different server means simple username-based authentication isn't allowed. I had to set up passwords for the postgres user (db administrator, via an environment variable) and the redash user (in /home/redash/.pgpass) to get it to run.

Dockerfile
tjwudi: hmm, I think we should allow users to build from their source, but not download the latest version of re:dash. Sure we build and publish official images, but users can have their own build.

arikfr:
The script will be executed using the redash container, creating the necessary users and tables. Proper instructions will be part of the docs.

Dockerfile
arikfr: We can create a separate image for running re:dash from their source (usually will be used for development), that can be based on this one. But I wouldn't want to build the "official" images from source (will require more setup on the image, and duplicate work the CI does already).

tjwudi: Not exactly. I think there is some miscommunication. Because I don't see any difference between copy from local source and download from latest code, besides that `COPY` allows user to build their own images based on their code, which doesn't change our workflow at all.

arikfr: The tar ball isn't just the packaged source code. It has the front end
assets already "compiled", which saves the need to bloat the image with
Node.js dependencies that are not needed in production.

So what I had in mind is to have a base image which we use for production.
And another image for development that extends it with the Node.js stack
and uses redash sources from an attached volume that will be mapped to
user's local source code (the use of volume instead of copy is to avoid the
need to rebuild the image everytime you edit the source code).

But if you prefer to start simpler with one image, without volumes, that
uses local files to build the image that's fine too. Because I can probably
run the Docker image creation from the CI machine. Just replace the tar
ball download with the copy command. The rest of my suggestions still make
sense, no?

tjwudi: It turns out we still need redash user. celery will complain about superuser privilege like this

```
Running a worker with superuser privileges when the
worker accepts messages serialized with pickle is a very bad idea!

If you really want to continue then you have to set the C_FORCE_ROOT
environment variable (but please think about this before you do).

User information: uid=0 euid=0 gid=0 egid=0
Running a worker with superuser privileges when the
worker accepts messages serialized with pickle is a very bad idea!

If you really want to continue then you have to set the C_FORCE_ROOT
environment variable (but please think about this before you do).

User information: uid=0 euid=0 gid=0 egid=0
```

arikfr: I see. We can make sure we don't use Pickle (we not supposed to anyway), but I guess that using a dedicated user should be an issue.

tjwudi: Yeah, I got your idea now. What I am thinking is to use Vagrant (which re:dash already has) for development and asset building, then user build docker image with compiled assets in place for production.

tjwudi: I am not familiar with Pickle at all. My current workaround is to run celery with `sudo -u redash`. I removed postgres user though.

bobrik: @tjwudi building from vagrant doesn't allow you to have automated build that can be repeated by anyone else without any environment requirements.

arikfr: @bobrik not that I want to keep Vagrant around for long, but Docker is a requirement too, isn't it?

bobrik: Docker should be the only requirement for self-contained builds. You do `docker build -t redash .` and you get your image. No other steps should be required.

It is totally fine to build dev image from already compiled assets, but release images are different.

arikfr: :+1:

arthurpsmith:
FYI below is the docker-compose.yml file I've been using with success. Some notes on differences:
* No need to expose the ports on the redis and postgres containers - these are automatically available to the redash container via the link, they do not need to be exposed on the docker host machine.
* with a POSTGRES_PASSWORD environment variable, the default admin user (postgres) on the postgres server can be accessed by username/password authentication. In my "redash_start.sh" script I have a line export PGPASSWORD=${POSTGRES_PASSWORD} when I need to log in as the postgres user via psql (ie. psql -h postgres -U postgres postgres ... works by pulling the password from the environment).
* Uses a volume on the host to have a persistent database between updates of the postgres container. A different volume choice might be suitable depending on your needs.
* exposes the redash port (5000) via a port map on the host. If you want nginx in front you could, (that would be one more container with the nginx image + a custom config file) but it doesn't seem  needed.

```
redis:
  image: redis

postgres:
  image: postgres
  environment:
    POSTGRES_PASSWORD: xxxxxxxx
  volumes:
   - /opt/postgres-data:/var/lib/postgresql/data

redash:
  build: .
  container_name: redash
  links:
   - redis
   - postgres 
  environment:
    POSTGRES_PASSWORD: xxxxxxxx
  ports:
   - "8080:5000"
```

bobrik:
I played with re:dash a bit and decided that it's nice to split celery and web into separate containers. It's easier to run on marathon and scale this way and you don't have to mess with init systems.

Dockerfile
tjwudi: Convinced. Will update later.

tjwudi:
So I decided to create a separate image for redash nginx. It's now live on <https://hub.docker.com/r/tjwudi/redash-nginx/> and repo <https://github.com/tjwudi/redash-nginx>.

arikfr:
Why not keep the Dockerfile and nginx.conf for the nginx image in this
repo? It will make it easier to manage access and edit Docker related
configuration.

Also make sure to add daemon off directive to nginx.conf.

tjwudi:
@arikfr Also makes sense. Though I prefer one image one repo, I can do it if that is what you guys want :)

tjwudi:
@arikfr daemon is off by default

tjwudi:
Well the thing is, the latest build does not contain **supervisord_docker.conf** so it can fail if you try to build. Anyway, it is now downloading from latest build instead :)

tjwudi:
@bobrik Separating celery is definitely what we want to do next, but probably in another branch maybe? b/c I don't know the complexity to do it yet.

arikfr:
I agree that we can split the images in another iteration/branch. Also the resulting image can still be used without the supervisord with custom run command.

setup/files/docker-redash-nginx/nginx.conf
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  server {
+    listen  80;
+
+    access_log  /var/log/nginx/log/access.log;
+    error_log  /var/log/nginx/log/error.log;
+
+    location / {
+      proxy_pass  http://redashapp;
```
arikfr: Unless in newer versions of nginx those options are on by default worth adding to the server directive:
```
  gzip on;
  gzip_types *;
  gzip_proxied any;
```

and to the `location /` directive:
```
    proxy_set_header Host $http_host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $scheme;
```

tjwudi: Thanks for this!

Dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM ubuntu:trusty
+MAINTAINER Di Wu <diwu@yelp.com>
+
+ENV FILES_BASE_URL /opt/redash/current/setup/files/
```
arikfr: this is no longer used.

tjwudi: Removed :)

arikfr:
Merged. I will probably tweak it some more myself, to make it work with CI and everything.

**Thanks** a lot for the effort making this and patience during the review. :-) :+1:

tjwudi:
My pleasure. Learned a lot here. Will go on contributing :) @arikfr ++

arikfr:
Awesome :+1:

tjwudi:
@arikfr are you working on separating celery tasks? If not, I am happy to take over.

arikfr:
@tjwudi you mean having Celery and gunicorn in separate containers?

PR #589, opened 2015-10-06
Fix: Google spreadsheet data source: cast values to their actual type from string 

shyamgopal:
Please review.

1) Tested for counter, chart (line/column) types of visualisation
2) Tested for integer, float and datetime datatypes


arikfr:
Merged. Thanks!

shyamgopal:
Thanks !

PR #590, opened 2015-10-07
Added "Select last used data source" to query view

twolfson:
We at Underdog.io are loving `redash` for setting up internal dashboards. We have 1 minor pain point which is `redash:metadata` is always selected as the first data source for a new query. While we could delete the `redash:metadata`, it would be really kicking the can down the road and not solving the problem.

We had an idea to re-use the selection that a user had from the last time they change the data source (e.g. if I use "db-a", then I will see "db-a" selected the next time I create a new query). This PR is working solution for that. In this PR:

- Added load/save `last_selected_data_source_id` in `localStorage` as part of `dataSourceId` interactions

**Notes:**

I am not too familiar with Angular so I wasn't sure if `localStorage` should be injected as a dependency. If you have any requests for how the code should be organized or about using library (e.g. `ded/kizzy`), please let me know.

### DIFF
--- a/rd_ui/app/scripts/controllers/query_view.js
+++ b/rd_ui/app/scripts/controllers/query_view.js
@@ -19,14 +19,35 @@
       $scope.queryResult = $scope.query.getQueryResult(maxAge, parameters);
     }
 
+    var getDataSourceId = function() {
+      // Try to get the query's data source id
+      var dataSourceId = $scope.query.data_source_id;
+
+      // If there is no source yet, then parse what we have in localStorage
+      //   e.g. `null` -> `NaN`, malformed data -> `NaN`, "1" -> 1
+      if (dataSourceId === undefined) {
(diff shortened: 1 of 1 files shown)


arikfr:
This is so great and much needed. Thanks!

Two small things:
1. I try to use camelCasing for the frontend stuff (aside from objects that come back from the server). So let's use `lastSelectedDataSourceId` as the key for localStorage.
2. It's a rare edge case, but the data source might not exist anymore so worth checking that it does (you can use `_.some` function for that).

twolfson:
Sure thing, taking a look at adding those now.

twolfson:
Alright, the PR has been updated to use `lastSelectedDataSourceId` and handle non-existant data sources.

arikfr:
Merged. Thanks!

twolfson:
Awesome, thanks for the quick turnaround =)

PR #593, opened 2015-10-08
Chart colour palette updated

atharvai:
Following new colours added to chart colour palette: Brown, Yellow, Black,Gray,Pink, Dark Blue.


arikfr:
Merged, thanks!

PR #594, opened 2015-10-08
Feature: date range selector support for charts

tjwudi:
screen shot 2015-10-08 at 2 57 34 pm

Date range selector allows you to zoom in & zoom out by date range. Although, we can use HighStock to achieve this, it is quite a big overhead, both to performance and engineering time. It is quite simple though.


rd_ui/app/scripts/visualizations/chart.js
```diff
@@ -27,11 +27,41 @@
...
       controller: ['$scope', function ($scope) {
         $scope.chartSeries = [];
         $scope.chartOptions = {};
+        $scope.dateRangeEnabled = $scope.options.xAxis && $scope.options.xAxis.type === 'datetime';
+        $scope.dateRange = { min: moment('1970-01-01'), max: moment() };
+
+        /**
+         * Update date range by finding date extremes
+         *
+         * ISSUE: chart.getExtreme() does not support getting Moment object out of box
```
arikfr: If `chart.getExtreme` works right but the only problem is that it returns non `Moment` objects, we can convert them. Might be easier than searching ourselves.

tjwudi: No, it simply returns something like `1`, `2`, `3` ...

rd_ui/app/scripts/services/resources.js
```diff
@@ -229,26 +255,18 @@
...
-          series[seriesName]['data'].push(point);
-        }
-
         if (seriesName === undefined) {
           _.each(yValues, function (yValue, seriesName) {
-            addPointToSeries(seriesName, {'x': xValue, 'y': yValue});
-          });
-        } else {
-          addPointToSeries(seriesName, point);
+            this._addPointToSeries({'x': xValue, 'y': yValue}, series, seriesName);
```
arikfr: You need to use `this._addPointToSeriesIfInDateRange` (if applicable) here too, otherwise when charting a query like "SELECT x, y" it won't apply the date filter.

rd_ui/app/scripts/visualizations/chart.js
```diff
@@ -39,19 +69,26 @@
...
                 if (!additional.name || additional.name == "") {
-                  additional.name = s.name;
+                  additional.name = series.name;
                 }
               }
-              $scope.chartSeries.push(_.extend(s, additional));
+              $scope.chartSeries.push(_.extend(series, additional));
             });
+
+            setDateRangeToExtreme(allSeries);
```
arikfr: This causes an infinite digest loop, because `reloadData` is called when `dateRange` changes, which results in `dateRange` change. Also, it overrides the user selection...

It seems like it needs to be called only when the chart data changes.

rd_ui/app/scripts/visualizations/chart.js
```diff
@@ -27,11 +27,41 @@
...
+              _.each(s.data, function (point) {
+                if (point.x.isBefore(minDateRange)) {
+                  // Use the copy of point.x to prevent side effects
+                  minDateRange = moment(point.x);
+                }
+                if (point.x.isAfter(maxDateRange)) {
+                  maxDateRange = moment(point.x);
+                }
+              });
+            });
```
arikfr: This logic can be simplified by using [`_.min`](http://underscorejs.org/#min), [`_.max`](http://underscorejs.org/#max) and [`_.pluck`](http://underscorejs.org/#pluck).

tjwudi: Oh actually I guess we should use `[moment.min](http://momentjs.com/docs/#/get-set/min/)` and its counterpart

rd_ui/app/scripts/visualizations/chart.js
tjwudi: No, it won't. `if (!minDateRange.isSame(oldMinDateRange)) { ... }` in `$watch` prevents it to do so. But it does do one unnecessary `setDateRangeToExtreme` call when data changed by date range filter. Regardless, I will try a better way.

arikfr:
@tjwudi this is ready for merge, right?

tjwudi:
Yup, for me it's ready :) @arikfr

toyama-h:
:+1:

shenyubao:
:+1:

shenyubao:
:+1:

Danier-Evens:
@arikfr  when to merge?

tjwudi:
@Danier-Evens probably in this week. @arikfr sent me an email saying that he is kinda busy working on something else. :)

Danier-Evens:
@tjwudi  @arikfr   OK! Thanks.

rd_ui/app/scripts/visualizations/chart.js
```diff
@@ -27,31 +27,62 @@
...
+        var setDateRangeToExtreme = function (allSeries) {
+          if ($scope.dateRangeEnabled && allSeries && allSeries.length > 0) {
+            $scope.dateRange = {
+              min: moment.min.apply(undefined, _.map(allSeries, function (series) {
+                return moment.min(_.pluck(series.data, 'x'));
+              })),
+              max: moment.max.apply(undefined, _.map(allSeries, function (series) {
+                return moment.max(_.pluck(series.data, 'x'));
+              }))
+            };
```
arikfr: Are you sure this is working?

I've just tried it, and I get min as 1/1/1970 and max as current date.

arikfr: I just noticed that you upgraded moment. I guess that's the issue.

arikfr:
Took me a while, but it's finally merged. 

:+1:

tjwudi:
Awesome! Thanks!

PR #599, opened 2015-10-11
Fix: don't send passwords back to the UI

### LINKED ISSUE #554: Don't return data source options via API
Currently, the route `api/data_sources/:id` returns the entire data source, including its various config options. Meaning, an authenticated admin essentially gains the direct database credentials.

While this might sound like a tiny issue, since its only accessible to the admin user, it is further complicated by the following:

- No 2 factor auth for admin user
- Getting the database credential will often give an attacker direct access to the database, completely bypassing the authentication flow. If you have made the mistake of connecting to redash via HTTP, your session cookie would be enough for this attack.

Add the fact that there is no brute force protection currently (#51), it becomes a really nice target.

Access Credentials should never be returned by any API. However, I am not sure what (if any) options could be securely returned via the API. Does the database name reveal anything? What if there is no password? I'd very much prefer if all you could do was see the name/type of the data source and delete it.

arikfr:
Currently when you ask for data source options as an admin, it was sending back the full data source configuration, including passwords. In case you have multiple admins in your system or you're using re:dash without HTTPS (not recommended), it might be a security concern. 

This pull request changes this: it will send a placeholder instead of the password (or any other field defined as "secret").

Closes #554.

### DIFF
--- a/rd_ui/app/views/data_sources/form.html
+++ b/rd_ui/app/views/data_sources/form.html
@@ -12,7 +12,7 @@
         <input name="input" type="{{input.type}}" class="form-control" ng-model="dataSource.options[name]" ng-required="input.required"
                 ng-if="input.type !== 'file'" accesskey="tab">
 
-        <input name="input" type="file" class="form-control" ng-model="files[name]" ng-required="input.required"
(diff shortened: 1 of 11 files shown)


arikfr:
@olgakogan WDYT?

PR #610, opened 2015-10-13
Feature: dimension support for charts

tjwudi:
So just as discussed with @arikfr via email ...

Dimensions is a set of features that data has. For example, let's say we have a query pulling UV of yelp.com from database.

```
SELECT time, COUNT(*) FROM uv_records
GROUP BY time
```

Often times the data should have many other properties. For example, for a single UV, let's say we magically know the user's gender. So now we have a query like this.

```
SELECT time, gender, COUNT(*) FROM uv_records
GROUP BY time, gender
```

The result cannot be shown as a chart in re:dash since, for every single date, we have multiple rows (points) corresponding to it. So the idea of dimension is here to help. A dimension is a certain property of data that can re-group current data into several groups. In our data, we can have two values (M, F) for gender, so if we use gender as a dimension, then we can have two groups of data. 

Here is a screenshot of how dimensions are set (not from the UV example though).

Once you have dimensions set up, you can view your data from different dimensions. So here is an example of the chart without applying any dimension.

image

Here we have two dimensions, has_canonical and is_indexable. Because we don't have any dimension enabled yet, the chart will still contain one series.

Now here is the fun part, we can turn on one of the dimensions and see the data being split into multiple series.

image

BTW I also added **.editorconfig** for the sake of making code more consistent. If you don't think it's a good idea to have it here, feel free to tell me so I can pull it out.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+# For more information about the properties used in
+# this file, please see the EditorConfig documentation:
+# http://editorconfig.org/
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[{.travis.yml,package.json}]
+# The indent size used in the `package.json` file cannot be changed
+# https://github.com/npm/npm/pull/3180#issuecomment-16336516
+indent_size = 2
+indent_style = space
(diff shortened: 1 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,22 @@
+# For more information about the properties used in
+# this file, please see the EditorConfig documentation:
+# http://editorconfig.org/
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
```
arikfr: Actually we use indent size of 4 for Python, but 2 for JS.

PR #617, opened 2015-10-20
Improve timezone handling:

### LINKED ISSUE #411: Add time zone support
Right now re:dash assumes that all timestamps are in the local time zone. However, Redshift (and many other databases, depending on the configuration) default to UTC, and thus re:dash parses the data incorrectly. What's confusing is that in Highcharts also formats the data in local time, leading to this awkward display:

image

Notice how the data is highlighted at the beginning of April 23 (which hasn't happened yet in GMT), but the date is showing April 22 4:00 pm (which hasn't happened yet in Pacific Standard Time).

I temporarily worked around this by patching ng_highchart.js:

```diff
         if (moment.isMoment(this.x)) {
-          var s = '<b>' + this.x.toDate().toLocaleString() + '</b>',
+          var s = '<b>' + this.x.toDate().toISOString() + '</b>',
               pointsCount = this.points.length;
```
resources.js:

```diff
             if (angular.isNumber(v)) {
               columnTypes[k] = 'float';
             } else if (_.isString(v) && v.match(/^\d{4}-\d{2}-\d{2}T/)) {
-              row[k] = moment(v);
+              row[k] = moment.utc(v);
               columnTypes[k] = 'datetime';
             } else if (_.isString(v) && v.match(/^\d{4}-\d{2}-\d{2}/)) {
-              row[k] = moment(v);
+              row[k] = moment.utc(v);
               columnTypes[k] = 'date';
             } else if (typeof(v) == 'object' && v !== null) {
               row[k] = JSON.stringify(v);
```

I'd imagine there will be cases where some users have their database configured to output a specific standard time and others will use UTC. What's the best way to support this feature? Options:

1. Store the UTC offset in the `query_result` (so moment will parse the right value)
2. Allow configuration option in re:dash for time zone (e.g. UTC, America/Los_Angeles) 

What do you think, @arikfr?

arikfr:
1. Load all date/datetime values with `moment.utc()` instead of `moment()`, which was assuming user's local timezone when time zone information wasn't available. 
2. Don't use `toLocaleString` to format strings (which was converting them to current timezone as well). Due to the need to allow showing dates with different formatting (DD/MM vs. MM/DD), there is a new setting to define the date format.

Ideally we should add a setting to the user's profile to allow him to select his default timezone and date/time format, which we can than use when showing timestamps.

Closes #411.

### DIFF
--- a/rd_ui/app/index.html
+++ b/rd_ui/app/index.html
@@ -200,7 +200,7 @@
 
 <script>
     // TODO: move currentUser & features to be an Angular service
-    var featureFlags = {{ features|safe }};
+    var clientConfig = {{ client_config|safe }};
     var currentUser = {{ user|safe }};
 
     currentUser.canEdit = function(object) {
--- a/rd_ui/app/scripts/app.js
+++ b/rd_ui/app/scripts/app.js
@@ -19,7 +19,7 @@ angular.module('redash', [
     'ngSanitize'
   ]).config(['$routeProvider', '$locationProvider', '$compileProvider', 'growlProvider', 'uiSelectConfig',
     function ($routeProvider, $locationProvider, $compileProvider, growlProvider, uiSelectConfig) {
-      if (featureFlags.clientSideMetrics) {
+      if (clientConfig.clientSideMetrics) {
         Bucky.setOptions({
           host: '/api/metrics'
         });
--- a/rd_ui/app/scripts/controllers/controllers.js
+++ b/rd_ui/app/scripts/controllers/controllers.js
@@ -3,7 +3,8 @@
     if (!value) {
       return "-";
     }
-    return value.toDate().toLocaleString();
+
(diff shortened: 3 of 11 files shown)


stanhu:
Thanks, @arikfr! This helps for those of us who configure the DB for UTC. However, if a DB is configured for a specific time zone, then the results will return in terms of that. I would think re:dash should use the adapter to determine this time zone before each DB session and store that in each query result.

arikfr:
@stanhu maybe instead of storing it in the query result, it should apply it to the date/time values it has in the results? (if there are any)

stanhu:
Good point. Normalizing to UTC internally in the query results is a good thing.

arikfr:
True. Although before adding more code to handle this, I would start with documenting that re:dash expects timestamps with either timezone data, or it assumes the data is UTC. And of course to implement #618.

PR #623, opened 2015-10-21
Feature: support MySQL over SSL

stanhu:
This PR allows an admin to use MySQL over SSL. Also fixes a bug where boolean values in the data sources tables would not be properly saved; a checkbox is now used to ensure the `true` or `false` values are properly sent.

image

Suggestion: the configuration schema should use an `OrderedDict` so that the options don't appear out of order as above.

### DIFF
--- a/redash/query_runner/mysql.py
+++ b/redash/query_runner/mysql.py
@@ -46,9 +46,25 @@ class Mysql(BaseQueryRunner):
                     'type': 'string',
                     'title': 'Database name'
                 },
-                "port": {
-                    "type": "number"
+                'port': {
+                    'type': 'number'
                 },
+                'use_ssl': {
+                    'type': 'boolean',
+                    'title': 'Use SSL'
+                },
+                'ssl_cacert': {
+                    'type': 'string',
+                    'title': 'CA certificate to verify peer against (SSL)'
+                },
+                'ssl_cert': {
+                    'type': 'string',
+                    'title': 'Client certificate file (SSL)'
+                },
+                'ssl_key': {
+                    'type': 'string',
+                    'title': 'Private key filename (SSL)'
+                }
             },
             'required': ['db'],
             'secret': ['passwd']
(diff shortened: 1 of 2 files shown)


redash/query_runner/mysql.py
```diff
@@ -46,9 +46,25 @@ def configuration_schema(cls):
...
+                'port': {
+                    'type': 'number'
                 },
+                'use_ssl': {
+                    'type': 'boolean',
+                    'title': 'Use SSL'
+                },
+                'ssl_cacert': {
+                    'type': 'string',
+                    'title': 'CA certificate to verify peer against (SSL)'
```
arikfr: I want to keep re:dash instances stateless as much as possible. This is why the BigQuery and Google Spreadsheets query runners store the credentials file in the DB. I prefer we do the same here.

If MySQL client only accepts file paths for SSL certificates, we can write those to a temp file before connecting. If this becomes too much of an overhead, we can "cache" the file (write it always to the same path, so it's not needed to be written more than once).

stanhu: Ok, but does it really make sense to store private keys in the database? Do you encrypt the data in any way?

arikfr: No, but we don't encrypt it on the filesystem either. Do you think it's still safer?

stanhu: If we stored the keys in the DB, anyone with redash DB access would have access to private, unencrypted keys stored in the DB, which may not be in line with an organization's security policies. Having the keys stored on the filesystem, which may be encrypted at rest, at least gives a system admin the ability to restrict access using filesystem permissions.

arikfr: :+1:

redash/query_runner/mysql.py
```diff
@@ -46,9 +46,25 @@ def configuration_schema(cls):
...
+                },
+                'ssl_cert': {
+                    'type': 'string',
+                    'title': 'Client certificate file (SSL)'
+                },
+                'ssl_key': {
+                    'type': 'string',
+                    'title': 'Private key filename (SSL)'
+                }
             },
```
arikfr: Let's just normalize the titles to:

ssl_cacert: "Path to CA certificate file to verify peer against (SSL)"
ssl_cert: "Path to client certificate file (SSL)"
ssl_key: "Path to private key file (SSL)"

stanhu: Done.

arikfr:
Merged. Thanks!

PR #625, opened 2015-10-23
Feature: Oracle query runner

jfyne:
Here is an Oracle query runner.

I had to put the `cx_Oracle` requirement in a separate file as it relies on some external packages being installed before pip will run. If it was in the `requirements_all_ds.txt` it would stop complete installation.

We have had it running on our redash for a few days and everything appears to be OK.

Please review! It's difficult to test without a working Oracle DB, but I imagine there aren't many people who use Oracle anyway

### DIFF
--- a/docs/datasources.rst
+++ b/docs/datasources.rst
@@ -201,3 +201,18 @@ Vertica
 -  **Additional requirements**:
 
    - ``vertica-python`` python package
+
+Oracle
+------
+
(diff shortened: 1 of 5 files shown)


arikfr:
Merged. Thanks!

And apologies for the delay.

jfyne:
No worries! Thanks for redash!

Issue #628, opened 2015-10-26
docker re:dash's database init should not rely on the tool used to manage the re:dash's container


gissehel:
The philosophy behind "not putting postgresql into containers", is to let you use any postgres server (either one dedicated to re:dash within a docker-compose, or one already used for other databases on your infrastructure).

The fact that, in order to init the database, you HAVE to use docker-compose doesn't provide that flexibility having postgres outside of the container should bring you.

gissehel:
I'll try to make something else based on the current image.

The best way would be to have an init script into the redash image that can be called the first time instead of supervisor.

The problem:
- It would require to have psql installed on the redash image, and it's not.

The possible solutions:
* Install psql on the image
  * **pro**: Simple only 1 image
  * **con**: Install psql while not needed except for once for init
* Use an "init redash" image, that derive from the redash image and that contain psql.
  * **pro**: The init database use a container that launched with barely the same configuration that the redash image
  * **con**: Need one more image
* Don't use psql and use python script to execute basic queries
  * **pro**: Simple only 1 image, no big useless installation, just small python script
  * **con**: May be a bit more complex to write (have no idea what modules are installed)

gissehel:
I'll try to make a POC for the 3rd or the 2nd solution.

arikfr:
You have a very good point. I like the 1st and 3rd solution. The 3rd should be quite easy to implement considering that Python used for the largest chunk of that script (creating tables).

gissehel:
pull request #629 implement the 3rd solution.

arikfr:
Fixed in #1530.

PR #629, opened 2015-10-27
Simple db init with docker

gissehel:
I added to re:dash the following parameters:
- REDASH_DATABASE_NAME
- REDASH_DATABASE_HOST
- REDASH_DATABASE_PORT
- REDASH_DATABASE_USER
- REDASH_DATABASE_PASSWORD

They are to be used instead of REDASH_DATABASE_URL. REDASH_DATABASE_URL still works and override REDASH_DATABASE_* other keys, so it should not be used if REDASH_DATABASE_* are used.

So given the **re:dash main container** is used with those keys, if you start the **re:dash init container** with the same image, with `/init.sh` as command and with the following env var set, it will create the database you need, the role (user), grant access to the new role, create tables, and populate tables. 

The new init env var:
- REDASH_DATABASE_USER_ADMIN
- REDASH_DATABASE_PASSWORD_ADMIN
- REDASH_DEFAULT_ADMIN_NAME
- REDASH_DEFAULT_ADMIN_LOGIN
- REDASH_DEFAULT_ADMIN_PASSWORD
(DEFAULT_ADMIN is the properties used to create the first user in re:dash database)

typical usage would be:
- REDASH_DATABASE_NAME=redash
- REDASH_DATABASE_HOST=postgres
- REDASH_DATABASE_PORT=5432
- REDASH_DATABASE_USER=redash
- REDASH_DATABASE_PASSWORD=redash_password
- REDASH_DATABASE_USER_ADMIN=postgres
- REDASH_DATABASE_PASSWORD_ADMIN=********
- REDASH_DEFAULT_ADMIN_NAME=Admin
- REDASH_DEFAULT_ADMIN_LOGIN=admin
- REDASH_DEFAULT_ADMIN_PASSWORD=admin

The init script will do the following:
- Create the database if not already exists
- Create the role if not already exists
- Grant right on database for role if either the database or the role was created
- Check if table "dashboards" is present to call or not the tables creations and add the admin user
- Create the reader role
- Grant rights for the reader role
- Check if re:dash's metadata is present in data_sources in order to insert the field if not present

The "init" container "best practices":
- The container is started with exact same parameters than the re:dash container except:
  - is interactive
  - has more env var (cf: before)
  - doesn't use REDASH_DATABASE_URL
  - force command "/init.sh"


gissehel:
Should close #628

arikfr:
1. Why do we need the extra env variables instead of the DB url?
2. Maybe we should add those db init commands to the `manage.py` commands? This way you don't need to deal with CLI parsing.

And sorry for the delay in reviewing this...

gissehel:
### 1. About the REDASH_DATABASE_* variables:

This is in deed the controversial part of my commit. 

I need access to both superuser and redashuser on the database. I didn't like the idea to repeat hostname/database name, and to let them be different. But if you don't like the idea, I can change that to REDASH_DATABASE_URL and REDASH_DATABASE_ADMIN_URL.

### 2. About manage.py

I'm not used to flask development and conventions. I'll take a look at it. Doesn't seems complex, but I might end up violating flask best practices...

gissehel:
By the way, I put the files in the docker folder, but there is nothing related to docker in those files, they might be used to init database on other environments too.

tclass:
would be really nice if we could merge this and get a proper docker(-compose) setup. I got it working for now, but it was really a PITA to get started.

arikfr:
@gissehel was planning to make more changes, hence I didn't merge. But I see that it's out of sync with latest master. Closing this for now. 

@gissehel if you still plan to update it, please reopen.

PR #631, opened 2015-10-30
Feature: alert rearm setting which allows periodic resending of alert messages.

nielsdraaisma:
This pull request adds a setting which matches against a rearm timer after query execution, this allows users to receive daily emails for alerts that are trigger over a longer period of time.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -613,6 +613,7 @@ class Alert(ModelTimestampsMixin, BaseModel):
     options = JSONField()
     state = peewee.CharField(default=UNKNOWN_STATE)
     last_triggered_at = DateTimeTZField(null=True)
+    rearm = peewee.IntegerField(null=True)
 
     class Meta:
         db_table = 'alerts'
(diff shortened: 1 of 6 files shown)


redash/tasks.py
```diff
@@ -240,10 +241,13 @@ def check_alerts_for_query(self, query_id):
...
         new_state = alert.evaluate()
-        if new_state != alert.state:
+        passed_rearm_threshold = False
+        if alert.rearm and alert.last_triggered_at:
+            passed_rearm_threshold = alert.last_triggered_at + datetime.timedelta(seconds=alert.rearm) < datetime.datetime.now().replace(tzinfo=alert.last_triggered_at.tzinfo)
+        if new_state != alert.state or (alert.state == models.Alert.TRIGGERED_STATE and passed_rearm_threshold ):
             logger.info("Alert %d new state: %s", alert.id, new_state)
             old_state = alert.state
-            alert.update_instance(state=new_state)
+            alert.update_instance(state=new_state, last_triggered_at=datetime.datetime.now())
```
arikfr: I think that if you use the `utcnow` function from `redash.utils` module here and on line 246, you won't need to replace the timezone.

redash/models.py
```diff
@@ -613,6 +613,7 @@ class Alert(ModelTimestampsMixin, BaseModel):
     options = JSONField()
     state = peewee.CharField(default=UNKNOWN_STATE)
     last_triggered_at = DateTimeTZField(null=True)
+    rearm = peewee.IntegerField(null=True)
```
arikfr: Maybe instead of allowing nulls, just set the default to `False`?

arikfr: Oops, ignore that - it's not a boolean (maybe I shouldn't do code reviews at 11PM...).

redash/tasks.py
arikfr: @nielsdraaisma can you make this change?

blu35ky:
@arikfr thanks for the PR, sorry I just noticed it. @nielsdraaisma is busy with his new role.

arikfr:
@blu35ky thanks! do you mind also rebasing it w/ latest master, so I can easily merge it?

blu35ky:
@arikfr np, done. :+1:

arikfr:
Merged now. Apologies that it took so long.

PR #636, opened 2015-11-03
Limit access to query using access groups

tjwudi:
I've created a minimal demo of Re:dash Access Group Model, in respond to our previous concern on re:dash security issues. The idea of Access Group Model works like this: 

1. every user belongs to one or more groups (set groups_a)
2. every query has its one or more access groups  (set groups_b)
3. an user can access a query iff groups_a ∩ groups_b != empty_set
4. only admins can manage user's groups
5. each query defaults to have access group ['default']
6. each user defaults to have group ['default']
7. only admins can manage groups in admin dashboard (host-of-redash.com/admin)
8. every group has its permissions (among create_dashboard, create_query, edit_dashboard, edit_query, view_query, view_source, execute_query, list_users and admin)

Screenshots:

(1) Editing user groups
image

(2) Editing query access groups
768bc09a-8234-11e5-91a6-607bc96b5631

(3) Managing Groups

image


redash/handlers/groups.py
```diff
@@ -0,0 +1,17 @@
...
+
+from redash.handlers.base import BaseResource
+from redash.wsgi import api
+from flask import make_response
+from redash import utils
+from redash.models import Group
+
+class GroupsAPI(BaseResource):
+    def get(self):
+        return self.make_json_response(Group.select())
```
arikfr: Why not just return the groups array? flask-restful takes care of serialization. 

Also, please return the groups as an array, instead of a property in an object to remain consistent with the other APIs.

redash/models.py
```diff
@@ -571,6 +581,9 @@ def pre_save(self, created):
         self.query_hash = utils.gen_query_hash(self.query)
         self._set_api_key()
 
+        if not self.access_groups:
+            self.access_groups = self.DEFAULT_ACCESS_GROUPS
```
arikfr: Why this is needed? I think the `default` value setting takes care of it.

redash/models.py
```diff
@@ -548,11 +557,12 @@ def recent(cls, user_id=None, limit=20):
...
             where(Event.object_type == 'query'). \
             where(cls.is_archived == False).\
+            where(cls.access_groups.contains_any(*current_user.groups)).\
             group_by(Event.object_id, Query.id).\
             order_by(peewee.SQL("count(0) desc"))
 
-        if user_id:
-            query = query.where(Event.user == user_id)
+        if limit_to_current_user:
+            query = query.where(Event.user == current_user.id)
```
arikfr: `current_user` is a global that's only available in the context of a web request. Coupling the model layer with the web one will limit the usefulness of it.

If a user is needed, just pass it to this method (instead of passing `user_id` we can pass the user object).

redash/handlers/queries.py
```diff
@@ -50,7 +50,8 @@ def get(self):
 class QueryRecentAPI(BaseResource):
     @require_permission('view_query')
     def get(self):
-        recent = [d.to_dict() for d in models.Query.recent(current_user.id)]
+        recent = [d.to_dict() for d in models.Query.recent(
+                    limit_to_current_user=True)]
```
arikfr: While here you're limiting to the current user's groups, in the lines below we retrieve filler recents (in case the user doesn't have enough) where you don't apply this limit. You need to filter by groups there too...

migrations/0014_add_access_groups_to_queries.py
```diff
@@ -0,0 +1,17 @@
...
+
+if __name__ == '__main__':
+    db.connect_db()
+    migrator = PostgresqlMigrator(db.database)
+
+    with db.database.transaction():
+        migrate(
+            migrator.add_column('queries', 'access_groups', models.Query.access_groups),
+        )
+        models.Query.update(access_groups=['default']).execute()
```
arikfr: I think the default value setting already sets this.

rd_ui/bower.json
```diff
@@ -33,7 +33,8 @@
     "angular-ui-select": "0.8.2",
     "angular-bootstrap-show-errors": "~2.3.0",
     "angular-sanitize": "1.2.18",
-    "d3": "3.5.6"
+    "d3": "3.5.6",
+    "ng-tags-input": "2.1.1"
```
arikfr: We're already using `angular-ui-select` which I believe has comparable features, see this example: http://plnkr.co/edit/juqoNOt1z1Gb349XabQ2?p=preview

tjwudi: Didn't notice, will update :+1:

redash/models.py
tjwudi: I don't know why it just doesn't work. Probably issue with peewee.

redash/handlers/queries.py
tjwudi: In `Query.recent`, I always constraint query with `where(cls.access_groups.contains_any(*current_user.groups))`, no matter what value `limit_to_current_user`. So it will still be limited to those visible to current user.

migrations/0014_add_access_groups_to_queries.py
tjwudi: L15 is for adding value to existing rows. Shouldn't I need it?

tjwudi: So the thing is, if you specified `default` value, and then run `\d+ queries` in Postgres, you will see it didn't actually apply default value for `ArrayField`. I suspect this is a bug of Playhouse extensions.

tjwudi:
Updated!

rd_ui/app/views/query.html
```diff
@@ -143,6 +143,19 @@
...
+            <p>
+                <i class="fa fa-lock"></i>
+                <span class="text-muted">Access Groups</span>
+                <div>
+                  <ui-select multiple ng-model="query.access_groups">
+                      <ui-select-match placeholder="Select access groups">{{$item}}</ui-select-match>
+                      <ui-select-choices repeat="group in availableGroups | filter:$select.search">
+                        {{group}}
+                      </ui-select-choices>
+                  </ui-select>
```
arikfr: It makes sense to turn this into a directive, and then reuse in the new/edit user forms.

tjwudi: Sure! Though there is an issue with ui-select (I've documented in code), it can be done by applying a little bit hack :smile:

tjwudi:
OK, now it comes with `tags-input` directive and used in new user, show(edit) user forms, and also query access groups form.

tjwudi:
I am wondering whether we could provide certain level of abstraction to avoid writing `.where(cls.access_groups.contains_any(*current_user.groups))` everywhere you query `Query`... Just like [scopes in Rails](http://guides.rubyonrails.org/active_record_querying.html#scopes).

So the problem here is you cannot write a `where` clause before `select`, which makes taking `where` clause out is impossible.

Seems like **peewee** does not have this capability. Any thoughts? @arikfr

tjwudi:
ping @arikfr ?

arikfr:
Maybe it's possible to achieve something similar using the ability to chain
multiple where clauses and that the query doesn't execute until you access
the data. So, you'll have a method like "filter_by_groups" that takes a
select query and groups to filter by.

Problem is that sometimes the order of how you apply the where clauses has
meaning so we might not be able to apply it at the end. But worth a try.

Another thought I had: while you restrict access to the query, you didn't
restrict access to the data source. Meaning that anyone can write a similar
query and run it. This is why initially I thought of managing access with
data sources. Have you thought of that?

On Friday, 6 November 2015, John Wu <notifications@github.com> wrote:

> I am wondering whether we could provide certain level of abstraction to
> avoid writing .where(cls.access_groups.contains_any(*current_user.groups))
> everywhere you query Query... Just like scopes in Rails
> <http://guides.rubyonrails.org/active_record_querying.html#scopes>.
>
> So the problem here is you cannot write a where clause before select,
> which makes taking where clause out is impossible.
>
> Seems like *peewee* does not have this capability. Any thoughts? @arikfr
> <https://github.com/arikfr>
>
> —
> <https://github.com/EverythingMe/redash/pull/636#issuecomment-154234058>.
>


-- 
-- 
Arik Fraimovich

tjwudi:
Hah, haven't thought about it before. Are you implying that rather than provide access restrictions to queries, we should provide it to data source?

tjwudi:
Well, we might still need access control over queries. Because one can view query results from query view. :confused:

arikfr:
Sometime ago I wrote how I thought access restriction via data sources
might work: https://gist.github.com/arikfr/8aa56aba9198980fd435

But maybe a combination of the above and what you done is needed. Because
in some cases you don't want to give access to query some datasource but
you do want to share some of the resulting queries...

alonho:
I also agree access control should be implemented in both levels. Some users should be able to invoke any query, others should be able to invoke pre-defined queries.
The same group mechanism should be applied to data sources.
I would even let users create data sources and optionally share them using groups. Imagine a single database where each redash user has a different database user with different permissions.

Perhaps we should split the data source management to a different ticket.

tjwudi:
@alonho yeah, I now provided access control both over queries and data sources.

rghose:
A really useful feature. Hope this gets merged soon :smile:

redash/models.py
arikfr: I just tested and the default value works for new objects and also in the migration. Can you verify that you have the correct version of peewee (2.4.7)?

redash/handlers/query_results.py
```diff
@@ -98,6 +99,11 @@ def get(self, query_id=None, query_result_id=None, filetype='json'):
         if query_result_id:
             query_result = models.QueryResult.get_by_id(query_result_id)
 
+        if not query:
+            query = models.Query.get(query_hash=query_result.query_hash)
+        if not (set(query.access_groups) & set(current_user.groups)):
```
arikfr: This prevents from running adhoc (non saved) queries... 
I suggest the following logic:
1. If there is query specified, check if the user has access to it.
2. If there is no query specified, check if the user has access to the data source.

redash/handlers/groups.py
```diff
@@ -0,0 +1,10 @@
+from redash.handlers.base import BaseResource
+from redash.wsgi import api
+from redash.models import Group
+
+
+class GroupsAPI(BaseResource):
+    def get(self):
+        return [group.to_dict() for group in Group.select()]
```
arikfr: If the user isn't an admin, it should only return the list of groups the user belongs to.

rd_ui/app/views/users/show.html
```diff
@@ -36,6 +36,10 @@
                         <input name="email" type="email" class="form-control" ng-model="user.email" required/>
                         <input-errors errors="userSettingsForm.email.$error"/>
                     </div>
+                    <div class="form-group required" ng-if="showSettings">
```
arikfr: It should check if the user is admin (`ng-if="currentUser.hasPermission('admin')"`).
Also you can remove the admin checkbox.

rd_ui/app/scripts/controllers/query_view.js
```diff
@@ -270,9 +278,19 @@
       }
       $scope.selectedTab = hash || DEFAULT_TAB;
     });
+
+    /**
+     * Access groups related
+     */
+    $scope.$watch('query.access_groups', function (newValue, oldValue) {
```
arikfr: Watch is being fired on the first check even if there was no change, so you need to make sure that `newValue !== oldValue`, otherwise it will try to save a new query.

arikfr: Also, you need to update the `Query.newQuery` function to set the `access_groups` value (should probably first copy the value of the data source).

arikfr:
Aside from the comments I made, there is another use case to handle: what happens when you change data source's access groups? It should probably update all the queries that were created with this data source. But what if you changed a query's groups (added some)?

This leads me to think that:
1. Every query should be accessible to the groups the data source is accessible to.
2. You can only add *additional* groups to open the query to, on top of the ones it inherits from the data source. 

This will reduce confusion, and also cover the case I described in the first paragraph.

toyama0919:
Awesome feature :blush:

sromku:
+1

vishesh92:
+1

arikfr:
I've implemented most of this pull request features (and some others) in #724. 

Let's move the discussion there.

PR #639, opened 2015-11-07
Feature: permission to schedule query

alexdebrie:
We had a problem on Redshift where people were scheduling really expensive queries and slowing down our whole cluster. This allows us to whitelist particular users with the ability to schedule queries.


arikfr:
Thanks!

PR #642, opened 2015-11-09
Docs: make migrating Vagrant box command a one-liner

tjwudi:


docs/dev/vagrant.rst
```diff
@@ -30,20 +30,7 @@ To get started with this box:
...
-        PYTHONPATH=. bin/run python migrations/0006_queries_last_edit_by.py
-        PYTHONPATH=. bin/run python migrations/0007_add_schedule_to_queries.py
-        PYTHONPATH=. bin/run python migrations/0008_make_ds_name_unique.py
-        PYTHONPATH=. bin/run python migrations/0009_add_api_key_to_user.py
-        PYTHONPATH=. bin/run python migrations/0010_create_alerts.py
-        PYTHONPATH=. bin/run python migrations/0010_allow_deleting_datasources.py
-        PYTHONPATH=. bin/run python migrations/0011_migrate_bigquery_to_json.py
-        PYTHONPATH=. bin/run python migrations/0012_add_list_users_permission.py
-        PYTHONPATH=. bin/run python migrations/0013_update_counter_options.py
+        export PYTHONPATH=. && find migrations/ -type f  | grep 00 --null | xargs -I file bin/run python file
```
arikfr: Is it guaranteed that `find` will return files in their order? (01 before 02, etc)

tjwudi: for those with number prefix, yes :)

arikfr:
Merged. Thanks!

tjwudi:
Np! FYI, I don't think it can be a sustainable solution to maintain migrations order manually by adding numbered prefixes. I think at some point we can use some other migration tools that persist migration histories in database.

arikfr:
Totally agree.

PR #644, opened 2015-11-10
Feature: send alert notifications to HipChat or web hook

toyama0919:
hello.

I want alert to hipchat.

### Need to set the following environment variables
* REDASH_HIPCHAT_API_TOKEN
* REDASH_HIPCHAT_ROOM_ID

redash

### DIFF
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -1,6 +1,7 @@
 import logging
 import urlparse
 import redis
+import hipchat
 from statsd import StatsClient
 from flask_mail import Mail
 
@@ -36,6 +37,7 @@ redis_connection = create_redis_connection()
 mail = Mail()
 mail.init_mail(settings.all_settings())
 statsd_client = StatsClient(host=settings.STATSD_HOST, port=settings.STATSD_PORT, prefix=settings.STATSD_PREFIX)
+hipchat_client = hipchat.HipChat(token=settings.HIPCHAT_API_TOKEN)
 
 import_query_runners(settings.QUERY_RUNNERS)
 
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -1,7 +1,6 @@
 import logging
 import urlparse
 import redis
-import hipchat
 from statsd import StatsClient
 from flask_mail import Mail
 
@@ -37,7 +36,6 @@ redis_connection = create_redis_connection()
 mail = Mail()
 mail.init_mail(settings.all_settings())
 statsd_client = StatsClient(host=settings.STATSD_HOST, port=settings.STATSD_PORT, prefix=settings.STATSD_PREFIX)
-hipchat_client = hipchat.HipChat(token=settings.HIPCHAT_API_TOKEN)
 
 import_query_runners(settings.QUERY_RUNNERS)
 
--- a/redash/tasks.py
+++ b/redash/tasks.py
@@ -7,7 +7,7 @@ import redis
 import hipchat
 import requests
 import json
-from bson import json_util
+from redash.utils import json_dumps
 from requests.auth import HTTPBasicAuth
 from celery import Task
 from celery.result import AsyncResult
@@ -361,7 +361,7 @@ def notify_webhook(alert, query, html, new_state):
         }
         headers = {'Content-Type': 'application/json'}
         auth = HTTPBasicAuth(settings.WEBHOOK_USERNAME, settings.WEBHOOK_PASSWORD) if settings.WEBHOOK_USERNAME else None
-        resp = requests.post(settings.WEBHOOK_ENDPOINT, data=json.dumps(data, default=json_util.default), auth=auth, headers=headers)
+        resp = requests.post(settings.WEBHOOK_ENDPOINT, data=json_dumps(data), auth=auth, headers=headers)
         if resp.status_code != 200:
             logger.error("webhook send ERROR. status_code => {status}".format(status=resp.status_code))
     except:
--- a/redash/tasks.py
+++ b/redash/tasks.py
@@ -6,6 +6,8 @@ from flask.ext.mail import Message
 import redis
 import hipchat
 import requests
+import json
+from bson import json_util
 from requests.auth import HTTPBasicAuth
 from celery import Task
 from celery.result import AsyncResult
@@ -353,15 +355,13 @@ def notify_mail(alert, html, new_state, app):
 def notify_webhook(alert, query, html, new_state):
     try:
         data = {
-            'new_state':new_state.upper(),
-            'alert_name':alert.name,
-            'alert_id':alert.id,
-            'query_id':query.id,
-            'url_base':settings.HOST
+            'event': 'alert_state_change',
+            'alert': alert.to_dict(full=False),
+            'url_base': settings.HOST
         }
-        headers = {'Content-type': 'application/x-www-form-urlencoded'}
+        headers = {'Content-Type': 'application/json'}
         auth = HTTPBasicAuth(settings.WEBHOOK_USERNAME, settings.WEBHOOK_PASSWORD) if settings.WEBHOOK_USERNAME else None
-        resp = requests.post(settings.WEBHOOK_ENDPOINT, data=data, auth=auth, headers=headers)
+        resp = requests.post(settings.WEBHOOK_ENDPOINT, data=json.dumps(data, default=json_util.default), auth=auth, headers=headers)
         if resp.status_code != 200:
             logger.error("webhook send ERROR. status_code => {status}".format(status=resp.status_code))
     except:
--- a/redash/tasks.py
+++ b/redash/tasks.py
@@ -4,10 +4,11 @@ import signal
 import traceback
 from flask.ext.mail import Message
 import redis
+import hipchat
 from celery import Task
 from celery.result import AsyncResult
 from celery.utils.log import get_task_logger
-from redash import redis_connection, models, statsd_client, settings, utils, mail, hipchat_client
+from redash import redis_connection, models, statsd_client, settings, utils, mail
 from redash.utils import gen_query_hash
 from redash.worker import celery
 from redash.query_runner import get_query_runner, InterruptException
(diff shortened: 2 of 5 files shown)


tjwudi:
Awesome, I really like this. It would probably better to have something like webhook for other apps to use in the future. Since everybody uses different toolsets and we might want to post notifications not only to HipChat but also IRC, email and so on.

redash/__init__.py
```diff
@@ -36,5 +37,6 @@ def create_redis_connection():
 mail = Mail()
 mail.init_mail(settings.all_settings())
 statsd_client = StatsClient(host=settings.STATSD_HOST, port=settings.STATSD_PORT, prefix=settings.STATSD_PREFIX)
+hipchat_client = hipchat.HipChat(token=settings.HIPCHAT_API_TOKEN)
```
arikfr: Let's move the hipchat client definition to where it's used.

redash/tasks.py
```diff
@@ -256,12 +257,24 @@ def check_alerts_for_query(self, query_id):
...
+            try:
+                with app.app_context():
+                    message = Message(recipients=recipients,
+                                      subject="[{1}] {0}".format(alert.name, new_state.upper()),
+                                      html=html)
+
+                    mail.send(message)
+            except:
+                tb = traceback.format_exc()
+                logger.error(tb)
```
arikfr: Why not just use `logger.exception`?

arikfr:
@tjwudi I agree. My plan was to add webhooks first, and then maybe a plugin model. I just haven't figured out yet how to allow plugins to supply their own UI, because ultimately the user might want to set a different HipChat (or Slack or whatever) room for different alerts.

tjwudi:
@arikfr or if that is complex, simply send a POST request to series of subscribers instead.

arikfr:
@tjwudi that's a webhook, no?

toyama0919:
> Awesome, I really like this. It would probably better to have something like webhook for other apps to use in the future. Since everybody uses different toolsets and we might want to post notifications not only to HipChat but also IRC, email and so on.

Nice Idea!
I will rewrite code for webhook. OK? 

@arikfr @tjwudi

arikfr:
@toyama0919 sure thing. But you can keep the HipChat integration. I suggest that the `check_alerts_for_query` will call different methods - one for mail, one for hipchat and one for webhook.

Later we can refactor it to be a plugin model, so we will have plugin for email, hipchat and in the future others (like Slack).

toyama-h:
@akariv 
> sure thing. But you can keep the HipChat integration. I suggest that the check_alerts_for_query will call different methods - one for mail, one for hipchat and one for webhook.

Ah. OK. Thank you for reply.
I keep the HipChat integration.

toyama0919:
I assume these parameters for webhook.
Is there it elsewhere?

```
{"alert_id"=>"69", "alert_name"=>"fuga: count(*) greater than 5476", "query_id"=>"3", "new_state"=>"TRIGGERED", "url_base"=>"http://localhost:9001"}
```

toyama0919:
@akariv @tjwudi 
added webhook.
We can webhook with basic auth.

```
export REDASH_WEBHOOK_ENDPOINT="http://10.0.2.2:4567/hi"
export REDASH_WEBHOOK_USERNAME="hoge"
export REDASH_WEBHOOK_PASSWORD="fuga"
```

※REDASH_WEBHOOK_USERNAME AND REDASH_WEBHOOK_PASSWORD is optional.

redash/tasks.py
```diff
@@ -317,3 +318,47 @@ def execute_query(self, query, data_source_id, metadata):
...
+        with app.app_context():
+            message = Message(recipients=recipients,
+                              subject="[{1}] {0}".format(alert.name, new_state.upper()),
+                              html=html)
+            mail.send(message)
+    except:
+        logger.exception("mail send ERROR.")
+
+def notify_webhook(alert, query, html, new_state):
+    http_client = httplib2.Http(timeout=30)
```
arikfr: [`requests`](http://docs.python-requests.org/en/latest/) is already in the project dependencies, please use it.

redash/tasks.py
```diff
@@ -317,3 +318,47 @@ def execute_query(self, query, data_source_id, metadata):
...
+    if settings.WEBHOOK_USERNAME:
+        http_client.add_credentials(settings.WEBHOOK_USERNAME, settings.WEBHOOK_PASSWORD)
+    try:
+        data = {
+            'new_state':new_state.upper(),
+            'alert_name':alert.name,
+            'alert_id':alert.id,
+            'query_id':query.id,
+            'url_base':settings.HOST
+        }
```
arikfr: I think it will be better to use `Alert`'s `to_dict` method and nest it, so data becomes:

```python
{
    'event': 'alert_state_change', # this to support other webhook calls in the future
    'alert': alert.to_dict(),
    'url_base': settings.HOST
}
```

What do you think?

toyama0919: this way is post-json?

```
headers = {'Content-Type': 'application/json'}
```

toyama0919: alert.to_dict , I feel that there is very much unnecessary information...

data example

```
{'state': 'triggered', 'options': {u'column': u'count(*)', u'value': 7, u'op': u'greater than'}, 'name': u'fuga: count(*) greater than 7', 'updated_at': datetime.datetime(2015, 11, 18, 1, 35, 37, 973331), 'query': {'latest_query_data_id': 108, 'schedule': None, 'is_archived': False, 'updated_at': datetime.datetime(2015, 11, 17, 4, 10, 40, 182108, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'user': {'auth_type': 'password', 'created_at': datetime.datetime(2015, 11, 4, 4, 43, 55, 314074, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'name': u'toyama0919', 'gravatar_url': 'https://www.gravatar.com/avatar/1933de2ab395fc2f64a20fae9f77029a?s=40', 'updated_at': datetime.datetime(2015, 11, 17, 1, 58, 3, 611870, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'id': 2, 'groups': [u'default'], 'email': u'toyama-h@ipros.jp'}, 'query': u'select count(*) from customer', 'id': 3, 'description': None, 'name': u'fuga', 'created_at': datetime.datetime(2015, 11, 17, 4, 10, 40, 182019, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'last_modified_by': {'auth_type': 'password', 'created_at': datetime.datetime(2015, 11, 4, 4, 43, 55, 314074, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'name': u'toyama0919', 'gravatar_url': 'https://www.gravatar.com/avatar/1933de2ab395fc2f64a20fae9f77029a?s=40', 'updated_at': datetime.datetime(2015, 11, 17, 1, 58, 3, 611870, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'id': 2, 'groups': [u'default'], 'email': u'toyama-h@ipros.jp'}, 'query_hash': u'dd8bb0018aea7d1892151fd9478bfc0e', 'api_key': u'42858af3cdbf80fe90e55d218a74f1f907d1a5d7', 'data_source_id': 2}, 'created_at': datetime.datetime(2015, 11, 18, 1, 35, 31, 868308, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'last_triggered_at': None, 'id': 103, 'user': {'auth_type': 'password', 'created_at': datetime.datetime(2015, 11, 4, 4, 43, 55, 314074, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'name': u'toyama0919', 'gravatar_url': 'https://www.gravatar.com/avatar/1933de2ab395fc2f64a20fae9f77029a?s=40', 'updated_at': datetime.datetime(2015, 11, 17, 1, 58, 3, 611870, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'id': 2, 'groups': [u'default'], 'email': u'toyama-h@ipros.jp'}}
```

arikfr: We can change `to_dict` not to embed user/query objects if not asked for:

```python
def to_dict(self, full=True):
    d = {
            'id': self.id,
            'name': self.name,
            'options': self.options,
            'state': self.state,
            'last_triggered_at': self.last_triggered_at,
            'updated_at': self.updated_at,
            'created_at': self.created_at
    }

    if full:
        d['query'] = self.query.to_dict()
        d['user'] = self.user.to_dict()
    else:
        d['query_id'] = self._data['query']
        d['user_id'] = self._data['user']

    return d
```

And then in your code call it with `full=False`.

toyama-h:
fixed.

post nested data, so 
```
{'Content-Type': 'application/json'}
```

toyama0919:
fixed.

post nested data, so 
```
{'Content-Type': 'application/json'}
```

redash/tasks.py
```diff
@@ -2,6 +2,11 @@
 import logging
 from flask.ext.mail import Message
 import redis
+import hipchat
+import requests
+import json
+from bson import json_util
```
arikfr: This will require users to install PyMongo, which isn't a requirement for re:dash. Please use `json_dumps` from `redash.utils` instead of `json.dumps` with `json_util`.

toyama0919: I think that the following one is better, do you think?
Should nest it?

```py
def notify_webhook(alert):
    try:
        data = alert.to_dict(full=False)
        data['event'] = 'alert_state_change'
        data['url_base'] = settings.HOST
        auth = HTTPBasicAuth(settings.WEBHOOK_USERNAME, settings.WEBHOOK_PASSWORD) if settings.WEBHOOK_USERNAME else None
        resp = requests.post(settings.WEBHOOK_ENDPOINT, data=data, auth=auth)
        if resp.status_code != 200:
            logger.error("webhook send ERROR. status_code => {status}".format(status=resp.status_code))
    except:
        logger.exception("webhook send ERROR.")

```

arikfr: I prefer to keep the structure as it was before (alert in a field of its own, instead of top level). Just don't use `bson`'s `json_util` but rather re:dash's utilities for dumping json.

toyama0919: Sorry later for reply.
OK!

Please wait a moment.

toyama0919:
I applied review from arikfr.

arikfr:
Great. It's ready for a merge. But you need to rebase with latest master, for me to be able to merge it.

toyama-h:
rebased.

arikfr:
Merged! Thank you for your contribution and patience with the code review process.

toyama-h:
:blush:

PR #645, opened 2015-11-11
Upgrade to Highcharts 4.1.9

stanhu:
Bumped version to support `panning` option.

Issue #273, opened 2014-08-28
Add open source alternative to HighCharts


arikfr:
Possible candidates:
1. http://c3js.org/examples.html

arikfr:
http://www.chartjs.org/

arikfr:
Site comparing all the JS graph libraries: http://www.jsgraphs.com/

shenyubao:
https://github.com/ecomfe/echarts

alonho:
c3js looks great (it's interactive which most aren't).
dimple has a good variety of features but isn't interactive.
chartjs doesn't have a lot of features and is practically a one-person project.
echarts has the best variety of the bunch.

Adding another option: nvd3 http://nvd3.org/examples/index.html 
I personally used it and it works great but isn't very rich in features.

We also need to look how well these options are maintained whether they have angular directives or we need to implement them.

alonho:
c3js has one maintainer that isn't very active.
echarts is well maintained but looks intimidating in terms of documentation. the angular binding isn't very popular (in terms of github stars)
nvd3 is maintained by several people and is active. the angular bindings area also popular and mature: http://krispo.github.io/angular-nvd3/#/

alonho:
The final two contenders are nvd3 and plotly. nvd3 has cool animations but plotly seems to have a lot of features and would be a better investment for the future.

Plotly was just open sourced so I've created the angular-plotly library in its [own repo](https://github.com/alonho/angular-plotly).

I'm considering several options: 
1. Merging #610 while I'm refactoring this area.
2. Using the ui-select widget for better multi select.

arikfr:
HighCharts are no more in v0.9.0 and replaced by Plotly.js.

PR #646, opened 2015-11-11
Support pan and zoom

stanhu:
I opened a separate PR for #645, but perhaps it can just be folded into this.

### DIFF
--- a/rd_ui/app/index.html
+++ b/rd_ui/app/index.html
@@ -142,7 +142,6 @@
 <script src="/bower_components/codemirror/mode/python/python.js"></script>
 <script src="/bower_components/codemirror/mode/javascript/javascript.js"></script>
 <script src="/bower_components/highcharts/highcharts.js"></script>
-<script src="/bower_components/highcharts/modules/exporting.js"></script>
 <script src="/bower_components/gridster/dist/jquery.gridster.js"></script>
 <script src="/bower_components/angular-growl/build/angular-growl.js"></script>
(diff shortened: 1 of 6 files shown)


tjwudi:
Well if we are using HighChart 4 why not just switch to HighStock?

I added date picker in #594. It was a really simple date picker with many hacky code. Actually HighStock supports date picker but because the version of HighChart we used were not compatible with any HighStock revision so I didn't use HighStock.

But if we are going to have date picker and pan&zoom support then it makes sense to switch to HighStock force instead. Since we can have many new nice features without hand crafting them. @arikfr Any thoughts?

arikfr:
@tjwudi actually, the plan is to replace HighCharts, because it's not open source (see #273). Maybe we will leave it as an optional plugin, but the core will have a different visualization option. Therefore I prefer not to invest much more into it by switching to HighStock.

arikfr:
@stanhu do you know if there were any backward incompatible changes between version 3 and 4?

stanhu:
I couldn't find any documentation about upgrading from Highcharts 3 to 4. The changes I made in this PR seem to make basic line graphs work fine for 4, but I may have missed something.

arikfr:
We decided to switch to a different charting library (nvd3 or plot.ly new open source lib), so I don't want to invest time in testing this, but also don't want to merge it without testing.

Therefore I'm closing this. I apologize for the time you wasted on this. :(

PR #647, opened 2015-11-11
Fix: bind Redis to localhost

runa:
Having it bound to the public addresses is a security problem.
See http://antirez.com/news/96

### DIFF
--- a/setup/files/redis.conf
+++ b/setup/files/redis.conf
@@ -62,7 +62,7 @@ tcp-backlog 511
 # Examples:
 #
 # bind 192.168.1.100 10.0.0.1
-# bind 127.0.0.1
+bind 127.0.0.1
 
 # Specify the path for the Unix socket that will be used to listen for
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks!

matangover:
I just found out my re:dash server was compromised multiple times last month because it used the previous (default) configuration which didn't include the 'bind' directive -- apparently some botnets (specifically XorDDOS) are going around taking advantage of this redis 'feature'. See e.g. [here](http://kevinchen.co/blog/postmortem-server-compromised/). I was not aware that leaving the redis port open to the world creates a security hole. With the fixed configuration by @runa (thanks!), it wouldn't be a problem for new installations, even if the redis port is left open.

Is there any way to alert current users of older re:dash versions (before v0.8.2.b1181) about this?

arikfr:
Unfortunately there is no way to really reach everyone, but I will mention this on the mailing list with the next update.

matangover:
Great. Thanks.

PR #653, opened 2015-11-14
Fix date range selector does not show data of last day when user timezone is not UTC

hakobera:
Related to #594 and #617.
In following conditions, re:dash does not show the data of last day.

- some timestamp value group by day
- user in non UTC timezone

Like this:

```
SELECT
  to_char(created_at, 'YYYY-MM-DD') as day,
  COUNT(*) AS cnt
FROM
  queries
GROUP BY
  day
ORDER BY
  day
```

I think this is caused `dateRange` is treated as local timezone, but `point.x` is treated as UTC.
Following screenshot is a result of `console.log(dateRange, point.x);` before https://github.com/EverythingMe/redash/blob/master/rd_ui/app/scripts/services/resources.js#L197

This pull request fix this issue, but I'm not sure that this change is OK for timezone handling in re:dash.
@arikfr What do you think?

### DIFF
--- a/rd_ui/app/scripts/visualizations/date_range_selector.js
+++ b/rd_ui/app/scripts/visualizations/date_range_selector.js
@@ -34,8 +34,8 @@
             scope.dateRangeHuman = oldDateRangeHuman;
             return;
           }
-          scope.dateRange.min = newDateRangeMin;
-          scope.dateRange.max = newDateRangeMax;
+          scope.dateRange.min = newDateRangeMin.startOf('day');
+          scope.dateRange.max = newDateRangeMax.endOf('day');
         }, true);
       }]
     }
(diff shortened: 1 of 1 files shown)


arikfr:
Are you sure it's due to time zone differences and not because of the time value of the max value? It looks like your fix actually does that: moves the the hour to be of end of day.

What you did looks ok to me, but I will make sure that we indeed create all time values in the same time zone as you mentioned.

Thanks.

hakobera:
@arikfr Thanks for review. I realized that this is not a issue of the time value of the max value, it's only related timezone difference issue. So I change my code to treat `dateRange` as UTC, it is as same as `point.x`. 

Is this OK for you?

arikfr:
Merged. Thanks!

PR #660, opened 2015-11-18
Fix: strings that start with a date wrongly treated as date fields

alexdebrie:
Tiny fix to avoid displaying a result that starts with a date string as only a date string.


arikfr:
Merged. Thanks!

PR #661, opened 2015-11-18
Fix cancelling queries for Redshift/Postgres

### LINKED ISSUE #598: Cancel doesnt actually cancel a query in redshift.
I was using redash and wrote a pretty bad query. Soon after I hit cancel and proceeded to run other queries but they wouldn't go through. I suspected redshift to be busy running my previous erroneous query so logged into the aws console and sure enough it was running. I had to terminate the query manually there.

alexdebrie:
Fixes #598.

I tested this on our stage machine and it worked. I've only actually implemented a handler for Postgres, but any query_runner can catch the `InterruptException` and handle it as desired.

I'm open to any and all comments. We don't have any other datasources hooked up, so not sure how this affects the other query runners.

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -1,5 +1,6 @@
 import logging
 import json
+import signal
 
 import jsonschema
 from jsonschema import ValidationError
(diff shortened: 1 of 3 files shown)


arikfr:
When I was testing this I noticed that it crashed my server, because it was trying to install the handler on non main thread. I think it has to do with PyCharm's debugger, but it made me realize that installing the handler every time we create the a query runner object is not necessary (and maybe even wrong).

I moved the `signal.signal` call to the `execute_query` function of `redash.tasks` and it works but prevents the issue I experienced (ideally we will call it once when the Celery worker starts, but didn't have the chance to look into it, and it felt good enough).

As for impact on other query runners: I've tested it only with MySQL, and it looks like it doesn't change the behavior, so looks like it's safe on that front. But queries don't get really cancelled... we will need to address it at some point, but once you move the `signal.signal` call it's good for merging.

alexdebrie:
Thanks for debugging and tips. I moved the signal handler as requested and tested on our machine. As a note, I left the `InterruptException` within `query_runners` as this is something other runners can import and catch as they see fit.

Let me know if you need any other changes.

arikfr:
Thanks! Merged.

PR #664, opened 2015-11-19
Feature: ability to embed visualizations in external sites

### LINKED ISSUE #320: Widget embedding
Allow embedding an redash widget iframe.
Should keep authentication in mind.

akariv:
1. Adding a 'public' bit to every query
2. Adding UI in the query editor for modifying it
3. Creating an embed page to show a specific visualization (by id)
4. Modifying the api-server for query-results to check the query for permissions as well
5. Show the embed code in the query editor

Closes #320.


arikfr:
It will make everything much simpler if you use query's API key for embed authorization, instead of adding a toggle on the query or having to change the permission model.

I suggest the following:
1. Change the embed URL to be: `/embed/<query_id>/visualizations/<visualization_id>`.
2. Change the embed handler to return the query result as part of the visualization dict or change the UI code to forward the API key when requesting the query result.

WDYT?

akariv:
@arikfr I totally agree.

I'll change it so it's as you suggested.

Just one thing - this means that this method won't work in case you're using hmac authentication, correct?

arikfr:
Correct.

On Sunday, 22 November 2015, Adam Kariv <notifications@github.com> wrote:

> @arikfr <https://github.com/arikfr> I totally agree.
>
> I'll change it so it's as you suggested.
>
> Just one thing - this means that this method won't work in case you're
> using hmac authentication, correct?
>
> —
> <https://github.com/getredash/redash/pull/664#issuecomment-158795465>.
>


-- 
-- 
Arik Fraimovich

akariv:
@arikfr please re-review.

redash/handlers/embed.py
```diff
@@ -0,0 +1,47 @@
...
+
+from redash import models, settings
+from redash.wsgi import app
+from redash.utils import json_dumps
+
+@app.route('/embed/query/<query_id>/visualization/<visualization_id>', methods=['GET'])
+@login_required
+def embed(query_id, visualization_id):
+
+    v = models.Visualization.get(models.Visualization.id == visualization_id)
```
arikfr: I think the following is a bit better:
```
query = models.Query.get_by_id(query_id)
vis = query.visualizations.where(models.Visualization.id == visualization_id).first()
```
If `visualization_id` doesn't belong to query, it will return `None`.

redash/handlers/embed.py
```diff
@@ -0,0 +1,47 @@
...
+    qr = {}
+    if v:
+        q = v.query
+        if q.id == int(query_id):
+            v = v.to_dict()
+            query_result_id = q._data['latest_query_data']
+            if query_result_id:
```
arikfr: `q.latest_query_data` will return `None` if there are no results, so you can assign it directly to `qr` instead of doing this check.

Also it might be better to do the check if there are results in the view, and then show a proper message.

redash/handlers/embed.py
```diff
@@ -0,0 +1,47 @@
...
+                qr = qr.to_dict()
+            else:
+                abort(400, message="No Results for this query")
+        else:
+            logging.error("%r != %r" % (q.id,query_id))
+            abort(404, message="Invalid visualization for query")
+    else:
+        abort(404, message="Visualization not found.")
+
+    client_config = {
```
arikfr: Let's move `client_config` to somewhere where it will be shared between this function and the index one.

redash/handlers/embed.py
arikfr: Another alternative is: `vis = models.Visualization.get(models.Visualization.id == ..., models.Visualization.query == ...)` - this will throw `models.Visualization.DoesNotExist` if not found (you can do `select()` with `first()` instead, which won't throw iirc).

redash/handlers/embed.py
```diff
@@ -0,0 +1,47 @@
...
+        'allowScriptsInUserInput': settings.ALLOW_SCRIPTS_IN_USER_INPUT,
+        'highChartsTurboThreshold': settings.HIGHCHARTS_TURBO_THRESHOLD,
+        'dateFormat': settings.DATE_FORMAT,
+        'dateTimeFormat': "{0} HH:mm".format(settings.DATE_FORMAT)
+    }
+
+    return render_template("embed.html",
+                           name=settings.NAME,
+                           client_config=json_dumps(client_config),
+                           visualization=json_dumps(v),
```
arikfr: It might be "safer" to dump the query without its SQL. While in your use case it doesn't matter, for others it might be beneficial to "hide" the SQL.

arikfr: But you can leave it as is for now.

akariv: I agree, left it as is for now.

redash/handlers/embed.py
akariv: Moved to settings.py

redash/handlers/embed.py
akariv: Done on the backend.

When can it happen that no results are available?

redash/handlers/embed.py
akariv: done

akariv:
@arikfr fixed all CR comments

redash/handlers/embed.py
akariv: ?

victoruiz:
Whenever I embed a visualization, I get a problem with highcharts dependency within bower components.

http://localhost:9001/bower_components/highcharts/highcharts.js


captura de pantalla 2015-12-15 a las 16 02 13

arikfr:
@victoruiz try running `bower install` again. I assume that you cloned master branch, ran bower install and then switched to this branch. master no longer uses HighCharts, while this one does.

akariv:
Seems I should probably update the dependencies for this branch... @arikfr
do you have any suggestion on how we can do it with less code duplication?

victoruiz:
Hi @arikfr I cloned akariv:master which contains the embed functionality, moved into /opt/redash/current/rd-ui/, then executed bower install... and there is no highcharts directory within rd_ui/app/bower_components.

Will look a little bit deeper.

akariv:
@victoruiz Hi I fixed the issues you encountered (i rebased from the upstream and didn't account for some of the changes). Now it should work for you.

PR #666, opened 2015-11-20
Fix: Specifying field type in the field name using __ didn't work

alonho:
It works for '::' but probably didn't work for '__' due to a a copy-paste


arikfr:
Nice catch.

shyamgopal:
@arikfr , may I know if this will affect using '__' in Filter or multiFilter in BigQuery ? I am concerned because of the pull request number being '666' ;-)

PR #668, opened 2015-11-22
Fix: AlertSubscription missing in all_models

cou929:
Hi.

I think `AlertSubscription` need to be included in `all_models`. Does this make sense?


arikfr:
Nice catch. Thanks!

PR #677, opened 2015-11-30
Fix: don't partially save a query if it hasn't been saved yet (with all the data)

### LINKED ISSUE #669: User can't save query without query text
(seems like it doesn't pass the `data_source_id`)

alonho:
Closes #669.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -621,18 +621,25 @@ class Alert(ModelTimestampsMixin, BaseModel):
     def all(cls):
         return cls.select(Alert, User, Query).join(Query).switch(Alert).join(User)
 
-    def to_dict(self):
-        return {
-            'id': self.id,
-            'name': self.name,
-            'query': self.query.to_dict(),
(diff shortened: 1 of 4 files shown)


rd_ui/app/scripts/controllers/query_view.js
```diff
@@ -96,7 +96,10 @@
 
     $scope.saveQuery = function(options, data) {
       if (data) {
-        data.id = $scope.query.id;
+        if ($scope.query.id)
+          data.id = $scope.query.id;
+        else // this is a new query, pass initial params (like datasource) and data (like name)
+          data = _.extend($scope.query, data)
```
arikfr: We use parentheses even for single line `if` statements.

But maybe just avoid saving (i.e. exit `saveQuery`) if `query.id` is `null`?

alonho:
Accepted your proposal. We won't save if `query.id` is `null`.

PR #680, opened 2015-12-01
Feature: show rows count per table

alexanderlz:
Add the ability to see tables row count.


rd_ui/app/views/query.html
```diff
@@ -90,7 +90,7 @@
             <div class="schema-browser">
                 <div ng-repeat="table in schema | filter:schemaFilter track by table.name">
                     <div class="table-name" ng-click="table.collapsed = !table.collapsed">
-                        <i class="fa fa-table"></i> <strong><span title="{{table.name}}">{{table.name}}</span></strong>
+                        <i class="fa fa-table"></i> <strong><span title="{{table.name}}">{{table.name}} ({{table.size}})</span></strong>
```
arikfr: The size should only be available if there is actually size... Something like this: `<span ng-if="table.size"> ({{table.size}})</span>`.

redash/query_runner/__init__.py
```diff
@@ -104,6 +105,25 @@ def to_dict(cls):
...
+        return schema_dict.values()
+
+    def _get_tables(self, schema_dict):
+        return []
+
+    def _get_tables_stats(self, tables_dict):
+        for t in tables_dict.keys():
+            if type(tables_dict[t]) == dict:
+                res = self._run_query_internal('select count(*) as cnt from %s' % t)
+                tables_dict[t]['size'] = res[0]['cnt']
```
arikfr: Currently we might load the schema the first time as part of a web request, hence it needs to be relatively fast, so this can't be the default implementation, specially not for Postgres (because of Redshift).

We can either change this default implementation, or have it run as a Celery task.

redash/models.py
```diff
@@ -320,7 +320,7 @@ def get_schema(self, refresh=False):
 
         if cache is None:
             query_runner = self.query_runner
-            schema = sorted(query_runner.get_schema(), key=lambda t: t['name'])
+            schema = sorted(query_runner.get_schema(get_stats=True), key=lambda t: t['name'])
```
arikfr: Isn't this the context where it's called from the web request?

PR #698, opened 2015-12-10
Fix: When stacking is enabled show both the relative value (in %) and the absolute value

### LINKED ISSUE #694: Charts: when stacking enabled, the tooltip shows the "stacked" value instead of real one

alonho:
Closes #694.


alonho:
screen shot 2015-12-10 at 5 59 40 pm

alonho:
By displaying only the value that is hovered on and not all the values in the display changes: plotly shows (x, y) instead of just the y value. I think this improves readability overall and specifically this issue as well.

screen shot 2015-12-10 at 10 30 20 pm

PR #701, opened 2015-12-10
Fix: #694: When stacking is enabled show both the relative value (in %) and the absolute value (attempt #2)

alonho:
screen shot 2015-12-10 at 10 45 19 pm

Finally made customization of onhover events work!


arikfr:
I think this handles only the percent stacking.  What about regular stacking?

arikfr:
@alonho merged. but check my comment.

PR #710, opened 2015-12-16
Feature: BigQuery: limit amount of MB processed per query

ryotarai:
Sometimes, a query executed on BigQuery highly costs. To prevent that, I added `totalMBytesProcessedLimit` parameter to BigQuery runner.
If `totalMBytesProcessedLimit` is set, the runner will dry-run a query and check data size which will be processed, and if the data is larger than `totalMBytesProcessedLimit`, an error occur like the following image.

re dash new query 2015-12-16 20-31-56

This PR contains indent changes, so compare view ignoring whitespace is more understandable: https://github.com/getredash/redash/compare/master...ryotarai:bq-max-mb-processed?w=

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -95,6 +95,10 @@ class BigQuery(BaseQueryRunner):
                 'jsonKeyFile': {
                     "type": "string",
                     'title': 'JSON Key File'
+                },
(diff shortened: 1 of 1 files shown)


arikfr:
Awesome! Will be useful for the demo instance too (although until now it doesn't look like anyone been abusing it).

PR #714, opened 2015-12-20
Feature: load schema for MongoDB data source

erans:
Added schema support for MongoDB. Collections will be shown as tables and we merge the first and last documents (sorted by Natural order) to show the properties of the document. Since MongoDB is document based it might miss a few fields but it should give a good enough reference.

### DIFF
--- a/redash/query_runner/mongodb.py
+++ b/redash/query_runner/mongodb.py
@@ -102,14 +102,55 @@ class MongoDB(BaseQueryRunner):
 
         return None
 
-
-    def run_query(self, query):
+    def _get_db(self):
         if self.is_replica_set:
             db_connection = pymongo.MongoReplicaSetClient(self.configuration["connectionString"], replicaSet=self.configuration["replicaSetName"])
         else:
             db_connection = pymongo.MongoClient(self.configuration["connectionString"])
 
-        db = db_connection[self.db_name]
+        return db_connection[self.db_name]
+
+    def _merge_property_names(self, columns, document):
+        for property in document:
+              if property not in columns:
+                  columns.append(property)
+
+    def _get_collection_fields(self, db, collection_name):
+        # Since MongoDB is a document based database and each document doesn't have 
+        # to have the same fields as another documet in the collection its a bit hard to
+        # show these attributes as fields in the schema.
+        #
+        # For now, the logic is to take the first and last documents (last is determined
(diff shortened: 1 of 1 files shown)


redash/query_runner/mongodb.py
```diff
@@ -102,14 +102,55 @@ def _get_column_by_name(self, columns, column_name):
...
         else:
             db_connection = pymongo.MongoClient(self.configuration["connectionString"])
 
-        db = db_connection[self.db_name]
+        return db_connection[self.db_name]
+
+    def _merge_property_names(self, columns, document):
+        for property in document:
+              if property not in columns:
+                  columns.append(property)
```
arikfr:
```python
return [prop for prop in document if prop not in columns]`
```

but you can use `set`s here (which will also be more performant, although it's probably negligible): 
```python
set(columns) | set(document.keys()) # or columns can be a set from the get-go.
```

redash/query_runner/mongodb.py
```diff
@@ -102,14 +102,55 @@ def _get_column_by_name(self, columns, column_name):
...
+
+        for d in db[collection_name].find().sort([("$natural", 1)]).limit(1):
+            first_document = d
+
+        for d in db[collection_name].find().sort([("$natural", -1)]).limit(1):
+            last_document = d
+
+        columns = []
+        if first_document: self._merge_property_names(columns, first_document)
+        if last_document: self._merge_property_names(columns, last_document)
```
arikfr: If I'm not mistaken, this can be replaced with (make the `_merge_property_names` method redundant too):

```python
documents  = db[collection_name].find().sort([("$natural", 1)]).limit(1) 
documents += db[collection_name].find().sort([("$natural", -1)]).limit(1)

columns  = set([document.keys() for document in documents])
```

erans: Usually yes but I am known for preferring plain code vs syntax trickery :-)

arikfr: I'm not sure the above qualifies as "syntax trickery" :-)

erans: :-)

Anyhow the reason for the merge  methods is that documents can contain sub
dictionary and while I didn't implement it yet I want to show subdocuments
in the from of collection.fieldname (similar to what Bigquery does to json)

PR #715, opened 2015-12-20
added a refresh flag (value 0 or 1) to force a refresh of a data sour…

erans:
…ce schema in /api/data_sources/[data-source-id]/schema

### DIFF
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -3,7 +3,7 @@ import json
 from flask import make_response, request
 from flask.ext.restful import abort
 
-from redash import models
+from redash import models, settings
 from redash.wsgi import api
 from redash.permissions import require_permission
 from redash.query_runner import query_runners, validate_configuration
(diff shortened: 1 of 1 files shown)


redash/handlers/data_sources.py
```diff
@@ -75,8 +75,10 @@ def post(self):
 
 class DataSourceSchemaAPI(BaseResource):
     def get(self, data_source_id):
+        refresh = request.args.get("refresh", "0", type=bool)
```
arikfr: I don't think this works the way you think it works. From what I've seen, if no refresh value is passed, it will return "0" (as string), which is a truth value... also, calling `bool` on any string value (including 0) will return True.

erans: I thought it was smarter than it was. Will hard code it to 0 and 1.

arikfr: You can use the parse_bool function from settings module.

erans: Cool.  Will do.

erans: Fixed it. 

In Python 2.x, True is 1 and 0 is False, however it can get overridden (unlike Python 3.x).
http://stackoverflow.com/questions/2764017/is-false-0-and-true-1-in-python-an-implementation-detail-or-is-it-guarante

But its better to be safe than sorry.

redash/handlers/data_sources.py
```diff
@@ -75,8 +75,11 @@ def post(self):
 
 class DataSourceSchemaAPI(BaseResource):
     def get(self, data_source_id):
+        refresh = request.args.get("refresh", "false")
+	refresh = settings.parse_boolean(refresh)
```
arikfr: looks like an indentation issue?

arikfr: ping.

arikfr:
Please reopen after rebase & fixing indentation.

PR #722, opened 2015-12-24
Fix: use second y axis for line charts while stacking

### LINKED ISSUE #721: Charts: can not use right yaxis if stackable is enabled
Hi.

I want to make chart that has stacked bars graph on left axis and has lines on right axis 
such as shown in the figure below.
However, now cant make that.
So, I want to make right axis enabled if above case.

ninneko:
Fixes #721.


arikfr:
Thanks :+1:

PR #725, opened 2015-12-27
Fix: Google OAuth - support for next

### LINKED ISSUE #720: User should be redirected to his original destination after login with Google OAuth
If the user tried to open a page before being logged in, he should be redirected to this page after successful login.

akariv:
Fixes #720.


arikfr:
Thanks!

PR #727, opened 2015-12-29
Fix mispelling

josegonzalez:


arikfr:
OMG. I can't believe I overlooked this. Thanks!

PR #746, opened 2016-01-07
Feature: add an option to update a query every 30 days

Xangis:
Add an option to update a query every 30 days for use with things like monthly reports and month-over-month comparisons.

I use this in my Redash deployment and thought it might be useful for others.


arikfr:
Thank :+1:

PR #750, opened 2016-01-09
Docs: how to setup SSL using Let's Encrypt SSL certs

JohnConnell:


docs/misc/letsencrypt.rst
```diff
@@ -0,0 +1,133 @@
...
+
+  sudo rm /etc/nginx/sites-available/redash.bak
+
+
+**Important Note:** letsencrypt certificates only remain valid for 90 days. To renew your certificate, simply run the same command again (below) and make sure you provide the same values you originally entered in step 4 when prompted.
+
+.. code::
+
+  cd /opt/letsencrypt
+  ./letsencrypt-auto certonly --standalone
```
arikfr: When renewing there is no need to turn off nginx? (to free port 80)

PR #765, opened 2016-01-15
Feature: support HipChat Server

nakechi:
Failed to send alert for Hipchat Server(on-premise), because api's url is different.
So, added support for Hipchat Server.

#### Setting environment variables
- HIPCHAT_API_URL


arikfr:
Thanks!

Issue #768, opened 2016-01-18
Empty cells in google sheets displayed as datetime values


JohnConnell:
When querying a Google Sheet, re:dash is displaying empty cells as datetime values equal to the current date.

Google Sheet with cell D2 empty:
![google_sheets](https://cloud.githubusercontent.com/assets/2201507/12387820/59c7da6e-bd8c-11e5-8438-00354c8f3f48.png)

redash:
![google_sheets_redash](https://cloud.githubusercontent.com/assets/2201507/12387819/59c3da5e-bd8c-11e5-8d13-dabe086b3b78.png)

shyamgopal:
@arikfr , in the json_data generated in redash/query_runner/google_spreadsheets.py, 'int' type is indicated by TYPE_INTEGER, 'float" is indicated by TYPE_FLOAT etc etc. Similarly, if the cell is an empty string ('') or 'None' what should be the type indicated in json_data; TYPE_NONE or something like that ? So in response to a cell being an empty string (''), what should be the data type in json_data ?

arikfr:
The default is `TYPE_STRING`, so I guess it should be it...

arikfr:
Although, maybe we should add `TYPE_NONE`.

shyamgopal:
1) Thanks so much @JohnConnell for reporting the bug.
2) Made some elementary changes in current/redash/query_runner/google_spreadsheet.py  , following are screenshots of results that I obtained. If results are satisfactory to you and @arikfr , I will checkout latest redash code, patch it and issue a pull request.
3) Actual Google spreadsheet:
<img width="831" alt="actualgs" src="https://cloud.githubusercontent.com/assets/11750143/12451079/4fe4f19e-bfad-11e5-8fcd-fb9102a5fce5.png">
4) resulting Redash table widget:
![redashtable](https://cloud.githubusercontent.com/assets/11750143/12451108/7b06855e-bfad-11e5-84ec-e064c6078c87.png)
5) resulting bar graph of Xvalue vs IntCol:
![intcol](https://cloud.githubusercontent.com/assets/11750143/12451139/9ad3b88e-bfad-11e5-8d01-af7f755bd38b.png)
6) resulting bar graph of Xvalue vs FloatCol:
![floatcol](https://cloud.githubusercontent.com/assets/11750143/12451153/b097dff6-bfad-11e5-801b-84af667d02bf.png)
[...]

PR #782, opened 2016-01-21
Fix: Empty cells in google sheets displayed as datetime values #768

shyamgopal:
A bug fix for #768  ; where empty cells in Google spreadsheet of type TYPE_STRING and TYPE_DATETIME were getting replaced by current timestamp.


arikfr:
Thanks!

PR #808, opened 2016-02-01
Feature: Microsoft SQL Server query runner

joeharris76:
Uses `pymssql` which in turn uses `FreeTDS`. Note that the data type
support is somewhat limited (see “datasources” page in docs).

### DIFF
--- a/redash/query_runner/mssql.py
+++ b/redash/query_runner/mssql.py
@@ -7,11 +7,11 @@ from redash.utils import JSONEncoder
 
 logger = logging.getLogger(__name__)
 
- try:
-     import pymssql
-     enabled = True
- except ImportError:
-     enabled = False
+try:
+    import pymssql
+    enabled = True
+except ImportError:
+    enabled = False
 
 # from _mssql.pyx ## DB-API type definitions & http://www.freetds.org/tds.html#types ##
 types_map = { 
--- a/redash/query_runner/mssql.py
+++ b/redash/query_runner/mssql.py
@@ -1,0 +1,141 @@
+import json
+import logging
+import pymssql
+import sys
+
+from redash.query_runner import *
+from redash.utils import JSONEncoder
+
+logger = logging.getLogger(__name__)
+
+# from _mssql.pyx ## DB-API type definitions & http://www.freetds.org/tds.html#types ##
+types_map = { 
+    1: TYPE_STRING,
+    2: TYPE_BOOLEAN,
+    3: TYPE_INTEGER,
+    4: TYPE_DATETIME,
+    5: TYPE_FLOAT,  
+}
+
+class SqlServer(BaseSQLQueryRunner):
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "user": {
+                    "type": "string"
+                },
+                "password": {
+                    "type": "string"
+                },
+                "server": {
+                    "type": "string",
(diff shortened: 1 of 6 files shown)


redash/query_runner/mssql.py
```diff
@@ -0,0 +1,141 @@
+import json
+import logging
+import pymssql
```
arikfr: You should wrap the import of `pymssql` with a try/except block to avoid breaking existing deployments that don't have this package:

```python
try:
    import pymssql
    enabled = True
except ImportError:
    enabled = False
```

And then use the `enabled` value for the `enabled` method of the query runner:

```python
    @classmethod
    def enabled(cls):
        return enabled
```

setup/ubuntu/bootstrap.sh
```diff
@@ -157,6 +157,10 @@ apt-get install -y libffi-dev libssl-dev
 # MySQL dependencies:
 apt-get install -y libmysqlclient-dev
 
+# Microsoft SQL Server dependencies (`sudo` required):
+sudo apt-get install -y freetds-dev
+sudo pip install pymssql==2.1.1
```
arikfr: No need to add it explicitly as the bootstrap script for Ubuntu install all requirements from `requirements_all_ds.txt`.

setup/ubuntu/bootstrap.sh
```diff
@@ -157,6 +157,10 @@ apt-get install -y libffi-dev libssl-dev
 # MySQL dependencies:
 apt-get install -y libmysqlclient-dev
 
+# Microsoft SQL Server dependencies (`sudo` required):
+sudo apt-get install -y freetds-dev
```
arikfr: You should add this to the Dockerfile as well, although I can do that myself if you prefer.

setup/ubuntu/bootstrap.sh
joeharris76: Can we guarantee the bootstrap will be run as `sudo` because the install will definitely fail without it.

setup/ubuntu/bootstrap.sh
joeharris76: I had a look at this but I can't really see where it would go. I guess you'll need to do it.

setup/ubuntu/bootstrap.sh
arikfr: [Lines 7-16](https://github.com/joeharris76/redash/blob/master/setup/ubuntu/bootstrap.sh#L7-L16) verify that we run as root (sudo).

redash/query_runner/mssql.py
```diff
@@ -0,0 +1,150 @@
+import json
+import logging
+import sys
+
+from redash.query_runner import *
+from redash.utils import JSONEncoder
+
+logger = logging.getLogger(__name__)
+
+ try:
```
arikfr: There is a misplaced space before `try` and `except`.

setup/ubuntu/bootstrap.sh
arikfr: :+1:

setup/ubuntu/bootstrap.sh
joeharris76: OK, resolved this in a new commit

redash/query_runner/mssql.py
joeharris76: OK, resolved this in a new commit

arikfr:
Thanks!

PR #813, opened 2016-02-06
Fix shellcheck issues in bootstrap.sh scripts

ojarva:
Fix shellcheck complaints. These changes are not particularly important,
but spotting new/real issues is easier when checker output is empty by
default.


arikfr:
Thanks!

PR #823, opened 2016-02-11
Feature: Button toggle to display dashboard in at full screen width

matthew-sochor:
We display dashboards on televisions and would like to better utilize the screen space.  This pull request adds a simple toggle button to the dashboard that allows the user to switch between `container` and `container-fluid` in the dashboard section that displays widgets.

By default the toggle is set to `false`


arikfr:
Thanks! I'm planning to improve the dashboard layout options, but this is a nice solution for the meantime.

PR #824, opened 2016-02-12
Add comment about what set_google_apps_domains actually does

agamrafaeli:
Had to debug the code to figure out what domain I should be adding here. Open to hear a different phrasing of this comment.

### DIFF
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -118,7 +118,9 @@ file.
    sudo -u redash bin/run ./manage.py org set_google_apps_domains {{domains}}
 
 
-If you're passing multiple domains, separate them with commas.
+If you're passing multiple domains, separate them with commas. 
+
+These are the domains you will allow users to login from. I.e. `gmail.com` will allow all users trying to login with a gmail.com address to login. Other addresses will not be allowed. 
 
 
 5. Restart the web server to apply the configuration changes:


docs/setup.rst
```diff
@@ -118,7 +118,9 @@ file.
    sudo -u redash bin/run ./manage.py org set_google_apps_domains {{domains}}
 
 
-If you're passing multiple domains, separate them with commas.
+If you're passing multiple domains, separate them with commas. 
+
+These are the domains you will allow users to login from. I.e. `gmail.com` will allow all users trying to login with a gmail.com address to login. Other addresses will not be allowed.
```
arikfr: Thanks for submitting an update for the docs. This is always appreciated :-) How about this alternative for the phrasing:

These are the (Google Apps) domains you will allow users to create accounts with; i.e. `redash.io` will allow all users trying to create an account with a `@redash.io` Google account to create an account and login. Other addresses will not be allowed. 

(Main change is to use a non Gmail domain and some clarification that we talk about Google accounts)

jeffwidman:
Ping @agamrafaeli

jeffwidman:
@arikfr Was this closed because no longer relevant, or because @agamrafaeli has been AWOL?

arikfr:
Mainly the later.

PR #835, opened 2016-02-17
Ubuntu bootstrap script - move update before upgrade

benmanns:
Running update before upgrade will fetch the latest sources, so we
can be sure that the upgrades will bring the box to the latest
versions of everything. Otherwise, this is often a no-op because
the box's sources will be cached at time of generation, meaning
there is nothing to upgrade.


arikfr:
Thanks.

PR #836, opened 2016-02-17
Tasks/c 5871 redash deploy scripts

bsiegel:
Partially completes https://youtrack.mobiledefense.com/issue/c-5871

### DIFF
--- a/rd_ui/app/index.html
+++ b/rd_ui/app/index.html
@@ -183,6 +183,7 @@
 <script>
     // TODO: move currentUser & features to be an Angular service
     var clientConfig = {{ client_config|safe }};
+    var basePath = "{{base_href}}";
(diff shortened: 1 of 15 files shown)


arikfr:
@bsiegel if you're installing from git, you should "compile" the frontend assets as well. Otherwise the load time in the browser will be slower.

bsiegel:
@arikfr Ah yep, I've been running through by trial and error, and did manage to find that I needed to run `make deps` to get thing working. I opened this PR by mistake (github loves to open PRs against upstream rather than against our fork's master branch) but in case you or anyone else is interested, I have bot bootstrap and upgrade scripts here: https://github.com/mobiledefense/redash/pull/2

Note, our provisioning system takes care of installing dependencies (like nodejs used for compiling the assets with gulp).

Issue #763, opened 2016-01-14
Flexible Notification Setting


CzechJiri:
I feel current implementation of email/webhook/hipchat is very kosher and a bit too deeply integrated into redash .env setting.

What might work better is having datasource-like approach for notification setting (notification setting done similar way datasources are configured) and would include next to email/hipchat also generic python code option.

This way the actual configuration is done in the web interface and users can have multiple webhook endpoints or multiple slack channels or any custom python implementation(s).

arikfr:
I completely agree. The way it's implemented right now was the quickest solution and probably gives some value. But eventually, as you described, there should be definitions per alert with some global defaults.

arikfr:
#842 implements this.

PR #842, opened 2016-02-22
Flexible notifications

alexdebrie:
Addresses #763.

This PR adds the ability to add `NotificationDestinations`, which are similar to data sources but for alerts. The idea is that you could register a particular destination, like a Slack room or a webhook, and then use it as a subscriber to a particular alert. 

General items:

- [x] Add `NotificationDestination` and `NotificationDestinationGroup` models.
- [x] Create `BaseDestination` and convenience functions for registering and using destinations.
- [x] Add `destination` handlers on backend.
- [x] Add destination CRUD elements on frontend.
- [x] Allow specifying which destination should be used when creating or subscribing to an alert.
- [x] Update `check_alerts_for_query` task to conform with new setup.
- [x] Implement Slack, Hipchat, and external webhook Destination types.

Re: `AlertSubscriptions`, I've added a `destination` field which maps to a `NotificationDestination`. My thought is that the UI will have a dropdown where a user can pick a destination when subscribing. For emails, we could either leave the `destination_id` null or just create an initial row in the `notification_destinations` table that is for email.

I'll probably need a little help on the Angular portion of creating AlertSubscriptions. Once we settle on how subscriptions will be created, I should be able to implement most of the remaining backend items.

I'm open to any and all suggestions/comments/changes on what I have so far and plans going forward.


redash/models.py
```diff
@@ -1034,7 +1012,134 @@ def record(cls, event):
...
+    def all(cls, alert_id):
+        return AlertSubscription.select(AlertSubscription, User).join(User).where(AlertSubscription.alert==alert_id)
+
+    @classmethod
+    def unsubscribe(cls, alert_id, user_id):
+        query = AlertSubscription.delete().where(AlertSubscription.alert==alert_id).where(AlertSubscription.user==user_id)
+        return query.execute()
+
+
+all_models = (Organization, Group, DataSource, DataSourceGroup, User, QueryResult, Query, Alert, AlertSubscription, Dashboard, Visualization, Widget, Event, NotificationDestination, NotificationDestinationGroup)
```
arikfr: Because of the foreign keys relationships the order here is important, and `NotificationDestination` and `NotificationDestinationGroup` should be before `AlertSubscription`.

arikfr:
I just merged #844. It's part of my work on encrypting secrets we store in the database, but merged early as it introduces change to how we deal with configuration.

Basically encapsulated most of the logic in the `ConfigurationContainer` class and also see the `ConfigurationField` which you need to use in the model.

redash/models.py
```diff
@@ -1034,7 +1012,134 @@ def record(cls, event):
...
+
+        return notification_destinations
+
+    @property
+    def groups(self):
+        groups = NotificationDestinationGroup.select().where(NotificationDestinationGroup.notification_destination==self)
+        return dict(map(lambda g: (g.group_id, g.view_only), groups))
+
+
+class NotificationDestinationGroup(BaseModel):
```
arikfr: Do we really need to manage per group permissions of notifications?

Maybe it will be simpler to have the following permission model:
1. Everyone can see all `NotificationDestination`s and use them.
2. Only the owner (the user who created it) and admins can edit them.

WDYT?

redash/models.py
```diff
@@ -1034,7 +1012,134 @@ def record(cls, event):
...
 
-all_models = (Organization, Group, DataSource, DataSourceGroup, User, QueryResult, Query, Alert, AlertSubscription, Dashboard, Visualization, Widget, Event)
+class NotificationDestination(BelongsToOrgMixin, BaseModel):
+
+    id = peewee.PrimaryKeyField()
+    org = peewee.ForeignKeyField(Organization, related_name="notification_destinations")
+    name = peewee.CharField()
+    type = peewee.CharField()
+    options = peewee.TextField()
+    created_at = DateTimeTZField(default=datetime.datetime.now)
```
arikfr: Need to have a reference to the user who created it.

arikfr:
> Re: AlertSubscriptions, I've added a destination field which maps to a NotificationDestination. My thought is that the UI will have a dropdown where a user can pick a destination when subscribing. For emails, we could either leave the destination_id null or just create an initial row in the notification_destinations table that is for email.

Both are good, but let's try the second one (having a notification destination of email) and see how it goes. If it results in too much code, we can revert to having null as destination_id.

Also note that you need to change the signature of the `notify` method to accept: `alert` & `user` along with the `query` object.

redash/models.py
alexdebrie: Yep, good call. I'll get that updated.

arikfr:
@alexdebrie I'll be happy to help, but it will be only next week (Sunday or Monday).

alexdebrie:
Thanks, @arikfr. I'll spend some time on it tomorrow and Saturday to see if I can get it figured out before then.

alexdebrie:
@arikfr I managed to hack together the ability to add a destination when creating an Alert. Changing the one click subscribe button will take a little more work, and we'll have to change the subscribers list since it's not just a user anymore.

I also added new task logic to call `notify` on each of the subscriptions for a triggered alert and implemented three destination types. I tested Slack and the webhook one but don't have email set up. It should be identical to the original email notification.

Feel free to let me know of any aspects you'd like changed.

alexdebrie:
@arikfr: Below is a screenshot of the UI for adding/removing subscriptions to an alert. I'm happy to make any changes.

Also, I'm handling backwards compatibility with the current alert system [here](https://github.com/getredash/redash/pull/842/files#diff-c078d895901960aef9f60f32e6a6f9a8R367). Basically, we allow None for the `destination` foreign key on the AlertSubscription model. All newly created models will have a value for that. The only ones that will have None are old AlertSubscriptions that were tied to email. This just imports the email destination and sends the email to the the creator of the AlertSubscription.

Let me know if there are any changes you'd like. I'll work on fixing the merge conflicts.

screen shot 2016-03-14 at 1 30 43 pm

rd_ui/app/scripts/controllers/alerts.js
```diff
@@ -120,6 +122,40 @@
       },
       controller: function ($scope) {
         $scope.subscribers = AlertSubscription.query({alertId: $scope.alertId});
+        $scope.subscription = {};
+        $scope.destinations = Destination.query();
+
+        $scope.destinationsDisplay = function(destination) {
+            var icons = {
+                'slack': 'fa-slack',
+                'email': 'fa-envelope',
```
arikfr: The icon should be part of the destination type definition.

redash/handlers/destinations.py
```diff
@@ -0,0 +1,110 @@
...
+    def get(self):
+        destinations = models.NotificationDestination.all(self.current_org)
+
+        response = {}
+        for ds in destinations:
+            if ds.id in response:
+                continue
+
+            d = ds.to_dict()
+            d['view_only'] = is_admin_or_owner(ds.id)
```
arikfr: I'm pretty sure this is not needed.

redash/models.py
```diff
@@ -1028,7 +1006,87 @@ def record(cls, event):
...
+    def destination(self):
+        return get_destination(self.type, self.options)
+
+    @classmethod
+    def all(cls, org):
+        notification_destinations = cls.select().where(cls.org==org).order_by(cls.id.asc())
+
+        return notification_destinations
+
+    def notify(self, alert_id, query_id, user_id, new_state, app, host):
```
arikfr: Why take ids instead of actual objects?

redash/handlers/destinations.py
```diff
@@ -0,0 +1,110 @@
...
+                                                     user=self.current_user)
+        destination.save()
+
+        return destination.to_dict(all=True)
+
+api.add_org_resource(DestinationListAPI, '/api/destinations', endpoint='destinations')
+api.add_org_resource(DestinationAPI, '/api/destinations/<destination_id>', endpoint='destination')
+
+
+class DestinationSchemaAPI(BaseResource):
```
arikfr: Copy paste from data sources left over?

arikfr:
The UI looks good. I wonder if we should change the terminology (subscribe and destination).

As for the migration, I will leave a few comments tomorrow.

rd_ui/app/scripts/controllers/alerts.js
alexdebrie: Just for clarification -- do you mean include icon on the model in the backend that gets sent as part of the API response?

arikfr: It will be defined in `BaseDestination` derived classes, a method that returns the icon. And it will be returned in the API calls both for types and destination instances.

rd_ui/app/scripts/directives/destination_directives.js
```diff
@@ -0,0 +1,80 @@
+(function () {
+  'use strict';
+
+  var directives = angular.module('redash.directives');
+
+  directives.directive('destinationForm', ['$http', 'growl', function ($http, growl) {
```
arikfr: We should have a single component (directive) of "DynamicForm", that will be used both here and in the data source form, to reduce the code duplication.

Will you need help with this?

rd_ui/app/scripts/services/resources.js
```diff
@@ -554,6 +554,18 @@
     return DataSourceResource;
   };
 
+  var Destination = function ($resource) {
+    var actions = {
+      'get': {'method': 'GET', 'cache': false, 'isArray': false},
+      'query': {'method': 'GET', 'cache': false, 'isArray': true},
+      'getSchema': {'method': 'GET', 'cache': true, 'isArray': true, 'url': 'api/destinations/:id/schema'}
```
arikfr: Copy paste...

rd_ui/app/scripts/services/resources.js
```diff
@@ -594,7 +606,12 @@
   };
 
   var AlertSubscription = function ($resource) {
-    var resource = $resource('api/alerts/:alertId/subscriptions/:userId', {alertId: '@alert_id', userId: '@user.id'});
+    var actions = {
+      'get': {method: 'GET'},
+      'save': {method: 'POST'},
+      'delete': {method: 'DELETE'}
+    };
```
arikfr: It doesn't look like this `actions` object is used anywhere.

redash/tasks.py
```diff
@@ -357,18 +357,20 @@ def check_alerts_for_query(self, query_id):
...
-                notify_webhook(alert, query, html, new_state)
+            host = base_url(alert.query.org)
+            
+            for subscription in alert.subscriptions:
+                try:
+                    if subscription.destination:
+                        subscription.destination.notify(alert.id, query.id, subscription.user.id, new_state, app, host) 
+                    else:
+                        # For backwards compatibility with prior alert system that was primarily email-based
+                        from redash.destinations import get_destination
```
arikfr: If this is just for backwards compatibility, let's handle this in the migrations:

- Create default email NotificationDestination.
- Assign it to all existing subscriptions.

redash/tasks.py
```diff
@@ -357,18 +357,20 @@ def check_alerts_for_query(self, query_id):
...
+                        config = {}
+                        destination = get_destination('email', config)
+                        destination.notify(alert.id, query.id, subscription.user.id, new_state, app, host) 
+                except Exception as e:
+                    logger.warn("Exception: {}".format(e))
 
 
 def notify_hipchat(alert, html, new_state):
```
arikfr: Need to migrate this one as well.

redash/tasks.py
arikfr: Also in the migration let's create a webhook/hipchat destination, if it's setup in settings.py

rd_ui/app/scripts/controllers/alerts.js
alexdebrie: Thanks. Updated.

alexdebrie:
@arikfr: I think I've got the migration and the dynamicForm directive, as well as the other small tweaks that were needed. 

I'm running into one last issue. For email subscriptions, I'm making on making a single `NotificationDestination` for email to avoid having to create a different email destination for every single user. My thought was that the user attached to an `AlertSubscription` could be used to send the email. However, I was also planning on making a unique constraint for alert and destination on `AlertSubscription`. This would make it so you don't have the one Slack room subscribed to the same alert multiple times. However, it would limit to only one user being able to get a notification by email. 

Do you have any thoughts on how to fix that? One option would be to manually handle uniqueness by allowing multiple subscriptions to same alert if destination is email but not for others, which I don't love. Another way would be to require users that want email to each register their addresses as separate destinations, which I also don't love.

arikfr:
Let's automatically create a destination for each user. At first, I thought it sucks (requires "book-keeping"), but on second thought, I'm not sure it's that bad and not that hard to "book-keep".

arikfr:
It has the added benefit of allowing a user to subscribe other users to some alert.

alexdebrie:
Sounds good. I'm going to add a setting to not create an email destination for each user. We don't have an email server configured on our instance, so I'd like to avoid listing 300+ destinations that will never be used. I'll also add a search filter to the "Add a destination" form.

arikfr:
That's a good point. Maybe it's not a very good idea to have an email destination for every user, because it will make the list very long -> meaning we need to start supporting pagination and such.

How about the following:
1. `AlertSubscription` without a destination is email subscription for the user who created it.
2. In terms of UI, we will keep the current one (with the faces) for user subscriptions and have your new one for destinations?

alexdebrie:
Was out for a few days but finally got back to this. Below is an image of the new Alert page. It allows for both individual email subscription or destination subscription. Let me know what you think.
screen shot 2016-03-23 at 11 18 47 am

arikfr:
I suspect some users will think the big blue "Subscribe" button is for them to subscribe, as it is easy to miss the "eye" button now.

How about the following:
1. Add <hr> between user subscription and destination subscriptions.
2. Change the button text from "Subscribe" to "Add" and have it inline with the dropdown.
3. Give the users a button as well (with the text "Subscribe") -> not sure about this one.

alexdebrie:
How about this:

screen shot 2016-03-23 at 1 45 11 pm

alexdebrie:
To de-clutter it and make it less confusing, I'm hiding the user subscriptions if email is not enabled:

screen shot 2016-03-24 at 9 28 21 am

arikfr:
Hi @alexdebrie,

Just wanted to apologize that your pull request is still waiting... the work on the redesign took longer than expected and then a few unexpected things happen (as they always do). 

But I'm releasing v0.10.0 today and rebase/update your pull request is my next item on the to do list. So hopefully early next week it will be merged and will be part of 0.11.0. 

Thank you for your patience.

alexdebrie:
No problem! Let me know if there's anything I can help with.

arikfr:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;9zgRtNeA&#x2F;62-more-destinations-for-alert-notifications"> More destinations for alert notifications</a>

toyama0919:
Awesome feature 😊

arikfr:
Almost there... updating the design of the subscriptions area and then I'm done.

I think we need a better name than "Notification Destinations". Maybe something like "Integrations"?"Targets"?

arikfr:
And finally after over 2 months it's merged! 

I did another change to the UI and it's no longer as in the GIF above.

romone22:
Great !! Thanks

PR #849, opened 2016-02-23
Fix: Presto - deduplicate column names

### LINKED ISSUE #847: presto query runner merges columns with same name
The query `SELECT 'foo' AS same_name, 'not foo' AS same_name` against a Presto data source generates surprising results: ('foo', 'foo') instead of ('foo', 'not foo')

The issue is in https://github.com/getredash/redash/blob/d74442184e30a8aeb19edacb39e1ca22cbb523c1/redash/query_runner/presto.py#L87. From experimenting with other data sources, it looks like redash itself supports showing result sets that have duplicated column names.

cldellow:
Fixes #847

### DIFF
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -78,24 +78,12 @@ class Presto(BaseQueryRunner):
 
         cursor = connection.cursor()
 
-        # If a column appears more than once with the same name, subsequent
-        # instances are numbered 1, 2, , ... N.
-        columns_seen = defaultdict(int)
-        def make_unique(name):
-          columns_seen[name] += 1
-          if columns_seen[name] == 1:
-            return name
-          return name + str(columns_seen[name] - 1)
 
         try:
             cursor.execute(query)
-            columns_data = [(make_unique(row[0]), row[1]) for row in cursor.description]
-
-            columns = [{'name': col[0],
-                'friendly_name': col[0],
-                'type': PRESTO_TYPES_MAPPING.get(col[1], None)} for col in columns_data]
-
-            rows = [dict(zip(([c[0] for c in columns_data]), r)) for i, r in enumerate(cursor.fetchall())]
+            column_tuples = [(i[0], PRESTO_TYPES_MAPPING.get(i[1], None)) for i in cursor.description]
+            columns = self.fetch_columns(column_tuples)
+            rows = [dict(zip(([c[0] for c in column_tuples]), r)) for i, r in enumerate(cursor.fetchall())]
             data = {'columns': columns, 'rows': rows}
             json_data = json.dumps(data, cls=JSONEncoder)
             error = None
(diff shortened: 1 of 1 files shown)


cldellow:
My Python is pretty basic, if there's a cleaner way to achieve this, LMK. This mimics the behaviour of the Postgres query runner (which seems to get it for free from the psycopg2 library.)

redash/query_runner/presto.py
```diff
@@ -76,9 +78,18 @@ def run_query(self, query):
...
+        def make_unique(name):
+          columns_seen[name] += 1
+          if columns_seen[name] == 1:
+            return name
+          return name + str(columns_seen[name] - 1)
+
         try:
             cursor.execute(query)
-            columns_data = [(row[0], row[1]) for row in cursor.description]
+            columns_data = [(make_unique(row[0]), row[1]) for row in cursor.description]
```
arikfr: Actually you can use the [`fetch_columns` method](https://github.com/getredash/redash/blob/master/redash/query_runner/__init__.py#L74) this class inherits from `BaseQueryRunner`. 

In the case of Presto, it will be something like this:
```python
columns = self.fetch_columns([(i[0], PRESTO_TYPES_MAPPING.get(i[1], None)) for i in cursor.description])
```

As a side note, your implementation of deduplication logic is nicer than the one in the `fetch_columns` method, so feel free to replace it there. :+1:

cldellow: Thanks for pointing that method out, I've made the change.

I'm hesitant to change BaseQueryRunner's fetch_columns. Turns out my approach renumbers things differently than BaseQueryRunner (me: foo, foo1, bar, bar1, basequeryrunner: foo, foo1, bar, bar2). This renaming of columns would likely break visualizations that had stored the old name.

redash/query_runner/presto.py
```diff
@@ -76,15 +78,12 @@ def run_query(self, query):
...
-            columns_data = [(row[0], row[1]) for row in cursor.description]
-
-            columns = [{'name': col[0],
-                'friendly_name': col[0],
-                'type': PRESTO_TYPES_MAPPING.get(col[1], None)} for col in columns_data]
-
-            rows = [dict(zip(([c[0] for c in columns_data]), r)) for i, r in enumerate(cursor.fetchall())]
+            column_tuples = [(i[0], PRESTO_TYPES_MAPPING.get(i[1], None)) for i in cursor.description]
+            columns = self.fetch_columns(column_tuples)
+            rows = [dict(zip(([c[0] for c in column_tuples]), r)) for i, r in enumerate(cursor.fetchall())]
```
arikfr: Don't you need to use `columns` here (which has the deduplicated names)?

redash/query_runner/presto.py
arikfr: Good point on being consistent with existing results.

redash/query_runner/presto.py
cldellow: Ugh, yes. Thanks :) I was too focused on the column name being correct, and failed to notice that the actual data was now missing.

arikfr:
Thanks!

PR #850, opened 2016-02-24
Kibana: add support for limiting # of results returned

toyama-h:
Hello :blush: 

query_runner/elasticsearch.py included kibana.
kibana default limit  is 500.

size is 500, but this 「size」 is selected count.
「size」 is not limit.

default limit to 500.
We can define limit param, bellow.

```
{
  "index" : "logstash-2016.02.19",
  "query" : "_type: batchapp",
  "limit" : 3000
}
```

bellow is 6 query per 500.
```
{
  "index" : "logstash-2016.02.19",
  "query" : "_type: batchapp",
  "size" : 500,
  "limit" : 3000
}
```


arikfr:
Thanks!

PR #852, opened 2016-02-24
Small fixes for new ConfigurationContainer use

alexdebrie:
This contains two small fixes related to the new ConfigurationContainer from #844. 

TL;DR:
* Set the schema type before retrieving the options on a saved data source
* Don't require a `'secrets'` key on schemas.

The first was discussed in Slack -- basically retrieving an existing `DataSource` model with `options` failed because it didn't have a schema set. This gets the schema from the `type` attribute, passes it to the `ConfigurationContainer`, then grabs the config.

The second fix applies to converting `options` to a dict in the `ConfigurationContainer` class. If the schema doesn't have a `secrets` key, it just returns the options immediately without attempting to mask. This way we don't need to include an empty `'secrets'` key on every configuration schema.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -15,7 +15,7 @@ from flask.ext.login import UserMixin, AnonymousUserMixin
 from permissions import has_access, view_only
 
 from redash import utils, settings, redis_connection
-from redash.query_runner import get_query_runner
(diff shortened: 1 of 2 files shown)


arikfr:
Thanks!

PR #854, opened 2016-02-24
Minors fixes for MongoDB and Python query runners

erans:
- Python: added 'additionalModulesPaths' config key to add additional paths (comma separated) from which an import can occur. It still requires white listing the module in 'allowedImportModules'

- MongoDB - Added support for bson.timestamp.Timestamp JSON serialization as it sometimes gets returned in newer PyMongo versions.

### DIFF
--- a/redash/query_runner/python.py
+++ b/redash/query_runner/python.py
@@ -77,7 +77,7 @@ class Python(BaseQueryRunner):
         if self.configuration.get("additionalModulesPaths", None):
             for p in self.configuration["additionalModulesPaths"].split(","):
                 if p not in sys.path:
-                    sys.path.append(p) 
+                    sys.path.append(p)
 
     def custom_import(self, name, globals=None, locals=None, fromlist=(), level=0):
         if name in self._allowed_modules:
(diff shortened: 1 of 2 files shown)


redash/query_runner/python.py
```diff
@@ -144,7 +153,9 @@ def execute_query(self, data_source_name_or_id, query):
         except models.DataSource.DoesNotExist:
             raise Exception("Wrong data source name/id: %s." % data_source_name_or_id)
 
-        data, error = data_source.query_runner.run_query(query)
+        query_runner = get_query_runner(data_source.type, data_source.options)
```
arikfr: get_query_runner doesn't exist anymore.

erans: Really? When did it go? Is it in a specific branch?

I'll fix it.

erans:
Updated the code

redash/query_runner/python.py
erans: I've updated the code.

redash/query_runner/python.py
```diff
@@ -56,8 +60,8 @@ def enabled(cls):
     def annotate_query(cls):
         return False
 
-    def __init__(self, configuration):
-        super(Python, self).__init__(configuration)
+    def __init__(self, configuration_json):
+        super(Python, self).__init__(configuration_json)
```
arikfr: You reverted a change (it should be `configuration`).

PR #855, opened 2016-02-24
Docs: document settings.py

sreynen:
This is mostly just a template so far, as I'm only able to describe the settings I already understand, which isn't many.


arikfr:
Thanks!

PR #856, opened 2016-02-25
Feature: running queries monitor

arikfr:
First iteration of adding views that show currently executing queries, queries that waiting in queue and ones that recently finished. Also showing the "outdated queries" list (queries that are scheduled to refresh).

TODO:
- [x] Add periodic job to remove old finished tasks.
- [x] Cancel buttons.

re_dash___running_queries


toyama-h:
+1 +1 +1!!!!

arikfr:
There is still an issue here, that we inherit from Celery: if the worker are forcefully killed (`kill -9`) we won't detect this. I'm not sure how to handle this, aside from:
1. Adding some explanation around this in documentation.
2. Marking jobs as cancelled after 24 hours, assuming that no one is running such long queries.

PR #865, opened 2016-02-29
Replace deprecated flask.ext.* with flask_*

jeffwidman:
Importing flask extensions using flask.ext.* is deprecated in favor of flask_*
For background, see: https://github.com/mitsuhiko/flask/issues/1135


arikfr:
Thanks!

PR #867, opened 2016-02-29
Fix non-standard indentation to conventional four spaces

jeffwidman:
Not sure what happened in this section of code, but it was incorrectly indented by two spaces rather than four in several places.


arikfr:
:+1:

PR #877, opened 2016-03-01
Correct Ubuntu Bootstrap.sh to a working Base URL

ink-adavison:
The redash/docker/setup/... path is returning 404, but redash/master/setup/... works.


arikfr:
Thanks!

PR #878, opened 2016-03-01
List only dashboards available to current user based on user's groups

yanenok:
User can only see dashboards that he has access to in lists on the main page and in header. Whether user has access is determined by the data sources available to him through his groups. If a dashboard contains widgets accessing different data sources, user will be able to see this dashboard in list, however, inside dashboard he will only be able to view widgets that he has access to, others will be locked. Dashboards with no widgets or with text-only widgets do not have data sources but may contain private information that should not be shared across groups, therefore they are only visible to their creator.

### DIFF
--- a/redash/handlers/dashboards.py
+++ b/redash/handlers/dashboards.py
@@ -11,18 +11,18 @@ from redash.handlers.base import BaseResource, get_object_or_404
 
 class DashboardRecentAPI(BaseResource):
     def get(self):
-        recent = [d.to_dict() for d in models.Dashboard.recent(self.current_org, self.current_user.id)]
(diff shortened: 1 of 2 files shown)


arikfr:
> Dashboards with no widgets or with text-only widgets do not have data sources but may contain private information that should not be shared across groups, therefore they are only visible to their creator.

Disabling access to dashboards with only text widgets will break existing functionality and not something I can accept. I don't mind limiting access to dashboards with no widgets, but text widgets are public by default for now.

In the future we can implement more elaborate permissions for dashboards.

arikfr:
Closing in favor of #957 (it includes your commits too).

PR #882, opened 2016-03-04
CLI: Restore ability to pass JSON options string for a new data source

tsibley:
Commit "Encapsulate data source/query runner configuration in an
object." (ed99b84) accidentally removed that functionality by not
inflating the passed in JSON into a ConfigurationContainer object.

This led to errors of the form if you passed -o:

      ...
      File "/opt/redash/redash-git/redash/models.py", line 321, in db_value
	return value.to_json()
    AttributeError: 'unicode' object has no attribute 'to_json'


arikfr:
Thanks!

PR #883, opened 2016-03-04
Support for REMOTE_USER authentication

tsibley:
Via the environment or a proxy-added request header.  This allows
redash to use nearly any custom authentication setup.

Note that, like existing authentication providers, if you disable
password authentication and the remote user auth is unsuccessful
(probably due to a misconfiguration), then you'll get a redirect
loop of /login ⇄  /remote_user/login.

### DIFF
--- a/redash/authentication/remote_user_auth.py
+++ b/redash/authentication/remote_user_auth.py
@@ -1,0 +1,34 @@
+import logging
+from flask import redirect, url_for, Blueprint, request
+from redash.authentication.google_oauth import create_and_login_user
+from redash.authentication.org_resolving import current_org
+from redash import settings
+
+logger = logging.getLogger('remote_user_auth')
+
+blueprint = Blueprint('remote_user_auth', __name__)
+
+@blueprint.route("/remote_user/login")
+def login():
+    next_path = request.args.get('next')
+
+    if not settings.REMOTE_USER_LOGIN_ENABLED:
+        logger.error("Cannot use remote user for login without being enabled in settings")
+        return redirect(url_for('redash.index', next=next_path))
+
+    email = request.headers.get(settings.REMOTE_USER_HEADER)
+
+    # Some Apache auth configurations will, stupidly, set (null) instead of a
+    # falsey value.  Special case that here so it Just Works for more installs.
+    # '(null)' should never really be a value that anyone wants to legitimately
+    # use as a redash user email.
+    if email == '(null)':
+        email = None
+
+    if not email:
+        logger.error("Cannot use remote user for login when it's not provided in the request (looked in headers['" + settings.REMOTE_USER_HEADER + "'])")
+        return redirect(url_for('redash.index', next=next_path))
+
+    logger.info("Logging in " + email + " via remote user")
+    create_and_login_user(current_org, email, email)
+    return redirect(next_path or url_for('redash.index'), code=302)
--- a/redash/settings.py
+++ b/redash/settings.py
@@ -90,6 +90,32 @@ SAML_METADATA_URL = os.environ.get("REDASH_SAML_METADATA_URL", "")
 SAML_LOGIN_ENABLED = SAML_METADATA_URL != ""
 SAML_CALLBACK_SERVER_NAME = os.environ.get("REDASH_SAML_CALLBACK_SERVER_NAME", "")
 
+# Enables the use of an externally-provided and trusted remote user via an HTTP
+# header.  The "user" must be an email address.
+#
+# By default the trusted header is X-Forwarded-Remote-User.  You can change
+# this by setting REDASH_REMOTE_USER_HEADER.
+#
+# Enabling this authentication method is *potentially dangerous*, and it is
+# your responsibility to ensure that only a trusted frontend (usually on the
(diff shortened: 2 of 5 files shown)


redash/authentication/remote_user_auth.py
```diff
@@ -0,0 +1,34 @@
...
+
+@blueprint.route("/remote_user/login")
+def login():
+    next_path = request.args.get('next')
+
+    if not settings.REMOTE_USER_LOGIN_ENABLED:
+        logger.error("Cannot use REMOTE_USER for login without being enabled in settings")
+        return redirect(url_for('index', next=next_path))
+
+    email = request.environ.get("REMOTE_USER", request.headers.get(settings.REMOTE_USER_HEADER))
```
arikfr: Do you have a real use for authenticating with a `REMOTE_USER` env variable?

redash/settings.py
```diff
@@ -90,6 +90,27 @@ def all_settings():
...
+#
+# In proxy situations, it may be desirable to forward REMOTE_USER via a HTTP
+# request header, which you can specify by setting REDASH_REMOTE_USER_HEADER.
+# Doing so is *potentially dangerous*, and it is your responsibility to ensure
+# that only a trusted frontend (usually on the same server) can talk to the
+# redash backend server, otherwise people will be able to login as anyone they
+# want by directly talking to the redash backend.
+#
+# If you set REDASH_PASSWORD_LOGIN_ENABLED to false, then your authentication
+# will be seamless.  Otherwise a link will be presented on the login page to
```
arikfr: > If you set REDASH_PASSWORD_LOGIN_ENABLED to false, then your authentication will be seamless.

This doesn't seem to be accurate. On line 15 of `remote_user_auth.py` you redirect the user to the index page (which will result in redirect to the login page). But this seems to be the right thing to do.

redash/settings.py
```diff
@@ -90,6 +90,27 @@ def all_settings():
...
+# Note that redash will only check REMOTE_USER once, upon the first need for a
+# login, and then set a cookie which keeps the user logged in.  Dropping
+# REMOTE_USER after subsequent requests won't automatically log the user out.
+# Doing so could be done with further work, but usually it's unnecessary.
+#
+# In proxy situations, it may be desirable to forward REMOTE_USER via a HTTP
+# request header, which you can specify by setting REDASH_REMOTE_USER_HEADER.
+# Doing so is *potentially dangerous*, and it is your responsibility to ensure
+# that only a trusted frontend (usually on the same server) can talk to the
+# redash backend server, otherwise people will be able to login as anyone they
```
arikfr: It's not only making sure that a trusted frontend can talk to the redash server, but that this frontend filters the authentication header from external requests.

redash/authentication/remote_user_auth.py
tsibley: No, I'm using a header.  I added it for completeness.  I imagine someone could be using mod_wsgi, or another server instead of gunicorn, that would provide `REMOTE_USER` through the environment, although I realize you may not be interested in supporting that setup.  I don't have a vested interest in keeping the env check if you want to remove it.

redash/settings.py
tsibley: Yes, that's very true.  I'll edit the comment to that effect.  I'd glossed over that detail since I was using a frontend which always sets the header (overwriting any provided in the original request).  Thanks for the catch!

redash/settings.py
tsibley: The redirect to the index page on line 15 is conditionalized on `not settings.REMOTE_USER_LOGIN_ENABLED`.  Perhaps we're using a different idea of seamless?  I mean only that the user won't be prompted to click any link to use remote user auth.  That is, under:

    settings.PASSWORD_LOGIN_ENABLED = False;
    settings.REMOTE_USER_LOGIN_ENABLED = True;

then requests needing login will redirect to `/login`, which will redirect to `/remote_user/login`, which will perform the login and redirect the client back to the original page (or the index, if no `next_page`).

Am I missing something?

redash/settings.py
tsibley: Added!

tsibley:
I should note that I've been using this in production and it's been working fine.

arikfr:
Looking good and I appreciate the detailed comment with details on how this works.

Please remove the env variable support - I doubt anyone will run Re:dash in a setting where it will make sense. (also you need to rebase)

I appreciate your cooperation and patience with this pull request.

tsibley:
I removed environment support and rebased the branch onto the latest master.

redash/authentication/remote_user_auth.py
```diff
@@ -0,0 +1,34 @@
...
+
+blueprint = Blueprint('remote_user_auth', __name__)
+
+@blueprint.route("/remote_user/login")
+def login():
+    next_path = request.args.get('next')
+
+    if not settings.REMOTE_USER_LOGIN_ENABLED:
+        logger.error("Cannot use remote user for login without being enabled in settings")
+        return redirect(url_for('index', next=next_path))
```
arikfr: Sorry, I overlooked it in previous review, but this should be `redash.index` now.

arikfr:
Merged. Thanks!

PR #888, opened 2016-03-07
Improve TreasureData query runner

toru-takahashi:
ChangeLog

1. Update td-client version to v0.4.1
2. Add new option to enable auto get schema on query editor.
    - This option avoid slow page loading when a number of table in TreasureData Database is too many. 
3. Fix case sensitive issue for query type.
    - This fix allow upper-case for query type.

### DIFF
--- a/redash/query_runner/treasuredata.py
+++ b/redash/query_runner/treasuredata.py
@@ -48,9 +48,14 @@ class TreasureData(BaseQueryRunner):
                 'type': {
                     'type': 'string'
                 },
-                "db": {
-                    "type": "string",
-                    "title": "Database Name"
+                'db': {
+                    'type': 'string',
+                    'title': 'Database Name'
+                },
+                'getschema': {
+                    'type': 'boolean',
+                    'title': 'Auto Get Schema',
+                    'Default': True
                 }
             },
             'required': ['apikey','db']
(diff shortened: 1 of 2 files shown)


redash/query_runner/treasuredata.py
```diff
@@ -48,9 +48,14 @@ def configuration_schema(cls):
...
-                    "type": "string",
-                    "title": "Database Name"
+                'db': {
+                    'type': 'string',
+                    'title': 'Database Name'
+                },
+                'getschema': {
+                    'type': 'boolean',
+                    'title': 'Auto Get Schema',
+                    'Default': True
```
arikfr: It should be `default` (not `Default`).

redash/query_runner/treasuredata.py
```diff
@@ -71,23 +76,24 @@ def type(cls):
...
         schema = {}
-        try:
-            with tdclient.Client(self.configuration.get('apikey')) as client:
-                for table in client.tables(self.configuration.get('db')):
-                    table_name = '{}.{}'.format(self.configuration.get('db'), table.name)
-                    for table_schema in table.schema:
-                        schema[table_name] = {'name': table_name, 'columns': table.schema}
-        except Exception, ex:
-            raise Exception("Failed getting schema")
+        if self.configuration.get('getschema') == True:
```
arikfr: The `default` property in the schema is just a hint for the UI, it doesn't set actual values. So in your case change this line to: `if self.configuration.get('getschema', True):` to have `True` as default value.

redash/query_runner/treasuredata.py
```diff
@@ -48,9 +48,14 @@ def configuration_schema(cls):
...
                     'type': 'string'
                 },
-                "db": {
-                    "type": "string",
-                    "title": "Database Name"
+                'db': {
+                    'type': 'string',
+                    'title': 'Database Name'
+                },
+                'getschema': {
```
arikfr: Please change to `get_schema` for consistency.

arikfr:
Thanks! See comments.

toru-takahashi:
Thank you for your review. I'll fix it.

toru-takahashi:
I have a question about this point.
> The default property in the schema is just a hint for the UI, it doesn't set actual values.

When I change nothing in checkbox, `self.configuration.get('get_schema')` returns `None`, not `False`.



Is it a expected behavior?

If so, I'll rewrite from `if self.configuration.get('get_schema', True):` to `if self.configuration.get('get_schema', False):`

arikfr:
> When I change nothing in checkbox, self.configuration.get('get_schema') returns None, not False.

When you change nothing, then `get_schema` has no value yet in the dictionary and the `get` method returns `None` in such case. When you pass the second argument to `get`, it will be the value it returns in case no value is defined yet.

But it looks like the UI for checkboxes doesn't respect the default value so maybe it's better to have this setting to disable getting schema rather than enable. So it becomes "Disable Schema Retrieval" with the default of false.

toru-takahashi:
I fixed the matter pointed out by you. Could you review it again?

arikfr:
Looking good. :+1:

toru-takahashi:
Thanks!

PR #891, opened 2016-03-07
WIP: Use 'latest' tag on postgresql docker image

jeffwidman:
This is an example file, so no need to pin to an old Postgresql version

### DIFF
--- a/docker-compose-example.yml
+++ b/docker-compose-example.yml
@@ -15,7 +15,7 @@ redash:
 redis:
   image: redis:2.8
 postgres:
-  image: postgres:9.3
+  image: postgres:latest
   volumes:
    - /opt/postgres-data:/var/lib/postgresql/data
 redash-nginx:


jeffwidman:
Actually, I'm getting this error: 
    postgres_1     | LOG:  skipping missing configuration file "/var/lib/postgresql/data/postgresql.auto.conf"
    postgres_1     | FATAL:  database files are incompatible with server
    postgres_1     | DETAIL:  The data directory was initialized by PostgreSQL version 9.3, which is not compatible with this version 9.5.1.

I'm not understanding where these PG data files are created... I assumed they were created using DB commands so it would work fine with a newer PG version... is this 9.3 data file hardcoded in a setup script somewhere?

arikfr:
We intentionally use version 9.3, as this is the version that most users have. Even if 9.4/9.5 are fully backward compatible, the "danger" of using newer version is that we will use features version 9.3 doesn't support (which will break existing environments, when upgrading).

Either way, I don't think it's right to use the latest tag, because then you can't control which version you will get. This will result in different users running different versions of Postgres, which is a nightmare to support or debug issues. So even if we decide to upgrade, let's pin to a specific version.

If you still think we should upgrade to newer version, let's continue the discussion. But closing the pull request for now.

arikfr:
The setup doesn't come with any PG related files hardcoded. Maybe your environment started once with PG 9.3 and only after you upgraded to the latest? Try deleting the volume folder.

PR #892, opened 2016-03-07
Switch to a better redis image for docker compose

jeffwidman:
I don't think we need to be back on Redis 2. 

When I went to update, I noticed there's also an optimal slimmed-down build using Alpine linux. See info here: https://hub.docker.com/_/redis/
I stuck with `latest` for simplicity, plus it's not clear in the how well maintained Alpine is: https://news.ycombinator.com/item?id=10782897
But switching to the alpine image might be an optional improvement here.


arikfr:
Same reasoning as with the  Postgres version - https://github.com/getredash/redash/pull/891#issuecomment-193624797.

PR #894, opened 2016-03-07
Dockerfile: use redash's docker-hub nginx image

jeffwidman:
This is not fully workable yet, as running `docker-compose up` throws:

    ~/C/redash ❯❯❯ docker-compose up
    redash_redis_1 is up-to-date
    redash_postgres_1 is up-to-date
    redash_redash_1 is up-to-date
    Starting redash_redash-nginx_1
    ERROR: Cannot start container e3577332ca4ea334daa78012fbc14d1331fec966c1bb621b4a131faf7aea3373: [9] System error: not a directory

@arikfr any idea what's going on here? My redash image was created locally with `docker build -t redash .` using the default `Dockerfile` in the repo...

PR #895, opened 2016-03-07
Use redash's docker-hub redash image

jeffwidman:
I wasn't sure whether to add this as an inline comment or a change, but decided to use the remote image for two reasons:

1. People who are new to docker and just trying to get up and running quickly won't realize that they need to build the docker image locally. Conversely, anyone comfortable with docker will already know how to switch this to a local image built using the Dockerfile.

2. This is more consistent with #894

### DIFF
--- a/docker-compose-example.yml
+++ b/docker-compose-example.yml
@@ -1,5 +1,5 @@
 redash:
-  image: redash
+  image: redash/redash:latest
   ports:
     - "5000:5000"
   links:


arikfr:
The question is whether the user wants to run his local version or just an environment of redash. Anyway the compose file wasn't defined to build locally, so your change is better.

Thanks :+1:

PR #898, opened 2016-03-08
Switch to docker-compose version 2 format

jeffwidman:
https://docs.docker.com/compose/compose-file/#upgrading

### DIFF
--- a/docker-compose-example.yml
+++ b/docker-compose-example.yml
@@ -1,28 +1,30 @@
-redash:
-  image: redash
-  ports:
-    - "5000:5000"
-  links:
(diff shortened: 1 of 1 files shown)


arikfr:
> Version 2 files are supported by Compose 1.6.0+ and require a Docker Engine of version 1.10.0+.

Same as with the networking change - it's too new for now and doesn't offer real benefits (right?).

jeffwidman:
That's fine. Additionally, many hosting services that support docker may not support this `version 2` format for a bit, so another reason to hold off on doing this.

PR #910, opened 2016-03-10
Better full screen support

arikfr:
- Remove header 
- Add some padding


arikfr:
@matthew-sochor FYI.

matthew-sochor:
awesome :+1:

The thought of removing the header occurred to me but I never tried to get it to work.  Thanks for this!

PR #911, opened 2016-03-10
Docs: Docker compose instructions

jkatsnelson:


jeffwidman:
:+1: 
Really appreciate this @jkatsnelson I was banging my head against this earlier this week trying to figure it out.

docs/setup.rst
```diff
@@ -63,6 +63,19 @@ BigQuery access)
...
+Docker Compose
+------
+
+1. Make sure you have a docker machine up and running.
+2. Make sure your current working directory is the root of this github repository.
+3. Run `docker-compose up postgres`
+4. Run `./setup/docker/create_database.sh` (chmod +x this file if necessary).
+    * This will access the postgres container and set up the database.
+5. Run `docker compose up`
+6. Run `docker-compose ls`, take note of the ip for the docker machine you are using, and open the web browser.
```
jeffwidman: Shouldn't this be `ps` rather than `ls`? The latest version of `docker-compose` doesn't have an `ls` command...

jkatsnelson: Thanks!

jeffwidman:
These instructions work great for me. Hopefully @arikfr chooses to merge them.

xingh:
Mostly works on Windows. Can't run the bash script so I ran the commands in the bash script manually.. had to make sure to add -d to run commands , and escape the metadata in the last command. 

I may make a Powershell version if it doesn't work on Cygwin bash / Git bash

arikfr:
As @xingh commented on Gitter, we shouldn't rely on external bash script but rather encapsulate everything as commands in the Docker container. Originally there was a reason for having a bash script, but after revisting it, I see that it can all be a single entrypoint in the Redash docker container.

arikfr:
@jkatsnelson sorry for keeping this open for so long. I will merge it right before closing v0.10 (should happen soon).

jeffwidman:
> we shouldn't rely on external bash script but rather encapsulate everything as commands in the Docker container. Originally there was a reason for having a bash script, but after revisting it, I see that it can all be a single entrypoint in the Redash docker container.

True, but I'd hate to see that hold up this PR. Rather see this merged since it's factually correct right now and better than the current docs state. Easy enough to update it later as the process changes.

arikfr:
Merged 🎉 
Thank you @jkatsnelson and sorry for the long wait.

jkatsnelson:
👍 😄

PR #914, opened 2016-03-11
Adjust the UI on the New Query Page

aleesley:
The results table is full width.
Query info has been adjusted to full width below the query to allow for more room for the query results.
The schema browser has been moved to the left side, with visual adjustments to the expand and collapse.


arikfr:
I've been working practically on the same thing last night.

Re:dash design is going to get a major upgrade in the coming weeks -- 
1. I'm planning to make all the screens full width, including the query page but also the dashboards and others.
2. There is a designer coming on board that will help to make it look better.

If you're into improving the design and UX as well, I'll be happy to collaborate on this.

PR #919, opened 2016-03-13
Fix: schema fetching for Vertica & SQLite was broken (fix #916)

### LINKED ISSUE #916: Vertica doesn't show schema
Vertica query runner returns error due to incompatible get_schema method

alexanderlz:
The bug existed in sqlite also, fixed both.


arikfr:
Thanks.

PR #920, opened 2016-03-13
Feature: MongoDB - add support for $date and relative dates.

arikfr:
This pull request adds two imporvements to our MongoDB query parsing:

* Full support for [MongoDB Extended JSON](https://docs.mongodb.org/manual/reference/mongodb-extended-json/), which has [`$date`](https://docs.mongodb.org/manual/reference/mongodb-extended-json/#data_date) that can replace our `ISODate` implementation.
* Relative dates support:  passing an object of the form: `{"$relative_ts": "one hour ago"}` will result in a `datetime.datetime` object that represents one hour ago at the time of the query execution.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,6 +11,7 @@ dependencies:
   pre:
     - pip install -r requirements_dev.txt
     - pip install -r requirements.txt
+    - pip install pymongo==3.2.1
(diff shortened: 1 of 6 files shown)


arikfr:
@erans wdyt?

erans:
Looks good. It's actually a cool that Mongo doesn't have (yet) :-)

PR #924, opened 2016-03-14
Endless loop query.

toyama0919:
Do not update query_result.retrieved_at.
if not exists query by query_hash and data_source_id, query.latest_query_data_id update by query_id.

We exists wrong query_hash and query.

If exists Another solution, please help me.

ps.
We has 600 query, these query modify GUI for very tired.
We want query.query update for sql.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -487,7 +487,7 @@ class QueryResult(BaseModel, BelongsToOrgMixin):
         return query.first()
 
     @classmethod
-    def store_result(cls, org_id, data_source_id, query_hash, query, data, run_time, retrieved_at):
+    def store_result(cls, org_id, data_source_id, query_hash, query, data, run_time, retrieved_at, query_id = 'adhoc'):
         query_result = cls.create(org=org_id,
                                   query_hash=query_hash,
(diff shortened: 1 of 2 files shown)


arikfr:
I'm not sure I fully understand, but let's see -- you mean you have some queries that have a different `query_hash` between results and query? How is that possible?

I think that it will be better for you to understand how you got to his situation and fix it, as it might introduce issues in other places as well. 

Also, note that every time you save a query it recalcualtes `query_hash`, so it's enough to save a query once again to resolve this.

toyama0919:
>How is that possible?

Perhaps, query update by sql.


> Also, note that every time you save a query it recalcualtes query_hash, so it's enough to save a query once again to resolve this.

Ah, Model's callback(pre_save)?
I will try Model's save return ids.
It currently uses a sql.

arikfr:
> Perhaps, query update by sql.

Why did you have to do it?

> Ah, Model's callback(pre_save)?

Yes. You can write a simple loop, that loops all the Query objects and calls `save()` on them to trigger `query_hash` calculation.

PR #932, opened 2016-03-16
Adjust the UI on the New Query page

alexdebrie:
This PR makes a few design changes to the New Query page. 

A summary:

* Schema explorer is moved to the left of query box for better workflow.
* Schema explorer is expandable if needed
* Tables don't automatically expand while searching. They expand on click only.
* Query results are put in own row to provide a wider area for results.


arikfr:
Cool! Can you add a screenshot of what the end result looks like?

alexdebrie:
Sure!

screen shot 2016-03-16 at 8 56 42 am

arikfr:
I've added some changes and merged. Thanks!

PR #936, opened 2016-03-17
Filter enhancements

tsibley:
Two small enhancements to filters in the dashboard and on the query page.

### DIFF
--- a/rd_ui/app/scripts/controllers/query_view.js
+++ b/rd_ui/app/scripts/controllers/query_view.js
@@ -230,12 +230,20 @@
       $scope.$parent.pageTitle = $scope.query.name;
     });
 
+    var firstLoad = true;
     $scope.$watch('queryResult && queryResult.getData()', function(data, oldData) {
       if (!data) {
         return;
       }
 
       $scope.filters = $scope.queryResult.getFilters();
+      if (firstLoad) {
+        _.each($scope.filters, function (queryFilter) {
+          if (_.has($location.search(), queryFilter.name))
+            queryFilter.current = $location.search()[queryFilter.name];
+        });
+        firstLoad = false;
+      }
     });
 
     $scope.$watch("queryResult && queryResult.getStatus()", function(status) {
--- a/rd_ui/app/views/dashboard.html
+++ b/rd_ui/app/views/dashboard.html
@@ -43,7 +43,7 @@
                     </h3>
                 </div>
 
-                <visualization-renderer visualization="widget.visualization" query-result="queryResult"></visualization-renderer class="panel-body">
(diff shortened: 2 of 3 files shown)


rd_ui/app/views/dashboard.html
```diff
@@ -42,7 +42,7 @@ <h3 class="panel-title">
                     </h3>
                 </div>
 
-                <visualization-renderer visualization="widget.visualization" query-result="queryResult"></visualization-renderer class="panel-body">
+                <visualization-renderer visualization="widget.visualization" query-result="queryResult" hide-filters="dashboard.dashboard_filters_enabled"></visualization-renderer class="panel-body">
```
arikfr: The problem with this is that now you don't know if a widget is filtered or not. Also, sometimes the user might like to select different values on different widgets (that's actually less of a concern).

rd_ui/app/scripts/controllers/query_view.js
```diff
@@ -230,12 +230,20 @@
...
     $scope.$watch('queryResult && queryResult.getData()', function(data, oldData) {
       if (!data) {
         return;
       }
 
       $scope.filters = $scope.queryResult.getFilters();
+      if (firstLoad) {
+        _.each($scope.filters, function (queryFilter) {
+          if (_.has($location.search(), queryFilter.name))
+            queryFilter.current = $location.search()[queryFilter.name];
```
arikfr: If we're reading values from the URL, we should also write them to the URL to keep them in sync.
I think we have this functionality on dashboards - maybe we can share the code?

tsibley: As far as I can tell, only searching for queries updates the URI params.

However, I agree that updating it is better than not.  I'll look into it.

rd_ui/app/views/dashboard.html
tsibley: Fair enough. I expected that it wouldn't be necessary to know which queries are filtered or not, but making it clear is a good idea. How about if the widget still indicates the filter state of itself without including all the bulky filter UI repeated over and over?

arikfr: > How about if the widget still indicates the filter state of itself without including all the bulky filter UI repeated over and over?

Sounds good, what do you suggest? :)

tsibley: Basically showing a "**\<Name\>:** \<Value\>" pair in text for each dashboard-wide filter applying to a widget in place of where the inputs normally go.  It's really the styling of the thing that's cumbersome when dashboard-wide filters are used, and plain text is easier to style concisely and compactly than the input elements that are used normally.

PR #937, opened 2016-03-19
Add hide_schema option to data sources to disable schema sidebar

benmanns:
I'm not sure if this is the best place to put this, but it worked for me as a test of the feature.

One of the database servers I am testing Re:dash has several thousand tables with many columns each. When using the query editor, the schema tool in the right side slows down the page considerably and consumes a bunch of RAM.

This adds the ability to update a data source and add the `hide_schema` setting (`False` by default) to return an empty array from the schema API call.

### DIFF
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -102,7 +102,10 @@ class DataSourceSchemaResource(BaseResource):
     def get(self, data_source_id):
         data_source = get_object_or_404(models.DataSource.get_by_id_and_org, data_source_id, self.current_org)
         require_access(data_source.groups, self.current_user, view_only)
-        schema = data_source.get_schema()
+        if data_source.configuration.get('hide_schema', False):
+            schema = []
+        else:
+            schema = data_source.get_schema()
 
         return schema
 


arikfr:
I guess we should switch to storing the schema, and have a proper API for searching/browsing it to handle such cases.

I'm cool with merging this, but can you add a comment explaining this in the code?
Also, maybe it will be better to move the `if` into the `get_schema` method of the `DataSource` model, to avoid wasting  time on fetching the schema in the first place.

arikfr:
Ping?

arikfr:
Feels abandoned - closing.

a-grealish:
I have another use case for this feature which is not covered by https://github.com/getredash/redash/pull/1246. I would like to provide access to a read-only view for people to view dashboards, queries and download data. However, with a read-only view the users can click 'Edit Source' and view the SQL (not ideal but I could deal with this) but also view the entire scheme of our database. Perhaps a good solution is to revoke permission to view to 'edit source' screen (I can raise a new issue for this if you want) but this feature would help me address some issues we have with giving people read-only access.

arikfr:
@a-grealish please note that even if the UI doesn't show the SQL, most of the APIs will likely to show them at the moment (we do want to address this at some point).

PR #940, opened 2016-03-23
Nginx configuration: hide version

kirikiriyamama:
Hi, I think that nginx version should be hidden for security if we have no reason. What about this?


arikfr:
Thanks!

PR #945, opened 2016-03-25
Update color palette to add more pretty colors

ChiragKParmar:
Inspired from ggplot2 color palette, update ColorPalette to make charts
more pretty.


arikfr:
Thanks!

I merged this (d6a474b9f1c755fb8bc250080edecf3a8a007082), but because of the way I did it, it doesn't show as merged here...

I also added a small modification: the new colors are added as colors the user can pick from, but not applied automatically. This to prevent from chart colors changing for existing users.

PR #957, opened 2016-03-29
Hide dashboards that current user cannot see based on his groups

gseva:
This pr is based on https://github.com/getredash/redash/pull/878

The difference is that text widgets are public. And i also added some tests.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -886,14 +886,15 @@ class Dashboard(ModelTimestampsMixin, BaseModel, BelongsToOrgMixin):
     @classmethod
     def all(cls, groups, user_id):
         query = cls.select().\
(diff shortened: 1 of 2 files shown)


arikfr:
I've added a few more tests and merged. Thank you!

Btw, I'm wondering if we should enforce this when trying to access the dashboard directly (with the URL)?

arikfr:
@yanenok note that this includes your work as well, so thank you both!

gseva:
@arikfr 
Thanks for merging!
Regarding the direct access, i think it makes sense to redirect to or show nothing at all. But actually it wouldn't show data in that case, so i guess it's ok

arikfr:
The title might leak information, but I guess it's not that big of a deal. Will leave as is for now.

PR #962, opened 2016-03-30
InfluxDB: Added support for multiple queries and tags

groveriffic:
Multiple queries are separated by semicolons and could possibly have
different columns and tags.
https://docs.influxdata.com/influxdb/v0.11/guides/querying_data/#multiple-queries

Tags come in to the result set when you use a GROUP BY statement.

### DIFF
--- a/redash/query_runner/influx_db.py
+++ b/redash/query_runner/influx_db.py
@@ -19,16 +19,30 @@ def _transform_result(results):
     result_rows = []
 
     for result in results:
-        if not result_columns:
-            for c in result.raw['series'][0]['columns']:
-                result_columns.append({ "name": c })
+        for series in result.raw['series']:
+            for column in series['columns']:
+                if column not in result_columns:
+                    result_columns.append(column)
+            for key in series['tags'].keys():
+                if key not in result_columns:
+                    result_columns.append(key)
 
-        for point in result.get_points():
-            result_rows.append(point)
+    for result in results:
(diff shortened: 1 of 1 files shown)


arikfr:
Nice! Does it change the minimal version of Influxdb that we support?

groveriffic:
I've been working against InfluxDB version 0.10.0.  I didn't see any info on minimum version support, could you point me in the right direction?

I'm also noticing that I've broken support for result sets *without* tags.  I should have that fixed shortly.

redash/query_runner/influx_db.py
```diff
@@ -19,16 +19,32 @@ def _transform_result(results):
...
-            for c in result.raw['series'][0]['columns']:
-                result_columns.append({ "name": c })
+        for series in result.raw.get('series', []):
+            for column in series['columns']:
+                if column not in result_columns:
+                    result_columns.append(column)
+            tags = series.get('tags', {})
+            for key in tags.keys():
+                if key not in result_columns:
+                    result_columns.append(key)
```
arikfr: Can't there by name collision between column names and tag names?

groveriffic: Yes, that's possible and it causes problems when attempting to query based on the tag.

See also:
https://docs.influxdata.com/influxdb/v0.11/concepts/schema_and_data_layout/
bottom of the page.

arikfr:
Merged. Thanks!

PR #964, opened 2016-03-30
Allow passing branch name to bootstrap.sh

ink-adavison:
This ensures that it is possible to checkout a tag or branch and install from that tag using the same versioned files, for reliable repeatable builds.

e.g.

`bootstrap.sh`
will still install from the base url 
https://raw.githubusercontent.com/getredash/redash/master/setup/ubuntu/files/
but
```
git checkout v0.9.2.b1536
bootstrap.sh v0.9.2.b1536
```
will install from the base url:
https://raw.githubusercontent.com/getredash/redash/v0.9.2.b1536/setup/ubuntu/files/postgres_apt.sh

### DIFF
--- a/setup/amazon_linux/bootstrap.sh
+++ b/setup/amazon_linux/bootstrap.sh
@@ -8,7 +8,7 @@ REDASH_BRANCH="${REDASH_BRANCH:-master}"
 
 # Install latest version if not specified in REDASH_VERSION env var
 REDASH_VERSION=${REDASH_VERSION-0.9.2.b1536}
-LATEST_URL="https://github.com/getredash/redash/releases/download/${REDASH_BRANCH}/redash.${REDASH_VERSION}.tar.gz"
+LATEST_URL="https://github.com/getredash/redash/releases/download/v${REDASH_VERSION}/redash.${REDASH_VERSION}.tar.gz"
 VERSION_DIR="/opt/redash/redash.${REDASH_VERSION}"
 REDASH_TARBALL=/tmp/redash.tar.gz
 
--- a/setup/amazon_linux/bootstrap.sh
+++ b/setup/amazon_linux/bootstrap.sh
@@ -2,7 +2,17 @@
 set -eu
 
 REDASH_BASE_PATH=/opt/redash
-FILES_BASE_URL=https://raw.githubusercontent.com/getredash/redash/master/setup/amazon_linux/files/
+
+# Default version to master
+version=master
+
+# If a version is specified on the command line, use it instead
+if [[ -n "$1" ]]; then
+  version=$1
+fi
+
+FILES_BASE_URL=https://raw.githubusercontent.com/getredash/redash/$version/setup/amazon_linux/files/
+
 # Verify running as root:
 if [ "$(id -u)" != "0" ]; then
     if [ $# -ne 0 ]; then
--- a/setup/amazon_linux/bootstrap.sh
+++ b/setup/amazon_linux/bootstrap.sh
@@ -3,15 +3,16 @@ set -eu
 
 REDASH_BASE_PATH=/opt/redash
 
-# Default version to master
-version=master
+# Default branch/version to master if not specified in REDASH_BRANCH env var
+REDASH_BRANCH="${REDASH_BRANCH:-master}"
 
-# If a version is specified on the command line, use it instead
-if [[ -n "$1" ]]; then
-  version=$1
-fi
+# Install latest version if not specified in REDASH_VERSION env var
+REDASH_VERSION=${REDASH_VERSION-0.9.2.b1536}
+LATEST_URL="https://github.com/getredash/redash/releases/download/${REDASH_BRANCH}/redash.${REDASH_VERSION}.tar.gz"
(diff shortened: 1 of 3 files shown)


arikfr:
I'm not sure it's worth the hassle. The files it pulls are mainly configuration, which doesn't change much between versions. 

Also it's confusing, as I would expect it to install the version specified, which it won't do. The installed version is controlled via the `REDASH_VERSION` env variable. So to really install b1536 you will need to type

```
REDASH_VERSION=v0.9.2.b1536 bootstrap.sh v0.9.2.b1536
```

If someone needs a reliable environment for an older version, I recommend using the relevant Docker image.

ink-adavison:
We've had a couple of occasions where using a static base path has broken our builds. The first was when this path pointed at a docker branch which was deleted and then again when a broken nginx config was committed to master.

This doesn't break any existing behaviour as it still defaults to master and adds flexibility for those who would like to use configurations and scripts from a previously known working version or an experimental branch etc.

I've submitted this as a nice to have, but of course there may be some better method to this.

On 31 Mar 2016, at 08:41, Arik Fraimovich <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm not sure it's worth the hassle. The files it pulls are mainly configuration, which doesn't change much between versions.

Also it's confusing, as I would expect it to install the version specified, which it won't do. The installed version is controlled via the REDASH_VERSION env variable. So to really install b1536 you will need to type

REDASH_VERSION=v0.9.2.b1536 bootstrap.sh v0.9.2.b1536


If someone needs a reliable environment for an older version, I recommend using the relevant Docker image.

-
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub<https://github.com/getredash/redash/pull/964#issuecomment-203801512>


________________________________
This e-mail (including any attachments) is intended only for the recipient(s) named above. It may contain confidential and/or privileged information and should not be read, copied or otherwise used by any other person. If you are not a named recipient, please notify tech.support@ink-global.com and permanently delete the original message. Any views or opinions expressed in this message are those of the author and do not necessarily represent those of Ink. No reliance may be placed on this message without written confirmation from an authorised representative of its contents. We reserve the right to monitor all e-mail messages passing through our network. Please note that calls to and from Ink may be recorded for the purposes of training and development, dispute resolution and to evidence business transactions.

eSubstance Limited t/a Ink is a limited company registered in England and Wales. Registered number: 3829537. VAT No.: GB 907988664. Registered office: Blackburn House, Blackburn Road, West Hampstead, London, NW6 1RZ, United Kingdom.
________________________________

arikfr:
Fair enough. How about then we use this parameter for the tarball path as well (when it's not `master`)?

ink-adavison:
Hmm, maybe this would be better as an environment variable too, rather than a command line parameter. Keep it consistent.

Unfortunately, I'm not at a computer till next week, so I'll revisit this then.

On 31 Mar 2016, at 13:28, Arik Fraimovich <notifications@github.com<mailto:notifications@github.com>> wrote:


Fair enough. How about then we use this parameter for the tarball path as well (when it's not master)?

-
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub<https://github.com/getredash/redash/pull/964#issuecomment-203906903>


________________________________
This e-mail (including any attachments) is intended only for the recipient(s) named above. It may contain confidential and/or privileged information and should not be read, copied or otherwise used by any other person. If you are not a named recipient, please notify tech.support@ink-global.com and permanently delete the original message. Any views or opinions expressed in this message are those of the author and do not necessarily represent those of Ink. No reliance may be placed on this message without written confirmation from an authorised representative of its contents. We reserve the right to monitor all e-mail messages passing through our network. Please note that calls to and from Ink may be recorded for the purposes of training and development, dispute resolution and to evidence business transactions.

eSubstance Limited t/a Ink is a limited company registered in England and Wales. Registered number: 3829537. VAT No.: GB 907988664. Registered office: Blackburn House, Blackburn Road, West Hampstead, London, NW6 1RZ, United Kingdom.
________________________________

arikfr:
Env var sounds good to me, let's name it `BRANCH` and update the references in code to `branch` as well, to make sure people aren't confused.

ink-adavison:
Ok, I went with `REDASH_BRANCH` instead of just `BRANCH`, to be consistent with `REDASH_VERSION` (but can go with `BRANCH` if you prefer)

This means a previous release (well, from the point this is merged onwards) can be installed reliably by running:

    git checkout v0.9.2.b1536
    env REDASH_BRANCH=v0.9.2.b1536 REDASH_VERSION=0.9.2.b1536 ./bootstrap.sh

Note that Branch includes the "v" while Version omits it. It would be possible to merge these variables as one, but then that would remove the flexibility of being able to specify an experimental feature branch based of a specific or current version etc.

Of course running bootstrap.sh on its own without any branch or version still behaves exactly as before.

I also moved the REDASH_VERSION declaration to the top of the bootstrap files as I noticed it defaults to a specific version that needs to be updated whenever there's a new release and this hadn't been done. With it up there at the top of the file it's more noticeable to anyone casually glancing through looking for reasons why they had an unexpected outcome to an installation.

Does this seem cleaner and easier to follow?

arikfr:
Looking good. Thanks!

PR #966, opened 2016-03-31
Fix dashboard permissions, if user doesn't have access to any of the …

machira:
Fix dashboard permissions, if user doesn't have access to any of the datasources used in the widgets (except text widget) of a dashboard then do not show that dashboard to the user.

This PR shows users only the Dashboards with at least one of the widgets accessible to the user (based on the datasource used in the widget). If a Dashboard contains any text widget, and then that Dashboard is also visible to the user.

### DIFF
--- a/redash/handlers/dashboards.py
+++ b/redash/handlers/dashboards.py
@@ -11,11 +11,11 @@ from redash.handlers.base import BaseResource, get_object_or_404
 class RecentDashboardsResource(BaseResource):
     @require_permission('list_dashboards')
     def get(self):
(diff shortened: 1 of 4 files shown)


arikfr:
Isn't this a duplicate of #966 and #878?

machira:
@arikfr Sorry, we missed #957. Feel free to close (assuming you will be merging #957). 
Btw. we included `pymongo` in `requirements_dev.txt`.

arikfr:
Closing in favor of #957.

PR #967, opened 2016-04-01
Add: extend ElasticSearch query_runner to support aggregations

lloydw:


redash/query_runner/elasticsearch.py
```diff
@@ -132,6 +130,48 @@ def add_column_if_needed(mappings, column_name, friendly_name, result_columns, r
                     "type" : mappings.get(column_name, "string")})
                 result_columns_index[friendly_name] = result_columns[-1]
 
+        def getRow(rows, row):
```
arikfr: We're using [PEP8](https://www.python.org/dev/peps/pep-0008/) as a style guide. The TL;DR is that we use `snake_case` for names (~~`getRow`~~ -> `get_row`).

redash/query_runner/elasticsearch.py
```diff
@@ -132,6 +130,48 @@ def add_column_if_needed(mappings, column_name, friendly_name, result_columns, r
...
+            if row == None:
```
arikfr: `if row is None`

redash/query_runner/elasticsearch.py
```diff
@@ -132,6 +130,48 @@ def add_column_if_needed(mappings, column_name, friendly_name, result_columns, r
...
+            if isinstance(data, dict):
+
+                for key, value in data.iteritems():
+                    val = collect_aggregations(mappings, rows, parentKey if key == 'buckets' else key, value, row, result_columns, result_columns_index)
+                    if val:
+                        row = getRow(rows, row)
+                        collect_value(mappings, row, key, val, 'long')
+
+                for dataKey in ['value', 'doc_count']:
+                    if not dataKey in data:
```
arikfr: `if dataKey not in data`

redash/query_runner/elasticsearch.py
```diff
@@ -139,27 +179,49 @@ def add_column_if_needed(mappings, column_name, friendly_name, result_columns, r
...
+                    add_column_if_needed(mappings, column, column, result_columns, result_columns_index)
+
+                    value = h[column_name][column]
+                    row[column] = value[0] if isinstance(value, list) and len(value) == 1 else value
 
-            if row and len(row) > 0:
                 result_rows.append(row)
+        else:
+
+            raise Exception('Seems you\'ve hit an unsupported feature')
```
arikfr: You can use "" here instead to avoid escaping (`raise Exception("Seems you've hit...")`).

Also considering this message will be relayed back to the user, let's rephrase it to something like:
"Redash failed to parse the results it got from ElasticSearch.", to make sure it's clear where the error is coming from.

redash/query_runner/elasticsearch.py
arikfr: (and of course `data_key`)

arikfr:
I've made mainly style comments as I'm not familiar with ElasticSearch's results format that much. If you can add a few examples of possible results from ElasticSearch, I can help with reviewing the logic and also writing a few test cases.

lloydw:
Hey, sorry about the delay, been busy with other things. I've updated the styles as requested.

ElasticSearch uses a tree style response, so the code basically does a depth first traversal to flatten each response.

How do you want me to send you the example results?

arikfr:
> How do you want me to send you the example results?

You can create a gist with them or any other way that works for you.

chrsbats:
I have been playing with this. Line 161 seems to be missing a "type" parameter in the call to collect_value.

lloydw:
Yes, you'd be right

lloydw:
https://gist.github.com/lloydw/865bbcc0508288576648bff17c871234 - Here are some sample ElasticSearch outputs we use that the code correctly parses

arikfr:
Merged! Thanks and sorry it took so long.

lloydw:
No worries, you had slackbots to build :)

Issue #860, opened 2016-02-27
Feature request: Manage.py support group operation


ninneko:
For provisioning script, I want to add group, modify permission ... etc by manage.py.

Slach:
+1

sc0rp10:
+1

ninneko:
Il try this issue on this weekend.

arikfr:
Closed in #979.

PR #979, opened 2016-04-10
Feature: Add CLI to edit group permissions

ninneko:
#860

### DIFF
--- a/redash/cli/groups.py
+++ b/redash/cli/groups.py
@@ -1,7 +1,7 @@
 from flask_script import Manager, prompt_pass
 from redash import models
 
-manager = Manager(help="Users management commands. This commands assume single organization operation.")
+manager = Manager(help="Groups management commands. This commands assume single organization operation.")
 
 @manager.option('name', help="Group's name")
 @manager.option('--permissions', dest='permissions', default=None, help="Comma seperated list of permissions ('create_dashboard', 'create_query', 'edit_dashboard', 'edit_query', 'view_query', 'view_source', 'execute_query', 'list_users', 'schedule_query', 'list_dashboards', 'list_alerts', 'list_data_sources') (leave blank for default).")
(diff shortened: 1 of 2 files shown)


redash/cli/groups.py
```diff
@@ -0,0 +1,34 @@
+from flask_script import Manager, prompt_pass
+from redash import models
+
+manager = Manager(help="Users management commands. This commands assume single organization operation.")
```
arikfr: It should probably say "Group management..." instead of "Users management".

redash/cli/groups.py
```diff
@@ -0,0 +1,34 @@
...
+@manager.option('id', help="Group's id")
+@manager.option('--permissions', dest='permissions', default=None, help="Comma seperated list of permissions ('create_dashboard', 'create_query', 'edit_dashboard', 'edit_query', 'view_query', 'view_source', 'execute_query', 'list_users', 'schedule_query', 'list_dashboards', 'list_alerts', 'list_data_sources') (leave blank for default).")
+def change_permission(id, permissions=None):
+  print "Change permissisons of group %s ..." % (id)
+
+  if permissions is None:
+    permissions = models.Group.DEFAULT_PERMISSIONS
+  else:
+    permissions = permissions.split(',')
+    permissions = [p.strip() for p in permissions]
```
arikfr: You should factor this out into a helper function, instead of duplicating the code.

redash/cli/groups.py
```diff
@@ -0,0 +1,34 @@
...
+  print "permissions: [%s]" % ",".join(permissions)
+  
+  models.Group.update(permissions=permissions).where(models.Group.id==id).execute()
```
arikfr: I would split this into finding the group `group = models.Group.get_by_id(id)` and updating. While it will result in two queries it has the benefit of throwing an error if the user had a non existing ID. Also you can print out the previous permissions in case the user updated the wrong group.

arikfr:
Thank you for making the pull request! Please see comments.

redash/cli/groups.py
```diff
@@ -0,0 +1,50 @@
+from flask_script import Manager, prompt_pass
+from redash import models
+
+manager = Manager(help="Groups management commands. This commands assume single organization operation.")
+
+@manager.option('name', help="Group's name")
+@manager.option('--permissions', dest='permissions', default=None, help="Comma seperated list of permissions ('create_dashboard', 'create_query', 'edit_dashboard', 'edit_query', 'view_query', 'view_source', 'execute_query', 'list_users', 'schedule_query', 'list_dashboards', 'list_alerts', 'list_data_sources') (leave blank for default).")
+def create(name, permissions=None):
+  print "Creating group (%s)..." % (name)
```
arikfr: One last thing: we're using tab size of 4 and not 2.

arikfr:
Merged. Thanks!

Sorry for the delay, GitHub doesn't notify me of new commits so I need to manually check if there were updates...

PR #980, opened 2016-04-11
Add environment variable to override secure_flag on session cookie

machira:
This commit creates an environment variable that allows Developers to set the secure flag[1] on browser cookies. By default, this is set to False by flask[2]. 
1. More on Secure Flag here: https://www.owasp.org/index.php/SecureFlag
2. See SESSION_COOKIE_SECURE here: http://flask.pocoo.org/docs/0.10/config/

### DIFF
--- a/redash/settings.py
+++ b/redash/settings.py
@@ -122,7 +122,7 @@ STATIC_ASSETS_PATHS = [fix_assets_path(path) for path in os.environ.get("REDASH_
 
 JOB_EXPIRY_TIME = int(os.environ.get("REDASH_JOB_EXPIRY_TIME", 3600 * 6))
 COOKIE_SECRET = os.environ.get("REDASH_COOKIE_SECRET", "c292a0a3aa32397cdb050e233733900f")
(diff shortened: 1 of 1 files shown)


redash/settings.py
```diff
@@ -122,6 +122,8 @@ def all_settings():
 
 JOB_EXPIRY_TIME = int(os.environ.get("REDASH_JOB_EXPIRY_TIME", 3600 * 6))
 COOKIE_SECRET = os.environ.get("REDASH_COOKIE_SECRET", "c292a0a3aa32397cdb050e233733900f")
+SESSION_COOKIE_SECURE = os.environ.get("REDASH_SESSION_COOKIE_SECURE", False)
```
arikfr: 1. You should wrap it with `parse_boolean` like we do for the other boolean values (or it will be the string "True"/"true" instead of boolean value).

2. I think a nice default would be to set this to the value of `ENFORCE_HTTPS`. What do you think?

arikfr:
Thanks!

machira:
Hey @arikfr. Thanks for the feedback. I have incorporated it. Let me know what you think.

arikfr:
👍

PR #983, opened 2016-04-12
Add link to embeds to open query page in new window

tdawber:
At Atlassian most of our reports will be consumed through Embeds.  This little tweak to add a pop-out link back to re:dash will help tie the embed with redash itself.  I used a button that is used in a similar fashion on dashboards.

Also fixed a bug in embed.py that was pulling visualisation attributes into the visualisation.query object.

screen shot 2016-04-12 at 12 10 25 pm

### DIFF
--- a/redash/handlers/embed.py
+++ b/redash/handlers/embed.py
@@ -38,7 +38,7 @@ def embed(query_id, visualization_id, org_slug=None):
 
     qr = project(qr, ('data', 'id', 'retrieved_at'))
     vis = project(vis, ('description', 'name', 'id', 'options', 'query', 'type', 'updated_at'))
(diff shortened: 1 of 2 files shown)


redash/handlers/embed.py
```diff
@@ -38,7 +38,7 @@ def embed(query_id, visualization_id, org_slug=None):
 
     qr = project(qr, ('data', 'id', 'retrieved_at'))
     vis = project(vis, ('description', 'name', 'id', 'options', 'query', 'type', 'updated_at'))
-    vis['query'] = project(vis, ('created_at', 'description', 'name', 'id', 'latest_query_data_id', 'name', 'updated_at'))
+    vis['query'] = project(vis['query'], ('created_at', 'description', 'name', 'id', 'latest_query_data_id', 'name', 'updated_at'))
```
arikfr: 😮  oops.

arikfr:
Thanks!

PR #987, opened 2016-04-13
Feature: ability to change logo

AntoineAugusti:


redash/handlers/__init__.py
```diff
@@ -32,6 +32,14 @@ def status_api():
     return jsonify(status)
 
 
+@routes.app_context_processor
+def inject_variables():
+    return dict(
+        name=settings.NAME,
+        logoUrl=settings.LOGO_URL,
```
arikfr: `logoUrl` -> `logo_url`
(we're using `camelCase` only for Javascript and `snake_case` for Python)

arikfr:
Merged as is, because I want to merge #984 after this is merged already. I will fix the casing myself :+1:

PR #988, opened 2016-04-15
Feature: add CLI to invite user

AntoineAugusti:
It's in the title :)

### DIFF
--- a/redash/cli/users.py
+++ b/redash/cli/users.py
@@ -7,6 +7,21 @@ from redash.handlers.users import invite_user
 manager = Manager(help="Users management commands. This commands assume single organization operation.")
 
 
+def build_groups(groups, is_admin):
+    org = models.Organization.get_by_slug('default')
+    if isinstance(groups, basestring):
+        groups= groups.split(',')
+        groups.remove('') # in case it was empty string
+        groups = [int(g) for g in groups]
+
+    if groups is None:
+        groups = [org.default_group.id]
+
+    if is_admin:
+        groups += [org.admin_group.id]
+
+    return groups
+
 @manager.option('email', help="email address of the user to grant admin to")
 def grant_admin(email):
     try:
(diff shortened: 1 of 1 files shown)


redash/cli/users.py
```diff
@@ -76,6 +70,31 @@ def password(email, password):
...
+    try:
+        userFrom = models.User.get_by_email_and_org(inviterEmail, org)
+        user = models.User(org=org, name=name, email=email, groups=groups)
+
+        try:
+            user.save()
+        except IntegrityError as e:
+            if "email" in e.message:
+                print "Cannot invite. User already exists [%s]" % email
+                return
```
arikfr: Move the `invite_user` call and the message print to be inside the try clause, and then you can avoid the early return. Also please add an `else` clause in case the integrity error is for another reason.

redash/cli/users.py
```diff
@@ -84,3 +103,19 @@ def list():
             print "-" * 20
 
         print "Id: {}\nName: {}\nEmail: {}".format(user.id, user.name.encode('utf-8'), user.email)
+
+
+def build_groups(groups, is_admin):
```
arikfr: Please move the utility function to be above the functions that use it.

redash/cli/users.py
```diff
@@ -84,3 +103,19 @@ def list():
...
+    org = models.Organization.get_by_slug('default')
+    if isinstance(groups, basestring):
+        groups= groups.split(',')
+        groups.remove('') # in case it was empty string
+        groups = [int(g) for g in groups]
+
+    if groups is None:
+        groups = [models.Group.get(models.Group.name=="default", models.Group.org==org).id]
```
arikfr: If we already touching this code, then you can use `org.default_group` and `org.admin_group` helper methods.

redash/cli/users.py
```diff
@@ -76,6 +70,31 @@ def password(email, password):
...
+@manager.option('email', help="The invitee's email")
+@manager.option('name', help="The invitee's full name")
+@manager.option('inviterEmail', help="The email of the inviter")
+@manager.option('--admin', dest='is_admin', action="store_true", default=False, help="set user as admin")
+@manager.option('--groups', dest='groups', default=None, help="Comma seperated list of groups (leave blank for default).")
+def invite(email, name, inviterEmail, groups, is_admin=False):
+    org = models.Organization.get_by_slug('default')
+    groups = build_groups(groups, is_admin)
+    try:
+        userFrom = models.User.get_by_email_and_org(inviterEmail, org)
```
arikfr: Please make sure to comply with pep8 and use `snake_case`.

AntoineAugusti:
@arikfr Just tackled your comments

redash/cli/users.py
```diff
@@ -76,6 +85,31 @@ def password(email, password):
...
+@manager.option('email', help="The invitee's email")
+@manager.option('name', help="The invitee's full name")
+@manager.option('inviterEmail', help="The email of the inviter")
+@manager.option('--admin', dest='is_admin', action="store_true", default=False, help="set user as admin")
+@manager.option('--groups', dest='groups', default=None, help="Comma seperated list of groups (leave blank for default).")
+def invite(email, name, inviterEmail, groups, is_admin=False):
+    org = models.Organization.get_by_slug('default')
+    groups = build_groups(groups, is_admin)
+    try:
+        user_from = models.User.get_by_email_and_org(inviterEmail, org)
```
arikfr: `inviterEmail`... (here and in the option def)

AntoineAugusti:
@arikfr Done. Fuuu, snake case!

arikfr:
Merged. Thanks!

Issue #929, opened 2016-03-16
Support parameters in embeds (iframes)


rohanpd:
We would like to be able to append parameters to embeds so that we can allow for filterable reports in our internal applications.

rohanpd:
As part of this request it would be nice to have support for the maxAge parameter too. 

To some extent the maxAge parameter is more important to us then query parameters as we have filters and multi filters available as a work around for query params but no available work around for maxAge.

rohanpd:
We can close this one now. :)

PR #995, opened 2016-04-18
Feature/params on embeds

tdawber:
Hey @arikfr,

Here's a pull request fulfilling #929.  Given the potential sensitivity around maybe exposing the query sql, I added a parameter, defaulting to off, to enable this feature.  Added comments about potential side-effects of this feature.

Also enabled maxAge on embeds.

Happy to have a chat if this isn't quite how you would've done it.

Tom


redash/settings.py
```diff
@@ -201,6 +201,12 @@ def all_settings():
 # Enhance schema fetching
 SCHEMA_RUN_TABLE_SIZE_CALCULATIONS = parse_boolean(os.environ.get("REDASH_SCHEMA_RUN_TABLE_SIZE_CALCULATIONS", "false"))
 
+# Allow Parameters in Embeds
+# Warning: This will expose the full sql query and query attributes in the embed javascript
+# If you are exposing embeds publicly, be aware that this setting can leak your sql queries
```
arikfr: Worth mentioning that there is also the danger of SQL injection.

rd_ui/app/scripts/embed.js
```diff
@@ -43,12 +43,21 @@ angular.module('redash', [
...
        query = new Query(visualization.query);
-       $scope.queryResult = new QueryResult({query_result:query_result});
+
+       //max age from querystring, default to -1
+       maxAge = $location.search()['maxAge'];
+       if (maxAge === undefined) {
+          maxAge = -1;
+       }
+
+       qr = query.getQueryResult(maxAge, Query.collectParamsFromQueryString($location, query));
```
arikfr: Note that you don't use the `query_result` object, which means:
1. This won't work for unautheticated users (the call to query result will be without the API key).
2. No point in passing it or preloading it.

I think your direction is the right one, but needs a bit more work to work in all scenarios.

My suggestion is:
Rewrite the [`embed`](https://github.com/getredash/redash/blob/master/redash/handlers/embed.py#L19) endpoint to be more like [`public_dashboard`](https://github.com/getredash/redash/blob/master/redash/handlers/embed.py#L52) and use the same template (`public.html`). We can drop the `embed.js` file and move the routes definitions into `app.js`. 

Pass in the `seed_data` only the query object, but use the API to receive the query result (either a new one or existing one with the query result id). In app.js we already [take care of passing the API key](https://github.com/getredash/redash/blob/master/rd_ui/app/scripts/app.js#L28) to subsequent API calls so it will work both for signed in and signed out users.

tdawber: Thanks for the feedback, @arikfr.  I'll have a dig into your suggestion and see what I can come up with.

redash/handlers/embed.py
```diff
@@ -46,12 +57,24 @@ def embed(query_id, visualization_id, org_slug=None):
...
+    if settings.ALLOW_PARAMETERS_IN_EMBEDS == True:
+        #this will enable embedding parameters, but in doing so will expose the query sql
+        #the query sql will be passed to clients with the security token
+        vis['query'] = project(vis['query'], ('created_at', 'description', 'name', 'id', 'latest_query_data_id', 'name', 'updated_at', 'latest_query_data', 'data_source_id', 'query'))
+    else:
+        vis['query'] = project(vis['query'], ('created_at', 'description', 'name', 'id', 'latest_query_data_id', 'name', 'updated_at'))
+        
+    
+    response = render_template("public.html",
+                               headless='embed' in request.args,
```
arikfr: In the context of a visualization embed, you can always pass here `True`.

arikfr:
Played around with it and found several issues, that are actually issues in the code you're using and not in your pull request directly:

1. 
In case `ALLOW_PARAMETERS_IN_EMBEDS` is `False`, calling `Query#getQueryResult` returns immediately because Query's `query` value is `undefined` (and `Query#getQueryResult` checks that `query` is truthy).

Possible solutions: simple workaround of passing an empty string as `query` value (I've did it in some other place in the code before) or adding a parameter to `Query#getQueryResult` to skip this check.

2. 
The API key we use is of a query, therefore we need a `query_id` in the request to associate it with a query. But the call `Query#getQueryResult` makes has no `query_id` in the URL. 

This one is trickier to solve and will require some refactoring both on the server side and in the `QueryResult` implementation on the client side. I'm happy if both will be done, but the question is this something you can invest your time in? If you do, I'll be happy to elaborate on possible solutions.

arikfr:
Closing in favor of #1014. We will eventually implement something along the lines of my comment above, but until then...

nicokras:
Hi, can you give me an example on how to embed with parameters? 

I have tried several ways like the one below, but could not make it work. 

http://server/embed/query/3/visualization/4?api_key=apikey&p_start_date=2017-03-01

Also enabled REDASH_ALLOW_PARAMETERS_IN_EMBEDS=true in the .env file, and restarted. 

I will appreciate any help.
Thanks.

Aleyasen:
How can we embed with parameters? I'll be thankful if anyone can give an example.

rexuli:
I found a while and cannot get the answer. So I just try and it works.
Example:
```
<iframe src="http://redash_url/embed/query/11/visualization/33?api_key=API_KEY&p_Start%20Date=2018 04-10&p_End%20Date=2018-04-18"
  width="720" height="391"></iframe>
```
Params: Start Date & End Date in the query in redash

The &p_Start%20Date=2018 04-10&p_End%20Date=2018-04-18 will be the params in the query inside redash.

Remarks:
The embed needs to log in to redash.
Enabled REDASH_ALLOW_PARAMETERS_IN_EMBEDS=true in the .env

PR #998, opened 2016-04-18
Fix #994: query description was rendered with bigger font than name

### LINKED ISSUE #994: Embed query description appearing larger than vizualization name
The query description is appearing larger then the visualization name:

image

arikfr:
Result:

re_dash__

I've also added the query name to the embed. Anyone sees an issue with showing it?

### DIFF
--- a/rd_ui/app/scripts/embed.js
+++ b/rd_ui/app/scripts/embed.js
@@ -43,12 +43,13 @@ angular.module('redash', [
     }
   ])
    .controller('EmbedCtrl', ['$scope', function ($scope) {} ])
-   .controller('EmbeddedVisualizationCtrl', ['$scope', 'Query', 'QueryResult',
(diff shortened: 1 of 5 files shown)


arikfr:
@rohanpd WDYT?

rohanpd:
I like it but I have two concerns:
1. Some people might have a rather large query description and this could really push down the visualisation
2. We were using it several places currently without a query description just because we are trying to reduce real estate. 

It would be nice if we had a way to toggle some of these things off so we could have the visualisation as close as possible to the top of the embed. In our case we us the wiki markup of confluence to do the descriptions etc.

arikfr:
How about adding a url param to hide the description?

rohanpd:
I'm happy with this or having a URL parameter to show the description so it is by default off.

¯\_(ツ)_/¯

PR #1005, opened 2016-04-20
Update bootstrap.sh to include saml dependency

staritza:
Required for saml to work using an ubuntu server


arikfr:
Thanks!

arikfr:

Issue #1006, opened 2016-04-20
X axis labels get truncated when long enough in visualization/dashboard view


staritza:
We created [this visualization ](https://demo.redash.io/embed/query/1545/visualization/2060?api_key=27990352780ec9339a6c6b3e4d643599e5239a71) in demo.redash.io which has the latest plotly.js update.

Downloaded images also have the labels truncated.

What we would like to see is that the visualization has room to show full labels for items on the x-axis (scrollbar, adjustable graph, etc...).

arikfr:
Yeah, I'm aware of this. I thought that maybe if I let Plot.ly autosize the chart it will solve it, but looks like not. Seems like there is no automatic solution, so I'm thinking of letting the user control the margin size so they can decide how much space to give and if truncation is important to them.

arikfr:
Fixed in #1007.

PR #1007, opened 2016-04-22
Issue#1006:  Make bottom margin editable for Chart visualization

vorakumar:
This addresses Issue #1006 by allowing user to modify bottom margin for Chart visualization. When the margin is modified, it adjusts the hight accordingly.

### DIFF
--- a/rd_ui/app/scripts/directives/plotly.js
+++ b/rd_ui/app/scripts/directives/plotly.js
@@ -121,6 +121,8 @@
   angular.module('plotly', [])
     .constant('ColorPalette', ColorPalette)
     .directive('plotlyChart', function () {
+      var baseHeight = 250;
+      var bottomMargin = 50;
       return {
         restrict: 'E',
         template: '<div></div>',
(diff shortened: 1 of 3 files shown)


rd_ui/app/scripts/directives/plotly.js
```diff
@@ -121,6 +121,8 @@
   angular.module('plotly', [])
     .constant('ColorPalette', ColorPalette)
     .directive('plotlyChart', function () {
+      var baseHeight = 250;
```
arikfr: Why not calculate the height based on the height that is passed to the directive?

arikfr: Another option is to skip setting height here and delegate this to Plotly (by setting the height of the div and only specifying the margin size).

arikfr:
Thanks!

rd_ui/app/scripts/directives/plotly.js
ChrisKozak: Hello Arik,
Kumar has gone on vacation and I'm going to pick up where he left off.  Which alternative approach would you prefer we take?  Delegating to Plotly sounds more elegant.  What do you think?

MoonShadow70: [image: Inline image 2]
I'm having the same difficulty.

vorakumar:
@arikfr Removed the logic to calculate the height, and instead letting plotly to choose the right height based on the bottom margin. It seems to be working.

vorakumar:
@arikfr any feedback on the latest change?

arikfr:
Thanks!

PR #1014, opened 2016-04-27
Add server-side parameter handling for embeds

whummer:
Hey @arikfr,

This is an updated pull request fulfilling #929, taking a slightly different approach than our previous PR (https://github.com/getredash/redash/pull/995). Instead of retrieving the query template and filling in the parameter on the client side, this code receives the parameters and renders the query template on the server side (via `pystache.render`) and then returns the result.

It seems that this is a much simpler approach which integrates nicely with the existing `embed.html`. It is potentially also safer because the SQL query is not exposed to the client. SQL injection is still a potential threat, though, therefore we keep a warning message in the `settings.py` and disable this feature by default.

In general, it seems that other parts of the framework are also affected by potential SQL injection (e.g.,`QueryResultListResource::post` uses plain parameters). As part of a future work item, maybe we could use a proper SQL templating engine to prevent SQL injection altogether. However, for our current use cases this solution is sufficient (it is used mostly internally).

Please let us know what you think. Thanks

### DIFF
--- a/redash/handlers/embed.py
+++ b/redash/handlers/embed.py
@@ -7,11 +7,34 @@ from flask_restful import abort
 
 from redash import models, settings
 from redash import serializers
-from redash.utils import json_dumps
+from redash.utils import json_dumps, collect_parameters_from_request
 from redash.handlers import routes
 from redash.handlers.base import org_scoped_rule, record_event
 from redash.permissions import require_access, view_only
 from authentication import current_org
+
+#
+# Run a parameterized query synchronously and return the result
+# DISCLAIMER: Temporary solution to support parameters in queries. Should be
+#             removed once we refactor the query results API endpoints and handling
+#             on the client side. Please don't reuse in other API handlers.
+#
+def run_query_sync(data_source, parameter_values, query_text):
+    query_parameters = set(collect_query_parameters(query_text))
+    missing_params = set(query_parameters) - set(parameter_values.keys())
+    if missing_params:
+        raise Exception('Missing parameter value for: {}'.format(", ".join(missing_params)))
+
+    if query_parameters:
+        query_text = pystache.render(query_text, parameter_values)
+
+    try:
+        data, error = data_source.query_runner.run_query(query_text)
+        if error:
+            return None
+        return data
+    except Exception, e:
+        return None
 
 
 @routes.route(org_scoped_rule('/embed/query/<query_id>/visualization/<visualization_id>'), methods=['GET'])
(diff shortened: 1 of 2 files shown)


arikfr:
I don't want to introduce synchronous execution of queries with current implementation, as it will lead to some long running queries choking the frontend making the system unresponsive. I do want to introduce it in the future when we make the API server async (either by utilizing gevent or some other way).

Adding non standard query execution way makes this pull request less simpler than the other pull request. While #995 requires some more work (I will comment there soon on the results of my testing), it is the way to go, as it uses existing components.

`QueryResultListResource::post` allows plain parameters because it verifies you have permission to execute a query first. If you have a permission to run any query there is no risk of SQL injection (you can just run this query directly). Regardless we should consider moving to safe templating option, it just that we need to take into account that not all the data sources we support are SQL based...

redash/handlers/query_results.py
```diff
@@ -37,6 +37,22 @@ def run_query(data_source, parameter_values, query_text, query_id, max_age=0):
...
+def run_query_sync(data_source, parameter_values, query_text):
+    query_parameters = set(collect_query_parameters(query_text))
+    missing_params = set(query_parameters) - set(parameter_values.keys())
+    if missing_params:
+        raise Exception('Missing parameter value for: {}'.format(", ".join(missing_params)))
+
+    if query_parameters:
+        query_text = pystache.render(query_text, parameter_values)
+
+    data, error = data_source.query_runner.run_query(query_text)
```
arikfr: `run_query` might raise an exception, so better wrap it with a try/except.

redash/handlers/query_results.py
```diff
@@ -37,6 +37,22 @@ def run_query(data_source, parameter_values, query_text, query_id, max_age=0):
         job = enqueue_query(query_text, data_source, metadata={"Username": current_user.name, "Query ID": query_id})
         return {'job': job.to_dict()}
 
+#
+# Run a parameterized query synchronously and return the result
+#
+def run_query_sync(data_source, parameter_values, query_text):
```
arikfr: Let's move this function to `embed.py` and add the following disclaimer: 
> Temporary solution to support parameters in queries. Should be removed once we refactor the query results API endpoints and handling on the client side. Please don't reuse in other API handlers.

redash/handlers/query_results.py
whummer: done. changed in the updated PR

redash/handlers/query_results.py
whummer: done.

whummer:
@arikfr Thanks for the update on this PR. Please see my changes corresponding to your comments.

arikfr:
Merged. 🎉

PR #1016, opened 2016-04-28
Feature: Apache Drill query runner

dharmicdata:
New query runner for connecting to Apache drill.
Added drill as a query runner in settings.py.
Added drill_setup.sh in in the setup/ubuntu folder. It is not run automatically in the bootstrap script.

### DIFF
--- a/redash/query_runner/drill.py
+++ b/redash/query_runner/drill.py
@@ -1,0 +1,179 @@
+import json
+import logging
+import sys
+import os
+
+from redash.query_runner import *
+from redash.utils import JSONEncoder
+
+logger = logging.getLogger(__name__)
+
+try:
+    import pyodbc
+    enabled = True
+    logger.warning("LD_LIBRARY_PATH: {}".format(os.environ.get('LD_LIBRARY_PATH')))
+    logger.warning("ODBCINI: {}".format(os.environ.get('ODBCINI')))
+    logger.warning("MAPRDRILLINI: {}".format(os.environ.get('MAPRDRILLINI')))
+except ImportError, e:
+    logger.warning("Missing dependencies. Please install pyodbc.")
+    logger.warning("You can use pip: pip install pyodbc")
+    enabled = False
+
+COLUMN_NAME = 0
+COLUMN_TYPE = 1
+
+types_map = {
+    'BIGINT': TYPE_INTEGER,
+    'INTEGER': TYPE_INTEGER,
+    'INT': TYPE_INTEGER,
+    'DOUBLE': TYPE_FLOAT,
+    'long': TYPE_INTEGER,
+    'DECIMAL': TYPE_FLOAT,
+    'DEC': TYPE_FLOAT,
+    'NUMERIC': TYPE_FLOAT,
+    'FLOAT': TYPE_FLOAT,
+    'REAL': TYPE_FLOAT,
+    'BOOLEAN': TYPE_BOOLEAN,
+    'bool': TYPE_BOOLEAN,
+    'TIMESTAMP': TYPE_DATETIME,
+    'TIME': TYPE_DATETIME,
+    'DATE': TYPE_DATETIME,
+    'CHAR': TYPE_STRING,
+    'STRING': TYPE_STRING,
+    'str': TYPE_STRING,
+    'VARCHAR': TYPE_STRING,
+    'CHARACTER': TYPE_STRING,
+    'CHARACTER VARYING': TYPE_STRING,
+    # Unsure about the interval type
+    'INTERVAL': TYPE_DATETIME,
+}
+
+
+class Drill(BaseQueryRunner):
+    @classmethod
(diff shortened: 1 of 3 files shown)


redash/query_runner/drill.py
```diff
@@ -0,0 +1,179 @@
...
+
+try:
+    import pyodbc
+    enabled = True
+    logger.warning("LD_LIBRARY_PATH: {}".format(os.environ.get('LD_LIBRARY_PATH')))
+    logger.warning("ODBCINI: {}".format(os.environ.get('ODBCINI')))
+    logger.warning("MAPRDRILLINI: {}".format(os.environ.get('MAPRDRILLINI')))
+except ImportError, e:
+    logger.warning("Missing dependencies. Please install pyodbc.")
+    logger.warning("You can use pip: pip install pyodbc")
```
arikfr: We no longer use these messages. Please remove.

redash/query_runner/drill.py
```diff
@@ -0,0 +1,179 @@
...
+from redash.utils import JSONEncoder
+
+logger = logging.getLogger(__name__)
+
+try:
+    import pyodbc
+    enabled = True
+    logger.warning("LD_LIBRARY_PATH: {}".format(os.environ.get('LD_LIBRARY_PATH')))
+    logger.warning("ODBCINI: {}".format(os.environ.get('ODBCINI')))
+    logger.warning("MAPRDRILLINI: {}".format(os.environ.get('MAPRDRILLINI')))
```
arikfr: Please remove these warnings or move them to `debug` level.

redash/query_runner/drill.py
```diff
@@ -0,0 +1,179 @@
...
+}
+
+
+class Drill(BaseQueryRunner):
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "ConnectionType": {
```
arikfr: Please use consistent naming for the configuration properties. Preferably snake cases, i.e.: `connection_type`.

Also provide a `title` to give a friendly name, specially for the ZK ones.

redash/query_runner/drill.py
```diff
@@ -0,0 +1,179 @@
...
+
+                for table_name in map(lambda a: a['name'], self._run_query_internal(query_to_run % schema_name)):
+                    columns = map(lambda a: a['friendly_name'], self._run_query_internal_columns(columns_query % (schema_name, table_name)))
+
+                    if schema_name != 'INFORMATION_SCHEMA' and schema_name != 'sys':
+                        table_name = '{}.`{}`'.format(schema_name, table_name)
+
+                    schema[table_name] = {'name': table_name, 'columns': columns}
+        except Exception, e:
+            raise sys.exc_info()[1], None, sys.exc_info()[2]
```
arikfr: If you don't handle any exception and just re-raise it, why have the try/except at all?

arikfr:
Thanks! 

I've added a few comments, mainly on style. Later on I will setup Drill to give it a proper test so might have some more comments.

breath103:
👍 love to see get merged. any progress? i'm running re:dash on production AND apache drill on production too so let me know if need help on testing

arikfr:
@breath103 if you can test it on your setup and give feedback on if you had issues (as I never had the chance to test it myself :-(). There are a few style fixes to do, but I can do them myself after merging, once we know it's OK.

breath103:
@arikfr ok let me try out
@dharmicdata by the way, how do you think about implementing this with Apache Drill HTTP API? i feel like using JDBC / ODBC seems making this pretty much complicated, and since queries on apache drill usually take few minutes to be solved, that would be no big difference i feel like

arikfr:
@breath103 using the HTTP API is an interesting idea, and will simplify this greatly. Do you know if there are any options that we lose if we switch to it?

breath103:
well i can't really think of any... yet. basically Apache Drill supports ANSI standard for SQL, so even it's REST API is about passing SQL through the HTTP and getting result with JSON.
it also supports command like "SHOW SCHEMAS;", "SHOW TABLES".. etc so i guess it would work. even though that means we force people to open the extra port (8047 usually) but not sure that's the major problem here 

+ the format of Apache Drill's REST API returns is same format with the JSON format the Re:dash URL datasource supports, {columns: [], rows: []},
so basically, we could even use Apache Drill if the URL datasource supports POST call and Body.

arikfr:
@breath103 it's very similar, but the columns object is different. From the example in their docs, all the values returned as strings. Is it really the case? Is there some support for authentication?

breath103:
oh i see. types.. that could be tricky i guess. as of authentication, i don't think there is one. setting IP whitelist from the apache drill side would be the only way

jeffwidman:
@dharmicdata Any updates here?

arikfr:
Closing for inactivity :(

fcastillao:
nobody tested this?

PR #1021, opened 2016-05-01
Vagrant docs: add purging the cache step

ariarijp:
I got an error with Vagrant setup procedure in [Setting up development environment (using Vagrant)](http://docs.redash.io/en/latest/dev/vagrant.html).

This error to solve with purging the cache.

```
12:32:33 worker.1 | [2016-05-01 12:32:33,191: ERROR/MainProcess] Task redash.tasks.record_event[0c4fbabf-77b0-42ed-9bf6-52f071f15d82] raised unexpected: KeyError('org_id',)
12:32:33 worker.1 | Traceback (most recent call last):
12:32:33 worker.1 |   File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
12:32:33 worker.1 |     R = retval = fun(*args, **kwargs)
12:32:33 worker.1 |   File "/opt/redash/current/redash/tasks/base.py", line 13, in __call__
12:32:33 worker.1 |     return super(BaseTask, self).__call__(*args, **kwargs)
12:32:33 worker.1 |   File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 437, in __protected_call__
12:32:33 worker.1 |     return self.run(*args, **kwargs)
12:32:33 worker.1 |   File "/opt/redash/current/redash/tasks/general.py", line 13, in record_event
12:32:33 worker.1 |     models.Event.record(event)
12:32:33 worker.1 |   File "/opt/redash/current/redash/models.py", line 1046, in record
12:32:33 worker.1 |     org = event.pop('org_id')
12:32:33 worker.1 | KeyError: 'org_id'
```

Please review my changes.

Thanks,

PR #1027, opened 2016-05-04
Improve the Vagrant flow

staritza:

Hey..is it possible to ignore the changes to the deploy and notifications part of circle.yml? We had to take that out for our private use, but everything else can be merged in okay.


arikfr:
This pull request picked up all your other changes on your fork. You need to create a branch that has only the Vagrant related changes and create the pull request from it.

Vagrantfile
```diff
@@ -8,4 +8,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box = "redash/dev"
   config.vm.synced_folder "./", "/opt/redash/current"
   config.vm.network "forwarded_port", guest: 5000, host: 9001
+  config.vm.provision "shell" do |s|
+    s.inline = "/opt/redash/current/install_requirements.sh"
```
arikfr: Let's use a single script (`vagrant_provision.sh`) that will also install the requirements. Also in the case of the Vagrant box, let's install all 3 files -- `requirements.txt`, `requirements_dev.txt` and `requirements_all_ds.txt` (this also means that there is no need for manual install of the PyMongo package).

vagrant_provision.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
arikfr: Please move this file into `setup/vagrant`.

run_redash_locally.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+#Provision VM
+vagrant up
+
+#start server and background workers
+vagrant ssh -c "cd /opt/redash/current; bin/run honcho start -f Procfile.dev" &
```
arikfr: Let's have a single script to control vagrant and place it in `bin/` (we can call it `vagrant_ctl`) and use subcommands:

```bash
#!/bin/bash
set -e

help() {
  echo "Usage: "
  echo "`basename "$0"` {start, test}"
}

case "$1" in
  run)
    vagrant up
    vagrant ssh -c "cd /opt/redash/current; bin/run honcho start -f Procfile.dev"
    ;;
  test)
    vagrant up
    vagrant ssh -c "cd /opt/redash/current; make test"
    ;;
  *)
    help
    ;;
esac
```

vagrant_provision.sh
```diff
@@ -0,0 +1,13 @@
...
+cp /opt/redash/.env /opt/redash/current
+cd /opt/redash/current/rd_ui
+bower install
+cd /opt/redash/current
+
+#update database
+bin/run ./manage.py database drop_tables
+bin/run ./manage.py database create_tables
+bin/run ./manage.py users create --admin --password admin "Admin" "admin"
+
```
arikfr: Please add the Redis flush command from #1021.

Vagrantfile
staritza: We wanted the install_requirements.sh to be the same as what is used in circle.yml, so that there isn't a configuration drift (where circle.yml is updated but the vagrant provisioning file isnt). I guess we could still keep this file separate and call it from both circle.yml and vagrant_provision.sh. What do you think?

run_redash_locally.sh
staritza: Sounds good!

Vagrantfile
arikfr: Thing is that the dependencies we need to install in `circle.yml` are not the same as what we would install in Vagrant. In Vagrant we can install a larger set of dependencies to make it easier to develop and test wider range of scenarios, while for Circle I prefer to install as less as possible to make the build faster.

I really don't expect there will be many other requirements files added (if any), so this is a "risk" we can take.

staritza: okay. got ya.

PR #1043, opened 2016-05-09
Improve the vagrant flow

staritza:

To run locally, run ./bin/vagrant_ctl.sh start 
To test, ./bin/vagrant_ctl.sh test


arikfr:
Looks like you need to rebase with updated master.

docs/dev/vagrant.rst
```diff
@@ -14,27 +14,11 @@ To get started with this box:
     `Vagrant <https://www.vagrantup.com/>`__ installed.
 2.  Clone the Re:dash repository:
     ``git clone https://github.com/getredash/redash.git``.
-3.  Change dir into the repository (``cd redash``) and run run
-    ``vagrant up``. This might take some time the first time you run it,
+3.  Change dir into the repository (``cd redash``)
+4a. To execute tests, run ``./run_tests.sh``
+4b. To run the app, run ``./run_redash_locally.sh``.
```
arikfr: Notice the docs mention the old scripts and not `vagrant_ctl`.

setup/vagrant/provision.sh
```diff
@@ -0,0 +1,22 @@
...
+cd /opt/redash/current
+cp /opt/redash/.env /opt/redash/current
+cd /opt/redash/current/rd_ui
+bower install
+cd /opt/redash/current
+
+#install requirements
+sudo pip install -r /opt/redash/current/requirements_dev.txt
+sudo pip install -r /opt/redash/current/requirements.txt
+sudo pip install pymongo==3.2.1
```
arikfr: It's better to run `sudo pip install -r /opt/redash/current/requirements_all_ds.txt` here.

docs/dev/vagrant.rst
staritza: My bad. Good catch!

setup/vagrant/provision.sh
staritza: Running requirements_all_ds.txt has some dependencies that are not able to be installed. I could spend some time trying to fix all the dependencies. But they are not really required for anything on the vagrant box as far as I can tell. So I just decided to go with installing pymongo, which is the only thing we really need (one test fails when we dont have it). I could be wrong though? The dependencies can be sorted out in a later PR? Is not using requirements_all_ds.txt introducing a regression?

staritza:
Please hold off on this pr for a while. We are trying to work on our team git/redash work flow.

staritza:
@arikfr you can ignore this pr. we created [this](https://github.com/getredash/redash/pull/1044) with only the commits that matter and changes to reflect your comments

PR #1044, opened 2016-05-09
Improve vagrant flow

staritza:

To run locally, run ./bin/vagrant_ctl.sh start 
To test, ./bin/vagrant_ctl.sh test


arikfr:
Thanks!

PR #1045, opened 2016-05-09
Feature: allow user to change editor theme

ChiragKParmar:
Adding theme for editor: screen shots below
theme5
theme4
theme3
theme2


ChiragKParmar:
Adding theme for editor: screen shots below
theme2
theme5
theme4
theme3

rd_ui/app/embed.html
```diff
@@ -12,6 +12,12 @@
     <!-- build:css /styles/embed.css -->
     <link rel="stylesheet" href="/styles/superflat_redash.css">
     <link rel="stylesheet" href="/bower_components/codemirror/lib/codemirror.css">
+    <link rel="stylesheet" href="/bower_components/codemirror/theme/3024-day.css">
+    <link rel="stylesheet" href="/bower_components/codemirror/theme/3024-night.css">
+    <link rel="stylesheet" href="/bower_components/codemirror/theme/eclipse.css">
+    <link rel="stylesheet" href="/bower_components/codemirror/theme/monokai.css">
+    <link rel="stylesheet" href="/bower_components/codemirror/theme/neat.css">
+    <link rel="stylesheet" href="/bower_components/codemirror/theme/elegant.css">
```
arikfr: No need to add this in `embed.html` as we don't use the editor in embeds.

arikfr:
This is cool, but I don't feel it's worth making the UI more crowded for this feature. Also without persisting the option, it will be somewhat frustrating for the user to keep selecting the theme over and over again.

Maybe we can move this to user configuration? Or "hide" under some menu in the query screen (although still need to add persistence to the option).

ChiragKParmar:
I guess you are right! I like the option of moving to some menu in the query screen. And agree that unless we persist users choice, this adds little value. Let me work on that part. You can either decline this pull request or I will make changes on the same. Thanks

arikfr:
OK, I will wait for the updated version.

arikfr:
Please re-open when relevant. Thanks!

PR #1046, opened 2016-05-10
add change items by page to ng_smart_table.js

kissrobber:
My boss requested

### DIFF
--- a/rd_ui/app/scripts/ng_smart_table.js
+++ b/rd_ui/app/scripts/ng_smart_table.js
@@ -338,6 +338,23 @@ function getKeyFromObject(obj, key) {
           });
         }
       };
+    }])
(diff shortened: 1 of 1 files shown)


arikfr:
This is a great addition that many will like (and not only your boss), _but_ unfortunately I can't merge it as is :-\ It needs some style/design update to be consistent with the rest of the page. I would try pinning it to one of the sides and apply Bootstrap's style classes to the control.

If you don't have the time to do it, I will do it myself but it will be later on after I finish some other things I'm working on...

kissrobber:
Thank you.
I'm waiting your update, since I'm not good at Bootstrap.

arikfr:
I was over optimistic about my time :( eventually didn't implement it myself and the UI code changed so much that this is no longer relevant. Sorry.

PR #1054, opened 2016-05-13
Adding oracle on default settings

joaoleite:
If you are not using Oracle, you will have no problem in application start.

I needed to make this change to work with docker. I hope that already in a nearby docker image.

tks

### DIFF
--- a/redash/settings.py
+++ b/redash/settings.py
@@ -145,6 +145,7 @@ default_query_runners = [
     'redash.query_runner.oracle',
     'redash.query_runner.sqlite',
     'redash.query_runner.mssql',
+    'redash.query_runner.oracle',
 ]
 
 enabled_query_runners = array_from_string(os.environ.get("REDASH_ENABLED_QUERY_RUNNERS", ",".join(default_query_runners)))


arikfr:
We're not including in the default settings any query runner that requires packages that aren't installed by default on the AMI or Docker container we provide.

But you don't need to change the code to enable it, just set the following environment variable:
`REDASH_ADDITIONAL_QUERY_RUNNERS=redash.query_runner.oracle`.

ebertti:
But each `query_runner` has a `enable()` method to check if can or not run. Why this can be added by default?

arikfr:
It's a user experience issue - when a query runner enabled but missing dependencies, it prints a warning message. Past experience shows people get confused by these messages, so I prefer to minimize them.

gclsoft:
@arikfr I've set the REDASH_ADDITIONAL_QUERY_RUNNERS, but not work:
```
root@ubuntu:~/work/redash# export REDASH_ADDITIONAL_QUERY_RUNNERS=redash.query_runner.oracle
```
How to install oracle support?

joaoleite:
Maybe this can help you @gclsoft 

https://hub.docker.com/r/joaoleite/redash_oracle/

source: https://github.com/joaoleite/redash_oracle

PR #1060, opened 2016-05-17
Feature: support configuring user's groups with SAML

staritza:
This PR allows Redash users to manage group assignments in using their SAML provider.

To do this, configure the SAML response to include an attribute with key 'RedashGroups', and value as a list of the names of the groups the user should be in, from Redash.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -244,6 +244,13 @@ class Group(BaseModel, BelongsToOrgMixin):
     def members(cls, group_id):
         return User.select().where(peewee.SQL("%s = ANY(groups)", group_id))
 
+    @classmethod
+    def find_by_name(cls, org, group_names):
+        if len(group_names) == 0:
+            return []
+        result = cls.select().where(cls.org == org, cls.name << group_names)
+        return list(result) if result.count() > 0 else []
+
     def __unicode__(self):
         return unicode(self.id)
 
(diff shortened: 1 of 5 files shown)


redash/models.py
```diff
@@ -330,6 +337,12 @@ def hash_password(self, password):
     def verify_password(self, password):
         return self.password_hash and pwd_context.verify(password, self.password_hash)
 
+    def update_group_assignments(self, group_names, org):
+        groups = Group.find_by_name(org, group_names)
```
arikfr: Why not use `self.org` here?

arikfr:
I didn't look into this in depth yet, but nice touch with adding docs and tests! 💯 👍

redash/models.py
vorakumar: sure! will use it

redash/models.py
```diff
@@ -244,6 +244,13 @@ def all(cls, org):
     def members(cls, group_id):
         return User.select().where(peewee.SQL("%s = ANY(groups)", group_id))
 
+    @classmethod
+    def find_by_name(cls, org, group_names):
+        if len(group_names) == 0:
+            return []
+        result = cls.select().where(cls.org == org, cls.name << group_names)
+        return list(result) if result.count() > 0 else []
```
arikfr: `result.count()` will do another query... wouldn't it return an empty list anyway?

arikfr:
@vorakumar I can merge now, right?

vorakumar:
I believe so

arikfr:
Done. Thanks!

arikfr:

Issue #948, opened 2016-03-27
Area chart stacking doesn't work


arikfr:

machira:
This is related to this issue in plotly: https://github.com/plotly/plotly.js/issues/344

PR #1061, opened 2016-05-17
Fix: area chart stacking doesn't work

machira:
This PR addresses issue #948 as recommended by Plotly maintainers [here](https://github.com/plotly/plotly.js/issues/344). 

Specifically, we recalculate the data appearance when the legend is clicked. This ensures that we show existing series at the updated heights. 

We also fix the behaviour of percentage area plots. Here, if the legend is clicked to disable a given series, the visualization is refreshed so that only enabled series are plotted. The percentage total is still calculated over the all series. This is designed to be consistent with the percentage bar charts, where after disabling a given series, the plotted visualizations add up to less than 100%.

### DIFF
--- a/rd_ui/app/scripts/directives/plotly.js
+++ b/rd_ui/app/scripts/directives/plotly.js
@@ -314,8 +314,7 @@
           element.on('plotly_afterplot', function(d) {
             if(scope.options.globalSeriesType === 'area' && (scope.options.series.stacking === 'normal' || scope.options.series.stacking === 'percent')){
               d3.selectAll('.legendtoggle').each(function(d, i) {
-                // override plotly's default legend item on-click behaviour
-                d3.select(this).on('click', function() {
+                d3.select(this).on('click.legend_hack', function(){
                   var maxIndex = scope.data.length - 1;
                   var itemClicked = scope.data[maxIndex  - i];
 
@@ -327,6 +326,10 @@
                    }
                   Plotly.redraw(element);
                 });
+              });
+            } else {
+              d3.selectAll('.legendtoggle').each(function(d, i) {
+                d3.select(this).on('click.legend_hack', null);
               });
             }
           });
--- a/rd_ui/app/scripts/directives/plotly.js
+++ b/rd_ui/app/scripts/directives/plotly.js
@@ -144,7 +144,6 @@
   angular.module('plotly', [])
     .constant('ColorPalette', ColorPalette)
     .directive('plotlyChart', function () {
-      var bottomMargin = 50;
       return {
         restrict: 'E',
         template: '<div></div>',
@@ -185,10 +184,6 @@
           var recalculateOptions = function() {
             scope.data.length = 0;
             scope.layout.showlegend = _.has(scope.options, 'legend') ? scope.options.legend.enabled : true;
-            if(_.has(scope.options, 'bottomMargin')) {
-              bottomMargin = parseInt(scope.options.bottomMargin);
-              scope.layout.margin.b = bottomMargin;
-            }
             delete scope.layout.barmode;
             delete scope.layout.xaxis;
             delete scope.layout.yaxis;
@@ -309,8 +304,7 @@
           scope.$watch('series', recalculateOptions);
           scope.$watch('options', recalculateOptions, true);
 
-
-          scope.layout = {margin: {l: 50, r: 50, b: bottomMargin, t: 20, pad: 4}, autosize: true, hovermode: 'closest'};
+          scope.layout = {margin: {l: 50, r: 50, b: 50, t: 20, pad: 4}, height: scope.height, autosize: true, hovermode: 'closest'};
           scope.plotlyOptions = {showLink: false, displaylogo: false};
           scope.data = [];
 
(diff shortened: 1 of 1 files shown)


rd_ui/app/scripts/directives/plotly.js
```diff
@@ -45,59 +45,82 @@
...
+    _.each(seriesList, function(series) {
+      if(!_.has(series,'visible')){
+        series.visible = true;
+        series.original_y = series.y.slice();
+      }
+    });
+  };
+
+  var getEnabledSeries = function(seriesList){
+    return _.filter(seriesList, function(series) {
```
arikfr: You can write this as `_.filter(seriesList, 'visible');`.

arikfr:
Overall looks good. Did you test it with a chart having multiple types of series (bars & line, for example)?

rd_ui/app/scripts/directives/plotly.js
```diff
@@ -288,6 +311,25 @@
           var element = element[0].children[0];
           Plotly.newPlot(element, scope.data, scope.layout, scope.plotlyOptions);
 
+          element.on('plotly_afterplot', function(d) {
+            if(scope.options.globalSeriesType === 'area' && (scope.options.series.stacking === 'normal' || scope.options.series.stacking === 'percent')){
+              d3.selectAll('.legendtoggle').each(function(d, i) {
+                // override plotly's default legend item on-click behaviour
```
arikfr: What's the default behavior? Do we "un-override" it when the types changes?

rd_ui/app/scripts/directives/plotly.js
machira: Actually, this would not be the same. Because of Javascript truthiness, when `series.visible = 'legendonly'` would evaluate to `true`.

rd_ui/app/scripts/directives/plotly.js
machira: Thank you for the suggestion. We went back and took a look at it. It turns out we can use a custom listener through d3, and remove it when chart type changes.

machira:
It looks like multiple series (bar and Area) breaks the stacking. I will keep looking at it, and update you if I find a workaround.

rd_ui/app/scripts/directives/plotly.js
arikfr: :+1:

arikfr:
@machira ok, let me know if you find a solution. We can merge as is if you prefer, because it's still better than current situation.

vorakumar:
Hey Arik, we have updated the PR with some new changes. Some additional tests here showed some edge cases we hadn't thought of- such as edit mode. The most recent push works as expected. And multiple series (area and line) also seem to work appropriately.

rd_ui/app/scripts/directives/plotly.js
```diff
@@ -121,7 +144,6 @@
   angular.module('plotly', [])
     .constant('ColorPalette', ColorPalette)
     .directive('plotlyChart', function () {
-      var bottomMargin = 50;
```
arikfr: Did you mean to remove this and the related code?

rd_ui/app/scripts/directives/plotly.js
```diff
@@ -286,14 +304,31 @@
...
           scope.plotlyOptions = {showLink: false, displaylogo: false};
           scope.data = [];
 
           var element = element[0].children[0];
           Plotly.newPlot(element, scope.data, scope.layout, scope.plotlyOptions);
 
+          element.on('plotly_afterplot', function(d) {
+            if(scope.options.globalSeriesType === 'area' && (scope.options.series.stacking === 'normal' || scope.options.series.stacking === 'percent')){
+              $(element).find(".legendtoggle").each(function(i, rectDiv) {
+                  d3.select(rectDiv).on('click', function () {
```
arikfr: Using `click` without a "namespace" mean we override the default?

arikfr:
@vorakumar thanks! I appreciate the effort you put here.

rd_ui/app/scripts/directives/plotly.js
vorakumar: yes. We tried using `namespace` but it didn't work out. The new listener method was never called.

vorakumar:
major work has been done by @machira

arikfr:
By "you" I meant all of you at ThoughtWorks :-)

rd_ui/app/scripts/directives/plotly.js
arikfr: And is there a way to get the previous click handler and call it inside our own?

I'm worried that at some point in the future, Plotly will add some additional logic there that we will miss.

machira: Yes, it is a risk that if Plotly updates that area, our stuff breaks. On the other hand, such an update may also fix the plotly issue, so we may want to remove this behaviour entirely. 
We tried a few things, but it didn't seem to work. Ultimately, we decided not to spend too much time on it since Plotly may be soon replaced by a different charting library. If it becomes an issue, we can revisit and rework this.

rd_ui/app/scripts/directives/plotly.js
arikfr: @machira @vorakumar ?

rd_ui/app/scripts/directives/plotly.js
arikfr: OK, sounds reasonable.

rohanpd:
I have seen splunk do this in a different way that might be worth considering:

image

When you hover over the legend it just highlights the slice that you are analysing:

image

In saying that, I'm happy with what you guys have suggested.

rd_ui/app/scripts/directives/plotly.js
machira: @arikfr, No, we didn't mean to do that. It's now fixed.

arikfr:
Merged :tada:

PR #1070, opened 2016-05-22
Strip HTML tags from Excel file download

danielerapati:
Html tags in result values make Excel files produced by drill-down queries (such as [this one](http://demo.redash.io/api/queries/514/results/1096872.xlsx) from [this query](http://demo.redash.io/queries/514#table)) a bit unreadable and not too useful.

This change removes the html tags and preserves the value (the text attribute inside the tag).

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,4 +1,4 @@
-from redash.utils import build_url, collect_query_parameters, collect_parameters_from_request
+from redash.utils import build_url, collect_query_parameters, collect_parameters_from_request, clean_from_html_tags
 from collections import namedtuple
 from unittest import TestCase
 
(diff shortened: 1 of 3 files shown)


tests/test_utils.py
```diff
@@ -49,3 +49,15 @@ def test_ignores_non_prefixed_values(self):
...
+class TestCleanFromHtmlTags(TestCase):
+    def test_removes_html_tags(self):
+        dirty_string = '<a href="some/site" target="_brank">value</a>'
+        self.assertEqual('value', clean_from_html_tags(dirty_string))
+
+    def test_ignores_number(self):
+        self.assertEqual(3, clean_from_html_tags(3))
+
+    def test_ignores_string_with_no_html_tag(self):
+        self.assertEqual('value', clean_from_html_tags('value'))
```
arikfr: Will it handle this case: `< whatever` properly? (and return `< whatever` back)

danielerapati: Good catch: haven't tried yet but I bet it raises some malformed html error
...

arikfr: If it raises an error, you might want to look at an alternative way to clean the HTML. I've seen a few very simple code samples on StackOverflow that do it with no external dependencies.

arikfr:
You're welcome to reopen if you ever get to rebasing & addressing the comments.

Thanks.

PR #1089, opened 2016-06-01
Add support for serialising UUID type within MSSQL #961

### LINKED ISSUE #961: Support for UUID type in MSSQL query runner
Currently it fails on serializing to JSON (need to add fallback in the serializer to cast into string).

James226:
Closes #961.


arikfr:
Thanks!

PR #1091, opened 2016-06-02
Add caching for queries used in embeds

whummer:
This PR adds caching of query results for queries used in embeds.

Following the notation already used in `query_view.js`, users can specify a *maxAge* parameter in the URL to limit the result's maximum cache age.

Tests have been updated to cover this new functionality.

### DIFF
--- a/redash/handlers/embed.py
+++ b/redash/handlers/embed.py
@@ -84,7 +84,7 @@ def embed(query_id, visualization_id, org_slug=None):
             # WARNING: Note that the external query parameters
             #          are a potential risk of SQL injections.
             #
-            max_age = int(request.args['maxAge']) if 'maxAge' in request.args else 0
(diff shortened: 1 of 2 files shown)


redash/handlers/embed.py
```diff
@@ -59,7 +84,8 @@ def embed(query_id, visualization_id, org_slug=None):
             # WARNING: Note that the external query parameters
             #          are a potential risk of SQL injections.
             #
-            results = run_query_sync(query.data_source, parameter_values, query.query)
+            max_age = int(request.args['maxAge']) if 'maxAge' in request.args else 0
```
arikfr: Can be simplified: `max_age = int(request.args.get('maxAge', 0))`.

whummer: updated, thanks.

arikfr:
Merged.

PR #1092, opened 2016-06-03
Add Heroku support

adamlwgriffiths:
This adds Heroku support, whilest trying to not inject any Heroku specific code in the way of other installation paths (ie, using bin/pre_compile to aggregate / manipulate files, heroku-postbuild in package.json, etc) as per [this comment](https://github.com/getredash/redash/issues/896#issuecomment-222597132).
The requirements handling is a bit more fragile than if cffi was just included directly.

Move .bowerrc, bower.json, gulpfile.js, package.json
down to root level.
Update paths in .bowerrc, gulpfile.js, Makefile
Add a heroku-postbuild step to package.json which
installs devDependencies and runs the build.
Add step in bin/pre_compile which adds the
requirements_all_ds.txt to requirements.txt to ensure that
cffi is installed. Also removes pymssql as this is
not supported on Heroku.
Add content from rd_ui/.gitignore to .gitignore and
remove rd_ui/.gitignore.
Add section in setup.rst about Heroku deployments.


adamlwgriffiths:
I've also noted that there are a number of obsolete files in rd_ui.
.gitattributes, .editorconfig, .jshint, .travis.yml.
I didn't remove those as I only wanted to do the minimum for the PR.

arikfr:
@adamlwgriffiths actually `.jshint` and `.editorconfig` are being used. `.gitattributes` and `.travis.yml` can be removed.

Procfile.heroku
```diff
@@ -0,0 +1,2 @@
+web: ./manage.py runserver -p $PORT --host 0.0.0.0
+worker: ./bin/run celery worker --app=redash.worker --beat -Q queries,celery,scheduled_queries
```
arikfr: No need to prefix with `bin/run` because Heroku loads the env variables for us.

bin/pre_compile
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# heroku post_compile script
```
arikfr: pre_compile :)

bin/pre_compile
```diff
@@ -0,0 +1,17 @@
...
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+pushd $DIR/..
+
+# heroku requires cffi to be in requirements.txt in order for libffi to be installed.
+# https://github.com/heroku/heroku-buildpack-python/blob/master/bin/steps/cryptography
+# to avoid making it a requirement for other build systems, we'll inject it now
+# into the requirements.txt file
+# remove pymssql, it doesn't compile on heroku
+grep -v '^pymssql' requirements_all_ds.txt >> requirements.txt
```
arikfr: The only concern here is that if in the future we will add some other library to requirements_all_ds that breaks Heroku deployment we won't know about it... :\

docs/setup.rst
```diff
@@ -79,6 +79,44 @@ Docker Compose
 Now proceed to `"Setup" <#setup>`__.
 
 
+Heroku
+------
```
arikfr: Worth adding a warning that currently they will need to figure out how to do migrations between upgrades on their own.

arikfr:
Thank you for your effort on this!

docs/setup.rst
```diff
@@ -79,6 +79,44 @@ Docker Compose
...
+    $ heroku addons:create heroku-postgresql:hobby-dev
+
+4. Add Redis addon::
+
+    $ heroku addons:create heroku-redis:hobby-dev
+
+5. Set environment variables::
+
+    $ heroku config:set REDASH_DATABASE_URL=`heroku config:get DATABASE_URL`
+    $ heroku config:set REDASH_REDIS_URL=`heroku config:get REDIS_URL`
```
arikfr: I'm totally cool with changing the relevant lines in `settings.py` to:

```python
DATABASE_CONFIG = parse_db_url(os.environ.get("REDASH_DATABASE_URL", os.environ.get("DATABASE_URL", "postgresql://postgres")))
```

And:

```python
REDIS_URL = os.environ.get('REDASH_REDIS_URL', os.environ.get('REDIS_URL', "redis://localhost:6379/0"))
```

I can do it myself after merging this pull request, if you prefer.

Procfile.heroku
adamlwgriffiths: I took that file verbatim from Procfile.dev
https://github.com/getredash/redash/blob/master/Procfile.dev
Not sure if that one needs updating too

bin/pre_compile
adamlwgriffiths: I could either add a `requirements-heroku.txt`, or split each db into it's own `requirements.txt`, and make `requirements_all_ds.txt` do a -r for each one. Either way, I can specify which libs I want.

docs/setup.rst
adamlwgriffiths: I don't understand the issue here. Is there a migration path that is handled in other deployments but not by this?
I haven't read anything about migrations in the docs, is there somewhere I can look at to understand the issue?

adamlwgriffiths:
I'll update based on your comments hopefully on Monday.
Let me know your thoughts on the `requirements.txt` and documentation.

bin/pre_compile
arikfr: I prefer to avoid having to maintain another requirements file. Let's keep as is, and revisit in the future.

Procfile.heroku
arikfr: Procfile.dev is used for Vagrant, where `bin/run` is actually needed.

docs/setup.rst
arikfr: For other deployments we recommend using the upgrade script, as documented here: http://docs.redash.io/en/latest/upgrade.html.

The way it runs migrations is by comparing existing migrations with new ones. Which is awful and can't be done in stateless environments like Heroku.

adamlwgriffiths:
Ok, I've made some changes as per your comments.
Let me know if there's anything not quite right.

arikfr:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Ac5bwzYA&#x2F;9-ability-to-run-re-dash-on-heroku"> Ability to run re:dash on Heroku</a>

arikfr:
Merged. Thank you!

adamlwgriffiths:
Awesome, thanks!

Just a reminder that this relies on the REDASH_DATABASE_URL and REDASH_REDIS_URL environment variable retrieval (fallback to DATABASE_URL and REDIS_URL) which has yet to be committed (for anyone else out there reading this).

arikfr:
I committed these changes before merging this pull request ;-)

adamlwgriffiths:
Sweet! Cheers

PR #1099, opened 2016-06-08
Fix RST syntax for links

adamlwgriffiths:
I derped with the link syntax in #1092.
It seems RST wants a space between the link text and the `<url>`.
My RST previewer didn't show any issue, but readthedocs doesn't like it and the link text displays the RST syntax and shows an invalid link.
I've tested this locally with sphinx and it resolves the issue.


arikfr:
Thanks.

(just another reminder that I should never used RST in the first place...)

PR #1100, opened 2016-06-08
fix vagrant provisioning script

ninneko:
This PR fix provision.sh for vagrant to add re:dash meta data to data sources as default.
This make it easy to create test query on development environment.


arikfr:
Thanks, but I prefer to keep the provision script as simple as possible to reduce chances it will fail.

If I'm not mistaken, adding Redash's own database is as simple as creating a new datasource of type Postgres and database name "redash" without any other parameters. Maybe a nice alternative to this pull request will be to mention this in the Vagrant documentation page?

PR #1102, opened 2016-06-08
Switch to HipChat V2 API

arikfr:
If you were using Re:dash with HipChat and already upgraded to 0.11 you will need to update manually HipChat's settings.


arikfr:
@rohanpd FYI.

PR #1103, opened 2016-06-08
Docs: add section about monitoring

AntoineAugusti:
Mention the `/status.json` endpoint. We use it with [Grafana](http://grafana.org) and it's super useful.

image

### DIFF
--- a/docs/usage/maintenance.rst
+++ b/docs/usage/maintenance.rst
@@ -70,3 +70,43 @@ Version
 See current version:
 
 ``bin/run ./manage.py version``
+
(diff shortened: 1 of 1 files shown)


AntoineAugusti:
@arikfr Happy to change the text if you've got comments

arikfr:
:+1: thanks!

Btw, we also report some stats using StatsD (if one is provided). Are you using Grafan with InfluxDB or Graphite? (or something else)

AntoineAugusti:
@arikfr With InfluxDB :)

PR #1105, opened 2016-06-09
Added support for "ObjectId" for querying fields that are of type ObjectId

erans:
For example:
{
	"collection" : "mycol",
	"query" : {
		"_id" : "ObjectId(\"xxxxxxxx\")"
	}
}

### DIFF
--- a/redash/query_runner/mongodb.py
+++ b/redash/query_runner/mongodb.py
@@ -43,14 +43,18 @@ class MongoDBJSONEncoder(JSONEncoder):
 
 
 date_regex = re.compile("ISODate\(\"(.*)\"\)", re.IGNORECASE)
+objectid_regex = re.compile("ObjectId\(\"(.*)\"\)", re.IGNORECASE)
 
-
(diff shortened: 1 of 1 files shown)


arikfr:
I believe we already support this with the support for Mongo Extended JSON.
You should be able to write this query as:

```json
"_id": {"$oid": "..."}
```

https://docs.mongodb.com/manual/reference/mongodb-extended-json/#oid

erans:
This is true when using Strict Mode - but not it's not always the case:
http://stackoverflow.com/questions/26938598/mongodb-oid-vs-objectid

This supports both modes.

arikfr:
It's not something that depends on your MongoDB server -- it's in the [implementation of PyMongo](https://github.com/mongodb/mongo-python-driver/blob/master/bson/json_util.py#L140-L141).

erans:
Heh... maybe its time I upgrade my PyMongo version :-|

OK. Scrape that.

PR #1111, opened 2016-06-10
Escape HTML characters before rendering in results table, fixes #755

### LINKED ISSUE #755: Results containing opening bracket (<) are not displayed
I have this query:

SELECT keywords from recent_searches;

Some of these contain HTML tags or quotes. Like so (output from postgres console):

```
 keywords      
```

---

 10916</cite><span
 11576</cite><span
 12739</cite><span
 1506'-'1
 21666</cite><span
 29002</cite><span
 30328</cite><span
 32710</cite><span
 44035</cite><span
 04289+and+1=(select+0)--+-
 04289+and+1=(select+1)--+-
 62244</cite><span
 93201</cite><span

(ok markdown chokes a bit on that... I'm not sure how to convey it better)

In the Redash query results, the "keywords" field shows up as blank for those items (which were search log entries from bots trying to perform SQL injection attacks). Every item listed above shows up as blank. Which is better than breaking something, but if that data can be escaped cleanly and displayed it would be an improvement.

James226:
This should solve HTML/XML being rendered in table results.

### DIFF
--- a/rd_ui/app/scripts/filters.js
+++ b/rd_ui/app/scripts/filters.js
@@ -120,4 +120,11 @@ angular.module('redash.filters', []).
           filtered.push(items[i])
       return filtered;
     };
+  })
+
+  .filter('escapeHtml', function () {
+    var element = angular.element('<pre/>');
+    return function (text) {
+      return element.text(text).html();
+    };
   });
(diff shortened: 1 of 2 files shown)


arikfr:
Although it's crude, allowing some HTML in table results is a *feature* to enable creating drill downs and other useful things. This pull request breaks it.

The ideal way will be to allow the user to decide if he wants to render HTML in the table row or not. This will happen when we implement [options for the table visualization](https://trello.com/c/tQxDwU4s/50-configuration-for-table-visualization).

For now to prevent breaking existing functionality, I will have to pass on this pull request.

Issue #876, opened 2016-02-29
Discussion: Switch from Peewee to SQLAlchemy


jeffwidman:
Can you clarify a bit why you went with Peewee instead of SQLAlchemy? 

The latter is a much more power ORM, versus the former focuses on simplicity. Peewee was probably the right choice in the early days of the project, but since you're looking to build this into a full-fledged business, wouldn't it be better to switch to SQLAlchemy? SQLAlchemy also offers the benefit of being able to drop into Core when the ORM isn't flexible enough. 

This assumes you're using the ORM for MySQL/Postgres connectors... If not, then I simply misunderstand how the architecture/connectors work.

arikfr:
We're using Peewee as an ORM for the metadata (users, queries, results, dashboards, etc). I chose Peewee because it was simple but powerful. I did give SQLAlchemy a try, but actually it failed me when I did want to run a raw SQL query and get back ORM objects (something that is really simple w/ Peewee).

But today I'm more open to trying out SQLAlchemy again:
1. Peewee doesn't have migrations support. There are projects like [arnold](https://github.com/cam-stitt/arnold), but they are very bare bones.
2. No optimistic-locking support.
3. SQLAlchemy is the de-facto ORM for Python, resulting in much wider support in other libraries.

Having said that, we're not really limited by Peewee's feature set. So I'm not sure what will be more work: switching to SQLAlchemy or creating our own support for migrations with Peewee.

jeffwidman:
dunno, but I personally am very comfortable with sqlalchemy and much prefer to hack on projects that use it because it's so powerful/flexible/commonly used--between Django ORM and SQLAlchemy, that's all I need to know for 99% of Python ORM projects... 

btw, I don't think `alembic` is tied to sqlalchemy, but I might be wrong... 

In general, I'd vote for SQLAlchemy just because it gives you options down the road if you decide you want to do something fancy for some reason

arikfr:
Just to follow up on this, `alembic` is SQLAlchemy specific. Also, for reference, another `peewee`/migrations project: https://github.com/klen/peewee_migrate.

I really want to address the issue of migrations already, and basically we need to decide between -
1. Keep Peewee and add migrations (either our own or one of the mentioned projects).
2. Switch to SQLAlchemy.

I suspect the first will be faster to implement, but moving to SQLAlchemy will probably pay off in the future.

jeffwidman:
[...]

PR #1113, opened 2016-06-13
Add: share modify/access permissions for queries and dashboard

whummer:
Hi @arikfr,

This PR allows users to share edit permissions for a given query or dashboard. This feature request has previously been [discussed here](https://discuss.redash.io/t/ability-to-contribute-to-someone-elses-query/14).

A new button *"Share Edit Permissions"* has been added to the "..." details popup when viewing the source of a query (or dashboard). 
image

To manage the access permissions, a modal dialog is used (similar to the dialog used for managing group members). To give an existing user "modify" access, the user can simply be selected from the search/dropdown list. To revoke access, use the "Remove" button and changes will be effected immediately. Note that the same modal code (view and controller) is used for both, query permissions and dashboard permissions, to avoid code duplication.
image

At the core of the approach is a new table and model entity `Change`, which keeps track of the historical changes made to model entities (i.e., Query, and Dashboard). Whenever a change is made, a new record is added to the `changes` table. The UI is aware of the current "version" of the query/dashboard it is working with, and the approach implements "optimistic locking" to avoid concurrent changes to override each other. If a conflict is detected, a warning message is displayed to the user (see below). This can be easily tested by using two browser sessions with two different users, attempting to modify the same query.
image

The access permissions are handled via a new API endpoint `/api/access/`, which allows to 1) list, 2) grant, and 3) revoke access permissions, as well as to 4) attempt an access and receive an access "granted" or "denied" decision.

A set of tests has been added to cover the basic functionality of managing access permissions.


arikfr:
Thanks! That's a lot of changes... I am hoping to start going through it this week, but it will take some time.

whummer:
Hi @arikfr, we rebased this branch to current `master`. Any chance we can get this released for the next milestone? Cheers

redash/handlers/access.py
```diff
@@ -0,0 +1,77 @@
...
+        req = request.get_json(True)
+        grantee = req['user_id']
+        access_type = req['access_type']
+
+        query = AccessPermission.delete()\
+                .where(AccessPermission.object_id == object_id)\
+                .where(AccessPermission.object_type == object_type)\
+                .where(AccessPermission.access_type == access_type)\
+                .where(AccessPermission.grantee == grantee)\
+                .where(AccessPermission.grantor == self.current_user)
```
arikfr: Does this mean that only the user who gave the permission can revoke it?

I think we can limit the look up here to `(object_id, object_type, grantee)`.

arikfr: Also this should be encapsulated in the model, so this becomes:

```python
permission = AccessPermission.get_by_object_and_grantee(object_type, object_id, grantee)
permission.delete_instance()
```

redash/handlers/access.py
```diff
@@ -0,0 +1,77 @@
...
+        deleted = query.execute()
+        result = {'deleted': deleted}
+        return result
+
+class AccessAttemptResource(BaseResource):
+
+    def post(self, object_type, object_id, access_type):
```
arikfr: Why is this a `POST` request and not `GET`?

redash/handlers/api.py
```diff
@@ -69,6 +70,11 @@ def json_representation(data, code, headers=None):
 api.add_org_resource(QueryRefreshResource, '/api/queries/<query_id>/refresh', endpoint='query_refresh')
 api.add_org_resource(QueryResource, '/api/queries/<query_id>', endpoint='query')
 
+api.add_org_resource(AccessListResource, '/api/access/<object_type>/<object_id>', endpoint='list_access')
+api.add_org_resource(AccessGrantResource, '/api/access/<object_type>/<object_id>', endpoint='grant_access')
+api.add_org_resource(AccessRevokeResource, '/api/access/<object_type>/<object_id>', endpoint='revoke_access')
```
arikfr: How about we name this resource as `access_permission`?

Also you can merge the last 3 classes into one (assuming you change the access check to GET). So it becomes `AccessPermissionListResource` and `AccessPermissionResource`.

redash/models.py
```diff
@@ -757,6 +759,94 @@ def __unicode__(self):
         return unicode(self.id)
 
 
+class AccessPermission(BaseModel):
+
+    id = peewee.PrimaryKeyField()
+    object_type = peewee.CharField()
+    object_id = peewee.IntegerField()
```
arikfr: We should add an index on `object_id, object_type`.

redash/models.py
```diff
@@ -757,6 +759,94 @@ def __unicode__(self):
...
+    object = GFKField('object_type', 'object_id')
+    access_type = peewee.CharField()
+    grantor = peewee.ForeignKeyField(User, related_name='grantor')
+    grantee = peewee.ForeignKeyField(User, related_name='grantee')
```
arikfr: How about we make `grantee` a GFK field as well? So we can later on add the option to grant to permission to an `ApiKey` object or a `Group` object.

It's NTH so we can skip this for now, as it should be an easy migration.

redash/models.py
```diff
@@ -757,6 +759,94 @@ def __unicode__(self):
...
+                .where(AccessPermission.access_type == access_type)
+
+        return q.execute()
+
+
+class Change(BaseModel):
+
+    id = peewee.PrimaryKeyField()
+    object_id = peewee.CharField()
+    object_type = peewee.CharField()
```
arikfr: Same here re. index for object.

redash/handlers/access.py
```diff
@@ -0,0 +1,77 @@
...
+            if perm.access_type not in result:
+                result[perm.access_type] = []
+            result[perm.access_type].append(perm.grantee.to_dict())
+
+        return result
+
+
+class AccessGrantResource(BaseResource):
+
+    def post(self, object_type, object_id):
```
arikfr: You need to make sure here that the user is admin or owner of the granted object.

redash/handlers/access.py
```diff
@@ -0,0 +1,77 @@
...
+        if permissions.count() > 0:
+            return {'result': 'already_granted'}
+
+        perm = AccessPermission()
+        perm.object_type = object_type
+        perm.object_id = object_id
+        perm.access_type = access_type
+        perm.grantor = self.current_user
+        perm.grantee = grantee
+        perm.save()
```
arikfr: The logic for granting permission should be encapsulated in a model method of AccessPermission.

redash/handlers/access.py
```diff
@@ -0,0 +1,77 @@
...
+        return result
+
+class AccessAttemptResource(BaseResource):
+
+    def post(self, object_type, object_id, access_type):
+        access = AccessPermission.exists(user=self.current_user, access_type=access_type, object_id=object_id, object_type=object_type)
+        if access:
+            return {'result': 'access_granted'}
+        abort(403)
+        return False
```
arikfr: `abort` throws an exception, so this line is not needed.

redash/handlers/dashboards.py
```diff
@@ -8,6 +11,19 @@
 from redash.handlers.base import BaseResource, get_object_or_404
 
 
+def _save_change(user, dashboard_id, old_dashboard, new_dashboard, change_type):
```
arikfr: I think we can generalize this and have a single method for all change-tracking objects. Also it should be at the model level.

Ideally, it will be a hook that is being called on the model's `#save` method, but at that point we don't know who the user who made the change :-\ Maybe we can add another save method named `tracked_save` that will also take as a parameter the user who made the change.

One benefit of doing this at the model level, is that we can leverage `dirty_fields` property of the model (that will return the fields that were changed).

redash/handlers/dashboards.py
```diff
@@ -25,6 +41,11 @@ class DashboardListResource(BaseResource):
     def get(self):
         dashboards = [d.to_dict() for d in models.Dashboard.all(self.current_org, self.current_user.groups, self.current_user)]
 
+        for dashboard in dashboards:
+            last_change = models.Change.get_latest(object_id=dashboard['id'], object_type=models.Dashboard.__name__)
+            if last_change:
+                dashboard['latest_version'] = last_change.id
```
arikfr: The current version of a dashboard should be the tracked object's property. It will save us from doing this lookup, but also will be used for the locking.

redash/handlers/dashboards.py
```diff
@@ -49,18 +77,45 @@ def get(self, dashboard_slug=None):
...
         return response
 
     @require_permission('edit_dashboard')
     def post(self, dashboard_slug):
         dashboard_properties = request.get_json(force=True)
         # TODO: either convert all requests to use slugs or ids
         dashboard = models.Dashboard.get_by_id_and_org(dashboard_slug, self.current_org)
+
+        # check access permissions
+        if self.current_user.id != dashboard.user.id:
```
arikfr: We should use [`is_admin_or_owner`](https://github.com/getredash/redash/blob/master/redash/permissions.py#L64) here (this is the helper I mentioned in another comment).

redash/handlers/dashboards.py
```diff
@@ -49,18 +77,45 @@ def get(self, dashboard_slug=None):
...
+                    object_id=dashboard.id,
+                    object_type=models.Dashboard.__name__):
+                abort(403)
+
+        # Optimistic locking: figure out which user made the last
+        # change to this dashboard, and bail out if necessary
+        last_change = models.Change.get_latest(object_id=dashboard.id, object_type=models.Dashboard.__name__)
+        if last_change and 'latest_version' in dashboard_properties:
+            if last_change.id > dashboard_properties['latest_version']:
+                abort(409) # HTTP 'Conflict' status code
```
arikfr: This is *very* optimistic locking :-) As it hopes that no one else will save anything in the time that passes from this check until line 110. While usually it's OK, it's still a risk.

The optimistic locking I was referring to is what [Charles suggested in this comment](https://github.com/getredash/redash/issues/876#issuecomment-223488522):

```python
class BaseVersionedModel(BaseModel):
    version = IntegerField(default=0)

    def save_optimistic(self):
        if not self.version:
            return self.save()  # Since this is an `INSERT`, just call regular save method.

        # Update any data that has changed and bump the version counter.
        field_data = dict(self._data)
        current_version = field_data.pop('version', 0)
        field_data = self._prune_fields(field_data, self.dirty_fields)
        if not field_data:
            raise ValueError('No changes have been made.')

        ModelClass = type(self)
        field_data['version'] = ModelClass.version + 1  # Atomic increment

        query = ModelClass.update(**field_data).where(
            (ModelClass.version == current_version) &
            (ModelClass.id == self.id))

        nrows = query.execute()
        if nrows == 0:
            # It looks like another process has updated the version number.
            raise ConflictDetectedException()  # Raise exception? Return False?
        else:
            self.version += 1  # Update in-memory version number.
            return nrows
```

Note the use of update + where. It might be needed some changes, to use our callbacks for pre/post save.

redash/handlers/api.py
arikfr: After looking at the usage, I think we should change this to be: `/api/<object_type>/<object_id>/acl`, and `object_type` should be `queries` and `dashboards` to be consistent with the rest of the API. In the handler itself, we can map it to `Query` and `Dashboard` (or the object class).

arikfr:
Finally reviewed the code. I mostly reviewed the server code... I will go over the frontend code in more detail as we progress.

Sorry for the lots of comments, it just that this is a critical piece of the system...

whummer:
Awesome, thanks for the detailed feedback. We'll try our best to address your comments in the next couple of days... Cheers

redash/handlers/access.py
whummer: Intuitively, `GET` to me is more about getting the current state from an API, whereas `POST` is more where you actively try to run an action against the API (in this case posting a request to try and obtain an access permission). That's why we chose `POST` here. But I'm happy to change this to `GET` instead.

arikfr: It's not really trying to obtain access, but really checking the access status. No resources get changed and it's a "safe" action, and hence semantically it doesn't feel right (to me) to use POST here.

redash/handlers/access.py
whummer: Done, this has been moved to the model, and we're now verifying that the current user is the object owner or admin (using the helper function).

redash/handlers/api.py
whummer: Done. Adapted the API layout accordingly.

redash/models.py
whummer: Done.

redash/models.py
whummer: Done.

redash/handlers/access.py
whummer: Done.

redash/handlers/access.py
whummer: Done. Moved to `AccessPermission` model.

redash/handlers/access.py
whummer: good catch, thanks. removed.

redash/handlers/dashboards.py
whummer: Very good point, our initial implementation was indeed a bit messy and not generic at all. 

This has been refactored and completely generalized. The `_save_change` function has been removed from `dashboards.py` and `queries.py`, and instead we now handle this functionality in the Model directly (see `Change.save_change`). This way things are much cleaner and more re-usable.

redash/handlers/dashboards.py
whummer: Done. This change also impacted other parts of the code (e.g., the UI code where we previously tracked a propery `latest_version` which is now simply the `version` property of the model instance).

We've re-run all tests and did additional manual testing to ensure that the original functionality is still preserved.

redash/handlers/dashboards.py
whummer: Good point. Done.

redash/handlers/dashboards.py
whummer: This part of the handler has also been updated to work with the new `version` property in the model. Conceptually, it should provide the same functionality as the snippet that Charles had suggested.

whummer:
Hey @arikfr, we've carefully considered the feedback and refactored the implementation based on your suggestions.

The major changes are around adding the `version` field to the model classes (`Query` and `Dashboard` for now), as well as moving most of the core functionality from the handlers to the model classes, which indeed makes the code much more readable and reusable.

The PR is now ready for a second round of review. Looking forward to receiving more feedback. Cheers.

arikfr:
@whummer super. I will review the UI code now and then give the server code a second pass. Thank you for the rapid follow up.

redash/handlers/access.py
```diff
@@ -0,0 +1,103 @@
...
+        # make sure the current user is permitted to perform this operation
+        target_object = clazz.select().where(clazz.id == object_id).get()
+        require_admin_or_owner(target_object.user.id)
+
+        req = request.get_json(True)
+        grantee = req['user_id']
+        access_type = req['access_type']
+
+        permissions = AccessPermission.find(grantee=grantee, object_id=object_id,
+            object_type=object_type, access_type=access_type, grantor=self.current_user)
```
arikfr: Usage of `grantor` here has the same issue - permission might be granted already by another user.

redash/handlers/access.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if permissions.count() > 0:
+            return {'result': 'already_granted'}
+
+        AccessPermission.grant_permission(object_type=object_type,
```
arikfr: The model code (AccessPermission) should take care of converting object to `object_type`, `object_id` pair. This goes both for `grant_permission` and `find`. There is an example in the `ApiKey` model how to do this:

```python
    @classmethod
    def get_by_object(cls, object):
        return cls.select().where(cls.object_type==object._meta.db_table, cls.object_id==object.id, cls.active==True).first()

    @classmethod
    def create_for_object(cls, object, user):
        return cls.create(org=user.org, object=object, created_by=user)
```

In create operations the `GFKField` class already takes care of it, and in read operations we need to do it.

redash/handlers/dashboards.py
```diff
@@ -56,11 +62,29 @@ def post(self, dashboard_slug):
...
+        if not redash.permissions.is_admin_or_owner(object_owner_id=dashboard.user.id):
+            if not self.current_user.has_access(
+                    access_type=models.AccessPermission.ACCESS_TYPE_MODIFY,
+                    object_id=dashboard.id,
+                    object_type=models.Dashboard.__name__):
+                abort(403)
+
+        # Optimistic locking: figure out which user made the last
+        # change to this dashboard, and bail out if necessary
+        last_change = models.Change.get_latest(object_id=dashboard.id, object_type=models.Dashboard.__name__)
```
arikfr: Why do we still use the `Change`'s version?

redash/handlers/dashboards.py
```diff
@@ -56,11 +62,29 @@ def post(self, dashboard_slug):
...
         # TODO: either convert all requests to use slugs or ids
         dashboard = models.Dashboard.get_by_id_and_org(dashboard_slug, self.current_org)
+
+        # check access permissions
+        if not redash.permissions.is_admin_or_owner(object_owner_id=dashboard.user.id):
+            if not self.current_user.has_access(
+                    access_type=models.AccessPermission.ACCESS_TYPE_MODIFY,
+                    object_id=dashboard.id,
+                    object_type=models.Dashboard.__name__):
+                abort(403)
```
arikfr: Let's have this as another helper function in `redash.permissions`:

```python

def can_modify(object, user):
    return is_admin_or_owner(object.user_id) or user.has_access(object, models.AccessPermission.ACCESS_TYPE_MODIFY)

def require_modify_permission(object, user):
    if not can_modify(object, user):
        abort(403)
```

redash/handlers/queries.py
```diff
@@ -94,17 +112,21 @@ def post(self, query_id):
 
         query_def['last_modified_by'] = self.current_user
 
-        query.update_instance(**query_def)
+        old_query = copy.deepcopy(query.to_dict())
+        new_change = query.update_instance_tracked(changing_user=self.current_user, old_object=old_query, **query_def)
```
arikfr: Why not `tracked_save`?

redash/models.py
```diff
@@ -344,6 +343,10 @@ def update_group_assignments(self, group_names):
         self.groups = map(lambda g: g.id, groups)
         self.save()
 
+    def has_access(self, access_type, object_id, object_type):
```
arikfr: Maybe access_type should be last?

redash/models.py
```diff
@@ -757,6 +776,190 @@ def __unicode__(self):
...
+            q = q.where(AccessPermission.grantor == grantor)
+        return q
+
+    @staticmethod
+    def exists(object_id, object_type, grantee, access_type=None):
+        try:
+            permissions = AccessPermission.find(object_id=object_id,
+                object_type=object_type, grantee=grantee, access_type=access_type)
+            return permissions.count() > 0
+        except Exception, e:
```
arikfr: Why would we have an exception here?

redash/models.py
```diff
@@ -757,6 +776,190 @@ def __unicode__(self):
...
+    def grant_permission(object_type, object_id, access_type, grantee, grantor):
+        perm = AccessPermission()
+        perm.object_type = object_type
+        perm.object_id = object_id
+        perm.access_type = access_type
+        perm.grantor = grantor
+        perm.grantee = grantee
+        perm.save()
+
+    @staticmethod
```
arikfr: It should be a `@classmethod` and then use the `cls` function parameter instead of `AccessPermission`.

redash/models.py
```diff
@@ -757,6 +776,190 @@ def __unicode__(self):
...
+            permissions = AccessPermission.find(object_id=object_id,
+                object_type=object_type, grantee=grantee, access_type=access_type)
+            return permissions.count() > 0
+        except Exception, e:
+            logging.warn("Unable to query for AccessPermission: %s" % e)
+            return False
+
+    @staticmethod
+    def reset(object_type, object_id, access_type):
+        revoke_permission(object_type=object_type, object_id=object_id, access_type=access_type)
```
arikfr: Why have both `reset` and `revoke_permission`?

redash/models.py
```diff
@@ -757,6 +776,190 @@ def __unicode__(self):
...
+
+    def delete_instance(self, *args, **kwargs):
+        super(AccessPermission, self).delete_instance(*args, **kwargs)
```
arikfr: ?

redash/models.py
```diff
@@ -757,6 +776,190 @@ def __unicode__(self):
...
+
+    def to_dict(self):
+        d = {
+            'id': self.id,
+            'object_id': self.object_id,
+            'object_type': self.object_type,
+            'access_type': self.access_type,
+            'grantor': self.grantor.id,
+            'grantee': self.grantee.id
```
arikfr: Use `self.grantor_id` and `self.grantee_id` or it will query the database for the user models here...

redash/models.py
```diff
@@ -757,6 +776,190 @@ def __unicode__(self):
...
+        try:
+            query = Change.select(Change)\
+                .where(Change.object_id == object_id)\
+                .where(Change.object_type == object_type)
+            if change_type:
+                query = query.where(Change.change_type == change_type)
+            query = query.order_by(Change.created_at.desc())
+            for change in query:
+                return change
+        except Exception, e:
```
arikfr: Why would we have an exception here?

redash/models.py
```diff
@@ -757,6 +776,190 @@ def __unicode__(self):
...
+    @staticmethod
+    def get_latest(object_id, object_type, change_type=None):
+        try:
+            query = Change.select(Change)\
+                .where(Change.object_id == object_id)\
+                .where(Change.object_type == object_type)
+            if change_type:
+                query = query.where(Change.change_type == change_type)
+            query = query.order_by(Change.created_at.desc())
+            for change in query:
```
arikfr: It will query the database for all the changes, create objects for them and then use only the first...

Just do: `return query.limit(1).first()` (`first()` by itself doesn't limit the result set).

redash/models.py
```diff
@@ -945,6 +1150,13 @@ def recent(cls, org, groups, user_id, for_user=False, limit=20):
     def get_by_slug_and_org(cls, slug, org):
         return cls.get(cls.slug == slug, cls.org==org)
 
+    def tracked_save(self, changing_user, old_object=None, *args, **kwargs):
+        self.version += 1
+        self.save(*args, **kwargs)
+        # save Change record
+        new_change = Change.save_change(user=changing_user, old_object=old_object, new_object=self)
+        return new_change
```
arikfr: Instead of duplicating the `tracked_save` logic, let's have a mixin for this like we have for timestamps. 

Also I'm not happy with the added complexity of the way the actual change is tracked. We currently have no use for the change itself, how about we don't implement this as part of this pull request?

arikfr: Also I just noticed that you didn't implement the actual optimistic locking here...

whummer:
Thanks @arikfr again for the detailed comments. Will try to address them within the next couple of days. Also, I added you as collaborator to the forked repo, please feel free to add the tests there. Thanks

whummer:
Hey @arikfr , I saw that you did a couple of updates, thanks for the changes. How do you suggest we split up the remaining work for this PR? I can reserve some time to work on it this week. Please let me know. Cheers

arikfr:
@whummer Let's see what I manage to do tomorrow -- I might just finish all the need updates myself.

arikfr:
phew... almost done. all that's left:

- [ ] Event tracking for permissions modifications
- [ ] Reimplement _change tracking_ with the new Mixin.
- [ ] Wrap the UI for managing permissions with a feature flag. I want to more gradually roll this out.
- [ ] Make sure that partial updates to the query pass the current version.
- [ ] Dashboards: adding widget works regardless of version. Need to decide if it's a wanted behavior.
- [ ] Assert access to the data source when changing the query.

arikfr:
OK, I'm done (!). The UI for managing permissions is behind a feature flag and turned off by default. There are few "open" issues that we will address before we "graduate" this feature:

- Multiple users can add widgets to the dashboard without getting the conflict error. Need to decide if it's an issue.
- We need to decide how to handle the case when someone gives you edit permissions to a query you don't have access to its data source. We should probably not allow this in the first place.
- We need to send an email / notification to the grantee once he's granted access.

Anything I forgot?

I'm not merging this yet, as I want to merge a few other things first, but I will merge it tomorrow.

whummer:
Wow @arikfr , awesome work. Thanks for putting so much effort into finishing off this PR. Super excited to see this merged tomorrow.

adamlwgriffiths:
We've noticed the 'the dashboard has been changed by another user' message come up when there should be no reason. Not sure what has triggered it, but attempting a second save 'fixed it'.

arikfr:
@adamlwgriffiths does it come  up every time? if not, any ideas on how to reproduce?

adamlwgriffiths:
I haven't been able to pin it down, so these may not be correct as I may have performed other actions prior / in-between.

The first time I saved a dashboard (created prior to this patch) the message appeared.
Opening a dashboard, then editing a widget / query in another tab, would cause this message when I then went to re-order widgets using the 'edit dashboard' dialog.
I've also received it when editing queries, can't put a finger on when. I've also noticed that a 'query saved' dialog will appear below the 'changed by another user' dialog.

Most of these were with a single user on the system, so any conflict is either from multiple tabs or bugs. It's still a frustrating dialog and really drags down the UX.

arikfr:
I think I know what happened in your case:
1. Add widget to the dashboard.
2. Try to rearrange them.
3. You get the conflict message.

That's obviously not the intended behavior and will be fixed before wrapping up v0.12.

In normal situations you won't see this message often (or at all). I think one improvement we can introduce is to give you the option to override the changes anyway.

ssoltane:
Hi,
I updated my redash version from v0.11 to v0.12. But i can't see the feature. 
Is there a flag that i need to activate ? 

Thanks

adamlwgriffiths:
Which feature specifically?

ssoltane:
I can't see the "Share edit permission" for queries or dashboards when i click on "..."

adamlwgriffiths:
I believe this feature is in the new UI refactor / migration to Angular 1.5 (see PR #1376) and merged in November 28 2016.
This wasn't part of the v0.12 release which was created on November 20 2016.
It is only in the master branch at this time.

arikfr:
@ssoltane It's part of 0.12. As mentioned in the comments, the UI is disabled with a feature flag. You need to add the environment variable FEATURE_SHOW_PERMISSIONS_CONTROL and set it to `true`.

If you used our AMIs or bootstrap script, then just add this to the `/opt/redash/.env` file.

ssoltane:
Super thks for the information :).

adamlwgriffiths:
@arikfr I'm not sure if this is still relevant (I'm using angular 1.2).

I'm able to reliablyreplicate the 'query changed' dialog by changing the data source from the drop down and hitting save.

arikfr:
@adamlwgriffiths I don't see it happen in v1.0.0.

galgoogle:
@arikfr , I've added the FEATURE_SHOW_PERMISSIONS_CONTROL, but still can't see the edit permissions. I'm using version 2.0.0.

satyadeepk:
This is the right environment variable to be added to `/opt/redash/.env`
```
export REDASH_FEATURE_SHOW_PERMISSIONS_CONTROL="true"
```
It works with this.
image

galgoogle:
@satyadeepk I've added it, and I see now "Manage Permissions". But when using it, it doesn't affect anything.

PR #1127, opened 2016-06-15
Mark basic_auth_password as secret

adamlwgriffiths:
Correctly marks the Elasticsearch datasource password as being secret.
Without this, passwords are included in serialized elasticsearch datasources.


arikfr:
Thanks!

PR #1136, opened 2016-06-17
Feature: add --org option to all relevant CLI commands.

adamlwgriffiths:
--org is the organization slug, not the name.
Allows the management of users, datasources and groups
with respect to organisations.
All commands default to 'default' slug, or None where
relevant, which means the commands will still work
as they did before without any changes.

### DIFF
--- a/redash/cli/users.py
+++ b/redash/cli/users.py
@@ -23,7 +23,7 @@ def build_groups(org, groups, is_admin):
 
 @manager.option('email', help="email address of the user to grant admin to")
 @manager.option('--org', dest='organization', default='default', help="The organization the user belongs to")
-def grant_admin(email, organization):
+def grant_admin(email, organization='default'):
     try:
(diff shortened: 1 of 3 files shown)


redash/cli/users.py
```diff
@@ -7,8 +7,7 @@
 manager = Manager(help="Users management commands. This commands assume single organization operation.")
```
AntoineAugusti: You will need to update this :)

adamlwgriffiths:
I've updated that text for both `users` and `groups` managers.
I added a missing organization parameter to the `users.list` function.
I added appropriate defaults to the organization parameters to allow calling the functions manually without having to specify the organization.
I also moved the organization parameter to the end of each function it was added to, to avoid breaking any existing code.

arikfr:
Merged. Thanks!

I'll be happy to hear on how you're using this. There is no email in your profile, so you're welcome to reach out (arik at redash.io).

PR #1143, opened 2016-06-19
Fix: use the email input type where needed

ariarijp:
I think, use the email input type for a good UX.

### DIFF
--- a/rd_ui/app/login.html
+++ b/rd_ui/app/login.html
@@ -44,7 +44,7 @@
   <form role="form" method="post" name="login">
     <div class="form-group">
       <label for="inputEmail">Email</label>
-      <input type="text" class="form-control" id="inputEmail" name="email" value="{{email}}">
+      <input type="email" class="form-control" id="inputEmail" name="email" value="{{email}}">
     </div>
     <div class="form-group">
       <label for="inputPassword">Password</label>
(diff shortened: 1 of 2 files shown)


rd_ui/app/login.html
```diff
@@ -44,7 +44,7 @@
   <form role="form" method="post" name="login">
     <div class="form-group">
       <label for="inputEmail">Email</label>
-      <input type="text" class="form-control" id="inputEmail" name="email" value="{{email}}">
+      <input type="email" class="form-control" id="inputEmail" name="email" value="{{email}}">
```
arikfr: While it makes sense, it's a problem as sometimes "email" is used just as a username -- like the default `admin` user. Changing type here will prevent people from logging in...

redash/templates/forgot.html
```diff
@@ -13,7 +13,7 @@
 <form role="form" method="post">
     <div class="form-group ">
         <label for="email">Enter the email address you used for this account:</label>
-        <input type="text" class="form-control" name="email" value="{{email}}">
+        <input type="email" class="form-control" name="email" value="{{email}}">
```
arikfr: Here we can keep the email type, as it makes no sense to use the "forgot password" page for non emails.

My suggestion: update the pull request to change the type only here.

arikfr:
Thanks! See my comments.

redash/templates/forgot.html
ariarijp: Thank you for your review! I will do.

ariarijp:
Updated this pull request.

Please review it again.

arikfr:
Thanks!

PR #1146, opened 2016-06-20
Docs: add a link to settings documentation.

adamlwgriffiths:
The settings page doesn't seem to have any links to it.
The only way to find it is through searching for it.
So we'll add one to the setup.rst page.


arikfr:
Thanks!

PR #1147, opened 2016-06-20
Control celery worker count with environment variables, defaults to 2

adamlwgriffiths:
Add to, or replace all instances of -c2, in celery commands
with -c${REDASH_CELERY_WORKER_COUNT:-2}.
Some commands had no -c value, most had -c2, 1 had -c1.
Add REDASH_CELERY_WORKER_COUNT to the settings page.

### DIFF
--- a/Procfile.heroku
+++ b/Procfile.heroku
@@ -1,2 +1,2 @@
 web: ./manage.py runserver -d -r -p $PORT --host 0.0.0.0
-worker: celery worker --app=redash.worker -c2 --beat -Q queries,celery,scheduled_queries
+worker: celery worker --app=redash.worker -c${REDASH_HEROKU_CELERY_WORKER_COUNT:-2} --beat -Q queries,celery,scheduled_queries
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -14,6 +14,7 @@ The follow is a list of settings and what they control:
 - **REDASH_DATABASE_URL**: *default "postgresql://postgres"*
 - **REDASH_CELERY_BROKER**: *default REDIS_URL*
 - **REDASH_CELERY_BACKEND**: *default CELERY_BROKER*
+- **REDASH_HEROKU_CELERY_WORKER_COUNT**: *default 2*
 - **REDASH_QUERY_RESULTS_CLEANUP_ENABLED**: *default "true"*
 - **REDASH_QUERY_RESULTS_CLEANUP_COUNT**: *default "100"*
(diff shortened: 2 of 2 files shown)


adamlwgriffiths:
It may be more desirable to define a per-queue variable, Ie. one variable for queries, and another for scheduled_queries.

arikfr:
Let's keep this only for the Heroku configuration (and document it in the Heroku section).

Many people have already the existing Supervisor configuration files, and having to maintain the [documentation](http://docs.redash.io/en/latest/usage/maintenance.html#changing-the-number-of-workers) on how to change number of workers for both versions feels to me like more confusing than adding value.

adamlwgriffiths:
I'm happy to modify the supervisord config files to use the [environment option](http://supervisord.org/subprocess.html#subprocess-environment).
I can also update the documentation to reflect this.

That said, I can understand the desire to not break existing deployments.
If you still want me to, I can back-out the non-Heroku changes.
But I think it is important to perform this change eventually, for consistencies sake.

adamlwgriffiths:
Changing environment / variables should be preferred over changing scripts.
For me, modifying the Procfile would mean that I cannot always do a merge without conflict. It's easier for me to move as much into variables as possible, so the deployment / running scripts remain vanilla.

For supervisord, moving the variable into the `environment` option doesn't prevent user's from having to alter the script to change the value, which could again lead to merge conflicts.
The variable can be set in the user's environment, but would need to be documented as to how.
The `ubuntu` and `amazon_linux` supervisord configs don't specify a user to run as, which I assume means it defaults to either `root` or `supervisord` users.

This is beyond my usage at the moment, so I'll defer this to you.

arikfr:
> For supervisord, moving the variable into the environment option doesn't prevent user's from having to alter the script to change the value, which could again lead to merge conflicts.

In the context when supervisor is used, those files aren't used from the repo but rather placed "statically" in the user's setup. They don't change when upgrading versions.

> The ubuntu and amazon_linux supervisord configs don't specify a user to run as, which I assume means it defaults to either root or supervisord users.

Supervisor itself is started with the `redash` user, and therefore we don't specify the user to use in supervisor's configuration.

> If you still want me to, I can back-out the non-Heroku changes.

Yes, please. 👍

adamlwgriffiths:
Sorry for the delay, I've been sick.

I've reduced the commit back to just `Procfile.heroku` and `settings.rst`.
I've changed the variable to be `REDASH_HEROKU_CELERY_WORKER_COUNT`.

arikfr:
Thanks!

I hope you feel better now.

PR #1151, opened 2016-06-22
Change: ability to set pymssql TDS version (to support Azure Database) #backward-incompatible

masaedw:
Azure Database requires tds_version 7.0 but pymssql 2.1.1 doesn't have option of tds_version.

* Update pymssql to 2.1.2 (added tds_version option from this version)
* Add configuration of tds_version (At least azure requires 7.0 so for now default version is it. However I don't know how is the best choice.)

### DIFF
--- a/docs/datasources.rst
+++ b/docs/datasources.rst
@@ -258,4 +258,4 @@ Microsoft SQL Server
 -  **Additional requirements**:
 
    - ``freetds-dev`` C library
-   - ``pymsssql`` python package, requires FreeTDS to be installed first
+   - ``pymssql`` python package, requires FreeTDS to be installed first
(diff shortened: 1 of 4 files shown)


arikfr:
Thank you!

arikfr:
According to [pymssql's documentation](http://pymssql.org/en/latest/ref/pymssql.html#pymssql.connect), Azure requires TDS version 7.1. Are you sure about the 7.0 version?

masaedw:
When I written that codes,  if I remember right, 7.1 is not accepted. But now, both 7.0 and 7.1 (and 7.2) are OK. 7.3 is NG. I'm using V12 Database.

According to [Azure SQL Database General Limitations and Guidelines](https://azure.microsoft.com/en-us/documentation/articles/sql-database-general-limitations/), Azure Database supports tds version 7.3 or later. :confused:

You can change the default version.

PR #1154, opened 2016-06-23
improve dashboard UI.

ninneko:
modify dashboard and embed visualization UI for emphasizing visualization name above query name.
dashboard
embeded query


arikfr:
Because not every widget will show visualization name (it's hidden in case it's the default), it doesn't feel right to show it first. We can make the font larger though.

PR #1156, opened 2016-06-23
Docs: add SSL parameters to nginx configuration

jcox92:
These parameters prevent some common exploits and vulnerabilities. My primary reference was here: https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html


arikfr:
Thanks!

PR #1157, opened 2016-06-24
Add annotate_query method for oracle runner.

nobuhito:


arikfr:
Do you mind adding some context for reasoning?

nobuhito:
When I let SQL run for the first time, an error was indicated.
When the "annotate_query" method was added with that, an error went off.
But a comment out did its method, an error isn't indicated now.
It may be the different cause, so if it's unnecessary, please close.

redash_oracle_error

It's machine translation. I'm sorry to be difficult to understand.

arikfr:
Yeah, the issue you had is different and not related to having the method in the class (it has it anyway, derived from parent).

PR #1170, opened 2016-07-05
Fix: change parameters for 'manage.py ds new' CLI

WesleyBatista:
When running the setup script `redash/setup/docker/create_database.sh` we get the following error:
```
usage: manage.py [-h]
                 {status,shell,users,database,send_test_mail,runserver,version,groups,org,ds,runworkers,check_settings}
                 ...
manage.py: error: unrecognized arguments: -n
```

This PR fixes the error removing the **-n** parameter and using **--options** and **--type** as well, instead of _-o_ and _-t_.


arikfr:
Thanks!

PR #1171, opened 2016-07-06
Fix: "download dataset" in embed with parameters

fabito:
Missing "id" and "retrieved_at" properties in JSON serialized QueryResul was affecting "download dataset" in embed with parameter pages.

### DIFF
--- a/redash/handlers/embed.py
+++ b/redash/handlers/embed.py
@@ -17,6 +17,7 @@ from redash.handlers.query_results import collect_query_parameters
 from redash.permissions import require_access, view_only
 from authentication import current_org
 
+
 #
 # Run a parameterized query synchronously and return the result
 # DISCLAIMER: Temporary solution to support parameters in queries. Should be
@@ -41,7 +42,7 @@ def run_query_sync(data_source, parameter_values, query_text, max_age=0):
 
     if query_result:
         logging.info("Returning cached result for query %s" % query_hash)
-        return query_result.data
+        return query_result
 
     try:
         started_at = time.time()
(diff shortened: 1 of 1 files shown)


arikfr:
Looks good to me, but @whummer it touches your code - mind reviewing?

whummer:
LGTM, thanks @fabito

redash/handlers/embed.py
```diff
@@ -55,8 +56,10 @@ def run_query_sync(data_source, parameter_values, query_text, max_age=0):
             query_result, updated_query_ids = models.QueryResult.store_result(data_source.org_id, data_source.id,
                                                                                   query_hash, query_text, data,
                                                                                   run_time, utils.utcnow())
+        else:
+            query_result = models.QueryResult(data=data, data_source=data_source, query_hash=query_hash)
```
fabito: @arikfr and @whummer ,

There is still an issue here. When `query_result` is None (max_age <= 0), I'm creating this new QueryResult instance but since there's no `id` the csv/xlsx download won't work.

arikfr: Why do we have different logic here based on the value of `max_age`? What difference does it make?

fabito: Good question. The "if" on line 53 was already there. I added the "else" block just for transforming data into a QueryResult.

@whummer could you help us here ?

redash/handlers/embed.py
```diff
@@ -89,7 +92,7 @@ def embed(query_id, visualization_id, org_slug=None):
             if results is None:
                 abort(400, message="Unable to get results for this query")
             else:
-                qr = {"data": json.loads(results)}
+                qr = results.to_dict()
```
arikfr: I was about to merge this, but then I realized -- we intentionally return only the data field to avoid leaking information via the query.

Can you change this into `project(results.to_dict(), ('data', 'retrieved_at'))` (you will need to import `project` from `funcy`)?

PR #1176, opened 2016-07-11
Add: 'list' command for org and groups CLI

adamlwgriffiths:
Adds a `list` subcommand for both `org` and `group` commands.
`group list` can be filtered based on organziation with the `--org` parameter.
Output is in the format of the existing `users list` command.


arikfr:
Thanks!

PR #1177, opened 2016-07-11
Fix: users list CLI command was broken

adamlwgriffiths:
a66e182f73dc58f203325a3da5b9e71453c5fed1 accidentally broke `users list` command.
This fixes that.


arikfr:
Thanks!

Issue #1178, opened 2016-07-11
Left Pane is not shown when database contains table that name has unicode characters (MSSQL)


kitsuyui:
<img width="703" alt="not shown" src="https://cloud.githubusercontent.com/assets/2596972/16723136/17afa844-4786-11e6-833b-e9bcdfc1b3b0.png">

```pyexception
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python2.7/dist-packages/flask_restful/__init__.py", line 477, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flask_login.py", line 792, in decorated_view
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flask/views.py", line 84, in view
    return self.dispatch_request(*args, **kwargs)
  File "/opt/redash/redash.0.11.0.b1903/redash/handlers/base.py", line 25, in dispatch_request
    return super(BaseResource, self).dispatch_request(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flask_restful/__init__.py", line 587, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/opt/redash/redash.0.11.0.b1903/redash/handlers/data_sources.py", line 105, in get
    schema = data_source.get_schema()
  File "/opt/redash/redash.0.11.0.b1903/redash/models.py", line 412, in get_schema
    schema = sorted(query_runner.get_schema(get_stats=refresh), key=lambda t: t['name'])
  File "/opt/redash/redash.0.11.0.b1903/redash/query_runner/__init__.py", line 117, in get_schema
    self._get_tables(schema_dict)
  File "/opt/redash/redash.0.11.0.b1903/redash/query_runner/mssql.py", line 95, in _get_tables
    table_name = '{}.{}'.format(row['table_schema'], row['table_name'])
```

## Steps to Reproduce

- Using mssql query runner
- Ref: https://github.com/getredash/redash/issues/1101
- Ref: https://github.com/getredash/redash/pull/1104
- The table name is " `テスト`"

## Technical details:

[...]

PR #1179, opened 2016-07-11
Fix: schema not loading when table name has unicode characters

kitsuyui:
Fix: Left Pane is not shown when database contains table that name has unicode characters (MSSQL)

- Update table_name to be unicode (not str).


kitsuyui:
Ref: #1178

arikfr:
Thanks!

PR #1183, opened 2016-07-12
Swapping SystemRandom for a cryptographically strong random number generator in token generation

ChrisKozak:
Hey Arik,
We ran [RATS](https://security.web.cern.ch/security/recommendations/en/codetools/rats.shtml) and this was one of the results:

Problem:  redash//redash/utils/__init__.py:49: Medium: choice
Standard random number generators should not be used to generate randomness used for security reasons.  For security sensitive randomness a crytographic randomness generator that provides sufficient entropy should be used.

Solution: Replaced SystemRandom's 'choice' with pycrypto 'choice'.  [SystemRandom's documentation](https://docs.python.org/3/library/random.html) warn "The pseudo-random generators of this module should not be used for security purposes."

### DIFF
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -4,14 +4,13 @@ import codecs
 import decimal
 import datetime
 import json
-import random
 import re
 import hashlib
 import pytz
 import pystache
 
 from funcy import distinct
-
+from Crypto.Random import random
 from .human_time import parse_human_time
 from redash import settings
 
@@ -48,9 +47,7 @@ def generate_token(length):
     chars = ('abcdefghijklmnopqrstuvwxyz'
              'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
              '0123456789')
-
-    rand = random.SystemRandom()
-    return ''.join(rand.choice(chars) for x in range(length))
+    return ''.join(random.choice(chars) for x in range(length))
 
 
 class JSONEncoder(json.JSONEncoder):


arikfr:
Actually the documentation for Python 2 (what you linked to is Python 3) states:

> Warning The pseudo-random generators of this module should not be used for security purposes. Use os.urandom() or SystemRandom if you require a cryptographically secure pseudo-random number generator.

([source](https://docs.python.org/2/library/random.html#random.SystemRandom))

I'm not sure why they have different wording in the Python 3 docs, but I'm not sure the warning applies to `SystemRandom`. Technically it's not a pseudo-random generator that is implemented in the `random` module but rather uses `os.urandom` that uses a pseudo-random generator implemented by the operating system. The `cryptography` library also [recommends `os.urandom`](https://cryptography.io/en/latest/random-numbers/).

Reading the discussion [here](http://stackoverflow.com/questions/20460061/pythons-pycrypto-library-for-random-number-generation-vs-os-urandom) it looks like PyCrypto uses `os.urandom` too, so again - no clear benefits on `SystemRandom`.

ChrisKozak:
You're right.  I misunderstood the documentation.  They're saying that "random" is pseudo-random.  "random" provides SystemRandom for true random-number generation. 

"The random module also provides the SystemRandom class which uses the system function os.urandom() to generate random numbers from sources provided by the operating system."

Thanks!

arikfr:
:+1: 

Closing this one then.

PR #1192, opened 2016-07-19
Add the REDASH_GOOGLE_APPS_DOMAIN to the document

ken880guchi:
@arikfr 
Hi.

As of the title, and then add the value of the REDASH_GOOGLE_APPS_DOMAIN to document.
More users to be able to know this option.


arikfr:
This setting is no longer in use and kept only for the migration. I don't think we should document it.

ken880guchi:
Thank you for your reply.

PR #1193, opened 2016-07-19
Change: modify the argument order of moment.add function call

ken880guchi:
@arikfr 

Hi.

Argument of moment.add function is using the order in which become Deprecated in 2.8.0.
http://momentjs.com/docs/#/manipulating/add/

fix because uses the version 2.8.4 of moment.js.

pasted image at 2016_07_20 12_15 am

image


arikfr:
Thanks!

PR #1198, opened 2016-07-21
Change: add support for Standard SQL in BigQuery query runner

mystelynx:
add checkbox to specify whether to use legacy or standard sql.

legacy: https://cloud.google.com/bigquery/query-reference
standard: https://cloud.google.com/bigquery/sql-reference/

updated:
image

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -103,6 +103,10 @@ class BigQuery(BaseQueryRunner):
                 'userDefinedFunctionResourceUri': {
                     "type": "string",
                     'title': 'UDF Source URIs (i.e. gs://bucket/date_utils.js, gs://bucket/string_utils.js )'
+                },
+                'useLegacySql': {
+                    "type": "boolean",
+                    'title': "use BigQuery's legacy SQL: https://cloud.google.com/bigquery/query-reference",
                 }
             },
             'required': ['jsonKeyFile', 'projectId'],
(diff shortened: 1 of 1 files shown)


redash/query_runner/big_query.py
```diff
@@ -132,6 +136,7 @@ def _get_total_bytes_processed(self, jobs, query):
         job_data = {
             "query": query,
             "dryRun": True,
+            "useLegacySql": self.configuration.get('useLegacySql', False),
```
arikfr: Unless I misunderstand this, the default should be `True` because otherwise it will break all existing queries...

arikfr:
Thanks! See my comment.

redash/query_runner/big_query.py
arikfr: And because it goes unchecked by default, let's change the configuration name to "Use Standard SQL (Beta)".

mystelynx: Thanks you for your comment!

I'll try to fix that.

redash/query_runner/big_query.py
```diff
@@ -103,6 +103,10 @@ def configuration_schema(cls):
                 'userDefinedFunctionResourceUri': {
                     "type": "string",
                     'title': 'UDF Source URIs (i.e. gs://bucket/date_utils.js, gs://bucket/string_utils.js )'
+                },
+                'useStandardSql(Beta)': {
```
arikfr: Why have the `(Beta)` part in the key name?

redash/query_runner/big_query.py
```diff
@@ -103,6 +103,10 @@ def configuration_schema(cls):
...
+                    "type": "boolean",
+                    'title': "use BigQuery's standard SQL: https://cloud.google.com/bigquery/sql-reference/",
```
arikfr: Please change the title to `Use Standard SQL (Beta)`.

redash/query_runner/big_query.py
mystelynx: Sorry, I misunderstand your comment. I'll remove `(Beta)` from key name.

PR #1203, opened 2016-07-22
Add: map marker cluster visualization

WesleyBatista:
This pull request adds [Leaflet.markercluster](https://github.com/Leaflet/Leaflet.markercluster) visualization, using [ui-leaflet](https://github.com/angular-ui/ui-leaflet) directive from [angular-ui](https://github.com/angular-ui/).

screenshot 2016-07-22 at 14 20 16
------
## Known issues:
- **We have a limit of 10000 rows.** To prevent the web browser to freeze building the points we don't let the script runs if the query results has more than 10k lines. Keep in mind that even with this limit your browser or the final user browser can freeze because it depends on the RAM memory available on the device used. A reasonable number on my tests was 5~6K :/
- **The screen to edit the visualization has UI/UX problems.** Trying to provide a fully configurable visualization I exposed all the settings available. This led the UI to have options that doesn't make sense. Also, changing the values do not make them appear on the pre-visualization aside.

## Screenshots
screenshot 2016-07-22 at 14 14 49
screenshot 2016-07-22 at 14 14 22
screenshot 2016-07-22 at 14 08 14
screenshot 2016-07-22 at 15 32 52

### Mobile
screenshot 22 de julho de 2016 15-08


rd_ui/app/vendor_scripts.html
```diff
@@ -39,6 +39,9 @@
 <script src="/bower_components/canvg/StackBlur.js"></script>
 <script src="/bower_components/canvg/canvg.js"></script>
 <script src="/bower_components/leaflet/dist/leaflet.js"></script>
+<script src="/bower_components/angular-simple-logger/dist/angular-simple-logger.js"></script>
```
arikfr: I think you forgot to remove this.

rd_ui/app/views/visualizations/markercluster.html
```diff
@@ -0,0 +1,12 @@
+<div class="markercluster-viz markercluster-container">
+
+  <leaflet
+    lf-center="center"
+    markers="markers"
+    layers="layers"
+    defaults="visualization.options.leaflet"
+    width={{visualization.options.general.width+'%'}}
```
arikfr: I think it will be better to allow it take 100% of the container element, as usually no one would like to leave empty space and there is no point in making it bigger than the container.

arikfr:
Thanks! Looking good and definitely will be useful.

The points limit is reasonable and it's good you put a "safe guard" in place.

> The screen to edit the visualization has UI/UX problems. Trying to provide a fully configurable visualization I exposed all the settings available. This led the UI to have options that doesn't make sense. 

I suggest to review the options and remove the ones that are less likely to be used. Also, let's split it into several tabs, like the charts visualization has. So the less common options won't confuse the "regular" user and be hidden in their own tab.

> Also, changing the values do not make them appear on the pre-visualization aside.

That should be easy to fix, let's revisit it after implementing the tabs?

Btw, do you think it will be possible to merge this with the other map visualization?

rd_ui/app/vendor_scripts.html
WesleyBatista: This is a dependency from [angular-ui/ui-leaflet](https://github.com/angular-ui/ui-leaflet/blob/master/package.json#L76) :/

rd_ui/app/views/visualizations/markercluster.html
WesleyBatista: 👍

WesleyBatista:
I will try to implement the tabs and remove some options until tomorrow and send the changes.

WesleyBatista:
Regarding merging with the nowadays solution I think that is a good idea and it was a thing that I was intended to do on the conception. 
But I found a little bit tricky since the map visualization is currently using the Leaflet library directly on the visualization code to implement it, and my solution takes the angular-ui's directive.

WesleyBatista:
I think it is possible to merge (and for sure I do like to help on this), but for now maybe it would be better to do this later in another PR.
Do you agree?

WesleyBatista:
@arikfr, can you help me with the options to be updated on the pre-visualization?
It is something with the `$scope.$watch`. I am getting an infinite loop :/

arikfr:
> can you help me with the options to be updated on the pre-visualization? It is something with the $scope.$watch. I am getting an infinite loop :/

Sure I'll give it a look.

WesleyBatista:
@arikfr, I was thinking about do not call the type of this visualization 'MAP_MARKERCLUSTER'.
What do you think about something more general like 'MAP_PLOT', 'MAP_VIZ'... I don't know. But call it something more general and start a smoothly migration from the old to this new one.

zlu:
+1

arikfr:
I finally had the time to look into it:

> Also, changing the values do not make them appear on the pre-visualization aside.

There are two issues here:

1. You are changing the Scope's `visualization` object to be the `visualization.options` object, hence breaking the reference in the `height` setting.
2. It seems that this directive doesn't support changing options after rendering: it applies the `defaults` options at the `link` stage of the controller, [which is being run only once](http://www.codelord.net/2015/09/16/when-does-a-directive-link-function-get-called-again/).

While reviewing their code, it looks like 5K LOC that have built in support for many Leaflet's plugins. Including duplicate plugins: like support for 5 different icons plugins, that do pretty much the same thing.

The lack of support for "live preview", redundant code and other factors lead me to think of dropping `angular-ui-leaflet` and keep using Leaflet directly. If you don't mind, I would like to take a stab at implementing the Marker Cluster functionality directly on Leaflet (along with refactoring the current map visualization). OK?

arikfr:
As discussed with @WesleyBatista, I'm closing this in favor of #1292, where I implement clustering using Leaflet.markersCluster directly instead of this Angular wrapper. The main reason being that the wrapper doesn't support some features we need (like live preview).

PR #1204, opened 2016-07-24
Add: show views in schema browser for Vertica data sources

falling-down:
Currently only metadata from tables show up for Vertica. View metadata has to be queried from another system view.


arikfr:
Merged. Thanks.

PR #1214, opened 2016-07-26
Enable site usage to be tracked by Google Analytics

amasare:
Supply your GA tracking code to the Redash .env file using this env variable: REDASH_GOOGLE_ANALYTICS_TRACKING_CODE

### DIFF
--- a/rd_ui/app/app_layout.html
+++ b/rd_ui/app/app_layout.html
@@ -28,7 +28,18 @@
     <link rel="icon" type="image/png" sizes="32x32" href="/images/favicon-32x32.png">
     <link rel="icon" type="image/png" sizes="96x96" href="/images/favicon-96x96.png">
     <link rel="icon" type="image/png" sizes="16x16" href="/images/favicon-16x16.png">
+{% if ga_tracking_code %}
+  <script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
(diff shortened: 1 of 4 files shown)


rd_ui/app/app_layout.html
```diff
@@ -28,7 +28,18 @@
...
     <link rel="icon" type="image/png" sizes="16x16" href="/images/favicon-16x16.png">
+{% if ga_tracking_code %}
+  <script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
+  ga('create', '{{ga_tracking_code}}', 'auto');
+  ga('send', 'pageview');
```
arikfr: This will *not* be called on every page change, but only on the first load (like tab open or actual page refresh with Cmd+R). This is the problem I mentioned on Slack.

arikfr:
Worth looking at [analytics.js](https://github.com/segmentio/analytics.js) to allow others to use different analytics provider. Although it's not clear to me if it's still possible to use this library without Segment. If not, there might be other libraries.

amasare:
Good point about allowing different analytics providers. I just read up on analytics.js and it seems you have to use Segment with the library. Will look around for others.

rd_ui/app/app_layout.html
amasare: Just found this https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications, which might help with tracking single page apps and dynamic page loading.

arikfr:
Should we keep this open? Do you want to reopen when you resume working on it?

amasare:
Thanks Arik. We have decided to not continue work on this. I will close it.

PR #1216, opened 2016-07-28
Change: bootstrap.sh - use non interactive dist-upgrade

atsaki:
`apt-get -y dist-upgrade` may prompt grub config and stop process.
This patch makes `apt-get -y dist-upgrade` non-interactive.


arikfr:
Thanks!

PR #1217, opened 2016-07-28
Change: install libffi-dev for Cryptography (Ubuntu setup script)

atsaki:
This fixes error `libffi-dev` is missing to install `Cryptography`.


arikfr:
Thanks.

PR #1220, opened 2016-07-29
Add: ability to enable global search in table visualizations

fabito:
Allow users to enable smart-table's global search in table visualizations by appending the suffix `__globalSearch` to any of the columns in the query.

Since global search functionality is very similar to filter's, I thought that using the same mechanics would be a good idea.

Thoughts ?

### DIFF
--- a/rd_ui/app/scripts/services/resources.js
+++ b/rd_ui/app/scripts/services/resources.js
@@ -368,7 +368,15 @@
       });
 
       this.filters = filters;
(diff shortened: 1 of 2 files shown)


arikfr:
While this works it's a terrible workaround that will create even more legacy to support in the future. Sadly I'm closing this, and we will add this feature when we implement #1356 (configuration for table visualization).

Sorry.

PR #1223, opened 2016-08-03
Fix: Alert: when hipchat Alert.name is multibyte character, occur error.

toyama0919:


redash/destinations/hipchat.py
```diff
@@ -42,7 +42,7 @@ def notify(self, alert, query, user, new_state, app, host, options):
                 query_url=query_url)
 
             data = {
-                'message': message,
+                'message': message.encode('utf-8', 'ignore'),
```
arikfr: JSON is unicode, so wouldn't it be enough to construct a unicode string on [line 39](https://github.com/toyama0919/redash/blob/ae3af64c0966f95ef087d0be1741e5d9bb1de32b/redash/destinations/hipchat.py#L39)? I.e. change `message = "<a href..."` to `message = u"<a href..."`?

toyama0919:
@akariv 
Thanks.
applied review.

arikfr:
@toyama0919 I meant that if you create `message` as a unicode string, then you might not need the `encode` call later in the code. Are you sure it's necessary after your last fix? What is the string that makes it fail?

toyama0919:
@arikfr 
>I meant that if you create message as a unicode string, then you might not need the encode call later in the code.

OOps, sorry!!
Fixed again.

arikfr:
Thanks!

PR #1227, opened 2016-08-06
Fix: Bower install fails in vagrant

hokaccha:
When I run the "vagrant provision", "bower install" failed.

```
$ vagrant provision
==> default: Running provisioner: shell...
    default: Running: inline script
==> default: bower                           ENOENT No bower.json present
...
```

Now, the path of `bower.json` is not `/opt/redash/current/rd_ui` but `/opt/redash/current`.


arikfr:
Thanks!

PR #1230, opened 2016-08-09
Change: annotate Presto queries with metadata

kataring:
I want to use the annotate_query in Presto.
There is no reason for the "False".


arikfr:
Thanks!

PR #1232, opened 2016-08-09
Fix: don't show warning when query string (parameters value) changes

### LINKED ISSUE #1174: Changing a parameter value in the UI causes an alert() when you haven't saved the query
Open up a query's source and modify it.  If it doesn't have a parameter already, add one.  Without saving, modify the parameter's value in the input field.  The real-time location updating while typing into the parameter input field triggers the "If you leave this page you'll lose your changes" alert.  Save the query and then try typing in the parameter's input field.  It updates the location without triggering an `alert()`.

hokaccha:
This pull-request fixes #1174

### DIFF
--- a/rd_ui/app/scripts/directives/directives.js
+++ b/rd_ui/app/scripts/directives/directives.js
@@ -58,7 +58,7 @@
         }
 
         $scope.$on('$locationChangeStart', function (event, next, current) {
-          if (next.split("#")[0] == current.split("#")[0]) {
+          if (next.split("?")[0] == current.split("?")[0] || next.split("#")[0] == current.split("#")[0]) {
             return;
           }
 


rd_ui/app/scripts/directives/directives.js
```diff
@@ -58,7 +58,7 @@
         }
 
         $scope.$on('$locationChangeStart', function (event, next, current) {
-          if (next.split("#")[0] == current.split("#")[0]) {
+          if (next.split("?")[0] == current.split("?")[0]) {
```
arikfr: We need the previous check (with the `#`) too, for cases when the user changes the current visualization and there are no parameters present (and hence no query string parameters).

hokaccha: Thank you for your review. fixed https://github.com/getredash/redash/pull/1232/commits/7843d2ee84a91a316a5d048851d9bb75ead216f5

arikfr:
Thanks!

PR #1236, opened 2016-08-16
Add: query runner for Cassandra and ScyllaDB

yershalom:


requirements_all_ds.txt
```diff
@@ -16,3 +16,4 @@ botocore==1.4.4
 sasl>=0.1.3
 thrift>=0.8.0
 thrift_sasl>=0.1.0
+cassandra-driver==3.1.1
```
arikfr: Is it pure Python or has some C/system package dependencies?

yershalom: gcc python-dev

arikfr: 👍

redash/query_runner/cass.py
```diff
@@ -0,0 +1,137 @@
...
+            'type': 'object',
+            'properties': {
+                'host': {
+                    'type': 'string',
+                },
+                'port': {
+                    'type': 'number',
+                    'default': 9042,
+                },
+                'Keyspace': {
```
arikfr: Let's keep all properties snake_cased (i.e. `Keyspace` -> `keyspace`).

redash/query_runner/cass.py
```diff
@@ -0,0 +1,137 @@
...
+                'username': {
+                    'type': 'string',
+                    'title': 'Username'
+                },
+                'password': {
+                    'type': 'string',
+                    'title': 'Password'
+                }
+            },
+            'required': ['Keyspace']
```
arikfr: I would assume host is required too, no?

redash/query_runner/cass.py
```diff
@@ -0,0 +1,137 @@
...
+    @classmethod
+    def type(cls):
+        return "Cassandra"
+
+    @classmethod
+    def enabled(cls):
+        try:
+            from cassandra.cluster import Cluster
+        except ImportError:
+            return False
```
arikfr: You can move this whole block to the top of the file (to avoid duplicate import code), and set a variable (`enabled`) which you use here.  You can see the [BigQuery module](https://github.com/getredash/redash/blob/master/redash/query_runner/big_query.py#L17-L26) as an example.

redash/query_runner/cass.py
```diff
@@ -0,0 +1,137 @@
...
+        try:
+            if self.configuration.get('username', '') and self.configuration.get('password', ''):
+                auth_provider = PlainTextAuthProvider(username='{}'.format(self.configuration.get('username', '')),
+                                                      password='{}'.format(self.configuration.get('password', '')))
+                connection = Cluster([self.configuration.get('host', '')], auth_provider=auth_provider)
+            else:
+                connection = Cluster([self.configuration.get('host', '')])
+
+            session = connection.connect()
+            logger.info("Cassandra running query: %s", query)
```
arikfr: Let's move this to `debug` level.

redash/query_runner/cass.py
```diff
@@ -0,0 +1,137 @@
...
+            for i in result.column_names:
+                column_names.append(i)
+
+            for column_name in column_names:
+
+                columns.append({
+                    'name': column_name,
+                    'friendly_name': column_name,
+                    'type': 'string'
+                })
```
arikfr: You can replace this code with the `fetch_columns` method from the `BaseQueryRunner` class.

Something like `columns = self.fetch_columns(map(lambda c: (c, 'string'), column_names))`.

redash/query_runner/cass.py
```diff
@@ -0,0 +1,137 @@
...
+            json_data = json.dumps(data, cls=JSONEncoder)
+
+            error = None
+
+        except cassandra.cluster.Error, e:
+            error = e.args[1]
+        except KeyboardInterrupt:
+            error = "Query cancelled by user."
+        except Exception as e:
+            raise sys.exc_info()[1], None, sys.exc_info()[2]
```
arikfr: This is not necessary.

redash/query_runner/cass.py
```diff
@@ -0,0 +1,137 @@
...
+                   'type': 'string',
+                   'title': 'Username'
+               },
+               'password': {
+                   'type': 'string',
+                   'title': 'Password'
+               }
+           },
+           'required': ['Keyspace']
+       }
```
arikfr: Is it any different than Cassandra's `configuration_schema`?

arikfr:
Looks great! Only few style things to fix.

I also have some ideas on a few improvements -- like supporting multiple nodes in the host param or correctly setting the type of the column, but this can be left for another iteration.

redash/query_runner/cass.py
yershalom: Which lines?

arikfr: 99-100, you catch an exception and then raise it again. Might as well not catch it and let it bubble up on its own.

arikfr:
@yershalom can I merge?

yershalom:
@arikfr
Yes you may sir

arikfr:
Merged. Thanks!

PR #1240, opened 2016-08-18
Fix: when shared dashboard token not found, return 404

WesleyBatista:
Today, if we click on a link that points to a dashboards shared publicly and the link is not available anymore, the user gets a 500 error page. 
Which actually is not the case, we just do not found the dashboard with the given token.

This PR aims to better report to the user what really happened. 
We respond with the 404 error page.

### DIFF
--- a/redash/handlers/embed.py
+++ b/redash/handlers/embed.py
@@ -125,7 +125,12 @@ def embed(query_id, visualization_id, org_slug=None):
 def public_dashboard(token, org_slug=None):
     # TODO: verify object is a dashboard?
     if not isinstance(current_user, models.ApiUser):
-        api_key = models.ApiKey.get_by_api_key(token)
+        try:
+            api_key = models.ApiKey.get_by_api_key(token)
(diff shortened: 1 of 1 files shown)


redash/handlers/embed.py
```diff
@@ -125,7 +125,12 @@ def embed(query_id, visualization_id, org_slug=None):
 def public_dashboard(token, org_slug=None):
     # TODO: verify object is a dashboard?
     if not isinstance(current_user, models.ApiUser):
-        api_key = models.ApiKey.get_by_api_key(token)
```
arikfr: You can use the [`get_object_or_404`](https://github.com/getredash/redash/blob/master/redash/handlers/base.py#L68-L72) helper method: 

`api_key = get_object_or_404(models.ApiKey.get_by_api_key, token)`.

WesleyBatista:
Did not remembered it. done! :)

arikfr:
Thanks!

PR #1241, opened 2016-08-19
Fix: Notification text on Slack

kataring:
We can be identified by the color.
I think the text is unnecessary.


arikfr:
You don't see the color in notifications... and there is some value to being explicit.

Sorry, but I think I'll pass on this one.

PR #1242, opened 2016-08-19
Docs: add warning re. quotes on column names and BigQuery

ereli:
Bigquery (unlike PostgresQL) doesn't work with "Action__filter" as it doesn't support quoted field names. in order to use the filter function you need to use an unquoted column name.
image


arikfr:
Thanks!

PR #1244, opened 2016-08-19
Verify deparment with bamboo api

d-ray:
We don't want to allow access for all users with an appfolio.com email address.  This allows us to limit access based on what department a user belongs to (determined using the BambooHR API).


d-ray:
Doh!  This was supposed to be a PR to appfolio/redash, not the upstream.  Apologies.

PR #1246, opened 2016-08-19
Fix: support large schemas in schema browser

### LINKED ISSUE #1065: datasource with a lot of schemas cause browser to hang
I've got a PG datasource with ~200 schemas that have ~100 tables each with quite a few columns. When I use the query editor the section that displays the schemas/tables/columns causes my browser to hang. Probably because the ng-repeat loop get out of control. 

I'm wondering about what would be a good way to fix this? Perhaps there's optimizations to the angular repeat that would alleviate loop problems? My first fix was to add a `LIMIT 1000` to the schema query in pg.py  https://github.com/getredash/redash/blob/fbb81b62d6c79f8e9381ac8973ac34739bfa6b41/redash/query_runner/pg.py#L93 

But since all my schemas have the same structure I just changed the WHERE to be like `WHERE table_schema = 'public';`

arikfr:
We can now enable search for schemas of all sizes and the browser doesn't hang when loading large schema (I tested it with 1669 tables and 38871 columns across all tables).

Closes #1065.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -35,7 +35,8 @@
     "angular-resizable": "^1.2.0",
     "material-design-iconic-font": "^2.2.0",
     "plotly.js": "^1.9.0",
-    "angular-ui-ace": "bower"
+    "angular-ui-ace": "bower",
+    "angular-vs-repeat": "^1.1.7"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


benmanns:
Fixes #937 as well. Thanks!

PR #1249, opened 2016-08-22
Add: additional options for Slack webhook

mystelynx:
users can specify custom username, icon_emoji, icon_url or channel in slack notification.

see: https://api.slack.com/incoming-webhooks


arikfr:
Thanks!

PR #1252, opened 2016-08-23
Add: Schema loading support for Presto query runner (using information_schema)

rohanpd:
I have tested the PR that @ninneko submitted and the result was a worker timeout because the query took over 30 seconds to return. This is probably because we have over 1800 tables currently in our Presto instance.

This PR implements the suggestion that @hussainbohra raised in this PR [1233](https://github.com/getredash/redash/pull/1233).


arikfr:
Thanks!

PR #1256, opened 2016-08-25
Change: when forking a query, copy all visualizations

### LINKED ISSUE #439: When forking a query, charts are not copied

ninneko:
Closes #439.

### DIFF
--- a/rd_ui/app/scripts/controllers/query_source.js
+++ b/rd_ui/app/scripts/controllers/query_source.js
@@ -14,7 +14,8 @@
     var isNewQuery = !$scope.query.id,
         queryText = $scope.query.query,
         // ref to QueryViewCtrl.saveQuery
-        saveQuery = $scope.saveQuery;
+        saveQuery = $scope.saveQuery,
+        forkQuery = $scope.forkQuery;
 
     $scope.sourceMode = true;
     $scope.canEdit = currentUser.canEdit($scope.query) || $scope.query.can_edit;// TODO: bring this back? || clientConfig.allowAllToEditQueries;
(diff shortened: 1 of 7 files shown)


redash/models.py
```diff
@@ -715,6 +715,32 @@ def recent(cls, groups, user_id=None, limit=20):
 
         return query
 
+    @classmethod
+    def fork(cls, id, user, org):
```
arikfr: Why class method and not an instance method?

rd_ui/app/scripts/controllers/query_source.js
```diff
@@ -70,12 +71,28 @@
...
+      var savePromise = forkQuery(options, data);
+
+      if (!savePromise) {
+        return;
+      }
+
+      savePromise.then(function(savedQuery) {
+        queryText = savedQuery.query;
+
+        if (isNewQuery) {
```
arikfr: `isNewQuery` will never be true, because you can't fork a new query.

But don't we need to update the `$location.path` in any case, as the query id changes to the forked one?

arikfr:
@ninneko have you seen my comments?

ninneko:
Sorry, I was not able to commit because I have been sick. 
I will commit this PR until tomorrow.

arikfr:
Sorry to hear that. I hope you're feeling better now!

No pressure with the pull request. Just wanted to make sure you noticed the
comments.

Thanks.

redash/models.py
ninneko: I think that fork logic (fetch query and chart data by query id, then change their name and id, then insert them) should be model's method.

arikfr: Anyway the `fork` method shouldn't fetch the query. It will either operate on itself (if it's an instance method) or on a given query (if we keep it as a class method).

What benefit we have from having this as a class method?

ninneko: ok, change instance methdo

arikfr:
@ninneko you need to update the tests, because they still use the class method instead of instance method.

PR #1269, opened 2016-09-01
Add: Google Drive federated tables support in BigQuery query runner

kurtgooden:
Google Big Query supports adding Google Drive files such as Google Spreadsheets as federated tables. To query these from Redash an additional scope is required during the OAuth.

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -119,6 +119,7 @@ class BigQuery(BaseQueryRunner):
     def _get_bigquery_service(self):
         scope = [
             "https://www.googleapis.com/auth/bigquery",
+	    "https://www.googleapis.com/auth/drive"
             ]
 
         key = json.loads(b64decode(self.configuration['jsonKeyFile']))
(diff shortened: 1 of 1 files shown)


redash/query_runner/big_query.py
```diff
@@ -230,7 +231,7 @@ def _get_project_id(self):
         return requests.get('http://metadata/computeMetadata/v1/project/project-id', headers={'Metadata-Flavor': 'Google'}).content
 
     def _get_bigquery_service(self):
-        credentials = gce.AppAssertionCredentials(scope='https://www.googleapis.com/auth/bigquery')
+        credentials = gce.AppAssertionCredentials(scope='https://www.googleapis.com/auth/bigquery,https://www.googleapis.com/auth/drive')
```
arikfr: I'm not sure this will work; I think the user needs to boot the instance with this API enabled.

If you didn't have the chance to test this with existing instance, let's just keep this fix limited to the "regular" BigQuery query runner.

arikfr:
Thanks!

Aside from updating the query runner, did you have to do anything else?

kurtgooden:
@arikfr I also had to specifically give view access on the file to the service account. (Clicked the share button and shared with the email address of the service account.)

redash/query_runner/big_query.py
kurtgooden: I did not get a chance to fully test the GCE version.

arikfr: Let's keep it as it was then , to avoid breaking existing installations and we will revisit this when we get the chance.

arikfr: Ping? I would to merge this, once we revert this change. Thanks :)

arikfr:
Brilliant, thanks!

PR #1270, opened 2016-09-02
Add: CTRL+K hotkey for dashboards search

WesleyBatista:
Nowadays, if we have a lot of dashboards, the menu list grows and at some time you cannot see all dashboards.

This PR aims to address this issue and add the hotkey `CMD + K`/`CTRL + K`, used for search on Slack/Mattermost

captura de tela 2016-09-02 as 17 58 29

### DIFF
--- a/rd_ui/app/scripts/controllers/controllers.js
+++ b/rd_ui/app/scripts/controllers/controllers.js
@@ -150,7 +150,37 @@
     });
   }
 
-  var MainCtrl = function ($scope, $location, Dashboard) {
+  var DashboardSearchCtrl = function($scope, $modalInstance, $http, $location, Dashboard) {
+    $scope.dashboardSelected = {};
+
+    Dashboard.query(function (dashboards) {
+        $scope.dashboards = _.sortBy(dashboards, "name");
+    });
+
+    $scope.close = function() {
+      $modalInstance.close();
+    };
+
+    $scope.openSelectedDashboard = function(dashboard){
+      $scope.close();
+      $location.path('/dashboard/' + dashboard.slug).replace();
+    }
+
+    $scope.findDashboard = function(search){
+      if (search == "") {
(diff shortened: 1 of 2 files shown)


ThamaraHessel:
👍

rohanpd:
+1!

Is this help context available all over the app? Would be great to add some more stuff to it for the query editor too.

WesleyBatista:
@rohanpd yes, it is! :)

The (awesome) [hotkey module](https://github.com/chieffancypants/angular-hotkeys) was added as dependency on the main [redash module](https://github.com/getredash/redash/pull/1270/files#diff-08de1c90bcd3796927dfe5f0cc799843R23) and the search hotkey was added in the `MainCtrl`, which holds all the application.

WesleyBatista:
I just found out that we already have a service to bind hotkeys, `KeyboardShortcuts`.
I removed the module from the stack :)

breath103:
:+1:

arikfr:
Sorry for letting it dry here... 

I think it's a good feature, but it has a big UX problem: it's very confusing that you search for queries from the nav bar but you search for dashboards with this keyboard shortcut.

For this to go in we need to have a unified search for both object types. We should probably use the opportunity to switch to using PostgresSQL's full text search capabilities instead of abusing `ILIKE`.

This will require much more effort so I can understand if you're not into it right now.

If you prefer not to handle the work on unified search but still push this feature we can do:
1. Add a feature flag to enable it.
2. Add a toggle to select between searching for queries or dashboards.

What do you think?

arikfr:
@WesleyBatista should I just close this one?

WesleyBatista:
The project evolved a lot since the PR (and that is great :D ) and I do not have the time to take on this now. You can close it :)

Maybe in the future we can improve on this.

PR #1272, opened 2016-09-03
Fix: TreasureData get_schema method was returning array instead of string as column name

ariarijp:
Fix schema information widget on Query Source View.

For details, Please see below screenshots.

---

Before

before

After

after

### DIFF
--- a/redash/query_runner/treasuredata.py
+++ b/redash/query_runner/treasuredata.py
@@ -84,7 +84,10 @@ class TreasureData(BaseQueryRunner):
                     for table in client.tables(self.configuration.get('db')):
                         table_name = '{}.{}'.format(self.configuration.get('db'), table.name)
                         for table_schema in table.schema:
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks! Is it TD version dependent or a bug we had since forever?

cc: @toru-takahashi

toru-takahashi:
Thank you for the implement.

The reason that column name show Array type with column type is my expected behavior.
(And, it's not TD version's bug.)
Because, I though that showing column type with column name is useful for users to write a query at that time. (And, I think re:dash doesn't have a feature to show column type. Is it right?)

Anyway, It's ok to apply this fix to become same behavior as other plugins.

arikfr:
@toru-takahashi got it. It's better to pass only the column name as it will be more readable and useful for auto complete.

I agree that adding support for types has its benefits, but we need to do it properly.

toru-takahashi:
I see. Thank you!

Issue #578, opened 2015-09-29
Dashboard list not scrollable


jcox92:
My team has quite a few dashboards (~40-50) and the ones later in the alphabet aren't visible in the dashboard list because the list can't be scrolled.

This is occurring on OS X in both Chrome and Firefox.

![scrollable](https://cloud.githubusercontent.com/assets/844493/10164045/1381735c-6685-11e5-8770-6d2c90891130.gif)

arikfr:
There is an undocumented feature of dashboards grouping. For example, if you rename all your "Feed Engagement - *" dashboards into "Feed Enagement: *" (i.e. "Feed Engagement - Top 5" -> "Feed Engagement: Top 5"), you will notice that they get grouped in a "folder" (named "Feed Engagement") in the dashboards menu.

Of course we still need to solve the long list issue, but in your case, I guess it will solve the problem.

jcox92:
Thanks @arikfr! That's helpful!

alonho:
There's a conflict between submenu and scroll.. they just don't go together.

arikfr:
@alonho ack. Let's keep this open and just revisit when we do the top menu.

zlu:
We have several hundred dashboards that are grouped already.  Unable to scroll them is causing an operational headache.  Any update on why this is happening?

rohanpd:
+1
This is starting to become a problem for us as well. Our dashboards are already grouped.

I don't think that the dropdown needs to display all dashboards but being able to click through and see the full list (like you can for queries) would be useful.

PGuiv:
I think I found a fix for that issue, using bootstrap3:

Adding "pre-scrollable" to the ul element class fixed it for me.
```
<ul class="dropdown-menu pre-scrollable" dropdown-menu="">
...
```

I only tested it on Chrome, though, since this is what my whole team uses. So, not sure how it will react on other browsers.

WesleyBatista:
@PGuiv, If you have chance to do it, would be good to test with submenus. 
I tried [forking](http://www.bootply.com/mzZvWwPoKG#) a demo of a html code using the **dropdown-submenu pre-scrollable** classes and putting a submenu inside a `li`, and it did not worked :disappointed:

PGuiv:
@WesleyBatista That worked when I tried on http://demo.redash.io/.

I edited the HTML to add more subsections under the Examples link in the dashboard menu, to have enough links to scroll.

Then, I added pre-scrollable to the Examples ul HTML tag:
[...]

PR #1273, opened 2016-09-04
Enable scrolling in nested dashboards

bakunyo:
Enabled scrolling in nested dashboards.
This PR relates to issue #578 .

redash_scroll

This isn't solve all problems of long list, but it's better than now, I think.

### DIFF
--- a/rd_ui/app/styles/redash.css
+++ b/rd_ui/app/styles/redash.css
@@ -232,6 +232,8 @@ to add those CSS styles here. */
 .dropdown-submenu > .dropdown-menu {
   top: 0;
   left: 100%;
+  overflow: auto;
+  max-height: 300px;
   margin-top: -6px;
   margin-left: -1px;
   -webkit-border-radius: 0 6px 6px 6px;


arikfr:
As you mentioned, the real issue is the height of the "main list". I'm not sure if people have problems with having sub sections too long to scroll -- did this happen to you?

While I don't mind adding support for inner scrolling of the sub section, I prefer to keep it as it is today, where it will use the full height of the screen (starting from its position) rather than setting a "static" max-height. Is it a simple change?

As for a fix for the main issue, I was thinking on replacing the current menu with the recent dashboards list and having a dedicated page for showing all dashboards, like we have for queries.

bakunyo:
Thank you for comment.
Yes, I actually have too long sub sections.

As you say, Setting to the full height of the screen is better than static height. It's simple, but its difficult to me..(Could you please any advise about my css code?)

I'm looking forward to see the page for showing all dashboards. It's very nice!

arikfr:
Closing in favor of #1399.

PR #1275, opened 2016-09-06
Fixed graphite bug, no target was setup

yershalom:
Tested on Our redash

### DIFF
--- a/redash/query_runner/graphite.py
+++ b/redash/query_runner/graphite.py
@@ -61,7 +61,7 @@ class Graphite(BaseQueryRunner):
             self.auth = None
 
         self.verify = self.configuration.get("verify", True)
-        self.base_url = "%s/render?format=json&" % self.configuration['url']
+        self.base_url = "%s/render?format=json&target=" % self.configuration['url']
 
     def run_query(self, query):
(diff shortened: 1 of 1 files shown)


arikfr:
I think we were expecting you to pass the `target=` part in your query. I don't mind changing this although if someone already using this data source it will break his queries :-/

arikfr:
After some more thought -- it's better to maintain backward compatibility here. I guess you can check if the query starts with `target=` and then do nothing and if it isn't, add it. Then we can safely merge this one.

redash/query_runner/graphite.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, configuration):
             self.auth = None
 
         self.verify = self.configuration.get("verify", True)
-        self.base_url = "%s/render?format=json&" % self.configuration['url']
+        self.base_url = "%s/render?format=json&target=" % self.configuration['url']
```
arikfr: After some more thought -- it's better to maintain backward compatibility here. I guess you can check if the query starts with `target=` and then do nothing and if it isn't, add it. Then we can safely merge this one.

arikfr:
Closing for now, please re open when relevant.

PR #1280, opened 2016-09-09
Add: configuration flag to disable scheduled queries

hirotaka-s:
Add a setting to disable refresh of queries during development. While doing development using re:dash, I want to be able to avoid unwanted processing cost in the query runner(e.g. B
igQuery) by preventing queries from being executed automatically.


arikfr:
Are you sure the UI changes are necessary if it's just for development? If you implemented them just for "feature completeness", but don't need them yourself, I prefer we skip them.

hirotaka-s:
Thank you for comment. 
Yes, that makes sense. I remove this pull request's UI changes.

arikfr:
Thanks. I appreciate the contribution and your cooperation!

PR #1285, opened 2016-09-12
Change: do not display Oracle tablespace name in schema browser

falling-down:
Does not make sense to list tablespace name in schema browser as this is not used when writing a SQL query. Should use the owner instead.

Should list all tables visible to the user for querying. Joining to USER_TABLES limits schema browser to only tables owned by user. Not very useful if you create a read-only redash user to connect with.

But still should remove known system tables from schema browser. Oracle 12c introduced a ORACLE_MAINTAINED field on ALL_USERS table that would have been a more elegant solution, but would not be compatible with older versions of Oracle obviously. Best option I could find was to simply exclude the known system users explicitly.


arikfr:
Thanks!

PR #1292, opened 2016-09-22
Change: improvements to map visualization

### LINKED ISSUE #557: Cannot set colors in "classify by" in "map visualization"
1) Cannot set color values that I want, for "Classify by column". I set the html color value, the background of that form field turns to appropriate color, but when I hit "Save" (Save visualization) button, all the form fields of "Classify by column" values, turn back to default colors and map shows original default colors.
2) When I login as "admin" and query the "visualization" table of reddash::metadata, the color of appropriate "Classify by column" seems to have got the correct html color code set by me, but this is not reflected in actual map visualization.

arikfr:
* Fix #557: user can set his own colors for the different groups on the map.
* Add layer selection control.
* Inspired by #1203, added support for markers clustering.
* Changed the map setting screen (to hopefully) be simpler.
* Disabled automatic zoom when scrolling past the window.

image

image

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -36,7 +36,8 @@
     "material-design-iconic-font": "^2.2.0",
     "plotly.js": "^1.9.0",
     "angular-ui-ace": "bower",
-    "angular-vs-repeat": "^1.1.7"
+    "angular-vs-repeat": "^1.1.7",
+    "leaflet.markercluster": "^0.5.0"
   },
   "devDependencies": {
     "angular-mocks": "1.2.18",
--- a/rd_ui/app/app_layout.html
+++ b/rd_ui/app/app_layout.html
@@ -20,6 +20,8 @@
     <link rel="stylesheet" href="/bower_components/font-awesome/css/font-awesome.css">
     <link rel="stylesheet" href="/bower_components/leaflet/dist/leaflet.css">
     <link rel="stylesheet" href="/bower_components/angular-resizable/src/angular-resizable.css">
+    <link rel="stylesheet" href="/bower_components/leaflet.markercluster/dist/MarkerCluster.css">
+    <link rel="stylesheet" href="/bower_components/leaflet.markercluster/dist/MarkerCluster.Default.css">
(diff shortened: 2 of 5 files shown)


arikfr:
@WesleyBatista I didn't implement any special settings. If there are any settings you want me to port from your code, let me know and I will add it. Thanks!

WesleyBatista:
Really thanks for implementing it! 

I just had time to test it today. Here are my thoughts:

**1**. I really like that now we can use `group by` in the visualization settings, and now we are able to select the groups that we want to see in the visualization. 👏 👏 👏 
captura de tela 2016-09-26 as 17 58 20

**2**. The colors used by the marker cluster library and ours is not managed very well. for example, take a look at this 5 blue points: 
captura de tela 2016-09-26 as 17 46 59
captura de tela 2016-09-26 as 17 46 39
When zooming out, the 4 closer to each other are clustered and the color is set to green.

The same issue within another use case: looking at the yellow clusters in the next image, I find clusters with the same colors closer to each other, and start to wondering "why they were not grouped?", but I know that we have 5 different groups because I used the `group by`. 
captura de tela 2016-09-26 as 18 02 06
It would be more meaningful if the colors used was the group's color, so I would also know which groups they belongs.


**3**. the custom layers is something that I miss. The OpenStreet Map default tiles is messy by showing a lot of information in the map, and the colors it uses mixes with the ones used by the clusters/groups. For example, look at this two images: 
captura de tela 2016-09-26 as 18 29 29

captura de tela 2016-09-26 as 17 55 40
The **Stamen Toner** layers are more clear by showing only the necessary. That way the user which is taking care of a dashboard, for example, can focus on the main information which is the clusters.

For now that is what I found 😄

arikfr:
@WesleyBatista thanks! I'll implement both.

arikfr:
@WesleyBatista added both things.

shyamgopal:
Thank you so much @arikfr and @WesleyBatista !!

PR #1295, opened 2016-09-23
add a few sudos

marc151:
I hit wall on my VM untill I added the 3 sudos, I don't know if this will help others but i'm proposing it.


marc151:
I hope this is the correct procedure, this is my first git commit. let me know if i did something wrong

arikfr:
This is indeed the correct procedure, well done 👍 

But something doesn't make sense -- the whole script runs as root (see [L16-L26](https://github.com/marc151/redash/blob/9924973f7866b287054b01439b746a36d914fc38/setup/ubuntu/bootstrap.sh#L18-L26)), so the `sudo` calls you added are a no-op.

How did you run the script when you encountered the issue?

marc151:
Oh never mind i realise now it's cause my firewall blocked the download 
from github so i had to run that part of the script manually. I'll undo my 
pull request.

Sent with AquaMail for Android
http://www.aqua-mail.com

arikfr:
I thought it might be the case. Closing it now 👍

PR #1298, opened 2016-09-26
Add: JIRA (JQL) query runner

arikfr:
Add ability to query a JIRA instance. Currently it takes JSON as the query. Some example queries:

*Simple query, just return issues with no filtering:*
```json
{
}
```

*Return only specific fields:*
```json
{
    "fields": "summary,priority"
}
```

*Return only specific fields and filter by priority:*
```json
{
    "fields": "summary,priority",
    "jql": "priority=medium"
}
```

*Count number of issues with `priority=medium`:*
```json
{
    "queryType": "count",
    "jql": "priority=medium"
}
```

### DIFF
--- a/docs/datasources.rst
+++ b/docs/datasources.rst
@@ -63,7 +63,7 @@ Graphite
 
 -  **Options**:
 
-   -  Url (mandatory)
+   -  URL (mandatory)
    -  User
    -  Password
    -  Verify SSL certificate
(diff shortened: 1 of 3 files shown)


akakimeladze:
Hi guys. There's an issue in retrieving the "component" field in the jql.
For example when I write:

`{
    "fields": "key,issuetype,component",
    "jql": "project=SSD"
}`

it doesn't retrieve "component" field, which is a default JIRA field (not a custom one) and is directly queryable and returned in issue details without any additional things.

PR #1299, opened 2016-09-27
Change: send Content-Type header (application/json) in query results responses

tatsukawa:
I found lack of Content-Type header in the response of QueryResultResource.


arikfr:
Thank you!

PR #1301, opened 2016-09-29
Add: "test connection" button for data sources

### LINKED ISSUE #942: Have a test button when adding databases
Have a button when adding databases to test the connection.

washort:
This adds a 'test' command to `manage.py ds` and a 'test' button to the `/data_sources` forms.

I'm not very familiar with Angular so the JS code probably needs a good bit of help. 

Closes #942 .

### DIFF
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -141,3 +141,15 @@ class DataSourcePauseResource(BaseResource):
         })
 
         return data_source.to_dict()
+
+
+class DataSourceTestResource(BaseResource):
+    @require_admin
+    def post(self, data_source_id):
+        data_source = get_object_or_404(models.DataSource.get_by_id_and_org, data_source_id, self.current_org)
+        try:
+            data_source.query_runner.test_connection()
+        except Exception, e:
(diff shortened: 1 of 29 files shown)


arikfr:
Amazing :-)

redash/handlers/data_sources.py
```diff
@@ -141,3 +141,15 @@ def delete(self, data_source_id):
...
+
+
+class DataSourceTestResource(BaseResource):
+    @require_admin
+    def post(self, data_source_id):
+        data_source = get_object_or_404(models.DataSource.get_by_id_and_org, data_source_id, self.current_org)
+        try:
+            data_source.query_runner.test_connection()
+        except Exception, e:
+            return {"msg": "failure: {}".format(e), "ok": 0}
```
arikfr: Why not use `True`/`False` instead of 0/1?

And let's use `message` instead of `msg` to be consistent with other responses.

arikfr:
I've fixed the UI issue & updated it to show a spinner and disable the button while in progress. Also allowing to do a test connection only after saving (otherwise it tests with the old configuration).

👍

PR #1303, opened 2016-09-30
Set SESSION_COOKIE_HTTPONLY

yuya-takeyama:
Related to https://github.com/getredash/redash/pull/980

## Problem

I deployed Re:dash to HTTPS environment and set `REDASH_ENFORCE_COOKIE=true`.
Then I noticed that it returns weird session cookie.

```
set-cookie:session=***; Secure; HttpOnly; Path=/
```

It contains both of `Secure` and `HttpOnly`.

## Solution

Flask has a configuration named `SESSION_COOKIE_HTTPONLY`.
http://flask.pocoo.org/docs/0.10/config/

I think it should always be inverse of `SESSION_COOKIE_SECURE`.

With this change, now Re:dash returns session cookie correctly.

```
set-cookie:session=***; Secure; Path=/
```

---

Please review, thanks!


arikfr:
I don't understand why one should be the inverse of the other. `Secure` defines how the cookie can be transmitted while `HttpOnly` defines how it can be accessed, there is nothing that prevents using them both. Even one of the examples for [`Set-Cookie` header on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) includes them both.

If you think you need this, you can add this setting with default value of True, but it shouldn't be related to the value of `SESSION_COOKIE_SECURE`.

yuya-takeyama:
@arikfr 
Ah sorry, I misunderstood the meaning of `HttpOnly`!
Then current implementation has no problem.

Let me close now.

arikfr:
:+1:

PR #1304, opened 2016-10-03
Only execute the query that is highlighted

smeriwether:
Hello! I've just started using `redash` and this is my first PR. 

Basically, while using the query-editor I frequently type out & save many queries but I can only "execute" the query at the very bottom of the editor (at least those are the only results I get back). 

This change will, if present, run the selected query, else keep the existing behavior. 
- If multiple queries are selected the server will be sent all queries (as it currently does) and only the last query will be executed
- If a partial query is selected the server will return the database error and display it to the user (as it currently does)

Please let me know if there is any other information you need from me, thanks!

### DIFF
--- a/rd_ui/app/scripts/directives/query_directives.js
+++ b/rd_ui/app/scripts/directives/query_directives.js
@@ -138,6 +138,7 @@
               $scope.$parent.$on("angular-resizable.resizing", function (event, args) {
                 editor.resize();
               });
+              $scope.query.editor = editor;
 
               editor.focus();
             }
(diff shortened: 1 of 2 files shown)


smeriwether:
Screenshot of behavior:

screen shot 2016-10-03 at 11 08 52 am

smeriwether:
I have not (and do not expect) time to open up to fix this so I am just going to close this PR.

sbecainfo:
I would love this feature to be implemented. Is it included in any product roadmap?

Issue #530, opened 2015-08-13
When a TextBox is added on dashboard it cannot be updated


matiascarazzo:
When a TextBox is added on dashboard it cannot be updated

satyadeepk:
Yeah, I feel like having an edit button too on the Text Box.
Looks like this is the PR which added the Text Box. We will probably need to add an edit button with an edit dialog.
https://github.com/EverythingMe/redash/pull/192/files

jeffwidman:
+1 for an edit button on the text box.

mathemagica:
+1 for an edit button

JonahTurnquist:
+1

gcasey-atvi:
+1

RichardLitt:
@gcasey-atvi This should have been fixed; are you still finding this to be an issue?

Lpember:
+1

PR #1305, opened 2016-10-04
Add: UI to edit dashboard text box widget

hokaccha:
This PR wil fix https://github.com/getredash/redash/issues/530

Demo:

redash2

### DIFF
--- a/redash/handlers/widgets.py
+++ b/redash/handlers/widgets.py
@@ -57,7 +57,6 @@ class WidgetResource(BaseResource):
         require_admin_or_owner(widget.dashboard.user_id)
         widget_properties = request.get_json(force=True)
         widget.text = widget_properties['text']
-        widget.width = widget_properties['width']
         widget.save()
 
         return widget.to_dict()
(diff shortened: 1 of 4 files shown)


redash/handlers/widgets.py
```diff
@@ -52,10 +52,20 @@ def post(self):
 
 class WidgetResource(BaseResource):
     @require_permission('edit_dashboard')
+    def post(self, widget_id):
+        widget = models.Widget.get_by_id_and_org(widget_id, self.current_org)
+        require_admin_or_owner(widget.dashboard.user_id)
+        widget_properties = request.get_json(force=True)
+        widget.text = widget_properties['text']
+        widget.width = widget_properties['width']
```
arikfr: Allowing to change widget size, means that we need to handle the case of the layout needs to change (in case we change it from "Regular" to "Double").

I'm totally fine with limiting the editing functionality to only the text itself (without size).

arikfr: Also I would add a comment here mentioning that it currently handles Text Box widgets only.

arikfr:
This was quick! :-)

hokaccha:
Thanks!

PR #1317, opened 2016-10-06
Cassandra query_runner

scotth1:
A working Cassandra query runner.
The PR provides the new query runner, and updates the settings.py to add it into the list of datasources.
The DataStax Cassandra python driver appears to already be in requirements_ds.txt

I've only learnt python this week, so any style or optimisation comments might need to be explicitly spelt out.


arikfr:
This is very appreciated but we already have Cassandra query runner..

scotth1:
ah! I couldn't see it, what is it called.

arikfr:
https://github.com/getredash/redash/blob/master/redash/query_runner/cass.py

PR #1336, opened 2016-10-10
Add fixed width for each column

smeriwether:


smeriwether:
I did not mean to create a pull request just yet, my apologies.

PR #1339, opened 2016-10-11
add http://

guaguaxiche:
base_url add http://


guaguaxiche:
the old code is 
<pre>
def base_url(org):
    if settings.MULTI_ORG:
        return "https://{}/{}".format(settings.HOST, org.slug)
    return settings.HOST
</pre>
how to set the settings.HOST?  "www.google.com" or "http://www.google.com" or "https://www.google.com"?

PR #1342, opened 2016-10-13
Add: support for passing entityid in SAML requests

zoetrope:
OpenAM requires EntityId to send to IdP.
So I add new configuration 'REDASH_SAML_ENTITY_ID'


arikfr:

PR #1343, opened 2016-10-13
Add: support for local SAML metadata file

zoetrope:
In current implementation, re:dash requests to IdP to fetch metadata.
I think re:dash should not communicate directly with IdP, because IdP may be on intranet.
So I add new configuration 'REDASH_SAML_LOCAL_METADATA_PATH', and use local metadata instead of communication with IdP.


arikfr:

PR #1344, opened 2016-10-13
SAML: use NameID value, if SAML Assertion is missing

zoetrope:
I want to use NameID value instead of Assertion Attribute.


jeffwidman:
Any reason this was never merged? It seems straightforward

arikfr:
@jeffwidman mainly because I never had the time to verify that it's "safe" or correct :-( For example this line: `email = authn_response.ava['EmailAddress'][0]` -- I don't know if it's guaranteed to be a list every time :\

arikfr:
Hi,

Thanks.

jeffwidman:
@arikfr I'm surprised you haven't had customers ask for this?

arikfr:

PR #1345, opened 2016-10-13
changes to pass metadata through query runner

J314Y:


J314Y:
@arikfr I opted to pass through metadata to the query runner because I was thinking that maybe down the future more than just the name may want to be passed down to query execution level. Lmk your thoughts

PR #1359, opened 2016-10-22
Fix: schema loading query for Hive was wrong for non default schema

laughingman7743:
Table not found errors occurred in other than the default schema.
Add schema name to columns query.


arikfr:
Thanks!

PR #1360, opened 2016-10-24
Fix: connection test was broken for MySQL

shinji19:
Hello, I use Re:dash. Re:dash is very cool !

I have a problem...
Mysql test connection is always failed of data source page.
Maybe "noop_method" is invalid, because query_runner/\__init__.py defined "noop_query" for test_connection.
Please check my pull request. Thank you.


arikfr:
You're right. Thanks!

PR #1369, opened 2016-10-28
Fix: missing format call in Elasticsearch test method

adamlwgriffiths:
Calling 'test' on an elasticsearch datasource displays an error along the lines of "could not connect to {0}/_cluster/health, did you mean http://{0}/_cluster/health".
This string is obviously intended to be a format string, but format is not called on it before it is used.
This commit adds the missing 'format' call with the 0'th parameter being `self.server_url` as per all the other urls.


arikfr:
Thanks!

PR #1371, opened 2016-11-01
Change: display user's password reset link to the admin when mail server disabled

vitorbaptista:
This is work in progress, I'm sending just to ask for some help. Adding the reset link is quite easy, I'm just wondering what would be the best way to display it to the user. We have a similar functionality when creating a new user, where we display the invite URL to the admin inside the template. Would that be the best approach for this?

### DIFF
--- a/rd_ui/app/views/users/show.html
+++ b/rd_ui/app/views/users/show.html
@@ -34,9 +34,10 @@
         <p>
           <strong>The user should receive a link to reset his password by email soon.</strong>
         </p>
-        <p ng-if="clientConfig.mailSettingsMissing">
-          You can use the following link to reset the password yourself:<br/>
-          {{passwordResetLink}}
+        <p ng-if="!clientConfig.mailSettingsMissing">
+          You don't have mail server configured, please send the following link
+          to {{user.name}} to reset their password:<br/>
+          <a ng-href="passwordResetLink">{{passwordResetLink}}</a>
         </p>
       </div>
     </div>
--- a/rd_ui/app/views/users/show.html
+++ b/rd_ui/app/views/users/show.html
@@ -30,11 +30,11 @@
         </button>
       </div>
 
-      <div ng-if="passwordResetLink && !disablePasswordResetButton" class="alert alert-success">
-        <p>
+      <div ng-if="passwordResetLink" class="alert alert-success">
+        <p ng-if="!clientConfig.mailSettingMissing">
           <strong>The user should receive a link to reset his password by email soon.</strong>
         </p>
-        <p ng-if="!clientConfig.mailSettingsMissing">
+        <p ng-if="clientConfig.mailSettingsMissing">
           You don't have mail server configured, please send the following link
           to {{user.name}} to reset their password:<br/>
           <a ng-href="passwordResetLink">{{passwordResetLink}}</a>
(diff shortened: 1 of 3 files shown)


arikfr:
Yes, I think that doing it in a similar manner to how we did with the invite link is preferable than showing it briefly in the growl message.

Also, when implementing this please show the link only when `clientConfig.mailSettingsMissing` is true.

vitorbaptista:
@arikfr Please take a look. I wanted to make the link an actual `<a>` element, but didn't know how to generate a URL using the base path, and ended up following the pattern of the invite link. I also removed the growl notification, as the user now is notified via the message in the body.

This is in a couple commits just to make it easier to understand the PR, but would expect them to be squashed when merged.

rd_ui/app/views/users/show.html
```diff
@@ -24,9 +24,21 @@ <h3 class="p-l-5">{{user.name}}</h3>
...
+      </div>
+
+      <div ng-if="passwordResetLink && !disablePasswordResetButton" class="alert alert-success">
+        <p>
+          <strong>The user should receive a link to reset his password by email soon.</strong>
+        </p>
+        <p ng-if="clientConfig.mailSettingsMissing">
+          You can use the following link to reset the password yourself:<br/>
+          {{passwordResetLink}}
+        </p>
```
arikfr: 1. Let's show the original message ("should receive by email...") only if email is enabled (`ng-if="!clientConfig.mailSettingsMissing"`).

2. Change the text here to: `You don't have mail server configured, please send the following link to {{user.name}} to reset their password:`.

3. To have the reset link as an `<a>` tag: `<a ng-href="passwordResetLink">{{paswordResetLink}}</a>`. IIRC, it's a full link so the base href doesn't matter.

vitorbaptista:
@arikfr Thanks for the review! Just added your comments in and this is good to go AFAIK.

rd_ui/app/views/users/show.html
```diff
@@ -24,9 +24,22 @@ <h3 class="p-l-5">{{user.name}}</h3>
...
-        Password Reset Email
-      </button>
+      <div class="form-group">
+        <button class="btn btn-default" ng-click="sendPasswordReset()" ng-disabled="disablePasswordResetButton">Send
+          Password Reset Email
+        </button>
+      </div>
+
+      <div ng-if="passwordResetLink && !disablePasswordResetButton" class="alert alert-success">
+        <p>
```
arikfr: Let's add here `ng-if="clientConfig.mailSettingsMissing"`.

arikfr: No. We show one message if there is no mail server configured and the other otherwise. But I just noticed that you have the ng-if backwards: it should be `clientConfig.mailSettingsMissing` for your message and `! clientConfig.mailSettingsMissing` for the original one.

vitorbaptista: Oh, I see. Just changed it. To be honest, as a user, I'd prefer to have the link shown regardless if I have an e-mail server configured or not, similar to the user invites. I find them useful even when the user will receive an e-mail, and it also gives admins the ability to change a user's password herself. Happy to go with whatever you think is best, though.

Just to let you know, I'm adding new commits and not changing the original ones just because I think it's easier to review this way, but I expect that you (or me) will squash them before merging.

arikfr:
Thanks!

PR #1374, opened 2016-11-02
Add: allow '*' in REDASH_CORS_ACCESS_CONTROL_ALLOW_ORIGIN

washort:
Mozilla currently uses this in its redash deployment: https://github.com/robotblake/redash-tweaks/blob/master/cors-fix.diff


AntoineAugusti:
Cool that Mozilla uses Redash 👍

PR #1375, opened 2016-11-04
Fix: Download Dataset does not work when not logged in

### LINKED ISSUE #1247: Download Dataset does not work when not logged in
### Issue Summary

Using a query public URL, trying to download the dataset as a CSV when not logged in fails
### Steps to Reproduce
1. Create a query that returns a table
2. Create a public link using "embed" (use the src argument)
3. Open the link from an **anonymous** browser window
4. Click Download Dataset CSV

Getting 
image
### Technical details:
- Redash Version:  0.11.1+b2095
- Browser/OS: Chrome / Mac

phpcodecrafting:
This fixes #1247 where the download links on a visualization's public embed URL do not work if logged out of redash.

### DIFF
--- a/rd_ui/app/scripts/embed.js
+++ b/rd_ui/app/scripts/embed.js
@@ -49,6 +49,7 @@ angular.module('redash', [
        $scope.embed = true;
        $scope.visualization = visualization;
        $scope.query = visualization.query;
+       $scope.api_key = $location.search()['api_key'];
        query = new Query(visualization.query);
        $scope.queryResult = new QueryResult({query_result: query_result});
      }])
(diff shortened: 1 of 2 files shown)


rd_ui/app/scripts/embed.js
```diff
@@ -49,6 +49,7 @@ angular.module('redash', [
        $scope.embed = true;
        $scope.visualization = visualization;
        $scope.query = visualization.query;
+       $scope.api_key = $location.search()['api_key'];
```
arikfr: Should be `apiKey`. Other than - perfect. 👍 Thanks!

phpcodecrafting: Good point. Corrected and pushed.

arikfr:
Thanks!

PR #1380, opened 2016-11-09
Doc improvement for docker-compose

antgel:


docs/setup.rst
```diff
@@ -69,13 +69,17 @@ run the :doc:`upgrade process <upgrade>` and then proceed to `"Setup" <#setup>`_
...
-2. Make sure your current working directory is the root of this GitHub repository.
-3. Run ``docker-compose up postgres``.
-4. Run ``./setup/docker/create_database.sh``. This will access the postgres container and set up the database.
-5. Run ``docker-compose up``
-6. Run ``docker-machine ls``, take note of the ip for the Docker machine you are using, and open the web browser.
-7. Visit that Docker machine IP at port 80, and you should see a Re:dash login screen.
+(``$REDASH_DIR`` refers to the top-level directory of your Redash git repository clone.)
+
+#. Install docker and docker-compose.
+#. ``cd $REDASH_DIR``.
```
arikfr: Why not keep the original instruction (#2 above)? I find it much more understandable than this. You don't use $REDASH_DIR anywhere else anyway.

docs/setup.rst
```diff
@@ -140,16 +144,17 @@ should have a running Re:dash instance with everything you need to get
 started . Re:dash should be available using the server IP or DNS name
 you assigned to it. You can point your browser to this address, and login
 with the user "admin" (password: "admin"). But to make it useful, there are
-a few more steps that you need to manually do to complete the setup:
+a few manual steps to complete the setup:
 
-First ssh to your instance and change directory to ``/opt/redash``. If
-you're using the GCE image, switch to root (``sudo su``).
+First connect to your instance using ssh, and switch to root (``sudo su``).
```
arikfr: why the change?

docs/setup.rst
```diff
@@ -69,13 +69,17 @@ run the :doc:`upgrade process <upgrade>` and then proceed to `"Setup" <#setup>`_
...
-5. Run ``docker-compose up``
-6. Run ``docker-machine ls``, take note of the ip for the Docker machine you are using, and open the web browser.
-7. Visit that Docker machine IP at port 80, and you should see a Re:dash login screen.
+(``$REDASH_DIR`` refers to the top-level directory of your Redash git repository clone.)
+
+#. Install docker and docker-compose.
+#. ``cd $REDASH_DIR``.
+#. ``cp docker-compose-example.yml docker-compose.yml``
+#. ``docker-compose up postgres``.
+#. Change to another terminal.
```
arikfr: why?

docs/setup.rst
```diff
@@ -140,16 +144,17 @@ should have a running Re:dash instance with everything you need to get
...
 you assigned to it. You can point your browser to this address, and login
 with the user "admin" (password: "admin"). But to make it useful, there are
-a few more steps that you need to manually do to complete the setup:
+a few manual steps to complete the setup:
 
-First ssh to your instance and change directory to ``/opt/redash``. If
-you're using the GCE image, switch to root (``sudo su``).
+First connect to your instance using ssh, and switch to root (``sudo su``).
+
+If you're using the Docker image, ``docker exec -i -t redash_redash_1 /bin/bash``, will put you in ``/opt/redash/current`` as root.
```
arikfr: 1. They should use Compose's exec. This way no need to assume that the name will be `redash_redash_1`.
2. I'm not sure this is needed.  Is there a change they can'tdo via the env variables?

docs/setup.rst
```diff
@@ -140,16 +144,17 @@ should have a running Re:dash instance with everything you need to get
...
 
 Users & Google Authentication setup
 -----------------------------------
 
-Most of the settings you need to edit are in the ``/opt/redash/.env``
-file.
+Most of the settings you need to edit are in the ``/opt/redash/current/.env``
+file. If it doesn't exist, ``cp /opt/redash/current/.env.example /opt/redash/current/.env``.
```
arikfr: The instructions assume you either use one of the provided images or the bootstrap script. The `.env` File will always be there and `/opt/redash/current/.env` is a symlink to `/opt/redash/.env` (so it's not lost when upgrading).

Note that for Docker you should *not* use the .env file but rather Docker's env variables.

PR #1381, opened 2016-11-09
Add: documentation links for data sources

washort:
Add a customizable link to documentation for a data source, defaulting to the reference docs for the query language for that source.

### DIFF
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -100,7 +100,7 @@
               <span class="text-muted">Data Source</span>
               <select ng-disabled="!isQueryOwner" ng-model="query.data_source_id" ng-change="updateDataSource()"
                       ng-options="ds.id as ds.name for ds in dataSources"></select>
-
+              <a ng-if="dataSource.options.doc_url" ng-href={{dataSource.options.doc_url}}>{{dataSource.type_name}} documentation</a>
               <div class="pull-right">
                 <button class="btn btn-s btn-default" ng-click="togglePublished()" ng-if="query.is_draft && query.id != undefined && (isQueryOwner || currentUser.hasPermission('admin'))">
                   <span class="fa fa-paper-plane"></span> Publish
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -100,7 +100,7 @@
               <span class="text-muted">Data Source</span>
               <select ng-disabled="!isQueryOwner" ng-model="query.data_source_id" ng-change="updateDataSource()"
                       ng-options="ds.id as ds.name for ds in dataSources"></select>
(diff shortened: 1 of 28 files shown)


rd_ui/app/views/query.html
```diff
@@ -110,7 +110,7 @@ <h4 class="modal-title">Query Archive</h4>
               <span class="text-muted">Data Source</span>
               <select ng-disabled="!isQueryOwner" ng-model="query.data_source_id" ng-change="updateDataSource()"
                       ng-options="ds.id as ds.name for ds in dataSources"></select>
-
+              <a ng-href={{dataSource.doc_url}}>{{dataSource.type_name}} documentation</a>
```
arikfr: can you add a screenshot of how it looks like?

arikfr: also worth showing only if the url isn't empty.

washort: screenshot from 2016-11-11 11-38-55

Fixed so that it should only show when doc_url is present.

rd_ui/app/views/query.html
```diff
@@ -110,7 +110,7 @@ <h4 class="modal-title">Query Archive</h4>
               <span class="text-muted">Data Source</span>
               <select ng-disabled="!isQueryOwner" ng-model="query.data_source_id" ng-change="updateDataSource()"
                       ng-options="ds.id as ds.name for ds in dataSources"></select>
-
+              <a ng-if="dataSource.options.doc_url" ng-href={{dataSource.options.doc_url}}>{{dataSource.type_name}} documentation</a>
```
arikfr: As I'm not sure how I feel about this, can you hide it behind a feature flag (with default being off)?

(check how I did it for the [Manage Permissions](https://github.com/getredash/redash/blob/master/rd_ui/app/views/query.html#L128) button)

Thanks.

washort: Feature flag added.

arikfr:
Coming to thin of it -- why do we need this as a configuration value for each data source?

washort:
The reason for configuration values is so that individual data sources can link to data-specific docs rather than just the language reference. (Or to the version of the docs matching the backend DB version, etc.)

rafrombrc:
Hi! Just following up on this. We'd really like to get this deployed for our local installation within the next week. Is there a chance of getting this PR merged upstream soon, or should we add it to our own fork for now?

Thanks for your attention!

arikfr:
Let me review this again on Sunday and get back to you.

arikfr:
I don't want to make it hard on you to maintain your fork, but on the other hand don't feel 100% with merging this. As even when it's behind a feature flag, it introduces code that we will have to maintain that won't benefit the general users population.

Let's keep this on your fork for now, and if we see it makes syncing with upstream a hassle, will revisit to think what we can do about this.

How's that sound?

washort:
We're always going to have some stuff that needs to be in our own fork. I don't think this will be a burden for now. Thanks for having a look.

PR #1399, opened 2016-11-16
Add: dashboards list page (with #tags)

luke14free:
test

### DIFF
--- a/frontend/app/assets/css/redash.css
+++ b/frontend/app/assets/css/redash.css
@@ -679,3 +679,14 @@ div.table-name:hover {
     stroke-opacity: .2;
 }
 
+/*Dashboard list view */
+.no-margin{
+  margin:0px;
+}
+.two-px-margin{
+  margin:2px;
+}
+
+.taglist{
+  margin-top:20px;
+}
--- a/frontend/app/assets/css/redash.css
+++ b/frontend/app/assets/css/redash.css
@@ -680,13 +680,6 @@ div.table-name:hover {
 }
 
 /*Dashboard list view */
-.no-margin{
-  margin:0px;
-}
-.two-px-margin{
+.m-2{
   margin:2px;
-}
-
-.taglist{
-  margin-top:20px;
 }
--- a/frontend/app/pages/dashboards/dashboard-list.html
+++ b/frontend/app/pages/dashboards/dashboard-list.html
@@ -3,8 +3,8 @@
   <div class="col-lg-3">
     <input type='text' class='form-control' placeholder="Search Dashboards..." 
       ng-change="$ctrl.tableParams.reload()" ng-model="searchText"/>
-    <div class='list-group taglist'>
-      <h3 class='list-group-item no-margin'>Tags</h3>
+    <div class='list-group m-t-20'>
+      <h3 class='list-group-item m-0'>Tags</h3>
       <a ng-repeat='tag in $ctrl.allTags' ng-class='{"active": $ctrl.tagIsSelected(tag)}' 
       class='list-group-item' ng-click='$ctrl.toggleTag(tag)'>
         {{ tag }}
@@ -21,7 +21,7 @@
               {{ dashboard.untagged_name }}
             </a>
           </td>
-          <td data-title="'Tags'"><span class="label label-primary two-px-margin" ng-bind="tag" ng-repeat="tag in dashboard.tags"></span></td>
+          <td data-title="'Tags'"><span class="label label-primary m-2" ng-bind="tag" ng-repeat="tag in dashboard.tags"></span></td>
           <td data-title="'Created At'">{{ dashboard.created_at | dateTime }}</td>
         </tr>
       </table>
--- a/frontend/app/pages/dashboards/dashboard-list.html
+++ b/frontend/app/pages/dashboards/dashboard-list.html
@@ -1,0 +1,30 @@
+<div class='container row'>
+  <page-header title="Dashboards"></page-header>
+  <div class="col-lg-3">
+    <input type='text' class='form-control' placeholder="Search Dashboards..." 
+      ng-change="$ctrl.tableParams.reload()" ng-model="searchText"/>
+    <div class='list-group taglist'>
+      <h3 class='list-group-item no-margin'>Tags</h3>
+      <a ng-repeat='tag in $ctrl.allTags' ng-class='{"active": $ctrl.tagIsSelected(tag)}' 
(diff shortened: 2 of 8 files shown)


frontend/app/pages/dashboards/dashboard-list.js
```diff
@@ -0,0 +1,96 @@
+import template from './dashboard-list.html';
+import {_} from 'underscore';
+
+function DashboardListCtrl($scope, Dashboard, $location, currentUser, clientConfig, NgTableParams) {
+  const self = this;
```
arikfr: You don't need `self` - when using arrow functions, they preserve `this`.

redash/handlers/dashboards.py
```diff
@@ -25,8 +25,11 @@ def get(self):
 class DashboardListResource(BaseResource):
     @require_permission('list_dashboards')
     def get(self):
-        dashboards = [d.to_dict() for d in models.Dashboard.all(self.current_org, self.current_user.groups, self.current_user)]
-        return dashboards
+        results = models.Dashboard.all(self.current_org, self.current_user.groups, self.current_user)
+        page = request.args.get('page', 1, type=int)
+        page_size = request.args.get('page_size', 25, type=int)
+        dashboards = models.Dashboard.all(self.current_org, self.current_user.groups, self.current_user)
+        return paginate(results, page, page_size, lambda q: q.to_dict())
```
arikfr: If you return paginated results, then you can't implement tag filtering or search in the client side. Otherwise it will only apply to the current page you're on.

frontend/app/assets/css/redash.css
```diff
@@ -679,3 +679,14 @@ div.table-name:hover {
...
 
+/*Dashboard list view */
+.no-margin{
+  margin:0px;
+}
+.two-px-margin{
+  margin:2px;
+}
+
+.taglist{
```
arikfr: You can use `m-t-20` instead.

frontend/app/assets/css/redash.css
```diff
@@ -679,3 +679,14 @@ div.table-name:hover {
     stroke-opacity: .2;
 }
 
+/*Dashboard list view */
+.no-margin{
+  margin:0px;
+}
+.two-px-margin{
```
arikfr: For consistency with the other classes, can you name it `m-2`?

frontend/app/assets/css/redash.css
```diff
@@ -679,3 +679,14 @@ div.table-name:hover {
     stroke-opacity: .2;
 }
 
+/*Dashboard list view */
+.no-margin{
```
arikfr: You can use `m-0`.

frontend/app/pages/dashboards/dashboard-list.js
```diff
@@ -0,0 +1,100 @@
...
+  const page = parseInt($location.search().page || 1, 10);
+  const count = 25;
+
+  this.defaultOptions = {};
+  this.dashboards = Dashboard.query({}); // shared promise
+
+  $scope.selectedTags = []; // in scope because it needs to be accessed inside a table refresh
+  $scope.searchText = "";
+
+  $scope.$watch(function(){
```
arikfr: Why do you need this watch? You already call `reload` with `ng-change`.

Also it might be better to have an explicit search button.

frontend/app/pages/dashboards/dashboard-list.js
```diff
@@ -0,0 +1,100 @@
...
+    return $scope.searchText;
+  }, function(){
+    if(this){
+      this.defaultOptions.reload()
+    }
+  });
+
+  this.tagIsSelected = (tag) => {
+    return $scope.selectedTags.indexOf(tag) > -1;
```
arikfr: How about changing `selectedTags` into an object (map), so all these checks become much simpler?

frontend/app/pages/dashboards/dashboard-list.js
```diff
@@ -0,0 +1,100 @@
...
+
+  this.logoUrl = clientConfig.logoUrl;
+  const page = parseInt($location.search().page || 1, 10);
+  const count = 25;
+
+  this.defaultOptions = {};
+  this.dashboards = Dashboard.query({}); // shared promise
+
+  $scope.selectedTags = []; // in scope because it needs to be accessed inside a table refresh
+  $scope.searchText = "";
```
arikfr: Use `this` instead of `$scope`.

frontend/app/pages/dashboards/dashboard-list.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  this.tableParams = new NgTableParams({ page, count }, {
+    getData(params) {
+      const options = params.url();
+      $location.search('page', options.page);
+
+      const request = {};
+
+      return self.dashboards.$promise.then((data) => {
```
arikfr: If I understand this right, it means we will load the same dashboards from the server over and over again, whenever the user changes tags selection/search term. 

Let's load the dashboards once, and then only re-apply the filters.

Maybe use the router's resolve mechanism here to make sure we have dashboards loaded before the controller gets called?

frontend/app/pages/dashboards/dashboard-list.html
```diff
@@ -0,0 +1,30 @@
...
+    <div class='list-group taglist'>
+      <h3 class='list-group-item no-margin'>Tags</h3>
+      <a ng-repeat='tag in $ctrl.allTags' ng-class='{"active": $ctrl.tagIsSelected(tag)}' 
+      class='list-group-item' ng-click='$ctrl.toggleTag(tag)'>
+        {{ tag }}
+      </a>
+    </div>
+  </div>
+  <div class="col-lg-9">
+    <tab-nav tabs="$ctrl.tabs"></tab-nav>
```
arikfr: Are the tabs really needed?

redash/handlers/dashboards.py
```diff
@@ -7,7 +7,7 @@
 from redash import models
 from redash.models import ConflictDetectedError
 from redash.permissions import require_permission, require_admin_or_owner, require_object_modify_permission, can_modify
-from redash.handlers.base import BaseResource, get_object_or_404
+from redash.handlers.base import BaseResource, get_object_or_404, paginate
```
arikfr: no longer needed, right?

arikfr:
I might change a few things later on, but looks good to me 👍 Unless there is something you might want to add, I will merge it a bit later after I finish something on the webpack branch.

PR #1400, opened 2016-11-16
full-text search for query text, name, description

washort:
Adds a tsvector field to `queries` and a trigger to automatically fill it when the query text or metadata changes.

### DIFF
--- a/migrations/0029_index_queries.py
+++ b/migrations/0029_index_queries.py
@@ -1,0 +1,28 @@
+from playhouse.migrate import PostgresqlMigrator, migrate
+
+from redash.models import db
+from redash import models
+
+if __name__ == '__main__':
+    db.connect_db()
+    migrator = PostgresqlMigrator(db.database)
+
+    with db.database.transaction():
+        migrate(
+            migrator.add_column('queries', 'tsv', models.Query.tsv)
+        )
+        db.database.execute_sql("CREATE INDEX tsv_idx ON queries USING gin(tsv)")
+        db.database.execute_sql("UPDATE queries SET tsv = setweight(to_tsvector(coalesce(name, '')), 'A') || setweight(to_tsvector(coalesce(description, '')), 'B') || setweight(to_tsvector(coalesce(query, '')), 'C')")
+        db.database.execute_sql("""
+CREATE FUNCTION query_search_trigger() RETURNS trigger AS $f$
+begin
+  new.tsv :=
+      setweight(to_tsvector(coalesce(new.name, '')), 'A') ||
+      setweight(to_tsvector(coalesce(new.description, '')), 'B') ||
+      setweight(to_tsvector(coalesce(new.query, '')), 'C');
+  return new;
+end
+$f$ LANGUAGE plpgsql;
+""")
+        db.database.execute_sql("CREATE TRIGGER query_tsvectorupdate BEFORE INSERT OR UPDATE ON queries FOR EACH ROW EXECUTE PROCEDURE query_search_trigger();")
+    db.close_db(None)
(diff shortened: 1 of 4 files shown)


migrations/0029_index_queries.py
```diff
@@ -0,0 +1,28 @@
...
+    migrator = PostgresqlMigrator(db.database)
+
+    with db.database.transaction():
+        migrate(
+            migrator.add_column('queries', 'tsv', models.Query.tsv)
+        )
+        db.database.execute_sql("CREATE INDEX tsv_idx ON queries USING gin(tsv)")
+        db.database.execute_sql("UPDATE queries SET tsv = setweight(to_tsvector(coalesce(name, '')), 'A') || setweight(to_tsvector(coalesce(description, '')), 'B') || setweight(to_tsvector(coalesce(query, '')), 'C')")
+        db.database.execute_sql("""
+CREATE FUNCTION query_search_trigger() RETURNS trigger AS $f$
```
arikfr: The migration will work for existing deployments. But what about new ones? Have you considered an before_save/after_save hook instead of a DB trigger?

Also, did you check if it's too much of a performance hit if we do the calculation of the vector in "realtime" instead of storing it?

arikfr: Also, need to update all existing queries, right?

washort: Good points. Might be worth leaving this off until after #1417 is done.

arikfr: Sounds good to me. 👍

washort:
Updated to use alembic, tested upgrade with some sample queries. I didn't benchmark it but I'm fine with not adding the `queries.tsv` field if you'd rather leave it out.

migrations/0029_index_queries.py
washort: Ready for re-review.

arikfr:
I will review this (and other pending pull requests) after master is stable and v1 is released. Just don't have the bandwidth to do both :(

jezdez:
@arikfr Hey there, I'd like to help out with this feature and wondered if there is anything blocking from your side for the feature in general. Since I've got time to spend on this, please let me know how we can proceed.

In terms of general direction of the feature I'd like to propose to move in a different direct than this PR and adopt [SQLAlchemy-Searchable](https://sqlalchemy-searchable.readthedocs.io/) since it solves the full text search problem on top of PostgreSQL more complete, e.g. [flask-sqlalchemy integration](https://sqlalchemy-searchable.readthedocs.io/en/latest/integrations.html), advanced search queries with [search operators](https://sqlalchemy-searchable.readthedocs.io/en/latest/search_query_parser.html) (AND, OR, paratheses, negatives etc) and [alembic integration](https://sqlalchemy-searchable.readthedocs.io/en/latest/alembic_migrations.html).

If that direction makes sense to you, I'd be happy to work on a PR to get the ball rolling again. 
Otherwise I'm ready to pick this PR up and finish it as well. What do you prefer?

arikfr:
Hi @jezdez 👋

SQLAlchemy-Searchable looks good and seems to make implementation much easier! Should I close this PR then?

In the future it will be nice to have a unified search experience for all objects (queries, dashboards, alerts, users). I was wondering whether we should have a single table with the search vectors for all of them. But realized that we can continue with the current approach (each table has its own vector), but later create a view the unifies them.

Thanks!

jezdez:
@arikfr Hello :) Thanks for your quick reply!

I think it's fair to close the PR, I'll keep an eye on it when I work on the implementation based on SQLAlchemy-Searchable so there are no surprises for the review. I'd like to suggest to collect some experience with the package with the queries first and then work from there to add a unified search, especially since there is likely some work to be done from the UX perspective to get the search results rendering right, since the objects are quite different in terms of what people may want to search for.

arikfr:
> I'd like to suggest to collect some experience with the package with the queries first and then work from there to add a unified search

Agreed. 👍

arikfr:
@washort thank you for "paving" the way 👏

PR #1402, opened 2016-11-17
Change: correctly propagate ElasticSearch errors to the UI

adamlwgriffiths:
All erronous ES queries have been displaying an error stating the JSON could not be decoded, regardless of the actual error (HTTP, syntax, auth, etc).

This modification ensures any request error is propagated back.
It seems that the main issue was requests errors being non-json-serializable due to the usage of both " and ' characters.
The abstraction of the `_get_mapping` function also made it difficult to get the root error instead of just the exception, this has been modified to be similar to `run_query` and now returns a tuple of `(mappings, error)`.

### DIFF
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -17,27 +17,27 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 ELASTICSEARCH_TYPES_MAPPING = {
-    "integer" : TYPE_INTEGER,
-    "long" : TYPE_INTEGER,
-    "float" : TYPE_FLOAT,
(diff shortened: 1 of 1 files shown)


arikfr:
Thank you for doing some house keeping there as well! The ElasticSearch data source definitely needs some more love.

adamlwgriffiths:
Just noticed that I've mixed logging and logger usage (always do that =\). I'll fix this tomorrow and update the PR.

adamlwgriffiths:
Ok, should be good to go.
Changed incorrect usage of `logging` to `logger`.
Moved some of the syntax changes from the error commit back to the PEP-8 commit.

Issue #1188, opened 2016-07-14
[feature] continuous-error-bar plot


bobcolner:
Would love to see this plot.ly type supported:

https://plot.ly/javascript/continuous-error-bars/

Thanks

arikfr:
Implemented in #1433.

PR #1404, opened 2016-11-17
Add: error bands support to graphs (all except pie)

### LINKED ISSUE #795: feature request: allow to chart confidence intervals for line graphs
Suppose every row contains the error, it would be very helpful if the width of the line would be dynamically defined with that value.

luke14free:
f122d760-ac24-11e6-9821-a5c5935b3be5

Closes #795, #1188.


frontend/app/services/query-result.js
```diff
@@ -260,13 +266,13 @@ function QueryResultService($resource, $timeout, $q) {
 
         if (seriesName === undefined) {
           each(yValues, (yValue, ySeriesName) => {
-            addPointToSeries({ x: xValue, y: yValue }, series, ySeriesName);
+            errors.push(eValue);
+            addPointToSeries({ x: xValue, y: yValue, error_y: eValue }, series, ySeriesName);
```
arikfr: Should be `errorY` or `yError`.

frontend/app/services/query-result.js
```diff
@@ -218,12 +218,14 @@ function QueryResultService($resource, $timeout, $q) {
 
     getChartData(mapping) {
       const series = {};
+      const errors = [];
```
arikfr: Seems unused?

frontend/app/services/query-result.js
```diff
@@ -248,6 +249,10 @@ function QueryResultService($resource, $timeout, $q) {
             yValues[name] = value;
             point[type] = value;
           }
+          if (type === 'error_y') {
```
arikfr: for consistency let's change this too to `yError`?

luke14free: Mh, I am a bit confused by the implementation of the whole method. But I believe it needs to be `error_y` because it's already the data structure that is feed to plotly (and the parameter that plotly want as input is called `error_y`). I will double check anyways

frontend/app/visualizations/chart/plotly.js
```diff
@@ -299,17 +300,23 @@ function PlotlyChart() {
 
           if (useUnifiedXaxis && index === 0) {
             const yValues = {};
+            const eValues = {};
 
-            data.forEach((row) => { yValues[row.x] = row.y; });
+            data.forEach((row) => {
+              yValues[row.x] = row.y;
+              eValues[row.x] = row.yError;
```
arikfr: Shouldn't we set eValues only if they are actually exist?

arikfr:
Seems like it picked up some of my commits... anyway you need to rebase :|

luke14free:
some stuff was seriously wrong with this commit, I was totally unable to rebase, I am closing this one and opening https://github.com/getredash/redash/pull/1430

PR #1405, opened 2016-11-17
[Datasources] Add: simple Google Analytics query runner

denisov-vlad:
To get data via this query runner you should copy&paste url from `API Query URI` field on this page:  https://ga-dev-tools.appspot.com/query-explorer/

for example:
`https://www.googleapis.com/analytics/v3/data/ga?ids=ga%3A84987497&start-date=2016-09-01&end-date=2016-09-30&metrics=ga%3AnewUsers&dimensions=ga%3AsubContinent&filters=ga%3AsubContinent%3D%3DAustralasia&samplingLevel=HIGHER_PRECISION&max-results=100`

### DIFF
--- a/redash/query_runner/google_analytics.py
+++ b/redash/query_runner/google_analytics.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 from base64 import b64decode
 import json
 import logging
@@ -32,7 +34,7 @@ types_conv = dict(
 )
 
 
-class GoogleAnalytics(BaseQueryRunner):
+class GoogleAnalytics(BaseSQLQueryRunner):
     @classmethod
     def annotate_query(cls):
         return False
@@ -61,6 +63,21 @@ class GoogleAnalytics(BaseQueryRunner):
 
     def __init__(self, configuration):
         super(GoogleAnalytics, self).__init__(configuration)
+
+    def _get_tables(self, schema):
+        accounts = self._get_analytics_service().management().accounts().list().execute().get('items')
+        if accounts is None:
+            raise Exception("Failed getting accounts.")
+        else:
+            for account in accounts:
+                schema[account['name']] = {'name': account['name'], 'columns': []}
+                properties = self._get_analytics_service().management().webproperties().list(
+                    accountId=account['id']).execute().get('items', [])
+                for property_ in properties:
+                    schema[account['name']]['columns'].append(
+                        u'{0} (ga:{1})'.format(property_['name'], property_['defaultProfileId'])
+                    )
+        return schema.values()
 
     def _get_analytics_service(self):
         scope = ['https://www.googleapis.com/auth/analytics.readonly']
--- a/redash/query_runner/google_analytics.py
+++ b/redash/query_runner/google_analytics.py
@@ -1,0 +1,111 @@
+from base64 import b64decode
+import json
+import logging
+from redash.query_runner import *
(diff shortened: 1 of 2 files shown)


redash/query_runner/google_analytics.py
```diff
@@ -0,0 +1,112 @@
...
+        super(GoogleAnalytics, self).__init__(configuration)
+
+    def _get_analytics_service(self):
+        scope = ['https://www.googleapis.com/auth/analytics.readonly']
+        key = json.loads(b64decode(self.configuration['jsonKeyFile']))
+        credentials = SignedJwtAssertionCredentials(key['client_email'], key["private_key"], scope=scope)
+        return build('analytics', 'v3', http=credentials.authorize(httplib2.Http()))
+
+    def run_query(self, query, user):
+        logger.info("Analytics is about to execute query: %s", query)
```
arikfr: `logger.debug`

redash/query_runner/google_analytics.py
```diff
@@ -0,0 +1,112 @@
...
+        params = parse_qs(urlparse(query).query, keep_blank_values=True)
```
arikfr: How about we expect a JSON object instead of a query string? It will make parsing simpler, but also less error prone (most users won't know how to properly escape the query string).

redash/query_runner/google_analytics.py
```diff
@@ -0,0 +1,112 @@
...
+            rows = []
+            for r in response['rows']:
+                d = {}
+                for c, value in enumerate(r):
+                    column_name = response['columnHeaders'][c]['name']
+                    column_type = filter(lambda col: col['name'] == column_name, columns)[0]['type']
+                    if column_type == TYPE_DATE:
+                        value = datetime.strptime(value, '%Y%m%d')
+                    elif column_type == TYPE_DATETIME:
+                        if len(value) == 10:
```
arikfr: and if not?

denisov-vlad: There's only one TYPE_DATETIME column on it has 10 chars length. So I can add raise Exception if length is different.

redash/query_runner/google_analytics.py
```diff
@@ -0,0 +1,136 @@
...
+import json
+import logging
+from redash.query_runner import *
+from redash.utils import JSONEncoder
+from urlparse import urlparse, parse_qs
+from datetime import datetime
+logger = logging.getLogger(__name__)
+
+try:
+    import gspread
```
arikfr: you don't need gspread here.

redash/query_runner/google_analytics.py
```diff
@@ -0,0 +1,136 @@
...
+                    column_type = filter(lambda col: col['name'] == column_name, columns)[0]['type']
+                    if column_type == TYPE_DATE:
+                        value = datetime.strptime(value, '%Y%m%d')
+                    elif column_type == TYPE_DATETIME:
+                        if len(value) == 10:
+                            value = datetime.strptime(value, '%Y%m%d%H')
+                        elif len(value) == 12:
+                            value = datetime.strptime(value, '%Y%m%d%H%M')
+                        else:
+                            raise Exception('Wrong datetime format')
```
arikfr: Let's make it more friendly: `"Unknown date/time format in results: '{}'".format(value)`.

arikfr:
👍 thanks for the quick follow up.

PR #1406, opened 2016-11-17
Add: custom js visualizations

luke14free:
test

### DIFF
--- a/frontend/app/visualizations/chart/index.js
+++ b/frontend/app/visualizations/chart/index.js
@@ -65,7 +65,6 @@ function ChartEditor(clientConfig) {
         scatter: { name: 'Scatter', icon: 'circle-o' },
       };
 
-      console.log(clientConfig);
       if (clientConfig.allowCustomJSVisualizations) {
         scope.chartTypes.custom = { name: 'Custom', icon: 'code' };
       }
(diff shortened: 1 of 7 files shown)


frontend/app/visualizations/chart/index.js
```diff
@@ -75,6 +76,11 @@ function ChartEditor(ColorPalette) {
         });
       };
 
+      scope.options.customCode = "// Available variables are x, ys, element, and Plotly.\n";
```
arikfr: Please use ES6 template strings here (created with backticks - "``"), which supports multi line strings.

frontend/package.json
```diff
@@ -53,7 +53,7 @@
     "ng-annotate": "^1.2.1",
     "ng-annotate-loader": "^0.2.0",
     "ng-table": "^2.1.0",
-    "plotly.js": "^1.16.0",
+    "plotly.js": "^1.2.0",
```
arikfr: ?

luke14free:
@arikfr I added the value to the clientConfig, but it was somehow defined as a constant in the `index.js` file. I guess that's it's linked to the fact that you want to separate backend and frontend and are still mocking that stuff, so I also added temporarily the switch to that constant in order to test the feature.

frontend/app/visualizations/chart/index.js
```diff
@@ -66,6 +65,11 @@ function ChartEditor(ColorPalette) {
         scatter: { name: 'Scatter', icon: 'circle-o' },
       };
 
+      console.log(clientConfig);
+      if (clientConfig.allowCustomJSVisualizations) {
```
arikfr: Need to use this in the render function too, so that someone clever won't manually submit this type.

arikfr:
@luke14free you understood it right about the `clientConfig` 👍

agenorfbp:
in my redash don't have many files listed in this pull, example: index.js.
I need create those files ?

agenorfbp:
This customization is very important for me and my organization, this Should be including in version 1.0

PR #1407, opened 2016-11-17
Change: options to the script runner for fully capable shell execution

luke14free:
This is something we find very useful in our company. It basically allows you to execute any script from your machine. It allows us to run custom Python/R scripts that interact with external services and do various things. The data from the script needs to be outputted to the stdout in the same JSON format that redash uses with all the other data sources.

Note that this is __very insecure__ and should be used very carefully.


arikfr:
Merged. Thanks!

PR #1409, opened 2016-11-18
[Datasources] Add: Add query runner for Yandex ClickHouse

denisov-vlad:
Query runner for this database: https://clickhouse.yandex

It works via HTTP interface and gets answer as JSON: https://clickhouse.yandex/reference_en.html#JSON

### DIFF
--- a/redash/query_runner/clickhouse.py
+++ b/redash/query_runner/clickhouse.py
@@ -104,7 +104,7 @@ class ClickHouse(BaseSQLQueryRunner):
         except Exception as e:
             data = None
             logging.exception(e)
-            error = str(e)
+            error = unicode(e)
         return data, error
 
 register(ClickHouse)
--- a/redash/query_runner/clickhouse.py
+++ b/redash/query_runner/clickhouse.py
@@ -1,0 +1,119 @@
+import json
+import logging
+from redash.query_runner import *
+from redash.utils import JSONEncoder
+import pprint
+logger = logging.getLogger(__name__)
+
+pp = pprint.PrettyPrinter()
+
+try:
+    import requests
+    enabled = True
+except ImportError as e:
+    logger.info(str(e))
+    enabled = False
+
+
+class ClickHouse(BaseSQLQueryRunner):
+    noop_query = "SELECT 1"
+
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "user": {
+                    "type": "string",
+                    "default": "default"
+                },
+                "password": {
+                    "type": "string"
+                },
+                "host": {
+                    "type": "string",
+                    "default": "127.0.0.1"
+                },
+                "port": {
+                    "type": "number",
+                    "default": 8123
+                },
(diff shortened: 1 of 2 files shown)


redash/query_runner/clickhouse.py
```diff
@@ -0,0 +1,120 @@
...
+
+            if table_name not in schema:
+                schema[table_name] = {'name': table_name, 'columns': []}
+
+            schema[table_name]['columns'].append(row['name'])
+
+        return schema.values()
+
+    def _send_query(self, data, stream=False):
+        url = 'http://{host}:{port}'.format(host=self.configuration['host'], port=self.configuration['port'])
```
arikfr: http and not https? why not let the user specify the full url instead of host/port pair?

redash/query_runner/clickhouse.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _clickhouse_query(self, query):
+        query += ' FORMAT JSON'
+        result = self._send_query(query)
+        columns = [{'name': r['name'], 'friendly_name': r['name'],
+                    'type': self._define_column_type(r['type'])} for r in result['meta']]
+        return {'columns': columns, 'rows': result['data']}
+
+    def run_query(self, query, user):
+        logger.info("Clickhouse is about to execute query: %s", query)
```
arikfr: Let's change this to `logger.debug`.

redash/query_runner/clickhouse.py
```diff
@@ -0,0 +1,120 @@
...
+        if query == "":
+            json_data = None
+            error = "Query is empty"
+            return json_data, error
+        try:
+            q = self._clickhouse_query(query)
+            data = json.dumps(q, cls=JSONEncoder)
+            error = None
+        except Exception as exc:
```
arikfr: If we just return the exception as the `error` value, we can let it bubble up and handle it there.

redash/query_runner/clickhouse.py
```diff
@@ -0,0 +1,120 @@
+import json
+import logging
+from redash.query_runner import *
+from redash.utils import JSONEncoder
+logger = logging.getLogger(__name__)
+
+try:
+    import requests
```
arikfr: No need to wrap this in `try`/`except`. `requests` is a direct dependency of the project.

redash/query_runner/clickhouse.py
```diff
@@ -112,9 +101,10 @@ def run_query(self, query, user):
             q = self._clickhouse_query(query)
             data = json.dumps(q, cls=JSONEncoder)
             error = None
-        except Exception as exc:
+        except Exception as e:
             data = None
-            error = str(exc)
+            logging.exception(e)
+            error = str(e)
```
arikfr: Btw, it's possible the exception might have some unicode characters, no? (like fragments of the query) So maybe better to use `unicode` here.

denisov-vlad: of course. changed.

arikfr:
Thanks!

PR #1410, opened 2016-11-18
Add: auto-refresh to dashboards

luke14free:


arikfr:
❤️  

Screenshot?

luke14free:
sorry, the dashboards contain confidential data, hope this is enough

screen shot 2016-11-18 at 4 29 25 pm

arikfr:
Sure. Just watned to see how the refresh control looks like.

arikfr:
I had in mind a UI that uses a button & drop down like this:

image

So when you click on the button - it refreshes all. But when you click on the arrow it gives you the auto refresh options. When you select one of them, the button changes to be the option you selected.

image

The markup is something like this:

```html
<div class="btn-group" uib-dropdown>
  <button id="split-button" type="button" class="btn btn-default">Refresh</button>
  <button type="button" class="btn btn-default" uib-dropdown-toggle>
    <span class="caret"></span>
    <span class="sr-only">Split button!</span>
  </button>
  <ul class="dropdown-menu" uib-dropdown-menu role="menu" aria-labelledby="split-button">
    <li role="menuitem"><a href="#">Every ...</a></li>
  </ul>
</div>
```

(and the `btn-default` class changes to `btn-primary` when toggled)

luke14free:
test

should be good to go!

rohanpd:
Can we disable or set certain refresh frequencies with this? I'm worried about added load on our cluster if there are many groups of people setting high refresh frequencies.

arikfr:
I think that we will move the refresh frequencies for the dashboards and queries to a configuration, so you can set what's the minimum you allow.

Also note that if several people refresh the same dashboard at the same time, it will run only once.

rohanpd:
I've created a Issue #1412 to followup my concern around minimum refresh cycles.

Issue #1124, opened 2016-06-14
Switch database code to SQLAlchemy instead of Peewee


arikfr:
Following the discussion in #876, we should switch the database code (models) to SQLAlchemy instead of Peewee.

(Related: #1123)

arikfr:
Done! :-)

PR #1417, opened 2016-11-21
Change: Replace Peewee with SQLAlchemy/Alembic

washort:
- [x] Describe current DB schema using SQLAlchemy's ORM
- [x] Switch test setup to SQLAlchemy
- [x] Rewrite all queries using SQLAlchemy query API
    - [x] Rewrite queries in CLI commands
    - [x] Rewrite queries in HTTP handlers
- [x] Fix admin views to work with SQLAlchemy
- [ ] Set up Alembic for schema migrations
    - [x] Add to the project (using Flask-Migrate)
    - [x] Transfer migrations from v1 to Alembic (only the `is_draft` one)
    - [x] Rename old migrations folder to `old_migrations` (might just delete it)
    - [ ] Add some code in startup script to check if we're at a recent version of the DB, and if not stop execution (with the option to disable this with a configuration override).
    - [x] Add "protection" script in the new migrations folder that prints a message to those who try to run migration "old style".
    - [x] Check if we need to call `alembic stamp head` when creating the DB at the first time.
- [x] Replace `MeteredModel` with SQLAlchemy timing events


redash/handlers/base.py
```diff
@@ -66,10 +66,11 @@ def require_fields(req, fields):
 
 
 def get_object_or_404(fn, *args, **kwargs):
-    try:
-        return fn(*args, **kwargs)
-    except DoesNotExist:
+    rv = fn(*args, **kwargs)
+    if rv is None:
```
arikfr: SQLA will never raise an exception for missing rows?

requirements.txt
```diff
@@ -4,6 +4,7 @@ Flask-Admin==1.1.0
 Flask-RESTful==0.3.5
 Flask-Login==0.3.2
 Flask-OAuthLib==0.9.2
+Flask-SQLAlchemy==2.1
```
arikfr: In another project I used [alchy](https://github.com/dgilland/alchy) which also has Flask-Alchy which is a drop-in replacement for it. The benefit is that we won't need Flask session whenever we're using the DB. The downside is that alchy never seemed to gain mind share unlike Flask-SQLAlchemy.

TL;DR: we can keep Flask-SQLA and see if it adds too much boilerplate to the jobs code. If it isn't, keep it. Otherwise consider alchy.

redash/models.py
```diff
@@ -689,24 +581,45 @@ def should_schedule_next(previous_iteration, now, schedule):
...
+
+class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
+    id = Column(db.Integer, primary_key=True)
+    version = Column(db.Integer)
+    org_id = Column(db.Integer, db.ForeignKey('organizations.id'))
+    org = db.relationship(Organization, backref="queries")
+    data_source_id = Column(db.Integer, db.ForeignKey("data_sources.id"), nullable=True)
+    data_source = db.relationship(DataSource)
+    latest_query_data_id = Column(db.Integer, db.ForeignKey("query_results.id"), nullable=True)
+    latest_query_data = db.relationship(QueryResult)
```
arikfr: Why do we need both `latest_query_data_id` and `latest_query_data`? (applies to all similar fields) SQLA doesn't have a convenience method to get the object id instead of loading the object itself otherwise?

redash/handlers/base.py
washort: its `.get()` and `.first()` methods return None when there's no entry, so I believe that's correct.

requirements.txt
washort: My suspicion is that it won't add boilerplate, calling `create_app()` should be most of it.

redash/models.py
washort: SQLA is a bit more explicit about this stuff; the `foo_id` field is the actual db column, and `foo` is the attribute for the related ORM object.

redash/handlers/base.py
arikfr: Maybe we can get rid of this helper, as it seems that Flask-SQLAlchemy has its own: [first_or_404](http://flask-sqlalchemy.pocoo.org/2.1/api/#flask.ext.sqlalchemy.BaseQuery.first_or_404) and [get_or_404](http://flask-sqlalchemy.pocoo.org/2.1/api/#flask.ext.sqlalchemy.BaseQuery.get_or_404).

requirements.txt
arikfr: 👍

redash/models.py
arikfr: 👍

requirements.txt
jeffwidman: Re: alchy, it seems the author's mostly moved onto https://github.com/dgilland/sqlservice#history

redash/handlers/base.py
jeffwidman: > SQLA will never raise an exception for missing rows?

See also `.one()` and `.one_or_none()`

redash/cli/database.py
```diff
@@ -2,13 +2,15 @@
...
 @manager.command
 def create_tables():
     """Create the database tables."""
-    from redash.models import create_db, init_db
-
-    create_db(True, False)
+    from redash.models import db, create_db, init_db
+    create_db(True, True)
     init_db()
+    db.session.commit()
```
jeffwidman: Why do you need the `db.session.commit()`? 

I assume `create_db` or `init_db` calls Flask-SQLAlchemy's `create_all()`, which will implicitly call a commit:
http://stackoverflow.com/questions/34410091/flask-sqlalchemy-how-can-i-call-db-create-all-and-db-drop-all-without-trigg

redash/models.py
```diff
@@ -1,238 +1,133 @@
...
-
-def _simple_value(v):
-    if isinstance(v, BaseModel):
-        return v.id
-
-    return v
+class TimestampMixin(object):
+    updated_at = Column(db.DateTime(True), default=db.func.now(),
+                           onupdate=db.func.now(), nullable=False)
+    created_at = Column(db.DateTime(True), default=db.func.now(),
```
jeffwidman: Maybe this should use `server_default`? 

http://stackoverflow.com/a/33532154/770425

redash/models.py
```diff
@@ -1,238 +1,133 @@
...
-            self._dirty.clear()
-            self.post_save(False)
-            return nrows
-
-
 class BelongsToOrgMixin(object):
     @classmethod
     def get_by_id_and_org(cls, object_id, org):
-        return cls.get(cls.id == object_id, cls.org == org)
+        return cls.query.filter(cls.id == object_id, cls.org == org).first()
```
jeffwidman: Will this query ever return more than one result? If not, should probably use `one()` or `one_or_none()`

redash/models.py
```diff
@@ -277,33 +172,27 @@ def has_access(self, obj, access_type):
...
 
-    # When Organization is used with LocalProxy (like the current_org helper), peewee doesn't recognize it as a Model
-    # and might call int() on it. This method makes sure it works.
-    def __int__(self):
-        return self.id
-
     @classmethod
     def get_by_slug(cls, slug):
-        return cls.get(cls.slug == slug)
+        return cls.query.filter(cls.slug == slug).first()
```
jeffwidman: I suspect this should also be a `one()` or `one_or_none()` as I think you want errors if more than one result is returned for a given slug.

redash/models.py
```diff
@@ -315,29 +204,30 @@ def is_public(self):
...
-    permissions = ArrayField(peewee.CharField, default=DEFAULT_PERMISSIONS)
-    created_at = DateTimeTZField(default=datetime.datetime.now)
+    id = Column(db.Integer, primary_key=True)
+    org_id = Column(db.Integer, db.ForeignKey('organizations.id'))
+    org = db.relationship(Organization, backref="groups")
+    type = Column(db.String(255), default=REGULAR_GROUP)
+    name = Column(db.String(100))
+    permissions = Column(postgresql.ARRAY(db.String(255)),
+                         default=DEFAULT_PERMISSIONS)
+    created_at = Column(db.DateTime(True), default=db.func.now())
```
jeffwidman: Probably want `server_default()`

jeffwidman:
Fixes #1124

redash/models.py
washort: Probably, but I don't want to change the schema until after I get things working as-is. (Hi Jeff! It's been a while! Never expected to see someone from SFSH commenting on my code :-)

arikfr: Now I'm curios what's SFSH :-)

arikfr:
@washort I finished with my work on the frontend (for now) and want to give you a hand here. I rebased you branch with the latest master & fixed  an issue with the settings/DATABASE_URL. Do you have some unpushed work or can I do a force push with these changes?

arikfr:
I updated the tests code and now we get real failures, but still many tests fail just because the database runs out of connections. I tried to compare how we manage the connection/session in `setUp`/`tearDown` with other projects and couldn't spot an obvious issue. :-(

arikfr:
Another SQLA question:

```python
    @classmethod
    def get_by_id_and_org(cls, visualization_id, org):
        return cls.query.join(Query).filter(cls.id == visualization_id, Query.org == org).one()
```

With peewee I could pass to such method either `Organization` object or just an Organization id. But in SQLA it seems that for `Query.org` I can only use `Organization` object and for `Query.org_id` I can use only integers.

Any middle ground?

washort:
No, you have to match up the right value with the right attribute. This is only really an issue in unit tests though, I think, because as far as I can tell, in the rest of the code you should only be using object ids when they're in query parameters; the rest of the time you can just pass around objects.

arikfr:
Working today on this branch was a reminder why I never liked SQLAlchemy in the first place :-\ It's very powerful, but why the simple stuff are so hard and verbose?

arikfr:
@washort 
When I created the getredash version of your branch, it somehow got messed up and included lots of unrelated commits. I tried rebasing, which resulted in a lot of wasted time and weird result. Eventually I just cherry-picked our commits on top of the latest master and it worked.

But I had to force push the result over your branch... I hope you didn't have anything uncommitted.

washort:
Looks like we're experiencing test failures due to webpack not running in CircleCI.

arikfr:
> Replace GFKBase usage with bridge tables

I'm not 100% sure about this. How will it work? Any reference implementation/documentation?

arikfr:
All tests pass now (I changed configuration to run Webpack) 💯 

But I greped the code for things like `update_instance` and `save` and still found usage of it. Mainly in the CLI but also in places where we were too lazy to write the tests ;-)

washort:
That's why I fixed the tests first -- to see where it'd be profitable to write more tests :-)

Re bridge tables - this is what I was looking at: http://docs.sqlalchemy.org/en/latest/_modules/examples/generic_associations/table_per_related.html

arikfr:
> Re bridge tables - this is what I was looking at: http://docs.sqlalchemy.org/en/latest/_modules/examples/generic_associations/table_per_related.html

This will work for `AccessPermission`, but won't work for `ApiKey`. For ApiKey we need to do the opposite lookup - given an api key, find the object associated with it.

redash/models.py
```diff
@@ -1,238 +1,140 @@
...
-
-    return v
+class TimestampMixin(object):
+    updated_at = Column(db.DateTime(True), default=db.func.now(),
+                           onupdate=db.func.now(), nullable=False)
+    created_at = Column(db.DateTime(True), default=db.func.now(),
+                           nullable=False)
 
 
 class ChangeTrackingMixin(object):
```
arikfr: How about implementing this with a an `before_update` or `after_update` event?

It seems that SQLA has the tools to determine if something was changed in the event. I tried to experiment with it, but couldn't get the event to trigger. :\

arikfr: I found out why the event didn't trigger and implemented it: https://github.com/getredash/redash/commit/e8739b39678db57a1ab31ef4d2747294d448590c

If you have no comments, I will push this change to your branch.

The part I'm not happy about is how we deduce the user who changed the object, but I'm not sure it's that bad. Eventually this code is only relevant for the API, which is Flask based... and I added some safeguards to make sure it doesn't cause harm outside of Flask context.

washort: I'll have a look at this next. What was the issue with the way it's done now?

arikfr:
@washort I did some updates to the CLI tests:

- Switch to using the "main" manager instance (the FlaskGroup) object, to ensure that the Flask app is set (there were tests failing because it was expecting the FLASK_APP env variable).
- Added missing `Session.commit` calls in the CLI commands.
- Switch to using `flask.cli.AppGroup` to create the subcommands so we don't have to wrap everything with `with_app_context` decorator.

There are still some (4?) tests failing because the CLI creates its own app_context/db session. I'm not sure how to solve this :-( One option is to create our own `with_app_context` decorator that will reuse an existing app context if one is available, but it feels wrong to change production code to accommodate tests...

arikfr:
Added Alembic (with Flask-Migrate). Updated the tasks to reflect integration status.

redash/cli/data_sources.py
```diff
@@ -2,13 +2,15 @@
...
+from flask.cli import AppGroup
+from sqlalchemy.orm.exc import NoResultFound
 
 from redash import models
 from redash.query_runner import query_runners
 from redash.query_runner import get_configuration_schema_for_query_runner_type
 from redash.utils.configuration import ConfigurationContainer
 
-manager = click.Group(help="Data sources management commands.")
+manager = AppGroup(help="Data sources management commands.")
```
washort: Oh nice, I didn't see that.

arikfr:
Aside from more fixes to broken functionality (if there is anything left) the only thing I want to add in this branch before merging is "Replace MeteredModel with SQLAlchemy timing events". All the rest can be a follow up IMO.

washort:
Had a look at the query-timing stuff available in SQLAlchemy - the main difficulty with replicating the current behavior is that by the time queries are executed, the only information available is the query text and its parameters; model class, method name etc. aren't accessible.

Any thoughts on how you want to handle this? Obviously we can parse the SQL to retrieve table names and operation type, if that's the path you prefer.

arikfr:
Let's keep the functionality of overall timing per request and # of queries
executed. Until now this is the only infrormation I actuallly used.

washort:
In that case I think we're done. I still want to change the schema a bit but that can happen in a new branch.

arikfr:
I've added some more metrics and .... it's merged! :)

jeffwidman:
Major congrats you two, this was a lot of work!

PR #1418, opened 2016-11-21
properly handle view_only permission in groups API

washort:
Cleans up an oversight in `DataSource.to_dict`.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -482,7 +482,7 @@ class DataSource(BelongsToOrgMixin, BaseModel):
             (('org', 'name'), True),
         )
 
-    def to_dict(self, all=False, with_permissions=False):
+    def to_dict(self, all=False, with_permissions_for=None):
         d = {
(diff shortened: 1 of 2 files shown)


redash/models.py
```diff
@@ -500,8 +500,10 @@ def to_dict(self, all=False, with_permissions=False):
             d['scheduled_queue_name'] = self.scheduled_queue_name
             d['groups'] = self.groups
 
-        if with_permissions:
-            d['view_only'] = self.data_source_groups.view_only
+        if with_permissions_for is not None:
+            d['view_only'] = DataSourceGroup.get(
```
arikfr: This way we need to make another database query, while we already have the object from calling `add_group`/`remove_group`. How about we change those methods to return the `DataSourceGroup` object, and then assign the `view_only` value in the handler code?

arikfr:
Implemented in #1417.

PR #1419, opened 2016-11-21
Change: use redash.utils.json_dumps instead of json.dumps in Python query runner

ehfeng:
Currently, python runner is returning datetime objects as unicode strings. `redash.utils.json_dumps` returns them as native python datetime objects.

### DIFF
--- a/redash/query_runner/python.py
+++ b/redash/query_runner/python.py
@@ -4,6 +4,7 @@ import logging
 import sys
 
 from redash.query_runner import *
+from redash.utils import json_dumps
 from redash import models
 
 import importlib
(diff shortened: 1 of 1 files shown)


ehfeng:
Still figuring out how to set this up locally, so I made this change blind, based on our gitter discussion.

arikfr:
Ping me on the chat if you want a walkthrough of the setup.

ehfeng:
@arikfr realized this changes behavior—for hosted redash, it may be worth messaging your users, in case their old queries break.

arikfr:
It's only an extension of the previous behavior - the data types `json_dumps` handles make the regular `json.dumps` call raise an exception.

But anyway we don't have the Python query runner enabled in the hosted version, as it's a security risk. While it runs in a sandbox, I quite easily managed to by pass it. And I'm not some hacker...

Thanks for the concern though :)

ehfeng:
Dang, lucky. Breaking changes and migrations are always a pain :P

arikfr:
True that 😱

PR #1420, opened 2016-11-22
Avoid cross site scripting risk point in query formatter

johnkearney:
Our internal security analysis found the potential for an XSS vulnerability in the query formatter. If not supplied with SQL, the query formatter returned exactly what was sent to it.


arikfr:
And how's that an issue?

PR #1424, opened 2016-11-23
Added support for running MongoDB queries on secondary in replicaset mode

erans:


SebC99:
Why is this PR still open? Is there any other way to set the mongodb read preference in redash?

erans:
Not sure why this hasn't been merged before. The main different is adding the "readPreference" flag so that by default, if there is a replicaSet configured you can tell it to prefer reading data from the replicaSets that are not the current master.

I've merged some minor conflicts that were there since this is a super old PR.

ilkarataev:
What about pull request?

erans:
I've fixed the conflicts and I think it should be merged.

arikfr:
@erans thanks for updating this PR. But it looks like `readPreference` is no longer used after the update with latest master?

redash/query_runner/mongodb.py
```diff
@@ -130,11 +130,24 @@ class MongoDB(BaseQueryRunner):
...
+                    'title': "Database Name"
+                },
+                'replicaSetName': {
+                    'type': 'string',
+                    'title': 'Replica Set Name'
+                },
+                'readPreference': {
+                    'type': 'string',
+                    'title': 'Read preference (when replica set exists): primary, primaryPreferred, secondary, secondaryPreferred, or nearest'
+                }
```
arikfr: You can now specify an enum (that will create a dropdown in the UI):

```
                "readPreference": {
                    "type": "string",
                    "extendedEnum": [
                        {"value": "primary", "name": "Primary"},
                        {"value": "primaryPreferred", "name": "Primary Preferred"},
                        {"value": "secondary", "name": "Secondary"},
                        {"value": "secondaryPreferred", "name": "Secondary Preferred"},
                        {"value": "nearest", "name": "Nearest"},
                    ],
                    "title": "Replica Set Read Preference",
                },
```

erans: Ooo... nice... It's been a while since I've contributed (and this PR was sitting around for far too long :-) ).

erans: Changed to the enum + made sure readPreference is being passed to MongoClient.

I set the default as "primaryPreferred" so that it doesn't change the original logic too much. If someone will set a different value than that would be used.

arikfr: :+1: thanks! This feature (enums in configuration schema) is what delayed this PR... it just took way longer than I expected to have this feature. Then by the time we had it, this PR was out of sync.

redash/query_runner/mongodb.py
```diff
@@ -158,10 +178,12 @@ def __init__(self, configuration):
         )
 
     def _get_db(self):
+        kwargs = {}
         if self.is_replica_set:
             db_connection = pymongo.MongoClient(
                 self.configuration["connectionString"],
                 replicaSet=self.configuration["replicaSetName"],
+                readPreference=self.configuration.get("readPreference", "primaryPreferred")
```
arikfr: The UI will use the first value of the enum as the default, which in this case is `primary`. If you want to have `primaryPreferred` as the default, you need to have it as the first one in the list.

erans: Updated it to work like that.

arikfr:
Congrats! That's my birthday gift for you :gift:

erans:
:-D Thanks!

redash/query_runner/mongodb.py
```diff
@@ -130,11 +130,31 @@ class MongoDB(BaseQueryRunner):
...
+                    'title': 'Connection String'
+                },
+                'dbName': {
+                    'type': 'string',
+                    'title': "Database Name"
+                },
+                'replicaSetName': {
+                    'type': 'string',
+                    'title': 'Replica Set Name'
+                },
```
jezdez, line 146: @arikfr I'm wondering why restyled didn't catch this change of style if black would have turned the single quotes into double quotes?

arikfr: 🤷 no idea.

Issue #1357, opened 2016-10-21
Add time of the day option when refreshing queries once a week/month


MarcoMene:
It's nice to have top possibility to refresh queries once per week/month.
It would be nicer to have the possibilty to choose the day of the week/month and time of refresh.

luke14free:
Since with celery we can achieve almost cron-like scheduling (http://docs.celeryproject.org/en/2.0-archived/getting-started/periodic-tasks.html#crontab-like-schedules), it would be nice to expose this to the end user, adding the possibility of running the query on certain days of the week or of the month.

For the days of the week we could show a tick-list of all the days of the week and then the user can chose when to refresh the query (usecase -> I don't want to refresh my dashboard on weekends).

For the days of the month we can have a simple text input asking the user to type, separated with a comma, all the days of the month in which he wants the query to be run.

Another, less user friendly idea, would be to let the user input a crontab-like string to decide when the queries can be updated.

luke14free:
An idea of UI could be the following:

<img width="629" alt="screen shot 2016-11-22 at 2 13 56 pm" src="https://cloud.githubusercontent.com/assets/166602/20525071/f0525610-b0bd-11e6-922f-b33d2662b6e2.png">

<img width="625" alt="screen shot 2016-11-22 at 2 15 02 pm" src="https://cloud.githubusercontent.com/assets/166602/20525102/18df7ba8-b0be-11e6-8467-a0a1a7437756.png">

<img width="634" alt="screen shot 2016-11-22 at 2 15 08 pm" src="https://cloud.githubusercontent.com/assets/166602/20525105/1ae712bc-b0be-11e6-9bcd-4dcb38b05ed9.png">

wdyt @arikfr? there would need to be some changes in the structure of the queries models.. etc

arikfr:
> Since with celery we can achieve almost cron-like scheduling (http://docs.celeryproject.org/en/2.0-archived/getting-started/periodic-tasks.html#crontab-like-schedules), it would be nice to expose this to the end user, adding the possibility of running the query on certain days of the week or of the month.

We don't use Celery's scheduler. The main reason being that our scheduler predates the use of Celery in Redash, but also because it's not clear how dynamic Celery's scheduler is. It seems to be improved to be more dynamic in Celery v4.0 (was recently released).

(I will comment about the UI next)

arikfr:
[...]

PR #1425, opened 2016-11-23
Add: advanced query scheduling (cron-like with simple UI)

luke14free:
test

This addresses issue: https://github.com/getredash/redash/issues/1357


migrations/0027_change_query_schedule_to_text.py
```diff
@@ -0,0 +1,9 @@
+from redash.models import db
+import peewee
+from playhouse.migrate import PostgresqlMigrator, migrate
+
+if __name__ == '__main__':
+    migrator = PostgresqlMigrator(db.database)
+    with db.database.transaction():
+        db.database.execute_sql('ALTER TABLE queries ALTER COLUMN schedule TYPE text;')
```
arikfr: Why not just change the length?

redash/models.py
```diff
@@ -685,6 +688,9 @@ def should_schedule_next(previous_iteration, now, schedule):
             previous_iteration = normalized_previous_iteration - datetime.timedelta(days=1)
 
         next_iteration = (previous_iteration + datetime.timedelta(days=1)).replace(hour=hour, minute=minute)
+    else:
```
arikfr: Don't you prefer to have only cron-style syntax in the DB and use only croniter?

Also this needs more tests to make sure it acts as intended.

frontend/app/pages/queries/schedule-dialog.html
```diff
@@ -3,16 +3,43 @@
...
+            ng-if="$ctrl.refreshType === 'advanced'" 
+            save-query="$ctrl.saveQuery">
+        </advanced-refresh>
+        <basic-refresh 
+            refresh-type="$ctrl.refreshType" 
+            ng-if="$ctrl.refreshType === 'basic'"
+            query="$ctrl.query" 
+            save-query="$ctrl.saveQuery">
+        </basic-refresh>
+        <div ng-if="!$ctrl.refreshType">
```
arikfr: Check how the other dialogs are structured, and use the same footer for the button.
Also I think it's better to have the button always presented, but disabled instead of hidden when the user is not allowed to save.

frontend/app/filters/index.js
```diff
@@ -32,6 +32,8 @@ export function durationHumanize(duration) {
 export function scheduleHumanize(schedule) {
   if (schedule === null) {
     return 'Never';
+  } else if (schedule.split(' ').length === 5) {
+    return `CRON tab: ${schedule}`;
```
arikfr: How about returning "Advanced Schedule" and having a pop over show description of the schedule?

migrations/0027_change_query_schedule_to_text.py
luke14free: 1) because the crontab string could grow very long (and I don't exactly know how long)
2) because text and char have the same performances (read the tip from https://www.postgresql.org/docs/9.1/static/datatype-character.html)

redash/models.py
luke14free: I thought about it. The issue with that is: 
1) we would need a migrator to migrate legacy queries, many things could go potentially wrong here.
2) we would need to be able to parse cron in the frontend and understand what a cron query mean e.g. -> */5 * * * * is "Every 5 minutes". This is might not be always very straightforward.

frontend/app/pages/queries/schedule-dialog.html
luke14free: 👍

frontend/app/filters/index.js
luke14free: We could but I am not a fan at all of pop-over because they work very poorly with touch devices. But if you really want it, I can implement it.

luke14free:
closing in favor of https://github.com/getredash/redash/pull/1434

PR #1428, opened 2016-11-23
Fixed Error import from cassandra

yershalom:
Hi,

Added import Error by mistake.
Replaced it by DriverException

### DIFF
--- a/redash/query_runner/cass.py
+++ b/redash/query_runner/cass.py
@@ -88,8 +88,10 @@ class Cassandra(BaseQueryRunner):
             error = None
         except DriverException as e:
             error = e
+            json_data = None
         except KeyboardInterrupt:
             error = "Query cancelled by user."
+            json_data = None
 
         return json_data, error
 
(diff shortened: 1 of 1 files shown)


redash/query_runner/cass.py
```diff
@@ -7,8 +7,9 @@
 logger = logging.getLogger(__name__)
 
 try:
-    from cassandra.cluster import Cluster, Error
+    from cassandra.cluster import Cluster
     from cassandra.auth import PlainTextAuthProvider
+    from cassandra import DriverException
```
arikfr: I think that the Cassandra driver version we have in the requirements file doesn't have this :-( Can you confirm?

arikfr: ?

yershalom: Hi @arikfr 
I got confused.
I didn't notice that I worked with cassandra-driver==3.4.0 on my test server.

yershalom:
@arikfr Made a new one

PR #1430, opened 2016-11-25
Add: error bands support to graphs (all except pie)

luke14free:
a06403f0-acae-11e6-8feb-1d00f4e002c9

(fixing problem with picked up commits in https://github.com/getredash/redash/pull/1404)


client/app/services/query-result.js
```diff
@@ -224,6 +224,7 @@ function QueryResultService($resource, $timeout, $q) {
         let seriesName;
         let xValue = 0;
         const yValues = {};
+        let eValue = 0;
```
arikfr: I just realized that we will always have an eValue... did you check how it affects queries that don't have a error column set?

luke14free: yes, they work just fine.

arikfr: if the error is 0 , plotly ignores it?

luke14free: @arikfr for some reason they do look fine even though with the errors = 0 they should look slightly different. I want to review this throughly on Monday, also I find very annoying that after I set an error/group by column I cannot clear it, so I want to add that feature as well.

luke14free:
closing as this must now go into master and there were some commits imported that weren't mine..

PR #1435, opened 2016-11-28
Whitelisting more builtin primitives

mattrobenolt:
This is also a slight refactor to make it easier to expand on this list
of builtins we allow.

Also, doesn't mutate the global `safe_builtins` dict every time from
`RestrictedPython`, instead, opts for copying the dict first.

### DIFF
--- a/redash/query_runner/python.py
+++ b/redash/query_runner/python.py
@@ -38,6 +38,11 @@ class CustomPrint(object):
 
 
 class Python(BaseQueryRunner):
+    safe_builtins = (
+        'sorted', 'reversed', 'min', 'max',
+        'sum', 'set',
+    )
+
     @classmethod
     def configuration_schema(cls):
         return {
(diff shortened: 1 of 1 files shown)


redash/query_runner/python.py
```diff
@@ -38,6 +38,11 @@ def __call__(self):
 
 
 class Python(BaseQueryRunner):
+    safe_builtins = (
+        'sorted', 'reversed', 'min', 'max',
+        'sum', 'set',
```
mattrobenolt: To be explicit, `sum` and `set` are the additions here.

mattrobenolt:
It seems these test failures are unrelated since other PRs are failing for the same reason.

arikfr:
Yes, the failures are for a different reason. I'll sort it out. I assume you tested these changes locally?

mattrobenolt:
> I assume you tested these changes locally?

You would be assuming incorrectly. :( I haven't set up re:dash locally to run against. I can do that later. We run it in a docker container ourselves for production.

mattrobenolt:
I'm building docker container locally and testing this out now. In theory, this should work just fine. :)

arikfr:
👍 it's quite easy to run the query runner alone, as it's just Python code.

mattrobenolt:
Alright, couldn't get docker container running, but just got it running through python enough to get into the CLI.

I did:

```
In [1]: from redash.query_runner.python import Python
In [2]: Python({}).run_query('print(sorted, reversed, min, max, sum, set)', None)
Out[2]:
('{"rows": [], "log": ["[2016-11-28T19:38:07.050052] (<built-in function sorted>, <type \'reversed\'>, <built-in function min>, <built-in function max>, <built-in function sum>, <type \'set\'>)"], "columns": []}',
 None)
```

as a test, so it appears that everything is patched in correctly.

fwiw, when building from master in docker, I get the following JavaScript error when trying to go to the `/new/query` page:

image

arikfr:
How do you "build" the frontend code in your Docker image?

Note that I wouldn't recommend using master for "production" usage at the moment. I merged earlier today big refactor of the frontend code, so there are some issues still.

mattrobenolt:
I just did `docker build --rm -t redash .` and ran this which should compile from whatever is in my local directory, so I guess it'd just be this: https://github.com/getredash/redash/blob/master/Dockerfile#L32

arikfr:
Strange, that I would expect to work. I'll give it a look tomorrow.

PR #1438, opened 2016-12-01
[ElasticSearch] Change: use simplejson for better error descriptions

adamlwgriffiths:
The standard json library only provides an error
that the json failed to parse (“No JSON object could be decoded”).
simplejson provides actual character and line numbers
along with an expected character in it's errors.
This makes writing elasticsearch queries much easier.

I have only applied this to the ElasticSearch query runner.
There are other modules that could use it for error-checking queries (MongoDB, etc).

screen shot 2016-12-01 at 11 41 46 am

### DIFF
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -1,4 +1,4 @@
-import json
+import simplejson as json
 import logging
 import sys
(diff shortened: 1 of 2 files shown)


arikfr:
Thanks.

PR #1441, opened 2016-12-02
Fix symlinks to /frontend. Should be /client

adamlwgriffiths:
This was preventing the latest code from working.


arikfr:
Thanks! 

Btw, I don't recommend using the current master branch for anything but development.

PR #1474, opened 2016-12-19
[Cassandra] Change: test connection query

yershalom:
…onnection

### DIFF
--- a/redash/query_runner/cass.py
+++ b/redash/query_runner/cass.py
@@ -9,7 +9,6 @@ logger = logging.getLogger(__name__)
 try:
     from cassandra.cluster import Cluster
     from cassandra.auth import PlainTextAuthProvider
-    from cassandra import ReadTimeout
     enabled = True
 except ImportError:
     enabled = False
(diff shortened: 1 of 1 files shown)


redash/query_runner/cass.py
```diff
@@ -85,6 +86,9 @@ def run_query(self, query, user):
             json_data = json.dumps(data, cls=JSONEncoder)
 
             error = None
+        except ReadTimeout as e:
+            error = e
```
arikfr: Error needs to be a string. But why handle ReadTimeout in a special way?

redash/query_runner/cass.py
```diff
@@ -9,13 +9,14 @@
...
     from cassandra.auth import PlainTextAuthProvider
+    from cassandra import ReadTimeout
     enabled = True
 except ImportError:
     enabled = False
 
 
 class Cassandra(BaseQueryRunner):
-    noop_query = "SELECT * FROM system"
+    noop_query = "SELECT * FROM system.compactions_in_progress"
```
arikfr: Every user has access to this? Is there something like `SELECT 1`?

yershalom: No  `SELECT 1` for CQL.
Most of test queries goes in the same manner like `SELECT now() FROM system.local;`

redash/query_runner/cass.py
arikfr: What do you mean by pass it?

PR #1475, opened 2016-12-20
[Cassnadra] Added set_keyspace for easier query cassandra

yershalom:


redash/query_runner/cass.py
```diff
@@ -72,6 +72,7 @@ def run_query(self, query, user):
                 connection = Cluster([self.configuration.get('host', '')])
 
             session = connection.connect()
+            session.set_keyspace('{}'.format(self.configuration['keyspace']))
```
arikfr: Why not `session.set_keyspace(self.configuration['keyspace'])`?

yershalom:
@arikfr 
Changed it, thanks for the python lesson :-)

arikfr:
Thanks!

PR #1478, opened 2016-12-21
[ElasticSearch] Fix: connection test was always succesfful

adamlwgriffiths:
Error was being created, but not returned. Which caused it
to always display as success.


arikfr:
Thanks!

PR #1479, opened 2016-12-21
[ElasticSearch] Add: enable schema browser

adamlwgriffiths:
Split _get_mappings into 2 functions.
Add _get_query_mappings which is used by existing logic
and parses the mappings json for query mappings.
Add get_schema which uses basic _get_mappings and recurses
through the mappings json to build a schema for the indices.


arikfr:
Thanks!

PR #1482, opened 2016-12-22
[Cassandra] Add: schema browser support & explicit protocol version

yershalom:
…o 3 (Also working with cassandra version 2.1 and above)

### DIFF
--- a/redash/query_runner/cass.py
+++ b/redash/query_runner/cass.py
@@ -1,7 +1,7 @@
 import json
 import logging
 
-from redash.query_runner import BaseSQLQueryRunner, register
+from redash.query_runner import BaseQueryRunner, register
 from redash.utils import JSONEncoder
 
 logger = logging.getLogger(__name__)
@@ -14,7 +14,7 @@ except ImportError:
     enabled = False
 
 
-class Cassandra(BaseSQLQueryRunner):
+class Cassandra(BaseQueryRunner):
     noop_query = "SELECT dateof(now()) FROM system.local"
 
     @classmethod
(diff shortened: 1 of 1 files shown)


redash/query_runner/cass.py
```diff
@@ -54,22 +54,36 @@ def type(cls):
...
-        return results, error
+        results, error = self.run_query(CF_query, None)
+        results = json.loads(results)
+        for row in results['rows']:
+            table_name = row['columnfamily_name']
+            if table_name not in schema:
+                schema[table_name] = {'name': table_name, 'columns': []}
+            CN_query = """
+            SELECT column_name FROM system.schema_columns where keyspace_name ='{}' and columnfamily_name ='{}';
+            """.format(self.configuration['keyspace'], table_name)
```
arikfr: Why not run: `SELECT columnfamily_name, column_name FROM system.schema_columns where keyspace_name ='{}'` and split them to column families in the code? This way you need to run only a single query instead of N queries.

redash/query_runner/cass.py
```diff
@@ -54,22 +54,36 @@ def type(cls):
         return "Cassandra"
 
     def _get_tables(self, schema):
-        query = """
+        CF_query = """
```
arikfr: `cf_query`

redash/query_runner/cass.py
```diff
@@ -14,7 +14,7 @@
     enabled = False
 
 
-class Cassandra(BaseQueryRunner):
+class Cassandra(BaseSQLQueryRunner):
```
arikfr: Why?

yershalom: Because when I use BaseQueryRunner i'm getting empty list `[]` for get_tables

arikfr: Correct, because you need to implement `get_schema` instead (`get_tables` is a method BaseSQLQueryRunner uses for `get_schema`).

arikfr:
What's the need for protocol version

yershalom:
I added this because I want to avoid from the warning log. If you are using Cassandra 2.1 so it throws a warning about downgrading the protocol version.
screen shot 2017-01-03 at 10 26 11 am

This is from the cassandra driver code - https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L1118

arikfr:
OK, if anyone will have a problem with this, we will add this as a parameter.

arikfr:
Btw, Scylla supports this too, right?

yershalom:
Yes.

arikfr:
Thanks ! :)

PR #1483, opened 2016-12-22
Added global params to dashboards

rockwotj:
I think master is currently in a broken state, so I couldn't actually make this change there, but I did start from the last release `v0.12.0.b2449` and made this change. I can't propose a PR from a tag or propose a new branch though.

If you prefix a parameter with `$` then when viewing it at the dashboard level it gets promoted to be a dashboard level param and queries can share these params.

Feature request here: https://discuss.redash.io/t/add-ui-for-query-params-at-dashboard-level/236/8

Video of the feature here: https://goo.gl/photos/UEjDv3fDhCrb2AnU6

I guess in order to merge this, we'd need a new branch for a `v0.13.0` release or something similar. When master is stable and I would be more than happy to add this there too (and make it slightly less hacky). Our team needs this to adopt redash, and we don't want to be stuck on an old version 😄

### DIFF
--- a/rd_ui/app/scripts/controllers/dashboard.js
+++ b/rd_ui/app/scripts/controllers/dashboard.js
@@ -14,14 +14,27 @@
     $scope.isFullscreen = false;
     $scope.refreshRate = 60;
     $scope.showPermissionsControl = clientConfig.showPermissionsControl;
+    $scope.globalParams = [];
+    $scope.reloadChildren = [];
 
     var renderDashboard = function (dashboard) {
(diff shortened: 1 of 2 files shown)


rockwotj:
There needs to be some way to update the global params list when a widget is added or removed. Any thoughts on this? Should there be a watch on the length of widgets?

arikfr:
Watch on widgets will work, although it's bad for performance. On the master branch, you can just do this in a callback after the add widget dialog is closed.

rockwotj:
SG

PR #1492, opened 2016-12-27
Fixing dynamic table

rockwotj:
If I entered a query for `SELECT 1` then I wasn't getting a result.

### DIFF
--- a/client/app/components/dynamic-table/index.js
+++ b/client/app/components/dynamic-table/index.js
@@ -12,8 +12,8 @@ function DynamicTable($sanitize) {
   this.orderByReverse = false;
 
   this.pageChanged = () => {
-    const first = this.count * (this.page - 1);
-    const last = this.count * (this.page);
+    const first = this.itemsPerPage * (this.page - 1);
+    const last = this.itemsPerPage * (this.page);
 
     this.rows = this.allRows.slice(first, last);
   };


client/app/components/dynamic-table/index.js
```diff
@@ -12,8 +12,8 @@ function DynamicTable($sanitize) {
   this.orderByReverse = false;
 
   this.pageChanged = () => {
-    const first = this.count * (this.page - 1);
-    const last = this.count * (this.page);
+    const first = this.itemsPerPage * (this.page - 1);
+    const last = this.itemsPerPage * (this.page);
```
arikfr: I'm not sure why we have both, but we [set them both to the same value](https://github.com/rockwotj/redash/blob/bc3691b5aedf5ec069690c27274953e01b083ee7/client/app/components/dynamic-table/index.js#L8). Why would this make a difference?

rockwotj: I think it's because `this.count` [is bound](https://github.com/rockwotj/redash/blob/bc3691b5aedf5ec069690c27274953e01b083ee7/client/app/components/dynamic-table/index.js#L68) and since the parent component is not specifying a value, it's undefined. Maybe a better fix is to remove `itemsPerPage` and have this be `(this.count || 15)`?

arikfr: But we do assign a value to `this.count` in the component initialization, so even if the parent doesn't provide a value we give it a default.

Also I couldn't reproduce this behavior (ran `select 1` and saw results). Maybe there is a different issue here?

arikfr: This issue just happened to me when npm installed Angular 1.6 instead of 1.5. Is it possible you're using Angular 1.6 too?

I'm going to commit npm-shrinkwrap file to the project to avoid such issues in the future.

rockwotj: I am using angular 1.6, good call. I'll close this then.

PR #1493, opened 2016-12-28
Fix: more `db.commit()` calls

washort:
Some missing commits in handlers and tasks that write to the db.

### DIFF
--- a/redash/handlers/alerts.py
+++ b/redash/handlers/alerts.py
@@ -28,7 +28,7 @@ class AlertResource(BaseResource):
             'object_id': alert.id,
             'object_type': 'alert'
         })
-
+        models.db.session.commit()
         return alert.to_dict()
 
     def delete(self, alert_id):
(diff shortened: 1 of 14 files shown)


redash/handlers/alerts.py
```diff
@@ -28,7 +28,7 @@ def post(self, alert_id):
             'object_id': alert.id,
             'object_type': 'alert'
         })
-
+        models.db.session.commit()
```
arikfr: The commit should be before the event, because if the commit fails we shouldn't log an event. In case you thought the event writes to the DB - the event just sends a task to Celery.

redash/handlers/events.py
```diff
@@ -1,12 +1,13 @@
...
+from redash.models import db
 
 
 class EventResource(BaseResource):
     def post(self):
         events_list = request.get_json(force=True)
         for event in events_list:
             self.record_event(event)
-
+        db.session.commit()
```
arikfr: No need for commit here.

arikfr:
Changed order and merged. Thanks!

PR #1494, opened 2016-12-29
[Impala] enable schema browser

yuua:
Thank you for this great tool

line 89, in the "redash/query_runner/impala_ds.py" causes 500 error
The impala schema list was not displayed.

This error is `show column stats %s`  impala query 
`columns_query = "show column stats %s" ` to `columns_query = "show column stats %s.%s" ` 

For this reason, I have sent a pull request.
Thank you.

### DIFF
--- a/redash/query_runner/impala_ds.py
+++ b/redash/query_runner/impala_ds.py
@@ -82,11 +82,11 @@ class Impala(BaseSQLQueryRunner):
     def _get_tables(self, schema_dict):
         schemas_query = "show schemas;"
         tables_query = "show tables in %s;"
-        columns_query = "show column stats %s;"
+        columns_query = "show column stats %s.%s;"
 
-        for schema_name in map(lambda a: a['name'], self._run_query_internal(schemas_query)):
-            for table_name in map(lambda a: a['name'], self._run_query_internal(tables_query % schema_name)):
-                columns = map(lambda a: a['Column'], self._run_query_internal(columns_query % table_name))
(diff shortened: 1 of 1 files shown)


redash/query_runner/impala_ds.py
```diff
@@ -82,11 +82,11 @@ def __init__(self, configuration):
     def _get_tables(self, schema_dict):
         schemas_query = "show schemas;"
         tables_query = "show tables in %s;"
-        columns_query = "show column stats %s;"
+        columns_query = "show column stats %s.%s;"
 
-        for schema_name in map(lambda a: a['name'], self._run_query_internal(schemas_query)):
-            for table_name in map(lambda a: a['name'], self._run_query_internal(tables_query % schema_name)):
-                columns = map(lambda a: a['Column'], self._run_query_internal(columns_query % table_name))
+        for schema_name in filter(lambda a: len(a) > 0, map(lambda a: str(a['name']), self._run_query_internal(schemas_query))):
```
arikfr: It's better to use `unicode` instead of `str`, in case it has non ascii characters. 

Also, in what cases it will be null/ of length 0?

yuua: Thank you!

Since there is no character other than ASCII , Change ```str``` to ```unicode```

>Also, in what cases it will be null/ of length 0?

I'm sorry... , table scheme is length 0 scheme no case

arikfr:
Thanks!

PR #1497, opened 2017-01-01
Fix #16: when updating dashboard name refresh dashboards dropdown

### LINKED ISSUE #16: When updating dashboard name, the dashboards list won't be updated until refresh of the page

arikfr:


AntoineAugusti:
@arikfr Could you explain why you chose to flatten the dashboard list in the menu? This looks like an important regression. Can we go back to the ability to group dashboards by their name?

We at Drivy heavily use dashboard grouping as we > 70 dashboards and I'd like 2 levels of groups instead of one.

arikfr:
@AntoineAugusti the new dropdown shows only recent dashboards, while all the dashboards are presented on the new dashboards page. Because the recent dashboards is limited in size (10?), there is no need for grouping.

AntoineAugusti:
@arikfr Oh, I missed the introduction of the new dashboards page. Could you show me the PR / a screenshot?

toyama-h:
Hmm.
We want grouping dashbord too..
We use 10 grouping dashboard.

PR #1504, opened 2017-01-04
Add: global parameters for dashboards

rockwotj:
Any variable prefixed with `$` will be elevated to a global param, check out the video below for a demo.

https://goo.gl/photos/3eXQRnJFuoyh4TRs5

This PR: #1483 for the master branch and much cleaner. Good work on the refactor the new codebase is much cleaner than before!

Feature request here: https://discuss.redash.io/t/add-ui-for-query-params-at-dashboard-level/236/8

/cc @tonymeng

### DIFF
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -31,27 +31,25 @@ function DashboardCtrl($rootScope, $scope, $routeParams, $location, $timeout, $q
     let globalParams = {};
     this.dashboard.widgets.forEach(row =>
       row.forEach((widget) => {
-        widget.getQuery().getParametersDefs().forEach((param) => {
-          if (param.name[0] === '$') {
-            const defaults = {};
-            defaults[param.name] = _.clone(param);
-            defaults[param.name].locals = [];
-            globalParams = _.defaults(globalParams, defaults);
-            globalParams[param.name].locals.push(param);
-          }
+        widget.getQuery().getGlobalParametersDefs().forEach((param) => {
+          const defaults = {};
+          defaults[param.name] = _.clone(param);
+          defaults[param.name].locals = [];
+          globalParams = _.defaults(globalParams, defaults);
+          globalParams[param.name].locals.push(param);
         });
       })
     );
     this.globalParameters = _.values(globalParams);
   };
 
-  $scope.$watch(() => this.globalParameters, (parameters) => {
-    _.each(parameters, (global) => {
+  this.onGlobalParametersChange = () => {
+    _.each(this.globalParameters, (global) => {
       _.each(global.locals, (local) => {
         local.value = global.value;
       });
     });
-  }, true);
+  };
 
   $scope.$on('deleteDashboardWidget', extractGlobalParameters);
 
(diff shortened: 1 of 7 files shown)


stevben:
+1 for this feature following this post (https://discuss.redash.io/t/add-ui-for-query-params-at-dashboard-level/236/10).
Any idea if this will be accepted soon ?

rockwotj:
Not sure it's up to @arikfr, I haven't heard from him yet.

client/app/pages/dashboards/dashboard.js
```diff
@@ -26,6 +27,32 @@ function DashboardCtrl($rootScope, $routeParams, $location, $timeout, $q, $uibMo
...
+
+  this.onGlobalParametersChange = () => {
+    _.each(this.globalParameters, (global) => {
+      _.each(global.locals, (local) => {
+        local.value = global.value;
+      });
+    });
+  };
+
+  $scope.$on('deleteDashboardWidget', extractGlobalParameters);
```
arikfr: Instead of using an event, let's add a callback function to the widget component.

client/app/pages/dashboards/dashboard.js
```diff
@@ -26,6 +27,32 @@ function DashboardCtrl($rootScope, $routeParams, $location, $timeout, $q, $uibMo
...
+          globalParams = _.defaults(globalParams, defaults);
+          globalParams[param.name].locals.push(param);
+        });
+      })
+    );
+    this.globalParameters = _.values(globalParams);
+  };
+
+  this.onGlobalParametersChange = () => {
+    _.each(this.globalParameters, (global) => {
```
arikfr: You can use `forEach` here and for `global.locals`.

rockwotj:
I kind of like the implicit, but I understand the importance of importance of feature discovery. I'll add it in.

rockwotj:
@arikfr added in an explicit toggle: https://goo.gl/photos/c94rHWcgHKyXSXET9

rockwotj:
@arikfr added in the explicit toggle: https://goo.gl/photos/8HWUN7nYro6Xv8Qq5

client/app/services/query.js
```diff
@@ -293,6 +294,21 @@ function QueryResource($resource, $http, $q, $location, currentUser, QueryResult
     return this.getParameters().get();
   };
 
+  Query.prototype.getLocalParametersDefs = function getLocalParametersDefs() {
+    if (!this.$localParameters) {
+      this.$localParameters = this.getParametersDefs().filter(p => !p.global);
```
arikfr: It's safe to cache because it's used only in dashboards, where the query won't change - right?

Maybe we should move these two functions to the dashboard code, to avoid someone using it "by accident" in another context?

rockwotj: It must be cached or angular can't detect that the state isn't changing. The normal parameters are cached, so why is this case any different?

https://github.com/getredash/redash/pull/1504/files/cb2e6fddf44d32fc22d54e5f2a2adffcbb44dce1#diff-35fc04aab5b4f115145224036f087c3cR285

arikfr: Normal parameters aren't cached -- whenever we call `Parameters.get` it calls `Parameters.updateParameters` (which is a no-op in case the query text didn't change). Or are we talking about different things?

rockwotj: Oh I see. The parameters object is cached, but I didn't look into what `get` did. I'll move it to dashboard.

rockwotj:
@arikfr PTAL

arikfr:
👌 thanks! :-)

rockwotj:
Whoop! Thank you

PR #1505, opened 2017-01-06
Add: get_by_id method to Query class

denisov-vlad:
It makes python query runner's get_query_result() works

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -830,6 +830,10 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
 
         return query
 
+    @classmethod
+    def get_by_id(cls, _id):
+        return db.session.query(cls).filter(cls.id == _id).one()
+
     def fork(self, user):
         forked_list = ['org', 'data_source', 'latest_query_data', 'description',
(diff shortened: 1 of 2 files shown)


redash/models.py
```diff
@@ -830,6 +830,10 @@ def recent(cls, group_ids, user_id=None, limit=20):
 
         return query
 
+    @classmethod
+    def get_by_id(cls, _id):
+        return db.session.query(cls).filter(cls.id == _id).one()
```
arikfr: This can be `cls.query.get(query_id)`.

denisov-vlad: @arikfr Maybe `cls.query[_id]` is better? `.get()` returns `None` and that's why it shows wrong exception message.

arikfr: The `[]` accessor, turns `cls.query` into a `SELECT *` query...

Your original approach is the way to go then. Just if you don't mind change it to `cls.query` instead of `db.session.query`.

denisov-vlad:
Fixed it in new PR: https://github.com/getredash/redash/pull/1513/commits/ab0a448c8799ab7a90b557b54b64ac37dd813399

PR #1507, opened 2017-01-06
Replace "generic foreign key" with table-per-association

washort:
This removes the `object_type` field from a few tables and replaces them with a table per model type. So instead of using `Change` you'd use `Query.Change` or `Dashboard.Change`. New models can participate in this by subclassing (e.g.) `ChangeTrackingMixin` and get their own associated table.


arikfr:
This is too "cold" to ever be merged. If you still think some of the changes here are relevant, please open an issue to discuss first. Thanks.

Closing this now.

PR #1509, opened 2017-01-09
[Presto/Athena] Change: remove special rule around public schema

GAwadhwalAtlassian:
Both the Presto and Athena query runners currently have a special rule for the "public" schema. But you do need to define the schema in all situations (even if it's public). I think this rule was copied across when we copied the code from the Postgres query runner.

This PR removes this special rule and treats "public" as just a normal schema.


arikfr:
Thanks 👍

PR #1512, opened 2017-01-10
[Python] Add: define more safe_builtins

denisov-vlad:


arikfr:
Are you sure all of them are safe? :)

denisov-vlad:
How do you understand function's safety?

I haven't added methods which can interact with filesystem.

arikfr:
> How do you understand function's safety?

Not opening a way to do unsafe operations :)

PR #1514, opened 2017-01-10
[Athena] Change: enable query annotations

GAwadhwalAtlassian:
The Athena query runner currently has the annotation (comment on top of query) disabled. This PR will enable it so we can see who is running what from Athena history.


arikfr:
Thanks.

PR #1516, opened 2017-01-11
fix alembic fail when run alembic upgrade head

Slach:
fix this behavior
    cd /opt/redash/migrations && alembic upgrade head

output:   FAILED: No 'script_location' key found in configuration.

### DIFF
--- a/migrations/alembic.ini
+++ b/migrations/alembic.ini
@@ -1,8 +1,10 @@
 # A generic, single database configuration.
 
 [alembic]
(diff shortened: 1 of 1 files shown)


arikfr:
You're not supposed to use the `alembic` command directly, but rather `manage.py db upgrade`. Do you get the same error when you use `manage.py`?

Slach:
manage.py working fine
this PR is mistake 
thanks ;)

PR #1524, opened 2017-01-15
Allow params once again in embeds

ziahamza:
An attempt to re-enable params in embeds.


stevben:
+1 !
I can be a beta tester if you want. I have a great use case for that.

ziahamza:
@stevben That would be great. We have also deployed this internally, and is working without bugs for the most part :)

arikfr:
"for the most part" -- there is a part it has bugs with? :)

ziahamza:
@arikfr The page shows an `Internal failure` when the params are not passed, as it crashes it. Should probably show a more appropriate error message as its not obvious. 

I can add a fix for ^ if the rest of the PR looks good

client/app/pages/queries/visualization-embed.js
```diff
@@ -33,7 +33,8 @@ export default function (ngModule) {
     return session($http, $route, Auth).then(() => {
       const queryId = $route.current.params.queryId;
       const query = $http.get(`/api/queries/${queryId}`).then(response => response.data);
-      const queryResult = $http.get(`/api/queries/${queryId}/results.json`).then(response => response.data);
+      // pass down any params from the top level url
+      const queryResult = $http.post(`/api/queries/${queryId}/results.json${location.search}`).then(response => response.data);
```
arikfr: Let's keep the old behavior and only use the new one, if it's enabled via [the feature flag](https://github.com/getredash/redash/blob/master/redash/settings.py#L236).

redash/handlers/query_results.py
```diff
@@ -101,6 +151,28 @@ def options(self, query_id=None, query_result_id=None, filetype='json'):
         return make_response("", 200, headers)
 
     @require_permission('view_query')
+    def post(self, query_id=None, filetype='json'):
```
arikfr: This should return error 400 if the feature flag is disabled, as it's a security concern. Worth adding a comment here about this as well.

client/app/pages/queries/visualization-embed.js
ziahamza: @arikfr Question: whats the best way to access featureFlags from the client side? Or should I pass in a param to the backend, and check the featureFlag from there?

ziahamza: resolve it by checking it on the server side 👍

ziahamza:
@arikfr updated the PR to only run queries when the featureFlag is set. Should be good to go now 👍

ziahamza:
@arikfr any updates on getting this PR upstream?

rohanpd:
We've been beta testing this and have come across some embeds that appear to time out still (regardless on whether they have parameters or not):

Failed to load resource: the server responded with a status of 502 (Bad Gateway)

ziahamza:
@rohanpd There is a timeout somewhere I think. For that I think it will break either way (either by redash or the http reverse proxies in general in front of redash).

ziahamza:
@arikfr would love to get your comments on this and push this to upstream.

morganizeit:
What's the status of this pull request? No updates for a month.

My org has ~160 parameterized embeds to facilitate operations. This is for fast reference of account details, eliminating multiple clicks.

Clicks become seconds become minutes... Since you can't put {{params}} into embedded views the GET params are the only way to filter a view without sending people out to the ReDash site, losing the context of where they were at. Embedding is important for getting the most from a reporting platform like ReDash.

It worries me that big parts of the embedding API can go missing without warning. For now our workaround is to manually edit all the iframes to point at an older version of ReDash in Staging. Lots of work and must be repeated when the bugfix is ready.

arikfr:
Now that we have a stable 1.0 release, we can get back to working on getting this into master.

@ziahamza:

> @rohanpd There is a timeout somewhere I think. For that I think it will break either way (either by redash or the http reverse proxies in general in front of redash).

There is a timeout in Gunicorn and might be in other proxies upstream (like ELB/ALB). But the problem introduced here is that previously queries without parameters were using the cached version of the query, hence not hitting the timeout. This pull request changed this behavior for all queries, instead of changing it for only queries with parameters.

Basically this pull request should work like the previous solution: introduce a change in behavior only when the feature flag is toggled and only to parameterized queries.

arikfr:
@morganizeit I understand your frustration, but just want to clarify something:

> It worries me that big parts of the embedding API can go missing without warning. 

1. This part of the API was always considered non production ready and was behind a feature flag for this reason.
2. "without warning" - it was made clear in the release notes, that the 1.0 release doesn't have support for this feature and that it will be introduced back after the 1.0 release.

morganizeit:
Hey thanks Arik! Just back from travel and catching up. Re-reading my comment it sounds pretty nasty. Yuck, please accept my apologies.

I'm not an admin of the system so I did not know this was a beta feature, off by default. I came across this thread and the release notes but did not understand that it was not _previously_ supported either, I thought it simply changed to unsupported in this release when it was dropped: https://discuss.redash.io/t/embed-visualization-with-parameters/477

Thanks for the clarification and for warming up the feature again. Parameterized embed enables showing data where people are already looking, increasing awareness of ReDash and what it can do.

Cheers,
Morgan.

nt3478:
+1

morganizeit:
Hi folks! Any updates RE sorting out obstacles for this feature?

ziahamza:
@arikfr updated the PR to only run queries from cache if query parameters are not passed. https://github.com/getredash/redash/pull/1524/files#diff-f734d32676ec526e17e5ae92d41b5787R193

does this work?

problem:
Another +1 on this!

arikfr:
Thanks @ziahamza , I will review it again.

morganizeit:
Aaaaalmost there... all the CodeClimate errors are about whitespace and line length. @ziahamza added the feature flag. Are there any other obstacles?

azpeters:
greatly anticipating this. any estimate for its planned release date?

morganizeit:
Our data team applied the PR changes to our internal fork and deployed, looking good so far. Thank you for working to bring this beta feature back!

Cheers,
Morgan.

redash/handlers/query_results.py
```diff
@@ -9,14 +10,63 @@
...
 
 
+#
+# Run a parameterized query synchronously and return the result
+# DISCLAIMER: Temporary solution to support parameters in queries. Should be
+#             removed once we refactor the query results API endpoints and handling
+#             on the client side. Please don't reuse in other API handlers.
+#
+def run_query_sync(data_source, parameter_values, query_text, max_age=0):
+    query_parameters = set(collect_query_parameters(query_text))
```
ziahamza: its the same as for the usual `run_query` as well: https://github.com/getredash/redash/pull/1524/files#diff-f734d32676ec526e17e5ae92d41b5787R71

phpcodecrafting:
Can we please prioritize this? It's been open for 5+ months. Additionally the last 3 releases have all incorrectly stated that the "next" release would bring `ALLOW_PARAMETERS_IN_EMBEDS` back.

arikfr:
@phpcodecrafting I meant the next major release (v2).

Anyway, merged.

arikfr:
@ziahamza thank you for your help here 👍

phpcodecrafting:
Awesome!  Many thanks to all involved!

PR #1521, opened 2017-01-13
[ElasticSearch] Fix: wrong variable name.

arikfr:
Fixes #1519.

@adamlwgriffiths FYI.

PR #1526, opened 2017-01-16
[ElasticSearch] Fix for #1521

adamlwgriffiths:
The merge from #1521 fixed the invalid variable error
but it fixed in in a way that introduced another error:
"dictionary changed size during iteration".
The correct way to apply the fix is to add a new `mappings`
dict which takes the result schema.

These errors crept through because of caching. I believed I was testing the code but I was actually inspecting cached query results and cached schemas. My apologies for this.


arikfr:
Thanks!

arikfr:
We should really add tests to the ES parser. Will help in future refactors :)

Issue #1527, opened 2017-01-16
timeout setting


lab79:
On long queries now we get an error running query - Connection timed out.
Please, add a timeout setting.

Database - Vertica

https://github.com/getredash/redash/blob/master/redash/query_runner/vertica.py
...
                'read_timeout': self.configuration.get('read_timeout', 600)
...

arikfr:
can you add full stacktrace from the log to understand where you get this timeout error?

PR #1528, opened 2017-01-16
[Vertica] Change: set longer read_timeout

lab79:
https://github.com/getredash/redash/issues/1527


lab79:
Yes, seconds https://pypi.python.org/pypi/vertica-python/ (mistakenly thought that millisecond)
Propose leave the default - 10 minutes (600)

arikfr:
Just noticed you added this to configuration, then yes - let's keep the default value same as the default. Also you need to add this to the `configuration_schema` so it will show up in the UI.

arikfr:
Thanks!

PR #1540, opened 2017-01-23
[MySQL] Change: faster retrieval of schema

yaningzhu:
Because `information_schema.tables` and `information_schema.columns` do not have indexes on them, when they get large, joining them becomes painful. We have a database with over 10k tables and more than 200k columns. The old query takes forever.

The new query do not look at `information_schema.tables` at all. This works if your `information_schema.columns` table is in good shape in terms of data integrity (and it should be).

Tested on MySQL 5.5, 5.6(Percona), 5.7(Percona). The schema has not changed since then.


arikfr:
Merged. Thanks!

PR #1546, opened 2017-01-25
Add: API docstrings

washort:
Some descriptions of API methods, suitable for use with sphinx.httpdomain.

I've got an example of the generated docs here: https://people-mozilla.org/~ashort/redash_docs/api.html


arikfr:
Thanks!!

PR #1553, opened 2017-01-29
Fixing deep links for query parameters

### LINKED ISSUE #1544: Deep linking broken for all non text query parameters
### Issue Summary

Deep linking to a date/time param will not cause the spinner to pick up the correct time

### Steps to Reproduce

1. Create a query with a date parameter
2. Deep link to it with a parameter for the date
3. The display for the query will not show

### Technical details:

* Redash Version: v12
* Browser/OS: Chrome
* How did you install Redash: From source

rockwotj:
Closes #1544 

I'd like to make the date format for a parameter configurable, but not sure how to do that. Is there a good way? I chose the one that worked best for [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#canonical-format_1).


rockwotj:
Nvm, found the client config value, using that!

arikfr:
While trying to understand why the datepicker doesn't work in v1 anymore, I found out that in newer Angular version they changed the input[date] directive to return a `Date` object. While handling this I practically solved the same issue as you have: #1649, but realized it only when I was practically done.

My approach there is a bit different:
1. The `value` property always has a string representation of the value. I think that in your implementation the `value` property will have a `Date` object, which I'm not sure how will be serialized (for default values) when saving the query.
2. I'm using ISO standard to serialize the dates instead of the client config value. This is both backward compatible with the behavior of previous implementation and seems to support most databases.

What do you think?

rockwotj:
I do like your approach more, however why not use the REDASH_DATE_FORMAT? Isn't that what the config value is for? While most databases might support it, it would be a shame if one did not, and had no way to change it without forking.

rockwotj:
Has for your first point, that is a problem. I fixed that but didn't seem to update this branch. I overrode the resource in the query to serialize to the correct string, but I don't quite like that approach

rockwotj:
Closing this in favor of #1649

rockwotj:
@arikfr this is the code I added to fix serialization when it gets saved. I don't really like having it, but this approach does let you pass the `clientConfig` value to the `Parameter` class and then not have it saved to the database.

```
const Query = $resource('api/queries/:id', { id: '@id' },                               
    {                                                                                     
      save: {                                                                             
        method: 'POST',                                                                   
        transformRequest: (data) => {                                                     
          const toSave = clone(data);                                                     
          toSave.options = clone(toSave.options);                                         
          if (has(toSave.options, 'parameters')) {                                        
            toSave.options.parameters = toSave.options.parameters.map((param) => {        
              const parameter = clone(param);                                             
              parameter.value = param.getValue();                                         
              return parameter;                                                           
            });                                                                           
          }                                                                               
          return angular.toJson(toSave);                                                  
        },                                                                                
      },         

```

PR #1555, opened 2017-01-31
Change: generate sourcemaps for modules (original source files)

44px:
This enables sourcemaps for original files instead of transpiled code.

Before (this is `app/components/filters.js`):
image

After:
image


arikfr:
Thanks!

Issue #1556, opened 2017-01-31
0027_add_draft_toggle.py makes every existing dashboards hide (in master branch)


kitsuyui:
Sorry, I found a nits bug in master branch.
(I know recent master branch is the status for very experimental stage. 😄 )

### Issue Summary

The old_migration scripts make every existing dashboards hide.
https://github.com/getredash/redash/blob/master/old_migrations/0027_add_draft_toggle.py#L11-L17

It is adding is_draft column both queries and dashboards.
However its update cares only queries.

So every dashboard will be `is_draft = true` .

```python
        migrate(
            migrator.add_column('queries', 'is_draft', models.Query.is_draft)
        )
        migrate(
            migrator.add_column('dashboards', 'is_draft', models.Query.is_draft)
        )
        db.database.execute_sql("UPDATE queries SET is_draft = (name = 'New Query')")
```

arikfr:
This migration was replaced by a [new one](https://github.com/getredash/redash/blob/master/migrations/versions/65fc9ede4746_add_is_draft_status_to_queries_and_.py) which takes care of this issue.

kitsuyui:
@arikfr Okay! I understood.

arikfr:
I kept the `old_migrations` folder for reference mainly. Not expected to be used... before upgrading to v1.0, users should upgrade to 0.12 (using the old method) and then to v1.0 with new one.

Later today I'm going to post a blog post with a new upgrade script that handles this.

PR #1557, opened 2017-01-31
Fix: #1556 0027_add_draft_toggle.py makes every existing dashboards hide

kitsuyui:
- https://github.com/getredash/redash/issues/1556


arikfr:
https://github.com/getredash/redash/issues/1556#issuecomment-276317723
^ Closing.

PR #1563, opened 2017-02-02
Change: send events to webhook as JSON with a schema.

### LINKED ISSUE #1552: [Feature Request] Update the Redash webhook payload to be a Snowplow-compatible JSON
This is a proposal to take the existing Redash webhook and make some minor changes to make it compatible with [Snowplow](https://github.com/snowplow/snowplow).

As well as allowing a Redash+Snowplow user to warehouse all of their Redash activity in Snowplow for audit and compliance, this has the additional benefit that it ties the Redash webhook to an agreed [JSON Schema format](http://json-schema.org/).

### JSON Schema

We have published a JSON Schema for the Redash webhook as part of Iglu Central, here:

https://github.com/snowplow/iglu-central/blob/master/schemas/io.redash.webhooks/event/jsonschema/1-0-0

The exact naming of the webhook schema (`io.redash.webhooks/event`) was checked with @arikfr and should allow Redash to add further more-specific webhooks in the future as needed.

The schema has been carefully specified so that the generated Redshift table maps very closely onto Redash's underlying event table:

https://github.com/snowplow/iglu-central/blob/master/sql/io.redash.webhooks/event_1.sql

### Steps to make the Redash webhook compatible

Currently the Redash webhook is a POST request with the event data encoded in the body as form data.

We want to change this to JSON, but specifically Iglu-compatible self-describing JSON (Snowplow's schema system is called [Iglu](https://github.com/snowplow/iglu)).

In place of the current POST payload, the Redash webhook should look like this:

```json
{
  "schema": "iglu:io.redash.webhooks/event/jsonschema/1-0-0",
  "data": {
    "id": 23,
    "user_id": null,
    "action": "DELETE",
    ...
  }
}
```

Notes:

* The event's integer types should be kept as numerics in JSON
* The `created_at` field should be a valid ISO8601 date-time

We can easily check any test webhook JSON against the JSON Schema to verify that they match (if there is an error in the JSON Schema, we can patch it in a new Iglu Central release).

### Outcome

When this has been implemented, it will be possible to warehouse all Redash webhooks in Snowplow **with zero coding**, using Snowplow's built-in [Iglu webhook adapter](https://github.com/snowplow/snowplow/wiki/Iglu-webhook-adapter).

There are some really interesting use cases here:

* Full audit log for Redash in Snowplow (allowing the events table in Redash to be regularly truncated?)
* Real-time detection (ML-driven?) of a bad actor based on their SQL queries (e.g. PII theft, corporate espionage)

arikfr:
Example of resulting event:

```json
{
   "data":{
      "user_id":null,
      "created_at":"2017-02-02T08:08:15+02:00",
      "object_type":"test",
      "org_id":1,
      "object_id":null,
      "additional_properties":{
         "test":1
      },
      "action":"test"
   },
   "schema":"iglu:io.redash.webhooks/event/jsonschema/1-0-0"
}
```

Closes #1552.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## v1.0.0-rc.2 - Unreleased
+
+### Changed
+
+- [#1563](https://github.com/getredash/redash/pull/1563) Send events to webhook as JSON with a schema.
+
 ## v1.0.0-rc.1 - 2017-01-31
 
 This version has two big changes behind the scenes:
--- a/redash/models.py
+++ b/redash/models.py
@@ -1316,13 +1316,24 @@ class Event(db.Model):
     action = Column(db.String(255))
     object_type = Column(db.String(255))
     object_id = Column(db.String(255), nullable=True)
-    additional_properties = Column(db.Text, nullable=True)
+    additional_properties = Column(MutableDict.as_mutable(PseudoJSON), nullable=True, default={})
     created_at = Column(db.DateTime(True), default=db.func.now())
 
     __tablename__ = 'events'
 
     def __unicode__(self):
         return u"%s,%s,%s,%s" % (self.user_id, self.action, self.object_type, self.object_id)
+
+    def to_dict(self):
+        return {
+            'org_id': self.org_id,
+            'user_id': self.user_id,
+            'action': self.action,
+            'object_type': self.object_type,
+            'object_id': self.object_id,
+            'additional_properties': self.additional_properties,
+            'created_at': self.created_at.isoformat()
+        }
 
     @classmethod
     def record(cls, event):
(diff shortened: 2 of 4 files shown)


arikfr:
@alexanderdean there are two changes here from your schema:

- No `id` field. I prefer to not include it, in case in the future we might not store these events to the local database. If you still want an ID for deduplication purposes, we can add UUID maybe?
- `additional_properties` is an object and not a string field.

alexanderdean:
Sounds good @arikfr - I've added https://github.com/snowplow/iglu-central/issues/484

We can't (yet) deduplicate webhook events based on webhook-specific fields, so no rush to add in an event `UUID` from our side (I agree however that it's better to expose a UUID than a current local database sequence)...

alexanderdean:
Not a deal-breaker but I would probably emit the `created_at` using UTC - so in this case it would be:

```
"created_at":"2017-02-02T06:08:15Z",
```

It's slightly more compact, and less brittle - the TZ of the instance running Redash shouldn't be significant at analytics time...

arikfr:
In most deployments the server time will be UTC anyway. I'll change it to be UTC regardless, but I guess it's less urgent.

alexanderdean:
Makes sense - thanks @arikfr !

PR #1564, opened 2017-02-02
Fix: options selection in the map visualization editor (lat, long, groupBy)

janusd:
Currently (1.0.0-rc1) column options of a result query won't show up in the map visualization editor (see attached images). Changing column name handling similar to the chart visualization editor.

screen shot 2017-02-02 at 15 56 01

screen shot 2017-02-02 at 15 54 42


arikfr:
Thanks!

Issue #25, opened 2013-10-28
Query versioning


arikfr:
- Each query should link to its parent, in case it's a fork of an existing query.
- When changing a query, store last N versions and let the user go back to an older version.

tsibley:
:+1: This would be great, and was one of the key cons on my list of deciding whether or not to adopt redash.

ChiragKParmar:
+1

tsibley:
Along the same lines, comparing results between queries could be very useful too.  A "Compared to previous version" tab next to the "Results" tab would be neat-o.  I often export results to CSV files and compare them with [daff](https://github.com/paulfitz/daff).

tsibley:
Daff has a JS library, so it could perhaps even be integrated directly into the frontend.

rohanpd:
@whummer has a PR #1113 to implement a changes table. This will be storing the versions of the query but not currently presenting a UI to compare versions.

arikfr:
@tsibley I'm planning to give access to past query results (they are stored anyway). Will keep daff in mind when it happens.

WesleyBatista:
I was wondering if would not be better change the approach. 
The main objective is to have the queries versioned, right? 
We know that Git is a great versioning system. We could use it.

In my conception we can use GitHub as a "queries source".

Basically it would work as follows: In Redash settings we would have to set up:

- REDASH_GITHUB_**REPO**
- REDASH_GITHUB_**USERNAME**
- REDASH_GITHUB_**ACCESS_TOKEN**

In the UI, on the query creation page, we would have a field where we would put the path to the query 
For example: `path/to/query.sql`.

Our handler will hit the GitHub servers at `https://api.github.com/repos/{repo}/contents/{path_to_query}`, given the auth previously get.

![redash_proposal](https://cloud.githubusercontent.com/assets/755254/16391130/18f8db72-3c7c-11e6-9468-d558512352ea.png)

I even started to make a prototype, but not finished it.
If someone find it is worth, please feel free to [fork me](https://gist.github.com/WesleyBatista/71ca5841515f81cf8523ea645a8fa0cc).

I think we could save a good development time with this idea and get the queries versioning working 😄

tsibley:
@WesleyBatista It's a good idea, and one I've sketched out on paper for working with a local clone with bidirectional sync rather than assuming Github.  I definitely like the idea of being able to look at query history in Git with my normal tools.

[...]

PR #1569, opened 2017-02-06
add compare query version support

spasovski:
How to trigger:

The modal itself:


redash/handlers/queries.py
```diff
@@ -264,3 +265,23 @@ def post(self, query_id):
...
+            models.Change.object_type == 'queries',
+            models.Change.object_id == query_id)
+        page = request.args.get('page', 1, type=int)
+        page_size = request.args.get('page_size', 25, type=int)
+        return paginate(results, page, page_size, lambda c: c.to_dict(full=False))
+
+
+class QueryVersionResource(BaseResource):
+    @require_permission('view_query')
+    def get(self, query_id, version_id):
```
arikfr: Why not use the `Change` object's primary key? It will make the query much more efficient and simpler.

redash/handlers/queries.py
```diff
@@ -264,3 +265,23 @@ def post(self, query_id):
         parameter_values = collect_parameters_from_request(request.args)
 
         return run_query(query.data_source, parameter_values, query.query_text, query.id)
+
+
+class QueryVersionListResource(BaseResource):
+    @require_permission('view_query')
+    def get(self, query_id):
+        results = models.Change.query.filter(
```
arikfr: This logic (finding changes for an object) should be in the model (i.e. a method like `Change.for` that takes the model).

Also you need to fetch the query first & check for proper permissions. You can see [an example in alerts](https://github.com/getredash/redash/blob/master/redash/handlers/alerts.py#L47-L49).

client/app/pages/queries/compare-query-dialog.js
```diff
@@ -0,0 +1,59 @@
...
+          span.appendChild(document.createTextNode(part.value));
+          fragment.appendChild(span);
+        });
+
+        return fragment;
+      }
+
+      document.getElementById('current-query-diff').appendChild(getDiffContent(currentDiff));
+      document.getElementById('previous-query-diff').appendChild(getDiffContent(previousDiff));
+    });
```
arikfr: This whole logic belongs in the template. You can easily express it with Angular:

```html
<span ng-class="{'diff-added': part.added, 'diff-removed': part.removed}" ng-repeat="part in diff">
{{part.value}}
</span>
```

client/app/assets/css/redash.css
```diff
@@ -667,3 +667,36 @@ div.table-name:hover {
 .m-2{
   margin:2px;
 }
+
+/* Compare Query Version container */
+/* Offers slight visual improvement (alignment) to modern UAs */
```
arikfr: You can place this along side the component. [Example](https://github.com/getredash/redash/blob/master/client/app/components/app-header/index.js#L5).

client/app/pages/queries/query.html
```diff
@@ -74,6 +74,22 @@
...
+        <select id="version-choice">
+          <option ng-repeat="version in versions" value="{{version.object_version}}">
+            {{version.object_version}}
+          </option>
+        </select>
+        <button class="btn btn-s btn-default" ng-click="compareQueryVersion()">
+          <span class="zmdi zmdi-play"></span>
+          Compare
+        </button>
+      </div>
```
arikfr: Next time, it will be great if you could include a screenshot of all the UI changes. But based on the markup -- I think it will take too much space.

How about linking the `updated at` timestamp to a UI similar to what Dropbox does in their Paper product:

image

Basically they show a list of all the recent changes. It's both efficient on the screen real estate, but also efficient in reviewing the changes, as you don't have to pick each individual version.

You just need to change the changes list API to return the full change set and then we save on multiple API calls. Another thing to consider is that not all changes are changes to the query text, so need to filter those out (can be done in the client side).

spasovski: Yea good call on more screenshots I was a bit too focused on the modal. The dropbox example wasn't the way we'd intended to do it though I can see some of its merits - we'd gone for a more general "compare current to..." approach. 

This change would require a more significant time investment so we might need to see if it fits Mozilla's needs. Tagging @rafrombrc on this.

client/app/pages/queries/compare-query-dialog.js
spasovski: Good call it shaved quite a few lines of JS.

client/app/pages/queries/query.html
rafrombrc: Hrm. I do like the UI that you're proposing, @arikfr. There are some cases where it might be a bit unwieldy.. if a query is >100 lines of code and has multiple significant edits then the single page stream might be a bit long, but even so we can probably work around that by having the diffs be collapsible or something.

Might you be open to proceeding with our current UI and then treating this change as a separate issue?

arikfr: Regardless of how we implement the preview in the modal, the triggering UI can't be as it is in this pull request. Maybe a middle ground is to have the "updated at" timestamp trigger the comparison UI, and in the modal window show the version picker (instead of having it on the main page). This way the impact is minimal.

spasovski:
Thanks for the comments - updated for re-review.

client/app/pages/queries/view.js
```diff
@@ -254,14 +254,28 @@ function QueryViewCtrl($scope, Events, $route, $routeParams, $location, $window,
 
     updateSchema();
     $scope.dataSource = find($scope.dataSources, ds => ds.id === $scope.query.data_source_id);
-    $scope.executeQuery();
```
alison985: Are you sure removing this line doesn't break anything?

package.json
```diff
@@ -38,6 +38,7 @@
     "d3": "^3.5.17",
     "d3-cloud": "^1.2.1",
     "debug": "^2.2.0",
+    "diff": "^3.2.0",
```
alison985: Given the issue we were having, should this be 3.3.0?

redash/handlers/queries.py
```diff
@@ -271,3 +273,19 @@ def post(self, query_id):
...
+
+
+class QueryVersionListResource(BaseResource):
+    @require_permission('view_query')
+    def get(self, query_id):
+        results = models.Change.list_versions(models.Query.get_by_id(query_id))
+        page = request.args.get('page', 1, type=int)
+        page_size = request.args.get('page_size', 25, type=int)
+
+        return paginate(results, page, page_size, lambda q: q.to_dict(full=False))
```
alison985: I don't see paginate options in the UI/html so I'm not sure how this paginate will react at query version 26?

redash/models.py
```diff
@@ -166,10 +166,6 @@ class ChangeTrackingMixin(object):
     skipped_fields = ('id', 'created_at', 'updated_at', 'version')
     _clean_values = None
 
-    def __init__(self, *a, **kw):
```
alison985: Why remove this? How is it related?

redash/models.py
```diff
@@ -194,13 +190,19 @@ def record_changes(self, changed_by):
...
+                prev = self._clean_values[col.name]
+                current = getattr(self, attr.key)
+                if prev != current:
+                    changes[col.name] = {'previous': prev, 'current': current}
 
-        db.session.add(Change(object=self,
-                              object_version=self.version,
-                              user=changed_by,
-                              change=changes))
+        if changes:
```
alison985: What do you return if this is false?

redash/models.py
```diff
@@ -947,6 +949,7 @@ def fork(self, user):
         kwargs = {a: getattr(self, a) for a in forked_list}
         forked_query = Query.create(name=u'Copy of (#{}) {}'.format(self.id, self.name),
                                     user=user, **kwargs)
+        forked_query.record_changes(changed_by=user)
```
alison985: This is the only place you refer to the concept of forking. Curious: what ends up getting written to the `changes` database table in this case?

tests/handlers/test_queries.py
```diff
@@ -158,3 +158,30 @@ def test_refresh_query_you_dont_have_access_to(self):
         user = self.factory.create_user(group_ids=[group.id])
         response = self.make_request('post', self.path, user=user)
         self.assertEqual(403, response.status_code)
+
```
alison985: You could also add tests for QueryVersionListResource, for the fork of a query, and for a revert if you wanted.

redash/handlers/dashboards.py
```diff
@@ -57,6 +57,7 @@ def post(self):
                                      user=self.current_user,
                                      is_draft=True,
                                      layout='[]')
+        dashboard.record_changes(changed_by=self.current_user)
```
alison985: I didn't see any additional dashboard references. Is recording a change to a dashboard already supported?

redash/models.py
```diff
@@ -748,7 +750,7 @@ def should_schedule_next(previous_iteration, now, schedule, failures):
 
 class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
     id = Column(db.Integer, primary_key=True)
-    version = Column(db.Integer, default=1)
+    version = Column(db.Integer, default=0)
```
alison985: Are there any current instances that have this set to 1 that would get an upgrade bug/unexpected behavior because of this?

redash/models.py
```diff
@@ -1075,7 +1078,6 @@ def to_dict(self, full=True):
             'id': self.id,
             'object_id': self.object_id,
             'object_type': self.object_type,
-            'change_type': self.change_type,
```
alison985: Why is this removed?

client/app/pages/queries/view.js
alison985: Tagging @ashort to answer my python-related questions for this PR.

redash/handlers/dashboards.py
washort: Yes, `Dashboard` subclasses `ChangeTrackingMixin`.

redash/models.py
washort: `None`

redash/models.py
washort: No, because defaults only affect new instances, and the bugfixes in `ChangeTrackingMixin` result in the version number being bumped immediately.

redash/models.py
washort: The initial change, same as any other creation of a query or dashboard

redash/models.py
washort: There's no attribute named `change_type`.

redash/handlers/queries.py
spasovski: Yep really good catch I think an earlier draft was going to have a paginated versions list.

arikfr:
As there is a new pull request for the same feature from @spasovski, closing this one.

shakismyname:
@arikfr Hi there! Just quickly curious - I saw that the pr got merged into the [mozilla fork](https://github.com/mozilla/redash/pull/149). But it appears that #25 is still open. Sounds like you're all still designing/working on it?

Thanks!

PR #1570, opened 2017-02-06
Update docker-compose.production.yml

btmc:
Nginx doesn't have "redash" upstream. This fixes the error.


iplaman:
thanks! should be merged

arikfr:
Thanks!

PR #1584, opened 2017-02-10
Add: scheduled query backoff in case of errors

washort:
This tracks scheduled query model objects all through the celery task for execution so that failure can be recorded. Number of failures are then used to decide how much longer to delay query execution.

### DIFF
--- a/migrations/versions/d1eae8b9893e_.py
+++ b/migrations/versions/d1eae8b9893e_.py
@@ -1,0 +1,25 @@
+"""add Query.schedule_failures
+
+Revision ID: d1eae8b9893e
+Revises: 65fc9ede4746
+Create Date: 2017-02-03 01:45:02.954923
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'd1eae8b9893e'
+down_revision = '65fc9ede4746'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('queries', sa.Column('schedule_failures', sa.Integer(),
+                                       nullable=False, server_default='0'))
+
+
+def downgrade():
+    op.drop_column('queries', 'schedule_failures')
--- a/redash/tasks/queries.py
+++ b/redash/tasks/queries.py
@@ -197,7 +197,8 @@ class QueryTask(object):
         return self._async_result.revoke(terminate=True, signal='SIGINT')
 
 
-def enqueue_query(query, data_source, user_id, scheduled=False, metadata={}):
+def enqueue_query(query, data_source, user_id, scheduled_query=None,
+                  metadata={}):
     query_hash = gen_query_hash(query)
     logging.info("Inserting job for %s with metadata=%s", query_hash, metadata)
     try_count = 0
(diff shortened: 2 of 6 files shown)


redash/tasks/queries.py
```diff
@@ -441,9 +457,13 @@ def run(self):
 
     def _annotate_query(self, query_runner):
         if query_runner.annotate_query():
+            if self.task.request.delivery_info:
```
arikfr: Why is this needed?

migrations/versions/d1eae8b9893e_.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'd1eae8b9893e'
+down_revision = '65fc9ede4746'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('queries', sa.Column('schedule_failures', sa.Integer(),
```
arikfr: I wonder if we should store this in the database. Maybe in Redis instead?

The main benefit of keeping it in the database is that we can easily query for it, to generate a report of failing queries (although then maybe we need `last_failure_at` column too?).

The downside is that it makes the table bloated with operational data and adds more calls to the database.

redash/tasks/queries.py
```diff
@@ -425,7 +435,13 @@ def run(self):
         if error:
             self.tracker.update(state='failed')
             result = QueryExecutionError(error)
+            if self.scheduled_query:
+                self.scheduled_query.schedule_failures += 1
+                models.db.session.add(self.scheduled_query)
         else:
+            if self.scheduled_query:
+                self.scheduled_query.schedule_failures = 0
```
arikfr: How about we do this update only if `schedule_failures` is `> 0`? Otherwise we generate unneeded update.

redash/tasks/queries.py
washort: this is None if invoked outside of celery (such as in unit tests)

migrations/versions/d1eae8b9893e_.py
washort: I wondered about using Redis but wasn't clear on how it should be done. I'll have another crack at it if that's what you'd prefer.

redash/tasks/queries.py
arikfr: :\ I don't like changing code to accommodate for tests (this will never be executed outside of a celery context otherwise). Two alternatives:

- Pass the routing key & task id to the constructor instead of the task directly. 
- Mock this.

WDYT?

migrations/versions/d1eae8b9893e_.py
arikfr: Both options have their benefits. Using Redis will allow us to store more metadata without feeling we're abusing the queries table :) I mean metadata like:

- What the error was.
- When the last error occurred.

washort:
Fixed the things you mentioned; mocking the celery stuff for test turned out to be easier than I thought.

arikfr:
Just realized something: we need to reset the `schedule_failures` counter when the user changes the query text.

washort:
Good point! Added.

arikfr:
Thanks!

PR #1588, opened 2017-02-10
Fix docker initialization script

ylatuya:
Fix initialization for docker removing the use of env variables like POSTGRES_PORT_5432_TCP_ADDR in favour of the service name "postgres"


arikfr:
Actually we no longer use this script. Did you find it referenced somewhere in the docs?

To avoid confusion, removed it (along with other unused setup files) in 9ef68361755a57bb951965e253497fce47c91e54.

ylatuya:
No references, I just found the scripts and though those would be needed. Thanks

PR #1592, opened 2017-02-13
Change: use https instead of git protocol for git repositories in npm-shrinkwrap.json

deecay:
Some proxies do not allow git:// protocol. There is a workaround for this such as `git config --global url."https://".insteadOf git://`. 

However, npm from Shrinkwrap now has limitations and the above workaround is being ignored. So an shrinkwrap file with git:// fails when installing Redash behind a proxy. Replacing git:// to https:// manually will make it work.


arikfr:
Merged. Thanks 👍

PR #1597, opened 2017-02-14
[SQL Server] Fix: schema won't load if the server is case sensitive

### LINKED ISSUE #1571: mssql get_schema() fails on case sensitive object name setups
### Issue Summary

get_schema() for MS-SQL fails on case sensitive object name setups.

### Steps to Reproduce

1. Define datasource for  case-sensitive-object-name MSSQL database instance.

### Technical details:

* Redash Version: 0.12.0

### How to fix

change
`        SELECT table_schema, table_name, column_name
        WHERE ...`
to
`        SELECT table_schema, table_name, column_name
        FROM INFORMATION_SCHEMA.COLUMNS
        WHERE...`
fixed my case.

### Where to go from here
I'm not really sure if this is the best way to fix this issue. If any MSSQL instance can be case-sensitive AND has this tablename in lower case, then this fix will not work. But I don't know if this can happen.

deecay:
Fixes _get_tables() for MSSQL with case sensitive object setup.

Changing the object name INFORMATION_SCHEMA.COLUMNS to uppercase makes the query compatible with both case insensitive and case sensitive setups.

Closes #1571.

### DIFF
--- a/redash/query_runner/mssql.py
+++ b/redash/query_runner/mssql.py
@@ -94,7 +94,7 @@ class SqlServer(BaseSQLQueryRunner):
     def _get_tables(self, schema):
         query = """
         SELECT table_schema, table_name, column_name
-        FROM information_schema.columns
+        FROM INFORMATION_SCHEMA.COLUMNS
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks.

PR #1600, opened 2017-02-14
Feature: User Deactivation

vishesh92:


migrations/versions/94450534392b_add_is_active_column_to_users.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '94450534392b'
+down_revision = '65fc9ede4746'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('users', sa.Column('is_active', sa.Boolean, default=True, index=False))
```
arikfr: This should be indexed.

redash/handlers/authentication.py
```diff
@@ -118,8 +118,9 @@ def login(org_slug=None):
             user = models.User.get_by_email_and_org(request.form['email'], org)
             if user and user.verify_password(request.form['password']):
                 remember = ('remember' in request.form)
-                login_user(user, remember=remember)
-                return redirect(next_path)
+                if login_user(user, remember=remember):
```
arikfr: 👍  We should do this everywhere else where we use `login_user` as well.

client/app/pages/users/show.html
```diff
@@ -59,6 +59,9 @@ <h3 class="p-l-5">{{user.name}}</h3>
           <input name="email" type="email" class="form-control" ng-model="user.email" required/>
           <error-messages input="userSettingsForm.email" form="userSettingsForm"></error-messages>
         </div>
+        <div class="form-group required">
+          <label class="control-label checkbox-inline"><input name="is_active" type="checkbox" ng-model="user.is_active" />Is Active</label>
+        </div>
```
arikfr: This should be shown only to admins.

Also maybe have it as a button on the main tab next to the password reset button?

arikfr: The button can be named "Disable" and make this a one way action for now (I doubt there will be many cases when a user gets disabled and then reinstated).

migrations/versions/94450534392b_add_is_active_column_to_users.py
vishesh92: I am not sure if it's a good idea to index Boolean fields.
http://stackoverflow.com/questions/12025094/performance-tuning-create-index-for-boolean-column

arikfr: Fair point 👍

client/app/pages/users/show.html
vishesh92: Someone might want to try out this feature. If we allow only disabling, then they won't be able to revert it back.

arikfr: I'm not sure we should worry much about this case, but it should be just as easy to show an "Enable" button in case of a disabled user.

yejianye:
Any updates on this PR? This would be a very useful admin feature.

gauravarora:
This is a useful PR. Can we get this merged please?

Issue #1602, opened 2017-02-15
Demo site not working for IE11


deecay:
Hi, the demo site loads with blank page via IE11.

SCRIPT438: Object doesn't support property or method 'assign'

We may need `es6-object-assign` or some other polyfill?

deecay:
Merged #1650

PR #1610, opened 2017-02-18
Bring back IE11 support

deecay:
This attempts to bring back IE11 support.

This PR adds polyfill es6-object-assign, which defines Mozilla foundation's polyfill definition of Object.assign() only if it is not defined.

Tested with IE11 and Chrome 56.0.2924.87 by pulling from my branch and then docker-compose.

I don't like the fact that I had to write polyfill() in pages/index.js, but this was the only place I could find. Please review.

[https://www.npmjs.com/package/es6-object-assign](https://www.npmjs.com/package/es6-object-assign)

### DIFF
--- a/client/app/pages/index.js
+++ b/client/app/pages/index.js
@@ -1,3 +1,7 @@
+import ObjectAssign from 'es6-object-assign';
+
+ObjectAssign.polyfill();
+
 export { default as home } from './home';
 export { default as queriesList } from './queries-list';
 export { default as alertsList } from './alerts-list';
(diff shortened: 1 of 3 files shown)


deecay:
And this closes #1602.

arikfr:
Let's use Babel's ability to handle this for us. As I don't have a way to test this, can you try the following: https://babeljs.io/docs/plugins/transform-object-assign/?

thanks.

deecay:
OK I will.

Issue #1327, opened 2016-10-07
[feature request] Hide customizing parameters after adding pivot table to dashboard


TonyStarkBy:
![pivot_table](https://cloud.githubusercontent.com/assets/16562082/19186624/d53ec05e-8c8f-11e6-8958-a23e2b2e4640.jpg)

zapatel:
+1

arikfr:
Closed by #1615.

PR #1615, opened 2017-02-20
Add: option to hide pivot table controls

deecay:
This PR adds a checkbox to hide control elements from pivottable visualization, allowing us to show only the table part in visualizations and dashboards.

Please see #1327 to get an idea of what this does.


client/app/visualizations/pivot/index.js
```diff
@@ -12,48 +15,65 @@ function pivotTableRenderer() {
...
 
-          if ($scope.visualization) {
-            Object.assign(options, $scope.visualization.options);
+            if ($scope.visualization) {
+              Object.assign(options, $scope.visualization.options);
+            }
+            $(element).pivotUI(data, options, true);
+            if ($scope.visualization.options.controls.enabled) {
+              const controls = $('.pvtAxisContainer, .pvtRenderer, .pvtVals');
+              for (let i = 0; i < controls.length; i += 1) { controls[i].style.display = 'none'; }
```
arikfr: I'm pretty sure that PivotTable has a built in capability to render only the results. Have you considered using it?

deecay: I have looked through the [documentation](https://github.com/nicolaskruchten/pivottable/wiki/Parameters) page before implementing, but couldn't find a way.

1. pivottable.js do have a method to display the table and table only, like [this](http://nicolas.kruchten.com/pivottable/examples/simple.html). But this is done by preparing calling pivot() instead of pivotUI().

2. If you prefer pivot() way, I will test if pivot() and pivotUI() can be rendered back and forth with the same `options` and `data`, to ensure that when someone tries to restore the pivotUI controls, everything is restored.

deecay: I have coded the following and it works, but found one issue.
When one chooses any Renderer other than Table, such as Heatmap, all the fancy coloring will be lost.
```
            if (options.controls && options.controls.enabled) {
              $(element).pivot(data, options, true);
            } else {
              $(element).pivotUI(data, options, true);
            }
```

So how shall we do this? display:none? pivot()? or some other way?

arikfr: Will be lost when he removes the controls? (the `.pivot()` call)

deecay: With .pivot() way, if a user creates pivot table with Heatmap renderer, and then hides the controls, the table will be rendered as a plain white pivot table. That's what .pivot() is. So it does more than just hiding the controls. There is no way to keep the table cell background colors.

My way is a little hackish, but retains the Heatmapped table.

arikfr: 👍 unless @nicolaskruchten will have a better suggestion will go with the workaround.

Thank you for the time and research!

arikfr:
Thanks!

Issue #1466, opened 2016-12-12
Add "Refresh Schema" button to the datasource


arikfr:
We already have an API to refresh to the schema, just need to expose UI for it.

PR #1617, opened 2017-02-20
Add "Refresh Schema" button to the datasource

44px:
Hello.

This PR closes https://github.com/getredash/redash/issues/1466
Adds "refresh schema" button to schema browser on query edit page:

image

This is my first PR for Redash interface, so I have some questions:
1. Is this a good place for "refresh schema" button?
2. Does it make sense to refactor directives into components (like this: https://github.com/getredash/redash/commit/b2e747caef0ac5a4f4f08f5dd7ef04b5c415f346)? In my experience components are easier to maintain and use because they have clearer API defined in bindings. Also lifecycle methods (`$onInit` and `$onChanges`) are quite useful.
3. Do you have some coding style guidelines for CSS? What do you think about BEM-like class names for component:
`.schema-browser` instead of `.schema-container` (CSS class name == component name)
`.schema-browser__control` instead of `.schema-control`
`.schema-browser__list` instead of `.schema-browser`

Thanks!


arikfr:
Nice! Thanks!

As for your questions:

> 1. Is this a good place for "refresh schema" button?

Well, it doesn't feel 100% right, but I don't have a better idea :)

> Does it make sense to refactor directives into components (like this: b2e747c)? 

Yes! I actually started refactoring them when I was working on the Angular1.5/ES6/webpack migration, but realized it will take too much time so decided to leave as is. But the goal is to rewrite every component/directive when we "touch" them.

> Do you have some coding style guidelines for CSS? 

Unfortunately not really. Just trying to be consistent with previous choices. :)

> What do you think about BEM-like class names for component:

The concept makes sense, but the double underscores separators look very ugly 😕 

>  .schema-browser instead of .schema-container (CSS class name == component name)

Regardless - this totally makes sense (class name == component name) and you're welcome to make this change.

> .schema-browser__control instead of .schema-control
> .schema-browser__list instead of .schema-browser

Maybe a middle ground will be `.schema-browser-control`? :)

I will try to read some more about BEM to form an opinion. If you have any good reference materials or well known project examples, it will be great.

client/app/pages/queries/schema-browser.html
```diff
@@ -1,13 +1,21 @@
 <div class="schema-container">
-  <div class="p-t-5 p-b-5">
```
arikfr: I'm curious about this change (`p-t-5 p-b-5` -> `schema-control`). It has obvious benefits, but it has the downside of making the CSS code larger. Any thoughts on this? When would you use a generic class vs. specific one? 

(this is just for discussion, no need to change anything)

client/app/pages/queries/view.js
```diff
@@ -43,26 +43,31 @@ function QueryViewCtrl($scope, Events, $route, $routeParams, $http, $location, $
...
-      if (data && data.length > 0) {
+  function toggleSchemaBrowser(hasSchema) {
+    $scope.hasSchema = hasSchema;
+    $scope.editorSize = hasSchema ? 'col-md-9' : 'col-md-12';
+  }
+
+  function getSchema(refresh = undefined) {
+    DataSource.getSchema({ id: $scope.query.data_source_id, refresh }, (data) => {
+      const hasSchema = data && (data.length > 0);
+      if (hasSchema) {
```
arikfr: It's possible now that because of some temporary error, an existing schema will disappear on refresh. Maybe we should check for the existence of previous schema, and in such case preserve it and show some error message via growl?

client/app/services/data-source.js
```diff
@@ -3,7 +3,7 @@ function DataSource($resource) {
     get: { method: 'GET', cache: false, isArray: false },
     query: { method: 'GET', cache: false, isArray: true },
     test: { method: 'POST', cache: false, isArray: false, url: 'api/data_sources/:id/test' },
-    getSchema: { method: 'GET', cache: true, isArray: true, url: 'api/data_sources/:id/schema' },
+    getSchema: { method: 'GET', cache: false, isArray: true, url: 'api/data_sources/:id/schema' },
```
arikfr: Maybe we can trigger non cached request when requesting a refresh only?

arikfr:
Ping :)

44px:
Hello!
I apologize for the delay with response.

> If you have any good reference materials or well known project examples, it will be great.

As for articles, I can recommend https://css-tricks.com/bem-101/
And maybe the most well known project is Material Design Lite (https://getmdl.io/components/#layout-section)

BEM is good for projects with large amount of custom styles. Since Redash is based on Bootstrap maybe BEM doesn't bring much benefit here. Maybe it will be more convenient to keep naming scheme same as in Bootstrap :)

client/app/pages/queries/schema-browser.html
44px: Well, we needed some specific style declarations here, so I replaced two general classes with one more specific. We got a bit larger CSS, but lesser HTML :) I thought having just `schema-control` vs `p-t-5 p-b-5 schema-control` or `p-t-5 p-b-5 flex-container` will be a bit cleaner.

One of the biggest problems of CSS is that you constantly need to name things and all these names live in global namespace. So I totally understand your desire to keep CSS as little as possible :)

But there is a way to have many specific classes and keep CSS maintainable:
For example, we have `some-component`. All its styles goes to `some-component.css`. And this file have classes prefixed with component name. So when we see `.some-component` class name we know where to find its styles and we can be sure that changes in this file wouldn't affect other components.

client/app/pages/queries/view.js
44px: Yes, it would be great!
Which HTTP code we'll get when error occurs? Will it be 5xx or just 200 with empty data?

client/app/services/data-source.js
44px: Yes, we can add separate `refreshSchema` method which will do two things:
1) Perform non cached request with `refresh=true` in url.
2) Remove cache for `getSchema`

Like here: http://stackoverflow.com/a/25124772

The main drawback here that we'll get extra request with next `getSchema` call.

Unfortunately, we don't have much control on how Angular stores values in cache (it uses full url as cache key: https://github.com/angular/angular.js/blob/v1.5.11/src/ng/http.js#L1251). It seems like there is no simple way to perform request on one url (with `refresh=true`), but renew cache data for another url.

client/app/pages/queries/view.js
arikfr: It will be 200 with empty data, hence the issue :(

client/app/services/data-source.js
arikfr: Got it. Let's keep as you did then and revisit if we fill the cache is really necessary.

client/app/pages/queries/view.js
44px: Thank you! Added check for old schema existence on refresh.

arikfr:
Thanks!

Issue #889, opened 2016-03-07
Alerting/visualizing on row counts rather than just field values


tsibley:
It'd be great to have alerts which could trigger when the number of rows a query returns crosses a certain threshold.  This would better support queries which return “suspicious” rows that need further investigation by a human.  Currently to accomplish that, you basically need two queries: one for actually grabbing the data (that the human can look at) and one for triggering the alert which just aggregates the first with `count()` (or does something clever like inspect redash's own metadata about the first query).  With two queries, there's lots of potential for divergence over time.  You can also hack it by including a window function in your select list (such as `row_number() over ()`) and alerting on that, but it's kinda ugly.

tsibley:
The ability to use the number of rows a query returns, instead of a field value from a row, would also be useful for the Counter visualization type.  It can be useful to show just the row count of a query on a dashboard, and disclose the query results upon clicking through.

gizmodude4:
I'd like to second this feature -- redash is almost perfect for our use case except we have non-SQL literate people using it who need to be able to visualize aggregates as well as values based on queries engineering writes.

deecay:
Closing. #1619

yoavbls:
Hey, this feature still isn't resolved,
the row counter is only in counter visualization but not in alerts.
Can you reopen it, please?

cjpit:
It would be great to have this on the query api as well.  For example I have redash integrated with our own internal reporting system.. The internal reporting app can schedule running of reports, and exporting to excel and attaching to a email.  If the result of the query is zero rows, I'd like to not send out the email, so that it can be used for exception reporting, when there is actual useful data available.

Thanks!

puttehi:
Would also be nice to have as a limiter for queries in general so users don't OOM the instance on some troublesome datasets with light data but a lot of rows

PR #1619, opened 2017-02-21
Add: use results row count as the value for the counter visualization.

deecay:
This PR is a response to #889

Now Counter visualization can show the row count of resultset.

When one sees a counter value in a dashboard, he often wants to obtain detailed information through table data. With this PR, we can now provide table data and its row counter with one query.

### DIFF
--- a/client/app/visualizations/counter/counter-editor.html
+++ b/client/app/visualizations/counter/counter-editor.html
@@ -2,13 +2,13 @@
   <div class="form-group">
     <label class="col-lg-6">Counter Value Column Name</label>
     <div class="col-lg-6">
-      <select ng-options="name for name in queryResult.getColumnNames()" ng-model="visualization.options.counterColName" class="form-control"></select>
+      <select ng-options="name for name in queryResult.getColumnNames()" ng-model="visualization.options.counterColName" class="form-control" ng-disabled="visualization.options.countRow"></select>
     </div>
   </div>
   <div class="form-group">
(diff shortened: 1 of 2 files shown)


arikfr:
Can you give a real world example where you might use this?

arikfr:
Also, why not just run another query that returns the count?

deecay:
Sure!

Two-query method has 2 downsides.

1. You have to find a way to guide people to the matching query with table result.
Lets say there is one dashboard that shows 3 counters. Number of unfulfilled backorders for today, this week, and last week. When one dashboard widget shows that there are "12 unfulfilled backorders last week", they often click on the query name, hoping to find out more details about those 12 backorders. When they click on "TABLE", they only see "12".
 I couldn't find a sensible way to construct a dashboard with many KPI figures, where people can easily drill down to each KPI's individual details. Issuer of #889 said that he may add a counter column to every row with row_cont(), but it's not the way to go; I agree. But with this PR, people can click on "TABLE" to see the details of the exact query the counter showed. 

2. Two queries means more maintenance and more computing resource. 
Two queries for the same data means twice the number of queries (of essentially the same data but only different in representation), twice the maintenance cost for queries (if you change the where clause of one query, you must update the corresponding query, and you *must not* forget to update it), twice the worker process for redash, and twice the session load for the data source (you would want to minimize this when the data source is not your own). All these extra costs can be avoided by just one checkbox.

Please let me know your thoughts.

arikfr:
Both are great use cases actually!

PR #1621, opened 2017-02-21
[DynamoDB] Fix: handle count(*) queries

kopanitsa:
This is pull request to fix below problem.

## Problem

- When I send a query as "SCAN Count(*) FROM Foobar;", an error message is shown (`Error running query: 'str' object has no attribute 'iteritems'`)

## Expected

- When I send above query, it returns counted result.
- Without "count" function ("SCAN * FROM Foobar;"), it works.
- According to Dynamo DB spec (DQL), count is defined.
https://dql.readthedocs.io/en/latest/topics/queries/select.html

## Route cause

- `count(*)` is not correctly handled in `dynamo_sql.py`

## Related issue

https://discuss.redash.io/t/dynamodb-count-query/385
This issue reports same issue, I guess.

### DIFF
--- a/redash/query_runner/dynamodb_sql.py
+++ b/redash/query_runner/dynamodb_sql.py
@@ -2,7 +2,6 @@ import json
 import logging
 import sys
 
-
 from redash.query_runner import *
 from redash.utils import JSONEncoder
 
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks!

PR #1623, opened 2017-02-21
Bugfix: race condition in query task status reporting

washort:
We encountered this traceback recently:

    AttributeError: 'exceptions.TypeError' object has no attribute 'get'
     File "flask/app.py", line 1639, in full_dispatch_request
       rv = self.dispatch_request()
     File "flask/app.py", line 1625, in dispatch_request
     File "flask_restful/__init__.py", line 477, in wrapper
       resp = resource(*args, **kwargs)
     File "flask_login/utils.py", line 228, in decorated_view
       return func(*args, **kwargs)
     File "flask/views.py", line 84, in view
       return self.dispatch_request(*args, **kwargs)
     File "redash/handlers/base.py", line 28, in dispatch_request
       return super(BaseResource, self).dispatch_request(*args, **kwargs)
     File "flask_restful/__init__.py", line 587, in dispatch_request
       resp = meth(*args, **kwargs)
     File "redash/handlers/query_results.py", line 233, in get
       return {'job': job.to_dict()}
     File "redash/tasks/queries.py", line 159, in to_dict
       updated_at = self._async_result.result.get('start_time', 0)

I believe this is a result of the task status changing from STARTED to FAILED after the initial status check. This patch changes it to fetch the status info only once.

### DIFF
--- a/redash/tasks/queries.py
+++ b/redash/tasks/queries.py
@@ -155,23 +155,25 @@ class QueryTask(object):
         return self._async_result.id
 
     def to_dict(self):
-        if self._async_result.status == 'STARTED':
-            updated_at = self._async_result.result.get('start_time', 0)
+        task_info = self._async_result._get_task_meta()
+        result, task_status = task_info['result'], task_info['status']
+        if task_status == 'STARTED':
+            updated_at = result.get('start_time', 0)
         else:
             updated_at = 0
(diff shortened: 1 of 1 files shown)


arikfr:
Nice! I was wondering where this is coming from :)

PR #1636, opened 2017-02-27
[Feature] add: the propertyOrder field to specify order in DataSource settings

rmakulov:
```json
            "properties": {
                "user": {
                    "type": "string"
                },
                "password": {
                    "type": "string"
                }
```

screen shot 2017-02-27 at 15 53 41

```json
            "properties": {
                "user": {
                    "type": "string",
                    "propertyOrder": 0
                },
                "password": {
                    "type": "string",
                    "propertyOrder": 1
                }
```

screen shot 2017-02-27 at 15 56 53

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -45,13 +45,6 @@ class InterruptException(Exception):
     pass
 
 
-def index(l, x):  # Allow non-full key list to be used in sorting
-    try:
-        return l.index(x)
-    except ValueError:
-        return sys.maxint, x
-
-
 class BaseQueryRunner(object):
     noop_query = None
 


arikfr:
Thanks! This is something that was on my mind for quite some time. But I wonder if we should use this approach and rely on the fact that OrderedDict gets serialized in order (and then the browser keeps this order)? This is how it works, but to the best of my knowledge this is not a guaranteed behavior. 

An alternative I was thinking of is a field that defines an order (something like `order: ["host", "username", "password", ...]`).

What do you think?

rmakulov:
I also thought about that. I prefer the alternative you mentioned above.
Fields that are not specified are added at the end.

```json
            "order": ['host', 'port', 'user', 'password'],
            "required": ["dbname"],
            "secret": ["password"]

```

screen shot 2017-02-27 at 15 56 53

arikfr:
But we're still depend on unguaranteed behavior of JSON. I was thinking of handling the order on the client side (using this extra property to sort the keys on the client side).

rmakulov:
What about such implementation?

redash/query_runner/__init__.py
```diff
@@ -43,6 +45,13 @@ class InterruptException(Exception):
     pass
 
 
+def index(l, x):  # Allow non-full key list to be used in sorting
+    try:
+        return l.index(x)
+    except ValueError:
+        return sys.maxint, x
```
arikfr: This is no longer needed, right?

rmakulov: Right :)

arikfr:
Thanks!

Apologies for the long wait.

Issue #1338, opened 2016-10-11
a bug


luotasha:
redash/redash/utils/**init**.py

def base_url(org):
    if settings.MULTI_ORG:
        return "https://{}/{}".format(settings.HOST, org.slug)

```
return settings.HOST
```

MULTI_ORG branch return url with 'http' while another branch return not. i can define settings.HOST with 'http://' prefix but it's ugly

PR #1637, opened 2017-02-27
uniform return values in base_url

r0fls:
Address #1338.

### DIFF
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -153,8 +153,8 @@ def collect_parameters_from_request(args):
 
 def base_url(org):
     if settings.MULTI_ORG:
-        return "https://{}/{}".format(settings.HOST, org.slug)
+        return "http://{}/{}".format(settings.HOST, org.slug)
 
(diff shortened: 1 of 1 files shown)


r0fls:
Is this not running the circle ci tests?

redash/utils/__init__.py
```diff
@@ -155,6 +155,6 @@ def base_url(org):
     if settings.MULTI_ORG:
         return "https://{}/{}".format(settings.HOST, org.slug)
 
-    return settings.HOST
+    return "https://{}".format(settings.HOST)
```
arikfr: Unfortunately not everyone is using `https`... I think uniforming should be in the opposite direction, and make the MULTI_ORG branch of the if not use https by default.

r0fls:
@arikfr updated

arikfr:
You kind of missing the point: some will have it as `http` others as `https`. So either way this will break it for someone. Also anyone who has this set already will have the protocol set there already so this change will break it for them.

I don't see why it's important to change this line. Let's leave it as is...

Thanks.

PR #1643, opened 2017-03-01
[Data Sources] Add: SalesForce query runner

msnider:
Adds a Salesforce query runner for the SOQL (Salesforce Object Query Language) API (for those of us who have to interact with Salesforce!).

As written it does require a new dependency on [simple_salesforce](https://github.com/simple-salesforce/simple-salesforce).

### DIFF
--- a/redash/query_runner/salesforce.py
+++ b/redash/query_runner/salesforce.py
@@ -1,0 +1,177 @@
+# -*- coding: utf-8 -*-
+
+import re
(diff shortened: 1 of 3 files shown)


arikfr:
Good timing! Just a week or so ago I learned about SOQL and thought, "we should have support for this". :)

msnider:
Awesome! Thanks!

PR #1649, opened 2017-03-02
Fix: datetime parameters were not using a date picker.

arikfr:


client/app/services/query.js
```diff
@@ -43,6 +43,40 @@ class QueryResultError {
...
+    }
+
+    return this.value;
+  }
+
+  set ngModel(value) {
+    if (this.type === 'date') {
+      this.value = moment(value).format('YYYY-MM-DD');
+      this.$$value = moment(this.value).toDate();
+    } else if (this.type === 'datetime-local' || this.type === 'datetime-with-seconds') {
```
rockwotj: Why is there two versions of this if they are handled the same? If a deep link has seconds in the query parameters,then the datetime-local type will show seconds in the datepicker.

client/app/services/query.js
```diff
@@ -43,6 +43,40 @@ class QueryResultError {
...
+      this.$$value = this.$$value || parseInt(this.value, 10);
+      return this.$$value;
+    }
+
+    return this.value;
+  }
+
+  set ngModel(value) {
+    if (this.type === 'date') {
+      this.value = moment(value).format('YYYY-MM-DD');
```
rockwotj: I dislike having this hardcoded, why not use the environment variable?

client/app/services/query.js
arikfr: Wasn't aware of this - fixed.

client/app/services/query.js
arikfr: For now this maintains backward compatibility with v0.12.0 and fixes a regression. We can revisit this and add option to set the format in a future pull request.

But I don't think it should be in an env variable, but rather as configuration for the parameter, as users might use different databases on the same Redash deployment.

client/app/services/query.js
```diff
@@ -43,6 +43,43 @@ class QueryResultError {
...
+
+    return this.value;
+  }
+
+  set ngModel(value) {
+    if (value && this.type === 'date') {
+      this.value = moment(value).format('YYYY-MM-DD');
+      this.$$value = moment(this.value).toDate();
+    } else if (value && this.type === 'datetime-local') {
+      this.value = moment(value).format('YYYY-MM-DDTHH:mm');
```
rockwotj: BigQuery sadly doesn't support this format (it doesn't like the `T`) so we won't be able to use this (we've been using the fix in #1553) until it's configurable.

arikfr: What format BQ supports?

rockwotj: `YYYY-MM-DD HH:mm:ss`

arikfr: I changed it to be the current format :)

rockwotj:
Whoop! Thanks Arik :)

rockwotj:
Thanks Arik :)

arikfr:
Sure thing. Thank you for the feedback and help here!

PR #1650, opened 2017-03-03
Add babel-plugin-transform-object-assign for IE support

deecay:
This PR adds Babel plugin transform-object-assign.

This is necessary for v1 to work with IE11 (and probably other versions of some older IEs).

This is alternate version of #1610.


arikfr:
Thanks!

PR #1651, opened 2017-03-03
Fix: set default values in options to enable 'default: True' for checkbox

rmakulov:
The configuration schema includes a default value field, but unfortunately does not work for checkboxes because of placeholder usage. We suggest setting default values for all inputs directly on the datasource initial page via options, which can be saved without typing for convenience.


arikfr:
Thanks!

PR #1653, opened 2017-03-04
Fix: query embed dialog close button wasn't working

r0fls:
The `x` button on the query embed modal was not working. This fixes it.

screen shot 2017-03-04 at 2 52 09 pm


arikfr:
Thanks!

PR #1654, opened 2017-03-05
Added RPM specfiles for building Redash

kagesenshi:
Built and tested in Fedora25

installs redash in /opt/redash/
uses config file for configuring redash /etc/redash.cfg
includes a runner script for executing manage.py and daemons using the configured environment variables /usr/bin/redash 
installs supervisor config in /etc/supervisord.d/
logs in /var/log/redash/

### DIFF
--- a/setup/fedora/redash-runner.py
+++ b/setup/fedora/redash-runner.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+
+from ConfigParser import ConfigParser
+import os
+import argparse
+import sys
+
+LIST_OPTS = [
+    'enabled_query_runners',
+    'enabled_destinations'
+]
+
+def main():
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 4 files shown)


arikfr:
Sorry for letting it dry for so long and thank you for your effort on this ! This is both interesting and educating on how RPM packages are crafted.

As this is not something I can commit to maintain, how about we open a separate repository for it under the `getredash` organization on GitHub and you will maintain it?

Btw, why did you use supervisor and not systemd to manage the processes?

kagesenshi:
Hi,

That works for me i guess. Just a repo for the RPM spec. I'm thinking of bringing this to Fedora copr too. 

Using supervisord simply because i'm familiar with it coming from Zope/Plone stack. Haven't actually studied SystemD configurations.

arikfr:
I've created https://github.com/getredash/fedora_rpm for this and added you as an admin. Let me know if you didn't receive the invite.

Thanks!

PR #1656, opened 2017-03-07
Fix Docker file ownership issues:

arikfr:
1. Simplify user creation to create a non system user (so the uid will usually
   match the host user).
2. Set the user to redash & remove the need to change user in docker entrypoint.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,5 +8,6 @@ RUN pip install -r requirements.txt -r requirements_dev.txt -r requirements_all_
 COPY . ./
 RUN npm install && npm run build && rm -rf node_modules
 RUN chown -R redash /app
+USER redash
 
 ENTRYPOINT ["/app/bin/docker-entrypoint"]
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -1,19 +1,12 @@
 #!/bin/bash
 set -e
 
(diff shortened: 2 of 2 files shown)


44px:
@arikfr, can these changes broke redash_worker container?

After pulling latest changes I noticed that queries stuck in queue.

`docker-compose ps` shows this:
```
redash_worker_1     /app/bin/docker-entrypoint ...   Exit 1
```

`redash_worker_1` shuts down soon after start (I checked multiple times)

`docker logs redash_worker_1` ends with:
```
worker: Warm shutdown (MainProcess)
[2017-03-16 21:04:13,956: INFO/MainProcess] beat: Shutting down...
Starting scheduler and 2 workers for queues: queries,scheduled_queries,celery...
Running a worker with superuser privileges when the
worker accepts messages serialized with pickle is a very bad idea!

If you really want to continue then you have to set the C_FORCE_ROOT
environment variable (but please think about this before you do).

User information: uid=0 euid=0 gid=0 egid=0
```

I restored `sudo -E -u redash` in `bin/docker-entrypoint` and queries queue works as before now.

arikfr:
You need to rebuild the container (`docker-compose rm server` / `docker-compose rm worker` and then `create` or `up`).

44px:
Thanks, it works!
A bit strange - I tried `docker-compose rm` and then `docker-compose up` previously, but it didn't help.

PR #1657, opened 2017-03-08
Show vertical scrollbar for IE

deecay:
Vertical scrollbar for HTML shown in Chrome and Firefox, but not IE.

### DIFF
--- a/client/app/assets/css/superflat_redash.css
+++ b/client/app/assets/css/superflat_redash.css
@@ -6430,7 +6430,7 @@ a {
 }
 html {
   overflow-x: hidden\0/;
-  -ms-overflow-style: none;
+  -ms-overflow-style: scroll;
 }
 html,
 body {
--- a/client/app/assets/css/superflat_redash.css
+++ b/client/app/assets/css/superflat_redash.css
@@ -6430,7 +6430,7 @@ a {
 }
 html {
   overflow-x: hidden\0/;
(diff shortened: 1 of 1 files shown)


arikfr:
Can you add screenshots of before/after please?

deecay:
Current:
now


Scrollbar is missing from from every page.


Fixed:
patched1


* If `-ms-overflow-style` set to 'scroll', the bar appears anytime mouse is in the window. Disappears when mouse is out. 
* If set to 'auto', the bar appears when mouse is in the window and near the vertical scrollbar. Disappears when mouse moves away from the bar.

This PR adopts 'scroll'. But any value but 'none' is fine for me.

arikfr:
The behavior I see in Chrome on OSX is that the scrollbar appears when you start scrolling. Is this the same behavior with "auto"?

deecay:
With 'auto':
 1. Moving mouse around will show wide scrollbar. Staying still for a while will auto hide the wide scrollbar.
 2. Pressing down key and scrolling down while wide scrollbar is shown, wide scrollbar is shown.
 3. Pressing down key and scrolling down while wide scrollbar is NOT shown, narrow scrollbar is shown.

out

arikfr:
Let's use `auto` then. Thanks!

deecay:
Done.

arikfr:
Cool, thanks!

PR #1662, opened 2017-03-09
Add: Trimmed ; from query as it causes trouble when using pyhive

benmagro:
When using the pyhive connector, queries fail if they have a colon at the end. This can be pretty annoying.

Not sure if this is the best way to fix the problem. Happy to take feedback on the PR.

Thanks


arikfr:
@benmagro Ping?

benmagro:
On it. I'll move. Just been flat out :-)

arikfr:
Please reopen when updated. Thanks.

PR #1665, opened 2017-03-10
Fix: Set query filter to match dashboard level filters

benmagro:
Added:
When dashboard level parameters are present, we should set to the filter strings to match.

Removed:
Existing code that doesn't seem to work.

Video:
filter_params

### DIFF
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -84,8 +84,7 @@ function DashboardCtrl($rootScope, $routeParams, $location, $timeout, $q, $uibMo
             return;
           }
 
-          if ((hasQueryStringValue || dashboard.dashboard_filters_enabled)) {
-            // Sets query filters to match dashboard fiters when present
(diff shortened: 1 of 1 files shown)


MichaelJAndy:
This could make a big difference in de-duplicating our reports, +1

client/app/pages/dashboards/dashboard.js
```diff
@@ -84,13 +84,15 @@ function DashboardCtrl($rootScope, $routeParams, $location, $timeout, $q, $uibMo
             return;
           }
 
+          if ((hasQueryStringValue || dashboard.dashboard_filters_enabled)) {
+            // Sets query filters to match dashboard fiters when present
+            queryFilter.current = $location.search()[queryFilter.name];
```
arikfr: But if `hasQueryStringValue==false` then it will set the filter to empty value?

arikfr:
I applied the fix the condition myself. Thank you for the pull request 👍

Issue #216, opened 2014-05-14
Time series in pivot tables appear to be ordered by first letter of month - rather than by date


nickw-hailo:
When dropping a "week" time series to get weekly metrics into the re:dash pivot table, re:dash appears to order based on first letter of the month instead of date/time eg. April, February, January, March, May). The query orders by time correctly, however.

arikfr:
I don't we think should show ISO format in the UI, but rather just send it this way (or whatever format it needs) to the pivottable library.

matogertel:
Having the same issue, but with weekdays instead of months. I agree with arikfr on this, send it in ISO to the pivottable library, and maybe add a field to the UI to customize the formatting. (with some sensible default)

kimiyuki:
is it possible to apply redash_date_format for the pivot UI? I don't understand technical details for it, but it would be great to fix the issue.

olilogic77:
Just trying out pivot tables and seeing the same issue - Friday, Monday, Saturday, Sunday ....

![screen shot 2017-02-02 at 2 30 38 pm](https://cloud.githubusercontent.com/assets/5517304/22537206/351ce520-e954-11e6-9bdb-7f322beb066a.png)

dobesv:
I ran into this and tried using `DATE_FORMAT()` to change the format of the date - but redash parsed the date and converted it into that format and sorted everything wrong anyway.  I had to modify the date format to have a leading space to prevent it from being parsed and reformatted to the undesirable format starting with a week day.

dudanogueira:
Same here. even tried YYYYMM format and always unsorted. It would be nice to be possible, for datetime pivot dates, to allow group an drill down by week, month, quarter and years.

dudanogueira:
Obs: instalaled from the docker here: https://github.com/getredash/redash/blob/master/setup/docker-compose.yml

amirashakernady:
Same here.it would be better if it solved

PR #1666, opened 2017-03-10
Change: add support for filtered data in Pivot table visualization

deecay:
This PR changes how the `data` fed to pivottable visualization is prepared, and makes pivottable show filtered and formatted data just as you can see in Table visualization.

I have noticed that pivottable visualization:
* does not incorporate query/dashboard filter
* `::filter` annotation appears as part of pivottable attribute name and does not get removed
* date and datetime columns does not respect `REDASH_DATE_FORMAT`

This PR addresses all three points.

### DIFF
--- a/client/app/visualizations/pivot/index.js
+++ b/client/app/visualizations/pivot/index.js
@@ -21,7 +21,7 @@ function pivotTableRenderer() {
         if ($scope.queryResult.getData() !== null) {
           // We need to give the pivot table its own copy of the data, because it changes
           // it which interferes with other visualizations.
-          data = angular.copy($scope.queryResult.getRawData());
+          data = angular.copy($scope.queryResult.getData());
           const options = {
             renderers: $.pivotUtilities.renderers,
             onRefresh(config) {
--- a/client/app/visualizations/pivot/index.js
+++ b/client/app/visualizations/pivot/index.js
@@ -1,10 +1,9 @@
+import angular from 'angular';
 import $ from 'jquery';
 import 'pivottable';
 import 'pivottable/dist/pivot.css';
-import { formatValue } from '../table';
-import { getColumnCleanName } from '../../services/query-result';
 
-function pivotTableRenderer(clientConfig, $filter) {
+function pivotTableRenderer() {
   return {
     restrict: 'E',
     scope: {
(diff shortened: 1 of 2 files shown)


client/app/visualizations/pivot/index.js
```diff
@@ -21,7 +22,19 @@ function pivotTableRenderer() {
...
           // it which interferes with other visualizations.
-          data = angular.copy($scope.queryResult.getRawData());
+          data = [];
+          const columns = $scope.queryResult.getColumns();
+          const rows = $scope.queryResult.getData();
+          rows.forEach((row) => {
+            const rowObj = {};
+            columns.forEach((col) => {
+              const colName = getColumnCleanName(col.name);
+              const value = formatValue($filter, clientConfig, row[col.name], col.type);
```
arikfr: Two things that come to mind here:

1. Are we sure that converting numbers into strings will behave properly with PivotTable?
2. For dates, shouldn't we use some standard like (YYYY-MM-DD) regardless of configuration to solve #216?

deecay: - Comment 2

  I am thinking of adding a Sorting option to pivottable editor. Because passing a sort-friendly format will work for numerical dates, but it doesn't help for Monday/Tuesday/..., Mon/Tue/..., January/February/..., Jan/Feb/..., etc.
 If we pass `REDAH_DATE_FORMAT` to pivot, the problem below will go away because the dates won't be in ISO format,



and letting people define custom sorters will take care of the first comment of #216.

- Comment 1
I have to look into it a bit more for part 1 of your comment.

arikfr: They have a way to pass a sort function?

Anyway, how about we change this pull request to only handle the filters part and leave the formatting to another pull request?

deecay: Yes they do. Example is [here](http://nicolas.kruchten.com/pivottable/examples/mps_agg.html).
I have to come up with how to let people define it in pivottable editor, but currently have no idea.

I'm perfectly fine with dividing this PR into two, and making this only about the filters.
I assume what you mean is just replacing `.getRawData()` to `.getData()`?

deecay:
This PR has been changed and now it is about fixing pivottable of filtered data.

Formatting part will be addressed in a separate PR.

client/app/visualizations/pivot/index.js
arikfr: > I have to come up with how to let people define it in pivottable editor, but currently have no idea.

The user shouldn't have to define them. We should have default ones for the types we have (dates, numbers, etc - where needed).

PR #1673, opened 2017-03-16
Add: .editorconfig to keep code style consistent

44px:
[EditorConfig](http://editorconfig.org/) helps developers define and maintain consistent coding styles between different editors and IDEs.


arikfr:
Thanks 👍

Issue #1616, opened 2017-02-20
Problem with clickhouse connector


akonyaev90:
Hi!

When I select from clickhouse columns with type Array(String) - that looks good,
but when I select Array(IntX)  - I see empty cells.
I clickhouse-console Array(IntX) columns looks good.

Any idea?

arikfr:
Fixed in #1682.

PR #1682, opened 2017-03-22
[Clickhouse] Fix: better support for types

denisov-vlad:
#1616 Fixed issue with wrong type definition for array of integers


arikfr:
Thanks.

PR #1686, opened 2017-03-25
[Salesforce] Change: Sandbox cannot be required or it will force it to be True. 

msnider:
Very minor fixes for the Salesforce query runner:

1) If Sandbox is a required parameter then Redash will require it be True

2) SOQL queries should not be annotated since SOQL does not allow comments


arikfr:
Thanks!

PR #1687, opened 2017-03-26
[Data Source] Adds a Google Spreadsheet OAuth User

akiray03:
Adds a Google Spreadsheet DataSource connecting by OAuth User (NOT Service Account).

## Motivation

Some G Suite users, Restrict sharing of files outside the organization. They cannot grant permissions to users in different domains. The service accounts domain is different G Suite domains, it could not be accessed.
https://support.google.com/a/answer/60781?hl=en

## Screen Capture

redash-datasource-google-spreadsheet-oauth-user


denisov-vlad:
I think it will be good to add this option for all Google data sources used in redash. But the way with **new** datasource with different login type looks bad for me.

akiray03:
@denisov-vlad  Thanks. My idea is as follows. What do you think?

1. adds GoogleOAuth model.
1. adds google_oauth_id columns (nullable) to data_sources.
1. For a data source that uses Google authentication, let's choose the service account and Google OAuth.

image

denisov-vlad:
@akiray03 that's a good idea but i can't imagine how to make it in redash user interface.
It's possible to add google_oauth_credetials to redash user's profile. Then you can add datasource for Spreadsheets/Analytics/etc using service account or your redash user's oauth credentials. It looks better for me.

denisov-vlad:
maybe @arikfr knows a better solution

akiray03:
@denisov-vlad  Thanks. sounds good.
The idea I suggested for the first time was that we do not need to give permission to Redash's end users. But in many cases it can be realized with a service account.
I think that I will implement it with the idea you presented.

1. adds UserGoogleCredential model. (It has user_id, org_id, scopes, and credentials.)
2. adds the Service Account and the executor account selection settings to DataSource
3. adds account selection function to QueryRunner
4. add a mechanism to request authentication 
5. and other ..

akiray03:
@arikfr What do you think with this approach?

arikfr:
Apologies for leaving this untouched for so long... :-(

I have a few comments about the code, but my main concern is more a conceptual one: data sources in Redash are shared among users. Does it really make sense for a user to give access to all his spreadsheets in such environment?

It does make sense to have a flow where the user can easily import a spreadsheet into Redash but on a per spreadsheet basis (maybe using something like the [Picker API](https://developers.google.com/picker/docs/)?); without the need to create a special data source first. It does require changing some concepts in Redash a bit. :\

What was your use case for this?

--- 

As for the other discussion you had here, due to the need to authorize specific scopes when adding different data sources, I'm not sure what we gain by storing OAuth credentials in a central store (while adding complexity to the implementation).

---

Apologies again for the long period of silence. Will try to be more responsive now.

arikfr:
Hi,

Thanks.

PR #1688, opened 2017-03-26
[Refactoring] make_{csv,excel}_content move to models from handlers

akiray03:
I want to be able to use like: `python manage.py queries export <query_id>`
I think that QueryResultResource should not have data structure conversion processing.

### DIFF
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -1,13 +1,10 @@
-import csv
 import json
-import cStringIO
 import time
 
 import pystache
 from flask import make_response, request
 from flask_login import current_user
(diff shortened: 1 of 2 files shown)


arikfr:
Moving this logic out of the controllers makes sense. I might move it around later a bit more (to serializers).

About the intended usage -- why use the CLI instead of calling the API? The API for getting query results is simple enough to be almost as comfortable as a CLI, while preserving all the logic of keeping record of query usage, permissions, etc.

akiray03:
Oops. Thanks to your remarks, I thought it would be better to realize with a combination of API.

PR #1690, opened 2017-03-27
Fix: SQLite utf-8 error messages

deecay:
### Issue
SQLite does not show error message when error message contain unicode character.

With SQL like

`select 名称 from user`,

 we get

`raised unexpected: UnicodeDecodeError('ascii', 'no such column: \xe5\x90\x8d\xe7\xa7\xb0', 16, 17, 'ordinal not in range(128)')`

in log, and nothing on screen.

### Fix
Since the error instance is immutable, 

`sqlite3.OperationalError('no such column: \xe5\x90\x8d\xe7\xa7\xb0',)`

I had to create a new instance with utf-8 decoded arguments of the original error instance to get 

`sqlite3.OperationalError('no such column: \u540d\u79f0',)`.

Then we can see 'no such column: 名称' as on screen error.

### DIFF
--- a/redash/query_runner/sqlite.py
+++ b/redash/query_runner/sqlite.py
@@ -83,7 +83,11 @@ class Sqlite(BaseSQLQueryRunner):
             error = "Query cancelled by user."
             json_data = None
         except Exception as e:
-            raise sys.exc_info()[1], None, sys.exc_info()[2]
+            # handle unicode error message
+            err_class = sys.exc_info()[1].__class__
+            err_args = [arg.decode('utf-8') for arg in sys.exc_info()[1].args]
+            unicode_err = err_class(*err_args)
+            raise unicode_err, None, sys.exc_info()[2]
         finally:
             connection.close()
         return json_data, error


redash/query_runner/sqlite.py
```diff
@@ -83,7 +83,11 @@ def run_query(self, query, user):
             error = "Query cancelled by user."
             json_data = None
         except Exception as e:
-            raise sys.exc_info()[1], None, sys.exc_info()[2]
+            # handle unicode error message
+            err_class = sys.exc_info()[1].__class__
+            err_args = [arg.decode('utf-8') for arg in sys.exc_info()[1].args]
+            unicode_err = err_class(*err_args)
+            raise unicode_err, None, sys.exc_info()[2]
```
arikfr: What if we remove this whole `except Exception` block, will it resolve the issue?

deecay: No it doesn't. The error will then go straight up to line 423 of `redash/tasks/queries.py`, and encounter an error at `unicode(e)`. Without some kind of decoding in sqlite.py, it will not pass through `unicode(e)`.

```
worker_1    | [2017-03-26 19:33:13,394: ERROR/MainProcess] Task redash.tasks.execute_query[a53160d4-b95e-4be5-a41c-f854408a4440] raised unexpected: UnicodeDecodeError('ascii', 'no such column: \xe5\x90\x8d\xe7\xa7\xb0', 16, 17, 'ordinal not in range(128)')
worker_1    | Traceback (most recent call last):
worker_1    |   File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
worker_1    |     R = retval = fun(*args, **kwargs)
worker_1    |   File "/app/redash/worker.py", line 68, in __call__
worker_1    |     return TaskBase.__call__(self, *args, **kwargs)
worker_1    |   File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
worker_1    |     return self.run(*args, **kwargs)
worker_1    |   File "/app/redash/tasks/queries.py", line 500, in execute_query
worker_1    |     scheduled_query).run()
worker_1    |   File "/app/redash/tasks/queries.py", line 427, in run
worker_1    |     error = unicode(e)
worker_1    | UnicodeDecodeError: 'ascii' codec can't decode byte 0xe5 in position 16: ordinal not in range(128)
```

Changing `unicode(e)` to `unicode(e, 'utf-8')` at queries.py results in on-screen error of
`coercing to Unicode: need string or buffer, OperationalError found`.

arikfr:
Thanks!

PR #1695, opened 2017-03-29
Fixes Issue: multiple users with the same email

### LINKED ISSUE #1694: Multiple Redash users with the same email-id
- Currently in the Redash code, It is possible to create multiple users with the same email-id, because email is not case insensitive. 
- For example It is possible to create two users like InFo@reDaSh.io and INFO@REDASH.IO, Now when we try to login through Google Auth, Redash gets email id as info@redash.io which matches with none of the emails ( InFo@reDaSh.io and INFO@REDASH.IO ). This issue has been the source of several login failures at our company.

ShahSunny:
With current code, It is possible for the Admin to create two redash users with the same email-id, for example two users can be, REDASH@REDASH.IO and reDash@redash.io. 
Fixes #1694

### DIFF
--- a/redash/handlers/users.py
+++ b/redash/handlers/users.py
@@ -27,7 +27,7 @@ class UserListResource(BaseResource):
     def post(self):
         req = request.get_json(force=True)
         require_fields(req, ('name', 'email'))
-
+        req['email'] = req['email'].lower()
         user = models.User(org=self.current_org,
                            name=req['name'],
                            email=req['email'],


redash/handlers/users.py
```diff
@@ -27,7 +27,7 @@ def get(self):
     def post(self):
         req = request.get_json(force=True)
         require_fields(req, ('name', 'email'))
-
+        req['email'] = req['email'].lower()
```
arikfr: Shouldn't we transform email to lower case everywhere where we receive it as input?

casbeebc: I agree. Perhaps updating the [User](https://github.com/getredash/redash/blob/master/redash/models.py#L318) model and listen for a set event on the email field, would be better. Then just make sure it is lowercase: 
```
@listens_for(User.email, 'set')
def make_email_lowercase(target, val, oldval, initiator):
    target.email = val.lower()
```

_Note: Haven't tested the code._

ShahSunny: @arikfr I felt one central change would be better than changing it everywhere.

wichert:
Why not use the CIText type for the email column? That makes all searches case insensitive, will still preserving original case. More importantly it removes the need to do a `.lower()` throughout your codebase.

redash/handlers/users.py
arikfr: @ShahSunny I agree about one central change, but this one you did is not the central one.

@wichert's suggestion to use a case insensitive type for this column makes sense, although need to think how to deal with the migration - considering that some users might have the same email more than once already.

ShahSunny:
@arikfr Is this resolved in the new release?

arikfr:
Yes, it was implemented in #1985. Thank you for your effort here :+1:

PR #1696, opened 2017-03-29
Fix: wrong percent stacking math

spasovski:
Fixes issue as described in https://github.com/mozilla/redash/issues/41


arikfr:
Thanks!

Issue #1697, opened 2017-03-30
Queries list page is slow when some queries has large query result


shimpeko:
### Issue Summary

Queries listing page (`/queries`) is slow when some queries has large query result.
Because it execute admin queries which select all columns from query_results for each queries on the list.

### Steps to Reproduce

1. Create query which will return large result_set
2. Execute it
3. Access `/queries?page=1`

### Technical details:

* Redash Version: 1.0.0-rc2
* Browser/OS: Chrome 56
* How did you install Redash: docker-compose (used default docker-compose.yml for testing)

arikfr:
Thanks! This also uncovers a nasty N+1 queries issue there.

PR #1698, opened 2017-03-30
Make queries list page faster

shimpeko:
For https://github.com/getredash/redash/issues/1697

This PR is to reduce number of admin SQL and amount of data send from admin DB to App server when `/api/queries` or `/api/queries/my` is requested.

There are two reasons which make those two API slow.

- Executing extra admin SQL for each queries to be responded(listed) (N+1)
- Retrieving non-required data for query listing from DB to App server ( large query result data )

This PR solves both.

Changes on `Query#all_queries` is the main part. Before the change, it only retrieves columns of `queries` table. `query_results` and `users` are retrieved on succeeding process (N+1 part).
And also all columns are retrieved from admin DB to App server which will be slow with large query result data.

After the change, it retrieved minimal data for query listing with 1 SQL.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -807,14 +807,14 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
     @classmethod
     def all_queries(cls, group_ids, user_id=None, drafts=False):
         select_columns = [
-            Query.id
-            , Query.name
-            , Query.created_at
-            , Query.schedule
-            , Query.is_draft
-            , User.name.label("created_by")
-            , QueryResult.retrieved_at
-            , QueryResult.runtime
+            Query.id,
+            Query.name,
+            Query.created_at,
+            Query.schedule,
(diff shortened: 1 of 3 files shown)


redash/models.py
```diff
@@ -806,7 +806,18 @@ def create(cls, **kwargs):
...
+            Query.name,
+            Query.created_at,
+            Query.schedule,
+            Query.is_draft,
+            User.name.label("created_by"),
+            QueryResult.retrieved_at,
+            QueryResult.runtime
+            ]
+        q = (db.Query(select_columns, session=SignallingSession(db))
+            .join(User, Query.user_id == User.id)
```
arikfr: SQLAlchemy has a way to eagerly load associated models, see the implementation in my [pull request](https://github.com/getredash/redash/pull/1703/files#diff-2eb9bad8b59774e58b2c9a664440905aR811).

arikfr:
Thanks, but closing for #1703.

PR #1699, opened 2017-03-30
DB2 Support for Redash 

shrirenjith:
Added DB2 Support for re-dash.  Implemented the DB2 Query runner . Tested this on DB2-LUW v 11 . 
Boolean  Types and other esoteric types (CLOB,BLOB) etc are not supported . Will implement them in future . Also plan to add DB2 for Mainframe (z/OS) soon . 

Note
This required additional libxml dependencies ( Will have to add RUN apt-get -y build-dep python-lxml ) to the DockerFile  . 

(Ignore the Docker Files Changes, I had  to do change it to this running on windows docker) .

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,4 @@ RUN npm install && npm run build && rm -rf node_modules
 RUN chown -R redash /app
 USER redash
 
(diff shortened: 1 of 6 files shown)


shrirenjith:
Closing this . Have some more changes and a cleaner version

PR #1700, opened 2017-03-31
Change: upgrade Plotly.js.

deecay:
Update plotly from 1.21.2 to 1.25.0

Version 1.25.0 of plotly is especially attractive because we can now double-click on one series of a legend to "single-out" that series in the plot. This feature is very convenient for analyzing data.

[plotly Changelog](https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md)

I have tried creating all chart types with 1.25.0 and it seems okay to me. However, I haven't done any thorough testing.


arikfr:
While this was waiting they had a few more releases... let's bump to `1.26.1`?

deecay:
Sure. Let's do that.

arikfr:
Merged. Thanks :)

Issue #1370, opened 2016-10-31
Feature Request: Sorting on Multiple Columns


r-kells:
Feature Request: Provide ability to sort on multiple columns. This makes it much easier to gain insight from tables with multiple numeric fields. 

Example here using datatables.js (shift click to multi sort)
http://codepen.io/anon/pen/NRbkzN

tmozgach:
Hello, just want to be sure, do you need this feature in the following part of system, for example?
![2](https://cloud.githubusercontent.com/assets/26471428/24088187/c4d52e8a-0ce4-11e7-90e3-e533ad7c3dae.png)

arikfr:
@tmozgach I think the original issue was about sorting query results.

tmozgach:
Hello, I see. One question more, I am new with this system, if it is possible could you navigate me and tell which file is a major one and I can add js code there, which will influence on all table in the project? Thanks!

arikfr:
Implemented in v4.

PR #1712, opened 2017-04-07
add datatables library

KinGoverm:
add datatables jquery library for sorting tables.


tmozgach:
It is for Feature Request: Sorting on Multiple Columns #1370.

arikfr:
Thank you for your effort here, but I can't merge this as is:
1. You need to adapt this to use Angular rather than trying to hook on any table with jQuery events.
2. We manage dependencies with npm and not including direct script tags.
3. `styles.c93891a69f22f1523673.css` is from a specific build. This reference changes dynamically on every build and handled by the build system.

Take a look at #1723 as a possible solution for your needs.

PR #1713, opened 2017-04-07
Change: Box plot library from d3.js to Plotly.js

deecay:
This PR will change the box plotting library from using d3 directly, to Plotly.

out-box

The whiskers correspond to the box' edges +/- 1.5 times the interquartile range, a default behavior of plotly box plot.

You must uncheck sorting of the X-Axis values to draw meaningful boxes.

### Remaining issues
* What should I do with existing box plot menus and source files? Should I keep them for backwards compatibility?

### DIFF
--- a/client/app/visualizations/box-plot/index.js
+++ b/client/app/visualizations/box-plot/index.js
@@ -176,7 +176,7 @@ export default function (ngModule) {
 
     VisualizationProvider.registerVisualization({
       type: 'BOXPLOT',
-      name: 'Boxplot',
+      name: 'Boxplot (Deprecated)',
       renderTemplate,
       editorTemplate: editTemplate,
     });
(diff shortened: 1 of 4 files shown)


deecay:
Done.

arikfr:
Super, thanks! 👍  Can you please send me the results of the demo query you used in the gif so I can add it to the demo instance?

deecay:
Here is the python query used in the demo. Or did you want some specific png's?

```
from random import random

result = {}
categories = ["A", "B", "C", "D"]
subcategories = ["W", "X", "Y", "Z"]

for category in categories:
    for subcategory in subcategories:
        for j in range(10):
            add_result_row(result, {"Category" : category, "Subcategory" : subcategory, "Value" : random()})

add_result_column(result, "Category", "", "string")
add_result_column(result, "Subcategory", "", "string")
add_result_column(result, "Value", "", "float")
```

PR #1716, opened 2017-04-11
Fix: Non-ASCII dashboard tag

deecay:
Currently dashboard tag can only be activated if the tag (the string before colon of the dashboard name) is all ASCII alphabets.

This PR tries to include characters with umlauts and multi-byte characters to become part of tags.

I know this regexp may not be perfect. Please let me know if you know a better solution.

non-ascii-dashboard-tags


arikfr:
Thanks!

Surprisingly it seems like there is no better solution 😢 .

Issue #1720, opened 2017-04-14
JQL: add support for fetching all the results by way of pagination


jvanegmond:
### Issue Summary

The JQL integration returns only the first 50 issues. This is the default number of issues returned via JIRA REST API. A mechanism should be implemented where a query is executed multiple times to fetch subsequent issues from JIRA.

### Steps to Reproduce

1. Configure JIRA integration
2. Create any JQL query which returns more than 50 issues
3. Execute the query

Expected: More than 50 issues returned.
Actual: Only 50 issues returned.

### Technical details:

* Redash Version: 0.12.0+b2449
* Browser/OS: Ubuntu 16.4
* How did you install Redash: Provisioning script from https://redash.io/help-onpremise/setup/setting-up-redash-instance.html

denisov-vlad:
You can set `"maxResults"` option to get more than 50 rows. It is not a bug, it's default JQL limitation.

arikfr:
Merged #1721.

While @denisov-vlad is right that you can just specify this in a query, it looks like 1K is a reasonable limit.

jiteshsojitra:
Please pass &maxResults=1000 in jql (it doesn't return me more than 1000 results).

arikfr:
@justmiles I reopened this.

PR #1721, opened 2017-04-14
[JIRA JQL] Change: change default max results limit from 50 to 1000

jvanegmond:
The current JQL implementation is limited to the default returned items by JIRA which is only 50 issues. This is unsuited for most purposes so this PR raises the limit to 1000 issues returned. The count query works correctly.

To fully solve the limit, multiple queries should be made to JIRA with different startAt parameter. This PR does not implement this, as I didn't have time for that. Therefore I have raised issue #1720.


arikfr:
Thanks.

PR #1722, opened 2017-04-17
[bugfix] Setup a new workers for beat to avoid unpredictable stuck of scheduled queries

yoheimuta:
I think that this PR is close to a bug fix. Because setting up based on https: //redash.io/help-onpremise/setup/setting-up-redash-instance.html causes the problem to be written below.

### Problem

Currently, if redash_celery workers gets stuck, scheduled queries will not be executed either. Even if no redash_celery_scheduled workers are processing the query. Because the redash_celery workers are also responsible for the periodic execution.

### Solution

By delegating periodic execution to the new redash_celery_beat workers, the execution status of the redash_celery workers will not affect the execution of the scheduled query.

I think that this is the behavior expected by the users.

### DIFF
--- a/setup/ubuntu/files/supervisord.conf
+++ b/setup/ubuntu/files/supervisord.conf
@@ -10,12 +10,21 @@ numprocs=1
 autostart=true
 autorestart=true
 
-# There are two queue types here: one for ad-hoc queries, and one for the refresh of scheduled queries
+# There are three queue types here: one for celery periodic tasks, one for ad-hoc queries, and one for the refresh of scheduled queries
 # (note that "scheduled_queries" appears only in the queue list of "redash_celery_scheduled").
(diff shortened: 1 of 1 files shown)


yoheimuta:
Thanks.

> redash_celery_beat process -- change concurrency to 1.

OK, I fixed 4b6369b.

> redash_celery process - keep celery in its queues list. If it's not busy - why not let it handle those tasks as well?

It is because I think that it's easier to clearly separate tasks for each process.

Specifically, since each log is written to a separate file, it is convenient to see only the log of redash_celery_beat for the periodic execution log.

How about that?

yoheimuta:
I mention @arikfr  just in case.

arikfr:
Closing this, as we no longer use supervisor. Thanks regardless!

PR #1727, opened 2017-04-18
improve Salesforce error message

akiray03:
In Salesforce datasource, the error message is:

> Error running query: Malformed request {url}. Response content: {content}

image

----

changes to:

> Error running query: [{"errorCode":"INVALID_FIELD","message":"\n CreatedDate >= '2017-01-01'\n ^\nERROR at Row:8:Column:5\nvalue of filter criterion for field 'CreatedDate' must be of type dateTime and should not be enclosed in quotes"}]

image

### DIFF
--- a/redash/query_runner/salesforce.py
+++ b/redash/query_runner/salesforce.py
@@ -159,7 +159,7 @@ class Salesforce(BaseQueryRunner):
             data = {'columns': columns, 'rows': rows}
             json_data = json_dumps(data)
         except SalesforceError as err:
-            error = err.message
+            error = err.content
             json_data = None
         return json_data, error
 


arikfr:
Are both error messages in the screenshots for the same error?

akiray03:
I got those screenshots in the same situation. 
The only difference is the difference in this code.

redash/query_runner/salesforce.py
```diff
@@ -159,7 +159,7 @@ def run_query(self, query, user):
             data = {'columns': columns, 'rows': rows}
             json_data = json_dumps(data)
         except SalesforceError as err:
-            error = err.message
+            error = err.content
```
arikfr: How about we take it a step forward and make it really friendly with something like: 

```python
if isinstance(err.content, list):
    error = "\n".join([content.get('message', '') for content in err.content)
else:
    error = err.content
```

?

arikfr:
Thanks!

PR #1730, opened 2017-04-19
[Cassandra] Add: support for UUID serializing and setting protocol version

mfouilleul:
- Add two parameters (protocol and cqlversion)
- Handle UUID datatype

### DIFF
--- a/redash/query_runner/cass.py
+++ b/redash/query_runner/cass.py
@@ -94,13 +94,13 @@ class Cassandra(BaseQueryRunner):
             if self.configuration.get('username', '') and self.configuration.get('password', ''):
                 auth_provider = PlainTextAuthProvider(username='{}'.format(self.configuration.get('username', '')),
                                                       password='{}'.format(self.configuration.get('password', '')))
-                connection = Cluster([self.configuration.get('host', '')], 
-                                     auth_provider=auth_provider, 
-                                     protocol_version=self.configuration.get('protocol', ''), 
+                connection = Cluster([self.configuration.get('host', '')],
+                                     auth_provider=auth_provider,
+                                     protocol_version=self.configuration.get('protocol', ''),
                                      cql_version=self.configuration.get('cqlversion', ''))
             else:
-                connection = Cluster([self.configuration.get('host', '')], 
-                                     protocol_version=self.configuration.get('protocol', ''), 
+                connection = Cluster([self.configuration.get('host', '')],
+                                     protocol_version=self.configuration.get('protocol', ''),
                                      cql_version=self.configuration.get('cqlversion', ''))
             session = connection.connect()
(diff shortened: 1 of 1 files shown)


redash/query_runner/cass.py
```diff
@@ -77,10 +94,14 @@ def run_query(self, query, user):
             if self.configuration.get('username', '') and self.configuration.get('password', ''):
                 auth_provider = PlainTextAuthProvider(username='{}'.format(self.configuration.get('username', '')),
                                                       password='{}'.format(self.configuration.get('password', '')))
-                connection = Cluster([self.configuration.get('host', '')], auth_provider=auth_provider, protocol_version=3)
+                connection = Cluster([self.configuration.get('host', '')],
+                                     auth_provider=auth_provider,
+                                     protocol_version=self.configuration.get('protocol', ''),
```
arikfr: To be consistent with previous versions (and the stated default value), let's change this to: `self.configuration.get('protocol', 3)`.

redash/query_runner/cass.py
```diff
@@ -77,10 +94,14 @@ def run_query(self, query, user):
...
+                                     cql_version=self.configuration.get('cqlversion', ''))
```
arikfr: Are you sure this one is needed? According to the docs it knows to pick the right one: "If a specific version of CQL should be used, this may be set to that string version. Otherwise, the highest CQL version supported by the server will be automatically used." (https://datastax.github.io/python-driver/api/cassandra/cluster.html)

If it is, the default value needs to be `None` and not an empty string.

redash/query_runner/cass.py
mfouilleul: ...my bad dude, totally right.

redash/query_runner/cass.py
mfouilleul: I had misadventures with those proto and cql versions, we have several production versions and even in the official cqlsh client we should juggle with these params, it is a mess, they fixed the version in the code, it tells you what version to use but does not change it... 👍 
```
cat /usr/local/bin/cqlsh | grep -i DEFAULT_CQLVER
DEFAULT_CQLVER = '3.3.1'
parser.add_option('--cqlversion', default=DEFAULT_CQLVER,
                 cqlver=DEFAULT_CQLVER, keyspace=None
```
```
cqlsh 127.0.0.1 -u cassandra
Password:
Connection error: ('Unable to connect to any servers', {'127.0.0.1': ProtocolError("cql_version '3.3.1' is not supported by remote (w/ native protocol). Supported versions: [u'3.4.0']",)})
```
I don't know the python client, I will do some tests before editing these.

Anyway the default value to "None" should be a good choice.

mfouilleul:
Any news?  :)

arikfr:
Thanks!

Sorry it took so long.

Issue #1724, opened 2017-04-18
Scheduled queries continue to run 1h+ are executed duplicatively


shimpeko:
### Issue Summary

I found my scheduled queries are executed duplicatively.
Those queries are relatively long running ones which takes more than 1 hour.

After some investigation, I came to think `CELERY_TASK_RESULT_EXPIRES` is the cause.
`CELERY_TASK_RESULT_EXPIRES` is now set as `3600` sec which is shorter than default value 1 day.
This means an `AsyncResult` is deleted after 1 hour it created, and I *THINK* it includes an `AsyncResult` with `STARTED`  state, or `AsyncResult` with running query in another word.

In scheduler log, I found a task was cancelled after 1 hour it had been started and then new task was created after the cancellation. While the query related to the cancelled task is still running. This cause duplicated queries running on our DB.

This may be not a bug, but I'm not sure this behavior is expected. And also IMO 1 hour is too short for default value.

#### Scheduler log
```
2017/04/17 20:00 Created: ec87fac9-cae6-4070-b1ba-2b5ba8cd61b3
"{\"username\": \"Scheduled\", \"retries\": 0, \"started_at\": 1492459209.55685, \"task_id\": \"ec87fac9-cae6-4070-b1ba-2b5ba8cd61b3\", \"created_at\": 1492459209.53251, \"updated_at\": 1492494826.952597, \"state\": \"finished\", \"query_id\": 1732, \"run_time\": 35617.35370802879, \"error\": null, \"scheduled\": true, \"scheduled_retries\": 0, \"data_source_id\": 13, \"query_hash\": \"d6c7c4a76f63f391c8c82a30c409687f\"}"

2017/04/17 20:00 Started: ec87fac9-cae6-4070-b1ba-2b5ba8cd61b3

2017/04/17 21:00 [2017-04-17 21:00:52,010: INFO/Worker-2910] In progress tracker for d6c7c4a76f63f391c8c82a30c409687f is no longer enqueued, cancelling (task: ec87fac9-cae6-4070-b1ba-2b5ba8cd61b3).

2017/04/17 21:01 Created: 5086eefe-2594-4b3a-8e64-7a2231aa73d1
[...]

PR #1736, opened 2017-04-25
Extend expiry time to prevent duplicate execution

shimpeko:
for https://github.com/getredash/redash/issues/1724 

I think 3600 sec for CELERY_TASK_RESULT_EXPIRES is too short, since 1h+ query is not rare in DWH workload.

I don't have strong reason for the alternative value, 12 hour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,8 +39,11 @@
 - Fix: page header wasn't updating on dashboards page @MichaelJAndy
 - Fix: keyboard shortcuts didn't work in parameter inputs
 
-## v1.0.0-rc.2 - 2017-02-22
+### Other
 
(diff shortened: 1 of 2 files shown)


arikfr:
My main concern here is the impact increasing `CELERY_TASK_RESULT_EXPIRES` might have on the size of Redis in an environment with a lot of frequent queries :| 

It seems that although it expires after 1 hour, it doesn't stop the execution or change anything. The only negative side effect is that once it expires the UI shows the query is in queue instead of executing. Maybe a good balance will be to increase it but to something lower like 3 or 4 hours?

REDASH_JOB_EXPIRY_TIME can be changed to 12 hours.

shimpeko:
Thank you for review.
Changed `CELERY_TASK_RESULT_EXPIRES` to 4 hours.

arikfr:
Merged. Thanks!

PR #1738, opened 2017-04-27
[Athena] Fix: queries throwing errors except for SELECT

suemoc:
When post query to Athena via re:dash, like CREATE TABLE or DESCRIBE,
a following error occurs.

```
Error running query: FAILED: ParseException line 1:0 cannot recognize input near '/' '*' 'Username'
```
In Athena, re:dash annotation format (`/* ~ */`) fails with many Hive DDL except for SELECT.

In this PR, I disable annotations in Athena query runner to solve the problem.

### DIFF
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -40,6 +40,9 @@ class Athena(BaseQueryRunner):
             'secret': ['aws_secret_key']
         }
 
+    @classmethod
+    def annotate_query(cls):
+        return False
 
     def get_schema(self, get_stats=False):
         schema = {}
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -4,8 +4,10 @@ import os
 import requests
 
 from redash.query_runner import BaseQueryRunner, register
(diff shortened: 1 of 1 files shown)


redash/query_runner/athena.py
```diff
@@ -40,6 +40,9 @@ def configuration_schema(cls):
             'secret': ['aws_secret_key']
         }
 
+    @classmethod
+    def annotate_query(cls):
+        return False
```
arikfr: I know of several Redash users who rely on this comment for monitoring/auditing purposes. Instead of disabling it, let's add a settings FLAG to decide whether to disable it with the default being `True` (=> enable it).

It can be something like:

```python
from redash.settings import parse_boolean

...

ANNOTATE_QUERY = parse_boolean(os.environ.get('ATHENA_ANNOTATE_QUERY', 'true'))
```

And then use `ANNOTATE_QUERY` as the return value of `annotate_query`.

suemoc: Sure. I've fixed the code.

arikfr:
Thanks!

PR #1747, opened 2017-05-03
[JQL] Fix: allow to override maxResults

denisov-vlad:
check if maxResults is already in query


redash/query_runner/jql.py
```diff
@@ -181,10 +181,10 @@ def run_query(self, query, user):
             field_mapping = FieldMapping(query.pop('fieldMapping', {}))
 
             if query_type == 'count':
-                query['maxResults'] = 1
+                query['maxResults'] = query.get('maxResults', 1)
```
arikfr: Here it needs to be always 1, because we disregard the results and just take the count value.

arikfr:
Thanks!

PR #1753, opened 2017-05-04
[Presto] Add: query cancellation support

fbertsch:
PyHive finally cut a new version with my cancellation changes, so we can push this upstream :)


arikfr:
Thanks!

PR #1760, opened 2017-05-09
Add: static enum parameter type.

rockwotj:
This adds simple enums from the discussion [here](https://discuss.redash.io/t/add-enum-support-for-query-parameters/366/5).

### DIFF
--- a/client/app/components/parameter-settings.html
+++ b/client/app/components/parameter-settings.html
@@ -25,7 +25,7 @@
       </div>
       <div class="form-group" ng-if="$ctrl.parameter.type === 'enum'">
         <label>Values (newline delimited)</label>
-        <textarea class="form-control" rows="3" ng-model="$ctl.parameter.enumOptions">
+        <textarea class="form-control" rows="3" ng-model="$ctrl.parameter.enumOptions">
       </div>
     </div>
 </div>
--- a/client/app/components/parameter-settings.html
+++ b/client/app/components/parameter-settings.html
@@ -13,7 +13,7 @@
         <select ng-model="$ctrl.parameter.type" class="form-control">
           <option value="text">Text</option>
           <option value="number">Number</option>
-          <option value="enum">Enum</option>
+          <option value="enum">Dropdown List</option>
           <option value="date">Date</option>
           <option value="datetime-local">Date and Time</option>
           <option value="datetime-with-seconds">Date and Time (with seconds)</option>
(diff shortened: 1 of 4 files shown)


client/app/components/parameter-settings.html
```diff
@@ -22,5 +23,9 @@ <h4 class="modal-title">{{$ctrl.parameter.name}}</h4>
         <label>Global</label>
         <input type="checkbox" class="form-inline" ng-model="$ctrl.parameter.global">
       </div>
+      <div class="form-group">
```
arikfr: Shouldn't this show only if the user picked `enum` before?

client/app/components/parameter-settings.html
```diff
@@ -13,6 +13,7 @@ <h4 class="modal-title">{{$ctrl.parameter.name}}</h4>
         <select ng-model="$ctrl.parameter.type" class="form-control">
           <option value="text">Text</option>
           <option value="number">Number</option>
+          <option value="enum">Enum</option>
```
arikfr: I'm not a native English speaker, but will `Enum` be clear for non developers?

client/app/components/parameters.html
```diff
@@ -6,6 +6,11 @@
       <input ng-switch-when="datetime-with-seconds" type="datetime-local" step="1" class="form-control" ng-model="param.ngModel">
       <input ng-switch-when="datetime-local" type="datetime-local" class="form-control" ng-model="param.ngModel">
       <input ng-switch-when="date" type="date" class="form-control" ng-model="param.ngModel">
+      <span ns-switch-when="enum">
+        <select ns-model="param.value" class="form-control">
+          <option ng-repeat="option in param.enumOptions.split('\n')" value="{{option}}">{{option}}</option>
```
arikfr: It's cleaner to do the split in the controller code (also gives more flexibility in the future).

client/app/components/parameter-settings.html
rockwotj: I don't think so, but this text is exposed to people writing queries only, will non-devs be writing queries often? I would think if you can write a query that you would know what an enum is. MySQL at least has a [Enum type](https://dev.mysql.com/doc/refman/5.7/en/enum.html). I'm fine to rename it if you'd like

client/app/components/parameters.html
rockwotj: Is this what you intended?

client/app/components/parameter-settings.html
rockwotj: Oops.

client/app/components/parameter-settings.html
arikfr: Maybe not often, but they do and we should be friendly :-)

How about "Dropdown List"?

arikfr:
I went ahead and renamed enum to dropdown and merged.

Thanks again for another great feature :)

rockwotj:
Thanks for renaming and merging!

arikfr:
Sure thing! Are you planning working on the dynamic one as well? 🙏

rockwotj:
I have another feature I will add first having to do with the datepicker (we already use it internally), but after that I might, especially because updating the drop-down is annoying.

Issue #1725, opened 2017-04-18
Alert notifications fail (sometime) with a SQLAlchemy error


danielerapati:
### Issue Summary

After migrating to 1.0.1 alerts are notified only sporadically to Slack (I cannot pinpoint exactly the conditions in which it happens).

### Steps to Reproduce

1. Create a failing alert for an existing query
2. Add your email and a Slack destination to your alert (the order seems to matter: add first the email destination then the slack destination)
3. Refresh the query such that the alert would trigger
4. You will receive an alert email but no slack alert
5. In the back-end `celery_error.log` you will see an error message like
```
[2017-04-18 13:13:58,184: ERROR/Worker-4] redash.tasks.check_alerts_for_query[fb4fee06-3318-44f7-8aaf-7984d792a793]: Error with processing destination
Traceback (most recent call last):
  File "/opt/redash/redash.1.0.1.b2833/redash/tasks/alerts.py", line 23, in notify_subscriptions
    subscription.notify(alert, alert.query_rel, subscription.user, new_state, current_app, host)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/attributes.py", line 237, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/attributes.py", line 584, in get
    value = self.callable_(state, passive)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/strategies.py", line 530, in _load_for_state
    (orm_util.state_str(state), self.key)
DetachedInstanceError: Parent instance <AlertSubscription at 0x7fcd934f04d0> is not bound to a Session; lazy load operation of attribute 'user' cannot proceed
```

I think this error is caused by the `subscription.user` call.
My guess is that it is eventually caused by [`models.db.session.commit()`](https://github.com/getredash/redash/blob/master/redash/tasks/alerts.py#L51). 

This "might" be connected to #1706 .

### Technical details:

* Redash Version: 1.0.1+b2833
* Browser/OS: (not relevant, error is in the backend)
* How did you install Redash: ec2 from ami (in 2015), recently upgraded from 0.12 to 1.0.1

arikfr:
But we initialize SQLAlchemy with `expire_on_commit` set to `False`:
```
db = SQLAlchemy(session_options={
    'expire_on_commit': False
})
```
([source](https://github.com/getredash/redash/blob/master/redash/models.py#L35))

[...]

PR #1780, opened 2017-05-22
Fix multiple alert subscription detached

danielerapati:
references #1725 

When a triggered Alert had more than 1 AlertSubscription it could fail
firing the second subscription (and subsequent) due to the
`alert.subscriptions` being detached from the `models.db.session`.

We avoid the problem by explicitely loading the subscriptions by id.

Also implemented a safeguard for alerts when the query result is empty (it assumes alert state is unchanged).

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -1104,18 +1104,20 @@ class Alert(TimestampMixin, db.Model):
 
     def evaluate(self):
         data = json.loads(self.query_rel.latest_query_data.data)
-        # todo: safe guard for empty
-        value = data['rows'][0][self.options['column']]
-        op = self.options['op']
+        if data['rows']:
+            value = data['rows'][0][self.options['column']]
+            op = self.options['op']
 
-        if op == 'greater than' and value > self.options['value']:
-            new_state = self.TRIGGERED_STATE
-        elif op == 'less than' and value < self.options['value']:
-            new_state = self.TRIGGERED_STATE
-        elif op == 'equals' and value == self.options['value']:
-            new_state = self.TRIGGERED_STATE
+            if op == 'greater than' and value > self.options['value']:
+                new_state = self.TRIGGERED_STATE
+            elif op == 'less than' and value < self.options['value']:
+                new_state = self.TRIGGERED_STATE
+            elif op == 'equals' and value == self.options['value']:
+                new_state = self.TRIGGERED_STATE
+            else:
+                new_state = self.OK_STATE
         else:
-            new_state = self.OK_STATE
+            new_state = self.state
 
         return new_state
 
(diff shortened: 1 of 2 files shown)


arikfr:
Thanks!

Did you manage to reproduce #1725 reliably before fixing?

danielerapati:
Hi,
I did reproduce it very reliably on our installation.
I'm working on replicating in a test: the issue clearly depends on the order of database operations and I'm not sure how mock objects will behave after doing [`models.db.session.commit()`](https://github.com/getredash/redash/blob/master/redash/tasks/alerts.py#L51).

Maybe upgrading SQLalchemy would solve the issue more cleanly (rather than hacking how we list AlertSubscriptions).

arikfr:
> I did reproduce it very reliably on our installation.

Can you elaborate on how you reproduce it? I tried to reproduce it in code and in tests (while removing mocks) and failed :( Also had this reproduce at times on demo.redash.io, but now it stopped.

> Maybe upgrading SQLalchemy would solve the issue more cleanly

Why?

I don't mind upgrading, but did you see anything in the changelog that might be related?

danielerapati:
> Can you elaborate on how you reproduce it?
It happens on our installation whenever an alert with more than one subscription fires. I deployed a `scheduler` container with this patch and it would notify all alert subscriptions without failing. 

I agree that there should be a test for this patch to be considered complete.
It almost seems like there is a bit of information missing to reliably reproduce.

> did you see anything in the changelog that might be related?
I'm going through them now.

arikfr:
> It happens on our installation whenever an alert with more than one subscription fires. I deployed a scheduler container with this patch and it would notify all alert subscriptions without failing.

I see it happen on a query with a single subscriber 🤔

> I agree that there should be a test for this patch to be considered complete.
It almost seems like there is a bit of information missing to reliably reproduce.

The reason I want to reproduce this before merging is that I'm worried we're burying a bigger issue here that we better understand and fix. Meanwhile spent almost a day on this and still no reliable reproduction or understanding of the problem 😞 .

arikfr:
I think I found it 🎉  the call to `app.app_context()` [here](https://github.com/getredash/redash/blob/master/redash/destinations/email.py#L43) is not needed and probably closes the previous session.

Going to verify this properly now, but wanted to share so you don't waste any more time.

danielerapati:
Wow, awesome job.
This would answer why tests with mock alert subscriptions don't replicate the problem and why it seems to depend on the order in which the alert subscriptions are notified.
Thank you

arikfr:
@danielerapati mind giving #1792 test on your environment? 

Also, can you revert the change to the way we fetch subscriptions we have in this pull request so I can merge it for the other fix? 

Thanks.

danielerapati:
Tested, it works as expected :+1:

I have opened #1793  for the safeguard against empty results.

arikfr:
thanks!

PR #1781, opened 2017-05-23
Add: Go to now button in date parameters

rockwotj:
This is an thing that we've been using internally, what do you think? I'm happy to guard behind a feature flag if you don't think it's useful.


rockwotj:
It seems the failing check in codeclimate is that there is not an extension for the `moment` import. I don't see how that is an error?

arikfr:
Can you add a screenshot of it looks like?

Btw, I would love to somehow add support for something like [DateRangePicker](http://www.daterangepicker.com/) (specially with the predefined ranges). Problem is that it returns two values (start and finish)...

rockwotj:
If you ever think of a clever way to get around the issue with multi-value parameters, I'd love to help contribute the DateRangePicker :)

My initial thought is a mutli-value parameter named `foo` would be used in the query template as `foo.start` and `foo.end`, but then there isn't an obvious way to show this to users unless you have full autocomplete support in the query editor.

arikfr:
>  My initial thought is a mutli-value parameter named foo would be used in the query template as foo.start and foo.end, but then there isn't an obvious way to show this to users unless you have full autocomplete support in the query editor.

You're right that it will be hard to discover or understand. 

What I had in mind is to have a set of "magic" parameters, that are globally defined/available. One of them will be a date range, so you can always reference: `date_range_start` and `date_range_end` in your queries. And if you do, we render the date range parameter UI. Others can be `redash_user_id` (which won't have a UI).

 WDYT?

rockwotj:
Ideas on how to allow discovery of this? Also what if date times where needed as well? Does this balloon into `datetime_range` variables too? If there are answers to those questions it sounds good to me.

rockwotj:
FWIW, while we're in the topic, it'd be great to have support for timezones too. All our data is stored in UTC timestamps, but we have people who access the data all over the world and want to see the data in their local time.

rockwotj:
@arikfr are you not interested in this change? If so please close this PR, we'll just keep this forked internally.

Issue #1782, opened 2017-05-23
Queries are kept waiting even when there are available workers


shimpeko:
### Issue Summary
I have 4 workers for adhoc/user queries. 4 workers means number of concurrency should be 4.
But sometimes we see less than 4 queries are in progress while there are waiting queries in queue on `/admin/queries/tasks` page.

I think this happens because of celery worker's prefetch function. A celery worker fetches 4 tasks(queries) at a time by default. If a worker fetches 4 tasks at once and 1st query is long running, rest of 3 queries will be kept waiting for long time even if other workers are available.

Prefetch is good for many short queries but not efficient with long running queries. So I think it should be configurable through `setting.py`.

About celery prefetch: http://docs.celeryproject.org/en/latest/userguide/optimizing.html#reserve-one-task-at-a-time

### Steps to Reproduce

1. Set up more than 2 workers
2. Run some queries

### Technical details:

- Redash Version: 1.0.3
- Browser/OS: Chrome
- How did you install Redash: Docker

arikfr:
> I have 4 workers for adhoc/user queries.

Is it 4 Celery workers or one Celery worker with concurrency of 4 (`-c4`)?

shimpeko:
4 celery workers and worker's concurrency is 1 (`-c1`).

arikfr:
I'm using ECS too, but I use both multiple ECS services/tasks & Celery's concurrency, to better utilize memory allocations.

arikfr:
But anyway, this explains why you might experience this. I'm just not sure that the solution in #1783 is the right direction :\

nocode99:
@arikfr Can you post some example celery configurations you use in ECS with rough estimations on your workloads?  We are currently testing the latest version and running the workers in Docker.  We provisioned 3 containers with concurrency set to 2 (so a max of 6).  I also setup celery flower to monitor and we are seeing inconsistent behaviors on when workers pick up tasks.  Sometimes, we'll have 4 queries running while the rest are in queue.  Then, we saw 6 queries running simultaneously, but after a bunch of tasks finished, the # of simultaneous queries dropped to 4 while we still had queries in queue.  Flower showed that the 3rd worker had no queries running, but did not pick up any jobs.

We are planning to put the celery workers in ECS but our testing has been very inconsistent.  On average, our queries take anywhere between seconds to ~10 min.  We primarily use this towards our Redshift cluster where we allow up to 10 simultaneous queries.

vbajaria:
[...]

PR #1783, opened 2017-05-23
Add setting for celery prefetch

shimpeko:
This PR is for https://github.com/getredash/redash/issues/1782 .

Celery worker fetches four tasks at a time. With long running queries, this cause tasks kept waiting even when there are available workers. I think Number of prefetch should be configurable though setting.

ref. http://docs.celeryproject.org/en/latest/userguide/optimizing.html#reserve-one-task-at-a-time

### DIFF
--- a/redash/settings.py
+++ b/redash/settings.py
@@ -73,6 +73,8 @@ SQLALCHEMY_ECHO = False
 CELERY_BROKER = os.environ.get("REDASH_CELERY_BROKER", REDIS_URL)
 CELERY_BACKEND = os.environ.get("REDASH_CELERY_BACKEND", CELERY_BROKER)
 CELERY_TASK_RESULT_EXPIRES = int(os.environ.get('REDASH_CELERY_TASK_RESULT_EXPIRES', 3600))
+CELERYD_PREFETCH_MULTIPLIER = int(os.environ.get('REDASH_CELERYD_PREFETCH_MULTIPLIER', 4))
+CELERY_ACKS_LATE = parse_boolean(os.environ.get('REDASH_CELERY_ACKS_LATE', "false"))
 
 # The following enables periodic job (every 5 minutes) of removing unused query results.
 QUERY_RESULTS_CLEANUP_ENABLED = parse_boolean(os.environ.get("REDASH_QUERY_RESULTS_CLEANUP_ENABLED", "true"))
--- a/redash/worker.py
+++ b/redash/worker.py
@@ -47,7 +47,9 @@ if settings.QUERY_RESULTS_CLEANUP_ENABLED:
 celery.conf.update(CELERY_RESULT_BACKEND=settings.CELERY_BACKEND,
                    CELERYBEAT_SCHEDULE=celery_schedule,
                    CELERY_TIMEZONE='UTC',
-                   CELERY_TASK_RESULT_EXPIRES=settings.CELERY_TASK_RESULT_EXPIRES)
+                   CELERY_TASK_RESULT_EXPIRES=settings.CELERY_TASK_RESULT_EXPIRES,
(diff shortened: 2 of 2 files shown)


redash/settings.py
```diff
@@ -73,6 +73,8 @@ def all_settings():
 CELERY_BROKER = os.environ.get("REDASH_CELERY_BROKER", REDIS_URL)
 CELERY_BACKEND = os.environ.get("REDASH_CELERY_BACKEND", CELERY_BROKER)
 CELERY_TASK_RESULT_EXPIRES = int(os.environ.get('REDASH_CELERY_TASK_RESULT_EXPIRES', 3600))
+CELERYD_PREFETCH_MULTIPLIER = int(os.environ.get('REDASH_CELERYD_PREFETCH_MULTIPLIER', 4))
+CELERY_ACKS_LATE = parse_boolean(os.environ.get('REDASH_CELERY_ACKS_LATE', "false"))
```
arikfr: `ACKS_LATE` is a dangerous setting to add, as anyone who doesn't fully understand how Celery works can easily shoot himself in the foot.

If I understand it correctly, when Celery acknowledges tasks only after they finish (ACKS_LATE = True), then if the task takes > `visibility_timeout` (default = 1 hour) it will run the task again. Meaning that any query that takes longer than `visibility_timeout` will be executed multiple times.

The prefetch multiplier is only relevant in cases where there are more than one Celery workers for the same queue (not the same thing as one worker with a process pool). It only applies to _unacknowledged_ tasks, so in the case of ACKS_LATE=False, any task the worker fetches will be immediately acknowledged...

Relevant documentation:
http://docs.celeryproject.org/en/3.1/glossary.html
http://docs.celeryproject.org/en/3.1/getting-started/brokers/redis.html

This is another example of how Celery is just too clever for us and we might benefit from using a simpler library like [rq](http://python-rq.org/)...

jeffwidman: Not just too clever, but my experience at work has been that Celery is occasionally buggy when you start flipping more advanced options like this. The code is simply less well-tested and fewer folks have incentive to maintain these parts of the celery code. I like Celery for many things, but I agree with @arikfr that it's best to try and minimize usage of the non-standard options unless forced otherwise (which does happen from time to time).

arikfr: @jeffwidman have you considered alternatives?

jeffwidman: I've looked at `rq` for personal projects and liked what I saw, but never actually tried it. Also, I'm worried I'd miss just enough of celery's extra functionality that I'd regret it. 

At work, it really depends on the need... rq doesn't have enough tuning knobs for many of our more advanced use cases, celery hits a nice sweet spot when you need more than a worker queue but less than migrating the service into a full-blown kafka pipeline. The other big problem though we run into with celery is that we have some tasks where we can't lose the data for, but celery's best backend is redis, where getting enough durability guarantees require a fair bit of operational overhead. And a couple of those services also are often at risk of having redis run out of memory (even on very large boxes) because celery doesn't support redis-cluster... which forces us to load balance at a different layer of the stack where it's often more of a pain operationally... so I'd say we have a like/dislike relationship with celery that isn't strong enough either direction to be a love/hate relationship.

arikfr: Thanks @jeffwidman. Actually I'm not sure about "Celery's best backend is redis" -- at some point they even thought to deprecate the Redis support. I think the most compatible backend is RabbitMQ.

My main issue w/ Celery is visibility into what's going on (# of currently running jobs, active workers, etc). It has some APIs to get this information, but it's not stable from my experience if the workers are overloaded and can't respond to the messages.

I think going forward we will either have to better understand how to monitor Celery with Redis or switch to `rq` while implementing missing functionality.

arikfr:
I'm closing this now, but @shimpeko you're welcome to reopen if you have thoughts on the issues pointed out in the comments.

Thanks.

PR #1793, opened 2017-05-29
safeguard alerts against empty query results

danielerapati:
alert state will become Unknown

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -1143,18 +1143,20 @@ class Alert(TimestampMixin, db.Model):
 
     def evaluate(self):
         data = json.loads(self.query_rel.latest_query_data.data)
-        # todo: safe guard for empty
-        value = data['rows'][0][self.options['column']]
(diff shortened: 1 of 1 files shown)


redash/models.py
```diff
@@ -1143,18 +1143,20 @@ def to_dict(self, full=True):
...
+                new_state = self.TRIGGERED_STATE
+            elif op == 'less than' and value < self.options['value']:
+                new_state = self.TRIGGERED_STATE
+            elif op == 'equals' and value == self.options['value']:
+                new_state = self.TRIGGERED_STATE
+            else:
+                new_state = self.OK_STATE
         else:
-            new_state = self.OK_STATE
+            new_state = self.state
```
arikfr: Maybe it should switch back to UNKNOWN state in such case?

danielerapati: :+1:

arikfr:
Thanks!

PR #1797, opened 2017-05-30
Add orderBy:'name' to ng-repeat attribute on some pages

sdif:
Hi,
We are using Redash and we got a huge list of data sources, we wanted to sort data sources by name but it can't be done by default. 
I added it in a "hard way" but it seems to accommodate everybody like this.

Saâd

### DIFF
--- a/client/app/pages/data-sources/list.html
+++ b/client/app/pages/data-sources/list.html
@@ -5,7 +5,7 @@
         <a href="data_sources/new" class="btn btn-default"><i class="fa fa-plus"></i> New Data Source</a>
       </p>
       <div class="list-group">
-        <a ng-href="data_sources/{{dataSource.id}}" class="list-group-item" ng-repeat="dataSource in dataSources"><i class="fa fa-database"></i> {{dataSource.name}}</a>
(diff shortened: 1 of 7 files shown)


harmy:
I think sorting at the server side is the right way to do this.

arikfr:
I agree with @harmy, this should be done on the server side.

sdif:
Oh yes right ! 
I removed the order from the `*list.html` and added an `.order_by(models.Model.name)` on the `list` function

arikfr:
@GitSumito as you been using the CLI a lot recently, do you mind reviewing this?

GitSumito:
> @GitSumito as you been using the CLI a lot recently, do you mind reviewing this?

I'll check it!

GitSumito:
@sdif 
cc @arikfr 
Functionally, it is OK.
In my environment works well.

But why don't you add test code??

GitSumito:
@sdif 
cc @arikfr 
Functionally, it is OK.
In my environment works well.

But why don't you add test code??

GitSumito:
@sdif 
cc @arikfr 
Functionally, it is OK.
In my environment works well.

But why don't you add test code??

GitSumito:
@sdif 
cc @arikfr 
Functionally, it is OK.
In my environment works well.

But why don't you add test code??

GitSumito:
@sdif
@arikfr
I can make some test code for this.
But status of the repository that @sdif made is "unknown repository".

If you don't mind, can I make a new pull request with test code?

PR #1807, opened 2017-06-04
Fix redirect to /setup on the last setup step

44px:
If you try to set up developer environment from master branch, you'll get redirect to `http://localhost:8080/login?next=/` instead of `http://localhost:8080/setup` on the last step.

Seems like it caused by `current_org is not None` check.

PS: I'm not sure these changes are 100% correct (I'm not a python developer). Maybe there is a better way to fix problem.


arikfr:
Oops. Thanks!

Issue #1777, opened 2017-05-21
Athena query runner that uses the REST API


arikfr:
Amazon finally released documentation for Athena's REST API and @laughingman7743 created [PyAthena](https://pypi.python.org/pypi/PyAthena/1.0.0), DB API 2.0 compliant client for Athena. We can now create a query runner that doesn't need the JDBC proxy. 👌

arikfr:
Implemented :)

PR #1809, opened 2017-06-05
Implement Athena query runner using RestAPI

laughingman7743:
getredash/redash#1777

### DIFF
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -75,7 +75,8 @@ class Athena(BaseQueryRunner):
                     'title': 'KMS Key',
                 },
             },
-            'required': ['region', 's3_staging_dir']
+            'required': ['region', 's3_staging_dir'],
+            'secret': ['aws_secret_key']
         }
 
     @classmethod
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -39,31 +39,32 @@ class Athena(BaseQueryRunner):
     @classmethod
     def name(cls):
         return "Amazon Athena"
-    
+
     @classmethod
     def configuration_schema(cls):
         return {
             'type': 'object',
             'properties': {
-                'aws_access_key_id': {
+                'region': {
                     'type': 'string',
-                    'title': 'AWS Access Key ID',
+                    'title': 'AWS Region'
                 },
-                'aws_secret_access_key': {
+                'aws_access_key': {
                     'type': 'string',
-                    'title': 'AWS Secret Access Key',
+                    'title': 'AWS Access Key'
                 },
-                'schema_name': {
+                'aws_secret_key': {
                     'type': 'string',
-                    'title': 'Schema Name',
-                },
-                'region_name': {
-                    'type': 'string',
-                    'title': 'Region Name',
+                    'title': 'AWS Secret Key'
                 },
                 's3_staging_dir': {
                     'type': 'string',
-                    'title': 'S3 Staging Directory',
+                    'title': 'S3 Staging Path'
+                },
+                'schema': {
+                    'type': 'string',
+                    'title': 'Schema Name',
+                    'default': 'default'
                 },
                 'encryption_option': {
                     'type': 'string',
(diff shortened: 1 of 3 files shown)


redash/query_runner/athena.py
```diff
@@ -1,102 +1,127 @@
...
             error = None
-
-            return json_data, error
-        except requests.RequestException as e:
-            if e.response.status_code == 400:
-                return None, response.content
-
-            return None, str(e)
-        except KeyboardInterrupt:
-            error = "Query cancelled by user."
```
arikfr: Let's keep the special handler for `KeyboardInterrupt` to have the friendlier error message.

redash/query_runner/athena.py
```diff
@@ -1,102 +1,127 @@
...
                 },
-                'aws_access_key': {
+                'aws_secret_access_key': {
                     'type': 'string',
-                    'title': 'AWS Access Key'
+                    'title': 'AWS Secret Access Key',
                 },
-                'aws_secret_key': {
+                'schema_name': {
                     'type': 'string',
```
arikfr: Let's add a default value here (`default`) and then it can be no longer a mandatory field.

redash/query_runner/athena.py
```diff
@@ -1,102 +1,127 @@
...
-    def name(cls):
-        return "Amazon Athena"
-
-    @classmethod
     def configuration_schema(cls):
         return {
             'type': 'object',
             'properties': {
-                'region': {
+                'aws_access_key_id': {
```
arikfr: As this is supposed to be a drop in replacement for the existing query runner, let's keep the names same as they were in the previous query runner. 

I know this means that we will need to construct the kwargs dict manually in the connect call, but I think it's worth it here.

redash/query_runner/athena.py
```diff
@@ -1,102 +1,127 @@
...
+    'decimal': TYPE_FLOAT,
+}
+
 
 class Athena(BaseQueryRunner):
     noop_query = 'SELECT 1'
 
     @classmethod
-    def name(cls):
-        return "Amazon Athena"
```
arikfr: Why?

redash/query_runner/athena.py
```diff
@@ -1,102 +1,127 @@
...
+                    'title': 'Encryption Option',
+                },
+                'kms_key': {
+                    'type': 'string',
+                    'title': 'KMS Key',
+                },
             },
-            'required': ['region', 'aws_access_key', 'aws_secret_key', 's3_staging_dir'],
-            'secret': ['aws_secret_key']
+            'required': ['region_name', 'schema_name', 's3_staging_dir']
```
arikfr: AWS access/secret keys are no longer required as they can be taken from ENV variables?

redash/query_runner/athena.py
laughingman7743: OK. I will keep the same names as the previous query runner.

redash/query_runner/athena.py
laughingman7743: 😱

redash/query_runner/athena.py
laughingman7743: Yes, PyAthena support boto3 credentials. http://boto3.readthedocs.io/en/latest/guide/configuration.html
Besides environment variables, if you want to authenticate using the role of EC2 instance, you will not specify secret key and access key.

redash/query_runner/athena.py
laughingman7743: OK. I will add KeyboardInterrup handler.

laughingman7743:
@arikfr done

laughingman7743:
I would like to add an Athena query runner as the default query runners. Is this all right?

arikfr:
> I would like to add the Athena query runner as the default query runners. Is this all right?

Yes.

laughingman7743:
Thank you. I added the Athena query runner as the default query runners.

arikfr:
Thanks!

laughingman7743:
Sorry, the key name of the region name in the configuration schema was incorrect and it was fixed.

laughingman7743:
Thanks!

PR #1812, opened 2017-06-07
UI change: same view for input on search result page as in header

44px:
Just small markup change to make UI more consistent :)

before:
image

after:
image


arikfr:
Thanks!

PR #1813, opened 2017-06-07
UI change: add some space between parameters

44px:
before:
image

after:
image

### DIFF
--- a/client/app/components/parameters.html
+++ b/client/app/components/parameters.html
@@ -4,7 +4,7 @@
      ng-model="parameters">
   <div class="form-group m-l-10 m-r-10"
        ng-repeat="param in parameters">
-    <label>{{param.title}}</label>
+    <label class="parameter-label">{{param.title}}</label>
     <button class="btn btn-default btn-xs"
             ng-click="showParameterSettings(param)"
             ng-if="editable">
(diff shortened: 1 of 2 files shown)


client/app/components/parameters.html
```diff
@@ -1,7 +1,15 @@
-<div class="form-inline bg-white p-5" ng-if="parameters | notEmpty" ui-sortable="{ 'ui-floating': true, 'disabled': !editable }" ng-model="parameters">
-  <div class="form-group" ng-repeat="param in parameters">
+<div class="form-inline bg-white p-5"
+     ng-if="parameters | notEmpty"
+     ui-sortable="{ 'ui-floating': true, 'disabled': !editable }"
+     ng-model="parameters">
+  <div class="form-group m-l-10 m-r-10"
```
arikfr: Why add margin on both sides? Shouldn't m-r be enough?

Btw, did you  try putting the labels on top of the input boxes?

44px: Yes, at first I tried to add just `m-r-20` only but it looks bad on "mobile" view:
image

Adding 10px on both sides fixes the issue without extra CSS:
image

> Btw, did you try putting the labels on top of the input boxes?

Good idea! It looks more cleaner and saves extra space:

image

arikfr:
Thanks!

PR #1826, opened 2017-06-14
update if test for string instead of expecting boolean

alison985:
Hi. Right now if the environment variable REDASH_MULTI_ORG is unset the `if settings.MULTI_ORG` statement returns true and can't find the multi_org.html template, generating 5 test suite errors. 

By changing the if statement to look for the string "true" all tests pass regardless of whether REDASH_MULTI_ORG is set to "true", "false", or unset.

### DIFF
--- a/redash/handlers/static.py
+++ b/redash/handlers/static.py
@@ -27,7 +27,7 @@ def send_static(filename):
 
 
 def render_index():
-    if settings.MULTI_ORG:
+    if settings.MULTI_ORG == "true":
         response = render_template("multi_org.html", base_href=base_href())
     else:
         full_path = safe_join(settings.STATIC_ASSETS_PATHS[-2], 'index.html')


arikfr:
Hi 👋 

> Right now if the environment variable REDASH_MULTI_ORG is unset the if settings.MULTI_ORG statement returns true and can't find the multi_org.html template, generating 5 test suite errors.

This is how the value is being loaded ([line 88 in settings.py](https://github.com/getredash/redash/blob/master/redash/settings.py#L88)):

```python
MULTI_ORG = parse_boolean(os.environ.get("REDASH_MULTI_ORG", "false"))
```

If the value is unset it will default to `"false"` and then converted `False` by `parse_boolean`.

> By changing the if statement to look for the string "true" all tests pass regardless of whether REDASH_MULTI_ORG is set to "true", "false", or unset.

By doing this change it will make MULTI_ORG not work, because it can never be the "true" string. It's either `True` or `False`.

The reason MULTI_ORG is True in tests is because [it's set to be so on purpose](https://github.com/getredash/redash/blob/master/tests/__init__.py#L15). I don't remember the exact reasoning, but it made things easier when testing.

PR #1833, opened 2017-06-19
Update to Webpack 2

44px:
Hello!

Here are some highlights from this PR:
1) Webpack config updated to new format (https://github.com/getredash/redash/pull/1833/commits/202c53c7d77978ce4bfc183dead932f4f7e938fc)
2) Webpack Dev Server and loaders updated to the latest version (https://github.com/getredash/redash/pull/1833/commits/45f448e0d11fd4be6e2a9ad2ed0e64b6a0f9efb1, https://github.com/getredash/redash/pull/1833/commits/ee4b05eb9894bf425f171c4dd4ce5e2267df50a7)
3) ~~Also, I removed and recreated `npm-shrinkwrap.json` so some of the app' dependencies got minor updates. I didn't notice any problems at a first glance but maybe this change should be moved into separate PR anyway?~~ ng-annotate-loader moved to devDependencies and removed from `npm-shrinkwrap.json`

PS: it was quite funny to see [v3.0.0](https://github.com/webpack/webpack/releases/tag/v3.0.0) released a few hours before I was close to finish this PR =)


arikfr:
Amazing. Thanks!

> Also, I removed and recreated npm-shrinkwrap.json so some of the app' dependencies got minor updates. I didn't notice any problems at a first glance but maybe this change should be moved into separate PR anyway?

If it's possible to just update the webpack dependencies and keep everything else as is, I will prefer it. I want to release a stable version soon, and a bit worried by edge case annoying bugs that might result from updating the other dependencies. But it's a lot of hassle, to split this out, then keep as is.

44px:
Done =)

I did `git rebase -i`, restored `npm-shrinkwrap.json` and removed ng-annotate via `npm uninstall`. So production dependencies versions aren't affected now.

arikfr:
Merged. Thanks once again!

Based on their release blog post, upgrading to V3 shouldn't be hard from V2 🤞

44px:
Thanks!

Yes, this upgrade should be easier, but I would wait for v3.1.x at least ;)

Anyway, now we can try to leverage tree-shaking and cut some bytes. I briefly tested it (honestly, without impressive results - just 11KB which is about 4% of current bundle size).

Also, I found that ng-annotate doesn't work with ES6 modules so I [replaced it with babel plugin](https://github.com/getredash/redash/pull/1863).

Issue #1790, opened 2017-05-29
LDAP/Basic Auth Support for the Redash itself


Civil:
This is a feature request. Maybe I've missed that in documentation, but I'd like to see either LDAP or at least Basic Auth support in Redash (e.x. if nginx sends user parameters, assume and autocreate a user with some default rights and also override current internal authorization).

yuokada:
Duplicated?
https://github.com/getredash/redash/issues/633

Civil:
Yeah, thx.

arikfr:
@amarjayr's #1836 was just merged 🙌

Issue #633, opened 2015-11-02
Support for LDAP authentification


gissehel:
Support for ldap auth would be usefull.

adamb0mb:
:+1:
LDAP, but specifically Active Directory would be nice for us.

pjrtb:
+1

TonyStarkBy:
+1 for jira

arikfr:
@TonyStarkBy what do you mean by JIRA? Use Jira's users?

rakeshsingh:
+1

Slach:
+1

test-account-0:
+1

ersiny:
+1

slakr:
+1

evilNullPointer:
+1

epaiva:
+1

adamb0mb:
@arikfr I haven't been keep track... is there any plans for this?

Huangsir:
+1

epaiva:
Hello everyone! Maybe we could help to get this issue solved by supporting it with donations. Do you think it helps @arikfr ? If it does, we can try using BountySource. https://www.bountysource.com/issues/27885192-support-for-ldap-authentification

swainr:
+1

ConnorChristie:
+1

gregwjacobs:
+1

cumthqb:
+1

deecay:
I would like to point out that Redash can handle LDAP since remote user auth by @tsibley #883 had been merged.

You need to setup a web server with LDAP  and reverse proxy capability. I chose apache/httpd, but I believe nginx can handle most of the cases too.

I use https://github.com/dariko/httpd-rproxy-ldap with a little modification for RequireAll/RequireAny. Works with my company AD setup.

agnoldo:
Hi @deecay, thanks for this tip. Sorry for my lack of knowledge about LDAP but I'd like to ask you if the following setup may work:

- I've installed re:dash in a VM running Ubuntu
- I've two different LDAP servers running in my organization. I can choose any of them (one is a little better)

Should I install mod_authnz_ldap Apache Module and then execute httpd-rproxy-ldap's run.sh on re:dash's apache server in order to make it connect with one of my LDAP servers?

Thanks in advance

Arnaldo

deecay:
Hi @agnoldo,

[...]

PR #1836, opened 2017-06-20
LDAP (Active Directory) implementation

amarjayr:
Adds support for LDAP integration (https://github.com/getredash/redash/issues/633 & https://github.com/getredash/redash/issues/1790). When enabled, users can login to Redash using their LDAP credentials by going to "/ldap_auth/login". 

Uses the [ldap3](https://github.com/cannatag/ldap3) library (v. 2.2.4 ~ [GNU Lesser General Public License](https://github.com/cannatag/ldap3/blob/master/LICENSE.txt))

Adds the following environment variables for configuration:
- `REDASH_LDAP_LOGIN_ENABLED`: when true LDAP login will be enabled. Users can login using ldap by going to "/ldap_auth/login". If `PASSWORD_LOGIN_ENABLED` is false, users will automatically be redirected to "/ldap_auth/login".
- `REDASH_LDAP_URL`: the full URL of the LDAP host (including port).
- `REDASH_LDAP_BIND_DN`: the DN to bind to
- `REDASH_LDAP_BIND_DN_PASSWORD`: the password for the bind DN
- `REDASH_LDAP_DISPLAY_NAME_KEY`: The key for the display name.
- `REDASH_LDAP_EMAIL_KEY`: The key for the email.
- `REDASH_LDAP_CUSTOM_USERNAME_PROMPT`: The custom username prompt. Some LDAP implementations use usernames whereas other ones use emails so the prompt needs to be customizable.
- `REDASH_LDAP_SEARCH_TEMPLATE`: The search template (ie. what the username entered into redash should match to in LDAP).
- `REDASH_SEARCH_DN`: The schema to bind to.


arikfr:
Thank you for making this!

My main concern here is that it will get to the state of the SAML support we have right now: where I have no idea how the code works or a way to test and can't help people who have issues. How about we turn this into a plugin to Redash, that people can separately install and will live in its own repository? 

Having it as a plugin will also help with the fact that the ldap3 package is LGPL3 licensed, which I'm not sure if it is compatible with our BSD3 license.

Do you have the bandwidth to do these changes?

AmarJayR:
Hi @arikfr, that's an interesting idea. @jeffwidman and I rather have it in the core, but understand the concern. Are you pulling native SAML integration out? How would a plugin be architected? Redash doesn't seem to support plugins yet. Can the plugin repo live under the `getredash` org on GitHub. We can pull LDAP into a plugin but want the visibility/officialness of being under the redash umbrella, even if it's marked as less/non-supported.

arikfr:
> Are you pulling native SAML integration out?

No, I'm actually planning to invest some time in improving it. But that's because I have customers who need it.

> How would a plugin be architected? Redash doesn't seem to support plugins yet.

Don't know :) Need to think about this, but I think there is a simple path to reach it. Maybe can learn from how Flask plugins work.

> Can the plugin repo live under the getredash org on GitHub.

Yes.

jeffwidman:
My preference is that any external authorization plugins have the same architecture/status. So if you're supporting SAML in core, then I'd prefer that LDAP was pulled into core. If you want a plugin, then can you show us the way by moving SAML into a plugin?

Since these are the two most popular protocols, between them and google auth you've covered 95%+ use cases and can tell folks who ask for additional protocols to modify their identity provider to support one of these two. We originally started down the SAML path, but our redash instance is behind a firewall and not externally routable so we can't plug it into our SAML provider. So personally if it were me I'd probably pull it into core since pretty much all this PR does is add some config flags and a new blueprint. The LDAP protocol stuff is all outsourced to an actively maintained project, plus it's not like LDAP protocol will change much.

Licensing is a good point--I discussed with @AmarJayR and he's researching several options. I know he originally picked `ldap3` library because it was pure python so easier to debug/contribute fixes to, plus a much stronger community than the alternatives which mostly seemed unmaintained. The best solution would be if they'd be willing to re-license themselves, that's a big ask of them, but never hurts to ask.

AmarJayR:
@cannatag (owner of ldap3) confirmed that it is okay to use his project (https://github.com/cannatag/ldap3/issues/374):

> I think you can use an LGPL3 licensed project in your BSD3 licensed project. You don't need to include only BSD3 licensed projects in a BSD3 licensed project. Just declare that the ldap3 project uses the LGPL3 project and you're ok.

jeffwidman:
@arikfr Any update on this? @amarjayr is done with his Netskope internship at the end of this month, so it'd be nice to get this finalized before he moves on. I'll still be around to support this after he moves on.

strangeman:
Works like a charm, thank you!

jeffwidman:
Any update on this @arikfr?

jeffwidman:
@arikfr any update here?

We've been getting radio silence for the past month, despite this being a highly requested feature (#633) with seventeen thumbs-up.

redash/authentication/ldap_auth.py
```diff
@@ -1,20 +1,26 @@
...
 from flask_login import current_user, login_required, login_user, logout_user
-from ldap3 import Server, Connection, SIMPLE
 
-from redash import settings
+try:
+    from ldap3 import Server, Connection, SIMPLE
+except ImportError:
+    if settings.LDAP_LOGIN_ENABLED:
+        logger.error("ldap3 is required to use LDAP authentication (see requirements.txt)")
+        raise
```
arikfr: Maybe just `exit()` instead of `raise`?

arikfr:
Merged 👍  Thank you for your cooperation and patience, @amarjayr.

For those following from home: to avoid including GPL licensed library the LDAP library is commented out in `requirements.txt`. So if you want to use the LDAP integration, you will need to install it manually.

Just to clarify: `ldap3`'s author is right -- there is **no** issue with using a GPL licensed library. But some corporates have a real fear of GPL and try to avoid it. To make their life easier we make this optional.

cannatag:
Hi I'm the author of the ldap3 library. ldap3 is released under the LGPL3. The "L" stands for "lesser" and means that the use of the library in proprietary software is permitted.

arikfr:
@cannatag thank you for chiming in! I know and personally I don't mind. But I had enough times companies ask me "do you use any GPL software?" (and I don't think they make the distinction between GPL, LGPL or any other kind of GPL), so I just try to avoid it when I can (in Redash). This makes Redash's adoption much easier in such companies.

cannatag:
You're right @arikfr. This has become a problem with version 3 of GPL. It's a pity that they created a such limited license. Probably I will change to the MIT or BSD license in the future.

jeffwidman:
@cannatag +1 for switching `ldap3` to at least a dual-license that supports MIT/BSD. I understand the context here and why technically GPL should be fine, but also empathize with @arikfr's position as a one-man shop trying to put food on the table that he doesn't have time to educate lawyers about the finer points of software licenses...

noahrowland:
Note that there is an issue with this implementation on the front end.

App routes url to '/ldap/login' based on ldap_auth.py in authentication.
Template login page redirects to '/ldap_auth/login' which returns a 404 since there is no route pointing to that url. Either correct the route or correct the template redirect.

noahrowland:
Also note that the documentation should be updated for the following env variable:
export REDASH_LDAP_SEARCH_TEMPLATE='(sAMAccountName=%(user)s)'

The code in ldap_auth expects the key to be 'username' not 'user'. However, 'user' is the most common key convention.

jeffwidman:
Can you report these as new tickets so they don't get lost?

spuder:
For refence `RUN pip install ldap3` does not work as discussed here: https://discuss.redash.io/t/recommended-method-of-including-ldap-fails/4821/4

For those trying to build your own docker containers with ldap 3, use the `--user` option. 
```
FROM redash/redash:8.0.2.b37747

RUN pip install --user ldap3
```

PR #1846, opened 2017-06-28
Athena updates: bring back ability to disable annotations and disable formatter

arikfr:
* use simple formatter to avoid the need to escape "%" character.
* bring back the option to disable query annotations.

### DIFF
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import os
 
 from redash.query_runner import *
 from redash.settings import parse_boolean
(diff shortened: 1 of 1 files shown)


redash/query_runner/athena.py
```diff
@@ -33,6 +34,11 @@
 }
 
 
+class SimpleFormatter(object):
+    def format(self, operation, parameters=None):
+        return operation
```
arikfr: @laughingman7743 I'm using this to avoid escaping the '%' character in Athena queries. WDYT?

laughingman7743: LGTM!

PR #1860, opened 2017-06-29
Remove /forgot endpoint if REDASH_PASSWORD_LOGIN_ENABLED is false

AmarJayR:
Users should not be able to reset their password if `REDASH_PASSWORD_LOGIN_ENABLED` is false. Instead, the `/forgot` endpoint should return a `404`.


arikfr:
Thanks 👌

PR #1862, opened 2017-07-01
signed embed parameters

talgryn:
enable signing urls with embedded parameters

### DIFF
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -10,7 +10,7 @@ from redash import models, settings, utils
 from redash.tasks import QueryTask, record_event
 from redash.permissions import require_permission, not_view_only, has_access, require_access, view_only
 from redash.handlers.base import BaseResource, get_object_or_404
-from redash.utils import collect_query_parameters, collect_parameters_from_request, gen_query_hash
+from redash.utils import collect_query_parameters, collect_parameters_from_request, gen_query_hash, gen_signature_hash
 from redash.tasks.queries import enqueue_query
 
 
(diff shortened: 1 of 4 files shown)


redash/handlers/query_results.py
```diff
@@ -187,10 +187,24 @@ def get(self, query_id=None, query_result_id=None, filetype='json'):
...
+                    if settings.EMBED_KEY:
+                        is_valid = False
+                        provided_signature = request.args.get('signature')
+                        if provided_signature is not None:
+                            expected_sig = gen_signature_hash(parameter_values, None)
+                            if expected_sig == provided_signature:
+                                is_valid = True
+                        elif request.args.get('gensig'):
+                            expected_sig = gen_signature_hash(parameter_values, request.args.get('gensig'))
+                            print "signature %s " % expected_sig
```
arikfr: Please don't leave debug code...

redash/handlers/query_results.py
```diff
@@ -187,10 +187,24 @@ def get(self, query_id=None, query_result_id=None, filetype='json'):
             query_result = get_object_or_404(models.QueryResult.get_by_id_and_org, query_result_id, self.current_org)
         elif query_id is not None:
             query = get_object_or_404(models.Query.get_by_id_and_org, query_id, self.current_org)
-
             if query is not None:
                 if settings.ALLOW_PARAMETERS_IN_EMBEDS and parameter_values:
-                    query_result = run_query_sync(query.data_source, parameter_values, query.to_dict()['query'], max_age=max_age)
+                    is_valid = True
```
arikfr: Instead of using the `is_valid` flag, just abort the request with a relevant status code (401?).

redash/utils/__init__.py
```diff
@@ -42,6 +42,10 @@ def dt_from_timestamp(timestamp, tz_aware=True):
 def slugify(s):
     return re.sub('[^a-z0-9_\-]+', '-', s.lower())
 
+def gen_signature_hash(parameters, key):
+    if key is None:
+        key = settings.EMBED_KEY
```
arikfr: I think it will be safer if we use a distinct key for each query. We can reuse the current query API key, we just need to make sure we don't send it to "guest" users (it might be returned by default in the query API response now).

redash/utils/__init__.py
```diff
@@ -42,6 +42,10 @@ def dt_from_timestamp(timestamp, tz_aware=True):
...
+    return hashlib.sha256(str(parameters)+key).hexdigest()
```
arikfr: You better sort the parameters, to make sure the order doesn't make a difference. Also look into using the [`hmac`](https://docs.python.org/2/library/hmac.html) library.

redash/handlers/query_results.py
talgryn: It's more complicated than that since the validation of a request is a complex condition (if we use embed key, and the signature was passed, and the signature is valid).
I think it's more readable this way.

arikfr:
Seems like this is abandoned. Once #2904 is implemented it will make implementing this feature much easier.

PR #1865, opened 2017-07-03
Fix bug getting the Salesforce sandbox parameter

msnider:
Needed to set a default value for the boolean parameter.


arikfr:
Thanks!

PR #1870, opened 2017-07-06
Add the option to configure log stream

### LINKED ISSUE #1869: Log lines go to `stderr` and not to `stdout`
### Issue Summary

Running redash (dockerized) is logging everything to `stderr`. I think this is the default for python's `logging` module, although I'm not entirely sure. For me it makes sense that `INFO` level log lines should go into `stdout`, so having some kind of option to configure that would be nice.

### Steps to Reproduce

1. run redash in docker and redirect both `stderr` and `stdout` to separate files:
`docker run --rm redash/redash server 1> ~/out.log 2> ~/error.log`

2. `out.log` file remains empty whereas all the log lines go to `error.log`.

### Technical details:

* Redash Version: 1.0.3
* Browser/OS: -

eyalzek:
by default when using python's `logging` module and a `StreamHandler`,
the stream is directed to `sys.stderr`
(https://docs.python.org/2/library/logging.handlers.html)

By setting the `REDASH_LOG_STREAM` environment variable to `"STDOUT"` we
enable the option to stream the logs to `sys.stdout`. Setting this
configuration option to anything else (or leaving it as default) will
initialize the `StreamHandler` to `sys.stderr` - meaning the original
behavior remains unchanged.

fixes #1869

### DIFF
--- a/redash/settings.py
+++ b/redash/settings.py
@@ -134,6 +134,7 @@ COOKIE_SECRET = os.environ.get("REDASH_COOKIE_SECRET", "c292a0a3aa32397cdb050e23
 SESSION_COOKIE_SECURE = parse_boolean(os.environ.get("REDASH_SESSION_COOKIE_SECURE") or str(ENFORCE_HTTPS))
 
 LOG_LEVEL = os.environ.get("REDASH_LOG_LEVEL", "INFO")
+LOG_STDOUT = parse_boolean(os.environ.get('REDASH_LOG_STDOUT', 'false'))
 
 # Mail settings:
 MAIL_SERVER = os.environ.get('REDASH_MAIL_SERVER', 'localhost')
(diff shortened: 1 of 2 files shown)


redash/settings.py
```diff
@@ -134,6 +134,7 @@ def all_settings():
 SESSION_COOKIE_SECURE = parse_boolean(os.environ.get("REDASH_SESSION_COOKIE_SECURE") or str(ENFORCE_HTTPS))
 
 LOG_LEVEL = os.environ.get("REDASH_LOG_LEVEL", "INFO")
+LOG_STREAM = os.environ.get("REDASH_LOG_STREAM", "STDERR")
```
arikfr: Because it's either `stdout` or `stderr`, let's make it simpler -- how about we change the configuration value to a boolean named `REDASH_LOG_STDOUT`?

eyalzek:
@arikfr that makes sense, it didn't feel exactly right when I was originally writing it :) fixed and rebased

arikfr:
Merged. Thanks!

PR #1871, opened 2017-07-06
docker: make gunicorn worker count configurable

unixwitch:
```
Allow $REDASH_WEB_WORKERS to be set in the environment to change the
number of Gunicorn workers that are started (currently hardcoded to
four).  If not set, the default is four, so this will not affect
existing users at all.

Documentated by example in docker-compose example manifests.
```

Use case: we want to run two copies of the web container for redundancy, but we don't want it to use twice as many resources.  Now we can run two workers per container, giving four in total.

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -19,7 +19,7 @@ scheduler() {
 }
 
 server() {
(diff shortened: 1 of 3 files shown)


arikfr:
Merged. Thanks!

(note the change I added in: 35cb0bc)

PR #1873, opened 2017-07-11
Custom JS code chart improvements

### LINKED ISSUE #1628: Loading custom JS chart into editor fails.
### Issue Summary

Saved custom JS chart visualization does not load in editor.

### Steps to Reproduce

1. Select Custom Plotly chart and select xAxis and a yAxis.
 I chose `select ID, org_id from events LIMIT 10` from Redash metadata.

2. Paste the following to Custom Code.
 Line chart correctly renders.
```
Plotly.plot( element, [{
    x: x,
    y: ys }]);
```

3. Save

4. Click Edit.
 The error is raised and I see default template comments in Custom Code. Am I doing something wrong?
```
angular.js?3437:13920 SyntaxError: Unexpected end of input
    at refresh (eval at <anonymous> (app.d5b5662….js:1196), <anonymous>:501:92)
    at eval (eval at <anonymous> (app.d5b5662….js:1196), <anonymous>:527:9)
    at Scope.$digest (eval at <anonymous> (vendor.d6c0067….js:486), <anonymous>:17524:23)
    at ChildScope.$apply (eval at <anonymous> (vendor.d6c0067….js:486), <anonymous>:17790:24)
    at HTMLButtonElement.eval (eval at <anonymous> (vendor.d6c0067….js:486), <anonymous>:25890:23)
    at defaultHandlerWrapper (eval at <anonymous> (vendor.d6c0067….js:486), <anonymous>:3497:11)
    at HTMLButtonElement.eventHandler (eval at <anonymous> (vendor.d6c0067….js:486), <anonymous>:3485:9)
(anonymous) @ angular.js?3437:13920
(anonymous) @ angular.js?3437:10467
$digest @ angular.js?3437:17542
$apply @ angular.js?3437:17790
(anonymous) @ angular.js?3437:25890
defaultHandlerWrapper @ angular.js?3437:3497
eventHandler @ angular.js?3437:3485
angular.js?3437:13920 SyntaxError: Unexpected end of input
    at refresh
```

### Technical details:

* Redash Version: 1.0.0 pulled from HEAD on Feb. 22nd
* Browser/OS: Chrome
* How did you install Redash: Docker with dev setup

deecay:
This PR introduces some fixes and changes to custom code chart visualization.

1. Fix: Currently, custom js code textarea does not load the saved code properly. This fixes the issue by setting the textarea contents only if the content is undefined.

2. Fix: Any update of the graph in editor causes appended trace. Trace gets added because the plotly graph is not initialized. Now data and calcdata will be initialized by setting them to [].

3. Change: Make textarea vertically resizable for readability and editability of code.

4. Change: Add toggle checkbox for auto-update graph. This is handy when you are dealing with big dataset which takes time to refresh.

This fixes #1628.

### DIFF
--- a/client/app/visualizations/chart/plotly.js
+++ b/client/app/visualizations/chart/plotly.js
@@ -499,6 +499,9 @@ const CustomPlotlyChart = (clientConfig) => {
         return;
       }
       const refresh = () => {
+        // Clear existing data
+        element[0].children[0].calcdata = [];
+        element[0].children[0].data = [];
         // eslint-disable-next-line no-eval
         const codeCall = eval(`codeCall = function(x, ys, element, Plotly){ ${scope.options.customCode} }`);
         codeCall(scope.x, scope.ys, element[0].children[0], Plotly);
--- a/client/app/visualizations/chart/plotly.js
+++ b/client/app/visualizations/chart/plotly.js
@@ -499,9 +499,9 @@ const CustomPlotlyChart = (clientConfig) => {
         return;
       }
       const refresh = () => {
-        // Clear existing data
-        element[0].children[0].calcdata = [];
-        element[0].children[0].data = [];
+        // Clear existing data with blank data for succeeding codeCall adds data to existing plot.
+        Plotly.newPlot(element[0].children[0]);
+
         // eslint-disable-next-line no-eval
         const codeCall = eval(`codeCall = function(x, ys, element, Plotly){ ${scope.options.customCode} }`);
         codeCall(scope.x, scope.ys, element[0].children[0], Plotly);
--- a/client/app/visualizations/chart/plotly.js
+++ b/client/app/visualizations/chart/plotly.js
@@ -514,9 +514,11 @@ const CustomPlotlyChart = (clientConfig) => {
           });
         });
       };
-      scope.$watch('options.customCode', () => {
+      scope.$watch('[options.customCode, options.autoRedraw]', () => {
         try {
-          refresh();
+          if (scope.options.autoRedraw) {
+            refresh();
+          }
         } catch (err) {
           if (scope.options.enableConsoleLogs) {
             // eslint-disable-next-line no-console
--- a/client/app/assets/css/superflat_redash.css
+++ b/client/app/assets/css/superflat_redash.css
@@ -1781,6 +1781,9 @@ fieldset[disabled] .form-control {
 textarea.form-control {
   height: auto;
 }
+textarea.v-resizable {
+  resize: vertical;
+}
 input[type="search"] {
   -webkit-appearance: none;
 }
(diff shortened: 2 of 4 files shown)


client/app/visualizations/chart/plotly.js
```diff
@@ -514,9 +517,11 @@ const CustomPlotlyChart = (clientConfig) => {
           });
         });
       };
-      scope.$watch('options.customCode', () => {
+      scope.$watch('[options.customCode, options.autoRedraw]', () => {
```
arikfr: Why do we need to watch `autoRedraw`? If it's disabled, on next refresh it won't redraw. If it's enabled, still we have nothing to do.

client/app/visualizations/chart/plotly.js
```diff
@@ -499,6 +499,9 @@ const CustomPlotlyChart = (clientConfig) => {
         return;
       }
       const refresh = () => {
+        // Clear existing data
+        element[0].children[0].calcdata = [];
+        element[0].children[0].data = [];
```
arikfr: Where is this used?

deecay: This piece is related to 

> 2. Fix: Any update of the graph in editor causes appended trace.

and used by

```
        const codeCall = eval(`codeCall = function(x, ys, element, Plotly){ ${scope.options.customCode} }`);
        codeCall(scope.x, scope.ys, element[0].children[0], Plotly);
```

The code above appends a trace to an already existing plot. So without clearing `data` from the already drawn plotly graph, every update in code editor will cause the trace in the graph to multiply. One-trace bar graph becomes two-trace bar graph, three-trace bar graph, and so on, each time you type in code editor.

Using Plotly.newPlot() might be another option to avoid appending, but haven't tried it yet.

client/app/visualizations/chart/plotly.js
deecay: I thought so too. But I realized that I would want the graph to redraw 'when I ticked the auto-redraw'. Without this watch, only the next change in code will update the graph. 

I think, writing a lengthy code without autoRadraw, then placing a tick on autoRedraw, should trigger the graph redraw immediately.

client/app/visualizations/chart/plotly.js
deecay: Please see https://github.com/getredash/redash/pull/1873/commits/1a8078ab0345108ef91ca5a17835b3c6816f15ba .

This less hackish way fixes the problem too.

client/app/visualizations/chart/plotly.js
arikfr: Then why not just watch `option.autoRedraw`?

deecay: When I am editing a small script, I would like to refresh graph as I type. This is done by watching `customCode`.
When I am editing a big script, I would like to turn off the refresh-as-I-type, but when I click on the auto redraw checkbox, I want so see the graph refreshed immediately. This last part is done by watching `autoRedraw`.

arikfr:
Thanks!

PR #1874, opened 2017-07-12
Setting optional limit for done tasks

alexpekurovsky:
This PR provides ability to specify `?limit=` query to `/api/admin/queries/tasks` API and display requested amount of tasks in `done` state

### DIFF
--- a/redash/handlers/admin.py
+++ b/redash/handlers/admin.py
@@ -1,5 +1,6 @@
 import json
 
+from flask import request
 from flask_login import login_required
 from redash import models, redis_connection
 from redash.handlers import routes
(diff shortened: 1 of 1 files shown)


redash/handlers/admin.py
```diff
@@ -32,9 +32,10 @@ def outdated_queries():
 @require_super_admin
 @login_required
 def queries_tasks():
+    limit = request.args.get('limit') or 50
     waiting = QueryTaskTracker.all(QueryTaskTracker.WAITING_LIST)
     in_progress = QueryTaskTracker.all(QueryTaskTracker.IN_PROGRESS_LIST)
-    done = QueryTaskTracker.all(QueryTaskTracker.DONE_LIST, limit=50)
+    done = QueryTaskTracker.all(QueryTaskTracker.DONE_LIST, limit=limit)
```
arikfr: Maybe we should apply the limit to all of them?

alexpekurovsky: Implemented

alexpekurovsky:
@arikfr I added some improvements. Idea behind this code is following:
- Global limit is 50 by default. Can be overwritten by `?limit=` query
- Limit for waiting/done/in_progress tasks is None by default. Can be configured by setting `?waiting_limit=`/`?done_limit=`/`progress_limit=` respectively.
- If no limit for specific task state is provided, global limit is used

redash/handlers/admin.py
```diff
@@ -32,9 +33,14 @@ def outdated_queries():
 @require_super_admin
 @login_required
 def queries_tasks():
-    waiting = QueryTaskTracker.all(QueryTaskTracker.WAITING_LIST)
-    in_progress = QueryTaskTracker.all(QueryTaskTracker.IN_PROGRESS_LIST)
-    done = QueryTaskTracker.all(QueryTaskTracker.DONE_LIST, limit=50)
+    global_limit = int(request.args.get('limit')) if request.args.get('limit') is not None else 50
+    waiting_limit = int(request.args.get('waiting_limit')) if request.args.get('waiting_limit') is not None else None
+    progress_limit = int(request.args.get('progress_limit')) if request.args.get('progress_limit') is not None else None
+    done_limit = int(request.args.get('done_limit')) if request.args.get('done_limit') is not None else None
```
arikfr: `{}.get` can take a fallback value if the key isn't present, so I think you can simplify this with:

```python
global_limit = int(request.args.get('limit', 50))
waiting_limit = int(request.args.get('waiting_limit', global_limit))
progress_limit = int(request.args.get('progress_limit', global_limit))
done_limit = int(request.args.get('done_limit', global_limit))
```

And then you can pass the different limits as is without the `x or global_limit` part.

alexpekurovsky:
Your code really looks much better

arikfr:
Merged. Thanks!

PR #1876, opened 2017-07-12
Add SSL configuration option for PostgreSQL

TylerBrock:
This adds the `sslmode` configuration option for the Postgres data source.

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -70,6 +70,11 @@ class PostgreSQL(BaseSQLQueryRunner):
                 "dbname": {
                     "type": "string",
                     "title": "Database Name"
+                },
+                "sslmode": {
+                   "type": "string",
+                   "title": "SSL Mode",
+                   "default": "prefer"
                 }
             },
             "order": ['host', 'port', 'user', 'password'],


redash/query_runner/pg.py
```diff
@@ -70,6 +70,11 @@ def configuration_schema(cls):
                 "dbname": {
                     "type": "string",
                     "title": "Database Name"
+                },
+                "sslmode": {
+                   "type": "string",
+                   "title": "SSL Mode",
+                   "default": "require"
```
miketheman: Should the default mode be `require` without a semi-loud doc update? I know this would be fairly confusing if I didn't have ssl server side. What about `prefer`?

TylerBrock: Meh. It doesn't modify any existing connections just sets the defaults for new ones. I can change it to prefer though.

TylerBrock: Done.

TylerBrock:
Bump @arikfr can we get this one merged?

arikfr:
Thanks 👍 Will be merged right after the v2 release.

arikfr:
Thanks! Sorry for the long wait.

PR #1884, opened 2017-07-18
Add link to query page from admin view

miketheman:
When we have a long0running query, it is useful to look up what that
query definition is. Having a link directly from the admin page cuts
down on the manual step of opying the query id and pasting that into the
address bar.

Signed-off-by: Mike Fiedler <miketheman@gmail.com>


arikfr:
Thanks 👍

PR #1886, opened 2017-07-20
WIP: Counter value formatting

deecay:
This PR is still WIP, and I seek for your advise around handling locales.

I am trying to add a feature to specify locale for counter value. I am planning to use [`angular-dynamic-locale`](https://github.com/lgalfaso/angular-dynamic-locale) to change locale at runtime. However I have been facing two issues.

counter
counter-editor

1. I cannot mix locales in one dashboard.
When I try to place multiple counter widgets with multiple locales in one dashboard, locales for counter widgets are not displayed correctly. This is due to the nature of angular-dynamic-locale, where `tmhDynamicLocale.set()` sets locales asynchronously. This is fine in individual visualization and counter-editor, but loading multiple counter widgets with different locales in dashboard results in 'single locale for all widgets'. For example, if default locale was en-US:
* USD widget loads => USD currency shown as USD => EUR widget loads => EUR currency shown as USD => locale set to USD => locale set to EUR
　
This will not become a problem if a user sticks to 'one locale per dashboard', but would like to do something to allow one locale per widget. Do you have any idea how to handle this?

2. How to handle installation of angular-i18n.
Running `npm install -S angular-i18n` places locale javascript files under node_modules/angular-i18n/. The files in this directory has to be moved somehow to client/app folder to be able to be seen by angular-dynamic-locale via HTTP get. For now I create `locales` folder under client/app and copy the locale files manually. Does `mkdir locales && cp  node_modules/angular-i18n/* locales` in Dockerfile make sense to you for installation? Is it too much to copy all 801 locale files? Is this against the Webpack way?

### DIFF
--- a/client/app/index.js
+++ b/client/app/index.js
@@ -26,6 +26,7 @@ import 'brace';
 import 'angular-ui-ace';
 import 'angular-resizable';
 import ngGridster from 'angular-gridster';
+import 'angular-dynamic-locale/dist/tmhDynamicLocale';
 import { each } from 'underscore';
 
 import './sortable';
@@ -42,12 +43,14 @@ import registerDirectives from './directives';
 import registerVisualizations from './visualizations';
 import markdownFilter from './filters/markdown';
 import dateTimeFilter from './filters/datetime';
+import customCurrencyFilter from './filters/customCurrency';
+import customNumberFilter from './filters/customNumber';
 
 const logger = debug('redash');
 
 const requirements = [
   ngRoute, ngResource, ngSanitize, uiBootstrap, ngMessages, uiSelect, 'angularMoment', toastr, 'ui.ace',
-  ngUpload, 'angularResizable', vsRepeat, 'ui.sortable', ngGridster.name,
+  ngUpload, 'angularResizable', vsRepeat, 'ui.sortable', ngGridster.name, 'tmh.dynamicLocale',
 ];
 
 const ngModule = angular.module('app', requirements);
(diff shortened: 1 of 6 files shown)


arikfr:
Wouldn't it be easier to change the counter visualization to support strings, and then we can just let the user format it with a query?

Like, `SELECT '$' || value`?

deecay:
I was thinking of extending counter visualization in the future so that it can have coloring such as "less is better" and "closer to certain value is better". There, it would be better to have number as a value, so that we can do number comparison like target >= value. Otherwise one would have to define the target string like '$1,234.00' to be able to color the value red or green.

However, we can always have 2 columns in the resultset, one column a pure number, and one column a formatted string for counter display. Then use number column as target values.

If you agree, I will prepare the PR with string support.

deecay:
I am now thinking of the following.

   * Counter visualization can show string value. Use it to show numbers that are already formatted with SQL.
   * Counter visualization can format numbers. 1000 => "1000", 1000 => "1,000", 1234.56 => "1.234,56"
   * Counter visualization numbers can have prefix or suffix. 100 => "$100", 99 => "99 EUR", 99.9 => "99.9%", 99 => "99sheeps"

counter-editor
counter



Is this method better?

Please see https://github.com/deecay/redash/commit/c2782098839e5a586b641d957ef86b80ffaaa207 for possible PR.

arikfr:
It is better 👍 It will make it even better if all the formatting stuff go into a tab of their own to avoid confusion :)

deecay:
Thank you for the input. Closing now.

Issue #1888, opened 2017-07-20
BUG: Alerts can't be editted


alison985:
Please see https://github.com/mozilla/redash/issues/139 for details. I am opening this issue so I can reference it in a PR with the fix.

### Steps to Reproduce

1. Go to Alerts list page.
2. Click on an alert.

You expect to be taken to the alerts specific page, but instead you stay on the alerts list page.

### Technical details:

* Redash Version: 2.0.0
* Browser/OS: Chrome, Mac
* How did you install Redash: Docker Compose

alison985:
https://github.com/mozilla/redash/commit/e7ddfcdfb3b5901bf0342032db613cbf2c792de7

The id for the alert isn't currently being passed so the URL can't be properly made on the list page.

arikfr:
Fixed in #1889.

PR #1889, opened 2017-07-21
Send alert ID to UI

alison985:
This will allow alert detail page links on the alert list page to work again.

Fixes bug #1888


arikfr:
Thanks!

Issue #1572, opened 2017-02-08
[Presto Datasource] Use user's login mail address instead of Datasource's username for Impersonating 


zz22394:
Dear team,

We are using  Re:dash 0.12.0+b2449 against Presto 157t.

### Issue Summary
We found when redash run queries on Presto , it sends SQLs to Presto,
with Datasource's username.
How about using  user's login mail address instead of Datasource's username for Impersonating?

### Steps to Reproduce

1. Using Datasource's username:
We can see like this:
![datasource_username_in_redash](https://cloud.githubusercontent.com/assets/516268/22720698/af37ca6a-edef-11e6-9e40-6a00fd99f003.png)
Display in Presto:
![before](https://cloud.githubusercontent.com/assets/516268/22720701/b39e3b5c-edef-11e6-8692-e6dbbf45c8ec.png)

2. Use user's login mail address for Impersonating:
<img width="1197" alt="after" src="https://cloud.githubusercontent.com/assets/516268/22720726/e8b3166e-edef-11e6-8cb1-692c0b537c0b.png">
I Changed presto.py and tasks/queries.py ,
![image](https://cloud.githubusercontent.com/assets/516268/22720744/066809c6-edf0-11e6-82b1-e5653b569ae7.png)

yukimaeno:
+1

gppeixoto:
+1 and thank you for sharing your solution. We have a local deployment of Redash and are working on managing resource and queues of Presto queries on an user level -- this would not be feasible without changing the default user

Ralnoc:
I have opened PR #3723 that resolves this issue and allow for a default user to still be assigned without conflict.

* Username: It uses the configured field for username over all else, then uses the email address to extract the username based on the left side of the `@` symbol in the email address and finally, if all of that is missing, defaults to `redash`.
* Password: is pulled from a new password field in the configuration page for Presto
* Source Identifier: It changes the `source` identifier for `pyhive` so that instead of reporting itself as `pyhive` to Presto, it declares itself as `redash`.

PR #1892, opened 2017-07-25
Change user name to login mail address for Impersonating (presto)

zz22394:
When redash run queries on Presto , it sends SQLs to Presto,
with Datasource's username.

This PR will use user's login mail address instead of Datasource's username for Impersonating,
to make it more clear that who ran the query on presto.
(#1572)

### DIFF
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -95,7 +95,7 @@ class Presto(BaseQueryRunner):
         return schema.values()
 
     def run_query(self, query, user):
-        if user==None:
+        if user is None:
             param_username = self.configuration.get('username', 'redash')
         else:
             param_username = user.email
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -86,7 +86,7 @@ class Presto(BaseQueryRunner):
 
         for row in results['rows']:
             table_name = '{}.{}'.format(row['table_schema'], row['table_name'])
-            
+
             if table_name not in schema:
                 schema[table_name] = {'name': table_name, 'columns': []}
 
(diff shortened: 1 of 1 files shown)


arikfr:
(and apologies for the long time it took to comment...)

zz22394:
image

@arikfr Thanks for your comment.
Acoording to your comment, I will add an checkbox , when adding a new presto datasource.

With this, we can make this behavior optional.
How do you think about this change?

HimanshPal:
@arikfr - Would be awesome if this PR can be merged into Master As we are in need to of this use-case as to restrict users on Presto bases on certain limits

bdgeise:
Is it possible to get this merged for the next release?  We really could use this functionality.

arikfr:
Hi 👋 

While it's a small change, there is a bigger picture to consider: Redash caches the last query execution result. This means it will cache the last execution of whatever user who executed it last, and hence might leak data to other users while giving the admin a false sense of protection.

I feel that we can't introduce this change as is without addressing the caching issue. 

Of course I might be wrong here and misunderstand the use case, so I'm happy to hear your feedback.

Thanks.

bdgeise:
That @arikfr that makes sense and is my ignorance into the complexities involved in the entire project.

Currently, for our specific use case probably not an issue and more around having controls in place to limit problematic DDL that could be issued on specific schemas that users should only have read only.  Managing permissions from the Presto side scales better because we can also have users using other tools to access the cluster.

Can you point me to the cache code?  Maybe I can try to help out so this PR can be merged in the near future.

arikfr:
> Currently, for our specific use case probably not an issue and more around having controls in place to limit problematic DDL that could be issued on specific schemas that users should only have read only. 

Why not create the data source with a Presto user that has read only access?

> Can you point me to the cache code? Maybe I can try to help out so this PR can be merged in the near future.

I appreciate the will to help, but caching is an integral part of how Redash works right now. Unless you have days to spend on this, I doubt a quick win is possible.

(I'm closing this PR now, but you're welcome to keep the discussion going)

PR #1895, opened 2017-07-27
add data_scanned to Athena

alison985:
Athena costs are based on how much data is scanned by the query. This PR leverages the PyAthena library functionality to report that information to the user. It also saves it to the database so it can be analyzed later.


redash/models.py
```diff
@@ -617,18 +617,25 @@ class QueryResult(db.Model, BelongsToOrgMixin):
...
     retrieved_at = Column(db.DateTime(True))
+    data_scanned = Column(db.String(255), nullable=True)
 
     __tablename__ = 'query_results'
 
     def to_dict(self):
+        if hasattr(self, 'data_scanned'):
+            data_scanned_info = self.data_scanned
+        else:
+            data_scanned_info = 'to_dict'
```
arikfr: Why `to_dict`? 

Btw, this `if` check will be always true...

redash/models.py
```diff
@@ -663,13 +670,20 @@ def get_latest(cls, data_source, query, max_age=0):
 
     @classmethod
     def store_result(cls, org, data_source, query_hash, query, data, run_time, retrieved_at):
+        try:
+            data_scanned_information = json.loads(data)['data_scanned']
+        except (ValueError, TypeError) as e:
```
arikfr: You can use `json.loads(data).get('data_scanned', 'error')` which will return `None` if it has no value. No need for the `try`/`except`.

But why make every other query runner return some value here when this is probably relevant for 2-3 data sources at the moment? 

I think a simpler approach will be to return nothing when it's not supported, and just not show any value in the UI.

---

Another issue is that calling `json.loads` here can be a performance/memory issue for large datasets. I have a different approach to suggest, but will outline it in the review summary.

redash/query_runner/athena.py
```diff
@@ -149,7 +149,13 @@ def run_query(self, query, user):
             column_tuples = [(i[0], _TYPE_MAPPINGS.get(i[1], None)) for i in cursor.description]
             columns = self.fetch_columns(column_tuples)
             rows = [dict(zip(([c['name'] for c in columns]), r)) for i, r in enumerate(cursor.fetchall())]
-            data = {'columns': columns, 'rows': rows}
+            qbytes = 'N/A'
+            try:
+                cursor._poll()
+                qbytes = cursor._data_scanned_in_bytes
+            except AttributeError as e:
+                debug("Athena Upstream can't get data_scanned_in_bytes: %s", e)
```
arikfr: What's `debug`?

rohanpd:
I'd love to see something like this implemented. At the moment we see users scanning lots of data from Athena hiking up our bills without them even knowing it.

arikfr:
Replaced by #2204.

PR #1896, opened 2017-07-27
Textbox editing fix

alison985:
Combines mozilla/redash PR’s 86 and 95.

There was a bug that saved textbox content on a dashboard when you
tried to close without saving. This fixes it.

### DIFF
--- a/client/app/pages/dashboards/widget.js
+++ b/client/app/pages/dashboards/widget.js
@@ -8,7 +8,7 @@ const EditTextBoxComponent = {
     close: '&',
     dismiss: '&',
   },
-  controller(toastr) {
+  controller($rootScope, $location, $http, toastr) {
     'ngInject';
 
     this.saveInProgress = false;
@@ -23,6 +23,10 @@ const EditTextBoxComponent = {
         this.saveInProgress = false;
       });
     };
+    this.closeWithoutSave = () => {
+      this.widget.text = this.widget.existing_text;
+      this.close();
+    };
   },
 };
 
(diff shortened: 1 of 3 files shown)


client/app/pages/dashboards/widget.js
```diff
@@ -8,7 +8,7 @@ const EditTextBoxComponent = {
     close: '&',
     dismiss: '&',
   },
-  controller(toastr) {
+  controller($rootScope, $location, $http, toastr) {
```
arikfr: Why the extra injects are needed?

client/app/pages/dashboards/widget.js
```diff
@@ -23,18 +23,25 @@ const EditTextBoxComponent = {
...
+      this.close();
+    };
   },
 };
 
 function DashboardWidgetCtrl($location, $uibModal, $window, Events, currentUser) {
   this.canViewQuery = currentUser.hasPermission('view_query');
 
   this.editTextBox = () => {
+    this.widget.existing_text = this.widget.text;
```
arikfr: I think it will be better to keep this logic in the component. So the component will make a copy of the original value and restore to it when closed instead of saved.

Another option is to use a different object as the `ng-model` of the `<textarea>` and only on save assign it to the widget. This way no extra logic is needed on close and you can bring back keyboard support.

client/app/pages/dashboards/widget.js
alison985: They aren't. Good eye, thanks.

alison985:
@arikfr  Updated.

arikfr:
Thanks!

PR #1897, opened 2017-07-27
Fix google auth domain message

alison985:
Fixes https://github.com/mozilla/redash/issues/79

### DIFF
--- a/redash/authentication/google_oauth.py
+++ b/redash/authentication/google_oauth.py
@@ -88,7 +88,17 @@ def login():
     next_path = request.args.get('next', url_for("redash.index", org_slug=session.get('org_slug')))
     logger.debug("Callback url: %s", callback)
     logger.debug("Next is: %s", next_path)
-    return google_remote_app().authorize(callback=callback, state=next_path)
+    extra = {}
+    if 'org_slug' in session:
+        org = models.Organization.get_by_slug(session.pop('org_slug'))
+    else:
+        org = current_org
+    if org.google_apps_domains:
+        extra['hd'] = org.google_apps_domains[0]
+    if session.get('relogin') == '1':
+        extra['prompt'] = 'consent'
+        session['relogin'] = '0'
+    return google_remote_app().authorize(callback=callback, state=next_path, **extra)
 
 
 @blueprint.route('/oauth/google_callback', endpoint="callback")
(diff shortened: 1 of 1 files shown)


redash/authentication/google_oauth.py
```diff
@@ -88,7 +88,17 @@ def login():
...
     logger.debug("Callback url: %s", callback)
     logger.debug("Next is: %s", next_path)
-    return google_remote_app().authorize(callback=callback, state=next_path)
+    extra = {}
+    if 'org_slug' in session:
+        org = models.Organization.get_by_slug(session.pop('org_slug'))
+    else:
+        org = current_org
+    if org.google_apps_domains:
+        extra['hd'] = org.google_apps_domains[0]
```
arikfr: This breaks support for multiple Google Apps domains. See #1924 for relevant discussion.

redash/authentication/google_oauth.py
```diff
@@ -113,7 +123,7 @@ def authorized():
 
     if not verify_profile(org, profile):
         logger.warning("User tried to login with unauthorized domain name: %s (org: %s)", profile['email'], org)
-        flash("Your Google Apps account ({}) isn't allowed.".format(profile['email']))
+        session['relogin'] = '1'
         return redirect(url_for('redash.login', org_slug=org.slug))
```
arikfr: Instead of relying on a session variable, let's pass a query string option to the `redash.login` route, which will tell it to ask the user to pick a new account (I guess this is what `extra['prompt'] = 'consent'` does).

jezdez: @alison985 What he means is that we can pass a query parameter to the redirected URL here (Flask will append any param that isn't matching a route path parameter as a query parameter instead), e.g.:

```python
# will turn into /user/login?prompt=consent
return redirect(url_for('redash.login', org_slug=org.slug, force='true'))
```

Then in the `handlers/authentication.py:login` function when it creates the `google_auth_url` variable using the `get_google_auth_url` function we can pass the `force` value along as a query parameter to the `google_oauth.authorize_org` or `google_oauth.authorize` view.

Then in the specific Google login view instead of checking the session check the query parmas:

```python
if 'force' in request.args:
    extra['prompt'] = 'consent'
```

redash/authentication/google_oauth.py
jezdez: Hm, I'm not sure how we could solve this other than hardcoding the hosted domain to chose here using a new setting? What do you think @arikfr?

PR #1900, opened 2017-07-31
Change: pivottable design to match other components

deecay:
This PR changes css for pivot table to make it blend in better with other components.

pivot-css

Too monotonous?


arikfr:
How does it look before the change?

deecay:
before the change:
snapcrab_noname_2018-1-26_16-51-38_no-00

I hope this, or some kind of similar design alignment, be a part of recent design renewal.

arikfr:
Replaced by #2264.

Thanks, @deecay for your effort here. 🙏

PR #1906, opened 2017-08-03
Query Runner for Azure SQL Data Warehouse

kitsuyui:
- This is almost copied from mssql.py.
- Microsoft's driver installation is here: https://www.microsoft.com/en-us/sql-server/developer-get-started/node/ubuntu/

https://discuss.redash.io/t/support-for-azure-sql-data-warehouse/980

### DIFF
--- a/redash/query_runner/azure_sql_data_warehouse.py
+++ b/redash/query_runner/azure_sql_data_warehouse.py
@@ -55,11 +55,6 @@ class SQLServerODBC(BaseSQLQueryRunner):
                     "type": "number",
                     "default": 1433
                 },
-                "tds_version": {
-                    "type": "string",
-                    "default": "7.0",
-                    "title": "TDS Version"
-                },
                 "charset": {
                     "type": "string",
                     "default": "UTF-8",
@@ -136,7 +131,6 @@ class SQLServerODBC(BaseSQLQueryRunner):
             password = self.configuration.get('password', '')
             db = self.configuration['db']
             port = self.configuration.get('port', 1433)
-            tds_version = self.configuration.get('tds_version', '7.0')
             charset = self.configuration.get('charset', 'UTF-8')
             driver = self.configuration.get('driver', '{ODBC Driver 13 for SQL Server}')
 
--- a/redash/query_runner/azure_sql_data_warehouse.py
+++ b/redash/query_runner/azure_sql_data_warehouse.py
@@ -33,7 +33,7 @@ class MSSQLJSONEncoder(JSONEncoder):
         return super(MSSQLJSONEncoder, self).default(o)
 
 
-class SqlServer(BaseSQLQueryRunner):
+class SQLServerODBC(BaseSQLQueryRunner):
     noop_query = "SELECT 1"
 
     @classmethod
--- a/redash/query_runner/azure_sql_data_warehouse.py
+++ b/redash/query_runner/azure_sql_data_warehouse.py
@@ -1,0 +1,192 @@
+import json
+import logging
+import sys
+import uuid
+
+from redash.query_runner import *
+from redash.utils import JSONEncoder
+
+logger = logging.getLogger(__name__)
+
+try:
+    import pyodbc
+    enabled = True
+except ImportError:
+    enabled = False
+
+# from _mssql.pyx ## DB-API type definitions & http://www.freetds.org/tds.html#types ##
+types_map = {
+    1: TYPE_STRING,
+    2: TYPE_BOOLEAN,
+    # Type #3 supposed to be an integer, but in some cases decimals are returned
+    # with this type. To be on safe side, marking it as float.
+    3: TYPE_FLOAT,
+    4: TYPE_DATETIME,
+    5: TYPE_FLOAT,
+}
+
+
+class MSSQLJSONEncoder(JSONEncoder):
+    def default(self, o):
+        if isinstance(o, uuid.UUID):
+            return str(o)
+        return super(MSSQLJSONEncoder, self).default(o)
+
+
+class SqlServer(BaseSQLQueryRunner):
+    noop_query = "SELECT 1"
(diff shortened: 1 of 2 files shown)


arikfr:
Thanks!

> This is almost copied from mssql.py.

What are the differences then?

kitsuyui:
The main difference is changing driver to pyodbc from pymssql.

Because pymssql does not adapted to Azure Data Warehouse yet.
Some users disruptions can see [in this forum](https://social.msdn.microsoft.com/Forums/SqlServer/en-US/4bfb4b20-440f-4c87-93c0-1442946c47a0/connect-data-warehouse-from-python?forum=ssdsgetstarted).
(I tried some ways to understanding. But I could't detect what happened its inside finally.)

For now, the only way to solve is to use pyodbc.
Fortunately its type definitions are almost same.
So I changed simply about connection and name only.
The "almost copied" means that.

kitsuyui:
Hi, @arikfr .
Does need to brush up source codes?
Or are more descriptions needed?

redash/query_runner/azure_sql_data_warehouse.py
```diff
@@ -0,0 +1,192 @@
...
+                },
+                "server": {
+                    "type": "string",
+                    "default": "127.0.0.1"
+                },
+                "port": {
+                    "type": "number",
+                    "default": 1433
+                },
+                "tds_version": {
```
arikfr: This is no longer needed.

redash/query_runner/azure_sql_data_warehouse.py
```diff
@@ -0,0 +1,192 @@
...
+    def run_query(self, query, user):
+        connection = None
+
+        try:
+            server = self.configuration.get('server', '')
+            user = self.configuration.get('user', '')
+            password = self.configuration.get('password', '')
+            db = self.configuration['db']
+            port = self.configuration.get('port', 1433)
+            tds_version = self.configuration.get('tds_version', '7.0')
```
arikfr: Same as above (no longer needed).

redash/query_runner/azure_sql_data_warehouse.py
```diff
@@ -0,0 +1,192 @@
...
+                    "default": "UTF-8",
+                    "title": "Character Set"
+                },
+                "db": {
+                    "type": "string",
+                    "title": "Database Name"
+                },
+                "driver": {
+                    "type": "string",
+                    "title": "Driver Identifier"
```
arikfr: Worth adding the default value here, to avoid confusing people.

redash/query_runner/azure_sql_data_warehouse.py
```diff
@@ -0,0 +1,192 @@
...
+    4: TYPE_DATETIME,
+    5: TYPE_FLOAT,
+}
+
+
+class MSSQLJSONEncoder(JSONEncoder):
+    def default(self, o):
+        if isinstance(o, uuid.UUID):
+            return str(o)
+        return super(MSSQLJSONEncoder, self).default(o)
```
arikfr: You can import the duplicated code:

```
from redash.query_runner.mssql import MSSQLJSONEncoder, types_map
```

redash/query_runner/azure_sql_data_warehouse.py
```diff
@@ -0,0 +1,192 @@
...
+# from _mssql.pyx ## DB-API type definitions & http://www.freetds.org/tds.html#types ##
+types_map = {
+    1: TYPE_STRING,
+    2: TYPE_BOOLEAN,
+    # Type #3 supposed to be an integer, but in some cases decimals are returned
+    # with this type. To be on safe side, marking it as float.
+    3: TYPE_FLOAT,
+    4: TYPE_DATETIME,
+    5: TYPE_FLOAT,
+}
```
arikfr: Are you sure the types_map stays the same? It usually changes between DB-API drivers.

redash/query_runner/azure_sql_data_warehouse.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class MSSQLJSONEncoder(JSONEncoder):
+    def default(self, o):
+        if isinstance(o, uuid.UUID):
+            return str(o)
+        return super(MSSQLJSONEncoder, self).default(o)
+
+
+class SqlServer(BaseSQLQueryRunner):
```
arikfr: Worth changing the class name, just in case (`SQLServerODBC` maybe?).

redash/query_runner/azure_sql_data_warehouse.py
```diff
@@ -0,0 +1,192 @@
...
+            "secret": ["password"]
+        }
+
+    @classmethod
+    def enabled(cls):
+        return enabled
+
+    @classmethod
+    def name(cls):
+        return "Azure SQL Data Warehouse"
```
arikfr: Maybe we should call it "Microsoft SQL Server (ODBC)"?

I saw weird issues with pymssql and Azure (not only Data Warehouse), so I'm thinking of deprecating the other connector and using this one for all SQL Server connections in the future.

kitsuyui:
@arikfr Thank you so much! I'll fix it.

redash/query_runner/azure_sql_data_warehouse.py
kitsuyui: It seems to be same.
I had tested both pymssql and pyodbc.
Integers, double, varchar, nvarchar, bigint, and UUID.
And it worked.

So I will use imports from  mssql.

kitsuyui:
@arikfr Please re-review. I did it!

kitsuyui:
@arikfr 
I've fixed. Please re-review.

And I'm sorry that it is becoming many revising commits.
Is it OK? Or should I rebase it?

arikfr:
Thanks!

PR #1912, opened 2017-08-06
Snowflake support is no longer enabled by default

arikfr:
We don't install snowflake connector by default, as it's causing conflicts with
other packages. To properly support it we probably need to switch from pyOpenSSL
to the other package snowflake is using (that's compatible with it).


jacopoch:
Maybe we should think about fixing this issue, Snowflake is getting a lot of traction now and removing this seemed a bit arbitrary 👍

PR #1920, opened 2017-08-08
Counter value string formatting

deecay:
This PR lets you define various counter value formatting by introducing new tab "FORMAT" for counter visualization editor.

counter

counter-editor


arikfr:
Thanks!

PR #1925, opened 2017-08-09
Fix "custom visualization" code box behavior

RoyWiggins:
Only set the code box to the default string if it doesn't already have code entered already. It was blowing away the previously-entered code on load.


arikfr:
Thanks, but I believe this was fixed in #1873. 

(please reopen if not)

PR #1928, opened 2017-08-14
Feature/query based parameter

rohithmenon:
Feature: Query based parameter (drop-down)
=================================
The feature is useful for our use-cases where parameters tend to be dynamic and comes from other queries. We have couple of use-cases where we will need such support. This is a feature available in a comparable product offering from Mode analytics (https://help.modeanalytics.com/articles/using-dynamic-parameters/).

The implementation is basically to use a saved query within redash as a parameter. Restriction for this query is that only first column will be used for options of the parameter and it must of a string format. By using saved queries for parameter, we get all the goodies of redash queries (caching, refreshing etc.)

### DIFF
--- a/client/app/components/parameters.html
+++ b/client/app/components/parameters.html
@@ -19,9 +19,8 @@
           <option ng-repeat="option in extractEnumOptions(param.enumOptions)" value="{{option}}">{{option}}</option>
         </select>
       </span>
-      <span ng-switch-when="query" ng-controller="QueryBasedParameterController">
-        <select ng-model="param.value" class="form-control" ng-options="queryResult for queryResult in queryResults">
-        </select>
+      <span ng-switch-when="query">
+        <query-based-parameter param="param"></query-based-parameter>
       </span>
       <input ng-switch-default type="{{param.type}}" class="form-control" ng-model="param.ngModel">
     </span>
--- a/client/app/components/parameters.html
+++ b/client/app/components/parameters.html
@@ -19,7 +19,11 @@
           <option ng-repeat="option in extractEnumOptions(param.enumOptions)" value="{{option}}">{{option}}</option>
         </select>
       </span>
+      <span ng-switch-when="query" ng-controller="QueryBasedParameterController">
+        <select ng-model="param.value" class="form-control" ng-options="queryResult for queryResult in queryResults">
+        </select>
+      </span>
       <input ng-switch-default type="{{param.type}}" class="form-control" ng-model="param.ngModel">
     </span>
-  </div>
+  </
 </div>
--- a/client/app/components/parameters.html
+++ b/client/app/components/parameters.html
@@ -25,5 +25,5 @@
       </span>
       <input ng-switch-default type="{{param.type}}" class="form-control" ng-model="param.ngModel">
     </span>
-  </
+  </div>
 </div>
--- a/client/app/components/parameters.js
+++ b/client/app/components/parameters.js
@@ -1,5 +1,14 @@
 import template from './parameters.html';
+import queryBasedParameterTemplate from './query-based-parameter.html';
 import parameterSettingsTemplate from './parameter-settings.html';
+
+const QueryBasedOption = (query) => {
+  const queryBasedOption = {
+    id: query.id,
+    name: query.name,
+  };
+  return queryBasedOption;
+};
 
 const ParameterSettingsComponent = {
   template: parameterSettingsTemplate,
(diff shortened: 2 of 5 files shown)


client/app/components/parameters.js
```diff
@@ -8,13 +8,45 @@ const ParameterSettingsComponent = {
...
     dismiss: '&',
   },
-  controller() {
+  controller($sce, Query) {
     'ngInject';
 
+    this.trustAsHtml = html => $sce.trustAsHtml(html);
     this.parameter = this.resolve.parameter;
+    this.onQuerySelect = (query) => {
+      this.parameter.query = query;
```
arikfr: Wouldn't it be more efficient to save only the `query.id` here?

client/app/components/parameters.html
```diff
@@ -19,7 +19,11 @@
...
         </select>
       </span>
+      <span ng-switch-when="query" ng-controller="QueryBasedParameterController">
+        <select ng-model="param.value" class="form-control" ng-options="queryResult for queryResult in queryResults">
+        </select>
+      </span>
       <input ng-switch-default type="{{param.type}}" class="form-control" ng-model="param.ngModel">
     </span>
-  </div>
+  </
```
arikfr: deleted by mistake?

awm33:
OMG yes! Would love this feature :)

client/app/components/parameters.js
rohithmenon: First of all, thanks for review. I am totally new to frontend stuff, and this seemed like something we needed and hence gave it a shot. My knowledge of angular is as much as w3school, stackoverflow and reading redash code.

Regarding the query.id, I was assuming query.id would be sufficient, but I was not sure if the watch on paramter would trigger when the query changes.

client/app/components/parameters.html
rohithmenon: Good catch. Not sure how to catch this in dev. It did not break stuff when I deployed it locally. Are there some good tricks? I realize that frontend does not have good test coverage, but are there tools that I can use?

rohithmenon:
> convert the use of controller into a separate component. I'm not sure how much you're experienced with Angular or how much time you have... 
I can do that. Like mentioned before controller examples I saw on the web used ng-controller. I thought about refactoring it into a component with template for query-based dropdown, but did not do that because it was pretty small and contained within the scope of parameter-settings. That being said, I don't know when to choose between the two. Let me know what you think.

> will you have the time to do this or prefer that I merge it as is?
I should be able to convert it into a component. There are enough examples of code snippets within redash. Let me know of your thoughts for components.

client/app/components/parameters.js
arikfr: 👍 

I think the putting only the id should be enough for the watch, if you encounter issues -- let me know and I will give a look.

client/app/components/parameters.html
arikfr: Because it works, tests might not help here either. But there might be a linter for HTML we can add to the build process.

rohithmenon:
Fair enough. I will try to make query-dropdown a component.
--
About heuristics, I did not follow the name/value part. I got the idea of dropdown from values and to check for a column named value if there are more than one columns.

The feature looks pretty similar to dropdown values except that instead of manually entering values in the textarea, it gets it from the output of a query. The concept of name is not very clear to me.

rohithmenon:
Parameter Settings
image

---

Dropdown
image

arikfr:
> About heuristics, I did not follow the name/value part. I got the idea of dropdown from values and to check for a column named value if there are more than one columns.

The idea is that sometimes you would like to display a different value to the user than the one you use in a query. Think of users for example -- you would like to show in the dropdown the user name, while use the user id as the value for the parameter.

rohithmenon:
Sounds reasonable. Will add.

rohithmenon:
@arikfr Lemme know if the changes look good.

azymnis:
@arikfr any chance we could get this merged? Would be great to have this as a feature in our stack

arikfr:
I'm on vacation and will be back next week (29/8). Merging this is going to be one of the first things I do ;)

rohithmenon:
Thanks @arikfr Didn't know that you were out of office.

highlycaffeinatedcode:
Just wondering if queries used for parameters will support parameters themselves?

e.g. if the parameter query were:
 
`select distinct cc.currency_code, cc.currency_name`
`from currency_codes cc`
`join transactions t on cc.currency_code = t.currency_code`
`where t.date = '{{param_date}}' `

When viewing a Dashboard for a param date, the dropdown would be filtered to only show the relevant currencies used on that date.

rohithmenon:
A query with parameters will be supported as long as query with parameters has some results with default values. I don't believe your usecase is supported by this change. Params used in drop down does not show up as params themselves.

Also a feature like that would require parameter dependency. Drop-down (A) depends on Drop-down (B). This also means that cyclic dependency should be avoided. This change is pretty simple with the assumption that query results can be used for drop down values and does not recursively check for dependencies.

rohithmenon:
@arikfr Gentle ping

arikfr:
Merged. Thank you for your patience and for the work done here!

rohithmenon:
@vabanin From your response, I believe the query is not getting an id assigned which is strange.

Are you on the latest master of redash without no other changes? You don't have to pull my change now that the change is merged to master. Could you try the same steps with just redash master?

vabanin:
@rohithmenon 
> Could you try the same steps with just redash master?
Yes, I have made this on redash master -  the result is the same...

rohithmenon:
I believe I know what is happening. Can you try a data source that is not sqlite?

vabanin:
> Can you try a data source that is not sqlite?

My Redash (it's internal DB) works on Postgres, data sources - Postgres and Redshift.

rohithmenon:
To ensure that the problem that I am able to reproduce with sqlite is the same as that you are facing, could you try this:
In file parameters.js, lines 63 to 74 (https://github.com/getredash/redash/pull/1928/files#diff-9fece99dbc1acba86ef0205621bedb48R63) remove the condition (column.type === 'string')

vabanin:
> In file parameters.js, lines 63 to 74 (https://github.com/getredash/redash/pull/1928/files#diff-9fece99dbc1acba86ef0205621bedb48R63) remove the condition (column.type === 'string')

Bingo!  In drop-down has been appeared the result of the query!

rohithmenon:
I am using BigQuery as the backend, and column.type === 'string' works. In some other backends, it is not the case I believe. If you are up for it, you can send a PR to @arikfr

rockwotj:
It seems that sqlite sends a column type of `null` always, so I am not able to get this to work with sqlite.

rohithmenon:
@rockwotj I have made a PR (https://github.com/getredash/redash/pull/1975) with the fix. Most of my tests were of BigQuery and hence did not realize this.

vabanin:
@rohithmenon thank you very much!

arikfr:
@rohithmenon 
Finally had time to properly review this and found some new things:

* Everytime you change the param value, it will load again query result (because of the `$watch` which is triggered for any changed in `param`).
* On a related note, we prefer not to use `$watch` but rather `$onChanges`.
* The query name is saved with the parameter. This might create an inconsistency when the query is renamed. Not a big deal, but I think I will just change it to store the query id and load the query name when rendering.
* When you first setup the parameter, if there is no value for it, there is an empty value selected. Which disappears after you select something (the same happens with the Dropdown List option too). I think we should select the first item in this case.

I will fix those, just wanted to share though.

rohithmenon:
@arikfr Sounds good!

idoDavid:
Any reason I don't see this "Query based dropdown list" type? I'm on 4.0.1 and I see this:

image

arikfr:
@idoDavid no reason I can think of. Are you sure your UI is updated?

tiagoajacobi:
Hey guys, this feature is being very useful for me.

I would like to discuss the possibility of adding one extra thing on it, which is to make the "value" column variable.

For example I have a query that results 4 columns:
 - The Month which can be the "name"
 - days_in_month
 - transactions
 - amount_spend

So I would like to have only one "Query based dropdown list", but use the different comuns in each parameter of my query. For example:
```
SELECT {{month_result}}.days_in_month,
               {{month_result}}.transactions,
               {{month_result}}.amount_spend
```
OR
```
SELECT {{month_result.days_in_month}},
               {{month_result.transactions}},
               {{month_result.amount_spend}}
```
Do you think that something like that can be supported?
Thanks

PR #1933, opened 2017-08-18
131  v toggle schema browser upstream

alison985:
mozilla/redash#131

### DIFF
--- a/client/.babelrc
+++ b/client/.babelrc
@@ -1,4 +1,4 @@
 {
   "presets": ["es2015", "stage-2"],
-  "plugins": ["angularjs-annotate", "transform-object-assign"]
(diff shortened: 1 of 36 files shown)


arikfr:
The idea of giving the user the ability to select which tables to show is a good one, but I feel that this implementation both in concept (using a table suffix) and implementation is too Mozilla specific:

- It uses "versioned table" terminology in code and UI.
- Includes the doc_url feature, which was also rejected from upstream.
- Has some unwanted changes to models code -- currently we send the data source configuration only to admins, it seems that now it will send it to all users.
- Includes the ActiveData query runner.

Thank you for the effort, but closing this for now. Will be happy to discuss a more general approach to this as part of the work on a better schema browser.

PR #1935, opened 2017-08-21
Support writes in mysql

mfouilleul:
Avoid trigging errors during insert, update, deletes.

### DIFF
--- a/redash/query_runner/mysql.py
+++ b/redash/query_runner/mysql.py
@@ -140,37 +140,40 @@ class Mysql(BaseSQLQueryRunner):
                                          ssl=self._get_ssl_parameters())
             cursor = connection.cursor()
             logger.debug("MySQL running query: %s", query)
-            queries = query.split(';')
+            queries = query.rstrip(';').split(';')
 
             transaction = False
             transaction_rows = 0
 
             for query in queries:
-                if len(query) > 0:
-                    rows_count = cursor.execute(query)
-                    data = cursor.fetchall()
-                    if cursor.description is not None:
-                        columns = self.fetch_columns([(i[0], types_map.get(i[1], None)) for i in cursor.description])
-                        rows = [dict(zip((c['name'] for c in columns), row)) for row in data]
-
-                        data = {'columns': columns, 'rows': rows}
-                        json_data = json.dumps(data, cls=JSONEncoder)
-                        error = None
-                    else:
-                       if rows_count >= 0:
-                           transaction = True
-                           columns = [{'name': 'Row(s) Count', 'type': types_map.get(str(type(rows_count)).upper(), None)}]
-                           transaction_rows += rows_count
+                 cursor.execute(query)
+                 rows_count = cursor.rowcount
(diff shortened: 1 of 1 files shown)


redash/query_runner/mysql.py
```diff
@@ -140,21 +140,39 @@ def run_query(self, query, user):
...
-            cursor.execute(query)
-
-            data = cursor.fetchall()
-
-            # TODO - very similar to pg.py
-            if cursor.description is not None:
-                columns = self.fetch_columns([(i[0], types_map.get(i[1], None)) for i in cursor.description])
-                rows = [dict(zip((c['name'] for c in columns), row)) for row in data]
-
+            queries = query.rstrip(';').split(';')
```
arikfr: That's not such a good idea, think of the following (contrived) example:

```
>>> "SELECT ';'".split(';')
["SELECT '", "'"]
```

But see the fix here: https://github.com/getredash/redash/pull/1955/commits/e02fdb3e37e5b6fefab591a44c74ecf4ce3b6223 for supporting multiple queries.

Once you rebase with latest, you can simplify the logic you implemented here.

redash/query_runner/mysql.py
```diff
@@ -140,21 +140,39 @@ def run_query(self, query, user):
...
-            else:
-                json_data = None
+
+            if json_data is None:
                 error = "No data was returned."
+            else:
+                error = None
+
+            if transaction:
+                connection.commit()
```
arikfr: Why not commit above?

redash/query_runner/mysql.py
```diff
@@ -140,21 +140,39 @@ def run_query(self, query, user):
...
+                if cursor.description is not None:
+                    columns = self.fetch_columns([(i[0], types_map.get(i[1], None)) for i in cursor.description])
+                    rows = [dict(zip((c['name'] for c in columns), row)) for row in data]
+                    data = {'columns': columns, 'rows': rows}
+                    json_data = json.dumps(data, cls=JSONEncoder)
+                    break
+                else:
+                    transaction = True
+                    columns = [{'name': 'Row(s) Affected',
+                                'type': types_map.get(str(type(rows_count)).upper(), None)}]
```
arikfr: Wouldn't it be always an integer? In which case you can use `TYPE_INTEGER` constant.

mfouilleul:
Do we need more changes for this?

arikfr:
Is this still needed following the changes in #3003?

Jakdaw:
I'm not sure how much need there is for writes from redash - with the #3003 changes you can now achieve the same thing; eg with a redash query like:-

```
UPDATE mytable SET a=a+1 WHERE b={{whatever}};
SELECT ROW_COUNT() as "Row(s) Affected";
```

arikfr:
Following #3003, closing this. @mfouilleul, if you still think this is needed (and willing to put in the work to update it to the changes in `redash.query_runner.mysql`) -- let me know. I will review it again before you do some extra work.

Thanks.

PR #1949, opened 2017-09-06
Run queries with no cached result in public dashboards

washort:


arikfr:
I think you opened it for the wrong repo :)

PR #1954, opened 2017-09-13
Upgraded dql version to 0.5.24

aterreno:
This allows querying tables in dynamo DB which have keys with dashes in the name. 

Please see more here about this bug report on DQL: https://github.com/stevearc/dql/issues/12


arikfr:
Thanks!

PR #1955, opened 2017-09-13
MySQL: multiple queries support & connection timeout 

arikfr:
* Added support for multiple queries (shows results only for the last one).
* Added connection timeout for flaky hosts.


zhongjiajie:
Why use hardcode here? should we use configure like other paramters?

Issue #811, opened 2016-02-03
Feature request: setup layouts for charts


TonyStarkBy:
We have some chart with big numbers. And for us is difficult to read "6.234454M" rather then "6,234,454"
You can see difference here: http://codepen.io/anon/pen/NxLPqe

It can be fixed with:
`var layout = {                       
    yaxis: {                 
      title: 'time'  ,
      tickformat: 's',
      hoverformat: 'n',
    }
};`

PR #1958, opened 2017-09-20
Add hoverformat support to yAxis

TonyStarkBy:
Add: support hoverformat for Y Axis #811


arikfr:
Thank you. We never got to land this specific pull requestin master, but you can now control the hover text in charts.

PR #1965, opened 2017-09-22
Prevent line breaks in EditInPlace description when using Firefox

alexmuller:
Before this change, pressing enter in Firefox 55 would insert a line break into the description field and then save it.

This change prevents the line break from being inserted before saving.

There's no change to Chrome's behaviour from this change.


alexmuller:
Looks like there are some unresolved security vulnerabilities in [marked](https://npmjs.com/package/marked) which are causing the tests to fail.

alexmuller:
#1966 should resolve this.

arikfr:
Thanks!

PR #1966, opened 2017-09-22
Use a different markdown library

alexmuller:
`marked` has some security vulnerabilities which have been unresolved for a while. `markdown` seems to be better supported.


arikfr:
Thanks!

PR #1967, opened 2017-09-22
ODBC query runner

wankdanker:
Based on #1906, this pull request implements a generic ODBC query runner. Schema is obtained using pyodbc's cursor.colums() method.

I'm not super knowledgeable in Python, so I may have missed some error checking or something else.


arikfr:
Hi,

Thanks.

PR #1972, opened 2017-09-25
Configure supervisor.conf logging to /opt/redash/logs/redash.log…

bradrushworth:
… with a rollover at 100MB.

I couldn't get production logging to work until I added this configuration.


arikfr:
> I couldn't get production logging to work until I added this configuration.

How did you setup things?

arikfr:
Thank you for submitting this pull request. We redid all the setup scripts, so this is no longer relevant.

PR #1975, opened 2017-09-27
Bugfix/query based param

rohithmenon:
column.type not set by many data sources [sqlite, postgres etc]. Therefore removing the additional check.


rohithmenon:
@rockwotj Nope!
Here is an example:
image

arikfr:
Thanks 👍

PR #1976, opened 2017-09-27
users using docker-compose require restart always

muddydixon:
# Summary
* added `restart always` to each services in `docker-compose.yml`
* perhaps almost users using docker-compose require restart always.


arikfr:
Thanks. How about we add this only to the Redis & Postgres services in `docker-compose.yml` and to all services in `docker-compose.production.yml`?

muddydixon:
OK! I fixed this.

arikfr:
Merged. THanks!

yutannihilation:
> Production yml: Line 34 has to be below 35

Agreed. It seems this PR broke the YAML syntax...

PR #1985, opened 2017-10-02
Ensure email is case-insensitive

### LINKED ISSUE #1980: Login page for user/pass is case sensitive for the username
### Issue Summary

When logging in with a username/password, the username is case-sensitive, and disallows a user from logging in when there is a mismatch.

### Steps to Reproduce

1. Create a user `foo@example.com` with a password
2. Try to log in with `Foo@example.com` and get a failure

It is common to have usernames/email addresses converted to lowercase (or possibly uppercase if you lean that way) to ensure uniqueness and consistent user behavior.

The current implementation appears to allow both `foo@example.com` and `Foo@example.com` to be two different users, despite having the same effective email address.

I think from an implementation perspective, this could live either in the form-side (some sort of javascript conversion `.toLowerCase()` or (preferably) an Authentication API-side `.lower()` for the input.

I'd be happy to hack on this, however in a quick, cursory evaluation it isn't immediately apparent where to implement this kind of change, so any pointers would be helpful.

miketheman:
While RFC 5321 alows for the username/mailbox art of the address to be
case-sensitive, it is highly discouraged.

    The local-part of a mailbox MUST BE treated as case
    sensitive. Therefore, SMTP implementations MUST take care to
    preserve the case of mailbox local-parts. In particular, for
    some hosts, the user "smith" is different from the user
    "Smith". However, exploiting the case sensitivity of mailbox
    local-parts impedes interoperability and is discouraged.
    Mailbox domains follow normal DNS rules and are hence not
    case sensitive.
    -- https://tools.ietf.org/rfc/rfc5321.txt

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

Resolves #1980

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -349,15 +349,15 @@ class Group(db.Model, BelongsToOrgMixin):
         return unicode(self.id)
 
 
-class EmailType(TypeDecorator):
+class LowercasedString(TypeDecorator):
     """
-    A case-insensitive string
+    A lowercased string
     """
     impl = db.String
     comparator_factory = CaseInsensitiveComparator
 
     def __init__(self, length=320, *args, **kwargs):
-        super(EmailType, self).__init__(length=length, *args, **kwargs)
+        super(LowercasedString, self).__init__(length=length, *args, **kwargs)
 
     def process_bind_param(self, value, dialect):
         if value is not None:
@@ -374,7 +374,7 @@ class User(TimestampMixin, db.Model, BelongsToOrgMixin, UserMixin, PermissionsCh
     org_id = Column(db.Integer, db.ForeignKey('organizations.id'))
     org = db.relationship(Organization, backref=db.backref("users", lazy="dynamic"))
     name = Column(db.String(320))
-    email = Column(EmailType)
+    email = Column(LowercasedString)
     password_hash = Column(db.String(128), nullable=True)
     # XXX replace with association table
     group_ids = Column('groups', MutableList.as_mutable(postgresql.ARRAY(db.Integer)), nullable=True)
@@ -386,7 +386,8 @@ class User(TimestampMixin, db.Model, BelongsToOrgMixin, UserMixin, PermissionsCh
     __table_args__ = (db.Index('users_org_id_email', 'org_id', 'email', unique=True),)
 
     def __init__(self, *args, **kwargs):
-        kwargs["email"] = kwargs.get("email", None).lower()
+        if kwargs.get('email') is not None:
+            kwargs['email'] = kwargs['email'].lower()
         super(User, self).__init__(*args, **kwargs)
 
     def to_dict(self, with_api_key=False):
--- a/redash/models.py
+++ b/redash/models.py
@@ -22,6 +22,7 @@ from redash.permissions import has_access, view_only
 from redash.query_runner import (get_configuration_schema_for_query_runner_type,
                                  get_query_runner)
 from redash.utils import generate_token, json_dumps
+from redash.utils.comparators import CaseInsensitiveComparator
 from redash.utils.configuration import ConfigurationContainer
 from sqlalchemy import distinct, or_
 from sqlalchemy.dialects import postgresql
@@ -348,12 +349,32 @@ class Group(db.Model, BelongsToOrgMixin):
         return unicode(self.id)
 
 
+class EmailType(TypeDecorator):
+    """
+    A case-insensitive string
+    """
+    impl = db.String
+    comparator_factory = CaseInsensitiveComparator
+
+    def __init__(self, length=320, *args, **kwargs):
+        super(EmailType, self).__init__(length=length, *args, **kwargs)
+
+    def process_bind_param(self, value, dialect):
+        if value is not None:
+            return value.lower()
+        return value
+
+    @property
+    def python_type(self):
+        return self.impl.type.python_type
+
+
 class User(TimestampMixin, db.Model, BelongsToOrgMixin, UserMixin, PermissionsCheckMixin):
     id = Column(db.Integer, primary_key=True)
     org_id = Column(db.Integer, db.ForeignKey('organizations.id'))
     org = db.relationship(Organization, backref=db.backref("users", lazy="dynamic"))
     name = Column(db.String(320))
-    email = Column(db.String(320))
+    email = Column(EmailType)
     password_hash = Column(db.String(128), nullable=True)
     # XXX replace with association table
     group_ids = Column('groups', MutableList.as_mutable(postgresql.ARRAY(db.Integer)), nullable=True)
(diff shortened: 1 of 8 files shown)


miketheman:
Resolves #1980

I'm not 100% certain of the impact this would have on existing users already stored in the db - so this deserves a bit of discussion and thought before merging.

It's possible that writing a DB migration to run effectively `UPDATE users SET email=lower(email)` would do the trick - but I'm not certain of how to go about that.

miketheman:
@arikfr any thoughts?

miketheman:
Great idea. In my reading about this, there is a postgres extension that can enforce case insensitive columns as well.

For this solution, how extensive should it be?
Do we care if someone pops open the SQL shell and inserts values directly? Is it acceptable to require a postgres extension? (For reference, RDS postgres makes the extension available as of 9.3.x)

Or should the enforcement be entirely in the Python side, and we expect that anyone touching the DB directly "knows what they are doing"?

Either path can be chosen, and have different implications.

The pg extension one is the most "complete" as it also applies to the indexes created, and may pull in one more Python dependency.

arikfr:
I think that something on the Python side is enough, as we can't be certain that all existing deployments will have this extension. And we can assume that if someone is messing directly with the DB, he should know what he's doing.

miketheman:
@arikfr I took another pass at this, moving the logic to the model. I tried to add more tests to exercise the behaviors.

redash/models.py
```diff
@@ -348,12 +349,32 @@ def __unicode__(self):
         return unicode(self.id)
 
 
+class EmailType(TypeDecorator):
+    """
+    A case-insensitive string
+    """
+    impl = db.String
+    comparator_factory = CaseInsensitiveComparator
```
arikfr: Assuming that all emails already lower cased, why do we need a special comparator?

redash/models.py
```diff
@@ -365,6 +386,7 @@ class User(TimestampMixin, db.Model, BelongsToOrgMixin, UserMixin, PermissionsCh
     __table_args__ = (db.Index('users_org_id_email', 'org_id', 'email', unique=True),)
 
     def __init__(self, *args, **kwargs):
+        kwargs["email"] = kwargs.get("email", None).lower()
```
arikfr: Isn't this handled by the `EmailType`?

redash/models.py
```diff
@@ -348,12 +349,32 @@ def __unicode__(self):
         return unicode(self.id)
 
 
+class EmailType(TypeDecorator):
```
arikfr: Maybe let's name it `LowercasedString`?

miketheman: Sure, can rename. Wanted to keep this targeted at email, but if it's more generally useful, then will do.

redash/models.py
miketheman: The custom comparator is to handle when inputs from anywhere (api, cli) won't need to specify `.lower()` when performing equality comparison.

Here's an example of how the SQL would look, with and without the custom comparator:

```python
>>> from redash.models import User
>>> user1 = User.query.filter(User.email == "TEST@example.com")
>>> print(user1)
```

```sql
-- Before
SELECT users.groups AS users_groups, users.updated_at AS users_updated_at, users.created_at AS users_created_at, users.id AS users_id, users.org_id AS users_org_id, users.name AS users_name, users.email AS users_email, users.password_hash AS users_password_hash, users.api_key AS users_api_key
FROM users
WHERE users.email = %(email_1)s

-- After, custom comparator uses SQL `lower`
SELECT users.groups AS users_groups, users.updated_at AS users_updated_at, users.created_at AS users_created_at, users.id AS users_id, users.org_id AS users_org_id, users.name AS users_name, users.email AS users_email, users.password_hash AS users_password_hash, users.api_key AS users_api_key
FROM users
WHERE lower(users.email) = lower(%(lower_1)s)
```

So this prevents us needing to handle every case of anyone trying to compare against this column by ensuring the underlying SQL handles it.

redash/models.py
miketheman: I added this, to handle the case where the email as input is returned directly to the API response.
Without it, it triggers the following failure:

```
======================================================================
FAIL: test_creates_user_case_insensitive_email (tests.handlers.test_users.TestUserListResourcePost)
----------------------------------------------------------------------
  File "/app/tests/handlers/test_users.py", line 37, in test_creates_user_case_insensitive_email
    self.assertEqual(rv.json['email'], 'user@example.com')
AssertionError: u'User@Example.com' != 'user@example.com'
```

When a [User is created here](https://github.com/getredash/redash/blob/0c404fa6024569651110d21f27a41eab10f29419/redash/handlers/users.py#L31-L34), the input response is sent directly back, despite the value being stored in the DB with the lowercased value.

A reproduction via shell:

```python
>>> from redash import models
>>> org = models.Organization.query.first()
>>> user = models.User(org=org, name="Fred", email="FRED@flintstone.org")
>>> models.db.session.add(user)
>>> models.db.session.commit()
>>> print(user.email)
FRED@flintstone.org

# New shell session
>>> from redash import models
>>> user = models.User.query.filter_by(email="FRED@flintstone.org").one()  # Note: this is where the custom comparison comes into play
>>> print(user.email)
fred@flintstone.org
```

So either there's something to be done about resetting the session post-commit, or re-fetching the value from the DB. Using `kwargs` to ensure that when instantiating a `User()`, the email always gets lowered.

If it's acceptable that the response to a user being created contain the original case, then I'll remove it, and change the test case assertion - I'm not clear how the response is used yet, so wanted to ensure that whatever is returned is something the caller can depend on.

miketheman: Reading a little more, we could add a `refresh()` call, which expires the sqlalchemy object state and immediately performs a SELECT.

Which is preferable?

- Ensuring each time a `User()` object is created, a call to `lower()` happens in Python-land, and from then on the session is internally consistent
- When pushing a value to the DB, we trigger a second query to refresh the session object's state from the DB

miketheman:
@arikfr I think I answered the questions posed - can you please take a look?

miketheman:
Hey @arikfr ! I responded to the questions - could you please take another look?

redash/models.py
arikfr: 👍 sounds good. Let's keep it the way you implemented to avoid an extra query. One thing I noticed though: `kwargs.get('email', None)` might return `None`, in which case calling `.lower()` on it will fail.

We should probably call `.lower()` only if we get back a value. Maybe something like:

```python
if kwargs.get('email') is not None:
    kwargs['email'] = kwargs['email'].lower()
```

(`.get` returns `None` by default)

arikfr:
Thank you for the your patience here :-) See my new comment. Aside from that looks good.

miketheman:
@arikfr Completed class rename and extra guard against None.

miketheman:
Hi @arikfr ! Was hoping this wasn't forgotten.

arikfr:
Many thanks 🙏

PR #1990, opened 2017-10-02
Timezone support for human time

### LINKED ISSUE #618: User configuration for date/time format & default timezone
See #617 and #411 for more details.

jakesjews:
Partially resolves #618 

This is an initial implementation of support for setting organization timezones and using that timezone with the human time helper. I'm not too familiar with angular so there might be some improvements I need to make. It seemed like the best place for the timezone option was in organization settings but I can change it to anything else that works better. I can also add some documentation after the final design gets locked down.

screen shot 2017-10-02 at 2 50 57 pm


jakesjews:
I'm not sure why eslint is failing on the moment import. The other moment imports look the same and running eslint locally passes.

jakesjews:
Is there a different approach that would be better than this? I'm open to changing anything

arikfr:
Thanks! The approach is good and I was actually planning to do something similar to move more configuration from environment variables to the UI. 

I still need to look into this in detail and review properly. I'm currently focused on finishing the v3 release. But once I'm done with it, will get back to this. Is this OK?

Thanks.

jakesjews:
Totally OK. Thanks a bunch!

jakesjews:
@arikfr should I get this PR merged up with latest master or should I implement this in a different way?

arikfr:
@jakesjews Sorry for keeping this hanging for so long... I realized that we won't be able to introduce a time zone setting that only affects the behavior of `$humanTime`, while implementing proper support for timezones is not an effort we can do right now.

If you still want to implement this, I was thinking that maybe we can introduce an optional parameter to `$humanTime`? So it becomes:

```json
{
	"collection": "date_test",
	"query": {
		"lastModified": {
			"$gt": {
				"$humanTime": {"value": "3 years ago", "timezone": "Europe"}
			}
		}
	},
	"limit": 100
}
```

A bit verbose, but better than nothing...

jakesjews:
@arikfr that works for me I can close this and make another PR

PR #1995, opened 2017-10-04
Modernize Python 2 code to get ready for Python 3

cclauss:
Make the minimal, safe changes required to convert the repo's code to be syntax compatible with both Python 2 and Python 3. There is definitely more work required to complete the port to Python 3 but this is a minimal, safe first step.


arikfr:
Awesome, thanks! 

Really looking forward to being able to move to Python 3.

PR #1997, opened 2017-10-05
Working Py3 code that still supports Py2

cclauss:
@akariv This is the follow on PR to #1995  While #1995 was [futurize stage 1](http://python-future.org/futurize_cheatsheet.html#step-1-modern-py2-code), this PR is [futurize stage 2](http://python-future.org/futurize_cheatsheet.html#step-2-working-py3-code-that-still-supports-py2) plus some cleanup on [raise_with_traceback()](http://python-future.org/compatible_idioms.html#raising-exceptions).

Please let me know if it works for you in both Python 2 and in Python 3.

### DIFF
--- a/old_migrations/0003_update_data_source_config.py
+++ b/old_migrations/0003_update_data_source_config.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from past.builtins import basestring
 import json
 import jsonschema
 from jsonschema import ValidationError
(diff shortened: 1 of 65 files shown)


akariv:
@arikfr ^

cclauss:
Perhaps I did not need to add future to __requirements_all_ds.txt__ if it is already in __requirements.txt__.

arikfr:
I need to review this to make sure it's safe. Anyway won't be merged before the v3 release.

PR #1998, opened 2017-10-06
[FIX] Revoke permission should respect to given grantee and access type.

meinac:
The issue is, if you try to revoke the permission of a user from an
object, all the permissions on this object get removed. The fix is
assigning filtered query object to it's own reference.

According to SQLAlchemy documentation, `filter` method applies to
the **copy** of the query object which means calling filter doesn't
affect the object receiving filter call. For more information;
http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.filter

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -1041,13 +1041,13 @@ class AccessPermission(GFKBase, db.Model):
                              cls.object_type == obj.__tablename__)
 
         if access_type:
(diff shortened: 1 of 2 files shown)


arikfr:
Nice catch! Thanks.

PR #2004, opened 2017-10-09
Allow setting test file with docker test run

meinac:
During the development process, sometimes it makes sense to run only
the related tests of the file you've recently changed for speed.
By this change you can run only one test file, like so;

  `docker-compose run --rm -e TEST_FILE=tests/handlers/test_dashboards.py server tests`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ FULL_VERSION=$(VERSION)+b$(CIRCLE_BUILD_NUM)
 BASE_VERSION=$(shell python ./manage.py version | cut -d + -f 1)
 # VERSION gets evaluated every time it's referenced, therefore we need to use VERSION here instead of FULL_VERSION.
 FILENAME=$(CIRCLE_ARTIFACTS)/$(NAME).$(VERSION).tar.gz
+TEST_ARGS?=--with-coverage --cover-package=redash tests/
 
 deps:
 	if [ -d "./client/app" ]; then npm install; fi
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,6 +4,7 @@ FULL_VERSION=$(VERSION)+b$(CIRCLE_BUILD_NUM)
 BASE_VERSION=$(shell python ./manage.py version | cut -d + -f 1)
 # VERSION gets evaluated every time it's referenced, therefore we need to use VERSION here instead of FULL_VERSION.
 FILENAME=$(CIRCLE_ARTIFACTS)/$(NAME).$(VERSION).tar.gz
+TEST_FILE?=tests/
```
arikfr: It can be used to pass any argument/option to `nosetsts`; maybe we should give it a more generic name, like `TEST_ARGUMENTS` or something like this?

meinac: Thanks for the feedback. I think it's a good point, we can rename it as `TEST_ARGS` or `TEST_ARGUMENTS` but the variable has to be quoted like;
`docker-compose run --rm -e TEST_ARGS="--process=4 tests/handlers/test_dashboards.py" server tests`

Passing this argument will override the entire default arguments. The default one will be `--with-coverage --cover-package=redash tests/`

Does this make sense?

arikfr:
> Does this make sense?

Yes. :)

meinac:
Thanks for your quick response, I've updated the makefile as well as the commit message.

arikfr:
Thanks!

PR #2017, opened 2017-10-11
Fix docker-compose.production.yml

yutannihilation:
`docker-compose` fails on my local Windows with this error:

```console
> docker-compose -f docker-compose.production.yml run --rm server create_db
ERROR: yaml.scanner.ScannerError: mapping values are not allowed here
  in ".\docker-compose.production.yml", line 35, column 20
```

It seems #1976 broke `docker-compose.production.yml` by introducing some wrong syntax. This PR fixes the issue as suggested in https://github.com/getredash/redash/pull/1976#commitcomment-24882099.


arikfr:
Thanks!

yutannihilation:
Thanks for merging 👍

PR #2021, opened 2017-10-12
Make invitation token max age configurable

hhamalai:
Allows configuration the max life time of an invitation token which is currently fixed in code for one week.


arikfr:
Thanks!

Issue #1883, opened 2017-07-18
'NoneType' object has no attribute 'width' when adding 'Regular' width widget


albrycht:
### 

A summary of the issue and the browser/OS environment in which it occurs.

### Steps to Reproduce

1. Create new empty dashboard
2. Create new query, publish it
3. Add widget to dashbord with width "Regular"
4. Archive query
5. Try to add new widget to this dashboard with width "Regular"

Last step should fail. Logs from gunicorn:

     [2017-07-18 11:31:38,880][PID:8359][ERROR][redash] Exception on /api/widgets [POST]
     Traceback (most recent call last):
       File "/opt/redash/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask/app.py", line 1639, in full_dispatch_request
         rv = self.dispatch_request()
       File "/opt/redash/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask/app.py", line 1625, in dispatch_request
         return self.view_functions[rule.endpoint](**req.view_args)
       File "/opt/redash/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask_restful/__init__.py", line 477, in wrapper
         resp = resource(*args, **kwargs)
       File "/opt/redash/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask_login/utils.py", line 228, in decorated_view
         return func(*args, **kwargs)
       File "/opt/redash/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask/views.py", line 84, in view
         return self.dispatch_request(*args, **kwargs)
       File "/opt/redash/apps/redash/htdocs/redash/handlers/base.py", line 28, in dispatch_request
         return super(BaseResource, self).dispatch_request(*args, **kwargs)
       File "/opt/redash/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask_restful/__init__.py", line 587, in dispatch_request
         resp = meth(*args, **kwargs)
       File "/opt/redash/apps/redash/htdocs/redash/permissions.py", line 47, in decorated
         return fn(*args, **kwargs)
       File "/opt/redash/apps/redash/htdocs/redash/handlers/widgets.py", line 56, in post
         if neighbour_widget.width == 1:
     AttributeError: 'NoneType' object has no attribute 'width'

Additional info:
I was repeating those 5 steps over and over again I was able to generate dashboard which contained empty widgets. Field `widgets` returned by API contained list:

     [      
         [],
         [],
         [] 
     ]      

[...]

PR #2025, opened 2017-10-13
Fix error when adding widget to dashboard

kyoshidajp:
A detail of this error is #1883.

I changed that delete widgets containing a query from dashboard layouts when deleting the query.

And, check `neighbour_widget` exists or not for existing users who have a dashboard generated by this bug when they try to add new widget.


arikfr:
Thanks! But following #2062 we no longer store the layout in the dashboard, so this bug is no longer relevant.

PR #2028, opened 2017-10-14
Set auto focus in first input items

kyoshidajp:
In order to improve user experience, the following items should be focus automatically when they open the page. 

- **Type** in new Data Source form
- **Trigger** in new Query Snippet form
- **Group Name** in new Group dialog
- **User name** in new User form
- **Dashboard name** in edit Dashboard dialog
- **Search box** in Dashboard list

Because they set focus them manually in most cases.


arikfr:
Cool. Thanks!

PR #2038, opened 2017-10-16
Add support for AWS Redshift Spectrum (external) tables

atharvai:
New Feature:
Show external tables in schema for Redshift data source.

Redshift Spectrum's External tables are not stored in `information_schemas.tables` or `in pg_catalog.tables` so we need to use [`svv_columns`](http://docs.aws.amazon.com/redshift/latest/dg/r_SVV_COLUMNS.html) to get the tables and columns of external schemas. `svv_columns` includes both internal Redshift tables/views and external tables.


arikfr:
Thanks!

rajaraman84:
svv_tables is used to get the base table, external table  and view

PR #2040, opened 2017-10-17
Show query editor's Archive/Publish Query drop-down only on saved queries

cyriac:
image
changes to 
image

--
The drop-down shows up for saved queries as normal.


arikfr:
Thanks 👍

PR #2041, opened 2017-10-17
Add full text search for queries based on the Postgres tsvector type.

jezdez:
This needs some more tests with a prod like database and probably another
look for someone with more experience with custom alembic migrations.

This is related to #1400 of course and uses [SQLAlchemy-searchable](https://sqlalchemy-searchable.readthedocs.io/en/latest/).

### DIFF
--- a/migrations/versions/5ec5c84ba61e_.py
+++ b/migrations/versions/5ec5c84ba61e_.py
@@ -1,0 +1,35 @@
+"""Add Query.search_vector field for full text search.
+
+Revision ID: 5ec5c84ba61e
+Revises: 7671dca4e604
+Create Date: 2017-10-17 18:21:00.174015
+
+"""
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils as su
+import sqlalchemy_searchable as ss
+
+
+# revision identifiers, used by Alembic.
+revision = '5ec5c84ba61e'
+down_revision = '7671dca4e604'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    op.add_column('queries', sa.Column('search_vector', su.TSVectorType()))
+    op.create_index('ix_queries_search_vector', 'queries', ['search_vector'],
+                    unique=False, postgresql_using='gin')
+    ss.sync_trigger(conn, 'queries', 'search_vector',
+                    ['name', 'description', 'query'])
+
+
+def downgrade():
+    conn = op.get_bind()
+
+    ss.drop_trigger(conn, 'queries', 'search_vector')
+    op.drop_index('ix_queries_search_vector', table_name='queries')
+    op.drop_column('queries', 'search_vector')
--- a/redash/models.py
+++ b/redash/models.py
@@ -12,7 +12,7 @@ from funcy import project
 
 import xlsxwriter
 from flask_login import AnonymousUserMixin, UserMixin
-from flask_sqlalchemy import SQLAlchemy
+from flask_sqlalchemy import SQLAlchemy, BaseQuery
 from passlib.apps import custom_app_context as pwd_context
 from redash import settings, redis_connection, utils
 from redash.destinations import (get_configuration_schema_for_destination_type,
@@ -30,11 +30,13 @@ from sqlalchemy.dialects import postgresql
 from sqlalchemy.event import listens_for
 from sqlalchemy.ext.mutable import Mutable
 from sqlalchemy.inspection import inspect
-from sqlalchemy.orm import backref, joinedload, object_session, subqueryload
+from sqlalchemy.orm import backref, joinedload, object_session
 from sqlalchemy.orm.exc import NoResultFound  # noqa: F401
 from sqlalchemy.types import TypeDecorator
 from sqlalchemy.orm.attributes import flag_modified
 from functools import reduce
+from sqlalchemy_searchable import SearchQueryMixin, make_searchable, vectorizer
+from sqlalchemy_utils.types import TSVectorType
 
 
 class SQLAlchemyExt(SQLAlchemy):
@@ -49,6 +51,21 @@ class SQLAlchemyExt(SQLAlchemy):
 db = SQLAlchemyExt(session_options={
     'expire_on_commit': False
 })
+# Make sure the SQLAlchemy mappers are all properly configured first.
+# This is required by SQLAlchemy-Searchable as it adds DDL listeners
+# on the configuration phase of models.
+db.configure_mappers()
+
+# listen to a few database events to set up functions, trigger updates
+# and indexes for the full text search
+make_searchable(options={'regconfig': 'pg_catalog.simple'})
+
+
+class SearchBaseQuery(BaseQuery, SearchQueryMixin):
+    """
+    The SQA query class to use when full text search is wanted.
+    """
+
 
 Column = functools.partial(db.Column, nullable=False)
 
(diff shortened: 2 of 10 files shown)


migrations/versions/5ec5c84ba61e_.py
```diff
@@ -0,0 +1,35 @@
...
+                                       nullable=True))
+
+    ss.sync_trigger(conn, 'queries', 'search_vector',
+                    ['name', 'description', 'query'])
+
+
+def downgrade():
+    conn = op.get_bind()
+
+    ss.drop_trigger(conn, 'article', 'search_vector')
```
arikfr: `article`?

redash/cli/database.py
```diff
@@ -7,6 +7,8 @@
 def create_tables():
     """Create the database tables."""
     from redash.models import db
+    # make sure the additional search mappers are configured
+    db.configure_mappers()
```
arikfr: What are the search mappers?

redash/models.py
```diff
@@ -915,27 +931,33 @@ def outdated_queries(cls):
...
-    def search(cls, term, group_ids, include_drafts=False):
-        # TODO: This is very naive implementation of search, to be replaced with PostgreSQL full-text-search solution.
-        where = (Query.name.ilike(u"%{}%".format(term)) |
-                 Query.description.ilike(u"%{}%".format(term)))
-
-        if term.isdigit():
-            where |= Query.id == term
-
-        where &= Query.is_archived == False
+    def search_queries(cls, term, group_ids, include_drafts=False, limit=20):
```
arikfr: Why append `_queries`? It's a method on the `Query` class isn't it obvious that it searches for queries? And in the handlers code you still call `.search`...

redash/models.py
```diff
@@ -28,10 +28,12 @@
 from sqlalchemy.event import listens_for
 from sqlalchemy.ext.mutable import Mutable
 from sqlalchemy.inspection import inspect
-from sqlalchemy.orm import backref, joinedload, object_session, subqueryload
+from sqlalchemy.orm import backref, joinedload, object_session
 from sqlalchemy.orm.exc import NoResultFound  # noqa: F401
 from sqlalchemy.types import TypeDecorator
 from functools import reduce
+from sqlalchemy_searchable import SearchQueryMixin, make_searchable, search
+from sqlalchemy_utils.types import TSVectorType
```
arikfr: Not that I mind much about adding `sqlalchemy_utils` as a dependency, but doesn't `sqlalchemy_searchable` provide something for this?

migrations/versions/5ec5c84ba61e_.py
jezdez: Yikes!

redash/models.py
jezdez: Yeah, the [SQLAlchemy-Utils](https://github.com/kvesteri/sqlalchemy-utils) package is from the same author as [SQLAlchemy-Searchable](https://github.com/kvesteri/sqlalchemy-searchable) and contains a whole bunch of things.

The code behind [TSVectorType](https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/types/ts_vector.py) is trivial relatively speaking, so I guess we could also just ship our own. The only risk I see with this is that we could trade one dependency less for having to keep an eye on changes in SQLAlchemy-Searchable that require changes in the TSVectorType class as well.

On the other hand, although quite out of scope of this PR, is the question if there are things in SQLAlchemy-Utils that may be useful for Redash in general, given its swiss-army-knife-ness. I'm fine with whatever way you want to go with this.

redash/models.py
jezdez: Ugh, yeah, this is just a leftover from figuring out how to use the `search` method that the `SearchQueryMixin` adds to the `query_class` of the `Query` model. Will fix.

redash/cli/database.py
jezdez: The comment is misleading, this basically makes sure that all mappers that have been registered with the SQLAlchemy ORM have been properly configured. Since SQLAlchemy-Searchable installs database triggers that have to act on when ORM model instances are created (e.g. to track updates), this call is needed.

Here's what the SA-Searchable docs say:

> It is very important here that you either access your searchable class or call configure_mappers before the creation of tables. SA-Searchable adds DDL listeners on the configuration phase of models.

Since this code happens in the phase in which tables are created, we can't really access the searchable class, but have to call `configure_mappers`. I'll update the comment to reflect this.

redash/models.py
arikfr: 👍 ok, let's use it. No need to find more places to use it, we can gradually introduce it where applicable.

migrations/versions/5ec5c84ba61e_.py
```diff
@@ -0,0 +1,35 @@
...
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    op.add_column('queries', sa.Column('search_vector', su.TSVectorType(),
+                                       nullable=True))
+
+    ss.sync_trigger(conn, 'queries', 'search_vector',
+                    ['name', 'description', 'query'])
```
arikfr: Will this also update all existing queries?

migrations/versions/5ec5c84ba61e_.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def downgrade():
+    conn = op.get_bind()
+
+    ss.drop_trigger(conn, 'queries', 'search_vector')
+    op.drop_index('ix_queries_search_vector', table_name='queries')
```
arikfr: This is a byproduct of `ss.sync_trigger`?

migrations/versions/5ec5c84ba61e_.py
jezdez: Yes, looking at the [docs](https://sqlalchemy-searchable.readthedocs.io/en/latest/alembic_migrations.html#sqlalchemy_searchable.sync_trigger) it says: "Updates all rows for given search vector by running a column=column update query for given table."

I've also manually added a bunch of queries against Redash's own Postgres database in the master branch, then switch to this branch, did the Alembic upgrade and then was able to use the search feature right away.

migrations/versions/5ec5c84ba61e_.py
jezdez: Yep, although weirdly it raised a ProgrammingError exception when I tried roll it back, will look into this.

jezdez: This was an oversight when setting up SA-Searchable. The migration that I fixed in https://github.com/getredash/redash/pull/2041/commits/eb6966fa83d2b44978bf5b49db0df052d11c8170 now reflects adding an index for the search vector column. Another call to `db.configure_mappers()` was needed to have the mapper processor of SA-Searchable to correctly add the index.

arikfr: 👍 

> Another call to db.configure_mappers() was needed

It seems that it's at the top level of `redash.models`. So I guess the one in `redash.cli.database` is redundant?

jezdez: True indeed, good catch, removed it in https://github.com/getredash/redash/pull/2041/commits/36aeef776d610a266d1c9dfa166be10bb5d1726a.

jezdez:
@arikfr Just added a bunch of tests for the more advanced search queries as described on https://sqlalchemy-searchable.readthedocs.io/en/latest/search_query_parser.html.

arikfr:
@jezdez this is finalyl ready to be merged, but I think I saw some fixes you did to this code in Mozilla's fork. Are they present here already or this need an update?

Thanks.

jezdez:
@arikfr They aren't in this PR yet, I'll get them in today and let you know. Thanks for reaching out again!

arikfr:
Super, thanks!

jezdez:
@arikfr Okay, I've updated the PR with the recent things in our fork and updated the design to match the query-list and my-queries page as well. Does this work?

screen-shot-2018-01-08-22-48-58

jezdez:
@arikfr BTW, apologies for taking so long :(

jezdez:
@arikfr Any idea how we can retrigger the code climate check? It seems to be stuck.

arikfr:
Thanks ! 👍

PR #2042, opened 2017-10-18
Make /opt/redash directory if it's not exist.

isomura:
This change make this setup script easy to retry.


setup/ubuntu/bootstrap.sh
```diff
@@ -50,7 +50,7 @@ install_system_packages() {
 }
 
 create_directories() {
-    mkdir /opt/redash
+    mkdir -p /opt/redash
```
arikfr: Did it ever happen to you that `/opt` wasn't there? Otherwise why do we need this?

isomura: @arikfr 
When this script was terminated abnormally after /opt/redash directory was made, "/opt/redash is exists." error occurred everytime with this mkdir command.
P option ignore if /opt/redash is exists.

arikfr: In general this script is not idempotent and I would be careful to run it again "blindly".

arikfr: But it's safe enough to make this change, so merging. Thanks.

isomura: Thank you!

PR #2045, opened 2017-10-18
Added 'Use Glue Data Catalog' options in Athena

myouju:
Enabled to get schema data from glue data catalog with get_schema


arikfr:
Thanks!

PR #2048, opened 2017-10-21
Elasticsearch: Collect doc_count field from aggregation

arjan:
The `doc_count` field, which is part of the Elasticsearch aggregation result, was ignored.
This fix makes the following class of queries return a `doc_count` column, which then can be used for histograms and the like:

```
{
    "aggs": {
        "sessions": {
            "date_histogram": {
                "field": "start_time",
                "interval": "day"
            }
        }
    },
    "size": 0
}
```

### DIFF
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -219,7 +219,7 @@ class BaseElasticSearch(BaseQueryRunner):
                     result_row = get_row(rows, row)
                     collect_aggregations(mappings, rows, parent_key, value, result_row, result_columns, result_columns_index)
                     if 'doc_count' in value:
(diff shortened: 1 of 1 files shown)


redash/query_runner/elasticsearch.py
```diff
@@ -218,6 +218,8 @@ def collect_aggregations(mappings, rows, parent_key, data, row, result_columns,
                 for value in data:
                     result_row = get_row(rows, row)
                     collect_aggregations(mappings, rows, parent_key, value, result_row, result_columns, result_columns_index)
+                    if 'doc_count' in value:
+                        collect_value(mappings, result_row, 'doc_count', value['doc_count'], 'long')
```
arikfr: It should be `integer` and not `long`.

arikfr:
Thanks!

PR #2053, opened 2017-10-24
SQL Query Runner for MAPD database

cdessanti:
The query Runner support querying tables/views, catalog and cardinality of objects


arikfr:
Thanks!

PR #2054, opened 2017-10-25
Allow running any command inside the container via the docker entrypoint script.

jezdez:
This allows running generic things like opening a bash shell: ```docker-compose run server bash```


arikfr:
🤘

PR #2057, opened 2017-10-26
Fix #1881 Cutting-off minutes when schedule_failures is too large

### LINKED ISSUE #1881: OverflowError: date value out of range in should_schedule_next
Hi,

My redash never refresh all my queries. And I found that there was an "OverflowError: date value out of range" exception raised at:
https://github.com/getredash/redash/blob/v1.0.3/redash/models.py#L703
```
    if failures:
        next_iteration += datetime.timedelta(minutes=2**failures)
```

I think maybe it should check the date value range before applying the delta.

kitsuyui:
This is for https://github.com/getredash/redash/issues/1881#issuecomment-339656216 (1) .


arikfr:
Thanks! Closing this as a similar fix was merged in #3899.

Issue #2051, opened 2017-10-24
Internal server when creating Excel file from dataset with a field containing array


thijsc:
### Issue Summary

Internal server when downloading a dataset that contains a field containing an array in Excel format. In our case this field was created using an aggregation query on MongoDB.

### Steps to Reproduce

1. Create a query with a field that contains an array.
2. Click "Download dataset > Download as Excel file"
3. The Excel file contains the text `{"message": "Internal Server Error"}%` and the log on the server tracks this error:

```
[2017-10-24 14:25:44,407][PID:1468][ERROR][redash] Exception on /api/queries/6/results/128.xlsx [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python2.7/dist-packages/flask_restful/__init__.py", line 477, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flask_login/utils.py", line 228, in decorated_view
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flask/views.py", line 84, in view
    return self.dispatch_request(*args, **kwargs)
  File "/opt/redash/redash.2.0.1.b3080/redash/handlers/base.py", line 27, in dispatch_request
    return super(BaseResource, self).dispatch_request(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flask_restful/__init__.py", line 587, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/opt/redash/redash.2.0.1.b3080/redash/permissions.py", line 47, in decorated
    return fn(*args, **kwargs)
  File "/opt/redash/redash.2.0.1.b3080/redash/handlers/query_results.py", line 224, in get
    response = self.make_excel_response(query_result)
  File "/opt/redash/redash.2.0.1.b3080/redash/handlers/query_results.py", line 252, in make_excel_response
    return make_response(query_result.make_excel_content(), 200, headers)
  File "/opt/redash/redash.2.0.1.b3080/redash/models.py", line 720, in make_excel_content
    sheet.write(r + 1, c, row.get(name))
[...]

PR #2060, opened 2017-10-28
Fix error when exporting list data as Excel file

kyoshidajp:
About #2051.

XlsxWriter's `.write()` is not support list data. My solution is convert to string.

And, I have already checked the following BSON Types in MongoDB is OK.

- Arrays
- String
- Integer
- Boolean
- Object
- Date
- Timestamp
- Symbol


arikfr:
Thanks!

thijsc:
Awesome, thanks!

Issue #2055, opened 2017-10-25
Membership not deleted from users table in the database after group is removed


jeduardo:
### Issue Summary

When a group is removed from Redash its id remains in the user membership list in the database.

A summary of the issue and the browser/OS environment in which it occurs.

### Steps to Reproduce

0. Before starting with the process, the database looks like this for 2 users and the default groups:

```
postgres=# select id, groups from users;
 id | groups 
----+--------
  2 | {2}
  1 | {1,2}
(2 rows)

postgres=# select id, name from groups;
 id |  name   
----+---------
  1 | admin
  2 | default
(2 rows)
```
1. Create a new group through the Redash web interface. It shows up on the database as expected. Membership is not yet assigned or changed.

```
postgres=# select id, groups from users;
 id | groups 
----+--------
  2 | {2}
  1 | {1,2}
(2 rows)

postgres=# select id, name from groups;
 id |      name       
----+-----------------
  1 | admin
  2 | default
 15 | Test Membership
(3 rows)
```
2. Assign one of the users to the new group through the Redash web interface. Membership is reflected in the database:

``` 
postgres=# select id, groups from users;
 id |  groups  
----+----------
  2 | {2}
  1 | {1,2,15}
(2 rows)

```

3. Delete the group using the Redash interface. The membership from the assigned user to the now non-existing group is not reflected in the database.

```
postgres=# select id, name from groups;
 id |  name   
----+---------
  1 | admin
  2 | default
(2 rows)

postgres=# select id, groups from users;
 id |  groups  
----+----------
  2 | {2}
  1 | {1,2,15}
(2 rows)
```
The user with id=1 remains a member of group 15 (Test Membership), however this group does not exist anymore.

I would expect that when the group is removed, membership references to the group are also removed.

### Technical details:

* Redash Version: 2.0.0+b2990
* Browser/OS: Chrome/Linux
* How did you install Redash: Docker compose (dev environment)

kyoshidajp:
@arikfr Could you close this PR? I already fixed #2061. However, I didn't describe closing keyword like `Fixed`. Sorry.

arikfr:
Fixed in #2061. Thank, @kyoshidajp .

PR #2061, opened 2017-10-29
Delete the group id from user when deleting a group

kyoshidajp:
About #2055.

Since there was no deletion processing, I added it.

### DIFF
--- a/redash/handlers/groups.py
+++ b/redash/handlers/groups.py
@@ -70,6 +70,7 @@ class GroupResource(BaseResource):
         members = models.Group.members(group_id)
         for member in members:
             member.group_ids.remove(int(group_id))
+            models.db.session.add(member)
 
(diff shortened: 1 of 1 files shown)


redash/handlers/groups.py
```diff
@@ -67,6 +67,10 @@ def delete(self, group_id):
         if group.type == models.Group.BUILTIN_GROUP:
             abort(400, message="Can't delete built-in groups.")
 
+        members = models.Group.members(group_id)
+        for member in members:
+            member.group_ids.remove(int(group_id))
```
arikfr: Need to add `member` to the session too?

kyoshidajp: @arikfr It was my mistake. Fixed it. Thanks.

arikfr:
Thanks!

PR #1543, opened 2017-01-24
Bring the Less source of superflat's theme into the project

arikfr:
- [x] Replace Sass loader with Less
- [x] Add SuperFlat's less sources into the project
- [x] Move redash.css definitions to Less
- [ ] Cleanup the Less sources and remove unused ones
- [ ] Move stylesheets imports from index.js to main.less.

PR #2062, opened 2017-10-29
Add: flexible dashboard layout

### LINKED ISSUE #335: More flexible widgets grid (allow finer control over height/width)
### Current implementation

Currently editing a dashboard works as follow:

1. When adding a widget you select its width: regular (half width) or double (full width). Height is set based on the content.
2. When opening the "Edit Dashboard" dialog, you can rearrange the order of the widgets.

Edit Dashboard Dialog

Issues with the current approach:

* You can't change widget size after placing it.
* You can't change widget's height.
* You can't control widget's width beyond the current two options.

### Solution

To solve this we want to have a flexible (_but still responsive_) grid that the user can freely place and move widgets on. 

* Each visualization will define minimum width and height sizes (derived from the visualization type). This will be the default sizes for the widget when placed on the grid, but the user will be able to resize it to be larger (but not smaller) afterwards.
* Once entering dashboard "edit mode", the user will be able to move around and resize the widgets in place instead of a dedicated UI.
* We should probably change other behaviors in the UI (move add widget to be part of the edit mode, change dashboard title editing to be in place, etc), but we will leave it to second iteration after the grid behavior is implemented.

### Technical Notes

* We currently use `angular-gridster` to allow moving around the widgets in the Edit Dashboard dialog. Maybe we can use it for the grid or a similar library.
* Currently the widget knows nothing about its placement, but only its size (width). The dashboard has a layout property, which is an array of arrays of widgets ids (each item in the array is a row/array of widgets). This is error prone and makes layout changes harder. If possible, it will be better if each widget retains its own placement information.
* The API to update a widget currently supports only updating a textbox widget. We will need to extend it to support all widget types.

kravets-levko:
Dashboard implementation based on `angular-gridster`. Backward compatible as much as it was possible, on first dashboard update layout will be converted to new format. Dashboard will be saved implicitly when adding and removing widgets, and explicitly - when editing by user.

When adding new widget and prior to first save widget will automatically fit it's contents. Once saved, height can be only explicitly set. Also, auto-sizing is enabled for existing dashboards for backward compatibility.

Visualizations can set constraints - min/max row/column count. Legacy option `height` is also supported, but `minRows` has a priority over it. By default widget has no constraints set (minimal size is 1x1 cell, maximal is unlimited).

Removed unnecessary controls from Add widget and Edit dashboard dialogs.

Closes #335.


arikfr:
> about defaults. Currently I set dashboard to have 3 columns (we can set any, it's up to you), 

Let's have 6, as counter widgets can be really small. Usually people will use at most 3 columns, but for things like the counter widget, having more columns is great. One thing we will need to do is to remove some UI elements on the smaller widgets - but this can be a future improvement (unless can be easily achieved with CSS media queries or similar).

> and default dimensions for widgets are initialized from width/height (does not exist currently). As I understand, we should make some changes to backend so it will return default dimensions for widgets. So I'd like to talk about it, if it's possible.

Currently widget take as much height as they need -- is it possible to preserve this? 
If not, we will need to define default heights (and widths) per visualization type.

If you like to chat, you can either ping me on [Gitter](gitter.im/getredash/redash) or [Slack](slack.redash.io).

kravets-levko:
@arikfr Can you review this, please? Thanks!

client/app/assets/css/superflat_redash.css
```diff
@@ -7913,6 +7913,44 @@ fieldset[disabled] .btn-inverse.active {
   position: relative;
   box-shadow: 0 1px 1px rgba(0, 0, 0, 0.07);
 }
+
+.dashboard-wrapper {
```
arikfr: Let's keep all new styles next to the components JS/HTML code.

client/app/config/index.js
```diff
@@ -46,8 +46,42 @@ const requirements = [
...
+    handles: ['n', 'e', 's', 'w', 'ne', 'se', 'sw', 'nw'],
+  },
+  draggable: {
+    enabled: true, // whether dragging items is supported
+  },
+};
+
 const ngModule = angular.module('app', requirements);
 
+ngModule.constant('dashboardGridOptions', dashboardGridOptions);
```
arikfr: How about we move this to `app/components/dashboard_grid_options.js` and just import it?

client/app/config/index.js
```diff
@@ -46,8 +46,42 @@ const requirements = [
...
+  floating: true,
+  swapping: true,
+  width: 'auto',
+  colWidth: 'auto',
+  rowHeight: 100,
+  margins: [0, 0],
+  outerMargin: false,
+  sparse: false,
+  isMobile: false,
+  mobileBreakPoint: 1200,
```
arikfr: We need to change this to be lower. I will play around with a few values to see what works better.

client/app/pages/dashboards/public-dashboard-page.html
```diff
@@ -15,7 +15,11 @@
     <filters ng-if="$ctrl.dashboard.dashboard_filters_enabled"></filters>
   </div>
 
-    <div ng-repeat="row in $ctrl.dashboard.widgets" class="row">
-      <dashboard-widget ng-repeat="widget in row" widget="widget" dashboard="$ctrl.dashboard" public="true"></dashboard-widget>
+
+  <div gridster="$ctrl.dashboardGridOptions">
+    <div ng-repeat="widget in $ctrl.dashboard.widgets">
+      <dashboard-widget widget="widget" gridster-item="widget.options.position"
```
arikfr: Don't we need the auto-height directive here too?

It feels like we should encapsulate the whole dashboard grid logic in its own component that will be reused here and in the regular dashboard.

client/app/pages/dashboards/dashboard.js
```diff
@@ -3,9 +3,58 @@ import template from './dashboard.html';
...
+      handles: ['n', 'e', 's', 'w', 'ne', 'se', 'sw', 'nw'],
+      stop: saveDashboard,
+    },
+    draggable: {
+      enabled: false,
+      stop: saveDashboard,
+    },
+  });
+
+  $scope.$watch(
```
arikfr: We try to move away from using `$watch` whenever possible. In this case, I believe there are two cases when the widgets list might change: when a widget is added or when it's removed. We can have callbacks in the code for both cases instead of constantly checking the state of the array.

Another question is why do we need to save the dashboard in such case? To update the locations of the widgets?

client/app/services/dashboard.js
```diff
@@ -1,8 +1,33 @@
...
+          widget.options.position = extend({}, {
+            row: rowIndex,
+            col: widgetIndex * dashboardV1ColumnSize,
+            sizeX: dashboardV1ColumnSize * widget.width,
+            sizeY: -1, // auto
+          }, widget.options.position);
+          return widget;
+        }),
+      );
+    }
```
arikfr: I wonder if we should do the migration here or just do it once as a database migration. 🤔

client/app/assets/css/superflat_redash.css
kravets-levko: Sorry, I'm a bit new to the codebase, so probably chose inappropriate place for the styles. Where do you usually keep component-specific CSS?

client/app/pages/dashboards/dashboard.js
kravets-levko: Yes, because when we add/remove widgets, other widgets' positions may change, so we need to save it all. I know about cons of using `$watch`, therefore I'm just watching `length` property, not entire list of objects, - it's fast and will not lead to performance issues. My opinion is - if we need this in the single place, and is clear and fast enough - it can be implemented in (such) a simple way.

client/app/pages/dashboards/public-dashboard-page.html
kravets-levko: Got it, will take a look.

client/app/services/dashboard.js
kravets-levko: It's up to you. This code works as expected, it's also simple and fast, but it exists only to support legacy data structures. So if it's safe to perform migration at once - we can remove this.

client/app/assets/css/superflat_redash.css
arikfr: Most of the CSS code is still in this `superflat_redash.css` file 😞 There is #1543 where there was a try to bring in the LESS sources of the Superflat into the project, but it wasn't merged in time and now requires some rebase.

But all new CSS I try to place along side the component code, for example: https://github.com/getredash/redash/tree/master/client/app/components/app-header

client/app/services/dashboard.js
arikfr: OK, I will think about it some more and reach a conclusion before we finalize this.

client/app/pages/dashboards/dashboard.js
arikfr: There is the performance concern, which you address nicely with watching only the length. But there are other concerns of maintainability and making the code clear -- future reader might not understand why we need this.

But there is a worse issue here: because of the watch, `saveDashboard` is called every time you load the dashboard.

We already have everything "provisioned" to run code after adding a widget:

https://github.com/getredash/redash/blob/900d5588578eae027dd7a6caf5f91383d8893da2/client/app/pages/dashboards/dashboard.js#L181

And after delete:

https://github.com/getredash/redash/blob/900d5588578eae027dd7a6caf5f91383d8893da2/client/app/pages/dashboards/dashboard.html#L65

Maybe add new controller function for post widgets change tasks and run there `extractGlobalParams` and `saveDashboard`.

client/app/assets/css/superflat_redash.css
kravets-levko: Okay, got it. If I'll have some more time - I try to review that PR and fix it. LESS is awesome :wink:

client/app/pages/dashboards/dashboard.js
arikfr: Btw, another reason why I try to stay away from things like `$watch`, is that it makes the code more portable to other frameworks. 😉

kravets-levko: You're right. Will think how to improve this code.

client/app/assets/css/superflat_redash.css
arikfr: 👌 btw, instead of actually rebasing that branch I would copy over the changes to a new branch of master and then compare what was done in `app/assets/css` since the time this branch was created and apply those.

client/app/pages/dashboards/dashboard.js
```diff
@@ -3,9 +3,58 @@ import template from './dashboard.html';
 import shareDashboardTemplate from './share-dashboard.html';
 
 function DashboardCtrl(
-  $rootScope, $routeParams, $location, $timeout, $q, $uibModal,
+  $rootScope, $scope, $routeParams, $location, $timeout, $q, $uibModal,
   Title, AlertDialog, Dashboard, currentUser, clientConfig, Events,
+  dashboardGridOptions, toastr,
 ) {
+  this.saveInProgress = false;
+  const saveDashboard = _.debounce(() => {
```
arikfr: Maybe `saveDashboardLayout`?

arikfr:
@kravets-levko finally had the chance to write down all the UI comments:

- [ ] It seems like every time you add a new widget, it gets added on a new row instead of using free space available on an existing row.
- [ ] Instead of saving on each move, let's show two buttons: [Cancel | Done Editing] -- "Cancel" will restore previous layout, while "Done Editing" will save. Otherwise on slow connections/when having many widgets it freezes the editing for too long (and people might want to cancel what they done).
- [ ] When clicking on refresh, the spinner should be centered.
- [ ] When we are in responsive mode (everything stacked), angular-gridster disabled editing as well. We need to disable the edit button in this case.
- [ ] When resizing, there is a dark silhouette that shows the size the widget will actually take (according to the next column). It needs to be somewhat darker, as on some monitors I didn't see it.
- [ ]  The resize arrow is out of the widget (see screenshot). This might be because of the scrollbars (Chrome on Linux).
image
- [ ] Widgets get scrollbars, even when not needed - like the "Queries Count" widget on the `/dashboard/counters` dashboard.
- [ ] I couldn't resize a widget that contains a table and it has huge padding/margin at the bottom (see example in the counters dashboard).

Not part of the dashboard layout behavior, but related and need to be addressed:

- [ ] Chart widgets don't resize when the widget changes size.
- [ ] Counter widget don't resize their text. I think we need to set the text size to be relative to the container and have some maximum width after which it will get cut with an ellipsis.

client/app/config/index.js
kravets-levko: Current version has a breakpoint at Bootstraps `-lg-` - `1200px`, so it's just kept as-is.
I think, the best decision it to make it adjustable (via dashboard options) - because the best breakpoint value depends on dashboard layout and configuration of widgets. How do you this - does it make any reason?

client/app/config/index.js
kravets-levko: I will extract it into own file, but leave in the `config` folder (not `components`) - because it's actually a config, not a component.

client/app/pages/dashboards/public-dashboard-page.html
kravets-levko: I think, we should add some issues labeled as `refactoring`, and this one should be among them (as well as moving to LESS) - otherwise this PR tends to grow too large.

kravets-levko:
So much updates :smile: 
The first set of fixes is already here, and now I'll go through your list and fix all the possible.

arikfr:
👍 

This introduces a big behavior change, hence the amount of feedback. But most of those are small stuff.

client/app/pages/dashboards/public-dashboard-page.html
arikfr: 👍 
It's the "Tech Debt" label. You're welcome to add those.

client/app/config/dashboard-grid-options.js
```diff
@@ -0,0 +1,35 @@
...
+    handles: ['n', 'e', 's', 'w', 'ne', 'se', 'sw', 'nw'],
+  },
+  draggable: {
+    enabled: true, // whether dragging items is supported
+  },
+};
+
+export default function init(ngModule) {
+  ngModule.constant('dashboardGridOptions', dashboardGridOptions);
+}
```
arikfr: Actually had something simpler in mind:

1. In this file:

```
export default {
... settings ...
};
```

2. Where we need this value: `import dashboardGridoptions from '@config/dashboard-grid-options'`.

But it's working, so let's leave as is.

kravets-levko:
@arikfr I pushed some more fixes. First 6 notices were fixed; about last 4 ones: I fixed some (more or less) to make it working smoother; but all of them are related not to dashboard layout, but to widget contents - widgets have purely implemented layout. I think, I'll review all of them, and properly fix within another PR (this one it already large).

client/app/config/dashboard-grid-options.js
```diff
@@ -0,0 +1,35 @@
...
+  floating: true,
+  swapping: true,
+  width: 'auto',
+  colWidth: 'auto',
+  rowHeight: 50,
+  margins: [15, 15],
+  outerMargin: false,
+  sparse: false,
+  isMobile: false,
+  mobileBreakPoint: 1200,
```
arikfr: Let's use 800 here. 

In the future we might want to implement a better logic here, which will take into account the widgets sizes when stacking them. So small ones won't get stacked, but that's for the future.

client/app/pages/dashboards/dashboard.html
```diff
@@ -42,6 +57,7 @@
               </button>
               <ul class="dropdown-menu pull-right" uib-dropdown-menu>
                 <li ng-if="!$ctrl.dashboard.is_archived"><a ng-click="$ctrl.editDashboard()">Edit Dashboard</a></li>
+                <li ng-if="!$ctrl.dashboard.is_archived" ng-class="{disabled: $ctrl.isGridDisabled}"><a ng-click="$ctrl.editLayout(true)">Edit Layout</a></li>
```
arikfr: It seems that the visual cue it gives that it's disabled is the cursor type, but on mobile/tablets there is no cursor... Let's just hide this button when it's not supposed to be available.

Btw, can we use `.visible-md` and `.visible-lg` classes instead of the custom JS logic, no?

kravets-levko: In Boostrap, by default, that class also changes look off the item. But seems that `superflat_redash` does too much customization :smile: 

About custom JS logic: at this moment - yes, we can use `.visible-*` classes, but it's bad idea for another reason: it's not synchronized with `angular-gridster`. Suppose someone in future will decide to set breakpoint to `900px` - and this menu item will start showing/hiding at a wrong time. So it will look a bug. Therefore it's better to use events provided by `angular-gridster` to know exactly when it's in mobile mode or not. In addition, that event is triggered only when grid mode changes, not on every resize event.

arikfr: 👍 

got it. makes sense, so let's just hide it based on the gridster's event.

arikfr:
🚀

guzhenping:
hello, everyone ! My redash is from master branch. but, when l archived a query  existed in a dashboard and added a new chart to that dashboard, it raised:

[2018-01-10 15:50:31,256][PID:42][ERROR][redash] Exception on /api/widgets [POST]
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1625, in dispatch_request
  File "/usr/lib/python2.7/site-packages/flask_restful/__init__.py", line 477, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/flask_login/utils.py", line 228, in decorated_view
    return func(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/flask/views.py", line 84, in view
    return self.dispatch_request(*args, **kwargs)
  File "/med/redash/handlers/base.py", line 27, in dispatch_request
    return super(BaseResource, self).dispatch_request(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/flask_restful/__init__.py", line 587, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/med/redash/permissions.py", line 48, in decorated
    return fn(*args, **kwargs)
  File "/med/redash/handlers/widgets.py", line 56, in post
    if neighbour_widget.width == 1:
AttributeError: 'NoneType' object has no attribute 'width'

arikfr:
@guzhenping this is because the old logic is still there 🤔 We need to remove it, I opened #2218 to track this.

PR #2066, opened 2017-10-31
replaced -y option since it didn't perform any action with -qq

eorochena:
Replaced -y option since it didn't perform any action with the apt-get update command; the -qq option will only display errors and warning when updating the repositories.

Thanks.

### DIFF
--- a/setup/ubuntu/bootstrap.sh
+++ b/setup/ubuntu/bootstrap.sh
@@ -37,7 +37,7 @@ create_redash_user() {
 }
 
 install_system_packages() {
-    apt-get -y update
+    apt-get -qq update
     # Base packages
     apt install -y python-pip python-dev nginx curl build-essential pwgen
     # Data sources dependencies:


setup/ubuntu/bootstrap.sh
```diff
@@ -37,7 +37,7 @@ create_redash_user() {
 }
 
 install_system_packages() {
-    apt-get -y update
+    apt-get -qq update
```
arikfr: Are we sure that `apt-get update` will never ask any questions? It seems harmless to keep `-y` so maybe have both?

eorochena: I'm positive, the only time that I get prompted is when installing, removing or upgrading the system but I found something interesting in the man page for apt-get

"-q, --quiet
           Quiet; produces output suitable for logging, omitting progress
           indicators. More q's will produce more quiet up to a maximum of 2.
           You can also use -q=# to set the quiet level, overriding the
           configuration file. Note that quiet level 2 implies -y; you should
           never use -qq without a no-action modifier such as -d, --print-uris
           or -s as APT may decide to do something you did not expect.
           Configuration Item: quiet."

arikfr:
Thank you for submitting this pull request. We redid all the setup scripts, so this is no longer relevant.

PR #2068, opened 2017-10-31
Copy parameters value when forking a query

### LINKED ISSUE #2036: Parameters are loosing their configuration on fork
### Issue Summary
When forking a query, its parameters are loosing their configuration.

### Steps to Reproduce

1. Create query with parameters
2. Fork it

### Technical details:

* Redash Version: commit 6982f48b34a66ef1aeb9fe42065e2e2a412e0d54

kyoshidajp:
Fix #2036.

Now, new query does not have their parameters value  when forking a query. I would like to copy the parameters value from an original query.


arikfr:
Super. Thank you once again! 🏅

PR #2072, opened 2017-11-01
🎉  Add: data source to run queries on top of query results

### LINKED ISSUE #1977: Queries using other queries as a data source
I couldnt find an existing ticket for this so made this one. 

Redash seems to advertise that it supports adding queries as first class data sources here: https://redash.io/help/queries/using-datasets-as-data-sources.html

It would be really cool to have this functionality land in redash. @arikfr any plans on open sourcing the preview, or would you accept any PRs that add this functionality directly into redash?

arikfr:
This is the long awaited Query Results data source, which let's the user run query on top of existing query results (documented [here](https://redash.io/help/queries/using-datasets-as-data-sources.html)).

Closes #1977.

### DIFF
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -1,0 +1,167 @@
+import json
+import logging
+import numbers
+import re
+import sqlite3
+
+from dateutil import parser
+
+from redash import models
+from redash.permissions import has_access, not_view_only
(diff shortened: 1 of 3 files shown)


dncpax:
Would like to propose that forcing to refresh the base queries removes control from the user. While not forcing the refresh delegates control to the user and what he/she setup for the base queries.
So what I'm doing really is proposing to not force refresh the base queries and trust the user knows best what he's doing. As it is, redash overules what we are trying to do.

Issue #1356, opened 2016-10-21
Table visualization v2


arikfr:
The goal is to address multiple issues we have with the current table visualization. We will do this in several steps:

Step 1:
~Replace current simple table visualization with [ag-Grid](https://www.ag-grid.com/). The main reason for picking ag-Grid is the fact that it's a robust library, with good support and it's framework agnostic.~

Step 2:
* Add configuration for the table visualization.
  - Select number of rows to show per page.
  - Select which column to show.
  - Select column types (how to render the column).
  - Select whether to render HTML in a column (default: false).
* Add option to create multiple tables for a single query.

Step 3:
More advanced configurations for the table visualization:

- Ability to convert a column into a link (#582).
- Enable search on results (#1220, this should be a toggle users can enable).

Future improvements: 
Conditional formatting?
Spark lines?

fernandosjp:
Hi @arikfr,

Would things like conditional fomatting be included here? I think it would be really interesting to be able to have informative tables such as the one bellow. Sparklines might be more complex but conidtional formatting with colors would already be a great step. 

![image](https://cloud.githubusercontent.com/assets/4928401/20709589/73785742-b61d-11e6-865b-5658c7734292.png)

PR #2077, opened 2017-11-07
Add ITEMS_PER_PAGE parameter

kyoshidajp:
Now, only 15 items display per query result page. I would like to check more items without page transition.

I add the parameter `ITEM_PER_PAGE` which enable to specify how many items can display in result list per page.


arikfr:
Before finishing release v4 we're going to have #1356 implemented, which will include the ability to control # of rows to display.

At first I thought of closing this pull request in favor of #1356, but maybe there is still a place for a global setting? WDYT?

kyoshidajp:
My opinion is that it would be better to control number of rows as a default configuration, aside from each query.

kyoshidajp:
This feature was implemented by #2993.

jezdez:
Thanks @kyoshidajp :)

Issue #1180, opened 2016-07-11
Is it possible to label pie charts after grouping?


fredthedead:
After grouping the data based on a column and having 2 pie charts side by side there is no labeling on top of each pie chart saying to which group value is matches... I didn't notice a way to enable this - is it possible?

arikfr:
Can you create an example on demo.redash.io (login with any Google account) or at least attach a screenshot?

fredthedead:
![screen_shot_2016-07-14_at_5_54_19_pm](https://cloud.githubusercontent.com/assets/19219041/16832455/a69e7d46-49ed-11e6-8e8d-81f2fee36c97.png)

(Sorry had to blur labels)

![screen_shot_2016-07-14_at_5_55_03_pm](https://cloud.githubusercontent.com/assets/19219041/16832440/96322fde-49ed-11e6-8b49-90a0450ce6e3.png)

dpreuss:
Any update on this issue? I appear to have run into the same problem...

arikfr:
Implemented in #2082.

PR #2082, opened 2017-11-12
Add: Pie chart subplot title

### LINKED ISSUE #2035: Pie Chart Title
Hi there,

there does not seem to be any way to add a title to the pie chart when one is displaying multiple pie charts per visualization. Could the title (which is shown on hover) be added above each of the charts?

deecay:
pie-subplot-title

Closes #2035 and #1180.

### DIFF
--- a/client/app/visualizations/chart/plotly.js
+++ b/client/app/visualizations/chart/plotly.js
@@ -252,7 +252,9 @@ const PlotlyChart = () => {
           const cellWidth = 1 / cellsInRow;
           const cellHeight = 1 / rows;
           const xPadding = 0.02;
-          const yPadding = 0.05;
+          const yPadding = 0.1;
+
+          scope.layout.annotations = [];
 
           each(scope.series, (series, index) => {
             const xPosition = (index % cellsInRow) * cellWidth;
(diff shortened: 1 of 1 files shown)


arikfr:
Nice!

client/app/visualizations/chart/plotly.js
```diff
@@ -252,7 +252,9 @@ const PlotlyChart = () => {
           const cellWidth = 1 / cellsInRow;
           const cellHeight = 1 / rows;
           const xPadding = 0.02;
-          const yPadding = 0.05;
+          const yPadding = 0.1;
```
arikfr: Why did we need to change this?

deecay: We need a little more space than now to make the second-row title to fit. The gap between the first row and second row is too tight for a row of text to fit nicely.

arikfr:
Thanks!

deecay:
FYI, Upcoming Plotly will have native support for Pie subplot titles.
[link](https://github.com/codrut3/plotly.js/blob/bce5add399f962241b6fd482f1eefa19b0d601d7/test/image/baselines/pie_title_variations.png)

arikfr:
Nice :) Thanks for the update, @deecay .

PR #2083, opened 2017-11-13
Responsive visualizations

kravets-levko:
Make visualizations to look better with new dashboard layout.

1. **Counter**: fit text to widget's width (not to height because widget's height may be dynamic; but, in general, if text of counter fits width - most likely it will fit height too due to nature of this viz).
2. **Table**: added default dimensions and constraints; if table will be too large - scrollbars will appear (only if needed);
3. **Map**: fixed position, added default dimensions and constraints.
4. **Pivot**: added default dimensions and constraints. There are still some UI issues (need to look closer), but if widget will have enough size - it will work well.
5. **Chart**: added default dimensions and constraints. Fits to container (both width and height).

**Refactoring**: moved some methods to better places; each directive is now in own file.

### DIFF
--- a/client/app/assets/css/redash.css
+++ b/client/app/assets/css/redash.css
@@ -705,3 +705,7 @@ div.table-name:hover {
 .dropdown-menu > .disabled > a{
  pointer-events: none;
 }
+
+.map-visualization-container {
+  height: 500px;
+}
--- a/client/app/assets/css/redash.css
+++ b/client/app/assets/css/redash.css
@@ -345,6 +345,15 @@ counter-renderer counter-target {
   display: block;
 }
 
+counter-renderer value .ruler,
+counter-renderer counter-target .ruler {
+  display: inline-block;
+  font: inherit;
+  line-height: inherit;
+  margin: 0;
+  padding: 0;
+}
+
 counter-renderer counter-target {
   color: #ccc;
 }
@@ -714,4 +723,8 @@ div.table-name:hover {
 .plotly-chart-container {
   height: 400px;
   overflow: hidden;
+}
+
+visualization-renderer .pagination {
+  margin: 0;
 }
--- a/client/app/assets/css/redash.css
+++ b/client/app/assets/css/redash.css
@@ -321,7 +321,7 @@ to add those CSS styles here. */
 }
 
 pivot-table-renderer > table, grid-renderer > div, visualization-renderer > div {
-  overflow: auto;
+  /*overflow: auto;*/
 }
 
 counter-renderer {
--- a/client/app/assets/css/redash.css
+++ b/client/app/assets/css/redash.css
@@ -339,6 +339,7 @@ counter-renderer counter {
 }
 
 counter-renderer value,
+counter-renderer counter-name,
 counter-renderer counter-target {
   font-size: 1em;
   line-height: 1em;
@@ -346,8 +347,12 @@ counter-renderer counter-target {
 }
 
 counter-renderer value .ruler,
+counter-renderer counter-name .ruler,
 counter-renderer counter-target .ruler {
-  display: inline-block;
+  display: block;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
   font: inherit;
   line-height: inherit;
   margin: 0;
--- a/client/app/assets/css/redash.css
+++ b/client/app/assets/css/redash.css
@@ -341,6 +341,7 @@ counter-renderer counter {
 counter-renderer value,
 counter-renderer counter-target {
   font-size: 1em;
+  line-height: 1em;
   display: block;
 }
 
@@ -708,4 +709,9 @@ div.table-name:hover {
 
 .map-visualization-container {
   height: 500px;
+}
+
+.plotly-chart-container {
+  height: 400px;
+  overflow: hidden;
 }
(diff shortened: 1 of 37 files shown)


arikfr:
I played with the last version and here's some things I noticed:

* I created a new dashboard (http://localhost:8080/dashboard/resize-test). After adding the second widget there it got stuck. It still stuck after reloading the page, seems to be busy with something? The same happens with this dashboard: http://localhost:8080/dashboard/counters.
* On the resize-test dashboard, the Dashboards Count widget has a very large padding on the top, which creates a scroll.
* On the resize-test dashboard, the Queries Count widget is the default size it gets added with -- and still it has scroll to see all the content...

arikfr:
The map visualization doesn't seem to auto resize after finishing with the resize. It does update after reloading.

arikfr:
Reviewed the fixed version. Some new comments:

* When opening the visualizations example dashboard, it keeps vertically resizing itself.
* We should apply a minimum rows size of 4 for all widgets, it doesn't make sense to make smaller widgets.
* Counter visualization: the font size should be a function of the _height_ rather than _width_. For width we can use `text-overflow: ellipsis`, so the user will decide if they want the text to be cut or to increase size (when relevant).
* Table visualization: there is a big white space at the bottom with the current default size. How about we decrease `defaultRows` to 14 and decrease the spacing between the pagination control and the table? This way everything will fit nicely.
* Map visualization: when I change it's size, it seems that the Leaflet control expands but the map itself doesn't so there is some grey area. If you don't see this on your machine, I will send a gif showing what I mean.

kravets-levko:
@arikfr Fixed everything (even map - I reproduced that behavior), except of counter: we cannot calculate font size relative to height, because it's not compatible auto-height feature and will make circular dependency: 1. we'll calculate font size depending on height 2. height will reduce 3. auto-size will reduce widget's height 4. goto step 1. So, I think, user just should set large enough height (comparing to width - widget should be almost square) - so text will fit both with and height. In addition, I think that `text-overflow: ellipsis` is not so good solution as it hides a part of text without ability to view it; but when we have vertical scroll - user will always see amount, and he can reach description if needed.

kravets-levko:
@arikfr One more set of changes: counters now fits to height of container; if text is too wide - it's showing with ellipsis (`...`). Also, now widgets in old-style dashboards will not have auto-height by default, but use visualizations defaults (that means - if visualization has, say, height of 10 rows by default - it will be used, otherwise auto-height will be enabled).

client/app/assets/css/redash.css
```diff
@@ -321,26 +321,44 @@ to add those CSS styles here. */
 }
 
 pivot-table-renderer > table, grid-renderer > div, visualization-renderer > div {
-  overflow: auto;
+  /*overflow: auto;*/
```
arikfr: At this point I assume we can just delete the commented lines, no?

client/app/components/dashboards/add-widget-dialog.js
```diff
@@ -58,74 +58,45 @@ const AddWidgetDialog = {
...
+            { position: {} },
+          );
+          const w = new Widget(response.widget);
+
+          const position = this.dashboard.calculateNewWidgetPosition(w);
+          w.options.position.col = position.col;
+          w.options.position.row = position.row;
+
+          // Save it with new position
+          return w.$save().then(() => w);
```
arikfr: Why not calculate the position before saving, thus eliminating the need to call the server twice?

kravets-levko: Because before the first save, `visualization` property is not available (and, as i understand, only server can populate it), but it's needed to properly calculate dimensions of widget.

client/app/directives/resize-event.js
```diff
@@ -0,0 +1,71 @@
...
+    });
+  }
+}
+
+function resizeEvent() {
+  return {
+    restrict: 'A',
+    link($scope, $element, attrs) {
+      const node = $element[0];
+      const exists = _.find(items, item => item.node === node);
```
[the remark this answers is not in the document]
kravets-levko: `Map` is [not well supported](https://developer.mozilla.org/uk/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility) by browsers. In addition, this line will be executed only on first widget render, so it shouldn't lead to performance issues (I think, only when dashboard will have about few thousands widgets...)

client/app/assets/css/redash.css
kravets-levko: If it's not used - yes, it definitely should be removed. But I'll check it to be sure (ATM I don't remember why I commented it out instead of deleting)

client/app/directives/resize-event.js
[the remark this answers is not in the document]
arikfr: OK. 👍 

Although "not well supported" -> that's what we have Babel for :)

client/app/components/dashboards/add-widget-dialog.js
arikfr: If I'm not mistaken, `this.selectedVis` is the same object the server will return.

client/app/directives/resize-event.js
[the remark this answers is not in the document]
kravets-levko: Ahahah :smiley: I missed that we're using Babel :smiley: Will change this, of course :+1:

client/app/components/dashboards/add-widget-dialog.js
kravets-levko: I'll check, and if they are the same - I'll remove second request.

arikfr:
👍 thanks :-)

PR #2089, opened 2017-11-17
Add Qubole query runner

msumit:
Adding support for Qubole as a data source and running hive queries against it.

### DIFF
--- a/redash/query_runner/qubole.py
+++ b/redash/query_runner/qubole.py
@@ -1,0 +1,132 @@
+from __future__ import absolute_import
+import time
+import requests
+import logging
+from cStringIO import StringIO
+
+from redash.query_runner import BaseQueryRunner, register
+from redash.query_runner import TYPE_STRING
+from redash.utils import json_dumps
+
+try:
+    import qds_sdk
+    from qds_sdk.qubole import Qubole as qbol
+    from qds_sdk.commands import Command, HiveCommand, PrestoCommand
+    enabled = True
+except ImportError:
+    enabled = False
+
+
+class Qubole(BaseQueryRunner):
+
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "endpoint": {
+                    "type": "string",
+                    "title": "API Endpoint",
+                    "default": "https://api.qubole.com"
+                },
+                "token": {
+                    "type": "string",
+                    "title": "Auth Token"
+                },
+                "cluster": {
+                    "type": "string",
+                    "title": "Cluster Label",
+                    "default": "default"
+                },
+                "query_type": {
+                    "type": "string",
+                    "title": "Query Type (hive or presto)",
+                    "default": "hive"
+                }
+            },
+            "order": ["endpoint", "token", "cluster"],
+            "required": ["endpoint", "token", "cluster"],
+            "secret": ["token"]
+        }
+
+    @classmethod
+    def enabled(cls):
+        return enabled
+
+    @classmethod
+    def annotate_query(cls):
+        return False
+
+    def test_connection(self):
+        headers = self._get_header()
+        r = requests.head("%s/api/latest/users" % self.configuration['endpoint'], headers=headers)
+        r.status_code == 200
+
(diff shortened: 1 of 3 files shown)


redash/query_runner/qubole.py
```diff
@@ -0,0 +1,130 @@
...
+        headers = {"Content-type": "application/json", "Accept": "application/json",
+                   "X-AUTH-TOKEN": self.configuration['token']}
+        h = httplib2.Http()
+        resp, content = h.request("%s/api/latest/users" % self.configuration['endpoint'],
+                                  "HEAD", headers=headers)
+        resp.status == 200
+
+    def run_query(self, query, user):
+        try:
+            import qds_sdk
```
arikfr: Please use a similar pattern to what we have here:
https://github.com/getredash/redash/blob/1b7cc3731c6ad5337855bcadbd5fe1ea64287691/redash/query_runner/big_query.py#L17-L26

This way it will also properly indicate whether the data source type is really supported.

redash/query_runner/qubole.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @classmethod
+    def annotate_query(cls):
+        return False
+
+    def test_connection(self):
+        headers = {"Content-type": "application/json", "Accept": "application/json",
+                   "X-AUTH-TOKEN": self.configuration['token']}
+        h = httplib2.Http()
+        resp, content = h.request("%s/api/latest/users" % self.configuration['endpoint'],
```
arikfr: Please use the `requests` library as everywhere else in the code.

redash/query_runner/qubole.py
```diff
@@ -0,0 +1,130 @@
...
+            from qds_sdk.qubole import Qubole as qbol
+            from qds_sdk.commands import Command, HiveCommand
+        except ImportError as e:
+            raise(e)
+
+        qbol.configure(api_token=self.configuration['token'],
+                       api_url='%s/api' % self.configuration['endpoint'])
+
+        hc = HiveCommand.create(query=query, label=self.configuration['cluster'])
+        print("Qubole command created with Id: %s and Status: %s", hc.id, hc.status)
```
arikfr: Please change those into `logging.debug` calls instead of `print` statements.

redash/query_runner/qubole.py
```diff
@@ -0,0 +1,130 @@
...
+            print("Qubole command Id: %s and Status: %s", hc.id, hc.status)
+
+        rows = []
+        columns = []
+        error = None
+
+        if(hc.status == 'done'):
+            iso = datetime.datetime.utcnow().isoformat()
+            fp = open('/tmp/%s_%s' % (hc.id, iso), 'wb')
+            hc.get_results(fp, True, None, True, None, ['true'])
```
arikfr: Instead of reading/writing a file, did you consider using [cStringIO](https://docs.python.org/2/library/stringio.html#module-cStringIO)?

arikfr: Also to make the code clearer, please use the names of the arguments, i.e.:

```python
hc.get_results(fp=fp, inline=True, delim=None, fetch=True, qlog=None, arguments=['true])
```

Also, from reading their documentation I think we should use `fetch=False`.

msumit:
Thanks for the review @arikfr. will work on the comments.

redash/query_runner/qubole.py
msumit: done

redash/query_runner/qubole.py
msumit: done

redash/query_runner/qubole.py
msumit: done.. but used `logging.info`..

redash/query_runner/qubole.py
msumit: both are done

msumit:
@arikfr plz review, fixed all the suggestions.

msumit:
Hi @arikfr can this PR be merged?

arikfr:
Thanks!

PR #2090, opened 2017-11-19
Add a description field to a dashboard

kyoshidajp:
Add a description field to a dashboard like a query's one.

# Screenshot

## Edit dialog

before
dashboard_before

after
dashboard_after2

## Detail page

| before | after |
| ------ | ----- |
| dashboard_detail_before | dashboard_after |

### DIFF
--- a/client/app/components/dashboards/edit-dashboard-dialog.html
+++ b/client/app/components/dashboards/edit-dashboard-dialog.html
@@ -6,6 +6,9 @@
   <p>
     <input type="text" class="form-control" placeholder="Dashboard Name" ng-model="$ctrl.dashboard.name" autofocus>
   </p>
+  <p>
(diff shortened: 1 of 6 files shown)


deecay:
Can this handle Markdown? I was thinking of using this feature like "For more details go [here](dummy)".

arikfr:
@deecay I'm not sure it should. And you can always add a textbox widget for such thing.

deecay:
@arik Yeah that's true, but textbox widget always takes 185px hight (one 'block' of grid, I guess?). It's a bit of an overkill for a oneliner description. Maybe a lower min-height of textbox widget is what I want.

arikfr:
Maybe :)

arikfr:
Hi,

Thanks.

PR #2092, opened 2017-11-22
Fix doesn't work  mysql query cancel button

jasonsmithj:
Cancel button is not working when I use mysql. 
In Postgres version is working correctly.


arikfr:
Thanks!

PR #2096, opened 2017-11-23
Display profile image to each page

kyoshidajp:
To increase visibility, I think it would be better to display image next to user name.

# Screenshot

 Queries list, for example.

## before
before_670320

## after
after_670320

# List of all changed page

- queries list
- queries search result list
- query detail
- query snippets list
- query snippet edit

# Other

Add `image_url` field to `User` model. If user logged in via Google Authentication, then picture url got as a callback will save as image_url and return as `gravatar_url`. Otherwise, Gravatar url will return as `gravatar_url`.

BTW a classname `profile__image_thumb` may not be appropriate.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -395,7 +395,7 @@ class User(TimestampMixin, db.Model, BelongsToOrgMixin, UserMixin, PermissionsCh
     org = db.relationship(Organization, backref=db.backref("users", lazy="dynamic"))
     name = Column(db.String(320))
     email = Column(LowercasedString)
-    image_url = Column(db.String(320), nullable=True)
+    _profile_image_url = Column('profile_image_url', db.String(320), nullable=True)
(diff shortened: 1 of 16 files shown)


kyoshidajp:
Sorry, I am going to fix the test errors later.

kocsmy:
This is a cool idea, thanks @kyoshidajp

redash/models.py
```diff
@@ -436,6 +437,9 @@ def is_api_user(self):
 
     @property
     def gravatar_url(self):
```
arikfr: Let's rename this to `profile_image_url` (here and everywhere else we use it) to be consistent.

kyoshidajp:
@arikfr Thanks. I already changed. However, codeclimate doesn't finish. I'm not sure the reason.

arikfr:
And it's in. Thanks!

PR #2097, opened 2017-11-23
Handling whitespace characters in Query Results data source

ariarijp:
I would like to fix a problem with whitespace characters on Query Results data source.


arikfr:
Thanks!

Issue #2095, opened 2017-11-22
Sankey/sunburst rendering issue on a shared dashboard


arikfr:
They seem to render only after the first resize event. You can try this URL:

http://localhost:8080/public/dashboards/AZvtvhObBmraBGcZw77WXWAHie7uflFBYSUAl1PR?org_slug=default

PR #2098, opened 2017-11-23
 Fixed sunburst/sankey issue with rendering

kravets-levko:
Issue getredash/redash#2095

1. Fixed Sankey and Sunburst issue with rendering (did not render on page load).
2. Removed unused (due to removed options) fields from viz editors.
3. Small UI improvements to Cohort viz.

### DIFF
--- a/client/app/assets/less/inc/visualizations/sankey.less
+++ b/client/app/assets/less/inc/visualizations/sankey.less
@@ -12,3 +12,8 @@
   stroke: #000;
   stroke-opacity: .2;
 }
+
+.sankey-visualization-container {
+  height: 300px;
+  overflow: hidden;
+}
--- a/client/app/assets/less/inc/visualizations/sankey.less
+++ b/client/app/assets/less/inc/visualizations/sankey.less
@@ -14,6 +14,6 @@
 }
 
 .sankey-visualization-container {
-  height: 300px;
+  height: 500px;
   overflow: hidden;
 }
--- a/client/app/components/app-header/app-header.html
+++ b/client/app/components/app-header/app-header.html
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-default navbar-fixed-top app-header" role="navigation">
+<nav class="navbar navbar-default app-header" role="navigation">
   <div class="container">
     <div class="navbar-header">
       <button type="button" class="navbar-toggle" ng-click="isNavOpen = !isNavOpen">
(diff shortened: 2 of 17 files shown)


client/app/assets/less/inc/visualizations/sankey.less
```diff
@@ -12,3 +12,8 @@
   stroke: #000;
   stroke-opacity: .2;
 }
+
+.sankey-visualization-container {
+  height: 300px;
```
arikfr: Does this make Sankey fixed height?

client/app/components/app-header/app-header.html
```diff
@@ -1,5 +1,5 @@
-<nav class="navbar navbar-default app-header" role="navigation">
-  <div class="container"">
+<nav class="navbar navbar-default navbar-fixed-top app-header" role="navigation">
```
arikfr: Actually we changed it to be not-fixed on purpose. 

I think it's still fixed on shared dashboards -- that's a mistake.

client/app/visualizations/sunburst/index.js
```diff
@@ -40,7 +43,7 @@ export default function init(ngModule) {
 
     const editTemplate = '<sunburst-sequence-editor></sunburst-sequence-editor>';
     const defaultOptions = {
-      height: 300,
```
arikfr: If we remove this, we need to remove the option from the viz editor. I'm actually fine with it, if it now occupies the size it is given.

(this is the same for Sankey)

client/app/assets/less/inc/visualizations/sankey.less
kravets-levko: Here (and in other components) - it's the default height (when dashboard is in mobile mode or when auto-height enabled) - just moved from component's config.

client/app/components/app-header/app-header.html
kravets-levko: My bad - I had to ask which is right. Will fix now

client/app/visualizations/sunburst/index.js
kravets-levko: Need to check (probably for other components too). I think, in a case when user can set height of viz directly on dashboard - we don't need such option in editors, it's redundant.

client/app/assets/less/inc/visualizations/sankey.less
arikfr: 👌 this is the same for embeds, right?

client/app/visualizations/sunburst/index.js
arikfr: I agree. While I can think of uses for this, it's not worth the extra complexity.

client/app/assets/less/inc/visualizations/sankey.less
kravets-levko: Yes, sure

kravets-levko: And for preview in editors too

PR #2104, opened 2017-11-27
Replace dynamic-table renderer with ag-grid

kravets-levko:
Issue getredash/redash/issues/1356 (Step 1)

Some screenshots:

Dashboard: 
image

Query editor: 
image

Embedded: 
image


arikfr:
Can you please add a few screenshots? 🙏

arikfr:
Some notes:

* Vertical scrollbar on the results page:

image

* When applying sorting to the columns, there is a horizontal scrollbar being added:

image

* The user can resize/drag columns, but it doesn't get saved. Once step #2 is done, we can save this configuration, but until then - is it hard to disable? 

* Can we apply the same look & feel as our previous table had?

* Looks like there is an error in the pagination counter and the last page gets duplicated?

kravets-levko:
- scrollbars: as I mentioned earlier, library manages all dimensions by itself, so I cannot control when (and why) scrollbars become visible.
- I decided that it may be useful even if size/position is not saved - user may decrease column width to make more columns visible, or, say, put some columns near each other to have a better view, etc. But yes - it can be easily disabled.
- the same look & feel - yes, but it will take some time (cannot estimate right now).
- pagination - will take a look.

arikfr:
I played around with it a bit, and the problem with the way it's setup today is that there will be large white space in case of small number of rows, because the table has fixed height.

arikfr:
Oops. Accidentally clicked on "Close" 😳

arikfr:
Following a discussion with Levko we decided that at the moment agGrid introduces more issues than the value it brings.

PR #2110, opened 2017-11-30
Table v2

kravets-levko:
Issue getredash/redash/issues/1356

Improvements to table visualization:
- multi-column sort;
- custom cell rendering + JSON cell renderer as example (see screenshot);
- search through data;
- editor:
  - allow to choose columns to display;
  - allow to re-order column;
  - allow to edit column title;
  - allow to allow/disallow HTML contents;
  - allow to change content alignment in column;
  - allow to specify format for date/time and numeric columns, and false/true labels for boolean columns.

Screenshots:

image

image

image

image


arikfr:
Some comments on the visualization editor:

* It should use tabs like the chart visualization. For now it can be two groups: Grid and Columns.
* The columns tab should be formatted like a table. 
* The alignment option should be 3 grouped buttons that use the relevant icons (`fa-align-right`, `fa-align-left`, `fa-align-center`).

I'm not 100% about showing the column options as a table -- as it might be too crowded, as we add options, but I don't like the list option either. wdyt? @kocsmy thoughts?

package.json
```diff
@@ -37,13 +37,15 @@
...
     "angular-vs-repeat": "^1.1.7",
+    "angularjs-dragula": "^2.0.0",
     "bootstrap": "^3.3.7",
     "brace": "^0.10.0",
     "core-js": "https://registry.npmjs.org/core-js/-/core-js-2.4.1.tgz",
     "cornelius": "git+https://github.com/restorando/cornelius.git",
     "d3": "^3.5.17",
     "d3-cloud": "^1.2.4",
     "debug": "^3.1.0",
+    "dragula": "^3.7.2",
```
arikfr: Did you see we already have [sortable](https://github.com/getredash/redash/blob/master/client/app/lib/sortable.js)? I don't mind switching, but prefer to keep only one lib for the same functionality.

kravets-levko:
@arikfr, good point. Technically it does not matter if columns will be aligned as list or as table, just keep in mind that there will be more fields for each column related to formatting. ATM I plan to add dropdown with possible format types ('number', 'date', 'string' etc.), and each type will have additional options (say, for number it will be decimal separator, thousands separator, decimal digits, etc., for date/time - date/time format string, ans so on). cc: @kocsmy

package.json
kravets-levko: Missed it. Will remove dragula and use it.

arikfr:
> @arikfr, good point. Technically it does not matter if columns will be aligned as list or as table, just keep in mind that there will be more fields for each column related to formatting. ATM I plan to add dropdown with possible format types ('number', 'date', 'string' etc.), and each type will have additional options

Yeah, that's why I'm not 100% sure about it. Maybe the format column will span on a few lines when needed? I have a feeling it will be easier to scan and see as a table rather than a list.

> (say, for number it will be decimal separator, thousands separator, decimal digits, etc., for date/time - date/time format string, ans so on).

Maybe to simplify number formatting we will take a single string that describes the formatting?
There is http://numbrojs.com/getting-started.html that handles this, but not sure if we need another dependency just for this.

kravets-levko:
I was thinking about it, and didn't decide yet. But I think that single formatting string should be better option from UX (and development :-) ) point

arikfr:
👍 

We can also copy this implementation to the counter viz too.

kravets-levko:
Yes, but let do that a bit later :smile:

arikfr:
Looks great and exciting to see it finally happen! 

Some comments:

* The order of the visualizations seem to change when adding new ones or updating them. It looks like the first one becomes the last edited. We should retain the concept of a default table. Maybe a simple solution will be to sort the visualizations by id, so the default table will always be the first (as it has the lowest id)?
* When I opened the editor the first time for an existing table, the "items per page" dropdown had no value. We need to make sure it reflects the default.
* Multiple column sorting, I found it to be confusing a bit. Some ideas on how to make it less confusing: 
    * Enable it only when Shift+Clicking on a column (while using regular sort otherwise).
    * Show the order with a small indicating number like agGrid does.
* The columns tab should be the default/first rather than the Grid tab.
* Columns editor:
  * We capitalize the column header by default. The columns options should reflect that (show the title the same way). Maybe we should stop doing this?
  * Maybe we should have a static label with the column name? So regardless of the changes you make to the title you always know what is it about?
  * There needs to be a visual hint that you can reorder the columns (probably just a different pointer).
  * The "Format specs" link of the number format can be hidden in a popover on a `?` icon next to the "Number format" title.
  * Maybe we should use a vertical order for the column "boxes", so it's easier to scroll with a wheel mouse? :)

kravets-levko:
@arikfr,

- there are strange things with viz sorting in addition to what you said: new viz is added to the end of list, but after page reload, server returns them in reversed order. I think we should sort them explicitly to have any stable order of visualizations (I think, sort by id will be a good choice).
- items per page dropdown - need to take a look, it should pick the default value.
- let's show small numbers - it's really good idea to dive an indication of columns order in sort.
- tabs - NP, will swap them.
- columns editor:
  - I think we should stop using auto-capitalize (anyway usually it's not working well...). If user does not like default title - he can always click "Edit" button and type whatever he wants.
  - static label - do you mean something like a description? can you explain this idea a bit more?
  - format specs - NP, will do.
  - as we're talking earlier - both layouts are possible :-) if you want to try how it feels with vertical layout - just set `flex-direction: column` for div with column editors, - this will allow you to compare and choose the best option. As for me - horizontal layout looks great, except of scrolling (but I'm a cheater - I have mouse with two wheels, so I can scroll in both directions :smile: )

arikfr:
>  I think we should stop using auto-capitalize (anyway usually it's not working well...)

👍 I agree.

> static label - do you mean something like a description? can you explain this idea a bit more?

I meant a label that indicates what column you're currently editing, because once you changed the title you have no idea what it is. 

> As for me - horizontal layout looks great, except of scrolling (but I'm a cheater - I have mouse with two wheels, so I can scroll in both directions 😄 )

I agree that it looks good (and represents the look of the table), it just that more often you will need to scroll to see the next column. Yet many use a trackpad these days, so...

@kocsmy any opinion on this?

kravets-levko:
> I meant a label that indicates what column you're currently editing, because once you changed the title you have no idea what it is.

In this case, we can show original column names (but I'm not sure for UI now).

> As for me - horizontal layout looks great, except of scrolling

I'll push a final set of changes today, and then we'll think on it. Maybe, @kocsmy will have some ideas, - anyway it's better to play with "live" app, then talking (IMHO)

arikfr:
👌 

I still need to finish reviewing the code itself. Started yesterday, but didn't have the chance to finish.

kocsmy:
Regarding the layout: it made more sense for me when we placed the column editor above the table, instead of left: column editor, right: table. This way, you see a lot of data examples which isn't really important: you're just checking how the data in the table will look like, maybe even 1 row should be enough.

In that case, the horizontal scrolling was also more intuitive as it was sitting right on top of the table and you very quickly made a visual connection.

arikfr:
@kravets-levko what about (multi column sorting) "Enable it only when Shift+Clicking on a column (while using regular sort otherwise)."?

kravets-levko:
@arikfr I forgot about it. Added now.

client/app/visualizations/table/formats.js
```diff
@@ -0,0 +1,60 @@
...
+      // Only `true`
+      return value => (value ? values[0] : '');
+    }
+  }
+  return value => (value ? 'true' : 'false');
+}
+
+function createNumberFormatter(format) {
+  if (_.isString(format) && (format !== '')) {
+    const n = numeral(0); // cache `numeral` instance
```
arikfr: How is it being cached?

arikfr:
Some new comments:
* Try creating a new query and running it -> no table is shown.
* The tooltip for format specs should show up without a click.
* Search seems to work strangely on dates.
* When adding a new column to the table, it gets added at the end -- this is unexpected (specially if I didn't change default order).

client/app/visualizations/table/formats.js
kravets-levko: in the next line, the function is returned, which later will be called for each row in a table. so placing it here instead of inside that function is a kind of cache.

client/app/components/dynamic-table/default-cell/index.js
```diff
@@ -0,0 +1,16 @@
+import template from './template.html';
+
+export default function init(ngModule) {
+  ngModule.directive('dynamicTableDefaultCell', $sanitize => ({
```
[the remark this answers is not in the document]
kravets-levko: yes. in `<dynamic-table>` controller there is a function that compiles template for rendering single row; if we don't create this directive - we'll need to put its template in that function. so this directive makes code clean (also, please read the comment to that function too). and another reason to have it - it's making no difference between "custom" and generic cell renderer, - it's like defining "entity" - "cell renderer", and then making two implementations of it.

client/app/visualizations/table/formats.js
arikfr: 👍 now I see it.

client/app/components/dynamic-table/default-cell/index.js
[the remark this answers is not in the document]
arikfr: I meant, why a directive and not a component?

kravets-levko:
> Try creating a new query and running it -> no table is shown.
Fixed, will ship with next set of changes.

> The tooltip for format specs should show up without a click.
There is an issue: if show it on mouseover and hide on mouseout - user cannot click the link (popover gets hidden before user can reach it); when show in on mouseover and hide on outside click - it's a bit strange and may confuse user.

> Search seems to work strangely on dates.
Probably it does searching by "raw" data, will fix.

> When adding a new column to the table, it gets added at the end -- this is unexpected (specially if I didn't change default order).
When user saves viz - an unique index is assigned to all columns; that index is used to sort column later. If we'll assign that index only after re-ordering columns - the behavior will not be clear to user, because from his point new columns will be sometimes added at proper place (when not re-ordering), and sometimes - at the end (when columns were re-ordered earlier). So I decided to make the behavior to be the same for any case.

arikfr:
> > Search seems to work strangely on dates.

> Probably it's because it does searching by "raw" data, will fix.

That's what I thought, but even when searching for something that was supposed to be found in the raw data it wasn't found. 

We should think of adding Jest to the project, and start adding unit tests for things like this function. Can be easy to test and bring a lot of value.

> > When adding a new column to the table, it gets added at the end -- this is unexpected (specially if I didn't change default order).

> When user saves viz - an unique index is assigned to all columns; that index is used to sort column later. If we'll assign that index only after re-ordering columns - the behavior will not be clear to user, because from his point new columns will be sometimes added at proper place (when not re-ordering), and sometimes - at the end (when columns were re-ordered earlier). So I decided to make the behavior to be the same for any case. Hope I explained my motives good enough

I can see two possible solutions:
1. Always reset the order when seeing new columns.
2. Assign unique index only when the user reorders the columns.

Both might surprise users, but I think that 2 will result in less frustration.

client/app/components/dynamic-table/dynamic-table.css
```diff
@@ -1,3 +0,0 @@
-th.sortable-column {
```
arikfr: We're using this in other places too -- like `alerts-list.html`.

kravets-levko: 1. It's not removed, just renamed to `.less`.
2. I was thinking that this file contains component-specific CSS; it there are some styles shared with other components or some common CSS - it should be moved to the component or `assets/less` respectively. Will do it a bit later.

kravets-levko:
I fixed search by date (and in general not by date) - by applying formatting function to data before search; the idea is like WYSIWYG - what user does see in the table - that is used for search.

For column ordering - I think all three (including current one) solutions are not great, but probably you're right - 2 is less evil. Will change the implementation to it.

client/app/components/dynamic-table/default-cell/index.js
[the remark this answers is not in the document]
kravets-levko: See my comment a line below - AngularJS does not allow wrong tags in the table (it cannot be configured anyhow), therefore this directive should be used as attribute; but component can be only element (`restrict` is ignored - always `E` used). That's why :-)

client/app/components/dynamic-table/dynamic-table.css
arikfr: 1. But in the `less` file it's scoped to the dynamic-table container, isn't it?
2. True, my bad probably.

client/app/components/dynamic-table/default-cell/index.js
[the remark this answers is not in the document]
arikfr: Bummer :(

client/app/components/dynamic-table/dynamic-table.css
kravets-levko: Moved that styles to `assets/less/inc/table.less` - hope it's better place for it.

arikfr:
Looks good. One last thing: on a new (unsaved) query, when showing the table it's missing the "Table" title.

kravets-levko:
@arikfr Ready

arikfr:
👏 👏 🏆 :-)

PR #2111, opened 2017-12-01
Upgrade psycopg2 for support PostgreSQL 10.0

kyoshidajp:
Run `pip install` after install PostgreSQL 10.0, the following error has occurred.

```
% pip install -r requirements.txt -r requirements_dev.txt
Collecting psycopg2==2.5.2 (from -r requirements.txt (line 21))
  Using cached psycopg2-2.5.2.tar.gz
    Complete output from command python setup.py egg_info:
    running egg_info
    creating pip-egg-info/psycopg2.egg-info
    writing pip-egg-info/psycopg2.egg-info/PKG-INFO
    writing top-level names to pip-egg-info/psycopg2.egg-info/top_level.txt
    writing dependency_links to pip-egg-info/psycopg2.egg-info/dependency_links.txt
    writing manifest file 'pip-egg-info/psycopg2.egg-info/SOURCES.txt'
    warning: manifest_maker: standard file '-c' not found

    Error: could not determine PostgreSQL version from '10.1'

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/_4/_g1sfpnx2tzf_nck0dqzc7xh0000gp/T/pip-build-wEEbyQ/psycopg2/
```

According to psycopg/psycopg2#489, psycopg2 already fixed this issue.

How about upgrading psycopg2 version. Or with Docker image.

PR #2113, opened 2017-12-02
Design/refinements 2

kocsmy:
There are a couple of design improvements in this PR:

**Tables:**

- headers are now Title Case instead of UPPERCASE
- numerous padding and margin improvements
- adjusted table header colors

**Query screen:**

- Rearranged query editor control buttons: Execute query moved to the right, Fork moved under ...
- Show/Hide Source button now is better connected to its area
- Query editor auto-suggest dropdown now highlights nicer (than the previous bulky dropshadow)
- Numerous spacing fixes so it feels more consistent across the product

**STYLING-README:**

- Added a basic styling readme for contributors so we can try to be a bit more consistent and write better quality code (and avoid inline css) - will add more details to this

**General styling:**

- Bug fix: fullscreen not hiding navigation now hides navigation again
- Now, we're using system-fonts as default and Roboto has been removed: font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif !important;
- Pagination is now rectangular instead of circular to better align with our styling direction
- Public Dashboards are now navigation free for better data consuming experience. Powered by Redash also added to the footer.
- Edit Layout now disables Query Headers to link to its query page and replaced with cursor: move to indicate that you can move the widgets in edit mode.
- Line height fixes so longer dashboard and query names will look better when they break to multiline.


kocsmy:
Query Before:
query-old

Query After:
query-new

arikfr:
1. Looks a bit strange that the `...` button is in the middle, no?
2. The "Data Source" label got truncated to "Data", not sure if it's intentional. Maybe the icon is enough?
3. "headers are now Title Case instead of UPPERCASE" -> we can skip this (apply on transformation).
4. How does the results page look like?

kocsmy:
1. I don't think it's strange (especially because the Execute button is more prominent and btn-primary).
2. Yes, I made it shorter but we can even use the icon only.
3. Ok.
4. Which one if the result page?

kyoshidajp:
@kocsmy An error `Error: could not determine PostgreSQL version from '10.0'` in CircleCI has already fixed on master branch by #2111. If you rebase master, then the error would not be reproduced :sunny:

kocsmy:
Further improvements added, see screenshots:
screenshot 2017-12-05 13 05 48

screenshot 2017-12-05 13 05 25

PR #2120, opened 2017-12-05
prevent remove self

negibouze:
If an administrator mistakenly removed him/herself from a built-in group, it can not recover.
In order to prevent this, the administrator made it impossible to remove him/herself from the built-in group.

・current built-in group view
current view

・built-in group view after change
after_built-in_group

・regular group view after change
after_regular_group


arikfr:
Super :) Thanks!

PR #2126, opened 2017-12-05
Layout fixes (mainly schema browser)

arikfr:
* Maintain consistent size of query editor, regardless of availability of the schema.
* Change order of Create menu (Query is first now).
* "View Query" menu item in widget's menu is now linking to the query source view.
* Fix some accidental issues in the UI.

The 3 states of schema browser:
3tgjdpdsqw

* Showing
* Hidden (when not supported)
* Showing but empty (when failed to load)

### DIFF
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -250,10 +250,6 @@
                 <div ng-if="!query.visualizations.length">
                   <filters filters="filters"></filters>
                   <grid-renderer query-result="queryResult" items-per-page="50"></grid-renderer>
-                  <!-- the ng-repeat is a lame hack to find the table visualization... -->
-                  <div class="p-15" ng-repeat="vis in query.visualizations" ng-if="vis.type == 'TABLE'">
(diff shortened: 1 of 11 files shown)


client/app/pages/queries/query.html
```diff
@@ -238,19 +237,17 @@ <h3>
...
                 </ul>
               </div>
             </div>
             <div class="row">
               <div class="col-lg-12">
-                <div ng-show="selectedTab == 'table'">
+                <div ng-if="!query.visualizations.length">
                   <filters filters="filters"></filters>
                   <grid-renderer query-result="queryResult" items-per-page="50"></grid-renderer>
                   <!-- the ng-repeat is a lame hack to find the table visualization... -->
```
kravets-levko: As I understand, this is a fix for broken rebase of previous PR. If I'm right - there is some code missing below (there shouldn't be a block with `ng-if="vis.type == 'TABLE'"` and probably some other changes).

arikfr: 👍 checking.

PR #2133, opened 2017-12-06
feat(QueryRunner): add exasol

jomar83:


m1racoli:
Since the codeclimate tests failed independent of our changes, will this be merged anytime soon?

jezdez:
FWIW, the codeclimate tests seem to pass (now?).

As to whether this lands, I'll let @arikfr make that call, but I'm inclined to encourage you to release this as a separate package to PyPI instead and use the `REDASH_ADDITIONAL_QUERY_RUNNERS` environment variable to extend the list of query runners.

Especially shipping binary libraries like the .so file that is part of this pull request is a maintenance burden for an Open Source project such as Redash.

As for your separate package, maybe using [python-exasol](https://github.com/EXASOL/python-exasol) would suffice for your needs and simplify the package?

arikfr:
> Especially shipping binary libraries like the .so file that is part of this pull request is a maintenance burden for an Open Source project such as Redash.

Maybe we can download this during the Dockerfile build process instead of including in our repo?

> As for your separate package, maybe using python-exasol would suffice for your needs and simplify the package?

It requires the same .so file :)

jezdez:
> Maybe we can download this during the Dockerfile build process instead of including in our repo?

Sure I guess, but again, this would duplicate efforts maintaining code to interact with a database. Next to the official python-exasol there is also a sqlalchemy_exasol package (https://github.com/blue-yonder/sqlalchemy_exasol) maybe that could be a dependency for us instead? Or of a 3rd party query runner package?

> It requires the same .so file :)

Right, but it would allow maintaining the project separately instead of increasing the maintenance burden on the main Redash project.

arikfr:
Closing in favor of #4378, which does not require an ODBC driver.

PR #2140, opened 2017-12-07
Fix LDAP href URL issue & Support missing LDAP mail entry

LiviuJianu:
**URL issue:**  
When activating the LDAP / SSO authentication, the URL the user clicks
on returns a 404 not found.

In the login.html, the href value is: /ldap_auth/login which is not the
same one used for the blueprint route in ldap_auth.py: /ldap/login

**Missing LDAP mail entry**
When the LDAP user has no email entered in the mail attribute, login is
not possible due to the fact that the current implementation uses the
email field.

Therefore, the email needs to be composed from other fields.
For this, we need to get the username of the user and the company
domain, concatenate the values and use the resulting email.

e.g.
To get the username, we make use of  the already existing key, but ask
it to query the ldap server for the “uid” attribute:
`LDAP_EMAIL_KEY = “uid” —> johndoe`

To get the company domain we add an optional environment key read in
the settings file.
The enviromnent key would hold the domain name:
`export REDASH_LDAP_CUSTOM_MAIL_DOMAIN=“redash.com"`

The end result is: `johndoe@redash.com`

When the key is not defined, the flow is not changed.


LiviuJianu:
Creating a branch to have only those commits in the pull requests.

PR #2141, opened 2017-12-07
Fix ldap url mail

LiviuJianu:
When the LDAP user has no email entered in the mail attribute, login is
not possible due to the fact that the current implementation uses the
email field.

Therefore, the email needs to be composed from other fields.
For this, we need to get the username of the user and the company
domain, concatenate the values and use the resulting email.

e.g.
To get the username, we make use of  the already existing key, but ask
it to query the ldap server for the “uid” attribute:
LDAP_EMAIL_KEY = “uid” —> johndoe

To get the company domain we add an optional environment key read in
the settings file.
The enviromnent key would hold the domain name:
export REDASH_LDAP_CUSTOM_MAIL_DOMAIN=“redash.com"

The end result is: johndoe@redash.com

When the key is not defined, the flow is not changed.

### DIFF
--- a/redash/authentication/ldap_auth.py
+++ b/redash/authentication/ldap_auth.py
@@ -36,7 +36,12 @@ def login(org_slug=None):
         user = auth_ldap_user(request.form['email'], request.form['password'])
 
         if user is not None:
-            create_and_login_user(current_org, user[settings.LDAP_DISPLAY_NAME_KEY][0], user[settings.LDAP_EMAIL_KEY][0])
+            email = ''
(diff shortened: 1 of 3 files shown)


gramakri:
Would be good to have atleast https://github.com/getredash/redash/pull/2141/commits/4c99e4377b66ee181b2e8453c1fb25cab847a49e merged since the LDAP login page currently shows a 404

gramakri:
The same fix is made in PR 2170 - https://github.com/getredash/redash/pull/2170/commits/840f04850516f66481b8d6615825cb907f942c8b

PR #2143, opened 2017-12-08
Fix an exception when displaying Presto error messages.

visualage:
When the db.message is a string or a non-dictionary, db.message.get throws an exception. In my specific case, db.message is a unicode string.
This patch checks if the db.message.get attribute exists before invoking.


redash/query_runner/presto.py
```diff
@@ -116,7 +116,10 @@ def run_query(self, query, user):
         except DatabaseError as db:
             json_data = None
             default_message = 'Unspecified DatabaseError: {0}'.format(db.message)
-            message = db.message.get('failureInfo', {'message', None}).get('message')
+            if isattr(db.message, 'get'):
```
arikfr: Are you sure `isattr` is a Python function?

As an alternative, you can just check if it's a `dict` (`if isinstance(db.message, dict)`).

visualage: Yes. My mistake. Should be hasattr function. I must accidentally made the typo when changed the if from the one-line form to the multi-line form (due to the line-too-long stylecheck error).

PR #2148, opened 2017-12-10
Replace invalid date time strings

kyoshidajp:
When `dateTime` filter got invalid date value, it returned `Invalid date`.

Screenshot when a query has not executed is the following, for example.

before_datetime

I believe the date value is not valid in this situation. This PR make `dateTime` filter to return `-`. 

after_datetime

Invalid date is too.

### DIFF
--- a/client/app/filters/datetime.js
+++ b/client/app/filters/datetime.js
@@ -5,7 +5,7 @@ export default function init(ngModule) {
 
   ngModule.filter('dateTime', clientConfig =>
     function dateTime(value) {
-      if (!value || !value.isValid()) {
+      if (!value || !moment(value).isValid()) {
         return '-';
       }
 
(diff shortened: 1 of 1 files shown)


client/app/filters/datetime.js
```diff
@@ -5,7 +5,7 @@ export default function init(ngModule) {
 
   ngModule.filter('dateTime', clientConfig =>
     function dateTime(value) {
-      if (!value) {
+      if (!value || !value.isValid()) {
```
arikfr: I noticed that we call `moment` on `value` below, which implies that we might send a string to this function. Maybe check `isValid` after the conversion to a `moment` object?

kyoshidajp:
@arikfr Thank you for your comment. I changed calling `isValid` method after conversion to a `moment` object.

arikfr:
Thanks!

PR #2152, opened 2017-12-11
Fix: Appearance collapse of outdated-queries.

bamboo-yujiro:
There seems to be no `<th>` element for Id of "out dated queries", so I fixed it.

before:

bb

after:

aft


arikfr:
Thanks!

PR #2155, opened 2017-12-12
[WIP] Minimum change on filter default

tonyjiangh:
(Changed on Dec. 17)
Originally aimed to add default value for filter, but decided to leave out default for this PR.

**Task lists**
- [x] Set default value for multi-filter to all value
  - that's more natural I suppose?
- [x] Use one single `*` tag instead of showing all tags when [SELECT ALL].
  - Click on the `*` tag to transform back to all tags
- [x] Only list not selected choices
-  ~~Save default value to options~~
    - No idea how to do that now. New new AngularJS actually.

Note's below.

### DIFF
--- a/client/app/components/filters.html
+++ b/client/app/components/filters.html
@@ -16,8 +16,8 @@
 
       <ui-select ng-model="filter.current" multiple ng-if="filter.multiple" on-select="$ctrl.filterChangeListener(filter, $model)"
         on-remove="$ctrl.filterChangeListener(filter, $model)" remove-selected="false">
-        <ui-select-match placeholder="Select value for {{filter.friendlyName}}...">{{$item | filterValue:filter}}</ui-select-match>
-        <ui-select-choices repeat="value in filter.values | filter: $select.search" group-by="$ctrl.itemGroup">
+        <ui-select-match placeholder="Select value for {{filter.friendlyName}}..." ng-click="$ctrl.filterExpandAllValue(filter)">{{$item | filterValue:filter}}</ui-select-match>
+        <ui-select-choices ui-disable-choice="filter.current[0] === '*'" repeat="value in filter.values | filter: $select.search" group-by="$ctrl.itemGroup">
           <span ng-if="value == '*'">
             Select All
           </span>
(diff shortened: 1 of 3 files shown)


tonyjiangh:
**Notes**
- I guess we need an alternative to current SELECT ALL option for multi-filter. ↓ This could be a problem when there's more values.
image

- I guess other plans for default value would eventually come to implementing filter objects and persist them in db, which is kind of a big job. I guess a minimum fix before that would be nice?

@arikfr 
I would like to consult on this before writing more code. 
Any thought on the task list? Do they seem right to you?

arikfr:
Moving forward I want to avoid adding more annotations to the query in order to drive the UI. 

As for default value: we can do what we do with parameters and save the default value in Query's `options` object. With parameters the current value of the parameter when you save the query gets saved as the default, I think doing the same for filters is a good start.

As for the UX issues of having lots of values in a multi filter, I agree that it's needs to be addressed. But I think it's in the domain of the UI component and we don't need to change the data for this. @kocsmy @kravets-levko do you have any ideas about this?

tonyjiangh:
### Default value
Ok, no more annotations then. I'll check how parameter works.

### Lots of tag issue
This is what I'm thinking about so far.
How Select ALL will be:
image

Click on the tag, if we want the multiple labels back, (in order to remove some from all selected state maybe).

Further enhancement maybe setting the maximum size of filter and allow scrolling?

tonyjiangh:
Updates: 
- Decided to leave out the adding filter default value part.
   - Is it better to migrate to another PR for [SELECT ALL] change, instead of editing this one?
- Checked the part of code of saving parameters to options, no idea how to do that for filters.
   - Operations on filters are currently in query_result, I'm not sure how to save that into query's options. (Completely new to AngularJS here.)
   - So I think it's better to not touch it myself.

kocsmy:
Personally, I like the * to select all idea. It's kinda technical but I can imagine working well here.

tonyjiangh:
I've checked the code around but couldn't came up with a solution for the default value. I'll close with PR and maybe open another for the SELECT ALL tag mentioned above only.

The problem I met was that
- filters belongs to query_result but `options` object belongs to query. But query_result object doesn't seem to have a foreign key pointing to query directly.
- both query_editor and dashboard visulization uses filters so it seems better to implement default value handling in service. But they're acquiring filter information from different api in a different way, so it seems that there's a lot to handle.

arikfr:
> filters belongs to query_result but options object belongs to query. But query_result object doesn't seem to have a foreign key pointing to query directly.

We can pass the options as another property to the `<filters>` component.

> But they're acquiring filter information from different api in a different way

This is indeed trickier... we can do what parameters do at the moment: use the first (or is it last?) query's default. But we definitely need to handle it better.

Issue #1902, opened 2017-08-02
Wrong column display when stacking with negative values


grzlybear:
### Issue Summary
I'm trying to create a column chart with series values that contain both positive and negative values.
For example, the number of new users is shown as positive while the number of churned users is shown as negative. They both have the same X value (same time frame).

When selecting the stacking option I expect the positive values to be above the X line and the negative values to be below the X line.

### Technical details:

* Redash Version: 1.0.3
* Browser/OS: Chrome / OSX
* How did you install Redash: self hosted on google cloud

reality
![screen shot 2017-08-02 at 14 43 09](https://user-images.githubusercontent.com/4466323/28872287-36a95c18-7791-11e7-8c32-7f2e64e060f2.png)

expected
![screen shot 2017-08-02 at 14 45 42](https://user-images.githubusercontent.com/4466323/28872313-4bfc7c3a-7791-11e7-95bc-af0fac63c168.png)

deecay:
@arikfr 
Introduce barmode ['stack', 'overlay', 'group', 'relative'] somewhere in the editor-ui?
Since the list of options is not too long, maybe radio control is better that dropdown, which would require more clicks. Any preference or thoughts?

Example:
[https://codepen.io/etpinard/pen/mPGLrY](https://codepen.io/etpinard/pen/mPGLrY)

arikfr:
We already have some of those options implemented with custom code, so it's nice we might be able to drop it and use their implementation. 

But looks like they have a bug with "stack" too (in your example) or did I misunderstand how it supposed to work?

deecay:
Related Plotly community discussion [here](https://community.plot.ly/t/stacked-bar-with-positive-and-negative-values/920/2).
So the behavior of "stack" seems intentional?

Issue #2117, opened 2017-12-04
Revisit Plotly integration


arikfr:
Our Plotly integration contains a lot of custom code, mainly around stacking. It created some issues over the years (#1902, #1832, #1326, #2116). We also miss out on new features like annotations (#1254).

Let's revisit the integration code, clean it up and mainly:
* remove our custom stacking code (if possible).
* add support for annotations.

Once this is done, we can start adding new features (like #728 for example).

deecay:
I would like to add
* Supporting graph height/width and margins.

as one very basic yet important point. Related controls and codes are scattered in couple of places and this is a good opportunity to review what should be done for graph heights and margins.

arikfr:
Height can be changed by adjusting the widget height now. And why would someone need to change the margins?

deecay:
Is height adjustable in query page?
Lets say an author created bar chart and wants it to be 1000px high. Does he have to put it in dashboard to see if the query and chart config are just right?

I will post screenshots later for margins.

kravets-levko:
Hi @deecay! Your idea definitely makes sense, but, IMHO, we should think a step forward:

1. can this use case be applied to another visualizations? I think - yes.
2. should that width/height option be applied  in other places - not only on query page, but in editor's preview, shared dashboard, shared viz, etc.? Definitely yes, for a common behavior.

We already talked with @arikfr about making embedded visualizations to fit size of `<iframe>` (in the same manner as they fit widgets in dashboard). So, probably we should work out this idea and apply it to query page somehow. WDYT?

arikfr:
To expand on @kravets-levko's comment: I think that we need to make visualization take the space they are given. We already have this behavior on dashboard widgets, we just need to expand it to embeds and query page too.

Then to accommodate your use case we can make the visualization section on a query page resizable. 

The only question remains is whether the size of the visualization section should be persisted with the visualization/query. For the sake of simplicity, I would go with no persistence, and revisit this decision after we have this "in the wild".

WDYT?

deecay:
Hi,
Now I understand that there is a plan that "fit height to container" is going to be applied to places other than dashboards. That is awesome.

Then I just wanted to point out query page is important to consider, and I agree with @kravets-levko that there are even more places to consider, such as editor's preview. 

[...]

Issue #2116, opened 2017-12-04
Percentage stacking doesn't work for line charts


arikfr:

arikfr:
Resolved in #2156 (part of the v4 release).

kravets-levko:
@innovia Please add screenshots of settings for this chart. Thanks!

kravets-levko:
@innovia it's strange. anyway - thanks for this report, I'll try to reproduce it. if there is a bug - we definitely should fix it :wink:

Issue #1832, opened 2017-06-18
Auto-Maximum does not work in stacking graph


vladimirbaranov:
### Issue Summary

Auto-Maximum does not work in stacking graph. (Chrome)

### Steps to Reproduce

1. Select stacking chart
2. Choose Y-axis minimum
3. Leave Y-axis maximum at Auto
4. The top of the chart gets cut off and auto-maximum does not get set.

![image](https://user-images.githubusercontent.com/6749278/27264417-d1fa6e3c-544b-11e7-87eb-6d8040cc0024.png)

### Technical details:

* Redash Version: 1.0.3+b2850
* Browser/OS: Chrome
* How did you install Redash: Command Line

alexmuller:
I can reproduce this in Redash 1.0.0+b2815 with Firefox 55.

arikfr:
Resolved in #2156 (part of the v4 release).

PR #2156, opened 2017-12-12
Revisit Plotly integration

kravets-levko:
Issue getredash/redash#2117

- Improved code quality of chart component.
- Implemented auto-size for axis (and removed unnecessary `X-Axis Height` option from editor).
- Add new stacking options (fixes getredash/redash#1902).
- Fix y-axis min/max options (getredash/redash#1832).
- Fix chart labels + better formatting (getredash/redash#1326).
- Fix bug with staking negative values (bar, line and area charts). Fix percent value stacking (bar, line and area charts). Implement stacking for line charts (getredash/redash#2116).
- Option to always show labels on chart (bar, line and area charts - getredash/redash/issues/1254).


client/app/config/index.js
```diff
@@ -57,7 +57,7 @@ function registerAll(context) {
     .map(context)
     .map(module => module.default);
 
-  return modules.map(f => f(ngModule));
+  return modules.filter(isFunction).map(f => f(ngModule));
```
arikfr: I'm assuming this is because of the utils.js file... let's just move `app/visualizations/chart/plotly/` content to `app/lib/visualizations`.

client/app/visualizations/chart/chart-editor.html
```diff
@@ -13,7 +13,7 @@
       <a ng-click="changeTab('series')">Series</a>
     </li>
   </ul>
-  <div ng-show="currentTab == 'general'">
+  <div ng-if="currentTab == 'general'" class="m-t-10 m-b-10">
```
arikfr: Any special reason to switch to `ng-if`? I think that because those are tabs, it will be simpler/quicker to have `ng-show`, no?

client/app/visualizations/chart/index.js
```diff
@@ -53,9 +56,26 @@ function ChartEditor(ColorPalette, clientConfig) {
...
-        Percent: 'percent',
+        Stack: 'stack',
+        Overlay: 'overlay',
+        Group: 'group',
+        Relative: 'relative',
       };
+      scope.allowedStackingOptions = [];
+      scope.isStackingOptionAllowed = value => scope.allowedStackingOptions.indexOf(value.value) >= 0;
+
+      scope.$watch('options.globalSeriesType', () => {
```
arikfr: Can we use `ng-change` on the input field instead of `$watch`?

client/app/visualizations/chart/chart-editor.html
kravets-levko: It's a king of holywar in AngularJS world - `ng-show` vs. `ng-if`. Generic recommendation is to use `ng-show` for more or less static fragments that should frequently be shown/hidden. `ng-if` works better if sub-view has a lot of bindings - because when `ng-if` condition becomes false, `ng-if` removed subtree from DOM and detaches all event listeners, watchers, destroys directives, etc. This also means that `ng-if` is bad if its condition changes too frequently. But in case of tabs, IMHO, `ng-if` is better solution - because it matches all mentioned pros - condition changes rarely, subtree contains a log of angular bingins.

client/app/config/index.js
kravets-levko: I prefer to leave it - not especially because of `utils.js`, but rather for avoiding a sort of strange bugs - when just adding an empty file, app starts crashing, and it's a surprise for ones who doesn't know about this _autoloader_

client/app/visualizations/chart/index.js
kravets-levko: Okay, will do

client/app/config/index.js
kravets-levko: P.S. This also affects not just visualizations, but components, directives, services...

arikfr: Good point, let's keep it. 👍

client/app/visualizations/chart/chart-editor.html
arikfr: I guess it depends on what's defined as frequent... but ok, let's use ng-if.

kravets-levko: I think, *frequent* should be measured with comparing time of rendering subtree with interval of changing condition. In case with tabs, subtree will render, say, in 50-100ms, but user switches tabs once in, say, 5 or more seconds (I think this assumptions are more or less correct). So, comparing that values, we can see that subtree renders fast enough, but it is shown or hidden very rarely - because user will spend tens of seconds and more while editing options on tab.

arikfr:
image

It's hard to read the values. Can we place them on top of the bars?

kravets-levko:
Need to try, but not sure.

arikfr:
I'm not sure if it's related to this pull request but on a query page, when you open the API key dialog the chart resizes and then resizes back to previous size.

kravets-levko:
It's because bootstrap hides scrollbar on `body`; chart is not resized to previous state (there will be a difference in few pixels). BTW the same behavior may be noticed on dashboards too - when opening any modal dialog, grid will also try to fit new size and all widgets will resize a bit.

arikfr:
> It's because bootstrap hides scrollbar on body; chart is not resized to previous state (there will be a difference in few pixels). BTW the same behavior may be noticed on dashboards too - when opening any modal dialog, grid will also try to fit new size and all widgets will resize a bit.

What can we do about this?

arikfr:
Btw, it also flickers on first load.

client/app/visualizations/chart/chart-editor.html
arikfr: 👍

client/app/visualizations/chart/index.js
```diff
@@ -53,9 +56,13 @@ function ChartEditor(ColorPalette, clientConfig) {
 
       scope.stackingOptions = {
         Disabled: null,
-        Enabled: 'normal',
-        Percent: 'percent',
+        Stack: 'stack',
+        Overlay: 'overlay',
+        Group: 'group',
+        Relative: 'relative',
```
arikfr: I couldn't see the difference between `Group` and `Disabled` or `Relative` and `Stack`. At least not with bar charts (maybe it works different with other chart types?).

Unless it's very important let's keep those settings (and any other) as they were to ease the rollout of this code, as otherwise we will need to start applying data migrations which is better to avoid in this case.

kravets-levko: Do your data have negative values? With negative values all combinations have clear difference

arikfr: OK, now I can see the difference between Relative and Stack. But still not between group and disabled.

Why not always use relative mode when stacking and then we don't need a different option?

kravets-levko: All that modes really depends on data:

image

image

About relative mode: we cannot always use it, because it displays data in different way: instead of absolute values, it normalizes values against sum, so range of y-axis will be 0..100 (%). And relative mode is not a stacking option itself, it's an option to prepare data before displaying.

arikfr: If it's just the way the data prepared, just the more reason not to have it as an option in the UI :)

The end result should be: the user has three options: grouped, stacked or percent stacked. They shouldn't need to care about how we achieve this, we should just show it correctly (within reason) :)

In the first screenshot above it looks super weird because of two Y axes, right? I'm not sure this is a huge use case to be concerned about. But don't think we need to stack across Y axes (if it's something we can control).

client/app/visualizations/chart/chart-editor.html
```diff
@@ -32,6 +32,9 @@
       </div>
     </div>
 
+    <div ng-if="['line', 'area', 'column'].indexOf(options.globalSeriesType) >= 0" class="checkbox">
```
arikfr: Why scatter plot doesn't support annotations?

client/app/visualizations/chart/plotly/utils.js
```diff
@@ -0,0 +1,515 @@
...
+  });
+
+  return seriesList;
+}
+
+function enableAnnotations(layout, seriesList, options) {
+  if (options.annotations) {
+    if (options.globalSeriesType === 'column') {
+      seriesList.forEach((series) => {
+        series.textposition = 'inside';
```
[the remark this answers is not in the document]
kravets-levko: > the user has three options: grouped, stacked or percent stacked. 

Then for bar chart we will have 8 options + "Disabled" instead of 4 + "Disabled" - because then there will be "Stack", "Percent stack", "Overlay", "Percent overlay"...

Maybe, we need to choose better label for that checkbox ("Percent stacking" or so), but IMHO it's not good idea to transform it into more options.

client/app/visualizations/chart/chart-editor.html
kravets-levko: You're right, it supports. Looks a bit strange, but will add - maybe, someone will need it

client/app/visualizations/chart/plotly/utils.js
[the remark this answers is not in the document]
arikfr: I'm not sure I understand why the user needs to see more than only 3 options (disabled/grouped, stacked, percent)?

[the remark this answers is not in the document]
kravets-levko: Hmm. Earlier we had only tree options: "Disabled", "Stacked", "Percent stacked" - which was exactly the same as "Stacked" but with additional data processing. Now for bar charts we have more stacking variants - "Stack", "Group", "Overlay", "Relative", and each of them can have the same "percent" modifier. I decided to make it a checkbox and do not flood the dropdown with large set of values.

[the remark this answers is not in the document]
kravets-levko: About "Percent" modifier - it allows to compare relative values, in a way like pie chart. An example:

Suppose we have three series with a single value each: 100, 200, 700. For regular stacking, chart will render bar of height 100, then bar of height 200 above the first one, and the last bar - of height 700. Max value of Y-axis will be `100+200+700 = 1000`.

In relative mode, it will normalize values against their sum: `100 / 1000 * 100% = 10%`, `200 / 1000 * 100% = 20%`, `700 / 1000 * 100% = 70%`, and then will render them in the same way as described above, but max value of Y-axis will be `100%`.

For a single value probably it does not make a lot of sense, but when all series contain more that one value - there is a clear difference between regular and "percent" node - in the first mode, every bar will have different height, in the second one - every bar will have `100%`-height with values distributed in that bound - very similar to pie chart, just with bars.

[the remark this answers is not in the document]
arikfr: What you described is percent stacking :-) And I still don't see a difference between Stack/Relative when Percent is enabled.

I think I will just review the code itself and try to understand.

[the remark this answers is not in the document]
kravets-levko: Haha :-) There is difference too, but it will be visible if there are more than one bar, and there are bars with all positive and bars some negative values. Then in "Stacking" mode all bars will be placed above x-axis (probably negative values will not be handled properly), and in "Relative" mode bars with negative values will have absolute height "100%", but will be displayed with some offset below x-axis (depending on sum of negative values).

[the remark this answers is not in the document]
kravets-levko: I was also wondering what is the difference between all that modes, but while digging deeper into code and docs, I realized that they work differently with edge cases - like mixed positive and negative values, etc. And, as I mentioned, "Percent" modifier - it's just a modifier, it translates values, but not changes rendering algorithm.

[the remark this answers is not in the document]
arikfr: ## Stacking

I found the documentation for the different `barmode` values [in Plotly's source](https://github.com/plotly/plotly.js/blob/5bc25b490702e5ed61265207833dbd58e8ab27f1/src/traces/bar/layout_attributes.js#L14-L29). 

Disabled is the same as `group`, because `group` is the default. `relative` seems to be the same as `stack`, but supports negative values. I tried to find a case when `stack` might be better than `relative` but couldn't think of one.

I also found [`barnorm` config value](https://github.com/plotly/plotly.js/blob/5bc25b490702e5ed61265207833dbd58e8ab27f1/src/traces/bar/layout_attributes.js#L31-L43), which allows to enable "built in" percent based rendering. Because we need to have our own calculation for area stacking, I guess we should just stick to it for bars as well.

To summarize the discussion we had so far and what I think we should do:

The user should have only 2 stacking options:

* No Stacking (we will use `group` for bars, which we can either pass explicitly now or keep as no value)
* Stacked (which for bars will use `barmode`=`relative`)

And then a checkbox to trigger "Normalize values to percentage", as you added.

These options should be only available when the user selects "Bar" or "Area" as the chart type (right? I don't see a reason to show them in other modes).

### Percent based interaction

When disabling one of the series, the 100% value should be updated - it doesn't happen at the moment.

Here's an example of what I mean:

Both series showing:



Only one enabled:

Only One Enabled

### Migration

Moving to 2 values for stacking + toggle for percentage will require a migration. I think it's simple enough to handle it in the chart code instead of having to do a DB migration... I think I even noticed you already implemented something for it?

## Data Labels

1. Let's rename the toggle to "Show Data Labels" instead of "Annotations" and update the option name in code accordingly. Annotations have a different meaning.
2. Scatter plots (which is what we use for both lines, area and scatter) have built in support for showing data labels. You need to pass `text` to the `mode` value (and if it has a value than you use the `+` modifier, i.e. `lines+text`).

kravets-levko:
@arikfr One more set of changes (based on our discussion):

- reduced count of stacking options;
- annotations for scatter plots replaced with data labels;
- toggling legend item will re-calculate percent values;
- checkboxes for new options now have better titles.

"Migration" on client-side was already implemented.

kravets-levko:
It was mostly support for "annotations" - and we decided to remove it.

arikfr:
Few issues I found:

* When you double click on a series in the legend it supposed to disable all the other series. But in bar chart it disables everything, it does seem to work for other chart types (like area).
* Area chart get messed up when stacking is enabled (see http://localhost:8080/queries/4/source?p_min_count=10#22 on the dev server).
* Why percentage normalization option only visible when stacking enabled?
* When hovering over points, we show 0 values for series that don't have data at this point. It's not the way it used to work previously and not the default Plotly behavior. I assume this is because at some stage of preparing the data, you make sure that we have a Y value for every X value -- we don't need this (except for some cases - this is what the unifiedXAxis was for, iirc).

kravets-levko:
First two items - need to take a look.

> Why percentage normalization option only visible when stacking enabled?

Because earlier it was a stacking option. But you're right, now it can be standalone option.

> When hovering over points, we show 0 values for series that don't have data at this point.

I kept all the code related to `unifiedXAxis` option. Probably this behavior is related to stacking, need to take a look.

kravets-levko:
@arikfr Updated PR - fixed all your notices + fixed chart flickering on resize.

arikfr:
Feels like we're getting there, but still a few things:

* There is still flickering when opening a chart visualization.
* http://localhost:8080/queries/13/source#21 (dev server) - when you disable stacking the chart "breaks". But if you create a new chart for the same query, stacking/no stacking works. Any idea if this is an actual issue or something broken specifically with this chart?

kravets-levko:
> There is still flickering when opening a chart visualization.

This is one-time chart update - it's related to auto-computing chart margins. The cart shouldn't flicker after that.

> when you disable stacking the chart "breaks". 

Probably it's related to that "front-end migration" - will check ASAP. Thanks for pointing.

kravets-levko:
@arikfr Can you please explain (or show somehow) what do you mean saying that chart breaks? Maybe, in your example, chart looks a bit strange because there is right y-axis? See how it looks with and without right axis:

image

image

arikfr:
Yep, the double Y axis explains it. I wonder if there is a way for them to have the 0 value at the same place?

client/app/visualizations/chart/plotly/index.js
```diff
@@ -0,0 +1,144 @@
...
+    const plotlyElement = element[0].querySelector('.plotly-chart-container');
+    const plotlyOptions = { showLink: false, displaylogo: false };
+    let layout = {};
+    let data = [];
+
+    const updateChartDimensions = debounce(() => {
+      if (updateDimensions(layout, plotlyElement, calculateMargins(plotlyElement))) {
+        Plotly.relayout(plotlyElement, layout);
+      }
+    }, 100);
```
arikfr: When I pass the optional `immediate` parameter to `debounce` here it stops the flickering (as it doesn't wait the first 100ms). Is there a reason not to do that?

kravets-levko: Thanks for pointing, I'll focus on this and check it

kravets-levko:
About double Y axis - I think they have different position of `0` because they have different min/max values - plotly just interpolates them independently so they have different start position. But user can manually specify min/max value if he is not happy with the result.

arikfr:
double Y - 👍

kravets-levko:
@arikfr, I updated my comment - added some things. Didn't expect that you'll read it so fast :-)

arikfr:
> Also, we have some code that calculates min/max for y axis I can check when is it used, and probably, we can calculate unified margins for both axis, I think it makes some sense.

The two axes might not have the same values (for example one in the [-100, 100] range and the other in the [-100,000, 50,000] range). Also sometimes one of them might not have negative values at all. If there is an easy way to make sure that regardless they are symetrical, then let's give it a shot. Otherwise we can leave as is (unless this happens every time the user uses two Y axes).

kravets-levko:
@arikfr Fixed flickering issue, thank you for help. About double Y-axis - I played with charts a bit, and I think it's better to leave it as it is, because (as you mentioned earlier) there is no universal solution, there always will be some edge cases. Currently chart viz has enough options to deal with most cases. In example, the issue we talked earlier could be fixed by setting Y-axis ranges manually:

image

arikfr:
💃

denisov-vlad:
Looks like after this PR values on chart is missed.

kravets-levko:
Hi @denisov-vlad ! Can you also provide all options for your chart, and data types of columns used there? Also, I see that your chart has multiple Y-values for each X-value within each series (I see light horizontal lines on bars) - can you try to aggregate the data properly and check if something changes? Thanks!

denisov-vlad:
Hi @kravets-levko. Here is chart's options:



column types:
 {u'friendly_name': u'month', u'type': u'date', u'name': u'month'}, 
{u'friendly_name': u'country', u'type': u'string', u'name': u'country'}, 
{u'friendly_name': u'mpb', u'type': u'float', u'name': u'mpb'}, 
{u'friendly_name': u'adx', u'type': u'float', u'name': u'adx'}, 
{u'friendly_name': u'fb', u'type': u'float', u'name': u'fb'}, 
{u'friendly_name': u'fia', u'type': u'float', u'name': u'fia'}, 
{u'friendly_name': u'total', u'type': u'float', u'name': u'total'},
 {u'friendly_name': u'month_pivot', u'type': u'string', u'name': u'month_pivot'},
 {'type': 'string', 'friendly_name': 'country_name', 'name': 'country_name'}

I've tried to reformat query to have 1 value column but chart is still the same:

kravets-levko:
Thanks @denisov-vlad! The problem is that you have multiple rows with the same `month` value - chart cannot choose which y-value to show. You should use `GROUP BY` in your query by `month` column and `SUM` values in your `fb`, `fia`, `mpb` and `adx` solumns.

denisov-vlad:
@kravets-levko yes, that's it! :)

Is it possible to add checkbox to sum it on front-end? We use one query for different visualisations: pivot table and bar-chart for example. Some queries are really huge and runs about 10 minutes. That's why creating duplicates with different `GROUP BY` isn't a good solution

 Please look at this issue: https://github.com/getredash/redash/issues/1742.

arikfr:
ahhh... this issue keeps poping up :-)

If the Query Results data source was supporting loading cached results, you could solve it by running another query in Redash that doesn't require running the 10 minutes query again. 🤔

denisov-vlad:
@arikfr haha. :)

Query results is a good idea but we are using redash more than 1 year and have about 2000 queries and 15 datasources. Using query results will give x3 to queries count. %)

PR #2165, opened 2017-12-16
Fix unicode handling at query result

tonyjiangh:
About this https://github.com/getredash/redash/pull/2145

In the end, it was my complete misunderstanding of unicode handling and 1 single character fix.
Sorry about all the attention wasted...


arikfr:
:)

Thanks!

Issue #2129, opened 2017-12-06
Counter is too small on the query page


arikfr:
![image](https://user-images.githubusercontent.com/71468/33656583-23ebc786-da7f-11e7-9319-6bf64105b0b8.png)

PR #2166, opened 2017-12-18
Counter is too small on query page

kravets-levko:
Issue getredash/redash#2129

### DIFF
--- a/client/app/visualizations/counter/index.js
+++ b/client/app/visualizations/counter/index.js
@@ -16,7 +16,7 @@ function getRowNumber(index, size) {
   return size + index;
 }
 
-function CounterRenderer() {
+function CounterRenderer($timeout) {
   return {
     restrict: 'E',
     template: counterTemplate,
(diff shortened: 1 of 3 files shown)


client/app/visualizations/counter/index.js
```diff
@@ -97,6 +97,10 @@ function CounterRenderer() {
             $scope.stringSuffix = null;
           }
         }
+
+        $timeout(() => {
+          $scope.handleResize();
```
arikfr: Will there be "flickering"?

kravets-levko: Maybe, just a little bit. On my core i7 I didn't see any flickering. But I cannot change/remove this code - I need to wait for angular to render second line with number, otherwise it computes wrong font size.

arikfr: If it's has `display:none` will we be able to measure it?

kravets-levko: No.

PR #2171, opened 2017-12-20
Added API version to salesforce query runner 

mayconbordin:
Added API version to the Salesforce query runner, it is necessary for some queries that use fields added in newer versions of the Salesforce API.
Also fixed an issue when a row column value does not exist on the dictionary

### DIFF
--- a/redash/query_runner/salesforce.py
+++ b/redash/query_runner/salesforce.py
@@ -10,7 +10,7 @@ logger = logging.getLogger(__name__)
 
 try:
     from simple_salesforce import Salesforce as SimpleSalesforce
-    from simple_salesforce.api import SalesforceError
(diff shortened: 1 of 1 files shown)


redash/query_runner/salesforce.py
```diff
@@ -76,6 +76,10 @@ def configuration_schema(cls):
                 },
                 "sandbox": {
                     "type": "boolean"
+                },
+                "api_version": {
+                    "type": "string",
+                    "title": "Salesforce API Version"
```
arikfr: Can you please add a "default" property with the value of `DEFAULT_API_VERSION`?

Thanks.

arikfr:
Thanks!

PR #2172, opened 2017-12-20
Add page-specific class to body for each app page

kravets-levko:
Requested by @kocsmy

### DIFF
--- a/client/app/config/index.js
+++ b/client/app/config/index.js
@@ -111,4 +111,13 @@ registerComponents();
 registerPages();
 registerVisualizations(ngModule);
 
+ngModule.run(($rootScope) => {
+  $rootScope.$on('$routeChangeStart', () => {
+    $rootScope.additionalBodyClass = '';
+  });
+  $rootScope.$on('$routeChangeSuccess', ($event, $route) => {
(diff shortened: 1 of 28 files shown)


arikfr:
@kocsmy can you explain the usage for page specific classes?

kocsmy:
@arikfr sure: for example the Query page might need a body { overflow: hidden; } and the best way to do this is with a unique body class (or ID would work, too). 

It's also handy for other pages, in case we need to apply page specific styling.

arikfr:
@kocsmy does the `<body>` needs this style or some other top level component might be just as good?

kocsmy:
@arikfr the body needs this style afaik (I tried lower level elements)

arikfr:
Having a page specific class will lead us to having unique snowflakes instead of consistent pages. What we do need is support for different layouts (a closed set of different layouts).

Thank to the work @kravets-levko did with creating the top level `<app-view>` component, we can use it to create support for layouts.

As we don't really need this now, let's close this and revisit when we do need a different layout for some of our pages.

kocsmy:
But sometimes, we do need beautiful unique snowflakes so we can make snowballs that we can throw far far away.

PR #2175, opened 2017-12-21
Fix : User Form displays a incorrect error appearance.

bamboo-yujiro:
Regardless that user saving or updating is success, each form field of user has `error frame`.

**Image (gif) :**

before1 mov

The code which has the problem is in `client/app/filters/index.js`.

In `showError function`, if the `form` is submitted, it assume that there is an error.

I fixed it.

And when calling of `showErorr function`,  it's argument is no longer useless, so I removed it.

**After:**

**gif**
after1 mov

**image (when any error occur)**
aft1

### DIFF
--- a/client/app/filters/index.js
+++ b/client/app/filters/index.js
@@ -92,7 +92,7 @@ export function notEmpty(collection) {
   return !isEmpty(collection);
 }
 
-export function showError(field, form) {
-  return (field.$touched && field.$invalid) || form.$submitted;
+export function showError(field) {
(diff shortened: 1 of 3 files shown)


kocsmy:
thanks for fixing this, it was bothering me for a while :)

arikfr:
Thanks!

I think that originally it was supposed to be `&& form.$submitted`, but I don't remember why we had this. And looks like without it, it still works as you would expect, so 🤷‍♂️

PR #2180, opened 2017-12-23
Feature: Add an unarchive dashboard button

kyoshidajp:
Fix #982.

It would be easy to restore without directly updating DB :relaxed:

# Screenshot

Replaced "archived" label to "Unarchive" button.
unarchive_button

Confirm modal:
unarchive_dialog

PR #2181, opened 2017-12-24
Feature: Add an unarchive query button

kyoshidajp:
Similar to #2180.

This is a query version.

# Screenshot

Confirm dialog:
unarchive_query_dialog

## Unarchive button

I can't make up my mind which one to choose where the button should be placed.

### Plan A(current version)

Same as #2180. Next to the query name.
unarchive_query_button

### Plan B

By the control button area.
unarchive_query_button2

### Plan C

Inside the botton group.
unarchive_query_botton3

### DIFF
--- a/client/app/pages/queries/view.js
+++ b/client/app/pages/queries/view.js
@@ -288,6 +288,7 @@ function QueryViewCtrl(
       },
       () => {
         $scope.query.is_archived = false;
+        toastr.success('Query unarchived');
       },
       () => {
         toastr.error('Query could not be unarchived.');
--- a/client/app/pages/queries/view.js
+++ b/client/app/pages/queries/view.js
@@ -282,21 +282,12 @@ function QueryViewCtrl(
   $scope.unarchiveQuery = () => {
     const queryId = $scope.query.id;
     const url = `api/queries/${queryId}/unarchive`;
-    const unarchive = () => {
-      Events.record('unarhive', 'queries', queryId);
-      $http.patch(url).success(() => {
-        $scope.query.is_archived = false;
-      }).error(() => {
-        toastr.error('Query could not be unarchived.');
-      });
-    };
-
-    const title = 'Unarchive Query';
-    const message =
-      'Are you sure you want to unarchive this query?';
-    const confirm = { class: 'btn-warning', title: 'Unarchive' };
-
-    AlertDialog.open(title, message, confirm).then(unarchive);
+    Events.record('unarhive', 'queries', queryId);
+    $http.patch(url).success(() => {
+      $scope.query.is_archived = false;
+    }).error(() => {
+      toastr.error('Query could not be unarchived.');
+    });
   };
 
   $scope.updateDataSource = () => {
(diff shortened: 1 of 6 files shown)


kocsmy:
Thanks for this @kyoshidajp —in order to keep this consistent with Unarchive Dashboard #2180 —I'd say Plan A makes the most sense.

kocsmy:
Maybe in the future it'd make sense to have the label [Archived] with a dropdown to unarchive (so it's slightly easier to understand). Currently we don't really have a label+dropdown component, so the current solution will work for know.

kyoshidajp:
@kocsmy Thanks for your comment.

@arikfr  Two issues are reported by codeclimate.  I think it's not a problem. WDYT?

arikfr:
Actually I think we should go with Plan C for both queries and dashboards.

Re. CodeClimate: you can safely ignore those.

kocsmy:
Oh, good point, we have ... menu there, too and that's where you archive already. Makes more sense.

kyoshidajp:
OK! I'm going to change both queries and dashboards code tonight.

kyoshidajp:
I've already changed.

However, I can't ignore CodeClimete error. "Disable check button" isn't shown.

codeclimate

client/app/pages/queries/view.js
```diff
@@ -278,6 +279,26 @@ function QueryViewCtrl(
...
+        toastr.error('Query could not be unarchived.');
+      });
+    };
+
+    const title = 'Unarchive Query';
+    const message =
+      'Are you sure you want to unarchive this query?';
+    const confirm = { class: 'btn-warning', title: 'Unarchive' };
+
+    AlertDialog.open(title, message, confirm).then(unarchive);
```
arikfr: I'm not sure confirmation is needed here... it's not that destructive and easy to revert.

client/app/pages/queries/view.js
```diff
@@ -278,6 +279,26 @@ function QueryViewCtrl(
     AlertDialog.open(title, message, confirm).then(archive);
   };
 
+  $scope.unarchiveQuery = () => {
+    const queryId = $scope.query.id;
+    const url = `api/queries/${queryId}/unarchive`;
+    const unarchive = () => {
+      Events.record('unarhive', 'queries', queryId);
+      $http.patch(url).success(() => {
```
arikfr: Why not use the `Query` resource?

redash/handlers/queries.py
```diff
@@ -239,6 +239,17 @@ def delete(self, query_id):
         query.archive(self.current_user)
         models.db.session.commit()
 
+    def patch(self, query_id):
```
arikfr: Feels non idiomatic.. :\ Maybe a regular update (POST) request with `is_archived` set to `false`?

client/app/pages/queries/view.js
kyoshidajp: I agree with you. I'll fix it.

kyoshidajp:
I changed to Plan C and resolved a conflict.

unarchive_query_menu_button

Also, add a message.

unarchived_message

@arikfr Could you review this?

gclsoft:
Why I still not find the unarchive query button button? @kyoshidajp 
image

gclsoft:
@kyoshidajp I see, thank you.

arikfr:
Hi,

Thanks.

PR #2184, opened 2017-12-27
rename REDASH_SEARCH_DN to REDASH_LDAP_SEARCH_DN

gramakri:
This makes the naming consistent.

### DIFF
--- a/redash/settings.py
+++ b/redash/settings.py
@@ -150,7 +150,7 @@ LDAP_CUSTOM_USERNAME_PROMPT = os.environ.get('REDASH_LDAP_CUSTOM_USERNAME_PROMPT
 # LDAP Search DN TEMPLATE (for AD this should be "(sAMAccountName=%(username)s)"")
 LDAP_SEARCH_TEMPLATE = os.environ.get('REDASH_LDAP_SEARCH_TEMPLATE', '(cn=%(username)s)')
 # The schema to bind to (ex. cn=users,dc=ORG,dc=local)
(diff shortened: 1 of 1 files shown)


gramakri:
I will submit a PR to https://github.com/getredash/website/blob/master/onpremise/setup/ldap-authentication.md once this is merged.

redash/settings.py
```diff
@@ -150,7 +150,7 @@ def all_settings():
 # LDAP Search DN TEMPLATE (for AD this should be "(sAMAccountName=%(username)s)"")
 LDAP_SEARCH_TEMPLATE = os.environ.get('REDASH_LDAP_SEARCH_TEMPLATE', '(cn=%(username)s)')
 # The schema to bind to (ex. cn=users,dc=ORG,dc=local)
-LDAP_SEARCH_DN = os.environ.get('REDASH_SEARCH_DN', None)
+LDAP_SEARCH_DN = os.environ.get('REDASH_LDAP_SEARCH_DN', None)
```
arikfr: We need to support both, just in case someone already using REDASH_SEARCH_DN, so something like this:
`os.environ.get('REDASH_LDAP_SEARCH_DN', os.environ.get('REDASH_SEARCH_DN'))`.

gramakri: Thanks for the review, I will update the PR.

gramakri:
https://github.com/getredash/website/pull/61 is the doc change

arikfr:
Thanks!

Issue #2162, opened 2017-12-13
User should see an error message when trying to open a query he has no permission for


arikfr:
Currently the user gets a blank page. Should see a proper error page.

PR #2187, opened 2017-12-27
Improve error handling mechanism

kravets-levko:
Related to issue related to getredash/redash#2162

Added root component for application, that manages visibility of header, footer and page area and shows error message depending on current route status and auth status. This allows to remove some ugly hacks related to router.


client/app/services/auth.js
```diff
@@ -24,6 +24,7 @@ function getLocalSessionData() {
 
 function AuthService($window, $location, $q, $http) {
   const Auth = {
+    logger,
```
arikfr: I'm not sure it's worth passing the logger around. It can be as easily created where needed, no?

client/app/components/app-view/index.js
```diff
@@ -0,0 +1,64 @@
...
+            this.showHeaderAndFooter = true;
+          } else {
+            event.preventDefault();
+            Auth.loadSession().then(() => {
+              if (Auth.isAuthenticated()) {
+                this.showHeaderAndFooter = true;
+                Auth.logger('Loaded session');
+                $route.reload();
+              } else {
+                throw new Error('Need to login');
```
arikfr: I'm assuming this will be caught by `catch` below, why not call `Auth.login` here?

client/app/components/app-view/index.js
```diff
@@ -0,0 +1,64 @@
...
+      $rootScope.$on('appViewError', (event, error) => {
+        if ((error !== null) && (error !== undefined) && (error !== '')) {
+          this.error = error instanceof Error ? error : new Error('' + error);
+        } else {
+          this.error = null;
+        }
+      });
+
+      $rootScope.$on('$routeChangeSuccess', () => {
+        $rootScope.$broadcast('appViewError', null);
```
arikfr: Why use an event here (and below) instead of updating the state of error value directly?

client/app/pages/dashboards/dashboard.js
```diff
@@ -164,12 +164,16 @@ function DashboardCtrl(
...
-      this.loadDashboard();
+    }, (error) => {
+      const statusGroup = Math.floor(error.status / 100);
+      if (statusGroup === 5) {
+        // recoverable errors - all 5** (server is temporarily unavailable
+        // for some reason, but it should get up soon).
+        this.loadDashboard();
+      } else {
+        // all kind of 4** errors are not recoverable, so just display them
+        $rootScope.$broadcast('appViewRejection', error);
```
arikfr: I really want to avoid using events... maybe we can use [`$exceptionHandler`](https://docs.angularjs.org/api/ng/service/$exceptionHandler) and just throw an exception here?

Then in the handler we can have a few known types that will have dedicated screens for and "Unknown error occurred".

kravets-levko:
@arikfr Implemented error handling using `$exceptionHandler`. Now it also will catch not only route and model errors, but, probably, it will help to write better code :-)

arikfr:
Because we never took client side error handling seriously enough I wonder if it won't result in users seeing the error screen too often for non important stuff 🤔 

I guess we should give it a try and find out as we try it.

client/app/components/app-view/index.js
```diff
@@ -0,0 +1,68 @@
+import debug from 'debug';
+import template from './template.html';
+
+const logger = debug('redash:app-view');
+
+export default function init(ngModule) {
+  let customHandler = null;
```
arikfr: How about having an object to handle errors:

```javascript

const errorHandler = {
    error: null,
    reset() {
        this.error = null;
    },
    processError(error) {
        this.error = error;
    }
}
```

Then we can use it without waiting for it to be initialized.

i.e.:

```javascript
ngModule.factory('$exceptionHandler', () => errorHandler.processError);
```

kravets-levko:
@arikfr Updated PR. I spent some time on testing this error handling - and I didn't see errors nowhere except cases when network down or 403 error - seems that all errors are handled more or less correctly. IMHO, even if we'll encounter some error that is not "global" - we should fix it (catch and process) instead of ignoring.

arikfr:
💃

PR #2188, opened 2017-12-28
Fix upgrade script for ci channel

rgjodekerken:
bin/upgrade --channel ci broke with the following error
ValueError: tmp/circle-artifacts.69TScQG/4.0.0%2Bb3384 is not valid SemVer string


arikfr:
Thanks!

PR #2196, opened 2018-01-02
Fix to error handling mechanism: fail only on promise rejections

kravets-levko:


client/app/components/app-view/error-handler.js
```diff
@@ -10,15 +10,17 @@ export class ErrorHandler {
   }
 
   process(error) {
+    this.reset();
     if (!(error instanceof Error)) {
       if (error.status && error.data) {
```
arikfr: I'm pretty sure that any $http error will have these properties... maybe we should throw an explicit error object that triggers the display?

kravets-levko: Aah...Yes, we can, definitely. It will be almost the same as it was in the first commit :-) But the first solution has not good architecture. Let me think a bit about it.

arikfr: We can utilize `$routeChangeError`, and then there either throw a new object that wraps the $http exception or just set the value of `error`.

The benefit of using the throw option, is that we can use it later in other places where we want the error to block the whole page.

kravets-levko: I already implemented it as custom error (we cannot use `$routeChangeError` because then we'll return back to a moment before I started with all that error handling - `$routeChangeError` sometimes will not work). Will do some tests and update this PR

kravets-levko:
@arikfr Updated PR. Now it works in next way: it will show all errors occurred while route is changing (between `$routeChangeStart` and `$routeChangeSuccess`/`$routeChangeError`), any error in grace period after `$routeChangeSuccess` (it allows to catch errors during controller instantiations, so user will see error instead of page with angular markup), and `PromiseRejectionError` - at any time (currently it is explicitly triggered in dashboard page controller in `loadDashboard` function). Other errors will be just logged in console.

client/app/components/app-view/error-handler.js
arikfr: But `$routeChangeError` wasn't working because we were not rendering the component that supposed to show the error.

But with the change you did it works properly now. So you can just set the error in the `$routeChangeError` event handler, instead of the change in effb212e4b309e5d35accdd6043299b71bcab928.

kravets-levko: Yes, but dashboard page errors will not be caught then - because dashboard page does not use `route.resolve` and therefore does not trigger `$routeChangeError`. So we return back to a need of having a single point of handling errors, and `$routeChangeError` is just one of (many) sources of errors - therefore there if `throw` there.

arikfr: For the dashboard, we can try one of the following: 
1. Switch to resolve.
2. Throw a specific error class we define (`LoadError`?), that for it we will trigger the error screen.

I would go with #2 just to minimize unrelated changes and to have this option in place, so we can utilize later when needed.

The change is that we don't always show the error to the user but rather for specific classes.

kravets-levko: We cannot do 1 because there is auto-reload functionality.

And 2 is already implemented :-)

arikfr: It's not. For 2 there is no need for the wait time, additional babel transformation, etc, etc.

kravets-levko: Please see my comment for #effb212 - I introduced custom error type - `PromiseRejectionError` - which is triggered in `$routeChangeError` or on-demand when needed (currently in dashboard controller). In addition, I decided to show some extra errors (conditions described in my comment) - because if, say, we'll have `TypeError` in controller - then user will see a view with all that `{{ ... }}` or blank page at all - because controller will fail to instantiate. `$routeChangeError` will not be triggered in such cases.

arikfr: I saw the comment. The thing is you added a lot of complexity along with that error type. I don't think the added complexity is worth it. Mixing timing issues and error handling is a recipe for disaster.

We don't need to worry about the user seeing all the `{{...}}`, because we will show him _specific_ errors we _choose_ to show. TypeError isn't one of them. I'm OK with paying the price the user might see broken markup in a rare occasion if in return we get simpler code.

Also once we properly log errors the chances of a user seeing some random error are lower.

kravets-levko: 1. Additional babel transformation is needed because without it babel cannot transpile `extends Error` - it will not have own class (all derived classes will be replaced with Error) and therefore will not pass `instanceof` checks.
2. I added wait time in order to make better user experience - it's not required, but will catch accidental errors in controller constructors (as I mentioned earlier - such errors will prevent views from rendering, so user will not see nothing). The core mechanism is built around custom error class.

kravets-levko: I see your comment with some delay, but got it. Will remove unnecessary code and ping you :+1:

kravets-levko:
@arikfr Updated PR: now showing only dedicated error type (`PromiseRejectionError` - triggered in `$routeChangeError` - covers most pages, and in dashboard controller)

client/app/components/app-view/error-handler.js
arikfr: babel transformation: what do we gain from extending `Error` instead of creating a simple class?

kravets-levko: I think it's just right - all errors should be derived from `Error` class, especially when remember that some tools like browser's console has special handling for `Error` class. I read a docs for `babel-plugin-transform-builtin-extend` - it does not add a lot of overhead - it will process only classes inherited from `Error` (specified in `.babelrc`), and will just add a simple wrapper around it to fix its behavior, that's it.

client/app/components/app-view/error-handler.js
```diff
@@ -1,27 +1,26 @@
+import { EventEmitter } from 'events';
+
 // eslint-disable-next-line import/prefer-default-export
-export class ErrorHandler {
+export class ErrorHandler extends EventEmitter {
```
arikfr: Why it needs to be an EventEmitter? What was wrong with just setting the `error` value when needed to show an error?

kravets-levko: Yep... :+1:

Issue #2183, opened 2017-12-24
UI- filter headers mixed


galgoogle:
When using more than 2 multi-filters, the headers are being shown on the first 2 rows, and not attached to their boxes.

![image](https://user-images.githubusercontent.com/29460916/34327369-25ee68ae-e8cb-11e7-9a07-6673b99977f8.png)

* Redash Version: 3.0.0+b3134
* Browser/OS: Chrome

kocsmy:
It's the [filters.html](https://github.com/getredash/redash/blob/master/client/app/components/filters.html) here:

`<div class="container bg-white p-b-15" ng-show="$ctrl.filters | notEmpty">
  <div class="row">
    <div class="col-sm-6 p-l-0" ng-repeat="filter in $ctrl.filters">
      <label>{{filter.friendlyName}}</label>
    </div>
  </div>
  <div class="row">
    <div class="col-sm-6 p-l-0" ng-repeat="filter in $ctrl.filters">
      <ui-select ng-model="filter.current" ng-if="!filter.multiple" on-select="$ctrl.filterChangeListener(filter, $model)" on-remove="$ctrl.filterChangeListener(filter, $model)"
        remove-selected="false">
        <ui-select-match placeholder="Select value for {{filter.friendlyName}}...">{{$select.selected | filterValue:filter}}</ui-select-match>
        <ui-select-choices repeat="value in filter.values | filter: $select.search">
          {{value | filterValue:filter }}
        </ui-select-choices>
      </ui-select>

      <ui-select ng-model="filter.current" multiple ng-if="filter.multiple" on-select="$ctrl.filterChangeListener(filter, $model)"
        on-remove="$ctrl.filterChangeListener(filter, $model)" remove-selected="false">
        <ui-select-match placeholder="Select value for {{filter.friendlyName}}...">{{$item | filterValue:filter}}</ui-select-match>
        <ui-select-choices repeat="value in filter.values | filter: $select.search" group-by="$ctrl.itemGroup">
          <span ng-if="value == '*'">
            Select All
          </span>
          <span ng-if="value == '-'">
            Clear
          </span>
          <span ng-if="value != '*' && value != '-'">
            {{value | filterValue:filter }}
          </span>
        </ui-select-choices>
      </ui-select>
    </div>
  </div>
</div>
`

galgoogle:
@kocsmy is it an easy fix?

kocsmy:
I'm checking this today and see if I can fix.

[...]

PR #2200, opened 2018-01-03
Fixing filter headers mixed

kocsmy:
Fixing issue #2183 

From now on, we won't mix filter headers. Everything is properly placed.

screenshot 2018-01-03 20 16 32

This PR also fixes misplaced filters on unsaved queries and improves #2199 previous commit.

### DIFF
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -248,7 +248,7 @@
             </div>
             <div class="row">
               <div class="col-lg-12">
-                <div ng-if="!query.visualizations.length">
+                <div ng-if="!query.visualizations.length" class="query__vis col-md-12 p-b-15">
                   <filters filters="filters"></filters>
                   <grid-renderer query-result="queryResult" items-per-page="50"></grid-renderer>
                 </div>
--- a/client/app/assets/less/redash/redash-newstyle.less
+++ b/client/app/assets/less/redash/redash-newstyle.less
@@ -204,6 +204,10 @@ page-header, .page-header--new {
   }
 }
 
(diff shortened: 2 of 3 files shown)


client/app/pages/queries/query.html
```diff
@@ -248,7 +248,7 @@ <h3>
             </div>
             <div class="row">
               <div class="col-lg-12">
-                <div ng-if="!query.visualizations.length">
+                <div ng-if="!query.visualizations.length" class="query__vis col-md-12 p-b-15">
```
arikfr: Did you need to apply anything similar on the dashboard page?

kocsmy: This one fixes the unsaved query alignment. The filter headers are fixed with this change: https://github.com/getredash/redash/pull/2200/commits/4c055c663be6c3b679622fc9b5f3034d1630af97#diff-e8061505c2069e286392a8ad98acdb53

arikfr: 👍

kocsmy: I messed up some stuff here so I'm fixing them here: https://github.com/getredash/redash/pull/2201

galgoogle:
@kocsmy In the current version we don't have client/app/assets/ folder (version 3 beta).
Will it be available only in the new version?

kocsmy:
@galgoogle the most important fix was made on the client/app/components/filters.html file, the fix on the css is just a minor styling improvement. You should be able to fix it on your version based on my commit. (but I think the assets folder was introduced pre-3, so I'm not sure what's the issue)

PR #2204, opened 2018-01-06
Report amount of data scanned for Athena queries

washort:
This supersedes #1895, which I don't have push access to. All review comments should be addressed.


arikfr:
Thanks!

PR #2205, opened 2018-01-06
Make Redshift sslmode configurable

sjakthol:
This makes it possible to configure Redash to require TLS and verify the certificate Redshift presents.

The setting works similarly to that of the PostgreSQL source except that Redshift explicitly passes 'prefer' to psycopg2 if sslmode is not configured (for compatibility).


arikfr:
Thanks!

PR #2208, opened 2018-01-07
feat: make SAML signature requirements configurable

### LINKED ISSUE #1669: SAML: Don't require signed assertions
Currently Redash requires signed SAML assertions.  This is mostly legacy functionality, and isn't needed when the response itself is signed.  Most IdPs will not sign assertions by default (requiring special configuration), but will sign the response, as that's what the vast majority of SPs expect. 

In `saml_auth.py`:

```
                'want_assertions_signed': True,
                'want_response_signed': False,
```

This should probably be reversed:

```
                'want_assertions_signed': False,
                'want_response_signed': True,
```

sjakthol:
These changes make it possible to configure if the SAML based authentication requires the IDP to sign individual assertions, whole response or both. These values can be configured both globally via environmental variables or through organization specific settings from the Admin UI / API.

The two environmental variables provide global defaults for these values:

* REDASH_SAML_WANT_ASSERTIONS_SIGNED
* REDASH_SAML_WANT_RESPONSE_SIGNED

If these variables are not present, they default to the previously hardcoded values (want_signed_assertions = True and want_response_signed = False). This is done to ensure existing installations do not break on upgrade. A test has been written to ensure this behavior persists.

Alternatively, these two variables can be modified on per-organization basis from the Admin UI. These configurations override the global defaults set via the environment.

Special care has been taken to ensure that it's not possible to disable signing requirements completely. If the two environmental variables are both disabled while SAML is enabled, the server will instruct the user to enable at least signing requirement and refuse to start. Similarly, the UI will not let the user to disable both signing requirements and the API will return a 400 response if signing requirements are going to be disabled.

Tests have also been written to test the backend functionality.

Closes #1669.

### DIFF
--- a/client/app/pages/settings/organization.html
+++ b/client/app/pages/settings/organization.html
@@ -45,6 +45,26 @@
                         <input name="input" type="string" class="form-control" ng-model="$ctrl.settings.auth_saml_nameid_format" accesskey="tab"
                                ng-change="$ctrl.update('auth_saml_metadata_nameid_format')">
                     </div>
+
+                    <label>SAML Response Signing Settings</label>
+                    <p ng-class="{'text-danger': !$ctrl.validateSamlSigningSettings(), 'help-block': $ctrl.validateSamlSigningSettings()}">
+                        Depending on your SAML IDP, you must require the IDP to sign
+                        assertions, whole response or both.
+                    </p>
+                    <div class="form-group">
+                        <label>
+                            <input name="input" type="checkbox" ng-model="$ctrl.settings.auth_saml_want_assertions_signed"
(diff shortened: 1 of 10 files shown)


RichardLitt:
Thanks, @sjakthol. Sometimes it takes us a while to respond to PRs. Is there a reason you closed this?

sjakthol:
There were some conflicts and I don't have time to resolve those in the foreseeable future. 

Furthermore, I don't think the approach I had here is good in the long run. Instead of adding individual toggles for every possible option the pysaml2 library has, we should instead let users provide arbitrary configuration blob directly to pysaml2 so that they can configure the SAML setup to match their requirements (if the defaults are not good enough).

arikfr:

PR #2210, opened 2018-01-08
Expose data source version info in UI

washort:
This supersedes #1932 which I don't have push access to.


jezdez:
@emtwo implemented this as a separate extension in https://github.com/mozilla/redash-stmo/pull/10 so we can close this for now.

PR #2211, opened 2018-01-09
Fixes#1625: Add redis connection Pooling

srinivasreddy:


srinivasreddy:
I am not sure of doing lazy execution of  function in a module. Anybody know of a way to do that?  AFAIK, it is not possible. @arikfr

PR #2217, opened 2018-01-09
Fix #2137: Stop lowercasing the query. 

### LINKED ISSUE #2137: Case insensitive parameters in Redash query result cache
### Issue Summary

Redash may display wrong results for query with case sensitive parameter. It looks like results cache is ignoring param letters size.

### Steps to Reproduce

1. Create a query with global text parameter.
2. Create dashboard with visualization for this query.
3. Open dashboard with parameter set to lowercase letter for example 'test'
4. Open dashboard on a new tab with parameter 'TEST'. 

In the last step you should receive visualization for parameter 'test' even when parameter value is set to 'TEST'

I have been able to reproduce this scenario on demo.redash.io. To reproduce do:

1. Open link: http://demo.redash.io/dashboard/ma_dashboard?p_dashboard_name=test
You should see rows from dashboard table with name `test`
2. Open link: http://demo.redash.io/dashboard/ma_dashboard?p_dashboard_name=Test
You should see rows from dashboard table with name `test` even when parameter `dashboard_name` is set to `Test`. After refreshing visualization result changed to proper one.

### Technical details:

* Redash Version: 3.0.0+b3134
* Browser/OS: Chrome/Firefox/Opera
* How did you install Redash: used demo.redash.io

srinivasreddy:
-  And also remove the unused import
-  Write migration script to migrate the existing query hashes

### DIFF
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -14,7 +14,6 @@ import os
 from funcy import distinct, select_values
 from sqlalchemy.orm.query import Query
 
-from .human_time import parse_human_time
 from redash import settings
 
 COMMENTS_REGEX = re.compile("/\*.*?\*/")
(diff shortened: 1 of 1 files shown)


srinivasreddy:
had difficulty in writing migration file. `alembic.ini` does not have `script_location` section. it is always complaining . How  would you create a script, targeting `migrations` folder?

PR #2221, opened 2018-01-11
correctly sort search result columns

washort:
I didn't catch this bug in search-result sorting (added in @jezdez 's fulltext search PR) until today.

### DIFF
--- a/client/app/pages/queries/queries-search-results-page.js
+++ b/client/app/pages/queries/queries-search-results-page.js
@@ -23,6 +23,22 @@ function QuerySearchCtrl($location, $filter, currentUser, Events, Query) {
     this.paginator.updateRows(queries);
   });
 
+  this.createdAtSort = row => row.created_at.valueOf();
+  this.createdBySort = row => row.user.name;
+  this.scheduleSort = function (row) {
+    if (!row.schedule) {
+      return null;
+    }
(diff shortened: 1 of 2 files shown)


jezdez:
I could have sworn the thing worked anyway 😬

washort:
Me too, unfortunately it was not actually sorting by name or date, just rearranging rows....

client/app/pages/queries/queries-search-results-page.js
```diff
@@ -23,6 +23,10 @@ function QuerySearchCtrl($location, $filter, currentUser, Events, Query) {
     this.paginator.updateRows(queries);
   });
 
+  this.createdAtSort = row => row.created_at.valueOf();
+  this.createdBySort = row => row.user.name;
+  this.scheduleSort = row => row.schedule && parseInt(row.schedule, 10);
```
arikfr: I'm not sure if it's a huge deal, but schedule might be an hour as well, in which case it will get random sort for minutes (i.e. 12:30 might come before 12:00).

washort: -- aha, true. We haven't used that feature so I didn't notice. Should be easy to fix

washort:
added fix for time-of-day schedules

arikfr:
Thanks!

Issue #1940, opened 2017-08-30
Use ?fullscreen to hide navbar in shared dashboards


arikfr:
Currently it uses `?embed`, should support `?fullscreen` too to be consistent with regular dashboards.

kravets-levko:
@arikfr I think this could be closed

PR #2226, opened 2018-01-15
 Remove ?embed-related code in public dashboards

kravets-levko:
Issue getredash/redash#1940


kravets-levko:
I spent some time on backend code - seems that there is no code related to `embed` param except of [this line](https://github.com/getredash/redash/blob/master/redash/handlers/embed.py#L98) which just records an event - I think we should keep it because some URLs with `embed` param may already exist.

PR #2231, opened 2018-01-18
100%-wide results table

### LINKED ISSUE #2215: 100% Wide Results Table
The auto width table layout has the benefit of properly spacing the columns, but in the case of small number of rows it just makes the table less visible/prominent.

This:
image

Vs this:
image

It seems that adding an empty column at the end of the table with set `width: 99%` has no negative impact, so let's give it a try.

kravets-levko:
Closes #2215.

Added "spacer" column at the end of table + tweaked data columns width based on data type.

With few columns:
image

With a lot of columns:
image

Scrolling also works:
image

### DIFF
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -93,7 +93,6 @@ edit-in-place p.editable:hover {
 
   table {
     border: 1px solid #f0f0f0;
-    width: auto;
   }
 
   .paginator-container {
--- a/client/app/components/dynamic-table/dynamic-table-row.js
+++ b/client/app/components/dynamic-table/dynamic-table-row.js
@@ -16,7 +16,7 @@ export default function init(ngModule) {
       $scope.$watch('render', () => {
         if (isFunction($scope.render)) {
           $scope.render($scope, (clonedElement) => {
-            $element.empty().append(clonedElement);
+            $element.empty().append(clonedElement).append('<td></td>');
           });
         }
       });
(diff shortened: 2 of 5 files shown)


arikfr:
> tweaked data columns width based on data type.

can you elaborate?

kravets-levko:
As we talked earlier about this improvement, the solution was to add empty column with 99%-width. But in this case columns with string values become too narrow - because text gets wrapped. So after some experiments, I found new solution - do not set `width="99%"` for last column, but set width to `1%` for numeric, boolean and datetime columns and for other columns set width to `auto`, - in this case, free space is distributed more or less properly.

arikfr:
Wouldn't `white-space:nowrap` on the `<td>` element do the same trick?

kravets-levko:
I also tried it. But in general it's better to allow multiline text. Imagine that we have a column with a long text. In a case of `white-space: nowrap` table will become too wide, horizontal scroll will appear, and user will need to scroll down, then scroll right and then scroll up to read all the text. With current solution table will try to fit container (and wrap too long lines), and horizontal scroll will appear only when there are too much columns/too long text, so in general it is easier for user to view data in the table.

kravets-levko:
BTW - About `white-space: nowrap` - I think it makes sense to add it to numeric, boolean and datetime columns: you can see on screenshots above that date/time can be wrapped, as well as numbers (browser can wrap them at decimal or thousands separator).

Also, as an improvement, I can add datatype-specific classes to table cells (i.e. `table-cell-numeric` or `table-cell-datetime`) - it will allow to apply some styles by column type, - don't know if we need it, but it may be useful.

arikfr:
💃

PR #2243, opened 2018-01-19
Disable "Download Dataset" button when query hasn't been saved

### LINKED ISSUE #2232: Download dataset for unsaved query causing Internal Server Error
### Issue Summary

A new, unsaved query dataset download responds with 
```
{ message: "Internal Server Error" }
```
Once the query is saved, the error goes away.

### Steps to Reproduce

1. **Open Queries** > **New Query**
2. Select your data source, write your query, hit **Execute**
3. Once you get the response, select **Download Dataset** > **CSV/Excel**

If you closely look at the URL for the downloads you would see an undefined in the URL. This section of the path is normally used for the query id.
screenshot 2018-01-18 17 01 54

### Technical details:

* Redash Version: Redash 3.0.0+b3134
* Browser/OS: Ubuntu 16.04.3 LTS
* How did you install Redash: via [bootstrap.sh](https://raw.githubusercontent.com/getredash/redash/master/setup/ubuntu/bootstrap.sh) at version 2

kyoshidajp:
Fix #2232

The reason why error raised is the query hasn't had an id. It should be saved before downloading a dataset.

I think this solution is simple and enough for this situation.

### DIFF
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -245,7 +245,7 @@
 
                     <div class="btn-group dropup" uib-dropdown>
                       <button type="button" class="btn btn-default dropup-toggle"
-                              ng-disabled="queryExecuting || !queryResult.getData()" aria-haspopup="true"
+                              ng-disabled="query.id === undefined || queryExecuting || !queryResult.getData()" aria-haspopup="true"
                               uib-dropdown-toggle aria-expanded="false">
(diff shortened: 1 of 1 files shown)


arikfr:
This issue was fixed already in #2134, so users can download results of an unsaved query.

kyoshidajp:
Thanks!

Issue #2234, opened 2018-01-18
Further Query Editor Improvements


kocsmy:
- [x] Fix Smaller screen glitches @kocsmy
- [x] Attach refresh schedule and created/edited by to the top, increase the size of schema browser and remove resizer there @kocsmy @kravets-levko 
- [x] Scroll in each region, instead of the entire page scrolling (fixed layout, with body overflow: hidden) @kravets-levko 
- [x] Clickable resizer toggle to quickly open and expand section @kravets-levko

kocsmy:
Current status:
![screenshot 2018-01-19 16 03 51](https://user-images.githubusercontent.com/2378022/35156866-74458854-fd32-11e7-8c6f-2a36051e0690.png)
![screenshot 2018-01-19 16 04 16](https://user-images.githubusercontent.com/2378022/35156874-77a2fdba-fd32-11e7-8a2c-db65c8d31c3d.png)

There are some layout glitches which will be resolved once we add "Scroll in each region, instead of the entire page scrolling".

kocsmy:
Idea: the refresh schedule is kinda alone there. I'd put the action on "Updated an hour ago". For example: Updated an hour ago (Every 12h) and the later one is a link to change that.

kocsmy:
Another idea: Publish button is inside the content, maybe we should move it out, next to the Show Data Only? WDYF @arikfr 
![screenshot 2018-01-19 16 08 15](https://user-images.githubusercontent.com/2378022/35157057-1d9a13de-fd33-11e7-9d4b-ca2229e0ddb0.png)

arikfr:
>Idea: the refresh schedule is kinda alone there. I'd put the action on "Updated an hour ago". For example: Updated an hour ago (Every 12h) and the later one is a link to change that.

Possible, but might make it less discoverable, no?

arikfr:
> Another idea: Publish button is inside the content, maybe we should move it out, next to the Show Data Only? 

Seems ok.

kocsmy:
> Possible, but might make it less discoverable, no?

Not sure, the current position is also out of context. It'd be in context at least.

PR #2245, opened 2018-01-19
Further Query Improvements

kocsmy:
Initial commit for #2234

### DIFF
--- a/client/app/assets/less/inc/flex.less
+++ b/client/app/assets/less/inc/flex.less
@@ -1,0 +1,38 @@
+.d-flex         { display: flex !important; }
+.d-inline-flex  { display: inline-flex !important; }
+
+.flex-row            { flex-direction: row !important; }
+.flex-column         { flex-direction: column !important; }
+.flex-row-reverse    { flex-direction: row-reverse !important; }
+.flex-column-reverse { flex-direction: column-reverse !important; }
+
+.flex-wrap         { flex-wrap: wrap !important; }
+.flex-nowrap       { flex-wrap: nowrap !important; }
+.flex-wrap-reverse { flex-wrap: wrap-reverse !important; }
+.flex-fill         { flex: 1 1 auto !important; }
+
+.justify-content-start   { justify-content: flex-start !important; }
+.justify-content-end     { justify-content: flex-end !important; }
+.justify-content-center  { justify-content: center !important; }
+.justify-content-between { justify-content: space-between !important; }
+.justify-content-around  { justify-content: space-around !important; }
+
+.align-items-start    { align-items: flex-start !important; }
+.align-items-end      { align-items: flex-end !important; }
+.align-items-center   { align-items: center !important; }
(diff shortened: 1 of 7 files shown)


kocsmy:
Thanks for the cleanup and layout improvements, @kravets-levko 

I can spot 2 issues:
- [ ] Minimalising the query editor makes the top part of the visualisation go below the hidden area and disappear:
screenshot 2018-01-25 00 15 30

- [ ] When the editor is closed, the metadata and meta refresh and history goes up. It should still stick to the bottom:
screenshot 2018-01-25 00 15 41

kravets-levko:
@kocsmy Fixed. Can you please check and confirm that everything is ok

kocsmy:
Thanks for the quick fixes, @kravets-levko — all look good now!

arikfr:
Some comments:

Important: 
The data area needs its own scroll - right now the scroll is shared between the data area and the query text. This has two down sides: when you scroll the data you can't see the query but more important: you need to scroll down to be able to scroll to the side.

Nice to have: (can be fixed later on, unless very simple)
* The refresh schedule is part of the query metadata, so it doesn't need a divider on top of it. The vertical divider on the right should be seen on it.
image

* In data only view, the left side is very empty. Maybe we should use a horizontal area for metadata in such case?

kocsmy:
Thanks @arikfr 

- Regarding the scroll: yes, we discussed that with @kravets-levko and we weren't sure, I think he can implement that version.
- Refresh schedule: You mean like this? 
screenshot 2018-01-28 13 20 02

- Too empty left side: I wouldn't move the bottom fixed metadata to the top, I actually tried that and having it move around is more annoying than having some empty space there. Maybe we can add something there later.

arikfr:
> Refresh schedule: You mean like this?

Yes and probably with less padding between it and the "created/updated" part.

> Too empty left side: I wouldn't move the bottom fixed metadata to the top, I actually tried that and having it move around is more annoying than having some empty space there. Maybe we can add something there later.

That's because you think about this as being the same page in two different states. But what if you think about it as an independent page?

kocsmy:
> That's because you think about this as being the same page in two different states. But what if you think about it as an independent page?

But this is the same page in different state?:) Okay, if I make myself think about that way, then we don't need the left sidebar at all and we can go with a totally different layout. But afaik we just moved away from that and made a decision to show the schema section always, even if it's empty? Anyway, I got what you're saying and I agree that left side feels very empty. Not sure how to address this, but let's discuss on our call.

arikfr:
This PR needs a rebase...

arikfr:
There is still one thing left and it's the issue with the "Refresh Schedule" not being part of the metadata above it. I merged regardless so we don't keep this large pull request hanging.

Created #2279 to track this.

kocsmy:
Thanks @arikfr — yes, that's totally kept in mind and we'll make that happen, too!

PR #2246, opened 2018-01-20
[WIP]Enable list query by one character when adding widget on Dashboard

kyoshidajp:
I think it would be better to list queries by one character on "Add Widget" dialog.

Before:
sample dashboard 2018-01-20 12-19-38

After:
sample dashboard 2018-01-20 12-14-22

This change will lead to increase search access. Another solution is delay refresh, or defining this behavior as Settings.


kocsmy:
@kyoshidajp thanks a lot for this, this is massively useful! We'll also list the most recent (or even all) queries here.

arikfr:
Thanks!

PR #2247, opened 2018-01-20
fix: commit saml user group assignment to the database

### LINKED ISSUE #1668: SAML group assignment doesn't seem to work
According to the documentation, sending a `RedashGroups` attribute should set the user's groups.  However, this doesn't seem to work.  

Adding `pprint` debugging in `saml_auth.py`, I can see the necessary attribute is being sent:

```
[2017-03-13 14:12:58,516][PID:923][INFO][saml2.response] Subject NameID: <?xml version='1.0' encoding='UTF-8'?>
<ns0:NameID xmlns:ns0="urn:oasis:names:tc:SAML:2.0:assertion" Format="urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress" NameQualifier="https://login.torchbox.com/idp/shibboleth" SPNameQualifier="https://redash.torchbox.com/">felicity@torchbox.com</ns0:NameID>
{'FirstName': ['Felicity'],
 'LastName': ['Tarnell'],
 'RedashGroups': ['wagtail',
                  'tech',
                  'developers',
                  'drupal',
                  'admin',
                  'sysadmin'],
 'mail': ['felicity@torchbox.com']}
[2017-03-13 14:13:00,362][PID:923][INFO][metrics] method=POST path=/saml/callback endpoint=saml_auth.idp_initiated status=302 content_type=text/html; charset=utf-8 content_length=209 duration=2076.67 query_count=4 query_duration=1527.05
```

However, while the user is logged in correctly, they are not a member of any groups besides `default`.   (I tried pre-creating a new group called `sysadmin`, which didn't make any difference; the `admin` group already exists anyway.)

sjakthol:
Relevant tests were also modified to reload the user
object from the database before asserting that the
user belongs to correct groups.

Fixes #1668.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -515,6 +515,7 @@ class User(TimestampMixin, db.Model, BelongsToOrgMixin, UserMixin, PermissionsCh
         groups.append(self.org.default_group)
         self.group_ids = [g.id for g in groups]
         db.session.add(self)
+        db.session.commit()
 
     def has_access(self, obj, access_type):
         return AccessPermission.exists(obj, access_type, grantee=self)
--- a/tests/models/test_users.py
+++ b/tests/models/test_users.py
@@ -1,12 +1,14 @@
 from tests import BaseTestCase
 
-from redash.models import User
+from redash.models import User, db
 
 class TestUserUpdateGroupAssignments(BaseTestCase):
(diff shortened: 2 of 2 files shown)


arikfr:
Thanks!

PR #2248, opened 2018-01-20
feat: make trusted header authentication compatible with multiorg mode

sjakthol:
The previous implementation of remote header login did not
support the multiorg mode of Re:Dash. These changes modify
the trusted header authentication to expose a per-organization
login endpoint that logs users in the specified organization.

The feature itself is not that interesting as multiorg is pretty
much impossible to use in a standalone Re:Dash installation. What's
more interesting is that all tests are executed in multiorg mode. It's
not possible to write tests for the trusted header authentication if
the method does not support multiorg mode.

To make benefits of these changes more concrete, some tests were
written to test the basic functionality of trusted header
authentication.


sjakthol:
Fixed

arikfr:
Merged. Thanks!

PR #2249, opened 2018-01-20
fix: update correct settings in saml settings form

sjakthol:
The entity_id and nameid_format form fields didn't work as they
tried to save a setting with an incorrect name. This has now
been fixed and the form can be used to configure these values.

A short debounce delay was also added to the form to avoid
saving the value after every keystroke (if values are inputted
manually).


arikfr:
Thanks!

arikfr:

PR #2260, opened 2018-01-26
Extend the Prometheus query runner to support the range query

jubel-han:
The Prometheus HTTP API supports the two type of expression queries, [instant queries](https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries) and [range queries](https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries)
But the current Prometheus query runner only supports the `instant` query which means that we can **NOT** visualize the time series metrics for the period of time.

This PR will help to meet it.

### DIFF
--- a/redash/query_runner/prometheus.py
+++ b/redash/query_runner/prometheus.py
@@ -56,10 +56,6 @@ class Prometheus(BaseQueryRunner):
     def annotate_query(cls):
         return False
 
-    @classmethod
-    def type(cls):
-        return "Prometheus"
-
     def test_connection(self):
         resp = requests.get(self.configuration.get("url", None))
         return resp.ok
--- a/redash/query_runner/prometheus.py
+++ b/redash/query_runner/prometheus.py
@@ -38,21 +38,6 @@ def get_range_rows(metrics_data):
 
 
 class Prometheus(BaseQueryRunner):
-    def __init__(self, configuration):
-        self.columns = [
-            {
-                'friendly_name': 'timestamp',
-                'type': TYPE_DATETIME,
-                'name': 'timestamp'
-            },
-            {
-                'friendly_name': 'value',
-                'type': TYPE_STRING,
-                'name': 'value'
-            },
-        ]
-
-        super(Prometheus, self).__init__(configuration)
 
     @classmethod
     def configuration_schema(cls):
@@ -70,6 +55,10 @@ class Prometheus(BaseQueryRunner):
     @classmethod
     def annotate_query(cls):
         return False
+
+    @classmethod
+    def type(cls):
+        return "Prometheus"
 
     def test_connection(self):
         resp = requests.get(self.configuration.get("url", None))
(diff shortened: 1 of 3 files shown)


jubel-han:
@RichardLitt when are we going to merge this PR?

RichardLitt:
Good question, @jubel-han. This PR needs to be reviewed first; I'm not technically able to help with that yet. Redash is run by volunteers, which means that we'll get to this when we can, but we can't make promises about merging it at any specific point in the future. 

This is not a huge PR, so I would think that we should be able to merge this relatively soon. You've also added tests, which help a lot - thank you!

redash/query_runner/prometheus.py
```diff
@@ -40,52 +89,66 @@ def get_schema(self, get_stats=False):
...
-            rows = []
-            for row in raw_data:
-                h = {}
-                for r in row['metric']:
-                    h[r] = row['metric'][r]
-                    h['value'] = row['value'][1]
-                    h['timestamp'] = datetime.datetime.fromtimestamp(row['value'][0])
-                rows.append(h)
 
+            rows = get_range_rows(metrics) if query_type == 'query_range' else get_instant_rows(metrics)
```
arikfr: I think this will be easier to read as a regular `if`:

```python
if query_type == 'query_range' :
    rows = get_range_rows(metrics) 
else:
    rows = get_instant_rows(metrics)
```

(there is no need to predefine `rows` as `if` doesn't create a new scope in Python)

redash/query_runner/prometheus.py
```diff
@@ -40,52 +89,66 @@ def get_schema(self, get_stats=False):
...
+
+            metrics = response.json()['data']['result']
+
+            if len(metrics) == 0:
+                return None, 'query result is empty.'
+
+            metric_labels = metrics[0]['metric'].keys()
+
+            for label_name in metric_labels:
+                self.columns.append({
```
arikfr: While usually we don't reuse a query runner, this is not guaranteed behavior, so it's better to have `columns` to be local to the function instead of a class variable.

jubel-han: fixed.

redash/query_runner/prometheus.py
jubel-han: fixed.

jubel-han:
@arikfr all fixed, please have another review.

redash/query_runner/prometheus.py
```diff
@@ -22,6 +56,10 @@ def configuration_schema(cls):
     def annotate_query(cls):
         return False
 
+    @classmethod
+    def type(cls):
+        return "Prometheus"
```
arikfr: This is not necessary and a bad idea, because it will break for all the existing users who already created data sources for Prometheus (where type is `prometheus`).

jubel-han: @arikfr fixed, please have another review

jubel-han:
@arikfr fixed `type`, please have another review.

arikfr:
Thanks :zap:

PR #2266, opened 2018-01-29
Url password can have quoted special characters.

javier-sanz:
As described on this Redash [forum support issue](https://discuss.redash.io/t/redis-quoted-password/1435) password can have special characters and those ones can be `quoted` (encoded). They must be `unquoted ` before being sent otherwise the validation process will fail.

### DIFF
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -55,10 +55,8 @@ def create_redis_connection():
         else:
             redis_db = 0
         # Redis passwords might be quoted with special characters
-        redisPassword = ""
-        if redis_url.password is not None:
-            redisPassword = urllib.unquote(redis_url.password)
-        r = redis.StrictRedis(host=redis_url.hostname, port=redis_url.port, db=redis_db, password=redisPassword)
+        redis_password = redis_url.password and urllib.unquote(redis_url.password)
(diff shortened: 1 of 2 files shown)


redash/__init__.py
```diff
@@ -53,8 +54,11 @@ def create_redis_connection():
             redis_db = redis_url.path[1]
         else:
             redis_db = 0
-
-        r = redis.StrictRedis(host=redis_url.hostname, port=redis_url.port, db=redis_db, password=redis_url.password)
+        # Redis passwords might be quoted with special characters
+        redisPassword = ""
+        if redis_url.password is not None:
+            redisPassword = urllib.unquote(redis_url.password)
+        r = redis.StrictRedis(host=redis_url.hostname, port=redis_url.port, db=redis_db, password=redisPassword)
```
arikfr: We use `snake_case` and not `camelCase`. And this can be simplified to:

```python
redis_password = redis_url.password and urllib.unquote(redis_url.password)
```

If the password is `None`, it will just assign `None`.

javier-sanz:
@arikfr I fixed it following your comments. I added another change on the helper.py that I think it has the same issue. Let me know if you thing that change is not needed.

arikfr:
Thanks !

javier-sanz:
NP 😃

Issue #2161, opened 2017-12-13
Remove loading indicator when updating query parameter value (before executing)


arikfr:
Pace gets triggered whenever History pushState is called. We actually need to trigger only when there is an HTTP request.

We can either configure Pace to ignore pushState or switch to: https://github.com/chieffancypants/angular-loading-bar.

kocsmy:
Yes, please! I believe this is massively disturbing. Also we'll need a Run/Execute button next to the parameter so users don't need spend time figuring out how it works (current Execute button is far from this element and very much disconnected so it's hard to figure out).

PR #2269, opened 2018-01-30
Remove loading indicator when updating query parameter value (before executing)

kravets-levko:
Issue getredash/redash#2161

When history event occurs (`pushState` or `replaceState`) do not show progress bar if only query or hash changed; show it only if URL path changed.

Depends on: https://github.com/getredash/pace/pull/1


arikfr:
👍 

Don't we need to update `package-lock.json` to pick up the new revision of `pace`?

kravets-levko:
Need to check. We use it via direct URL, but if it saves commits hash - then yes, we need to update it

kravets-levko:
Updated :+1:

PR #2270, opened 2018-01-30
Upgrade to Celery 3.1.25 in preparation to Celery 4.

jezdez:
To prepare for a future Celery upgrade to verson 4.x this upgrades
Redash to 3.1.25 which includes support for Celery's task protocol 2.
It's the first of two steps to get Redash upgraded to Celery 4.x.

More info:
- [3.1.4 release notes](http://docs.celeryproject.org/en/3.1/changelog.html#version-3-1-24) regarding the new task protocol.
- [4.0 release notes](http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#step-1-upgrade-to-celery-3-1-25) regarding the two-step-upgrade approach.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -32,7 +32,7 @@ wsgiref==0.1.2
 honcho==0.5.0
 statsd==2.1.2
 gunicorn==19.7.1
-celery==3.1.23
+celery==3.1.25
 jsonschema==2.4.0
 RestrictedPython==3.6.0
 pysaml2==4.5.0


jezdez:
Due to the way the task protocol changes, the upgrade to 4.x needs to happen in two steps. I propose to go to 3.1.25 in the next release to force the task protocol upgrade. And then to 4.x in the next after that.

I strongly suggest to way for Celery 4.2 to be released first though as it will include a critical fix for hybrid 3.x/4.x clusters that prevents 4.x workers to process task messages that were created by 3.x clients. More details in the [issue](https://github.com/celery/celery/issues/4356) and in this [excellent gist](https://gist.github.com/ewdurbin/ddf4b0f0c0a4b190251a4a23859dd13c).

arikfr:
Thanks!

We will still need to think about how to do the Celery 4.2 upgrade, because some users might upgrade straight from `Redash<=3` (which uses `Celery <3.1.25`) to the version where we introduce Celery 4.x.

`<wishful-thinking>` Maybe we end up switching to RQ before this happens 😆 `</wishful-thinking>`

Issue #2241, opened 2018-01-19
Insert node by click bug


denisov-vlad:
### Issue Summary

After this pull request https://github.com/getredash/redash/pull/2225 it's possible to insert table / column by click. 

Column is inserted in the following format: {schema_name}.{table_name}.{column_name}. It's not suitable for PostgreSQL, for example, and should be like this: {table_name}.{column_name} or {column_name}.

### Steps to Reproduce

1. Open query editor with PostgreSQL datasource.
2. Insert column by click
3. Try to run query.

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* Redash Version: master
* Browser/OS: Safari / MacOS
* How did you install Redash: script.

kravets-levko:
@denisov-vlad I checked the code, and it works as expected: it inserts `{table_name}` when you click top-level node (table name) in schema browser, and it inserts `{table_name}.{column_name}` when you click second-level nodes (column names). It work exactly in the same way as auto-completion dropdown.

But at the same time I found another bug that may cause errors during execution - it's already merged (https://github.com/getredash/redash/pull/2275). So I kindly ask you - can you please check that fix and give me some feed back - do you still see your issue, or was it fixed? Thanks a lot!

denisov-vlad:
It's ok now. Thanks.

PR #2275, opened 2018-01-31
Query editor insert node from schema browser update ngModel

kravets-levko:
[This PR](https://github.com/getredash/redash/pull/2225) added new feature - ability to insert schema node into query editor by click in schema browser. But editor did not update ngModel value immediately - only on blur or user input. This PR fixes that bug.

Probably related to issue getredash/redash/issues/2241

### DIFF
--- a/client/app/components/queries/query-editor.js
+++ b/client/app/components/queries/query-editor.js
@@ -44,6 +44,7 @@ function queryEditor(QuerySnippet) {
           onLoad(editor) {
             $scope.$on('query-editor.paste', ($event, text) => {
               editor.session.doc.replace(editor.selection.getRange(), text);
+              $scope.query.query = editor.session.getValue();
             });
 
             // Release Cmd/Ctrl+L to the browser


arikfr:
The issue in #2241 is a different one and it's with how we build the final value that gets inserted. I think that we shouldn't insert the full path (i.e. `table.column`), but just the current highlighted value - be it a table or a column.

kravets-levko:
@arikfr I used auto-completion dropdown as an examplt when implementing that - it also inserts `table` or `table.column`, and, actually, it works without any issues. But this PR fixes the but which looks very similar in UI - say, when you write `select * from ` and then insert table name from schema browser - it will not update model, therefore you'll execute only `select * from ` and will get syntax error. So I decided that author of getredash/redash/issues/2241 may just misinterpret the error message. I'll ask him to check his issue with this fix - if I'm right - his problem will gone, if I'm not - I'll check it again.

arikfr:
The auto complete works a bit different: it suggest both `column` and `table.column`:

image

kravets-levko:
You're right :-) but at the same time there are two things:

1. when you click `table.column` item - it inserts `table.column`; it's completely valid syntax in all SQL dialects, queries like `select table.id from table` will run successfully.
2. in schema browser we cannot offer both options (for obvious reasons). So I decided to always insert fully qualified name: it will not change anything if query contains the only table, and it will help if there are some joins.

arikfr:
> it's completely valid syntax in all SQL dialects, queries like select table.id from table will run successfully.

True, _but_ if you alias a table then it won't work. Here's an example:

```sql
select queries.id
from queries q
limit 1
```

Because there is no control over this and because sometimes the table name will include the schema name, I think it's safer bet to insert only the column name.

kravets-levko:
Makes sense. Will fix it.

PR #2276, opened 2018-01-31
Insert schema node by click - fixes

kravets-levko:
Fixes related to getredash/redash/pull/2275

- Insert only column name (without table name) - discussed with @arikfr.
- Save cursor position after table/column name inserted.


arikfr:
👍

PR #2280, opened 2018-02-01
[database] Add index on table events column object_id

marcmillien:
This PR adds a migration to add an `INDEX` on table `events` column `object_id` to avoid unresponsive redash due to the possible `events` table size.


RichardLitt:
Thanks @marcmillien. Was this tied to a specific issue?

marcmillien:
Yes, we currently have more than 400 000 rows on the `events` table, and our postgres server is a small one.

We had redash unresponsive due to a request with `JOIN` between `events`, `users` and an other one. 
Unfortunatly I haven't saved what was the exact request during the issue :sob: .

PR #2282, opened 2018-02-01
Improve visualisation colors

kocsmy:
We're doing a couple of improvements on the default visualisation colors.

This is an initial version with vivid colors, not final and we'll try different color schemes.

screenshot 2018-02-01 22 39 53
screenshot 2018-02-01 22 40 01


kocsmy:
Added a couple of improvements:
screenshot 2018-02-17 18 27 23

Less vivid colors and they should also look better next to each other.

Issue #2138, opened 2017-12-07
Add a new column type for images


yalisassoon:
In table visualizations there are a number of use cases where it is nice to embed images in a column e.g.:

1. In retail, for product reports, having a visualization of the product
2. Having a logo or a map if you're reporting by brand or greography
3. Having a "flag" if e.g. a metric is above or below a certain threshold

I believe that embedding images is possible now via a hack - it would be nice to make this feature "official"

arikfr:
Here's an example for the current workaround: http://demo.redash.io/queries/1896/source#table.

kravets-levko:
Hi @yalisassoon! I started working on your request, and here is my PR: getredash/redash/pull/2283 It's in progress, but if you have any notices or ideas - feel free to comment there. Thanks!

RichardLitt:
Thanks @kravets-levko! This was merged a couple of days ago, and should be all set now.

kravets-levko:
You're right @RichardLitt, it was merged few days ago, but also few day after my invitation to discuss it :smiley:

PR #2283, opened 2018-02-02
Table visualization: "Image" column type

kravets-levko:
Issue getredash/redash#2138

Implementation details:

Added new column type ("Image") and several options for it: "URL Template", "Width"/"Height", "Title template" (see screenshots below). All options allows to use values from another columns via `{{ column_name }}` syntax. It allows to construct image URL from many parts, or construct `data:` URLs when image data and MIME type are stored in different columns, etc. Also, it allows to use other columns for image size and `title` attribute.

image
image
image

### DIFF
--- a/client/app/components/dynamic-table/image-cell/index.js
+++ b/client/app/components/dynamic-table/image-cell/index.js
@@ -1,0 +1,66 @@
+import { isUndefined } from 'underscore';
+
+const hasOwnProperty = Object.prototype.hasOwnProperty;
+
+function trim(str) {
+  return str.replace(/^\s+|\s+$/g, '');
+}
+
+function processTags(template, data, defaultColumn) {
+  return template.replace(/{{\s*([^\s]+)\s*}}/g, (match, column) => {
+    if (column === '@') {
+      column = defaultColumn;
+    }
+    if (hasOwnProperty.call(data, column) && !isUndefined(data[column])) {
+      return data[column];
+    }
+    return match;
+  });
+}
+
+function buildImgTag(column, row) {
+  const url = trim(processTags(column.imageUrlTemplate, row, column.name));
+  const width = parseInt(processTags(column.imageWidth, row, column.name), 10);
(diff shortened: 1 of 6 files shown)


client/app/components/dynamic-table/image-cell/index.js
```diff
@@ -0,0 +1,66 @@
...
+  return template.replace(/{{\s*([^\s]+)\s*}}/g, (match, column) => {
+    if (column === '@') {
+      column = defaultColumn;
+    }
+    if (hasOwnProperty.call(data, column) && !isUndefined(data[column])) {
+      return data[column];
+    }
+    return match;
+  });
+}
```
arikfr: I'm not sure if we gain anything by it, but maybe use [`_.template`](http://underscorejs.org/#template)?

client/app/components/dynamic-table/image-cell/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+function buildImgTag(column, row) {
+  const url = trim(processTags(column.imageUrlTemplate, row, column.name));
+  const width = parseInt(processTags(column.imageWidth, row, column.name), 10);
+  const height = parseInt(processTags(column.imageHeight, row, column.name), 10);
+  const title = trim(processTags(column.imageTitleTemplate, row, column.name));
+
+  const result = [];
+  if (url !== '') {
+    result.push('<img src="' + url + '"');
```
arikfr: Can someone inject script tags this way?

i.e., make `url` something like `"><script src="evil.js"><img src="`? Or it will be cleaned out by `$sanitize` below?

kravets-levko: I checked it - `$sanitize` removes all "bad" things, so it's safe.

client/app/components/dynamic-table/image-cell/index.js
kravets-levko: I think `_.template` is too much for such a simple thing.

PR #2287, opened 2018-02-04
Table visualization: "Link" column type

kravets-levko:
Related to this PR getredash/redash/pull/2283 and works in a similar way.

Added new column type ("Link") and several options for it: "URL template", "Text template", "Title template", "Open in new tab" (adds `target="_blank"` attribute) (see screenshots below). All options allows to use values from another columns via `{{ column_name }}` syntax.

image
image
image

P.S. Sorry for a lot of changes comparing to PR getredash/redash/pull/2283 - I moved some code to better place, and somewhere `git` wrongly detected file renaming instead of delete + adding new file.

### DIFF
--- a/client/app/components/dynamic-table/default-cell/index.js
+++ b/client/app/components/dynamic-table/default-cell/index.js
@@ -1,3 +1,5 @@
+import { contains } from 'underscore';
+import { renderDefault, renderImage, renderLink } from './utils';
 import template from './template.html';
 
 export default function init(ngModule) {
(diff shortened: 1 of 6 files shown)


client/app/components/dynamic-table/default-cell/index.js
```diff
@@ -7,10 +9,35 @@ export default function init(ngModule) {
...
+      } else {
+        $scope.allowHTML = contains(['image', 'link'], $scope.column.displayAs);
+      }
+
+      let renderValue;
+      switch ($scope.column.displayAs) {
+        case 'image': renderValue = renderImage; break;
+        case 'link': renderValue = renderLink; break;
+        default: renderValue = renderDefault; break;
+      }
```
arikfr:
```javascript
const types = {image: renderImage, link: renderLink};
const renderValue = types[$scope.column.displayAs] || renderDefault;
```

How about this instead of a switch case?

Right now it's not a huge difference, but as the column types list grows it will become more manageable.

kravets-levko: :+1:

arikfr:
💃

Issue #2209, opened 2018-01-08
Blank page issue on Safari (iOS and MacOS)


arikfr:
Sometimes Safari caches empty Javascript files, which results in blank pages until the user clears cache.

It might be related to this Webkit issue:
https://bugs.webkit.org/show_bug.cgi?id=32829

(also related: http://bytes.schibsted.com/ios7-bug-shows-white-page-when-getting-304-not-modified-from-server/)

elipollak:
+1

PR #2296, opened 2018-02-07
Assets serving changes

arikfr:
Some changes to the way we serve assets:

* All assets served from `/static/`.
* All assets go through Webpack (in one way or another).
* Template helper for Flask to load assets using Webpack's manifest file (copied from Zeus).

Breaking change: `REDASH_STATIC_ASSETS_PATH` is a single value again and no longer supports multiple paths.

Might help with #2209.


webpack.config.js
```diff
@@ -57,8 +59,17 @@ const config = {
...
+      filename: 'styles.[chunkhash].css',
+      publicPath: '/assets/'
+    }),
+    new ManifestPlugin({
+      fileName: 'asset-manifest.json'
+    }),
+    new CopyWebpackPlugin([
+      { from: 'client/app/assets/robots.txt' },
+      { from: 'client/app/assets/css/login.css', to: 'styles/login.css' },
+      { from: 'client/app/assets/js/jquery.min.js', to: 'js/jquery.min.js' },
```
kravets-levko: Do we really need to store `jquery.min.js` in `app/assets` folder? We already have jquery and all pre-built bundles in `node_modules`

arikfr: Probably not. But then we need to create a new entry point that will export only jQuery?

kravets-levko: Can you try to replace `from` path to be `node_modules/jquery/dist/jquery.min.js`? Will it work?

arikfr: Brilliant. Done 👍 96ad3aaa259b8f97844eb0b9722b02c45ed970b4

arikfr:
Thanks! 

> P.S. It's cool that you finally removed old css files from client/app/assets/css 👍

It's been long overdue. Next thing can be to create a dedicated login page CSS, but maybe it's better to load the large one to save download time on the next page?

Issue #2295, opened 2018-02-07
Filters dropdown is "trapped" inside the widget area


arikfr:
![image](https://user-images.githubusercontent.com/71468/35921962-e1831430-0c24-11e8-8112-f69e9f47f769.png)

(apologies for the blue tint)

kravets-levko:
I already fixed it earlier, when we were implementing new dashboards. Strange that it happens again...

arikfr:
Maybe got lost in one of the bigger merges?

kravets-levko:
Probably yes. Or something happened during re-design. Anyway, it will be fixed ASAP :+1:

kravets-levko:
BTW - which page it that?

arikfr:
`/dashboard/filters-example`

kravets-levko:
Noticed that I was fixing parameters, not filters. So I take my words back - I have never fixed this bug :-)

PR #2297, opened 2018-02-07
Filters dropdown is "trapped" inside the widget area

kravets-levko:
Issue getredash/redash#2295

This PR also contains some UI tweaks for dropdowns (applies some common styles).

image


arikfr:
💃

PR #2300, opened 2018-02-08
Do not extract the topmost directory from tarball

devforfu:
The bootstrap script was creating a bit odd directory tree when I've called it, something like `~/redash/redash.3.0.0/redash-3.0.0`. Though I was using not an AWS link, but one of Github's release URLs.

### DIFF
--- a/setup/ubuntu/bootstrap.sh
+++ b/setup/ubuntu/bootstrap.sh
@@ -65,7 +65,7 @@ create_directories() {
 extract_redash_sources() {
     sudo -u redash wget "$LATEST_URL" -O "$REDASH_TARBALL"
     sudo -u redash mkdir "$VERSION_DIR"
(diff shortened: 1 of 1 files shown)


arikfr:
What is the tarball you were using?

devforfu:
@arikfr This one:https://github.com/getredash/redash/archive/v3.0.0.tar.gz), from [Releases](https://github.com/getredash/redash/archive/v3.0.0.tar.gz) page.

arikfr:
This is auto generated by GitHub, but not the actual release you should use. You need to use the releases from https://version.redash.io. You can see them already referenced in the bootstrap and upgrade scripts.

devforfu:
Ok, understood.

RichardLitt:
Thank you for being proactive, anyway, @devforfu. :)

PR #2301, opened 2018-02-08
Fix webpack config and some assets URLs

kravets-levko:


webpack.config.js
```diff
@@ -143,10 +143,14 @@ const config = {
...
-      { from: /./, to: '/static/index.html' }
-    ]},
-    contentBase: path.join(__dirname, 'client', 'app'),
+    index: '/static/index.html',
+    historyApiFallback: {
+      index: '/static/index.html',
+      rewrites: [
+        { from: /.?/, to: '/static/index.html' },
+      ],
+    },
```
arikfr: Maybe this is not needed anymore with the `index` setting?

kravets-levko: no, I tried - it does not work without it. actually, I don't understand why, but it does not work.

kravets-levko: hm, well, I'm wrong, it works without this line. will remove it

arikfr:
💃

PR #2303, opened 2018-02-09
Fix webpack config (server-rendered pages); fix assets URLs

kravets-levko:
Supplements PR getredash/redash/pull/2301

There are some caveats with config in previous PR related to server-rendered pages (like login page, invite user page, etc.). In short: all that pages contains assets URLs with hash (files are built with `npm run build` and are located in `client/dist` folder). Previous webpack config tries to mount that folder as path with static files (`contentBase` option), but it leads to some bugs in a case when `client/dist` folder does not exist (some resources does not load), or filename conflicts (say, there is "virtual" `index.html` generated by `webpack-dev-server`, and also "physical" one in `client/dist`. In some cases WDS serves "physical" one).

This PR fixes that issues in next way:
1. WDS will not serve any static files; everything needed for front-end code is already in WDS's memory and served by it (including fonts, images, etc.).
2. Files needed by server-rendered pages (it's only CSS and JS with "salt" in filename) are proxied to backend and served by it - so no more conflicts between generated and "physical" files.

Also this PR fixes some assets URLs missed in previous one, and removes unnecessary proxy URLs (now all assets are served from '/static', so no need to proxy all that '/styles', '/images' etc.).

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -6,9 +6,9 @@
     <base href="/">
     <title>Redash</title>
 
-    <link rel="icon" type="image/png" sizes="32x32" href="/images/favicon-32x32.png">
-    <link rel="icon" type="image/png" sizes="96x96" href="/images/favicon-96x96.png">
-    <link rel="icon" type="image/png" sizes="16x16" href="/images/favicon-16x16.png">
+    <link rel="icon" type="image/png" sizes="32x32" href="/static/images/favicon-32x32.png">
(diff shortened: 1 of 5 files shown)


arikfr:
> Files needed by server-rendered pages (it's only CSS and JS with "salt" in filename) are proxied to backend and served by it - so no more conflicts between generated and "physical" files.

Very cool 👍

Issue #2214, opened 2018-01-09
Add Widget Dialog UX Update


arikfr:
The Add Widget dialog should always show a list of queries (preferably the last ones you edited) and this list will update when you start searching.

arikfr:
@kravets-levko what are you planning to use to show the initial queries list? recent queries?

kravets-levko:
Yes, it should be the most relevant to user. If no recent queries - I'll try to choose one of:
- show empty list (probably with hint or so);
- show user's queries (few newest);
- show few newest queries (not just users ones).
I will try these variants and choose what looks the best.

arikfr:
Let's start with the recents API and if we feel it's not enough, find alternatives.

arikfr:
(you can see recents usage in the homepage)

PR #2311, opened 2018-02-13
Add Widget Dialog UX Update

kravets-levko:
Issue getredash/redash#2214

After opening modal, user see recent queries; when user starts typing in search input, search results are loaded and shown; when user clears search input - recent queries are shown. Everything related to text widgets left unchanged.

Screenshots:

image
image
image

### DIFF
--- a/client/app/components/dashboards/add-widget-dialog.html
+++ b/client/app/components/dashboards/add-widget-dialog.html
@@ -10,7 +10,7 @@
 
   <div ng-show="$ctrl.isTextBox">
     <div class="form-group">
-      <textarea class="form-control" ng-model="$ctrl.text" ng-model-options="{ debounce: 200 }" rows="3"></textarea>
+      <textarea class="form-control" ng-model="$ctrl.text" ng-model-options="{ debounce: 200 }" rows="3" autofocus></textarea>
     </div>
     <div ng-show="$ctrl.text">
       <strong>Preview:</strong>
@@ -20,9 +20,8 @@
 
   <div ng-show="$ctrl.isVisualization">
     <div class="form-group">
-      <input type="text" placeholder="Search a query by name" class="form-control" autofocus="autofocus"
-        ng-if="!$ctrl.selectedQuery" ng-change="$ctrl.searchQueries($ctrl.searchTerm)"
-        ng-model="$ctrl.searchTerm" ng-model-options="{ debounce: 200 }">
+      <input type="text" placeholder="Search a query by name" class="form-control" autofocus
+        ng-if="!$ctrl.selectedQuery" ng-model="$ctrl.searchTerm" ng-change="$ctrl.searchQueries($ctrl.searchTerm)">
       <div ng-if="$ctrl.selectedQuery" class="p-relative">
         <input type="text" class="form-control bg-white"
           ng-value="$ctrl.selectedQuery.name" readonly="readonly">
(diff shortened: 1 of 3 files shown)


client/app/components/dashboards/add-widget-dialog.html
```diff
@@ -4,46 +4,75 @@ <h4 class="modal-title">Add Widget</h4>
...
-          <p ng-bind-html="$ctrl.text | markdown"></p>
+    </div>
+
+    <div ng-if="!$ctrl.selectedQuery && ($ctrl.searchTerm == '')" class="scrollbox" style="max-height: 50vh">
+      <div class="list-group" ng-if="$ctrl.recentQueries.length > 0">
+        <a href="javascript:void(0)" class="list-group-item"
+          ng-repeat="query in $ctrl.recentQueries" ng-click="$ctrl.selectQuery(query.id)"
+        >
+          <span>{{query.name}}</span>
+          <span class="label label-default" ng-if="query.is_draft">Unpublished</span>
```
arikfr: It not supposed to show unpublished queries here. I guess they show up, because the recents API returns them. You can either apply filtering on the server side or on the client once we have the recents.

client/app/components/dashboards/add-widget-dialog.html
```diff
@@ -4,46 +4,75 @@ <h4 class="modal-title">Add Widget</h4>
...
+          class="d-flex align-items-center justify-content-center"
+          style="position: absolute; right: 1px; top: 1px; bottom: 1px; width: 30px; background: #fff; border-radius: 3px;"
+        ><i class="text-muted fa fa-times"></i></a>
       </div>
-      <div ng-show="$ctrl.text">
-          <strong>Preview:</strong>
-          <p ng-bind-html="$ctrl.text | markdown"></p>
+    </div>
+
+    <div ng-if="!$ctrl.selectedQuery && ($ctrl.searchTerm == '')" class="scrollbox" style="max-height: 50vh">
```
arikfr: Why not have a single top level `<div ng-if="!$ctrl.selectedQuery" class="scrollbox" style="max-height: 50vh">` and then inside have the logic of whether we have searchTerm/recents/results.

client/app/components/dashboards/add-widget-dialog.html
```diff
@@ -4,46 +4,75 @@ <h4 class="modal-title">Add Widget</h4>
...
+      <div class="list-group" ng-if="$ctrl.recentQueries.length > 0">
+        <a href="javascript:void(0)" class="list-group-item"
```
arikfr: Are you sure the `javascript:void(0);" part needed?

client/app/components/dashboards/add-widget-dialog.html
```diff
@@ -4,46 +4,75 @@ <h4 class="modal-title">Add Widget</h4>
...
-      <div ng-show="$ctrl.selected_query">
-          <div class="form-group">
-              <label for="">Choose Visualization</label>
-              <select ng-model="$ctrl.selectedVis" ng-options="vis as vis.name group by vis.type for vis in $ctrl.selected_query.visualizations" class="form-control"></select>
-          </div>
+    <div ng-show="$ctrl.selectedQuery">
+      <div class="form-group">
+        <label for="">Choose Visualization</label>
+        <select ng-model="$ctrl.selectedVis" class="form-control"
+          ng-options="vis as vis.name group by vis.type for vis in $ctrl.selectedQuery.visualizations"></select>
```
arikfr: Let's take the opportunity and make selected the visualization easier: instead of a dropdown, show small "cards" (like the ones we have for data sources/destinations) with the visualization name and type.

image

client/app/components/dashboards/add-widget-dialog.js
```diff
@@ -12,61 +13,76 @@ const AddWidgetDialog = {
...
+            }
           }
-        }
-      });
+        });
+      }
     };
 
-    this.searchQueries = (term) => {
+    // `ng-model-options` does not work with `ng-change`, so do debounce here
```
arikfr: If it doesn't, why keep the markup?

client/app/components/dashboards/add-widget-dialog.js
```diff
@@ -12,61 +13,76 @@ const AddWidgetDialog = {
...
+        this.searchedQueries = [];
         return;
       }
 
       Query.search({ q: term }, (results) => {
-        this.queries = results;
+        // If user will type too quick - it's possible that there will be
+        // several requests running simultaneously. So we need to check
+        // which results are matching current search term and ignore
+        // outdated results.
```
arikfr: 👌

client/app/components/dashboards/add-widget-dialog.html
kravets-levko: yes, otherwise links will not have their style (color, underline, etc.)

client/app/components/dashboards/add-widget-dialog.html
kravets-levko: Good idea. Should they be grouped somehow? Also I can add icons (same as in viz editor)

arikfr: In the drop down we group by type, but I don't think it's needed here.

Icons: I had the same thougth until I remembered that we have only icons for chart types, but not for top level visualization type :-( We probably should have some, but might take time until we collect for all of them.

client/app/components/dashboards/add-widget-dialog.html
arikfr: Here we have similar markup and I think it's ok, no?

https://github.com/getredash/redash/blob/17f3b1e92df35613677584958cc37e7aee783990/client/app/pages/dashboards/dashboard-list.html#L9-L12

client/app/components/dashboards/add-widget-dialog.html
kravets-levko: well, I forgot that we don't have icons for visualization types, just for chart types. So my question is not more relevant: the answer is - no any icons :-)

arikfr: :-)

client/app/components/dashboards/add-widget-dialog.html
arikfr: 👍

kravets-levko: ah, there is s small difference between links with and without `href`: link without `href` will have cursor like for regular text, but link with `href` will have `cursor: pointer` - it's defaults for most browsers. it will work also for `list-group`s (so I'll remove `href`s here) because bootstrap implicitly sets cursor for list items.

there also was a bug in old IE (up to IE8, I think) where links without `href` didn't have usual hover styling, but it's irrelevant for us.

client/app/components/dashboards/add-widget-dialog.html
kocsmy: @arikfr "show small "cards" (like the ones we have for data sources/destinations) with the visualization name and type" — we tried some designs there and it brings up a few questions that are non-trivial. Like these boxes doesn't look like you need to select one while the dropdown is very trivial. I'd recommend to drop this improvement for now as it increases the scope and we should focus on v4 release.

kocsmy: Here are some quick sketches to show the problem:

kravets-levko: @arikfr, let's leave it for now. we discussed it with @kocsmy yesterday, and ATM we have no idea that will match all conditions good enough (we need to show viz type, viz may have long name, it should look nice both when there are 1-2 viz and a lot of viz; if showing groups - we should remember that some groups may be large and some may contain 1-2 viz etc.). we will think about it and share our ideas, but it may take some time. I fixed other notices and will update PR in a moment.

arikfr: 👍

PR #2313, opened 2018-02-13
Fix: widget is not removed from dashboard, only after page refresh

kravets-levko:
Issue: when trying to remove widget from dashboard, it was removed on server, but not removed on client. All next attempts to update dashboard layout will fail because frontend will try to save non-existing widget and will get an error. This PR fixes it.

### DIFF
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -317,7 +317,11 @@ function DashboardCtrl(
   this.removeWidget = () => {
     this.extractGlobalParameters();
     if (!this.layoutEditing) {
-      saveDashboardLayout();
+      // We need to wait a bit for `angular-gridster` before it updates widgets,
+      // and only then save new layout
+      $timeout(() => {
+        saveDashboardLayout();
+      }, 50);
     }
   };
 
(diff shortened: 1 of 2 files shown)


client/app/pages/dashboards/dashboard.js
```diff
@@ -317,7 +317,11 @@ function DashboardCtrl(
   this.removeWidget = () => {
     this.extractGlobalParameters();
     if (!this.layoutEditing) {
-      saveDashboardLayout();
+      // We need to wait a bit for `angular-gridster` before it updates widgets,
+      // and only then save new layout
```
arikfr: Maybe there is some callback we can use instead of "magic timeout"?

kravets-levko: `angular-gridster` has not anything useful; there are some events like `gridster-item-transition-end` but to use it we need some complicated state management or every time bind/unbind event handler.  There is no event that is triggered when `angular-gridster` finished updating items - we can watch all items, but it's also very complicated. In addition, `gridster-item-transition-end` handler and watcher will be triggered for each changed item, and it's almost impossible to determine when they all are updated (some may not changed - no event, no watcher call). Therefore such a "magic"...

PR #2318, opened 2018-02-14
Support authentication for the URL query runner.

jezdez:
Adds a new BaseHTTPQueryRunner class and tests.

This is a backport from https://github.com/mozilla/redash/pull/336.

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -1,5 +1,8 @@
 import logging
 import json
+import sys
+
+import requests
 
 from redash import settings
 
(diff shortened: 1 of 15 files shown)


redash/query_runner/__init__.py
```diff
@@ -145,6 +147,93 @@ def _get_tables_stats(self, tables_dict):
...
+                )
+            )
+        except requests.RequestException as exc:
+            # Catch all other requests exceptions and return the error.
+            logger.exception(exc)
+            error = str(exc)
+        except Exception as exc:
+            # Catch any other exceptions, log it and reraise it.
+            logger.exception(exc)
+            raise sys.exc_info()[1], None, sys.exc_info()[2]
```
arikfr: We have this block of code in some places, but it's not really needed. Any unhandled exception is being taken care of by the execute query task.

jezdez: @arikfr I've removed those sections in https://github.com/getredash/redash/pull/2318/commits/8a4391ca436b15123caee9e78abb5d62dc8e103d (except the one in sqlite.py which does some extra unicode error handling).

Issue #2076, opened 2017-11-07
Refactoring: merge `dashboard` and `public-dashboard`


kravets-levko:
Currently there are two pages to display dashboard: `dashboard` and `public-dashboard`. First one is an extended version of the second, but with additional controls (editing, sharing, etc.). Since there is a lot of duplicated code, the idea is to merge that two pages (via extracting common code into components, services, etc.) and keep all logic related to dashboards in the single place.

deecay:
+1 for this. And by doing that, I would like to see dashboard level filters available in public dashboards. This is a key feature for me.

s1moe2:
+1 here for the exact same reasons that @deecay mentioned. Having the dashboard level filters available in public dashboards is super useful, specially because having individual filters on each visualisation completely breaks the user experience, in my opinion (if you have multiple visualisations)

kravets-levko:
@gabrieldutra Please close this issue after getredash/redash#4228 and getredash/redash#4203 merged. Thanks!

gabrieldutra:
Yeah, I'm kind of solving this issue in those PRs. Thanks for pointing this @kravets-levko! :)

PR #2320, opened 2018-02-15
Few bugfixes

kravets-levko:
1. https://discuss.redash.io/t/v4-beta-public-sharing-blank/1533
2. After generating "Invite user" link and "Reset password" link, UI showed relative link (only path fragment), and UI was inconsistent (one of them was shown in textarea and another - as link). Now it looks like this:

image
image

### DIFF
--- a/client/app/components/app-view/index.js
+++ b/client/app/components/app-view/index.js
@@ -27,9 +27,9 @@ const layouts = {
 function selectLayout(route) {
   let layout = layouts.default;
   if (route.layout) {
-    layout = layouts[route.layout];
+    layout = layouts[route.layout] || layouts.default;
   } else if (!route.authenticated) {
-    layout = layout.defaultSignedOut;
+    layout = layouts.defaultSignedOut;
   }
   return layout;
 }
--- a/client/app/services/utils.js
+++ b/client/app/services/utils.js
@@ -1,4 +1,4 @@
-function fullUrl(url) {
+function absoluteUrl(url) {
   const location = window.location;
 
   const scheme = location.protocol.toLowerCase();
@@ -9,6 +9,6 @@ function fullUrl(url) {
 
 export default function init(ngModule) {
   ngModule.factory('Utils', () => ({
-    fullUrl,
+    absoluteUrl,
   }));
 }
--- a/client/app/services/utils.js
+++ b/client/app/services/utils.js
@@ -1,14 +1,8 @@
-function absoluteUrl(url) {
-  const location = window.location;
+/* eslint-disable import/prefer-default-export */
 
-  const scheme = location.protocol.toLowerCase();
-  const host = location.host;
-
-  return scheme + '//' + host + url;
-}
-
-export default function init(ngModule) {
-  ngModule.factory('Utils', () => ({
-    absoluteUrl,
-  }));
+export function absoluteUrl(url) {
+  const urlObj = new URL(url, window.location);
+  urlObj.protocol = window.location.protocol;
+  urlObj.host = window.location.host;
+  return urlObj.toString();
 }
--- a/client/app/services/utils.js
+++ b/client/app/services/utils.js
@@ -1,18 +1,14 @@
-function fullUrl($location, url) {
-  const scheme = $location.protocol().toLowerCase();
-  const defaultPort = scheme === 'https' ? 443 : 80;
-  const host = $location.host();
-  const port = parseInt($location.port(), 10);
+function fullUrl(url) {
+  const location = window.location;
 
-  let result = scheme + '://' + host;
-  if (port !== defaultPort) {
-    result = result + ':' + port;
-  }
-  return result + url;
+  const scheme = location.protocol.toLowerCase();
+  const host = location.host;
+
+  return scheme + '//' + host + url;
 }
 
 export default function init(ngModule) {
-  ngModule.factory('Utils', $location => ({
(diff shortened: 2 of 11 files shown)


client/app/components/app-view/index.js
```diff
@@ -27,9 +27,9 @@ const layouts = {
 function selectLayout(route) {
   let layout = layouts.default;
   if (route.layout) {
-    layout = layouts[route.layout];
+    layout = layouts[route.layout] || layouts.default;
```
arikfr: 👌

client/app/components/app-view/index.js
```diff
@@ -27,9 +27,9 @@ const layouts = {
...
   } else if (!route.authenticated) {
-    layout = layout.defaultSignedOut;
+    layout = layouts.defaultSignedOut;
```
arikfr: 🤦‍♂️

client/app/services/utils.js
```diff
@@ -0,0 +1,18 @@
+function fullUrl($location, url) {
+  const scheme = $location.protocol().toLowerCase();
```
arikfr: Why not use `window.location` and drop the dependency on Angular?

redash/handlers/dashboards.py
```diff
@@ -181,7 +181,10 @@ def get(self, token):
         else:
             dashboard = self.current_user.object
 
-        return serializers.public_dashboard(dashboard)
+        return project(
+            dashboard.to_dict(with_widgets=True, user=self.current_user),
+            ('name', 'layout', 'dashboard_filters_enabled', 'updated_at', 'created_at', 'widgets')
```
arikfr: The serializer does more than that: it removes fields we don't want to expose publicly (like the query text) and includes fields we usually don't like the query result.

kravets-levko: I can agree with you, but my opinion (based on my experience) is that such code is very error-prone - when database schema or data structures (like dashboard layout in this case) changes - such code most likely will not be updated (like in in this case :-) ). So, can we implement this in some different way? say, `id` field is sensitive (BTW - is it *really* sensitive?) - maybe it's simpler just omit few fields? WDYT?

arikfr: How is it different when using `project`? The call to `project` will not be updated the same way the serializer wasn't updated. :-)

Some context on serializers: the idea is to stop putting serialization logic in the models code (the `to_dict` methods). It's not a concern of the models and it makes code harder to write, as we can't assume state we usually have when we serialize data (like `current_user`).

The mistake here (aside from not having good UI tests that could've caught this) was to duplicate widgets layout logic in the public dashboards serializer. As you suggested we need to call the non public serializer (currently the `Dashboard#to_dict` method), and omit data from the returned object (along with adding the query result). But we need to do it in the serializer.

arikfr: I realized that we suggested the same thing. The only difference is that current implementation missing the update of the widgets' properties and the location of the code (directly in handler or in serializer).

kravets-levko: I actually don't clearly understand why we need that serialized at all. Let me explain.

I see two reasons why we may need to transform data for public dashboards:

1. protect sensitive data - in this case, I think it's better not to pick "safe" fields (with `project`), but remove "unsafe" ones with `omit`. both options are error-prone; but we also should remember that front-end code expects data in some format, and when some fields are missing - it may lead to strange bugs.
2. save some traffic via removing unused fields - I think it's not a reason at all - we will save, say, 500 bytes, or even 1000 bytes (but it should be really large dashboard), but will make our code potentially buggy.

I addition, one day we're going to merge dashboard and public dashboard code, and in this case we (probably) will need the same data format for both (getredash/redash#2076).

If I'm missing something - just explain me please. I really want to make this better :-)

P.S. The only this I missed (noticed while typing this comment) - I removed code that processed hidden text boxes. It definitely should be restored.

arikfr: The main reason for omitting the fields is security concerns, because of that I actually think that `project` is the right choice. While both are error prone, it's better to have a bug than a data leak.

As for why put this in a serializer function rather inline in this handler:

* Better separation of concerns.
* To be able to reuse the code (the way we serialize visualizations/query can be reused in embed API calls for a visualization).

Hidden text boxes: I think we can drop this code. It was added by someone to allow adding `<script>` tags into the dashboard. That's not functionality that we want to keep supporting.

kravets-levko: Okay, got it. I'll amend this commit - it will be easier to restore code. But GitHub should keep our discussion.

About hidden textboxes - should we also remove the checkbox (and related JS) from UI?

kravets-levko:
@arikfr Updated this PR: restored serializers (but fixed them), removed code related to hidden text widgets (that flag is not used anywhere)

client/app/services/utils.js
```diff
@@ -0,0 +1,14 @@
...
+  const location = window.location;
+
+  const scheme = location.protocol.toLowerCase();
+  const host = location.host;
+
+  return scheme + '//' + host + url;
+}
+
+export default function init(ngModule) {
+  ngModule.factory('Utils', () => ({
```
arikfr: We don't need Angular's DI here. We can just `export` `absoluteUrl` and then where we need to use it:

```js
import { absoluteUrl } from '@/services/utils';
```

PR #2167, opened 2017-12-18
Fix for DPI-1054 error on Oracle driver

mayconbordin:
I was unable to use redash with Oracle, using cx_Oracle 5.3. I kept getting this error:
`DPI-1054: connection cannot be closed when open statements or LOBs exist`

To fix the issue I installed cx_Oracle version 6.1 and changed I few lines of code on the oracle query runner to delete the data after it has been used and close the cursor.

PR #2322, opened 2018-02-16
Fix for DPI-1054 error on Oracle driver #2167

alon710:
source https://github.com/oracle/python-cx_Oracle/issues/138


RichardLitt:
Thanks @alon710. Good catch. Are there any downsides to removing this? Is it needed in other environments?

alon710:
Not that I know 😄

arikfr:
Hi,

Thanks.

PR #2323, opened 2018-02-17
Added support for user impersonation in presto datasource

HimanshPal:
Currently while submitting a query to presto, the username is being sent is static and remains same for all users. This PR adds the feature of user impersonation which allows the username to be submitted as user's email address while submitting a query to presto. 

Adding this feature helps in taking advantage of presto grant/revoke authorization commands & as well as can be useful in limiting max concurrent queries being submitted on a per user basis.

This feature is turned off by default and can be turned on from presto's data source page. 

Attached is a screenshot

screen shot 2018-02-17 at 6 37 59 pm

### DIFF
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -53,8 +53,14 @@ class Presto(BaseQueryRunner):
                 },
                 'username': {
                     'type': 'string'
+                },
+                'impersonate_user': {
+                    'type': 'boolean',
+                    'title': 'Impersonate the currently logged on user (Submit users email address as username on presto)',
+                    'default': False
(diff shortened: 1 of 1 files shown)


HimanshPal:
@arikfr - Can you please review this PR ?
Also can confirm whether the text show on UI is fine or If it needs to changed to something else.

RichardLitt:
Thanks @HimanshPal! We'll review this when we have time. Don't worry. :)

arikfr:
Thanks, @HimanshPal for your effort here. I'm going to close this now, for context, please see the discussion in #1892.

Issue #2312, opened 2018-02-13
User shouldn't be able to change the type of the first table


arikfr:
Currently if you open the visualization editor for the main table viz, you can change its type. This should be disabled.

PR #2324, opened 2018-02-18
User shouldn't be able to change the type of the first table

kravets-levko:
Issue getredash/redash#2312

### DIFF
--- a/client/app/visualizations/edit-visualization-dialog.js
+++ b/client/app/visualizations/edit-visualization-dialog.js
@@ -19,6 +19,9 @@ const EditVisualizationDialog = {
     this.visualization = copy(this.originalVisualization);
     this.visTypes = Visualization.visualizationTypes;
 
+    // Don't allow to change type after creating visualization
+    this.canChangeType = !(this.visualization && this.visualization.id);
+
     this.newVisualization = () =>
       ({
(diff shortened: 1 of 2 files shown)


client/app/visualizations/edit-visualization-dialog.js
```diff
@@ -18,6 +18,20 @@ const EditVisualizationDialog = {
...
+    if (
+      this.visualization &&
+      isArray(this.query.visualizations) &&
+      (this.query.visualizations.length > 0)
+    ) {
+      // Forbid type change for first (min id) viz if it has type `TABLE`
+      this.canChangeType = !(
+        (this.visualization.type === 'TABLE') &&
+        (this.visualization.id === min(pluck(this.query.visualizations, 'id')))
+      );
```
arikfr: Maybe a simpler solution will be to not let the user change type after creating the visualization? I don't see much reason for a user to change type after creation and it will make it less of a guess work here :)

kravets-levko: Hm, makes sense. Will do :+1:

kravets-levko: Hi @deecay ! This code is outdated (@arikfr also didn't like it :smile: ) Now everything is much simpler :ok_hand:

PR #2325, opened 2018-02-19
Reduce restart to unless-stopped for Redis and Postgres

benmanns:
`restart: always` will cause containers to restart when the machine is rebooted or the docker process restarts. Setting this to `unless-stopped` will mean the container will restart only if the docker process/computer was rebooted without stopping the container first. Otherwise, running `docker-compose stop` and rebooting will cause the docker containers all to start again (and goodbye laptop battery!)

We might also consider `restart: on-failures` which would mean that rebooting would _not_ cause containers to start again unless requested via `docker-compose up`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,11 +33,11 @@ services:
       WORKERS_COUNT: 2
   redis:
     image: redis:3.0-alpine
-    restart: always
+    restart: unless-stopped
   postgres:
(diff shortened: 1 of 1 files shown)


arikfr:
Merged this as it's better than the current setting. But I agree that `restart: on-failures` makes more sense for a development setup -- you're welcome to implement this.

Thanks!

PR #2330, opened 2018-02-20
Sort widgets by col/row to make sure they are placed correctly

arikfr:


client/app/services/dashboard.js
```diff
@@ -2,7 +2,11 @@ import * as _ from 'underscore';
 
 function Dashboard($resource, $http, currentUser, Widget, dashboardGridOptions) {
   function prepareDashboardWidgets(widgets) {
-    return widgets.map(widget => new Widget(widget));
+    const widgetObjects = widgets.map(widget => new Widget(widget));
+    // This sorting is needed for converted dashboards, whose widgets don't have size yet. In such cases
+    // Gridster might position them wrong on the dashboard (unless sorted by col/row).
+    const sortedByCol = _.sortBy(widgetObjects, widget => widget.options.position && widget.options.position.col);
+    return _.sortBy(sortedByCol, widget => widget.options.position && widget.options.position.row * -1);
```
kravets-levko: @arikfr Did you check this fix? I mean - when I played with it, it worked when sorting was like `ORDER BY col ASC, row DESC`. This implementation does `ORDER BY col ASC; ORDER BY row DESC`, but I didn't try this variant.

arikfr: It did work, but I tested only with one case so it might have been a coincidence...

`sortBy` does a stable sort, so if we just change the order of the sorting, it will work as you wanted (see this https://codepen.io/arikfr/pen/OQZpNd?editors=1010).

kravets-levko: In my case, after second sort it should be:
```
2 / 1
2 / 0
1 / 1
1 / 0
0 / 1
0 / 0
```
that's what I tried and what worked. I think this also may work. If you're not sure - I'll check it a bit later, and we will know for sure.

arikfr: But isn't your example the result of `ORDER BY row DESC, col ASC`?

arikfr: Correction: `OREDER BY row DESC, col DESC`? (both descending)

kravets-levko: Crap. I missed that your output is in format `row / col` :smile: Yep, seems code is correct. Let's give it a try :+1:

PR #2341, opened 2018-02-24
Refactor extracting query id from query

ariarijp:
Hi, 

This code looks complex than current code.

It's because I have a plan to support query parameters on query results.
this change is prepare for that.

Here is my idea.

```sql
SELECT * FROM query_123('{"foo":123,"bar":"baz"}');
                        ~~~~~~~~~~~~~~~~~~~~~~~~~
                        Pass query parameters as JSON
```

If you don't want to support query parameters or you have more better ideas, please feel free to tell me.

Thank you
Takuya

### DIFF
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -4,6 +4,7 @@ import numbers
 import re
 import sqlite3
 
+import sqlparse
 from dateutil import parser
 
 from redash import models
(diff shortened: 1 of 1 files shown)


ariarijp:
I will send another PR when 4.0.0 released.

That will be included query parameter support.

Thank you,
Takuya

arikfr:
Thank you for the effort @ariarijp and apologies that I didn't have the chance to look into this before.

Adding support for query parameters is definitely a welcomed addition! I just wonder about the best way to implement it. To start a discussion, I shared my thoughts here: https://discuss.redash.io/t/thoughts-on-adding-support-for-queries-with-parameters-in-query-results-data-source/1709.

ariarijp:
Thanks! I'm going to check your thoughts!

PR #2342, opened 2018-02-25
Apply query format options from settings

ariarijp:
Hi, 

sqlparse has some [format options](http://sqlparse.readthedocs.io/en/latest/api/#formatting-of-sql-statements).

Query formatting rules are depends on team or some other reasons.
I think it is useful if these format options can be changed by settings.

Thank you
Takuya

### DIFF
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -223,6 +223,6 @@ ALLOW_PARAMETERS_IN_EMBEDS = parse_boolean(os.environ.get("REDASH_ALLOW_PARAMETE
 
 # sqlparse
 SQLPARSE_FORMAT_OPTIONS = {
-    'reindent': True,
-    'keyword_case': 'upper',
+    'reindent': parse_boolean(os.environ.get('SQLPARSE_FORMAT_REINDENT', 'true')),
+    'keyword_case': os.environ.get('SQLPARSE_FORMAT_KEYWORD_CASE', 'upper'),
 }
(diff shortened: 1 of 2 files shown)


ariarijp:
Hi, @bealdav

Thank you for your comment.

For now, edit settings.py directly If you want to change these formatting options.

bealdav:
Ok for now. 

Among options I think there are environment vars

ariarijp:
That makes sense.

I will try to code to change sqlparse options via env-vars on next PR.

arikfr:
@ariarijp maybe load the options as a JSON object?

Something like:

```python
SQLPARSE_FORMAT_OPTIONS = json.loads(os.environ.get('SQLPARSE_FORMAT_OPTIONS', json.dumps({
    'reindent': True,
    'keyword_case': 'upper',
}))
```

ariarijp:
@arikfr I had the same idea as you.

But I think .env file should not have JSON string.

Fir now, I prefer to override some options via SQLPARSE_FORMAT_OPTIONS_**** in .env file.

Thanks

redash/settings/__init__.py
```diff
@@ -220,3 +220,9 @@ def all_settings():
 # Allow Parameters in Embeds
 # WARNING: With this option enabled, Redash reads query parameters from the request URL (risk of SQL injection!)
 ALLOW_PARAMETERS_IN_EMBEDS = parse_boolean(os.environ.get("REDASH_ALLOW_PARAMETERS_IN_EMBEDS", "false"))
+
+# sqlparse
+SQLPARSE_FORMAT_OPTIONS = {
+    'reindent': parse_boolean(os.environ.get('SQLPARSE_FORMAT_REINDENT', 'true')),
```
ariarijp: @arikfr @bealdav 
`reindent` and `keyword_case` can be changed via env-vars.

ariarijp:
Hi team,

If you don't need its PR.
Please feel free to close it.

Thank you

arikfr:
Thanks!

PR #2343, opened 2018-02-25
Exporting query results as Google Spreadsheet

kyoshidajp:
In order to export query results to Google Spreadsheet, calling [IMPORTDATA](https://support.google.com/docs/answer/3093335?hl=en) with CSV format URL from Spreadsheet is the easiest way. However, the way is needed to make access public to allow access from Google.

This feature is export query results by using Google Spreadsheet API and add you as the owner. Also, after you export results, then open the sheet page.

## Steps to use

1. Submit **JSON Key File** which has `https://spreadsheets.google.com/feeds` and `https://www.googleapis.com/auth/drive` scope of Google Spreadsheet  at settings page
1. Run a query and get the result
1. Click **Export button** on query page

## Screenshot

**Export button on query page**

"Download" label may not be an appropriate name.

| Before | After |
| ------- | ------ |
| download_before | download_after |

**JSON key file input on settings page**

| Before | After |
| ------- | ------ |
| settings_before  | settings_after |

### DIFF
--- a/client/app/components/queries/query-results-link.js
+++ b/client/app/components/queries/query-results-link.js
@@ -22,6 +22,7 @@ function queryResultLink() {
             url = `api/query_results/${scope.queryResult.getId()}.${fileType}`;
           }
           element.attr('href', url);
+          if (fileType === 'spreadsheet') return;
           element.attr(
             'download',
             `${scope.query.name.replace(' ', '_') +
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -254,6 +254,11 @@
                   <span class="fa fa-file-excel-o"></span> Download as Excel File
                 </a>
               </li>
+              <li>
+                <a query-result-link file-type="spreadsheet" target="_blank">
+                  <span class="fa fa-file-o"></span> Export as Spreadsheet
+                </a>
+              </li>
             </ul>
           </div>
 
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -256,7 +256,7 @@
               </li>
               <li>
                 <a query-result-link file-type="spreadsheet" target="_blank">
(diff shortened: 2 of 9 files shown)


kyoshidajp:
Sorry, I'm going to fix error tests later.

kocsmy:
I like this feature, thanks @kyoshidajp 

Question: what happens if you don't have your JSON Key file and your try to hit the "Export as Spreadsheet" button? 

I think it'd be an elegant solution to show a modal with a little message that you'll need to upload your JSON Key File first and link to the Settings (or maybe support article).

The choice of word for Download/Export is indeed a bit confusing. What I'd do is keep it as Download (and we'll figure out something better for this later) but use "Export as Google Spreadsheet" instead to be clear about what's going to happen.

On the Settings page, shouldn't we use "Google Spreadsheet" (instead of Spreadsheet) to be more specific about what is this for?

denisov-vlad:
Is it possible to export to Google Spreadsheets without passing JSON key file when the user is registered via OAuth?

kyoshidajp:
@kocsmy Thanks. 

> Question: what happens if you don't have your JSON Key file and your try to hit the "Export as Spreadsheet" button?

I didn't consider that. Now, I changed that redirect to settings page with a warning message.

organization settings 2018-02-27 20-23-51

Also, I changed the label on the settings and on the query page.

new querym 2018-02-27 20-19-09

kyoshidajp:
@denisov-vlad No. Export by Service Account with API. JSON key file of Google Spreadsheet Account must be set.

kocsmy:
Thanks @kyoshidajp that all looks great now!

kyoshidajp:
I have resolved the conflict.

bealdav:
really interesting feature

arikfr:
Hi,

Thanks.

PR #2346, opened 2018-02-26
Authentication fixes (SAML and login page)

### LINKED ISSUE #2235: LDAP Auth /login_auth/login 404
### Issue Summary

With LDAP auth enabled, app routes url to '/ldap/login' based on ldap_auth.py in authentication.
Template login page redirects to '/ldap_auth/login' which returns a 404 since there is no route pointing to that url. 

### Steps to Reproduce

1. enable LDAP auth
2. navigate to localhost and click link to SSO login page
3. observe 404

Fix: adjust route in /authentication/ldap_auth.py

### Technical details:

* Redash Version: redash.3.0.0.b3134
* Browser/OS: Chrome
* How did you install Redash: ami

arikfr:
* Always show login page, even if password based login disabled.
* Make SAML work in MULTI_ORG mode.
* Generate URLs with `url_for` (should fix #2235).

### DIFF
--- a/redash/templates/login.html
+++ b/redash/templates/login.html
@@ -32,6 +32,7 @@
       <a href="{{ url_for('ldap_auth.login', next=next) }}" class="login-button btn btn-default btn-block">LDAP/SSO Login</a>
     {% endif %}
 
+    {% if show_password_login %}
     {% if show_google_openid or show_saml_login or show_remote_user_login or show_ldap_login %}
       <hr>
     {% endif %}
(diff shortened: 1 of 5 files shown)


redash/templates/login.html
```diff
@@ -21,17 +21,18 @@
     {% endif %}
 
     {% if show_saml_login %}
-      <a href="/saml/login" class="login-button btn btn-default btn-block">SAML Login</a>
+      <a href="{{ url_for('saml_auth.sp_initiated', org_slug=org_slug, next=next) }}" class="login-button btn btn-default btn-block">SAML Login</a>
```
kravets-levko: @arikfr [Here](https://github.com/getredash/redash/blob/master/redash/templates/invite.html#L24-L41) is similar code - should it also be fixed?

arikfr: Good catch! You're of course correct. I'll fix it (and also grep the code for more places).

PR #2354, opened 2018-02-28
Add a Redash extension system based on Python entrypoints.

jezdez:
This is along the lines of what Flask does for CLI plugins:

  http://flask.pocoo.org/docs/0.12/cli/#cli-plugins

The API allows specifying Python callbacks that receive the
Redash Flask app as the only argument and allow extending
the Redash process with the usual Flask API as needed. This
does not cover front-end specific extensions (yet).

This is a backport from https://github.com/mozilla/redash/pull/348
that should make our lives at Mozilla easier and maybe lead to
a more extensible Redash code base.

A suggested path forward would be investigating additional
extension points for adding front-end components/pages and
query runners (e.g. the Cassandra query runner which slows
build time down a lot).

### DIFF
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -91,7 +91,7 @@ class SlugConverter(BaseConverter):
 
 
 def create_app(load_admin=True):
-    from redash import handlers
+    from redash import extensions, handlers
     from redash.handlers.webpack import configure_webpack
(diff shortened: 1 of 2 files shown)


arikfr:
Thanks!

PR #2355, opened 2018-02-28
Use a paginator when iterating over glue tables

jezdez:
Without this you only get the first 100 tables in the glue database

Backport of https://github.com/mozilla/redash/pull/333.


arikfr:
Thanks :+1:

PR #2356, opened 2018-03-01
Dashboard fixes and improvements

kravets-levko:
- Improved `gridsterAutoHeight` directive: it will self-disable when user manually changes widget's height. As a result, auto-height feature will still enabled when user adds widget, updates other widgets or even position or width of that widget and save layout. Widget will become "locked" only if user will change its height.
- When saving dashboard adding or removing widgets, save only modified widgets (with changed size and/or position).
- Change default table viz width from 4 to 3 columns.


arikfr:
:dancer:

PR #2358, opened 2018-03-01
Fix query page long content unscrollable problem

tonyjiangh:
Before
----
- With long table, the query-metadata part and bottom-controller part disappeared(out of screen)
- It's not scrollable
image

After
-----
- I guess this is what we want?
image

### DIFF
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -118,7 +118,7 @@
     </nav>
 
     <div class="content">
-      <div class="flex-fill d-flex flex-column p-l-15 p-r-15">
+      <div class="flex-fill d-flex flex-column p-l-15 p-r-15" style="height: 0; overflow-y: hidden;">
         <div class="row editor" resizable r-directions="['bottom']" r-flex="true" resizable-toggle
           style="min-height: 11px; max-height: 460px;" ng-if="sourceMode">
           <section>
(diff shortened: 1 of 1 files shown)


kocsmy:
Oh, thanks so much for spotting this @tonyjiangh and applying a fix 👍

@kravets-levko we might want to move the inline CSS code, please review and advise.

kravets-levko:
@kocsmy We could keep it as inline CSS, it's fine in this case.

@tonyjiangh Thanks for this fix! Does it also work with smaller tables and other visualizations?

kravets-levko:
@tonyjiangh I think you should create an issue for this because in Chrome everything looks fine even without this fix, but in Firefox this fix does not work properly:

Chrome: 
image

Firefox without fix:
image

Firefox with fix (note that there is no scrollbar - it will appear at some point if I'll decrease window height. also scrollbar is cut):
image
image

I think we need more complex solution for this.

tonyjiangh:
@kravets-levko 
Hmm, that's weird because I'm actually using Firefox for development, maybe this also depends on browser version or OS.

I'll create an issue for this and check for some other platform as well.

kravets-levko:
@tonyjiangh It's my bad too - I mostly use Chrome and sometimes I forget to check markup in another browsers. And also it's strange that it does not work, because all visualizations have their own scrollbars on overflow. So, create an issue and will see what to do next. Anyway, thank you a lot that you noticed this!

tonyjiangh:
There's a more fundamental fix [here](https://github.com/getredash/redash/pull/2363), so this PR is closed.

arikfr:
Thanks again, @tonyjiangh for raising the issue and taking a stab at a solution. :+1:

PR #2262, opened 2018-01-26
Change multi-filter behavior and fix dashboard multi-filter bug

tonyjiangh:
Alternative is here: https://github.com/getredash/redash/pull/2359
I prefer the alternative solution

Before
===

* Dashboard level multi-filter acts weird

After
===
change ALL_VALUES representation
----
image
* Multi-filter shows `*` instead of rendering tags for all options
* All options except `Clear` disabled when `*` is selected
* Dashboard-level multi-filter and queryResult-level multi-filter synced
* Default value for multi-filter becomes ALL_VALUE(`*`)

Related Issues
----
https://github.com/getredash/redash/issues/1952
https://github.com/getredash/redash/issues/608

Related PR
----
https://github.com/getredash/redash/pull/2155

PR #2359, opened 2018-03-01
[Alternative] Change multi-filter behavior and fix dashboard multi-filter bug

tonyjiangh:
Alternative to this: https://github.com/getredash/redash/pull/2262
I prefer this one better.
We could keep one and close the other.

Before
===

* Dashboard level multi-filter acts weird

After
===
change Clear behavior (Thanks to darylerwin's idea)
---
image
* Multi-filter `Clear` removes all tag but shows all data
* Dashboard-level multi-filter and queryResult-level multi-filter synced
* Default value for multi-filter becomes NON_VALUE

### DIFF
--- a/client/app/components/filters.html
+++ b/client/app/components/filters.html
@@ -11,7 +11,7 @@
         </ui-select-choices>
       </ui-select>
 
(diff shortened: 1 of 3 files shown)


yoavbls:
+1 to this option

arikfr:
Hi,

Thanks.

Issue #2361, opened 2018-03-02
Query page long visualization is unscrollable


tonyjiangh:
### Issue Summary
As the title said
- With long table, the query-metadata part and bottom-controller part disappeared(out of screen)
- It's not scrollable

![image](https://user-images.githubusercontent.com/16881036/36874706-47bbaa14-1df0-11e8-8ac5-fe7768d7098d.png)

I thought this is a global problem, but it seems to be browser independent. https://github.com/getredash/redash/pull/2358#issuecomment-369855605

This does not happens for:
- Chrome Version 64.0.3282.186
- Safari Version 11.0.1

### Steps to Reproduce

1. Create a long visualization such as 25-rows table

### Technical details:

* Redash Version: v4.0.0beta (570187fc1f96d336ed63088c83efc1c56298a7cb)
* How did you install Redash: docker development env
* Browser: Firefox 58.0.2
* OS: macOS High Sierra 10.13.1

kravets-levko:
Discussion started here: getredash/redash#2358

More technical details: bug not reproduced in all Chromium-based browsers (could somebody check in Safari?).

tonyjiangh:
@kravets-levko 
I checked Safari as well. No fix needed.

As for Firefox, I couldn't reproduce the scrollbar problem you described, even if I tried another machine so I guess this is version related.

kravets-levko:
hm, really strange. I have an idea how to fix this, need to try first

kravets-levko:
Issue also reproduced on IE11 (cannot check Edge)

kravets-levko:
@tonyjiangh I created PR with my solution - getredash/redash#2363 and I kindly ask you to check it. I checked it in Chrome, Chromium, Firefox (Ubuntu 16.04) and IE11 (Win7) - seems it works.

BTW - the issue was not only with visualizations, but with some other blocks too (like schema browser).

PR #2363, opened 2018-03-02
Query page long visualization is unscrollable

kravets-levko:
Issue getredash/redash#2361

Some browsers do not take in account that blocks with `flex-grow: 1` may have `overflow: auto` blocks inside and always resize that blocks to fit their contents.

Fix: wrap content of blocks with `flex-grow: 1` with absolutely positioned block. This allows browsers to stretch them to fit all free space (technically they do not have contents), and since wrapper has strict dimensions - scrollbars will appear as expected.


tonyjiangh:
@kravets-levko Thanks a lot for the fix. It works for me.
I'll close my PR.

PR #2367, opened 2018-03-03

alon710:
Fix for Impala and Hive error
AttributeError: 'TSaslClientTransport' object has no attribute 'trans'
source https://github.com/cloudera/impyla/issues/234


RichardLitt:
Thanks, @alon710!

Issue #2317, opened 2018-02-14
Choropleth map visualization


arikfr:
![image](https://user-images.githubusercontent.com/71468/36198774-3402625a-1180-11e8-9228-cc6d9cae4e10.png)

Possible libraries:

* https://github.com/markmarkoh/datamaps
* http://jvectormap.com/
* Plotly

deecay:
- http://leafletjs.com/examples/choropleth/

SylvainMassol:
Yes leafletjs is a good library, support many formats like GeoJson, and it's already use in Redash.

deecay:
I'm fiddling around with `leaflet-choropleth` module, and found out visualization shown below can be done quite easily.
![image](https://user-images.githubusercontent.com/1698635/36972904-4dcc13a0-20b4-11e8-9c4d-0ea40675efd2.png)

However, I do not have real-life use cases, so I'm not sure what to do from here. 

1. Visualize one property from `"properties"` within geoJSON column in query result. ([geoJSON for above example](https://raw.githubusercontent.com/timwis/leaflet-choropleth/gh-pages/examples/basic/crimes_by_district.geojson)).
2. Query result column "value" has value to visualize and column geoJSON has the polygons.
3. Query result is just a table of columns "id" and "value", and this table should be merged with some other source of geoJSON.

arikfr:
The first use case we want to accomodate is to be able to visualize a dataset the looks like:

Country | Users Count
------------ | -------------
JP | 3201
US | 2912
UK | 215

I.e. take value per country and visualize it on a map.

Later we can expand this to be more granular (zoom in on specific continent, cities, etc).

This is why I feel like Leaflet might not be the try solution as it starts with too much details.

arikfr:
But it has different base maps we can use so maybe one of them is what we need?

deecay:
Yes. There are different base maps. 

The example polygons on the example page just happened to be district-size. Leaflet/geoJSON can have country level polygons too. There are several versions of country level polygons available, so I will try to find suitable one and come back with visual example.

arikfr:
Any "catalog" of those base maps?

kravets-levko:
@deecay FYI -  I'm already on this issue, and will create PR tomorrow (I see you're also experimenting with it, so I decided to tell you)

deecay:
Oh, that's nice. Can't wait to see it!

deecay:
[...]

PR #2374, opened 2018-03-06
Choropleth visualization

kravets-levko:
Issue getredash/redash#2317

Visualization is based on Leaflet since we already use it.

**Screenshots:**

Initial state:
image

With data:
image

Popups:
image

Colors:
image

Zoom:
image


kravets-levko:
@deecay Please take a look too :-)

kravets-levko:
@arikfr Currently I added country boundaries file (in GeoJSON format) - it contains boundaries in middle resolution (it's size is ~390kB and I decided not to take file with higher resolution). But I think we may need some external source. WDYT?

arikfr:
What do you mean by external source? 
How big is it gzipped?

kravets-levko:
> How big is it gzipped?
About 150kB

> What do you mean by external source?
I'm not sure if it's okay to store this file in codebase; and in future we may want to extend this viz, so we may need other geojson files. It's not a problem for now, but maybe you have some better ideas

deecay:
Hi @kravets-levko,

Awesome. I love it. I haven't tried it yet but I had one thought.

I think the tooltip and popup input would be nice for regular Map viz too.

deecay:
I think we need display area saving, just like Map viz's setBounds(). Also, I propose adding three new inputs, Initial Zoom Level, Default Latitude, and Default Longitude.

Now in Map viz, the zoom and display area depends on user adjustment by SetBound(), but I want to align the display area perfectly when I want to compare two maps in a Dashboard. So I think Default Lat/Lon approach is nice to have along with bounds approach.

client/app/visualizations/choropleth/index.js
```diff
@@ -0,0 +1,265 @@
...
+          map.fitBounds(choropleth.getBounds());
+        }
+      }
+
+      $scope.handleResize = () => {
+        resize();
+      };
+
+      $http.get(countriesDataUrl).then((response) => {
+        countriesData = response.data;
```
arikfr: How about we wrap loading `countriesData` in a module global function that returns a promise? This way it will be cached while the app is loaded.

Also it solves the issue of `render` being called before the data is ready: `render` can wait for this promise to resolve (which is immediate if it's already resolved).

arikfr:
> Now in Map viz, the zoom and display area depends on user adjustment by SetBound(), but I want to align the display area perfectly when I want to compare two maps in a Dashboard. So I think Default Lat/Lon approach is nice to have along with bounds approach.

It makes sense to have both: save the user adjusted bounds, but also reflect it in the properties if someone wants to adjust them manually.

arikfr:
(not sure it _have_ to be part of this pull request, though)

kravets-levko:
@arikfr @deecay Updated PR. Now it is possible to edit bounds:
image

If you want to compare two maps - you can select region on one of them using mouse, and then copy coordinates to second viz. Also there is useful feature from leaflet - you can zoom to arbitrary region by selecting it with mouse while holding `Shift` key (in a case you didn't know about this).

@arikfr I added cache for HTTP requests, but render will not wait for it - it's called from other places like angular watchers and I didn't want to increase code complexity - it's much simpler to check if data is available and exit if not (and it's also easier to understand how it works). Instead I did some profiling and optimized function calls a bit, so now it will do something useful only when necessary.

deecay:
@kravets-levko, I've tried the bounds and it looks good.

arikfr:
🕺

PR #2376, opened 2018-03-07
Fix docker-entrypoint broke for other name than "postgres"

valentin2105:
This commit broke our Redash installation because we use other name than "Postgres" : 

https://github.com/getredash/redash/commit/aa06e44ff8330bd4e79ef7fcd7b1390016d80a34#diff-c1111bd512b29e821b120b86446026b8

I think using the REDASH_DATABASE_URL env variable to make this check is better ? 

What do you think ?

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -23,7 +23,16 @@ server() {
 }
 
 create_db() {
-  while ! bash -c "echo > /dev/tcp/postgres/5432" &> /dev/null ; do
+  DATABASE_FULLURL=$(echo $REDASH_DATABASE_URL |cut -d '@' -f2 |sed 's,/.*,,g')
+
+  if [[ $DATABASE_FULLURL = *":"* ]]; then
+    DATABASE_URL=$(echo $DATABASE_FULLURL |cut -d ':' -f1)
+    DATABASE_PORT=$(echo $DATABASE_FULLURL |cut -d ':' -f2)
+  else 
+    DATABASE_URL=$(echo $DATABASE_FULLURL)
+    DATABASE_PORT="5432"
+  fi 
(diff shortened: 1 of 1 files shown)


RichardLitt:
Thanks, @valentin2105. I'm not entirely sure - but this might be a good question for the forum. Have you tried reaching out there, too?

valentin2105:
HI, @RichardLitt , I don't understand what I need to reach on the Forum ? 
The problem is : 

 the docker-entrypoint file use the name `postgres` to talk to the DB, hardcoded on it unless of the entiere project of redash that actually use `REDASH_DATABASE_URL`. 

`postgres` work for docker-compose installation only, in our case, we use Kubernetes.

This fix, will work with every Redash installation from compose to other stuffs.

RichardLitt:
@valentin2105 Ahh, I see. I misunderstood a bit. Great, carry on. We'll respond to this issue as soon as we are able.

arikfr:
Thanks @valentin2105! 

Considering the added complexity around this and the potential for bugs, I'm thinking of removing this check all together. What do you think?

valentin2105:
I don't see any problem to delete this check, 'cause the docker-compose integrate that : 
```
    depends_on:
      - postgres
      - redis
```
So the redash container wait for postgres, and in case of latency, it will crash and be reloaded by Docker.  
Would you I take it of in my commit ? or close the PR ?

Thanks

arikfr:
You can remove it in your PR, if you want. Thanks :)

valentin2105:
It's done :)

arikfr:
Thanks!

arikfr:
@ariarijp FYI. I think that if we want to bring back similar functionality we need to implement it in the Python code, where we already know the connection details. We can either add some general retry logic to the database connection code (maybe SQLAlchemy already has something for this that we can enable?) or add a simple `manage.py` command to check availability of database.

Issue #2375, opened 2018-03-07
Admin should be able to disable a user


arikfr:
A user object is referenced in many other objects (queries, dashboards, etc) so instead of deleting them, we will disable an inactive user. In the future we can implement full delete by implementing an ownership passing mechanism.

- [x] Admins should see a toggle for disabling a user on the user profile page
- [x] Admin can't disable his own user.
- [x] Disabled user page should indicate it's disabled, show only its name and email and a button to enable (instead of all the other components we have there now)
- [x] When loading a user we should check if it's disabled, and of course prevent login/access if it is disabled
- [x] Wherever we show the user in the UI, a disabled user should appear as disabled (some icon instead of avatar and grey color for name).
- [x] Disabled users will not appear in the users list. Admins will see a separate tab on the users list page, for disabled users.

Few thoughts on the backend implementation:

* The API should be: `POST /api/users/<id>/disable` to disable and `DELETE /api/users/<id>/disable` to enable.
* In the model we will add a `disabled_at` field, which is by default is `null`. When the user gets disabled we will set the timestamp it happened there.

arikfr:
Implemented and merged.

PR #2380, opened 2018-03-08
Admin should be able to disable a user

kravets-levko:
Issue getredash/redash#2375

Everything implemented according issue description (including notes about backend implementation).

Some screenshots:
image
image
image
image


client/app/services/user.js
```diff
@@ -16,18 +16,65 @@ function transform(data) {
   return data;
 }
 
-function User($resource, $http) {
+function enableUser(user, toastr) {
+  return user.$enable()
+    .then((data) => {
+      toastr.success(
+        `User <b>${user.name}</b> is now enabled.`,
```
arikfr: We need to sanitize `user.name` here, no?

redash/handlers/users.py
```diff
@@ -133,3 +134,23 @@ def post(self, user_id):
...
 
 
+class UserDisableResource(BaseResource):
+    @require_admin
+    def post(self, user_id):
+        user = models.User.get_by_id_and_org(user_id, self.current_org)
+        # admin cannot disable self; current user is an admin (`@require_admin`)
+        # so just check user id
+        if user.id == current_user.id:
+            abort(403)
```
arikfr: `400` feels more correct here than `403`.

client/app/pages/users/list.html
```diff
@@ -1,24 +1,47 @@
...
     <div class="col-md-12">
-      <p ng-if="$ctrl.currentUser.hasPermission('admin')">
-        <a href="users/new" class="btn btn-default"><i class="fa fa-plus"></i> New User</a>
-      </p>
+      <div class="clearfix m-b-10">
+        <a ng-if="$ctrl.currentUser.hasPermission('admin')"
+          href="users/new" class="btn btn-default pull-left"
+        ><i class="fa fa-plus"></i> New User</a>
+
+        <div class="btn-group pull-right">
```
arikfr: Only admins should be able to see disabled users (at least in the UI, it's OK it comes back in the API).

client/app/pages/users/list.js
```diff
@@ -1,16 +1,48 @@
+import { filter } from 'underscore';
 import startsWith from 'underscore.string/startsWith';
 import settingsMenu from '@/lib/settings-menu';
 import { Paginator } from '@/lib/pagination';
 import template from './list.html';
 
-function UsersCtrl(currentUser, Events, User) {
+function UsersCtrl($sce, currentUser, Events, User) {
```
arikfr: `$sce` looks unused?

kravets-levko:
@arikfr Updated PR

image

arikfr:
I still need to review the updates, but what's still missing is tests like:

- disabled user can't login
- logged in user that was disabled won't have access to API
- disabled user doesn't get forgot password emails

kravets-levko:
@arikfr Added more tests

client/app/filters/profile-image.js
```diff
@@ -0,0 +1,9 @@
+import defaultProfileImageUrl from '@/assets/images/avatar.svg';
+
+function getUserProfileImageUrl(user) {
+  return user.is_disabled ? defaultProfileImageUrl : user.profile_image_url;
+}
+
+export default function init(ngModule) {
+  ngModule.filter('profileImage', () => getUserProfileImageUrl);
```
arikfr: Instead of having a filter and then remembering to use it everywhere, why not return the right URL in the API?

redash/handlers/users.py
```diff
@@ -133,3 +136,23 @@ def post(self, user_id):
...
 
 
+class UserDisableResource(BaseResource):
+    @require_admin
+    def post(self, user_id):
+        user = models.User.get_by_id_and_org(user_id, self.current_org)
+        # admin cannot disable self; current user is an admin (`@require_admin`)
+        # so just check user id
+        if user.id == current_user.id:
+            abort(400, message="This user cannot be disabled.")
```
arikfr: Let's change the message to: `You cannot disable your own account. Please ask another admin to do this for you.`.

client/app/services/user.js
```diff
@@ -16,18 +16,67 @@ function transform(data) {
...
+      return data;
+    })
+    .catch((response) => {
+      let message = response instanceof Error ? response.message : response.statusText;
+      if (!isString(message)) {
+        message = 'Unknown error';
+      }
+      toastr.error(
+        `Cannot enable user <b>${userName}</b><br>${message}`,
+        { allowHtml: true },
```
arikfr: We have user generated content here (`userName`) and allowHtml... does it go through sanitization (by `toastr` maybe)?

kravets-levko: yes, `toastr` uses `$sanitize` somewhere inside (I checked its sources)

PR #2381, opened 2018-03-10
Add other aggregation and target value setting to counter

tonyjiangh:
Before
===

image

After
===
image

Diff
===
- Allow specifying target value directly in couonter viz edit
  - Activated when selected `Use no column`, (used to be `No target Value`)
  - Target value disabled when specifying null
- Change count rows to aggregate rows and add support for sum and average
  - Backward support rowCounter

I'm not sure what's the best plan here to handle the rowCounter setting. I chose the current plan because
- I think the parameter name should be changed to make it easier to understand.
- I also want to have a smooth migrate from the former `rowCounter` that users don't have to do any new setup


arikfr:
Closing this now as it is no longer relevant post the React migration :-( See relevant discussion here: https://discuss.redash.io/t/avoid-query-duplications-counter-and-table/4962/5?u=arikfr.

Thank you and apologies this was left out.

Issue #2386, opened 2018-03-13
[version 4.0 rc1]no login form in ldap login page 


idalin:
I add  a `show_password_login=True,` before
https://github.com/getredash/redash/blob/0410d834d149eb1416920aabbd4a76031e4fdb15/redash/authentication/ldap_auth.py#L48
to fix this.

RichardLitt:
Thanks, @idalin. This sounds like it would be a good fit for a PR. We can then review the relevant code there. Do you want to open one?

PR #2387, opened 2018-03-14
fix no login form in ldap login page #2386

idalin:


arikfr:
Good catch. Thank you !

PR #2390, opened 2018-03-14
add environment vairable REDASH_CELERY_TIMEZONE

Himenon:
I would like to add functionality to Celery 's TIMEZONE other than UTC, is it the way of this approach? Or is there a better way?

Or is it included in this PR https://github.com/getredash/redash/pull/1990?

Thanks.


arikfr:
The codebase no longer uses Celery :) But thank you for opening the Pull Request!

PR #2394, opened 2018-03-18
Test redash build security

yershalom:


RichardLitt:
Thanks @yershalom. Any reason this was closed? :)

yershalom:
Becuase the env vars are secured and its ok. So no need to run it.

PR #2405, opened 2018-03-21
Fix max_age behavior on embed graphs

inytar:
The behavior for the `max_age` parameter when doing a `POST` or `GET` with parameters on QueryResults was different. This PR makes it so that max_age of
- -1: Always return the last query result independent of age.
- 0: Execute the query.
- \> 0: Try to get latest query not older than max_age, else execute query.

I'd love some pointers on how to add tests for this.

### DIFF
--- a/redash/handlers/embed.py
+++ b/redash/handlers/embed.py
@@ -1,69 +1,14 @@
 from __future__ import absolute_import
-import logging
-import time
 
-import pystache
 from flask import request
 
 from .authentication import current_org
 from flask_login import current_user, login_required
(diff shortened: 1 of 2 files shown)


arikfr:
Thank you, but this changes `run_query_sync` which is not part of the code we provide support for or maintain. I hope that the day this code path will be removed is really near by (a month or two).

In the meantime, I prefer not to do any changes there so to avoid breaking it for those who do use it.

Thank you for taking the time in making this pull request, but unfortunately it won't be merged.

PR #2406, opened 2018-03-21
Replace dashboard engine with gridstack

kravets-levko:
`angular-gridster` has issues with handling large dashboards (20+ widgets - not so large actually...). This PR replaces `angular-gridster` with [gridstack.js](https://github.com/gridstack/gridstack.js) - pure `jQuery`-based dashboard engine, very fast and with all features we need.

This PR has a lot of code for binding `gridstack.js` with Angular code (two directives and helper class) and some helper functions - `gridstack.js` works differently from `angular-gridster` and this additional code is needed for compatibility.

Also I optimized dirty-checking in dashboard editor - now it is simpler and detects changes better.

All non-obvious code has comments added.

I don't add any screenshots - it looks and works 100% the same as with `angular-gridster`.


client/app/components/dashboards/gridstack/gridstack.js
```diff
@@ -0,0 +1,90 @@
+import $ from 'jquery';
+import _ from 'underscore';
+import 'jquery-ui/ui/widgets/draggable';
+import 'jquery-ui/ui/widgets/droppable';
+import 'jquery-ui/ui/widgets/resizable';
+import 'gridstack/dist/gridstack.css';
+
+window.$ = window.jQuery = $;
+window._ = _;
```
arikfr: Isn't this handled by the `webpack.ProvidePlugin` (which we already use for jQuery)?

kravets-levko: yes, probably accidentally missed it. but thanks for pointing

kravets-levko:
@arikfr this PR contains some "dirty" code (needed to use "private" methods for optimization reasons and to fix some bugs). in future, when `gridstack.js` will have rich API, I'll revisit it and update the code. ATM I carefully tested it, it works fine.

client/app/services/dashboard.js
```diff
@@ -1,12 +1,49 @@
...
+    .reduce((row, widgetsAtRow) => {
+      let height = 1;
+      _.each(widgetsAtRow, (widget) => {
+        height = Math.max(
+          height,
+          widget.options.position.autoHeight
+            ? defaultWidgetSizeY
+            : widget.options.position.sizeY,
+        );
+        widget.options.position.row = row;
```
arikfr: Won't `position.row` will already be `row`?

I don't see how `height` or `row + height` is being later used?

kravets-levko: This function is used to ensure that widgets are sorted by rows/columns and does not overlap - because otherwise they may be placed in unpredictable way. The idea is to compute new positions for all widgets so they will have some space between - `gridstack` will then "flow" them up and all widgets will be placed at proper positions. Algorithm:

0. we have widgets at arbitrary positions, possibly overlapping;
1. groups all widgets by rows and orders groups by row number ascending - now we have logical rows, each row has a list of widgets at that row, but rows may still overlap;
2. Assign new row number for widgets at each row: start from `0`, and for each row use previous row number plus height of highest widget of previous row (for auto-height widgets use value computed before the loop - it's height of highest widget among all, or 20 - just a large enough value). At this step we ensured that rows does not overlap. Yes, most likely here will be large space between rows, but it does not matter - `gridstack` has no limitation for grid height (in opposite to `angular-gridster`).
3. Finally sort widgets by rows (at this step we have new row numbers) ascending, and by column (also ascending) - his will ensure that widgets will be added to grid from top top bottom and from left to right - `gridstack` tries to reflow widgets after each added, so if we add, for example, bottom-most widget first, it will "flow" up, and most top-left widget will be placed below it.

arikfr: I got confused by the `_.chain` syntax, but it's clear now.

Will the user see this initial reflow? Or gridstack takes care of it before rendering?

kravets-levko: No, no initial reflow

PR #2410, opened 2018-03-22

alon710:
Fix for Impala and Hive error
source cloudera/impyla#234


arikfr:
Replaced by #3303.

PR #2414, opened 2018-03-26
Fix the create_root command to not ignore the organization parameter

mjgp2:
Previous code hardcodes the slug as 'default' rather than following the convention elsewhere that the org parameter passed in is the slug.

Should there be another parameter for the org name?


arikfr:
Seems like this is abandoned. I'm closing this now, but you're welcome to re-open after addressing the comments. Thanks!

PR #2415, opened 2018-03-27
Defensive check for retrieving words in wordcloud viz.

upman:


client/app/visualizations/word-cloud/index.js
```diff
@@ -9,7 +9,7 @@ function findWordFrequencies(data, columnName) {
   const wordsHash = {};
 
   data.forEach((row) => {
-    const wordsList = row[columnName].toString().split(' ');
+    const wordsList = typeof row[columnName] === 'string' ? row[columnName].toString().split(' ') : '';
```
arikfr: If we know it's a string, then why do we need to call `toString`?

upman: Whoops, my bad. Fixed it.

arikfr:
If you would like to implement this with the comments addressed, I suggest to wait for #2988 to land.

Closing this for now.

Issue #1743, opened 2017-04-28
problem canceling running queries


DanielBrCi:
Hello, Redash's GitHub team

I'm having a problem with my queries.
I have a series of queries stucked in my queries queue and it seems impossible to cancel them.
There's a cancel button for the queries in progress, but it doesn't do anything.

I've read in a Redash help center that you can do something called sudo (https://redash.io/help-onpremise/maintenance/ongoing-maintenance-and-basic-operations.html)

But I'm on the same level as a child when it comes to programming. I get that I can somehow access the Redash admin to foce stop the running queries, I just don't really understand how.

oyosc:
i want to know do you have solved it? i think i have the same issue;when i excute, it occurs Query in queue and don't return result, and i can't cancel it .

DanielBrCi:
@oyosc , luckly I have a friend who's an expert on these matters. I couldn't really explain what the solution, was, but basically he guided me on how to open the terminal and manually stop the queries. But i was basically copy and pasting his instructions. What I can tell you is you've got to be able to "edit" the redash page. Otherwise, the administrator must give you permission to do so before you can force shut the queries from your site

oyosc:
thanks,but the role is administrator,when i run the server and excute ,it occurs Query in queue… , i can't cancel it,i sure the data source is correct,and i can get schema,so whether you can help me solve it?
![image](https://cloud.githubusercontent.com/assets/13896491/25646890/753ffe26-2fef-11e7-94e9-12b7b0a172a4.png)

oyosc:
as you can see, it just send the request to the backend one by one,and no result return

ileibovich:
Hi, I am experiencing the same issue.

oyosc:
Celery: ./bin/run celery worker --app=redash.worker --beat -Qscheduled_queries,queries,celery -c2
you have to excute it!

ileibovich:
Run it where? On what machine? @oyosc

jolly7481:
hi! does anyone already resolved this issue? also have the same problem. i'm running redash on docker how can i clear queries queue?

killerdejavu:
`sudo supervisorctl stop redash_celery`
`sudo supervisorctl start redash_celery`
 This solved the problem for me

johnstoecker:
I had to run the above steps from @killerdejavu twice to clear the queries 👍

arikfr:
Closing this now as this is a very old issue and a lot has changed since.

PR #2416, opened 2018-03-28
Separation of Scheduler and Worker

Himenon:
This PR solves the following two problems.

When using `docker-compose`,

- https://github.com/getredash/redash/issues/1743
- Query schedule execution does not work

---

Background

When I used `docker-compose.production.yaml`, I faced a problem that no query was executed.

Apparently, the Worker seemed to have not worked well.

I simply used it as `./bin/docker-entrypoint worker` but now it is no longer scheduled to run, so I added a container for` scheduler`.

Are there other ideas to solve these problems?

thanks.

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -24,7 +24,7 @@ services:
     restart: always
   worker:
     image: redash/redash:latest
-    command: scheduler
+    command: worker
     environment:
       PYTHONUNBUFFERED: 0
       REDASH_LOG_LEVEL: "INFO"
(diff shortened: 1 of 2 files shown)


docker-compose.production.yml
```diff
@@ -33,6 +33,17 @@ services:
       QUEUES: "queries,scheduled_queries,celery"
       WORKERS_COUNT: 2
     restart: always
+  schceduler:
```
r-kells: spelling

arikfr:
I don't think we need this change in `docker-compose.yml` as this is a development setup that should be light on processes. 

As for `docker-compose.production.yml` see #3251.

Considering the above, I'm closing this, but thank you for the effort!

PR #2417, opened 2018-03-28
Query Result Datasource: Allow special characters in column names

deecay:
Currently for Query Result Datasource, if any of the source column has special character such as '(' in column name, CREATE_TABLE of the intermediate SQLite table will fail with syntax error.

This PR tries to fix that by quoting all column names with double quotes when executing CREATE_TABLE.

### DIFF
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -85,7 +85,7 @@ def create_table(connection, table_name, query_results):
                for column in query_results['columns']]
     safe_columns = [fix_column_name(column) for column in columns]
 
-    column_list = ", ".join(safe_columns)
+    column_list = ', '.join('"{0}"'.format(c) for c in safe_columns)
     create_table = u"CREATE TABLE {table_name} ({column_list})".format(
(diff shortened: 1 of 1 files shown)


arikfr:
Interesting and then things like brackets or white space is acceptable by SQLite?

The approach I was thinking about is to replace this characters, but maybe this one is better? The only downside I can think of is that it makes it harder to query (you need to qualify the table name in quotes as well).

deecay:
Yes. SQLite allows brackets and whitespaces.

Looks like any character is fine as long as its double-quoted.

>  Table names that begin with "sqlite_" are reserved for internal use. It is an error to attempt to create a table with a name that starts with "sqlite_".
https://www.sqlite.org/lang_createtable.html

---

The table which motivated me to create this PR looked like this. The data is from external source and i did not have control over column names.

Latitude (dd) | Longitude (dd) | Elevation (m)
-- | -- | --

I thought allowing was better than eliminating, in this situation. But I didn't realize the downside of quoting.

PR #2419, opened 2018-03-28
Add option to allow empty parameter value

tonyjiangh:
## The problem to tackle
The default value of parameter.
We have [this page about default value](https://help.redash.io/article/139-default) in knowledge base.
But since we currently don't allow null value, there has to be some value filled in first as an indicator of using default value.

I think using null for that is just more intuitive.

## The fix here
Add an 'Allow empty' option to parameter setting

## Functionality
Query like this
```sql
select case when '{{val}}' = '' then current_date else '{{val}}' end as value
```
Could work without preset value for parameter now

## New challenge
Empty value handling might be difficult and generate a lot of syntax error. We might want a better error message to help editor get the query right.
Also, if this is merged, I'll try to update the website for new example of usage.

### DIFF
--- a/client/app/components/parameter-settings.html
+++ b/client/app/components/parameter-settings.html
@@ -20,11 +20,19 @@
           <option value="datetime-with-seconds">Date and Time (with seconds)</option>
         </select>
       </div>
-      <div class="form-group">
-        <label>
-          <input type="checkbox" class="form-inline" ng-model="$ctrl.parameter.global">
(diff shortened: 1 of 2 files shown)


arikfr:
This needs to be updated with recent changes to parameters implementation. Also, we would like to properly support optional parameters, as described in #3011.

Considering the above, I'm closing this pull request. Thank you for you effort though!

PR #2427, opened 2018-03-30
PubNub integration for faster query results

drslump:
Ideally the client should have a websocket connection against the server to receive realtime updates about the tasks. However this is a bit hard to implement given the current server technology.

These changes take an alternative approach, relying on the third party,[PubNub](https://www.pubnub.com/) which has a very interesting free tier offering which should cover most use cases.

> FREE FOREVER up to 100 devices and 1M messages a month to get your app up and running. Basic support included.

⚠️ Since the *subscriber key* is passed on to the browser (under logged in sessions), it could be stolen and the thief could inspect all the task IDs that are being executed. I can't think of a way to directly exploit this but anyone using it should be aware of that potential information leak.

To enable the feature we need to install the `pubnub` Python package (uncomment on `requirements.txt`) and define the following environment variables:

      PUBNUB_SSL=true
      PUBNUB_PUBLISH_KEY=pub-c-xyz
      PUBNUB_SUBSCRIBE_KEY=sub-c-xyz

The current polling mechanism checks for completion every 3 seconds, so the user experience is greatly improved, specially for fast queries.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "pace-progress": "git+https://github.com/getredash/pace.git",
     "pivottable": "^2.15.0",
     "plotly.js": "1.30.1",
+    "pubnub": "^4.20.2",
     "ui-select": "^0.19.8",
     "underscore": "^1.8.3",
     "underscore.string": "^3.3.4"
--- a/client/app/services/pubnub.js
+++ b/client/app/services/pubnub.js
@@ -1,0 +1,48 @@
+import debug from 'debug';
+import PubNub from 'pubnub';
+
+const logger = debug('redash:services:PubNubSubscriber');
+
+function PubNubFactory() {
+  class PubNubSubscriber {
+    constructor(key, channel) {
+      this.callbacks = [];
+      this.channel = channel;
+
+      const pubnub = new PubNub({ subscribeKey: key });
+
+      pubnub.addListener({
+        status: (event) => {
+          logger('[' + channel + ']', event);
+        },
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -61,6 +61,7 @@
     "pace-progress": "git+https://github.com/getredash/pace.git",
     "pivottable": "^2.15.0",
     "plotly.js": "1.30.1",
+    "pubnub": "^4.20.2",
```
drslump: couldn't come up with a way to make this dependency optional

drslump:
I went with PubNub because I used it some years ago, didn't research alternatives to be honest. For the current usage pattern that 1M messages/month seem to be more than enough.

Trying to make the system secure when using a 3rd party is not worth it in my opinion. If some user fears the leak of information then she shouldn't enable the system. I thought about using the a channel per task (race conditions between subscriber and publisher) and also about encrypting the payload based on the session but in the end there is going to be some leakage.

In my opinion the service is lacking support for async IO on the server, checking the HTTP framework it uses I don't think it's worth it to hack something up with gevent/eventlet. Perhaps the most sensible solution would be to have an additional server that only handles the realtime messagin (backed by Redis). Not only for notifications but also to allow blocking HTTP calls in the API when refreshing a query. That complicates the deployment but it's becoming easier and easier to deploy servers every day 😄

drslump:
closing this one since the approach in https://github.com/getredash/redash/pull/2439 works good enough for most use cases and is much cleaner.

PR #2428, opened 2018-04-01
Improve a query failure message for Treasure Data Runner

toru-takahashi:
The current Treasure Data Query Runner doesn't show an error message if the query failed due to any errors (e.g. syntax error, etc...)

This PR improves a message to include the actual error message on the query editor by overriding [errors.InternalError](https://github.com/treasure-data/td-client-python/blob/bf8b8b550fd9e11afa88969f69c462b6f8739a5b/tdclient/cursor.py#L72)

- Current

- New

### DIFF
--- a/redash/query_runner/treasuredata.py
+++ b/redash/query_runner/treasuredata.py
@@ -8,6 +8,7 @@ logger = logging.getLogger(__name__)
 
 try:
     import tdclient
+    from tdclient import errors
     enabled = True
 
 except ImportError:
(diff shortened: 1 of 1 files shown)


redash/query_runner/treasuredata.py
```diff
@@ -102,22 +103,24 @@ def run_query(self, query, user):
...
+            if cursor.rowcount == 0:
+                rows = []
+            else:
+                rows = [dict(zip(([c[0] for c in columns_data]), r)) for i, r in enumerate(cursor.fetchall())]
+            data = {'columns': columns, 'rows': rows}
+            json_data = json.dumps(data, cls=JSONEncoder)
+            error = None
+        except errors.InternalError as e:
+            json_data = None
+            error = "%s: %s" % (e.message, cursor.show_job()['debug']['stderr'])
```
arikfr: Is `['debug']['stderr']` guaranteed to be always present? Maybe worth safe guarding with `.get()`?

toru-takahashi: Thank you for your review.
I think it's guaranteed. But, yeah, I agree with you.
I'll add .get() for it.

toru-takahashi:
Added the safeguard with .get().
Could you review it?

arikfr:
Thanks.

toru-takahashi:
Thank you!!

PR #2435, opened 2018-04-02
Docker query runner

drslump:
Docker query runner based on `docker-py`, which is a great library to interact directly with the daemon, avoiding launching a subprocess for the docker cli and offering much better control about how the containers are created.

Allows to run arbitrary code with the well defined isolation of containers.

The communication is done via standard streams using UTF-8 encoding. The
query is provided over stdin, the results are expected on stdout and any
errors should be writen to stderr.

Supported formats for providing the results:

 - Json (see https://discuss.redash.io/t/creating-a-new-query-runner-data-source-in-redash/347)
 - CSV (either with commas or tabs)

Regardless of the format a normalization pass is performed to enrich the
result with type information based on the column values. It also supports
deriving a type from the column names by using the `name:type` pattern


guidopetri:
@drslump same here, thanks for the PR! Would you mind rebasing + updating merge conflicts on this PR as well?

Same applies as the other PR, we'll close this in about a week if we don't hear back. If you're interested in reopening afterwards, we would also very much welcome that.

PR #2437, opened 2018-04-03
FIX: templating error was breaking dashboard and query editor

drslump:
I was demoing ReDash to my team and obviously it had to break 😃 

To reproduce:

  - Create a query with a parameter `{{{foo}}}`
  - Set a value for the parameter
  - Save, Publish and add to a dashboard
  - Edit the query and modify the parameter to `{{{foo}}` (**unbalanced braces**)
  - Now the dashboard is not rendered (error on console)
  - Now the query is not possible to edit (error on console)


arikfr:
What version were you demoing with?

drslump:
@arikfr 4.0rc1

arikfr:
This seems to be solved in latest master. Thanks.

PR #2438, opened 2018-04-04
Fix so that ALLOW_SCRIPTS applies to text widgets

jwestbrook:
When the env setting `REDASH_ALLOW_SCRIPTS_IN_USER_INPUT` or `ALLOW_SCRIPTS_IN_USER_INPUT` is set to True the text widgets take escaped HTML and "trust" it versus "trusting" the original text value.

This negates the actual purpose of the env setting of allowing the user to put <script> tags or other valid html tags in the text block

### DIFF
--- a/client/app/filters/markdown.js
+++ b/client/app/filters/markdown.js
@@ -9,7 +9,7 @@ export default function init(ngModule) {
 
       let html = markdown.toHTML(String(text));
       if (clientConfig.allowScriptsInUserInput) {
-        html = $sce.trustAsHtml(html);
+        html = $sce.trustAsHtml(text);
       }
 
       return html;


arikfr:
Thanks!

But your change disabled Markdown rendering, so that's not optimal either.

jwestbrook:
Noted, I'll put together a bigger fix that handles both scenarios

arikfr:
👍 thanks!

The issue was created by switching to another (more maintained/secure) markdown rendering library (named `markdown`). I would check if it has any configuration options to allow unsafe HTML.

sreynen:
There are a couple issues on preserving HTML in the markdown issue queue:

https://github.com/evilstreak/markdown-js/issues/295
https://github.com/evilstreak/markdown-js/issues/288

Both of those issues suggest the project is not maintained, and the previous maintainer seems to confirm that in https://github.com/evilstreak/markdown-js/issues/266

arikfr:
Thanks for the info @sreynen. I don't mind replacing the lib, given the new one is of similar size.

arikfr:
Hi,

Thanks.

PR #2439, opened 2018-04-04
Send a Retry-After header for running jobs with an estimate

drslump:
As commented on https://github.com/getredash/redash/pull/2427, the current polling algorithm is too simple and there is room of improvement there.

This changeset introduces a new header [Retry-After](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.37) (part of HTTP/1.1 standard although we twist it a bit). When a client performs the polling operation asking for a job, if the job is still running:

 - Calculate the average duration of that query if it was run previously
 - Check how far off it currently is from that estimation
 - Apply an easing function so that it spans the intervals accordingly
 - Return a number of seconds to wait on the `Retry-After` header

### DIFF
--- a/client/app/components/rd-timer.js
+++ b/client/app/components/rd-timer.js
@@ -8,12 +8,21 @@ function rdTimer() {
     controller($scope) {
       $scope.currentTime = '00:00:00';
 
+      const compute = () => {
+        $scope.currentTime = moment(moment() - moment($scope.timestamp)).utc().format('HH:mm:ss');
+      };
+
       // We're using setInterval directly instead of $timeout, to avoid using $apply, to
       // prevent the digest loop being run every second.
       let currentTimer = setInterval(() => {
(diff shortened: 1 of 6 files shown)


client/app/components/rd-timer.js
```diff
@@ -8,13 +8,22 @@ function rdTimer() {
...
       let currentTimer = setInterval(() => {
-        $scope.currentTime = moment(moment() - moment($scope.timestamp)).utc().format('HH:mm:ss');
+        compute();
         $scope.$digest();
       }, 1000);
 
+      // When the timestamp changes we need to recompute the time
+      $scope.$watch('timestamp', () => {
+        compute();
+      });
```
drslump: somehow related, the visual effect was that when executing a second time it was showing the timer from the previous one for a split second

drslump:
@arikfr PR rebased against latest changes.

arikfr:
Hi,

Thanks.

PR #2442, opened 2018-04-06
Add Selenium & Firefox based UI tests

hackebrot:
Hi there! 👋 

Thank you for making Redash! 📊

My team would love to have test coverage for screens that are most important to our users, so this PR aims at adding the foundation for developing automated UI tests to Redash. I checked the [trello board](https://trello.com/b/b2LUHU7A/redash-roadmap) and the GitHub issues for related proposals, but couldn't find any. Please feel free to point me to any previous conversations!

This PR adds a proof-of-concept implementation, that will hopefully provide a good basis for discussion. Here's a gist of what it does...

* install Firefox and Geckodriver into the development server image and add [pytest-selenium](https://github.com/pytest-dev/pytest-selenium) to the Python dependencies
* define a pytest fixture to create a Redash app, which handles setup and teardown code, such as app config and database clean up
* add a live server implementation using multiprocessing to run the Redash app, which is required to open pages with Selenium
* automatically add a [pytest marker](https://docs.pytest.org/en/latest/example/markers.html#marking-test-functions-and-selecting-them-for-a-run) to every test in the ``ui/`` subdirectory to easily select and deselect UI tests
* add two UI tests for the login page: one for a successful attempt and one for a failed attempt

I'd also suggest to integrate a library for working with page objects, such as [PyPOM](https://github.com/mozilla/PyPOM), to make it more convenient to interact with pages in tests. For the sake of only adding external dependencies which are strictly required, I did not PyPOM as part of this PR.

For context: I work on test automation on @jezdez's team and plan to gradually add more tests in the future, but would like to hear your thoughts on this first. Looking forward to your feedback! 🙇

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,37 @@
 FROM redash/base:latest
 
+ENV DEBIAN_FRONTEND=noninteractive \
+    MOZ_HEADLESS=1
+
+RUN dependencies=' \
+        bzip2 \
+        curl \
+        firefox \
+    ' \
+    && set -x \
+    && apt-get -qq update && apt-get -qq install --no-install-recommends -y $dependencies \
+    && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
+ENV FIREFOX_VERSION=59.0
+RUN FIREFOX_DOWNLOAD_URL=https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2 \
+    && apt-get -y purge firefox \
+    && rm -rf /opt/firefox \
+    && curl -sSL $FIREFOX_DOWNLOAD_URL -o /tmp/firefox.tar.bz2 \
+    && tar -C /opt -xjf /tmp/firefox.tar.bz2 \
+    && rm /tmp/firefox.tar.bz2 \
+    && mv /opt/firefox /opt/firefox-$FIREFOX_VERSION \
+    && ln -fs /opt/firefox-$FIREFOX_VERSION/firefox /usr/bin/firefox
+
+ENV GECKODRIVER_VERSION=0.20.0
+RUN GECKODRIVER_DOWNLOAD_URL=https://github.com/mozilla/geckodriver/releases/download/v$GECKODRIVER_VERSION/geckodriver-v$GECKODRIVER_VERSION-linux64.tar.gz \
+    && rm -rf /opt/geckodriver \
+    && curl -sSL $GECKODRIVER_DOWNLOAD_URL -o /tmp/geckodriver.tar.gz \
+    && tar -C /opt -zxf /tmp/geckodriver.tar.gz \
+    && rm /tmp/geckodriver.tar.gz \
+    && mv /opt/geckodriver /opt/geckodriver-$GECKODRIVER_VERSION \
+    && chmod 755 /opt/geckodriver-$GECKODRIVER_VERSION \
+    && ln -fs /opt/geckodriver-$GECKODRIVER_VERSION /usr/bin/geckodriver
+
 # We first copy only the requirements file, to avoid rebuilding on every file
 # change.
 COPY requirements.txt requirements_dev.txt requirements_all_ds.txt ./
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,11 @@
 FROM redash/base:latest
 
 ENV DEBIAN_FRONTEND=noninteractive \
-    MOZ_HEADLESS=1
+    MOZ_HEADLESS=1 \
+    FIREFOX_VERSION=59.0 \
+    GECKODRIVER_VERSION=0.20.0
 
+# Install requirements for UI tests
 RUN dependencies=' \
         bzip2 \
         curl \
@@ -10,27 +13,14 @@ RUN dependencies=' \
     ' \
     && set -x \
     && apt-get -qq update && apt-get -qq install --no-install-recommends -y $dependencies \
+    && apt-get -y purge firefox \
     && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
 
-ENV FIREFOX_VERSION=59.0
-RUN FIREFOX_DOWNLOAD_URL=https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2 \
-    && apt-get -y purge firefox \
-    && rm -rf /opt/firefox \
-    && curl -sSL $FIREFOX_DOWNLOAD_URL -o /tmp/firefox.tar.bz2 \
-    && tar -C /opt -xjf /tmp/firefox.tar.bz2 \
-    && rm /tmp/firefox.tar.bz2 \
-    && mv /opt/firefox /opt/firefox-$FIREFOX_VERSION \
-    && ln -fs /opt/firefox-$FIREFOX_VERSION/firefox /usr/bin/firefox
+COPY bin/install-firefox.sh /usr/local/bin/install-firefox
+COPY bin/install-geckodriver.sh /usr/local/bin/install-geckodriver
 
-ENV GECKODRIVER_VERSION=0.20.0
-RUN GECKODRIVER_DOWNLOAD_URL=https://github.com/mozilla/geckodriver/releases/download/v$GECKODRIVER_VERSION/geckodriver-v$GECKODRIVER_VERSION-linux64.tar.gz \
-    && rm -rf /opt/geckodriver \
-    && curl -sSL $GECKODRIVER_DOWNLOAD_URL -o /tmp/geckodriver.tar.gz \
-    && tar -C /opt -zxf /tmp/geckodriver.tar.gz \
-    && rm /tmp/geckodriver.tar.gz \
-    && mv /opt/geckodriver /opt/geckodriver-$GECKODRIVER_VERSION \
-    && chmod 755 /opt/geckodriver-$GECKODRIVER_VERSION \
-    && ln -fs /opt/geckodriver-$GECKODRIVER_VERSION /usr/bin/geckodriver
+RUN install-firefox
+RUN install-geckodriver
 
 # We first copy only the requirements file, to avoid rebuilding on every file
 # change.
--- a/bin/install-firefox.sh
+++ b/bin/install-firefox.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env sh
+
+set -e
+
+if [ -z "$FIREFOX_VERSION" ]; then
+    echo "Set FIREFOX_VERSION environment variable." >&2
+    exit 1
+fi
+
+# Remove existing Firefox files
+sudo rm -rf /opt/firefox
+
+FIREFOX_DOWNLOAD_URL=https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2
+curl -sSL $FIREFOX_DOWNLOAD_URL -o /tmp/firefox.tar.bz2
+
+# Extract tarball
+sudo tar -C /opt -xjf /tmp/firefox.tar.bz2
+
+# Remove bz2 archive
+rm /tmp/firefox.tar.bz2
+
+# Create symlink to Firefox bin
+sudo mv /opt/firefox /opt/firefox-$FIREFOX_VERSION
+sudo ln -fs /opt/firefox-$FIREFOX_VERSION/firefox /usr/bin/firefox
--- a/circle.yml
+++ b/circle.yml
@@ -5,6 +5,10 @@ machine:
   node:
     version:
       6.9.1
+  environment:
+    FIREFOX_VERSION: "59.0"
+    GECKODRIVER_VERSION: "0.20.0"
+    MOZ_HEADLESS: 1
 dependencies:
   override:
     - pip install --upgrade setuptools
@@ -12,6 +16,11 @@ dependencies:
     - pip install -r requirements.txt
     - npm install
     - npm run build
+    - sudo apt-get -qq update
+    - sudo apt-get -qq install --no-install-recommends -y bzip2 curl firefox
+    - sudo apt-get -y purge firefox
+    - ./bin/install-firefox.sh
+    - ./bin/install-geckodriver.sh
   cache_directories:
     - node_modules/
 test:
--- a/tests/ui/conftest.py
+++ b/tests/ui/conftest.py
@@ -1,0 +1,174 @@
+# -*- coding: utf-8 -*-
+
+"""Plugin for running UI tests."""
+
+import multiprocessing
+import time
+import socket
+
+import pytest
+
+from redash import (
+    create_app,
+    redis_connection,
+)
+from redash.models import db
+from tests.factories import Factory
+
+
+class LiveServer(object):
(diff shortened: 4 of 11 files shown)


arikfr:
Hi 👋 

Thank you for the detailed introduction and explanation.

Having proper integration/E2E tests is long overdue for a project of our size, and will have a very positive impact on our development/release velocity and quality. I have some comments on some of the implementation details, but before I dive into that, I wanted to have a higher level discussion on the tool choice.

I've been recently looking into E2E options for Redash, and been hearing very good feedback about [Cypress](https://www.cypress.io/). It looks like a very nice integrated solution for E2E/UI testing with lots of thought put into the overall development experience.

It's Chrome based at the moment, but they are working towards [cross browser support](https://github.com/cypress-io/cypress/issues/310) and [Firefox specifically](https://github.com/cypress-io/cypress/issues/1096). The linked proposal about cross browser support also details some of the limitation of WebDriver-only solutions and how Cypress is better.

Python vs. Javascript: Cypress's tests are in Javascript. It's a downside because it's not consistent with the backend and we can't reuse testing infrastructure. But it's also a benefit, as many of the frontend contributions coming from developers with Javascript only experience.

The first tests in this pull request feel very verbose, but I'm assuming this is due to lack of supporting code and tools like PyPOM? If there is a good open source project with a nice existing pytest/Selenium based test suite, I'll be happy to review.

Last time I did automated testing for the web was many years ago, using tools from Ruby's ecosystem, so I don't have much hands on experience. From just reviewing the "marketing" it looks like Cypress is the better choice. But as you have more experience, I will be happy to hear what you think (and others following the discussion as well, of course).

Once again, *thanks*. 🙏 And regardless of the tool we choose, looking forward in seeing a better test coverage for Redash. 🤩

arikfr:
I stumbled at this new Egghead course on doing E2E testing with Puppeteer and Jest: https://egghead.io/courses/end-to-end-testing-with-google-s-puppeteer-and-jest

At first I thought that it might be a viable solution for us as well, due to how simple and easy to use it looks. But Cypress seems to provide practically the same APIs for tests, but comes with the UI that gives you visibility on how the test works and where there might be issues.

I still thought it's worth sharing as someone might find this useful.

bahmutov:
As a member of Cypress team, I can say that we can help with testing this project - we are always there to answer any questions

- docs https://docs.cypress.io/guides/overview/why-cypress.html#
- chat https://gitter.im/cypress-io/cypress
- github http://github.com/cypress-io/cypress
- recipes https://github.com/cypress-io/cypress-example-recipes

jrbenny35:
@hackebrot You can use [mozdownload](https://pypi.org/project/mozdownload/) and [mozinstall](https://pypi.org/project/mozInstall/1.1/) to download and install firefox. See [this](https://github.com/mozilla/testpilot/blob/master/bin/circleci/install-test-dependencies.sh).

Also, I don't think you need to ```sudo``` your scripts. You can just change their permissions.

tests/ui/conftest.py
```diff
@@ -0,0 +1,144 @@
...
+    app.config['SERVER_NAME'] = '{host}:{port}'.format(host=host, port=port)
+
+    app_ctx = app.app_context()
+    app_ctx.push()
+
+    db.session.close()
+    db.drop_all()
+    db.create_all()
+
+    yield app
```
tarekziade: I would try.. finally this yield so the cleanup is always done no matter what happens in the test

tests/ui/conftest.py
```diff
@@ -0,0 +1,144 @@
...
+    """Return a host name as str."""
+    return 'localhost'
+
+
+@pytest.fixture(name='port')
+def fixture_port():
+    """Return a free port as int."""
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.bind(('', 0))
+    port = s.getsockname()[1]
```
tarekziade: I would add 

```python
socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```

here, so in case you create a server on the same port quickly, you don't get a
"address already in use" error

hackebrot: Thank you @tarekziade! I added this line 👍

tests/ui/conftest.py
hackebrot: pytest takes care of that

```python
import pytest

@pytest.fixture
def foobar():
    print('\n>setup')
    yield
    print('\n>teardown')

def test_foobar(foobar):
    raise RuntimeError
```

```text
$ pytest -s -v test_foobar.py
test_foobar.py::test_foobar
>setup
FAILED
>teardown
```

tests/ui/pages/login.py
```diff
@@ -0,0 +1,57 @@
...
+class LoginPage(Page):
+    """Page object model for the login."""
+
+    URL_TEMPLATE = '/{org}/login/'
+
+    @property
+    def title(self):
+        """Return the page title."""
+        self.wait.until(lambda s: self.selenium.title)
+        return self.selenium.title
```
davehunt: The wait should return the title, so you can just do:

```python
return self.wait.until(lambda s: self.selenium.title)
```

tests/ui/pages/login.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @property
+    def alert(self):
+        """Return the alert element."""
+        return self.wait.until(expected.visibility_of_element_located((
+            By.CSS_SELECTOR,
+            ".alert-danger",
+        )))
```
davehunt: Returning elements means the test author needs to be familiar with the Selenium client, which the page object model helps to abstract. Instead, you could return the text of the alert. If there are other attributes of the alert that you might want to access, then consider a `Region` with methods/properties.

tests/ui/pages/login.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @property
+    def profile_dropdown(self):
+        """Return the profile dropdown element."""
+        return self.wait.until(expected.visibility_of_element_located((
+            By.CSS_SELECTOR,
+            ".dropdown--profile__username",
+        )))
```
davehunt: As above, the page object should avoid returning web elements.

tests/ui/test_login.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+
+"""UI tests for the login page."""
+
+
+def test_login_wrong_user_credentials(login_page):
+    """Test for a failed login attempt."""
+    assert login_page.title == 'Login to Redash'
```
davehunt: You really only need this assertion in one of your tests.

hackebrot:
Hi @arikfr :wave:

Thank you for your feedback :bowing_man:

I refactored the UI login tests to use [PyPOM][pypom]. The tests itself are much more readable and less verbose now!

@jrbenny35 and I also resolved the issues around installing Firefox and Geckodriver on Circle CI. In the long run we probably want to use the same test setup for [local development][devguide] and CI to make debugging of installation errors and test failures more convenient:

``docker-compose run --rm server tests``

Cypress does looks interesting and I'll most definitely give it a try once Firefox supports lands 🦊 

[pypom]: https://github.com/mozilla/PyPOM
[devguide]: https://redash.io/help/open-source/dev-guide/docker

Dockerfile
```diff
@@ -1,5 +1,27 @@
 FROM redash/base:latest
 
+ENV DEBIAN_FRONTEND=noninteractive \
+    MOZ_HEADLESS=1 \
+    FIREFOX_VERSION=59.0 \
+    GECKODRIVER_VERSION=0.20.0
+
+# Install requirements for UI tests
+RUN dependencies=' \
```
jezdez: Any reason to do multi-line strings here instead of passing it to apt-get directly?

Dockerfile
```diff
@@ -1,5 +1,27 @@
...
+        curl \
+        firefox \
+    ' \
+    && set -x \
+    && apt-get -qq update && apt-get -qq install --no-install-recommends -y $dependencies \
+    && apt-get -y purge firefox \
+    && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
+COPY bin/install-firefox.sh /usr/local/bin/install-firefox
+COPY bin/install-geckodriver.sh /usr/local/bin/install-geckodriver
```
jezdez: To reduce the number changes to the orginal Dockerfile, let's just use the `COPY` to /app and call the installation scripts from there.

bin/install-firefox.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ -z "$FIREFOX_VERSION" ]; then
+    echo "Set FIREFOX_VERSION environment variable." >&2
+    exit 1
+fi
+
+# Remove existing Firefox files
+sudo rm -rf /opt/firefox
+
+FIREFOX_DOWNLOAD_URL=https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2
```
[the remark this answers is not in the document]
jezdez: Seems like a good idea to just make the whole URL configurable then.

circle.yml
```diff
@@ -5,13 +5,22 @@ machine:
...
 dependencies:
   override:
     - pip install --upgrade setuptools
     - pip install -r requirements_dev.txt
     - pip install -r requirements.txt
     - npm install
     - npm run build
+    - sudo apt-get -qq update
+    - sudo apt-get -qq install --no-install-recommends -y bzip2 curl firefox
+    - sudo apt-get -y purge firefox
```
jezdez: Why does this install Firefox here using apt first and then installs it again from the CDN?

tests/ui/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Avoid 'address already in use' errors
+    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+
+    s.close()
+    return port
+
+
+@pytest.fixture(name='app')
+def fixture_app(host, port):
```
jezdez: 😍 for not falling for the fixture naming trap

tests/ui/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+
+    app.config['TESTING'] = True
+    app.config['SERVER_NAME'] = '{host}:{port}'.format(host=host, port=port)
+
+    app_ctx = app.app_context()
+    app_ctx.push()
+
+    db.session.close()
+    db.drop_all()
+    db.create_all()
```
jezdez: Hmm, recreating the tables for every test that uses that fixture seems expensive. Have you done some tests with regard to change of time of the existing tests?

tests/ui/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+        email=user_email,
+        password=user_password,
+    )
+
+
+@pytest.fixture(name='login_page')
+def fixture_login_page(selenium, live_server, factory):
+    """Return a page object model for login."""
+    p = LoginPage(selenium, live_server.url, org=factory.org.slug)
+    return p.open()
```
jezdez: nit: one character variables

tests/ui/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+@pytest.fixture(name='host')
+def fixture_host():
+    """Return a host name as str."""
+    return 'localhost'
+
+
+@pytest.fixture(name='port')
+def fixture_port():
+    """Return a free port as int."""
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
jezdez: nit: one character variable

Dockerfile
hackebrot: I tend to use multi-line strings for dependency installs because it makes reviewing pull requests that add or remove dependencies easier. In this particular case, with only 3 dependencies, it doesn't make a huge difference.

circle.yml
hackebrot: As I understand, installing Firefox via apt-get creates certain folders that are required for a proper installation. Any thoughts @davehunt?

Dockerfile
hackebrot: We can speed up Docker builds by installing Firefox and Geckodriver in an extra step and making use of Docker's caching feature. Otherwise every code change will require a new Docker build which then downloads Firefox and Geckodriver.

tests/ui/conftest.py
hackebrot: https://twitter.com/hackebrot/status/981600024332259328 😁

arikfr:
@hackebrot,

> Cypress does look interesting and I'll most definitely give it a try once Firefox support lands 🦊

I'm afraid that if we don't start with it, then there is no way we will ever switch over (without significant effort).

> I refactored the UI login tests to use PyPOM. The tests itself are much more readable and less verbose now!

The tests themselves are indeed more readable, but you had to introduce 58 lines of setup code. While in some cases it will be reusable, for smaller features like the login page, that's a lot of setup...

For reference, I gave it a try and spent the last ~30 minutes setting up Cypress and writing the same login page tests:

```javascript
describe("Login Test", function() {
  it("Successful login", () => {
    cy.visit("http://localhost:5000/login");

    cy.title().should("eq", "Login to Redash");

    cy.get("#inputEmail").type("arik@redash.io");
    cy.get("#inputPassword").type("verysecret");
    cy.get("button[type='submit']").click();

    cy.title().should("eq", "Redash");
    cy.get(".dropdown--profile__username").should("contain", "Arik Fraimovich");
  });

  it("Shows message on failed login", () => {
    cy.visit("http://localhost:5000/login");
    cy.get("#inputEmail").type("arik@redash.io");
    cy.get("#inputPassword").type("badpassword");
    cy.get("button[type='submit']").click();

    cy.title().should("eq", "Login to Redash");
    cy.get(".alert-danger").should("contain", "Wrong email or password.");
  });
});
```

Note that I have no prior experience with Cypress, so most of those 30 minutes were spent on understanding their API and assertions.

And here's how the test runner shows you the test executions:
https://cl.ly/2C3g3i3E3L00

Also, it has many nice helpers in the UI, like highlighting the element you're querying:



This on top of visually being able to see what the query runner is doing (typing, clicking, etc).

---

Using Selenium with pytest has its benefits. The main one is being able to reuse the same fixtures between backend and frontend tests. But we can't neglect the fact that using Cypress will be significantly more friendly to future contributors, as it's Javascript based and has friendlier API.

I understand that you, as an organization, have a lot of experience with your stack. But I will appreciate if you give this a second thought and consideration.

I will do the same from my end.

Thanks.

hackebrot:
Hi @arikfr! :wave:

Apologies for the late response.

My goal for this project is to provide value to Redash developers as quickly as possible, by building up a suite of UI tests that will detect regressions and give immediate feedback on pull requests.

Being a OSS maintainer myself, I understand why you would prefer to use technologies that most of your contributors are familiar with. Cypress looks promising and I am open to give it a try once they support the browser that the majority of our users use.

Going back to what I want to achieve with this, I think it makes sense to use technologies that myself and folks on my team have the most experience with.  Doing so will allow us to be more productive and focus on implementing tests rather than getting acquainted with a new stack.

So here's what I came up with meet both our needs: I created a [new project][redash-ui-tests] and we will gradually automate [UI tests][tests] there. We will use Selenium, Python 3.6 and pytest and publish a Docker image with all the tests and dependencies. This will allow everyone to run these tests on CI against Redash, without you having to take on extra work of reviewing and merging pull requests for the tests.

Please feel free to create new issues for [redash-ui-tests][redash-ui-tests] and comment on existing issues. And thanks again for your work on Redash!

[redash-ui-tests]: https://github.com/mozilla/redash-ui-tests/
[tests]: https://github.com/mozilla/redash-ui-tests/issues?q=is%3Aissue+is%3Aopen+label%3Atests

arikfr:
Thank you for the update @hackebrot. I will close this pull request now and started following the new repository.

Once it's take more shape, I think we should update the developer documentation to point at the E2E tests repo.

PR #2444, opened 2018-04-06
feat: show error if saml response cannot be parsed

sjakthol:
Previously, exceptions from pysaml2 would've caused a 500 Internal Server Error response to be sent back to the user. Now, the user is redirected back to login page which includes an user friendly error message.


sjakthol:
I rebased this to get rid of some conflicts that had appeared since April. I would appreciate if we could get this merged / rejected before more conflicts arise...

arikfr:
Thanks 👍

arikfr:

PR #2452, opened 2018-04-10
Firefox: widget height continuously increases

kravets-levko:
Firefox, dashboard, auto-height feature: in some cases (loader shown and widget has params, but possibly other cases) widget starts continuously growing. Why: Firefox uses different rounding strategy for element's dimensions (`offsetHeight` is always integer (rounded up), `getBoundingClientRect` returns dimensions with a "garbage", i.e. if element has height `100.45px`, it will return something like `100.45006791123px`, but in fact will use only few fraction digits). Fix: explicitly round values before computing widget height.


arikfr:
Did you check this one on Chrome? 🤣

kravets-levko:
Yes, of course 😎

Issue #728, opened 2015-12-29
Friendlier tooltips for charts


arikfr:
![image](https://cloud.githubusercontent.com/assets/71468/12031296/046bcd24-ae13-11e5-9b5b-a3a451e4e08d.png)

* It will be better to switch to something more similar to the tooltip format we used with the HighCharts charts, where it was clearer what is the X axis and was is the Y.
* While at it, we should show dates as date only and time as date/time ("1/1/2016 vs. 1/1/2016 00:00").

alaincaltieri:
Hi Arikfr,

Linked to this issue, I think it may be really useful to have in the tooltips also the grand totals when there's a stacked data visualisation. Let me show an example:

![image](https://cloud.githubusercontent.com/assets/4777707/12265608/41e8eff2-b93e-11e5-92da-6ed2133be76b.png)

As you see the tooltip shows the amount of each group, but it doesn't show the total summed amount, which may really come in handy sometimes.

arikfr:
@alaincaltieri good point. I'll remember that once I get to this ticket. Thanks!

PR #2468, opened 2018-04-17
Better tooltips for charts; ability to change formats for numbers and dates (WIP)

kravets-levko:
Issue getredash/redash#728

**Don't merge it!** 
This PR introduces a lot of changes, I need to be sure that everything works correctly.

Some screenshots:

image
image
image
image

### DIFF
--- a/client/app/lib/value-format.js
+++ b/client/app/lib/value-format.js
@@ -2,6 +2,8 @@ import moment from 'moment/moment';
 import numeral from 'numeral';
 import _ from 'underscore';
 
+numeral.options.scalePercentBy100 = false;
+
 // eslint-disable-next-line
 const urlPattern = /(^|[\s\n]|<br\/?>)((?:https?|ftp):\/\/[\-A-Z0-9+\u0026\u2019@#\/%?=()~_|!:,.;]*[\-A-Z0-9+\u0026@#\/%=~()_|])/gi;
 
--- a/client/app/visualizations/chart/plotly/index.js
+++ b/client/app/visualizations/chart/plotly/index.js
@@ -1,10 +1,13 @@
 import { each, debounce, isArray, isObject } from 'underscore';
+import $ from 'jquery';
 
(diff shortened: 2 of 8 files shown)


client/app/lib/value-format.js
```diff
@@ -2,6 +2,8 @@ import moment from 'moment/moment';
 import numeral from 'numeral';
 import _ from 'underscore';
 
+numeral.options.scalePercentBy100 = false;
```
arikfr: What's this?

client/app/visualizations/chart/plotly/index.js
```diff
@@ -34,6 +40,10 @@ const PlotlyChart = () => ({
     let layout = {};
     let data = [];
 
+    const tooltip = $('<div>')
+      .addClass('plotly-chart-tooltip')
+      .appendTo('body');
```
arikfr: So we render our own tooltip... I can see how this can become a maintenance nightmare, chasing various Plotly edge cases 🤔 

What do we lose if we go with Plotly's builtin tooltip?

client/app/lib/value-format.js
kravets-levko: By default, for formats like `0[.]00%` `numeral` will multiply value by 100 - so `0.5` will become `50%`. This option disables this behavior so `0.5` will become `0.5%`, and `50` - `50%`.

client/app/visualizations/chart/plotly/index.js
arikfr: Past experience shows that there are always unexpected scenarios with charts :) Once we merge and deploy this, we're committed to it. But I don't think we need this distraction right now (having to maintain and fix issues related to the tooltip) -- we have more interesting stuff to do.

So let's shelf this for now, and implement something simpler:

* First step: expose the ability to set `hoverformat` for the X and Y axes.
* Second step: feature to set the `hovertext` property using data from the current point.

While it's not as complete as what you implemented here, it will serve most of the cases.

Issue #2459, opened 2018-04-15
When viewing a query (not editing), "fork" option is clickable but non-functional


ben-lantern:
Redash 4.0.0-rc.1

1) Open a query without editing its source
2) Click ... on the top-right

Expected:

Either "Fork" is not shown (since I'm not in "edit source" mode, that may be fine), or if it is shown it is functional

Actual:

"Fork" is shown but it is not functional.  When you click on it, the query is not forked. If you look in My Queries and All Queries, no copy is apparent.

RichardLitt:
Thanks, @ben-lantern! Hope this works for you.

PR #2472, opened 2018-04-19
Disable fork button for view_only user

tonyjiangh:
Fix https://github.com/getredash/redash/issues/2459

### Disabled view
image
With cursor disabled as well

### ~~Before/After~~
Permission | Before (Show Data Only) | After (Show Data Only) | Before(Edit Source) | After(Edit Source)
---|---|---|---|---
Creator/Admin | Active | Disabled | Active  | Active |
Full Access | Active | Disabled | Active | Active | 
View Only | Active(returns 403 response) | Disabled | Active(returns 403 response) | Disabled | 

Tested over Firefox 59.0.2

### DIFF
--- a/client/app/assets/less/inc/dropdown.less
+++ b/client/app/assets/less/inc/dropdown.less
@@ -11,6 +11,7 @@
         // The real magic ;)
         > a {
             pointer-events: none;
+            color: @dropdown-link-disabled-color;
         }
     }
 
@@ -29,12 +30,8 @@
 
     &:not([class*="bg-"]) {
         & > li > a {
-            &:not([disabled*="disabled"]) {
-                color: #4C4C4C;
-
-                &:hover {
-                    color: #000;
-                }
+            &:hover {
+                color: #000;
             }
         }
     }
--- a/client/app/assets/less/inc/dropdown.less
+++ b/client/app/assets/less/inc/dropdown.less
@@ -29,10 +29,12 @@
 
     &:not([class*="bg-"]) {
         & > li > a {
-            color: #4C4C4C;
+            &:not([disabled*="disabled"]) {
+                color: #4C4C4C;
 
-            &:hover {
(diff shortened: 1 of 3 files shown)


client/app/assets/less/inc/dropdown.less
```diff
@@ -29,10 +29,12 @@
 
     &:not([class*="bg-"]) {
         & > li > a {
-            color: #4C4C4C;
```
kravets-levko: @tonyjiangh I think, [this is a better place](https://github.com/getredash/redash/blob/master/client/app/assets/less/inc/dropdown.less#L9-L15) for this CSS. @kocsmy - WDYT?

arikfr:
Thanks! Why disable it (for creator/admin) in data only mode?

kravets-levko:
@arikfr As I see, @tonyjiangh didn't implement any new logic - he uses the same condition (`query.id === undefined || !canForkQuery()`), so `Fork` item will not be disabled in data only mode (as it works now).

client/app/assets/less/inc/dropdown.less
kocsmy: Yes, we have a disabled class for li already. We'll need to adjust the color a bit more gray to indicate it's disabled.

tonyjiangh:
@kravets-levko 
Thanks for the review. I'll try to fix the position for css.

@arikfr @kravets-levko 
Yes, I didn't implemented any new logic. But fork in data only mode would be disabled with this fix.
Here's why:
We're using `client/app/pages/queries/view.js` for data only view,
and `client/app/pages/queries/source-view.js` for query source view.
`$scope.canForkQuery()` is only currently implemented for the later.

It thought this was intentional. But based on the conversation above, I guess it's not. I'll fix that too.

arikfr:
Thanks!

PR #2463, opened 2018-04-15
Docker based setup scripts

pashaxp:

PR #2475, opened 2018-04-20
To avoid "cannot import name main" error

ariarijp:
Hi, 

I got below error when I ran bootstrap script for Ubuntu.

```
...snip...

  File "/usr/bin/pip", line 9, in <module>
    from pip import main
ImportError:
cannot import name main

...snip...
```

I guess this is related to https://github.com/pypa/pip/issues/5240 .

I think we should use pip 9.0.x for now.

Thank you,

### DIFF
--- a/setup/ubuntu/bootstrap.sh
+++ b/setup/ubuntu/bootstrap.sh
@@ -71,7 +71,7 @@ extract_redash_sources() {
 }
 
 install_python_packages() {
-    pip install --upgrade pip
+    pip install --upgrade pip==9.0.3
     # TODO: venv?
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks!

Following the discussion on the linked issue, there seems to be better solutions to this. But because we are going to deprecate this script soon, so I don't mind a quicker solution :-)

ariarijp:
Thank you for your comment.

Please let me know what is the plan to deprecate bootstrap scripts.
Should we run Redash on Docker containers?

arikfr:
The idea is to replace it with #2463.

PR #2482, opened 2018-04-23
add support ChatWork Alert Destination.

matsumo:
https://go.chatwork.com/


redash/destinations/chatwork.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def notify(self, alert, query, user, new_state, app, host, options):
+        try:
+            # Documentation: http://developer.chatwork.com/ja/endpoint_rooms.html#POST-rooms-room_id-messages
+            url = 'https://api.chatwork.com/v2/rooms/{room_id}/messages'.format(room_id=options.get('room_id'))
+
+            alert_url = '{host}/alerts/{alert_id}'.format(host=host, alert_id=alert.id)
+            query_url = '{host}/queries/{query_id}'.format(host=host, query_id=query.id)
+
+            message_template = options.get('message_template', ChatWork.ALERTS_DEFAULT_MESSAGE_TEMPLATE)
```
arikfr: It's better that we use some proper templating engine here to avoid possible security risk that can be introduced by using a user input with `String#format`.

We already use [`pystache`](https://github.com/defunkt/pystache), so you can easily use it.

arikfr: I've realized that we do the same thing in email destination... so will address both together.

arikfr:
Thanks!

PR #2496, opened 2018-04-27
Fixes the order of columns in the schema browser when Redshift is used as a DataSource.

akiray03:
Fixed the order of columns in the schema browser.
When Redshift is used as a data source, it appears in order of table definition.

Suppose that the following table exists:

```sql
create table yumiyama.sample_table (
  id int,
  first_name varchar(255),
  last_name varchar(255),
  created_at varchar(255),
  updated_at varchar(255)
);
```

The display of Redash's Schema Browser is changed as follows.

| before | after |
| --- | --- |
| image | image


arikfr:
Thanks!

PR #2503, opened 2018-05-01
[Athena] Fix: missing databases (apply pagination)

### LINKED ISSUE #2500: Athena query runner doesn't paginate schemas
### Issue Summary
Hi,

AWS Athena query runner paginates only tables but not schemas.
We have ~170 schemas, however, only the first 100 are shown in schema browser.
 
### Steps to Reproduce

1. Connect to Athena with Glue metastore which has more than 100 schemas
2. In the new query window, only 100 schemas are shown.

I've looked into the code and looks like you paginate only on tables but not on schemas (https://github.com/getredash/redash/blob/v4.0.0/redash/query_runner/athena.py#L130).
### Technical details:

* Redash Version: 4.0.0
* Browser/OS:

Kind regards,
Ievgen.

ialeinikov:
Closes #2500.


redash/query_runner/athena.py
```diff
@@ -125,18 +125,20 @@ def __get_schema_from_glue(self):
                 region_name=self.configuration['region']
                 )
         schema = {}
-        paginator = client.get_paginator('get_tables')
+        database_paginator = client.get_paginator('get_databases')
+        table_paginator = client.get_paginator('get_tables')
+        databases = (database for databases in database_paginator.paginate()
+                     for database in databases['DatabaseList'])
```
arikfr: something in these 2 lines doesn't look right. are you sure this is what you wanted?

ialeinikov: We want to paginate over both databases and tables.
Database paginator return list of lists, in the l#130 I flatten it inside a generator.

jezdez: Maybe it would suffice to just indent the double for-loops differently to make it more clear?

redash/query_runner/athena.py
```diff
@@ -125,18 +125,20 @@ def __get_schema_from_glue(self):
...
-                if table_name not in schema:
-                    column = [columns['Name'] for columns in table['StorageDescriptor']['Columns']]
-                    schema[table_name] = {'name': table_name, 'columns': column}
-                    for partition in table['PartitionKeys']:
-                        schema[table_name]['columns'].append(partition['Name'])
-
+        database_paginator = client.get_paginator('get_databases')
+        table_paginator = client.get_paginator('get_tables')
+        databases = (database for databases in database_paginator.paginate()
+                     for database in databases['DatabaseList'])
```
arikfr: Thank you for making the change into two for-loops. But don't we need to remove lines 130-131 now?

arikfr:
Thanks!

PR #2507, opened 2018-05-02
Log user's screen resolution

arikfr:


kyoshidajp:
@arikfr I think this `screen_resolution` will be logged to an events table, but the table doesn't have the field. Could you tell me where can I see this data?

arikfr:
It will be in the `additional_properties` column, here's an example query:

```sql
select additional_properties::json->>'screen_resolution', count(distinct user_id) as unique_users, count(0) as events_count
from events
where created_at > current_date - 7 and additional_properties::json->>'screen_resolution' <> ''
group by 1
```

kyoshidajp:
@arikfr Thanks. I agree to that.

PR #2509, opened 2018-05-02
Show custom date format on settings page

kyoshidajp:
# Issue

When I set custom date format like `YYYY/MM/DD` as `REDASH_DATE_FORMAT`, then it is not shown on settings page.

dateformat1

And if I selected another format, then never select `YYYY/MM/DD`.

dateformat3

# Changes

Add custom date format into format candidates.

dateformat2


kyoshidajp:
@arikfr Thanks. I moved it. And refactor to pass code climate error. Could you review it?

arikfr:
Thanks 👍

PR #2510, opened 2018-05-03
Display confirmation dialog when deleting a item

kyoshidajp:
# Description

Show confirm message when deleting the following items. 

- Data source
- Destination(Alert)
- Query snippet

The following items are already shown.

- Query
- Dashboard
- Widget
- Group

# Screenshot

delete_data_source
delete_destination
delete_snippet

### DIFF
--- a/client/app/pages/data-sources/show.js
+++ b/client/app/pages/data-sources/show.js
@@ -6,7 +6,7 @@ const logger = debug('redash:http');
 
 function DataSourceCtrl(
   $scope, $route, $routeParams, $http, $location, toastr,
-  currentUser, Events, DataSource,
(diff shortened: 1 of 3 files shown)


arikfr:
Thanks!

kocsmy:
@kyoshidajp thanks for this extra confirmation. I remember when you submitted this PR and I forgot to comment: it'd be a nice addition if we could say the exact name of the item you are deleting instead of "this data source/destination/snippet". 

For example: Are you sure you want to delete **Big Query (Redash)** data source?

kyoshidajp:
@kocsmy Thanks. It’s a good idea for users. I agree with you. So, can I submit that PR?

PR #2512, opened 2018-05-03
BigQuery: Replace project id with dataset in table list format

kyoshidajp:
# Issue

Using BigQuery data source, a name format of the table is `"Project ID"."Table Name"` in table list.

It is not strictly a table name in the BigQuery. So, if paste it to query as table name by clicking an arrow link on table list, then fail to run the query.

Like this.

schema1

# Changes

Change format to `"Dataset"."Table Name"`.

schema2


kyoshidajp:
Yes. Thanks.

PR #2513, opened 2018-05-04
Add location property to BigQuery data source settings

kyoshidajp:
# Feature

BigQuery began to support Tokyo region from April this year.

- [Google Cloud Platform Blog: BigQuery arrives in the Tokyo region](https://cloudplatform.googleblog.com/2018/04/BigQuery-arrives-in-the-Tokyo-region.html)
- [Dataset Locations  \|  BigQuery  \|  Google Cloud](https://cloud.google.com/bigquery/docs/dataset-locations)

This PR adds a location property to BigQuery data source settings. And calls job query with it.

If it is not specified, the value is `US` (which is default BigQuery location).

# Screenshot

bq_location

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -68,12 +68,15 @@ def _load_key(filename):
         f.close()
 
 
-def _get_query_results(jobs, project_id, job_id, start_index):
-    query_reply = jobs.getQueryResults(projectId=project_id, jobId=job_id, startIndex=start_index).execute()
+def _get_query_results(jobs, project_id, location, job_id, start_index):
+    query_reply = jobs.getQueryResults(projectId=project_id,
+                                       location=location,
+                                       jobId=job_id,
+                                       startIndex=start_index).execute()
     logging.debug('query_reply %s', query_reply)
     if not query_reply['jobComplete']:
         time.sleep(10)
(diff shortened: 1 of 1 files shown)


sekiyama58:
Thanks kyoshidajp!
@arikfr Could you merge this one?

redash/query_runner/big_query.py
```diff
@@ -148,10 +155,15 @@ def _get_bigquery_service(self):
     def _get_project_id(self):
         return self.configuration["projectId"]
 
+    def _get_location(self):
+        location = self.configuration.get("location", "")
+        return "US" if location == "" else location
```
arikfr: I think you can simplify this into:

```python
def _get_location(self):
    return self.configuration.get("location", "US")
```

No?

redash/query_runner/big_query.py
```diff
@@ -110,6 +113,10 @@ def configuration_schema(cls):
                     "type": "boolean",
                     'title': "Use Standard SQL (Beta)",
                 },
+                'location': {
+                    "type": "string",
+                    'title': 'Processing Location (default: US)',
```
arikfr: For default values we have the `default` property of the configuration item.

kyoshidajp:
@arikfr Thanks. I fixed. Could you review again?

arikfr:
Thanks!

PR #2520, opened 2018-05-05
Filter data sources in a data source input area

kyoshidajp:
# Feature

Filter Data Sources in a data source input area.

If a user has many data sources, this feature will be useful and there is almost no bad influence.

# Screenshot

| open | input `m` |
| ----- | ---------- |
| filter_ds1 | filter_ds2 |


arikfr:
Thanks!

PR #2524, opened 2018-05-09
Fix when use lower_case_table_names 0 in mysql

yamoe:


arikfr:
Can you explain a bit more about what's going on here? thanks :)

yamoe:
sorry, there is typo.

if set **lower_case_table_names** to 0 in mysql, column name is in upper case.
(0 is default value on linux mysql)
thus, there is an error in later code. 

row['table_schema'] will generate an error. (redash/redash/query_runner/mysql.py line 117)
This is because table_name is actually in upper case.

yamoe:
references about lower_case_table_names  : https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html

redash/query_runner/mysql.py
```diff
@@ -99,11 +99,11 @@ def enabled(cls):
 
     def _get_tables(self, schema):
         query = """
-        SELECT col.table_schema,
-               col.table_name,
-               col.column_name
+        SELECT col.TABLE_SCHEMA as table_schema,
+               col.TABLE_NAME as table_name,
+               col.COLUMN_NAME as column_name
```
arikfr: Isn't this breaking for anyone who didn't set `local_case_table_names` to 0?

kravets-levko:
@arikfr isn't this fix related to #4547?

arikfr:
> @arikfr isn't this fix related to #4547?

They seem to address similar things, but not sure if it's the same thing.

redash/query_runner/mysql.py
kravets-levko: I'll answer instead: no, it will not, because originally all column names in `information_schema` are upper-cased. When `lower_case_table_names == 0` - then query must use upper-cased names as well; when `lower_case_table_names != 0` - no matter what case the query use, MySQL will use CI comparison. So common denominator is to always use upper-cased names.

kravets-levko:
> They seem to address similar things, but not sure if it's the same thing.

Yeah, I see now :+1:

PR #2536, opened 2018-05-15
Chart: formats for values, labels and tooltips

kravets-levko:
Issue: getredash/redash#728

This PR is a replacement of getredash/redash#2468 and is based on discussion there.

Added tab with new settings: formats for numbers, date/time and tooltips. Tooltips now may include other fields from query records.

Screenshots:

image

image

### DIFF
--- a/client/app/visualizations/chart/index.js
+++ b/client/app/visualizations/chart/index.js
@@ -16,9 +16,10 @@ const DEFAULT_OPTIONS = {
   seriesOptions: {},
   columnMapping: {},
 
+  // showDataLabels: false, // depends on chart type
   numberFormat: '0,0[.]00000',
   percentFormat: '0[.]00%',
-  dateTimeFormat: 'DD/MM/YYYY HH:mm',
+  // dateTimeFormat: 'DD/MM/YYYY HH:mm', // will be set from clientConfig
   textFormat: '', // default: combination of {{ @@yPercent }} ({{ @@y }} ± {{ @@yError }})
 
   defaultColumns: 3,
@@ -36,7 +37,7 @@ function ChartRenderer() {
     },
     template,
     replace: false,
-    controller($scope) {
+    controller($scope, clientConfig) {
       $scope.chartSeries = [];
 
       function zIndexCompare(series) {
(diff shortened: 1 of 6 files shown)


deecay:
Hi, very nice work.

Does this change any tooltip functionality for scatter plot? I often have use cases where i want to show ID or a NAME of the row for scatter tooltip. In other words, I want to show a tooltip with column value that is not X-, Y-, Z- axis.

kravets-levko:
@deecay Yes, you can use values from other columns in format `{{ column_name }}` (there is `?` icons with short help about formats)

deecay:
That is awesome. Thank you.

2018年5月15日(火) 19:27 Levko Kravets <notifications@github.com>:

client/app/visualizations/chart/index.js
```diff
@@ -1,10 +1,32 @@
...
+  yAxis: [{ type: 'linear' }, { type: 'linear', opposite: true }],
+  xAxis: { type: 'datetime', labels: { enabled: true } },
+  error_y: { type: 'data', visible: true },
+  series: { stacking: null, error_y: { type: 'data', visible: true } },
+  seriesOptions: {},
+  columnMapping: {},
+
+  numberFormat: '0,0[.]00000',
+  percentFormat: '0[.]00%',
+  dateTimeFormat: 'DD/MM/YYYY HH:mm',
```
arikfr: The default here should be based on the value in `clientConfig`.

client/app/visualizations/chart/plotly/utils.js
```diff
@@ -69,9 +81,25 @@ function getFontColor(bgcolor) {
...
+  let result = 'text';
+  if (!hasX) result += '+x';
+  if (!hasName) result += '+name';
+  return result;
+}
+
+export function normalizeValue(value, dateTimeFormat = 'YYYY-MM-DD HH:mm:ss') {
   if (moment.isMoment(value)) {
-    return value.format('YYYY-MM-DD HH:mm:ss');
+    return value.format(dateTimeFormat);
```
arikfr: I'm not sure this is right. Plotly has its own rules on how to format date/time values on the axes (depends on the time range).

If we pass it a different format, it might not detect these values as datetime values. 

Did you try this?

arikfr: Maybe remove this setting for now and address it if we find the need in the future?

kravets-levko: I'll check if I can use it only for tooltips, if not - I'll remove it

client/app/visualizations/chart/index.js
arikfr: Ping.

kravets-levko: Sorry... Done

PR #2539, opened 2018-05-16
allow view_only permission to run query

goodwill:
I believe this is more appropriate for general use cases - running queries should equal to view_only rights, otherwise I don't really understand what is view only stands for.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ node_modules
 .tmp
 .sass-cache
 npm-debug.log
+.history
+postgresql-data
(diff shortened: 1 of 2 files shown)


michaelsolla:
Any chance do you know if this would allow users with view only permissions to the data source to see dashboard items as well?  This assuming they had already had the permission to view an existing board.

If not I will look into reporting an issue (if not already reported) where users can't view dashboard items of they only have view only on the dashboard but not the query or data source.

Thanks.

hades-6:
Is this PR getting merged in any time soon?

arikfr:
Thanks, @goodwill for your contribution. But it was implemented this way for a reason, to prevent allowing users to run any query on a data source when they don't have such permission.

We've started a project to redo permissions in Redash, which will also address this use case. To follow along you can watch #3284.

PR #2546, opened 2018-05-17
Configuration for incorporating React components

washort:
This provides the necessary parts for using React components in Angular templates.

The `footer` component is converted to React as a demonstration.


jezdez:
@arikfr Hey there, @washort has ported various other Angular components but I thought it would make sense to start small first and see if you have feedback for the general setup that we can incorporate now. In other words, I consider this to be the first step of many and would appreciate if you or someone else from the Redash team could take a look at it. Cheers!

washort:
I think the tests are failing due to the JS dependencies being different from master. Could a maintainer try CircleCI's "run without cache" option?

arikfr:
Actually I think it fails because of this: 

https://github.com/getredash/redash/blob/a6bc7c02bf64b612e47bf1e7d2550fbc057bb054/webpack.config.js#L32-L36

Looks like `ngcomponent` needs `lodash`, and `underscore` doesn't cut it anymore. I'm cool with replacing `underscore` with `lodash` (should've done it long ago). Any idea if it's a drop-in replacement?

washort:
Aha, and I didn't notice this because we haven't pulled that into our fork yet. Fixed.

arikfr:
> Looks like there may not be any dependencies on underscore other than dev dependencies and the direct uses in redash. Would you want those underscore usages converted to lodash for this PR?

It's only being used by `nomnom` which is a dependency of a dependency of a dependency of a library Plot.ly is using :) I'm not sure it affects the build, because it looks like code that isn't used in the client.

The only thing we should probably do in this pull request is remove our requirements of `underscore` and `underscore.string` from `package.json`.

arikfr:
@washort don't you need to commit an updated `package-lock.json` file?

arikfr:
@washort never mind -- I did it myself (& updated some new files to use lodash instead of underscore).

Just one last question: did you had the chance to do some performance tests? Any noticeable implication of including react in the project?

washort:
I had a look using Chrome's "Lighthouse" audit tool and load/render times look like they're no more than 5% slower.

arikfr:
https://discuss.redash.io/t/react-migration-an-experiment-with-less-di/2077

jezdez:
@arikfr ZOMG thank you!! 😍

arikfr:
@jezdez me? Let's thank @washort who does the heavy lifting here :-)

jezdez:
@arikfr That's true!

@washort:

1462065207_wxvke

PR #2549, opened 2018-05-21
Allow get_tables to see views and v10-style partitioned tables

coreyhuinker:
And it does so with just one query.

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -141,8 +141,7 @@ class PostgreSQL(BaseSQLQueryRunner):
         return schema.values()
 
     def _get_connection(self):
-    
-    connection = psycopg2.connect(user=self.configuration.get('user'),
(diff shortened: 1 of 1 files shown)


coreyhuinker:
tl;dr - yes

The catalog tables (pg_class, pg_attribute, pg_namespace) are present in all supported version of postgresql (and have probably been there since postgres started).

There have been no instances of postgres changing a `relkind` value from version to version, and it would be a major upgrade headache if they did try doing that.

While partitions don't exist prior to v10, there's no harm in running a query that would not filter out the nonexistent rows.

We may have to tweak the `IN` clause when new discoverable object types are made available in future version of postgres, but that's about it.

That, and the existing code risked duplicate entries for mviews if a future version of information_schema decided that mviews were tables.

This query should also work for the postgresql forks (Vertica, Redshift) though they may have chosen different relkind values for the object types.

arikfr:
Thank you for the pull request and the detailed explanation!

PR #2566, opened 2018-06-05
Fix #78 - Implement server side pagination and sorting for queries

### LINKED ISSUE #78: Better support for large query results
If a query has large result (~50K rows) it will make the UI to get stuck. We need to detect large results sets and handle them differently (server side pagination?).

jezdez:
This was initially reported in #78 but also applies to the queries search where the list of results can become quite large depending on the number of queries and search term used.

This was originally reported in https://github.com/mozilla/redash/issues/384 and https://github.com/mozilla/redash/issues/385 and also fixes also https://github.com/mozilla/redash/issues/323 (which was not reported here initially).

It deprecates the standalone backend query search handler and client query search resource and implements redirects for those.

I believe the live paginator should also be used in more list views and the non-live paginator deprecated or outright removed. I'm not sure how Redash is handling API backward compatibility, so I erred on the side of caution and followed a deprecation policy (that could mean removing old APIs after the next version for example).


client/app/pages/queries-list/index.js
```diff
@@ -43,16 +75,49 @@ class QueriesListCtrl {
...
 
-    this.paginator = new LivePaginator(queriesFetcher, { page });
+    this.paginator = new LivePaginator(
+      queriesFetcher,
+      {
+        page,
+        itemsPerPage: this.pageSize,
+        orderByField: pageOrder,
+        orderByReverse: pageOrderReverse,
+        params: this.parameters,
```
washort: does `this.parameters` have a value at this point? (I see where `update()` assigns to it later.)

client/app/pages/queries-list/index.js
```diff
@@ -43,16 +80,49 @@ class QueriesListCtrl {
...
+      this.searchTerm = '';
+      this.update();
+    };
+
+    this.update = () => {
+      if (!isString(this.searchTerm) || this.searchTerm.trim() === '') {
+        this.searchTerm = '';
+      }
+      this.paginator.itemsPerPage = this.pageSize;
+      this.paginator.params = this.parameters();
```
washort: any reason not to just inline the `parameters` function? I found the similar name to `params` confusing.

jezdez:
@arikfr This is the PR I talked about recently, please let me know if you want me to explain the implementation details if needed.

client/app/pages/queries-list/index.js
```diff
@@ -1,12 +1,23 @@
 import moment from 'moment';
+import { isString } from 'underscore';
+import startsWith from 'underscore.string/startsWith';
```
arikfr: You can use `String`'s own `startsWith`.

client/app/pages/queries-list/index.js
```diff
@@ -1,12 +1,23 @@
 import moment from 'moment';
+import { isString } from 'underscore';
```
arikfr: If the React pull request gets merged before this one, we need to remember to replace this with `lodash`...

redash/handlers/queries.py
```diff
@@ -18,6 +20,34 @@
...
+def order_results(results, default_order='-created_at'):
+    """
+    Orders the given results with the sort order as requested in the
+    "order" request query parameter or the given default order.
+    """
+    # See if a particular order has been requested
+    order = request.args.get('order', '').strip() or default_order
+    # and if it matches a long-form for related fields, falling
+    # back to the default order
+    selected_order = order_map.get(order, order)
```
arikfr: Maybe we should use `default_order` as the default value here as well, just to make sure that we only get predefined values for ordering?

redash/handlers/queries.py
```diff
@@ -18,6 +20,34 @@
...
+    # The query may already have an ORDER BY statement attached
+    # so we clear it here and apply the selected order
+    return sort_query(results.order_by(None), selected_order)
```
arikfr: Interesting. What would happen without this?

redash/handlers/queries.py
```diff
@@ -49,11 +80,14 @@ def get(self):
...
-                                             include_drafts=include_drafts,
-                                             limit=None)]
+        # this redirects to the new query list API that is aware of search
+        new_location = url_for(
+            'queries',
+            org_slug=current_org.slug,
+            search=term,
+            drafts='true' if include_drafts else 'false'
+        )
+        return {}, 301, {'Location': new_location}
```
arikfr: We always deploy API and client side together, so maybe we can just remove this?

redash/models.py
```diff
@@ -973,20 +974,50 @@ def create(cls, **kwargs):
...
+            # Adding outer joins to be able to order by relationship
+            .outerjoin(User, User.id == Query.user_id)
+            .outerjoin(
+                QueryResult,
+                QueryResult.id == Query.latest_query_data_id
+            )
+            .options(
+                contains_eager(Query.user),
+                contains_eager(Query.latest_query_data),
+            )
```
arikfr: Two `options` calls on purpose?

redash/handlers/queries.py
jezdez: By default, the `all_queries` class method sorts by `created_at`, which means we just reset it here to be able to apply the sorting according to the query parameter.

redash/handlers/queries.py
jezdez: Ah, that'd be nicer indeed. And you don't make any guarantees regarding the API stability?

redash/models.py
jezdez: Yeah, SQA generated different queries otherwise 😬

redash/handlers/queries.py
jezdez: Sure good idea!

client/app/pages/queries-list/index.js
jezdez: Hm, I thought it's not supported by IE11?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith#Browser_compatibility

client/app/lib/pagination/live-paginator.js
```diff
@@ -1,14 +1,34 @@
 export default class LivePaginator {
-  constructor(rowsFetcher, { page = 1, itemsPerPage = 20 } = {}) {
+  constructor(rowsFetcher, {
+    page = 1,
+    itemsPerPage = 20,
+    orderByField,
+    orderByReverse = false,
+    params = {},
```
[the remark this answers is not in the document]
jezdez: Oh, hm, maybe?

client/app/pages/queries-list/index.js
jezdez: It's a bit of a kerfuffle, I'll clean this up while porting this to https://github.com/getredash/redash/pull/2573

client/app/pages/queries-list/index.js
jezdez: Yeah, good idea.

client/app/pages/queries-list/index.js
arikfr: Thinking about IE11 is very benevolent of you :) But I would assume Babel takes care of this for us?

redash/handlers/queries.py
arikfr: But isn't SQLA smart enough to just replace the `order_by` when called again?

redash/handlers/queries.py
arikfr: Not really 🙄

I wonder if there are people who are using this API endpoint not in the application? 🤔

jezdez:
Closing in favor of #2686 which is much smaller since many of the changes here have been merged with #2573.

Issue #2567, opened 2018-06-06
Order Data Sources Alphabetically


alison985:
A Mozilla re:dash user requested that data sources be sorted alphabetically here: https://github.com/mozilla/redash/issues/367 Opening this issue to submit the PR against.

RichardLitt:
Thanks, @alison985! :)

PR #2568, opened 2018-06-06
sort datasources alphabetically

alison985:
fixes https://github.com/getredash/redash/issues/2567

Same as https://github.com/mozilla/redash/issues/367


arikfr:
Thanks .

PR #2570, opened 2018-06-06
fixes query pagination alignment

### LINKED ISSUE #2569: navigation displayed in the query source page is not centered
Details: https://github.com/mozilla/redash/issues/320

Opening issue to submit PR against.

alison985:
fixes #2569

Same as https://github.com/mozilla/redash/pull/413/ which fixes https://github.com/mozilla/redash/issues/320

### DIFF
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -209,7 +209,7 @@ edit-in-place p.editable:hover {
   }
 
   .paginator-container {
-    text-align: left;
+    text-align: center;
     margin-top: 10px;
 
     li:first-of-type {


arikfr:
I'm not sure about this one actually. Because the pagination is in the same "container" element as the table itself, if the table is wide enough, it might make the pagination not visible without scrolling.

Maybe the paginator needs its own container (which is 100% width of the parent), and then we can center it?

@kocsmy wdyt?

kocsmy:
I'm not sure either about the center position. The table can be very wide + the left alignment kinda makes sense here for me.

If you implement this, make sure it gets its own container as @arikfr was suggesting.

alison985:
This is related to #2584.

arikfr:
Thanks!

PR #2574, opened 2018-06-06
commit query result to db before evaluating alerts

mtrbean:
We have run into a problem where alert is triggering incorrectly because it is using stale query result. The race condition happens when the query result is relatively big and when the `check_alerts_for_query` task runs the result is not yet committed to the database.

Committing the db session before firing the celery tasks solves the problem for us.


RichardLitt:
Thanks for your first contribution, @mtrbean! We'll review this when we can.

mtrbean:
thanks!

arikfr:
Nice catch. Thanks!

PR #2578, opened 2018-06-10
Fix connection error when you run "create_tables"

ariarijp:
Hi @arikfr ,

Here is an another solution of #2736 .

Thank you,


arikfr:
Thanks 👍

PR #2582, opened 2018-06-11
Support elasticsearch-sql plugin for SQL query

korkeatw:
ElasticSearch version 6.3 and later supported SQL statement for querying  the data but not for the earlier versions. To use the SQL statement, you have to install [elasticsearch-sql](https://github.com/NLPchina/elasticsearch-sql) plugin.

The PR is to support SQL statement if user use ElasticSearch as data source and the plugin is already installed.

### DIFF
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -430,8 +430,8 @@ class ElasticSearch(BaseElasticSearch):
                 if sql_query.lower().startswith("select"):
                     logger.debug("Query using SQL statement")
                     index_name = get_index_from_sql(sql_query)
-                    url = "{0}/_sql".format(self.server_url)
-                    params = {"sql": sql_query}
+                    url = "{0}/_sql?sql={1}".format(self.server_url, sql_query)
+                    params = {}
 
             mapping_url = "{0}/{1}/_mapping".format(self.server_url, index_name)
             mappings, error = self._get_query_mappings(mapping_url)
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -1,9 +1,11 @@
 import logging
+import re
 import sys
 import urllib
 
 import requests
 import simplejson as json
+from simplejson import JSONDecodeError
 from requests.auth import HTTPBasicAuth
 
 from redash.query_runner import *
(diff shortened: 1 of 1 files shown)


RichardLitt:
Thanks for submitting your first pr, @korkeatw! We'll review this as soon as we have some bandwidth. :)

redash/query_runner/elasticsearch.py
```diff
@@ -402,25 +411,34 @@ def name(cls):
...
 
-            url = "{0}/{1}/_search".format(self.server_url, index_name)
-            mapping_url = "{0}/{1}/_mapping".format(self.server_url, index_name)
+            try:
+                query_dict = json.loads(query)
+                index_name = query_dict.pop("index", "")
+                result_fields = query_dict.pop("result_fields", None)
+                url = "{0}/{1}/_search".format(self.server_url, index_name)
+                params = {"source": json.dumps(query_dict), "source_content_type": "application/json"}
+            except JSONDecodeError:
```
arikfr: Relaying on `JSONDecodeError` to detect SQL queries, means that users who just have some issue with their JSON will get an unexplained error when the fallback behavior is triggered.

How about we create a dedicate query runner for ES SQL?

korkeatw: I see your point. My code may change normal behavior of the query runner. Thanks @arikfr 

If I understand correctly, you prefer creating new subclass for ES SQL right?

korkeatw: BTW. "ES with SQL plugin" and "Standard ES SQL" (version >=6.3) interfaces are different. For example:

**ES with SQL plugin**
```bash
GET /_sql?sql=SELECT * FROM table LIMIT 5
```

**Standard ES SQL**
```bash
POST /_xpack/sql?format=txt
{
    "query": "SELECT * FROM table LIMIT 5"
}
```

If we implement query runner for ES SQL plugin separately (separate into the new data source), it will be the problem because user may use both normal JSON query and SQL.

Any suggestions?

arikfr: > If I understand correctly, you prefer creating new subclass for ES SQL right?

Yes.

As for the second question, is there a simple way to check what version of ES we're querying? Or maybe, just add a checkbox configuration option to set whether we're using the plugin or not?

korkeatw: Good idea! I like the concept that we're having the checkbox to select between JSON query and SQL query. 

Will take care of this.

redash/query_runner/elasticsearch.py
```diff
@@ -43,6 +45,13 @@
 }
 
 
+def get_index_from_sql(sql):
+    matches = re.search(r"from ([\w\d]+)", sql, flags=re.IGNORECASE)
```
jean343: It does not match an index of the type `metadata_0_2018-08-29_18:40:21`, I would add `([\w\d-:]+)`

arikfr:

PR #2583, opened 2018-06-11
Query header improvements on widgets

kocsmy:
I'm improving the query header on widgets:

- Put the visualization name first, before the query name
- This saves quite a lot space + improves the visibility of the visualization name

Todo:

- [x] Make the entire header clickable (incl. visualization name)
- [x] Only render <visualization-name> if we have content, hide it otherwise (so empty - won't appear everywhere) 

screenshot 2018-06-11 17 18 21_preview


kravets-levko:
@kocsmy (cc @arikfr) BTW - we don't show visualization name if it's default for visualization type. WDYT about not to show it when it's the same as query name (see screenshot)?

image

kocsmy:
"WDYT about not to show it when it's the same as query name (see screenshot)?"
— Hmm, so my opinion is weird on this because I don't see the point of showing in that case either BUT I wouldn't implement to hide it for this edge case.

arikfr:
> BUT I wouldn't implement to hide it for this edge case.

+1 for this.

---

But I would change this implementation to have the first item (be it visualization name, if available or query name) to have the strong color. 

Future improvement can be to have a toggle for whether to show any of them at all, but that's for another pull request.

kocsmy:
This is now changed as you asked, @arikfr 

screenshot 2018-06-27 16 38 05

arikfr:
Actually my suggestion was:

> But I would change this implementation to have the first item (be it visualization name, if available or query name) to have the strong color.

But it seems that now both of them have the same weight?

kocsmy:
Oh, got it, I see.

kocsmy:
This is now implemented accordingly, @arikfr 
screenshot 2018-07-06 14 38 10

arikfr:
🙌

PR #2585, opened 2018-06-12
Show data only mode: allow to add and delete visualizations

kravets-levko:
Requested by @kocsmy 

On query page in data-only mode:
1. allow to delete visualizations (currently button is visible but does not work);
2. allow to add new visualizations (currently user should go to editing mode to add viz).

### DIFF
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -16,6 +16,17 @@ body.fixed-layout {
   }
 }
 
+.tab-nav .tab-new-vis {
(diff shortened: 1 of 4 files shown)


kocsmy:
Thanks for this, @kravets-levko — I think this will greatly improve the usability of the Query page. There was a lot of comments about people can't find how to add visualization as it was hidden until you went Edit Query mode.

I turned this tab to a button to further increase the visibility:
screenshot 2018-06-13 16 59 53

PR #2588, opened 2018-06-15
Removes hover for format button in query edit textarea

### LINKED ISSUE #2587: Format button and autocomplete checkbox should always be displayed
Copies https://github.com/mozilla/redash/issues/344

Opening an issue to submit a PR against.

alison985:
fixes #2587

### DIFF
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -175,16 +175,6 @@ edit-in-place p.editable:hover {
     margin-right: 10px;
     z-index: 999;
     position: relative;
-    opacity: 0;
-    transition: transform 0.12s ease-out;
-    transition-duration: 0.3s;
-    transition-property: opacity;
-  }
-
-  &:hover {
-    .btn__format {
-      opacity: 1;
-    }
   }
 }
 
--- a/client/app/components/dashboards/add-widget-dialog.html
+++ b/client/app/components/dashboards/add-widget-dialog.html
@@ -15,7 +15,7 @@
     </div>
     <div ng-show="$ctrl.text" class="m-t-15">
       <strong>Preview:</strong>
-      <p ng-bind-html="$ctrl.text | markdown"></p>
+      <p ng-bind-html="$ctrl.text | markdown" class="word-wrap-break"></p>
     </div>
   </div>
 
--- a/client/app/components/dashboards/add-widget-dialog.html
+++ b/client/app/components/dashboards/add-widget-dialog.html
@@ -15,7 +15,7 @@
     </div>
     <div ng-show="$ctrl.text" class="m-t-15">
       <strong>Preview:</strong>
-      <p ng-bind-html="$ctrl.text | markdown" class="word-wrap-break"></p>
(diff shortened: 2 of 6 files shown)


arikfr:
I think the original thought behind making them visible on hover was to make sure they don't block parts of the query text.

But I can see why always showing them can be beneficial in terms of feature discovery (and query lines are usually not that long).

@kocsmy wdyt?

kocsmy:
I hid this button because I don't think it's necessary to show all the time and click very often. My argument that you'd only click this a very few times during editing the query, therefore we don't need to show this, take up the space and make the interface look busy.

Regarding discoverability: the editor textarea is quite big and you move your mouse over it quite often so I'd assume people will find this feature. 

@alison985 why do you think we should have this always visible?

alison985:
I personally have no opinion either way. Our QA person @madalincm filed this request and @jezdez put it in the list of work for me to pull from. Gentlemen, could either of you weigh in please?

jezdez:
I can't speak for @madalincm but here's the original issue from downstream: https://github.com/mozilla/redash/issues/344

IIRC this was filed at the time when we merged the 4.x UI changes into our fork and it came up as a change in behavior from the previous version. I may be misremembering though, @madalincm probably knows.

Caveat: as you can see in the [issue screencast](https://user-images.githubusercontent.com/15685441/36735243-61989de8-1bde-11e8-86af-5466babb11ac.gif) we've added an additional checkbox in our fork that toggles the autocomplete feature of the query editor (since some users have reported performance issues with it for data sources with a large number of tables). But that shouldn't matter for this case.

The core of the issue is that the action button to reformat the query is not visible until hovering over the query content which [constraints the affordance of the button](https://uxdesign.cc/affordance-in-user-interface-design-3b4b0b361143) unnecessarily since the user's mouse cursor may already be or always have been hovering outside the query editor and in effect hiding the possibility of reformatting the query from the user.

In other words, if the query reformatting is a feature we want to offer to user (since it helps with maintaining query content due to the perceived quality improvement of an automated reformatting) then we shouldn't hide it behind a user input action but always show it just like the other actionable user interfaces like the `Save` or `Execute` button.

madalincm:
@jezdez described exactly what I had in mind when I logged the issue. +1 at everything he said.

kocsmy:
Thanks for providing more info for this, I'll do some research and think about this.

arikfr:
In #2644 we are discussing adding a button below the query editor (in the same line as "Save/Execute" buttons). If this finalizes at the final design, we can also place the "Format Query" button there (along with a toggle for autocomplete).

Sounds good?

kocsmy:
Yes, the format query button can be seen on the latest screenshot on #2644

jezdez:
@arikfr @kocsmy As mentioned in https://github.com/mozilla/redash/issues/344#issuecomment-403531622 I totally agree with your solution, thank you!

arikfr:
I implemented what we discussed in #2709 (moved the button next to the Add New Param button).

jezdez:
@arikfr Thank you!

Issue #2575, opened 2018-06-07
Encoding error on alert email : "Mail send error"


DependencyHell:
Hello Arik and the Redash team,

thanks for this awesome open source project!

I just want to submit a small issue

### Issue Summary

Alert is not sending email with characters such as "ô" in the name.

### Steps to Reproduce

1. Create a query
2. Create an alert on this query
3. In the name of the alert, set a character such as "ô"
4. Check the log file of the worker :

```
Mail send error.
Traceback (most recent call last):
  File "/app/redash/destinations/email.py", line 48, in notify
    subject=subject_template.format(alert_name=alert_name, state=state),
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 31: ordinal not in range(128)
```

[File "/app/redash/destinations/email.py", line 48](https://github.com/getredash/redash/blob/672c86efaa4eee7ed4295fe021431409e418de57/redash/destinations/email.py#L48)

I think it should be printed encoded into utf-8.

### Technical details:

* Redash Version: Redash 4.0.1+b4038 
* Browser/OS: Chrome/Linux
* How did you install Redash: setup using Docker

RichardLitt:
Thanks for the informative issue, @DependencyHell (nice GitHub name). This sounds like you know what the error is exactly; would you be able to make a PR for this?

arikfr:
@DependencyHell can you confirm if the change in #2605 fixes this?

Thanks!

DependencyHell:
@arikfr @RichardLitt whoops, seems I missed some notifications

The bug is now fixed, thank you!

PR #2605, opened 2018-06-20
Use unicode string for alert email subject.

arikfr:
#2575

### DIFF
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -123,7 +123,7 @@ MAIL_ASCII_ATTACHMENTS = parse_boolean(os.environ.get('REDASH_MAIL_ASCII_ATTACHM
 
 HOST = os.environ.get('REDASH_HOST', '')
 
-ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE = os.environ.get('REDASH_ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE', "({state}) {alert_name}")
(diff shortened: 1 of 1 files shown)


redash/settings/__init__.py
```diff
@@ -123,7 +123,7 @@ def all_settings():
 
 HOST = os.environ.get('REDASH_HOST', '')
 
-ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE = os.environ.get('REDASH_ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE', "({state}) {alert_name}")
+ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE = os.environ.get('REDASH_ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE', u"({state}) {alert_name}")
```
jezdez: I think you'll also want to `decode('utf-8')` the value in case `REDASH_ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE` is indeed set since it would be a bytestring on Python 2.

Speaking of which, is this the right time to ask about Python 3? 😬

arikfr: > Speaking of which, is this the right time to ask about Python 3? 😬

:-) there is actually a pull request to update the code to be compliant with Python 3. But I'm concerned on how to test everything, specially the different query runners?

arikfr:
Considering #3137 this is probably no longer needed.

PR #2606, opened 2018-06-20
requirements: added pyyaml

denisov-vlad:
New Metrika/Appmetrika datasources requires PyYAML lib. I forgot to add it to requirements.


arikfr:
Thanks!

PR #2607, opened 2018-06-20
Contributing to the docs

EmlynC:
Awesome project @arikfr! We've recently deployed this on to GKE; Kubernetes hosted on Google Cloud Platform (GCP). Pods for the redash, worker and redis (we'll probably use MemoryStore when it goes GA), CloudSQL for PostgreSQL and exposed via a Service (NodePort) so that we can use Ingress to terminate SSL at the loadbalancer, and forward 80 -> 443.

I'd be happy to provide exemplar configuration and documentation to how to do this — we've had to write this internally any way.

Would you consider pushing your documentation to a repo so that we can contribute to it? The content from https://redash.io/help/?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,1 @@
+Public docs, here perhaps?


arikfr:
Hi 👋 

The website and the documentation are already on GitHub: https://github.com/getredash/website. The documentation is here: https://github.com/getredash/website/tree/master/website/_kb.

Thank you for offering to document your experience. I'm sure it will be useful for others! You can add this under the `open-source` -> `setting-up` category (we will need to restructure the docs a bit as setting up is a single article right now).

PR #2610, opened 2018-06-20
Fix export query results output file name

gabrieldutra:
Hi guys, this is a small change on the output file name when you export a query result.

`queryName.replace(/ /g, '_')` will seek for all the occurrences of space in the file name. Not only the first one, as it was before.


arikfr:
Thanks!!

PR #2612, opened 2018-06-21
Issue 2534 Fix wrong behavior with empty query

### LINKED ISSUE #2534: When adding an empty query to a dashboard the dashboard stops rendering
1. Create a dashboard
2. Create an empty query (i.e. create a new query, write nothing in it, and save it).
3. Add above query to the dashboard.

Result: you can no longer edit this dashboard and it will no longer render its widgets.
Expected: everything works fine.

The issue seems to happen due to the code trying to load result but there is no query.

is2ei:
This will fix #2534 
It seems empty QueryResult will blocks rendering.
I changed the function to return QueryResult with status 4 "failure".
Please let me know if you have any concern! Thanks.

### DIFF
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -263,7 +263,11 @@ function QueryResource($resource, $http, $q, $location, currentUser, QueryResult
 
   Query.prototype.getQueryResult = function getQueryResult(maxAge) {
     if (!this.query) {
-      return new QueryResultError("Can't execute empty query.");
+      return new QueryResult({
+        job: {
+          status: 4,
+        },
+      });
     }
     let queryText = this.query;
 


client/app/services/query.js
```diff
@@ -263,7 +263,11 @@ function QueryResource($resource, $http, $q, $location, currentUser, QueryResult
 
   Query.prototype.getQueryResult = function getQueryResult(maxAge) {
     if (!this.query) {
-      return new QueryResultError("Can't execute empty query.");
+      return new QueryResult({
```
kravets-levko: @arikfr hm, interesting solution. WDYT about this?

arikfr:
Your solution works, but @kravets-levko's change in #2638 makes `QueryResultError` work for this case as well, which in the long term is a bit better.

Thank you, @is2ei for taking a stab at this. I will close this now, but your contribution is appreciated and we hope to see others in the future :)

PR #2616, opened 2018-06-21
Pie chart: sectors colors

kravets-levko:
Allow to choose colors for Pie chart sectors.

1. Add new tab for color settings: for Pie charts it shows list of values, for other types - list of series:
image

2. Hide some options (left/right axis, series type) for Pie chart on Series tab as that options are not applicable to Pie charts:
image
image


kocsmy:
@kravets-levko maybe should we show something like "Not applicable for Pie Charts" text if they are not relevant?

kravets-levko:
@kocsmy this text is too long to show it instead of hidden controls :-) but we can disable that inputs and add `title="Not applicable for Pie Charts"` so user will see it on mouseover

arikfr:
🍾

PR #2619, opened 2018-06-22
data source sort case insensitive

alison985:
#2567 
In QA on the Mozilla side Madalin noticed that the sort wasn't case insensitive. This fixes that.


jezdez:
LGTM+

arikfr:
Thanks!

PR #2632, opened 2018-06-28
react rules for eslint

washort:
This switches to the full airbnb eslint rule set, with a few exceptions. This enables a lot of rules and it's not clear which ones are best for redash. In particular I am not sure if all of the jsx-a11y rules should be turned on. These are the rules I've been using so far in https://github.com/mozilla/redash/tree/react .


arikfr:
Any idea is Airbnb's rules are still the community's standard or there is something else that took over?

washort:
Here's the configuration a recent Mozilla project is using; airbnb with some exceptions: https://github.com/iodide-project/iodide/blob/master/.eslintrc.js

arikfr:
OK, let's roll with this and revisit if we feel it doesn't serve us well.

arikfr:
Thanks!

PR #2634, opened 2018-06-28
React alert-subscriptions component

washort:
This was one of the first major pieces I converted, since it was somewhat standalone and needed some complex behavior. Good example of how to use customized selection UI and handle Angular DI.


arikfr:
Thanks ! Can you add a screenshot, so we can see how react-select looks like?

washort:
2018-06-28-014730_1743x713_scrot

washort:
(Note that the eslint errors here are because it's not using the React eslint support in https://github.com/getredash/redash/pull/2632 )

client/app/react-components/DestinationOption.js
```diff
@@ -0,0 +1,56 @@
...
+      >
+        <Destination destination={this.props.option.destination} />
+      </div>
+    );
+  }
+}
+
+DestinationOption.propTypes = {
+  className: PropTypes.string.isRequired,
+  // isDisabled: PropTypes.bool.isRequired,
```
openjck: Was this left over from testing?

client/app/components/alerts/alert-subscriptions.js
```diff
@@ -0,0 +1,161 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { react2angular } from 'react2angular';
+import toastr from 'angular-toastr';
+import Select from 'react-select';
```
arikfr: Need to use [Ant's Select component](https://ant.design/components/select/).

client/app/components/alerts/alert-subscriptions.js
```diff
@@ -0,0 +1,161 @@
+import React from 'react';
```
arikfr: Filename should be updated (`AlertSubscriptions.jsx`).

washort:
This now uses Ant and is up to date with master.

client/app/components/alerts/AlertSubscriptions.jsx
```diff
@@ -0,0 +1,154 @@
...
+  saveSubscriber = () => {
+    const sub = new this.props.AlertSubscription({ alert_id: this.props.alertId });
+    if (this.state.selection) {
+      sub.destination_id = this.state.selection;
+    }
+
+    sub.$save(
+      () => {
+        this.props.toastr.success('Subscribed.');
+        const destinations = this.state.destinations.filter(d => d.id !== this.state.selection);
```
spasovski: good use of `array.filter()`. Hopefully in time we can even get rid of lodash.

client/app/components/alerts/AlertSubscriptions.jsx
```diff
@@ -0,0 +1,154 @@
...
+        this.setState({
+          subscribers: this.state.subscribers.concat(sub),
```
spasovski: can also do `subscribers: [this.state.subscribers, ...sub]` as an alternative

client/app/components/alerts/AlertSubscriptions.jsx
```diff
@@ -0,0 +1,154 @@
...
+        });
+      },
+      () => {
+        this.props.toastr.error('Failed saving subscription.');
+      },
+    );
+  }
+
+  unsubscribe = (subscriber) => {
+    const destination = subscriber.destination;
```
spasovski: `const {destination, user} = subscriber` would also work

client/app/components/alerts/AlertSubscriptions.jsx
```diff
@@ -0,0 +1,154 @@
...
+      () => {
+        this.props.toastr.success('Unsubscribed');
+        let destinations;
+        const subscribers = without(this.state.subscribers, subscriber);
+        if (destination) {
+          destinations = this.state.destinations.concat(destination);
+        } else if (user.id === this.props.currentUser.id) {
+          destinations = this.state.destinations.concat({
+            id: null,
+            name: this.props.currentUser.name + ' (Email)',
```
spasovski: could use string interpolation here though you wouldn't save any bytes, i'd do the same for simple string concats

client/app/components/alerts/AlertSubscriptions.jsx
```diff
@@ -0,0 +1,154 @@
...
+
+  render() {
+    return (
+      <div className="p-5">
+        <h4>Notifications</h4>
+        <div>
+          <Select
+            value={this.state.selection || null}
+            disabled={this.state.destinations.length === 0}
+            onChange={this.updateSelection}
```
spasovski: I am confused why this would work. `onChange` should pass the event as its default argument so in the function above we'd be setting selection to the event object.

client/app/components/alerts/AlertSubscriptions.jsx
```diff
@@ -0,0 +1,154 @@
...
+          >
+            {this.state.destinations.map(d => <Select.Option value={d.id} key={d.name}><Destination destination={d} /></Select.Option>)}
+          </Select>
+        </div>
+        <div className="m-t-5">
+          <button
+            className="btn btn-default"
+            onClick={this.saveSubscriber}
+            disabled={this.state.destinations.length === 0}
+            style={{ width: '50%' }}
```
spasovski: Nit: this is probably inline since it was that way in angular but I still advise against it.

client/app/components/alerts/AlertSubscriptions.jsx
washort: Turns out Ant's `Select` component passes the selected value instead of an event.

spasovski: Sounds good...looked odd at a glance.

arikfr:
Replaced by #4081.

PR #2636, opened 2018-06-28
React ace editor component

washort:
Uses https://github.com/securingsincity/react-ace to wrap the editor as a React component.


package.json
```diff
@@ -69,6 +69,7 @@
     "plotly.js": "1.30.1",
     "prop-types": "^15.6.1",
     "react": "^16.3.2",
+    "react-ace": "^6.1.0",
```
arikfr: We should probably remove `brace` from the dependencies?

client/app/react-components/QueryEditor.js
```diff
@@ -0,0 +1,205 @@
...
+    } else if (nextProps.schema !== prevState.schema) {
+      return {
+        ...prevState,
+        schema: nextProps.schema,
+        keywords: buildKeywordsFromSchema(nextProps.schema),
+        autocompleteQuery: (nextProps.schema.reduce((totalLength, table) =>
+          totalLength + table.columns.length, 0) <= 5000),
+      };
+    }
+    return prevState;
```
openjck: `getDerivedStateFromProps` should return `null` if state doesn't need to be updated.

package.json
washort: done

client/app/react-components/QueryEditor.js
openjck: Thanks for that link. That's some really interesting reading.

package.json
washort: oops, we do import `brace` and use it directly, so leaving it in dependencies.

washort:
Updated to catch up with changes in 5.0 and fixed several bugs and style nits.

openjck:
> I know we won't be merging this until after the 5.0 release, but if you have time to look this over I'd appreciate it so I can have it ready to go as soon as possible.

I'm not sure who that's addressed to, but I'd be happy to re-review if it would help.

client/app/react-components/QueryEditor.js
```diff
@@ -0,0 +1,229 @@
...
+
+import AceEditor from 'react-ace';
+import ace from 'brace';
+import toastr from 'angular-toastr';
+
+import 'brace/ext/language_tools';
+import 'brace/mode/json';
+import 'brace/mode/python';
+import 'brace/mode/sql';
+import 'brace/theme/textmate';
```
arikfr: Need to import `'brace/ext/searchbox'` too to enable Ace's search function.

client/app/react-components/QueryEditor.js
```diff
@@ -0,0 +1,229 @@
...
+            break;
+          }
+          case 'paste': {
+            const [text] = args;
+            editor.session.doc.replace(editor.selection.getRange(), text);
+            const range = editor.selection.getRange();
+            this.props.updateQuery(editor.session.getValue());
+            window.setTimeout(() => {
+              editor.selection.setRange(range);
+            });
```
arikfr: If I'm not mistaken, [we're using $timeout in the Angular component](https://github.com/getredash/redash/blob/master/client/app/components/queries/query-editor.js#L57-L59) to sync with Angular's digest loop. Not sure if we need this here. But maybe @kravets-levko can shed more light?

package.json
arikfr: I don't think you need to declare it here to require from it, as it's a dependency of react-ace. Did you encounter an issue after removing it from `package.json`?

client/app/react-components/QueryEditor.js
```diff
@@ -0,0 +1,229 @@
...
+              }}
+              showPrintMargin={false}
+              wrapEnabled={false}
+              onLoad={this.onLoad}
+              onPaste={this.onPaste}
+              onChange={(queryText) => { this.props.updateQuery(queryText); }}
+            />
+          </div>
+
+          <div className="editor__control">
```
arikfr: It does make sense to move the editor controls from `query.html` into their own component, but why should they be part of the query editor component?

client/app/react-components/QueryEditor.js
```diff
@@ -0,0 +1,229 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { map } from 'lodash';
+
+import { OverlayTrigger, Tooltip } from 'react-bootstrap';
```
arikfr: We're adopting [Ant](https://ant.design/docs/react/introduce) as UI library (already used for date parameters). So in this case we should use [Ant's Tooltip component](https://ant.design/components/tooltip/).

Checkout the data parameters code on how to include its stylesheets.

client/app/components/queries/query-editor.js
```diff
@@ -1,160 +1,6 @@
...
-      },
-    },
-  };
-}
+import { react2angular } from 'react2angular';
+import QueryEditor from '@/react-components/QueryEditor';
 
 export default function init(ngModule) {
-  ngModule.directive('queryEditor', queryEditor);
+  ngModule.component('queryEditor', react2angular(QueryEditor, null, ['QuerySnippet', 'Query', 'KeyboardShortcuts']));
```
arikfr: Why not include the `react2angular` declaration with the component itself?

client/app/react-components/QueryEditor.js
```diff
@@ -0,0 +1,229 @@
+import React from 'react';
```
arikfr: Please change the filename extension to `.jsx`. Also as we have a dedicated extension for React component, I think we can move them back into `app/component` out of their `app/react-component` exile :)

client/app/react-components/QueryEditor.js
kravets-levko: @arikfr Actually, I don't remember why we need `$timeout` here :-( Probably, there were some issues when trying to restore selection, and `$timeout` fixed that, but it's better just to try - remove `window.setTimeout` and check if it works correctly.

client/app/react-components/QueryEditor.js
arikfr: I need to review the structure of this page to get a good idea, but maybe EditorControls?

Anyway, it shouldn't be coupled with the editor component. What's beautiful in React that you can nicely compose multiple small narrow scoped components.

client/app/components/queries/query-editor.js
washort: Oh, I see what you mean now.

package.json
washort: Removing it gives linter errors (and I'm reluctant in general to not explicitly list all dependencies).

washort:
Addressed review comments.

package.json
```diff
@@ -70,7 +70,9 @@
...
     "react-dom": "^16.3.2",
+    "react-bootstrap": "^0.32.3",
```
arikfr: This is not needed anymore, right?

client/app/components/QueryEditor.jsx
```diff
@@ -0,0 +1,240 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { map } from 'lodash';
+import { Tooltip } from 'antd';
```
arikfr: While webpack supposed to get rid of unused code, we did import specific components (`antd/lib/tooltip`) when we used it in other places. @kravets-levko do you remember why?

client/app/components/QueryEditor.jsx
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  render() {
+    // eslint-disable-next-line react/prop-types
+    const modKey = this.props.KeyboardShortcuts.modKey;
+    const parameterTooltip = <span>Add New Parameter (<i>{modKey} + P</i>)</span>;
+    const formatTooltip = <span>Format Query</span>;
+    const saveTooltip = <span>{modKey} + S</span>;
+    const executeTooltip = <span>{modKey} + Enter</span>;
+    const acTooltip = <span>Autocomplete</span>;
```
arikfr: Why not define it where we use it?

washort: Sounds good. Using the bootstrap style these were a good bit longer so I pulled them out of line.

package.json
washort: removed

client/app/components/QueryEditor.jsx
washort: fixed

client/app/components/QueryEditor.jsx
washort: fixed

washort:
Thanks. This is my first experience with React so I'm sure there's better ways to organize things. :-)

PR #2638, opened 2018-06-28
Dashboard stops rendering when adding widget with empty query

kravets-levko:
Issue getredash/redash#2534

When adding widget with empty query, dashboard tried to access query results. Empty query returns `QueryResultsError` instead of `QueryResults`, and the first one does not have some methods, for example `toPromise`. This PR fixes the issue by adding `QueryResultsError.toPromise` method which just returns rejected promise.


arikfr:
👍

PR #2635, opened 2018-06-28
Add tests for Query Results

ariarijp:
Hi team,

I Added some tests for Query Results Query Runner.

Please feel free to ask me about this PR.

Thanks,

PR #2639, opened 2018-06-28
Add tests for SpreadSheets

ariarijp:
Hi team,

I Added some tests for Google SpreadSheets Query Runner.

This PR is similar to #2635 .

Thanks,


arikfr:
Thanks!

> This PR is similar to #2635 .

We should probably merge the two utility functions into one, as they have similar function/purpose./

Issue #2402, opened 2018-03-20
Keyboard shortcut to collapse query editor/schema browser


arikfr:
To help users take full advantage of the collapsible nature of the areas in the query editor screen, we need to add quick keyboard shortcuts to collapse query editor + schema browser or only the schema browser.

I was thinking about:

* `Alt+D`: hide all
* `Alt+Shift+D`: hide only schema browser

In terms of implementation it looks like we can add this in the `resizable-toggle` directive. It will need to accept a list of keyboard shortcuts, and then bind with them.

arikfr:
@kravets-levko what do you think about the implementation plan?

idoDavid:
Can you add collapsable queries to the editor as well? so if i'm using a lot of `with x as..` I would be able to collapse individual queries.

Also, a full-screen editor mode would be very useful.

kocsmy:
@idoDavid We're improving the max-height of the query editor massively, would this help?

https://discuss.redash.io/t/increase-the-max-height-of-the-query-editor/1861/4

arikfr:
@kocsmy no, it won't :)

@idoDavid that's a nice idea. you can check if Ace Editor (the component we use for code editing) supports collapsable regions and maybe take a stab at implementing? :)

arikfr:
Implemented in #2640.

PR #2640, opened 2018-06-28
 Keyboard shortcut to collapse query editor/schema browser

kravets-levko:
Issue getredash/redash#2402

Combinations assigned as described in issue. `KeyboardShortcuts` updated to allow using multiple handlers for the same shortcut (needed to toggle both editor and schema browser). Changes are backward-compatible (all existing shortcuts work).


arikfr:
🕺

PR #2649, opened 2018-06-29
Add references to Airbnb's style guides in contribution guide.

arikfr:
As we begin [a big refactor](https://discuss.redash.io/t/react-migration-an-experiment-with-less-di/2077) of our frontend codebase, I thought that it's good time to revisit our style and make sure we're consistent.

As we already use Airbnb's `eslint` rules and it seems widely adopted, I thought that we should follow their general guide as well, mainly including filename convention:

https://github.com/airbnb/javascript#naming-conventions
https://github.com/airbnb/javascript/tree/master/react#naming

It's different than what we did so far, but seems reasonable.

@getredash/maintainers what do you think?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,7 +58,7 @@ If you would like to suggest an enhancement or ask for a new feature:
 - Please add [documentation](#documentation) for new features or changes in functionality along with the code.
 - Please follow existing code style:
   - Python: we use PEP8 for Python.
-  - Javascript: we use Airbnb's style guides for [JavaScript](https://github.com/airbnb/javascript#naming-conventions) and [React](https://github.com/airbnb/javascript/blob/master/react) (currently we don't follow Airbnb's convention for naming files, but we're gradually fixing this).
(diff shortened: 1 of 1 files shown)


jezdez:
@arikfr I can't particular comment on the style guide's content (other than it looks sensible to me) I want to say that I fully agree that adopting a style guide is critical for Redash's growth both in terms of maintaining a high level of code quality as well as lowering the barrier of entry for the potentially increasing influx of new contributors.

Related but not part of this ticket maybe we should also consider adopting a backend style guide like @ambv's [black](http://black.readthedocs.io/) (which also has a [bot to automate it](https://github.com/Mariatta/black_out)).

arikfr:
You mentioning `black` reminded me that I also apply [prettier](https://github.com/prettier/prettier) to all JS/HTML/CSS code I touch. I will mention it in the styleguide too.

As for `black` itself, my main concern about it is that to run it you need Python 3.6+. Might make it harder to run for new contributors. Maybe we should go with `autopep8` for now? Any idea on the difference between the two?

jezdez:
@arikfr black is a lot more opinionated about how to interpret PEP8 (and other PEPs) and goes a bit further in some areas, so I would say it's a lot closer to prettier than just using the autopep8 tool for example. I didn't notice black is only Python 3.6+, that's indeed a bummer. 

So maybe your recommandation to follow PEP8 suffices for now and we can pick up black once we're at 3.6?

arikfr:
> So maybe your recommandation to follow PEP8 suffices for now and we can pick up black once we're at 3.6?

I guess so... should we setup something automatic or keep this at a recommendation level for now?

PR #2650, opened 2018-06-30
Throw error when non-owner tries to add a user to dashboard permissions

### LINKED ISSUE #2592: A user that is granted permissions to a dashboard cannot grant permissions to others for the dashboard
Copies https://github.com/mozilla/redash/issues/346

Making an issue to submit a PR against.

alison985:
fixes #2592 

I choose this solution instead of changing `require_admin_or_owner` in `ObjectPermissionsListResource` since that would effect permissions for multiple operations.


client/app/components/permissions-editor/index.js
```diff
@@ -53,7 +53,10 @@ const PermissionsEditorComponent = {
       $http.post(this.aclUrl, body).success(() => {
         user.alreadyGrantee = true;
         loadGrantees();
-      });
+      })
+        .catch(() => {
```
arikfr: This is too general and might hide other errors. We should:

* Check for status code 403, in which case show the user the error message.
* In any other case, just say that something went wrong.

client/app/components/permissions-editor/index.js
```diff
@@ -53,7 +53,14 @@ const PermissionsEditorComponent = {
       $http.post(this.aclUrl, body).success(() => {
         user.alreadyGrantee = true;
         loadGrantees();
-      });
+      })
+        .catch((error) => {
+          if (error.status === 403) {
+            toastr.error('You can not add a user to this dashboard. Ask the dashboard owner to grant them permissions.');
```
jezdez: I think it needs to be `cannot` and not `can not`, right?

client/app/components/permissions-editor/index.js
```diff
@@ -53,7 +53,14 @@ const PermissionsEditorComponent = {
       $http.post(this.aclUrl, body).success(() => {
         user.alreadyGrantee = true;
         loadGrantees();
-      });
+      })
+        .catch((error) => {
```
jezdez: Also let's move this to the end of line 56, e.g. `}).catch((error..`

client/app/components/permissions-editor/index.js
alison985: Oxford Dictionary says both are acceptable: https://en.oxforddictionaries.com/usage/cannot-or-can-not However, since I had to do a commit to address your second comment, I changed it.

alison985:
@jezdez addressed

PR #2652, opened 2018-06-30
don't autoexecute query on changing datasource

### LINKED ISSUE #2651: Don't execute auto-execute query when changing datasources
Copies https://github.com/mozilla/redash/issues/436 to port https://github.com/mozilla/redash/issues/29 upstream.

Right now when you change datasource the query automatically executes. This is an issue to delete the automatic execution.

alison985:
fixes #2651 

This is a port of a Mozilla feature upstream.


alison985:
That makes sense to me Arik. I'm assuming you mean just hide the display of the results from the UI. 

There is already a `ng-if="showDataset"` in the query.html page for the tabs and data of query results. I tried adding `$scope.showDataset = false;` locally and adding a setter for `showDataset` to fix this error: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Getter_only but that didn't work so I added a new variable.

arikfr:
> That makes sense to me Arik. I'm assuming you mean just hide the display of the results from the UI.

No, actually the important thing is to clear the reference to the last result (`latest_query_data_id`). Because otherwise someone loading this query will see that it uses some data source that the results don't belong to.

alison985:
@arikfr That's already happening here: https://github.com/getredash/redash/blob/master/client/app/pages/queries/view.js#L305

alison985:
@arikfr Do I need to do anything else before this can be merged?

jezdez:
@alison985 Let's wait till Arik is back for this to merge.

client/app/pages/queries/view.js
```diff
@@ -120,6 +120,7 @@ function QueryViewCtrl(
     Notifications.getPermissions();
   };
 
+  $scope.dataSourceHasNotChanged = true;
```
jezdez: Let's turn around the logic and call this `dataSourceChanged` with a default of `false` so we don't have to deal with a double negative below.

alison985: @jezdez Updated.

jezdez: @alison985 Thank you! 😃

alison985:
@jezdez Ready for Review.

client/app/pages/queries/query.html
```diff
@@ -217,7 +217,9 @@ <h3>
               <!-- End of Query Execution Status -->
 
               <!-- tabs and data -->
-              <div ng-if="showDataset" class="flex-fill p-relative">
+              <div ng-if="showDataset && dataSourceChanged" class="flex-fill p-relative">
+                <div class="d-flex flex-column p-absolute" style="left: 0; top: 0; right: 0; bottom: 0;">
```
jezdez: This opens a new `div` without closing it, and also uses inline styles. Is this needed or just for debugging purposes?

alison985: These lines actually look like the result of a bad merge as they serve no purpose. I've removed them. The in-line styles below were there before me.

client/app/pages/queries/query.html
```diff
@@ -217,7 +217,7 @@ <h3>
               <!-- End of Query Execution Status -->
 
               <!-- tabs and data -->
-              <div ng-if="showDataset" class="flex-fill p-relative">
+              <div ng-if="showDataset && dataSourceChanged" class="flex-fill p-relative">
```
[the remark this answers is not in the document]
alison985: @arikfr This variable was originally `dataSourceHasNotChanged` but @jezdez wanted to remove the double negative. This is the variable added to handle your comments above to not show results when the datasource has changed. You're right to question this, I will add the ! now.

alison985:
Ping @jezdez and @arikfr

alison985:
Docker Hub issue made build fail. When Docker Hub is back up the build should be rerun.

alison985:
Anything else on this one @jezdez and/or @arikfr?

chang:
+1 This could be especially dangerous when developing a mutation query on a test database. If the schemas are similar enough, it could successfully and unintentionally auto-execute while switching over to a prod data source.

arikfr:
Hi,

Thanks.

jezdez:
@arikfr I'm a bit confused by your last review comment, you're right that changing the source is only possible in edit mode, but I still see a request happening against the server when the source is changed there, and I believe the query is actually executed. I'd appreciate if you'd double-check this on your side so we can close this.

Here's a screen recording of the current master executing the query when changing a data source (if it's the same type of data source): https://dsc.cloud/a46d84/RhVthKiRVrg1Od79fBzY

arikfr:
I just tried it out on https://redash-preview.netlify.com/ and couldn't reproduce it. And looking at your video again I can see that it's definitely not the latest master. Maybe you didn't rebuild the assets?

jezdez:
Lemme check again, thanks!

jezdez:
You right, @arikfr, that was it :) Sorry for the noise!

PR #2655, opened 2018-06-30
add partition key marker for Athena and Presto data sources

### LINKED ISSUE #2654: Add partition key marker to Athena and Presto columns
Copies https://github.com/mozilla/redash/issues/430 to port https://github.com/mozilla/redash/issues/185 upstream

This is an issue to open a PR against.

alison985:
fixes #2654 

This was originally to port the partition key marker upstream, but upstream also didn't have the column type in the schema for Athena and Presto data sources so I added that as well since it was the same lines being touched.


alison985:
@arikfr What do you think about this one?

redash/query_runner/athena.py
```diff
@@ -143,24 +143,17 @@ def get_schema(self, get_stats=False):
...
         if error is not None:
-            raise Exception("Failed getting schema.")
-
-        results = json.loads(results)
-        for row in results['rows']:
-            table_name = '{0}.{1}'.format(row['table_schema'], row['table_name'])
-            if table_name not in schema:
-                schema[table_name] = {'name': table_name, 'columns': []}
-            schema[table_name]['columns'].append(row['column_name'])
+            raise Exception("Failed getting schema from query.")
```
jezdez: The exception message could stay the same here to stay backward compatible.

redash/query_runner/__init__.py
```diff
@@ -177,3 +177,19 @@ def get_configuration_schema_for_query_runner_type(query_runner_type):
 def import_query_runners(query_runner_imports):
     for runner_import in query_runner_imports:
         __import__(runner_import)
+
+
+# for athena and presto
+def format_schema(results):
+    schema = {}
+    for row in results['rows']:
+        table_name = '{0}.{1}'.format(row['table_schema'], row['table_name'])
```
jezdez: No need to use `'{0}.{1}'` here instead of `'{}.{}'` since providing the order of the format string variables was only mandatory in Python 2.6 (and at least the Redash Docker container runs Python 2.7.x).

jezdez:
Oh, I meant "LGTM with comments" :)

redash/query_runner/__init__.py
```diff
@@ -177,3 +177,19 @@ def get_configuration_schema_for_query_runner_type(query_runner_type):
 def import_query_runners(query_runner_imports):
     for runner_import in query_runner_imports:
         __import__(runner_import)
+
+
+# for athena and presto
```
jezdez: Actually let's move this function into the presto module (since Athena is a product based on Presto) and import it in the athena module.

Also, could you please add a docstring to the function describing what it does please?

alison985:
@jezdez updated

PR #2660, opened 2018-07-01
Allows using unpublished queries in alerts

### LINKED ISSUE #2659: Allow using unpublished queries in alerts
copies https://github.com/mozilla/redash/issues/429 to port https://github.com/mozilla/redash/issues/97 upstream.

Opening an issue to submit a PR against.

alison985:
fixes #2659 

This is a port of Mozilla fork functionality upstream. 

Original PR commentary is here: https://github.com/mozilla/redash/issues/97 where Arik specifically said that unpublished queries are excluded from the Dashboard Add Widget dialog on purpose. Therefore, we only added unpublished queries to the alert search.


arikfr:
Actually, the same reasoning I mentioned for Add Widget applies here as well. Basically draft queries are not supposed to be used anywhere, to make sure people publish them and not keep them in this state forever.

Unless there is a very good reason for allowing draft queries in alerts, I don't think we will merge this in...

arikfr:
Closing this for now. Let's re open if there is a reason to change this behavior.

emtwo:
@arikfr, suppose we wanted to make this into an extension for the Mozilla fork. I imagine we could use the `Policy` object from here https://github.com/getredash/redash/pull/2560 to check if the object passed into `Query.search()` should have the `include_drafts` attribute.

This would be a small change in this codebase here to allow the attribute to be added if the policy (only turned on in the Mozilla fork) allows it.

What do you think of this?

Issue #2620, opened 2018-06-22
Syntax Error on nested queries from google spreadsheet with blank column names


metheos:
Redash 4.0.1+b4038
Pulling data from google spreadsheet
the spreadsheet has some columns with blank headers
If I run a query against the query results from pulling in the spreadsheet (e.g. query_45) I get an error 
```
File "/opt/redash/redash.4.0.1.b4038/redash/query_runner/query_results.py", line 92, in create_table
    connection.execute(create_table)
OperationalError: near ")": syntax error
```

This is caused by create_table with a create table string that contains a blank column header at the end.
e.g. 
```sql
CREATE TABLE table_46 (Name, Phone, )
```

I hacked a fix into my query_results.py to create named columns for blank columns.

------------------query_results.py:79--------------------
```python
def fix_column_name(name,columncounter):
    if name.strip() == '':
        name = u"BLANK{colno}".format(colno=columncounter)

    return name.replace(':', '_').replace('.', '_').replace(' ', '_')


def create_table(connection, table_name, query_results):
    columns = [column['name']
               for column in query_results['columns']]
    safe_columns = []
    colno = 0
    for column in columns:
        safe_columns.append(fix_column_name(column,colno))
        colno = colno + 1 
```

ariarijp:
@arikfr I have two solutions for it.

1. Raise an exception when blank column name found
2. Rename blank column names implicitly (example: `BLANK_1`, `BLANK_2`...)

Do you have another solutions?

ariarijp:
Good, `column_{col_num}` looks more better and user friendly.

However, I would like to clarify your idea.

Does **column number** means `A` (Excel style) or '1' (Numeric style) ?

metheos:
I would suggest Excel style with letters since that's what excel and google docs both use for the column id. It would make it easier for someone to refer back to their spreadsheet if it's a complex one.

arikfr:
Closed in #2663.

PR #2663, opened 2018-07-03
Set column name implicitly when column name is blank

ariarijp:
Hi @arikfr

This PR will resolve Issue #2620

I used Excel-style column letters instead of column number.

Please feel free to let me know your thoughts.

Thanks,


arikfr:
Thanks!

Issue #2602, opened 2018-06-19
Dashboard shows a spinner when query param has no value


arikfr:
Current behavior:
When opening a dashboard with queries that use parameters, that have no default value, it will show the spinner on the widget indefinitely.

The correct behavior:
Show a message saying parameter value is required.

arikfr:
@kravets-levko @kocsmy assigned to you both, as Levko will implement the functionality, but will need Zsolt's help for the UX/UI.

kravets-levko:
It's more generic case: dashboard shows spinner on widgets always when `QueryResults` promise was rejected for this widget.

PR #2664, opened 2018-07-03
Dashboard shows a spinner when query failed to load

kravets-levko:
Issue getredash/redash#2602 

As I mentioned in comment to the issue, bug is more generic: widget didn't process all query results errors correctly.

Screenshot how it works with this fix:
image

### DIFF
--- a/client/app/services/widget.js
+++ b/client/app/services/widget.js
@@ -116,17 +116,23 @@ function WidgetFactory($http, Query, Visualization, dashboardGridOptions) {
       // while widget is refreshing, `this.data` !== `this.queryResult`
 
       if (force || (this.queryResult === undefined)) {
+        this.loading = true;
+        this.refreshStartedAt = moment();
+
         if (maxAge === undefined || force) {
           maxAge = force ? 0 : undefined;
         }
         this.queryResult = this.getQuery().getQueryResult(maxAge);
-        this.loading = true;
-        this.refreshStartedAt = moment();
 
-        this.queryResult.toPromise().finally(() => {
-          this.loading = false;
-          this.data = this.queryResult;
-        });
+        this.queryResult.toPromise()
+          .then((result) => {
+            this.loading = false;
+            this.data = result;
+          })
+          .catch((error) => {
+            this.loading = false;
+            this.data = error;
+          });
       }
 
       return this.queryResult.toPromise();
--- a/client/app/services/widget.js
+++ b/client/app/services/widget.js
@@ -106,29 +106,27 @@ function WidgetFactory($http, Query, Visualization, dashboardGridOptions) {
     }
 
     load(force, maxAge) {
-      this.loading = true;
-      this.refreshStartedAt = moment();
-
       if (!this.visualization) {
         return undefined;
       }
 
-      if (force || this.queryResult === undefined) {
+      // Both `this.data` and `this.queryResult` are query result objects;
+      // `this.data` is last loaded query result;
+      // `this.queryResult` is currently loading query result;
+      // while widget is refreshing, `this.data` !== `this.queryResult`
+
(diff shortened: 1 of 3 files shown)


client/app/services/widget.js
```diff
@@ -106,29 +106,27 @@ function WidgetFactory($http, Query, Visualization, dashboardGridOptions) {
...
-          (queryResult) => {
-            this.data = queryResult;
-            this.loading = false;
-          },
-          () => {
-            this.loading = false;
-            this.data = null;
-          },
-        );
+        this.queryResult.toPromise().finally(() => {
```
arikfr: Wouldn't it be cleaner to reject the Promise with a `QueryResultError` object, and then here use a local reference to query result instead of relaying on having a reference to the query result object that triggered the process?

I think it will make it easier to refactor into promises only API later on.

kravets-levko: I was thinking about it. But if we'll decide to reactor it - it may require changes in other places that use query results. For sure, I can do it, but it may take some time (including testing).

arikfr: But until now we were not rejecting this promise in case of errors, so no other code depends on it being of specific type. Or did you mean something else?

kravets-levko: Actually, all code related to query results does not rely on state of promise; instead it's using `getStatus()`/`getError()` method of query result, which can be `QueryResult` or `QueryResultError` instance. Maybe the right way is to reject promise when there are errors with query result, but there are some issues. First of all, when you request query result, all its methods will modify original object. Therefore it's hard to return new  `QueryResult`/`QueryResultError` instance in `then/catch`. Another issue is that `Query.getQueryResult()` will return query result instance, not a promise; that means, if `QueryResult` will receive status `4/failed` it should turn into `QueryResultError`, which is impossible (without changing a way we return query results for a query). That's what I mean when saying that it may require massive refactoring for `Query`, `QueryResults`, `Widget` services and related UI code.

arikfr: I wasn't suggesting a global refactor or change... I was referring to the local code in `services/widget.js`. Previously it was using the QueryResult object from the Promise:

https://github.com/getredash/redash/pull/2664/files#diff-3fe10675d2975aacb62dabd70951b589L124

But now, it uses the local `this.queryResult`: 

https://github.com/getredash/redash/pull/2664/files#diff-3fe10675d2975aacb62dabd70951b589R128

My suggestion is to resolve the promise with a `QueryResultError` object (instead of a string) and keep using the resolved object as the value we assign to `this.data` -- in the error state as well.

kravets-levko:
@arikfr Updated this PR - changes are based on our discussion. Please review it again. Thanks!

client/app/services/widget.js
```diff
@@ -106,29 +106,32 @@ function WidgetFactory($http, Query, Visualization, dashboardGridOptions) {
...
+      // `this.queryResult` is currently loading query result;
+      // while widget is refreshing, `this.data` !== `this.queryResult`
+
+      if (force || (this.queryResult === undefined)) {
         if (maxAge === undefined || force) {
           maxAge = force ? 0 : undefined;
         }
         this.queryResult = this.getQuery().getQueryResult(maxAge);
+        this.loading = true;
+        this.refreshStartedAt = moment();
```
arikfr: Shouldn't this come before the line where we call `getQueryResult`?

kravets-levko: It doesn't matter, actually. But if you think it will look better - I can do it

kravets-levko: @arikfr Done

arikfr: It's not about looks :) It's about correctness. Currently the code here assumes something about `getQueryResult`'s implementation -- that it will return immediately. But it's not necessarily will always be the case.

PR #2669, opened 2018-07-09
Store Athena query id with result metadata

tdawber:
Here's a simple change that stores the Athena query id in the result metadata.

Here at Atlassian we are keen to be able to cross-reference Athena usage data with the internal Redash data, and this additional piece of metadata will help us achieve this.


tdawber:
Hey @arikfr, would you consider extending the metadata saved for Athena results like this?

If you remember, we would previously send some query information in a comment in the queries themselves, but this has proved difficult with Athena.  Storing the query id would allow us to do some cross referencing the other way.

arikfr:
👌 

Thanks!

PR #2670, opened 2018-07-09
Sort objects (datasources, users, etc.) and change number formatting

sdif:
Hi,

1. Sort objects (`all` method) by name
2. Change number formatting 

It's helpful for us and I think it will be helpful for other people too :)

### DIFF
--- a/client/app/visualizations/chart/index.js
+++ b/client/app/visualizations/chart/index.js
@@ -17,7 +17,7 @@ const DEFAULT_OPTIONS = {
   columnMapping: {},
 
   // showDataLabels: false, // depends on chart type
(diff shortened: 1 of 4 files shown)


sdif:
Hi Arik,

Regarding the default format of the numbers I'm going to take a look to do it with a configuration option in order to avoid breaking changes.

For objects name sorting, I think it will be acceptable for every user and even preferred but I'm open to read you opinion on this ! 
I just found those two merged PR : #2568 and #2619 
What we want to do is a bit wider than the two PR made by @alison985 because it will also affect user, group, alert, dashboard and query listing.
If you are ok with it, I can make a separate PR with only the modification regarding the `redash/models.py` file.

PR #2672, opened 2018-07-11
Initial update to circleci v2 config.

### LINKED ISSUE #2609: Migrate to Circle CI v2.0
Hi folks! 👋 

Circle CI is dropping support for version 1.0 builds end of August, so I'm suggesting to migrate to the new build config format soon. See note in the Circle CI documentation [here](https://circleci.com/docs/1.0/) ⚠️ 

They also provide documentation on [how to do the migration](https://circleci.com/docs/2.0/migrating-from-1-2/).

Would you be OK with a PR to also update the CI configuration to use the same mechanism for running the tests as for running the tests locally? Currently we install [dependencies manually](https://github.com/getredash/redash/blob/master/circle.yml#L10) and [run pytest directly](https://github.com/getredash/redash/blob/master/circle.yml#L19) as opposed to using [docker-compose and the tests entrypoint](https://redash.io/help/open-source/dev-guide/docker)

```bash
$ docker-compose run --rm server tests
```

jrbenny35:
Should fix #2609 

This new config uses the docker-compose configuration to run both the unit tests and a new suite of integration tests. The integration tests can be found [here](https://github.com/mozilla/redash-ui-tests).

The unit tests and integration tests run per commit. These also run on each master/release branch build as well as docker deploys.

I think I have most of the old config added but, of course, this can be changed as needed.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -74,7 +74,7 @@ jobs:
           name: Pull redash images
           command: |
             set -x
-            docker-compose -f docker-compose.yml up -d
+            docker-compose -f docker-compose.yml up -d --no-start
             sleep 10
       - run:
           name: Pull redash-ui-tests
@@ -88,6 +88,7 @@ jobs:
             docker-compose run --rm --user root server /app/manage.py users create_root root@example.com "rootuser" --password "IAMROOT" --org default
             docker-compose run --rm --user root server /app/manage.py ds new "ui-tests" --type "url" --options '{"title": "uitests"}'
             docker-compose run -d -p 5000:5000 --user root server
+            docker-compose start postgres
       - run:
           name: Run tests
           command: |
@@ -98,6 +99,11 @@ jobs:
   build-deploy: # build for master or release branches
     docker:
       - image: circleci/node:6
+    environment:
+      NAME: redash
+      VERSION: $(python ./manage.py version)
+      FULL_VERSION: $VERSION+b$CIRCLE_BUILD_NUM
+      FILENAME: $NAME.$FULL_VERSION.tar.gz
     working_directory: ~/redash
     steps:
       - attach_workspace:
@@ -105,6 +111,8 @@ jobs:
       - run:
           name: Build
           command: bin/pack "$CIRCLE_BRANCH"
+      - store_artifacts:
+          path: $NAME.$FULL_VERSION.tar.gz
   build-release: # build for tags
     docker:
       - image: circleci/node:6
--- a/circle.yml
+++ b/circle.yml
@@ -5,7 +5,7 @@ notify:
 jobs:
   build:
     docker:
-      - image: circleci/node:6
+      - image: circleci/node:8
     working_directory: ~/redash
     steps:
       - checkout
@@ -98,7 +98,7 @@ jobs:
           path: report.html
   build-deploy: # build for master or release branches
     docker:
-      - image: circleci/node:6
+      - image: circleci/node:8
     environment:
       NAME: redash
       VERSION: $(python ./manage.py version)
@@ -115,7 +115,7 @@ jobs:
           path: $NAME.$FULL_VERSION.tar.gz
   build-release: # build for tags
     docker:
-      - image: circleci/node:6
+      - image: circleci/node:8
     working_directory: ~/redash
     steps:
       - attach_workspace:
--- a/circle.yml
+++ b/circle.yml
@@ -86,7 +86,7 @@ jobs:
             docker-compose run --rm --user root server create_db
             docker-compose run --rm postgres psql -h postgres -U postgres -c "create database tests"
             docker-compose run --rm --user root server /app/manage.py users create_root root@example.com "rootuser" --password "IAMROOT" --org default
-            docker-compose run --rm --user root server /app/manage.py ds new "ui-tests" --type "url" --options '{"url": "uitests"}'
+            docker-compose run --rm --user root server /app/manage.py ds new "ui-tests" --type "url" --options '{"title": "uitests"}'
             docker-compose run -d -p 5000:5000 --user root server
       - run:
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -1,38 +1,151 @@
...
   webhooks:
     - url: https://webhooks.gitter.im/e/895d09c3165a0913ac2f
-general:
-  branches:
-    ignore:
-      - gh-pages
+jobs:
+  build:
+    docker:
+      - image: circleci/node:6
```
arikfr: We were using Node v6.9 because that's the latest that was available in the old platform. Now I assume we can use something newer?

circle.yml
```diff
@@ -1,38 +1,151 @@
...
+            set -x
+            pip install --upgrade pip
+            pip install docker-compose>=1.18
+            docker-compose --version
+      - run:
+          name: Pull images and setup
+          command: |
+            set -x
+            docker-compose up -d
+            sleep 10
```
arikfr: This feels brittle... did you try the following:

1. `docker-compose up -d --no-start` -> this will create all the necessary containers.
2. `docker-compose start postgres` -> this will return when Postgres is running.
3. Now you can proceed to creating the database and running the tests.

?

circle.yml
```diff
@@ -1,38 +1,151 @@
...
+      - image: circleci/node:6
+    working_directory: ~/redash
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: Tag deploy
+          command: |
+            set -x
+            bin/pack "$CIRCLE_TAG"
```
arikfr: Don't we need to define the tar ball file as an artifact?

circle.yml
jrbenny35: I didn't specify an exact version in error, but It defaults to 6.14 if I remember correctly. You can view some builds/logs here: https://circleci.com/gh/mozilla/redash/tree/update-circleci

circle.yml
jrbenny35: That's pretty slick, I should give that a go since ```--rm``` will run the containers anyway.

circle.yml
jrbenny35: So this where I need some clarification. Do we just want the artifact as a circleci artifact? Or does it need to be something specific. Circleci 2.0 doesn't have much on the old ```tarball``` command, just stating that it was an artifact.

jrbenny35: Looking at current master builds on circle, there is an artifact with a ```tar.gz``` extension. Is this made with the command ```bin/pack``` ?

jezdez: @jrbenny35 Yep, bin/pack basically puts everything in the repo clone into a tarball (except a few things like git history and node_modules etc).

circle.yml
jezdez: @arikfr There are a lot more modern npm versions on https://hub.docker.com/r/circleci/node/tags/ but I would recommend not doing an update at the same time as this pull request to remove the chance for conflicting issues.

jezdez:
@arikfr Just to clarify, are you okay with this PR running the UI tests that are developed over in https://github.com/mozilla/redash-ui-tests here?

circle.yml
jrbenny35: @jezdez I don't think 6.9 is available as a docker image though. Unless you are talking about doing a major update to the code-base, outside of this PR.

arikfr: Let's update to `circleci/node:8`, which is the latest LTS and the version I've been working with for along time now, so doubt there will be any issues.

circle.yml
arikfr: Yes, it just needs to be available as a CircleCI artifact.

jrbenny35:
Just added some updates. Let me know if there is anything still missing.

hackebrot:
Hey @arikfr! 👋 

Can you please give this another look and see if we can merge this to master?

Thank you! 🙇

jrbenny35:
Any updates on this?

jezdez:
@jrbenny35 I've talked with @arikfr about this and we settled on this:

- If there is an option to configure Circle CI to *optionally* run the UI tests (e.g. by clicking a button) when needed, then we should do that.
- If there isn't an option we agreed that it's too early to use the UI tests in this repo and would like to ask you to remove them for now so we can merge this without it.

jrbenny35:
> If there is an option to configure Circle CI to optionally run the UI tests (e.g. by clicking a button) when needed, then we should do that.

What is the idea behind this?

jezdez:
> Am 24.08.2018 um 16:18 schrieb Benjamin Forehand Jr <notifications@github.com>:
> 
> If there is an option to configure Circle CI to optionally run the UI tests (e.g. by clicking a button) when needed, then we should do that.
> 
> What is the idea behind this?

Basically to make sure that the runtime of the tests isn’t increased and minor changes don’t trigger the UI tests by default. Instead when big features land or UI components are changed we could trigger the UI tests on demand from the Circle CI user interface.

jrbenny35:
> Basically to make sure that the runtime of the tests isn’t increased and minor changes don’t trigger the UI tests by default. Instead when big features land or UI components are changed we could trigger the UI tests on demand from the Circle CI user interface.

What about a branch or a commit string? I don't think there is a button to run tests on circle.

jezdez:
There is also the ability to [approve specific jobs of a workflow on Circle CI](https://circleci.com/docs/2.0/workflows/#holding-a-workflow-for-a-manual-approval)

jrbenny35:
I don't know if this change will work but from my research ```circleci step halt``` found [here](https://discuss.circleci.com/t/stop-build-with-success-failure-at-any-point-in-the-build-life-cycle/2691/4) allows the job to be cancelled without returning a failure code.

The string I used is ```!run-ui-tests```. Not sure if it will work until merging though.

PR #2674, opened 2018-07-12
Add support for the Chrome Logger extension

arikfr:
This adds support for logging to Chrome's console using the [Chrome Logger](https://craig.is/writing/chrome-logger) extension. 

This makes it easier to see server side information about the requests we're making (like what SQL queries it generates).

It's only enabled in debug mode.

### DIFF
--- a/redash/handlers/chrome_logger.py
+++ b/redash/handlers/chrome_logger.py
@@ -1,0 +1,54 @@
+import time
+import chromelogger
+from flask import g, request
+from flask_sqlalchemy import get_debug_queries
+
+
+def log_queries():
+    total_duration = 0.0
+    queries_count = 0
+
+    chromelogger.group("SQL Queries")
+
+    for q in get_debug_queries():
+        total_duration += q.duration
+        queries_count += 1
+        chromelogger.info(q.statement % q.parameters)
+        chromelogger.info("Runtime: {:.2f}ms".format(1000 * q.duration))
+
+    chromelogger.info("{} queries executed in {:.2f}ms.".format(queries_count, total_duration*1000))
+
+    chromelogger.group_end("SQL Queries")
+
+
+def chrome_log(response):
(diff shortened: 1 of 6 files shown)


redash/handlers/chrome_logger.py
```diff
@@ -0,0 +1,54 @@
...
+
+    chromelogger.info("{} queries executed in {:.2f}ms.".format(queries_count, total_duration*1000))
+
+    chromelogger.group_end("SQL Queries")
+
+
+def chrome_log(response):
+    request_duration = (time.time() - g.start_time) * 1000
+    queries_duration = g.get('queries_duration', 0.0)
+    queries_count = g.get('queries_count', 0.0)
```
jezdez: Maybe use `0` as a fall back for `queries_count`?

redash/handlers/chrome_logger.py
```diff
@@ -0,0 +1,54 @@
...
+        response.headers.add(*header)
+
+    return response
+
+
+def init_app(app):
+    if not app.debug:
+        return 
+
+    app.after_request(chrome_log)
```
jezdez: New line at the end of the file?

redash/metrics/request.py
```diff
@@ -32,7 +32,7 @@ def calculate_metrics(response):
                         request_duration,
                         queries_count,
                         queries_duration)
-
+
```
jezdez: No need for indentation here

jezdez:
BTW, there is also a Firefox extension to enable this for Firefox: https://addons.mozilla.org/en-US/firefox/addon/chromelogger/

redash/handlers/chrome_logger.py
arikfr: We really need to setup an auto formatter :-)

arikfr:
@jezdez thanks for the review and the pointer to the Firefox extension. 👍

PR #2677, opened 2018-07-12
Fix empty states for fav&tag

kocsmy:


arikfr:
Issues with empty states:

1.

When there are no results when selecting a tags combination, we show the "Let's get started" screen. Maybe a screen saying "No dashboards found tagged with 'sales' and 'marketing'."?

image

2. 

No favorites empty state can be a bit larger, maybe?

image

3. 

Search empty state can be larger too?

image

4. 

This is the initial loading animation when you open the dashboards page...

image

5. 

No indication that search is in progress.

---

Looks like we need a nice empty screen for the different states, for which we will use a different text based on the state.

kocsmy:
@arikfr thanks for the suggestions.
1. I'm preparing a state for this
2-3. I don't see the point of making it larger, why do you think we should make this larger?
4. Not sure what you mean, should we add more details here like "Loading dashboards..."?
5. The dashboard search seem to be very quick to me. Do we expect this to be slow (> 250-500ms)?

kocsmy:
FYI I believe the search should be top so I pushed the changes. Even though it'll disappear on the My Queries page and make the page "jump"—we should that fix in the future and make it searchable as well.

arikfr:
> 2-3. I don't see the point of making it larger, why do you think we should make this larger?

It's not really about making it larger, it's more about making it look nice and not like something a developer put in there as a place holder :)

> Not sure what you mean, shall we add a text like "Loading dashboards..." + the loader gif?

Same here: this spinner on a grey background just doesn't look good.

> The dashboard search seem to be very quick to me. Do we expect this to be slow (> 250-500ms)?

While I would hope it wouldn't be any slower than that, there are still cases when it might be (network or just issues with performance).

kocsmy:
> It's not really about making it larger, it's more about making it look nice and not like something a developer put in there as a place holder :)

As discussed, we'll go with the slightly advanced version:
screenshot 2018-07-16 15 47 44

The spinner now has a white background.

Not sure what to do with the search, I think we could apply the loading animation there (which might flash or not even appear as it's quick most of the time).

arikfr:
> Not sure what to do with the search, I think we could apply the loading animation there (which might flash or not even appear as it's quick most of the time).

Yes, loading animation is fine. We can try and optimize and show only after 500ms or some other threshold.

kocsmy:
Ok, here's what needs to be handled:
- Query search doesn't trigger the empty state (it shows the totally empty empty state)
- No dashboard/query found for this tag - this needs implementation
- My Queries needs search bar on top, too

arikfr:
I implemented the changes here in the original branch. It was easier to wire with logic while copying the markup then to merge and search for it afterwards.

Issue #2644, opened 2018-06-28
Explicit “Add Parameter” Button in Query Editor


arikfr:
Add a button to the interface to add a parameter to the query. This button will open the current parameter definition dialog, but will let you also define the parameter keyword. Once submitted, it will add the `{{parameter}}` string into the query.

* This will help with the date range issue (#2643), as it can define the parameter keyword and then in the query you can use `{{keyword.start}}` and `{{keyword.end}}` to get the start/end of the date range.
* Will help with discoverability of the feature (today there is no hint about the existence of parameters).
* Should be also triggered by a keyboard shortcut.

kocsmy:
Here's the proposed design for this:

- "+ Add Parameter" button added below the query editor.
- This opens the parameter dialog, see screenshot. We can change the form items based on the selected type, just like how it works now.
- The dialog closes and inserts the parameter to the query editor and the UI element for the parameter.

![query parameter](https://user-images.githubusercontent.com/2378022/42137210-a6635828-7d68-11e8-91d4-a8382f0e0c65.png)
![query parameter dialog](https://user-images.githubusercontent.com/2378022/42137211-a96ef96e-7d68-11e8-8596-24a2e68068e0.png)

arikfr:
Some notes:

* Maybe use an icon for the button instead of the long text (something like `{{ }}` maybe?)? While it will require some learning from the user, it will create a more streamlined UI.
* If we already add buttons there, let's mvoe the format button there.
* The Keyword should be the first item in the form, the Title should come after and Type last. Title will be auto filled (with some manipulation to make it human readble) based on Keyword value.
* Enter key in this form should submit it.
* We need a keyboard shortcut to trigger this UI.

kocsmy:
Maybe use an icon for the button instead of the long text (something like {{ }} maybe?)? While it will require some learning from the user, it will create a more streamlined UI.
- I was thinking about this and we could use the same for this: _"How to show if a widget is affected by a parameter? Maybe color code them, and show this color on the widget?"_ of https://github.com/getredash/redash/issues/2641

I'm a bit concerned that {{ }} might look like something is broken on the interface :)

The Keyword should be the first item in the form, the Title should come after and Type last. Title will be auto filled (with some manipulation to make it human readble) based on Keyword value.
- This is exactly what I did first, but then I was thinking what happens if you add a data-range. Then we'll need to add another keyword and it might be strange that something changes not below the type selector but on top of it.

arikfr:
[...]

PR #2678, opened 2018-07-12
Explicit "Add Parameter" Button in Query Editor

kravets-levko:
Implemented as described in getredash/redash#2644

Screenshots:

image
image
image

### DIFF
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -170,6 +170,12 @@ class Parameters {
     return this.query.options.parameters;
   }
 
+  add(parameterDef) {
+    this.query.options.parameters = this.query.options.parameters
+      .filter(p => p.name !== parameterDef.name);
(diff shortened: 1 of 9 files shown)


kravets-levko:
@kocsmy yes, you're right, there are so much screenshots in issue discussion :smile: Fixed

client/app/services/query.js
```diff
@@ -129,6 +129,14 @@ class Parameters {
     return this.query.options.parameters;
   }
 
+  add(parameterDef) {
+    if (isObject(parameterDef)) {
```
arikfr: When is it not an object?

kravets-levko: It's just for a case

arikfr: What case? :)

kravets-levko:
@arikfr Updated PR, please review it again. Thanks!

arikfr:
🍷

PR #2679, opened 2018-07-12
Add a required setting to parameters

smerrill:
I know that the UI/UX around parameters is getting an overhaul soon, but I hacked this together to allow parameters to be marked as required or not, and it appears to work in some quick testing.


arikfr:
This needs to be updated with recent changes to parameters implementation. Also, we would like to properly support optional parameters, as described in #3011.

Considering the above, I'm closing this pull request. Thank you for you effort though!

Issue #2484, opened 2018-04-24
Add ability to search table column names in schema browser


jezdez:
### Issue Summary

When using the schema browser to browse tables of data sources with a lot of columns it would be useful to in addition to optionally match the names of the columns in a specific table.

E.g.

In addition to search for "name-of-table" allow "name-of-table:column-name" or similar. Also "name-of-table:mn-n" would match "name-of-table:column-name" as well.

### Steps to Reproduce

1. Use table name to search in schema browser.
2. Try to find the name of a wanted column by continue to type.

A work-around right now is to use the page search on the browser side to find the column name.

Considering the use of a data source agnostic delimiter would be useful to prevent overlap with table names.

### Technical details:

* Redash Version: 4.0.0rc1
* Browser/OS: Firefox/macOS
* How did you install Redash: Docker

arikfr:
Sounds good to me. I like `:` as a delimiter, but BigQuery uses `:` in table names to separate project name from the dataset/table name. 

We can either ignore this fact or maybe we can just use space (` `) as a separator? 

Anyway, you (or anyone else) welcome to submit a pull request 👍

kocsmy:
+1, sounds like a nice improvement to me

jezdez:
Closed via #2681.

PR #2681, opened 2018-07-13
Add ability to search table column names in schema browser

alison985:
Re: #2484 and https://github.com/mozilla/redash/issues/380

This PR:
* Allows you to search for a table name, a column name, or both. A space should be used between table name and column name. If there is no space it is a table name search only, with a space anything after the space is for a column name.
* If more than one space is used it gets replaced with one space. This avoids having to give an error toastr message and the behavior that happens when there are multiple spaces.


arikfr:
If I remember right, the `filter` filter can take an object, and apply the search based on the object structure. Using this will require changing the data structure of the schema locally to the browser, but that shouldn't be much of an issue. And potentially will make implementation of filtering simpler.

jezdez:
@alison985 Remind me, is this closer to be solved? What can I do help you with this?

jezdez:
@alison985 The pull request description above says otherwise, which is why I'm asking. Could you update the description please?

alison985:
@jezdez Updated. Sorry about that.

PR #2682, opened 2018-07-15
Add Zeit's now support to have preview builds for every PR

arikfr:
Using [Zeit's Now GitHub integration](https://zeit.co/blog/now-for-github) we will now have a unique deployment for each Pull Request. 👌

We're using Now only to serve the HTTP requests. Redash worker, database and Redis are shared. Meaning that this is useful mainly for reviewing frontend changes, but considering that there are quite a few pull requests that are waiting for visual review, I think this will provide value too.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,3 +11,4 @@ RUN chown -R redash /app
 USER redash
 
 ENTRYPOINT ["/app/bin/docker-entrypoint"]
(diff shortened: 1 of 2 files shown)


arikfr:
@getredash/maintainers

jezdez:
Wow neat!

PR #2686, opened 2018-07-18
Implement server side pagination and sorting for queries lists

### LINKED ISSUE #78: Better support for large query results

jezdez:
This was initially reported in #78 but also applies to the queries search where the list of results can become quite large depending on the number of queries and search term used.

This was originally reported in mozilla#384 and mozilla#385.

## Notes:

- The sorting criteria are hardcoded and only allow specified columns (and related columns) to be used.

- It adds a page size dropdown to the query-list page (and the appropriate backend handling) and allows clearing the search input to reset the query list to the non search result status.

- It deprecates the standalone backend query search handler and redirects requests for it.

- I believe the live paginator should also be used in more list views and the non-live paginator deprecated or outright removed.

- I'm not sure how Redash is handling API backward compatibility, so I erred on the side of caution and followed a deprecation policy (that could mean removing old APIs after the next version for example).

## Screenshot:

screen-shot-2018-07-18-22-31-32

## TODOs:

- [ ] Check code if uses of old search API works with handler redirect

### DIFF
--- a/client/app/filters/index.js
+++ b/client/app/filters/index.js
@@ -4,7 +4,7 @@ import { capitalize as _capitalize, isEmpty } from 'lodash';
 export function durationHumanize(duration) {
   let humanized = '';
 
-  if (duration === undefined) {
+  if (duration === undefined || duration === null) {
     humanized = '-';
   } else if (duration < 60) {
     const seconds = Math.round(duration);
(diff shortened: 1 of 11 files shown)


kocsmy:
Thanks for the PR @jezdez 

Can we:
- Move it to the bottom of the left side of the results
- Only show this if we have more than 20 results

And is there any way to make the dropdown look better in Firefox?

arikfr:
Btw, issue #78 is about when a single query's result is very large and needs pagination.

kocsmy:
@arikfr yes, this is a really nice component. I'll make sure it fits to our current design.

arikfr:
🍾

jezdez:
Apologies for the late reply @kocsmy, I was out for a bit. Thank you for addressing the issues @arikfr! It's much appreciated 🎉 🎊 🍾

Issue #2642, opened 2018-06-28
Improve Date/DateTime type parameters


arikfr:
1. We rely on the browser interface for `<input type=“date”>`, which is not consistent between browsers. We should switch to a dedicated component.
2. We need support for setting a dynamic default value (“Current date”).

PR #2697, opened 2018-07-25
Date/Time parameters: dynamic default value (current date/time)

kravets-levko:
Issue getredash/redash#2642

Also fixes CSS issues introduced by Ant library and `babel-plugin-import`

### DIFF
--- a/client/app/components/parameter-settings.html
+++ b/client/app/components/parameter-settings.html
@@ -26,6 +26,12 @@
           Global
         </label>
       </div>
+      <div class="form-group" ng-if="['date', 'datetime-local', 'datetime-with-seconds'].indexOf($ctrl.parameter.type) >= 0">
+        <label>
+          <input type="checkbox" class="form-inline" ng-model="$ctrl.parameter.useCurrentDateTime">
+          Use Today/Now as default value if no other value is set
+        </label>
+      </div>
       <div class="form-group" ng-if="$ctrl.parameter.type === 'enum'">
         <label>Dropdown List Values (newline delimited)</label>
         <textarea class="form-control" rows="3" ng-model="$ctrl.parameter.enumOptions"></textarea>
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -1,9 +1,24 @@
 import moment from 'moment';
 import debug from 'debug';
 import Mustache from 'mustache';
-import { each, zipObject, isEmpty, map, filter, includes, union, uniq, has } from 'lodash';
+import {
+  each, zipObject, isEmpty, map, filter, includes, union, uniq, has,
+  isNull, isUndefined,
+} from 'lodash';
 
 const logger = debug('redash:services:query');
+
+const DATETIME_FORMATS = {
+  // eslint-disable-next-line quote-props
+  'date': 'YYYY-MM-DD',
+  'datetime-local': 'YYYY-MM-DD HH:mm',
+  'datetime-with-seconds': 'YYYY-MM-DD HH:mm:ss',
+};
+
+function normalizeNumericValue(value, defaultValue = null) {
+  const result = parseFloat(value);
+  return isFinite(result) ? result : defaultValue;
+}
 
 function collectParams(parts) {
   let parameters = [];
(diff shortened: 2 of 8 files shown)


client/app/components/parameter-settings.html
```diff
@@ -26,6 +26,12 @@ <h4 class="modal-title">{{$ctrl.parameter.name}}</h4>
           Global
         </label>
       </div>
+      <div class="form-group" ng-if="['date', 'datetime-local', 'datetime-with-seconds'].indexOf($ctrl.parameter.type) >= 0">
+        <label>
+          <input type="checkbox" class="form-inline" ng-model="$ctrl.parameter.useCurrentDateTime">
+          Use current date/time if no value set
```
arikfr: "Use Today/Now as default value if no other value is set."

client/app/services/query.js
```diff
@@ -36,6 +40,22 @@ class Parameter {
...
+  getValue() {
+    if (this.isEmpty && this.useCurrentDateTime) {
+      // eslint-disable-next-line default-case
+      switch (this.type) {
+        case 'date': return moment().format('YYYY-MM-DD');
+        case 'datetime-local': return moment().format('YYYY-MM-DD HH:mm');
+        case 'datetime-with-seconds': return moment().format('YYYY-MM-DD HH:mm:ss');
+      }
+    }
+    return this.isEmpty ? '' : this.value; // normalize empty value
```
arikfr: Wouldn't it be better to use `null` as the normalized value?

Consider that we have some non string values (like numbers).

client/app/services/query.js
```diff
@@ -130,7 +150,7 @@ class Parameters {
   }
 
   getMissing() {
-    return map(filter(this.get(), p => p.value === null || p.value === ''), i => i.title);
+    return map(filter(this.get(), p => p.getValue() === ''), i => i.title);
```
arikfr: Can you explain this please?

client/app/services/query.js
kravets-levko: Yes, right now I'm trying to simplify (or remove at all) `ngModel` getter/setter and move that logic to methods that can be used both with angulars `ng-model` and react components.

kravets-levko:
@arikfr I updated this PR and I'm pretty happy with the code. Please review it. Thanks!

client/app/services/query.js
```diff
@@ -24,44 +34,75 @@ class Parameter {
...
+      if (
+        includes(['date', 'datetime-local', 'datetime-with-seconds'], this.type) &&
+        this.useCurrentDateTime
+      ) {
+        return moment().format(DATETIME_FORMATS[this.type]);
+      }
+      return null; // normalize empty value
+    }
+    if (this.type === 'number') {
+      return parseInt(this.value, 10) || null; // normalize empty value
```
arikfr: Want to guess what will `parseInt(0, 10) || null` return? :)

client/app/services/query.js
```diff
@@ -24,44 +34,75 @@ class Parameter {
...
-      this.$$value = moment(this.value).toDate();
-    } else if (value && this.type === 'datetime-with-seconds') {
-      this.value = moment(value).format('YYYY-MM-DD HH:mm:ss');
-      this.$$value = moment(this.value).toDate();
+  setValue(value) {
+    if (includes(['date', 'datetime-local', 'datetime-with-seconds'], this.type)) {
+      value = moment(value);
+      if (value.isValid()) {
+        this.value = value.format(DATETIME_FORMATS[this.type]);
+        this.$$value = value.toDate();
```
arikfr: I wonder if we even need this anymore... Does Ant's datetime picker expects a date object?

kravets-levko: It expects `moment` instance, but I cannot set `PropTypes` for this

client/app/services/query.js
kravets-levko: Oooops... thanks you, will fix

client/app/services/query.js
kravets-levko: I'll try to learn how custom validators work and replace this with `moment` instance

kravets-levko:
@arikfr Updated PR, could you review again, please?

client/app/services/query.js
```diff
@@ -24,44 +39,75 @@ class Parameter {
...
-      this.$$value = this.$$value || parseInt(this.value, 10);
-      return this.$$value;
-    }
+  get isEmptyValue() {
+    return isNull(this.value) || isUndefined(this.value) || (this.value === '');
+  }
+
+  get isEmpty() {
+    // `isEmptyValue` may return true, but parameter may actually be not empty -
+    // in example, date/time params may have default value
```
arikfr: We use `isEmptyValue` only in `getValue`, so maybe just inline the logic there to avoid two similarly named functions?

client/app/services/query.js
```diff
@@ -1,10 +1,25 @@
...
 
+const DATETIME_FORMATS = {
+  // eslint-disable-next-line quote-props
+  'date': 'YYYY-MM-DD',
+  'datetime-local': 'YYYY-MM-DD HH:mm',
+  'datetime-with-seconds': 'YYYY-MM-DD HH:mm:ss',
+};
+
+function normalizeNumericValue(value, defaultValue = null) {
+  const result = parseInt(value, 10);
```
arikfr: Number typed parameter can be a float as well. I'm not sure why we used `parseInt` there in the first place, but maybe it was a non issue because we didn't use `ngModel` with input of type number. But worth fixing this now. Is there a reason to keep this parsing?

kravets-levko:
@arikfr Updated PR (seems there are no issues with replacing `parseInt` with `parseFloat`, so I did it). Please review once more. Thanks!

arikfr:
🍾

denisov-vlad:
@kravets-levko @arikfr is it possible to schedule query with dynamic default value?

PR #2710, opened 2018-07-31
Design/ant to redash

kocsmy:
This PR adjust some smaller design details so Ant components fit well into our app:

Data-time component:
screenshot 2018-07-31 14 34 03

### DIFF
--- a/client/app/assets/less/main.less
+++ b/client/app/assets/less/main.less
@@ -1,3 +1,8 @@
+@import '~antd/lib/style/core/iconfont.less';
+@import '~antd/lib/input/style/index.less';
+@import '~antd/lib/date-picker/style/index.less';
+@import 'redash/redash-ant';
+
 /** LESS Plugins **/
 @import 'inc/less-plugins/for';
 
@@ -76,3 +81,6 @@
 @import 'redash/redash-newstyle';
 @import 'redash/redash-table';
 @import 'redash/query';
+
+
+
--- a/client/app/assets/less/redash/redash-ant.less
+++ b/client/app/assets/less/redash/redash-ant.less
@@ -7,3 +7,4 @@
 @border-color-base      : #e8e8e8;
 
 @primary-color          : #2196F3;
+@input-color            : #6d6d6d;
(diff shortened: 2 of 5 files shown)


client/app/assets/less/main.less
```diff
@@ -1,3 +1,8 @@
+@import '~antd/lib/style/core/iconfont.less';
+@import '~antd/lib/input/style/index.less';
+@import '~antd/lib/date-picker/style/index.less';
+@import 'redash/redash-ant';
```
arikfr: It's better to keep the component specific stylesheets imports next to the components, so if we remove them we remove everything together.

kocsmy: Ok, if I move it back to the component, what's the correct path for 'redash/redash-ant'; ?

kravets-levko: I think we should keep it where it is, and import per-component:
```javascript
import 'antd/lib/style/core/iconfont.less';
import 'antd/lib/input/style/index.less';
import 'antd/lib/date-picker/style/index.less';
import '@/assets/redash/redash-ant.less';
```

@arikfr - WDYT?

arikfr: Is the order of imports important? Because of not, we can keep `redash-ant.less` in `main.less` to avoid repeating it everywhere.

arikfr: Also, it's a small thing, but `redash/redash-ant.less` feels repetitive... I think that naming it `ant.less` is enough.

kocsmy: redash-ant.less is renamed to ant.less, makes sense.

The order of imports seems to be important otherwise it won't work.

kocsmy: The order of import is even more important:

import '@/assets/less/redash/ant.less';
import 'antd/lib/style/core/iconfont.less';
import 'antd/lib/input/style/index.less';
import 'antd/lib/date-picker/style/index.less';

This works 👌

kocsmy: Ok, I tried these things and a variety of others but those won't work. The only thing worked when it was placed in the first lines of main.less. Maybe I'm missing something?

client/app/assets/less/redash/ant.less
```diff
@@ -0,0 +1,9 @@
+// Overwritting Ant Design defaults to fit into Redash current style
+@font-family-no-number  : -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif !important;
+@font-family            : -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif !important;
+@code-family            : -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif !important;
```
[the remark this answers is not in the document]
kocsmy: I used variables as you suggested but our styling in general requires a bigger maintenance and simplification.

Issue #2643, opened 2018-06-28
Date Range Type Parameter


arikfr:
We need support for adding a _Date Range_ typed parameter. The tricky part here is that date ranges have two values and not only one.

We should use [this](http://www.daterangepicker.com/) or similar as the UI. What I really like in this picker is the predefined date ranges.

nadavivry:
I would suggest looking into "Ant Desing" design system which comes with a wide set of components, including date-picker components (I recently reviewed a few design systems in order to find a general solution to my own app and chose this one).

You can take a look at multiple examples, including date-range selection examples here:
https://ant.design/components/date-picker/

Note that they support predefined date ranges as well ("Presetted Ranges").

Besides date-range picker and date-picker in general, I think that Ant Design components set answer most of the use cases for Redash, whether for query parameters or other parts.

Now that we can implement React components in Redash, it can work smoothly.

arikfr:
Thanks for the tip, @nadavivry.

Do you know if we were to use a single component from Ant, is there some extra baggage that comes from it being a full design system, or each component is standalone?

nadavivry:
According to the "Manually import" example in the following page, looks like it won't be a problem to import each component independently and avoid the extra baggage:
https://ant.design/docs/react/introduce#Use-modularized-antd

In any case, using Webpack with "Tree shaking", you shouldn't be worried of an extra baggage at all (in this case or others).

arikfr:
Cool! It's definitely one of the better options I've seen so far and at least some of the other components will be useful too.

PR #2712, opened 2018-08-01
Date/Time Range parameter type

kravets-levko:
Issue getredash/redash#2643

Also fixes minor issues with Date/Time inputs (trigger digest cycle on value change).

Screenshots:

image

image

image

### DIFF
--- a/client/app/components/parameter-settings.html
+++ b/client/app/components/parameter-settings.html
@@ -29,9 +29,14 @@
           <option value="number">Number</option>
           <option value="enum">Dropdown List</option>
           <option value="query">Query Based Dropdown List</option>
+          <option disabled>--------------------------------------------------</option>
           <option value="date">Date</option>
           <option value="datetime-local">Date and Time</option>
           <option value="datetime-with-seconds">Date and Time (with seconds)</option>
+          <option disabled>--------------------------------------------------</option>
+          <option value="date-range">Date Range</option>
+          <option value="datetime-range">Date and Time Range</option>
+          <option value="datetime-range-with-seconds">Date and Time Range (with seconds)</option>
         </select>
       </div>
       <div class="form-group">
--- a/client/app/components/parameters.js
+++ b/client/app/components/parameters.js
@@ -1,7 +1,8 @@
-import { find, includes, words, capitalize } from 'lodash';
+import { find, includes, words, capitalize, extend } from 'lodash';
 import template from './parameters.html';
 import queryBasedParameterTemplate from './query-based-parameter.html';
 import parameterSettingsTemplate from './parameter-settings.html';
+import parameterInputTemplate from './parameter-input.html';
 
 function humanize(str) {
   return capitalize(words(str).join(' '));
@@ -135,22 +136,14 @@ function ParametersDirective($location, $uibModal) {
           if (scope.changed) {
             scope.changed({});
           }
+          const params = {};
           scope.parameters.forEach((param) => {
-            if (param.value !== null || param.value !== '') {
-              $location.search(`p_${param.name}`, param.value);
-            }
+            extend(params, param.toUrlParams());
           });
+          $location.search(params);
         }, true);
       }
 
-      // These are input as newline delimited values,
-      // so we split them here.
-      scope.extractEnumOptions = (enumOptions) => {
-        if (enumOptions) {
-          return enumOptions.split('\n');
-        }
-        return [];
-      };
       scope.showParameterSettings = (param) => {
         $uibModal.open({
           component: 'parameterSettings',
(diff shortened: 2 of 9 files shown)


client/app/components/DateInput.jsx
```diff
@@ -10,24 +11,35 @@ import 'antd/lib/date-picker/style/index.less';
...
   const format = clientConfig.dateFormat || 'YYYY-MM-DD';
-  const defaultValue = moment(value, format);
+  const additionalAttributes = {};
+  if (value && value.isValid()) {
+    additionalAttributes.defaultValue = value;
+  }
+  if (isFunction(onSelect)) {
+    additionalAttributes.onChange = (v) => {
+      onSelect(v);
+      $scope.$applyAsync();
```
arikfr: I think it's better to keep Angular code outside of the React components as much as possible, to make it easier to transition.

Also, in this case -- all the different Date/Time inputs eventually call the same function (`param.setValue`), so it's even easier to introduce this change in a single place. Btw, I wouldn't do this in the `Parameter#setValue` method but rather have a `setValue` function in the `Parameters` component which takes care of this.

client/app/components/parameter-settings.html
```diff
@@ -29,9 +29,14 @@ <h4 class="modal-title">{{ $ctrl.isNewParameter ? 'Add Parameter' : $ctrl.parame
           <option value="number">Number</option>
           <option value="enum">Dropdown List</option>
           <option value="query">Query Based Dropdown List</option>
+          <option disabled>--------------------------------------------------</option>
           <option value="date">Date</option>
           <option value="datetime-local">Date and Time</option>
           <option value="datetime-with-seconds">Date and Time (with seconds)</option>
+          <option disabled>--------------------------------------------------</option>
+          <option value="date-range">Date Range</option>
+          <option value="datetime-range-local">Date and Time Range</option>
```
arikfr: The `-local` suffix is the result of laziness on my part (we were using the parameter type as the `<input>` tag type). No need to pass it as legacy to new types :)

`datetime-range` is enough.

client/app/services/query.js
```diff
@@ -103,6 +136,39 @@ class Parameter {
...
+      this.setValue([
+        query[`p_${this.name}.start`],
+        query[`p_${this.name}.end`],
+      ]);
+    } else {
+      this.setValue(query[`p_${this.name}`]);
+    }
+  }
+
+  toQuerySnippet() {
```
arikfr: We already have a concept named Query Snippet in the code base, so it might be confusing to have the same name here. Maybe `toQueryText`?

client/app/services/query.js
```diff
@@ -103,6 +136,39 @@ class Parameter {
   set ngModel(value) {
     this.setValue(value);
   }
+
+  toQueryParams() {
```
arikfr: This is a bit of a confusing name, maybe `toUrlQueryStringParams`? (same with the `from` variant of the function).

kravets-levko:
@arikfr I'm happy that most of the comment were about naming the things :smile: Updated PR according to  notices, please review again. Thanks!

client/app/components/parameters.js
```diff
@@ -135,14 +135,44 @@ function ParametersDirective($location, $uibModal) {
...
+            param.setValue(value);
+            scope.$applyAsync();
+          } else {
+            // Delete cached item if parameter with such name does not exist anymore
+            delete paramValueSetters[name];
+          }
+        });
+        paramValueSetters[name] = result;
+        return result;
+      };
```
arikfr: 😮 

How about we wrap the parameter input rendering in its own component (`<ParameterInput param=param>`), then we can have a "static" callback, that already has reference to the parameter and can call `$applyAsync` along with `setValue`?

kravets-levko: Good idea. Should it be React or Angular component? :smile:

arikfr: :-) Ideally React so we don't have to rewrite later on, but it's main purpose as glue for Angular, so can be Angular too...

kravets-levko: Let it be Angular component - currently it's more "angular" than "react": we still need "ngModel" there (ad I don't want to refactor it now), we will need "$scope" there and so on.

kravets-levko: @arikfr Done. Also noticed and fixed small bug

client/app/pages/queries/source-view.js
```diff
@@ -96,9 +96,10 @@ function QuerySourceCtrl(
         },
       })
       .result.then((param) => {
-        $rootScope.$broadcast('query-editor.command', 'paste', '{{ ' + param.name + ' }}');
-        $rootScope.$broadcast('query-editor.command', 'focus');
         $scope.query.getParameters().add(param);
+        param = $scope.query.getParameters().get(param.name);
```
arikfr: Is it different than what this function gets as an argument?

kravets-levko: Yes. An argument is just a plain object, but here I need a `Parameter` instance, and `.getParameters().get(name)` returns it

kravets-levko: @arikfr I changed this fragment a bit: I decided that adding an argument to `.getParameters().get()` just to use it in a single place wasn't a good idea, so `.getParameters().add(...)` now returns newly added `Parameter` instance.

arikfr:
🤩

PR #2717, opened 2018-08-02
Create TEMPLATE_PATH variable which gets value from new REDASH_TEMPLA…

MrCirca:
Hello,
I have setup redash in **python virtual environment** with **uwsgi** instead of gunicorn.
Doing many setup deployments i got an error.
Templates directory could not be found.
So i was searching about this problem in source code. I was checking the REDASH_STATIC_ASSETS envar changing the path consecutively .
 
So i had two choices:

  In handlers/base.py

1. **routes = Blueprint('redash', __name__, template_folder=settings.fix_assets_path('templates'))**
     to
    **routes = Blueprint('redash', __name__, template_folder=settings.fix_assets_path('/opt/redash/dist/redash/templates'))** (bad hardcoded choice)

2. Create variable TEMPLATES_PATH in settings/__init__.py
  **TEMPLATE_PATH = fix_assets_path(os.environ.get("REDASH_TEMPLATE_PATH", "../client/dist"))** which  does not affect the deployment from bash script, because if is defined as envar, will get the same value with STATIC_ASSETS_PATH

If you want we could discuss about it and find a solution.

Thank you

### DIFF
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -100,7 +100,7 @@ def create_app(load_admin=True):
     from redash.metrics.request import provision_app
 
     app = Flask(__name__,
-                template_folder=settings.STATIC_ASSETS_PATH,
+                template_folder=settings.TEMPLATES_PATH,
                 static_folder=settings.STATIC_ASSETS_PATH,
                 static_path='/static')
 
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -99,7 +99,7 @@ LDAP_SEARCH_DN = os.environ.get('REDASH_LDAP_SEARCH_DN', os.environ.get('REDASH_
 
 STATIC_ASSETS_PATH = fix_assets_path(os.environ.get("REDASH_STATIC_ASSETS_PATH", "../client/dist/"))
 
-TEMPLATES_PATH = fix_assets_path(os.environ.get("REDASH_TEMPLATES_PATH", "../client/dist/"))
+TEMPLATES_PATH = fix_assets_path(os.environ.get("REDASH_TEMPLATES_PATH", STATIC_ASSETS_PATH))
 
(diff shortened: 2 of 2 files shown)


arikfr:
It's not clear to me why you needed this option while it works properly without for other users?

Maybe uwsgi doens't set the current path correctly?

arikfr:
Why are you setting `REDASH_STATIC_ASSETS_PATH=/opt/redash/dist/client/dist/` in your uwsgi config?

Also, I just noticed that this pull request doesn't change the templates location for the `routes` Blueprint, so it not supposed even to fix the issue you're having.

My suggestion is that you add some logging to understand what path it currently resolves to, then it will help us understand why it can't find the templates (and probably what configuration we should adjust).

MrCirca:
I set REDASH_STATIC_ASSETS_PATH to do tries, changing path. But doesnt work too. By default, if i dont set REDASH_STATIC_ASSETS_PATH, and get the same error. Template not found. Yes, this pull request does not change something, but both template_folder and assets are related in flask application in  [__init_file__](https://github.com/getredash/redash/blob/master/redash/__init__.py) 102-105 lines. So i thought to separate assets path with templates path...So i will try logging them and come to a better conclusion

MrCirca:
I fixed it. It was my mistake.
```
[uwsgi]
socket = 127.0.0.1:3031
plugin = python
wsgi-file = /opt/redash/dist/redash/wsgi.py
callable = app
home = /opt/redash/pyenv
env = REDASH_LOG_LEVEL=INFO
env = REDASH_REDIS_URL=redis://localhost:6379/0
env = REDASH_DATABASE_URL=postgresql:///redash
env = REDASH_COOKIE_SECRET=Hn5FVC6NhTJKSYBMLry9NA6DEhr1shpg
pythonpath = /opt/redash/dist/
processes = 4
uid = redash
gid = 1000
```
I removed **chdir** option , and add **pythonpath** option.

PR #2719, opened 2018-08-04
Remove closing input tags

maxv:
>  Must have a start tag and must not have an end tag.

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input


arikfr:
Thanks.

PR #2720, opened 2018-08-05
Fix #2718: query search is broken.

### LINKED ISSUE #2718: Cannot select a query to use for a new alert or search queries when adding a widget to a dashboard
### Issue Summary

When creating a new alert, the query search doesn't find any queries to create alerts from. Similarly, the query search in "Add Widget" dialog of a dashboard finds no results.

### Steps to Reproduce

1. Create a query named "Alert"
2. Go to the "New Alert" page
3. Try to search for the query you just created ("Alert")

When the search happens, the following error is logged to the web console:

> Error: "[$resource:badcfg] Error in resource configuration for action `search`. Expected response to contain an array but got an object (Request: get api/queries/search)

It seems that #2686 changed the response format for the query search, but the model (https://github.com/getredash/redash/blob/a014df36d86ab78974befe979d68c9ef5378634b/client/app/services/query.js#L290-L295) and (some) callers of that .search() method were not converted to use the new API.

### Technical details:

* Redash Version: 5.0.0-beta (a014df3)
* Browser/OS: Firefox 61 / Linux
* How did you install Redash: Built from source

arikfr:


jezdez:
Yikes, good call! That was one last todo item in my #2686 PR. Thanks!

PR #2736, opened 2018-08-10
Remove X at the end of the input on Queries list page

kocsmy:
Before:
screenshot 2018-08-10 17 06 12

After:
screenshot 2018-08-10 17 12 25

I don't think this extra clean option adds anything to the interface other than unnecessary complexity.

### DIFF
--- a/client/app/pages/queries-list/queries-list.html
+++ b/client/app/pages/queries-list/queries-list.html
@@ -3,7 +3,7 @@
 
   <div class="row">
     <div class="col-md-3 list-control-t">
-      <div class="m-b-5">
+      <div class="m-b-10">
         <input type="text" class="form-control" placeholder="Search Queries..." autofocus ng-model="$ctrl.term" ng-model-options="{ allowInvalid: true, debounce: 200 }"
           ng-change="$ctrl.update()">
       </div>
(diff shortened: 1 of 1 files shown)


jezdez:
@kocsmy If it helps to understand the reasoning for me to add this:

When searching we see immediate in-page updates when pressing enter in the search field, but have to do multiple user interactions to get to the previous state of listing all queries again: selecting all, deleting, enter.

Having a separate user interface element serves as a direct interaction point to do that instead and I think it has precedence on some platforms such as iOS. It simply allows users to get back to the orginal state of the listing similar to how we allow selecting and deselecting tags.

Alternatives to the removal of the X could be:

- add a keyboard shortcut the allow pressing the `Escape` key when the search input is selected
- show the X only when search input is selected

kocsmy:
Thanks for sharing that, @jezdez but I still argue we need this.

Show the X only when search input is used and has value — this might be a good middle ground but we'll need to make the interface better around the X (the non-connecting border bothers me).

jezdez:
@kocsmy Sure, sweating the details is important, no doubt about that, and it's definitely a minor use case.

In terms of design language, there is also an inconsistency with the search input in the header (also visible in your screenshots) which doesn't have the X and behaves differently because of that.

kocsmy:
Exactly @jezdez — and that also works differently because you have to hit Enter or use the button to initiate the search. And they are so close to each other that this might be more confusing.

PR #2737, opened 2018-08-10
feat: support hiding Redshift tables from schema list

sjakthol:
These changes make it possible to configure a Redshift data source so
that certain tables are hidden from users in the schema browser.

This feature is useful because the Redshift SVV_COLUMNS table used to
retrieve the schema returns schema for every single table in the
database. This means that the schema for a Redshift data source also
includes tables that cannot be accessed from Redash. This clutters the
schema browser and confuses users of Redshift data sources.

The exclude filter can be used to manually hide the tables that are not
accessible from Redash. The filter is evaluated as a regular expression
which makes it rather flexible for various cases.

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -2,6 +2,7 @@ import os
 import json
 import logging
 import select
+import re
 
 import psycopg2
 
@@ -94,11 +95,17 @@ class PostgreSQL(BaseSQLQueryRunner):
 
         results = json.loads(results)
 
+        hidden_tables = self.configuration.get('hidden_tables')
+        hidden_tables_re = re.compile(hidden_tables) if hidden_tables else None
+
         for row in results['rows']:
             if row['table_schema'] != 'public':
                 table_name = '{}.{}'.format(row['table_schema'], row['table_name'])
             else:
                 table_name = row['table_name']
+
+            if hidden_tables_re and hidden_tables_re.match(table_name):
+                continue
 
             if table_name not in schema:
                 schema[table_name] = {'name': table_name, 'columns': []}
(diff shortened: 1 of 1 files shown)


arikfr:
Being able to filter the schema showed to the user is definitely something we would like to support. But I'm not sure that adding an adhoc support for this in a specific query runner is the right thing to do. :-\

About the actual issue you're trying to solve:

> This means that the schema for a Redshift data source also includes tables that cannot be accessed from Redash.

Maybe there is a way to query Redshift for only tables the user has access to?

sjakthol:
Thanks for the comment. I know this is an awful implementeation and I can already think of a second data source where similar adhoc filtering would be needed (Athena).

In theory, Redshift has an UDF called [HAS_TABLE_PRIVILEGE](https://docs.aws.amazon.com/redshift/latest/dg/r_HAS_TABLE_PRIVILEGE.html) which would be able to provide this information. The problem is that the function is a [leader node-only function](https://docs.aws.amazon.com/redshift/latest/dg/c_SQL_functions_leader_node_only.html) which means the query the function is used in can ONLY reference pg_ tables. Since the [query](https://github.com/getredash/redash/blob/master/redash/query_runner/pg.py#L252) Redash uses to determine the table schemas refer to a table called svv_columns, the query is not a leader node-only query and hence HAS_TABLE_PRIVILEGE cannot be used in that query. I haven't tested this in practice but the this is what the documentation indicates.

Alternatives (as far as I know) are either 1) parsing relacl strings from [pg_class](https://www.postgresql.org/docs/9.3/static/catalog-pg-class.html) table to determine who is allowed to access a table (quite complex) or 2) trying out if the user is allowed to run a query against a table (bad performance).

Anyway, I'll scrap this PR and see if I could come up with a way to somehow do this automatically in Redshift.

PR #2738, opened 2018-08-10
Added query.query_text fix and updated pyhive from 0.3.0 to 0.6.0

nasmithan:
Model.py had class Query's to_dict function removed back in May. See slack conversation with myself here: https://redash.slack.com/archives/C0H3V4RMW/p1533923229000319


jezdez:
Could you please split the update to the dependency into a separate commit and pull request?

Issue #2302, opened 2018-02-09
[mysql] Query can not cancel. worker process runaway


dozen:
### Issue Summary

MySQL query can not cancel.

When canceling a query whose data source is MySQL, celery runaway.

### Steps to Reproduce

1. Execute time-consuming queries
example query: 
```
SELECT SLEEP(10);
```

2. Push the Cancel button. Return error `Error running query: terminated`
![image](https://user-images.githubusercontent.com/802038/36024850-5feece58-0dd4-11e8-9cea-3d2380776dec.png)

3. worker process runaway. CPU usage exhaust.
```
top - 11:01:06 up 1 day,  4:18,  0 users,  load average: 0.78, 0.45, 0.38
Tasks:   5 total,   2 running,   3 sleeping,   0 stopped,   0 zombie
%Cpu(s): 28.2 us,  3.9 sy,  0.0 ni, 67.9 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem :  6100252 total,  1165948 free,   866780 used,  4067524 buff/cache
KiB Swap:  1048572 total,  1048572 free,        0 used.  4833908 avail Mem 

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                                                  
   12 redash    20   0  314296 126064   8184 R  99.7  2.1   0:17.35 celery                                                                                   
    1 redash    20   0  289876 119336  15844 S   0.3  2.0   0:07.48 celery                                                                                   
   13 redash    20   0  299428 111200   7464 S   0.0  1.8   0:01.18 celery                                                                                   
   18 redash    20   0   36640   3104   2676 R   0.0  0.1   0:00.06 top                                                                                      
   25 redash    20   0  308192 124056   6496 S   0.0  2.0   0:00.96 celery 
```

It seems that the process will run out of control if [SIGINT is sent to the worker process](https://github.com/getredash/redash/blob/master/redash/tasks/queries.py#L205), since it is not possible to trap KeyboardInterrupt while running [MySQLdb.execute()](https://github.com/getredash/redash/blob/master/redash/query_runner/mysql.py#L144).

In order to avoid this, need to execute MySQLdb.execute () as a sub thread using threading module, trap SIGINT on the main thread, and cancel the query with the MySQL KILL statement.

[...]

PR #2739, opened 2018-08-10
[MySQL] Query cancelling correctly

dozen:
related: #2302

1. change Mysql-python to PyMySQL
  libmysql is can't handling SIGINT. So, use pure python implementation.
2. kill MySQL thread by another connection when catch SIGINT.

`_cancel` may be ad hoc, but for the time being we can cancel the query of mysql correctly.


arikfr:
Thank you for the input @Jakdaw , I reviewed the original issue and @dozen's alternative implementation that uses threads. And honestly still not too excited to add threads to the mix.

> Please don't merge this - shifting to PyMySQL breaks support for MariaDB's 10.3 feature "Compound Statements Outside of Stored Programs"; see:

Considering that Redash is a tool to query and visualize data, and less for managing the schema -- is there a reasonable use case for compound statements?

> Additionally it breaks putting multiple queries into a single redash query; like:-
set @param = if("{{redashparam}}"="","default","{{redashparam}}"); select * from table where p = @param;

Why does it break them? Maybe we need to use a different API?

Jakdaw:
> Thank you for the input @Jakdaw , I reviewed the original issue and @dozen's alternative implementation that uses threads. And honestly still not too excited to add threads to the mix.

Why the fear of threads? This is starting one thread in the celery worker for the duration of the query and ensures that it's all cleaned up before the celery worker returns to the pool.

I'm sure that there are threads being created in some of the libraries used for other data sources (certainly Cassandra comes to mind, probably others) that Redash uses - why is this any different?

> > Please don't merge this - shifting to PyMySQL breaks support for MariaDB's 10.3 feature "Compound Statements Outside of Stored Programs"; see:
> 
> Considering that Redash is a tool to query and visualize data, and less for managing the schema -- is there a reasonable use case for compound statements?

Compound statements are not for managing the schema. They simply allow you to do some of the things that you might otherwise need to do in a stored procedure in a query - so you can, for instance, open a cursor (based on redash query parameters) and iterate over the entries to collate information from further queries (based on redash query parameters) into a temporary table to display in the Redash UI. All without having to first define any stored procedures in the DB engine.

> > Additionally it breaks putting multiple queries into a single redash query; like:-
> > set @param = if("{{redashparam}}"="","default","{{redashparam}}"); select * from table where p = @param;
> 
> Why does it break them? Maybe we need to use a different API?

I'm not sure - I've not dug into the details of what it's sending to the DB and how that differs from the C-library - I'll take a look. I believe it complained that everything after the first semicolon was unexpected. There may be a way to support it with pymysql; I've not checked, just observed that there are lots of queries that work with MySQLdb that don't work under this pull request. I'll have a dig and see whether I can find some more details.

arikfr:
> Why the fear of threads? This is starting one thread in the celery worker for the duration of the query and ensures that it's all cleaned up before the celery worker returns to the pool.

It's possible I'm a bit driven by FUD, more than trying to keep things simple 😳 

> I'm sure that there are threads being created in some of the libraries used for other data sources (certainly Cassandra comes to mind, probably others) that Redash uses - why is this any different?

True about Cassandra and I indeed saw cases where the thread it starts keep running even after the query was done.

Anyway if you happen to find a simple way to make the pure Python implementation to work -> great. If not, then let's give the thread implementation a try.

hoangphuoc25:
Can we get this merged in? We are facing the same problem with redash. Having a temporary solution is better is still better than nothing @arikfr

Jakdaw:
I agree that something should really be done about the ability to cancel queries but don't think that this PR is the answer as it breaks almost every query I have (which either have compound statements, multiple statements, or call stored procedures). @arikfr I'll rebase PR # 3003 against master and add the cancelling (yes, with a thread) functionality to that.

That way the MySQL query runner can cancel queries and properly support compound statements, multiple statements and stored procedures. This is what I've been running on my self-hosted instance for 6 months or so now.

arikfr:
@Jakdaw's pull request (#3003) was merged, so I'm closing this one. Thank you, @dozen for the effort.

PR #2741, opened 2018-08-10
fix broken query selection in alert page

sjakthol:
The following error occurs when a query is selected in the
alert page:

> TypeError: "item.getQueryResultPromise is not a function"

This is because .onQueryHandler() is given a plain object instead of a
Query when a query is selected from the dropdown.

These changes fix the issue by wrapping the plain object into a Query
object before calling .getQueryResultPromise() on it.


nasmithan:
Awesome thanks sjakthol! I noticed this was broken this morning as well.

jezdez:
Nice, thanks for the fix!

PR #2742, opened 2018-08-10
Fix query_results for Embedded Parameters (removed deprecated to_dict function)

nasmithan:
Model.py had class Query's to_dict function removed back in May. See slack conversation with myself here: https://redash.slack.com/archives/C0H3V4RMW/p1533923229000319


jezdez:
Thanks!

Issue #2744, opened 2018-08-11
Sort is not working properly when one of the cells in that column is undefined


Yossi-a:
### Issue Summary

When a column has a cell with no value, the sort in the table visualisation doesn't work properly.

### Steps to Reproduce

1. Insert rows to a datasource when one of the fields is missing:

db.getCollection('test').insertMany([
{ 'a': 11, 'b':21 },
{ 'a': 12 }, // b is missing
{ 'a': 13, 'b':23 }
])

2. Create query which show the data:
{
    "collection": "test",
    "fields": {
        "a": 1,
        "b": 1
    }
}

3. Click on the 'b' column to sort by it. 

4. The sort seems to be random.

### Expected result

In my opinion ascending sort in this case should place the undefined values first and result in: [row2, row1, row3] while descending sort ascending sort should place the undefined values last and result in: [row3, row1, row2]

### Technical details:

* Redash Version: 4.0.1, 5.0.0 beta
* Browser/OS: Chrome
* How did you install Redash: docker-compose

arikfr:
Thank you for the detailed bug report, but this is actually a duplicate of #2486.

PR #2745, opened 2018-08-11
Sort columns with undefined values

Yossi-a:
Issue #2744 

When a query has column with undefined value, the column sorting result in a random order.
This fix place the undefined values in the first or last place depends on the sort direction.

### DIFF
--- a/client/app/components/dynamic-table/index.js
+++ b/client/app/components/dynamic-table/index.js
@@ -34,11 +34,11 @@ function sortRows(rows, orderBy) {
     for (let i = 0; i < orderBy.length; i += 1) {
       va = a[orderBy[i].name];
       vb = b[orderBy[i].name];
-      if (!va || (va < vb)) {
+      if (va == undefined || (va < vb)) {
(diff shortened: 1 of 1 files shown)


client/app/components/dynamic-table/index.js
```diff
@@ -34,11 +34,11 @@ function sortRows(rows, orderBy) {
     for (let i = 0; i < orderBy.length; i += 1) {
       va = a[orderBy[i].name];
       vb = b[orderBy[i].name];
-      if (va < vb) {
+      if (!va || (va < vb)) {
         // if a < b - we should return -1, but take in account direction
         return orderBy[i].direction * -1;
       }
-      if (va > vb) {
+      if ((va > vb) || !vb) {
```
arikfr: I think you need explicit `null`/`undefined` check here instead of `!vb`/`!va`, because `!0 == !'' == true` in JavaScript...

Yossi-a:
Should be better now

arikfr:
Thanks!

PR #2746, opened 2018-08-12
port making autocomplete for large schemas faster

alison985:
Addresses https://github.com/mozilla/redash/issues/454 which fixes https://github.com/mozilla/redash/issues/232


client/app/components/queries/query-editor.js
```diff
@@ -119,31 +121,43 @@ function queryEditor(QuerySnippet, $timeout) {
 
         const schemaCompleter = {
           getCompletions(state, session, pos, prefix, callback) {
-            if (prefix.length === 0 || !$scope.schema) {
+            // make a variable for the auto completion in the query editor
+            $scope.autoCompleteSchema = $scope.schema; // removeExtraSchemaInfo(
```
jezdez: Could you elaborate what this does?

client/app/components/queries/query-editor.js
```diff
@@ -119,31 +121,43 @@ function queryEditor(QuerySnippet, $timeout) {
...
-                table.columns.forEach((c) => {
-                  keywords[c] = 'Column';
+                table.columns.forEach((c) => { // autoCompleteColumns
+                  if (c.charAt(c.length - 1) === ')') {
+                    let parensStartAt = c.indexOf('(') - 1;
+                    c = c.substring(0, parensStartAt);
+                    parensStartAt = 1; // linter complains without this line
+                  }
+                  // remove '[P] ' for partition keys
+                  if (c.charAt(0) === '[') {
```
jezdez: I don't think we have [P] in this code base yet.

client/app/components/queries/query-editor.js
```diff
@@ -119,31 +121,43 @@ function queryEditor(QuerySnippet, $timeout) {
...
+            if (!$scope.autoCompleteSchema.keywords) {
               const keywords = {};
 
-              $scope.schema.forEach((table) => {
+              $scope.autoCompleteSchema.forEach((table) => {
                 keywords[table.name] = 'Table';
 
-                table.columns.forEach((c) => {
-                  keywords[c] = 'Column';
+                table.columns.forEach((c) => { // autoCompleteColumns
```
jezdez: Could you leave a code comment what this does?

client/app/components/queries/query-editor.js
```diff
@@ -119,31 +121,43 @@ function queryEditor(QuerySnippet, $timeout) {
...
+                  if (c.charAt(c.length - 1) === ')') {
+                    let parensStartAt = c.indexOf('(') - 1;
+                    c = c.substring(0, parensStartAt);
+                    parensStartAt = 1; // linter complains without this line
+                  }
+                  // remove '[P] ' for partition keys
+                  if (c.charAt(0) === '[') {
+                    c = c.substring(4, c.length);
+                  }
+                  // keywords[c] = 'Column'; // dups columns
```
jezdez: Why the commented out line here?

client/app/components/queries/query-editor.js
alison985: Correct, but that is a pending PR and because it's an `if` statement it doesn't hurt anything prior to that PR being merged.

client/app/components/queries/query-editor.js
alison985: This is essentially my note to self that `keywords[c]` will duplicate columns in the list, that's why I had to make it `keywords[`${table.name}.${c}`]`. This will prevent me from trying to change it back later to reduce complexity and create a bug.

client/app/components/queries/query-editor.js
alison985: It essentially separates the interface schema (`schema`) and the auto complete schema (`autoCompleteSchema`) to handle UI differences between them.

client/app/components/queries/query-editor.js
```diff
@@ -103,8 +103,10 @@ function queryEditor(QuerySnippet, $timeout) {
                 // as it makes typing slower.
                 if (tokensCount > 5000) {
                   editor.setOption('enableLiveAutocompletion', false);
+                  editor.setOption('enableBasicAutocompletion', false);
                 } else {
                   editor.setOption('enableLiveAutocompletion', true);
+                  editor.setOption('enableBasicAutocompletion', true);
```
alison985: @jezdez to answer your question about how this makes the performance faster, the key to the performance improvement is in these 2 lines about `enableBasicAutocompletion`. The text editor redash uses is Ace Editor and without these lines there were autocompletion calls happening via the default configuration through Ace Editor that were causing a performance issue. 

Yes, a lot of other things seem to have ended up in the commit that this PR is porting upstream.

client/app/components/queries/query-editor.js
jezdez: Hm, I see, I'm afraid this will need to be removed for now since I got feedback from Arik about #2764 being problematic in terms of UI design. I will follow-up on this in the PR there and ask the Redash UX people to weigh in on a way forward. In the meantime we shouldn't hold back this PR here.

client/app/components/queries/query-editor.js
jezdez: Looking further into this, this seems to actually just add an additional command to the editor (https://github.com/ajaxorg/ace/blob/51a9b7c74671b86dceab01296a7d09e877853484/lib/ace/ext/language_tools.js#L175-L203) that triggers the auto-complete menu when typing `ctrl+space` ([et al.](https://github.com/ajaxorg/ace/blob/51a9b7c74671b86dceab01296a7d09e877853484/lib/ace/autocomplete.js#L433-L445)). Since this is *in addition* to the live auto-completion (where "live" means that is happens on any input change of the editor), I don't see how this improves performance per-se. Instead this seems like a drive-by change that was shipped in the original code change in the Mozilla fork. Not a big deal, but just want to state that this is a thing.

client/app/components/queries/query-editor.js
jezdez: To be able to modify the colums further down below?

If we remove the `[P]` code since #2764 isn't going to merged adding it, then what leaves this PR us with?

client/app/components/queries/query-editor.js
```diff
@@ -119,31 +121,43 @@ function queryEditor(QuerySnippet, $timeout) {
...
                 keywords[table.name] = 'Table';
-
+                // remove special characters from display schema that aren't okay in sql syntax
                 table.columns.forEach((c) => {
-                  keywords[c] = 'Column';
+                  if (c.charAt(c.length - 1) === ')') {
+                    let parensStartAt = c.indexOf('(') - 1;
+                    c = c.substring(0, parensStartAt);
+                    parensStartAt = 1; // linter complains without this line
+                  }
```
jezdez: So this removes columns that end with a closing parenthesis at the point where the opening parenthesis is? Can you elaborate which data sources have those kind of colums?

jezdez:
@alison985 Before we merge this I think we need to further review what the changes done after https://github.com/mozilla/redash/issues/232 mean for this repo that doesn't have all the changes made in the fork.

client/app/components/queries/query-editor.js
alison985: Given all of this, I'd be okay with closing this PR and closing the issue to port this change upstream. What do you think @jezdez?

client/app/components/queries/query-editor.js
jezdez: I see, let's remove this then.

alison985:
See comment https://github.com/getredash/redash/pull/2746#discussion_r216154344

Jannis and I agreed to close this one. Some of the changes may be relevant in the future, but not until the related features are upstream.

client/app/components/queries/query-editor.js
jezdez: Yeah, let's revisit this if the follow-up for #2764 is merged.

PR #2747, opened 2018-08-13
add databricks query runner

### LINKED ISSUE #2685: Add Databricks query runner
### Issue Summary

[Databricks](https://databricks.com/) offers a [Hive-like API](https://docs.databricks.com/user-guide/bi/jdbc-odbc-bi.html) that could be used to connect Redash to it to query tables over there.

### Steps to Reproduce

Mozilla has worked on a [POC code based on the existing Hive query runner](https://gist.github.com/robotblake/d0eff602664ab62959772dbb4f0cca22
) (curtesy of @robotblake) that could be the base for the code.

alison985:
Fixes #2685 

This is based on @robotblake's work here: https://gist.github.com/robotblake/d0eff602664ab62959772dbb4f0cca22

The test suite passed locally but did have 6 warnings which seem completely unrelated to this new data source query runner. I do not have a databricks connection to test this with.


redash/query_runner/databricks.py
```diff
@@ -0,0 +1,125 @@
...
+            },
+            "required": ["host"]
+        }
+
+    @classmethod
+    def type(cls):
+        return "databricks"
+
+    def __init__(self, configuration):
+        super(DataBricks, self).__init__(configuration)
```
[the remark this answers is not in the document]
alison985: Is it okay that if I don't override it from the parent Hive class that it will start `super(Hive,`?

redash/query_runner/databricks.py
```diff
@@ -0,0 +1,125 @@
+import base64
+import * from hive_ds
+
+
+class DataBricks(Hive):
+
+    @classmethod
+    def configuration_schema(cls):
```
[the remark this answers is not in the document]
alison985: I understand what you're trying to get at but have questions about what the syntax for this would look like.

requirements_all_ds.txt
```diff
@@ -22,6 +22,7 @@ simple_salesforce==0.72.2
 PyAthena>=1.0.0
 pymapd>=0.2.1
 qds-sdk>=1.9.6
+pybase64==0.4.0
```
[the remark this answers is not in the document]
alison985: Ah ha - it being part of the stdlib makes more sense.

redash/query_runner/databricks.py
```diff
@@ -0,0 +1,125 @@
...
+
+                # if username or password is set, add Authorization header
+                username = self.configuration.get('username', '')
+                password = self.configuration.get('http_password', '')
+                if username || password:
+                    auth = base64.b64encode(username + ':' + password)
+                    transport.setCustomHeaders({'Authorization': 'Basic ' + auth})
+
+                # create connection
+                connection = hive.connect(thrift_transport=transport)
```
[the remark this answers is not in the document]
alison985: It appears to me from googling that Hive supports an HTTP API, and again I know what you're going for but have implementation questions.

alison985:
@jezdez there is also a CodeClimate error on this code. Specifically, that there is missing whitespace around an operator, but a literal interpretation of that error and the line it's erroring on makes no sense to me.

redash/query_runner/hive_ds.py
```diff
@@ -97,7 +114,42 @@ def run_query(self, query, user):
...
+                if path and path[0] != '/':
+                    path = '/' + path
+
+                # if port is set prepend colon
+                port = self.configuration.get('port', '')
+                if port:
+                    port = ':' + port
+
+                # build uri
+                http_uri = "{}://{}{}{}".format(scheme, host, port, path)
```
arikfr: Why build the URL and not take a URL in the configuration?

alison985: I'm not sure what you mean. The http_uri is used to create the transport on line 135.

alison985: If your question is: why not make the URL the input fields instead of each separate piece (i.e. host, port) as an input field, the answer is that having each piece of the string makes it easier to use each piece elsewhere if needed.

redash/query_runner/hive_ds.py
```diff
@@ -97,7 +114,42 @@ def run_query(self, query, user):
...
+
+                # create transport
+                transport = THttpClient.THttpClient(http_uri)
+
+                # if username or password is set, add Authorization header
+                username = self.configuration.get('username', '')
+                password = self.configuration.get('http_password', '')
+                if username | password:
+                    auth = base64.b64encode(username + ':' + password)
```
jezdez: The base64 library needs to be import on top of the file

redash/query_runner/databricks.py
```diff
@@ -0,0 +1,18 @@
+import base64
```
jezdez: The import that is needed in the other module :)

redash/query_runner/databricks.py
```diff
@@ -0,0 +1,18 @@
...
+from .hive_ds import Hive
+
+try:
+    from pyhive import hive
+    from thrift.transport import THttpClient
+    enabled = True
+except ImportError:
+    enabled = False
```
jezdez: You'll need to define an `enabled` class method to use this in the `DataBricks` query runner below to make use of this enabled variable.

```python
    @classmethod
    def enabled(cls):
        return enabled
```

redash/query_runner/hive_ds.py
```diff
@@ -97,7 +114,42 @@ def run_query(self, query, user):
...
+                    transport.setCustomHeaders({'Authorization': 'Basic ' + auth})
+
+                # create connection
+                connection = hive.connect(thrift_transport=transport, **self.configuration.to_dict())
```
jezdez: Passing the configuration values as keyword arguments won't work, since the [`Connection` class](https://github.com/dropbox/PyHive/blob/d19cb0c2e14fce1888e327755526e94c30cb407d/pyhive/hive.py#L97-L102) which the `hive.connect` function calls, doesn't support values such as `use_http`, `http_scheme`, etc.

So similar to the case in which `use_http` is `False` we'll need to build the parameters to pass to hive.connect individually. Since the actual difference between both calls is the existence of `thrift_transport` in the case of using `use_http` being `True`, let's simply build the parameters once and only update them with the `thrift_transport` value when needed.

redash/query_runner/databricks.py
alison985: The `Hive` class has this method and the `DataBricks` class inherits from the `Hive` class so I don't think I have to re-define it.

redash/query_runner/hive_ds.py
alison985: So there's a couple of things going on here.

1. Issue: @robotblake wrote [this code](https://gist.github.com/robotblake/d0eff602664ab62959772dbb4f0cca22) 6 months ago. 

Action: I went looking for the version of the PyHive code that Blake would have been looking at. I didn't find `thrift_transport` [supported](https://github.com/dropbox/PyHive/blob/a064a496001192bccc47bafdef819ed90ea2529f/pyhive/hive.py) until a July commit which makes me think Blake was working off of a different implementation. Because passing configuration variables of host, port, auth, kerberos_service_name, and/or password as this code does in `**self.configuration.to_dict()` will [raise a ValueError](https://github.com/dropbox/PyHive/blob/a064a496001192bccc47bafdef819ed90ea2529f/pyhive/hive.py#L98-L108) in the Connection class.

2. Issue: The `host` variable used in [line 132](https://github.com/getredash/redash/pull/2747/files#diff-2b97dc03b6ae2c7943fe42487818ade3R132) is, as far as I can tell, never assigned before being used there. 

Action: Add host assignment above that line. 

3. Issue: No support for `use_http` or `http_scheme` in the Connection class.

Action: don't pass `**self.configuration.to_dict()` when `use_http` is True.

4. Issue: Don't rebuild parameters twice.

There's actually only one parameter used in both in the exact same manner and that's `host`, which I've placed above the if statement. Please note the default values between the `use_http` is True case versus the `use_http` is False case are different for port and username. The other configuration values are only used in either one case or the other.

5. Issue: There's a typo in configuration in line 149 below.

Action: Fixed.

redash/query_runner/databricks.py
jezdez: You'll still need to implement the method in the subclass since the `Hive` class is defined in the scope of the `hive_ds` module, which will make the enabled `Hive.enabled` classmethod use the module level variable from there, instead of a redefinition in a separate module. In other words, the subclass doesn't know about the scope of the module of the parent class. Here's an example to show: https://trinket.io/python/f5d33dff2e

alison985:
@jezdez PR updated

jezdez:
Oh, @alison985 do you mind rebasing again?

alison985:
@jezdez There aren't any conflicts showing for me. Why would I rebase it?

jezdez:
The pull request can't be merged according to GitHub:

ukowyod2r6grutkgnt5vtpxgt47gwwch-medium

jezdez:
Nevermind, I've squashed the thing in https://github.com/getredash/redash/commit/28a8525ce3fb24d303132c52109b5d62c67fae4f.

arikfr:
@alison985 @jezdez each data source needs an accompanying logo [here](https://github.com/getredash/redash/tree/master/client/app/assets/images/db-logos).

Issue #2494, opened 2018-04-26
Widgets crumble when trying to move / resize a widget


Caloulebeauf:
### Issue Summary

I have experienced several times a crumble of the Widgets (they all stack up at the top of the dashboard) when editing a widget, either resizing or moving. It does not happen on the first edition actions but only at some point.

### Steps to Reproduce

On a dashboard with severals widgets of various size and types :
1. Move the widgets / change the size
2. At some point (sorry for not being more precise) every widget moves to the top of the dashboard and superpose with the others 
3. A clic on Cancel moves everything back to normal

![9d8ef6371489ac67a611a6d38da63640705e3e71_1_690x128](https://user-images.githubusercontent.com/8491202/39319082-45c4b6ae-4980-11e8-831f-d96b04402c0c.png)

### Technical details:

* Redash Version: 4.0.0+b3948
* Browser/OS: Chrome 65.0.3325.181 / MacOS High Sierra
* How did you install Redash: This instance of Redash is built from the AWS AMI of Redash 2.0 and updated following the forum guidance.

arikfr:
@kocsmy I think I might have fixed this in my work on dashboard refresh -- what are the reproduction steps?

Caloulebeauf:
I have updated to the last version and i still have the issue :
![image](https://user-images.githubusercontent.com/8491202/39718957-18cdc4fe-5238-11e8-8d95-f6bff61d710c.png)
In a row of three elements with widgets above and under this row, i try to resize or move the left widget and before I released the click every other widget has moved to the top (respecting their horizontal position).

kravets-levko:
@arikfr  @kocsmy (cc @Caloulebeauf ) I think during next few days I'll find some time to fix this.

arikfr:
@kravets-levko did you manage to reproduce this?

kravets-levko:
No, but I have some ideas about why does this bug occur and what to do with
it. I'll ask Zsolt for a help if needed.

8 трав. 2018 р. 9:08 дп "Arik Fraimovich" <notifications@github.com> пише:

@kravets-levko <https://github.com/kravets-levko> did you manage to
reproduce this?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/getredash/redash/issues/2494#issuecomment-387295022>,
or mute the thread

kravets-levko:
@Caloulebeauf (cc @arikfr @kocsmy) Can you confirm that this is fixed? Thanks!

[...]

PR #2754, opened 2018-08-20
Widgets crumble when trying to move / resize a widget

kravets-levko:
Issue getredash/redash#2494

Also fixed bug in choropleth viz (introduced when replaced `underscore` with `lodash` - `lodash.max()` works differently from `underscore.max()`; also `underscore.pairs()` has another name in `lodash`)

### DIFF
--- a/client/app/components/dashboards/gridstack/index.js
+++ b/client/app/components/dashboards/gridstack/index.js
@@ -89,7 +89,8 @@ function gridstack($parse, dashboardGridOptions) {
             item.minSizeX, item.maxSizeX, item.minSizeY, item.maxSizeY,
             itemId,
           );
-          grid._updateStyles(grid.grid.getGridHeight());
+          const gridHeight = grid.grid.getGridHeight() + 10;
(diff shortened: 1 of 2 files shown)


client/app/components/dashboards/gridstack/index.js
```diff
@@ -89,7 +89,8 @@ function gridstack($parse, dashboardGridOptions) {
             item.minSizeX, item.maxSizeX, item.minSizeY, item.maxSizeY,
             itemId,
           );
-          grid._updateStyles(grid.grid.getGridHeight());
+          const gridHeight = grid.grid.getGridHeight() + 10;
```
arikfr: What's the purpose of the magic number (10)? :)

kravets-levko: Actual fix for the issue is few lines below (`grid._updateStyles` in `finally` section); I added this `10` because `gridstack` also adds it to preserve some extra space.

arikfr:
👍

Issue #2731, opened 2018-08-09
Change preferred sorting type


denisov-vlad:
### Issue Summary

Now Redash has ascending sort by creation time in dashboards / queries lists. Maybe descending could be better?

### Technical details:

* Redash Version: #1cafcc1

jezdez:
Woops, good catch. Also slightly related to #2757.

jezdez:
This was fixed in #2759 and I opened #2771 to track adding the dashboard sorting separately.

PR #2759, opened 2018-08-23
Sort queries in descending order.

jezdez:
Refs #2731.

### DIFF
--- a/client/app/pages/queries-list/index.js
+++ b/client/app/pages/queries-list/index.js
@@ -8,9 +8,13 @@ import './queries-list.css';
 class QueriesListCtrl {
   constructor($scope, $location, Events, Query, currentUser) {
     const page = parseInt($location.search().page || 1, 10);
+
     const orderSeparator = '-';
-    const pageOrder = $location.search().order || 'created_at';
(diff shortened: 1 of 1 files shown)


client/app/pages/queries-list/index.js
```diff
@@ -8,9 +8,13 @@ import './queries-list.css';
...
     const page = parseInt($location.search().page || 1, 10);
+
     const orderSeparator = '-';
-    const pageOrder = $location.search().order || 'created_at';
-    const pageOrderReverse = pageOrder.startsWith(orderSeparator);
+    this.pageOrder = $location.search().order || '-created_at';
+    this.pageOrderReverse = this.pageOrder.startsWith(orderSeparator);
+    if (this.pageOrderReverse) {
+      this.pageOrder = this.pageOrder.substr(1);
+    }
```
jezdez: This basically makes sure that thie orderByField parameter that is passed to the LivePaginator is always a value without the prefix `-` and allow the sort icon to render properly. Before this we've basically relied on implicit behavior of the ascending ordering that doesn't have a prefix.

jezdez:
@arikfr That's the PR I mentioned in our last meeting, basically a pretty obvious bug in the new server-side sorting feature. 😬

PR #2764, opened 2018-08-26
add partition key marker for Athena and Presto schemas

alison985:
for: https://github.com/mozilla/redash/issues/430

ports the following upstream: https://github.com/mozilla/redash/pull/201/files
for previous code review see: https://github.com/getredash/redash/pull/2655

### DIFF
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -30,6 +30,20 @@ PRESTO_TYPES_MAPPING = {
     "date": TYPE_DATE,
 }
 
+# for athena and presto
+def format_schema(results):
+    schema = {}
+    for row in results['rows']:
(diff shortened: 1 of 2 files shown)


redash/query_runner/presto.py
```diff
@@ -31,6 +31,21 @@
 }
 
 
+# for athena and presto
+def format_schema(results):
+    schema = {}
```
jezdez: I'm sorry but the docstring changes from https://github.com/getredash/redash/pull/2655/commits/17b9a0ace28d5b244787ef19a872c25cb1c3e457 didn't make it here, could you please add it?

redash/query_runner/presto.py
```diff
@@ -31,6 +31,21 @@
 }
 
 
+# for athena and presto
```
jezdez: No need to have this extra code comment.

PR #2656, opened 2018-06-30
add ability to add viz to dashboard from query page 

alison985:
fixes #955

This PR ports the Mozilla functionality upstream.

I had to make a minor tweak to get it to run in the current redash code. I also added feedback to the user that the widget was added successfully which is not in Mozilla's branch.

Original PR with code review is here: https://github.com/mozilla/redash/pull/184/

Issue #2641, opened 2018-06-28
Dashboard Parameters


arikfr:
Today to have a parameter at the dashboard level, we use the `global` flag for parameters.

New suggested flow:

* Deprecate the `global` flag.
* When adding a new widget to the dashboard, if the query has parameters we will show them to the user and let him pick what to do with them:
	1. Add the parameter to the dashboard (default option), creating a dashboard parameter.
	2. Map it to existing dashboard parameter (default option, if parameter with the same name exists).
	3. Give the parameter a static value.
	4. (Not sure we need this:) Keep it a standalone parameter at the widget level.

Things to think about:
* How to show if a widget is affected by a parameter? Maybe color code them, and show this color on the widget?

kocsmy:
This is related: https://github.com/getredash/redash/pull/2656

kocsmy:
This is also a great opportunity to improve the Add Widget modal and make it nicer.

kocsmy:
<img width="802" alt="screenshot 2018-07-13 17 44 43" src="https://user-images.githubusercontent.com/2378022/42700574-734abd2c-86c4-11e8-838d-234de9a41f26.png">

I've created this wireframe for the new modal to open the discussion.

- I'm wondering if we should move the "Text box" option to somewhere out of this model. My assumption is that text boxes are not as widely used as query visualizations. (first I was considering using an option to select which one you like on the [top of this modal](https://www.dropbox.com/s/8pz9rfuzadl0cl4/Screenshot%202018-07-13%2017.46.34.png?dl=0))
- The parameters section is the challenge here. I think we need to improve the separation of the parameters and options for the different methods of adding them.

arikfr:
👍 

How about we change the visualization selection from dropdown to multiple "cards" - one for each visualization, and show the parameters section after selecting the visualization?

Maybe even convert it to 3 steps:

1. Pick query (single screen, with recent queries and ability to search).
2. Pick visualization - at this point the query is selected and shown by name.
3. Setup parameters (if needed)

?

arikfr:
>Add text box will get it's own modal and we're adding a separate button in edit dashboard mode:

👍 

> Here's how it looks if we break it down to 3 steps. I'm not 100% convinced we need this, it feels the process gets a bit complicated. I created a quick prototype to demonstrate how this would work: https://marvelapp.com/18514ie6

[...]

PR #2767, opened 2018-08-26
add ability to add viz to dashboard from query edit page

### LINKED ISSUE #955: Add a button / drop-down from the query window to quickly add it to a dashboard.

alison985:
Fixes #955. Ports this feature from Mozilla's fork.

Previous PR on Mozilla: https://github.com/mozilla/redash/pull/184
Previous code review and discussion: https://github.com/getredash/redash/pull/2656 (Closed because I keep messing up merge/rebase from upstream master and while I did I found an existing dashboard search implementation in getredash/redash so I modified this feature to use that instead.)

Add to Dashboard modal:
add-to-dashboard-modal

iPhone6s screen size:
iphone6s

MacBook Air 13: full window:
macbookair 13 inch full


client/app/pages/queries/view.js
```diff
@@ -440,6 +440,19 @@ function QueryViewCtrl(
...
 
+  $scope.openAddToDashboardForm = (visId) => {
+    const visualization = getVisualization(visId);
+    $uibModal.open({
+      component: 'addToDashboardDialog',
+      size: 'sm',
+      resolve: {
+        query: $scope.query,
+        vis: visualization,
+        saveAddToDashboard: () => $scope.saveAddToDashboard,
```
arikfr: I couldn't find where this is used to where `$scope.saveAddToDashboard` is defined. Maybe a leftover?

client/app/pages/queries/add-to-dashboard.js
```diff
@@ -0,0 +1,52 @@
...
+      this.selected_query = this.resolve.query.id;
+      const widget = new Widget({
+        visualization_id: this.vis && this.vis.id,
+        dashboard_id: dash.id,
+        options: {},
+        width: 1,
+        type: 'visualization',
+      });
+      widget.save().then(() => {
+        this.selectedDashboard = Dashboard.get({ slug: dash.slug }, () => {});
```
arikfr: Why is this needed?

client/app/pages/queries/add-to-dashboard.html
```diff
@@ -0,0 +1,21 @@
+<div class="modal-header">
+    <button type="button" class="close" aria-label="Close" ng-click="$ctrl.close()"><span aria-hidden="true">&times;</span></button>
+    <h4 class="modal-title">Add to Dashboard</h4>
+</div>
+<div class="modal-body">
+    <form name="alertForm" class="form">
+      <div class="form-group">
+        <!-- <label>Limit Dashboard search to dashboards I've created? </label>
+        <input type="checkbox" ng-model="$ctrl.limitToUsersDashboards" on-change="$ctrl.dashboardList == $select.search"   ng-disabled="$ctrl.saveInProgress" /> -->
```
arikfr: Remove?

arikfr:
@kravets-levko FYI: we will need to support the new flow from #2641 here too. The easiest way to do this is to be able to use the component/modal you implement for #2641 in this modal as well. Only change being is that the first step will be to choose a dashboard instead of choosing query/visualization.

---

The bottom section becomes crowded:

> iPhone6s screen size:
iphone6s

> MacBook Air 13: full window:
macbookair 13 inch full

My suggestion:

* Replace {Download, Add to Dashboard, Embed} buttons with a single [Kebab menu](https://thenounproject.com/term/kebab-menu/198754/) that will have these options.
* Hide Add to Dashboard and Edit Visualization on small screen sizes.

@kocsmy @alison985 WDYT?

@alison985 you don't need to implement these changes as part of this pull request (unless you want to), but without them this pull request will need to wait for V5 to finalize before merging into master.

kocsmy:
Hiding these buttons behind a Kebab menu would make the interface lighter for sure. I'm wondering if that's fine and we're not hiding important features?

Also the download which has 2 option now should just open a modal for selecting the type, instead of further complicating the Kebab menu with submenus.

kravets-levko:
@kocsmy 

> Also the download which has 2 option now should just open a modal for selecting the type, instead of further complicating the Kebab menu with submenus.

Not necessarily: it could be two "Download ..." options in plain menu, like this:
```
Item 1
Item 2
Item 3
--------------------
Download as CSV file
Download as Excel file
--------------------
One more menu item
```

kocsmy:
Thanks, @kravets-levko that's also a good idea!

client/app/pages/queries/add-to-dashboard.html
alison985: Yes, all 3 of these sections are removed in the next commit.

alison985:
I've implemented the kebab menu. On an iPhone 6s it looks like this:
screenshot 2018-09-09 12 31 07

The problem is that there isn't enough vertical height for the full menu now. The Add to Dashboard option gets cut off.
screenshot 2018-09-09 12 32 38

I think the solution is to reduce the amount of screen space the query textarea takes up. I'm just not sure the best way to implement that.

Also, I'm not sure what you mean about the #2641 flow integration so I'm going to leave that part alone.

jezdez:
@alison985 Still needs a rebase. 
o5x49tdgjzo1ru8q831vuevdznmwnr3p-medium

PR #2774, opened 2018-08-29
Updates to CircleCI 2 configuration

arikfr:
* Split integration tests to their own workflow which is only executed on the master branch.
* Update names of jobs to be more descriptive.
* Fix an error in integration tests job setup.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,4 @@
 version: 2.0
-notify:
-  webhooks:
-    - url: https://webhooks.gitter.im/e/895d09c3165a0913ac2f
 jobs:
   build:
     docker:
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -63,14 +60,6 @@ jobs:
     steps:
       - attach_workspace:
           at: .
```
jrbenny35: I don't think this will work since it within another workflow. We need to add a ```checkout``` step. See [this](https://circleci.com/docs/2.0/configuration-reference/#checkout).

jrbenny35: Oh I didn't see that. I opened a PR for the checkout fix.

PR #2775, opened 2018-08-29
Add checkout to integration test workflow.

jrbenny35:
This is just a small change. I had designed the original config to work in one workflow, since we split the workflow we need to checkout the code branch to run the tests successfully.


arikfr:
Thanks! But as discussed already covered in #2777. I also moved the configuration from `.circleci.yml` to `.circleci/config.yml` so this file is removed.

PR #2780, opened 2018-09-02
port autocomplete toggle

alison985:
History of this change:
* Originally, a Mozilla redash user asked for a toggle to turn autocomplete off in https://github.com/mozilla/redash/issues/282 That was implemented in the Mozilla fork. 
* Madalin, the Mozilla QA person, then filed https://github.com/mozilla/redash/issues/344 to say that the autocomplete button should always be displayed, not just on hover over the query textarea. In the [PR to make that change](https://github.com/getredash/redash/pull/2588) discussion evolved to discuss the UI of this change. Now that https://github.com/getredash/redash/pull/2678 is merged, we can implement the solution.

This PR:
* Adds the autocomplete toggle as a button next to the Add a Parameter and the Format Query buttons.

Because this button has a state (on or off) I implemented it with a darker background color when off. This color change provides more information to the user and will communicate more in screenshots of bug reports.

When AutoComplete is on:
screenshot 2018-09-02 16 17 52

When AutoComplete is off:
screenshot 2018-09-02 16 16 52

### DIFF
--- a/client/app/components/queries/query-editor.js
+++ b/client/app/components/queries/query-editor.js
@@ -98,10 +98,11 @@ function queryEditor(QuerySnippet, $timeout) {
                 if (newSchema === undefined) {
                   return;
                 }
-                const tokensCount = newSchema.reduce((totalLength, table) => totalLength + table.columns.length, 0);
-                // If there are too many tokens we disable live autocomplete,
-                // as it makes typing slower.
-                if (tokensCount > 5000) {
+                const tokensCount =
+                  newSchema.reduce((totalLength, table) => totalLength + table.columns.length, 0);
+                // If there are too many tokens or if it's requested via the UI
+                // we disable live autocomplete, as it makes typing slower.
+                if (tokensCount > 5000 || !$scope.$parent.autoCompleteQuery) {
                   editor.setOption('enableLiveAutocompletion', false);
                 } else {
                   editor.setOption('enableLiveAutocompletion', true);
@@ -111,6 +112,10 @@ function queryEditor(QuerySnippet, $timeout) {
 
             $scope.$parent.$on('angular-resizable.resizing', () => {
               editor.resize();
+            });
+
+            $scope.$parent.$watch('autoCompleteQuery', () => {
+              editor.setOption('enableLiveAutocompletion', $scope.$parent.autoCompleteQuery);
             });
 
             editor.focus();
--- a/client/app/components/queries/query-editor.js
+++ b/client/app/components/queries/query-editor.js
@@ -26,6 +26,7 @@ function queryEditor(QuerySnippet, $timeout) {
       query: '=',
       schema: '=',
       syntax: '=',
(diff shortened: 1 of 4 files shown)


client/app/components/queries/query-editor.js
```diff
@@ -113,6 +114,10 @@ function queryEditor(QuerySnippet, $timeout) {
               editor.resize();
             });
 
+            $scope.$parent.$watch('autoCompleteQuery', () => {
+              editor.setOption('enableLiveAutocompletion', $scope.$parent.autoCompleteQuery);
+            });
```
kravets-levko: It's better to pass that option as component attribute binding and not access to parent scope

client/app/pages/queries/query.html
```diff
@@ -153,10 +153,14 @@ <h3>
                       <span ng-non-bindable>{{&nbsp;}}</span>
                     </button>
 
-                    <button type="button" class="btn btn-default" ng-click="formatQuery()" uib-tooltip="Format Query">
+                    <button type="button" class="btn btn-default m-r-5" ng-click="formatQuery()" uib-tooltip="Format Query">
                       <span class="zmdi zmdi-format-indent-increase"></span>
                     </button>
 
+                    <button type="button" class="btn btn-default m-r-5" ng-click="toggleAutoComplete()" uib-tooltip="Toggle AutoComplete" ng-class="onOrOff">
```
kravets-levko: Bootstrap already has a class for this - `active` - it makes button "pressed" and usually is used to implement such toggles. Also, usually "pressed" state is used to indicate that some feature is enabled, ad default state - when disabled (it's opposite behavior to implemented here).

client/app/components/queries/query-editor.js
kravets-levko: For this component: go to `scope: { ... }` definition object and add line: `autoCompleteQuery: '='` ([here](https://github.com/getredash/redash/blob/e2b41ebbb3dd36102b29381118f95522422a9c59/client/app/components/queries/query-editor.js#L28)). Then use it as `$scope.autoCompleteQuery` instead of `$scope.$parent.autoCompleteQuery`.

Then, [here](https://github.com/getredash/redash/blob/e2b41ebbb3dd36102b29381118f95522422a9c59/client/app/pages/queries/query.html#L146) add new attribute: `auto-complete-query="autoCompleteQuery"` (the variable from "parent" scope).

client/app/pages/queries/query.html
kravets-levko: `ng-class="{active: autoCompleteQuery}"` and remove `onOrOff` + related code

client/app/components/queries/query-editor.js
alison985: Thank you so much, that makes much more sense.

client/app/pages/queries/query.html
alison985: Updated.

Issue #789, opened 2016-01-22
Feature Request: Add 'last active' column to the /users page


JohnConnell:
While this can be accomplished with a query, having it on the /users page would make it more accessible to all users including non-technical ones that might only login to view dashboards and high-level pages. It would help promote accountability across some teams and give admins an easy way to view inactive accounts that they might need to disable or delete.

![redash_last_active](https://cloud.githubusercontent.com/assets/2201507/12505465/84a5741a-c0a4-11e5-9de9-e915ccc5232c.png)

arikfr:
Sounds good. We can update this column in the `record_event` background task when we process the events (I think most actions in the app will trigger some kind of event.

PR #2781, opened 2018-09-03
add last_active_at to users page

alison985:
This feature request was originally made in #789. It was implemented in the Mozilla fork in [issue 155](https://github.com/mozilla/redash/issues/155) with [PR 191](https://github.com/mozilla/redash/pull/191). This ports the feature to getredash/redash. It works with the revised pagination and table sort.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -473,6 +473,8 @@ class User(TimestampMixin, db.Model, BelongsToOrgMixin, UserMixin, PermissionsCh
         if with_api_key:
             d['api_key'] = self.api_key
 
+        d['last_active_at'] = Event.query.filter(Event.user_id == self.id).with_entities(Event.created_at).order_by(Event.created_at.desc()).first()
+
         return d
 
     def is_api_user(self):
--- a/client/app/pages/users/list.html
+++ b/client/app/pages/users/list.html
@@ -46,6 +46,10 @@
               Joined
               <sort-icon column="'created_at'" sort-column="$ctrl.paginator.orderByField" reverse="$ctrl.paginator.orderByReverse"></sort-icon>
             </th>
+            <th class="sortable-column" ng-click="$ctrl.paginator.orderBy('last_active_at')">
+              Last Active At
+              <sort-icon column="'last_active_at'" sort-column="$ctrl.paginator.orderByField" reverse="$ctrl.paginator.orderByReverse"></sort-icon>
+            </th>
             <th width="1%"></th>
           </tr>
         </thead>
@@ -61,6 +65,9 @@
             </td>
             <td>
               <span am-time-ago="user.created_at"></span>
+            </td>
+            <td>
+              <span am-time-ago="user.last_active_at[0]" uib-tooltop="user.last_active_at[0]"></span>
             </td>
             <td>
               <div ng-if="$ctrl.currentUser.hasPermission('admin') && (user.id != $ctrl.currentUser.id)">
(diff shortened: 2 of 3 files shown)


redash/models.py
```diff
@@ -473,6 +473,8 @@ def to_dict(self, with_api_key=False):
         if with_api_key:
             d['api_key'] = self.api_key
 
+        d['last_active_at'] = Event.query.filter(Event.user_id == self.id).with_entities(Event.created_at).order_by(Event.created_at.desc()).first()
```
arikfr: This probably explains why your users list was slow to load -- when you load the users list it makes a query to Event for each user, which can be very slow when loading hundreds of users. 

Also, for various reasons (mostly performance), we don't query the events table. 

If we want this feature, we will need to implement this in a different way (we can discuss, if needed).

jezdez: Ah, that would explain it indeed. 😬 

I think it would still make sense to implement this, maybe as a new column in the user table, which is updated every time users in a to-be-defined area of Redash (e.g. query page, dashboard view, etc)?

I guess updating it on every request would be a non-trivial amount of UPDATE queries, unless of course it would be done from a Celery task. But that seems like a good way to increase the risk of creating a trampling herd situation when lots of interactions happen and the Celery queues fills up without auto-scaling.

Alternatively we could use a Redis hash (user id -> last active date) and just update that on every request.

arikfr: I agree that updating it on every request will be brutal, even through a Celery job.

I think a reasonable workflow can be:

1. Update Redis whenever the user interacts with the system.
2. Have a periodic job that syncs this timestamp to the database.

I wouldn't add a dedicated column for this, but rather add a generic JSON blob to the users table and store it there (like we do with many other models).

alison985: @jezdez From what I know about Redis it doesn't store information if the process stops. Can you confirm we're okay with that behavior (i.e. small % chance of data being lost)? Also for the periodic job that syncs Redis to database is 10 minutes okay? Or an hour?

jezdez: @alison985 Ah interesting, according to [Redis' Docker page](https://hub.docker.com/_/redis/) uses an append only storage format only when passed the `--appendonly yes` option in Docker, so you're right about that being not ideal.

For Mozilla IIRC we're running on AWS' Elasticcache Redis instances, which persist by default, so it'd be not a huge deal on our side.

@arikfr Do you prefer switching to persistent storage in the Docker files (maybe not a bad idea any to be able to restore Celery queues on failure)?

arikfr: Actually Redis does persist information to disk on some schedule so it's not really lost.

jezdez: @arikfr Hm, are you sure the docker-compose will use the specified volume [`/data`](https://github.com/docker-library/redis/blob/0cf1c40cf2fe83c5b788d3b8472ffb78dc4d6449/4.0/alpine/Dockerfile#L61) automatically? I understand the default (when not using appendonly) is to simply fsync things to disk. But inside a container that's still ephemeral if the volume isn't mounted during container start. Which is the case I believe for the docker-compose setup. Or am I missing something?

arikfr: @jezdez you're right, it will be kept as long as the container lives but if the container gets rebuild it will be lost.

It does make sense to mount a volume to give the user a chance to back it up, but it's not a huge priority as the data there is (relatively) short lived anyway.

arikfr: And to clarify: when the container restarts the data is kept. It's only when the container gets removed the data will be lost.

jezdez: @arikfr Understood. I've filed https://github.com/getredash/redash/issues/2860 to track this.

arikfr:
Can we close this and reopen with the new implementation?

jezdez:
@arikfr Yep, I'll be working on the Redis implementation.

Issue #2544, opened 2018-05-17
Receive communication error when Athena returns NaN value


tdawber:
### Issue Summary

When Athena returns a NaN or Infinity value, generally as a result of dividing by 0.0, the following error is displayed:

“Error running query: failed communicating with server. Please check your Internet connection and try again.”

This error message is confusing, and the result could also be handled more nicely.

### Steps to Reproduce

1. execute the following query on an Athena connection:

select 1.0/0.0

When you divide an integer value by an integer zero (select 1/0), Athena throws the error "Error running query: DIVISION_BY_ZERO: / by zero" and this is presented nicely in Redash.

It would be good if either Redash throw a meaningful error upon being unable to handle the Infinity/Nan value, or if it displayed all results and presented null or some other value in place of Infinity/Nan.

### Technical details:

* Redash Version: 4.0.1

arikfr:
The issue is due to `json.dumps` sending the value as `Infinity`, which is considered an error by the JSON parser in the browser (see example below).

```json
{  
   "query_result":{  
      "data":{  
         "rows":[  
            {  
               "_col0":Infinity
            }
         ],
     }
   }
}
```

So there are two issues here:

1. We show the wrong message to the user: there wasn't a communication error, but we failed to parse the response.
2. We should send values the browser understands... Although I'm not sure what we should send in such case? `null`?

PR #2783, opened 2018-09-03
 Receive communication error when Athena returns NaN value

kravets-levko:
Issue getredash/redash#2544

Solution: use `simplejson` instead of `json` to dump query results; `simplejson` can replace `NaN`/`Infinity` values with `null`

### DIFF
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -1,10 +1,11 @@
 import json
 import logging
 import os
+import simplejson
 
 from redash.query_runner import *
 from redash.settings import parse_boolean
-from redash.utils import JSONEncoder
+from redash.utils import SimpleJSONEncoder
 
 logger = logging.getLogger(__name__)
 ANNOTATE_QUERY = parse_boolean(os.environ.get('ATHENA_ANNOTATE_QUERY', 'true'))
@@ -194,7 +195,7 @@ class Athena(BaseQueryRunner):
                     'athena_query_id': athena_query_id
                 }
             }
-            json_data = json.dumps(data, cls=JSONEncoder)
+            json_data = simplejson.dumps(data, ignore_nan=True, cls=SimpleJSONEncoder)
             error = None
         except KeyboardInterrupt:
             if cursor.query_id:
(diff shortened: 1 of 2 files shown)


arikfr:
We should probably switch to simple_json everywhere. I just wonder how can we make sure this is the right decision?

jezdez:
@arikfr I would argue that we should maybe look for a faster and/or better json implementation than simplejson given the amount of serialization/deserialization happening allover Redash.

There are some precedence discussing [benchmarks](https://python-rapidjson.readthedocs.io/en/latest/benchmarks.html), [other benchmarks](https://artem.krylysov.com/blog/2015/09/29/benchmark-python-json-libraries/), [memory consumption](https://blog.ionelmc.ro/2015/11/22/memory-use-and-speed-of-json-parsers/) (!) and [many](https://pypi.org/project/ujson/), [many](https://pypi.org/project/yajl/), [many](https://github.com/mre/hyperjson) more. OTOH [simplejson](https://github.com/simplejson/simplejson) is the real thing and has been maintained for a while now.

arikfr:
@jezdez thank you for the pointers! You're definitely right that we will benefit from a more performant (and memory efficient) JSON implementation.

From reading the [memory consumption blogpost](https://blog.ionelmc.ro/2015/11/22/memory-use-and-speed-of-json-parsers/), it looks like simplejson has the best memory consumption for a JSON that's more similar to what we usually have:

```
Peak memory usage (Python 2.7):

      simplejson: 1,171.7 Mb
           cjson: 1,304.2 Mb
            yajl: 1,457.1 Mb
            json: 1,468.0 Mb
       rapidjson: 1,561.6 Mb
        jsonlib2: 2,134.9 Mb
         jsonlib: 2,137.0 Mb
           ujson: 2,149.9 Mb
```

Also based on the same blog post, it seems like that only relevant alternative is `ujson` (considering stability), but it doesn't seem to support all the options we need...

jezdez:
@arikfr Excellent, I hadn't looked as closely, that seals it. Do you want me to work on a patch?

arikfr:
You're always welcome to :)

Issue #2757, opened 2018-08-21
Queries list should not default to ordering by creation date when search is used


jezdez:
### Issue Summary

A summary of the issue and the browser/OS environment in which it occurs.

### Steps to Reproduce

1. Search for a query on the queries list page
2. See that it applies the ordering by creation date (by `created_at`) instead of ordering by the scoring of the full text search

### Technical details:

* Redash Version: 5.0beta+
* Browser/OS: Firefox
* How did you install Redash: Docker

jezdez:
#2798 implements this, but needs #2790.

PR #2790, opened 2018-09-06
Refactor list based controllers.

jezdez:
This reduces the code duplication between the dashboard, user and queries list pages and normalizes many of the APIs between them.

This also:
- allows sorting query favorites
- adds a pagination size select to the dashboard list
- fixes a bunch of UI inconsistencies between the queries and dashboards list (e.g. margins)

The new ListCtrl class is subclassed in the various specific page controllers and extended as needed. New list pages can make use of the same pattern in the future.

### DIFF
--- a/redash/handlers/dashboards.py
+++ b/redash/handlers/dashboards.py
@@ -69,6 +69,18 @@ class DashboardListResource(BaseResource):
             serializer=serialize_dashboard,
         )
 
+        if search_term:
+            self.record_event({
+                'action': 'search',
+                'object_type': 'dashboard',
+                'term': search_term,
+            })
+        else:
+            self.record_event({
+                'action': 'list',
+                'object_type': 'dashboard',
+            })
+
         return response
 
     @require_permission('create_dashboard')
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -203,6 +203,18 @@ class QueryListResource(BaseResource):
             with_last_modified_by=False
         )
 
+        if search_term:
+            self.record_event({
+                'action': 'search',
+                'object_type': 'query',
+                'term': search_term,
+            })
+        else:
+            self.record_event({
+                'action': 'list',
+                'object_type': 'query',
+            })
+
         return response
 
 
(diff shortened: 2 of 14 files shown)


jezdez:
This will also simplify fixing #2757 since it's only happening in one code area instead of three.

arikfr:
I didn't have the chance to review this yet, but happy to see this :) 👏

jezdez:
@arikfr I've rebased this, in case you want to take another look.

arikfr:
@jezdez 👍 will do.

arikfr:
Ready to go, just needs rebasing.

redash/handlers/users.py
```diff
@@ -64,7 +65,7 @@ def serialize_user(user):
         if search_term:
             users = models.User.search(users, search_term)
             self.record_event({
-                'action': 'list',
+                'action': 'search',
```
jezdez: A tiny oversight in #2766.

redash/handlers/queries.py
```diff
@@ -203,6 +203,13 @@ def get(self):
             with_last_modified_by=False
         )
 
+        if search_term:
+            self.record_event({
+                'action': 'search',
+                'object_type': 'query',
+                'term': search_term,
+            })
```
jezdez: A missing event recording in #2766 for the new hybrid query view that also can be used for search.

redash/handlers/dashboards.py
```diff
@@ -69,6 +69,13 @@ def get(self):
             serializer=serialize_dashboard,
         )
 
+        if search_term:
+            self.record_event({
+                'action': 'search',
+                'object_type': 'dashboard',
+                'term': search_term,
+            })
```
jezdez: A missing event recording for #2766.

redash/handlers/queries.py
arikfr: 👍 
If we're at it, then also need to record a list event when it's not search. Also worth recording other parameters of the request - can be useful to analyze if people use tag based navigation or how deep they go.

jezdez: Good idea!

jezdez: Added.

Issue #2796, opened 2018-09-07
Improve mobile support


arikfr:
While introducing many improvements in V4, we made the mobile experience worse :( Here's some issues along with proposed solutions. I welcome everyone to share other issues they found on mobile.

## Dashboards

![image](https://user-images.githubusercontent.com/71468/45221561-77e5a500-b2ba-11e8-84e9-a3dfbca0d966.png)

* Header: we need to move the button to their own row instead of sharing the row with the title.
* The only button we need to keep in the header is Refresh (just refresh, without the auto refresh options) as the others are unlikely to be used in a mobile context.

### Counter
* (This is not only for mobile:) counter visualization should have a consistent font size - no more adjusting the size according to the widget size.
* On desktop we can show ellipsis when the counter text is too large for its space, as the user can resize the widget to fit the text. What can we do on mobile?

### Charts

* On mobile the legend sometimes overlaps the chart content and sometimes just pushes it "to the side" (as can be seen in the dashboard screenshot above). We need to move the legend to the bottom when the widget has less than some threshold of pixels width.

## Query View

![image](https://user-images.githubusercontent.com/71468/45221984-d7908000-b2bb-11e8-8c8f-b8a30d604b12.png)

* Hide the "..." button on mobile.
* Hide the "Embed" button.
* The page has static sections and you can only scroll the results area. This makes sense on the desktop, but not in mobile. We need to fix this.
* We need to improve the metadata section -- probably move it down and show full names along with the avatar.

kocsmy:
### Dashboard
- [ ] Header: we need to move the button to their own row instead of sharing the row with the title.
- [ ] The only button we need to keep in the header is Refresh (just refresh, without the auto refresh options) as the others are unlikely to be used in a mobile context

### Query
- [ ] Hide the "..." button on mobile.
- [ ] Hide the "Embed" button.
- [ ] The page has static sections and you can only scroll the results area. This makes sense on the desktop, but not in mobile. We need to fix this.
- [ ] We need to improve the metadata section -- probably move it down and show full names along with the avatar.
 
### Visualizations
- [ ] Counter (This is not only for mobile:) counter visualization should have a consistent font size - no more adjusting the size according to the widget size.
- [ ] On desktop we can show ellipsis when the counter text is too large for its space, as the user can resize the widget to fit the text. What can we do on mobile?

### Charts
[...]

PR #2800, opened 2018-09-09
Improved mobile design

kocsmy:
Improvements based on #2796


client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -804,6 +804,48 @@ text.slicetext {
...
+        line-height: 2;
+      }
+
+      .dashboard__control {
+        margin: 5px 0;
+      }
+    }
+
+    .btn-fullscreen, .btn-share, .btn-more, tags-control, .btn-refresh-dropdown {
+      display: none;
```
arikfr: You don't need specific classes for these buttons just to hide them. You can use the `hidden-xs` class instead.

client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -804,6 +804,48 @@ text.slicetext {
     text-align: center !important;
   }
 
+  dashboard-page {
```
arikfr: In the future, please place the markup for the component/page in a file next to the component/page.

client/app/assets/less/redash/redash-newstyle.less
kocsmy: Yes, but I'm trying to avoid those as it might lead to more complications on the long run (we had some issues with these in the past). Not sure which one makes more sense to be honest and I think we don't have a consistent pattern here. Happy to change if you think we should do so.

arikfr: What kind of complications?

kocsmy: If I remember correctly it was with flexboxes when we originally used different flex classes to control the flexbox but then we had to change things on different screen size. So we had to introduce specific classes instead and use media queries accordingly to control how it changes.

Not sure if I'm explaining it correctly.

Anyway, translating to this specific case: what if we need to change something else on this button in the future. Then we got to keep adding extra classes or just have 1 class and do the rest in Sass.

arikfr: _If_ we will need to change a button in a specific way, we can introduce an extra class. I don't see why should we add it when it's not needed though.

Also somewhat related: I want to avoid having unique snowflakes in our styling as much as possible. Of course it's not always possible, but we should aim there. So if we need to change a button, we should think whether it's a new style of general type of button.

kocsmy: Alright, I'll change it accordingly!

kocsmy:
I fixed the issues you suggested + moved the dashboard related styling to dashboard.less

arikfr:
👍

PR #2203, opened 2018-01-04
use 'hd' parameter for Google OAuth when only one domain is allowed

washort:
This is a fix for #1924.

Issue #1924, opened 2017-08-09
Add `hd` parameter for optimized Google Apps login


benmanns:
https://github.com/getredash/redash/blob/53268989c5fe5c619b399b488094c501cb286af4/redash/authentication/google_oauth.py#L91

Would it be possible to add the [hd parameter](https://developers.google.com/identity/protocols/OpenIDConnect#hd-param) to this authorization URL in the event that there's a single allowed Google Apps domain? It optimizes the sign in experience by limiting the suggested accounts to those for the passed Google domain.

If you're signed in to a single account on the domain and already authorized Redash: instant redirect to login.

If you're signed in on the domain and haven't authorized Redash: present only accounts from the authorized Google domain.

<img width="459" alt="screen shot 2017-08-09 at 3 36 34 pm" src="https://user-images.githubusercontent.com/27628/29140577-ecce0ca0-7d18-11e7-8001-f5ca4215df56.png">

If you're not signed in: present the login dialog with your domain pre-filled.

<img width="461" alt="screen shot 2017-08-09 at 3 33 25 pm" src="https://user-images.githubusercontent.com/27628/29140571-e98e1e2c-7d18-11e7-825f-4ad982aeb5c4.png">

If you're interested in merging something like this but can't prioritize this as a feature I can put together a PR with the changes required.

arikfr:
Just to make sure I understand: when there is only one Google Apps domain defined, we will pass the `hd` param but otherwise do what we do today?

benmanns:
Yeah, exactly.

arikfr:
It's practically the same, the only thing that changes is how you find current org:

https://github.com/getredash/redash/blob/53268989c5fe5c619b399b488094c501cb286af4/redash/authentication/google_oauth.py#L109-L112

PR #2801, opened 2018-09-10
re-prompt for login if unauthorized Google account with hd support

alison985:
Issues:
* https://github.com/mozilla/redash/issues/79 - request for reprompt for login
* https://github.com/mozilla/redash/issues/457 - port request for reprompt for login upstream
* https://github.com/getredash/redash/issues/1924 - request to show only emails from 1 google domain if there is only 1 domain

Previous PRs/relevant commits:
* https://github.com/mozilla/redash/commit/59366660e2c67686c545480c69f341e9cacb633f
* https://github.com/getredash/redash/pull/1897 (Closed, this PR replaces)
* https://github.com/getredash/redash/pull/2203 (can be closed when this PR is merged)

This PR:
* Changes the behavior when an unauthorized google account is used to login from giving an error message to reprompting for login. That message is carried through a query string instead of the session.
* Only limits the google domain accounts shown in the login prompt if there is 1 google domain for the org.

### DIFF
--- a/redash/authentication/google_oauth.py
+++ b/redash/authentication/google_oauth.py
@@ -66,10 +66,23 @@ def org_login(org_slug):
 @blueprint.route('/oauth/google', endpoint="authorize")
 def login():
     callback = url_for('.callback', _external=True)
-    next_path = request.args.get('next', url_for("redash.index", org_slug=session.get('org_slug')))
+    relogin = request.args.get('relogin', '0')
     logger.debug("Callback url: %s", callback)
     logger.debug("Next is: %s", next_path)
-    return google_remote_app().authorize(callback=callback, state=next_path)
+    extra = {}
+    org_slug = session.get('org_slug')
+    if org_slug:
+        org = models.Organization.get_by_slug(session.pop('org_slug'))
+    else:
+        org = current_org
+    next_path = request.args.get('next', url_for("redash.index", org_slug=org_slug))
+
+    hd = None
+    if len(org.google_apps_domains) == 1:
+        extra['hd'] = org.google_apps_domains[0]
+    if relogin == '1':
+        extra['prompt'] = 'consent'
+    return google_remote_app().authorize(callback=callback, state=next_path, **extra)
 
 
 @blueprint.route('/oauth/google_callback', endpoint="callback")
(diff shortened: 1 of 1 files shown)


redash/authentication/google_oauth.py
```diff
@@ -94,8 +107,7 @@ def authorized():
 
     if not verify_profile(org, profile):
         logger.warning("User tried to login with unauthorized domain name: %s (org: %s)", profile['email'], org)
-        flash("Your Google Apps account ({}) isn't allowed.".format(profile['email']))
-        return redirect(url_for('redash.login', org_slug=org.slug))
+        return redirect(url_for('redash.login', relogin='1', org_slug=org.slug))
```
jezdez: This can't redirect back to `redash.login`, the `redash/handlers/authentication.py::login` function, but back to either the [`org_login`](https://github.com/getredash/redash/blob/93b34bcbce2b3df328e1c1fac4261fa83e8717b3/redash/authentication/google_oauth.py#L60-L63) or [`login`](https://github.com/getredash/redash/blob/93b34bcbce2b3df328e1c1fac4261fa83e8717b3/redash/authentication/google_oauth.py#L66-L72) views in the `redash/authentication/google_auth.py` file. Otherwise the `relogin` parameter isn't picked up since it's lost when `redash.login` renders the button with the link to the Google login again.

I'm wondering if there is a chance that this ends in an endless loop, but I think Google will halt that when it happens too often.

The `org_login` view will need to also pass the relogin paramter along to `login` since it's really just a small facade to populate the `org_slug` URL path parameter.

Alternatively, we can also modify the `redash.login` main view (/login) and check if relogin was provided to it and pass it along the the Google auth link when it renders the button. That would mean that if users need to relogin, they'd always end up on the main login page with the button again, instead of being passed to Google right away.

Either way, this needs another pass, I'm afraid.

jezdez:
@alison985 This needs another pass, see my comment above.

alison985:
@jezdez Updated.

jezdez:
@washort Do you mind elaborating on the implementation for the patch (since you originally wrote it in https://github.com/mozilla/redash/commit/59366660e2c67686c545480c69f341e9cacb633f)?

washort:
My recollection is that even when a specific domain name is requested the user can override it. So if an incorrect account is used, this will re-prompt for a valid login instead of getting stuck in a redirect loop.

arikfr:
> So if an incorrect account is used, this will re-prompt for a valid login instead of getting stuck in a redirect loop.

I think it's worth trying reproducing this in a newer version. In the past we didn't show the login page if only Google Authentication was used, so it would redirect there immediately. Now we do show the login page, so I think it will show the error on the login page, and won't start a loop.

arikfr:
In light of my last comment, closing this.

PR #2803, opened 2018-09-10
address tag display on query list page

alison985:
addresses visual issues listed in https://github.com/mozilla/redash/issues/480

Revised query list page screenshot with these changes:
screenshot 2018-09-09 22 38 58

Let me know what you think. I thought it would be easier to start this discussion with a proposed solution even if it ends up completely different.


client/app/pages/queries-list/queries-list.html
```diff
@@ -97,8 +97,9 @@
               <td class="table-main-title">
                 <a href="queries/{{query.id}}">{{query.name}}</a>
                 <br>
-                <span class="label label-default" ng-if="query.is_draft">Unpublished</span>
-                <tags-control item="query"></tags-control>
+                <span class="maxwidth300">
```
kravets-levko: What is this for?

alison985: Without it the queries list will go long into the right sidebar.

alison985: There is further discussion in #2829

jezdez:
@kravets-levko Could you clarify if this is ready to be merged?

kravets-levko:
I have my doubts that it's the best solution to set a limit on tag's container width. Maybe, it's better to limit each tag width to some reasonable value (150-250px - not sure about exact value, it's better to try and see how it looks), and somehow cut too long tags (`text-overflow: ellipsis` or something like this), and let them wrap to multiple lines (IIRC they already wrap). @jezdez, @alison985 (cc @kocsmy) WDYT?

jezdez:
@kocsmy Makes sense to me.

So this would imply updating the date model to have a max length for the tag name and making sure the tag editing code can handle response where the tags being created may result in an error.

jezdez:
@arikfr WFM!

client/app/assets/less/inc/list-group.less
```diff
@@ -17,7 +17,12 @@
     }
 }
 
- .list-group-item {
+.max-character {
+  text-overflow: ellipsis;
+  overflow: hidden;
+}
```
kravets-levko: It also needs `white-space: nowrap` rule; even better is to replace all that with macro from Bootstrap: `.text-overflow();`

client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -449,10 +449,21 @@ page-header, .page-header--new {
...
+}
+
 .label-tag {
   background: fade(@redash-gray, 10%);
   color: fade(@redash-gray, 75%);
   margin-right: 3px;
+  display: inline-block;
+  margin-top: 2px;
+  text-overflow: ellipsis;
+  overflow: hidden;
```
kravets-levko: See my comment above - the same applies here

kravets-levko:
> If I were to add the maxwidth300 class back in I can get it to look like this 

Add `text-nowrap` class to all other cells except the first one in each row - it will do the job.

> Showing the all characters in the tag on hover using .label-tag:hover doesn't work, so I didn't include it.

Have you tried to add `title="full tag name here"`?

alison985:
Very cool @kravets-levko, thank you. Please review again.

kocsmy:
@alison985 I've extended your PR with an addition to handle the tag counter on the sidebar on #3006

arikfr:
Thanks!

PR #2808, opened 2018-09-12
UI Fixes for favorites and empty state

arikfr:
* Fixed an issue where the dashboards empty state won't show up for no dashboards available.
* Changed the empty state of queries/dashboards dropdown text to explicitly mention Favorites, as it's not entirely clear otherwise ("⭐️Dashboards will appear here" -> "⭐️Favorite Dashboards will appear here").
* Added an empty state on the homepage:

image


arikfr:
@kocsmy FYI. The empty state on the homepage is very basic, but better than just empty...

PR #2815, opened 2018-09-13
Fix QueryResultListResource docstring

tdsmith:
Change the docstring to reflect the behavior of the function.

The `QueryResultListResource` defaults max_age to -1. `run_query` passes max_age through to `models.QueryResult.get_latest`, where -1 means [any cached result](https://github.com/getredash/redash/blob/1e6d6bbeb3ad50f531fb017b8abec7e7a7ffb79c/redash/models.py#L754).

This behavior is inconsistent with the docstring, which claims it always executes and never uses a cached result by default (which would be true if max_age were defaulted to zero).


redash/handlers/query_results.py
```diff
@@ -104,7 +104,7 @@ def post(self):
 
         :qparam string query: The query text to execute
         :qparam number query_id: The query object to update with the result (optional)
-        :qparam number max_age: If query results less than `max_age` seconds old are available, return them, otherwise execute the query; if omitted, always execute
+        :qparam number max_age: If query results less than `max_age` seconds old are available, return them, otherwise execute the query; if omitted or -1, returns any cached result, or executes if not available. Set to zero to always execute.
```
jezdez: Would you mind adding linebreaks to improve readability a bit? I think the best practice is the indent at the first character of the parameter description. Thanks!

PR #2816, opened 2018-09-14
Use parent default method of simplejson encoder.

jezdez:
This is a regression introduced in #2783.


kravets-levko:
@arikfr I'll merge this

PR #2817, opened 2018-09-14
Consistently use simplejson to loading and dumping JSON.

### LINKED ISSUE #2807: Consistently use simplejson for speed/memory improvements
### Issue Summary

At the moment Redash uses the Python json module except for one area where it [recently started to use the also installed simplejson](https://github.com/getredash/redash/pull/2783) to be able to use the `ignore_nan` parameter. Simplejson is the original package that became the json stdlib module but which also has been further developed since then.

The standalone [simplejson](https://pypi.org/project/simplejson/) is faster and more effective in using memory on Python 2.7 according to a [set of benchmarks](https://blog.ionelmc.ro/2015/11/22/memory-use-and-speed-of-json-parsers/#lots-off-small-objects).

Once we switch to Python 3 we can revert to using the stdlib json module again as it seems to be very optimized there again. I'll make sure the patch uses a central `json_loads` and `json_dumps` function so we won't have to do another big patch to revert to json.

### Technical details:

* Redash Version: 5.0beta
* Browser/OS: Firefox

jezdez:
This introduces the new functions redash.utils.json_dumps and redash.utils.json_loads and simplifies the custom encoder setup.

UUIDs are now handled by the default encoder, too.

Fixes #2807.

### DIFF
--- a/redash/settings/helpers.py
+++ b/redash/settings/helpers.py
@@ -1,5 +1,4 @@
 import os
-import simplejson
 
 
 def fix_assets_path(path):
@@ -19,8 +18,15 @@ def set_from_string(s):
     return set(array_from_string(s))
 
 
-def parse_boolean(str):
-    return simplejson.loads(str.lower())
+def parse_boolean(s):
+    """Takes a string and returns the equivalent as a boolean value."""
+    s = s.strip().lower()
+    if s in ('yes', 'true', 'on', '1'):
+        return True
+    elif s in ('no', 'false', 'off', '0', 'none'):
+        return False
+    else:
+        raise ValueError('Invalid boolean value %r' % s)
 
 
 def int_or_none(value):
--- a/redash/settings/helpers.py
+++ b/redash/settings/helpers.py
@@ -1,5 +1,5 @@
-import json
 import os
+import simplejson
 
 
 def fix_assets_path(path):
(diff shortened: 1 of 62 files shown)


redash/settings/helpers.py
```diff
@@ -39,7 +39,7 @@ def set_from_string(s):
 
 
 def parse_boolean(str):
-    return json.loads(str.lower())
+    return simplejson.loads(str.lower())
```
kravets-levko: It's not related to this PR, but isn't it simpler to just compare value to `'true'`/`'false'` string instead of using JSON parser?

redash/utils/__init__.py
```diff
@@ -86,8 +90,8 @@ def process_default(self, o):
         return False, None  # default processing
 
 
-class JSONEncoder(JSONEncoderMixin, json.JSONEncoder):
-    """Adapter for `json.dumps`."""
+class JSONEncoder(JSONEncoderMixin, simplejson.JSONEncoder):
+    """Adapter for `simplejson.dumps`."""
```
kravets-levko: I think, if one of JSONEncoder classes is removed, JSONEncoderMixin can also be removed and code from `JSONEncoderMixin.process_default` can be moved to `JSONEncoder.default`

jezdez: Good idea, I didn't know the history of the class. Will change..

redash/settings/helpers.py
jezdez: yeah, that's easier and slightly faster, too

jezdez:
@kravets-levko Pushed the two things you asked about.

Issue #1194, opened 2016-07-20
Shared dashboard crashes if one of the queries has no results yet


arikfr:
When loading a shared dashboard, if one of the queries doesn't have results yet, the dashboard won't load.

NicWagenaar:
Having the same issue, this is the error message:
`QueryResultDoesNotExist: Instance matching query does not exist:`
`SQL: SELECT "t1"."id", "t1"."org_id", "t1"."data_source_id", "t1"."query_hash", "t1"."query", "t1"."data", "t1"."runtime
", "t1"."retrieved_at" FROM "query_results" AS t1 WHERE ("t1"."id" IS %s)`

lab79:
Empirically noticed the error when the name of dashboard contains ".", " ", etc.

PR #2823, opened 2018-09-15
Run queries with no cached result in public dashboards (re #1194)

washort:
When a public widget is requested whose query has no results yet, run the query.

### DIFF
--- a/redash/serializers.py
+++ b/redash/serializers.py
@@ -9,6 +9,7 @@ from funcy import project
 from flask_login import current_user
 from redash import models
 from redash.permissions import has_access, view_only
+from redash.handlers.query_results import run_query_sync
 
 
 def public_widget(widget):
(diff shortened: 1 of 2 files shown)


redash/serializers.py
```diff
@@ -9,6 +9,7 @@
 from flask_login import current_user
 from redash import models
 from redash.permissions import has_access, view_only
+from redash.handlers.query_results import run_query_sync
```
arikfr: > [Please don't reuse in other API handlers.](https://github.com/getredash/redash/blob/master/redash/handlers/query_results.py#L22-L25)

:-)

There are two issues with `run_query_sync`:

1. Security issue. It allows SQL injections for external users, which you avoid by not supporting parameters.
2. Performance: it runs synchronously, so if there are enough requests it can block the web server.

You avoided the security issue by not supporting parameters, but the performance (and in a way another security issue) is still there.

jezdez:
I'm closing this PR for now, since this is currently the wrong direction. I've opened #2844 to implement the short term solution as [mentioned above](https://github.com/getredash/redash/pull/2823#pullrequestreview-155734950).

PR #2824, opened 2018-09-15
feat: add support for expanding dashboard visualizations

sjakthol:
These changes implement support for expanding a dashboard visualization
into a larger modal dialog.

This is useful if you have a dashboard with lots of small widgets and
want to inspect one of the widgets more closely. In the past, this
would've required you to navigate to the query page to see a larger
version of the visualization. With these changes, visualizations can
be expanded right from the dashboard.

The implementation is simple as it just renders the visualization into
a modal dialog. Other parts of the widget (e.g. parameters) are not
included in this dialog.

--

Example Screenshot

### DIFF
--- a/client/app/components/dashboards/widget-dialog.html
+++ b/client/app/components/dashboards/widget-dialog.html
@@ -1,12 +1,11 @@
 <div class="modal-header">
   <button type="button" class="close" aria-hidden="true" ng-click="$ctrl.dismiss()">&times;</button>
   <div class="visualization-title">
-    <visualization-name visualization="$ctrl.widget.visualization"/>
-    <span>{{$ctrl.widget.getQuery().name}}</span>
+    <query-link query="$ctrl.widget.getQuery()" visualization="$ctrl.widget.visualization" readonly="true"></query-link>
   </div>
 </div>
 <div class="modal-body">
-    <visualization-renderer visualization="$ctrl.widget.visualization" query-result="$ctrl.widget.getQueryResult()" class="t-body"></visualization-renderer>
+  <visualization-renderer visualization="$ctrl.widget.visualization" query-result="$ctrl.widget.getQueryResult()" class="t-body"></visualization-renderer>
 </div>
 <div class="modal-footer">
   <button type="button" class="btn btn-default" ng-click="$ctrl.dismiss()">Close</button>
--- a/client/app/components/dashboards/widget-dialog.html
+++ b/client/app/components/dashboards/widget-dialog.html
@@ -1,0 +1,13 @@
+<div class="modal-header">
+  <button type="button" class="close" aria-hidden="true" ng-click="$ctrl.dismiss()">&times;</button>
+  <div class="visualization-title">
+    <visualization-name visualization="$ctrl.widget.visualization"/>
+    <span>{{$ctrl.widget.getQuery().name}}</span>
+  </div>
+</div>
+<div class="modal-body">
+    <visualization-renderer visualization="$ctrl.widget.visualization" query-result="$ctrl.widget.getQueryResult()" class="t-body"></visualization-renderer>
+</div>
+<div class="modal-footer">
+  <button type="button" class="btn btn-default" ng-click="$ctrl.dismiss()">Close</button>
+</div>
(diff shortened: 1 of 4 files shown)


sjakthol:
Thanks for the comments and sorry for the delay.

Here's a screenshot of how the expand button looks like (small button next to the refresh icon): https://user-images.githubusercontent.com/10127907/46026424-faac9380-c0f3-11e8-9453-7520c32cacc7.png

As for the second comment, I'm not sure if I was able to to parse your comment correctly. If you wanted to know whether I tried to use the Widget component in the modal dialog instead of the custom template, I would respond the following:

I thought about re-using the Widget component and hiding all the non-essential content when the component was used in the modal dialog but decided against it as eventually some new content will be added to the Widget component and it's very likely this won't be tested against the modal mode of the Widget (it's a small button after all) and cause some breakage in that view. Hence, I opted for creating a new component with the minimal required markup in it. While it's not optimal, IMO it's less likely to cause breakage in the future.

If my assumption about your second comment was not correct, please clarify and I'll try to address the correct concern :)

arikfr:
You understood it correctly :) But what content would you need to hide?

sjakthol:
The Widget component also includes bunch of functionality related to managing that widget in the dashboard.

For example, showing the top-right corner drop-down menu really only makes sense when they are part of a dashboard - not in a standalone modal dialog as some of the actions work on the underlying dashboard. As I'm not a huge Angular expert, I was afraid these might not work that well when taken out of the context they are expected to be in. Or at least some plumbing is required to make sure the modal dialog context is similar to the dashboard context the Widget component expects.

Similarly, I'm a bit afraid how well the parameter UI (part of the Widget component) works in this modal dialog if the same widget is rendered both in the dialog and in the dashboard below (e.g. will they stay  in sync). Hence, I would like to not have them visible in the modal view.

sjakthol:
I tried to use the dashboard-widget component to render the visualization in the modal dialog but couldn't really get it to work. The biggest problems were:

* The existing component didn't really fit the structure and style of modal dialog. For a nice-looking result, the title, body and the footer of the dashboard-widget component would've needed to be separated into the three sections of the modal dialog (header, body, footer).
* The content of the dashboard-widget contains links which allow the user to navigate to a different page. If one of these links are clicked, the page under the modal dialog navigates but the modal is left open. This is somewhat counter-intuitive, and I wasn't able to figure out how to make the modal close on navigation.

Hence, some duplication of code / logic is needed to get this feature to work without significant refactoring of the dashboard-widget component (and I don't really have time for doing that).

I made a small change to use the query-link component to render the modal title as a similar change was made to the dashboard-widget as well in ec17cc7eabdadf057a98bbf0684cc4c5ef077c5f.

I will leave this PR like this and it's up to you to decide if this is acceptable or not.

arikfr:
OK, it makes sense.

Last question: does it work properly when the query has parameters/filters? I'm not sure if it needs to render the parameters/filters UI, but does it retain the same state?

sjakthol:
Good point. I tested how it works with filters and parameters and the results were:
* If a widget has a parameter, its value is changed in the dashboard and the query is refreshed to contain results with that parameter, the the modal shows up-to-date version of the query with the parameter the query was last refreshed with (and rendered in the dashboard). The parameter selection box does not appear in the modal. So all good here.
* If a widget has a filter, the filter selection box is rendered in the modal also (it seems to be part of the visualization). If the filter value is changed in the dashboard, the modal popup inherits this state.

In the case of a filter, there was a minor problem with z-indexing of the modal dialog and the filter select dropdown. It appears that the default z-index of the ui-select widget is 1000 while the z-index of the modal dialog is at least 1050. Hence, if you tried to change the filter value from the modal dialog, it would render below the modal dialog.

i have made a fix for this to increase the z-index of the ui-select dropdown in case a modal dialog is open in this page. So filters should be good as well.

It should also be noted that if you change the filter value in the modal dialog, the change is reflected back to the dashboard. So all good here as well.

Please let me know if you can think of any other things that could go wrong (it's hard for me to check everything as I'm not so familiar with the different features Redash has).

arikfr:
Perfect. Thanks!

PR #2826, opened 2018-09-17
WIP: add API hook for schema browser display

alison985:
Original issue: https://github.com/mozilla/redash/issues/185
Original PR: https://github.com/mozilla/redash/pull/201

Issue to post this upstream: https://github.com/mozilla/redash/issues/430
Previous PR code reviews trying to port this upstream: https://github.com/getredash/redash/pull/2655 https://github.com/getredash/redash/pull/2764

Problem: [This javascript forEach loop](https://github.com/getredash/redash/compare/master...StantonVentures:alison_2654v3?expand=1#diff-ffb12a3b09d99f048b796d1abeae6d17R138) breaks the autoComplete feature/typing in the query textarea. Additional/random keys appear in it when typing. If you comment out this forEach loop the problem disappears. I have tried writing the loop different ways and this is the only one that really passes the linter. I have no idea why this interaction is occurring and would love any help or pointers in the right direction.

This PR:
* Adds the key font awesome icon to partition keys in the schema browser for Presto and Athena data sources.
* Adds a hook and API response to tell the front-end javascript which schema browser to use.
* Supports partition keys in the Athena Glue catalog generated schema.

### DIFF
--- a/client/app/components/queries/query-editor.js
+++ b/client/app/components/queries/query-editor.js
@@ -131,8 +131,18 @@ function queryEditor(QuerySnippet, $timeout) {
                 keywords[table.name] = 'Table';
 
                 table.columns.forEach((c) => {
-                  keywords[c] = 'Column';
-                  keywords[`${table.name}.${c}`] = 'Column';
+                  if (typeof c === 'string') {
+                    keywords[c] = 'Column';
+                    keywords[`${table.name}.${c}`] = 'Column';
+                  } else if (typeof c === 'object') {
+                    c.forEach((a, b) => {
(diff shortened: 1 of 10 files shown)


arikfr:
I don't remember the name right now, but there is a function in Lodash that does exactly this (returns only the properties you defined from a Javascript object).

While having support for different schema browser components makes sense, I don't think this is the case for it. To implement what's needed here it's enough to:

1. Extend the default schema browser to support showing icons next to the column names.
2. Extend the schema data structure to support having an icon metadata along with the column name.

PR #2833, opened 2018-09-17
Use server side sort order for tag list and show count of tagged items.

### LINKED ISSUE #2819: Improve tag list ordering on query and dashboard list
### Issue Summary

On the queries and dashboard list view the tags are ordered by usage count, but the sort order is ascending, meaning the tags with the *least* usage are shown on the top, the tags with the most tagged items on the bottom of the list.

I would suggest to do that to descending.

In terms of visualization, I wonder if it would make sense to show the number of items that individual tags have been assigned to?

Our QA team has filed a related issue to sort the tags alphabetical: https://github.com/mozilla/redash/issues/481

### Steps to Reproduce

1. Go to either queries or dashboard list with some tags assigned
2. See the list of tags being shown with the least tagged items first, not last.

### Technical details:

* Redash Version: 5.0beta
* Browser/OS: Firefox 62 / macOS 10.13.6

jezdez:
Fix #2819

It turned out to be just a problem of the client side sorting
the list of tags again even though the server side provided the
tags already correctly sorted.

### DIFF
--- a/client/app/components/NoTaggedObjectsFound.jsx
+++ b/client/app/components/NoTaggedObjectsFound.jsx
@@ -8,8 +8,8 @@ function NoTaggedObjectsFound({ objectType, tags }) {
     <BigMessage icon="fa-tags">
       No {objectType} found tagged with
       {Array.from(tags).map(tag => (
-        <span className="label label-tag" key={tag}>
-          {tag}
+        <span className="label label-tag" key={tag.name}>
+          {tag.name}
         </span>
       ))}.
     </BigMessage>
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -350,4 +350,16 @@ class QueryRefreshResource(BaseResource):
 
 class QueryTagsResource(BaseResource):
     def get(self):
-        return {t[0]: t[1] for t in models.Query.all_tags(self.current_user, True)}
+        """
+        Returns all query tags including those for drafts.
+        """
+        tags = models.Query.all_tags(self.current_user, include_drafts=True)
+        return {
+            'tags': [
+                {
+                    'name': name,
+                    'count': count,
+                }
+                for name, count in tags
+            ]
+        }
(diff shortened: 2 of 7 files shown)


client/app/components/NoTaggedObjectsFound.jsx
```diff
@@ -8,8 +8,8 @@ function NoTaggedObjectsFound({ objectType, tags }) {
     <BigMessage icon="fa-tags">
       No {objectType} found tagged with
       {Array.from(tags).map(tag => (
-        <span className="label label-tag" key={tag}>
-          {tag}
+        <span className="label label-tag" key={tag.name}>
+          {tag.name}
```
jezdez: I'm hoping that this is correct React? 😬

alison985:
LGTM, but I'm confused why there would be a distinct call and a groupby call [on this line](https://github.com/getredash/redash/blob/b1f5d60460664b4f604f2fa699b0814f8b701416/redash/models.py#L993) that this PR references.

redash/handlers/queries.py
```diff
@@ -350,4 +350,16 @@ def post(self, query_id):
 
 class QueryTagsResource(BaseResource):
     def get(self):
-        return {t[0]: t[1] for t in models.Query.all_tags(self.current_user, True)}
+        """
+        Returns all query tags including those for drafts.
+        """
+        tags = models.Query.all_tags(self.current_user, include_drafts=True)
+        return {
+            'tags': [
```
arikfr: Maybe it makes sense to change the model functions (`all_tags`) to return a dictionary object in the first place?

jezdez: I thought about that as well, but I tend to lean towards data model classes that return queries instead of serialized versions of the queries and do the transformation on the view layer depending on the needs of the view. That way the separation of concerns is kept and prevents the creation of "fat models", which deal with all kinds of app logic.

Would compacting this code help you?

```python
tags = models.Query.all_tags(self.current_user, include_drafts=True)
return {
    'tags': [dict(zip(['name', 'count'], tag)) for tag in tags]
}
```

jezdez:
The list-ctrl branch was merged, so this is only the actual changes now.

PR #2848, opened 2018-09-25
Add DB2 as a data source using ibm-db python package

nicof38:
Add DB2 as datasource using ibm-db as DB2 python driver.
This package is now added to the requirement-ds file and an DB2 icon has been added on client side.

May need to decide if this query runner is activated by default or not ? Right now, this is not the case (.settings/__init__.py has not been modified to enable this query runner)


redash/query_runner/db2.py
```diff
@@ -0,0 +1,147 @@
+import os
+import json
+import logging
+import select
+
+import ibm_db_dbi
```
arikfr: Please follow a pattern similar to this one:
https://github.com/getredash/redash/blob/99c73aef2d925cac269dbfd2b763b6961f68992f/redash/query_runner/big_query.py#L16-L25

This will make the import safe and the data source disabled in case of the import failing.

redash/query_runner/db2.py
```diff
@@ -0,0 +1,147 @@
...
+    ibm_db_dbi.FLOAT: TYPE_FLOAT,
+    ibm_db_dbi.DECIMAL: TYPE_FLOAT,
+    ibm_db_dbi.DATE: TYPE_DATE,
+    ibm_db_dbi.TIME: TYPE_DATETIME,
+    ibm_db_dbi.DATETIME: TYPE_DATETIME,
+    ibm_db_dbi.BINARY: TYPE_STRING,
+    ibm_db_dbi.XML: TYPE_STRING,
+    ibm_db_dbi.TEXT: TYPE_STRING,
+    ibm_db_dbi.STRING: TYPE_STRING
+}
```
arikfr: This should be moved into the try/catch block of the import too, as `ibm_db_dbi` might not be available in this case.

redash/query_runner/db2.py
```diff
@@ -0,0 +1,147 @@
...
+        results, error = self.run_query(query, None)
+
+        if error is not None:
+            raise Exception("Failed getting schema.")
+
+        results = json.loads(results)
+
+        for row in results['rows']:
+            if row['TABLE_SCHEMA'] != 'public':
+                table_name = '{}.{}'.format(row['TABLE_SCHEMA'], row['TABLE_NAME'])
```
arikfr: Please create a unicode string here (i.e. prefix with `u'` instead of `'`).

nicof38: OK, added

redash/query_runner/db2.py
nicof38: Ok, rewritten

arikfr:
Thanks!

benjaminsingleton:
I'm getting the same error when trying to connect to DB2 with Apache Superset so I suspect this has something to do with the DB2Express-C docker container. Not a Redash bug.

arikfr:
@benjaminsingleton it's probably configuration issue: the Docker container you launched for db2 is on a different network than your Redash setup, so you can't use `127.0.0.1` as the host. 

Also, if you have usage questions like these, please use [the forum](https://discuss.redash.io).

PR #2849, opened 2018-09-25
Fix invalid reference to alert.to_dict() in webhook

wankdanker:
It appears that alert.to_dict() is no longer a thing. In most other places
it has been replaced by serialize_alert(), but looks like this was missed.
The reference to alert.to_dict in destinations/webhook.py has been replaced
by a similar call to serialize_alert.

This fixes broken webhook processing for me and I assume others.


arikfr:
Thanks!

Issue #2445, opened 2018-04-06
Docker based setup script/cloud images


arikfr:
Following #896, it's time to update our [setup script](https://github.com/getredash/redash/blob/master/setup/ubuntu/bootstrap.sh) to use  🐳 Docker based deployment. This has many benefits among them easier and more stable upgrades (always using updated dependencies) and more standard setup across different platforms.

## New setup script

Once we move to use Docker, the setup script will do:

* Install Docker & Docker Compose.
* Download `docker-compose.yml` file and other supporting files.
* Start services.
* Done.

## New upgrade script

The upgrade script will probably be something along the lines of:

* Get latest version Docker image tag reference.
* Update `docker-compose.yml`.
* Rebuild services.
* Stop services.
* Apply migrations if needed.
* Start services.

## Documentation

Will have to update documentation accordingly. Mainly:

* [Setup guide](https://redash.io/help/open-source/setup) ([source](https://github.com/getredash/website/blob/master/website/_kb/open-source/setup.md))
* [How to upgrade](https://redash.io/help/open-source/admin-guide/how-to-upgrade) ([source](https://github.com/getredash/website/edit/master/website/_kb/open-source/admin-guide/how-to-upgrade.md))
* [On going maintenance](https://redash.io/help/open-source/admin-guide/maintenance) ([source](https://github.com/getredash/website/edit/master/website/_kb/open-source/admin-guide/maintenance.md))

## Control script

While controlling, checking logs, etc, can be done with the `docker-compose` command we should consider streamlining the process with a single control script.

I'm not 100% sure this is needed, but it can improve end user (admin) experience. On the otherhand, it makes the knowledge less portable.

---

This is an outline of the work required and some initial thoughts. Any feedback is welcome. Specially from users who already use Docker/Docker Compose for their Redash deployment.

lkraider:
The basic upgrade workflow I have been using is:

1. Edit `docker-compose.production.yml` and point to specific tag you want to deploy:
```diff
-    image: redash/redash:latest
+    image: redash/redash:4.0.0-rc.1
```

2. Pull the same tagged image:
```
docker pull redash/redash:4.0.0-rc.1
```

3. Redeploy the service:
```
docker-compose -f docker-compose.production.yml -p redash up -d
```

4. Run migrations
Still no clue how to do that correctly.

PR #2850, opened 2018-09-25
Docker based setup scripts

### LINKED ISSUE #896: Standardize on Docker as the primary deployment option

arikfr:
Docker based setup script for Redash on top of Ubuntu 18.04.

This extends the work done in #2463. Closes #896, #2445.


jeffwidman:
🍰

arikfr:
@jeffwidman 2.5 years after your vision is finally implemented 😅

PR #2852, opened 2018-09-26
Change placement (right/bottom) of chart legend depending on chart width

kravets-levko:
Issue getredash/redash#2796: When chart width is smaller than 600px (value chosen experimentally) place legend below the chart. 

Screenshots:

image
image

image
image


arikfr:
Thanks :)

Issue #2285, opened 2018-02-04
Add option to load cached results in Query Results queries


arikfr:
Currently when running a query that loads data from another query using the query results data source, we always execute the underlying query. Sometimes it makes sense to use a cached result.

We can start with a way to express it when specifying the query you want to load:

```sql
SELECT *
FROM cached_query_123
```

(instead of `FROM query_123`)

whitemark:
Thumb up!

ccwillia:
This would be a great feature to compare counts between the last time the query ran and the current run to see the percent change

chanansh:
I wish I could query based on another query WITH parameters. This way I can avoid running the same shared logic twice. A->B and A->C where A is the parent query and B and C are child queries and A has parameters.

PR #2855, opened 2018-09-26
Add option to query cached results

yoavbls:
#2285 
I Implemented it exactly like Arik's example in the issue 😃 
image

### DIFF
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -47,6 +47,11 @@ def extract_query_ids(query):
     return [int(q) for q in queries]
 
 
+def extract_cached_query_ids(query):
+    queries = re.findall(r'(?:join|from)\s+cached_query_(\d+)', query, re.IGNORECASE)
+    return [int(q) for q in queries]
+
+
 def _load_query(user, query_id):
     query = models.Query.get_by_id(query_id)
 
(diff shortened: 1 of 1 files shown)


redash/query_runner/query_results.py
```diff
@@ -60,21 +65,29 @@ def _load_query(user, query_id):
     return query
 
 
-def create_tables_from_query_ids(user, connection, query_ids):
-    for query_id in set(query_ids):
+def get_query_results(user, query_id, bring_from_cache):
         query = _load_query(user, query_id)
+        if bring_from_cache and query.latest_query_data_id is not None:
```
arikfr: When there are no results stored and the user asked for cache we shouldn't fallback to running the query but rather show an error message saying this ("No cached result available for query xxx.").

arikfr: Also seems like the indentation here is a bit off.

redash/query_runner/query_results.py
```diff
@@ -60,21 +65,29 @@ def _load_query(user, query_id):
...
-                "Failed loading results for query id {}.".format(query.id))
-
-        results = json.loads(results)
+def create_tables_from_query_ids(user, connection, query_ids, cached_query_ids=[]):
+    for query_id in set(query_ids):
+        results = get_query_results(user, query_id, False)
         table_name = 'query_{query_id}'.format(query_id=query_id)
         create_table(connection, table_name, results)
 
+    for query_id in set(cached_query_ids):
```
arikfr: It's worth loading the cached results first.

redash/query_runner/query_results.py
yoavbls: Sounds right, i fixed it

redash/query_runner/query_results.py
yoavbls: I hope you mean to run this for loop first, 
That's what I changed in my last commit

yoavbls:
Thank you for the code review,
I hope its fine now :)
(and sorry for the reference from another pull request, I fixed it but its still appears)

arikfr:
Thanks!

PR #2866, opened 2018-09-27
feat(redshift): hide tables the configured user cannot access

sjakthol:
The SVV_COLUMNS table used to determine the tables of a
Redshift database includes all tables, even those the
current user cannot access, by default. These tables clutter
the schema browser and make it harder for users to determine
which tables they should have access to.

These changes modify the Redshift query runner so that
tables the configured user cannot access are filtered
out from the database schema. The checks are two-fold:

* First, the query uses HAS_SCHEMA_PRIVILEGE to check if the
  current user has USAGE rights on a schema the given table
  belongs to. This privilege is required to access any of
  the tables in a schema.

* Second, the query tries to determine if the current user
  has SELECT access to the given table. Two cases need to
  be considered here:

  * First, we need to check if the table is part of an
    external schema. Access to tables in external schema
    is controlled at schema level - you cannot grant or
    revoke access to specific external tables. Additionally,
    the HAS_TABLE_PRIVILEGE returns an error if it is asked
    to give a verdict for an external table.

    Hence, the query checks if the schema a specific table
    belongs to is an external schema and if it is, the table
    is included in the list (if we got here the user already
    had USAGE on the given schema). This check short-circuits
    the table-level access check for external tables which
    means they are never passed to HAS_TABLE_PRIVILEGE().

  * Then, if the table was not part of an external schema,
    the HAS_TABLE_PRIVILEGE() function is used to determine
    if the user has SELECT access to the given table. The
    table is included in the schema if this check passes.

Together this condition ensures that tables the user definitely
cannot access are not included in the schema browser.

These changes have been tested to work in an environment that
includes normal and external schemas, normal and late-binding
views, and normal and external tables.

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -242,6 +242,11 @@ class Redshift(PostgreSQL):
         # https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_COLUMNS.html
         # Use PG_GET_LATE_BINDING_VIEW_COLS to include schema for late binding views data for Redshift
         # https://docs.aws.amazon.com/redshift/latest/dg/PG_GET_LATE_BINDING_VIEW_COLS.html
+        # Use HAS_SCHEMA_PRIVILEGE(), SVV_EXTERNAL_SCHEMAS and HAS_TABLE_PRIVILEGE() to filter
+        # out tables the current user cannot access.
+        # https://docs.aws.amazon.com/redshift/latest/dg/r_HAS_SCHEMA_PRIVILEGE.html
+        # https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_EXTERNAL_SCHEMAS.html
+        # https://docs.aws.amazon.com/redshift/latest/dg/r_HAS_TABLE_PRIVILEGE.html
         query = """
         WITH tables AS (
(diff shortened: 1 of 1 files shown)


sjakthol:
Rebased to resolve some merge conflicts.

redash/query_runner/pg.py
```diff
@@ -260,6 +265,12 @@ def _get_tables(self, schema):
         )
         SELECT table_name, table_schema, column_name
         FROM tables
+        WHERE
+            HAS_SCHEMA_PRIVILEGE(table_schema, 'USAGE') AND
+            (
+                table_schema IN (SELECT schemaname FROM SVV_EXTERNAL_SCHEMAS) OR
+                HAS_TABLE_PRIVILEGE(table_schema || '.' || table_name, 'SELECT')
```
arikfr: Just to make sure I understand: when a table is an external one, then `HAS_TABLE_PRIVILEGE` does not apply and having USAGE permission for the schema is enough?

arikfr: Never mind -- just read your superb description that explains all of that.

arikfr:
Merged. Thank you for the great work and detailed description!

PR #2867, opened 2018-09-28
Fix: TreasureData data source - deduplicate column names

zaimy:
The query `SELECT 'foo' AS column_name, 'bar' AS column_name;` against a TreasureData data source generates results: `('bar', 'bar')` instead of expected results: `('foo', 'bar')`.

image

It seems to be the same cause as the problem of https://github.com/getredash/redash/issues/847.
I fixed it to rename duplicate column names like other data sources.

### DIFF
--- a/redash/query_runner/treasuredata.py
+++ b/redash/query_runner/treasuredata.py
@@ -102,16 +102,13 @@ class TreasureData(BaseQueryRunner):
         cursor = connection.cursor()
         try:
             cursor.execute(query)
-            columns_data = [(row[0], cursor.show_job()['hive_result_schema'][i][1]) for i,row in enumerate(cursor.description)]
-
-            columns = [{'name': col[0],
-                'friendly_name': col[0],
-                'type': TD_TYPES_MAPPING.get(col[1], None)} for col in columns_data]
+            columns_tuples = [(row[0], cursor.show_job()['hive_result_schema'][i][1]) for i,row in enumerate(cursor.description)]
+            columns = self.fetch_columns(columns_tuples)
 
             if cursor.rowcount == 0:
                 rows = []
             else:
-                rows = [dict(zip(([c[0] for c in columns_data]), r)) for i, r in enumerate(cursor.fetchall())]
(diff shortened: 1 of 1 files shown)


zaimy:
Sorry, I did not map types. Fixed it with https://github.com/getredash/redash/pull/2867/commits/7026dde72576bb867f6d1e51b64029210a34bf37 .

zaimy:
Althought codeclimate is failing but I think refactoring is not the scope of this pull request. I would like to hear opinions from other people.

zaimy:
@arikfr 
Hello, could you tell me your opinion?

redash/query_runner/treasuredata.py
```diff
@@ -102,16 +102,13 @@ def run_query(self, query, user):
         cursor = connection.cursor()
         try:
             cursor.execute(query)
-            columns_data = [(row[0], cursor.show_job()['hive_result_schema'][i][1]) for i,row in enumerate(cursor.description)]
-
-            columns = [{'name': col[0],
-                'friendly_name': col[0],
-                'type': TD_TYPES_MAPPING.get(col[1], None)} for col in columns_data]
+            columns_tuples = [(i[0], TD_TYPES_MAPPING.get(i[1], None)) for i in cursor.show_job()['hive_result_schema']]
```
arikfr: Previous implementation was using `cursor.description` while current one is just using `cursor.show_job()`. I'm not familiar enough with TD to know if this is supported across all versions. Can you explain the difference?

zaimy: Yes, both call the same Treasure Data API request in td-client. `hive_result_schema` included in this response is an array of column name and type.

The difference between the two is that `cursor.show_job()` returns  `hive_result_schema` and other responce directly, but `cursor.description` retrieves and returns only the column names from `hive_result_schema`.

Previous implementation got column names from `cursor.description` and got types from `cursor.show_job()`, but I thought that using `cursor.show_job()` only is simpler.

Because the implementation of `cursor.show_job()` has not been changed since the DB API was introduced to td-client, it should work fine for all versions.

arikfr:
Thanks!

zaimy:
Thanks too!

Issue #2858, opened 2018-09-27
Cannot retrieve Athena/Glue schema


kadrach:
### Issue Summary

Redash cannot retrieve Athena(Glue) schema 

### Steps to Reproduce

- Setup an Athena data source using Glue catalog
- Attempt to list schemas in either UI or via API directly

```
% http redash.clasia.internal/api/data_sources/1/schema Authorization:redacted
HTTP/1.1 200 OK
Connection: keep-alive
Content-Encoding: gzip
Content-Type: application/json
Date: Thu, 27 Sep 2018 03:05:09 GMT
Server: nginx/1.9.10
Transfer-Encoding: chunked

{
    "error": {
        "code": 2,
        "message": "Error retrieving schema."
    }
}
```

Changing the redash log level to debug, and monitoring logs via `docker-compose logs --tail=0 --follow` I can see the boto debug logger listing and describing all Glue catalog tables and their data schemas. This does not appear to be an IAM permissions issue. There is no indication of an internal failure within redash.

### Technical details:

* Redash Version: 5.0.0+b4754
* Browser/OS: Backend issue
* How did you install Redash: From AMI ami-0fcac8599721ee168

kadrach:
https://github.com/getredash/redash/blob/53abc1678045c89fe619cb1581e52321e6abc8df/redash/query_runner/athena.py#L135

This entry is not present in all tables. Let me see if I can narrow it down.

arikfr:
Thank you for the extra details. Maybe it's a View or some other object that doesn't have a partition key?

kadrach:
Yeah that's correct. The patch below fixes the issue for me.

```diff
diff --git a/redash/query_runner/athena.py b/redash/query_runner/athena.py                          
index a0795293..18c6ed3b 100644                                                                     
--- a/redash/query_runner/athena.py                                                                 
+++ b/redash/query_runner/athena.py                                                                 
@@ -132,7 +132,7 @@ class Athena(BaseQueryRunner):                                                  
                 if table_name not in schema:                                                       
                     column = [columns['Name'] for columns in table['StorageDescriptor']['Columns']]
                     schema[table_name] = {'name': table_name, 'columns': column}                   
-                    for partition in table['PartitionKeys']:                                       
+                    for partition in table.get('PartitionKeys', []):                               
                         schema[table_name]['columns'].append(partition['Name'])                    
                                                                                                    
         return schema.values()                                                                     
```

arikfr:
[...]

PR #2868, opened 2018-09-28
Address edgecase when retrieving Glue schemas for Athena query runner.

kadrach:
Fixes getredash/redash#2858.


kadrach:
Something that just occured to me - Athena can't read JDBC tables, although they can be present in the Glue catalog. I'll need to look into how they can be identified.

arikfr:
Merged, thanks!

The added unit tests are very much appreciated. 👏

Issue #2854, opened 2018-09-26
pivot Visualization Name does not render on embeds 


jbmcclain:
### Issue Summary

pivot Visualization Name in the dashboard shared "embed into a webpage" does not render anymore 
One thing i did notice is that the version on the install now says 18.04 versus 16.04 should i update to the Ubuntu 18.04 first do you think this would make a diffrence?

### Steps to Reproduce

1. create the pivot table
2. create the dashboard
3. add the pivot to dashboard
4. share link and embed into a web-page

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* Redash Version: just updated to 5.0.0+b4755
* Browser/OS: chrome and firefox / ubuntu windows/ubuntu
* How did you install Redash: selfhosted on ubuntu 16.04 originally 

![image](https://user-images.githubusercontent.com/1903662/46095205-f6cb5080-c189-11e8-84ad-e696cb6387a5.png)

arikfr:
Ubuntu version won't make a difference here :)

It's probably a bug we introduced when updating the way titles are rendered. We will take a look.

Thanks for the report!

PR #2870, opened 2018-09-28
Widget titles wrong rendering on public dashboards

kravets-levko:
Issue getredash/redash#2854 (affects all types of visualizations - not only pivot table).

**Screenshots:**

In-app:
image

Public:
image

Print:
image


jbmcclain:
what is the best way to run npm so it updates the web pack?

PR #2872, opened 2018-09-28
Use Plotly's function to clean y-values (x may be category or date/time)

kravets-levko:
(Bug reported by Redash user).

Plotly can detect and use values with some extra characters (like `$6,123.88`). Our custom code that extends Plotly wasn't able to use such values and therefore some functions didn't work (tooltips, stacking etc.). This PR adds support for such values (using Plotly's function to clean data).


client/app/visualizations/chart/plotly/utils.js
```diff
@@ -36,6 +37,10 @@ export const ColorPalette = Object.assign({}, BaseColors, {
 
 const ColorPaletteArray = values(BaseColors);
 
+function cleanNumber(value) {
+  return isUndefined(value) ? value : (plotlyCleanNumber(value) || 0.0);
```
arikfr: What if value is not supposed to be a number? I think it will be safer to return `value` if `plotlyCleanNumber` return `undefined` (`BADNUM`).

Also why not have it as part of `normalizeValue` call?

Does the tooltip even take normalized values?

kravets-levko: Generally, all y-values (y, y-error, size) should be numbers; it means that actually we don't need to use `normalizeValue` there (noticed it after your comment), but just `cleanNumber`; in opposite, x-values may be any (date/time, category, number), so we should sanitize them with `normalizeValue`, but not with `cleanNumber` (it's not important for x-values because we dont do any computing with them - just use as-is; but we use y-values to implement stacking (line/area chars) and percent normalization, so it **should** be a number).

kravets-levko: @arikfr Updated the PR: there is the only chart type where `y` may be non-number - it's `bubble` chart; for other types non-numeric `y` does not make any sense.

client/app/visualizations/chart/plotly/utils.js
```diff
@@ -248,7 +253,7 @@ function preparePieData(seriesList, options) {
     }, 0);
     each(serie.data, (row) => {
       const x = normalizeValue(row.x);
-      const y = normalizeValue(row.y);
+      const y = cleanNumber(normalizeValue(row.y));
```
arikfr: This is not used in case of bubble charts?

kravets-levko: It's only for pie charts (y there can be only numeric). Also I noticed a bug nearby - so fixed and updated PR

PR #2877, opened 2018-09-28
Fix: wrong reference (EditInPlaceText -> EditInPlace)

arikfr:


jezdez:
Wait, wait, the class is called EditInPlaceText in the EditInPlace module. Shouldn't it be `import EditInPlaceText from './EditInPlace';`?

PR #2879, opened 2018-09-29
Circle CI: Add flake8 tests to find syntax errors and undefined names

cclauss:
Add [flake8](http://flake8.pycqa.org) tests to find Python syntax errors and undefined names.

__E901,E999,F821,F822,F823__ are the "_showstopper_" flake8 issues that can halt the runtime with a SyntaxError, NameError, etc. Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

PR #2878, opened 2018-09-28
Fixed the fix.

jezdez:
Fix for #2877.

Clicked merge too fast 😬

PR #2880, opened 2018-09-29
from redash.query_runner import register for line 23

cclauss:
Missing import discovered via #2879

### DIFF
--- a/redash/query_runner/databricks.py
+++ b/redash/query_runner/databricks.py
@@ -1,5 +1,6 @@
 import base64
-from .hive_ds import Hive
+from redash.query_runner import register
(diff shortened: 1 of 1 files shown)


redash/query_runner/databricks.py
```diff
@@ -1,5 +1,6 @@
 import base64
-from .hive_ds import Hive
+from redash.query_runner import register
```
jezdez: Good catch, could you please keep using relative imports though?

```python
from . import register
from .hive_ds import Hive
```

jezdez:
FTR, the unit-tests have failed because of #2878. A rebase should help.

cclauss:
Closing in favor of #2880

PR #2884, opened 2018-10-01
Rename Yandex Metrika to Metrica.

### LINKED ISSUE #2874: Fix Yandex Metrica name
Should be Yandex Metrica. Need to apply a fix, but keep the type name with a `k` to avoid breaking existing deployments.

jezdez:
Fix #2874.


arikfr:
Thanks, @jezdez .

PR #2887, opened 2018-10-02
Fix TypeError in edit-in-place component.

jezdez:
The issue being that the default export was called automatically and didn’t accept a ngModule.

### DIFF
--- a/client/app/components/edit-in-place.js
+++ b/client/app/components/edit-in-place.js
@@ -1,6 +1,0 @@
-import { react2angular } from 'react2angular';
-import EditInPlace from './EditInPlace.jsx';
-
-export default function init(ngModule) {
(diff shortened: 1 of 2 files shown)


client/app/components/edit-in-place.js
```diff
@@ -1,6 +0,0 @@
-import { react2angular } from 'react2angular';
```
jezdez: I don't see why this file was needed, since it automatically turns the React component into the edit-in-place Angular component. @washort Can you remember why you had both edit-in-place.js and EditInPlace.jsx?

arikfr:
https://github.com/getredash/redash/pull/2886

jezdez:
@arikfr Wow!

Issue #2861, opened 2018-09-27
PDF Export


dignati:
We would like the ability to export dashboards as PDFs both via UI and as an API call. 

I believe the only practical solution to generating PDFs would be to use a library like [pdfkit](https://pypi.org/project/pdfkit/) that basically screenshots websites. To generate PDFs directly we would need to replicate the javascript code that is rendering visualizations. 

Would you be open to a PR implementing this feature as described? Me and a coworker could start working on this soon.

arikfr:
I think a better solution will be to use [puppeteer](https://github.com/GoogleChrome/puppeteer), which uses Chrome. Otherwise you will probably have rendering issues with wkhtmltopdf, which probably won't support some of the Javascript we use.

We're definitely open to such PR, but would like to discuss the implementation first to make sure it's something we can support in the long run.

Thanks!

dignati:
Should we use [pyppeteer](https://github.com/miyakogi/pyppeteer) then?

arikfr:
Nope, pyppeteer works with Python 3.6+, so won't work in the same process as Redash anyway.

dignati:
So are you suggesting to run a nodejs process on the server to use puppeteer? I'm making process on adding basic UI to redash so that I can test a couple different backend approaches.

arikfr:
On the server or (preferably) as an additional Docker container.

> I'm making process on adding basic UI to redash

basic ui for this feature?

dignati:
Oh okay, additional container seems doable. 

> basic ui for this feature?

Yes. Basically just a download as pdf button in the dashboard context.

PR #2890, opened 2018-10-03
WIP: Add Dashboard PDF download 

dignati:
Heavy WIP implementation of #2861 because I want to get some early feedback and open a discussion. I'm very inexperienced in Angular and Python so any comments are appreciated.

- [ ] UI
  - [x] Download Button
  - [ ] Maybe progress indicator in case PDF generation takes a while
- [x] API call
- [ ] Backend Service
- [ ] Tests

The next step would be to use [puppeteer](https://github.com/GoogleChrome/puppeteer) to create a backend service that can be called from the API endpoint to actually generate the PDF. I would be happy to get some suggestions on how this should look. Probably a docker container with puppeteer and a tiny http API that takes a URL and returns a PDF? Or we could start the docker container in the API every time it is needed?

### DIFF
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -23,6 +23,7 @@ function getWidgetsWithChangedPositions(widgets) {
 }
 
 function DashboardCtrl(
+  $scope,
   $routeParams,
   $location,
   $timeout,
@@ -178,6 +179,7 @@ function DashboardCtrl(
       { slug: $routeParams.dashboardSlug },
       (dashboard) => {
         this.dashboard = dashboard;
+        $scope.dashboard = dashboard;
         this.isDashboardOwner = currentUser.id === dashboard.user.id || currentUser.hasPermission('admin');
         Events.record('view', 'dashboard', dashboard.id);
         renderDashboard(dashboard, force);
@@ -247,6 +249,11 @@ function DashboardCtrl(
         aclUrl: { url: `api/dashboards/${this.dashboard.id}/acl` },
       },
     });
+  };
+
+  this.downloadPdf = () => {
+    Events.record('donwloadPdf', 'dashboard', this.dashboard.id);
+    this.dashboard.$downloadPdf();
   };
 
   this.editLayout = (enableEditing, applyChanges) => {
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -179,7 +179,6 @@ function DashboardCtrl(
       { slug: $routeParams.dashboardSlug },
       (dashboard) => {
         this.dashboard = dashboard;
-        $scope.dashboard = dashboard;
         this.isDashboardOwner = currentUser.id === dashboard.user.id || currentUser.hasPermission('admin');
         Events.record('view', 'dashboard', dashboard.id);
         renderDashboard(dashboard, force);
--- a/redash/handlers/api.py
+++ b/redash/handlers/api.py
@@ -23,7 +23,7 @@ from redash.handlers.settings import OrganizationSettings
 from redash.handlers.favorites import QueryFavoriteListResource, QueryFavoriteResource, DashboardFavoriteListResource, DashboardFavoriteResource
 from redash.handlers.queries import QueryTagsResource
 from redash.handlers.dashboards import DashboardTagsResource
-
+from redash.handlers.pdf import DashboardPdfDownload
 
 class ApiExt(Api):
     def add_org_resource(self, resource, *urls, **kwargs):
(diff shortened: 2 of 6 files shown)


client/app/pages/dashboards/dashboard.js
```diff
@@ -178,6 +179,7 @@ function DashboardCtrl(
       { slug: $routeParams.dashboardSlug },
       (dashboard) => {
         this.dashboard = dashboard;
+        $scope.dashboard = dashboard;
```
dignati: This seems kind of ugly, `$scope` wasn't used in this file up until now. I haven't found any other way to be able to use the dashboard id in the angular template for the dashboard though.

client/app/pages/dashboards/dashboard.js
```diff
@@ -249,6 +251,11 @@ function DashboardCtrl(
     });
   };
 
+  this.downloadPdf = () => {
```
dignati: Have to check if this even runs when a click on the download button starts the download anyway.

redash/handlers/api.py
```diff
@@ -69,10 +69,10 @@ def json_representation(data, code, headers=None):
 
 api.add_org_resource(EventsResource, '/api/events', endpoint='events')
 
-api.add_org_resource(QueryFavoriteListResource, '/api/queries/favorites', endpoint='query_fovorites')
-api.add_org_resource(QueryFavoriteResource, '/api/queries/<query_id>/favorite', endpoint='query_fovorite')
-api.add_org_resource(DashboardFavoriteListResource, '/api/dashboards/favorites', endpoint='dashboard_fovorites')
-api.add_org_resource(DashboardFavoriteResource, '/api/dashboards/<object_id>/favorite', endpoint='dashboard_fovorite')
+api.add_org_resource(QueryFavoriteListResource, '/api/queries/favorites', endpoint='query_favorites')
```
dignati: I assume these were typos?

client/app/pages/dashboards/dashboard.js
kravets-levko: `$ctrl.dashboard.id`, no?

redash/handlers/pdf.py
```diff
@@ -0,0 +1,28 @@
...
+
+from flask import request, send_file
+from redash import models
+from redash.permissions import require_access, view_only, require_permission
+from redash.serializers import serialize_dashboard
+from redash.handlers.base import BaseResource, get_object_or_404
+
+class DashboardPdfDownload(BaseResource):
+    def get(self, dashboard_slug=None):
+        "dashboard = get_object_or_404(models.Dashboard.get_by_slug_and_org, dashboard_slug, self.current_org)"
```
kravets-levko: What is this for?

redash/handlers/pdf.py
```diff
@@ -0,0 +1,28 @@
...
+
+        self.record_event({
+            'action': 'download_pdf',
+            'object_type': 'dashboard',
```
kravets-levko: I think, it's useful to record also dashboard id here

redash/handlers/pdf.py
dignati: Nothing yet, this is copy paste leftover. I assume we need to get the url for the dashboard so that we can pass it on to puppeteer.

redash/handlers/pdf.py
dignati: Will do.

client/app/pages/dashboards/dashboard.js
dignati: Does not seem to work when I use it here: https://github.com/getredash/redash/pull/2890/files#diff-374cafdf3778c7e0169295ee5878667bR66

dignati:
> It will take a while :) Depends on how much resources the machine has and such, but a few seconds at least.
Yeah, seems likely. This might be outside of my UI Design capabilities though.

> My suggestion is let's make this implementation detail transparent to Redash, by using a configuration where you pass a URL of the snapshot service to Redash. This way the end user can deploy it with Docker, Serverless or whatever.

Okay, this sounds good. Should the new service be developed here, in-tree or in an extra repo?

arikfr:
> Okay, this sounds good. Should the new service be developed here, in-tree or in an extra repo?

I think a separate repo makes sense. I will open one tomorrow.

dignati:
The dashboard API endpoint does not seem to accept a `?api_key=` parameter like the query endpoint does. This needs to be added first so that the snap service can access the dashboard.

client/app/pages/dashboards/dashboard.js
dignati: This does in fact work - it was a problem with my IDE, which incorrectly highlighted {{ $ctrl.dashboard.id }} as a syntax error. Thank you 👍

arikfr:
> It is unclear to me how the snap service should authorize itself to the redash API to get access to the dashboard. There is a /embed endpoint for queries that accepts a api_key as URL parameter which works fine. The /public endpoint that provides something similar for dashboards however requires the dashboard to be public. Maybe we can extend /embed to support dashboards?

It's an open question for now :) The best solution will be to have some temporary API key that gives access to this dashboard. For now you can develop this using a user API key, until we figure out a better way.

arikfr:
> Okay, I'm having trouble using the User API key. 

What is the issue?

> Do I understand correctly that the snap service should be able to set HTTP Header Authorization to Key $user_api_key and then "see" the dashboard page?

Yes.

arikfr:
Hi,

Thanks.

Karthik-PlantSense:
hi anyone made the export dashboard to pdf on React
@arikfr @dignati

Issue #2430, opened 2018-04-01
Explicit configuration for Counter label


arikfr:
Currently the visualization name is used as the counter label. We need a separate explicit configuration option for it.

Vitaminl:
I found this issue this morning. Any solutions?

RichardLitt:
@Vitaminl Not yet! If you're keen to help out, we'd love a PR for this.

PR #2900, opened 2018-10-04
Add Counter label

ralphilius:
Issue getredash/redash#2430

**Screenshot**
image

### DIFF
--- a/client/app/visualizations/counter/counter-editor.html
+++ b/client/app/visualizations/counter/counter-editor.html
@@ -8,6 +8,12 @@
     </li>
   </ul>
   <div ng-show="currentTab == 'general'" class="m-t-10 m-b-10">
+      <div class="form-group">
+          <label class="col-lg-6">Counter Label</label>
+          <div class="col-lg-6">
+            <input type="text" ng-model="visualization.options.counterLabel" class="form-control" ng-trim="false">
+          </div>
+        </div>
     <div class="form-group">
       <label class="col-lg-6">Counter Value Column Name</label>
       <div class="col-lg-6">
--- a/client/app/visualizations/counter/index.js
+++ b/client/app/visualizations/counter/index.js
@@ -41,6 +41,11 @@ function CounterRenderer($timeout) {
           const targetRowNumber = getRowNumber($scope.visualization.options.targetRowNumber, queryData.length);
           const counterColName = $scope.visualization.options.counterColName;
           const targetColName = $scope.visualization.options.targetColName;
+          const counterLabel = $scope.visualization.options.counterLabel;
+          
+          if (counterLabel){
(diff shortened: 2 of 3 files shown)


client/app/visualizations/counter/counter-editor.html
```diff
@@ -8,6 +8,12 @@
     </li>
   </ul>
   <div ng-show="currentTab == 'general'" class="m-t-10 m-b-10">
+      <div class="form-group">
+          <label class="col-lg-6">Counter Label</label>
+          <div class="col-lg-6">
+            <input type="text" ng-model="visualization.options.counterLabel" class="form-control" ng-trim="false">
```
jezdez: Could you add a placeholder attribute that shows that the counter label defaults to the visualization name?

ralphilius: @jezdez Added :)

client/app/visualizations/counter/index.js
```diff
@@ -41,6 +41,11 @@ function CounterRenderer($timeout) {
           const targetRowNumber = getRowNumber($scope.visualization.options.targetRowNumber, queryData.length);
           const counterColName = $scope.visualization.options.counterColName;
           const targetColName = $scope.visualization.options.targetColName;
+          const counterLabel = $scope.visualization.options.counterLabel;
+
+          if (counterLabel) {
+            $scope.counterLabel = counterLabel;
+          }
```
jezdez: Just to clarify, and maybe I misunderstood the feature, does this actually keep falling back to the visualization name as I assumed when the `<counter-name>` element's content is changed to `counterLabel`?

jezdez: Currently master uses the viz name for the Counter label, and I wanted to make sure that this is backward compatible. In my previous comment about the placeholder I assumed that your patch would stay backward-compatible, but I think this needs another change so we don't see suddenly a lot of Counters no have a label anymore upon upgrade.

ralphilius: Makes sense to me. Please check the change I made.

jezdez: @ralphilius LGTM

kravets-levko:
@jezdez Indeed this code will work perfectly, but the only question - do we really need this option? IMHO it does not give us a lot of value because we already can set counter label by changing viz name, but now we will have two ways to do it, and in addition will have some code that we'll need to maintain. Once again - it's just my opinion. @arikfr - WDYT?

ralphilius:
@kravets-levko There are situations you need to have different counter label because the visualization name is long such as "Total Orders in Last 30 days" or "Total Items Sold by All Brands". These names are descriptive enough to see on shared/public boards. Without this option, it would be long and the card is repetitive with long names.

Example:
image

Issue #2875, opened 2018-09-28
Charts with long legend break when switching to horizontal legend


arikfr:
See gif:

![](https://cl.ly/77b748b27f71/Screen%252520Recording%2525202018-09-28%252520at%25252006.03%252520PM.gif)

Here's an example query to use to reproduce this:

```sql
select created_at::Date, action, count(0)
from events
group by 1,2
```

kravets-levko:
@arikfr Thanks, will check this ASAP

PR #2902, opened 2018-10-05
Charts with long legend break in horizontal mode

kravets-levko:
Issue: getredash/redash#2875

1. Update plotly.js to 1.41.3 which has fixed some bugs related to the legends (not all, sadly :-( ) and has feature we were waiting for: automatically compute margins to fully show axis labels (so our "hacky" code now removed).
2. Fix chart legend in horizontal mode (Plotly issue: plotly/plotly.js#1199) - "hacky" solution, hope we'll be able to remove it soon.
3. Some cleanup.

Screenshots:
image
image
image
image


package.json
```diff
@@ -67,7 +67,7 @@
     "numeral": "^2.0.6",
     "pace-progress": "git+https://github.com/getredash/pace.git",
     "pivottable": "^2.15.0",
-    "plotly.js": "1.30.1",
+    "plotly.js": "^1.41.3",
```
arikfr: Let's have a specific version here (`"1.41.3"`) as in the past there were cases of regressions in their releases, so we don't want random versions getting picked up in our builds.

PR #2903, opened 2018-10-05
add PagerDuty as an Alert Destination

alexanderlz:
Added PagerDuty as an Alert destination.
Will add Test method as part of separate PR


setup/setup.sh
```diff
@@ -61,11 +61,14 @@ setup_compose() {
     echo "export COMPOSE_PROJECT_NAME=redash" >> ~/.profile
     echo "export COMPOSE_FILE=/opt/redash/docker-compose.yml" >> ~/.profile
     source ~/.profile
-    docker-compose run --rm server create_db
-    docker-compose up -d
+    newgrp docker << END
```
arikfr: This is not needed anymore -- see the changes I did in #2908 .

redash/destinations/pagerduty.py
```diff
@@ -0,0 +1,66 @@
...
+                'summary': default_desc,
+                'severity': 'error',
+                'source': 'redash',
+            }
+        }
+
+        if new_state == "triggered":
+            data['event_action'] = 'trigger'
+        else:
+            data['event_action'] = 'resolve'
```
arikfr: Does PD support Unknown state?

redash/destinations/pagerduty.py
```diff
@@ -0,0 +1,69 @@
+import pypd
```
arikfr: This should be in a try/catch block and should set a boolean value which used in the `enabled` method of the destination like we do for data sources.

redash/destinations/pagerduty.py
```diff
@@ -0,0 +1,69 @@
...
+import logging
+
+from redash.destinations import *
+
+
+class Pagerduty(BaseDestination):
```
arikfr: It's spelled with a capital `D`, so the class name should be `PagerDuty`.

redash/destinations/pagerduty.py
```diff
@@ -0,0 +1,69 @@
...
+
+    KEY_STRING = '{user_name}_{user_email}_{query_id}_{query_name}'
+    DESCRIPTION_STR = 'redash_{query_id}_{query_name}'
```
arikfr: It's a string for humans not 🤖. I think `Redash Query #{query_id}: {query_name}` is a much friendlier default :)

redash/destinations/pagerduty.py
```diff
@@ -0,0 +1,69 @@
+import pypd
+import logging
+
+from redash.destinations import *
+
+
+class Pagerduty(BaseDestination):
+
+    KEY_STRING = '{user_name}_{user_email}_{query_id}_{query_name}'
```
arikfr: Wouldn't the alert id be a bette deduplication key?

redash/destinations/pagerduty.py
```diff
@@ -0,0 +1,69 @@
...
+    DESCRIPTION_STR = 'redash_{query_id}_{query_name}'
+
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            'type': 'object',
+            'properties': {
+                'integration_key': {
+                    'type': 'string',
+                    'title': 'Pagerduty Service Integration Key'
```
arikfr: PagerDuty

redash/destinations/pagerduty.py
```diff
@@ -0,0 +1,69 @@
...
+            "required": ["integration_key"]
+        }
+
+    @classmethod
+    def icon(cls):
+        return 'creative-commons-pd-alt'
+
+    def notify(self, alert, query, user, new_state, app, host, options):
+
+        default_desc = self.DESCRIPTION_STR.format(query_id=query.id, query_name=query.lowercase_name)
```
arikfr: Following the comment about the description text, I don't think we need to lower case the query name.

arikfr:
👍

PR #2905, opened 2018-10-07
Update docker-compose.yml

Rovel:
Hi,
just updating docker-compose dev to version 3, any special reason to keep redis on 3? and pg on 9.5? I could also add a volume to pg, any reason not to do so?
Since is such a basic PR (maintenance) I assumed there is no need for a thread, but I can open one if there is need

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 # This configuration file is for **development** setup. For production, refer to
 # docker-compose.production.yml.
-version: '2'
+version: '3'
 services:
   server:
     build: .
@@ -20,7 +20,7 @@ services:
   worker:
     build: .
     command: scheduler
-    volumes_from:
+    volumes:
       - server
     depends_on:
       - server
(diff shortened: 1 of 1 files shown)


Rovel:
thanks for the clarification @arikfr I just modified the docker compose, see if its ok....

off topic question: 
I would be great for me to add a postgres datasource with a schema param in it, but I couldn't find where the file that defines the form to add the pg datasource is located, so any chances that a PR of this kind is approvable (schema as a optional field)? and where you define the form fields for datasources & parsing that form to a db connection?

Thanks

arikfr:
@Rovel: note that I applied a small change to what you did -- I defined the code volume in the `worker` service directly. Your version was just creating a new volume for the `worker` service.

Rovel:
great! 👍 
I suggest to use 2 separates compose files, 1 for dev and 1 for deploys.
The dev you map volumes like those `".:/app"` and other like db and worker volumes in deploy usually I don't do this since the context is copied via dockerfile w/ the proper ignores I do 2 dockerfiles too 1 for dev an 1 for deploy  since needs are different specially around volumes.

arikfr:
We have two separate files already :)

Rovel:
oh shoots .  :)

PR #2906, opened 2018-10-07
Preventing open redirection

### LINKED ISSUE #2822: Open redirect due to the specific handling of the `state` parameter on the any redash instance
Hello.
I noticed, that redash instance is potentially vulnerable for Open Redirects and CSRF attacks due to specific handling of the `state` parameter on the `/oauth/google_callback` endpoint.
According to the https://tools.ietf.org/html/rfc6749#section-4.1.1 in OAuth2 `state` parameter should be used for preventing cross-site request forgery attacks (it must be secure random string), not the redirection (maybe additional parameter should be implemented instead, like `next`, allowing only on-site redirects)

Proof-of-concept: visit some test redash instance: https://redash.example.com/login?next=//google.com and authenticate through google. The  value of the `next` parameter will be taken into the `state` after google auth on the `/oauth/google_callback`, and you will end up on the `google.com` site instead `redash.example.com`.

I suggest to implement some kind of filtering or whitelisting to prevent such unvalidated redirects.

kyoshidajp:
Fix #2822

The issue is an Open Redirect Vulnerability.

## Changes

Change redirecting URL after logging in. Remove scheme and [netloc](https://docs.python.org/2.7/library/urlparse.html#module-urlparse)(domain and port) from next param value.

The following table shows examples of the login path and the redirect to path.

| login path | redirect to |
| ---- | ---------- |
| `/login` | `/{org.slug}` |
| `/login?next=queries` | `/queries` |
| `/login?next=https://redash.io` | `/` |
| `/login?next=//redash.io` | `/` |
| `/login?next=//{domain}/queries` | `/queries` |

### DIFF
--- a/redash/handlers/authentication.py
+++ b/redash/handlers/authentication.py
@@ -4,7 +4,7 @@ from flask import abort, flash, redirect, render_template, request, url_for
 
 from flask_login import current_user, login_required, login_user, logout_user
 from redash import __version__, limiter, models, settings
-from redash.authentication import current_org, get_login_url
+from redash.authentication import current_org, get_login_url, get_next_path
 from redash.authentication.account import (BadSignature, SignatureExpired,
                                            send_password_reset_email,
                                            validate_token)
@@ -12,7 +12,6 @@ from redash.handlers import routes
 from redash.handlers.base import json_response, org_scoped_rule
 from redash.version_check import get_latest_version
 from sqlalchemy.orm.exc import NoResultFound
-from urlparse import urlsplit, urlunsplit
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


redash/handlers/authentication.py
```diff
@@ -106,8 +120,7 @@ def login(org_slug=None):
     elif current_org == None:
         return redirect('/')
 
-    index_url = url_for("redash.index", org_slug=org_slug)
-    next_path = request.args.get('next', index_url)
+    next_path = get_next_path(org_slug)
```
arikfr: We use the same pattern of next_path in other places (`redash.google_oauth` and `redash.ldap_auth`) -- mind adding there was well?

kyoshidajp: @arikfr I have checked google oauth login with invalid `next` parameter works correctly. Not only by browsing but also by debugging with pdb.

When called `/oauth/google` in
https://github.com/getredash/redash/blob/v5.0.1/redash/authentication/google_oauth.py#L69
`next` parameter is already sanitized.

By via this code:
https://github.com/getredash/redash/blob/v5.0.1/redash/handlers/authentication.py#L131

I suppose LDAP login is the same, but I haven't checked it because I don't have the environment. Sorry.

arikfr: You mean it's sanitized because we're redirected there from `/login`? But an attacker can just send you to `/oauth/google` directly with an evil `next` URL.

I don't have a setup for LDAP either :) But I think it's ok to blindly apply your `get_next_path` function there instead of taking the value as is.

kyoshidajp: OK. I'll fix it.

arikfr:
Thanks!

PR #2910, opened 2018-10-08
Fix webpack build error about BigMessage.

jezdez:
This was also in https://github.com/getredash/redash/pull/2902 by @kravets-levko FTR


kravets-levko:
Thanks @jezdez! I'll remove it from my PR to keep the history clean.

PR #2912, opened 2018-10-08
WIP: Bail when webpack build errors.

### LINKED ISSUE #2895: webpack production build should fail on warning and errors
webpack's production build should fail (and stop image building) on warnings and errors to prevent issues like the one fixed in #2894.

jezdez:
Fixes #2895.

### DIFF
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -3,8 +3,10 @@ import debug from 'debug';
 import Mustache from 'mustache';
 import {
   each, zipObject, isEmpty, map, filter, includes, union, uniq, has,
-  isNull, isUndefined, isArray, isObject,
+  isNull, isUndefined, isArray, isObject, identity,
 } from 'lodash';
+
+Mustache.escape = identity; // do not html-escape values
 
 const logger = debug('redash:services:query');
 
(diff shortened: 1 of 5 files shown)


client/app/components/QueryEditor.jsx
```diff
@@ -208,7 +208,7 @@ class QueryEditor extends React.Component {
                 {this.props.dataSources.map(ds => <option label={ds.name} value={ds.id} key={`ds-option-${ds.id}`}>{ds.name}</option>)}
               </select>
               {this.props.canEdit ?
-                <Tooltip placement="top" title={modKey + ' + S'}>
+                <Tooltip placement="top" title={modKey + ' + S'>}>
```
kravets-levko: Seems it's invalid jsx (or am I wrong?)

arikfr: It's probably here to test that the build actually fails :)

kravets-levko: Probably :-) But I think we shouldn't keep it :-)

jezdez: Ooops, yeah, so this was hard to get right and I wanted to have a real-life example to test this with. Will remove once I have this working as expected.

PR #2915, opened 2018-10-09
Update docker-compose.production.yml

Rovel:
just update docker-compose version
I made myself a docker compose for my prod env that uses a RDS db so there is no need no manage a postgres container in prod.
r u interested in another pr with this alternative prod env with instructions?

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -4,7 +4,7 @@
 # Some other recommendations:
 # 1. To persist Postgres data, assign it a volume host location.
 # 2. Split the worker service to adhoc workers and scheduled queries workers.
-version: '2'
+version: '3'
 services:
   server:
     image: redash/redash:latest


docker-compose.production.yml
```diff
@@ -4,7 +4,7 @@
 # Some other recommendations:
 # 1. To persist Postgres data, assign it a volume host location.
 # 2. Split the worker service to adhoc workers and scheduled queries workers.
-version: '2'
+version: '3'
```
arikfr: If there is no change in functionality I wonder if maybe it's worth to keep it as "version: '2'" to have wider range of supported Docker Compose versions?

Rovel: in your file there is no api change at all, v3 add more stuff to manage deploy and networking.
you can check it here:
https://docs.docker.com/compose/compose-file/compose-versioning/#upgrading

changing it adhoc after a clone back and forth is no brainer in this case, but housekeeping in my opinion is always good specially considering compose is in v3.7 already.

Aaaand considering this is a prod env, having the latest patch in the infra in your first deploy is a good practice, deploying a new prod server with outdated stuff must be avoided even knowning its a reality. 💅 

but feel free reject this pr since is a easy change back and forth.  😃

arikfr:
Closing this because of #3251.

arikfr:
Thank you for your effort though!

Rovel:
U'r welcome, great project by the way!
Do you think that would be nice in the prod docker-compose persist the [env var settings](https://redash.io/help/open-source/admin-guide/env-vars-settings)? 
For example, if my container have some problem and I need to restart them I just lost a lot of stuff that I configured there, so I'm thinking of 2 options:
1. Persist a volume (Maybe bad, mostly because they'r env vars)
2. Have all those env vars in the docker-compose.prod (make sense a pr for this? adding a .env and adding it on the docker-compose, but first split secrets from configs)

Do you have another good suggestion on that?

arikfr:
Environment variables settings are passed using the env_file configured in https://github.com/getredash/redash/blob/master/setup/docker-compose.yml.

Rovel:
does make sense to split it in 2 files like 
env_file_params
env_file_secrets
?

arikfr:
I don't see a reason for this. But you're welcome to do in your setup whatever makes the most sense to you. Those are just example configuration.

PR #2916, opened 2018-10-09
Fix export query results output file name

gabrieldutra:
Hello again!

Well, this is actually the same issue as in my [PR#2610](https://github.com/getredash/redash/pull/2610), only that I missed one file :sweat_smile: 

Now I searched and the only place that has `replace(' ', '_')` is a `.py` file and in python it works as we want it to.

The problem can be seen in the following image:
underscore


arikfr:
Thanks!

kocsmy:
Love this little improvement, thanks @gabrieldutra !

PR #2917, opened 2018-10-09
Fix output file name not changing after rename query

gabrieldutra:
Hi,
This PR solves the following:
query-name

Note that changing the query name doesn't affect the filename when you download query results (unless you reload the page or execute the query again).

After the fix:
query-name-after


arikfr:
Thanks!

PR #2922, opened 2018-10-10
Query page layout improvements for small screens

kravets-levko:
Issue getredash/redash#2796 (items related to query page).

1. Change layout with fixed blocks to scrollable; some blocks (header, footer) kept fixed; editor, query metadata and results are scrollable (see screenshots).
2. Query metadata block now is a bit smaller and compact (to save some space), but still well-readable.

Screenshots:

image
image
image
image
image

### DIFF
--- a/client/app/assets/less/redash/ant.less
+++ b/client/app/assets/less/redash/ant.less
@@ -12,3 +12,20 @@
 @border-color-base      : #e8e8e8;
 
 @primary-color          : @blue;
+
+// Fix for disabled button styles inside Tooltip component.
+// Tooltip wraps disabled buttons with `<span>` and moves all styles
+// and classes to that `<span>`. This resets all button styles and
+// turns it into simple inline element (because now it's wrapper is a button)
+.btn {
+  button[disabled] {
+    -moz-appearance: none !important;
+    -webkit-appearance: none !important;
+    appearance: none !important;
+    border: 0 !important;
+    outline: none !important;
+    background: transparent !important;
+    margin: 0 !important;
+    padding: 0 !important;
+  }
+}
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -541,44 +541,25 @@ nav .rg-bottom {
     .schema-container {
       display: none;
     }
-  }
 
-  .query-page-wrapper {
-    .container {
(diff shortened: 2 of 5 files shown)


arikfr:
How does it look like in the following cases:

- Longer title? Try: "Latest Queries Executed By a User", "Where People Click on Get Started"
- Long title with 2-3 tags?
- Longer user name (like: "Arik Fraimovich" 😬)

Also:
- We should hide the "+ New Visualization" button, I doubt anyone will use it on mobile.
- Same goes with "Publish" button.
- There needs to be padding between the button and the data source selector.

arikfr:
Btw, how does the data only view looks like after these changes?

kravets-levko:
There are two screenshots with data-only  views (in portrait and landscape orientations)

kravets-levko:
1. Long title:
image

2. Tags are hidden on small screens (it was already implemented in this way, I didn't change anything).

3. (Very) long user name:
image


Other fixes will do ASAP and ping you. Also noticed that header looks not well in non-mobile layout:
image
I'm not sure - is it intentional or a bug? If the second -  I think I'll fix it too.

arikfr:
> Tags are hidden on small screens (it was already implemented in this way, I didn't change anything).

@kocsmy why?

> I'm not sure - is it intentional or a bug? If the second - I think I'll fix it too.

Worth fixing. Maybe always show the tags in the row below the title? 

Also, once we remove the publish button, we can give more space to the title on mobile.

kocsmy:
> Tags are hidden on small screens (it was already implemented in this way, I didn't change anything). Why?

I'm not sure if this is intentional. We can show them.

arikfr:
> I'm not sure if this is intentional. We can show them.

We should show them :)

kravets-levko:
1. Hide "New visualization" and "Publish" buttons in mobile mode.

2. Fix paddings: between the button and the data source selector (in mobile mode), between icon ant text in buttons (in desktop mode):
image
image


3. Fix "Execute" button appearance when running query:
image

4. Show tags in mobile mode and fix header layout: now tags are placed under the title on all screens:
image
image
image
image
image
image

kocsmy:
LGTM2

arikfr:
🙌

Issue #2603, opened 2018-06-19
The "dnspython" module must be installed to use mongodb+srv:// URIs


leorochael:
### Issue Summary

When attempting to test a connection to a MongoDB data source using a `mongodb+srv://` URL, the following message is displayed:

    The "dnspython" module must be installed to use mongodb+srv:// URIs

This message probably comes straight from `pymongo`, and I'm guessing that simply adding `dnspython` to `requirements_all_ds.txt` will be enough.

Or even better, inside `requirements_all_ds.txt`, simply replace this line:

    pymongo==3.6.1

With this line:

    pymongo[tls,srv]==3.6.1

The complete set of [pymongo extras, as of 3.6.1](https://github.com/mongodb/mongo-python-driver/blob/a9e219291915710e62aa6f3e1caea17f87b8c5c2/setup.py#L335), would be something like:

    pymongo[tls,srv,gssapi]==3.6.1

As of [pymongo 3.7.0b1](https://github.com/mongodb/mongo-python-driver/blob/3d8155d87d4ff6b95eba313e3803af6abfdd45c3/setup.py#L341), it will be

    pymongo[tls,srv,gssapi,snappy]>=3.7.0b1

### Steps to Reproduce

1. Create a new MongoDB Datasource
2. Type in a `mongodb+srv://` url, like the ones shown in the free tier of MongoDB Atlas
3. See the error message: 

    The "dnspython" module must be installed to use mongodb+srv:// URIs

### Technical details:

* Redash Version: 4.0.0+b3948
* Browser/OS: Firefox/Linux
* How did you install Redash: docker from getredash

RichardLitt:
Thanks @leorochael. This sounds like it might be a setup issue; have you asked about this bug on the forum?

arikfr:
@leorochael thank you for the detailed report. If there are no complications due to installing the extra dependencies, I don't mind updating requirements_all_ds.txt to include the pymongo extras. Did you try this setup already?

leorochael:
@arikfr I have run the pip command below (found on the Docker build) on my Ubuntu 17.10 machine, and found the same warnings with or without the extras:

```
dql 0.5.24 has requirement pyparsing==2.1.4, but you'll have pyparsing 2.2.0 which is incompatible.
pymapd 0.3.2 has requirement thrift==0.10.0, but you'll have thrift 0.11.0 which is incompatible.
[...]

PR #2928, opened 2018-10-10
Speed up builds by skipping installing requirements_all_ds.txt in CI unit tests

arikfr:


requirements_dev.txt
```diff
@@ -4,5 +4,8 @@ pytest-watch==4.1.0
 coverage==4.0.3
 mock==2.0.0
 
-# PyMongo is needed for one of the unit tests:
+# PyMongo and Athena dependencies are needed for some of the unit tests:
+# (this is not perfect and we should resolve this in a different way)
```
arikfr: @jezdez I guess this is another reason to split the query runners into their own packages.

arikfr: Why #2603 is another case?

arikfr: Anyway, I think it's worth merging this regardless. In parallel we need to layout a plan to streamline our release process and part of that separate the query runners into their own packages.

jezdez: I mean, #2603 shows that some of the query runners have exceedingly complicated dependencies on their own for which separation from the Redash core is useful. My underlying hypothesis is of course that a lean Redash core is easier to maintain than a "full" Redash core.

arikfr: I see it the other way: this shows that we should make sure to always test with all the dependencies, as some query runners might introduce conflicting packages or just require a different version than we would otherwise use.

_But_ we can always run the full test suite with all the dependencies before releasing.

jezdez: I'm not arguing against running the Redash tests will all the query runners, but for separating the code of the query runners in separate packages. They would still be installed via the requirements_all_ds.txt (before release) but simply install redash-cassandra, redash-mongo etc instead.

PR #2931, opened 2018-10-11
Google Spreadsheets: support for open by url

### LINKED ISSUE #2924: Google Spreadsheets: support for open by url
We currently use [gspread](https://github.com/burnash/gspread)'s [`open_by_key`](https://gspread.readthedocs.io/en/latest/#gspread.Client.open_by_key) method to load spreadsheets. But it also has the [`open_by_url`](https://gspread.readthedocs.io/en/latest/#gspread.Client.open_by_url) method.

Using the later we can add support for just providing the spreadsheet URL, which will be much more friendly to end users. 

To remain backward compatible we can check if the query is of format `http...` -> use the open by url method, otherwise use the open by key one.

alexanderlz:
Closes #2924.

### DIFF
--- a/redash/query_runner/google_spreadsheets.py
+++ b/redash/query_runner/google_spreadsheets.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from base64 import b64decode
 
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks!

PR #2933, opened 2018-10-12
Chnage: switch to Webpack 4

### LINKED ISSUE #2930: Upgrade from webpack 3 to webpack 4
The title pretty much tells it all.

dmonego:
Hi! There are still a few warnings this is throwing, but I'd appreciate if project maintainers had any input on this changeover! 

Overall it seems like the default behavior of webpack 4 is pretty close to the custom chunks you were maintaining, so a fair amount of code gets saved in webpack.config.

Closes #2930.


arikfr:
Amazing, thanks @dmonego ! 

It seems that the warnings about the size of bundle we create, but not about actual issues. Anyway, I will compare the generated files with what we currently have and then see if any changes are needed.

client/app/assets/less/main.less
```diff
@@ -11,7 +11,6 @@
 @import '~ui-select/dist/select.css';
 @import '~angular-toastr/src/toastr';
 @import '~angular-resizable/src/angular-resizable.css';
-@import '~pace-progress/themes/blue/pace-theme-minimal.css';
```
arikfr: Why?

dmonego: As far as I can tell, it's a dead link.

arikfr: It's not and it seems to be working for me with `MiniCssExtractPlugin`. I pushed a revert for this along with a few other fixes (see my commits in this PR). Once I see the build works -- will merge. 👍

arikfr:
Merged. Thanks!

PR #2934, opened 2018-10-12
Show dashboard creator on dashboard page

kocsmy:
46878258-18436200-ce43-11e8-99fb-2f500a496b1a

- I think this opens up some space for other details for example last edited time.
- I tried it without a background but I kind alike having this slight grey tone / this might be not the best choice though as it looks very similar to our secondary button design. Might need to make it less grey.


arikfr:
1. Can you show an example without the background?
2. How about aligning to the left?

arikfr:
Something I just thought about -- it will look really awkward on an empty dashboard, no?

kocsmy:
Hmm, it will indeed look awkward on an empty dashboard :) 

Another idea: maybe put it on the top right header (left to the refresh schedule) and only show it on header hover (just like add tags)?

kocsmy:
I'm not convinced this is better. It is not such an important information to show it there.

image

kocsmy:
How about adding an extra item here that opens a modal with extra details:
image

kocsmy:
Regarding empty dashboards / adding below the widgets: we could attach it to the bottom, similarly to our footer. That way it would be okay there. I think I like that idea the most.

arikfr:
> How about adding an extra item here that opens a modal with extra details:

Feels like too buried. 

> Regarding empty dashboards / adding below the widgets: we could attach it to the bottom, similarly to our footer. That way it would be okay there. I think I like that idea the most.

So far it's the best idea.

But how about showing the profile picture next to the dashboard title? Hovering on it, will show a text saying "Last edit 2 days ago by Zsolt Kocsmarszky (created 2 years ago)"?

kocsmy:
Here's how it'll look.

image

First, I though it'll be strange that we'll have the favorite and this sandwiching the title. But this could actually work fine.

I'm fine with both this and above the footer.

kocsmy:
Here's a version with tags next to it:
image

kocsmy:
I was so confident that I made the change in the markup and css and moved it next to the title.

If you are happy, it just needs implementation: actual user image, tooltip.

jezdez:
@kocsmy @arikfr Is there anything I can help move this along?

arikfr:
@jezdez it was missing the binding to the actual data. Fixed and merged.

jezdez:
@arikfr Thank you 😻

PR #2936, opened 2018-10-13
Add Kylin plugin for SQL query

Trigl:
Kylin support API to query data. This is a Kylin query runner to let redash connect to Kylin.

screen shot 2018-10-13 at 6 05 15 pm


redash/query_runner/kylin.py
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+    def run_query(self, query, user):
+        url = self.configuration['url']
+        kylinuser = self.configuration['user']
+        kylinpass = self.configuration['password']
+        kylinproject = self.configuration['project']
+
+        resp = requests.post(
+            os.path.join(url, "api/query"),
```
arikfr: I'm assuming this is to handle the cases where the URL might end with `/` or not?

Interesting use of `os.path.join` :)

redash/query_runner/kylin.py
```diff
@@ -0,0 +1,145 @@
...
+            auth=HTTPBasicAuth(kylinuser, kylinpass),
+            json={
+                "sql": query,
+                "offset": 0,
+                "limit": 50000,
```
arikfr: How about we add this to the configuration object with this value as the default?

redash/query_runner/kylin.py
Trigl: Sorry for late response, what you said is exactly true.

redash/query_runner/kylin.py
Trigl: Yes that's better, do I need change something?

arikfr: Yes, instead of using a static value, read one from configuration (like you do for the host/user/password/etc).

Trigl: Got it, thank you for your advice.

PR #2938, opened 2018-10-13
Auto focus tag input

kyoshidajp:
Set focus on a tag input box when opened the modal.

Before:
autofocus_tag_before

After:
autofocus_tag_after


arikfr:
Thanks, @kyoshidajp ! :)

PR #2947, opened 2018-10-14
Add SAML scheme override env var

leosunmo:
This setting will allow you to override the SAML Auth URL scheme that gets constructed by Flask. This is a useful feature if, for example, you're behind a Proxy Protocol enabled TCP load balancer (AWS ELB that terminates SSL) and your Nginx proxy or similar adds a X-Forwarded-Proto of HTTP even though your Redash URL for SAML auth is HTTPS. Ideally your proxy should pass along the actual forwarded protocol but it's impossible for it to know unless it analyses the Proxy Protocol and uses that to figure out the original scheme, which I'm not sure is possible.

I'm not quite sure where I should be adding documentation regarding this, but please point me in the right direction and I'll write something up.


leosunmo:
I'm unsure if I caused that unit test failure but it doesn't look like it. I can't really see what I would need to change to make it pass.

RyPeck:
@leosunmo Are you using the SSL settings found at https://redash.io/help/open-source/admin-guide/https-ssl-setup or are you using the SSL settings at the reference Git Stash https://gist.github.com/arikfr/64c9ff8d2f2b703d4e44fe9e45a7730e ?

leosunmo:
Hey @RyPeck , thanks for showing interest in the PR! I'm afraid neither of these settings will help me here. I could have missed something since it's been a few months since I last looked at this but the first link lands me in the same situation since the `X-Forwarded-Proto` is set to `http` (nginx doesn't know that it was once HTTPS, that ends at the AWS ELB).

The second link does help either as it refers to terminating SSL in Nginx itself, using Let's Encrypt. That happens at the AWS ELB level in my case.

The only way so far is to force Flask to think that it's an HTTPS connection so it constructs the correct URL.

leosunmo:
@arikfr Anything I need to do here? The CI fails seems random and in this case seems to be a visual change that doesn't involve my change at all.

arikfr:
👍

arikfr:
@leosunmo I apologize, but I had to revert this change because it breaks the SAML functionality. I think it's because you didn't import `settings`, but can't be sure.

If you still need this, please re-open with a new PR and please test it.

Thanks.

leosunmo:
@arikfr huh, strange! I'm running this in production (based on version 4.0.2) and it's the only thing that makes auth work at all! I will rebase with latest version, import settings and resubmit. There could have been changes that breaks it between version 4.0.2 and current master.

leosunmo:
@arikfr https://github.com/getredash/redash/pull/3614 fixed in this PR.

PR #2949, opened 2018-10-15
Port toggle to hide tables configurable substring (e.g. versioned tables)

alison985:
This ports Mozilla [Issue 31](https://github.com/mozilla/redash/issues/31) and [Issue 130](https://github.com/mozilla/redash/issues/130) as requested in Mozilla [Issue 584](https://github.com/mozilla/redash/issues/584).

Here's what the toggle button looks like in the schema browser (if one is specified in the data source config).
screenshot 2018-10-14 20 04 50
And toggled on:
screenshot 2018-10-14 20 04 56
And the taggle string input on a datasource config page:
screenshot 2018-10-14 20 05 02


kocsmy:
Thanks for the PR, @alison985 if you don't mind, I want to add some comments:
- I know this is the result of bad switch icon design but we have 2 exactly same looking button with very different functionality (single button vs toggle button).
- The icon itself does not imply what is this for. Is this really the best place to put this icon? Can we maybe put it somewhere else? Maybe add this as an option for the data source? 
- (There's a missing 5px margin between the 2 icons)

alison985:
@kocsmy The datasource provides a 'table_toggle_string' parameter. If that parameter is empty the toggle button doesn't show up in the schema browser UI. It only shows up if there is table_toggle_string provided. In terms of putting the button somewhere else, I could move the schema refresh button up to be in the same line as the data source drop-down and keep the new toggle button on the line with the search schema text field. This would make more sense as the refresh is connected to the data source and the toggle button is essentially a search filter.

I'll address the 5px margin between the 2 icons in the same commit I make the button move change after we've finalized a plan for the move.

rafrombrc:
We've decided to retract this PR and implement this as an extension once the upcoming changes to the schema browser implementation have landed.

PR #2953, opened 2018-10-15
Clickhouse fixes

denisov-vlad:
- add new line before FORMAT instead of whitespace (allows to comment last line of query)
- request timeout set to 5 seconds.


arikfr:
Thanks!

PR #2956, opened 2018-10-15
clickhouse: move timeout to params

denisov-vlad:
Previous commit with 5s timeout is not suitable for huge requests.

### DIFF
--- a/redash/query_runner/clickhouse.py
+++ b/redash/query_runner/clickhouse.py
@@ -71,7 +71,7 @@ class ClickHouse(BaseSQLQueryRunner):
             self.configuration['url'],
             data=data.encode("utf-8"),
             stream=stream,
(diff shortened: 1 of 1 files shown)


redash/query_runner/clickhouse.py
```diff
@@ -66,7 +71,7 @@ def _send_query(self, data, stream=False):
             self.configuration['url'],
             data=data.encode("utf-8"),
             stream=stream,
-            timeout=5,
+            timeout=self.configuration['timeout'],
```
arikfr: This should be `self.configuration.get('timeout', 30)` as otherwise it will break if no value is set.

denisov-vlad: Oops. Fixed

arikfr:
Thanks!

PR #2961, opened 2018-10-15
MongoDB: skip system collections when loading schema.

arikfr:


mwvaughn:
I have opened PR #3004 which fixes this, as the earlier commit merged by @arikfr did not resolve the error.

PR #2964, opened 2018-10-15
Fix no tags shown when having empty set

gabrieldutra:
I noticed that no tags were being shown (you can see in the image below)
captura de tela de 2018-10-15 16-13-19

This started happening after PR #2833 because `tag-list.html` now calls `toggleTag()` with `tag.name` instead of `tag`. I changed `NoTaggedObjectsFound.jsx` to work as it receives the Tag Set now.

## After the changes
captura de tela de 2018-10-15 16-14-17


arikfr:
Good catch, thanks!

Issue #2966, opened 2018-10-16
Add "Groups" users are belong to into users list


GitSumito:
### Issue Summary

Current users list

```
/usr/bin/docker exec -i redash_server_1 ./manage.py users list
```

* Current
```
---
Name: Fred Foobar
Email: foobar@example.com
Organization: Default
Active: True
````

there are no groups on the list.
It should be added on the same list.

* Shoule be
```
---
Name: Fred Foobar
Email: foobar@example.com
Organization: Default
Active: True
Group: admin, default
````

### Technical details:

* Redash Version:  5.0.1+b4850
* Browser/OS: Chrome
* How did you install Redash: Docker

GitSumito:
done!

PR #2967, opened 2018-10-16
Add "Groups" users are belong to into users list

GitSumito:
#2966

### DIFF
--- a/redash/cli/users.py
+++ b/redash/cli/users.py
@@ -247,3 +247,10 @@ def list(organization=None):
 
         print("Id: {}\nName: {}\nEmail: {}\nOrganization: {}\nActive: {}".format(
             user.id, user.name.encode('utf-8'), user.email, user.org.name, not(user.is_disabled)))
+
+        for group_id in user.group_ids:
+            try:
+                group = models.Group.query.get(group_id)
+                print("Group: {}".format(group.name))
+            except AttributeError:
+                pass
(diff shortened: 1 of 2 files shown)


GitSumito:
It seems that the function which I modified are too complex..
It should be fixed.

Does someone have a good idea??

redash/cli/users.py
```diff
@@ -247,3 +247,10 @@ def list(organization=None):
 
         print("Id: {}\nName: {}\nEmail: {}\nOrganization: {}\nActive: {}".format(
             user.id, user.name.encode('utf-8'), user.email, user.org.name, not(user.is_disabled)))
+
+        for group_id in user.group_ids:
+            try:
+                group = models.Group.query.get(group_id)
+                print("Group: {}".format(group.name))
+            except AttributeError:
```
kravets-levko:
```python
groups = models.Group.query.filter(models.Group.id.in_(user.group_ids)).all()
group_ids = [group.id for group in groups]
print("Groups: {}".format(", ".join(group_ids)))
```

GitSumito: But did not work well.

GitSumito: I'd like to display group name, not group id.

kravets-levko:
```python
group_names = [group.name for group in groups]
print("Groups: {}".format(", ".join(group_names)))
```
:-)

PR #2970, opened 2018-10-16
Change: encrypt data source options. 🔓

arikfr:
Something we're doing in the hosted version, but time to open source: encrypt data source options.

Only WIP thing is a proper migration file. Need to see how to use the EncryptedType without a model file.


redash/settings/__init__.py
```diff
@@ -107,6 +107,8 @@ def all_settings():
 COOKIE_SECRET = os.environ.get("REDASH_COOKIE_SECRET", "c292a0a3aa32397cdb050e233733900f")
 SESSION_COOKIE_SECURE = parse_boolean(os.environ.get("REDASH_SESSION_COOKIE_SECURE") or str(ENFORCE_HTTPS))
 
+SECRET_KEY = os.environ.get('REDASH_SECRET_KEY', COOKIE_SECRET)
```
jezdez: I would take the opportunity to make sure that this is a non empty value since this is a [critical setting in Flask](http://flask.pocoo.org/docs/1.0/config/#SECRET_KEY).

jezdez:
@arikfr Do you need help with this?

arikfr:
@jezdez I need to figure how to write the migration without using the `DataSource` model code (so the migration doesn't break in future releases). I think we can reuse some code from `sqlalchemy-utils`, but didn't have a chance to research this proper.

If you have any suggestions, I'm happy to apply this. Otherwise might push this feature to v7.

jezdez:
@arikfr Understood, I think we should move this to v7 since we have plenty of other big changes in v6 and I'd rather have this not done under time pressure.

arikfr:
Migration implemented and this is ready to go.

redash/settings/__init__.py
arikfr: I realized that this is 1) beyond the scope of the PR; 2) I don't have the time to implement it :-(. To avoid having this PR lingering any longer, went ahead without it.

jezdez: Makes sense.

denisov-vlad:
I think you've forgotten to update `SQLAlchemy-Utils` version here.

We had `ImportError: No module named encrypted_type` error. After update to `SQLAlchemy-Utils-0.33.11` the error has gone.

I'm not sure which version is OK to make a pull request.

arikfr:
@denisov-vlad I just noticed we have SQLAlchemy-Utils's version defined as `>=0.29.0` and indeed the Docker image has version `0.33.11`.

So no need to upgrade in this PR, but we should pin its version.

Issue #1443, opened 2016-12-02
Cassandra not selectable in Datasource type drop down


muglar:
Found a bug? Please fill out the sections below... thank you 👍

### Issue Summary

Fresh install of Re:dash 0.12.0+b2449 on Ubuntu 14.04 server using bootstrap.sh.  After installation is completed, unable to select Cassandra as a datasource in the type drop down menu.

### Steps to Reproduce

1. This is the first step
2. This is the second step, etc.

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* Redash Version: Re:dash 0.12.0+b2449
* Browser/OS: Chrome/Windows10
* How did you install Redash: bootstrap.sh on fresh ubuntu 14.04 server

![image](https://cloud.githubusercontent.com/assets/17802517/20852789/c46fb948-b8b5-11e6-8663-23203dcb7723.png)

sivasankar95:
did you find a solution for this issue???? I am too facing the same issue

dioptre:
Is the Cassandra support fake? It's still not an option in the latest docker.

PR #2975, opened 2018-10-17
Added missing Cassandra support (dropdown)

dioptre:
If it's advertised that the product supports it, we should add it.

#1443


arikfr:
Actually you can add it to the default query runners list: 

https://github.com/getredash/redash/blob/34fb3ac79f11cf2055f86cb359a2ce068c3c6c96/redash/settings/__init__.py#L156

arikfr:
@dioptre do you mind moving the change into settings, so we can lend this? Thanks!

arikfr:
As this seems to be abandoned, implemented the suggested change in #4058.

PR #2985, opened 2018-10-18
Add Jest based tests to our stack

### LINKED ISSUE #2847: Add Jest setup to the project
As we move to React, we should also one up our testing on the frontend. I think a good way to bootstrap this effort is to add Jest setup to the project. I feel that having the infrastructure in place will make it easier to add tests and actually encourage developers to write tests than do manual tests for new code or when fixing/updating existing stuff. 

We can also look into adding @jondot's [jest-pytest](https://github.com/jest-community/jest-pytest) to our backend testing setup. If we don't, we should at the very least look into using [snapshottest](https://github.com/syrusakbary/snapshottest).

arikfr:
This adds the needed setup to have Jest run tests in our project along with a simple test for the Footer component.

TODO: 

- [x] Make CI run these tests
- [x] Move jest related packages to devDependencies
- [x] Install and setup eslint-jest

Closes #2847.


.circleci/config.yml
```diff
@@ -45,6 +45,15 @@ jobs:
           path: /tmp/test-results
       - store_artifacts:
           path: coverage.xml
+  frontend-unit-tests:
+    docker:
+      - image: circleci/node:8
+    steps:
+      - checkout
+      - run: sudo apt install python-pip
```
jezdez: Do we need pip here?

.circleci/config.yml
```diff
@@ -125,6 +134,7 @@ workflows:
       - flake8-tests-on-python
       - flake8-tests-on-legacy-python
       - unit-tests
+      - frontend-unit-tests
```
jezdez: I think it would make sense to have a consistent naming scheme for the jobs, e.g.

- py2-flake8-tests
- py3-flake8-tests
- backend-unit-tests
- frontend-unit-tests

WDYT?

client/app/components/Footer.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import { react2angular } from 'react2angular';
 
-function Footer({ clientConfig, currentUser }) {
+export function Footer({ clientConfig, currentUser }) {
```
jezdez: I think we could make this the default export as well, in modules that have only one domain object to simplify the imports elsewhere, right?

package.json
```diff
@@ -10,7 +10,9 @@
     "build": "rm -rf ./client/dist/ && NODE_ENV=production webpack",
     "watch": "webpack --watch --progress --colors -d",
     "analyze": "rm -rf ./client/dist/ && BUNDLE_ANALYZER=on webpack",
-    "analyze:build": "rm -rf ./client/dist/ && NODE_ENV=production BUNDLE_ANALYZER=on webpack"
+    "analyze:build": "rm -rf ./client/dist/ && NODE_ENV=production BUNDLE_ANALYZER=on webpack",
```
jezdez: I notice there's lots of repetition in calling `rm -rf ./client/dist`, could we make this a `clean` step and use pre steps to trigger them?

From https://docs.npmjs.com/misc/scripts:

> Pre and post commands with matching names will be run for those as well (e.g. premyscript, myscript, postmyscript).

.circleci/config.yml
arikfr: Yes, for `bundle`.

.circleci/config.yml
arikfr: Good idea. I will update.

client/app/components/Footer.jsx
arikfr: Until we drop the Angular support, the default export has to be `init` function. Although maybe we can create another register function for `.jsx` files, which will call `react2angular` on these components by itself. Anyway, out of scope for this PR.

package.json
arikfr: The pre/post steps are action specific, so I'm not sure how can they save us any repetition  here. 

But we can define the `rm -rf` part as a `clean` command, just to avoid typos in the future.

jezdez: Cool, makes sense!

client/app/components/Footer.jsx
jezdez: Sigh, I always forget that shim :)

.circleci/config.yml
jezdez: D'oh!

client/app/components/Footer.jsx
arikfr: Hopefully soon, we will be able to forget it for good :D

.circleci/config.yml
arikfr: Updated. I ended up having a bit more verbose names for the flake8 tests.

package.json
```diff
@@ -5,12 +5,14 @@
...
-    "dev": "REDASH_BACKEND=https://dev.redashapp.com npm start",
     "bundle": "bin/bundle-extensions",
-    "build": "rm -rf ./client/dist/ && NODE_ENV=production webpack",
+    "clean": "rm -rf ./client/dist/",
+    "build": "npm run clean && NODE_ENV=production webpack",
     "watch": "webpack --watch --progress --colors -d",
-    "analyze": "rm -rf ./client/dist/ && BUNDLE_ANALYZER=on webpack",
-    "analyze:build": "rm -rf ./client/dist/ && NODE_ENV=production BUNDLE_ANALYZER=on webpack"
+    "analyze": "npm run clean && BUNDLE_ANALYZER=on webpack",
+    "analyze:build": "npm run clean && NODE_ENV=production BUNDLE_ANALYZER=on webpack",
```
jezdez: Thank you!

PR #2991, opened 2018-10-19
Add "Users" users are belong to into groups list

GitSumito:
### Issue Summary

Current users list

```
docker exec -it redash_server_1 ./manage.py groups list
```

* Current
```
---
Id: 1
Name: admin
Type: builtin
Organization: default
````

there are no users on the list.
It should be added on the same list.

* Shoule be
```
---
Id: 1
Name: admin
Type: builtin
Organization: default
Users: sumito
````

### Technical details:

* Redash Version:  5.0.1+b4850
* Browser/OS: Chrome


arikfr:
Thanks!

PR #2992, opened 2018-10-20
Fix an invalid prop type of `newVersionAvailable` warning

kyoshidajp:
Fix the following warning.

invalid_prop_type

According to the type check, `clientConfig.newVersionAvailable` must be boolean.

https://github.com/getredash/redash/blob/master/client/app/components/Footer.jsx#L34

### DIFF
--- a/redash/handlers/authentication.py
+++ b/redash/handlers/authentication.py
@@ -167,7 +167,7 @@ def date_format_config():
 def client_config():
     if not current_user.is_api_user() and current_user.is_authenticated:
         client_config = {
-            'newVersionAvailable': get_latest_version(),
+            'newVersionAvailable': bool(get_latest_version()),
             'version': __version__
         }
     else:


redash/handlers/authentication.py
```diff
@@ -167,7 +167,7 @@ def date_format_config():
 def client_config():
     if not current_user.is_api_user() and current_user.is_authenticated:
         client_config = {
-            'newVersionAvailable': get_latest_version(),
+            'newVersionAvailable': bool(get_latest_version()),
```
kravets-levko: Hi @kyoshidajp! As farr as I remember, `get_latest_version()` just returns latest available version, so it's not correct just to cast it to boolean; instead, value returned from `get_latest_version()` should be compared with current version and if latest version is greaterr than current - it should become `True`.

kyoshidajp: Oh, my bad. I will fix it. Thanks.

kyoshidajp: @kravets-levko It seems like if `new_version_available` key's value is in Redis, new version is available.

First, `reset_new_version_status()` is called when an application started. 
https://github.com/getredash/redash/blob/v5.0.2/redash/__init__.py#L75

Next, `new_version_available` key's value is set if new version is available. Otherwise the key is deleted.
https://github.com/getredash/redash/blob/master/redash/version_check.py#L48-L51

I think that my fix isn't wrong. Or should I add an alias method like the following and call it as value of `newVersionAvailable` in `client_config()`?

```
def is_newer_version():
    return get_latest_version()
```

kravets-levko: Oh, thanks @kyoshidajp! I missed some part of this process. I think your fix is good enough :+1:

arikfr:
Thanks!

PR #2994, opened 2018-10-20
Fix url scheme

kyoshidajp:
Fix URL scheme from `http` to `https`.

Some URLs were not changed due to an invalid certificate (like https://getbem.com/naming/).


arikfr:
Thanks!

PR #2995, opened 2018-10-23
Add an error message in query_results.

senshokazu:
There is a rule which converts colon( , ) and period( . ), white space into underscore( _ ) in query results.
But the users which unknown the rule get confused when the users select columns which include colon and period, white space.
So,  I added add an error message in query_results. The error message has occurred only when `no such column` has occurred.
`no such column: XX Note: If the column names contain colon, dot, or white space, they are converted to underscores`

### DIFF
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -183,6 +183,11 @@ class Results(BaseQueryRunner):
             connection.cancel()
             error = "Query cancelled by user."
             json_data = None
+        except Exception as error:
+            connection.cancel()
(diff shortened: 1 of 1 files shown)


arikfr:
@senshokazu why did you close this one?

senshokazu:
@arikfr 
I misunderstood this pull request is against the pull req's rule.
I'm a beginner, so I don't really now behavior in oss.

arikfr:
@senshokazu it's actually ok. what made you think it's against the rules?

senshokazu:
@arikfr  Just because. There is no clear reason in that.

PR #3002, opened 2018-10-25
Support LDAP servers where one doesn't first have to bind to the LDAP  server with a username/password

Jakdaw:
With this patch, for such LDAP servers one simply doesn't configure LDAP_BIND_DN and LDAP_BIND_DN_PASSWORD.


Jakdaw:
@arikfr Could this simple change please be merged?

arikfr:
Thanks.

PR #3003, opened 2018-10-25
Fix support for calling MySQL Stored Procedures and allow queries to be cancelled

Jakdaw:
There are several situations when MySQL might return multiple result sets - for instance calling Stored Procedures where a result set might be returned, followed by an empty (zero columns, zero rows) result set from the procedure itself.
This patch has Redash choose the last result set returned that has columns; enabling the use of Stored Procedures.


Jakdaw:
Per comment in PR # 2739, I've updated this PR to rebase against mater, and this now addresses:

- Calling stored procedures
- Using multiple statements in one query
- Cancelling MySQL queries

Jakdaw:
@arikfr could this please be merged?

arikfr:
YOLO! :-)

PR #3004, opened 2018-10-26
Implement filtering of Mongo 3.4+ read-only views

mwvaughn:
The fix in #356c874 only excludes `system.views` and thus fails to resolve the reported error which happens when loading the schema of a database with collections that are actually MongoDB views. 

Such collections have no obvious namespacing, so this patch detects whether a collection is a view by examining its `options()`. I do not believe there is any better way to filter out this class of collections in the PyMongo driver.

### DIFF
--- a/redash/query_runner/mongodb.py
+++ b/redash/query_runner/mongodb.py
@@ -153,7 +153,8 @@ class MongoDB(BaseQueryRunner):
 
         self.db_name = self.configuration["dbName"]
 
-        self.is_replica_set = True if "replicaSetName" in self.configuration and self.configuration["replicaSetName"] else False
+        self.is_replica_set = True if "replicaSetName" in self.configuration and self.configuration[
+            "replicaSetName"] else False
 
     def _get_db(self):
         if self.is_replica_set:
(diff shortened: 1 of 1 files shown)


mwvaughn:
I would like to fix this in general so that we can use views in building Redash queries, but this alleviates the blocker that people are having until then.

arikfr:
Why do we need to filter views? `system.*` views were filtered because of permissions issues, but not for them being views.

mwvaughn:
I reported, as did the other user, that views were not working for redash 5 (and specifically for my organization, mongo 3.6). You committed and closed the issue but it didn’t resolve the problem. I’m happy to dig into the general solution but this solved the issue in the short term. It may not have been ideal to issue a PR but at least it got a conversation going.

On Oct 29, 2018, at 3:52 PM, Arik Fraimovich <notifications@github.com<mailto:notifications@github.com>> wrote:


Why do we need to filter views? system.* views were filtered because of permissions issues, but not for them being views.

arikfr:
I reviewed the original issue again and indeed I misunderstood it. I fixed a different issue, where trying to read system.views will prevent the schema from loading. 

If you still have issues with loading the schema after the last fix, then can you please check where it fails and what is the error message?

arikfr:
I'm closing this now. But if you still have issues, encourage you to open an issue with more details -- specially stacktrace of where it fails.

Thanks.

PR #3005, opened 2018-10-26
Support AWS IAM profile for Amazon Elasticsearch

tilfin:
To use Elasticsearch managed by AWS, Redash supports only the way to be permitted by IP addresses currently. but we want to use policy permissions by AWS IAM profile on EC2/ECS generally.

* This PR adds `Use AWS IAM profile` option to Amazon Elasticsearch settings.
* If the option is true, applying the IAM profile of the running environment to the requests.

### DIFF
--- a/redash/query_runner/amazon_elasticsearch.py
+++ b/redash/query_runner/amazon_elasticsearch.py
@@ -2,7 +2,8 @@ from .elasticsearch import ElasticSearch
 from . import register
 
 try:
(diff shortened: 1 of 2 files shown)


tilfin:
@arikfr I fixed it! I confirmed that it correctly works by either IAM profile on AWS ECS or access/secret key on local PC.

arikfr:
Thanks, @tilfin !

PR #3007, opened 2018-10-27
Add permissions to the result of "manage.py groups list" command

Udomomo:
I often change group's permissions via CLI, because web interface has only two alternatives(full-access vs view-only) and I have to make more detailed configurations according to each group's use case. 
But there isn't a way to check each group's current permissions, except by writing a query to PostgreSQL directly. I'd like to check it on CLI too.


Udomomo:
Updated test case.

arikfr:
Thanks!

PR #3008, opened 2018-10-27
Google analytics fix fixes #2965

### LINKED ISSUE #2965: GA Data Source throws an error when no rows returned
### Issue Summary

Google Analytics Data Source throws `Error running query: 'rows'` when the query result is empty.
I have a pretty simple query with dimensions and filters, like:

```json
{
  "ids": "ga:177xxxxxx",
  "start_date": "2018-10-08",
  "end_date": "2018-10-12",
  "metrics": "ga:uniqueEvents",
  "dimensions": "ga:dimension1,ga:dimension3",
  "filters": "ga:dimension2==userrole;ga:eventCategory==eventcategory;ga:eventAction==enentaction;ga:dimension1!=demo"
}
```

Sometimes it returns empty result as there is no data. This results in error in redash.

### Steps to Reproduce

1. Create the Google Analytics Data Source
2. Make some query returning zero rows
3. Execute it in query editor

`Error running query: 'rows'` will be thrown. While this might be considered not a bug, I'd expect just an empty result with no errors.

### Technical details:

* Redash Version: 5.0.1
* Browser/OS: Chrome/macOS
* How did you install Redash: docker-compose

alexanderlz:
resolves #2965


redash/query_runner/google_analytics.py
```diff
@@ -43,36 +43,37 @@ def parse_ga_response(response):
...
-                    value = datetime.strptime(value, '%Y%m%d%H')
-                elif len(value) == 12:
-                    value = datetime.strptime(value, '%Y%m%d%H%M')
-                else:
-                    raise Exception("Unknown date/time format in results: '{}'".format(value))
-
-            d[column_name] = value
-        rows.append(d)
+    if response and 'rows' in response:
+        for r in response['rows']:
```
arikfr: We can assume `response` exists because otherwise this function will fail when trying to load the columns. In this case, maybe we can simplify this with a `for r in response.get('rows', []):`, so we don't need the extra indentation?

alexanderlz: ok.
done!

arikfr:
Thanks!

PR #3020, opened 2018-10-28
athena&presto support column type and partition key

alison985:
Original issue: mozilla#185
Original PR: mozilla#201

Issue to post this upstream: mozilla#430
Previous PR code reviews trying to port this upstream: #2655 #2764 #2826

This PR adds the key font awesome icon to partition keys and column type in the schema browser for Presto and Athena data sources.

Image (with forced partition key):
screenshot 2018-10-28 13 46 17


jezdez:
Diving into this more we found the overlap with #2990 too great to merge this PR first and will instead re-use the code later once #2990 has landed.

Issue #2998, opened 2018-10-24
Charts lose responsive features after refreshing the dashboard


arikfr:
> On mobile, when refreshing the dashboard widget query, the legend moves from the bottom to the right. The chart is easier to read if the legend stays on the bottom. Weird that refresh causes it to move...

PR #3024, opened 2018-10-29
 Charts lose responsive features after refreshing the dashboard

kravets-levko:
Fixes getredash/redash#2998


arikfr:
The size of the chart is known when we render it, why doesn't the layout get correct sizes/positioning from the get go?

kravets-levko:
@arikfr because to align legend, we also need to know it's size, which is unavailable until chart is rendered

PR #3026, opened 2018-10-29
Add "Active at" column to user list.

### LINKED ISSUE #789: Feature Request: Add 'last active' column to the /users page

jezdez:
Add user detail sync system based on Redis backend.

There is a periodic Celery task that updates a new “details” JSONB column in the “user” table with the data from Redis.

Currently this is only used for tracking the date of last activity of a user but can be extended with other user information later.

Updates a few dependencies.

A previous iteration of this by @alison985 was #2781 (also contained in this PR).

Fix #789.


jezdez:
@arikfr This uses [walrus](https://walrus.readthedocs.io/) as a high-level interface for storing data in Redis (based on the regular redis Python library) but stores the user details eventually as a JSONB column in Postgres.

I think both techniques are useful for Redash and could be good candidate to move some of the custom code over to. E.g. there is plenty of custom code to write to Redis and the pseudo JSON columns that don't allow doing database-side queries of values nested in the stored JSON. JSONB has the additional advantage of being index automatically. Here's some good introduction in case you need it: https://www.compose.com/articles/using-json-extensions-in-postgresql-from-python-2/

Walrus of course is less idiomatic than SQLA's JSONB support but is still useful for Redash since it offers a bunch of helpers that may be useful such as caching, locking, rate-limiting, full text search, containers and an ORM based on Redis. On top of that it has support for the recently released [Redis stream data type](https://redis.io/topics/streams-intro) which could be interesting to dive into as a light-weight replacement for some of the things we use Celery for.

arikfr:
@jezdez 

> the pseudo JSON columns that don't allow doing database-side queries of values nested in the stored JSON.

Of course they do. You just cast them to JSON first:

```sql
SELECT additional_properties::json->>'screen_resolution',
               count(0),
               count(DISTINCT user_id)
FROM events
GROUP BY 1
```

But this shouldn't stop us from using JSONB. One thing to consider -- what version of Postgres was it introduced at?

jezdez:
> @jezdez
> 
> > the pseudo JSON columns that don't allow doing database-side queries of values nested in the stored JSON.
> 
> Of course they do. You just cast them to JSON first:
> 
> ```sql
> SELECT additional_properties::json->>'screen_resolution',
>                count(0),
>                count(DISTINCT user_id)
> FROM events
> GROUP BY 1
> ```

Oh well, I should have stated, "..without an extra step" :) Which reminds me, I'll need to create the GIN index for this column additionally.

> 
> But this shouldn't stop us from using JSONB. One thing to consider -- what version of Postgres was it introduced at?

This was added in 9.4. [Some](https://blog.2ndquadrant.com/nosql-postgresql-9-4-jsonb/) [more](https://blog.2ndquadrant.com/jsonb-type-performance-postgresql-9-4/) [info](https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.4#JSONB_Binary_JSON_storage).

arikfr:
> Which reminds me, I'll need to create the GIN index for this column additionally.

Why do we need an index?

> This was added in 9.4. Some more info.

This might be an issue. A large portion of the open source deployment were using the bootstrap script, which was installing Ubuntu's default postgresql package. For whoever used Ubuntu 14.04 it means PostgreSQL 9.3... Maybe we can use `json` type instead?

jezdez:
> > Which reminds me, I'll need to create the GIN index for this column additionally.
> 
> Why do we need an index?

It'll improve query times dramatically. Here's some good background article on [JSONB query performance](https://blog.2ndquadrant.com/jsonb-type-performance-postgresql-9-4/).

> 
> > This was added in 9.4. Some more info.
> 
> This might be an issue. A large portion of the open source deployment were using the bootstrap script, which was installing Ubuntu's default postgresql package. For whoever used Ubuntu 14.04 it means PostgreSQL 9.3... Maybe we can use `json` type instead?

Oh dang, 14.04 is EOL'd in April 2019 though, which just means I'm worried that we'd need to support such an old distro (and Postgres version). FWIW the official APT packages still support 14.04 (https://www.postgresql.org/download/linux/ubuntu/), so maybe that'd be a good work around for people wanting to upgrade Redash on those old systems.

redash/models.py
```diff
@@ -1369,7 +1526,7 @@ def all(cls, org, group_ids, user_id):
                  (Dashboard.user_id == user_id) |
                  ((Widget.dashboard != None) & (Widget.visualization == None))),
                 Dashboard.org == org)
-            .distinct())
```
jezdez: I had to remove this since Postgres would otherwise throw a `SELECT DISTINCT ON expressions must match initial ORDER BY expressions` error. @arikfr Do you know a better fix?

Makefile
```diff
@@ -26,3 +26,9 @@ watch: bundle
 
 start: bundle
 	npm run start
+
+redis-cli:
+	docker-compose run redis redis-cli -h redis
+
+bash:
+	docker-compose run server bash
```
arikfr: Worth adding `--rm` to all the ephemeral container we start.

client/app/pages/users/list.html
```diff
@@ -46,6 +46,10 @@
               Joined
               <sort-icon column="'created_at'" sort-column="$ctrl.paginator.orderByField" reverse="$ctrl.paginator.orderByReverse"></sort-icon>
             </th>
+            <th class="sortable-column" ng-click="$ctrl.paginator.orderBy('active_at')">
+              Active at
```
arikfr: Maybe "Last Active At"?

migrations/versions/e7f8a917aa8e_add_user_details_json_column.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'e7f8a917aa8e'
+down_revision = '71477dadd6ef'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('users', sa.Column('details', postgresql.JSON(astext_type=sa.Text()), server_default='{"active_at": null}', nullable=True))
```
arikfr: Maybe a better default would be `{}`? Just in case in the future we change the implementation of this feature, we don't get stuck with this default.

redash/__init__.py
```diff
@@ -47,21 +49,23 @@ def create_redis_connection():
...
         redis_password = redis_url.password and urllib.unquote(redis_url.password)
-        r = redis.StrictRedis(host=redis_url.hostname, port=redis_url.port, db=redis_db, password=redis_password)
+        r = cls(host=redis_url.hostname, port=redis_url.port, db=redis_db, password=redis_password)
 
     return r
 
 
 setup_logging()
 redis_connection = create_redis_connection()
+walrus_db = create_redis_connection(cls=walrus.Database)
```
arikfr: Can we reuse the same connection?

redash/__init__.py
```diff
@@ -131,13 +133,14 @@ def create_app(load_admin=True):
     db.init_app(app)
     migrate.init_app(app, db)
     if load_admin:
-        init_admin(app)
+        admin.init_admin(app)
     mail.init_app(app)
-    setup_authentication(app)
+    authentication.init_app(app)
```
arikfr: 👌

redash/handlers/organization.py
```diff
@@ -10,7 +10,7 @@
 @login_required
 def organization_status(org_slug=None):
     counters = {
-        'users': models.User.all_not_disabled(current_org).count(),
+        'users': models.User.all(current_org).count(),
```
arikfr: Why?

redash/models.py
```diff
@@ -96,6 +106,131 @@ def get(self, query_id):
...
+            logger.info(
+                'syncing users: %s',
+                ', '.join([str(uid) for uid in user_ids])
+            )
+            # get all SQLA users that need to be updated
+            users = User.query.filter(User.id.in_(user_ids))
+            for i, user in enumerate(users):
+                update = to_sync[user.id]
+                updated_at = getattr(update, 'updated_at', None)
+                if updated_at is None:
```
arikfr: What this is protecting from?

redash/models.py
```diff
@@ -96,6 +106,131 @@ def get(self, query_id):
...
+            pass
+        finally:
+            user_ids = [str(user_id) for user_id in to_sync.keys()]
+            if user_ids:
+                logger.info(
+                    'Deleting temporary user details for users %s',
+                    ', '.join(user_ids)
+                )
+                delete_query = [
+                    UserDetail.user_id == str(user_id)
```
arikfr: Won't this be a list a booleans?

redash/models.py
```diff
@@ -96,6 +106,131 @@ def get(self, query_id):
...
+
+    def init_app(self, app):
+        """
+        Initializes this pseudo Flask extension with the given
+        Flask app.
+        """
+        request_started.connect(self.update_user_detail, app)
+
+
+user_details = UserDetailsExtension()
```
arikfr: Maybe be worth extracting all the user details related code into its own module? 

For a long time I wanted to change `redash.models` into a package and extract different modules from it. While this is to be of a task to fit into this pull request, maybe we can start by moving all the code from `redash/models.py` into the `redash/models/__init__.py` and then gradually move things into modules?

redash/models.py
```diff
@@ -431,9 +580,16 @@ class User(TimestampMixin, db.Model, BelongsToOrgMixin, UserMixin, PermissionsCh
                      unique=True)
 
     disabled_at = Column(db.DateTime(True), default=None, nullable=True)
+    details_default = {u'active_at': None}
+    details = Column(MutableDict.as_mutable(postgresql.JSON), nullable=True,
+                     server_default=json_dumps(details_default),
+                     default=details_default)
+    active_at = json_cast_property(db.DateTime(True), 'details', 'active_at', default=None)
```
arikfr: Doesn't having default value set for `active_at` enough so we don't have to set a default value in the `details` object?

redash/models.py
arikfr: In some places I think we just did a group by, but using distinct proved to be a better solution. Why did it start happening?

redash/tasks/general.py
```diff
@@ -57,3 +57,16 @@ def send_mail(to, subject, html, text):
...
+
+
+@celery.task(
+    name="redash.tasks.sync_user_details",
+    ignore_result=True,
+    soft_time_limit=60,
+    time_limit=120,
+    # let the task expire after 45 seconds since there will be
+    # another task 15 seconds later anyway
+    expires=45,
```
arikfr: Clever!

redash/models.py
jezdez: In the method it loads the dashboard owners via a joinedload and by that tries to load the user detail column as well. When applying `distinct()` to the query it seems as if it also tries to make the join query distinct, which again fails with that error above.

Makefile
jezdez: Good idea.

client/app/pages/users/list.html
jezdez: Ok.

migrations/versions/e7f8a917aa8e_add_user_details_json_column.py
jezdez: Right, this is based on the default I'm setting in the models.py. I think it's okay to have those differ, I basically added it in models.py to prevent KeyError when ordering the queries.

redash/models.py
arikfr: Maybe we can use a different strategy to load the users?

migrations/versions/e7f8a917aa8e_add_user_details_json_column.py
arikfr: See my other comment: https://github.com/getredash/redash/pull/3026#discussion_r232571363 :)

redash/__init__.py
jezdez: We sadly can't since the [`walrus.database.Database`](https://github.com/coleifer/walrus/blob/e9cd555a39fd9eb97d3301311f135a6c39df8b27/walrus/database.py#L58) class is a subclass of [`redis.Redis`](https://github.com/andymccurdy/redis-py/blob/3e3c96ffdbf44d6daba841b8b77f6dba4ec02e8a/redis/client.py#L2779) and not [`redis.StrictRedis`](https://github.com/andymccurdy/redis-py/blob/3e3c96ffdbf44d6daba841b8b77f6dba4ec02e8a/redis/client.py#L406) which was added to redis-py to fix some argument ordering differences compared to the Redis commands and other API and behavior inconsistencies.

I guess we could switch to using the redis.Redis based walrus Database class instead since the differences aren't that many and we simply could update our use of the methods with the different argument order. I don't mind either way.

redash/models.py
jezdez: Nah, it'll be a list of query expressions since `UserDetail.user_id` is a field instance, which subclasses the [Node class](https://github.com/coleifer/walrus/blob/e9cd555a39fd9eb97d3301311f135a6c39df8b27/walrus/query.py#L164-L195) from walrus' query module which uses operator overloading to return new expressions. That way they can be used in an OR query later below.

redash/models.py
jezdez: I like that, lemme do that.

redash/models.py
jezdez: Good question, lemme check.

redash/handlers/organization.py
jezdez: If you look at https://github.com/getredash/redash/pull/3026/commits/56cd0da386f458538fe7f01a473ced6d4e4fbdeb#diff-2eb9bad8b59774e58b2c9a664440905aR516 it shows that `User.all` was already doing that, so `User.all_not_disabled` was redundant.

redash/models.py
jezdez: Ah, that's a left-over from the previous cases where we didn't delete user details objects and key revocation policy might have deleted the value of the individual walrus field. This is moot now with the fact we delete the whole set of values from Redis.

migrations/versions/e7f8a917aa8e_add_user_details_json_column.py
jezdez: Yeah, that worked.

redash/__init__.py
jezdez: So ironically redis-py 3.0.0 was released today that removes the two different Redis and StrictRedis classes in favor of the StrictRedis class which will be the new default. More details: https://github.com/andymccurdy/redis-py#upgrading-from-redis-py-2x-to-30

redash/models.py
jezdez: Fixed.

redash/models.py
jezdez: I've done that now, and it feels good to do part of that. FWIW due to the import system I had to do a few more of the models that just the user details, to prevent circular imports. The majority of the models is still in redash/models/__init__.py though.

redash/models.py
jezdez: Yep, fixed now.

redash/tasks/general.py
jezdez: Yeah, it simply prevents from the queue filling up with similar tasks.

redash/__init__.py
arikfr: I wonder if there is a plan for walrus to support redis-py 3.0.0?

jezdez: Ah, d'oh, haden't caught up with my GitHub inbox: walrus 0.7.0 with redis-py 3.0.0 support is out for a few days now: https://github.com/coleifer/walrus/blob/master/CHANGELOG.md#070

jezdez: Okay, updated our code to work with the new redis-py version, but kombu needs an update to happen to land this: https://github.com/celery/kombu/commit/05152da53d169371b8b87718f33a6f47e97b7ba0

arikfr: Did they release a new version already?

jezdez: Sadly no, https://pypi.org/project/kombu/#history

jezdez: Ah, yeah, the [4.2.2 release](https://github.com/celery/kombu/compare/v4.2.1...v4.2.2) has it. I'll rebase the PR and add it to the requirements file.

jezdez:
@arikfr Given the size of changes, would you mind taking another look soonish?

arikfr:
Sure 👍

jezdez:
Another rebase needed :(

redash/models/users.py
```diff
@@ -0,0 +1,442 @@
...
+
+        The fields uid, rid and updated (timestamp) are
+        enforced and can't be overwritten.
+        """
+        # try getting the user detail with the given user ID
+        # or create one if it doesn't exist yet (e.g. when key was purged)
+        try:
+            user_detail = cls.get(cls.user_id == user_id)
+            # update the timestamp with the current time
+            user_detail.updated_at = utcnow()
```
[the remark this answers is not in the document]
rauchy: 1. I agree that adding another call to every request might seem expensive, but what scale in ms are we talking about? 5ms? 50ms? >100ms? My point is, if it's negligible - I prefer the simpler approach for readability.

2. I like `walrus.Hash` for readability.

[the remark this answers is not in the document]
rauchy: Reading through the rest of the walrus-related code, I'm thinking that a simpler implementation might actually be the `hset` one 😮

arikfr:
Merged 🍾 I'm going to do a few changes in a follow up pull request.

jezdez:
Oh hey, was about to tackle it this week, glad you got to it first 😍

redash/models/users.py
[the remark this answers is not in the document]
jezdez: FWIW, there is some level of premature optimization involved here as I implemented this feature just like @arikfr mentioned above first but saw it as another case of code that deals with update/refreshing Redis keys, which Redash has in many other places already. In other words I found the Walrus "models" to be a better abstraction for that than the low-level code, similar to how we use SQLAlchemy for RDB data. I don't think the runtime of individual Redis commands is a showstopper considering the difference in performance behavior compared to other parts of the request/response cycle. Of course that's just my opinion and you're free to have a different one :)

redash/utils/__init__.py
```diff
@@ -81,10 +81,25 @@ def default(self, o):
             return float(o)
         elif isinstance(o, (datetime.timedelta, uuid.UUID)):
             return str(o)
-        elif isinstance(o, (datetime.date, datetime.time)):
+        # See "Date Time String Format" in the ECMA-262 specification.
```
arikfr: In the future please open a separate pull request for such changes, so they can be probably reviewed and discussed. I completely missed this when reviewing #3026. 

My main concern here is line 98, which breaks queries like `select current_time` (on Postgres).

PR #3032, opened 2018-10-31
Makefile: Add make targets for test

koooge:
Hi. This is a proposal for development.

`make up`: build & up services
`make down`: down services
`make test`: up services, create DB table for test, and then exec tests. 
(I created based on .circleci/config.yml)

Existing targets was left because I don't know if they are used.


Makefile
```diff
@@ -12,12 +15,29 @@ create_database:
...
 bundle:
 	docker-compose run server bin/bundle-extensions
 
 tests:
 	docker-compose run server tests
 
+lint:
+	./flake8_tests.sh
+
+backend-unit-tests: up test_db
```
arikfr: Won't this try to run `test_db` everytime and hence fail on future tries?

koooge: Yes, it is not idempotent. 🤔 
Ho about *test_db* recreating the database tests?

Makefile
```diff
@@ -1,23 +1,43 @@
...
+	sleep 10
+	docker-compose run --name tests server tests
```
jezdez: IIUC this will not stop the Docker containers that are spun up by the `up` target. Could we add that please?

Makefile
```diff
@@ -1,23 +1,43 @@
...
+	docker-compose down
+
 bundle:
 	docker-compose run server bin/bundle-extensions
 
 tests:
 	docker-compose run server tests
 
+lint:
+	./flake8_tests.sh
```
jezdez: Which reminds me, could we move this script into the `bin/` directory as all the other scripts?

Makefile
koooge: Yes, `up(docker-compose up)` does not stop containers. But it will build and restart containers if your source have any changes.
`make test` will recreates the container, recreates database *tests* and does test everytime.

koooge: I misunderstood. I will add --rm option to `tests` container.

jezdez: The `--rm` option isn't enough to stop the containers that are launched in the `up` target since it doesn't work in detached mode.

Instead you'll have to call `docker-compose stop` explicitly at the end of the `test` target to make sure running `make test` starts and stops the whole thing.

That said personally I'd prefer if we'd not use `docker-compose up -d` at all since it tends to keep lingering docker containers when I run tests a lot locally that sometimes crash or are CTRL+C'd. You can achieve the same with replacing the `up` task the `compose_build` target as a dependency of the `backend-unit-tests` target. Then `--rm` will be enough.

arikfr: Actually I keep the "main" containers running all the time and only stop them when no longer needed. I don't think the tests command should start those containers anyway. Once it doesn't, it's up to each one to decide when to start or stop them, based on their preference.

koooge: I thought it would be better that the implicit test workflow is coded by `make test`. Redash experts can test whatever they prefer.

I think `make test` should outputs log of the test results. So "main" containers run in detached mode. Besides they will be still running after the test because Redash developers need their log to check the behavior.

jezdez: > I thought it would be better that the implicit test workflow is coded by `make test`. Redash experts can test whatever they prefer.

FWIW I would argue it's our goal to lower the barrier for contributors to run the tests, so no need for an expert level of knowledge to do that.

> I think `make test` should outputs log of the test results. So "main" containers run in detached mode. Besides they will be still running after the test because Redash developers need their log to check the behavior.

FYI when I need to debug something while working on a patch I usually run a bash inside the server container and work from there, e.g. runnning `export REDASH_DATABASE_URL="postgresql://postgres@postgres/tests"` there and calling pytest directly to be able to pass additional command line options, use pdb, opening the Python shell etc. Hence the reason to use something like `make test` would be to have a catch-all command that does what the Circle CI would do as well.

jezdez: > Actually I keep the "main" containers running all the time and only stop them when no longer needed. I don't think the tests command should start those containers anyway. Once it doesn't, it's up to each one to decide when to start or stop them, based on their preference.

IOW assuming a contributor wouldn't have a container running he'd be able to get the whole thing by running `make up test down`?

koooge: @jezdez 
> IOW assuming a contributor wouldn't have a container running they'd be able to get the whole > thing by running make up test down?
Yes.
I assume that developers can make sure all tests pass to just type `make test` before creating PR. (you have to call `make down` explicitly) 

@arikfr It causes errors because just `make tests` doesn't create database *test*. I'd like to just test by one command.

jezdez: > @jezdez
> 
> > IOW assuming a contributor wouldn't have a container running they'd be able to get the whole > thing by running make up test down?
> > Yes.
> > I assume that developers can make sure all tests pass to just type `make test` before creating PR. (you have to call `make down` explicitly)

Understood. When this is documented the `make down` step needs to be mentioned so people know what happened.

koooge:
#3038 :eyes:

jezdez:
@koooge I've left a comment there ;)

arikfr:
@jezdez can we hit the "Squash and merge"? :)

jezdez:
@arikfr Ok 🤷‍♂️

arikfr:
Thanks, @koooge !

PR #3036, opened 2018-10-31
Update Flask-Admin to 1.5.2

yoavbls:
I don't know if someone uses flask-admin but I upgraded it because of this bug:
https://github.com/flask-admin/flask-admin/issues/1588
In 1.4.2 you cant add and edit records/


arikfr:
Thanks!

PR #3047, opened 2018-11-07
Druid query runner

rauchy:
Here's my attempt at support for running SQL queries against a Druid.

Since I pretty much followed [this guide](https://discuss.redash.io/t/creating-a-new-query-runner-data-source-in-redash/347) for creating this query runner, I'm left with a feeling that there's nasty code duplication (in `run_query` and `get_schema`) - should we find a way to reuse these?


redash/query_runner/druid.py
```diff
@@ -0,0 +1,93 @@
...
+    from pydruid.db import connect
+    enabled = True
+except ImportError:
+    enabled = False
+
+from redash.query_runner import BaseQueryRunner, register
+from redash.utils import json_dumps, json_loads
+
+TYPES_MAP = {
+}
```
arikfr: image

?

redash/query_runner/druid.py
```diff
@@ -0,0 +1,93 @@
...
+                    "type": "number",
+                    "default": 8082
+                },
+                "scheme": {
+                    "type": "string",
+                    "default": "http"
+                }
+            },
+            "order": ['scheme', 'host', 'port'],
+            "required": ['host']
```
arikfr: No username/password support?

redash/query_runner/druid.py
kravets-levko: As far as I understand, Druid supports only one custom data type - `OTHER/COMPLEX`; however, I'm not sure how to represent it in Redash

arikfr: The types map is not only for custom stuff.

redash/query_runner/druid.py
rauchy: While there seems to be [support for basic auth](https://github.com/druid-io/pydruid/blob/master/pydruid/client.py#L40) in the client, the docs for Druid's "experimental" SQL dialect don't show anything regarding authentication. Perhaps they expect people to protect endpoints by other means?

arikfr: 👍 OK, let's just keep as is and if someone asks for it, revisit.

arikfr:
:shipit:

PR #3056, opened 2018-11-08
Support overriding the default Celery schedule database file

touchardv:
Support overriding the default Celery schedule database file via SCHEDULE_DB environment variable.

By default Celery will use a file celerybeat-schedule in the current directory.
This is an issue in a Kubernetes/Openshift environment as the file may be lost or even impossible to write.


arikfr:
Thanks!

PR #3057, opened 2018-11-08
New Celery/Queries Execution Status API

### LINKED ISSUE #2173: Runtime is inaccurate on active queries list
### Issue Summary
Runtime in active queries list does not reflect actual runtime.

### Steps to Reproduce

1. Go to the list of running queries: https://redash.prod.aws.pbinfra.com/admin/queries/tasks
2. Runtime is never updated from 0s

This differs from my expectations of a constantly incrementing runtime counter.

### Technical details:

Browser: Firefox Quantum 57.0.1 (64-bit)
Redash Version: Redash 4.0.0+b3169
How did you install Redash: Deployed as an ECS service

arikfr:
- [x] Remove QueryTaskTracker related code.
- [x] Add Celery introspection based API
   - [x] Show started_at timestamp
- [x] Update admin screens:
   - [x] Load data only once (or maybe cache it on the server?)
   - [x] Filter non query executions from queries list
- [x] Add a "queued at" timestamp to jobs, so we can track this too
- [x] Write a migration that clears QueryTaskTracker objects from Redis

Closes #2173.


redash/monitor.py
```diff
@@ -70,3 +74,66 @@ def get_status():
...
+        return query_tasks 
+
+    for task in itertools.chain(*tasks.values()):
+        data_source_id = None
+        query_id = None
+        user = None
+
+        # TODO: we should support parsing all the tasks, not only query ones.
+        if task['name'] == 'redash.tasks.execute_query':
+            args = ast.literal_eval(task['args'])
```
arikfr: 🤢

arikfr: It looks like they changed something in how this works and it no longer works. The good news are that because of this I found a better way to access the data: Celery gives you the option to set your own representation of the arguments, so I piggy back on this to provide a JSON object.

arikfr:
Updated UI:



The gif is a fail, but you get the idea.

Still WIP. Need to improve the UX + clean up the code.

migrations/versions/e5c7a4e2df4d_remove_query_tracker_keys.py
```diff
@@ -0,0 +1,51 @@
...
+    remove_count = min(max_keys, count - keep_count)
+    keys = redis_connection.zrange(list_name, 0, remove_count - 1)
+    redis_connection.delete(*keys)
+    redis_connection.zremrangebyrank(list_name, 0, remove_count - 1)
+    return remove_count
+
+
+def prune_all(list_name):
+    removed = 1000
+    while removed > 0:
```
rauchy: What's the importance of deleting in chunks here?

redash/monitor.py
```diff
@@ -18,29 +23,20 @@ def get_object_counts():
...
+    query = models.db.session.execute(
+        'select distinct queue_name from data_sources union all select distinct scheduled_queue_name from data_sources')
 
-    return queues
+    return ['celery'] + [row[0] for row in query]
 
 
 def get_queues_status():
-    queues = get_queues()
+    queues = {}
```
rauchy:
```suggestion
    return dict((q, {'size': redis_connection.llen(q)}) for q in get_queues())
```

client/app/components/admin/CeleryStatus.jsx
```diff
@@ -0,0 +1,220 @@
...
+    } else {
+      otherTasks.push(task);
+    }
+  });
+
+  return { queues: values(queues), queries, otherTasks, counters };
+}
+
+function QueuesTable({ loading, queues }) {
+  const columns = [
```
rauchy:
```suggestion
  const columns = ['Name', 'Active', 'Reserved', 'Waiting'].map(c => ({ title: c, dataIndex: c.toLowerCase() }));
```

client/app/components/admin/CeleryStatus.jsx
```diff
@@ -0,0 +1,220 @@
...
+    queues: [],
+  };
+
+  constructor(props) {
+    super(props);
+    this.fetch();
+  }
+
+  fetch() {
+    // TODO: handle error
```
rauchy: giphy

client/app/components/admin/CeleryStatus.jsx
```diff
@@ -0,0 +1,220 @@
...
+          );
+        },
+      },
+      Columns.timeAgo({ title: 'Start Time', dataIndex: 'start_time' }),
+    ];
+
+    const queryColumns = commonColumns.concat([
+      Columns.timeAgo({ title: 'Enqueue Time', dataIndex: 'enqueue_time' }),
+      {
+        title: 'Query ID',
```
rauchy: You can already guess what I want to suggest here 😝

migrations/versions/e5c7a4e2df4d_remove_query_tracker_keys.py
arikfr: I don't know what limits on number of keys `delete` has. In some cases this list can be large (think hundreds of thousands), so better to be safe.

redash/monitor.py
arikfr: I'll pass on this one. Feels less readable to me: even knowing what it should do, I got confused for a moment about what exactly it does.

client/app/components/admin/CeleryStatus.jsx
arikfr: Not trivial because then I need to add a function that will convert `query_id` to `Query ID`. We have something along this lines in code, but it will do `Query Id`.

rauchy: Oh, I was thinking the other way around ('Query ID' -> to lower -> snake case)

redash/monitor.py
```diff
@@ -18,29 +23,20 @@ def get_object_counts():
 
 
 def get_queues():
-    queues = {}
-    for ds in models.DataSource.query:
-        for queue in (ds.queue_name, ds.scheduled_queue_name):
-            queues.setdefault(queue, set())
-            queues[queue].add(ds.name)
+    query = models.db.session.execute(
+        'select distinct queue_name from data_sources union all select distinct scheduled_queue_name from data_sources')
```
jezdez: In case you want to write this a Python 😁: 
```python
queue_names = db.session.query(DataSource.queue_name).distinct()
scheduled_queue_names = db.session.query(DataSource.scheduled_queue_name).distinct()
query = models.db.session.execute(union_all(queue_names, scheduled_queue_name))
```

arikfr: Thanks!

arikfr: 212dcbb9b716b5efa3fcfa2bb626bbb70d974ac5

Although in retrospect I'm not sure what we gained by converting it to Python.

client/app/components/admin/CeleryStatus.jsx
arikfr: https://github.com/getredash/redash/pull/3057/commits/254106f1ef840876f2e068c7ecb4242788ddfd3f

client/app/components/admin/CeleryStatus.jsx
arikfr: https://github.com/getredash/redash/pull/3057/commits/b0f2248d6041f1087a109e40df5bba3b9822c49a

Issue #3042, opened 2018-11-06
Backslash in dropdown list value transformed to &#x2F; in query


GilAdirim:
I'm using parameters to filter a query. The values of the parameter are pre-calculated by a query. One of the values is `Other/Organic`.
When I select this value in the dropdown and execute the query, the underlying string in the SQL is changed to `Other&#x2F;Oranic`.

Redash Version: Hosted (master?)

kravets-levko:
Hi! Can you try to use three curly braces for the parameter (i.e. `{{{ param }}}`) and tell me if it works? Thanks!

GilAdirim:
Yes, that works - although I don't remember seeing anything about it in the documentation...

kravets-levko:
Well, we also noticed this few days ago - another user asked similar question on the forum. Therefore there are no anything about this in the docs. 

We use [Mustache](https://mustache.github.io/) to substitute parameter values, but by default it html-escapes some characters like `<`, `>`, `&` or `/`. Three curly braces allow to use unescaped value.

I'll close this issue because your problem is solved; the proper fix will be available soon.

kravets-levko:
@arikfr, hm, I missed that we use `pystache` on backend. Will fix it and then ping you

Issue #2904, opened 2018-10-07
Move parameters merging to the backend and make it safe


arikfr:
Currently the way parameters work in Redash is that we use the query text as a [Mustache template](http://mustache.github.io/) and merge the parameter values with it before running the query.

This very simple implementation has one major issue: it's open to SQL injections. It's a non issue when a "regular" Redash user uses parameters, as they can run any query anyway. But it limits the use of parameters to users with full access to the data source.

To provide any user with the ability to use parameters, we need to change our implementation to make it safe and SQL injections proof. 

The idea is:
1. Change the API to handle all parameters merging in the backend (today we send the final query from frontend).
2. Make the API safe.

While every database driver we use (that supports DB-API) has its own way of taking query text with parameters, it's better that we find a single way that is DB agnostic. Also need to remember that some of our datasources take queries as JSON/YAML.

It's also preferable that we can support current parameters syntax (`{{ parameter }}`), but we don't have to support the full range of Mustache functions (although it can be nice).

---

Some "prior art" that might be useful:

1. https://github.com/hashedin/jinjasql
2. [sqlalchemy.sql.expression.text](https://docs.sqlalchemy.org/en/latest/core/sqlelement.html#sqlalchemy.sql.expression.text)
3. [psycopg2.sql](https://github.com/psycopg/psycopg2/blob/master/lib/sql.py) module
4. https://github.com/deadpixi/dpdb

arikfr:
(I added another prior art reference I stumbled at today)

Jakdaw:
Is the intention that this would be optional?

Given the current behaviour I'm sure there are plenty of people who are using parameters not only for SQL-parameters but also for selecting tables, group by columns, order by fields etc. It's not going to be possible to avoid the SQL injection possibilities in these situations but do we really want redash to prevent such use-cases altogether?

arikfr:
@Jakdaw that's a good comment. There is no intention to prevent such scenarios. The idea is only to prevent SQL injections.

rauchy:
Are you referring to injections via the WHERE clause only? Because I think all the other clauses are susceptible to SQL injection. (`select 1; drop table midgets`)

rauchy:
Oh yeah, if we go down the JinjaSQL path, [we're on our own](https://github.com/hashedin/jinjasql#sql-safe-strings) for anything that can't be bound.

susodapop:
[...]

PR #3058, opened 2018-11-08
Mustache: don't html-escape query parameters values

kravets-levko:
When using `{{ param }}` syntax, Mustache by-default will html-escape them. Possible work-around is to use `{{{ param }}}` syntax to render unescaped value, but most users don't know about it. Also, html-escaping is meaningless in SQL or JSON. This fix allows to use `{{ param }}` syntax to render unescaped values.

`mustache.js` does not have any options to affect this behavior; it uses `Mustache.escape` function to escape values, so we'll monkey-patch it with function that just returns it's argument.

`pystache` has an option to disable html-escaping.

### DIFF
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -3,8 +3,10 @@ import debug from 'debug';
 import Mustache from 'mustache';
 import {
   each, zipObject, isEmpty, map, filter, includes, union, uniq, has,
-  isNull, isUndefined, isArray, isObject,
+  isNull, isUndefined, isArray, isObject, identity,
 } from 'lodash';
+
+Mustache.escape = identity; // do not html-escape values
 
 const logger = debug('redash:services:query');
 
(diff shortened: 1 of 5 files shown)


kravets-levko:
@arikfr AFAIK we don't substitute parameters on backend - API accepts ready-to-execute SQL. We'll need to remember about this when implementing #2904, but currently it's only frontend-related.

kravets-levko:
@arikfr, hm, I missed that we use `pystache` on backend. Will fix it and then ping you

kravets-levko:
@arikfr Updated PR

arikfr:
👌

kravets-levko:
Some related links:
Issue #3042
Forum [topic](https://discuss.redash.io/t/error-running-query-with-multiple-parameters-used-next-to-each-other/2615)

PR #3061, opened 2018-11-09
Table visualization horizontal scrollbar should not be always visible

kravets-levko:


arikfr:
👍

rollie42:
I think this change caused https://github.com/getredash/redash/issues/3290, the effect being for vertically large tables, you can't scroll horizontally unless you first scroll to the bottom of the table, but then you can't see the table header.

Should this be reverted? It makes sense to hide scrollbar when not needed, but having it when it is needed seems more important.

ranbena:
@rollie42 perhaps #4101 already fixed that behavior? (no need to scroll to bottom anymore)

rollie42:
@ranbena Indeed! Glad to see :) Maybe #3290 can just be closed as dupe then.

PR #3063, opened 2018-11-11
Colors can now be controlled for bubble charts in UI.

toph:
Bubble chart marker size override was previously clobbering seriesColor.


arikfr:
Thanks!

Issue #3055, opened 2018-11-08
Use Page Object model for Cypress tests


jrbenny35:
## Issue Summary

New cypress tests use locators within the tests to find elements. We should switch to using a Page Object Model to assist with upcoming UI changes, i.e.: react

### Technical details:

Some good information on the concept is found [here](https://medium.com/tech-tajawal/page-object-model-pom-design-pattern-f9588630800b).

jrbenny35:
Here is a test suite I wrote for an example: https://github.com/mozilla/send/tree/vnext/test/integration

gabrieldutra:
Hi @jrbenny35,
This does make tests look really better.  Nice example with your test suite. I also found [this example](https://medium.com/reactbrasil/deep-diving-pageobject-pattern-and-using-it-with-cypress-e60b9d7d0d91) with Cypress.

arikfr:
This was discussed in #3064 and we decided for now to wait until we have more tests before we implement some abstraction.

PR #3064, opened 2018-11-11
Update Cypress to use Page Objects

gabrieldutra:
## Description
This updates Cypress tests to have a start on using Page Objects (#3055).

## Changes
- `npm run cypress:server start` was failing on dev environment because `--build-args` is not supported in version 2 of docker-compose - `--build-args` is now only used when running CI;
- Most of selectors were updated to use `data-testid` which is Cypress' recommended approach to select elements;
- Some page objects were created in order to start abstracting Pages.

Please give me your thoughts on what could be improved :grin:

### DIFF
--- a/cypress/cypress-server.js
+++ b/cypress/cypress-server.js
@@ -17,10 +17,13 @@ function execSetup() {
   post(baseUrl + '/setup', { formData: setupData });
 }
 
-function startServer() {
+function startServer(skipDependencies) {
   console.log('Starting the server...');
-            
-  execSync('docker-compose -p cypress build --build-arg skip_ds_deps=true', { stdio: 'inherit' });
+
+  if (skipDependencies) {
+    execSync('docker-compose -p cypress build --build-arg skip_ds_deps=true', { stdio: 'inherit' });
+  }
+
   execSync('docker-compose -p cypress up -d', { stdio: 'inherit' });
   execSync('docker-compose -p cypress run server create_db', { stdio: 'inherit' });
 }
@@ -34,11 +37,11 @@ const command = process.argv[2];
 
 switch (command) {
   case 'start':
-    startServer();
+    startServer(false);
     execSetup();
     break;
   case 'start-ci':
-    startServer();
+    startServer(true);
     break;
   case 'setup':
     execSetup();
--- a/cypress/pages/LoginPage.js
+++ b/cypress/pages/LoginPage.js
@@ -1,0 +1,37 @@
+import Page from './Page';
+
+class LoginPage extends Page {
+  constructor() {
+    super('/login');
+    this.emailInput = '[data-testid=EmailInput]';
+    this.passwordInput = '[data-testid=PasswordInput]';
+    this.submitButton = '[data-testid=SubmitButton]';
+    this.errorMessage = '[data-testid=ErrorMessage]';
+  }
+
+  fillEmail(value) {
+    cy.get(this.emailInput)
+      .clear()
+      .type(value);
+
+    return this;
+  }
+
+  fillPassword(value) {
+    cy.get(this.passwordInput)
+      .clear()
+      .type(value);
+
+    return this;
+  }
+
+  submit() {
+    cy.get(this.submitButton).click();
+  }
+
+  validateTitle() {
+    cy.title().should('eq', 'Login to Redash');
+  }
+}
+
+export default LoginPage;
--- a/cypress/pages/NewDataSourcePage.js
+++ b/cypress/pages/NewDataSourcePage.js
@@ -1,0 +1,31 @@
+import Page from './Page';
+
+class NewDataSourcePage extends Page {
+  constructor() {
+    super('/data_sources/new');
+    this.databaseSourceList = '[data-testid=DatabaseSourceList]';
+    this.dynamicForm = '[data-testid=DynamicForm]';
+    this.saveButton = '[data-testid=SaveButton]';
+  }
+
+  selectSource(sourceName) {
+    cy.get(this.databaseSourceList)
+      .contains(sourceName)
+      .click();
+  }
+
+  fill(fieldName, value) {
+    cy.get(this.dynamicForm)
+      .get('[data-testid="' + fieldName + '"]')
+      .clear()
+      .type(value);
+
+    return this;
+  }
+
+  submit() {
+    cy.get(this.saveButton).click();
+  }
+}
+
+export default NewDataSourcePage;
--- a/client/app/components/app-header/app-header.html
+++ b/client/app/components/app-header/app-header.html
@@ -86,7 +86,7 @@
           <!--<a href="users" title="Settings"><i class="fa fa-cog"></i></a>-->
         <!--</li>-->
         <li class="dropdown" uib-dropdown>
-          <a href="#" class="dropdown-toggle dropdown--profile" uib-dropdown-toggle data-cy="dropdown-profile">
+          <a href="#" class="dropdown-toggle dropdown--profile" uib-dropdown-toggle data-testid="ProfileDropdown">
             <img ng-src="{{ $ctrl.currentUser.profile_image_url }}" class="profile__image--navbar" width="20"/>
             <span class="dropdown--profile__username" ng-bind="$ctrl.currentUser.name"></span> <span
             class="caret caret--nav"></span></a>
--- a/client/app/components/dynamic-form.html
+++ b/client/app/components/dynamic-form.html
@@ -1,7 +1,7 @@
-<form name="dynamicForm">
+<form name="dynamicForm" data-testid="DynamicForm">
     <div class="form-group required" ng-class='{"has-error": (dynamicForm.targetName | showError)}'>
         <label class="control-label" for="dataSourceName">Name</label>
-        <input type="string" class="form-control" name="targetName" ng-model="target.name" autofocus required>
+        <input type="string" class="form-control" name="targetName" ng-model="target.name" data-testid="TargetName" autofocus required>
         <error-messages input="dynamicForm.targetName" form="dynamicForm"></error-messages>
     </div>
     <hr>
@@ -9,12 +9,12 @@
         <label ng-if="field.property.type !== 'checkbox'" class="control-label">{{field.property.title || field.name | toHuman}}</label>
         <input name="input" type="{{field.property.type}}" class="form-control" ng-model="target.options[field.name]" ng-required="field.property.required"
                ng-if="field.property.type !== 'file' && field.property.type !== 'checkbox'" accesskey="tab" placeholder="{{field.property.default}}"
-               data-cy="{{field.property.title || field.name | toHuman}}">
+               data-testid="{{field.property.title || field.name | toHuman}}">
 
         <label ng-if="field.property.type=='checkbox'">
             <input name="input" type="{{field.property.type}}" ng-model="target.options[field.name]" ng-required="field.property.required"
                    ng-if="field.property.type !== 'file'" accesskey="tab" placeholder="{{field.property.default}}"
-                   data-cy="{{field.property.title || field.name | toHuman}}">
+                   data-testid="{{field.property.title || field.name | toHuman}}">
             {{field.property.title || field.name | toHuman}}
         </label>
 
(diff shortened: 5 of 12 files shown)


arikfr:
Thanks, @gabrieldutra. I have some concerns about the complexity and extra code POM adds to the implementation and it makes it much easier to discuss with actual code examples. 

I think there are two assumptions we can agree on:

* We maintain state by using the API and not driving the UI (think of `cy.login()` vs. explicitly visiting the Login page and submitting the details).
* All tests are isolated.

Considering these two assumptions, I think it becomes clear that most page objects _won't_ be reused across tests. So the result of creating page objects is just extra code/complexity and the need to use two (or more) files when authoring tests, while the gain is minimal.

Also worth noting that non of the page objects contain state, so why the user needs to create an instance every time? I think it made sense in Selenium tests, where you pass the driver to the Page object. But in Cypress there is no driver to pass -- you run in a managed environment where `cy` is global.

Anyway, we could have similar benefit from having helper functions in the test code. For example, the login tests become:

```javascript
function fillLogin(email, password) {
  cy.get('#inputEmail').type(email);
  cy.get('#inputPassword').type(password);
}


describe('Login', () => {
  beforeEach(() => {
    cy.visit('/login');
  });

  it('greets the user', () => {
    cy.contains('h3', 'Login to Redash');
  });

  it('shows message on failed login', () => {
    fillLogin('admin@redash.io', 'wrongpassword{enter}')

    cy.get('.alert').should('contain', 'Wrong email or password.');
  });

  it('navigates to homepage with successful login', () => {
    fillLogin('admin@redash.io', 'password{enter}')

    cy.title().should('eq', 'Redash');
    cy.contains('Example Admin');
  });
});
```

We no longer have duplicate code (filling out the form), but we also didn't have to introduce all this indirection. In no other test code we will ever use the login form, so why have a `LoginPage` object?

There are places where some reuse will happen across tests, like the NavBar and the Profile menu. But for this it's enough to have some helper function:

```javascript
Cypress.Commands.add('openProfileMenu', () => {
  cy.get('ProfileDropdown').click();
});
```

If we want to have more namespacing than just putting everything on a global `cy` object, we can import it from some file:

```javascript
// in cypress/common/nav_bar.js:

export function openProfileMenu() {
  cy.get('ProfileDropdown').click();
}

// In test code:
import openProfileMenu from '../../common/nav_bar.js';
```

I made be overlooking some benefit so happy to have this discussion. @jrbenny35 you're welcome to share your thoughts from your experience of using POM in the past as well.

Thanks!

cypress/pages/LoginPage.js
```diff
@@ -0,0 +1,38 @@
...
+
+    return this;
+  }
+
+  fillPassword(value) {
+    cy.get(this.passwordInput)
+      .clear()
+      .type(value);
+
+    return this;
```
jrbenny35: You could put ```fillPassword``` and ```fillEmail``` into one function, and maybe call it ```fillAndSubmit```, as there is probably very little chance we would want to just fill one or the other.

cypress/pages/NewDataSourcePage.js
```diff
@@ -0,0 +1,31 @@
...
+      .click();
+  }
+
+  fill(fieldName, value) {
+    cy.get(this.dynamicForm)
+      .get('[data-testid="' + fieldName + '"]')
+      .clear()
+      .type(value);
+
+    return this;
```
jrbenny35: You could also put the submit button in ```fill()``` as these actions are usually done at the same time.

jrbenny35:
@gabrieldutra really good job on this. Looks awesome. Very easy to follow as well.

So I think the main thing with Page Objects is better organization. A good test suite is easily readable and easy to follow, in order to aide with debugging and adding new tests.

> I think it becomes clear that most page objects won't be reused across tests. So the result of creating page objects is just extra code/complexity and the need to use two (or more) files when authoring tests, while the gain is minimal.

The "gain" would be organization and code quality. The POM model does help with reuse, yes, but its main goal is to represent pages in code for use with automation. Having one place where the representation of a page exists, that is easily defined, can prove useful when the actual HTML/CSS is changed. Since redash is going to get a full React update, we are actually ahead of the curve for once.

> Anyway, we could have similar benefit from having helper functions in the test code. For example, the login tests become:

While this could work, just imagine this for every sort of small action needed. Like for searching, or interacting with a drop down. You would have little helper functions for a lot of things and it would get quite confusing for someone who wanted to write UI tests. Having those actions on in their own .js files would keep the test suite clean and aide in readability.

> In no other test code we will ever use the login form, so why have a LoginPage object?

How true is this? In the python test suite I wrote for Mozilla's redash instance, we must login for each test. Now this is probably because selenium does not allow cookies, but is this true of cypress? We may actually need a ```beforeEach()``` that logs in for each test.

Also, there is a way to create the objects on import so that there isn't a need to create a new object each time, but my JS isn't quite there yet.

cypress/cypress-server.js
```diff
@@ -17,10 +17,13 @@ function execSetup() {
   post(baseUrl + '/setup', { formData: setupData });
 }
 
-function startServer() {
+function startServer(skipDependencies) {
   console.log('Starting the server...');
-            
-  execSync('docker-compose -p cypress build --build-arg skip_ds_deps=true', { stdio: 'inherit' });
+
+  if (skipDependencies) {
```
gabrieldutra: This is not necessary anymore since `docker-compose.yml` is in version 3 now.

arikfr:
Thank you for joining the conversation, @jrbenny35 .

> So I think the main thing with Page Objects is better organization. A good test suite is easily readable and easy to follow, in order to aide with debugging and adding new tests.

> The "gain" would be organization and code quality. The POM model does help with reuse, yes, but its main goal is to represent pages in code for use with automation. Having one place where the representation of a page exists, that is easily defined, can prove useful when the actual HTML/CSS is changed. Since redash is going to get a full React update, we are actually ahead of the curve for once.

I'm all for better organization and easy to follow test suite (and code in general), but I don't think POM achieves this in this case. If to use this small example, all we got is:

1. Two files to look at instead one (the test code & the Page code).
2. Unfamiliar concept for existing Cypress users. Anyone coming from a different repository that uses Cypress will need to learn a new concept before being able to write tests.

It might feel nicer, because each thing has its place, but we just added unnecessary complexity.

> Having those actions on in their own .js files would keep the test suite clean and aide in readability.

I agree. But for them to be in their own files, they don't have to be POM objects -- see my nav bar example in the previous comment. Which leads me to -->

> Also, there is a way to create the objects on import so that there isn't a need to create a new object each time, but my JS isn't quite there yet.

Why add even more complexity (in the form of auto creating instances), when we don't need an object instance? We can just have a simple object with functions on it or even simpler: we can just have a Javascript module... :-)

> How true is this? In the python test suite I wrote for Mozilla's redash instance, we must login for each test. Now this is probably because selenium does not allow cookies, but is this true of cypress? We may actually need a beforeEach() that logs in for each test.

The beforeEach action in Cypress just POSTS to the login action instead of rendering the form, filling it out and posting. See [cy.login implementation](https://github.com/getredash/redash/blob/master/cypress/support/commands.js).

jrbenny35:
I guess I don't see where the complexity you are talking about with a POM comes from. It is just organization of all the "helper functions" to relate to their actual usage locations, in code.

Code like ```navbar.userDropdown.editProfile()``` seems easy to follow and is specific enough to aide in a clean test. Creating objects that do that would be very much the same as a POM, and then you lose the capability to expand to different parts of the page in a clean way.

I am sure you could just have a .js file with ```fill()```, ```clickThis()```, ```clickThat()```, etc, but there would be very little organization, unless you follow a POM like structure, where you have a ```nav_bar.js``` that has a ```openProfile()```. What if I wanted to fork a query? Maybe something like ```forkQuery()```, within the ```query.spec.js```. But there are at least 10 actions you could do on a query page, so then we would have to move them to a ```query_page.js``` for organization, and then you are practically following a Page Object Model.

The example used here:
```
function fillLogin(email, password) {
  cy.get('#inputEmail').type(email);
  cy.get('#inputPassword').type(password);
}
```
and then in the test:
```
it('shows message on failed login', () => {
    fillLogin('admin@redash.io', 'wrongpassword{enter}')
```
Where does this function work? On just any page that has these elements? ```fillLogin``` doesn't really lend to a verbose test. ```loginPage.fillLogin()``` - on the Login Page, fill the login. The reader of the test knows exactly where this login is going. And like I said, this could be done without a POM, but it will kinda be a POM as well, just not as strict.

> Why add even more complexity (in the form of auto creating instances), when we don't need an object instance? We can just have a simple object with functions on it or even simpler: we can just have a Javascript module... :-)

I am with you, maybe I misunderstood your comment.

arikfr:
> I am with you, maybe I misunderstood your comment.

Reading your last comment, it does feel to me that we agree on the principals, but still need to work out the _how_.

I feel that Cypress is different enough from Selenium that it allows for things to be done differently; in a simpler way. Just like we do things differently in dynamically typed languages than statically typed ones.

My suggestion is: let's avoid any abstractions until we have more tests (20? 50?). Once we have more tests, the right abstractions will present themselves. 

@gabrieldutra what do you think?

jrbenny35:
> My suggestion is: let's avoid any abstractions until we have more tests (20? 50?). Once we have more tests, the right abstractions will present themselves.

It will be A LOT of work bringing 20-50 tests to this model. I feel that this is the right time for it. If we start strong now it will aide in the long run. As someone who writes a lot of automated ui tests, you won't want to come back and re-do this.

gabrieldutra:
First of all, thanks @jrbenny35 and @arikfr for your comments :smile: 

Well, POM itself makes the test code looks more organized and good-looking and that's why I opened the PR. However, Arik's comments made me think and search a bit more about it.

Cypress' position about POM:
> The page object pattern isn’t actually anything “special”. If you’re coming from Selenium you may be accustomed to creating instances of classes, but this is completely unnecessary and irrelevant.

Cypress basically makes it in a way that POM isn't needed at all. So I really think that if we wait, this won't be something that will come out. Based on that, the gain of POM with Cypress is entirely about code organization.

About the extra code, it makes it harder to write a test for a new page, but easier to write when the Page Object already exists.

So what I think: This is an unneeded pattern that is up to us to put in the code, if we want it to be like this, we should start it now, so it would be an example to be followed. POM makes me happy to see how the code seems after the abstractions and on the other hand it makes me sad that we lose Cypress' essence of simplicity.

The only part of these changes that I really would like to move forward are the ones related to the `data-testid` selector, since it's the recommended approach and it's easy to see in frontend what's being tested.

Anyway, happy to have this discussion, since it makes me learn :grin:

jrbenny35:
> The page object pattern isn’t actually anything “special”. If you’re coming from Selenium you may be accustomed to creating instances of classes, but this is completely unnecessary and irrelevant.

See this kinda stuff is what I don't like. The page object pattern isn't just a "Selenium" thing, its a good software thing. Sure, teaching someone how to use the tool Cypress would mean they wouldn't need to specifically use a POM, but then you are just using the TOOL. What if you go to another project that doesn't use Cypress? Our contributors who want to write tests now only will learn how to write them using Cypress and its API. I think Cypress does some cool things but it could always change, and we could always switch tools.

And its API is not really much easier than the alternative. Just looking [here](https://docs.cypress.io/api/cypress-api/custom-commands.html#Best-Practices), that search function looks quite complex to just return the search bar and allow input. And I couldn't actually tell you what its doing without going through the API docs and reading each command. I don't see how that is easier for anyone.

arikfr:
@gabrieldutra ,

> About the extra code, it makes it harder to write a test for a new page, but easier to write when the Page Object already exists.

We can still organize our code better, it just that --

1. It's not necessarily to implement POM the way it's implemented in other environments.
2. We don't have to do it right away.

> The only parts of these changes that I really would like to move forward are the ones related to the data-testid selector, since it's the recommended approach and it's easy to see in frontend what's being tested.

I agree, let's keep this but wait with the other changes.

@jrbenny35 ,

> It will be A LOT of work bringing 20-50 tests to this model.

If we feel the need earlier than we can do it earlier. My point wasn't about the exact number but about getting to a point where we feel some abstraction is useful.

> Just looking here, that search function looks quite complex to just return the search bar and allow input. 

It does more than that: it also stubs out the HTTP request the UI will make to get the search results.

> but then you are just using the TOOL. What if you go to another project that doesn't use Cypress?

The point of adopting Cypress is to use it. We're not going to write layers of abstractions around it just in case that we _might_ some day switch to a different tool. Abstracting specifics is beneficial some times, but in this case it's not cost effective.

Just like we don't plan on replacing Redis or Postgres. Or if we ever replace Flask, we will rewrite the relevant code -- we don't write abstractions on top of it to "shield" us from Flask.

---

Anyway, I think we invested enough time in this discussion. :) And I'm still under the impression that we generally agree on the larger concepts. We have some different ideas on when to apply abstractions though.

gabrieldutra:
I think we can close this for now and come back as soon as we feel we need it. Meanwhile, @jrbenny35 you are welcome to share your experience and thoughts :grin:. I created #3155 with the purpose of adding some more tests.

Issue #2593, opened 2018-06-17
Scheduled query failures should be easier to find and understand


alison985:
Copies https://github.com/mozilla/redash/issues/309

A user has reported "It's hard to track scheduled queries failing, because not only do they not alert, but the errors never show!"

Is this within scope for implementation in getredash/redash?

arikfr:
Sure!
But let's discuss the implementation first.

Thanks.

alison985:
@arikfr Please let us know what you think of the following proposal:

Priority 1:

* A new page for a user to see scheduled query runs of that query. Let's call this /queries/#/runs. This would be a version of http://0.0.0.0:5000/admin/queries/tasks#done filtered to that specific query. It would only be visible to users with access to that query.

Priority 2:

* Email alerts on query failure. In the Refresh Schedule dialog we would add field for how often you want error emails. The option should be, at least for this first implementation, "every run". In the future, other options could be: "once an hour", "once a week", "once a month". Email would include a link back to /queries/#/runs and include the name of the query.

Priority 3:

* A notification alert at the top of pages to identify there is an issue to look at. On clicking it a list of links to the /queries/#/runs page(s) when there is a "failed" state for a query in the last month(?). (Is the last 30 days an appropriate window?)
  * I thought about since they last logged in but since we don't know what page they will be seeing and you have to do a database query to get that piece of information I don't want to add that overhead to every page load.
  * From @jezdez: "we can think about async ways to load that list of alerts though, and optimize on the backend by using Redis' Sorted Set data type for caching (for example)."

Priority 4:

* Add a red warning icon to the view source view of the query page if the query schedule has failed.
  * I think the easiest thing to implement would be if the query has failed ever show the red icon and a link to the /queries/#/runs page, however, I can imagine that would start getting ignored. I think, ideally, it would make sense for it to be for "this version" of the query but the query version feature doesn't exist upstream yet. I think failure in X amount of time is reasonable but I think the X amount of time would have to be based on the refresh schedule of the query. Perhaps "if failure within the last 2 cycles+ 1 second" then show the red icon.

arikfr:
A lot of the suggested functionality relies on the data we use in `/admin/queries/tasks`, but unfortunately this is sometimes unreliable, gets cleared out periodically (so you might not see the error runs on a busy system) and will probably get removed in future versions.

What do you think about:

1. Start with the email functionality, as this is the easiest to implement with the current data/tools we have.

[...]

PR #3065, opened 2018-11-11
 save query errors to queryResults object & make retrievable

alison985:
This addresses the first part of https://github.com/mozilla/redash/issues/309 that as discussed and finalized in https://github.com/getredash/redash/issues/2593

From @arikfr 
> Store explicit metadata on failed scheduled queries: last successful execution time, last error time, number of error since last successful execution, last error message (or maybe all the errors?). This will be cleared out when the query executes normally again. Once we have this extra information we can use it in the email.

Instead of making a new object I added to the QueryResult object.


arikfr:
Looks abandoned.

jezdez:
Oh we just talked about it in our Redash meeting and will update again..

jezdez:
@arikfr Ah, totally agreed, I reread your longer comment above and agree about the implementation method, so closing again. Apologies for the noise. I'll follow up as part of #2593.

PR #3068, opened 2018-11-13
Add Rockset query runner

igorcanadi:
Per REST API documented here: https://docs.rockset.com/rest/


arikfr:
Thanks, @igorcanadi. I'll take it from here :)

PR #3073, opened 2018-11-13
Remove only Redash containers

ariarijp:
Hi, team

`make clean` removes all of docker containers for now.

I think `make clean` should be removed only Redash containers.

Thanks,


Makefile
```diff
@@ -10,7 +10,7 @@ create_database:
 	docker-compose run server create_db
 
 clean:
-	docker ps -a -q | xargs docker kill;docker ps -a -q | xargs docker rm
+	docker ps -a | grep `basename "$$PWD"` | awk '{print $$1}' | xargs docker kill;docker ps -a | grep `basename "$$PWD"` | awk '{print $$1}' | xargs docker rm
```
arikfr: Can we use docker-compose commands instead here?

ariarijp: Oh, what a simple solution!

I gonna fix it.

Makefile
```diff
@@ -17,7 +17,7 @@ create_database:
 	docker-compose run server create_db
 
 clean:
-	docker ps -a -q | xargs docker kill;docker ps -a -q | xargs docker rm
+	docker-compose down && docker-compose rm
```
ariarijp: @arikfr I've fixed it!

arikfr:
Thanks!

PR #3077, opened 2018-11-14
UI for the feature flag of the share edit permissions feature

arikfr:
image


client/app/pages/settings/organization.html
```diff
@@ -10,6 +10,15 @@ <h3 class="m-t-0">General</h3>
                     <option ng-repeat="date_format in $ctrl.dateFormatList" value="{{date_format}}">{{date_format}}</option>
                 </select>
             </p>
+
+            <p>
+                <h4>Feature Flags</h4>
+
+                <label>
+                    <input name="input" type="checkbox" ng-model="$ctrl.settings.feature_show_permissions_control" ng-change="$ctrl.update('feature_show_permissions_control')"
+                        accesskey="tab"> Enable Sharing Editing Permissions
```
arikfr: @kocsmy do you understand what this means? (the title)

kocsmy: @arikfr somehow I missed this 6 days ago :/ Yes, I was just looking at it thinking what does this even mean and I think it's not really clean. I don't have anything better at the moment though.

kocsmy: Maybe **Enable sharing "Edit Permissions"** — and add a little description what does this mean and where. People might not be aware of the details if its so little information provided.

As a feature flag, this might be also under a dedicated section, maybe even hidden by default and called "Advanced". Like Gmail "feature flags":
image

arikfr: When we get to the point we have multiple feature flags, then this is definitely a good design. 👍 For now, we have only one that will be short lived and used by people who specifically ask for this functionality.

Once we see it's stable, we will enable it for everyone.

Issue #2978, opened 2018-10-17
Query's updated_at timestamp gets updated when you refresh the query data


arikfr:
It shouldn't be.

PR #3082, opened 2018-11-14
Prevent Query's updated_at from changing when it is linked to new query results

rauchy:
Solves #2978, but in a way that makes chewing street chalk feel pleasant.

Basically, the updated_at column was important to keep for all changes _other than_ the change of associated query data, and the only way I could find to override `TimestampMixin`'s onupdate logic was to assign a new value in that same transaction. This is the closest value I could find 🤦‍♂️

Please let me know if you can find a better solution.

### DIFF
--- a/redash/models.py
+++ b/redash/models.py
@@ -790,7 +790,8 @@ class QueryResult(db.Model, BelongsToOrgMixin):
             Query.data_source == data_source)
         for q in queries:
             q.latest_query_data = query_result
-            q.updated_at = q.updated_at + datetime.timedelta(milliseconds=1) # ugly hack to avoid Query's updated_at from changing in this specific scenario
+            # ugly hack to avoid Query's updated_at from changing in this specific scenario
+            q.updated_at = q.updated_at + datetime.timedelta(milliseconds=1)
             db.session.add(q)
         query_ids = [q.id for q in queries]
         logging.info("Updated %s queries with result (%s).", len(query_ids), query_hash)
--- a/redash/models.py
+++ b/redash/models.py
@@ -190,10 +190,17 @@ class MutableList(Mutable, list):
 
 
 class TimestampMixin(object):
-    updated_at = Column(db.DateTime(True), default=db.func.now(),
-                           onupdate=db.func.now(), nullable=False)
-    created_at = Column(db.DateTime(True), default=db.func.now(),
-                           nullable=False)
+    updated_at = Column(db.DateTime(True), default=db.func.now(), nullable=False)
+    created_at = Column(db.DateTime(True), default=db.func.now(), nullable=False)
+
+
+@listens_for(TimestampMixin, 'before_update', propagate=True)
+def timestamp_before_update(mapper, connection, target):
+    # Check if we really want to update the updated_at value
+    if hasattr(target, 'skip_updated_at'):
+        return
+
+    target.updated_at = db.func.now()
 
 
 class ChangeTrackingMixin(object):
@@ -790,8 +797,7 @@ class QueryResult(db.Model, BelongsToOrgMixin):
             Query.data_source == data_source)
         for q in queries:
             q.latest_query_data = query_result
-            # ugly hack to avoid Query's updated_at from changing in this specific scenario
-            q.updated_at = q.updated_at + datetime.timedelta(milliseconds=1)
+            q.skip_updated_at = True
             db.session.add(q)
         query_ids = [q.id for q in queries]
         logging.info("Updated %s queries with result (%s).", len(query_ids), query_hash)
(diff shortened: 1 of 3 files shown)


tests/models/test_query_results.py
```diff
@@ -0,0 +1,72 @@
...
+
+from tests import BaseTestCase
+
+from redash import models
+from redash.utils import utcnow
+
+
+class QueryResultTest(BaseTestCase):
+    def setUp(self):
+        super(QueryResultTest, self).setUp()
```
jezdez: This method isn't needed.

redash/models.py
```diff
@@ -790,6 +790,8 @@ def store_result(cls, org, data_source, query_hash, query, data, run_time, retri
             Query.data_source == data_source)
         for q in queries:
             q.latest_query_data = query_result
+            # ugly hack to avoid Query's updated_at from changing in this specific scenario
+            q.updated_at = q.updated_at + datetime.timedelta(milliseconds=1)
```
jezdez: Let's not do that and instead fix the `TimestampMixin` class to optionally support not updating the `updated_at` value.

To do that we first have to change the mixin to stop using `onupdate` since it doesn't provide as with the hooks AFAIK to make educated guess, e.g. we don't have access to the ORM instance.

Instead we can use the [`before_update` ORM event](https://docs.sqlalchemy.org/en/latest/orm/events.html#sqlalchemy.orm.events.MapperEvents.before_update) to conditionally update the `updated_at` column:

```python
from sqlalchemy.event import listens_for

class TimestampMixin(object):
    updated_at = Column(db.DateTime(True), default=utcnow, nullable=False)
    created_at = Column(db.DateTime(True), default=utcnow, nullable=False)

@listens_for(TimestampMixin, 'before_update', propagate=True)
def timestamp_before_update(mapper, connection, target):
    # Check if we really want to update the updated_at value
    if hasattr(target.skip_updated_at):
        return
    target.updated_at = utcnow()
```

Then inside the `store_result` loop you can simply do:

```python
q.skip_updated_at = True
```

tests/models/test_query_results.py
```diff
@@ -0,0 +1,72 @@
...
+
+        self.assertEqual(found_query_result.id, qr.id)
+
+    def test_store_result_does_not_modify_query_update_at(self):
+        original_updated_at = utcnow() - datetime.timedelta(hours=1)
+        query = self.factory.create_query(updated_at=original_updated_at)
+
+        models.QueryResult.store_result(query.org_id, query.data_source, query.query_hash, query.query_text, "", 0, utcnow())
+
+        assert original_updated_at.strftime("%X") == query.updated_at.strftime("%X")
```
jezdez: Please use `self.assertEqual` here and compare `original_updated_at` with `query.updated_at` directly, without formatting it as a string.

redash/models.py
jezdez: @washort Just to be sure, do you know a reason not to use the event system for the timestamp mixin by any chance?

tests/models/test_query_results.py
arikfr: Isn't `assert` the way you assert when using py.test?

redash/models.py
rauchy: Thanks @jezdez, looks a lot better 👍

tests/models/test_query_results.py
jezdez: Yeah, but as long as we use a unittest base class it's better form to consistently use its assert methods to have a common assertion behavior and error reporting instead of mixing both. pytest transparently supports the unittest assert methods.

arikfr: Wouldn't it be more future proof to use `assert` in new code, even in unittest inheriting classes? This way when we're ready, we can change the base class and don't have to rewrite this test.

rauchy: I'd say that even for the sake of consistency within this module, it makes sense to stick to a single assert. Otherwise it just makes you wonder - what's different about _this_ test?

When we refactor, we'll probably do it as part of a massive search & replace, so one assert doesn't feel like it'll make things easier in the future.

jezdez: > Wouldn't it be more future proof to use `assert` in new code, even in unittest inheriting classes? This way when we're ready, we can change the base class and don't have to rewrite this test.

The compatibility code in pytest for unittest classes is an added layer of abstraction that makes debugging harder when being used side by side with test errors that didn't have the shim applied. In other words, the consistency of using one assertion method (in a module or class) is more important than the decision which to use. When (if?) we move away from unittest classes we can do as @rauchy says, do the conversion in one swap.

arikfr: OK, OK. 🏳️ :-)

I just find the camelCase assert function so ugly that I was happy to stop using them. But of course your arguments make sense (more than my dislike of the camelCase asserts).

jezdez: Oh I believe you and I can assure you the feeling about camelCase asserts is mutual. 😁

PR #3083, opened 2018-11-15
ClickHouse: added WITH TOTALS option support

denisov-vlad:
Clickhouse allows to add WITH TOTALS parameter to show aggregated values for all dataset.

https://clickhouse.yandex/docs/en/query_language/select/#with-totals-modifier

By default it returns default values which is not suitable for BI:
empty string for string, 0 for integers (in case of CH booleans are integers too)

Query runner converts strings to `Total` value, other types to `None`.


arikfr:
Looks good. I just wonder: maybe it will be more useful to add another column with the total value?

arikfr:
Thanks!

PR #3085, opened 2018-11-15
Switch BigQuery Standard SQL as the default

### LINKED ISSUE #2770: Switch to Standard SQL for BigQuery as the default
I think it's time to switch to use Standard SQL as the default syntax for BigQuery.

Also, we should show table names in the schema browser in the Standard SQL format instead of the legacy one (`.` as separator instead of `:`).

kyoshidajp:
Fix #2770

According to [the document](https://cloud.google.com/bigquery/docs/reference/legacy-sql),  project_name is optional in SQL. Moreover,  schema browser becomes simpler.

Because of that, deleted project_name in schema browser.

## Screenshot

Add Data Source

| Before | After |
| ------ | ----- |
| ds_before |  ds_after |

Schema Browser

| Before | After |
| ------ | ----- |
| table_name1 | table_name2 |

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -110,7 +110,8 @@ class BigQuery(BaseQueryRunner):
                 },
                 'useStandardSql': {
                     "type": "boolean",
-                    'title': "Use Standard SQL (Beta)",
+                    'title': "Use Standard SQL",
+                    "default": True,
                 },
                 'location': {
                     "type": "string",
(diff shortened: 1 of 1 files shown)


kyoshidajp:
@arikfr Thanks. 

I think that it will not be applied to existing BigQuery data sources. My change only affects when users add new data sources.

The value of `useStandardSql` of existing data sources is two patterns (`True`, `False`). Or an old setting which doesn't have `useStandardSql` key expects running as Legacy SQL.

In this case, `self.configuration.get('useStandardSql', False)` returns `True`, `False` and `False`. I think no need to migrate existing data sources and to change getting value of `useStandardSql`.

| value of `useStandardSql` | result of `self.configuration.get('useStandardSql', False)` |
| -------------------------- | -------------------------------------------------------- |
| `True` | `True` |
| `False` | `False` |
| (doesn't  have key) | `False` |

arikfr:
You were right, it works as needed. Thanks!

PR #3088, opened 2018-11-17
Add get_current_user() into Python Data Source

### LINKED ISSUE #1646: Access current user id/details from Python data source
First of all, thanks to the authors and contributors for a very interesting and promising tool.  
### Question:
Is it possible to access current user id/details from Python data source?
I would like to be able to control access to the viewed data at row level without a need of creating and maintaining multiple (in my case about 80) data sources and dashboards. My desired scenario:
    1. Store A logs into their account and accesses dashboard with KPI/Metrics.
    2. Dashboard calls Python data source(s).
    3. Python code loads data from a db table with all stores, identifies current user and filters out all records where store != Store A.
    4. Dashboard is dynamically loaded with entries for Store A only.

Also if there are any other ways to achieve the above, please advise.   

### Technical details:
* Redash Version: 0.12.0

kyoshidajp:
Fix #1646.

Get current user detail. For example:

```python
user = get_current_user()

# id
user["id"]

# email
user["email"]
```

### DIFF
--- a/redash/query_runner/python.py
+++ b/redash/query_runner/python.py
@@ -211,10 +211,15 @@ class Python(BaseQueryRunner):
 
         return json_loads(query.latest_query_data.data)
 
+    def get_current_user(self):
+        return self._current_user.to_dict()
+
     def test_connection(self):
         pass
 
     def run_query(self, query, user):
+        self._current_user = user
+
         try:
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks!

PR #3089, opened 2018-11-17
Open new tab when forking a query

### LINKED ISSUE #328: Allow to "Open in new tab" when forking a query

kyoshidajp:
Fix #328.

as_new_tab

I checked following environments.

- macOS High Sierra(10.13.6)
   - Chrome Version 70.0.3538.102 (Official Build) (64-bit)
   - Safari Version 12.0.1 (13606.2.104.1.2)

### DIFF
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -53,10 +53,7 @@
               </button>
               <ul class="dropdown-menu pull-right" uib-dropdown-menu>
                 <li ng-class="{'disabled': query.id === undefined || !canForkQuery() }">
-                  <a ng-click="duplicateQuery()"> Fork</a>
-                </li>
-                <li ng-class="{'disabled': query.id === undefined || !canForkQuery() }">
-                  <a ng-click="duplicateQuery(true)"> Fork (as new tab)</a>
+                  <a ng-click="duplicateQuery()"> Fork <i class="fa fa-external-link"></i></a>
                 </li>
                 <li class="divider"></li>
                 <li ng-if="!query.is_archived && query.id != undefined && (isQueryOwner || currentUser.hasPermission('admin'))">
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -55,6 +55,9 @@
                 <li ng-class="{'disabled': query.id === undefined || !canForkQuery() }">
                   <a ng-click="duplicateQuery()"> Fork</a>
                 </li>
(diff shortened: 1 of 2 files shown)


arikfr:
Cool! How about instead of adding another menu option, we use modifier keys (like Cmd + click to fork in a new tab)? We can change the menu item title when modifier key is pressed to assist with discovery.

@kocsmy what do you think?

Jakdaw:
Isn't that a function in the browser anyway? With Chrome on Linux I middle-class for that purpose.

arikfr:
@Jakdaw yes, except that this is not a real link but a button. But your assumption that this is a link and the reference to the default behavior, just reassures my idea of implementing it the same a browser does a link :)

kocsmy:
Although I understand why this is useful, I don't think it's a good idea to have another option here. It should be either a proper link and middle mouse button should open in a new tab (as default browser interaction) or we could introduce a modifier button that opens in a new tab. Alternatively we could add an [icon](https://fontawesome.com/icons/external-link-alt?style=solid) next to the link.

kyoshidajp:
Sorry, I'm late.

When forking, I usually compare it with the original query. If we don't add a new line for opening a new tab option, I want to change the operation of the fork link so that it always opens as a new tab. It's very simple.

I also agree to add an icon.

jezdez:
FWIW the native browser functionality of using the modifier key to open a link in a new tab also works for from submissions via a button. As such, the button that is used in the menu is definitely breaking the native behavior which makes this more like a bugfix than a new feature. So it's sensible to simply restore the ability use the modifier key.

arikfr:
> Sorry, I'm late.

Nothing to be sorry about. When there is no schedule, one can not be late :-)

> If we don't add a new line for opening a new tab option, I want to change the operation of the fork link so that it always opens as a new tab. 

If you prefer not to get into implementing modifier key support, I'm fine with this result as well (along with an icon). 

Thanks!

kyoshidajp:
Sure! I will do tomorrow.

kocsmy:
> I want to change the operation of the fork link so that it always opens as a new tab. It's very simple.

I think that makes sense for me, GDocs works the same when you use "Make a copy", I think is kind of natural behaviour to open a new tab when you are making a copy/forking.

kyoshidajp:
I did. How about this?

external-link

arikfr:
Looks good to me 👌

arikfr:
Thanks!

PR #3091, opened 2018-11-18
Autocomplete toggle improvements

### LINKED ISSUE #2984: Improve Autocomplete toggle
Currently it's not very clear if the toggle is active or not. We should improve its visual representation.

Also:

- [ ] The tooltip should represent the current state (say if it's active or not)
- [ ] Store the last state of the toggle
- [ ] If Live autocomplete is disabled because of number of tokens, the toggle should be disabled

arikfr:
- [x] Refactor to its own component.
- [x] Show state in tooltip (enabled/disabled).
- [x] Disable the toggle if autocomplete is not possible (too many tokens) & show hint that you can still trigger it with Ctrl+Space.
- [x] Store the last state of the toggle in localStorage.

Closes #2984.


arikfr:
Two alternatives to implementing the toggle in a way that it's clear what is the current state:

1. Button with two icons (one for enabled and one for disabled): https://cl.ly/5611e330efdd
2. Just icon without the button chrome: https://cl.ly/4120d07c29fd

@kocsmy what do you think?

arikfr:
@kocsmy figured this will be your preference and hence what I implemented :)

kocsmy:
Nice :)

Issue #3067, opened 2018-11-12
Autocomplete V2


arikfr:
We have several issues with our autocomplete:

1. 
It has performance issues, hence we disable it for schemas with too many tokens. This happens because we pass all the tokens available to Ace for it to figure it out. It's possible we can utilize their API better and only return relevant suggestions.

2. 
It's not aware of the text context, i.e. if you started typing a word and then continue, it ignores existing text (#2329).

3. 
No support for aliases: 

```sql
SELECT q.id
FROM queries q
```

In the above example it will not be aware that `q` is actually `queries` and won't suggest columns from it.

4. 
It's not aware of the code context, i.e. if I'm currently typing `SELECT ...` it makes sense to autocomplete almost anything, but if I'm typing `FROM ...` it should autocomplete table names or at least give them priority.

5.
Some people get annoyed by the autocomplete live behavior, but I'm not sure there is much to do about it. Maybe when the suggestions become more relevant it will be less of annoyance.

---

To implement 3 & 4, we need to understand the query. For this we can use this [Javascript SQL parser](https://github.com/codeschool/sqlite-parser).

We don't have to land it all together, but rather incremental improvements with each step.

AravindGopala:
Any update on this?

ivanzhukov07:
Confirmed on v26.3.0. Still affects the latest stable release.

Issue #2329, opened 2018-02-20
Improve Autocomplete Functionality


keanerobinson:
### Issue Summary

- This issue occurs on Chrome on High Sierra 10.13.3
- I can't currently make use of the predictive dropdown options in the query editor. The dropdown options need to fully replace the line being typed (I'll reproduce the steps in the following section to elaborate).

### Steps to Reproduce

1. Typing in a table name

![screen shot 2018-02-20 at 15 43 43](https://user-images.githubusercontent.com/22820802/36433703-0f67e9f4-1655-11e8-915e-7b8b5c8d4af6.png)

- I'd like to use one of the dropdown options to speed up the time it takes to write my query
- Let's assume that I want to select the first option in the dropdown and that I either hit enter or click on the option

2. Result after hitting enter / clicking

![screen shot 2018-02-20 at 15 47 27](https://user-images.githubusercontent.com/22820802/36433836-6c3b878a-1655-11e8-91df-1759335f264c.png)

- Obviously this is not a valid table name as there's a double reference to the schema
- The dropdown works fine when typing in field names when there is no join involved but in most cases I need to type the whole query manually / copy & paste from the list of table names.

PR #3092, opened 2018-11-18
Add dot behavior to autocomplete

gabrieldutra:
## Description
A start on #3067 - Issues 1 and 2.
This changes the dot behavior when writing a query and start separating keyword types as Table Keyword, Column Keyword or Table-column Keyword

With that we separate 2 situations of autocomplete, instead of rendering the entire schema:
1. **Default** - Show table and distinct column names;
2. **tableName.** - Show specific Table columns.

This will probably improve the performance and be useful when we introduce aliases or query context.

### Autocomplete not aware of the context (#3067 and #2329)
The dot by default breaks autocomplete words, so I changed its Regexp to include it.

### Repeated suggestions when navigating through query pages
repeated-suggestions
This seems to be something that could be affecting the performance.
Whenever would one navigate to a new query page, another completer would be added to **langTools**. I've changed `addCompleter` to `setCompleters` to fix this. 

### Before
autocomplete-dot-before
### After
autocomplete-dot

## To do
- [x] Start it
- [x] Test changes with a large schema and validate the performance
- [x] Refactor `QueryEditor` constructor
- Refactor other `QueryEditor` methods? :thinking: 

We'll probably have some conflicts with #3091, but nothing huge.

Let me know your thoughts :grin:


client/app/components/QueryEditor.jsx
```diff
@@ -81,92 +63,125 @@ class QueryEditor extends React.Component {
...
+
+        if (prefix.length === 0 || tableKeywords.length === 0) {
           callback(null, []);
           return;
         }
-        callback(null, this.state.keywords);
-      },
-    });
 
-    this.onLoad = (editor) => {
```
gabrieldutra: Was there a reason for those methods to be inside QueryEditor's constructor?

arikfr: 🤷‍♂️ 

@washort ?

arikfr:
👍

PR #3093, opened 2018-11-19
Remove Mustache templating from frontend

rauchy:
Remove Mustache templating from frontend and send all parameters to the API (as POST data)

### DIFF
--- a/client/app/services/query-result.js
+++ b/client/app/services/query-result.js
@@ -523,10 +523,16 @@ function QueryResultService($resource, $timeout, $q, QueryResultError) {
       return `${queryName.replace(/ /g, '_') + moment(this.getUpdatedAt()).format('_YYYY_MM_DD')}.${fileType}`;
     }
 
-    static get(dataSourceId, query, maxAge, queryId) {
+    static get(dataSourceId, parameters, query, maxAge, queryId) {
       const queryResult = new QueryResult();
 
-      const params = { data_source_id: dataSourceId, query, max_age: maxAge };
+      const params = {
+        data_source_id: dataSourceId,
+        parameters,
+        query,
+        max_age: maxAge,
+      };
+
       if (queryId !== undefined) {
         params.query_id = queryId;
       }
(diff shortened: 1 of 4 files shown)


redash/handlers/query_results.py
```diff
@@ -131,7 +132,7 @@ def post(self):
             'object_type': 'data_source',
             'query': query
```
arikfr: We should update the recorded event with the parameters if there are parameters.

arikfr: While we're at it, we should record the `query_id` as well.

client/app/services/query-result.js
```diff
@@ -523,10 +523,16 @@ function QueryResultService($resource, $timeout, $q, QueryResultError) {
       return `${queryName.replace(/ /g, '_') + moment(this.getUpdatedAt()).format('_YYYY_MM_DD')}.${fileType}`;
     }
 
-    static get(dataSourceId, query, maxAge, queryId) {
+    static get(dataSourceId, parameters, query, maxAge, queryId) {
```
arikfr: I'm curious, why have the parameters before the query?

rauchy: Yeah, found that peculiar ordinality [here](https://github.com/getredash/redash/blob/master/redash/handlers/query_results.py#L30) and thought I'd stay consistent.

arikfr: No reason to be consistent with that abomination. Once you're done , we will be able to finally get rid of it.

redash/handlers/query_results.py
```diff
@@ -5,22 +5,42 @@
...
+        text = query.text
+    except SQLInjectionError:
+        record_event({
+            'action': 'sql_injection',
+            'object_type': 'query',
+            'query': template,
+            'parameters': parameters,
+            'timestamp': time.time(),
+            'org_id': data_source.org_id
+        })
```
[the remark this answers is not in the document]
rauchy: It's just that the recorded event for query execution happens on a different level (in the handler body) and hooking that up would require refactoring the code for something quite temporary. I would rather just replace [this call](https://github.com/getredash/redash/pull/3093/files#diff-f734d32676ec526e17e5ae92d41b5787R30) with proper exception handling once we're done tracking.

Issue #3052, opened 2018-11-07
Tags autocomplete is broken


arikfr:
In the "Add Tag" dialog, when you start typing a prefix of an existing tag it doesn't show it.

PR #3094, opened 2018-11-19
Tags autocomplete: some tags not shown on search

kravets-levko:
Fixes getredash/redash#3052

### DIFF
--- a/client/app/components/tags-control/modal-template.html
+++ b/client/app/components/tags-control/modal-template.html
@@ -3,8 +3,7 @@
   <h4 class="modal-title">Add/Edit Tags</h4>
 </div>
 <div class="modal-body">
-  <!-- &#8203; is zero-width space - seems regular whitespace breaks components built-in search -->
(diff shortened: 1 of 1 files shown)


client/app/components/tags-control/modal-template.html
```diff
@@ -3,7 +3,8 @@
   <h4 class="modal-title">Add/Edit Tags</h4>
 </div>
 <div class="modal-body">
-  <ui-select multiple append-to-body tagging tagging-label=" "
+  <!-- &#8203; is zero-width space - seems regular whitespace breaks components built-in search -->
+  <ui-select multiple append-to-body tagging tagging-label="&#8203;"
```
arikfr: Let's use "(Create new tag)" here instead.

kravets-levko: Done. Just added `(create new tag)` (with small first letter - IMHO it looks a bit better):
image

arikfr:
Thanks!

PR #3097, opened 2018-11-19
Fix collection fields retreival bug when Views are present in MONGO DB

jodevsa:
Hi,

Retrieval of  collection fields will fail  when a collection of type View is present. This is because $natural sort is being used for all collection types although isn't supported in Views ([link](https://docs.mongodb.com/manual/reference/operator/meta/natural/))

This commit will add functionality to detect  Views and handle them accordingly (first 2 documents are considered the sample)

Thanks.


redash/query_runner/mongodb.py
```diff
@@ -174,6 +174,13 @@ def _merge_property_names(self, columns, document):
               if property not in columns:
                   columns.append(property)
 
+    def _is_collection_a_view(self, db, collection_name):
+        try:
+            db.command('collstats', collection_name)
+            return True
```
arikfr: When collstats succeeded doesn't it mean this is a Collection rather than a View?

jodevsa: Your absolutely right, forgot to switch them before committing

redash/query_runner/mongodb.py
```diff
@@ -184,19 +191,20 @@ def _get_collection_fields(self, db, collection_name):
...
-        last_document = None
-
-        for d in db[collection_name].find().sort([("$natural", 1)]).limit(1):
-            first_document = d
-
-        for d in db[collection_name].find().sort([("$natural", -1)]).limit(1):
-            last_document = d
+        collection_is_a_view = self._is_collection_a_view(db, collection_name)
+        documents_sample = []
+        if(collection_is_a_view):
```
arikfr: You don't need parentheses around the `if` statement:

```suggestion
        if collection_is_a_view:
```

jodevsa: Cool!

jodevsa:
@akariv  Not sure why the build failed this time, it's working from my side.

arikfr:
I'm not sure either. I triggered another execution.

arikfr:
Then I'll trigger another one. 🤷‍♂️

jodevsa:
finally :1st_place_medal:

arikfr:
Thanks!

PR #3099, opened 2018-11-20
force angular to update query editor properly (re #3098)

### LINKED ISSUE #3098: Any state change in QueryEditor seems to clear out the query
1. Write a query.
2. Click on Format Query or on autocomplete toggle.
3. The query text disappears.

washort:
Angular needs `$apply` called when event handlers (like the `onChange` in QueryEditor) modify stuff in scope. This results in the React component getting updated properly when query text changes.

Closes #3098.


arikfr:
Thanks.

PR #3105, opened 2018-11-20
Display frontend version

### LINKED ISSUE #3104: Set a version for the frontend code
In the footer we show the backend version, which is useful for verifying that you're using the version you expect to. But the frontend code should have a version too to allow similar verification (and in the future notification for users to reload for newer version).

We can use [git-revision-webpack-plugin](https://www.npmjs.com/package/git-revision-webpack-plugin) or similar.

rauchy:
Closes #3104 .

### DIFF
--- a/client/app/components/Footer.jsx
+++ b/client/app/components/Footer.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import { react2angular } from 'react2angular';
 
-import frontendVersion from '@/version.json';
+import frontendVersion from '../version.json';
 
 export function Footer({ clientConfig, currentUser }) {
   const backendVersion = clientConfig.version;
--- a/client/app/components/Footer.test.js
+++ b/client/app/components/Footer.test.js
@@ -10,7 +10,6 @@ test('Footer renders', () => {
   const currentUser = {
     isAdmin: true,
   };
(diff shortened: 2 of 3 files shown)


client/app/components/Footer.jsx
```diff
@@ -19,7 +19,7 @@ export function Footer({ clientConfig, currentUser }) {
 
   return (
     <div id="footer">
-      <a href="https://redash.io">Redash</a> {version}
+      <a href="https://redash.io">Redash</a> {version} ({COMMIT})
```
arikfr: To avoid confusion I would name them `backendVersion` and `frontendVersion`.

arikfr:
You probably need to update the jest snapshot.

arikfr:
webpack fails to run inside a container, because: `git describe --always` fails ("not a git repo") :-( Which makes sense... it's not a git repo when docker build runs webpack.

I guess we need an alternative approach here that won't involve using `git` directly.

rauchy:
@arikfr what do you think about this approach: when containers are created, write the latest SHA at creation into a VERSION file and read from it during webpack's build. If the file is not found, revert to using `git-revision-webpack-plugin` which will handle anyone cloning the repo directly.

If that makes sense, I just need a pointer to the scripts that create the different containers...

PR #3106, opened 2018-11-20
Fix: registerAll fails after minification

arikfr:
[This change[(https://github.com/getredash/redash/pull/3091/commits/e78912d6ef92b6d93c3cfea459ee9532308c60c0#diff-102261740816498432e605046c600993R71) in #3091 breaks after minification... I tried changing the Uglify settings to not mangle `init` functions, but this resulted in other issues.

So ended up with this solution: instead of filter by name, we look for `init` flag on the function name.


arikfr:
It might not be the best solution, but merging this to stop master branch from failing. We can revisit this.

PR #3109, opened 2018-11-21
Protect against SQL injections by using tree comparisons

rauchy:
This is part of moving parameter templating to the backend, but is pushed a separate PR just to get focused comments. This introduces an `SQLQuery` class which will handle templating and injection detection on templated queries. This PR doesn't integrate `SQLQuery` yet.

TODO:

- [x] Use parameter for column name or order clause.
- [x] Adding a whole different query. Like the union test just without UNION.
- [x] Move to separate module
- [x] Add some more test cases
- [x] Serialize `SQLQuery` after it is templated (throw on missing params?)


tests/test_utils.py
```diff
@@ -61,3 +61,35 @@ def test_skips_nones(self):
...
+        query = SQLQuery("SELECT * FROM users WHERE userid={{userid}}")
+        query.apply({"userid": "22 UNION SELECT body, results, 1 FROM reports"})
+
+        self.assertFalse(query.is_safe())
+
+    def test_marks_comment_attacks_as_not_safe(self):
+        query = SQLQuery("SELECT * FROM users WHERE username='{{username}}' AND password='{{password}}'")
+        query.apply({"username": "admin' --"})
+
+        self.assertFalse(query.is_safe())
```
arikfr: Additional cases to consider:

1. Use parameter for column name or order clause.
2. Adding a whole different query. Like the union test just without UNION.

Jakdaw: Aren't there Data Sources that aren't using SQL? Might sqlparse not fail for those?

Jakdaw:
Aren't there Data Sources that aren't using SQL? Might sqlparse not fail in those circumstances?

rauchy:
@jakdaw definitely, and the plan is to use this class only for the query runners that support SQL.

redash/utils/sql_query.py
```diff
@@ -0,0 +1,40 @@
...
+    def apply(self, parameters):
+        self.query = mustache_render(self.template, parameters)
+        return self
+
+    def is_safe(self):
+        template_tree = sqlparse.parse(self.template)
+        query_tree = sqlparse.parse(self.query)
+        return self._same_type(template_tree, query_tree)
+
+    def _same_type(self, a, b):
```
jezdez: No need for those extra parantheses in all the `if`s and `elif`s here.

redash/utils/sql_query.py
```diff
@@ -0,0 +1,40 @@
...
+            children_are_same = [self._same_type(child_a, child_b) for (child_a, child_b) in zip(a, b)]
+            return len(a) == len(b) and all(children_are_same)
+        elif (hasattr(a, 'tokens')):
+            template_tokens = filter(lambda t: t.ttype is not sqlparse.tokens.Error, a.tokens)
+            query_tokens = b.tokens
+            return self._same_type(template_tokens, query_tokens)
+        else:
+            return True
+
+    def text(self):
```
jezdez: This would be useful to be a property instead of a method.

jezdez: In case there is reason to use this multiple times, maybe even make this a [cached property](https://pypi.org/project/cached-property/).

redash/utils/sql_query.py
```diff
@@ -0,0 +1,45 @@
...
+from cached_property import cached_property
+
+
+class SQLQuery(object):
+    def __init__(self, template):
+        self.template = template
+        self.query = template
+
+    def apply(self, parameters):
+        if self.__dict__.has_key('text'):
```
arikfr:
```suggestion
        if 'text' in self.__dict__:
```

redash/utils/sql_query.py
```diff
@@ -0,0 +1,45 @@
...
+            children_are_same = [self._same_type(child_a, child_b) for (child_a, child_b) in zip(a, b)]
+            return len(a) == len(b) and all(children_are_same)
+        elif hasattr(a, 'tokens'):
+            template_tokens = filter(lambda t: t.ttype is not sqlparse.tokens.Error, a.tokens)
+            query_tokens = b.tokens
+            return self._same_type(template_tokens, query_tokens)
+        else:
+            return True
+
+    @cached_property
```
arikfr: This feels like a premature optimization at this point. Is the tree comparison such a "heavy" operation?

redash/utils/sql_query.py
```diff
@@ -0,0 +1,45 @@
...
+        self.query = template
+
+    def apply(self, parameters):
+        if self.__dict__.has_key('text'):
+            del self.__dict__['text']
+
+        self.query = mustache_render(self.template, parameters)
+        return self
+
+    def is_safe(self):
```
arikfr: `is_safe` should perform a few additional tests before doing the SQL tree comparison:

1. Test that the passed value is of the type the parameter expects (in case the parameter type isn't a string). 
2. If it's a dropdown (query based or static) parameter, then compare with the list of allowed values.

redash/utils/sql_query.py
rauchy: Already did this on a local commit and haven't pushed yet, so I'm politely ignoring.

redash/utils/sql_query.py
rauchy: yeah, you're right...

redash/utils/sql_query.py
rauchy: I tend to think that this should be in a different module. These checks are more related to the validity and compatibility of the provided parameters, and that's a different type of knowledge. I'd prefer to address this outside the current `SQLQuery` class (and maybe tree validations should be moved out as well, and `SQLQuery` should integrate these two distinct responsibilities - that's something that should be considered outside the scope of this PR, IMO)

arikfr: The implementation of this logic should be outside this class, as it will be reused in "JSONQuery" and others. Whether this class should call it some outside caller, is probably something we can decide when we get to integrate it with the actual codebase.

arikfr:
🕺

Issue #1742, opened 2017-04-28
Wrong column value in bar chart


denisov-vlad:
I think this bar will show the error: http://demo.redash.io/queries/4439/source#6248

I can't see real column values when using stacking.

<img width="260" alt="2017-04-28 10 38 30" src="https://cloud.githubusercontent.com/assets/7091907/25519053/0069ad34-2bff-11e7-89d6-fc13a6f8e35d.png">

arikfr:
Maybe it just a bad example, but in the example you sent you have multiple values for the same X value and hence the bad tooltip.

denisov-vlad:
@arikfr I think that it should me summed in graph. 

For example, I want to make some charts in one query: daily, weekly and monthly and it will be good to sum them in graph because it's better to create 3 different queries.

arikfr:
You can't always sum the values. The query in the demo is:

```sql
select created_at::date as day, date_trunc('week', created_at) as week, count(*)
from dashboards
where created_at >= current_date - interval '1 month'
group by day, week
```

But what if we wanted to know how many unique users we had each day? Then the query will be:
```sql
select created_at::date as day, date_trunc('week', created_at) as week, count(distinct user_id)
from dashboards
where created_at >= current_date - interval '1 month'
group by day, week
```

But you can't get the weekly unique number of users by just adding up the daily numbers, because if the same users visited each day, then the weekly number is the same as the daily one.

denisov-vlad:
@arikfr 
It was summed in a bar in `master` version but only the value was wrong.

The issue you provided is not visualization problem. If I want to count unique values I'll make a join with values `dau, wau`, for example, and create visualization without stacking.

But if I want to stack all values in a chart I understand that stacking is mostly summing.

arikfr:
> It was summed in a bar in master version but only the value was wrong.

Seems like it wasn't, see this: http://demo.redash.io/queries/4617/source#6595 or did I misunderstand you?

> But if I want to stack all values in a chart I understand that stacking is mostly summing.

I agree. But there is a difference between stacking values where for each series there is a single value vs. stacking where for each series there are multiple values. The way I see it: visualizations need to visualize the data, not to change it. Summing the values inside a series (not across multiple series) is changing the data it represents.

Also note that Plotly maintains the same behavior: it expects a single value for each series.

Having said all that, I think that we should either show an error message to the user in such case or implement something along the lines what you suggest.

denisov-vlad:
[...]

PR #3115, opened 2018-11-25
add warning_three_column_groupby

alison985:
See https://github.com/mozilla/redash/issues/555 - this is a port upstream of a Mozilla feature.

This PR will
* display a warning and not let a visualization be saved if there are 3 columns in the resultset of the query and there is not a group by field chosen in the visualization configuration.
* make the warning disappear and enable saving once a group by field is choosen.

Error state:
screenshot 2018-11-25 15 35 02

Post-error state:
screenshot 2018-11-25 15 36 25


alison985:
@jezdez I can't add you as a reviewer, but a heads up that this is ready to review.

arikfr:
Here's a valid chart this doesn't let me save:

image

(https://deploy-preview-3115--redash-preview.netlify.com/queries/22/source#32)

The data is: month, # of issues, # of pull requests. There is no problem of plotting each one of them on its own or both of them together. 

The issue you seem to try and overcome is when there are multiple Y values for a single X value (#1742). But this needs to be checked after the data is arranged into series. Also following the discussion in #1742 and in [the forum](https://discuss.redash.io/t/sum-average-x-axis-data/1788), I inclining towards conceding to applying an aggregation in such cases...

client/app/visualizations/edit-visualization-dialog.js
```diff
@@ -32,6 +38,24 @@ const EditVisualizationDialog = {
       this.visualization = this.newVisualization();
     }
 
+    this.has3plusColumnsFunction = () => {
+      let has3plusColumns = false;
+      if ((JSON.stringify(this.visualization.options.columnMapping).match(/,/g) || []).length > 1) {
```
arikfr: Why this and not [_.size](https://lodash.com/docs/4.17.11#size)?

client/app/visualizations/edit-visualization-dialog.js
```diff
@@ -32,6 +38,24 @@ const EditVisualizationDialog = {
...
+        has3plusColumns = true;
+      }
+      return has3plusColumns;
+    };
+
+    this.disableSubmit = () => {
+      if (this.has3plusColumnsFunction() &&
+        JSON.stringify(this.visualization.options.columnMapping).includes('unused') &&
```
arikfr: I have a query where one of the columns named "unused"... so this is not so reliable.

But this can be tested with `values(this.visualization.options.columnMapping).includes('unused')`.

(`values` is imported from `lodash`)

alison985: Thanks!

client/app/visualizations/edit-visualization-dialog.js
alison985: Great point - a way easier read. My brain just got stuck on it being a string.

alison985:
@arikfr I have to say I personally agree with the "applying an aggregation in such cases" route. It's a more elegant solution for the user as long as they get information about why they can't disable the aggregation if they attempt to disable it. Therefore I am not going to pursue the bug you found about multiple measures.

@jezdez FYI.

arikfr:
As this introduces more issues than solving anything, I'm closing this. I'm happy to discuss a potential solution along the lines of what I mentioned in the end of my previous comment.

Closing this now.

Issue #3108, opened 2018-11-21
Tags count for dashboards is wrong


arikfr:
![image](https://user-images.githubusercontent.com/71468/48824355-b395f400-ed6c-11e8-8fcb-7a04d6f2c2a4.png)

jezdez:
Huh!

jezdez:
@arikfr I'll take a look.

jezdez:
Hmm, I can't reproduce this @arikfr locally. But I added #3120 since that was the only difference to the query tags code. Do you mind checking if this fixes the issue for your environment?

arikfr:
I see it happens in the preview instance too. Do you have locally any dashboard with more than one widget?

jezdez:
Ah, my test dashboards only have one each! Thanks, will check..

jezdez:
I tried a bunch of variations of this, and I can't get the count to apply distinct correctly. @washort do you have any ideas how to express this with SQLA?

jezdez:
I'm wondering if this is possible by splitting the query in two separate ones, using a subquery for the tag counts, then one to get the list of dashboards using that subquery.

It'd also be splendid if we would be able to use `Dashboard.all` for the list of dashboards as the base query as well, to not have separate code paths that build the query :-/

arikfr:
> It'd also be splendid if we would be able to use Dashboard.all for the list of dashboards as the base query as well, to not have separate code paths that build the query :-/

I agree. It will also be useful later on to show tags that are relevant to the currently showing dashboards (search results, based on some tag filtering, etc).

> I'm wondering if this is possible by splitting the query in two separate ones, using a subquery for the tag counts, then one to get the list of dashboards using that subquery.

Probably the other way around, right? First get the list of dashboards, then extract tags + counts from it.

jezdez:
> > It'd also be splendid if we would be able to use Dashboard.all for the list of dashboards as the base query as well, to not have separate code paths that build the query :-/
> 
> I agree. It will also be useful later on to show tags that are relevant to the currently showing dashboards (search results, based on some tag filtering, etc).
> 
> > I'm wondering if this is possible by splitting the query in two separate ones, using a subquery for the tag counts, then one to get the list of dashboards using that subquery.
> 
> Probably the other way around, right? First get the list of dashboards, then extract tags + counts from it.

Yep, see the PR, I think that'll do the trick.

PR #3120, opened 2018-11-26
Fix tag counts for dashboards and queries.

jezdez:
Refs #3108.


jezdez:
FWIW, the main part is just b9f254a5b19dadfa4ab1b0792bc7086762d07758. Feel free to ignore the rest, which I only did to better understand the query.

docker-compose.yml
```diff
@@ -39,5 +39,7 @@ services:
     # The following turns the DB into less durable, but gains significant performance improvements for the tests run (x3
     # improvement on my personal machine). We should consider moving this into a dedicated Docker Compose configuration for
     # tests.
+    ports:
+      - "15432:5432"
```
arikfr: ☝️

redash/models.py
```diff
@@ -1390,26 +1390,19 @@ def search(cls, org, groups_ids, user_id, search_term):
...
-        query = query.filter(or_(Dashboard.user_id == user.id, Dashboard.is_draft == False))
-
-        return query.order_by(usage_count.desc())
+            db.session
+            .query(tag_column, usage_count)
+            .group_by(tag_column)
+            .filter(Dashboard.id.in_(dashboards.options(load_only('id'))))
+            .order_by(usage_count.desc())
+        )
+        return query
```
arikfr: Can I bother you with adding a test case for this bug & method?

docker-compose.yml
jezdez: Yeah, so question: I often do that to inspect the database during development using a macOS app ([Postico](https://eggerapps.at/postico/)). Do you mind having that exposed for development?

redash/models.py
jezdez: No bother :)

docker-compose.yml
arikfr: You could use Redash to inspect it ;-) But sure, I guess there is no harm in leaving this.

jezdez: Heh, I do that, too, in some cases. But in others it's easier to just use other software, plus it allows editing as well (e.g. reverting archiving queries and dashboards).

arikfr:
> The query looks like this btw:

Most of the complexity comes from the fact that permissions to dashboards are based on permissions to queries, which are based on permissions to data sources... In the not so far future I'm hoping to change this. Hopefully it will make the queries saner in the process.

arikfr:
Thank you for updating the Query query too. Once we make the subquery a parameter `all_tags` receives, it looks like we can move it into a mixin class, which we can reuse for both Query and Dashboard.

Probably not worth tackling now, but worth making a mental note for the future.

jezdez:
Yeah, I was about to do that, but then remembered that I've already split `models.py` in #3026 introducing a `redash/models/mixins.py`. Let's wait for this to land and then we can refactor this a tiny bit.

arikfr:
Thanks!

PR #3121, opened 2018-11-26
Fix forked query is opened in the same tab

### LINKED ISSUE #3118: Every forked query is opened in the same tab
Every forked query is being opened in the same tab. Probably because of the named window thing.

kyoshidajp:
Fix #3118.

The value of `tabName` must be unique for each browser. 

`Math.random()` actually isn't strictly unique. But I think it's enough in this case.

### DIFF
--- a/client/app/pages/queries/view.js
+++ b/client/app/pages/queries/view.js
@@ -178,7 +178,9 @@ function QueryViewCtrl(
   };
 
   $scope.duplicateQuery = () => {
-    const tabName = 'duplicatedQueryTab';
+    // To prevent opening the same tab, name must be unique for each browser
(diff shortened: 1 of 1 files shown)


gabrieldutra:
`window.open()` with the title omitted worked fine even though [Mozilla Docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) doesn't mention this as optional. If you feel you need to keep this parameter, you can take advantage of this:
> If the string is empty, the browser will create a new window every time (this behavior doesn't work when the string is replaced with NULL).

kyoshidajp:
If a name isn't specified, pop-up will be blocked. If user allows it, then will open. But I think it isn't a good behavior.

pop_up_blocked

pop_up_block2

PR #3128, opened 2018-11-26
Use lower timeout for the first 5 seconds when polling for results.

arikfr:
This will result in faster query loads for fast queries.

### DIFF
--- a/client/app/services/query-result.js
+++ b/client/app/services/query-result.js
@@ -35,8 +35,7 @@ export function getColumnCleanName(column) {
 }
 
 function getColumnFriendlyName(column) {
-  return getColumnNameWithoutType(column).replace(/(?:^|\s)\S/g, a =>
-    a.toUpperCase());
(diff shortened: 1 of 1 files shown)


client/app/services/query-result.js
```diff
@@ -493,22 +492,32 @@ function QueryResultService($resource, $timeout, $q, QueryResultError) {
...
-        // TODO: use QueryResultError, or better yet: exception/reject of promise.
-        this.update({ job: { error: 'failed communicating with server. Please check your Internet connection and try again.', status: 4 } });
-      });
+          if (this.getStatus() === 'processing' && this.job.query_result_id && this.job.query_result_id !== 'None') {
+            this.loadResult();
+          } else if (this.getStatus() !== 'failed') {
+            const waitTime = tryNumber > 10 ? 3000 : 500;
+            $timeout(() => {
+              this.refreshStatus(query, tryNumber + 1);
+            }, waitTime);
```
arikfr: These 4 lines are the actual change. 

The rest is auto formatting...

PR #3131, opened 2018-11-27
WIP: New visualization: Treemap with d3

deecay:
Adding new visualization to visualize hierarchical data.

treemap title

* Hierarchy data must be represented in "adjacency list" model.
### example of Adjacency List data
| Parent | Child | 
---| ---
| | root | 
| root| branchA| 
| root| branchB| 
| branchA| leafA| 
| branchA| leafB| 
| branchA| leafC| 
## sample query
Result is an adjacency list hierarchy of Redash dashboards and visualizations.
```
-- visualizations
select 'dashboard' || CAST(dashboard_id as TEXT) as "parent", 'viz' || CAST(visualization_id as TEXT) as "child", width as "width" from widgets

-- dashboards
union
select 'root' as "parent", 'dashboard' || CAST(id as TEXT) as "child", null from dashboards

-- root node
union
select null, 'root', null from widgets
```
### Editor
General
image

Colors
image


client/app/visualizations/treemap/treemap.less
```diff
@@ -0,0 +1,48 @@
+.node {
```
kravets-levko: Can you change these classes to be more specific/scoped? `.node`, `rect`, `text` and some others are too generic and may affect other parts of the app.

client/app/visualizations/treemap/treemap-editor.html
```diff
@@ -0,0 +1,122 @@
...
+    </div>
+    <div class="form-group">
+      <label class="ui-sortable-bypass text-muted" style="font-weight: normal; cursor: pointer;"
+        uib-popover-html="templateHint"
+        popover-trigger="'click outsideClick'" popover-placement="top-left">
+        Format specs <i class="fa fa-question-circle m-l-5"></i>
+      </label>
+    </div>
+  </div>
+  <div ng-show="currentTab == 'colors'" class="m-b-10">
```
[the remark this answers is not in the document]
deecay: Got it. Done.

client/app/visualizations/treemap/treemap.less
deecay: My bad. Done.

arikfr:
This comment is of course relevant here as well: https://github.com/getredash/redash/pull/3154#issuecomment-444054536.

deecay:
Updated React compliance. Please let me know if I've missed something.

kravets-levko:
Thanks @deecay! I'll review ASAP

client/app/visualizations/treemap/treemap.less
```diff
@@ -0,0 +1,33 @@
...
+  .treemap-tooltip {
+    display: none;
+    position: absolute;
+    padding: 6px;
+    background-color: #fff;
+    border: 1px solid #fff;
+    border-radius: 3px;
+    color: #222;
+    white-space: nowrap;
+    -webkit-user-select: none;
```
kravets-levko: No need to add browser-specific prefixes - webpack will do it

kravets-levko:
BTW - I created a demo of this visualization on preview instance, so whoever wants to play with it can use it: https://deploy-preview-3131--redash-preview.netlify.com/queries/189/source#332 (once this PR will be merged, I'll ad it to our "all visualizations" demo dashboard).

kravets-levko:
@ranbena I think that's value from root or some parents (?). Probably visualization should ignore "size" of all parents and aggregate it from leafs and propagate up to the root.

ranbena:
> Upd.: probably, there should be no tooltip/label at all on non-leaf blocks 🤔

Exactly!

kravets-levko:
One more thing that crossed my mind: do we really need a "root" record to be returned by query? It's actually not shown anywhere on visualization, so, maybe, it could be dynamically created (if needed internally)? Also, I guess, when user selects "Child" column (in fact - record unique key) - it's enough to build "flat" treemap without sub-blocks. WDYT?

ranbena:
> Upd.: probably, there should be no tooltip/label at all on non-leaf blocks 🤔

Another way to handle it is to remove the "Cell Padding" settings and hardcode it to `{ cellPadding: 0 }`. It looks fine and the problem doesn't surface 😬

deecay:
Hi @kravets-levko , @ranbena ,

Thanks so much for the interest and comments.

However, I think I should go this direction now. [plotly native treemap](https://codepen.io/MojtabaSamimi/pen/ExYwNYe)

arikfr:
Hi,

Thanks.

PR #3132, opened 2018-11-27
Add function python data source  to run query from specific query_id

k-tomoyasu:
Hi,

I added python data source function `execute_by_query_id(query_id, params)`.   
It run query from specific query id with query parameters.  

This enable using queries as functions in python data source. 

Thanks,

### DIFF
--- a/redash/query_runner/python.py
+++ b/redash/query_runner/python.py
@@ -4,7 +4,7 @@ import logging
 import sys
 
 from redash.query_runner import *
-from redash.utils import json_dumps, json_loads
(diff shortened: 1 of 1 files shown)


kamedono:
Hi @arikfr 
when will this be merge?

guidopetri:
@k-tomoyasu , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
There's another PR around which allows people to give names to other queries too (eg `get_the_stuff` instead of `query_55`).  Wonder how that will interact with this one?

PR #3137, opened 2018-11-28
Build custom alert message

k-tomoyasu:
Build custom alert messages with query results.  

Alert feature is useful but we cannot see detail about query result.
This feature enable us write custom alert messages with query results (using template engine `Jinja2`).

WIP yet. I would appreciate if  give me feedback.   

alert destination  
- [x] email
- [x] slack 
- [x] webhook
- [x] chatworks
- [x] pagerduty
- [x] mattermost
- [x] hangoutschat
-  ~hipchat~

setting  
setting

slack alert
slack


kocsmy:
Nice feature @k-tomoyasu — regarding the color of the preview button: we don't use that color so using btn-default would make sense. I'll think about having a better layout for this in the meantime.

k-tomoyasu:
> Nice feature @k-tomoyasu — regarding the color of the preview button: we don't use that color so using btn-default would make sense. I'll think about having a better layout for this in the meantime.

@kocsmy Thanks for your advice! 
Firstly, I changed button color (and some fixes) and update screenshots.

k-tomoyasu:
@jupe Thanks for your comment!
  >  Could rest of integrations be done in follow-up PR?  

That's a big help if it is OK.
I was not good progress of integration because rest destinations are not easy to debug in my environment.

jupe:
For our case email&slack would be enough. Most probably we doesn’t need anything else, at least in near future. Currently slack alert is almost useless because it gives just title, not even link.. so this would be really welcome feature :)

client/app/services/alert-template.js
```diff
@@ -0,0 +1,48 @@
...
+
+  return Alert;
+}
+
+
+export default function init(ngModule) {
+  ngModule.factory('AlertTemplate', AlertTemplateService);
+
+  ngModule.run(($injector) => {
+    AlertTemplate = $injector.get('AlertTemplate');
```
rauchy: I may be missing something, but why declare this variable in line 3 and assign it here? Why not simply 
```javascript
export const AlertTemplate = $injector.get('AlertTemplate');
```

rauchy: Hehe, disregard this. I was reviewing on a commit-level and saw you already addressed this in a later commit!

migrations/versions/ed7bf6adbd4d_add_alert_template.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'ed7bf6adbd4d'
+down_revision = '73beceabb948'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('alerts', sa.Column('template', sa.Text(), nullable=True))
```
arikfr: Let's put `template` in the `options` JSON object we already have in `alerts`.

redash/models/__init__.py
```diff
@@ -772,6 +773,10 @@ def evaluate(self):
     def subscribers(self):
         return User.query.join(AlertSubscription).filter(AlertSubscription.alert == self)
 
+    def render_template(self, showError=None):
```
arikfr: `showError` seems to be unused? Also should be `show_error`.

k-tomoyasu:
@arikfr Thanks for your review! Sorry for my late reply.

> 1. How about we use Mustache as the template language?  

Agree. I use `Mustache` both server and client.  

> 2. For Email type notification it will be beneficial to be able to edit the subject too.

I think it beneficial. I'll try to consider this feature!

redash/models/__init__.py
k-tomoyasu: thanks! I've removed.

migrations/versions/ed7bf6adbd4d_add_alert_template.py
k-tomoyasu: okay, fix logic and remove migration file.

k-tomoyasu:
@arikfr I've improved based on your feedback.

> How about we use Mustache as the template language?

I've modified this.

> For Email type notification it will be beneficial to be able to edit the subject too.

I added `Custom subject` option.  

setting

redash/models/__init__.py
```diff
@@ -772,6 +772,25 @@ def evaluate(self):
...
+            return self.options['template']
+        else:
+            return ""
+
+    @property
+    def custom_subject(self):
+        if 'subject' in self.options:
+            return self.options['subject']
+        else:
+            return ""
```
arikfr: You can simply call `self.options.get('subject', '')`.

redash/models/__init__.py
```diff
@@ -772,6 +772,25 @@ def evaluate(self):
     def subscribers(self):
         return User.query.join(AlertSubscription).filter(AlertSubscription.alert == self)
 
+    def render_template(self):
+        data = json_loads(self.query_rel.latest_query_data.data)
+        context = {'rows': data['rows'], 'cols': data['columns']}
+        return mustache_render(self.options['template'], context)
```
```suggestion
        return mustache_render(self.template, context)
```

redash/destinations/webhook.py
```diff
@@ -36,7 +36,8 @@ def notify(self, alert, query, user, new_state, app, host, options):
             data = {
                 'event': 'alert_state_change',
                 'alert': serialize_alert(alert, full=False),
-                'url_base': host 
+                'url_base': host,
+                "description": alert.render_template() if alert.template else ''
```
arikfr: Let's encapsulate this logic in the `render_template` function, i.e. just return an empty string if the template is empty.

client/app/pages/alert/index.js
```diff
@@ -70,6 +76,22 @@ function AlertCtrl($routeParams, $location, $sce, toastr, currentUser, Query, Ev
     );
   };
 
+  this.preview = () => {
+    const notifyError = () => toastr.error('Unable to render description. please confirm your template.', { timeOut: 10000 });
+    try {
+      const result = this.alertTemplate.render(this.alert.options.template, this.queryResult.query_result.data);
+      this.alert.preview = $sce.trustAsHtml(result.escaped);
+      this.alert.previewHTML = $sce.trustAsHtml(result.raw);
```
arikfr: This needs to be sanitized before trusting.

redash/models/__init__.py
```diff
@@ -772,6 +772,25 @@ def evaluate(self):
     def subscribers(self):
         return User.query.join(AlertSubscription).filter(AlertSubscription.alert == self)
 
+    def render_template(self):
+        data = json_loads(self.query_rel.latest_query_data.data)
+        context = {'rows': data['rows'], 'cols': data['columns']}
```
arikfr: How about we add to the context information about the alert like the state?

redash/models/__init__.py
k-tomoyasu: thanks. I've fixed.

redash/models/__init__.py
k-tomoyasu: thanks. I've fixed.

redash/models/__init__.py
k-tomoyasu: I added `alert.state` and fixed help message.

client/app/pages/alert/index.js
k-tomoyasu: thank you for pointing mistake!

redash/destinations/webhook.py
k-tomoyasu: refactored. thanks!

k-tomoyasu:
@arikfr 

> Thank you for updating this! Please see follow up comments.

Thank you for comments! 
I've updated.  it's ready to review.

k-tomoyasu:
@arikfr
I implemented on other destinations(except `hipchat`).  
(except hipchat).  

> Thank you for updating this! Please see follow up comments.

Thank you for comments!
I've updated. it's ready to review.

arikfr:
Apologies for letting this linger. I will try to get to it this week 🤞

buritto:
Hello, I am also looking forward to this functionality. How soon does this appear in redash?
Thank.

arikfr:
> Apologies for letting this linger. I will try to get to it this week 🤞

"this week"... 2 months later here I am 😅 

To avoid further review cycles, I did some changes myself:

1. Updated with latest master (there were some merge conflicts to fix).
2. Updated how the Webhook destination handles these new options (b078866).
3.  Add log message when checking alert (60cb858, not really related to this PR, just happened to include it).
4. Added a feature flag (`REDASH_FEATURE_EXTENDED_ALERT_OPTIONS`, disabled by default) to control whether to show the new options.

The reason for the feature flag is that the UI still needs more work. I will work on the product and UX definition with @ranbena and we will do these changes ourselves (along with converting this page to React).

But you can already use this in your setup 👍 

Thank you for your effort here, this great feature and patience during the process.

Once the build passes I will merge this.

arikfr:
@k-tomoyasu it's not a big thing, but the email address you signed this commits with (k.tomoyasu@geniee.co.jp) is not associated with your GitHub account. This means that you won't show up in the contributors page on GitHub :-( You can fix this by verifying this email address in your account: https://help.github.com/en/articles/setting-your-commit-email-address.

arikfr:
Merged 👍

k-tomoyasu:
Thanks for fixe and future plan:+1:
I'm looking forward to next release:smile:

> You can fix this by verifying this email address in your account

Thank you for your kindness. But this email is invalid now...
I'm glad to be listed on the page at the next opportunity!

Vaidas737:
Hi, how to enable this feature in version 8.0.0

ranbena:
@vaidas737 https://redash.io/help/user-guide/alerts/custom-alert-notifications

Vaidas737:
Thanks

Vaidas737:
@arikfr thanks a lot

auziel:
> @Vaidas737 in v8, you can enable it with an env var, but note that it won't be 100% compatible with the final implementation in v9.

Hi @arikfr I know it is a year old feature and v9 is already there, but I'm still on v8.0.0. I didn't find a clear explanation how to enable it in v8.
I try to set `FEATURE_EXTENDED_ALERT_OPTIONS=true` which didn't enable the feature. 
Is there a clear doc how to enable this feature for v8.0.0?
How can I customize the message body (I'm using Slack as the alert target)?

isdzulqor:
> > @Vaidas737 in v8, you can enable it with an env var, but note that it won't be 100% compatible with the final implementation in v9.
> 
> Hi @arikfr I know it is a year old feature and v9 is already there, but I'm still on v8.0.0. I didn't find a clear explanation how to enable it in v8.
> I try to set `FEATURE_EXTENDED_ALERT_OPTIONS=true` which didn't enable the feature.
> Is there a clear doc how to enable this feature for v8.0.0?
> How can I customize the message body (I'm using Slack as the alert target)?

try `REDASH_FEATURE_EXTENDED_ALERT_OPTIONS` not `FEATURE_EXTENDED_ALERT_OPTIONS ` @auziel

moonsub-kim:
Can I use this feature with the docker version `redash/redash:10.1.0.b50633`?

susodapop:
> Can I use this feature with the docker version `redash/redash:10.1.0.b50633`?

This feature was replaced in V9 and is now always enabled (the environment variable has no effect). You can read about how to use it [here](https://redash.io/help/user-guide/alerts/custom-alert-notifications).

PR #3138, opened 2018-11-28
Migrate all tags components to React

kravets-levko:
Tag list component looks as it was:
image

Edit Tags modal uses tags input from Ant library:

image
image
image


kravets-levko:
@arikfr Hm, yes. Seems there should be some animation, but instead it just waits and then removes node. I'll check this

arikfr:
OK, now it's better :) Can you make it autofocus on the input field?

webpack.config.js
```diff
@@ -170,7 +172,7 @@ const config = {
       }
     ]
   },
-  devtool: "cheap-eval-module-source-map",
+  devtool: isProduction ? false /* "source-map" */ : "cheap-eval-module-source-map",
```
kravets-levko: @arikfr At some point sourcemaps become to large and build on CircleCI fails. I tried to increase memory limit  (`NODE_OPTIONS="--max-old-space-size=4096"`) but it didn't help. With disabled sourcemaps in `production` mode build passes, but I think we need sourcemaps for production builds. Currently I have no good idea what to do with it. I can try different types of sourcemaps, but it does not guarantee that it will not fail next time. Another option is to detect CI builds somehow and disable sourcemaps only for them. WDYT?

arikfr: > I tried to increase memory limit (NODE_OPTIONS="--max-old-space-size=4096") but it didn't help. 

Did you try applying it the same way it was done in the visualizations pull request? Maybe `NODE_OPTIONS` gets ignored for some reason?

While we don't do much with source maps in production, they are still sometimes useful (recently I was debugging in production using them). 

I would try two things before disabling source maps:

1. Not to specify `devtool` at all and rely on webpack's defaults.
2. Pass the `max-old-space-size` option directly to the node process instead of relying on NODE_OPTIONS.

kravets-levko: Hm... Option 2 worked fine. I think we can keep it for now - sourcemaps should become much smaller when we'll finally remove Angular and related libraries.

client/app/components/tags-control/TagsControl.jsx
```diff
@@ -0,0 +1,103 @@
...
+          onEdit(newTags);
+        }
+      });
+  }
+
+  editTags() {
+    const { getAvailableTags } = this.props;
+    const tags = map(this.props.tags, trim);
+
+    if (isFunction(getAvailableTags)) {
```
arikfr: Why wouldn't it be a function? Same goes for all the protective code around the result from calling this function, instead of assuming it returns a Promise.

client/app/components/tags-control/ModelTagsControl.jsx
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  static defaultProps = extend({}, TagsControl.defaultProps, {
+    isDraft: false,
+    isArchived: false,
+  });
+
+  static archivedTooltip = '';
+
+  renderPrependTags() {
```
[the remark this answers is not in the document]
kravets-levko: That additional tags should be in the same container with regular tags - otherwise there are UI issues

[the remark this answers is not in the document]
arikfr: Got it 👍

arikfr:
👍

PR #3139, opened 2018-11-28
Better manage permissions modal

kocsmy:
I've improved the modal with the help of @kravets-levko 

- Dropdown now has proper spacing around its elements
- Query owner is no longer in the dropdown options (and can't be added either)
- Query owner is presented in the first line of the table by default

image

### DIFF
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -538,6 +538,16 @@ nav .rg-bottom {
   padding: 0 0 0 23px;
 }
 
+.table--permission {
+  .profile__image {
+    margin-right: 0;
+  }
+}
+
+.mp__permission-type {
+  text-transform: capitalize;
+}
+
 // Smaller screens
 
 @media (max-width: 880px) {
@@ -599,7 +609,7 @@ nav .rg-bottom {
         white-space: nowrap;
       }
     }
-    
+
     main {
       flex-direction: column-reverse;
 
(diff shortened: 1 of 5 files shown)


client/app/pages/queries/view.js
```diff
@@ -487,6 +487,7 @@ function QueryViewCtrl(
       component: 'permissionsEditor',
       resolve: {
         aclUrl: { url: `api/queries/${$routeParams.queryId}/acl` },
+        query: $scope.query,
```
arikfr: This same dialog is used for dashboards too. It makes more sense to pass in the id of the user who owns the object.

client/app/assets/less/redash/query.less
```diff
@@ -538,6 +538,16 @@ nav .rg-bottom {
   padding: 0 0 0 23px;
 }
 
+.table--permission {
+  .profile__image {
+    margin-right: 0;
+  }
+}
+
+.mp__permission-type {
```
arikfr: What's `mp`? And why two styles of naming? (`--` vs `__`)

kocsmy: mp refers to manage permission but ended up not using elsewhere, maybe the table should be mp__table but I chose to keep it aligned with the other table class names: "table table-condensed table-hover table--permission".

That's not 2 styles, it's BEM: http://getbem.com/naming/

client/app/pages/queries/view.js
kocsmy: Oh it's being used in dashboards, too! @kravets-levko can you help me fix it so it works on dashboards, too?

kravets-levko: @arikfr (cc @kocsmy) Fixed

kocsmy:
I checked it and all looks good 👌 Thanks, Levko!

arikfr:
👍

PR #3140, opened 2018-11-28
WIP: implement schema browser as a react component

spasovski:
Second attempt. init.init was missing in the new component. This will look different due to Antd though I can make it more closely resemble the current UI if requested.


arikfr:
Was there a reason that you didn't use the [Tree](https://ant.design/components/tree/) component?

washort:
Updated to master, now uses Tree component.

spasovski:
This is ready for a re-review. Please let me know if anything else needs to be addressed.

arikfr:
While you addressed the issues from my previous comment, you have introduced new issues:

image

1. Why the columns on the right?
2. When clicking on a column it gets blue background color. There is not supposed to be a click behavior for columns.
3. Search is very unresponsive -- maybe because of the 500ms debounce?
4. The search query seems to repeat key strokes: type some word like `Album` then delete it with the backspace, why a moment you will see `A` returns into the search query.

spasovski:
I don't see any of that locally though I am trying to repair my local environment since the worker processes kept erroring out after the rebase. It should be almost identical to the angular version granted the debounce can be removed/lowered and it also preserves the flex overflow ellipsis.

Appreciate the quick reply though I'll check if ashort and others can see the same issue in the branch before asking for rereview.

ranbena:
@spasovski I think AntD's `Menu` and `SubMenu` would be a better fit than `Tree`.

kravets-levko:
@spasovski This is a screenshot from this PR's [Netlify preview](https://deploy-preview-3140--redash-preview.netlify.com/):

image

Comparing to `master`, it also misses some margins around schema browser:

image

client/app/components/dynamic-table/index.js
```diff
@@ -69,12 +69,12 @@ function createRowRenderTemplate(columns, $compile) {
     switch (column.displayAs) {
       case 'json':
         return `
-          <dynamic-table-json-cell column="columns[${index}]" 
+          <dynamic-table-json-cell column="columns[${index}]"
```
kravets-levko: Is it somehow related to this PR?

ranbena:
> @spasovski I think AntD's Menu and SubMenu would be a better fit than Tree.

Lol I didn't realize you were asked to switch to Tree 😵
Lemme get back to you on that.

client/app/components/queries/schema-browser.html
```diff
@@ -1,30 +0,0 @@
...
-    <input type="text" placeholder="Search schema..." class="form-control" ng-model="$ctrl.schemaFilter" ng-model-options="{ debounce: 500 }" ng-disabled="$ctrl.isEmpty()" ng-change="$ctrl.splitFilter($ctrl.schemaFilter);">
-    <button class="btn btn-default"
-            title="Refresh Schema"
-            ng-click="$ctrl.onRefresh()">
-      <span class="zmdi zmdi-refresh"></span>
-    </button>
-  </div>
-
-  <div class="schema-browser" vs-repeat vs-size="$ctrl.getSize(table)">
-    <div ng-repeat="table in $ctrl.schema | filter:$ctrl.schemaFilterObject track by table.name">
```
kravets-levko: This file was the only place that was using `angular-vs-repeat` - consider removing that dependency too.

ranbena:
@spasovski which Ant component do you think would be best here?

kravets-levko:
@ranbena I think the best is to keep existing markup - it also doesn't require any CSS changes. In this case it's really simple component.

client/app/components/dynamic-table/index.js
spasovski: An earlier linter was whining about this though it is otherwise unrelated. Will gladly remove the whitespace change if needed.

kravets-levko: Interesting. What errors did you get from linter?

spasovski:
This is ready for another review pass. Uses antd's `<Icon>` but is regular markup otherwise.

client/app/assets/less/ant.less
```diff
@@ -247,3 +247,34 @@
 .ant-popover {
   z-index: 1000; // make sure it doesn't cover drawer
 }
+
+// schema browser table expand icon override
+.schema-container {
+  .schema-browser-title-item .copyArrow {
```
kravets-levko: Please use kebab-case for CSS (e.g. `.copy-arrow`)

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,134 @@
...
+        onClick={(evt) => { this.doCopy(evt, [title]); }}
+        aria-hidden="true"
+        type="double-right"
+        className="copyArrow"
+      />
+    </span>
+  );
+
+  doRowToggle = ({ target }) => {
+    target.closest('.table-item').classList.toggle('schema-table-expanded');
```
kravets-levko: I think it should be implemented in more "native" way: create a component that will represent a single table. That component will accept table name and columns via props (also add other necessary props), and will have the only state field - `isExpanded`. Also, it's a good case for using React hooks (`useState`).

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  doCopy(evt, hierarchy) {
+    // eslint-disable-next-line
+    this.props.$rootScope.$broadcast('query-editor.command', 'paste', hierarchy.join('.'));
+    evt.stopPropagation();
```
kravets-levko: Import `$rootScope` from `@/services/ng` - don't pass it via props (+ remove from `react2angular`)

client/app/components/dynamic-table/index.js
kravets-levko: ping :slightly_smiling_face: I'm really wondering why there are any codestyle notices for this code - because that's string literal and any linter wont touch it.

spasovski: I am probably not remembering right. I think (in november of last year) there was something the linter required here (possibly unused import or something) and after I made the change my code editor removed the trailing whitespace. I will go into vi and re-add it.

kravets-levko: Which editor do you use? Because my one also removes trailing whitespaces, but keeps in multiline string literals (because in general case it's unsafe to remove them)

kravets-levko: And while it's not so important, but I think you should revert this - because the entire file is not related to changes declared in this PR.

spasovski: Yep agreed.

client/app/components/queries/SchemaBrowser.jsx
kravets-levko: BTW - with this approach you can even remove css class, and just conditionally render columns only when table is expanded.

client/app/assets/less/ant.less
```diff
@@ -247,3 +247,34 @@
...
+    cursor: pointer;
+
+    svg {
+      transform: scale(0.6);
+    }
+  }
+  .table-open {
+    display: none;
+
+    .copyArrow {
```
kravets-levko: kebab-case

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,135 @@
+import React from 'react';
+import { react2angular } from 'react2angular';
+import PropTypes from 'prop-types';
+import { Icon } from 'antd';
+import { debounce } from 'lodash';
+import { $rootScope } from '@/services/ng';
+
+import { Schema } from '../proptypes';
+
+const isSchemaEmpty = schema => schema === undefined || !schema.length;
```
kravets-levko: `isArray(schema) && (schema.length > 0)`:

1. it's self-explaining: we expect that schema is an array with non-zero length.
2. it's more reliable (what if schema will be null? object?)

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+    // Candidate to be abstracted out to a set of config constants.
+    // Performs search filtering after the number of milliseconds set below.
+    this.doSearch = debounce(this.doSearch, 150);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevProps.schema && this.props.schema.length && (prevState.schema.length !== this.props.schema.length)) {
+      // setState() is fine here within the right conditions.
+      // eslint-disable-next-line
```
kravets-levko: Use this with a list of rules you want to disable

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,135 @@
...
+      if (!this.state.schemaInitialized) this.setState({ schema: this.props.schema });
```
kravets-levko: This is the line where eslint should fail (we require block even for a single statement).

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  handleSearch = (evt) => {
+    if (!evt.target.value) {
+      this.setState({ schema: this.props.schema, searchTerm: '' });
```
kravets-levko: `evt.target.value` is always string; here you should just update state and call `doSearch`, which should properly handle empty search term (return entire schema).

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,135 @@
...
+      return;
+    }
+    this.setState({ searchTerm: evt.target.value });
+    this.doSearch(evt.target.value);
+  }
+
+  doSearch = (searchTerm) => {
+    const schema = this.props.schema.filter(table => doesTableContainSearchTerm(searchTerm, table));
+    this.setState({ searchTerm, schema });
```
kravets-levko: Seems you don't need `handleSearch` at all. This function should look line this:
```javascript
doSearch = (searchTerm) => {
  let { schema } = this.props;
  if (searchTerm !== '') {
    schema = .filter(table => doesTableContainSearchTerm(searchTerm, table));
  }
  this.setState({ searchTerm, schema });
}
```

Then use it directly with input: ```onChange={event => this.doSearch(event.target.value)}.

Also, move debounce call here as well.

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  render() {
+    if (!this.props.schema) {
+      return null;
+    }
+    return (
+      <div className="schema-container">
+        <div className="schema-control">
+          <input
```
kravets-levko: Use `Input` and `Button` components from Ant

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,135 @@
...
+      schemaInitialized: false,
+    };
+
+    // Candidate to be abstracted out to a set of config constants.
+    // Performs search filtering after the number of milliseconds set below.
+    this.doSearch = debounce(this.doSearch, 150);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevProps.schema && (this.props.schema && prevState.schema.length !== this.props.schema.length)) {
```
kravets-levko: On query page, `$scope.schema` is set to a new array each time it's updated. So this check could be simplified to `prevProps.schema !== this.props.schema`. Also, it's the only condition needed - component should be ready that this prop may change - in this case filter should be applied to new schema.

spasovski:
The PR is not ready for re-review - I should've realized you'd get pinged. I can close until it's ready for next round if you'd prefer.

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+import { Schema } from '../proptypes';
+
+const isSchemaEmpty = schema => schema === undefined || !schema.length;
+
+// Return true if the search term is contained within the table name
+// or any of the table's columns.
+const doesTableContainSearchTerm = (searchTerm, table) => {
+  const regEx = new RegExp(searchTerm, 'gi');
+
```
kravets-levko: One more notice about this function: avoid using regex, especially in this form. First of all, it allows users to use arbitrary regex to filter schema, as well as it breaks previous behavior. Also, using `g` modifier makes regex "sticky" - it will remember last position where match was found, and next call to `test` will start matching from that position.

client/app/components/queries/SchemaBrowser.jsx
spasovski: what about: 
`const isSchemaValid = schema => schema.constructor === Array && schema.length;`

kravets-levko: `isArray` === `schema.constructor === Array`, but it's shorter and literally means what we need to check - that argument "is array" (vs. "argument's constructor is equal to Array") - that's rather a question of semantics.

PR #3141, opened 2018-11-29
Fix schema refresh fails to work on MySQL 8

hoangphuoc25:
Currently MySQL 8 returns capitalized column headers for schema refreshing query. This causes schema refreshing to fail for MySQL 8.
This PR adds a fix for that problem.

Tested on local MySQL v8.0.13 and RDS MySQL 8.


arikfr:
Thanks!

hoangphuoc25:
You're welcome!

Issue #3015, opened 2018-10-28
v6 Release Plan


arikfr:
Time to start aiming at releasing v6! I created the [v6 milestone](https://github.com/getredash/redash/milestone/14) to track any remaining work and this issue for tasks related to the release.

I hope that we can have a beta version by next week and a final release 2-3 weeks after.

- [x] Fix all issues and merge pull requests in the milestone
- [x] Update CHANGELOG ([diff](https://github.com/getredash/redash/compare/v5.0.0...master), @jezdez): #3112  
- [x] Sanity tests for the current master

When all issues are done:

- [x] Create a release branch
- [x] Create GitHub release and update `version.redash.io`
- [ ] Announce beta version
  - [x] Forum
  - [ ] Twitter
  - [ ] Facebook
- [x] Need to mention in upgrade notes:
   - [x] The need to upgrade first to v5.0.2, for smooth Celery migration
   - [x] The need to clear out the Celery queue (because by default v4 uses JSON instead of Pickle)

After the beta is stable:

- [ ] TBD.

jezdez:
@arikfr I'll do the changelog update!

arikfr:
@jezdez 👍 thanks.

jezdez:
@arikfr See #3112. I haven't added a blurb about the high level changes, or a summary. But generally speaking there are lots of UI/UX related changes and plenty of fixes and new additions of data sources. It's up-to-date up to babbeb79f0c9d7f8fe5a9f4e6b51c740b23e1e38.

arikfr:
Thanks, @jezdez. I will add the summary before wrapping it up.

jezdez:
I've asked @washort and @emtwo to test-drive master and see if anything obvious sticks out.

emtwo:
I've tried it out. Looks solid! Nothing obvious seems to stand out. Looking forward to the the release!

washort:
When editing a dashboard, changing the dashboard title results in the visualizations being replaced by the loading markers,

washort:
All new tags have "(create new tag)" appended to them,

arikfr:
The tags issue is something we’re aware of and @kravets-levko is working on a fix.

But the dashboard edit issue is nice find. 👍🏻

jezdez:
@arikfr Huzzah for the beta!

arikfr:
From what I seen so far the only open issue we should address before releasing the final version is #3181.

arikfr:
[...]

PR #3142, opened 2018-11-29
When editing a dashboard title results in the visualizations being replaced by the loading markers

kravets-levko:
Reported by @washort in the issue getredash/redash#3015 ([comment](https://github.com/getredash/redash/issues/3015#issuecomment-442516918))


arikfr:
👌

PR #3148, opened 2018-11-29
Fix label positioning on no found screen

kocsmy:
Before/After
image

image

Fixed for both dashboards and queries.

### DIFF
--- a/client/app/components/NoTaggedObjectsFound.jsx
+++ b/client/app/components/NoTaggedObjectsFound.jsx
@@ -6,9 +6,9 @@ import { BigMessage } from './BigMessage';
 function NoTaggedObjectsFound({ objectType, tags }) {
   return (
     <BigMessage icon="fa-tags">
-      No {objectType} found tagged with
+      No {objectType} found tagged with&nbsp;
       {Array.from(tags).map(tag => (
-        <span className="label label-tag" key={tag}>
+        <span className="label label-tag label--inline" key={tag}>
           {tag}
         </span>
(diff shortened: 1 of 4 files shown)


client/app/components/NoTaggedObjectsFound.jsx
```diff
@@ -6,9 +6,9 @@ import { BigMessage } from './BigMessage';
 function NoTaggedObjectsFound({ objectType, tags }) {
   return (
     <BigMessage icon="fa-tags">
-      No {objectType} found tagged with
+      No {objectType} found tagged with&nbsp;
       {Array.from(tags).map(tag => (
-        <span className="label label-tag" key={tag}>
+        <span className="label label-tag label--inline" key={tag}>
```
kravets-levko: I think we can now use [TagsControl](https://github.com/getredash/redash/blob/master/client/app/components/tags-control/TagsControl.jsx) here. If it's okay - I can implement it tomorrow (in this PR) (cc @arikfr)

arikfr: Will it also fix the alignment issue?

kravets-levko: Don't know :-) I just decided that it's more consistent to use existing component for displaying tags instead of rendering custom HTML

kravets-levko:
@kocsmy @arikfr The look didn't change:

image

Feel free to drop my commit if you're not happy with the result

PR #3155, opened 2018-12-01
WIP: Create a few more E2E Cypress tests

gabrieldutra:
## Description
The intention of this is to add a few more E2E tests in critical areas of Redash. (Suggestions accepted :grin:)

## Changes
- [x] Use `data-test` selectors for elements instead of id or class
- [x] Add DB Seed function
- [ ] Add PostgreSQL basic query test (I think we can take advantage of `docker-compose` for the most important Data Sources)
- [ ] Think of other possible tests to be added :thinking: 
- _Ad infinitum_...

LMK your ideas!! :grin:

### DIFF
--- a/client/app/components/app-header/app-header.html
+++ b/client/app/components/app-header/app-header.html
@@ -86,7 +86,7 @@
           <!--<a href="users" title="Settings"><i class="fa fa-cog"></i></a>-->
         <!--</li>-->
         <li class="dropdown" uib-dropdown>
-          <a href="#" class="dropdown-toggle dropdown--profile" uib-dropdown-toggle data-cy="dropdown-profile">
+          <a href="#" class="dropdown-toggle dropdown--profile" uib-dropdown-toggle data-test="ProfileDropdown">
             <img ng-src="{{ $ctrl.currentUser.profile_image_url }}" class="profile__image--navbar" width="20"/>
             <span class="dropdown--profile__username" ng-bind="$ctrl.currentUser.name"></span> <span
             class="caret caret--nav"></span></a>
--- a/client/app/components/dynamic-form.html
+++ b/client/app/components/dynamic-form.html
@@ -1,7 +1,8 @@
 <form name="dynamicForm">
     <div class="form-group required" ng-class='{"has-error": (dynamicForm.targetName | showError)}'>
         <label class="control-label" for="dataSourceName">Name</label>
-        <input type="string" class="form-control" name="targetName" ng-model="target.name" autofocus required>
+        <input type="string" class="form-control" name="targetName" ng-model="target.name"
+               data-test="TargetName" autofocus required>
         <error-messages input="dynamicForm.targetName" form="dynamicForm"></error-messages>
     </div>
     <hr>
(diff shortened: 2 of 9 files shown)


cypress/support/commands.js
```diff
@@ -13,3 +13,5 @@ Cypress.Commands.add('login', () => {
     body: users.admin,
   });
 });
+
+Cypress.Commands.add('getElement', element => cy.get('[data-test="' + element + '"]'));
```
arikfr: Maybe `getByTestId` to better convey the purpose?

gabrieldutra: Sure. That was actually one of those moments I was staring at the screen and couldn't think of any better, although unsatisfied with `getElement`  :man_facepalming:

gabrieldutra:
> True, but we should only use one (Postgres) type for running queries tests, to test the interface. To test the other data sources, it's enough to run an integration test at the Python code level.

Makes sense, since we are actually concerned about UI. Well, in case we at some point have a different Editor for a Data Source, we can use it.

> How about simple tests that visit each page we have (homepage, list pages, settings page, etc) and screenshot-test them with Percy? This doesn't test a lot, but it will make sure all routes are working and that no visual regressions were made. Over time, we can keep iterating and add actual tests.

Sounds good, it's simple and I think will benefit a lot. I have never used Percy, but what do you think about Cypress + CircleCI artifacts? I noticed that Cypress has a [screenshot feature](https://docs.cypress.io/api/commands/screenshot.html) and actually take screenshots after each test (we're just not taking use of that yet)

arikfr:
Just realized that the Percy tests will be super helpful with the React migration.

gabrieldutra:
> Just realized that the Percy tests will be super helpful with the React migration.

:grin:  I've already started testing and configuring it, so far I loved it. Anyway, I will continue later.

jezdez:
> > Just realized that the Percy tests will be super helpful with the React migration.
> 
> 😁 I've already started testing and configuring it, so far I loved it. Anyway, I will continue on that later.

YES! 😃

jrbenny35:
Could we split some of these tasks up? Maybe the Percy tests can be separate.

gabrieldutra:
> Could we split some of these tasks up? Maybe the Percy tests can be separate.

I think I went far with the tasks :sweat_smile:. Well, for now I think adding what's done + Percy setup will be enough. I also agree that Percy can be split in another PR, so if it's more convenient I can proceed with it.

arikfr:
Good point, @jrbenny35 .

@gabrieldutra I wouldn't bother with splitting, but I guess that it's worth to wrap up the Percy integration in this PR, and then further work in new PRs.

.dockerignore
```diff
@@ -3,4 +3,3 @@ client/dist/
 node_modules/
 .tmp/
 .venv/
-.git/
```
gabrieldutra: This was the easiest way I found to avoid `unknown-branch` on Percy. Does it impact elsewhere? :thinking:

arikfr: Yeah, it will unnecessarily increase the size of the Docker image. But I believe Percy can use env vars instead of using git directly: https://docs.percy.io/docs/environment-variables (not sure if this applies to the client we use though).

gabrieldutra: Those env vars were actually my second option haha. I'll handle it with them.

gabrieldutra:
To make use of `cy.percyScreenshot()` we have 2 options:
1. Use during the tests;
2. Make a specific file for screenshots, sth like:
```javascript
const pages = ['/', '/login', '/queries/1/source'];

describe('Screenshots', () => {
  pages.forEach((page) => {
    it(`take a screenshot of ${page}`, () => {
      cy.visit(page);
      cy.percySnapshot(page);
    });
  });
});
```
So far I'm with the first one, its advantage is that we can use the data from the tests. The second one would be better if we want better control of the location of these screenshots.

Let me know what you think :grin:

arikfr:
> To make use of cy.percyScreenshot() we have 2 options:
>
> 1. Use during the tests;
> 2. Make a specific file for screenshots, sth like:

Why not both? :-)

I think we should use it during tests, but until we have tests for everything we should have basic snapshot only tests ("populate a page with sample data" -> "visit page" -> "take snapshot").

Makes sense?

arikfr:
Looks good to me 👍 @gabrieldutra do you want to do anything else here or can we merge?

gabrieldutra:
> I think we should use it during tests, but until we have tests for everything we should have basic snapshot only tests ("populate a page with sample data" -> "visit page" -> "take snapshot").

Sounds good to me.

> Looks good to me +1 @gabrieldutra do you want to do anything else here or can we merge?

Considering a setup PR, for me it's fine to merge :grin:

PR #3156, opened 2018-12-02
json_dumps: add support for serializing buffer objects.

arikfr:


jezdez:
@arikfr For the record, buffer types are a Python 2-only feature. In Python 3 this needs to use memoryview objects.

jezdez:
> Is there something we can use from six to bridge?

Not really the [buffer built-in](https://docs.python.org/2.7/library/functions.html#buffer) was replaced with the memoryview built-in in Python 3, but they are not exactly the same. `memoryview` exists in Python 2.7 as well, so I'm wondering what datasource returns buffer objects? Could you elaborate what required this change?

There are a few PEPs with this, [3118](https://www.python.org/dev/peps/pep-3118/) and [3137](https://www.python.org/dev/peps/pep-3137/). And an [example change](https://github.com/python/cpython/commit/bae07c9baf3e53164de6f85a18ce747a76b9ffde) in CPython where Guido replaced buffer with memoryview form 2007.

arikfr:
It was psycopg2, but it supports both:

> Received data is returned as buffer (in Python 2) or memoryview (in Python 3).

([source](http://initd.org/psycopg/docs/usage.html#binary-adaptation))

I guess that when we switch to Python 3 we need to remember to update this function?

arikfr:
(of course instead of remembering, I will open a master issue to track Python 3 related work)

PR #3161, opened 2018-12-04
clickhouse: fixed int() conversion error

denisov-vlad:
Fixed errors:
1) int(None) from nullable integer value (now converts to None)
2) int([1,2,3]) from list of UInt64/Int64 values (now converts to string)


arikfr:
Thanks!

PR #3162, opened 2018-12-04
Remove API permissions for users who have been disabled

rauchy:


tests/test_models.py
```diff
@@ -508,3 +508,13 @@ def test_returns_dashboards_from_current_org_only(self):
 
         self.assertIn(w1.dashboard, models.Dashboard.all(self.u1.org, self.u1.group_ids, None))
         self.assertNotIn(w1.dashboard, models.Dashboard.all(user.org, user.group_ids, None))
+
+
+class ApiUserTest(TestCase):
+    def test_does_not_grant_permissions_to_disabled_users(self):
```
rauchy: I'm not sure this is the right place for this test, would appreciate feedback.

redash/authentication/__init__.py
```diff
@@ -110,7 +110,8 @@ def get_user_from_api_key(api_key, query_id):
             if query_id:
                 query = models.Query.get_by_id_and_org(query_id, org)
                 if query and query.api_key == api_key:
-                    user = models.ApiUser(api_key, query.org, query.groups.keys(), name="ApiKey: Query {}".format(query.id))
+                    name = "ApiKey: Query {}".format(query.id)
+                    user = models.ApiUser(api_key, query.org, query.groups.keys(), name, query.user)
```
arikfr: We don't use `ApiUser` for actual users. We load them here:

https://github.com/getredash/redash/blob/3ee7537a6c049a98e7284c67b98130556031e263/redash/authentication/__init__.py#L104

What we need to do is to check if it's a disabled user at this point and return `None` if it is.

rauchy: Could you explain the downsides of my approach? Not that I have a clear understanding of the areas of responsibilities that makes me think it's any better than your suggestion, but just because I lose a bunch of brain cells every time I introduce a null in an app.

arikfr: The main downside is that it's not working 😅... When you use a User's API key, the rest of the system sees this as a regular `User`, not as an `ApiKeyUser`. The later is only used when an API key of a dashboard or query is used...

rauchy: I would call that a winning argument 😆

redash/authentication/__init__.py
```diff
@@ -102,6 +102,8 @@ def get_user_from_api_key(api_key, query_id):
     org = current_org._get_current_object()
     try:
         user = models.User.get_by_api_key_and_org(api_key, org)
+        if user.disabled_at:
```
kravets-levko: [user.is_disabled](https://github.com/getredash/redash/blob/master/redash/models.py#L446)

tests/test_authentication.py
```diff
@@ -54,6 +55,14 @@ def test_user_api_key(self):
             rv = c.get(self.queries_url, query_string={'api_key': user.api_key})
             self.assertEqual(user.id, api_key_load_user_from_request(request).id)
 
+    def test_disabled_user_api_key(self):
+        user = self.factory.create_user(api_key="user_key")
+        user.disabled_at = utcnow()
```
kravets-levko: [user.disable()](https://github.com/getredash/redash/blob/master/redash/models.py#L449)

arikfr:
💃

PR #3163, opened 2018-12-04
Include correct version in production builds

rauchy:
This is my hypotheses on why production builds (i.e. demo.redash.io) are not showing correct frontend versions - `npm run build` needs to run after `update_version.sh`.


arikfr:
yes, this makes sense 👍

PR #3168, opened 2018-12-05
Add tag management commands

ariarijp:
Hi, I've added two commands.

These commands can add and remove tag from existing query via CLI. 

Usage:
- `./manage.py queries add_tag 1 foo`
- `./manage.py queries remove_tag 1 foo`


arikfr:
:+1:

Issue #2146, opened 2017-12-10
Cannot use html tag in description @version 3.0.0


orkShinnosuke:
### Issue Summary

Cannot use html tag in description

<img width="520" alt="2017-12-10 16 05 07" src="https://user-images.githubusercontent.com/3303281/33802824-eaffb76e-ddc3-11e7-93e9-2fb124d204c0.png">

P.S.
also cannot use in dashboards

### Steps to Reproduce

1. Look at below link
http://demo.redash.io/queries/5873/source
http://demo.redash.io/dashboard/test_1103

### Technical details:

* Redash Version: 3.0.0
* Browser/OS: google chrome/macOS Sierra(10.12.6)

kyoshidajp:
I figure HTML tag isn't be able to use after v3.0.0.

Markdown library has changed from [marked](https://www.npmjs.com/package/marked) to [markdown](https://www.npmjs.com/package/markdown) by #1966. markdown package do sanitize HTML, and can't disable it.

Actually, the PR disable sanitize option is already sent by evilstreak/markdown-js#132 in many years ago. However, it's not merged yet.

Why don’t you use markdown format?

arikfr:
@kyoshidajp is correct - this is due to the change in Markdown parser and we don't expect to bring this back (as this was never the intended behavior).

Just use Markdown.

orkShinnosuke:
@kyoshidajp @arikfr 
Thanks! I'll use Markdown

PR #3170, opened 2018-12-07
Change markdown module (markdown 0.5.0 to marked 0.5.2)

44int:
"markdown 0.5.0" will escape the entered HTML.
Therefore, even if "REDASH_ALLOW_SCRIPTS_IN_USER_INPUT = true" is enabled, it will not be rendered as intended.
By using marked 0.5.2, the above problem can be avoided and markdown can also be made effective.

Related pull requests are as follows.
https://github.com/getredash/redash/pull/2438

Marked documents
https://marked.js.org/#/USING_ADVANCED.md


deecay:
Related to #1966, #2146 ? Is vulnerability in marked resolved?

44int:
@deecay 
Thank you for finding related issues.
> Is vulnerability in marked resolved? 
Yes, i think resolved in 0.3.9. 
https://github.com/markedjs/marked/issues/997
Also, since "marked" is being maintained continuously, I think that there is no problem in use.

> I do not believe HTML will be allowed anyways?
No. It will not be effective unless the environment variable "REDASH_ALLOW_SCRIPTS_IN_USER_INPUT" is set to "true".
Default is false.


If you want to do the same as existing display, change to the following code.
what will you do?
```
#client/app/filters/markdown.js



```

arikfr:
Hi,

Thanks.

PR #3171, opened 2018-12-08
FIX: Reject empty query name

ariarijp:
Hi, folks

We can set empty query name for now on master branch.

I think this bug comes from migration from Angular to React.

The following link is previous version of implementation.

https://github.com/getredash/redash/blob/v5.0.2/client/app/components/edit-in-place.js#L62

Thanks,


arikfr:
Thanks!

PR #3172, opened 2018-12-08
Refactor alert add/edit page

yoavbls:
After:
image

Before: 
image

The title editable directly (like in dashboards and queries)
and I think its simpler and consistent in this form.
Its also could be better for future alert options like this one: #3137 
I Would like to get feedback 🙂


arikfr:
Thank you for the suggested updates. This page will undergo UX refactoring as part of the React migration.

PR #3173, opened 2018-12-09
Make it possible to move up one line by Ctrl+P on macOS

kyoshidajp:
Ctrl+P in Query Editor, the cursor will move up one line.

Other directions keyboard shortcuts are already available.

Ref: [Mac keyboard shortcuts \- Apple Support](https://support.apple.com/en-us/HT201236)
- Control-F: Move one character forward.
- Control-B: Move one character backward.
- Control-N: Move down one line.

I have only tested this feature on Chrome with macOS.

### DIFF
--- a/client/app/components/QueryEditor.jsx
+++ b/client/app/components/QueryEditor.jsx
@@ -104,7 +104,8 @@ class QueryEditor extends React.Component {
       editor.commands.bindKey('Cmd+L', null);
       editor.commands.bindKey('Ctrl+L', null);
 
-      editor.commands.bindKey('Ctrl+P', 'golineup');
+      // golineup only mac
+      editor.commands.bindKey({ win: null, mac: 'Ctrl+P' }, 'golineup');
 
       // eslint-disable-next-line react/prop-types
       this.props.QuerySnippet.query((snippets) => {
--- a/client/app/components/QueryEditor.jsx
+++ b/client/app/components/QueryEditor.jsx
@@ -102,8 +102,9 @@ class QueryEditor extends React.Component {
     this.onLoad = (editor) => {
       // Release Cmd/Ctrl+L to the browser
       editor.commands.bindKey('Cmd+L', null);
-      editor.commands.bindKey('Ctrl+P', null);
       editor.commands.bindKey('Ctrl+L', null);
+
+      editor.commands.bindKey('Ctrl+P', 'golineup');
 
       // eslint-disable-next-line react/prop-types
       this.props.QuerySnippet.query((snippets) => {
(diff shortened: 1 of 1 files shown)


gabrieldutra:
I tested it with a Linux notebook, the behavior was the same:
ctrl p

I usually do it with arrows haha.

Well, for `CTRL+P` there is no conflict, but for other keys that may have conflict you can use:
```javascript
editor.commands.bindKey({ win: null, mac: 'Ctrl+N' }, 'golinedown');
```

arikfr:
The reason we unbind Ctrl-P is because we use it to trigger new parameter dialog on Linux and Windows. We could unbind it only on these platforms and keep default behavior on MacOS, but I'm curious: in what cases do you use Ctrl-P instead of the up arrow?

kyoshidajp:
Thanks. My finger position is used to Ctrl-P to go up the line. I rarely use arrow keys.
I think this is not a rare case for people who use macOS.

kyoshidajp:
OK. I'm satisfied if it works on macOS.

client/app/components/QueryEditor.jsx
```diff
@@ -102,9 +102,11 @@ class QueryEditor extends React.Component {
     this.onLoad = (editor) => {
       // Release Cmd/Ctrl+L to the browser
       editor.commands.bindKey('Cmd+L', null);
-      editor.commands.bindKey('Ctrl+P', null);
```
gabrieldutra: I forgot about the Parameter dialog before :sweat_smile:. I tested here again, if this line is missing, it stops working...
Either keep this line or using `editor.commands.bindKey({ win: 'Ctrl+P', mac: null }, null);` make the dialog shortcut functional again.

client/app/components/QueryEditor.jsx
```diff
@@ -102,9 +102,11 @@ class QueryEditor extends React.Component {
...
       editor.commands.bindKey('Ctrl+L', null);
 
+      // golineup only mac
+      editor.commands.bindKey({ win: null, mac: 'Ctrl+P' }, 'golineup');
```
arikfr: I think you still need to unbind it for Windows, to allow for our shortcut to take over.

kyoshidajp:
@arikfr Thanks. I fixed it for Windows and Linux. Please check again.

client/app/components/QueryEditor.jsx
gabrieldutra: I was thinking, this line becomes unnecessary once you unbind the key only for Windows, since golineup is the [default behavior](https://github.com/ajaxorg/ace/wiki/Default-Keyboard-Shortcuts) for mac :thinking:

kyoshidajp: I think so, too. But it didn't work when I try.

gabrieldutra: pretty weird though :open_mouth:... Anyway, the parameter dialog is working with Ctrl+P in my Linux :+1:

arikfr:
Thanks!

PR #3174, opened 2018-12-09
Invalidate sessions after email or password change

rauchy:
Previously, when you changed your e-mail or password, all your existing session remained active. This change invalidates other sessions (which means that the session you use to change your e-mail or password will remain active, but if you are logged in through any other browser, you would get signed out).

Remaining work:

- [x] fix build
- [x] add test for session invalidation
- [x] backward compatibility (ensure existing sessions aren't invalidated when this rolls out)
- [x] test for backward compatibility
- [x] attempt to minimize the solution sprawl


redash/handlers/users.py
```diff
@@ -194,6 +195,16 @@ def post(self, user_id):
...
             self.update_model(user, params)
             models.db.session.commit()
+
+            @after_this_request
+            def update_identity_cookie(response):
+                session['user_id'] = user.get_id()
+
+                remember_token = request.cookies.get('remember_token')
+                if remember_token:
+                    response.set_cookie('remember_token', re.sub('.*\\|', user.get_id() + '|', remember_token))
```
arikfr: I think that you can simply call `login_user` here instead of manually updating things. It will take care of updating the `user_id` and the remember me cookie.

redash/authentication/__init__.py
```diff
@@ -40,12 +40,23 @@ def sign(key, path, expires):
 
 
 @login_manager.user_loader
-def load_user(user_id):
+def load_user(user_id_with_identity):
```
arikfr: Worth adding a comment here explaining the legacy support reasoning.

redash/handlers/users.py
rauchy: Good point regarding updating the session. However with this approach the `remember_token` value is not updated. Not sure what the implications of this though.

redash/handlers/users.py
```diff
@@ -194,6 +195,7 @@ def post(self, user_id):
         try:
             self.update_model(user, params)
             models.db.session.commit()
+            login_user(user, remember=True)
```
arikfr: Worth adding a comment as to why we call this here.

redash/models.py
```diff
@@ -553,6 +553,10 @@ def update_group_assignments(self, group_names):
     def has_access(self, obj, access_type):
         return AccessPermission.exists(obj, access_type, grantee=self)
 
+    def get_id(self):
+        identity = hashlib.md5("{0},{1}".format(self.email, self.password_hash)).hexdigest()
```
arikfr: I wonder if in some way we're revealing something about the password in this way? Maybe worth salting this? Maybe using the API key?

rauchy: Will salting really help here? I mean, the `password_hash` already acts as a salt for the email address. I don't mind appending the api key here, but I'm curious if it really adds anything.

arikfr: You're right. 👍

arikfr:
👌

rauchy:
@arikfr is it time to remove [this](https://github.com/getredash/redash/blob/823f172a9fde5170af0c137184bf135e4f45208f/redash/authentication/__init__.py#L47)?

PR #3175, opened 2018-12-09
Fix disable error message

rauchy:
Before:

image

After:

image

Also, I changed the status code to 403 ("The request was valid, but the server is refusing action. The user might not have the necessary permissions for a resource, or may need an account of some sort."), which feels to me more suitable than a 400 ("The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing)")


rauchy:
image

arikfr:
👍

PR #3180, opened 2018-12-11
Remove missing coverage from pytest terminal output

rauchy:
Don't know about you, but when I run pytest, I don't feel like seeing a wall of text with missing coverage is insightful in any way. I'd rather cleanly see the test output. WDYT?


arikfr:
I was thinking of stopping the code coverage report run by default and only using in in CI. WDYT?

jezdez:
Sure that works for me. It's still possible to see this by running `coverage report` after all.

arikfr:
Perfect.

PR #3199, opened 2018-12-17
Upgrade Node10

koooge:
redash/base is using node 6 now, but it will be EOL in 2019/4 (https://github.com/nodejs/Release#readme).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,10 @@ WORKDIR /frontend
 COPY package.json package-lock.json /frontend/
 RUN npm install
 
+COPY bin/bundle-extensions /frontend/bin/bundle-extensions
+RUN apt-get update -qq && apt-get install -y python-pip && pip install -U setuptools==23.1.0
+RUN npm run bundle
+
 COPY . /frontend
 RUN npm run build
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,8 @@
+FROM node:10.14.2 as builder
+
+COPY . /app
+RUN npm install && npm run bundle && npm run build && rm -rf node_modules
+
 FROM redash/base:latest
 
 # Controls whether to install extra dependencies needed for all data sources.
@@ -9,8 +14,7 @@ COPY requirements.txt requirements_dev.txt requirements_all_ds.txt ./
 RUN pip install -r requirements.txt -r requirements_dev.txt
 RUN if [ "x$skip_ds_deps" = "x" ] ; then pip install -r requirements_all_ds.txt ; else echo "Skipping pip install -r requirements_all_ds.txt" ; fi
 
-COPY . /app
-RUN npm install && npm run bundle && npm run build && rm -rf node_modules
+COPY --from=builder /app /app
 RUN chown -R redash /app
 USER redash
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ COPY requirements.txt requirements_dev.txt requirements_all_ds.txt ./
 RUN pip install -r requirements.txt -r requirements_dev.txt
 RUN if [ "x$skip_ds_deps" = "x" ] ; then pip install -r requirements_all_ds.txt ; else echo "Skipping pip install -r requirements_all_ds.txt" ; fi
 
-COPY . ./
+COPY . /app
 RUN npm install && npm run bundle && npm run build && rm -rf node_modules
 RUN chown -R redash /app
 USER redash
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,11 @@
-FROM node:10.14.2 as builder
+FROM node:10 as frontend-builder
 
-COPY . /app
-RUN npm install && npm run bundle && npm run build && rm -rf node_modules
+COPY package.json package-lock.json webpack.config.js /frontend/
+COPY client/ /frontend/client
+WORKDIR /frontend
+
+# TODO: RUN npm run bundle
+RUN npm install && npm run build
 
 FROM redash/base:latest
 
@@ -14,7 +18,9 @@ COPY requirements.txt requirements_dev.txt requirements_all_ds.txt ./
 RUN pip install -r requirements.txt -r requirements_dev.txt
 RUN if [ "x$skip_ds_deps" = "x" ] ; then pip install -r requirements_all_ds.txt ; else echo "Skipping pip install -r requirements_all_ds.txt" ; fi
 
-COPY --from=builder /app /app
+COPY . /app
+RUN rm -rf /app/client/*
+COPY --from=frontend-builder /frontend/client/dist /app/client/dist
 RUN chown -R redash /app
 USER redash
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,6 @@ WORKDIR /frontend
 COPY package.json package-lock.json /frontend/
 RUN npm install
 
-COPY bin/bundle-extensions /frontend/bin/bundle-extensions
-RUN apt-get update -qq && apt-get install -y python-pip && pip install -U setuptools==23.1.0
-RUN npm run bundle
-
 COPY . /frontend
 RUN npm run build
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,10 @@ COPY package.json package-lock.json webpack.config.js /frontend/
 COPY client/ /frontend/client
 WORKDIR /frontend
 
-# TODO: RUN npm run bundle
+COPY bin/bundle-extensions /frontend/bin/bundle-extensions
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,3 +1,8 @@
+FROM node:10.14.2 as builder
```
arikfr: Maybe `node:10` instead? Seem more future proof, no?

Dockerfile
```diff
@@ -1,3 +1,8 @@
...
+
+COPY . /app
+RUN npm install && npm run bundle && npm run build && rm -rf node_modules
```
arikfr: Let's split this into multiple steps, so we can benefit from Docker's layer caching when package.json doesn't change:

```
COPY package.json  package-lock.json ./
RUN npm install
COPY . ./
RUN npm run build
```

arikfr: There is an issue here though: `npm run bundle` needs the Python packages to really work. Right now we don't load any extensions in the default build, so it's a non issue, but this process will not work for people who use the codebase from extensions.

@emtwo @jezdez this is another case for moving the frontend extensions to use a frontend package manager rather than rely on pip.

Dockerfile
```diff
@@ -9,10 +14,9 @@ COPY requirements.txt requirements_dev.txt requirements_all_ds.txt ./
 RUN pip install -r requirements.txt -r requirements_dev.txt
 RUN if [ "x$skip_ds_deps" = "x" ] ; then pip install -r requirements_all_ds.txt ; else echo "Skipping pip install -r requirements_all_ds.txt" ; fi
 
-COPY . ./
-RUN npm install && npm run bundle && npm run build && rm -rf node_modules
+COPY --from=builder /app /app
```
arikfr: Let's be more specific here and copy only the client directory:

```
COPY --from=builder /app/client /app/client
```

Dockerfile
arikfr: Also, let's name it something more descriptive, like frontend-builder?

package.json
```diff
@@ -21,8 +21,8 @@
     "url": "git+https://github.com/getredash/redash.git"
   },
   "engines": {
-    "node": "^8.0.0",
-    "npm": "^5.0.0"
+    "node": "^10.14.2",
+    "npm": "^6.4.1"
```
arikfr: Maybe worth keeping previous versions for now? Otherwise existing users will need to upgrade?

Dockerfile
jezdez: Yeah, good example. I've looked into plugin patterns in Node land a bit and as always there are many answers to that question. Some examples I found interesting (and very much not exhausting the choices):

- https://github.com/nfarina/homebridge/blob/master/lib/plugin.js an example of a homegrown plugin system that looks for node modules with the "homebridge-" name prefix and loads them

- https://github.com/c9/architect a low-level "everything is a plugin" API that allows flexible customization and simple, native JS syntax, using package.json for configuration ([blogpost](https://medium.com/@ohbytheway/architecture-using-architect-8c8bb7d0277a))

- https://hapijs.com/tutorials/plugins another example of a homegrown system, this time not using the package.json API by default, and with the need to [manually](https://github.com/hapijs/hapi-auth-basic/blob/b73ab49bca00a4a5df1d2172d6cc062f7d56a82c/test/index.js#L26) [register the plugin](https://github.com/hapijs/hapi/blob/ae0255550a6c537f46eaee71b49bbcf3520a8bb2/lib/server.js#L363) (by default at least)

jezdez: I would suggest to reach out to the community and ask for people that may have more experience writing system that are extensible via plugins/extensions. I'm worried that I'm missing the elephant in the room to be honest 😬

package.json
koooge: I will revert this. Actually, these don't have to be updated, babel would transpile code well

Dockerfile
koooge: Okay, I feel `node:10` is easy to use under development

Dockerfile
koooge: Thank you for your reviews, and I'm sorry I have no good idea about `npm run bundle`!
So, I add installing step of pip and setuptools tentatively in frontend-builder. :pray:

arikfr: Thank you for the pointers, @jezdez. I will open an issue to bootstrap the discussion.

@koooge sorry for the mixed signals here: I wanted to point out the issue, but I didn't expect you to address it in this pull request. 

I think it's not worth to install the extra stuff for something we currently don't use. The only use of `npm run bundle` is in [mozilla/redash](https://github.com/mozilla/redash) and they have their own Dockerfile...

Dockerfile
```diff
@@ -9,10 +21,11 @@ COPY requirements.txt requirements_dev.txt requirements_all_ds.txt ./
 RUN pip install -r requirements.txt -r requirements_dev.txt
 RUN if [ "x$skip_ds_deps" = "x" ] ; then pip install -r requirements_all_ds.txt ; else echo "Skipping pip install -r requirements_all_ds.txt" ; fi
 
-COPY . ./
-RUN npm install && npm run bundle && npm run build && rm -rf node_modules
+COPY . /app
+RUN rm -rf /app/client/*
```
arikfr: I think this is not necessary. The COPY from the frontend-builder will override this anyway.

Dockerfile
```diff
@@ -1,3 +1,15 @@
...
+
+COPY package.json package-lock.json webpack.config.js /frontend/
+COPY client/ /frontend/client
+WORKDIR /frontend
+
+COPY bin/bundle-extensions /frontend/bin/bundle-extensions
+RUN apt-get update -qq && apt-get install -y python-pip && pip install -U setuptools==23.1.0
+RUN npm run bundle
+
+RUN npm install && npm run build
```
arikfr: It's better to split `npm install` from the build, so if there was no change in packages, it will cache this layer.

arikfr: Also the COPY of the non package files should happen between install and build (for similar caching reasons).

Dockerfile
```diff
@@ -1,3 +1,15 @@
+FROM node:10 as frontend-builder
+
+COPY package.json package-lock.json webpack.config.js /frontend/
+COPY client/ /frontend/client
+WORKDIR /frontend
+
+COPY bin/bundle-extensions /frontend/bin/bundle-extensions
+RUN apt-get update -qq && apt-get install -y python-pip && pip install -U setuptools==23.1.0
+RUN npm run bundle
```
arikfr: See my other comment: https://github.com/getredash/redash/pull/3199#discussion_r242450451.

Dockerfile
```diff
@@ -1,3 +1,15 @@
+FROM node:10 as frontend-builder
+
+COPY package.json package-lock.json webpack.config.js /frontend/
+COPY client/ /frontend/client
```
arikfr: Maybe copy the whole path, just to be on the safe side? It's a temporary container anyway.

Dockerfile
koooge: If this statement is not necessary, is it okay to remove?

Dockerfile
koooge: `/app` seems a little bit sensitive because `/app` is mount `.` on host. I tried `npm install` into /app in frontend-builder, but it disappeared in the resulting container. I'd like to talk about this in #3201

koooge:
sorry, I misunderstand. You mean it is okay to delete `npm run bundle` in Dockerfile only?

arikfr:
@koooge yes, I meant to remove it in the Dockerfile only.

docker-compose.yml
```diff
@@ -12,6 +12,7 @@ services:
       - "5000:5000"
     volumes:
       - ".:/app"
+      - /app/client/dist
```
arikfr: What's this for?

koooge: it needs for /dist created by `npm run build`. Please refer to **The node_modules Volume Trick** chapter on https://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html

Dockerfile
arikfr: @jezdez #3207

docker-compose.yml
arikfr: The way this `docker-compose.yml` file was designed to work is with the user running npm related stuff on the host, so `client/dist` will be part of the mounted volume.

See: https://redash.io/help/open-source/dev-guide/docker for details.

koooge: well.. we seem to need more hack for this point 🤔

koooge: Wait, I just realized /dist created by `npm run build` in Dockerfile will be overwritten by volume `.:/app`. So, `npm run build` is not necessary under development, right? 
Btw, is this Dockerfile used in production? (IOW, is this Dockerfile of redash/redash?)

arikfr: > So, npm run build is not necessary under development, right?

Correct.

PR #3201, opened 2018-12-18
Stop to mount volume

koooge:
Hi there, this volume mounting dirty both own dir and container without I knew it. It confuses me a bit. I'd like to clean own working dir.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,8 +10,6 @@ services:
       - redis
     ports:
       - "5000:5000"
-    volumes:
-      - ".:/app"
     environment:
       PYTHONUNBUFFERED: 0
       REDASH_LOG_LEVEL: "INFO"
@@ -20,8 +18,6 @@ services:
   worker:
     build: .
     command: scheduler
-    volumes:
-      - ".:/app"
     depends_on:
       - server
     environment:
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ create_database:
 
 clean:
 	docker-compose down && docker-compose rm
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -10,8 +10,6 @@ services:
       - redis
     ports:
       - "5000:5000"
-    volumes:
-      - ".:/app"
```
jezdez: Forgive me if I'm missing something, but removing this means that changes on the host machine won't be reflected in the running container, right? Since that's a central part of the developing workflow, to be able to change files on the host machine without having the rebuild the Docker image.

koooge: Yes, it would force you to rebuild docker. I was not sure you don't have to restart docker container when you edit code. I thought frontend code needs to run `npm run build`, and backend code needs to restart server(So it means almost the same as rebuild docker container, right?)

I know what restart docker container others for us. But It seems that this statement occurs some problem which overwrites `/app` directory on docker build. I will grasp this point.

arikfr: This docker-compose file is for the development environment, in this environment you will either use `npm run watch` or `npm start` -- both which auto build/reload the frontend code along with `dev_server` command of the Docker container, which auto reloads the web server on code changes.

This is why we mount the application code here. You can see the [production docker compose file](https://github.com/getredash/redash/blob/master/setup/docker-compose.yml), which doesn't do this.

koooge: Okay, I see. I just thought of deleting this for understandability if it's not necessary much. 🤝

PR #3206, opened 2018-12-19
Show Sort Y only when type is Heatmap

### LINKED ISSUE #3204: "Sort Y" should be shown only for Heatmap type charts
(or we need to support it for the other chart types)

deecay:
fixes #3204.

Could not think of other cases where y-axis values that benefit from sorting. 

We should come back when we think about horizontal bar charts and horizontal box plots, etc.


arikfr:
Thanks!

Issue #3179, opened 2018-12-10
Migrate DyanmicForm to React


arikfr:
- [ ] `<DynamicForm>` component in React
- [ ] Update usage (for Alert Destination and Data Sources)

PR #3209, opened 2018-12-20
Migrate DynamicForm to React

gabrieldutra:
## Description
This work is related to #3179. It creates a React DynamicForm component to replace the angular version of it.

## Component docs
The general idea of the React `DynamicForm` component is that you can create any form based on a list of fields (specified below).

I tried to create its props in an easy way to reuse along the code (you pass fields and a submit action, you have a form :ok_hand:).

### Props
| Property | Type | Required | Default |
| --- | --- | --- | --- |
| fields | `Field[]` | false | `[]` |
| actions | `Action[]` | false | `[]` |
| onSubmit | `Function(values: Object, success: Function, error: Function)` | false | `() => {}` |

### Field Prop Type
| Property | Type | Required |
| --- | --- | --- |
| name | `string` | true |
| title | `string` | false |
| type | one of `['text', 'email', 'password', 'number', 'checkbox', 'file']` | true |
| initialValue | `string`, `number` or `bool` | false |
| required | `bool` | false |
| placeholder | `string` | false |

### Action Type
| Property | Type | Required |
| --- | --- | --- |
| name | `string` | true |
| callback | `Function(onFinish: Function())` | true |
| class | `string` | false |
| disabledWhenDirty | `bool` | false |

## Changes
- [x] start creating the component :sweat_smile: 
- [x] render fields
- [x] understand and apply antd `getFieldDecorator` concept (+-)
- [x] save button
- [x] file inputs
- [x] render actions
- [x] create a transition angular1 -> React (basically adaptating props) (worked pretty well with a React component to "translate props)
- [x] adjust UI stuff to fit Redash
  - ~it seems to be showing 2 step buttons on InputNumber (browser and antd)~ :thinking: 
  - ~decide on which notification box to use~
- [ ] review and improve code stuff

### PostgreSQL
postgresql

### BigQuery (file upload)
bigquery

I think there are still improvements that can be made on this, but we already have a component that can be delivered.


gabrieldutra:
So far I could not use **toastr** in React (I haven't tried hard that). I found 2 other options with Antd and I thought it might be worthy of a discussion.

1. Find a way to use **toastr** (perhaps install react-toastr)
toastr

2. Use antd's message
antd-message

3. Use antd's notification
antd-notification


We also need to consider that this would be a change and most of Redash uses the first one (54 uses to be more accurate :joy:)

Well, let me know what you think :ok_hand:

arikfr:
I think you should be able to use toastr in React, in a similar way as I exposed the http service: 
https://github.com/getredash/redash/blob/master/client/app/services/http.js

If this doesn't work, then we can use Ant's notification.

gabrieldutra:
> I think you should be able to use toastr in React, in a similar way as I exposed the http service:
> https://github.com/getredash/redash/blob/master/client/app/services/http.js
> 
> If this doesn't work, then we can use Ant's notification.

Yep, that worked fine, we will have to update those notifications at some point, but it's better to change them all together, rather than keeping different styles around the code.

client/app/assets/less/redash/ant.less
```diff
@@ -36,3 +42,35 @@
 .ant-dropdown-in-bootstrap-modal {
   z-index: 1050;
 }
+
+// Make Ant labels more Redash like
+.ant-form-item-label label {
+  color: inherit;
+  white-space: initial;
+  line-height: initial;
+  text-align: left;
```
ranbena: You might want to either use `direction: rtl` instead (to flip the required mark location) or maybe ant-form's `layout="vertical"` (didn't try but sounds like it might work).

client/app/assets/less/redash/ant.less
```diff
@@ -36,3 +42,35 @@
 .ant-dropdown-in-bootstrap-modal {
   z-index: 1050;
 }
+
+// Make Ant labels more Redash like
+.ant-form-item-label label {
+  color: inherit;
```
ranbena: Why is it set to white in the first place?

client/app/assets/less/redash/ant.less
```diff
@@ -36,3 +42,35 @@
...
+  white-space: initial;
```
ranbena: What's this for?

client/app/assets/less/redash/ant.less
```diff
@@ -36,3 +42,35 @@
...
+  line-height: initial;
+  text-align: left;
+}
+
+// Fix ant input number showing duplicate arrows
+.ant-input-number-input::-webkit-outer-spin-button,
+.ant-input-number-input::-webkit-inner-spin-button {
+  -webkit-appearance: none;
```
ranbena: Looking at the [InputNumer](https://ant.design/components/input-number/) page on chrome, I don't see a duplicate button. Then why is this happening in Redash?

client/app/assets/less/redash/ant.less
```diff
@@ -36,3 +42,35 @@
...
+  &[disabled] {
+    color: @white !important;
+  }
+}
+
+// Remove green border from Ant inputs
+.has-success .ant-input,
+.has-success .ant-input:hover {
+  border-color: @border-color-base !important;
+}
```
ranbena: (I'm new to Redash code so I got some n00b questions)
What does `.has-success` mean in the form? Where is it coming from?

client/app/assets/less/redash/ant.less
```diff
@@ -36,3 +42,35 @@
...
+.ant-input-number-input {
+  -moz-appearance: textfield;
+}
+
+// Fix for text becoming black on disabled buttons
+.btn-danger, .btn-primary, .btn-success, .btn-info {
+  &[disabled] {
+    color: @white !important;
+  }
+}
```
ranbena: Can this be avoided by converting the related buttons from bootstrap to Ant?

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -0,0 +1,228 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Form, Input, InputNumber, Checkbox, Button, Upload, Icon } from 'antd';
```
ranbena: Unfortunately [this](https://github.com/getredash/redash/pull/3247#discussion_r245055244) 😓

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -0,0 +1,228 @@
...
+
+    this.state = {
+      isSubmitting: false,
+      inProgressActions: [],
+    };
+
+    this.actionCallbacks = this.props.actions.reduce((acc, cur) => ({
+      ...acc,
+      [cur.name]: cur.callback,
+    }), []);
```
ranbena: Nice trick :)
Btw, the `[]` argument should be `{}` or even better `null`.

client/app/assets/less/redash/ant.less
ranbena: Gotcha. I agree. 
Add this attribute `<Form layout="vertical">` and you won't need a css override.

client/app/assets/less/redash/ant.less
gabrieldutra: `.has-success` is added in [Ant's Form validation](https://ant.design/components/form/). For some reason the css for this came with a green border and I did it as an overwrite for this.

client/app/components/dynamic-form/DynamicForm.jsx
gabrieldutra: I saw it a few days ago :confused:, anyway I'll separate those...

client/app/assets/less/redash/ant.less
ranbena: That's weird..
According to [their less file](https://github.com/ant-design/ant-design/blob/4378454495a9c82543036c57cd10ef92ec08948f/components/form/style/index.less#L434), it should color only the success icon. Could it be our Ant package is not up to date?

Might want to research this more cause it seems the problem still persists (taken from [current netlify preview](https://deploy-preview-3209--redash-preview.netlify.com/data_sources/new):
screen shot 2019-01-05 at 20 58 11

gabrieldutra: I missed the color for `:hover` :man_facepalming: (should be blue instead). Anyway, I think the best option is to update our Ant package and see how it affects elsewhere. I remember I could find it somewhere in one of their `less` files located in `node_modules`...

client/app/components/dynamic-form/DynamicForm.jsx
gabrieldutra: Good catch, for some reason I thought with `null` this wouldn't work...

client/app/assets/less/redash/ant.less
gabrieldutra: Might be related to our `antd` version, I'll test it once I try to update our version

client/app/assets/less/redash/ant.less
gabrieldutra: To avoid these cuts with long labels (happens in their docs site if you inspect and change a form):
vertical-cut

client/app/assets/less/redash/ant.less
gabrieldutra: No idea :confused:. In `antd` files it comes from `@label-color: @heading-color;` (but should be `fade(#000, 85%)`, not white). Anyway, I'll see what happens after update `antd`.

client/app/assets/less/redash/ant.less
gabrieldutra: I think so, they are actually using Ant, however the color is still given through a `class` (bootstrap format) argument from the object. Basically what's needed is to rethink this (Data-sources angular page): 
```javascript
$scope.actions = [
    { name: 'Delete', class: 'btn-danger', callback: deleteDataSource },
    {
      name: 'Test Connection', class: 'btn-default pull-right', callback: testConnection, disableWhenDirty: true,
    },
  ]
```
Then we would either need to map this in DynamicForm Transition or actually change it.

My first idea on that was to simply use css to fix it till we actually migrate the parent. But thinking better, this would not be the best since I would be generating some debt :thinking:.

client/app/assets/less/redash/ant.less
gabrieldutra: Unfortunately updating Ant was not the solution for any css problem :confused:. I'll research a bit more tomorrow to at least fix the focused and hovered border.

ranbena: I'll take a deeper look if that's ok with you.

Also, I think it is worth upgrading the package anyway in a separate PR.

client/app/assets/less/redash/ant.less
gabrieldutra: I agree, that's certainly sth I would change when handling the Pages.

client/app/assets/less/redash/ant.less
gabrieldutra: Go ahead and let me know 🙂. If I find out anything I'll mention as well.

It's of course better to find the root problem for this rather than just css it.

client/app/assets/less/redash/ant.less
ranbena: I think this should do the trick:
```js
// show.js
$scope.actions = [
    { name: 'Delete', type: 'danger', callback: deleteDataSource },
    …
]
```
```jsx
// DynamicForm.jsx
renderActions() {
  return this.props.actions.map((action) => {
    …
    const actionProps = {
      key: action.name,
      htmlType: 'button',
      className: `${action.class} m-t-10`,
      type: action.type, // <--- defaults to 'default'
      …
    };

    return (<Button …>…</Button>);
  });
}
```

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -0,0 +1,234 @@
...
+    return this.props.fields.map((field) => {
+      const [firstItem] = this.props.fields;
+      const FormItem = Form.Item;
+      const { name, title, type } = field;
+      const fieldLabel = title || helper.toHuman(name);
+
+      const formItemProps = {
+        key: name,
+        className: 'm-b-10',
+        hasFeedback: false,
```
ranbena: This could be a kindova cool little feature, nay?

ranbena:
@gabrieldutra Some UX issues:
1. Field's placeholder value also acts as the field's initial value - should instead be left empty. This causes UX problems:
  -  user needs to first clear the field before entering a value.
  -  `.has-success` classname gets added automatically (making all fields green bordered) since its validator is immediately fulfilled.
2. The field text color is set to `@gray` which makes it undistinguishable from the placeholder text. I think Ant's default should be `rgba(0,0,0,0.65)`.

client/app/assets/less/redash/ant.less
ranbena: Plz add a comment to remove this line when [this PR](https://github.com/ant-design/ant-design/pull/14135) lands.

client/app/assets/less/redash/ant.less
ranbena: Or perhaps just give it the class `ant-btn-danger` instead of `btn-warning`?

gabrieldutra: Well, remove the `class` prop was already in my plans, so I used a `type` and a `purgeRight` prop. However, my guess was wrong and the color for disabled danger buttons keeps turning to black...

ranbena:
@gabrieldutra @arikfr Finally found the reason for the weird Ant color issues - Ant and  Bootstrap have many `less` variable names in common and the two clash. Changing order in `main.less` fixes all issues but presents others.

Perhaps we should split the implementation efforts of React and Ant.

arikfr:
@ranbena 

> Ant and Bootstrap have many less variable names in common and the two clash. 

😱 

> Perhaps we should split the implementation efforts of React and Ant.

You mean first migrate to Bootstrap and then migrate to Ant?

ranbena:
> You mean first migrate to Bootstrap and then migrate to Ant?

I mean that React migration commits shouldn't include Ant migration efforts.
In parallel, we need to make a cross-product Ant migration sweep and remove Bootstrap completely.

arikfr:
> I mean that React migration commits shouldn't include Ant migration efforts.

I'm not sure this is possible. For some of the migrations, we need UI widgets. This means we will either have to port JS logic from Angular or utilize Ant. Utilizing Ant feels like a more future proof strategy.

> In parallel, we need to make a cross-product Ant migration sweep and remove Bootstrap completely.

This sounds good, but has the issue that it requires that all of our code will be React based, as Ant has no Angular parallel.

🤔

gabrieldutra:
> @gabrieldutra @arikfr Finally found the reason for the weird Ant color issues - Ant and Bootstrap have many `less` variable names in common and the two clash. Changing their order in `main.less` fixes all issues but presents others.
Nice that you found and not so nice we have it :confused:. I knew there could be css conflicts from different styling libs, but variable names were not in my thoughts. Changing their order fixes the issues related, as well as importing the css library instead of less for the components.

ranbena:
> importing the css library instead of less for the components.

You mean like this?

```js
// DynamicForm.jsx
import React from 'react';
import Form from 'antd/lib/form';
…

// import ant component style
import 'antd/lib/input/style/css';
import 'antd/lib/input-number/style/css';
import 'antd/lib/checkbox/style/css';
import 'antd/lib/button/style/css';
import 'antd/lib/upload/style/css';
import 'antd/lib/icon/style/css';
```

Indeed this seems to work.

gabrieldutra:
> You mean like this?

Yes, no variables, no variable conflict :joy:, that could be in `ant.less` as well:
```css
/* ant.less */
...
@import '~antd/lib/form/style/index.css';
@import '~antd/lib/button/style/index.css';
...
```
However, we lose some control on Ant's by doing it.

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -0,0 +1,238 @@
...
+  renderFields() {
+    return this.props.fields.map((field) => {
+      const [firstItem] = this.props.fields;
+      const FormItem = Form.Item;
+      const { name, title, type } = field;
+      const fieldLabel = title || helper.toHuman(name);
+
+      const formItemProps = {
+        key: name,
+        className: 'm-b-10',
```
gabrieldutra: If we are not gonna keep bootstrap, I guess those classes shouldn't be used

ranbena:
> Yes, no variables, no variable conflict 😂, that could be in `ant.less` as well:

In `ant.less` sounds good to me. Let's wrap this up!

client/app/assets/less/redash/ant.less
ranbena: The change made it to [antd@3.12.0](https://github.com/ant-design/ant-design/releases/tag/3.12.0). Consider upgrading ⬆️💯

gabrieldutra: I upgraded locally to `3.12.1` and it didn't work. Also, I didn't see your PR mentioned in the changelog, so I guess it hasn't been released yet :confused:

ranbena: > and it didn't work

Do you mean the upgrade didn't work? Or the `white-space` fix?

gabrieldutra: > It should be there. Here's the real changelog ant-design/ant-design@3.12.0...master. I'll take a look and see why it's not applying.

At first I thought that as well. However, this is the exactly opposite, this is a comparison between 3.12.0 and master, so all the stuff on that link are not in this version.

ranbena: Ok ok I'm in! [3.12.2](https://github.com/ant-design/ant-design/releases/tag/3.12.2)

gabrieldutra: haha, nice! :grin:

package.json
```diff
@@ -46,7 +46,7 @@
     "angular-ui-ace": "^0.2.3",
     "angular-ui-bootstrap": "^2.5.0",
     "angular-vs-repeat": "^1.1.7",
-    "antd": "^3.7.1",
+    "antd": "^3.12.3",
```
gabrieldutra: Note for future me: [link](https://github.com/getredash/redash/pull/2756#issuecomment-453991364).

arikfr:
Good to go?

gabrieldutra:
> Good to go?

There is still a "ant input number showing duplicate arrows" problem that I fixed with css I'm trying to figure out the root cause. Also the styling has changed after #3279 (more Ant like). If it's up to me, those things don't affect the "mergeability" of this.

gabrieldutra:
Also a few things I'm postponing to ListPages migration:
- `dynamicFormHelper.js` - I didn't think it was worthy to spend time rewriting those functions yet;
- CodeClimate issues;

arikfr:
Merged.

arikfr:
Thanks :)

gabrieldutra:
screenshot_20190115-125224
I missed the color difference between labels and Checkboxes. Mind to fix this along with the other Ant color changes.

Issue #3044, opened 2018-11-06
Support Flask-Login cookie security settings


stanhu:
### Issue Summary

A security audit showed that the cookie settings on the Flask-Login `remember_token` didn't have the `HttpOnly` and `Secure` flag set. Redash uses Flask-Login, but there doesn't appear to be a way to pass along the `REMEMBER_` settings: https://flask-login.readthedocs.io/en/latest#cookie-settings

### Steps to Reproduce

1. Enable HTTPS.
2. Login to Redash.
3. Inspect the cookies.

The HTTP-only and Secure flag are not set.

### Technical details:

Here is the hack we added to the source code:

```diff
root@redash:/opt/redash/current/redash/authentication# diff __init__.py.orig __init__.py
143a144
>     app.config.update(REMEMBER_COOKIE_SECURE = True, REMEMBER_COOKIE_HTTPONLY = True)
```

PR #3214, opened 2018-12-23
Set remember_token cookie to be HTTP-Only and Secure

rauchy:
#3044


rauchy:
Heh, I wasted _way_ too much time trying to convert the local dev setup to https. Eventually realized it was much easier to setup an ngrok tunnel and use that one in order to test this issue.

redash/authentication/__init__.py
```diff
@@ -239,6 +239,7 @@ def setup_authentication(app):
     login_manager.anonymous_user = models.AnonymousUser
 
     app.secret_key = settings.COOKIE_SECRET
+    app.config.update(REMEMBER_COOKIE_SECURE=True, REMEMBER_COOKIE_HTTPONLY=True)
```
arikfr: Does `REMEMBER_COOKIE_SECURE` works with `http`?

rauchy: Couldn't find a better way to test for HTTPS at this level than [this](https://github.com/getredash/redash/pull/3214/files#diff-f42630f8986c45dd8f9a209502bf440fR242). The caveat of this is that if someone sets `ENFORCE_HTTPS` to false, but still hosts their Redash over HTTPS, they won't get secure cookies. WDYT?

arikfr: What do you think about the following suggestion:

1. Always set `REMEMBER_COOKIE_HTTPONLY` to True (I'm assuming this works both on HTTP and HTTPS, right?).
2. Have an environment variable to control `REMEMBER_COOKIE_SECURE` that has the default value of whatever `ENFORCE_HTTPS` is set to, but whoever wants can override it.

?

arikfr: Ping.

rauchy: Yeah, I guess that makes sense. Can't think of why someone would host over HTTPS and avoid `ENFORCE_HTTPS=true` though 🤷‍♂

jezdez:
@rauchy Turns out the session and rememberme cookies [have separate config variables to set](https://github.com/maxcountryman/flask-login/commit/312d0e4554d3e0dfb7cd3973d79a7f342d645bff) from the session cookie settings that #3404 is changing, so I think this should be reopened and merged, or maybe appended to #3404. What do you think?

arikfr:
@jezdez good catch. Let's append into #3404?

jezdez:
@arikfr Works for me. @rauchy Feel free to just append away 🛫

rauchy:
@jezdez b5c8e02 hope I didn't break everything 😝

PR #3224, opened 2018-12-27
Add regenerate function of user's API key

kyoshidajp:
# Description

Add regenerate function of user's API key. I think this feature is useful when an user's API key has leaked out.

Non-admins can regenerate only their own key. Admins can regenerate keys for all user's.

# Screenshot
regenerate_button

regenerate_modal

regenerate_success

regenerate_fail


redash/handlers/users.py
```diff
@@ -152,6 +152,28 @@ def post(self, user_id):
         }
 
 
+class UserRegenerateApiKeyResource(BaseResource):
+    def post(self, user_id):
+        user = models.User.get_by_id_and_org(user_id, self.current_org)
+        if user.is_disabled:
+            abort(404, message='Not found')
+        if not is_admin_or_owner(user_id):
+            abort(403, message="You cannot regenerate API Key."
```
arikfr: I think just `abort(403)` is enough. We will not show this UI to non admins once #3234 is resolved.

client/app/pages/users/show.js
```diff
@@ -114,6 +115,33 @@ function UserCtrl(
...
+        .success((data) => {
+          toastr.success('API Key is regenerated.');
+          user.api_key = data.api_key;
+          $scope.disableRegenerateApiKeyButton = false;
+        })
+        .error((response) => {
+          const message =
+            response.message
+              ? response.message
+              : `Cannot regenerate API Key ${response.statusText}`;
```
arikfr:
```suggestion
              : `Failed regenerating API Key: ${response.statusText}`;
```

client/app/pages/users/show.js
```diff
@@ -4,9 +4,10 @@ import { absoluteUrl } from '@/services/utils';
 import template from './show.html';
 import './settings.less';
 
+const regenerateConfirm = { class: 'btn-warning', title: 'Regenerate' };
```
arikfr: Why define it so far from where it's used?

kyoshidajp: I referred to `deleteConfirm` in  https://github.com/getredash/redash/blob/v6.0.0/client/app/pages/data-sources/show.js#L6 as a reference. But `regenerateConfirm` is used only once in `regenerateUserApiKey`. Should I write as an inline?

kyoshidajp:
@arikfr Could you review again?

client/app/pages/users/show.js
```diff
@@ -122,6 +122,34 @@ function UserCtrl(
...
     User.disableUser(user);
   };
+
+  $scope.regenerateUserApiKey = (user) => {
+    const doRegenerate = () => {
+      $scope.disableRegenerateApiKeyButton = true;
+      $http
+        .post(`api/users/${$scope.user.id}/regenerate_api_key`)
+        .success((data) => {
+          toastr.success('API Key is regenerated.');
```
arikfr:
```suggestion
          toastr.success('The API Key has been updated.');
```

redash/handlers/users.py
```diff
@@ -149,6 +149,27 @@ def post(self, user_id):
...
+
+        user.regenerate_api_key()
+        models.db.session.commit()
+
+        self.record_event({
+            'action': 'edit',
+            'object_id': user.id,
+            'object_type': 'user',
+            'updated_fields': ['api_key']
+        })
```
arikfr: I would use `regnerate_api_key` as `action` and remove the `updated_fields` field.

arikfr:
Thanks!

PR #3233, opened 2019-01-01
[WIP] Use CircleCI official image

kyoshidajp:
Use official images (using RAM). I would like to check speedup or not.


arikfr:
It didn't speedup? :)

kyoshidajp:
Both frontend-e2e-tests and backend-unit-tests are failed. But I couldn't solve.

arikfr:
Seem like issues with this image. Maybe need to reference it by another name?

PR #3239, opened 2019-01-01
Fix: Query editor duplicates keystrokes (#2972)

### LINKED ISSUE #2972: Query editor duplicates keystrokes
### Issue Summary

When typing in the query editor seemingly random keystrokes are duplicated, this makes it almost impossible to use.

Imgur

### Technical details:

* Redash Version: 5.0.0+b4754
* Browser/OS: Chrome Version 69.0.3497.100 (Official Build) (64-bit) & Firefox 62.0.3 (64-bit). Windows 10 Pro
* How did you install Redash: Self hosted running on Kubernetes

ranbena:
Bug in code caused js error which messed up the query editor.
`keywords` was set to empty array instead of the `table/column/tableColumn` structure.

Closes #2972.

### DIFF
--- a/client/app/components/QueryEditor.jsx
+++ b/client/app/components/QueryEditor.jsx
@@ -108,7 +108,14 @@ class QueryEditor extends React.Component {
 
   static getDerivedStateFromProps(nextProps, prevState) {
     if (!nextProps.schema) {
-      return { keywords: [], liveAutocompleteDisabled: false };
+      return {
+        keywords: {
+          table: [],
+          column: [],
+          tableColumn: [],
+        },
+        liveAutocompleteDisabled: false,
+      };
     } else if (nextProps.schema !== prevState.schema) {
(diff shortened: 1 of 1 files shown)


gabrieldutra:
I don't think #2972 was caused due to this (since this was not in that version). However, I did miss it and this could cause a similar issue at some flow 😅

ranbena:
> I don't think #2972 was caused due to this (since this was not in that version). However, I did miss it and this could cause a similar issue at some flow 😅

Hmm.. yeah if such a generic breakage in the flow causes this, it could be caused by other bugs as well. FYI, I'm able to consistently reproduce this on latest and this definitely fixes it.

arikfr:
👍

arikfr:
@gabrieldutra you're right that the original issue was probably due to something else, but I doubt it still exists.

PR #3240, opened 2019-01-01
Use the debian flavor of the redash/base image

arikfr:


justinclift:
As a thought, would it make sense to go the alpine route in order to reduce image size as much as possible?

Alpine's package manager is `apk` instead of `apt`, and seems to work decently.

In my usage with it (so far) for a different (Go based) project, it's been fairly easy to figure out the right packages for things.

arikfr:
@justinclift 
The switch to a Debian based image allows to reduce image size and use a standard Python Docker image with minimal effort.

But I don't mind switching over to Alpine. It just that no one had the time to invest in this effort :-)

jezdez:
@arikfr ZOMG!!! 🐍🥉

PR #3242, opened 2019-01-03
Feature: Refresh schedule - code optimizations

ranbena:
This is part of a PR batch to finalize the refresh-schedule feature.
After merging all changes in to `feature/refresh-schedule/emtwo` it should be merged into `master`.

#### What changed
This commit introduces no change in style or functionality.
Prepares code for upcoming changes.

#### Optimizations
1. Moved `intervalToSeconds` to `@/filters` since the corresponding method resides there.
2. Removed unused or unnecessary code in favor of `moment` and `lodash`, 
3. Used advanced js features for clean, lean and perf - getters/setters, memoization and array deconstruction.

🤘

### DIFF
--- a/client/app/components/queries/ScheduleDialog.jsx
+++ b/client/app/components/queries/ScheduleDialog.jsx
@@ -3,80 +3,35 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Modal from 'antd/lib/modal';
 import DatePicker from 'antd/lib/date-picker';
-import { map, range, partial } from 'lodash';
+import { range, padStart } from 'lodash';
 import moment from 'moment';
-import { secondsToInterval, IntervalEnum } from '@/filters';
+import { secondsToInterval, intervalToSeconds, IntervalEnum } from '@/filters';
 
 import './ScheduleDialog.css';
 
-const WEEKDAYS_SHORT = ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
-const WEEKDAYS_FULL = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday'];
-const INTERVAL_OPTIONS_MAP = {};
-INTERVAL_OPTIONS_MAP[IntervalEnum.NEVER] = 1;
-INTERVAL_OPTIONS_MAP[IntervalEnum.MINUTES] = 60;
-INTERVAL_OPTIONS_MAP[IntervalEnum.HOURS] = 24;
-INTERVAL_OPTIONS_MAP[IntervalEnum.DAYS] = 7;
-INTERVAL_OPTIONS_MAP[IntervalEnum.WEEKS] = 5;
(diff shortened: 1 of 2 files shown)


client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -158,104 +127,96 @@ class ScheduleDialog extends React.Component {
...
               ))}
             </select>
           ) : null}
-          <select value={this.state.interval} onChange={this.setInterval}>
-            {getAcceptableIntervals(this.props.refreshOptions).map(iv => (
-              <option value={iv.name || ''} key={iv.name}>
-                {String(iv.name)}
-              </option>
+          <select value={interval} onChange={this.setInterval}>
+            {this.intervals.map(iv => (
```
arikfr: We need to take into account the case where there won't be any option for some interval. The allowed intervals is a setting, the admin can change and some might remove all the per minute intervals all together.

ranbena: Cool. I'll test it and open an issue if it falls short.

ranbena: @arikfr so you mean that memoization needs to depend on the interval settings?

arikfr: memoization? maybe I commented in the wrong place in the code 😳 but what I meant is that if we don't have any minute level intervals available, then we should offer "Minute" as one of the option in the dropdown.

ranbena: Nm. I'll make sure it works as expected 🤝

client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -158,104 +127,96 @@ class ScheduleDialog extends React.Component {
...
         <div className="schedule-component">
           <div>Refresh every</div>
-          {schedule.interval ? (
-            <select value={this.state.count} onChange={this.setCount}>
-              {this.getAcceptableCounts().map(count => (
-                <option value={String(count)} key={count}>
-                  {String(count)}
-                </option>
+          {interval !== IntervalEnum.NEVER ? (
+            <select value={count} onChange={this.setCount}>
```
kravets-levko: How about using Ant's `Select` component instead of plain `<select>` tag?

ranbena: That's the next PR :P

client/app/components/queries/ScheduleDialog.jsx
ranbena: Yeah it works as requested.

The only thing to take into consideration is that I put **memoization** on it (or else it reconstructs the array on any scheduler state change). This means that on the off-chance that these specific settings have changed for the user while a query is open, it would update only on re-entering/refreshing a query page.

Can either remove memoization, or leave as is @arikfr.

arikfr: Can leave as is -- this is not something that changes that frequently.

Issue #3234, opened 2019-01-01
When non admin user viewing a user page, they should see a read only view


arikfr:
Currently we show them something which looks like an editable form:

![image](https://user-images.githubusercontent.com/71468/50570904-f18f2a80-0da4-11e9-9fcc-9d92bae36bdc.png)

But they can't edit anything.

PR #3244, opened 2019-01-03
Simple user view

rauchy:
This PR shows a simple user view (with avatar, name and e-mail) when opening up a profile that isn't yours (as opposed to showing a broken edit form as reported on #3234). Admins will keep seeing an edit form for all users.

image

### DIFF
--- a/client/app/components/UserShow.jsx
+++ b/client/app/components/UserShow.jsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { react2angular } from 'react2angular';
+
+export const UserShow = ({ name, email, profileImageUrl }) => (
+  <div className="col-md-4 col-md-offset-4 profile__container">
+    <img
+      alt="profile"
+      src={profileImageUrl}
+      className="profile__image"
+      width="40"
+    />
+
+    <h3 className="profile__h3">{name}</h3>
+
+    <hr />
+
+    <dl className="profile__dl">
+      <dt>Name:</dt>
+      <dd>{name}</dd>
+      <dt>Email:</dt>
+      <dd>{email}</dd>
+    </dl>
+  </div>
+);
+
+UserShow.propTypes = {
+  name: PropTypes.string.isRequired,
+  email: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


client/app/components/UserShow.jsx
```diff
@@ -0,0 +1,38 @@
...
+      <dd>{email}</dd>
+    </dl>
+  </div>
+);
+
+UserShow.propTypes = {
+  name: PropTypes.string.isRequired,
+  email: PropTypes.string.isRequired,
+  profileImageUrl: PropTypes.string.isRequired,
+};
```
arikfr: I wonder if it wouldn't be more "future proof" to pass the user object?

rauchy: Kinda meaningless because we would want to describe the user object's shape using proptypes, and that's kind of the same effort when (read: "if") we choose to add anything in the future, no?

kravets-levko: > that's kind of the same effort when (read: "if") we choose to add anything in the future, no?

Yes, but then you'll need only change proptypes; currently you'll also need to update all the places where you use this component.

rauchy: @kravets-levko kinda scratching my head with passing objects from Angular to React (as opposed to simple values). Do we have any examples of this happening in our code?

arikfr: What @kravets-levko said + you define the PropType shape of the User object one time and can reuse in multiple components.

ranbena: FYI, static typing allows declaring props which can be used across the app with **no import**.

```ts
// declare.js

declare UserProfile {
  name: string,
  image: string,
  email: string,
  tags?: Tag[],
}
```
```ts
// SomeClass.jsx

type Props = {
  someProp: string,
  profile: UserProfile,
}

class SomeClass extends extends Component<Props> {
  ...
}

```

arikfr: It's supposed to be `declare.ts` in your example, right?

ranbena: I assume (never used TS but rather Flow)

PR #3250, opened 2019-01-03
Fix: Refresh schedule phrase overlaps title

### LINKED ISSUE #3248: Bug: Refresh schedule phrase overlaps title
### Issue Summary
Refresh schedule phrase overlaps title when long enough.

screen shot 2019-01-03 at 18 09 19

(this bug originates from #2426 which has yet to land)

ranbena:
Closes #3248

#### The fix

Made phrase shorter - longest string fits both versions:
screen shot 2019-01-03 at 20 48 29

screen shot 2019-01-03 at 20 48 38

#### Bonus

Added tooltip to show full details on hover
screen shot 2019-01-04 at 8 05 38

Added css style to ensure string doesn't break like so:
screen shot 2019-01-03 at 20 58 00

### DIFF
--- a/client/app/filters/index.js
+++ b/client/app/filters/index.js
@@ -9,6 +9,16 @@ export const IntervalEnum = {
   WEEKS: 'week(s)',
 };
 
+export function localizeTime(time) {
+  const [hrs, mins] = time.split(':');
+  return moment
+    .utc()
+    .hour(hrs)
+    .minute(mins)
+    .local()
+    .format('HH:mm');
+}
+
 export function secondsToInterval(seconds) {
   let interval = IntervalEnum.MINUTES;
   let count = seconds / 60;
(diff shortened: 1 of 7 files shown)


client/app/filters/index.js
```diff
@@ -57,22 +56,7 @@ export function scheduleHumanize(schedule) {
...
-      .local()
-      .format('HH:mm');
-    scheduleString += `at ${localTime} `;
-  }
-
-  if (schedule.day_of_week) {
-    scheduleString += `on ${schedule.day_of_week}`;
-  }
-  return scheduleString;
+  return `Every ${count} ${interval} `;
```
arikfr: What will it print for "Every Tuesday at 16:00"?

ranbena: "Every 1 week(s)"

ranbena: Sure I can make it into a new component 👍🏻

arikfr: The benefit of implementing the component is that users with read only access to the query will be able to see the real schedule (otherwise it's available only in the dialog).

ranbena: Done!

ranbena:
Percy found [a bug](https://percy.io/Redash/redash/builds/1296181?utm_campaign=Redash&utm_content=redash&utm_source=github_status) in my implementation! @arikfr 😃

arikfr:
> Percy found a bug in my implementation! @arikfr 😃

Nice!!

PR #3253, opened 2019-01-04
frontend lint update

koooge:
Hi ✋ This is client-side lint update with change.

- Add npm-scripts `npm run lint` for just client-side lint
- Upgrade eslint & eslint-config-airbnb
- Fix some lint-rule & warn


koooge:
it downgrades eslint 5>4 in codeclimate due to https://github.com/codeclimate/codeclimate-eslint/issues/426

client/app/pages/data-sources/show.js
```diff
@@ -3,12 +3,12 @@ import debug from 'debug';
...
 const logger = debug('redash:http');
-const deleteConfirm = { class: 'btn-warning', title: 'Delete' };
-function logAndToastrError(deleteObject, httpResponse, toastr) {
+export const deleteConfirm = { class: 'btn-warning', title: 'Delete' };
+export function logAndToastrError(deleteObject, httpResponse, toastr) {
   logger('Failed to delete ' + deleteObject + ': ', httpResponse.status, httpResponse.statusText, httpResponse.data);
   toastr.error('Failed to delete ' + deleteObject + '.');
 }
-function toastrSuccessAndPath(deleteObject, deletePath, toastr, $location) {
+export function toastrSuccessAndPath(deleteObject, deletePath, toastr, $location) {
```
ranbena: What's the reason for this?

.codeclimate.yml
```diff
@@ -3,7 +3,7 @@ engines:
     enabled: true
   eslint:
     enabled: true
-    channel: "eslint-5"
+    channel: "eslint-4"
```
ranbena: @koooge can this cause codeclimate redundant warnings?

client/app/pages/data-sources/show.js
koooge: Fix import/named
```
/path/to/redash/client/app/pages/destinations/show.js
  3:10  error  deleteConfirm not found in '../data-sources/show'         import/named
  3:25  error  logAndToastrError not found in '../data-sources/show'     import/named
  3:44  error  toastrSuccessAndPath not found in '../data-sources/show'  import/named
```

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/named.md

.codeclimate.yml
koooge: Yes, it causes lots of `html-webpack-plugin was not found` error. It seems a codeclimate-eslint's issue.

ranbena: I mean the other way around - will the downgrade create discrepancies between the codebase linting (v5) and codeclimate's results (v4)?

koooge: Oh, it's absolutely yes. But I think it's not a big problem since codeclimate can disable specific rules.
https://github.com/getredash/redash/blob/master/.codeclimate.yml#L9-L11
Actually, this PR upgrades codeclimate's eslint3 to 4, and codebase's eslint4 to 5. So, their vesions are already different.

client/.eslintrc.js
```diff
@@ -26,15 +26,22 @@ module.exports = {
     "no-lonely-if": "off",
     "consistent-return": "off",
     "no-control-regex": "off",
+    'no-multiple-empty-lines': 'warn',
     "no-script-url": "off", // some <a> tags should have href="javascript:void(0)"
+    'react/button-has-type': 'warn',
+    'react/destructuring-assignment': 'warn',
```
ranbena: I dunno if I like this one @koooge. Do think this rule is valuable?

client/.eslintrc.js
```diff
@@ -26,15 +26,22 @@ module.exports = {
...
     "react/jsx-filename-extension": "off",
+    'react/jsx-one-expression-per-line': 'warn',
```
ranbena: Same here. I'm not sold on this. More of a nuisance I think.

client/.eslintrc.js
koooge: I fixed one `react/button-has-type` error and delete this warning rule 👌  Basically, I think we should obey the recommended rule.
But there are many (164) `react/button-has-type` warnings. It may be annoying..

koooge:
> You found a solution for this?
Yes, it seems to fix this by https://github.com/codeclimate/codeclimate-eslint/issues/426
But codeclimate(eslint-5) became to regard no-multiple-empty-lines as error😓

client/.eslintrc.js
ranbena: I was actually referring to line 32 `'react/deconstructing-assignment': 'warn'`.

koooge: Oops, I supposed we should obey `react/deconstructing-assignment` rule "in basically", too. But now I disabled this rule to obey the current code base.

ranbena:
@koooge so this is good to go?

koooge:
yes it is!

PR #3255, opened 2019-01-06
Feature: Refresh Schedule - Ant TimePicker

ranbena:
Ant `<TimePicker>` simply feels/looks better than hour/minute dropdowns.

screen shot 2019-01-06 at 9 13 02


ranbena:
Gonna merge it into #3247 instead

PR #3256, opened 2019-01-07
Redis based implementation of user active_at timestamp update

arikfr:
Following the discussion in #3026 this is the direct Redis implementation. Still WIP, because I need to fix the tests.

### DIFF
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -3,7 +3,7 @@ import logging
 import urlparse
 import urllib
 
-import walrus
+import redis 
 from flask import Flask, current_app
 from flask_sslify import SSLify
 from werkzeug.contrib.fixers import ProxyFix
@@ -48,7 +48,7 @@ def create_redis_connection():
         else:
             db = 0
 
-        client = walrus.Database(unix_socket_path=redis_url.path, db=db)
+        client = redis.StrictRedis(unix_socket_path=redis_url.path, db=db)
     else:
         if redis_url.path:
             redis_db = redis_url.path[1]
@@ -56,7 +56,7 @@ def create_redis_connection():
             redis_db = 0
         # Redis passwords might be quoted with special characters
         redis_password = redis_url.password and urllib.unquote(redis_url.password)
-        client = walrus.Database(host=redis_url.hostname, port=redis_url.port, db=redis_db, password=redis_password)
+        client = redis.StrictRedis(host=redis_url.hostname, port=redis_url.port, db=redis_db, password=redis_password)
 
     return client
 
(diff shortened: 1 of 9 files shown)


jezdez:
For the record, I'm sad that this happens to be honest, as I outlined it in the pull request.

1) The Walrus models are a high-level abstraction for data manipulation similar to the SQLAlchemy models and improve the maintainability of a critical feature (updating user specific data per request).
2) A goal of the feature was to be able to store additional user specific data (e.g. user configuration in the meta data blob) in the future instead of having to add additional columns to the user model for every configuration option.
3) Introducing Walrus allows using some high-level data structures for other Redis-based features and reduces TCO of those parts of the code.

arikfr:
> The Walrus models are a high-level abstraction for data manipulation similar to the SQLAlchemy models and improve the maintainability of a critical feature (updating user specific data per request).

Our use of Postgres is far more intensive and complex compared to our use of Redis, and hence the difference in abstractions employed. 

Switching away from walrus, we no longer need:

1. Extra dependency.
2. Extra model.
3. Our implementation of Walrus UTC Timestamp field.

IMO, this improves maintainability of the code. Just consider the fact that #3026 was blocked for a long time, as we were waiting for a compatible release of kombu.

> A goal of the feature was to be able to store additional user specific data (e.g. user configuration in the meta data blob) in the future instead of having to add additional columns to the user model for every configuration option.

True and this is still there in the form of the `details` blob on the User object. This isn't coupled with the Redis implementation.

> Introducing Walrus allows using some high-level data structures for other Redis-based features and reduces TCO of those parts of the code.

I'm not against Walrus in general. I just didn't feel this is a good use case for it. If we were not removing the `QueryTaskTracker` object (#3057), then it could be a good candidate to use Walrus Models. We can always reintroduce it when there is a compelling use case.

jezdez:
> > The Walrus models are a high-level abstraction for data manipulation similar to the SQLAlchemy models and improve the maintainability of a critical feature (updating user specific data per request).
> 
> Our use of Postgres is far more intensive and complex compared to our use of Redis, and hence the difference in abstractions employed.
> 
> Switching away from walrus, we no longer need:
> 
>     1. Extra dependency.
> 
>     2. Extra model.
> 
>     3. Our implementation of Walrus UTC Timestamp field.
> 
> 
> IMO, this improves maintainability of the code. Just consider the fact that #3026 was blocked for a long time, as we were waiting for a compatible release of kombu.
> 
> > A goal of the feature was to be able to store additional user specific data (e.g. user configuration in the meta data blob) in the future instead of having to add additional columns to the user model for every configuration option.
> 
> True and this is still there in the form of the `details` blob on the User object. This isn't coupled with the Redis implementation.
> 
> > Introducing Walrus allows using some high-level data structures for other Redis-based features and reduces TCO of those parts of the code.
> 
> I'm not against Walrus in general. I just didn't feel this is a good use case for it. If we were not removing the `QueryTaskTracker` object (#3057), then it could be a good candidate to use Walrus Models. We can always reintroduce it when there is a compelling use case.

Understood. I'll make sure to discuss this in more detail the next time.

arikfr:
You conceded too quickly. 😅

jezdez:
> You conceded too quickly. 😅

Eh, first day of work 🤓

PR #3265, opened 2019-01-10
Fix: Refresh schedule interval count doesn't adhere to permission rules

ranbena:
Fixes #3263.

#### The bug
Not a programatic bug but rather a wrong feature implementation.
Interval counts were set as hardcoded ranges instead of deriving from `props.refreshOptions`.

#### The fix
Now `intervals` returns grouped count by interval type.
```js
{
  [IntervalEnum: NEVER]: [],
  [IntervalEnum: MINUTES]: [1, 5, 10, 15, 30],
  [IntervalEnum: HOURS]: [1, .. , 12],
  [IntervalEnum: DAYS]: [1],
  [IntervalEnum: WEEKS]: [1, 2, 4.28],
}
```
:raised_back_of_hand: 'Days' has only one value, 'Weeks' has a fraction (30 days)

#### Bonus fix
The code assumed all interval types included a "1" count value.

PR #3268, opened 2019-01-10
Fix: Refresh schedule durations not pluralized sufficiently

ranbena:
"Week(s)" is nice and all but let's do this right. It's especially prominent in the refresh interval dropdown as of #3265.

#### Result
Refresh Scheduler
screen shot 2019-01-10 at 15 41 16

Refresh phrase
screen shot 2019-01-10 at 15 43 25

Query list
screen shot 2019-01-10 at 15 46 06

NOTE: "24 hours" is now "1 day". @arikfr lmk if you want this corrected.
screen shot 2019-01-10 at 15 44 19

### DIFF
--- a/client/app/components/queries/ScheduleDialog.jsx
+++ b/client/app/components/queries/ScheduleDialog.jsx
@@ -8,7 +8,7 @@ import Select from 'antd/lib/select';
 import Radio from 'antd/lib/radio';
 import { capitalize, clone, isEqual } from 'lodash';
 import moment from 'moment';
-import { secondsToInterval, IntervalEnum, localizeTime } from '@/filters';
+import { secondsToInterval, durationHumanize, pluralize, IntervalEnum, localizeTime } from '@/filters';
 
 import './ScheduleDialog.css';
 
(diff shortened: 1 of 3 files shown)


arikfr:
> NOTE: "24 hours" is now "1 day". @arikfr lmk if you want this corrected.

It's not perfect in the dashboard, but it makes sense in the queries (because then you can set a specific hour for it). Maybe we should remove the 24 hours option from the dashboard refresh dropdown 😆 

> image

I wonder if we should drop the "1" on the singular schedules? So it becomes "refreshes every week" instead of "refreshes every 1 week". Not a big deal though.

ranbena:
> I wonder if we should drop the "1" on the singular schedules? So it becomes "refreshes every week" instead of "refreshes every 1 week". Not a big deal though.

I wanted to do that actually. `Moment` has a humanize method which does that exactly. I'll create a followup PR.

arikfr:
> I wanted to do that actually. Moment has a humanize method which does that exactly. I'll create a followup PR.

I'm not sure their humanize does what we need. But on second thought, I think it's not worth spending more time on this. Let's move on to bigger fish.

ranbena:
https://github.com/getredash/redash/pull/3274 😬

arikfr:
📦

PR #3273, opened 2019-01-12
handlers: Fix post users

koooge:
Fix an error following

```
  File "/app/redash/handlers/users.py", line 97, in post
    name, domain = req['email'].split('@', 1)
ValueError: need more than 1 value to unpack
```


redash/handlers/users.py
```diff
@@ -94,6 +94,8 @@ def post(self):
         req = request.get_json(force=True)
         require_fields(req, ('name', 'email'))
 
+        if '@' not in req['email']:
+            abort(400, message='Bad email address.!!!')
```
kravets-levko: Please fix error message (no need for exclamation marks :-) )

koooge: Wow I forgot to delete the debug message! 😭

arikfr:
👍

PR #3282, opened 2019-01-14
Pin pyparsing version

arikfr:
We need to pin the version of pyparsing, as newer versions break SQLAlchemy-Searchable-10.0.6 (newer versions no longer depend on it). We can't upgrade SQLAlchemy-Searchable version as newer versions require PostgreSQL > 9.6, but we target older versions at the moment.


jezdez:
@arikfr Good catch!

PR #3283, opened 2019-01-14
Handling empty query.schedule

ranbena:
Preparing `<SchedulerPhrase>` and `<SchedulerDialog>` for `null` prop.


client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -21,21 +22,24 @@ const { Option, OptGroup } = Select;
 export class ScheduleDialog extends React.Component {
   static propTypes = {
     show: PropTypes.bool.isRequired,
-    // eslint-disable-next-line react/forbid-prop-types
-    query: PropTypes.object.isRequired,
+    schedule: RefreshScheduleType,
```
arikfr: I like it that it takes a `schedule` object now. Will become handy when we upgrade the dashboard scheduler or when we introduce scheduling for alerts which is separate from the query. 👌

client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -21,21 +22,24 @@ const { Option, OptGroup } = Select;
...
+
   constructor(props) {
     super(props);
     this.state = this.initState;
     this.modalRef = React.createRef(); // used by <Select>
   }
 
   get initState() {
-    const newSchedule = clone(this.props.query.schedule);
+    const newSchedule = clone(this.props.schedule || ScheduleDialog.defaultProps.schedule);
```
arikfr: Won't React set the value to the default value on its own, when the passed property is `null`?

ranbena: I WISH! The whole trouble here is that `null` is considered a value (as opposed to `undefined`). WAT?!
Also, `defaultProps` is a shallow nester which means the Object props won't have any meaning.

I do believe this deserves the `typescript-ftw` label here but I need to make sure.

ranbena:
> Looking good. Only thing is missing is setting the value of schedule to `null` when the query is not scheduled.

Good call. Done f170426b995974bd187e2f6fec265dca08784c8d

client/app/components/queries/ScheduleDialog.jsx
arikfr: 🤦‍♂️

arikfr:
Looks good, but how about we merge this into the `none_schedule` branch so it lands along with the other related fixes?

arikfr:
Oh, and looks like you need to update the tests.

ranbena:
> Looks good, but how about we merge this into the `none_schedule` branch so it lands along with the other related fixes?

I think the best strategy is to land it now.

arikfr:
> I think the best strategy is to land it now.

Why? It really serves no purpose without the other pull request and will actually break things for others: the migration in #2426 changes the `schedule` column to not accept null values. 

I know you have null schedule values in your local environment, but I'm pretty sure this is a local issue :-)

arikfr:
> General approach of purging brain queue.
> Avoiding future rebasing issues.

This is why we're going to merge it already, just not into master.

> Ease of mind while working and landing other PR.

I really hope the next thing you're going to do is not this feature related.

> How come I don't get an error saving queries with null in my local machine?

See my comment above about your local setup being broken :-)

Also, you still need to fix the tests here.

client/app/components/queries/ScheduleDialog.test.js
```diff
@@ -78,7 +71,7 @@ describe('ScheduleDialog', () => {
       const [wrapper] = getWrapper({
         interval: 1209600,
         time: '22:15',
-        day_of_week: 2,
+        day_of_week: 'Monday',
```
ranbena: Setting the prop types actually alerted a mistake in my test! This deserves the label :P

arikfr: I suspect you did this on purpose, just so you can tag another pull request :P

PR #3288, opened 2019-01-16
Allow execution of highlighted subquery

chang:
This PR allows the user to execute just the highlighted part of an adhoc query, which often comes in handy when iterating on subqueries.

query_section


chang:
Thanks for the quick review @arikfr! I'd be happy to iterate on this.

> I can add a pointer to where this happens in code.

That would be great.

arikfr:
Great, I appreciate the effort!

We update the reference here:

https://github.com/getredash/redash/blob/823e4ccdd6fcfee5d0df0d919d87af3100876549/client/app/pages/queries/view.js#L375

You can compare the query text of the query result object with the one of the query, and update if needed. This will mean that parameterized queries will no longer store this reference, but this was wrong anyway.

As for executing the correct query text, this is the handler for the Execute button:

https://github.com/getredash/redash/blob/823e4ccdd6fcfee5d0df0d919d87af3100876549/client/app/pages/queries/view.js#L108

Which in turn calls this `getQueryResult`:

https://github.com/getredash/redash/blob/823e4ccdd6fcfee5d0df0d919d87af3100876549/client/app/pages/queries/view.js#L29-L40

Which eventually calls the Query service `getQueryResult` method. It currently takes only the `maxAge` parameter, but for this feature we can add an optional `query` parameter, which will override the current query text.

client/app/components/QueryEditor.jsx
```diff
@@ -129,6 +129,7 @@ class QueryEditor extends React.Component {
 
   onLoad = (editor) => {
     // Release Cmd/Ctrl+L to the browser
+    this.editor = editor;
```
ranbena: No need to rely on this. Can access the editor with:
```js
const { editor } = this.refEditor.current;
```

ranbena:
@arikfr Perhaps missing an indication that current execution is from selection and not full query. Not a must. Thinking of current users who'd be unaware the editor functionality changed.

arikfr:
@ranbena 

> Perhaps missing an indication that current execution is from selection and not full query. Not a must. Thinking of current users who'd be unaware the editor functionality changed.

Good point. It makes sense to change the button text to "Execute Highlighted" when there is a selection.

> The "Save" button shows "dirty*" indication upon selection, which is wrong, as the query content remains unchanged (and saving has no affect).

This is because we update the Query's `query` value... Indeed not an expected behavior.

ranbena:
> This is because we update the Query's query value... Indeed not an expected behavior.

Another bug it introduces - when saving query while text selected, the selected text gets saved as the full query.

ranbena:
> It makes sense to change the button text to "Execute Highlighted" when there is a selection.

I think it could prove too "jumpy" and distracting.
Another approach is an indication on execution instead of on selection.
For instance, change selection color, making it obvious that the selection is the one getting executed.

webp net-gifmaker

arikfr:
@ranbena,

> I think it could prove too "jumpy" and distracting.
Another approach is an indication on execution instead of on selection.
For instance, change selection color, making it obvious that the selection is the one getting executed.

Maybe this + change the tooltip on the Execute button?

client/app/components/QueryEditor.jsx
chang: Hm, `this.refEditor` doesn't seem to be assigned. I can't find it in code and the browser throws an error when trying to access it.

chang:
I've refactored to maintain state of the highlighted text in `view.js` and stop modifying `queryText`. This is so the `mod+enter` shortcut can also get the highlighted text: 
https://github.com/getredash/redash/blob/40c6a2621cae6ea0e09b2868b57d4d728fe6e040/client/app/pages/queries/view.js#L131-L133

That should fix the save-related bugs. Couple questions about the UI:

@ranbena How do you change the text highlight color as it's executing? I couldn't figure this out.
@arikfr What do you mean by "change the tooltip"? The tooltip says `CTRL+ENTER` right now, so I don't think you meant to change that to "Execute Selected".

Currently I have the text change to "Execute Selected" but we can get rid of that and just have the highlight color change.

Also, the CI failure seems to be unrelated. I'm a javascript novice so thanks for the help, and any suggestions for refactoring are welcome.

client/app/components/QueryEditor.jsx
ranbena: Whoops, I dunno how I missed it..
Create the ref in the class constructor and you're good to go.

```jsx
class QueryEditor extends React.Component {
  static propTypes = {
    …
  }

  static defaultProps = {
    …
  }

  constructor(props) {
    super(props);

    this.refEditor = React.createRef(); // <-- right here
   
    …
```

client/app/services/query.js
```diff
@@ -431,12 +431,12 @@ function QueryResource(
...
   };
 
-  Query.prototype.getQueryResult = function getQueryResult(maxAge) {
+  Query.prototype.getQueryResult = function getQueryResult(maxAge, highlightedQueryText) {
     if (!this.query) {
       return new QueryResultError("Can't execute empty query.");
     }
-    const queryText = this.query;
 
+    const queryText = (highlightedQueryText == null) ? this.query : highlightedQueryText;
```
ranbena: Nicer syntax:
```js
const queryText = highlightedQueryText || this.query;
```

arikfr:
> Currently I have the text change to "Execute Selected" but we can get rid of that and just have the highlight color change.

It's actually good this way 👍

arikfr:
👏 

Well done, @chang. Thank you for this great feature and the quick follow ups. 💯

PR #3296, opened 2019-01-16
Update form text colors

gabrieldutra:
## Description
This updates colors for better readability in forms (and in standalone inputs). Also creates an external `inc/ant-variables.less` file.

## Color changes
- Input text color: `#9E9E9E` -> `#595959`
- Ant base text color: `#767676` -> `#595959`

Both are now (kind of) Ant's default `rgba(0, 0, 0, 0.65)`

### DynamicForm (entirely Ant form)
dynamic-form

## Query Params (mixed)
params

Extra: I took the opportunity to add one more Percy Snapshot since I was changing some part of UI, this should be pointed :)


client/app/assets/less/inc/ant-variables.less
```diff
@@ -0,0 +1,29 @@
...
+@font-family-no-number:                @redash-font;
+@font-family:                          @redash-font;
+@code-family:                          @redash-font;
+@font-size-base:                       13px;
+
+
+/* --------------------------------------------------------
+    Typograpgy
+-----------------------------------------------------------*/
+@text-color:                           #595959;
```
gabrieldutra: I didn't find a place where this is noticeable (lmk if you see this anywhere), but this creates a slight difference between `@text-color` for Ant styling and for the rest (`#595959` vs `#767676`).
 
I changed this color because it seemed better for Ant Checkbox labels and Buttons:

**Before**
checkbox-before
**After**
checkbox-after 1

It's probably better to pick one of those and set in both scopes, but I let this open for discussion :slightly_smiling_face:

ranbena: #595959 looks more suitable to me.

ranbena: @gabrieldutra I've noticed form labels have `font-weight: 500` which doesn't look suitable for checkbox and radio button labels (like in this screenshot). Doesn't look like it's an Ant style. Do you know what's causing this?

gabrieldutra:
```less
// inc/form.less 
label {
    font-weight: 500;
}

// ~bootstrap/less/forms.less
label {
  display: inline-block;
  max-width: 100%; // Force IE8 to wrap long content (see https://github.com/twbs/bootstrap/issues/13141)
  margin-bottom: 5px;
  font-weight: bold; // <- This
}
```

gabrieldutra: Replacing the first label selector with `font-weight: normal` fixes this.

gabrieldutra:
Same preview after changing `font-weight`:
**DynamicForm**
dynamic-form-fw

**Query Params**
params-fw

arikfr:
In the form, aren't the labels now a bit less prominent?

cypress/integration/data-source/create_data_source_spec.js
```diff
@@ -14,5 +14,6 @@ describe('Create Data Source', () => {
     cy.getByTestId('Database Name').type('postgres{enter}');
 
     cy.contains('Saved.');
+    cy.percySnapshot('Create Data Source page');
```
arikfr: 😍

gabrieldutra:
> In the form, aren't the labels now a bit less prominent?

Yes, they are :thinking:. That actually didn't bother me, but another option is to change only for labels and checkboxes:
```less
// ant.less
...
// Remove bold in labels for Ant checkboxes and radio-buttons
.ant-checkbox-wrapper, .ant-radio-wrapper {
  font-weight: normal;
}
...
```
### Result
checkbox-fw

ranbena:
> ... change only for labels and checkboxes:

I think that's the best option.

gabrieldutra:
> > ... change only for labels and checkboxes:
> 
> I think that's the best option.

Done :ok_hand:

arikfr:
I agree.

PR #3303, opened 2019-01-17

alon710:


alon710:
Continue of PR https://github.com/getredash/redash/pull/2410

arikfr:
The reason it wasn't merged until now is that we don't have a way to know the impact of pinning the Thrift version on all the data sources that use it.

Until we have better test coverage (#3807) for our query runners or someone from the community can test it with the different ones that use Thrift, we won't be able to merge it :-(

arikfr:
Hi,

Thanks.

PR #3304, opened 2019-01-18
support for fetching all JQL results by way of pagination (fixes #1720)

### LINKED ISSUE #1720: JQL: add support for fetching all the results by way of pagination

justmiles:
This PR adds support for fetching all JQL results by way of pagination. I was able to test this against a Jira Cloud instance with successful results.

This resolves #1720


redash/query_runner/jql.py
```diff
@@ -179,6 +181,19 @@ def run_query(self, query, user):
...
+                while data['total'] > index:
+                    query['startAt'] = index
+                    response, error = self.get_response(jql_url, params=query)
+                    if error is not None:
+                        return None, error
+
+                    data = response.json()
+                    index = data['startAt'] + data['maxResults']
+
+                    addl_results = parse_issues(data, field_mapping)
```
arikfr: Is it possible that the next pages of results might have different columns than the previous ones?

justmiles: Looking at [their docs](https://developer.atlassian.com/server/jira/platform/rest-apis/#pagination) it appears `startAt`, `maxResults`, and `total` will always be in the response.

justmiles: Ah, I see. That wouldn't be possible. They are additional results to the same query. The query defines the columns. If the column is empty it'll still come back but with a null value.

arikfr:
Thanks!

PR #3307, opened 2019-01-19
Override default number formats with environment variables

chang:
This change lets the user override the default number formats with the environment variables `REDASH_INTEGER_FORMAT` and `REDASH_FLOAT_FORMAT`, following the pattern of `REDASH_DATE_FORMAT`.

Looking at: 
https://github.com/getredash/redash/blob/e8120c5f79cca01b5a011e8197c4cd604bf718d9/client/app/visualizations/table/index.js#L61-L64

It actually seems like this was a planned feature, but was just missing the option to populate `clientConfig` with the env variables.

Tested by setting:
```bash
export REDASH_INTEGER_FORMAT="00"
export REDASH_FLOAT_FORMAT="00.00"
```
image

### DIFF
--- a/redash/handlers/authentication.py
+++ b/redash/handlers/authentication.py
@@ -199,6 +199,13 @@ def date_format_config():
     }
 
 
+def number_format_config():
+    return {
+        'integerFormat': current_org.get_setting('integer_format'),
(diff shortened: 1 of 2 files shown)


arikfr:
Thanks!

Issue #3284, opened 2019-01-15
Complete redo of permissions in Redash


arikfr:
**This is a first draft to describe this project. We invite interested parties to share their thoughts. As we progress with better planning this, I will update the issue description and add relevant sub issues for the steps.**

---

Until now access to data in Redash was governed by access to the data source this data came from. It makes sense as a good default, specially when we want to promote data democracy. But it introduces many challenges in less trivial use cases, for example:

* You want to share high level revenues KPIs dashboard with everyone, but give access to the raw data only to a limited group of people.
* You want to share a few dashboards with your investors without giving them access to all your data.

Also today you can share a single query using the embed link or a whole dashboard using the share link. This is currently limited only to queries/dashboards without parameters, as the later require access to run _any_ query on the underlying data source.

The goal of this project is to redo our permissions model and introduce safety to parameters execution to allow for the ability to share a query or dashboard with anyone the user's chooses - internal or external.

This is a huge change in how Redash works and will be implemented in steps:

1. Add support for "feature detection", to allow for the application to decide whether a specific object can be shared or not. We will start with disabling current sharing options (embed/shared link) when the object uses parameters.
2. Add support for safe execution of parameters: #2904 
3. Add support for sharing queries/dashboards with other users, through the current permissions dialog.
4. Add support for a public link/access to an object, which will replace the current share/embed features in queries/dashboards.
5. Add an API to create shareable links for objects with some parameters predefined (to allow Embedded Analytics use cases).

Some things to consider:

* What happens with an object that was shared already when you add a parameter?
* Add a concept of workspaces/teams instead of groups? So anything you create in this workspace has some default permissions?
* Until now you had to create data sources of type Query Results, URL, CSV, etc for to purpose of managing permissions. But once we have permissions decoupled from data sources, we can have all these data source types as built in types, so you can use them to load data into Redash and then decide who you share it with. ([related discussion](https://discuss.redash.io/t/add-an-in-app-csv-upload-option/2809))

ice2038:
I am interested. I have a problem with access to dashboards for viewing. I would like to have access to such dashboards and to such requests only from the assigned group of users.

I propose to introduce collections which may include dashboards, queries, tags, etc. These collections can be assigned to groups of users who see only them and nothing else.

pachecobruno:
Great! Today I  need to publish some results/insights for specific users, who should not access the queries or data sources. Will it be possible to create groups with more restrictive permissions? What you think about granularities of roles / permissions? Tks!

[...]

PR #3311, opened 2019-01-20
Textless query result endpoint

rauchy:
This PR adds an endpoint for running saved queries by POSTing to /api/queries/{id}/results and optionally providing parameter values. This is related to #3284 and will be used as the endpoint to be shared.

Parameter type validation, which is required for safe sharing of queries, will be applied on top of this.


redash/handlers/api.py
```diff
@@ -92,6 +92,7 @@ def json_representation(data, code, headers=None):
 api.add_org_resource(QueryResultResource,
                      '/api/query_results/<query_result_id>.<filetype>',
                      '/api/query_results/<query_result_id>',
+                     '/api/queries/<query_id>/results',
                      '/api/queries/<query_id>/results.<filetype>',
                      '/api/queries/<query_id>/results/<query_result_id>.<filetype>',
                      endpoint='query_result')
```
arikfr: It bothers me that we have so many different URLs on a single handler. One side effect of this, is now someone can call `POST /api/queries/<query id>/results.json` or `POST /api/queries/<query id>/results/<query result id>.json` and generate a server error (500) instead of receiving a method not supported error.

Let's start splitting them by adding this new one to its own handler?

redash/handlers/query_results.py
```diff
@@ -154,6 +154,26 @@ def options(self, query_id=None, query_result_id=None, filetype='json'):
...
+                                any cached result, or executes if not available. Set to zero to
+                                always execute.
+        """
+        params = request.get_json(force=True)
+        parameters = params.get('parameters', {})
+        max_age = int(params.get('max_age', 0))
+
+        query = get_object_or_404(models.Query.get_by_id_and_org, query_id, self.current_org)
+
+        return run_query(query.data_source, parameters, query.query_text, query_id, max_age)
```
arikfr: We need to do a permissions check before running the query, just like we do in `QueryResultListResource#post`.

redash/handlers/api.py
rauchy: You're right, it bothered me while adding the endpoint. I just couldn't find a proper definition for such handler.

arikfr: I just realized that we're limited by what Flask-Restful allows. This is the case where if we had simple function based Flask handlers it would probably be less limited. :|

PR #3312, opened 2019-01-20
Add support for setup.sh on CentOS 7

justinclift:
Tested on CentOS 7 x64 1810, and Ubuntu 18.04.01.

Should be reasonably robust too. :smile:

### DIFF
--- a/setup/README.md
+++ b/setup/README.md
@@ -1,6 +1,6 @@
-# Setup script for Redash with Docker on Ubuntu 18.04.
+# Setup script for Redash with Docker on CentOS 7 and Ubuntu 18.04.
 
-This is a reference setup for Redash on a single Ubuntu 18.04 server, which uses Docker and Docker Compose for deployment and management.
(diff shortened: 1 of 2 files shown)


arikfr:
Thank you, @justinclift! This is greatly appreciated, but unfortunately we're a small team, and won't have the resources to continue testing the setup script both on Ubuntu & CentOS 😔 

How about you move the CentOS version to a different repo (or just a Gist), and we add a link to it in the README and instructions?

justinclift:
Ahhh.  I've been clearing out my GitHub account today (deleting all my repo's, gists, etc), with the intention of finally killing my account here.

Lets just close this then, and when I set up a new location (probably spin up Gitea on a server) then we can figure it out then. :smile:

arikfr:
Just so the effort is not lost, maybe post this version to the forum?

Also, if you are going to keep the account itself but just without your own content, I can open the repo for you.

Thanks.

justinclift:
No worries.  I'll post it to the forum.

PR #3314, opened 2019-01-20
Add seamless redirect when SAML is only authentication option

mikeclarke:
For the case where SAML is the only option to authenticate, this changeset avoids rendering the template with the "SAML Login" link and starts the SAML login redirect flow directly. This allows for a seamless login experience without having to render the intermediate page with the single link. Works great for my use case, maybe this is handy for others as well.

### DIFF
--- a/migrations/versions/969126bd800f_.py
+++ b/migrations/versions/969126bd800f_.py
@@ -24,15 +24,20 @@ def upgrade():
     # Update widgets position data:
     column_size = 3
     print("Updating dashboards position data:")
-    for dashboard in Dashboard.query:
-        print("  Updating dashboard: {}".format(dashboard.id))
-        layout = simplejson.loads(dashboard.layout)
+    dashboard_result = db.session.execute("SELECT id, layout FROM dashboards")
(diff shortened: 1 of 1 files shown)


mikeclarke:
Environment variable works for me, what do you prefer - maybe `REDASH_SEAMLESS_SSO`?

> 1. The user can't logout (because when logging out, you get redirected to the login page, which redirects you to the SSO provider).

Would it make sense to add a logged-out view that behaves just like the login page, but without the immediate redirect?

> 2. With Google Apps, it sometimes created endless redirect loops.

Hm, not sure how to avoid this.

arikfr:
> Environment variable works for me, what do you prefer - maybe REDASH_SEAMLESS_SSO?

If it's going to be only for SAML, then maybe `REDASH_SEAMLESS_SAML_SSO`?

> Would it make sense to add a logged-out view that behaves just like the login page, but without the immediate redirect?

Makes sense. Although if you don't mind the immediate redirect, and it's behind an ENV var, I don't mind merging without it :)

PR #3315, opened 2019-01-21
Server-side parameter validation

rauchy:
This PR adds backend validation to query parameters. This validation will be used in textless endpoints, in which only query IDs and parameter values are passed in. Validation is done against the parameter schema (under `Query.options["parameters"]`) and is handled inside `ParameterizedQuery`.

Outside `ParameterizedQuery`, one preprocessing stage can be found under the query result handler which handles query-based parameters. The preprocessing stage transforms the `query`-type parameter to an `enum`-type parameter with all the values. This feels a bit like Solution Sprawl, but the alternative is to pass in `current_org` to `ParameterizedQuery` and have it fetch the values itself, which also feels strange. Would love to hear opinions on which sucks less.

#3284


redash/utils/parameterized_query.py
```diff
@@ -33,17 +33,59 @@ def _parameter_names(parameter_values):
...
         return self
 
+    def _valid(self, name, value):
+        definition = next((definition for definition in self.schema if definition["name"] == name), None)
+
+        if not definition:
+            return True
+
+        validators = {
+            "text": lambda x: type(x) in (str, unicode),
```
jezdez:
```suggestion
            "text": lambda x: isinstance(x, basestring),
```

redash/utils/parameterized_query.py
```diff
@@ -33,17 +33,59 @@ def _parameter_names(parameter_values):
...
+            "number": lambda x: type(x) == int,
```
jezdez:
```suggestion
            "number": lambda x: isinstance(x, int),
```

redash/utils/parameterized_query.py
```diff
@@ -33,17 +33,59 @@ def _parameter_names(parameter_values):
...
+        self.schema = schema
         self.query = template
         self.parameters = {}
 
     def apply(self, parameters):
-        self.parameters.update(parameters)
-        self.query = mustache_render(self.template, self.parameters)
+        invalid_parameter_names = [key for (key, value) in parameters.iteritems() if not self._valid(key, value)]
+        if invalid_parameter_names:
+            message = u"The following parameter values are incompatible with their type definitions: {}".format(",".join(invalid_parameter_names))
```
jezdez:
```suggestion
            message = u"The following parameter values are incompatible with their type definitions: {}".format(", ".join(invalid_parameter_names))
```

redash/handlers/query_results.py
```diff
@@ -64,7 +64,7 @@ def run_query_sync(data_source, parameter_values, query_text, max_age=0):
         return None
 
 
-def run_query(data_source, parameter_values, query_text, query_id, max_age=0):
+def run_query(data_source, parameter_values, query_text, query_id, max_age=0, parameter_schema={}):
```
jezdez: Mutable types such as dictionaries shouldn't be used for parameter defaults, since the dictionary is persisted between function calls, more [info here](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html).

```suggestion
def run_query(data_source, parameter_values, query_text, query_id, max_age=0, parameter_schema=None):
    if parameter_schema is None:
        parameter_schema = {}
```

redash/utils/parameterized_query.py
```diff
@@ -33,17 +33,59 @@ def _parameter_names(parameter_values):
...
+def _is_date_range(obj):
+    try:
+        return _is_date(obj["start"]) and _is_date(obj["end"])
+    except (KeyError, TypeError):
+        return False
+
+
 class ParameterizedQuery(object):
-    def __init__(self, template):
+    def __init__(self, template, schema={}):
```
jezdez:
```suggestion
    def __init__(self, template, schema=None):
        self.schema = schema or {}
```
Same issue with mutable types as above.

redash/utils/parameterized_query.py
```diff
@@ -33,17 +33,59 @@ def _parameter_names(parameter_values):
...
         self.template = template
+        self.schema = schema
```
jezdez:
```suggestion
```

rauchy:
@jezdez keep those nits coming 👍

redash/handlers/query_results.py
```diff
@@ -154,6 +156,20 @@ def options(self, query_id=None, query_result_id=None, filetype='json'):
 
         return make_response("", 200, headers)
 
+    def _fetch_rows(self, query_id):
+        query = models.Query.get_by_id_and_org(query_id, self.current_org)
+        query_result = models.QueryResult.get_by_id_and_org(query.latest_query_data_id, self.current_org)
```
arikfr: 🤔 do we need to check execution permission to this query here?

redash/utils/parameterized_query.py
```diff
@@ -33,17 +33,59 @@ def _parameter_names(parameter_values):
     return names
 
 
+def _is_date(string):
+    try:
+        parse(string)
```
arikfr: We should parse based on a specific format.

redash/utils/parameterized_query.py
```diff
@@ -33,17 +33,59 @@ def _parameter_names(parameter_values):
...
         self.template = template
         self.query = template
         self.parameters = {}
 
     def apply(self, parameters):
-        self.parameters.update(parameters)
-        self.query = mustache_render(self.template, self.parameters)
+        invalid_parameter_names = [key for (key, value) in parameters.iteritems() if not self._valid(key, value)]
+        if invalid_parameter_names:
+            message = u"The following parameter values are incompatible with their type definitions: {}".format(", ".join(invalid_parameter_names))
```
arikfr: Maybe contract the message in `InvalidParameterError` and pass it the parameter names?

redash/utils/parameterized_query.py
```diff
@@ -33,17 +33,59 @@ def _parameter_names(parameter_values):
...
 
+    def _valid(self, name, value):
+        definition = next((definition for definition in self.schema if definition["name"] == name), None)
+
+        if not definition:
+            return True
+
+        validators = {
+            "text": lambda x: isinstance(x, basestring),
+            "number": lambda x: isinstance(x, int),
```
arikfr: Doesn't `number` support floats?

redash/handlers/query_results.py
rauchy: The implication of _not_ checking for execution permissions here is that someone could potentially check that specific values are within the result set. Also, this is restricted within the organization. However, it's probably better to double check.

redash/utils/parameterized_query.py
rauchy: What's the risk here? A format is a data intake/display concern, not a security concern.

redash/handlers/query_results.py
```diff
@@ -154,6 +156,23 @@ def options(self, query_id=None, query_result_id=None, filetype='json'):
...
+
+        require_access(query_result.data_source.groups, self.current_user, view_only)
```
arikfr: Better to do this before we went through the trouble of loading the query and query_result.

redash/handlers/query_results.py
```diff
@@ -154,6 +156,23 @@ def options(self, query_id=None, query_result_id=None, filetype='json'):
...
+
+        return json_loads(query_result.data)["rows"]
+
+    def _convert_queries_to_enums(self, definition):
+        if definition["type"] == "query":
+            definition["type"] = "enum"
+
+            rows = self._fetch_rows(definition.pop("queryId"))
+            definition["enumOptions"] = [row.get('value', row.get(row.keys()[0])) for row in rows]
```
arikfr: That's much less code than the client side version 😮 _But_ it's possible that the first column here and on the client side won't be the same one... For now we can ignore this, but we should not wait too long with implementing #3316.

Also to be on the safe side, I would check that rows is not empty.

redash/handlers/query_results.py
```diff
@@ -171,8 +190,10 @@ def post(self, query_id):
         max_age = int(params.get('max_age', 0))
 
         query = get_object_or_404(models.Query.get_by_id_and_org, query_id, self.current_org)
+        parameter_schema = map(self._convert_queries_to_enums,
+                               query.options.get("parameters", {}))
```
arikfr: `parameters` is a list, so the default value should be `[]`, right? (I think this needs to be updated in other places in the code too)

redash/utils/parameterized_query.py
```diff
@@ -33,17 +34,58 @@ def _parameter_names(parameter_values):
...
+            "enum": lambda x: x in definition["enumOptions"],
+            "date": _is_date,
+            "datetime-local": _is_date,
+            "datetime-with-seconds": _is_date,
+            "date-range": _is_date_range,
+            "datetime-range": _is_date_range,
+            "datetime-range-with-seconds": _is_date_range,
+        }
+
+        return validators[definition["type"]](value)
```
arikfr: Should handle the case of an unknown type with appropriate error instead of just blowing up.

redash/handlers/query_results.py
rauchy: The best I can think of here is moving this one line up and running ```require_access(query.data_source.groups, self.current_user, view_only)```, otherwise I can't see how we get the groups to require access to.

arikfr: I overlooked the fact it needs to know the data source, so yes -- let's move it before query_result.

redash/handlers/query_results.py
rauchy: Keep in mind that this code doesn't care about 'name' columns, which is a presentation concern so it won't be as short.

redash/handlers/query_results.py
rauchy: Good catch!

redash/handlers/query_results.py
```diff
@@ -64,7 +64,9 @@ def run_query_sync(data_source, parameter_values, query_text, max_age=0):
         return None
 
 
-def run_query(data_source, parameter_values, query_text, query_id, max_age=0):
+def run_query(data_source, parameter_values, query_text, query_id, max_age=0, parameter_schema=None):
+    if parameter_schema is None:
+        parameter_schema = {}
```
arikfr:
```suggestion
        parameter_schema = []
```

No?

rauchy: you're right, but I think I have something better

redash/handlers/query_results.py
```diff
@@ -64,7 +64,7 @@ def run_query_sync(data_source, parameter_values, query_text, max_age=0):
         return None
 
 
-def run_query(data_source, parameter_values, query_text, query_id, max_age=0):
+def run_query(data_source, parameter_values, query_text, query_id, max_age=0, parameter_schema=None):
```
arikfr: You removed the handling of the `None` case here, because I assume that the callers will always pass a `parameter_schema`? If that's the case, then should probably remove the default value.

rauchy: This is actually intentional. Handling of `None` is done in `ParameterizedQuery`.

arikfr: I thought that's the case, but didn't find it. Apparently I was looking in the wrong place. 🤦🏻‍♂️

PR #3317, opened 2019-01-21
Use Ant's Paginator component; migrate SortIcon to React

kravets-levko:
Migrate SortIcon to React: look and behavior didn't change.

Use Ant's Paginator component:

image


client/app/components/Paginator.jsx
```diff
@@ -0,0 +1,46 @@
...
+Pagination.propTypes = {
+  page: PropTypes.number,
+  itemsPerPage: PropTypes.number,
+  totalCount: PropTypes.number,
+  onChange: PropTypes.func,
+};
+
+Pagination.defaultProps = {
+  page: 0,
+  itemsPerPage: 0,
```
arikfr: 0?

client/app/components/Paginator.jsx
```diff
@@ -0,0 +1,46 @@
...
+  totalCount: 0,
```
arikfr: I think it's better to make this a required property.

client/app/components/Paginator.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+Pagination.propTypes = {
+  page: PropTypes.number,
+  itemsPerPage: PropTypes.number,
+  totalCount: PropTypes.number,
+  onChange: PropTypes.func,
+};
+
+Pagination.defaultProps = {
+  page: 0,
```
arikfr: Maybe make it into a required property?

client/app/lib/list-ctrl.js
```diff
@@ -60,6 +60,8 @@ export default class ListCtrl {
       orderByReverse: this.pageOrderReverse,
     });
 
+    this.setCurrentPage = page => this.paginator.setPage(page);
```
arikfr: In `outdated-queries/index.js` you call `$applyAsync` after calling `setPage`. It's not needed here?

client/app/components/Paginator.jsx
```diff
@@ -0,0 +1,46 @@
...
+    </div>
+  );
+}
+
+Pagination.propTypes = {
+  page: PropTypes.number,
+  itemsPerPage: PropTypes.number,
+  totalCount: PropTypes.number,
+  onChange: PropTypes.func,
+};
```
arikfr: The Angular version of the Paginator component took as input a Paginator model object. This saved a lot of repeated code where it was used. Why not do the same with the React version?

client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -831,6 +807,56 @@ text.slicetext {
   text-shadow: 1px 1px 5px #333;
 }
 
+.pagination {
```
arikfr: This is styling AntPagination to look like our old paginator? If so, this should be in `ant.less`, no?

kravets-levko: There were two CSS overrides in this file, I just merged them together and added some new rules for Ant. I'm not sure that we should move it to `ant.less`

client/app/components/Paginator.jsx
arikfr: Probably a better idea. I'm not sure what it will take to use the Pagination model there. Two other alternatives are:

1. Make `<dynamic-table>` have the same "interface" as the pagination model.
2. Have another component that uses the one you have here, that takes the model.

client/app/assets/less/redash/redash-newstyle.less
arikfr: But are these overrides apply to the Ant paginator too?

client/app/lib/list-ctrl.js
kravets-levko: Hm, good catch. It works here without `$applyAsync` (I tested all pages except of outdated queries - because I don't know how to fill it with some items), and everywhere I had to add `$applyAsync`. It's strange that it works here

client/app/assets/less/redash/redash-newstyle.less
kravets-levko: Yes

client/app/components/Paginator.jsx
kravets-levko: It will not work - `react2angular` does no detect changes of object properties, so we cannot pass model object here because everywhere it's populated after the first render

client/app/lib/list-ctrl.js
kravets-levko: It works without `$applyAsync`, but it shouldn't. Probably it's some side-effect, so I added `$applyAsync` to ensure that it will not accidentally break

client/app/assets/less/redash/redash-newstyle.less
kravets-levko: If it matters - we don't use "raw" bootstrap paginator anywhere - just our Paginator component

client/app/components/Paginator.jsx
```diff
@@ -0,0 +1,43 @@
...
+      />
+    </div>
+  );
+}
+
+Pagination.propTypes = {
+  page: PropTypes.number.isRequired,
+  itemsPerPage: PropTypes.number.isRequired,
+  totalCount: PropTypes.number.isRequired,
+  onChange: PropTypes.func,
```
ranbena: Any reason not to require `onChange` unless we have a need for a "view only" mode?

client/app/assets/less/redash/redash-newstyle.less
arikfr: In this case it's worth removing unneeded styles and moving this to the Ant file. 

cc: @ranbena

client/app/components/Paginator.jsx
arikfr: We can have a simple wrapper in Angular that passes the properties from the object to React, and once all the Pagination component users are in React get rid of it. This is an isolated change that will be easier to cleanup than bad API used everywhere.

client/app/components/Paginator.jsx
kravets-levko: Well, I don't know. Currently we use this handler everywhere, but unlike other properties, this one is not required to render the component, so I decided to make it optional.

client/app/assets/less/redash/redash-newstyle.less
kravets-levko: I think our styles need more solid refactoring and it's not a subject of this PR

client/app/components/Paginator.jsx
kravets-levko: We will need to cleanup bad API anyways. Currently we have a lot of places where we use Paginator from the react templates; we can create wrapper to pass model objects, but later we will move all that pages to React, and there we will use Paginator component directly with its bad API - so I see no profit.

client/app/components/SortIcon.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { react2angular } from 'react2angular';
+
+export function SortIcon({ column, sortColumn, reverse }) {
```
ranbena: FYI, once tables are converted to Ant `<Table>`, the "Filter and sorter" feature would render this component obsolete.

kravets-levko: But until that time we need this component

kravets-levko:
> The font-family on the paginator is arial instead of our own definition

Thanks for this. I'm going to move overrides to a better place and fix this and some other issues with styling.

> Notice, the ellipsis is horizontally misaligned in it's containing box.

If I'll make a transparent background there (as you suggested) - I think we can leave it as is. WDYT?

client/app/components/Paginator.jsx
arikfr: No, the idea is that once we move to React we keep using the object API. Basically we will start with:

1. React component that takes 4 properties.
2. Angular component that uses the React one and accepts an object.

Then when we start moving tables to React, we add a React component that accepts an object. When all using code is in React, we can remove the Angular middleware.

This is all isolated changes in a single place, instead of tracking down all the places that use this component. Also, if we had done this from the beginning, you wouldn't have to repeat the $applyAsync code and wouldn't have discrepancies ;-)

As a side note, considering how often the issue of passing an object comes up -- maybe we can make it work somehow?

ranbena:
> I think we can leave it as is. WDYT?

I have time on my hands https://github.com/ant-design/ant-design/pull/14473

kravets-levko:
@ranbena @arikfr 

image

client/app/assets/less/inc/ant-variables.less
```diff
@@ -1,8 +1,38 @@
...
+@light-gray:                            #FCFCFC;
+@gray-light:                            #828282;
+@ace:                                   #f8f8f8;
+
 @primary-color:                        #2196F3;
 
+@redash-gray: rgba(102, 136, 153, 1);
+@redash-orange: rgba(255, 120, 100, 1);
+@redash-black: rgba(0, 0, 0, 1);
+@redash-yellow: rgba(252, 252, 161, 0.75);
```
ranbena: Not important nor a blocker but perhaps this list could be minimized to the bare minimum - `@redash-gray` and `@lightblue`? Doesn't look like the other colors are being called.

kravets-levko: Removed

kravets-levko:
@ranbena prev/next arrows fixed (there is border, so `line-height: @pagination-item-size - 2px;`)

PR #3319, opened 2019-01-21
Add: Uptycs query runner

### LINKED ISSUE #2932: Uptycs Query_runner for Redash
Hi Team, 

I want to add new query_runner for Uptycs in Redash. 
I did some testing with query_runner for Uptycs. However before I share the code, please let me know the process of posting the query_runner.

Thanks,
Vibhor

vibhorkum:
Hi Team,

I have added the query_runner for uptycs. Can you please verify and let me know if anything we need to add to make it part of Redash.

Thanks,
Vibhor

Closes #2932.


vibhorkum:
Hi @arikfr,
Can you review the changes and let me know if we need to modify for redash?

Thanks,

requirements.txt
```diff
@@ -55,6 +55,7 @@ python-geoip-geolite2==2015.303
 chromelogger==0.4.3
 pypd==1.1.0
 disposable-email-domains
+urllib3==1.24.1
```
arikfr: Why this is needed?

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+            },
+            "order": ['url', 'customerid', 'key', 'secret'],
+            "required": ["url", "customerid", "key","secret"],
+            "secret": ["secret", "customerid", "key"]
+        }
+
+
+    @classmethod
+    def enabled(cls):
+        return enabled
```
arikfr: The default value for this classmethod is `True`, so you can just remove it.

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "url": {
+                    "type": "string"
+                },
+                 "customerid": {
```
arikfr: It's will be more consistent with our style to use `customer_id`.

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+                "key": {
+                    "type": "string"
+                },
+                "secret": {
+                    "type": "string",
+                },
+            },
+            "order": ['url', 'customerid', 'key', 'secret'],
+            "required": ["url", "customerid", "key","secret"],
+            "secret": ["secret", "customerid", "key"]
```
arikfr: Are you sure customer id and key should be secret too?

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+        return enabled
+
+    @classmethod
+    def annotate_query(cls):
+        return False
+
+
+    @classmethod
+    def type(cls):
+        return "uptycs"
```
arikfr: You can remove this, as this will be the default anyway.

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+    def generateHeaders(self, key, secret):
```
arikfr: We're following [PEP-8](https://www.python.org/dev/peps/pep-0008/) as our styleguide, so this should be `generate_headers`.

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+        header={}
+        utcnow=datetime.datetime.utcnow()
+        date=utcnow.strftime("%a, %d %b %Y %H:%M:%S GMT")
+        authVar = jwt.encode({'iss': key}, secret, algorithm='HS256')
```
arikfr: `auth_var` (PEP-8)

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+        transformed_columns = []
+        rows = []
+        # convert all type to JSON string
+        # In future we correct data type  mapping later
+        if 'columns' in data:
+            for json_each in data['columns']:
+                name = json_each['name']
+                new_json = {"name": name, 
+                            "type": "string", 
+                            "friendly_name": "null"}
```
arikfr:
```suggestion
                            "friendly_name": name}
```

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+                transformed_columns.append(new_json)
+        # Transfored items into rows.
+        if 'items' in data:
+            rows = data['items']
+
+        redash_json_data = {"columns": transformed_columns, 
+                            "rows": rows}  
+        return redash_json_data
+
+    def restcall(self, sql):
```
arikfr:
```suggestion
    def api_call(self, sql):
```

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+             self.configuration.get('customerid')));
+
+        # post data base sql
+        post_data_json = { "query": sql }
+
+        response = requests.post(url, headers=header, 
+                json=post_data_json, verify=False )
+        response_output = json.loads(response.content)
+     
+        json_data = self.transformed_to_redash_json(response_output)
```
arikfr: It's better to check for error before transforming the results and only assign value to json_data if there was no error.

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,141 @@
...
+        error = None
+        # if we got error from Uptycs include error information
+        if 'error' in response_output:
+            error = response_output['error']['message']['brief']
+            error = error + '\n' + response_output['error']['message']['detail']
+        return json_data, error
+
+    def run_query(self, query, user):
+        data, error = self.restcall(query)
+        json_data = json.dumps(data, cls=JSONEncoder)
```
arikfr: Please use `json_dumps` from `redash.utils`.

requirements.txt
vibhorkum: This one is needed for ignoring any SSL certificate issue.

redash/query_runner/uptycs.py
vibhorkum: Yes. This is intentional. We want to keep it as secret.

requirements.txt
arikfr: Can you provide more details on what do you mean?

redash/query_runner/uptycs.py
arikfr: Why?

redash/query_runner/uptycs.py
vibhorkum: Made the changes as per your comment

redash/query_runner/uptycs.py
vibhorkum: fixed

redash/query_runner/uptycs.py
vibhorkum: This is purely based on Uptycs way of keeping combinations secret. However, will check with the team one more time.

arikfr: I think that atleast having one of them as non secret is better, because otherwise the admin has no way of knowing what configuration this is about. Values marked as secret are never sent back to the UI.

Also note that the other values are only available to the admin.

vibhorkum: Based on the discussion, removed customer_id from secret

requirements.txt
vibhorkum: I will be testing this and will update the thread accordingly.

vibhorkum:
Hi @arikfr,  Fixed all the issues based on your comments. Can you please re-verify and let me know if anything is missing?

Thanks

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,133 @@
...
+        # URL form using API key file based on GLOBAL
+        url = ("%s/public/api/customers/%s/query" %
+               (self.configuration.get('url'),
+                self.configuration.get('customer_id')))
+
+        # post data base sql
+        post_data_json = {"query": sql}
+
+        response = requests.post(url, headers=header, json=post_data_json,
+                                 verify=False)
```
arikfr: `verify=False` is a bad practice. Is uptycs' API always using self signed certificates or something? Maybe we should move this to a configuration option?

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,133 @@
...
+from redash.utils import json_dumps
+
+import json
+import jwt
+import datetime
+import requests
+import logging
+logger = logging.getLogger(__name__)
+
+enabled = True
```
arikfr: This seems to be unused.

vibhorkum: Removed the enabled variable

redash/query_runner/uptycs.py
vibhorkum: Fixed and added the SSL and option

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,136 @@
...
+                "url": {
+                    "type": "string"
+                },
+                "customer_id": {
+                    "type": "string"
+                },
+                "key": {
+                    "type": "string"
+                },
+                "SSL": {
```
arikfr: Should be lowered cases (like `ssl`) and `verify_ssl` is probably more descriptive.

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,136 @@
...
+                    "type": "boolean",
```
arikfr:
```suggestion
                    "type": "boolean",
                    "default": True,
                    "title": "Verify SSL Certificates",
```

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,136 @@
...
+        # URL form using API key file based on GLOBAL
+        url = ("%s/public/api/customers/%s/query" %
+               (self.configuration.get('url'),
+                self.configuration.get('customer_id')))
+
+        # post data base sql
+        post_data_json = {"query": sql}
+
+        response = requests.post(url, headers=header, json=post_data_json,
+                                 verify=self.configuration.get('SSL'))
```
arikfr:
```suggestion
                                 verify=self.configuration.get('SSL', True))
```

(default value is True)

redash/query_runner/uptycs.py
```diff
@@ -0,0 +1,136 @@
...
+        logger.debug("%s", json_data)
+        return json_data, error
+
+    def get_schema(self, get_stats=False):
+        header = self.generate_header(self.configuration.get('key'),
+                                      self.configuration.get('secret'))
+        url = ("%s/public/api/customers/%s/schema/global" %
+               (self.configuration.get('url'),
+                self.configuration.get('customer_id')))
+        response = requests.get(url, headers=header, verify=False)
```
arikfr: You need to use the verify configuration value here.

redash/query_runner/uptycs.py
vibhorkum: Fixed.

redash/query_runner/uptycs.py
vibhorkum: Ah! I missed this one. Fixed

redash/query_runner/uptycs.py
vibhorkum: Added the title and also change SSL to verify_ssl

redash/query_runner/uptycs.py
vibhorkum: Renamed to verify_ssl

vibhorkum:
Hi @arikfr Based on your valuable comments fixed all the issues.

Thanks,
Vibhor

arikfr:
Merged 👍 

Thank you for the the contribution and for the patience during the review :)

vibhorkum:
Thank you! And Thank you for all the feedback!

PR #3320, opened 2019-01-21
Fix Percy data source page always showing diff

gabrieldutra:
## Description
This snapshot is taken after submitting the form and a redirect happens at this moment. This was causing Percy diffs due to loading differences (such as the loading bar). I added some waiting time to avoid this.

I don't think the best practice would be to use `cy.wait` for this purpose, but I still haven't figured a way out of this haha.

~WIP: Testing if this actually works~ :slightly_smiling_face: 
Edit: Yep, no loading bar :grin:


arikfr:
Thanks!

PR #3323, opened 2019-01-22
Fix an error of exporting dict value as Excel

### LINKED ISSUE #3321: Exporting to Excel file fails when one of the columns is a dictionary
### Issue Summary

get error when exporting query results to excel file:
image
environment:ec2 on ecs

### Steps to Reproduce

1. create a new query
2. excute the query,save it ,then download it as excel file

### Technical details:

* Redash Version:6.0.0+b8537
* Browser/OS:chrome
* How did you install Redash:run redash by  ecs  on aws

kyoshidajp:
Fixes #3321


arikfr:
Thanks!

PR #3324, opened 2019-01-22
Migrate PageHeader component to React

kravets-levko:
Migrate `PageHeader` component to React; look and behavior didn't change.


client/app/components/PageHeader.jsx
```diff
@@ -0,0 +1,27 @@
...
+  );
+}
+
+PageHeader.propTypes = {
+  title: PropTypes.string,
+};
+
+PageHeader.defaultProps = {
+  title: '',
+};
```
arikfr: Just make title required. Why would anyone use PageHeader without a title?

PR #3330, opened 2019-01-23
[Widget params] Filtering out incompatible dashboard params

ranbena:
#### 🐞Bug
Reported by @arikfr https://github.com/getredash/redash/pull/3306#issuecomment-456383734
When editing or adding a new widget, under the "map to existing param", the dropdown lists all current dashboard params regardless of their input type. So it's possible to map a text param to a date param.

#### Fix
Sending `<ParameterMappingListInput>` an array of `{ name, type }` and then filtering out the non-matching ones for `<ParameterMappingInput>`.

### DIFF
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -223,7 +223,10 @@ export class ParameterMappingInput extends React.Component {
 export class ParameterMappingListInput extends React.Component {
   static propTypes = {
     mappings: PropTypes.arrayOf(PropTypes.object),
-    existingParamNames: PropTypes.arrayOf(PropTypes.string),
+    existingParams: PropTypes.arrayOf(PropTypes.shape({
+      name: PropTypes.string,
+      type: PropTypes.string,
+    })),
     onChange: PropTypes.func,
     clientConfig: PropTypes.any, // eslint-disable-line react/forbid-prop-types
     Query: PropTypes.any, // eslint-disable-line react/forbid-prop-types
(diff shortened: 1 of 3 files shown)


arikfr:
image

😒

arikfr:
@ranbena but... React would've told you that there is a mismatch.

arikfr:
@ranbena maybe I don't understand the example, but isn't the whole purpose of PropTypes is to define what data the component expects and raise an error if it doesn't match the definition?

ranbena:
Yeah but it's only in the component scope. For example, this should but won't yield a compile error:

```jsx
class Class1 extends React.Component {
  static propTypes = {
    prop1: PropTypes.number,
  }
  
  render() {
    return this.props.prop1 * 2; // <-- expects number, returns NaN
  }
}

ReactDOM.render(
  <Class1 prop1="SomeString" />, // <-- wrong data type, no error
  mountNode
);
```

ranbena:
While we're on the subject - all these type checks and risky assumptions in https://github.com/getredash/redash/pull/3331 are needless with static typing.

https://github.com/getredash/redash/blob/c3a7cdd737acd92b304c08c19307c49ee58cd89e/client/app/components/DateTimeRangeInput.jsx#L20-L24

arikfr:
@ranbena 

> Yeah but it's only in the component scope. For example, this should but won't yield a compile error:

Huh? It generates the following:

> Warning: Failed prop type: Invalid prop `prop1` of type `string` supplied to `Class1`, expected `number`.

https://codesandbox.io/s/4z3x78wll4

arikfr:
Oh, you mentioned a "compile error". Yes, it's not a compile error. But still noticeable.

arikfr:
> While we're on the subject - all these type checks and risky assumptions in #3331 are needless with static typing.

That's true. 🤔

ranbena:
> But still noticeable.

Only if you inspect the console while manually testing all scenarios - that's risky business.

PR #3331, opened 2019-01-23
Convert Angular services to CommonJS-style and use them in React components instead of injecting

kravets-levko:
1. Refine `Auth` services (`currentUser`, `clientConfig`):
1.1. remove dead code (related to `window.sessionStorage` which is not used);
2.2. fix race condition: `clientConfig` and `currentUser` object where filled with data when Angular needed to inject them (=== when first component that uses them was creating). Now both objects are always valid, but empty prior to a call to `Auth` service's methods.
2. Refine services:
2.1. Users (cleanup);
2.2. ~~Events (move `recordEvent` to this service);~~ move `recordEvents` from `lib` to `services`;
2.3. OfflineListener: remove Angular service and just subscribe to events;
2.4. Notifications: rewrite in CommonJS style.
3. Components: use imports instead of Angular DI


client/app/services/notifications.js
```diff
@@ -1,8 +1,10 @@
 import debug from 'debug';
 
+export let Notifications = null; // eslint-disable-line import/no-mutable-exports
+
 const logger = debug('redash:notifications');
 
-function Notifications(currentUser, Events) {
+function NotificationsService(currentUser, Events) {
```
arikfr: Because `Events` can be replaced with `recordEvent` this doesn't have to be an Angular service at all. We can import it directly in `view.js` (its only user).

client/app/services/offline-listener.js
```diff
@@ -1,4 +1,6 @@
-function OfflineListener(toastr) {
+export let OfflineListener = null; // eslint-disable-line import/no-mutable-exports
```
arikfr: No need to export this one. Nothing uses it, it just registers itself to some event and then shows a notification when the user is offline.

client/app/services/auth.js
```diff
@@ -1,32 +1,40 @@
 import debug from 'debug';
+import { includes, extend } from 'lodash';
 
-const logger = debug('redash:auth');
```
arikfr: It's a bit hard to follow what changed here from the diff, can you summarize the changes?

kravets-levko:
Thank you @arikfr for review. Tomorrow I planned to test everything and fix tests (that's why it's still WIP :slightly_smiling_face:), and also will check and fix the things you mentioned.

ranbena:
@arikfr Percy false-negative? https://percy.io/Redash/redash/builds/1354890

arikfr:
@ranbena re. Percy: looks like it. Been seeing this more recently. I pinged their support about this.

kravets-levko:
@ranbena 

> The PR is too huge for me to thoroughly run through all the changes though :/
It would be easier if it was split into commit per service.

I don't think it's actually needed - the only major change is in Auth service (and a bit later I'll describe what I changed there), in all others I just follow the template:
```javascript
export let ServiceName = null;
...
  ...
  ngModule.run(($injector) => {
    ServiceName = $injector.get('ServiceName');
  });
}
```

@arikfr - WDYT - should I split this PR into smaller ones or keep as is?

arikfr:
You can keep as is. It could be nice to split the Auth changes into : 1) changes to functionality, 2) support for import. But 🤷‍♂️

kravets-levko:
Okay, I'll split it, even into three PRs:

1. Auth changes to functionality
2. support for import (all services)
3. fixes to React components (replace DI with imports).

I'll convert this one into _"Auth changes to functionality"_, and will create new ones for the rest

arikfr:
@kravets-levko no need for 3 PRs. Just 3 commits.

client/app/services/events.js
```diff
@@ -1,13 +1,38 @@
...
+const sendEvents = debounce(() => {
+  const eventsToSend = events;
+  events = [];
+
+  $http.post('api/events', eventsToSend);
+}, 1000);
+
+function EventsService() {
   this.record = (action, objectType, objectId, additionalProperties) => {
-    recordEvent(action, objectType, objectId, additionalProperties);
```
arikfr: Why move it back into `Events`? There is no point in having an object with a single function in it. The idea was to delete this service when we drop Angular and stay only with `recordEvent`.

client/app/index.js
```diff
@@ -13,8 +13,7 @@ ngModule.config(($locationProvider, $compileProvider, uiSelectConfig, toastrConf
 });
 
 // Update ui-select's template to use Font-Awesome instead of glyphicon.
-// eslint-disable-next-line no-unused-vars
-ngModule.run(($templateCache, OfflineListener) => {
```
arikfr: Now that you removed `OfflineListener` here you need to explicitly import it.

client/app/services/events.js
arikfr: It should be its own file, because eventually there will be no `Events` service. At this point, we can just delete the `events.js` file and not worry about updating imports.

Unlike other services, here the name is different.

client/app/index.js
kravets-levko: No (actually, I tested it - it works :smile:) - it's inside init() function and is automatically invoked on application startup

arikfr: Oh, right. I forgot about my own magic... ✨

client/app/services/events.js
kravets-levko: @arikfr Done

client/app/services/events.js
```diff
@@ -1,30 +1,9 @@
-import { debounce, extend } from 'lodash';
-import { $http } from '@/services/ng';
+import recordEvent from '@/services/record-event';
 
 export let Events = null; // eslint-disable-line import/no-mutable-exports
```
arikfr: This is no longer needed, right?

client/app/services/record-event.js
```diff
@@ -0,0 +1,25 @@
+import { debounce, extend } from 'lodash';
```
arikfr: Filename should be `recordEvent.js`. See: 
https://github.com/airbnb/javascript#naming--filename-matches-export

kravets-levko: Well, we don't follow this rule almost nowhere :smile:

client/app/services/events.js
kravets-levko: Correct, thanks

client/app/services/record-event.js
arikfr: We follow it in all _new_ files since we adopted Airbnb's styleguide/rules. Old files followed a different rule.

arikfr:
📦

PR #3332, opened 2019-01-23
[Widget Params] Switched parameter list to table style (all parts)

ranbena:
Play with it in https://deploy-preview-3332--redash-preview.netlify.com.

### What changed and why?
* Params in Table view and edit functions in popovers. Better UX.
* Source type dropdown changed to radio buttons. I think it's extremely valuable for clarity to have all options laid out instead of hidden inside the dropdown.
* When no existing params, instead of hiding the option, it is disabled and has info available.
* "OK" button disables on form error.
* Source type labels are more coherent. @susodapop lmk if you have any suggestions here.
* If current keyword not in existing params, don't show it as an option. This was possible before but makes no sense as it essentially means adding a new dashboard param.

screen shot 2019-01-31 at 12 24 01

### DIFF
--- a/client/app/assets/less/ant.less
+++ b/client/app/assets/less/ant.less
@@ -13,6 +13,10 @@
 @import '~antd/lib/radio/style/index';
 @import '~antd/lib/time-picker/style/index';
 @import '~antd/lib/pagination/style/index';
+@import '~antd/lib/table/style/index';
+@import '~antd/lib/popover/style/index';
+@import '~antd/lib/icon/style/index';
+@import '~antd/lib/tag/style/index';
 @import 'inc/ant-variables';
 
 // Remove bold in labels for Ant checkboxes and radio buttons
--- a/client/app/assets/less/ant.less
+++ b/client/app/assets/less/ant.less
@@ -17,6 +17,7 @@
 @import '~antd/lib/popover/style/index';
 @import '~antd/lib/icon/style/index';
 @import '~antd/lib/tag/style/index';
+@import '~antd/lib/grid/style/index';
 @import 'inc/ant-variables';
 
 // Remove bold in labels for Ant checkboxes and radio buttons
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -1,11 +1,19 @@
 /* eslint react/no-multi-comp: 0 */
 
 import { extend, map, includes, findIndex, find, fromPairs } from 'lodash';
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import Select from 'antd/lib/select';
+import Table from 'antd/lib/table';
+import Popover from 'antd/lib/popover';
+import Button from 'antd/lib/button';
+import Icon from 'antd/lib/icon';
+import Tag from 'antd/lib/tag';
 import { ParameterValueInput } from '@/components/ParameterValueInput';
 import { ParameterMappingType } from '@/services/widget';
+import { Parameter } from '@/services/query';
+
+import './ParameterMappingInput.less';
 
 const { Option } = Select;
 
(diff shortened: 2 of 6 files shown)


client/.eslintrc.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
     'no-param-reassign': 0,
     'no-mixed-operators': 0,
     'no-underscore-dangle': 0,
+    "object-curly-newline": "off",
```
ranbena: What a nuisance this is.. doesn't allow:
```js
const { var1, var2, var4, var4, var5 } = this.props;
```

client/app/components/ParameterMappingInput.jsx
```diff
@@ -206,20 +214,83 @@ export class ParameterMappingInput extends React.Component {
   render() {
     const { mapping } = this.props;
     return (
-      <div key={mapping.name} className="row">
-        <div className="col-xs-5">
-          <div className="form-control-static">{'{{ ' + mapping.name + ' }}'}</div>
-        </div>
```
ranbena: Moved to "Keyword" table column

client/app/components/ParameterMappingInput.jsx
```diff
@@ -253,28 +359,81 @@ export class ParameterMappingListInput extends React.Component {
...
+                      Dashboard parameter{' '}
+                      <Tag className="tag">{mapping.mapTo}</Tag>
+                    </Fragment>
+                  );
+                case MappingType.WidgetLevel:
+                  return 'Widget parameter';
+                case MappingType.StaticValue:
+                  return 'Static value';
+                default:
+                  return ''; // won't happen (typescript-ftw)
```
ranbena: Typescript knows if your [switch case is exhaustive](https://www.typescriptlang.org/play/#src=enum%20ENUM%20%7B%0D%0A%20%20%20%20enum1%2C%0D%0A%20%20%20%20enum2%2C%0D%0A%7D%3B%0D%0A%0D%0Afunction%20foo(arg%3A%20ENUM)%3A%20string%20%7B%0D%0A%20%20%20%20switch(arg)%20%7B%0D%0A%20%20%20%20%20%20%20%20case%20ENUM.enum1%3A%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20return%20'enum1'%3B%0D%0A%20%20%20%20%20%20%20%20case%20ENUM.enum2%3A%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20return%20'enum2'%3B%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%20), so it's safe to ditch `default`.

client/app/components/ParameterMappingInput.jsx
```diff
@@ -240,6 +311,41 @@ export class ParameterMappingListInput extends React.Component {
...
+    // rest
+    return value.toString();
+  }
+
+  static getDefaultValue(mapping) {
+    const value = mapping.type === MappingType.StaticValue
+      ? mapping.value || mapping.param.normalizedValue
+      : mapping.param.normalizedValue;
+
+    return this.getStringValue(value);
```
ranbena: Enum, string, date, date range, number
screen shot 2019-01-23 at 21 49 48

I made it so _all_ date types would format to DD/MM/YY. Lmk if you think it needs refining.

client/app/components/ParameterMappingInput.less
```diff
@@ -0,0 +1,43 @@
...
+    font-weight: 500;
+    color: @heading-color;
+  }
+
+  footer {
+    border-top: @border-width-base @border-style-base @border-color-split;
+    padding: 10px 16px 0;
+    margin: 20px -16px 0;
+    text-align: right;
+  }
```
ranbena: Made the `<Popover>` look like a `<Modal>`

arikfr:
> This is the first of a batch of 3 commits that change UX of widget params.

Are you sure we should be testing this without the other two?

client/.eslintrc.js
kravets-levko: It requires newlines if there are too much fields (IIRC, 4 and more) :slightly_smiling_face:

client/app/assets/less/ant.less
```diff
@@ -13,6 +13,10 @@
 @import '~antd/lib/radio/style/index';
 @import '~antd/lib/time-picker/style/index';
 @import '~antd/lib/pagination/style/index';
+@import '~antd/lib/table/style/index';
+@import '~antd/lib/popover/style/index';
+@import '~antd/lib/icon/style/index';
+@import '~antd/lib/tag/style/index';
```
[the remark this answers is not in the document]
kravets-levko: Let's review this when we remove Angular and Bootstrap :smiling_imp:

client/app/components/ParameterMappingInput.jsx
kravets-levko: Add `// no default` [comment](https://eslint.org/docs/rules/default-case) instead

client/app/components/ParameterMappingInput.jsx
kravets-levko: You should use format from `clientConfig.dateFormat`:
https://github.com/getredash/redash/blob/master/client/app/components/DateInput.jsx#L14
https://github.com/getredash/redash/blob/master/client/app/components/DateTimeInput.jsx#L15-L16

client/app/components/ParameterMappingInput.jsx
ranbena: But then if a new `MappingType` is introduced, this would yield a runtime error. So I'd rather leave it.

(What static typing allows is to omit `default` safely - by getting a compile error `Not all code paths return a value.` if not all cases have been handled.)

ranbena:
> Are you sure we should be testing this without the other two?

Each part would demand different testing. But it doesn't really matter - I won't land them separately anyhow.

client/app/components/ParameterMappingInput.jsx
ranbena: @kravets-levko I found a way around it, leveraging the already formatted `param.value` instead of the raw `param.normalizedValue`. So now, not only is the date correctly formatted, but also shows time and seconds when relevant.

https://github.com/getredash/redash/pull/3332/commits/48743c7eda6da7aa15d1698ee30d69e3bc4c80e5

client/.eslintrc.js
ranbena: Yeah, it makes no sense.. that's what we got `max-len` for.

client/app/components/ParameterMappingInput.jsx
arikfr: Please make sure that you're using `clientConfig.dateFormat` only when showing it to the user. When setting the value for the parameter you need to use the static format we were using (`YYYY-MM-DD HH:mm:ss`).

ranbena: In 48743c7 there's no formatting being done. Just taking the already formatted `mapping.param.value`.

ranbena: Damn, there's a case where static value is not yet normalized and it turns to shit. I hope I don't have to resort to dupe-formatting 😭

ranbena:
I've been delayed with completing this cause of a bug I found, but now realized it happens in master #3355.

client/app/components/ParameterMappingInput.jsx
ranbena: This has been resolved :)

arikfr:
:shipit:

Issue #2810, opened 2018-09-12
Allow installing additional dependencies in Docker deployment


jezdez:
### Issue Summary

To be able to use the Python extension API (and soon the frontend extension API) it would be useful to be able to easily install extra dependencies when using the Redash docker image. I wonder if we could add an optional environment variables, e.g. `REDASH_EXTRA_PYTHON_REQUIREMENTS` and the analog way for Node dependencies, to be picked up when inheriting from the main Redash Docker image.

It's simple enough to do that right now by simply inheriting from the Redash image and running additional commands, e.g.:

```dockerfile
FROM redash/redash:latest

COPY requirements_extra.txt /tmp/
RUN pip install -r /tmp/requirements_extra.txt
```

If we want to encourage users to use the extension APIs in their projects, a guideline how to integrate it in the Docker based deployment workflow would alleviate the risk of the application bundle build process from breaking.

The above example is simple enough, but when considering that the [current Dockerfile](https://github.com/getredash/redash/blob/93b34bcbce2b3df328e1c1fac4261fa83e8717b3/Dockerfile#L6-L9) installs the Node dependencies only to run build the application bundle and then deletes `node_modules/` again, it's a lot more brittle that I think is worthwhile going forward.

For the sake of consistency between the Python and Node dependencies it would make sense to have a common API to add additional dependencies using the parent Redash Docker image.

Maybe an ideal way would be to use the [`ON BUILD`](https://docs.docker.com/engine/reference/builder/#onbuild) command to do the installation and build process if certain environment variables are defined? OTOH the `*-onbuild` images of the [various language images have recently been deprecated](https://github.com/docker-library/official-images/issues/2076), so maybe this is betting on the wrong horse.

Also mentioned in that ticket, Sentry uses `ON BUILD` in their ["onpremise" Docker image](https://github.com/getsentry/onpremise/blob/master/Dockerfile) though, which seems to be an accepted use case. The base onbuild image contains a few [`ON BUILD` commands to install additional dependencies if available](https://github.com/getsentry/docker-sentry/blob/f648178adf33bec68be47550a82850e18deb0cd8/9.0/onbuild/Dockerfile#L9-L13):

```dockerfile
FROM sentry:9.0

# ...

# Hook for installing additional plugins
ONBUILD RUN if [ -s requirements.txt ]; then pip install -r requirements.txt; fi

# Hook for installing a local app as an addon
ONBUILD RUN if [ -s setup.py ]; then pip install -e .; fi
```

[...]

PR #3333, opened 2019-01-23
adding gevent worker in requirements.txt, adding some gunicorn config…

ialeinikov:
* gevent is added to the requirements.txt
* some gunicorn configuration parameters are added with defaults


jezdez:
We had a similar case in #2743 where we noted that it doesn't require a new env var and that `GUNICORN_CMD_ARGS`is enough to enable this ability. @ialeinikov Could you try it out please?

From http://docs.gunicorn.org/en/stable/settings.html:

> Settings can be specified by using environment variable GUNICORN_CMD_ARGS.
> All available command line arguments can be used. For example, to specify the bind 
> address and number of workers:
>
> $ GUNICORN_CMD_ARGS="--bind=127.0.0.1 --workers=3" gunicorn app:app
>
> New in version 19.7.

arikfr:
Oops, I completely forgot about this. 😳

arikfr:
Although this Pull Request adds `gevent` to the requirements, so maybe worth keeping.

jezdez:
Yeah, the extra dependency is useful I guess, until we solve #2810 😃 OTOH using a custom `Dockerfile` with an added requirements installation step would work as well.

arikfr:
@jezdez it's true that a custom Dockerfile with a single `RUN pip install gevent==...` line will do the trick. I think we will move to using the gevent workers as the default setup, so I don't mind adding it already.

tonidas:
The performance is WAY better with gevent, there is no comparison between delegating a process to each request than delegating a thread. I've created a new redash image to add the dependency

Looking forward for this merge @ialeinikov !

arikfr:
Thanks!

PR #3334, opened 2019-01-24
[Widget Params] Split title and mapping editing (2 of 3)

ranbena:
This is the second of a batch of 3 commits that change UX of widget params.

#### Why?
Currently, there is one edit button per parameter, opening a popover that allows editing both source type and title. A better user experience is to have edit buttons per editable content - one for title, one for source type.

#### How it looks
Title edit
screen shot 2019-01-24 at 13 50 03

Source type edit
screen shot 2019-01-24 at 13 51 33

#### Implementation notes
Since now popovers open to the right of the content, affecting data as you type makes the popover jump horizontally as the position of the edit button shifts.
So, I implemented a "cancel/save" to each popover, hence the underlying table stays put during editing.

### DIFF
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -229,9 +229,14 @@ class EditMapping extends React.Component {
     existingParamNames: PropTypes.arrayOf(PropTypes.string).isRequired,
     onChange: PropTypes.func.isRequired,
     getContainerElement: PropTypes.func.isRequired,
-    clientConfig: PropTypes.any.isRequired, // eslint-disable-line react/forbid-prop-types
-    Query: PropTypes.any.isRequired, // eslint-disable-line react/forbid-prop-types
+    clientConfig: PropTypes.any, // eslint-disable-line react/forbid-prop-types
+    Query: PropTypes.any, // eslint-disable-line react/forbid-prop-types
   };
+
+  static defaultProps = {
+    clientConfig: null,
+    Query: null,
+  }
 
   constructor(props) {
     super(props);
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -3,6 +3,7 @@
 import { extend, map, includes, findIndex, find, fromPairs } from 'lodash';
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
+import moment from 'moment';
 import Select from 'antd/lib/select';
 import Table from 'antd/lib/table';
 import Popover from 'antd/lib/popover';
@@ -12,6 +13,8 @@ import Tag from 'antd/lib/tag';
 import { ParameterValueInput } from '@/components/ParameterValueInput';
 import { ParameterMappingType } from '@/services/widget';
 import { Parameter } from '@/services/query';
+
+import './ParameterMappingInput.less';
 
 import './ParameterMappingInput.less';
 
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -1,11 +1,19 @@
 /* eslint react/no-multi-comp: 0 */
 
 import { extend, map, includes, findIndex, find, fromPairs } from 'lodash';
-import React from 'react';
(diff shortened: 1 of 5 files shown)


client/app/components/ParameterMappingInput.jsx
```diff
@@ -149,7 +150,7 @@ export class ParameterMappingInput extends React.Component {
       <div className="m-t-10">
         <Select
           className="w-100"
-          defaultValue={mapping.mapTo}
+          value={mapping.mapTo}
```
ranbena: This is so that if edit is cancelled (making `state.mapping` reset to original `props.mapping`), it would affect these input components, or else they would have stayed as left.

client/app/components/ParameterMappingInput.jsx
```diff
@@ -191,32 +192,12 @@ export class ParameterMappingInput extends React.Component {
...
-          type="text"
-          className="form-control"
-          value={mapping.title}
-          onChange={event => this.updateParamMapping(mapping, { title: event.target.value })}
-          placeholder={mapping.param.title}
-        />
-      </div>
-    );
-  }
-
```
ranbena: Bye bye!

client/app/components/ParameterMappingInput.jsx
```diff
@@ -412,22 +492,23 @@ export class ParameterMappingListInput extends React.Component {
...
-                      <Tag className="tag">{mapping.mapTo}</Tag>
-                    </Fragment>
-                  );
-                case MappingType.WidgetLevel:
-                  return 'Widget parameter';
-                case MappingType.StaticValue:
-                  return 'Static value';
-                default:
-                  return ''; // won't happen (typescript-ftw)
-              }
```
ranbena: Moved to `getSourceTypeLabel`

arikfr:
@ranbena I can review this one directly instead of the previous one, right?

ranbena:
Yeah, although it's still incomplete until the dependency is fixed.
And I need to rebase it.

ranbena:
@arikfr ready!

arikfr:
1.

I think it makes sense to use a wider Modal:

image

Update: I actually noticed we already use a wider one when you edit parameters.

2.

image

Instead of "Change parameter value" maybe "Set Parameter value"?

3. 

When adding the widget to a dashboard with a static value it complaints that no value is set for `param`, while if you refresh it (with the refresh button) it works:

image

4.

(NTH)

image


When the Popover is open, clicking on Esc closes the Modal instead of the Popover.

5. 

*Completely unrelated* but I think we should pin to top (along with the header) the parameters input. I added a new widget that takes a dashboard parameter, but the input was not even seen on screen. Feels a bit disorienting. I will open an issue for this.

6. 

(should be fixed but not a blocker:)

1. Edit parameters of an existing widget.
2. Map it to a different parameter that has a different value.
3. The widget shows the old value. Expected: refresh and show value with the new parameter.

7.

We use the `p_` prefix for all parameter in the URL, but for widgets we use only `w{widget-id}_`. Shouldn't it be `p_w{widget-id}_`?

ranbena:
> I think it makes sense to use a wider Modal:

My only reservation is that the above inputs (query and visualization) would be _incredibly_ long.
Lemme think on this.

> Instead of "Change parameter value" maybe "Set Parameter value"?

This is trampled in the next commit #3377.

> When adding the widget to a dashboard with a static value it complaints that no value is set for `param`, while if you refresh it (with the refresh button) it works:

That's a bug on master. I'll open a separate issue.

> When the Popover is open, clicking on Esc closes the Modal instead of the Popover.

Right. I'll have a separate PR to solve this with an Ant Modal.

> (should be fixed but not a blocker:)
> 1. Edit parameters of an existing widget.
> 2. Map it to a different parameter that has a different value.
> 3. The widget shows the old value. Expected: refresh and show value with the new parameter.

Definitely a bug. I'll look into it.

> We use the `p_` prefix for all parameter in the URL, but for widgets we use only `w{widget-id}_`. Shouldn't it be `p_w{widget-id}_`?

Definitely.  Will open a separate issue.

ranbena:
Moved to #3332

PR #3337, opened 2019-01-24
Unify query based dropdown population

### LINKED ISSUE #3316: Combine frontend/backend query-based dropdown parameters
Currently, query-based dropdown lists are populated in the frontend using an API call that fetches the values and some logic which is implemented in the frontend. The same kind of logic is introduced in #3315 so it's probably best to unite those and create a dedicated endpoint that returns query-based dropdown list values.

rauchy:
This PR consolidates the backend and frontend logic for fetching a list of permitted values for a query-based dropdown parameter.

TODO:

- [x] add tests 🤭

Fixes #3316


redash/handlers/api.py
```diff
@@ -69,10 +69,10 @@ def json_representation(data, code, headers=None):
 
 api.add_org_resource(EventsResource, '/api/events', endpoint='events')
 
-api.add_org_resource(QueryFavoriteListResource, '/api/queries/favorites', endpoint='query_fovorites')
-api.add_org_resource(QueryFavoriteResource, '/api/queries/<query_id>/favorite', endpoint='query_fovorite')
-api.add_org_resource(DashboardFavoriteListResource, '/api/dashboards/favorites', endpoint='dashboard_fovorites')
-api.add_org_resource(DashboardFavoriteResource, '/api/dashboards/<object_id>/favorite', endpoint='dashboard_fovorite')
+api.add_org_resource(QueryFavoriteListResource, '/api/queries/favorites', endpoint='query_favorites')
```
rauchy: 🤪

redash/handlers/api.py
```diff
@@ -89,6 +89,7 @@ def json_representation(data, code, headers=None):
 api.add_org_resource(CheckPermissionResource, '/api/<object_type>/<object_id>/acl/<access_type>', endpoint='check_permissions')
 
 api.add_org_resource(QueryResultListResource, '/api/query_results', endpoint='query_results')
+api.add_org_resource(QueryResultDropdownResource, '/api/queries/<query_id>/dropdown.json', endpoint='query_result_dropdown')
```
arikfr: Why the .json suffix?

redash/handlers/query_results.py
```diff
@@ -131,6 +131,13 @@ def post(self):
 ONE_YEAR = 60 * 60 * 24 * 365.25
 
 
+class QueryResultDropdownResource(BaseResource):
+    def get(self, query_id):
+        data = dropdown_values(query_id, self.current_org)
+        headers = {'Content-Type': "application/json"}
+        return make_response(json_dumps(data), 200, headers)
```
arikfr: `BaseResource` takes care of converting an object to JSON. So you can do just:

```python
    return dropdown_values(...)
```

redash/utils/parameterized_query.py
```diff
@@ -1,9 +1,29 @@
...
-from redash.utils import mustache_render
+from redash import models
+from redash.utils import mustache_render, json_loads
+from redash.permissions import require_access, view_only
 from funcy import distinct
 from dateutil.parser import parse
 
+
+def dropdown_values(query_id, org):
+    def _pluck_name_and_value(row):
```
arikfr: I would move this function to be at the module level. Defining functions inside functions always felt dirty to me (not that I haven't done it...).

redash/utils/parameterized_query.py
```diff
@@ -1,9 +1,29 @@
...
+        name_column = "name" if "name" in row.keys() else row.keys()[-1]
```
arikfr: In the JS version we were:

1. Performing case insensitive search (converting row keys to lower case).
2. Using the _first_ column name by default (-1 is the last, no?).

tests/utils/test_parameterized_query.py
```diff
@@ -131,3 +158,21 @@ def test_raises_on_unexpected_param_types(self):
 
         with pytest.raises(InvalidParameterError):
             query.apply({"bar": "baz"})
+
+    @patch('redash.models.QueryResult.get_by_id_and_org', return_value=namedtuple('query_result', 'data')('{"rows": [{"id": 5, "name": "John", "value": "John Doe"}]}'))
+    @patch('redash.utils.parameterized_query.require_access')
+    @patch('redash.utils.parameterized_query.view_only')
+    @patch('redash.utils.parameterized_query.current_user')
+    @patch('redash.models.Query')
+    def test_dropdown_values_prefers_name_and_value_columns(self, *_):
```
arikfr: And here's why moving that pluck function to the module level would've been beneficial: you could've tested it directly without all these mocks.

arikfr: (I know you prefer to test the external API, but in this case it feels just too much)

redash/handlers/query_results.py
rauchy: 95b56109

redash/utils/parameterized_query.py
rauchy: 1. Roger that.
2. It is taking the first column (see [this test](https://github.com/getredash/redash/pull/3337/files#diff-373dcffefcbd523d468cf8afa3a45720R176)). I believe this has to do with the order of key additions in `json_loads`. Anyway, it works as expected.

redash/utils/parameterized_query.py
arikfr: Lambdas make sense if it's a short logic (one line), which in this case it isn't.

I don't see a reason to place functions in the scope of another function aside for the purpose of closures.

redash/utils/parameterized_query.py
arikfr: > It is taking the first column (see this test). I believe this has to do with the order of key additions in json_loads. Anyway, it works as expected.

It's not. Try changing the JSON into:

```
'{"rows": [{"zid": 5, "fish": "Clown", "poultry": "Hen"}]}'
```

You will notice that `zid` is no longer the last entry, but the 2nd (⁉️).

The order of keys in a Python dictionary can't be trusted. Considering that in the view we order the columns by their order in the `columns` array, maybe we should just use that here as well?

redash/utils/parameterized_query.py
rauchy: I don't mind taking this out to a private module function, but I'm still against testing it directly 😛

tests/utils/test_parameterized_query.py
arikfr: I agree with the general idea here of not testing implementation details. But maybe this function should be two functions composed together when used?

```python
def load_query_latest_result(query_id):
    ... magic ...

def dropdown_values_from_query_result(query_result):
    ... darker magic ...

# then in the dropdown handler:
class QueryDropDownResource(BaseResource):
    def get(query_id):
          # QueryResult#json is a new property that will return data post json_load
          return dropdown_values_from_query_result(load_query_latest_result(query_id).json)
```

Each function on its own has a clear purpose. You can say that "extracting dropdown values" is an implementation detail that should be hidden, because that's the whole thing here. So having it in its own function won't slow down changes. If it all, it will make them easier, because you don't need to worry about all the aspects of loading the query result.

Then you need the heavy mocking only once, in the integration test for `QueryDropDownResource`.

redash/utils/parameterized_query.py
arikfr: I'll take this :)

tests/utils/test_parameterized_query.py
rauchy: *** cracks knuckles ***

Yes, separation of concerns would be beneficial here. I definitely think we should extract all the data loading to a helper function and then `dropdown_values` would be basically:

```
rows = _load_result(query_id, org)
return map(_pluck_name_and_value, rows)
```

Regarding testing of pluck logic - when I find myself wanting to test private stuff, my method has always been asking myself “is this such an important piece of logic that I have to test it directly?”. If the answer is yes, it means that the test subject is _too important_ to be hidden inside something else - it is probably a domain concept that needs to be promoted to its own place.

In this case, I think that your motivation for promoting plucking logic is not due to importance (dropdown values is a really small concept for Redash), but due to the fuglyness caused by multiple calls to `patch`.

I’m going to guess you will find 739c65e3 more pleasing.

arikfr: It looks like we're in an agreement here on the concepts. It's just the details we're seeing in a different way. It's not about how ugly the tests are (we have plenty of those...). They were just the signal.

I'm not saying the tests should test the pluck logic, I'm saying the tests should test the "convert query results to dropdown values" logic. Pluck is just an implementation detail that is definitely not interesting on its own.

Also, the query result loading logic is very important as well (data security) and should be tested as well...

arikfr: Anyway, up to you how to handle this, but the important thing is to test the security aspects.

redash/utils/parameterized_query.py
```diff
@@ -1,9 +1,34 @@
...
+    row = {k.lower(): v for k, v in row.items()}
+    name_column = "name" if "name" in row.keys() else row.keys()[-1]
+    value_column = "value" if "value" in row.keys() else row.keys()[-1]
+
+    return {"name": row[name_column], "value": row[value_column]}
+
+
+def _load_data(query_id, org):
+    query = models.Query.get_by_id_and_org(query_id, org)
+    require_access(query.data_source.groups, current_user, view_only)
```
arikfr: If you're using `current_user`, then I guess you can use `current_org` as well.

It does mean that we won't be able to run this outside the context of a Flask request, but at the moment I think there are no such cases?

rauchy: I didn't know there was a `current_org`. Yeah, I can't think of a case where we would use this outside a request right now. That's addressable if the need ever comes up.

PR #3339, opened 2019-01-25
Refine SettingsMenu service and <settings-screen> component

kravets-levko:
- Move some logic from `<settings-screen>` component to `SettingsMenu` service;
- refine `SettingsMenu` service and move it to `/services` folder.

### DIFF
--- a/client/app/services/settingsMenu.js
+++ b/client/app/services/settingsMenu.js
@@ -19,21 +19,12 @@ class SettingsMenuItem {
 
 class SettingsMenu {
   constructor() {
-    this._items = [];
-    this._dirty = false;
-  }
-
-  get items() {
-    if (this._dirty) {
-      this._items = sortBy(this._items, 'order');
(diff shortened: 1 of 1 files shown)


client/app/services/SettingsMenu.js
```diff
@@ -0,0 +1,40 @@
...
+
+class SettingsMenu {
+  constructor() {
+    this._items = [];
+    this._dirty = false;
+  }
+
+  get items() {
+    if (this._dirty) {
+      this._items = sortBy(this._items, 'order');
```
arikfr: Why not sort when adding a new item and just use `this.items`? Feels like a premature optimization a bit.

kravets-levko: Yes, I agree. But - why not? This service is very small and probably will not grow at all. I just feel a bit uncomfortable when I see a code that sort something when new items added :slightly_smiling_face:

PR #3345, opened 2019-01-25
Refactor Policy and OrganizationStatus services

kravets-levko:
1. Convert `Policy` and `OrganizationStatus` services to CommonJS-style services (and fix usages).
2. `Policy` service should be configurable (previously Angular's DI did the trick)  ~~, therefore some new classes were introduced:
2.1. `AbstractPolicy`: abstract common parent (think about it as an interface; es6 has no a "true" interfaces support);
2.2. `DefaultPolicy`: default implementation for `AbstractPolicy` interface;
2.3. `services/policy` exports an instance of `PolicyProxy` which can accept a descendant of `AbstractPolicy` and forwards all method calls to it; it allows to transparently change policies implementation on the fly (native `Proxy` could make this part more clean, but `Proxy` still is poorly supported, so...).~~


ranbena:
@kravets-levko Soon.. [Typescript Interfaces](https://www.typescriptlang.org/docs/handbook/interfaces.html)

ranbena:
@kravets-levko perhaps consider a Proxy polyfill?

kravets-levko:
@ranbena It's impossible to implement a true `Proxy` polyfill due to its nature. Also from [Babel docs](https://babeljs.io/docs/en/learn/#ecmascript-2015-features-proxies):

> Due to the limitations of ES5, Proxies cannot be transpiled or polyfilled. See support in various JavaScript engines.

ranbena:
> it allows to transparently change policies implementation on the fly

@kravets-levko can you show me an example code usage of how to do that?

client/app/services/policy/AbstractPolicy.js
```diff
@@ -0,0 +1,56 @@
+export default class AbstractPolicy {
```
arikfr: We have a single use case for the Policy object and that we override/extend it in the SaaS version. The SaaSPolicy object inherits from Policy (now DefaultPolicy). I don't see any future case we will ever bother extending AbstractPolicy...

client/app/services/policy/index.js
```diff
@@ -0,0 +1,68 @@
...
+  getDashboardRefreshIntervals() {
+    return this.policy ? this.policy.getDashboardRefreshIntervals() : super.getDashboardRefreshIntervals();
+  }
+
+  getQueryRefreshIntervals() {
+    return this.policy ? this.policy.getQueryRefreshIntervals() : super.getQueryRefreshIntervals();
+  }
+}
+
+export default new PolicyProxy(new DefaultPolicy());
```
arikfr: Having this PolicyProxy adds a lot of complexity, where we could simple have:

```javascript
class Policy {
    ...

    updateRules(newPolicy) {
        extend(this, newPolicy);
    }
    
    ...
}
```

kravets-levko:
@arikfr I simplified implementation, can you please review again?

P.S. Something wrong with Percy snapshots, is it because of getredash/redash#3338?

client/app/services/policy/index.js
```diff
@@ -0,0 +1,8 @@
+import DefaultPolicy from './DefaultPolicy';
+
+// eslint-disable-next-line import/no-mutable-exports
+export let policy = new DefaultPolicy();
+
+export function setPolicy(newPolicy) {
+  policy = newPolicy instanceof DefaultPolicy ? newPolicy : new DefaultPolicy();
```
arikfr: If @ranbena sees this, he will immediately tag this as #typescript-ftw... 🤓 

But that aside, I'm worried this will create some race conditions that will be hard to spot, in case some other object holds on to the DefaultPolicy and "ignores" updates. I know this is what we do with the others, but in the case of the others, the default value is null so it's easier to spot issues.

Why not extend the default object with the new one?

gabrieldutra:
> P.S. Something wrong with Percy snapshots, is it because of #3338?

Actually with both #3338 and #3356, their CircleCI build on master probably run by the same time and for some reason #3356 finished last while it shouldn't. Anyway, I triggered some builds again and this should be fine soon here :slightly_smiling_face: (except for a diff for User API Key that I didn't see coming before :sweat_smile: )

client/app/services/policy/index.js
kravets-levko: Because just `extend`ing will lose `this` context for all methods - this is even more error-prone. I think I'll just throw an exception here.

kravets-levko: ...or just remove this check - hope nobody will put any garbage there

kravets-levko:
@arikfr Updated PR, can you please review it again?

kravets-levko:
@arikfr Yes, I tested it - I created a file with a fake implementation of SaaSPolicy (where each method just trows an error), added `init()` method and called `setPolicy` there (exactly the same as we should later do in SaaS fork) - it works as expected.

arikfr:
👍

PR #3348, opened 2019-01-25
Fix some incompatible dependencies

jezdez:
Not sure if this solves #3343 and it's unclear how much of boto3/botocore's API has been updated. The amount of changes in those is... staggering.

### DIFF
--- a/requirements_all_ds.txt
+++ b/requirements_all_ds.txt
@@ -11,7 +11,8 @@ td-client==0.8.0
 pymssql==2.1.3
 dql==0.5.24
 dynamo3==0.4.7
-botocore==1.10.2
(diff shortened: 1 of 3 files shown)


requirements_all_ds.txt
```diff
@@ -11,7 +11,8 @@ td-client==0.8.0
 pymssql==2.1.3
 dql==0.5.24
 dynamo3==0.4.7
-botocore==1.10.2
+boto3==1.9.85
+botocore==1.12.85
```
arikfr: Maybe we should specify only boto3 and let it install botocore for us?

arikfr:
Thanks.

PR #3354, opened 2019-01-26
React version of UserEdit

gabrieldutra:
## Description
React version for `UserEdit`. Not much yet, but I'll update this PR with my ideas.

## Changes
- [x] Update `DynamicForm` export to work in other React components
- [x] Migrate profile header
- [x] Render Basic Form with `DynamicForm`
- [x] API Key (I'm probably gonna move the regenerate button somewhere else and put a Copy Button where it is now - or even create an `InputWithCopy` component with this purpose, since this could be used in other places)
- [x] Change Password Modal
- [x] Add `minLength` to Password modal and validation for matching password (Can be improved)
- [x] Send Password Reset Email
- [x] Enable/Disable button
- [x] Resend Invitation
- [x] Disabled user behavior
- [x] Review behavioral rules
- [x] Fix UserShow snapshot test

**Extra**
- [ ] Use Ant `Avatar` instead of `profile__stuff` classes
- [x] Cypress tests :ok_hand:

**Code Review**
- [x] Small fixes
- [x] Fix possible errors when network is off
- [x] Improve User Feedback for Resend invitation - 50%

## Preview
**Overall**
overall

**Toggling User**
toggling-user

**Regular user view**
screenshot from 2019-02-07 23-05-56


arikfr:
I just merged your Percy PR. Can you update this one with master, so we can start seeing the changes to the user page on Percy? :)

client/app/components/users/UserEdit.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+    return (
+      <Modal
+        visible={this.state.changePassword}
+        title="Change Password"
+        onCancel={() => { this.setState({ changePassword: false }); }}
+        footer={null}
+        destroyOnClose
+      >
+        <DynamicForm fields={fields} saveText="Update Password" onSubmit={this.handleSave} />
```
arikfr: Interesting. I never thought we will use `DynamicForm` to render forms in the app (that are not based on backend properties). But I guess it makes sense... I would draw the line in cases where rendering with DynamicForm requires too many tweaks and it's better to use the components directly.

gabrieldutra: My thoughts are that components are a concept to avoid reuse of code. Forms can be pretty annoying to handle, so I tried to make `DynamicForm` in a way it would be an easy-to-use Object-To-Form component.

As for backend properties, it currently needs some mapping (improving how this is done is on my plans when migrating Data Sources & Destinations Page).

The biggest problem I see is that it's a lot easier to break a page you were not expecting to, and that's where creating tests take part :slightly_smiling_face:

gabrieldutra:
I created an `UserPolicy` to share the idea since a lot of "render logic" didn't seem very good, thought I have 2 concerns about it:
- I don't know where exactly to put this and it seemed to have some relevant discussion on #3345 
-  `canEditUser(user)` <- This user currently relies on a transition User (backend snake case attrs -> frontend camel case attrs) and this makes this Policy only usable in `UserEdit`

arikfr:
> I created an UserPolicy to share the idea since a lot of "render logic" didn't seem very good, thought I have 2 concerns about it:

Considering it's only used in this component, I'm not sure we should put it globally, but rather locally in the same file.

But we probably don't need this at all: in the past we used the same page both for admins/owner and for regular users. But now we have the UserShow component which we show to regular users, so anyone who sees this form can edit it (right?).

gabrieldutra:
> But we probably don't need this at all: in the past we used the same page both for admins/owner and for regular users. But now we have the UserShow component which we show to regular users, so anyone who sees this form can edit it (right?).

Yep, but I was more thinking about the specific view rules, e.g: User can edit his password or User can see the API key. I think it's interesting to have centralized rules (which makes changes easier), however for the User I can let it go since they are simple (basically `currentUser.isAdmin` and `user.isDisabled` permutations) and don't seem to be shared elsewhere.

arikfr:
> Yep, but I was more thinking about the specific view rules, e.g: User can edit his password or User can see the API key. I think it's interesting to have centralized rules (which makes changes easier), however for the User I can let it go since they are simple (basically currentUser.isAdmin and user.isDisabled permutations) and don't seem to be shared elsewhere.

Yep. Basically: 
1. User & admin can edit user properties.
2. For disabled users we hide some of the stuff.

Most of the complexity there is legacy we don't need.

cypress/integration/percy/page_screenshots.js
```diff
@@ -6,16 +6,26 @@ const pages = [
...
       cy.visit(page.url);
       cy.wait(1000);
       cy.percySnapshot(page.name);
     });
   });
+
+  it('takes a screenshot of User Profile', () => {
+    cy.visit('/users/me');
+    cy.getByTestId('ApiKey').then(($apiKey) => {
+      $apiKey.val('secret');
```
gabrieldutra: FYI @arikfr 
Revisited for a good reason :slightly_smiling_face:

client/app/components/users/UserEdit.jsx
```diff
@@ -0,0 +1,279 @@
...
+            The mail server is not configured, please send the following link
+            to {user.name} to reset their password:
+            <Input.TextArea value={absoluteUrl(passwordResetLink)} readOnly />
+          </p>
+        ) : 'The user should receive a link to reset their password by email soon.'}
+        type="success"
+        className="m-t-20"
+        afterClose={() => { this.setState({ passwordResetLink: null }); }}
+        closable
+      />
```
ranbena: Wdyt about having this in a Modal or Popover instead?

client/app/components/users/UserEdit.jsx
```diff
@@ -0,0 +1,279 @@
...
+    const data = {
+      id: this.props.user.id,
+      ...values,
+    };
+
+    User.save(data, (user) => {
+      successCallback('Saved.');
+      this.setState({ user: User.convertUserInfo(user) });
+    }, (error) => {
+      errorCallback(error.data.message || 'Failed saving.');
```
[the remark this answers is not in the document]
ranbena: Or maybe sth like?
```js
(error = {}) => {
  errorCallback(error.data && error.data.message || 'Failed saving.');
}
```

client/app/services/user.js
```diff
@@ -65,6 +65,65 @@ function deleteUser(user) {
...
+    });
+}
+
+function resendInvitation(user) {
+  return $http
+    .post(`api/users/${user.id}/invite`)
+    .then(() => {
+      toastr.success('Invitation sent.');
+    })
+    .catch((response) => {
```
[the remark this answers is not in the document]
gabrieldutra: I could not reproduce this with Resend Invitation, but I could with: User Save (change password too) and `toggleUser` methods. Anyway, I'll test this further by tomorrow and try to find out what is going on :thinking:

client/app/components/users/UserEdit.jsx
```diff
@@ -0,0 +1,279 @@
...
+    });
+  };
+
+  onUpdatePassword = (values, successCallback, errorCallback) => {
+    if (values.password === values.password_repeat) {
```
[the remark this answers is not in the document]
gabrieldutra: My first thought was to have it as a field rule, it's much better for UX. Antd offers an option for a Custom Validator and I've seen [a Chinese related discussion](https://github.com/ant-design/ant-design/issues/5155) with a sort of an example of what seemed to be what both of us want. By that point, it seemed to require some tweaks to make it work, thus I postponed.

Anyway, I'll try to implement this and I'll keep you updated :slightly_smiling_face:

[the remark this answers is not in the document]
ranbena: Let's put this at lowest priority.

client/app/components/users/UserEdit.jsx
ranbena: > Toastr with a shorter message for when the email server is configured

Definitely 👍 

> both a modal or a popover seem unexpected to me when I click on that button

I agree that the popover is irrelevant.
But I also think the current alert doesn't communicate the right message to the user.

We need to communicate:
1) Email not sent!
2) The reason is misconfiguration

And offer the user:
1) To fix configuration -  link to the relevant setting, perhaps a dedicated knowledge base help page.
2) Alternative by sending the link manually. With a [copy button](https://codesandbox.io/s/onxv7zxy5) perhaps.

client/app/components/users/UserEdit.jsx
```diff
@@ -0,0 +1,279 @@
...
+
+    User.resendInvitation(this.state.user).finally(() => {
+      this.setState({ resendingInvitation: false });
+    });
+  };
+
+  onClickRegenerateApiKey = () => {
+    const doRegenerate = () => {
+      User.regenerateApiKey(this.state.user).then((apiKey) => {
+        this.setState(prevState => ({ user: { ...prevState.user, apiKey } }));
```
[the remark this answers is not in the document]
gabrieldutra: `react/no-access-state-in-setstate` :confused:. That's either the `prevState` thing or:
```javascript
const { user } = { ...this.state.user, apiKey };
this.setState({ user: { ...user, apiKey } });
```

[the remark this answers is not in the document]
ranbena: Ok then leave as is 👍

client/app/components/users/UserEdit.jsx
gabrieldutra: I made some changes on the existing `Alert` to make it communicate better:

send-password-reset-email

Some notes:
- I took the opportunity to create an `InputWithCopy` component (I thought at first with `Input` instead of `<code>`, but LMK if you have another opinion).
- The page I found for related settings was [this](https://redash.io/help/open-source/setup#Mail-Configuration). My only concern is that this page will probably not be relevant to most Admins that will see this Alert, so I would not put this as the first option. I'll try to think somewhere to put it, though I think it could be interesting to add it to the existing Settings Screen -> Email Settings Warning.
- BTW, I'll check how it looks with a Modal instead of an Alert

client/app/services/user.js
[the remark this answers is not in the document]
gabrieldutra: It seems that when network is off this `response` is not `null`, but `response.data` is:
network-off

Anyway, I've fixed all the errors I could reproduce and also added `response = {}` to be safe in `$http` error cases.

cypress/integration/percy/page_screenshots.js
ranbena: Can this be done with stubs as well? Would be a more standard way of doing this.

client/app/components/users/UserEdit.jsx
```diff
@@ -0,0 +1,279 @@
...
+    super(props);
+    this.state = {
+      user: this.props.user,
+      changingPassword: false,
+      sendingPasswordEmail: false,
+      resendingInvitation: false,
+    };
+  }
+
+  onSaveUser = (values, successCallback, errorCallback) => {
```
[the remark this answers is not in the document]
ranbena: But also have this convention with the rest. `sendPasswordReset`, `resendInvitation`, `generateAPIKey`, etc.

client/app/components/users/UserEdit.jsx
ranbena: >  `Input` instead of `<code>`

It's perfect.

> this page will probably not be relevant to most Admins

How so?
@arikfr I think your input could be valuable here.

> BTW, I'll check how it looks with a Modal instead of an Alert

I think it's good enough the way it is. Let's move on :)

arikfr: I agree that the last version you show in the gif looks perfect 👌 

> * My only concern is that this page will probably not be relevant to most Admins that will see this Alert,

By the assumption that not all admins manage the server itself?

arikfr: Btw, <InputWithCopy> is useful for the API key too :)

ranbena: Yeah, perhaps copy button where the regenerate button is now, and move regenerate button  underneath.

(Also, add title "Password" above the two password buttons)

gabrieldutra: > By the assumption that not all admins manage the server itself?

Yes, not all admins will have access to environment stuff.

> Yeah, perhaps copy button where the regenerate button is now, and move regenerate button underneath.

I had that even written on the description before, I guess I was just waiting for another use for a `InputWithCopy` component to do it. Let's try this thing :)

cypress/integration/percy/page_screenshots.js
gabrieldutra: > Can this be done with stubs as well? Would be a more standard way of doing this.

Stubs were my first try, anything similar to Rspec's `allow(obj).to receive(:message).and_return(:value)`. However, I could not find a way to replace it without stubbing the whole request to the user. Since it would be more complex, I went with this Javascript change.

client/app/components/users/UserEdit.jsx
gabrieldutra: **Overall looking**
overall-apikey

**Pending Invitation view**
pending-invitation

I was in doubt if I would keep Resend Invitation here, but then I was thinking: does it make sense to have a Change Password for Pending Invitation users?

ranbena: The way I see it:
Change password - user.id === currentUser.id
Resend invitation - currentUser.isAdmin && user.isInvitationPending
Reset password - currentUser.isAdmin && user.id !== currentUser.id

client/app/components/users/UserEdit.jsx
```diff
@@ -0,0 +1,287 @@
...
+          className="profile__image"
+          width="40"
+        />
+        <h3 className="profile__h3">{user.name}</h3>
+        <hr />
+        {this.renderBasicInfoForm()}
+        {!user.isDisabled && (
+          <Fragment>
+            {this.renderApiKey()}
+            <h3>Password</h3><hr />
```
ranbena: Whoa too big. I thought similar to:
https://github.com/getredash/redash/blob/389d9ee3a69340e85692f160fb78cf05e48fe969/client/app/components/users/UserEdit.jsx#L170-L171

client/app/components/users/UserEdit.jsx
```diff
@@ -0,0 +1,287 @@
...
+    return (
+      <Fragment>
+        <Modal
+          visible={this.state.passwordModalIsOpen}
+          title="Change Password"
+          onCancel={() => { this.setState({ passwordModalIsOpen: false }); }}
+          footer={null}
+          destroyOnClose
+        >
+          <DynamicForm fields={fields} saveText="Update Password" onSubmit={this.updatePassword} />
```
ranbena: @gabrieldutra can you make the form submit button be the modal ok button?

ranbena: (reason: the button looks inconsistent with the rest of the modals)

gabrieldutra: Makes sense, I could by using `wrappedComponentRef` with `onOk` pointing to `handleSubmit` (sth similar to [this](https://ant.design/components/form/#components-form-demo-form-in-modal)). Also, we will need a workaround to trigger the enter key submit.

ranbena: I wonder if our [latest dialog opening convention](https://github.com/getredash/redash/commit/23f5dde4887eb6a9baa58214e3ac2c0d787b2e98) could be utilized here. This is how it operates currently:
```jsx
import ChangePassDialog from './ChangePassDialog';

openModal() {
  const { result, close, dismiss } = ChangePassDialog.showModal(optionalData);

  // Promise with onOK, onCancel callbacks
  result.then((result) => {
      // modal confirmed with optional result
    })
    .catch(() => {
      // modal dismissed
    });

  close(result); // force dialog confirm (with optional result)
  dismiss(); // force dialog dismiss
}

<button onClick={openModal}></button>
```

gabrieldutra: Interesting... I missed that PR :sweat_smile:. Anyway, I'll try to apply this convention.

gabrieldutra: @ranbena I created a `ChangePasswordDialog` and moved the related logic.

I didn't use `DynamicForm` to be able to make it work with Ant modal buttons and have the "Password don't match" in the field. Unfortunately it's now way more complex, so if you have any ideas to simplify it, LMK :slightly_smiling_face:

change-password

ranbena: Wise choice. Looks perfect.

ranbena:
@gabrieldutra if ready for merge - hit the S&M button.

gabrieldutra:
:rocket:

arikfr:
👏

PR #3356, opened 2019-01-28
Move BigQueryGCE to its own file

arikfr:


arikfr:
> And may I ask the reason why BigQuery GCE is not enabled by default? Will it be removed soon?

Its `enabled` check was sometimes causing issues when trying to start Redash and because it's not commonly used, it was simpler to remove from the default list.

There are no plans to remove it at the moment.

Issue #3355, opened 2019-01-27
[Widget Params] Bug: Date value empty in static param input


ranbena:
### Issue Summary

Static date value loads empty when editing parameter.

### Steps to Reproduce

1. Add new widget to dashboard, with date range as static value.
2. Click "edit parameters" on widget

**Expected**: initial value should be preselected in static value input.
**Actual**: input remains empty

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/486954/51801864-e61c0b80-224b-11e9-9a93-c461e3b771cb.gif)

The reason for this is hinted at in the open console

arikfr:
@ranbena does it happen to any static value or only date ranges?

ranbena:
@arikfr yeah it manifests in every value type.
In some scenarios even a `[object Object]` value is prefilled. A date (non-range) generates a js error.

@kravets-levko wdyt?

arikfr:
So strange, I'm pretty sure this was working. One more thing to try: does it reproduce if you pick a new value (different than the default one)?

ranbena:
> does it reproduce if you pick a new value?

Yep, pick a new value, save, click edit params, it's gone.

ranbena:
Fixed by #3357

PR #3357, opened 2019-01-28
Widget params: Date/Date range value empty in static param input

kravets-levko:
Issue getredash/redash#3355


arikfr:
@ranbena merged. Let me know when I can start poking around your PR.

PR #3358, opened 2019-01-28
Cypress tests: preset the admin API key to a static value

arikfr:
(this to avoid diffs in screenshots)

@gabrieldutra FYI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ ARG skip_ds_deps
 
 # We first copy only the requirements file, to avoid rebuilding on every file
 # change.
-COPY requirements.txt requirements_dev.txt ./
+COPY requirements.txt requirements_dev.txt requirements_all_ds.txt ./
 RUN pip install -r requirements.txt -r requirements_dev.txt
 RUN if [ "x$skip_ds_deps" = "x" ] ; then pip install -r requirements_all_ds.txt ; else echo "Skipping pip install -r requirements_all_ds.txt" ; fi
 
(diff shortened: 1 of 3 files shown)


cypress/cypress.js
```diff
@@ -1,72 +1,87 @@
...
+    console.log("POST " + request.route + " - " + result);
     if (seedValues.length) {
       seedDatabase(seedValues);
     }
   });
+
+  // Make sure the admin user has the same API key on every execution
+  execSync(
+    "docker-compose -p cypress run postgres psql -h postgres -c \"update users set api_key = 'secret' where email ='admin@redash.io';\""
+  );
```
arikfr: This is the actual test. The rest is Prettier doing its thing.

gabrieldutra: Nice! That was one of the solutions I thought about (the other one would require me to use Cypress Stubs).

My idea is to create a few Cypress tests in #3354, probably one for regenerating the API Key. If needed in there, I'll revisit this.

BTW, I thought the pattern was to use singlequoted :sweat_smile:

arikfr: > BTW, I thought the pattern was to use singlequoted 😅

🤔 you're correct. Something went wrong there. I'll fix.

Dockerfile
```diff
@@ -14,7 +14,7 @@ ARG skip_ds_deps
 
 # We first copy only the requirements file, to avoid rebuilding on every file
 # change.
-COPY requirements.txt requirements_dev.txt ./
+COPY requirements.txt requirements_dev.txt requirements_all_ds.txt ./
```
gabrieldutra: :eyes:

arikfr: https://github.com/getredash/redash/pull/3358/commits/ec9117c6e53b727e841da2d041badbac8b852065

Issue #3326, opened 2019-01-23
Set filename when calling the /results endpoint directly


arikfr:
Probably by using the `Content-Disposition` header.

aidarbek:
Hi! I am new to the project, could you clarify few things? As far as I understood, the [https://github.com/getredash/redash/blob/master/redash/handlers/query_results.py](https://github.com/getredash/redash/blob/master/redash/handlers/query_results.py) file (`QueryResultResource` class) should be changed and `Content-Disposition` header should be added to response containing xlsx, csv and json files. What the name of the file should be? Can it be `query_result_id` or `query_id` variables?

arikfr:
Hi @aidarbek, welcome :)

This is how we generate the filename on the frontend:

https://github.com/getredash/redash/blob/ff42ec2cc6700023e48c85fde387a955700b507a/client/app/components/queries/query-results-link.js#L27-L28

Basically: query name (spaces replaced with `_`) + date of query result execution + file extension.

I guess we can do the same?

arikfr:
Implemented in #3359 by @aidarbek 🙇

PR #3359, opened 2019-01-29
Filename set when /results called directly

aidarbek:
Fix for #3326


redash/handlers/query_results.py
```diff
@@ -276,6 +277,16 @@ def get(self, query_id=None, query_result_id=None, filetype='json'):
             if should_cache:
                 response.headers.add_header('Cache-Control', 'private,max-age=%d' % ONE_YEAR)
 
+            str_date = query_result.retrieved_at.strftime("%Y_%m_%d")
+
+            if query is not None:
+                filename = query.name.replace(" ", "_") + "_" + str_date
+            else:
+                filename = str(query_result.id) + "_" + str_date
+            filename += "." + filetype
```
arikfr: On second thought, I think that on the backend it will be safer to just use the query id (when available) instead of the name to avoid any unicode issues 😔 (for example, at the moment, line 288 will probably throw an error if the query name has unicode characters)

I would also use `.format` here. Something along the lines of:

```python
if query is not None:
    prefix = query.id
else:
    prefix = query_result.id

filename = "{}_{}.{}".format(prefix, str_date, filetype)
```

What do you think?

aidarbek: Oh silly me! I literally had `format` method used right below that code (when using `add_header`). I've made some changes according to it.

arikfr:
Thanks!

PR #3360, opened 2019-01-29
Yandex Metrica: changed Auth from params to headers

denisov-vlad:
Previous auth type via params will be deprecated on February, 13th.


ranbena:
@denisov-vlad have you tested this manually?

denisov-vlad:
@ranbena Thanks. I've tested it both Metrica and Appmetrica connections.

arikfr:
Thanks, @denisov-vlad ! ✌️

PR #3362, opened 2019-01-29
Fix: make ClickHouse password and username truly optional

### LINKED ISSUE #3289: Clickhouse: password is optional but we try to access it anyway
For Clickhouse type data sources, we don't require a password. But the code does require it by trying to directly access the value in the options dictionary, instead of using `get`:

https://github.com/getredash/redash/blob/823e4ccdd6fcfee5d0df0d919d87af3100876549/redash/query_runner/clickhouse.py#L77

aidarbek:
Fix #3289

### DIFF
--- a/redash/query_runner/clickhouse.py
+++ b/redash/query_runner/clickhouse.py
@@ -68,12 +68,12 @@ class ClickHouse(BaseSQLQueryRunner):
 
     def _send_query(self, data, stream=False):
         r = requests.post(
-            self.configuration['url'],
+            self.configuration.get('url', "http://127.0.0.1:8123"),
             data=data.encode("utf-8"),
             stream=stream,
             timeout=self.configuration.get('timeout', 30),
             params={
-                'user': self.configuration['user'],
+                'user': self.configuration.get('user', "default"),
                 'password':  self.configuration.get('password', ""),
                 'database': self.configuration['dbname']
             }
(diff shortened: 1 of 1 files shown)


redash/query_runner/clickhouse.py
```diff
@@ -74,7 +74,7 @@ def _send_query(self, data, stream=False):
             timeout=self.configuration.get('timeout', 30),
             params={
                 'user': self.configuration['user'],
```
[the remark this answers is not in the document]
aidarbek: I faced an error while building PR with CircleCI

> ERROR: Service 'redash' failed to build: The command '/bin/sh -c pip install -r requirements.txt -r requirements_dev.txt' returned a non-zero code: 2

I guess the problems is not in my PR and there is some error in dependencies? Could you please rerun workflow at some point in future?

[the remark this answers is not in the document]
arikfr: It's definitely unrelated. I triggered another execution.

redash/query_runner/clickhouse.py
```diff
@@ -74,7 +74,7 @@ def _send_query(self, data, stream=False):
...
-                'password':  self.configuration['password'],
+                'password':  self.configuration.get('password', ""),
                 'database': self.configuration['dbname']
```
denisov-vlad: Maybe we will use `.get('dbname', 'default')`?

arikfr: Well, at the moment `dbname` is a required field, so it's guaranteed we will have a value for it.

denisov-vlad: I didn't want to leave the same comment on 2 lines. :)

I think that dbname param should be not required if we agree that user and password fields could be not required too.

arikfr: 👍 got it. I'm going to merge this as is. But I don't mind making `dbname` optional. I just don't want to throw more work at @aidarbek :)

PR #3363, opened 2019-01-29
Fix: Escape button in tag edit modal

### LINKED ISSUE #3189: Tags input control has confusing keyboard behavior
image

1. When the tag suggestion list is open, I expect the `Esc` key to close it.
2. Also, when the suggestions are closed, I expect the `Enter` key to submit the changes.

What in fact happens is that `Enter` has no effect and `Esc` cancels the dialog. I assume this is in part due to the hybrid of a React control (tags input) and Angular modal. But once we move to only React in this area, we should revisit the UX.

ranbena:
Fixes #3189 

#### What
Replaced `uibModal` with Ant Modal.
Which allowed to swap select focus method for a simple `autoFocus` prop.

Also did some code cleaning:
* Chaining instead of func nesting. It's just lovelier.
* Perf optimization by running some methods in `constructor()` instead of `render()`.

More deletions than additions 😅 

#### Tests
Manually testes query edit source/data view and queries list.

### DIFF
--- a/client/app/components/tags-control/TagsControl.jsx
+++ b/client/app/components/tags-control/TagsControl.jsx
@@ -1,7 +1,7 @@
 import { map, trim } from 'lodash';
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
-import { $uibModal } from '@/services/ng';
+import TagsEditorModal from './TagsEditorModal';
 
 export default class TagsControl extends React.Component {
   static propTypes = {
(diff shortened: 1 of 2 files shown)


ranbena:
10x @arikfr I forgot to put in the `autoFocus` attribute 😅

Using a popover has one disadvantage - it's pre-rendered which makes props/state management more difficult. Better to attend to modal styling in a separate issue.

ranbena:
@arikfr does it matter to you if modal shows "cancel"/"ok" instead of "close"/"save"? It's easily configurable.

arikfr:
@ranbena let's put it this way: I didn't even notice.

PR #3366, opened 2019-01-30
Modal style adjustment

ranbena:
The default ant modal style has lines and spacing that the team doesn't favor. Adjusted to a more of the previous modal look.

Before | After
------------ | -------------
screen shot 2019-01-30 at 11 51 56 | screen shot 2019-01-30 at 11 26 22

Before | After
------------ | -------------
screen shot 2019-01-30 at 11 52 32 | screen shot 2019-01-30 at 11 46 44

### DIFF
--- a/client/app/assets/less/ant.less
+++ b/client/app/assets/less/ant.less
@@ -178,3 +178,21 @@
     }
   }
 }
+
+// styling for short modals (no lines)
+.@{dialog-prefix-cls}.shortModal {
+  .@{dialog-prefix-cls} {
+    &-header, &-footer {
+      border: none;
+      padding: 16px;
+    }
+    &-body {
+      padding: 10px 16px;
+    }
+    &-close-x {
+      width: 46px;
+      height: 46px;
+      line-height: 46px;
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


arikfr:
Seeing it in comparison like this I notice that while for the tags modal it looks better I'm not so sure about the scheduler. 🤔

ranbena:
@arikfr I gave it minimum height.
Also, for tagging, since the modal is so devoid of content, I added some help text that appears if no tags yet.

screen shot 2019-02-03 at 19 08 45

ranbena:
> Maybe the tags shouldn't be a modal?

As in a popover? Or simply allow editing in place?

arikfr:
Either :)

Simpler solution: use the border less version for smaller modals and the the bordered one for the bigger ones?

-- 
Arik Fraimovich

ranbena:
> use the border less version for smaller modals and the bordered one for the bigger ones?

Let's say a classname is added according to a height threshold.
1) Determined automatically -  CSS doesn't supply hooks on element dimensions, so this would have to be done with JS. I'm not a fan.
2) Determined manually - I'm not a fan either, but I'd go with this.

Wdyt?

ranbena:
@arikfr I went for `className="shortModal"` so currently only tags modal has the borderless styling.
Let's move on with our lives 🥵

PR #3367, opened 2019-01-30
Use textless endpoint for pristine queries

rauchy:
This PR directs pristine (i.e. saved) query executions to the new textless endpoint (/api/queries/:id/results) which accepts a query ID and set of parameters and applies server-side validation on them.

This is still marked as WIP as I am still trying to understand if there other parts of the codebase that are affected by this.


client/app/services/query.js
```diff
@@ -454,26 +454,30 @@ function QueryResource(
...
+      } else if (this.data_source_id) {
+        this.queryResult = QueryResult.get(this.data_source_id, queryText, parameters.getValues(), maxAge, this.id);
+      } else {
+        return new QueryResultError('Please select data source to run this query.');
       }
-    } else if (this.data_source_id) {
-      this.queryResult = QueryResult.get(this.data_source_id, queryText, parameters.getValues(), maxAge, this.id);
     } else {
-      return new QueryResultError('Please select data source to run this query.');
+      this.queryResult = QueryResult.getByQueryId(this.id, parameters.getValues(), maxAge);
```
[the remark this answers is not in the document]
arikfr: Not sure actually, because of the parameters. :-\ But we need to fully review how this works, because it has implications in dashboards. Maybe in other places, but iirc this was added for the benefit of widgets.

This is one of the oldest pieces of code in the UI codebase that survived many rewrites, but overtime mutated into a monster.

client/app/pages/queries/view.js
```diff
@@ -118,7 +118,7 @@ function QueryViewCtrl(
       return;
     }
 
-    getQueryResult(0, $scope.selectedQueryText);
+    getQueryResult(0, $scope.selectedQueryText, $scope.isDirty);
```
arikfr: isDirty might be true even if the query text isn't dirty. FYI.

arikfr: Whenever anything on query object changes (iirc).

PR #3368, opened 2019-01-30
React migration: FavoritesControl and TagsList components

kravets-levko:
1. Migrate `FavoritesControl` component to React and fix usages;
2. Migrate `TagsList` component to React and fix usages.

P.S. `FavoritesControl` has some logic (broadcasting event on root scope) that should be reviewed in future.


client/app/components/FavoritesControl.jsx
```diff
@@ -0,0 +1,89 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import { react2angular } from 'react2angular';
+import { $rootScope } from '@/services/ng';
+
+function toggleItem(event, item, callback) {
+  event.preventDefault();
+  event.stopPropagation();
+
+  if (item) {
```
arikfr: In what case `item` will be falsey?

client/app/components/FavoritesControl.jsx
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+      item.$favorite().then(() => {
+        item.is_favorite = true;
+        $rootScope.$broadcast('reloadFavorites');
+        if (isFunction(callback)) {
+          callback();
+        }
+      });
+    }
+  }
```
arikfr: This can be simplified:

```javascript
const action = item.is_favorite ? item.$unfavorite : item.$favorite;
action().then(() => {
  item.is_favorite = !item.is_favorite;
  $rootScope.$broadcast('reloadFavorites');
  if (isFunction(callback)) { callback() };
});
```

client/app/components/FavoritesControl.jsx
kravets-levko: Thanks, missed that

client/app/components/FavoritesControl.jsx
```diff
@@ -0,0 +1,89 @@
...
+    `,
+    bindings: {
+      item: '=',
+    },
+    controller($scope) {
+      // See comment for FavoritesControl.propTypes.forceUpdate
+      this.forceUpdateTag = 'force' + Date.now();
+      $scope.$on('reloadFavorites', () => {
+        this.forceUpdateTag = 'force' + Date.now();
+      });
```
ranbena: Wow. Without this the button won't update after click?!

kravets-levko: Yep. But it's bug of `react2angular` (it does not detect changes in objects). But in this case I decided to pass an object instead of few attributes because in this case it was very easy to detect when we should update component. When we'll get rid of `react2angular`, we can just remove this code too.

client/app/components/FavoritesControl.jsx
kravets-levko: @arikfr Fixed.

client/app/components/FavoritesControl.jsx
```diff
@@ -0,0 +1,89 @@
...
+          callback();
+        }
+      });
+    }
+  }
+}
+
+export function FavoritesControl({ item, onChange }) {
+  const icon = item.is_favorite ? 'fa-star' : 'fa-star-o';
+  const title = item.is_favorite ? 'Remove from favorites' : 'Add to favorites';
```
ranbena: A trick I like to do to avoid duplicate condition:
```js
const [icon, title] = item.is_favorite
  ? ['Remove from favorites', 'fa-star']
  : ['Add to favorites', 'fa-star-o'];
```

Not really suggesting to swap for this. Just another way of doing the same thing 😬

kravets-levko: IMHO both variants are more or less equal, so let's keep my one :wink:

client/app/components/FavoritesControl.jsx
```diff
@@ -0,0 +1,80 @@
...
+  event.preventDefault();
+  event.stopPropagation();
+
+  const action = item.is_favorite ? item.$unfavorite.bind(item) : item.$favorite.bind(item);
+  const savedIsFavorite = item.is_favorite;
+
+  action().then(() => {
+    item.is_favorite = !savedIsFavorite;
+    $rootScope.$broadcast('reloadFavorites');
+    if (isFunction(callback)) {
```
ranbena: Since there's a default value to `onChange` isn't this check redundant?

client/app/components/FavoritesControl.jsx
arikfr: Considering how we come back to this everytime, would it be possible to just fix this issue in `react2angular`?

client/app/components/FavoritesControl.jsx
kravets-levko: @ranbena Fixed

client/app/components/FavoritesControl.jsx
```diff
@@ -0,0 +1,77 @@
...
+  );
+}
+
+FavoritesControl.propTypes = {
+  item: PropTypes.shape({
+    is_favorite: PropTypes.bool.isRequired,
+  }).isRequired,
+  onChange: PropTypes.func,
+  // Force component update when `item` changes.
+  // Remove this when `react2angular` will finally go to hell
```
ranbena: axe

client/app/components/TagsList.jsx
```diff
@@ -0,0 +1,83 @@
...
+      // A set of tag names
+      selectedTags: new Set(),
+    };
+  }
+
+  componentDidMount() {
+    getTags(this.props.tagsUrl).then((allTags) => {
+      this.setState({ allTags });
+    });
+  }
```
ranbena: An opportunity for _this_  neat syntax 😋
```js
async componentDidMount() {
  this.setState = await getTags(this.props.tagsUrl);
}
```

kravets-levko: Try to look at the code generated for this by `babel` - probably you'll be surprised

client/app/components/FavoritesControl.jsx
arikfr: Probably all of us. Burn it all!

client/app/components/TagsList.jsx
arikfr: YES! I wanted to suggest this in another place in this PR, but was too lazy :)

client/app/components/TagsList.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+  render() {
+    const { allTags, selectedTags } = this.state;
+    if (allTags.length > 0) {
+      return (
+        <div className="list-group m-t-10 tags-list tiled">
+          {map(allTags, tag => (
+            <a
+              key={tag.name}
+              className={'list-group-item max-character ' + (selectedTags.has(tag.name) ? 'active' : '')}
```
ranbena: Btw, there is a popular package for classname manipulation https://www.npmjs.com/package/classnames

So it's more readable
```jsx
const className = classNames(
  'list-group-item',
  'max-character',
  {active: selectedTags.has(tag.name)}
);
return <a className={className}>...</a>
```

kravets-levko:
@ranbena @arikfr While I'm not 100% about using `async/await`, but `classnames` library is definitely something we need, so I decided to add it.

client/app/components/TagsList.jsx
```diff
@@ -0,0 +1,84 @@
...
+    };
+  }
+
+  componentDidMount() {
+    getTags(this.props.tagsUrl).then((allTags) => {
+      this.setState({ allTags });
+    });
+  }
+
+  toggleTag(event, tag) {
```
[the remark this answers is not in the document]
ranbena: Didn't we tell him to shut up already?!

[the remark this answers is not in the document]
arikfr: Actually after tuning the settings, I enabled it to comment on PRs. But old PRs come with old CodeClimate settings... :-(

PR #3369, opened 2019-01-30
Fix: Static param value not editable for Text/Number

ranbena:
Static parameters text and number cannot be edited.

STR:
1. Create a query with a textual static paramater
2. Create a widget from the this query
3. Choose static value as parameter type.
4. Type any value into bottom value text field.

Expected: field is editable.
Actual: you type but value stays the same.

#### Why?
#3357 changed the param value from `value` to `param.normalizedValue` but the `<input>` updates the `value` prop so it doesn't update it's content that still shows `param.normalizedValue`.

#### Why is this affecting only text and number types?
Because other param type elements implement an Ant component which gets fed the `defaultValue`, not `value`.

#### What's the fix?
I don't wanna risk touching the variable that's being fed or updated, instead I swapped the `<input>` for an Ant `<Input`> which takes a `defaultValue`. Had to split text and number types since they require different Ant components.

### DIFF
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -1,8 +1,9 @@
-import { isNull, isUndefined } from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
 import Select from 'antd/lib/select';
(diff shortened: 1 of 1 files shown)


arikfr:
🐞

PR #3370, opened 2019-01-30
Tune CodeClimate's config to make it less annoying

arikfr:
We should revisit this when the Angular code is gone.


arikfr:
Instant success:

image

😈

ranbena:
@arikfr pr-triage and waffle are confused and misleading.
After I clicked "update branch", wafflebot assigned me (wat!), and then pr-triage switched to "unreviewed" (wrong!).

PR #3371, opened 2019-01-30
Fix: static param text value is `[object Object]`

ranbena:
@kravets-levko just saved me... ⛑

Wrongly implemented `<Input onChange={onChange} />` expecting value argument instead of `event`..

### DIFF
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -145,7 +145,7 @@ export class ParameterValueInput extends React.Component {
       <Input
         className={'form-control ' + className}
         defaultValue={value || ''}
-        onChange={onSelect}
+        onChange={event => onSelect(event.target.value)}
       />
     );
   }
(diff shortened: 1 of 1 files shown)


client/app/components/ParameterValueInput.jsx
```diff
@@ -170,12 +170,12 @@ export class ParameterValueInput extends React.Component {
...
-        type="$ctrl.param.type"       
-        value="$ctrl.param.normalizedValue"       
-        enum-options="$ctrl.param.enumOptions"       
+      <parameter-value-input-impl
+        type="$ctrl.param.type"
+        value="$ctrl.param.normalizedValue"
+        enum-options="$ctrl.param.enumOptions"
         query-id="$ctrl.param.queryId"
-        on-select="$ctrl.setValue"       
+        on-select="$ctrl.setValue"
```
ranbena: I let this sneak in...

PR #3377, opened 2019-01-31
[WIP] [Widget Params] Restyled source editing (3 of 3)

ranbena:
Code work is done, but labeled WIP to allow feedback first on the user experience.
Play with it in https://deploy-preview-3377--redash-preview.netlify.com.

### What changed and why?
* Source type dropdown changed to radio buttons. I think it's extremely valuable for clarity to have all options laid out instead of hidden inside the dropdown.
* When no existing params, instead of hiding the option, it is disabled and has info available.
* "OK" button disables on form error.
* Source type labels are more coherent. @susodapop lmk if you have any suggestions here.
* If current keyword not in existing params, don't show it as an option. This was possible before but makes no sense as it essentially means adding a new dashboard param.

**Before**:
screen shot 2019-01-31 at 12 25 14

**After**:
screen shot 2019-01-31 at 12 24 01

### DIFF
--- a/client/app/assets/less/ant.less
+++ b/client/app/assets/less/ant.less
@@ -17,6 +17,7 @@
 @import '~antd/lib/popover/style/index';
 @import '~antd/lib/icon/style/index';
 @import '~antd/lib/tag/style/index';
+@import '~antd/lib/grid/style/index';
 @import 'inc/ant-variables';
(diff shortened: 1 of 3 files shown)


ranbena:
Moved to #3332

Issue #3341, opened 2019-01-25
Query schedule fails if there "Until" field equals None


denisov-vlad:
### Issue Summary

Query schedule doesn't work if there "until" field equals None:

```
[2019-01-25 10:36:45,288][PID:12495][ERROR][ForkPoolWorker-3448] Task redash.tasks.refresh_queries[dc3a6d03-fc3e-4e14-8fb9-d5de71ac2b59] raised unexpected: TypeError("'NoneType' object has no attribute '__getitem__'",)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 382, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/opt/redash/redash.origin.latest/redash/worker.py", line 75, in __call__
    return TaskBase.__call__(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 641, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/redash/redash.origin.latest/redash/tasks/queries.py", line 286, in refresh_queries
    for query in models.Query.outdated_queries():
  File "/opt/redash/redash.origin.latest/redash/models/__init__.py", line 564, in outdated_queries
    query.schedule['until'], '%Y-%m-%d')) if query.schedule['until'] else None
TypeError: 'NoneType' object has no attribute '__getitem__'
```

Schedule from queries table: `{"interval": 300, "until": null, "day_of_week": null, "time": null}`

It can't parse null value. 

Is it ok to remove "until" field if it equals null ?

### Technical details:

* Redash Version: master
* Browser/OS: –
* How did you install Redash: –

ranbena:
@arikfr How bow compacting before saving? Would be more space efficient as well?

Before:
```js
save() {
  ...
  if (newSchedule.interval) {
    this.props.updateQuery({ schedule: clone(newSchedule) });
  } else {
    this.props.updateQuery({ schedule: null });
  }
  ...
}
```
After:
```js
import { chain, isNil } from 'lodash';

save() {
  ...
  const schedule = chain(newSchedule).clone().omitBy(isNil);
  this.props.updateQuery({ schedule });
  ...
}
```

arikfr:
@ranbena it sounds like a good idea to compact the values. But will the new version you suggested result in `schedule` = `null` when it's empty?

ranbena:
[...]

PR #3378, opened 2019-01-31
Compact refresh schedule data

ranbena:
Removing any null values params from schedule object for efficiency.
Tested and works fine.


ranbena:
@arikfr are you talking about a change to frontend or backend?

ranbena:
#3341 has already been resolved

PR #2888, opened 2018-10-02
Add archived queries section to queries list.

jezdez:
This was originally filed in https://github.com/mozilla/redash/issues/150 where [@arikfr agreed](https://github.com/mozilla/redash/issues/150#issuecomment-398843209) to add this feature eventually.

screen-shot-2018-10-02-20-37-45 21

PR #3381, opened 2019-01-31
Migrate Dashboards/Queries/Users list pages to React

kravets-levko:
- [x] Create new base component based on `/lib/list-ctrl.js` (which is a common parent for mentioned pages);
- [x] Migrate Dashboards list page;
- [x] Migrate Queries list page;
- [x] Migrate Users list page;
  - [x] Error handling
  - [x] `<users-list-extra>`
- [x] Cleanup.

## Screenshots:

image

image

image

image

image

image

image


arikfr:
Have you considered using Ant's [<List>](https://ant.design/components/list/) component instead of a table for these pages?

jezdez:
@kravets-levko Could we please merge #2888 first since there is some overlap with the routes refactor there?

client/app/pages/queries-list/QueriesList.jsx
```diff
@@ -0,0 +1,247 @@
...
+import { $location, $rootScope } from '@/services/ng';
+import { Query } from '@/services/query';
+import { currentUser } from '@/services/auth';
+import { durationHumanize } from '@/filters';
+import { formatDateTime } from '@/filters/datetime';
+import './queries-list.css';
+
+class QueriesList extends ItemsList {
+  static sidebarMenu = [
+    {
```
jezdez: I think we should be able to use this list to hook up the routing as well, similar how I did this in #2888?

ranbena:
> Have you considered using Ant's component

What did you have in mind @arikfr ?

client/app/lib/pagination/live-paginator.js
```diff
@@ -6,7 +8,7 @@ export default class LivePaginator {
     this.itemsPerPage = itemsPerPage;
     this.totalCount = 0;
     this.orderByField = orderByField;
-    this.orderByReverse = orderByReverse;
+    this.orderByReverse = !!orderByReverse;
```
ranbena: Why is this needed?

kravets-levko:
@arikfr @jezdez 
> Yes, I'll merge yours first. (Sorry, Levko...)

It's totally fine, I didn't expect anything else :-)

@ranbena 
> the favorite icon in first table column is misaligned

I'll do the final cleanup on the last step, will fix this too :+1:

client/app/lib/pagination/live-paginator.js
kravets-levko: I'll check, thanks

client/app/pages/queries-list/QueriesList.jsx
kravets-levko: Yes, I'll revisit it after #2888 will be merged :+1:

client/app/pages/ItemsList.jsx
```diff
@@ -0,0 +1,320 @@
...
+import { clientConfig } from '@/services/auth';
+
+export default class ItemsList extends React.Component {
+  static sidebarMenu = [];
+
+  static listColumns = [];
+
+  static orderSeparator = '-';
+
+  static defaultOrder = '-created_at';
```
ranbena: FYI, there is a popular convention of defining such properties as "configuration" vars outside the class.

```js
import { $location, $route } from '@/services/ng';

const ORDER_SEPARATOR = '_';
const DEFAULT_ORDER = '-created_at';

export default class ItemsList extends React.Component {
  ...
}
```

It also allows to easily export certain vars from other locations that may need it.
```js
import { DEFAULT_ORDER } from '@/pages/ItemsList';
```

client/app/pages/ItemsList.jsx
```diff
@@ -0,0 +1,320 @@
...
+      isLoaded: false,
+      isEmpty: false,
+
+      searchTerm,
+      selectedTags: [],
+
+      pageSizeOptions: clientConfig.pageSizeOptions,
+
+      paginator: new LivePaginator(() => {}, {
+        page: parseInt(urlQueryParams.page, 10) || 1,
```
ranbena: `Number(str)` instead of `parseInt(str, 10)` is nicer :)

client/app/pages/ItemsList.jsx
```diff
@@ -0,0 +1,320 @@
...
+      <div className="text-center">
+        <BigMessage icon="fa-search" message="Sorry, we couldn't find anything." />
+      </div>
+    );
+  }
+
+  renderSearchInput(placeholder = null) {
+    return (
+      <div className="m-b-10">
+        <Input
```
ranbena: If you use `<Input.Search` then a search icon is added to the side. Could be nice.

client/app/pages/ItemsList.jsx
```diff
@@ -0,0 +1,320 @@
...
+          onChange={event => this.update({ searchTerm: event.target.value })}
+          autoFocus
+        />
+      </div>
+    );
+  }
+
+  renderSidebarMenu() {
+    const sidebarMenu = this.getSidebarMenu();
+    if (sidebarMenu.length > 0) {
```
ranbena: Can you explain this to me?
Where in the code is `sidebarMenu` being manipulated?

client/app/pages/ItemsList.jsx
```diff
@@ -0,0 +1,320 @@
...
+      return (
+        <div className="list-group m-b-10 tags-list tiled">
+          {map(sidebarMenu, item => (
+            <a
+              key={item.key}
+              href={item.href}
+              className={'list-group-item ' + (this.props.currentPage === item.key ? 'active' : '')}
```
ranbena: Better use `classnames` here?

client/app/pages/ItemsList.jsx
ranbena: Oh this is an abstract component? Damn we need interfaces!

ranbena:
@kravets-levko is it possible to avoid re-rendering the table when changing ordering? I thought Ant Table could do the whole thing on its own without intervention, no?

client/app/pages/ItemsList.jsx
kravets-levko: Yep, I tried it. But this input is already small, and icon will just use an extra space.

kravets-levko:
@ranbena 
> is it possible to avoid re-rendering the table when changing ordering? I thought Ant Table could do the whole thing on its own without intervention, no?

No, because sorting (and pagination) is server-side. On each click we fetch new set of items and update the table

client/app/pages/ItemsList.jsx
kravets-levko: I planned it for overriding in descendants (similar as `getListColumns` which is used in `UsersList` components). But this one will be probably removed, not sure ATM.

ranbena:
> No, because sorting (and pagination) is server-side

So you're saying it is possible if we load all the items into the table?

kravets-levko:
> So you're saying it is possible if we load all the items into the table?

Hm, of course :slightly_smiling_face: But IIRC (@arikfr - am I?) server-side pagination was introduced intentionally - for performance reasons.

ranbena:
> Hm, of course 🙂 But IIRC (@arikfr - am I?) server-side pagination was introduced intentionally - for performance reasons.

Cool cool. Didn't mean to suggest it, just wanted to understand.

client/app/pages/users/UsersList.jsx
```diff
@@ -0,0 +1,248 @@
+import { extend, map, filter } from 'lodash';
+import moment from 'moment';
+import React from 'react';
+import { react2angular } from 'react2angular';
+import classNames from 'classnames';
+import TimeAgo from 'react-timeago';
```
ranbena: Maybe `moment`'s Duration can be of use here https://momentjs.com/docs/#/durations/humanize/

kravets-levko:
@ranbena and @arikfr I decided to change users list page a bit. Please feel free to give your opinion about this.

client/app/pages/users/UsersList.jsx
kravets-levko: This component (as well as `am-time-ago` which we use in Angular code) is interactive. Do you think it's better to implement own component instead of `react-timeago`?

client/app/pages/ItemsList.jsx
kravets-levko: > Oh this is an abstract component? Damn we need interfaces!

No, it's just basic implementation of common functionality - a way to write reusable code. Descendants will just override some methods instead of implementing everything from scratch.

client/app/pages/users/UsersList.jsx
ranbena: Well in this case timeago only renders text so it would be more efficient to use the already installed moment. You could make a util out of it (the moment duration code is a bit long). Whatever you choose.

kravets-levko: I'll try to implement own component, and if it will become too massive - I'll leave `react-timeago`, otherwise I'll uninstall it.

client/app/pages/ItemsList.jsx
```diff
@@ -0,0 +1,325 @@
...
+      return (
+        <div className="m-b-10">
+          <TagsList tagsUrl={tagsUrl} onUpdate={selectedTags => this.update({ selectedTags })} />
+        </div>
+      );
+    }
+    return null;
+  }
+
+  renderPageSizeSelect() {
```
arikfr: How about we remove this and use Pagination's built in support for showing page size? Just need to add the following options:

```
showSizeChanger onShowSizeChange={} pageSize={} pageSizeOptions={}
```

(note that their `pageSizeOptions` takes a list of strings...)

client/app/pages/users/UsersList.jsx
kravets-levko: @ranbena getredash/redash#3385

client/app/pages/ItemsList.jsx
kravets-levko: @arikfr Do you mean this one?
image
The only issue is that currently we hide entire component when there is the only page. Now we will need to show it, otherwise page size selector will be hidden too (example: 10 items total, page size was 5 = 2 pages; user changed it to 20 = 1 page - if pagination is hidden - no way to change page size).

arikfr: #3386.

arikfr:
The thing that bothers me the most in this PR is that we use inheritance instead of composition. It feels "unnatural" in React. Apparently there is a whole page in React's docs on [why composition is better than inheritance](https://reactjs.org/docs/composition-vs-inheritance.html). 📖 

As a practical example: think of the Alerts list page. Obviously it should be able to reuse the code here. But it has only an items list with no sidebar, search, etc. While we could force a sidebar on it or start changing `ItemsList` to support such cases, it's an indication that might be smaller components should be used.

I think it should be something like:

0. `<ListPage>` - here we will put the logic of loading the data. It will render only basic things like the title, but the rest will be rendered via children prop. The actual pages will render `<ListPage>` and not inherit from it.
1. `<ItemsList>` - renders only the table with the objects.
2. `<TagsList>`
3. `<SidebarMenu>`

`<ListPage>` will hold the actual state and pass it along to the other components. This way each page has more control over what it shows. So for dashboard and queries, we will use all 3. For alerts we will start with `ItemsList` only and in the future add SidebarMenu.

Another thing I noticed is the repeated code in [column definitions](https://github.com/getredash/redash/pull/3381/files#diff-fe21b7c5f50a7d21325b8cf0bfdd9f29R31). I suggest we extract the ones that are shared (like favorite control, name, etc) into their module and import them where needed.

ranbena:
I favor Arik's approach.

kravets-levko:
@arikfr @ranbena I re-implemented this using composition instead of inheritance. I still need to do some cleanup and check everything, but you can start to review new changes.

client/app/components/items-list/components/EmptyState.jsx
```diff
@@ -2,10 +2,10 @@ import React from 'react';
 import { BigMessage } from '@/components/BigMessage';
 
 // Default "list empty" message for list pages
-export default function LoadingState() {
+export default function LoadingState(props) {
```
ranbena: Prolly rename to `EmptyListState`?

kravets-levko: It's copy-paste :man_facepalming: Thanks!

client/app/components/DynamicComponent.jsx
```diff
@@ -0,0 +1,53 @@
...
+  registerComponent(alias, null);
+}
+
+export default class DynamicComponent extends React.Component {
+  static propTypes = {
+    is: PropTypes.string.isRequired,
+    children: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.arrayOf(PropTypes.node),
+    ]),
```
ranbena: I believe `PropTypes.node` already accepts single or array.

client/app/pages/users/UsersList.jsx
```diff
@@ -0,0 +1,251 @@
...
+      // Admin
+      <div className="m-b-10">
+        <a
+          href="users/new"
+          className={classNames('btn', 'btn-default', 'm-b-10', { disabled: !policy.isCreateUserEnabled() })}
+        >
+          <i className="fa fa-plus m-r-5" />
+          New User
+        </a>
+        <DynamicComponent is="UsersListExtra" />
```
ranbena: I tried really hard, but I have no idea what this does..

client/app/pages/dashboards/DashboardListEmptyState.jsx
```diff
@@ -0,0 +1,38 @@
...
+  switch (page) {
+    case 'favorites': return (
+      <BigMessage message="Mark dashboards as Favorite to list them here." icon="fa-star" />
+    );
+    default: return (
+      <EmptyState
+        icon="zmdi zmdi-view-quilt"
+        description="See the big picture"
+        illustration="dashboard"
+        helpLink="https://help.redash.io/category/22-dashboards"
```
ranbena: For future consideration - links to external sites are better kept in a `constants` config file, so that updating them can be made easily and safely.

package.json
```diff
@@ -74,9 +74,9 @@
     "pivottable": "^2.15.0",
     "plotly.js": "1.41.3",
     "prop-types": "^15.6.1",
-    "react": "^16.3.2",
+    "react": "^16.7.0",
```
ranbena: Interesting - what did you need this for?

client/app/pages/queries-list/QueriesList.jsx
```diff
@@ -0,0 +1,205 @@
...
+          isArchived={item.is_archived}
+        />
+      </React.Fragment>
+    ), {
+      title: 'Name',
+      field: 'name',
+      width: null,
+    }),
+    Columns.avatar({ field: 'user', className: 'p-l-0 p-r-0' }, name => `Created by ${name}`),
+    Columns.dateTime.sortable({ title: 'Created At', field: 'created_at' }),
```
ranbena: If we add filters in the future, how would that look like?
Wouldn't it make sense to move it to the override obj?
```js
Column({
  dataType: Column.DATETIME,
  sortable: true,
  filterable: true,
  title: 'Created At',
  field: 'created_at'
}),
```

client/app/pages/queries-list/QueriesList.jsx
```diff
@@ -0,0 +1,205 @@
...
+      all: Query.query.bind(Query),
+      my: Query.myQueries.bind(Query),
+      favorites: Query.favorites.bind(Query),
+      archive: Query.archive.bind(Query),
+    };
+    const resource = resources[this.props.currentPage];
+
+    this.controller = new LiveItemsList({
+      doRequest: LiveItemsList.createFetcher(resource, Query),
+      onChange: ({ state }) => this.setState(state),
```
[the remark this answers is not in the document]
ranbena: Giving control of the state to a single controller (not talking about politics here) seems non-ideal. What I like about hoc is that each "ability" comes with it's own set of props so you can add as much as you like - but with this implementation, an additional ability alongside `LiveItemsList` won't be possible. Wdyt?

client/app/pages/users/UsersList.jsx
kravets-levko: Leave this for @arikfr

package.json
kravets-levko: Ooops... I needed it for react contexts. Will revert

client/app/pages/queries-list/QueriesList.jsx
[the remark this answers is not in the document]
kravets-levko: > One disadvantage about this is that we won't be able to add QueriesList specific state props, cause they'll get overwritten when live items change.

You can always set all data that came from controller as a single object. In this case I have no additional props, so decided to do it this way.

> What I like about hoc is that each "ability" comes with it's own set of props so you can add as much as you like

The common issue of all this pages is that you cannot just have own "controller" for, say, pagination, sorting, search, etc. - all that parts are coupled. Even if I'll implement it as HoC - it will be the single wrapper with all that logic. Even more - I tried to group state props somehow, but the only group I can definitely create - pagination; finally I decided just to put all state props into a plain object. I can consider refactoring `LiveItemsList` to HoC, but only as a single component - otherwise I'll put too much effort probably with no extra value from it.

client/app/pages/users/UsersList.jsx
kravets-levko: It's a replacement for [this](https://github.com/getredash/redash/blob/master/client/app/pages/users/list.html#L7).

package.json
kravets-levko: ...or maybe keep?

client/app/pages/queries-list/QueriesList.jsx
kravets-levko: So, yes - it's a good point. You already can mark column as sortable with adding property `sorter: true`, and you already can use it without calling that helper. But I decided to add it because then it's easy to see which columns are sortable. If you think that it's redundant - I'll remove it and will just use property everywhere.

client/app/pages/queries-list/QueriesList.jsx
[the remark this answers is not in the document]
kravets-levko: @ranbena Implemented `LiveItemsList` as HOC - really looks almost perfect :thinking:

package.json
ranbena: Good question. No real answer 😅

client/app/pages/users/UsersList.jsx
ranbena: I don't understand it but I'll leave it be 👍🏻

client/app/components/DynamicComponent.jsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+export function unregisterComponent(alias) {
+  registerComponent(alias, null);
+}
+
+export default class DynamicComponent extends React.Component {
+  static propTypes = {
+    is: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
kravets-levko: What about `alias` (the same as in `registerComponent`/`unregisterComponent`?

package.json
kravets-levko: I decided to revert because it's not really related to this PR.

client/app/components/DynamicComponent.jsx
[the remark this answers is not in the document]
arikfr: The way I see it this property defines the name of the extension you're providing. Hence the idea of `name`/`key`.

It's not an alias, because it doesn't have another name. `use` is almost there, but it's not like DynamicComponent has any purpose of its own beside being a placeholder.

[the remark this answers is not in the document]
kravets-levko: okay, then `name` (because `key` is reserved by React)

ranbena:
@kravets-levko two things stand out for me:

#### 1. Actions in usersList
The API is really uncomfortable :(
```jsx
const wrapUserAction = action => (
  (event, user) => {
    event.preventDefault();
    event.stopPropagation();
    // User service will handle errors, no need to do it here
    action(user).then(() => {
      this.props.controller.update();
    });
  }
);

this.actions = {
  deleteUser: wrapUserAction(User.deleteUser),
};

<button onClick={event => context.deleteUser(event, user)}>Delete</button>
```
This stuff (updating context, preventing event bubbling) should ideally be transparent to the implementation:
```jsx
<button onClick={() => User.deleteUser()}>Delete</button>

// or if row was a component on it's own
<button onClick={() => this.props.delete()}>Delete</button>
```

(I need to keep thinking about the the second thing, stay tuned)

kravets-levko:
@ranbena It's possible to implement 2nd option if columns definitions will be non-static (i.e. defined in constructor). I think it's okay too (a bit less discoverable) - if it looks fine for you - I'll implement it in few minutes.

ranbena:
#### 2. HOC props THIS IS ONLY FOR FURTHER DISCUSSION, NOT FOR THIS PR
Right now, the implementation acts as a wiring between the data and components which imo is unnecessary and mandates code duplication. It's common to keep the component rendering "dumb" and do the wiring in `component/index.js` using something like [recompose](https://github.com/acdlite/recompose).

Then for instance, an implementer would render sidebar like this:
```js
// UsersList/index.js
export default compose(
  withUsersData(),
  withLiveItems(),
  withSearchBar(),
  withPagination(),
)(UsersList);
```

```jsx
// UsersList/UsersList.jsx
import { Menu } from '@/components/items-list/components/Sidebar';

renderSidebar(isAdminView) {
  if (!isAdminView) {
    return null;
  }
  const { Searchbar, currentPage, PageSizeSelect } = this.props;
  return (
    <React.Fragment>
      <Searchbar />
      <Menu items={this.constructor.sidebarMenu} selected={currentPage} />
      <PageSizeSelect />
    </React.Fragment>
  );
}
```

arikfr:
@ranbena this is probably a bad venue for having this discussion: it's hard to follow and it will be lost. Maybe a forum post with references to the relevant code?

In general, I'm not 100% sold about using HOCs for passing presentation logic. But need specific examples to discuss.

kravets-levko:
@ranbena Sometimes this approach will work, sometimes not. For example, on pages from this PR search and sorting are coupled: when you start searching something, it resets sorting, and when you clear search input - it should return previously selected search column/order. It's much easier to maintain both functions together because you can see this dependency explicitly. As well as almost everything depends on pagination (`LivePaginator` is the one who rules over pagination, sorting and loads data). If we'll split this into smaller parts - they'll need to communicate somehow - it will add a lot of complexity (and I doubt that part will be reusable without others). IMHO the only right decision in this case is to use common sense.

ranbena:
> convention for PR commits

Like what?

kravets-levko:
@ranbena :man_shrugging: Yesterday we decided that we'll mark PRs with new features/bug fixes/etc. and describe what's changed, but didn't discuss how exactly it should look like.

arikfr:
Small bug 🐞 : when selecting "Demo Set" the table width changes and overlaps with the menu:

image

kravets-levko:
@arikfr How could I miss that :man_facepalming: Fixed

ranbena:
@kravets-levko perhaps give each column an individual `min-width` or else they squish.

screen shot 2019-02-05 at 17 10 46

kravets-levko:
@ranbena Thanks! Should be fine now

kravets-levko:
@ranbena No, it's okay (and BTW - on 1980px it's fine):

image

arikfr:
Merged ✌️

PR #3383, opened 2019-01-31
Be more permissive when parameters are safe

rauchy:
This PR allows users who have datasource view-only permissions to execute queries with safe parameters. (A safe query is currently a parameterized query with any type of param except 'text')

Todo: 

- [x] Merge #3367 
- [x] Allow these queries to run in the UI (currently you cannot click the Execute button on such queries)
- [x] Accept `ParameterizedQuery` in `run_query` instead of creating it inside
- [x] add tests 🦄


client/app/pages/queries/view.js
```diff
@@ -157,7 +157,7 @@ function QueryViewCtrl(
   $scope.canEdit = currentUser.canEdit($scope.query) || $scope.query.can_edit;
   $scope.canViewSource = currentUser.hasPermission('view_source');
 
-  $scope.canExecuteQuery = () => currentUser.hasPermission('execute_query') && !$scope.dataSource.view_only;
+  $scope.canExecuteQuery = () => $scope.query.is_safe || (currentUser.hasPermission('execute_query') && !$scope.dataSource.view_only);
```
arikfr: What if the user changes the query and it's no longer safe?

tests/utils/test_parameterized_query.py
```diff
@@ -159,6 +159,18 @@ def test_raises_on_unexpected_param_types(self):
...
+        schema = [{"name": "bar", "type": "text"}]
+        query = ParameterizedQuery("foo", schema)
+
+        self.assertFalse(query.is_safe)
+
+    def test_is_safe_if_not_expecting_text_parameter(self):
+        schema = [{"name": "bar", "type": "number"}]
+        query = ParameterizedQuery("foo", schema)
+
+        self.assertTrue(query.is_safe)
```
arikfr: Worth adding another test case: `is_safe = True` when query has no parameters.

rauchy: fca4f1e

client/app/pages/queries/view.js
rauchy: Then he will [get a 403 back](https://github.com/getredash/redash/pull/3383/files#diff-f734d32676ec526e17e5ae92d41b5787R195) from the server. I'll say this is an edge case, which while not handled in the most elegant way, is a good enough solution.

We might want to consider a better error message there, though.

arikfr: I think we should prevent editing query source when the user doesn't have Full Access permission to the data source. We might actually be doing this already.

Then this becomes a non issue. We can open a separate issue for this.

redash/serializers.py
```diff
@@ -100,6 +101,7 @@ def serialize_query(query, with_stats=False, with_visualizations=False, with_use
         'options': query.options,
         'version': query.version,
         'tags': query.tags or [],
+        'is_safe': ParameterizedQuery(query.query_text, query.options.get("parameters", [])).is_safe,
```
arikfr: > query.options.get("parameters", [])

Not the first time I see this. Maybe we should have a `parameters` property on the `Query` model?

rauchy:
> I think we should prevent editing query source when the user doesn't have Full Access permission to the data source. We might actually be doing this already.

Doesn't seem so. I'll [tackle this separately](https://github.com/getredash/redash/issues/3496) to keep this PR clean.

redash/serializers.py
rauchy: At first I thought 5bdab05, but then I thought 9d5e231. WDYT?

PR #3384, opened 2019-01-31
EmptyState alternative implementation with a Step component

arikfr:
@kravets-levko @ranbena wdyt?

### DIFF
--- a/client/app/components/empty-state/EmptyState.jsx
+++ b/client/app/components/empty-state/EmptyState.jsx
@@ -1,10 +1,9 @@
-import { some, map, filter } from 'lodash';
+import { some } from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
 import classNames from 'classnames';
 import { $uibModal } from '@/services/ng';
-import { currentUser } from '@/services/auth';
 import organizationStatus from '@/services/organizationStatus';
 import './empty-state.less';
 
@@ -17,11 +16,26 @@ function createDashboard() {
   });
 }
 
-function getSteps({
-  showAlertStep,
-  showDashboardStep,
-  showInviteStep,
-}) {
+export function Step({ completed, text, url = null, onClick = null, urlText, show = true }) {
+  if (!show) {
+    return null;
+  }
+
+  return (
+    <li className={classNames({ done: completed })}>
+      {onClick && (
+        <a href="#" onClick={onClick}>
+          {urlText}
+        </a>
+      )}
+      {url && <a href={url}>{urlText}</a>}
+      {text}
+    </li>
+  );
+}
+
+export function EmptyState({ icon, title, description, illustration, helpLink, onboardingMode, ...showSteps }) {
+  const { showAlertStep, showDashboardStep, showInviteStep } = showSteps;
   const {
     data_sources: dataSourceCount,
     queries: queriesCount,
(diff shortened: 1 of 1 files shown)


client/app/components/empty-state/EmptyState.jsx
```diff
@@ -17,78 +17,45 @@ function createDashboard() {
   });
 }
 
-function getSteps({
-  showAlertStep,
-  showDashboardStep,
-  showInviteStep,
-}) {
+function Step({ completed, text, url = null, onClick = null, urlText, show = true }) {
```
gabrieldutra: One question: Does it make a difference setting `url` and `onClick` to `null`, isn't that a default?

I liked it with `Step` component, though I myself am tendentious to use ES6 syntax haha:
```javascript
const Step = ({ completed, text, url, onClick, urlText, show = true }) => (
  show &&
  <li className={classNames({ done: completed })}>
    {onClick && (
      <a href="#" onClick={onClick}>
        {urlText}
      </a>
    )}
    {url && <a href={url}>{urlText}</a>}
    {text}
  </li>
);
```

kravets-levko: > Does it make a difference setting url and onClick to null

By default they will be `undefined`. In this case there's no difference, because `undefined == false` and `null == false`.

ranbena:
@arikfr I need some minimal detail - what's the STR / in which page can I see this in action?

client/app/components/empty-state/EmptyState.jsx
ranbena: This is my pref:

```jsx
function Step({ show, completed, text, url, urlText, onClick }) {
  if (!show) {
    return null;
  }

  return (
    <li className={classNames({ done: completed })}>
      <a href={url || 'javascript:void(0)'} onClick={onClick}>{urlText}</a> {text}
    </li>
  );
}
```
No need for conditional spacing, empty onClick simply doesn't register, clear conditional exit at top of method.

ranbena:
> @arikfr I need some minimal detail - what's the STR / in which page can I see this in action?

No need now, I got it. But some deets in PR means I'm wasting less time orienting myself.

client/app/components/empty-state/EmptyState.jsx
arikfr: @gabrieldutra you're correct. The `null` default is pointeless. That's what I get for writing code at the middle of the night... :)

@ranbena 👍

arikfr:
> No need now, I got it. But some deets in PR means I'm wasting less time orienting myself.

THis one

arikfr:
> No need now, I got it. But some deets in PR means I'm wasting less time orienting myself.

@ranbena in this case there is no visual change. The PR is only about the code itself.

> fixed eslint warnings (yep, Step component also should have own propTypes);

I know 😞 ... I was too lazy to add it yesterday. Thanks for taking care of this.

client/app/components/empty-state/EmptyState.jsx
```diff
@@ -17,71 +17,64 @@ function createDashboard() {
...
 }
 
+Step.propTypes = {
+  show: PropTypes.bool,
+  completed: PropTypes.bool,
+  text: PropTypes.string,
+  url: PropTypes.string,
+  urlText: PropTypes.string,
+  onClick: PropTypes.func,
+};
```
ranbena: You sure none of these should be required?

kravets-levko: Yep. Probably just `text`, but even `text` may be optional (why not?)

arikfr: Because a step without text makes no sense. :) I updated some props to be required (here and in EmptyState itself) in 78df7e7cc9f66eab8de97505371083a29ea0ab92.

client/app/components/empty-state/EmptyState.jsx
```diff
@@ -18,29 +18,30 @@ function createDashboard() {
...
-      {onClick && urlText && <a href="javascript:void(0)" onClick={onClick}>{urlText}</a>}
-      {url && urlText && <a href={url}>{urlText}</a>}
-      {text && urlText ? ' ' : ''}
+      <a href={url || 'javascript:void(0)'} onClick={onClick}>
+        {urlText}
+      </a>{' '}
       {text}
     </li>
-  ) : null;
+  );
```
arikfr: I switched to a version of @ranbena's suggested implementation. I agree that having an early return statement when not showing is better than the ternary operator. I know I mentioned LOC in our discussions, but I still prefer readable code over short code :)

Also, in one of the cases (data source step for non admins) we render an empty <a> tag. I think it's a reasonable price to pay to keep the logic here simple.

client/app/components/empty-state/EmptyState.jsx
```diff
@@ -17,131 +17,164 @@ function createDashboard() {
...
+        <h4>Let&apos;s get started</h4>
+        <ol>
+          {currentUser.isAdmin && (
+            <Step
+              show={isAvailable.dataSource}
+              completed={isCompleted.dataSource}
+              url="data_sources/new"
+              urlText="Connect"
+              text="a Data Source"
+            />
```
kravets-levko: That's an example of what I mentioned earlier: in Angular template we'd put everything in a single line :smile:

client/app/components/empty-state/EmptyState.jsx
```diff
@@ -76,50 +77,90 @@ export function EmptyState({
...
-              See our Help
-              <i className="fa fa-external-link m-l-5" aria-hidden="true" />
-            </a>
-          </p>
-        </div>
-      </div>
-    );
+  if (!shouldShow) {
+    return null;
   }
```
arikfr: Returning early here as well.

PR #3385, opened 2019-02-01
Migrate "time ago" components to React

kravets-levko:
- [x] Implement `TimeAgo` React component;
- [x] Replace `am-time-ago` with `TimeAgo` and remove `angular-moment` dependency :tada:
- [x] Replace `rd-time-ago` with `TimeAgo`.


client/app/config/index.js
```diff
@@ -16,7 +16,6 @@ import ngMessages from 'angular-messages';
 import toastr from 'angular-toastr';
 import ngUpload from 'angular-base64-upload';
 import vsRepeat from 'angular-vs-repeat';
-import 'angular-moment';
```
arikfr: 👋

client/app/components/TimeAgo.jsx
```diff
@@ -0,0 +1,91 @@
+import moment from 'moment';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+
+const interactiveComponents = new Set();
+
+function updateInteractiveComponents() {
+  interactiveComponents.forEach(component => component.forceUpdate());
+  setTimeout(updateInteractiveComponents, 3000);
```
arikfr: The value will update at most every minute. Why check every 3 seconds?

client/app/components/TimeAgo.jsx
```diff
@@ -0,0 +1,91 @@
...
+    date: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
+      PropTypes.instanceOf(Date),
+      // `moment`
+      (props, propName, componentName) => {
+        const value = props[propName];
+        if ((value !== null) && !moment.isMoment(value)) {
+          return new Error('Prop `' + propName + '` supplied to `' + componentName + '` should be a Moment.js instance.');
+        }
```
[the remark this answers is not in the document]
ranbena: Then perhaps make it into an exported proptype? (the are 2 other instances in the project using the same code). Something like:
```js
// proptype.js

export const Moment = (props, propName, componentName) => {
  const value = props[propName];
  if ((value !== null) && !moment.isMoment(value)) {
    return new Error('Prop `' + propName + '` supplied to `' + componentName + '` should be a Moment.js instance.');
}
```
```js
// TimeAgo.jsx

static propTypes = {
  date: PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.number,
    PropTypes.instanceOf(Date),
    Moment,
  ]),
}
```

client/app/components/TimeAgo.jsx
kravets-levko: It's not related to how often value will be updated. It's for updating an amount of time between value and now. For example, when you execute query - it's `executed_at` will be updated at once, but component should periodically show new value: `1 minute ago`, then `2 minutes ago` and so on. `angular-moment` has a progressive refresh interval, depending on time difference (in first seconds it's 1000ms, after a minute it's 1min, after an hour - 60min and so on).

client/app/components/TimeAgo.jsx
```diff
@@ -0,0 +1,91 @@
...
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+
+const interactiveComponents = new Set();
+
+function updateInteractiveComponents() {
+  interactiveComponents.forEach(component => component.forceUpdate());
+  setTimeout(updateInteractiveComponents, 3000);
+}
+updateInteractiveComponents();
```
ranbena: Might this introduce a memory leak since the timeout never expires?
(One way to go about it is to store the timeout and then clear it when unmounting)

kravets-levko: tldr: No :-)

This is a single timer to update all `TimeAgo` instances. `TimeAgo` components add themselves to a queue when they are mounted and remove when unmounted, so no any memory leak. I use this pattern sometimes because it's much easier than managing timers (you have to store timer ID somewhere when you create a timer, you should set that variable to `null` when you destroy a timer because browser may reuse that ID, and so on). In this case, component instances does not need different timers, queue will contain at most ~100 elements (users list page with max. page size), so I decided to implement it in this way. The only thing I may need to change - interval (see @arikfr comment above) - updating every `3000ms` may be too often, but I don't think we need progressive timers for each component instance as in `angular-moment`.

client/app/config/index.js
kravets-levko: I think it's the first Angular dependency we remove

client/app/components/TimeAgo.jsx
kravets-levko: Increased polling interval to 15 seconds.

arikfr: I meant how often we need to update the value on screen... :)

client/app/components/TimeAgo.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+  shouldComponentUpdate(nextProps) {
+    this._value = toMoment(nextProps.date);
+    if (nextProps.interactive) {
+      interactiveComponents.add(this);
+    } else {
+      interactiveComponents.delete(this);
+    }
+    return true; // default behaviour
+  }
```
ranbena: Although this works well, I believe `shouldComponentUpdate`'s purpose is to allow to block rendering depending on a condition. Perhaps `componentDidUpdate` would be more fitting (using `this.props` instead of `nextProps`).

kravets-levko: AFAIK `componentDidUpdate` is called after component receive new props and re-render. Therefore I'd need to call `forceUpdate` from it (to re-render it again with new `this._value`), which will lead to infinite loop because I have no a good condition when to stop calling it.
`componentWillUpdate` is deprecated, but (if I understand it correctly) `componentWillUpdate` was called when `shouldComponentUpdate` returns `true`, so technically it's a proper place for this.

client/app/components/TimeAgo.jsx
kravets-levko: Yes. But is that a problem? When queue is empty it returns almost immediately, and the next tick is in 30 seconds.

ranbena: True, there's no perf hit.
It's just a red flag for me. From my experience, these things tend to come back with a vengeance 😓 But I'm cool with it if you're cool with it.

One last suggestion (but only if it's an easy implementation) - observe the Set and start/stop the timer according to its length?

kravets-levko: While I agree that this implementation is a bit tricky, but in this case I'm totally fine with it (even if timer will stay working while queue is empty).

client/app/components/TimeAgo.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+  componentDidMount() {
+    if (this.props.interactive) {
+      interactiveComponents.add(this);
+    }
+  }
+
+  shouldComponentUpdate(nextProps) {
+    this._value = toMoment(nextProps.date);
```
arikfr: We call `forceUpdate` every 30 seconds. Does this mean we will call `toMoment` every 30 seconds?

arikfr: How about an implementation that uses state for the formatted value, so we don't need to use `forceUpdate`?

Something like -- https://codesandbox.io/s/jnmjq1p2ny.

kravets-levko: > We call forceUpdate every 30 seconds. Does this mean we will call toMoment every 30 seconds?

No, `shouldComponentUpdate` is called only when props changed.

> How about an implementation that uses state for the formatted value, so we don't need to use forceUpdate?

Interesting idea, I'll give it a try and if it looks fine - will update PR :+1:

client/app/components/TimeAgo.jsx
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export class TimeAgo extends React.Component {
+  static propTypes = {
+    date: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
+      PropTypes.instanceOf(Date),
+      Moment,
+    ]),
```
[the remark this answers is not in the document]
kravets-levko: It makes component more universal - so why not? We anyways convert this value to moment (yes, we can skip this, but it will not reduce LoC count).

ranbena:
> @arikfr @ranbena I refined implementation a bit, can you please review it again?

LGTM

client/app/components/TimeAgo.jsx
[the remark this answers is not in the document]
kravets-levko: I'll add a comment; you can pass everything accepted by `moment`

PR #3387, opened 2019-02-01
 [Widget Params] Migrated edit params + new widget dialog to Ant Modal

ranbena:
#### The bug
In #3332 there is a use of popovers, which on "escape" keypress close the whole modal..

#### The fix
Converted the two relevant modals to Ant Modal which won't allow this to happen.
Notice this PR is meant to merge to master, as preparation for #3332.

#### A few visual changes:
* Dialogs not as wide as original. I think this default width looks better.

* When clicking "ok" or "Add to Dashboard", I added a loading indicator till the modal self closes. (Slowed down)

* The previous modal disabled "cancel" buttons while widget was being saved - but I didn't implement that cause it's more code for sth unnecessary. Lmk if you think otherwise.
* Before, closing the modal would make it disappear immediately - now it has a pleasant closing animation.

@arikfr 117 additions / 178 deletions :P

### DIFF
--- a/client/app/components/dashboards/AddWidgetDialog.jsx
+++ b/client/app/components/dashboards/AddWidgetDialog.jsx
@@ -3,6 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
 import Select from 'antd/lib/select';
+import Modal from 'antd/lib/modal';
 import highlight from '@/lib/highlight';
 import {
   MappingType,
@@ -19,15 +20,9 @@ const { Option, OptGroup } = Select;
 
 class AddWidgetDialog extends React.Component {
   static propTypes = {
-    dashboard: PropTypes.object, // eslint-disable-line react/forbid-prop-types
-    close: PropTypes.func,
-    dismiss: PropTypes.func,
-  };
-
-  static defaultProps = {
-    dashboard: null,
-    close: () => {},
-    dismiss: () => {},
+    dashboard: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
+    onClose: PropTypes.func.isRequired,
+    onAdded: PropTypes.func.isRequired,
   };
 
   constructor(props) {
@@ -40,12 +35,14 @@ class AddWidgetDialog extends React.Component {
       searchedQueries: [],
       selectedVis: null,
       parameterMappings: [],
+      showModal: false, // show only after recent queries populated to avoid height "jump"
     };
 
     // Don't show draft (unpublished) queries in recent queries.
     Query.recent().$promise.then((items) => {
       this.setState({
         recentQueries: items.filter(item => !item.is_draft),
+        showModal: true,
       });
     });
 
@@ -143,7 +140,8 @@ class AddWidgetDialog extends React.Component {
       .save()
       .then(() => {
         dashboard.widgets.push(widget);
-        this.props.close();
+        this.props.onAdded();
+        this.close();
       })
       .catch(() => {
         toastr.error('Widget can not be added');
@@ -151,6 +149,10 @@ class AddWidgetDialog extends React.Component {
       .finally(() => {
         this.setState({ saveInProgress: false });
       });
+  }
+
+  close = () => {
+    this.setState({ showModal: false });
   }
 
   updateParamMappings(parameterMappings) {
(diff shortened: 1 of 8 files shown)


client/app/components/dashboards/AddWidgetDialog.jsx
```diff
@@ -143,7 +140,8 @@ class AddWidgetDialog extends React.Component {
       .save()
       .then(() => {
         dashboard.widgets.push(widget);
-        this.props.close();
+        this.props.onAdded();
+        this.close();
```
kravets-levko: Will this also trigger `props.onClose`?

client/app/components/dashboards/EditParameterMappingsDialog.jsx
```diff
@@ -45,7 +39,8 @@ class EditParameterMappingsDialog extends React.Component {
     widget
       .save()
       .then(() => {
-        this.props.close();
+        this.props.onChange();
```
kravets-levko: Maybe rename this prop to something like `onAccept` (and use this name in all modals)? WDYT?

client/app/components/dashboards/AddWidgetDialog.jsx
ranbena: Yep!
https://github.com/getredash/redash/blob/009cde7c347f0a5a1010c7af9ec22723c3487479/client/app/components/dashboards/AddWidgetDialog.jsx#L297

This allows the closing animation to happen, and only then `ng-if` removes the modal from the dom.

client/app/components/dashboards/EditParameterMappingsDialog.jsx
ranbena: 👍🏻onConfirm?

client/app/components/dashboards/widget.js
```diff
@@ -77,18 +77,20 @@ function DashboardWidgetCtrl($location, $uibModal, $window, $rootScope, Events,
...
-    });
+  this.mappingsDialogOpened = false;
+
+  this.openMappingsDialog = () => {
+    this.mappingsDialogOpened = true;
+  };
+
+  this.closeMappingsDialog = () => {
+    this.mappingsDialogOpened = false;
+    $rootScope.$applyAsync();
```
kravets-levko: No need to call `$applyAsync` on root scope - it will trigger too much checks. Inject `$scope` into controller and use `$scope.$applyAsync`

client/app/pages/dashboards/dashboard.html
```diff
@@ -110,8 +110,9 @@ <h2>
       <span class="hidden-xs hidden-sm">Widgets are individual query visualizations or text boxes you can place on your dashboard in various arrangements.</span>
     </h2>
     <div>
-      <a class="btn btn-default" ng-click="$ctrl.addWidget('textbox')">Add Textbox</a>
-      <a class="btn btn-primary m-l-10" ng-click="$ctrl.addWidget('widget')">Add Widget</a>
+      <a class="btn btn-default" ng-click="$ctrl.addTextBox()">Add Textbox</a>
+      <a class="btn btn-primary m-l-10" ng-click="$ctrl.openAddWidgetDialog()">Add Widget</a>
+      <add-widget-dialog ng-if="$ctrl.addWidgetDialogOpened" dashboard="$ctrl.dashboard" on-close="$ctrl.closeAddWidgetDialog" on-added="$ctrl.onWidgetAdded" />
```
kravets-levko: It's weird that similar actions are triggered by handlers with so different names (`addTextBox` vs. `openAddWidgetDialog`), but (as I understand) it's because "Add Textbox" modal is still Bootstrap modal - am I right?

kravets-levko:
@ranbena While code looks good to me, I still have feeling that something's missing there. WDYT if we'll have some API to open these modals, something like:
```javascript
static openModal(props) {
  ReactDOM.render(<AddWidgetDialog {...props} />, document.body);
}

// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

AddWidgetDialog.openModal({ dashboard, onConfirm, onClose })
```

kravets-levko:
(explaining previous comment)

Such API will allow to:
1. remove `react2angular` wrapper for dialogs;
2. remove state variable that manages visibility;
3. make template and code more clear.

client/app/components/dashboards/widget.js
ranbena: Since reverting to uibmodal api, no need for these no more :)

ranbena:
> 1. remove `react2angular` wrapper for dialogs;
> 2. remove state variable that manages visibility;
> 3. make template and code more clear.
> 
> P.S. `openModal` may even return promise - then no need for `onConfirm`/`onClose` props.
> P.P.S. You can try to create HoC for this 😉

All suggestions implemented! The code looks great and functions well.
Notice in the hoc I have implemented only the parts of ui.bootstrap.modal api that I needed for this PR. We can complete it on demand.

client/app/pages/dashboards/dashboard.js
```diff
@@ -318,26 +319,37 @@ function DashboardCtrl(
...
+  this.onWidgetAdded = () => {
+    this.extractGlobalParameters();
+    // Save position of newly added widget (but not entire layout)
+    const widget = _.last(this.dashboard.widgets);
+    if (_.isObject(widget)) {
+      return widget.save();
+    }
+  };
+
+  this.closeAddWidgetDialog = () => {
```
kravets-levko: Seems it's not used - can you please check for unused code and remove it as well?

ranbena: Beat you to it

client/app/hoc/asUIBModal.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+const asUIBModal = (WrappedComponent) => {
```
arikfr: Do you think this won't be used when we're 100% React? Because otherwise calling this `asUIBModal` is probably not the best name.

Also, are you sure this API is better over just `openModal()` function that will take the Modal to open along with whatever props we want it have?

arikfr: Actually, your way we only need to import the Modal...

Anyway, the important thing is the name :)

kravets-levko: @arikfr Indeed name is not super good, but we both with @ranbena  have no better idea ATM. But it be definitely used when we're 100% React.

> Also, are you sure this API is better over just openModal() function

It's more or less equal; this implementation is more OOP-way: `SomeModal.open(...)` vs. `openModal(SomeModal, ...)`

ranbena: > Do you think this won't be used when we're 100% React?

My intention is to remove the hoc once dashboard is reactified, but we'll discuss it then.

> Also, are you sure this API is better over just `openModal()` function that will take the Modal to open along with whatever props we want it have?

It's basically the same in every way. You either import the utility and the modal or the modal with the utility.

kravets-levko: I think we can merge it, but before that let's try to choose better name for this utility. How about `callerFriendlyModal`?

kravets-levko:
:space_invader:

PR #3374, opened 2019-01-31
Fix eslint errors

kravets-levko:
Mostly just removed extra empty lines at EoF.

@ranbena and @gabrieldutra There are also some fixes to components you were recently working on, so please check it. Thanks!

Issue #3207, opened 2018-12-19
RFC: Frontend Extensions


arikfr:
(previous discussion: #2565)

There is an undocumented support for extensions in Redash, both for the backend (#2354, #2740) and frontend (#2799). For both of them we use Python packages Entry Point functionality to discover the extensions. 

This works very well for backend extensions, but for frontend extensions it introduces several challenges:

* The extension can't have its own dependencies.
* The development workflow is a bit involved, because the extension code needs to be copied into the Redash code folder you need Python where you run the bundling process. Also in development, this means this part of the process needs to run in the container, while the webpack build process runs on the host (just explaining it is hard 😓).

We would like to define a new method for loading frontend extensions that will solve the above challenges.

Several directions and questions to address:

* Frontend extensions to be installed and managed using frontend tools (i.e. npm instead of Python entry points).
* How extensions will be used? Need to consider the fact that by default, we run `npm run build` when we build the image. How additional frontend code can be loaded into a container which "inherits" from our default one?
* Developer workflow: ideally you should be able to work on your extension enjoying the same ergonomics (i.e. refresh on changes) you would when developing the main code.

Any comments, ideas and feedback are welcome.

arikfr:
Relevant work: Grafana's plugin model.

jdbranham:
Any work being done on this? 
It would be great to add visualizations as a plugin - but I'm not sure where to start.

I have a current project that Redash can almost support, so I've ended up writing a separate reporting app to support the visualizations and query transformations I need.

I've worked on Grafana plugins and the plugin loading mechanism.  
Also, 'borrowed' some of the code and adapted it for my own platform, after removing all Angular code [using all React (TS) now]
https://github.com/savantly-net/sprout-platform

Essentially the UI get's a list of plugins from the API, then loads the plugins' bundles at runtime.
This solves some of the build problems @arikfr mentions, and allows plugins to have their own dependencies. 
Not sure if this is helpful for Redash, but maybe there's something useful there.

PR #3390, opened 2019-02-02
Added pre-push hook

ranbena:
Pre-push hook that lints and tests

Meta!
screen shot 2019-02-02 at 19 50 20

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3054,6 +3054,23 @@
         "unset-value": "^1.0.0"
       }
     },
+    "caller-callsite": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/caller-callsite/-/caller-callsite-2.0.0.tgz",
+      "integrity": "sha1-hH4PzgoiN1CpoCfFSzNzGtMVQTQ=",
+      "dev": true,
+      "requires": {
+        "callsites": "^2.0.0"
+      },
+      "dependencies": {
+        "callsites": {
+          "version": "2.0.0",
+          "resolved": "https://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
+          "integrity": "sha1-BuuE8A7qQT2oav/vrL/7Ngk7PFA=",
+          "dev": true
+        }
+      }
+    },
     "caller-path": {
       "version": "0.1.0",
       "resolved": "https://registry.npmjs.org/caller-path/-/caller-path-0.1.0.tgz",
@@ -3623,7 +3640,6 @@
       "resolved": "https://registry.npmjs.org/combined-stream/-/combined-stream-1.0.5.tgz",
       "integrity": "sha1-k4NwpXtKUd6ix3wV1cX9+JUWQAk=",
       "dev": true,
-      "optional": true,
       "requires": {
         "delayed-stream": "~1.0.0"
       }
@@ -3914,6 +3930,46 @@
     "cornelius": {
       "version": "git+https://github.com/restorando/cornelius.git#24d935811186c165c8ba63244ff363da71f32dcf",
       "from": "git+https://github.com/restorando/cornelius.git"
+    },
+    "cosmiconfig": {
+      "version": "5.0.7",
+      "resolved": "https://registry.npmjs.org/cosmiconfig/-/cosmiconfig-5.0.7.tgz",
+      "integrity": "sha512-PcLqxTKiDmNT6pSpy4N6KtuPwb53W+2tzNvwOZw0WH9N6O0vLIBq0x8aj8Oj75ere4YcGi48bDFCL+3fRJdlNA==",
+      "dev": true,
+      "requires": {
+        "import-fresh": "^2.0.0",
+        "is-directory": "^0.3.1",
+        "js-yaml": "^3.9.0",
+        "parse-json": "^4.0.0"
+      },
+      "dependencies": {
+        "esprima": {
+          "version": "4.0.1",
+          "resolved": "https://registry.npmjs.org/esprima/-/esprima-4.0.1.tgz",
+          "integrity": "sha512-eGuFFw7Upda+g4p+QHvnW0RyTX/SVeJBDM/gCtMARO0cLuT2HcEKnTPvhjV6aGeqrCB/sbNop0Kszm0jsaWU4A==",
+          "dev": true
+        },
+        "js-yaml": {
+          "version": "3.12.1",
+          "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.12.1.tgz",
+          "integrity": "sha512-um46hB9wNOKlwkHgiuyEVAybXBjwFUV0Z/RaHJblRd9DXltue9FTYvzCr9ErQrK9Adz5MU4gHWVaNUfdmrC8qA==",
+          "dev": true,
+          "requires": {
+            "argparse": "^1.0.7",
+            "esprima": "^4.0.0"
+          }
+        },
+        "parse-json": {
+          "version": "4.0.0",
+          "resolved": "https://registry.npmjs.org/parse-json/-/parse-json-4.0.0.tgz",
+          "integrity": "sha1-vjX1Qlvh9/bHRxhPmKeIy5lHfuA=",
+          "dev": true,
+          "requires": {
+            "error-ex": "^1.3.1",
+            "json-parse-better-errors": "^1.0.1"
+          }
+        }
+      }
     },
     "country-regex": {
       "version": "1.1.0",
@@ -6789,8 +6845,7 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -6808,13 +6863,11 @@
         },
         "balanced-match": {
           "version": "1.0.0",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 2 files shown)


arikfr:
The fact it only lints frontend code triggers my OCD, but... merged it anyway and let's see how it works.

jezdez:
Hi there, can I still vote to not enable this by default please?

ranbena:
Can you elaborate @jezdez?

kravets-levko:
@jezdez While I wasn't sure about running tests (but it doesn't feel wrong as well), but `eslint` is a must-have because very often we receive PRs with unformatted code, and eventually have to do something like [this](getredash/redash#3374) which is not good - much better when everyone takes care of their own code. And sometimes the only way to it is to enforce people to do it  :smile:

jezdez:
@ranbena Basically I find the enforced pre-push checks to be impractical and possibly an anti-pattern for contribution.

It brings in my experience:

1) increased drain on local resources, e.g. battery consumption,
2) slower review cycles while working on feature branches,
3) unwanted linting of unchanged or unrelated code (e.g. backend code),
4) running eslint on the host machine instead of inside the container, which is contrary to our goal of having dev/prod parity (one of the 12 factors)
5) added complexity of the development setup, effectively increasing the barrier of entry for new contributors

Of course all of those points would simply not be deal breakers if it would be optional to run eslint on push (e.g. by leaving it to the developers to enable the pre-push hook). That's what I would suggest to do in this case.

For cases in which eslint isn't run, we simply have to make sure the CI is catching the eslint errors and raising them as part of the pull request checks. The GitHub project could also be configured to require passing individual CI checks before being able to be merged.

@kravets-levko I don't disagree with running eslint, in fact I'm a huge proponent of it and any other tooling that improves code quality. E.g. I very much look forward to the day we can do the same for the backend code with [black](https://black.readthedocs.io/) once we're on Python 3.x.

I agree that responsibility for taking care of code quality is for everyone with the caveat that it can backfire in certain cases as mentioned above.

Lowering the level of complexity to lint code before it's merged is definitely the way to go. Just not at the cost of increasing the effort required for contribution. That's why I think GitHub's pull request checks (and the integration with 3rd party services) are the better place to do linting than on local machines.

kravets-levko:
@jezdez

For 1-4 use `git push --no-validate`. But by default lint/tests should run. By typing long command you tells us that you know what you do, and you bypassed check intentionally.

For 5 - I don't see any complexity. You had to run `npm install` before - you have to do it now. No any new steps.

ranbena:
Hey @jezdez thanks for the detailed reasoning.
The following is my own opinion and experience of coding culture:
> increased drain on local resources, e.g. battery consumption,

Reasonable trade-off. A good reason not to go with pre-commit. The weak resourced can choose to disable the hook. 

> slower review cycles while working on feature branches,

Depends. Pre-push is where this kind of feedback has most value hence getting post-push ci errors can have a negative effect on iterations.

> unwanted linting of unchanged or unrelated code (e.g. backend code)

A non-issue - up to us to configure it correctly and make sure error-riddled code never makes it to master.

> running eslint on the host machine instead of inside the container, which is contrary to our goal of having dev/prod parity (one of the 12 factors)

I'm unfamiliar with this topic. @arikfr wdyt?

> added complexity of the development setup

That's why we went for `husky`. No setup.

jezdez:
> @jezdez
> 
> For 1-4 use `git push --no-validate`. But by default lint/tests should run. By typing long command you tells us that you know what you do, and you bypassed check intentionally.

I disagree that this solves the mentioned problems, it just puts the burden on the side of contributors instead of automating something that can be easily automated and handled as a continuous iteration factor: If there is a linting error, it's similar to other test errors, not similar to other git push errors. It's my distinct experience that git based workflows break down for various reasons and are harder to debug than simply using a CI service to do the linting. I acknowledge that your experience may be different, so providing the option to enable it makes sense to me. 

I'm just asking to turn it off by default.

> For 5 - I don't see any complexity. You had to run `npm install` before - you have to do it now. No any new steps.

You're missing my point, let me elaborate:

The current setup to run `npm install` on the host machine is known to be problematic since it means we don't have parity between the development and the production (read: Docker image) environment, which is an anti-pattern for building maintainable software.

AFAIK running `npm install` on the host machine and not inside the container is solely for historically reasons (correct me if I'm wrong @arikfr, IIRC literally just an issue with how we mount the CWD during development) and prevents us from moving ahead with a few things like frontend extensions.

IIRC eventually everything will need to happen inside the Docker container so we don't have to deal with the differences between the build environment on the host machine and inside the container/image. @arikfr can probably speak more about this since he's had strong opinions about it.

ranbena:
@jezdez important to note that post-push checks would **still run**. The reasoning behind this is to have early-feedback in order to reduce push iterations which would lead to better engagement and less repo "noise".

jezdez:
> Hey @jezdez thanks for the detailed reasoning.
> The following is my own opinion and experience of coding culture:
> 
> > increased drain on local resources, e.g. battery consumption,
> 
> Reasonable trade-off. A good reason not to go with pre-commit. The weak resourced can choose to disable the hook.

How about we err on the side of the "weak resourced" and ask the powerful to run a one-time command instead of asking everyone else to disable it every time?

> > slower review cycles while working on feature branches,
> 
> Depends. Pre-push is where this kind of feedback has most value hence getting post-push ci errors can have a negative effect on iterations.

My take on this is that it's more valuable to have one place to see errors, than two. Having to wait for Circle (or whatever CI) once pushed to GitHub is annoying of course, but that's a orthogonal problem of optimizing the build process.

> > unwanted linting of unchanged or unrelated code (e.g. backend code)
> 
> A non-issue - up to us to configure it correctly and make sure error-riddled code never makes it to master.

I've pushed a bunch of changes that only had backend code changes where running eslint was a waste of my time. I don't understand how running eslint is more important than say flake8 that is run on Circle.

> > running eslint on the host machine instead of inside the container, which is contrary to our goal of having dev/prod parity (one of the 12 factors)
> 
> I'm unfamiliar with this topic. @arikfr wdyt?
> 
> > added complexity of the development setup
> 
> That's why we went for `husky`. No setup.

See the reply to @kravets-levko above.

jezdez:
> @jezdez important to note that post-push checks would **still run**. The reasoning behind pre-push is to have early-feedback in order to reduce push iterations which would lead to better engagement and less repo "noise".

I understand that the CI checks run either way and the theory that pre-push hooks help having "clean" contributions. It just doesn't match my experience with using those hooks.

jezdez:
FTR, as a Redash team member, I wish this would have been given more time for discussion since I now get the feeling the decision has already been made. Not encouraging while we're growing the project.

ranbena:
> ask the powerful to run a one-time command

Come on @jezdez that's unrealistic.

> a waste of my time

Perhaps I'm misunderstanding the harshness of your own experience. Here's mine - currently pre-push takes ~35s on my 6y/o laptop and as a full-time developer on the project, I don't push very often, so it simply doesn't feel like a pain. For you?

> the decision has already been made

Nah. Everything is debatable and there is zero ego involved. We want what's best for users and devs.

jezdez:
> > ask the powerful to run a one-time command
> 
> Come on @jezdez that's unrealistic.

I beg to differ, erring on the side of those who want to contribute, removing barriers and not adding them, helps Redash to stay maintainable. Since there are other alternatives (CI) there is no need to have this feature enabled by default.

> 
> > a waste of my time
> 
> Perhaps I'm misunderstanding the harshness of your own experience. Here's mine - currently pre-push takes ~35s on my 6y/o laptop and as a full-time developer on the project, I don't push very often, so it simply doesn't feel like a pain. For you?

Now you're just ignoring my feedback. Of course my machine can run eslint. But as I elaborated extensively above there is more to it than just the convenience of running a command. Do you need a git hook to run eslint? Can we turn it off by default and document in the contribution documentation how to enable it if needed?

> > the decision has already been made
> 
> Nah. Everything is debatable and there is zero ego involved. We want what's best for users and devs.

I'm happy to hear that!

So how do you feel about reverting the pull request and open a ticket (or forum post?) to find out how we can improve your development workflow that initially led to this pull request?

ranbena:
Looks like you and I don't share the same mentality (apart for the subject of dev/prod parity which I have no input on) and I'd rather halt. I'll let others weigh in if they'd like.

ranbena:
Glanced at how the big boys use hooks and it seems the standard is running lints (mostly eslint/prettier/tslint) [only on staged files](https://www.npmjs.com/package/lint-staged) as a pre-commit hook, which really makes sense for efficiency.

arikfr:
Glad to see this important discussion. I'll read it all and share my thoughts either later today or early next week. Just wanted to comment here so y'all know it didn't go unnoticed.

arikfr:
Not sure about reviving this discussion, but I did promise to follow up :-)

> running eslint on the host machine instead of inside the container, which is contrary to our goal of having dev/prod parity (one of the 12 factors)

While dev/prod parity is important, we should also be pragmatic about it. For the frontend side, using same/compatible version of Node.js is usually enough to have this parity.

> AFAIK running `npm install` on the host machine and not inside the container is solely for historically reasons (correct me if I'm wrong @arikfr, IIRC literally just an issue with how we mount the CWD during development) 

The reason is actually performance. In the past it seemed that running the frontend build process in a Docker container is slower. I think it’s still the case unless you give adequate resources to your Docker VM (in case of MacOS as host).

Considering how good the Node.js support across platforms, it seemed easier to keep running on the host.

> and prevents us from moving ahead with a few things like frontend extensions.

Even if we move frontend tooling into a Docker container, it will probably be in its own container, so the same problem remains… I think it’s inevitable to switch to a different flow for frontend extensions (probably using npm, #3207).  

> Just not at the cost of increasing the effort required for contribution.

At this stage of the project, we shouldn’t be shy about requiring a bit more effort from contributors (within reason, of course).

>  I don't understand how running eslint is more important than say flake8 that is run on Circle.

It’s not and it’s one of the reasons why I didn’t merge this pull request right away. It bothered me that we only lint frontend assets, but decided to go with the pragmatic approach. The problem with linting Python code is that we don’t have certainty about how Python is setup for the developer (are they using system Python? Are they using our Docker setup?).

I think that a nice solution would be to detect if the developer is using our Docker setup, and in this case run linting using the Docker container.

jezdez:
@arikfr I believe running eslint and npm test on every git commit (after merging #3410) is an anti-pattern that is better solved by services and optional tooling during development. I'd appreciate if you could extend the development docs with ways to disable it if possible.

arikfr:
@jezdez based on your reaction to #3410 I was under the impression you're in favor of that solution?

jezdez:
@arikfr I was at first, but after I left my comment the PR was updated to also run the tests on commit.

PR #3398, opened 2019-02-05
Fix: Selected item in dropdown unreadable

ranbena:
Regression from #3347.

####The bug
The "disabled" look leaked into the regular selected item style.

#### The fix
Made it disabled specific.
Tested manually in query page->data source selection (selected item) and Alerts (unpublished query item).


arikfr:
👍

PR #3399, opened 2019-02-05
Refactor TagsControl; fix TagsEditorModal animation

kravets-levko:
1. `TagsControl` refactoring: use composition instead of inheritance, use HOC to implement `QueryTagsControl` and `DashboardTagsControl`;
2. Fix `TagsEditorModal` animation on close.

### DIFF
--- a/client/app/components/tags-control/TagsControl.jsx
+++ b/client/app/components/tags-control/TagsControl.jsx
@@ -1,4 +1,4 @@
-import { isObject, map, filter, trim, extend } from 'lodash';
+import { map, trim, extend } from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
@@ -24,23 +24,12 @@ export class TagsControl extends React.Component {
     children: null,
   };
 
-  static Prepend({ children }) {
-    return <React.Fragment>{children}</React.Fragment>;
-  }
-
-  static Append({ children }) {
-    return <React.Fragment>{children}</React.Fragment>;
-  }
-
   state = {
     showModal: false,
-    availableTags: [],
   };
 
   openEditModal = () => {
-    // load available tags every time modal is shown; show modal only when tags loaded
-    this.props.getAvailableTags()
-      .then(availableTags => this.setState({ showModal: true, availableTags }));
+    this.setState({ showModal: true });
   };
 
   closeEditModal = () => {
@@ -64,7 +53,7 @@ export class TagsControl extends React.Component {
         {this.state.showModal && (
           <TagsEditorModal
             tags={tags}
-            availableTags={this.state.availableTags}
+            getAvailableTags={this.props.getAvailableTags}
             onConfirm={this.onTagsChanged}
             onCancel={this.closeEditModal}
           />
@@ -74,14 +63,12 @@ export class TagsControl extends React.Component {
   }
 
   render() {
-    const children = filter(React.Children.toArray(this.props.children), isObject);
     return (
       <div className={'tags-control ' + this.props.className}>
-        {filter(children, child => child.type === this.constructor.Prepend)}
+        {this.props.children}
         {map(this.props.tags, tag => (
           <span className="label label-tag" key={tag} title={tag}>{tag}</span>
         ))}
-        {filter(children, child => child.type === this.constructor.Append)}
         {this.props.canEdit && this.renderEditButton()}
       </div>
     );
@@ -94,21 +81,19 @@ function modelTagsControl({ archivedTooltip }) {
   function ModelTagsControl({ isDraft, isArchived, ...props }) {
     return (
       <TagsControl {...props}>
-        <TagsControl.Prepend>
-          {!isArchived && isDraft && (
-            <span className="label label-tag-unpublished">Unpublished</span>
-          )}
-          {isArchived && (
-            <Tooltip placement="right" title={archivedTooltip}>
-              <span className="label label-tag-archived">Archived</span>
-            </Tooltip>
-          )}
-        </TagsControl.Prepend>
+        {!isArchived && isDraft && (
+          <span className="label label-tag-unpublished">Unpublished</span>
+        )}
+        {isArchived && (
+          <Tooltip placement="right" title={archivedTooltip}>
+            <span className="label label-tag-archived">Archived</span>
+          </Tooltip>
+        )}
       </TagsControl>
     );
   }
 
-  // `extend` needed just for `react2angular`, so remove it when `react2angular` no longer needed
+  // ANGULAR_REMOVE_ME `extend` needed just for `react2angular`, so remove it when `react2angular` no longer needed
   ModelTagsControl.propTypes = extend({
     isDraft: PropTypes.bool,
     isArchived: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


client/app/components/tags-control/TagsControl.jsx
```diff
@@ -18,87 +21,118 @@ export default class TagsControl extends React.Component {
...
+        <TagsControl.Prepend>
+          {!isArchived && isDraft && (
+            <span className="label label-tag-unpublished">Unpublished</span>
+          )}
+          {isArchived && (
+            <Tooltip placement="right" title={archivedTooltip}>
+              <span className="label label-tag-archived">Archived</span>
+            </Tooltip>
+          )}
+        </TagsControl.Prepend>
```
arikfr: Instead of all the append/prepend logic, why not render the two special tags and then call `<TagsControl>`?

```jsx
render() {
  return (
    <React.Fragment>
        {...draft label...}
        {...archive label...}
        <TagsControl ...props/>
    </React.Fragment>
  );
}
```

Also, how about turning ModelTagsControl into a regular component that accepts an `archiveToolTip` prop and then QueryTagsControl/DashboardsTagsControl become:

```jsx
function QueryTagsControl(props) {
    return <ModalTagsControl archiveTooltip="..." ...props/>
}
```

WDYT?

client/app/components/tags-control/TagsControl.jsx
```diff
@@ -18,87 +21,118 @@ export default class TagsControl extends React.Component {
...
+  ModelTagsControl.defaultProps = extend({
+    isDraft: false,
+    isArchived: false,
+  }, TagsControl.defaultProps);
+
+  return ModelTagsControl;
+}
+
+export const QueryTagsControl = modelTagsControl({
+  archivedTooltip: 'This query is archived and can\'t be used in dashboards, and won\'t appear in search results.',
```
ranbena:
```suggestion
  archivedTooltip: 'This query is archived and can\'t be used in dashboards, or appear in search results.',
```

client/app/components/tags-control/TagsControl.jsx
kravets-levko: > Instead of all the append/prepend logic, why not render the two special tags and then call `<TagsControl>`?

Because it will break nesting level of elements (new elements will be outside of wrapper).

> Also, how about turning ModelTagsControl into a regular component that accepts an `archiveToolTip` prop

Indeed it will work as expected, but there will be no huge difference with current implementation; HOC just hides that intermediate component.

kravets-levko: @ranbena What's you opinion about this?

client/app/components/tags-control/TagsControl.jsx
```diff
@@ -18,87 +21,118 @@ export default class TagsControl extends React.Component {
...
+});
+
+export const DashboardTagsControl = modelTagsControl({
+  archivedTooltip: 'This dashboard is archived and and won\'t appear in the dashboards list or search results.',
```
ranbena:
```suggestion
  archivedTooltip: 'This dashboard is archived and won\'t be listed in dashboards nor search results.',
```

client/app/components/tags-control/TagsEditorModal.jsx
```diff
@@ -4,54 +4,66 @@ import PropTypes from 'prop-types';
...
     return (
       <Modal
-        visible
+        visible={this.state.isVisible}
         title="Add/Edit Tags"
-        onOk={() => close(result)}
-        onCancel={dismiss}
+        onOk={() => this.onConfirm(this.state.result)}
+        onCancel={() => this.onCancel()}
+        afterClose={() => this.state.onAfterClose()}
```
ranbena: Very nice modal refactor

client/app/components/tags-control/TagsControl.jsx
```diff
@@ -18,87 +21,118 @@ export default class TagsControl extends React.Component {
...
+            <Tooltip placement="right" title={archivedTooltip}>
+              <span className="label label-tag-archived">Archived</span>
+            </Tooltip>
+          )}
+        </TagsControl.Prepend>
+      </TagsControl>
+    );
+  }
+
+  // `extend` needed just for `react2angular`, so remove it when `react2angular` no longer needed
```
ranbena: Perhaps prefix comments with easy to find `ANGULAR_REMOVE_ME`.

client/app/components/tags-control/TagsControl.jsx
```diff
@@ -18,87 +21,118 @@ export default class TagsControl extends React.Component {
...
+    showModal: false,
+    availableTags: [],
+  };
+
   openEditModal = () => {
-    this.setState({ showModal: true });
-  }
+    // load available tags every time modal is shown; show modal only when tags loaded
+    this.props.getAvailableTags()
+      .then(availableTags => this.setState({ showModal: true, availableTags }));
```
ranbena: Not sure this is the best approach.
It might mean that there would be a noticeable (even if slight) irresponsiveness to the click event.

Perhaps have the modal open but the `<Select>` disabled and indicating loading until promise resolved.
```js
openEditModal = () => {
  this.setState({ showModal: true, loading: true });
  this.props.getAvailableTags()
      .then(availableTags => this.setState({ loading: false, availableTags }));
};

<Select disabled={this.state.loading} loading={this.state.loading} >
  {this.selectOptions}
</Select>
```

client/app/components/tags-control/TagsControl.jsx
arikfr: > Because it will break nesting level of elements (new elements will be outside of wrapper).

Right. 👍 It seems we don't use `TagsControl.Append` anywhere... Maybe use `children` directly to avoid the extra complexity?

> Indeed it will work as expected, but there will be no huge difference with current implementation; HOC just hides that intermediate component.

The main difference is a more familiar pattern.

client/app/components/tags-control/TagsControl.jsx
kravets-levko: Then I should move this to modal. But not a big deal

client/app/components/tags-control/TagsControl.jsx
ranbena: I favor @arikfr's suggestion cause it's more readable
```jsx
function QueryTagsControl(props) {
  return <ModalTagsControl archiveTooltip="..." ...props/>
}
```

Regarding prepend - is append used at all?
Doesn't look like it.. so why not just go with:
```jsx
<TagsControl ...props/>
  {...draft label...}
  {...archive label...}
</TagsControl>
```
And then:
```jsx
render() {
  return (
    <div className={'tags-control ' + this.props.className}>
      {this.props.children}
      {this.props.tags.map...}
      {this.renderEditButton()}
    </div>
  );
}
```

client/app/components/tags-control/TagsEditorModal.jsx
```diff
@@ -1,4 +1,4 @@
-import { map, trim, chain } from 'lodash';
+import { map, trim, uniq } from 'lodash';
```
ranbena: chains

client/app/components/tags-control/TagsControl.jsx
ranbena: When I wrote this ^^ I didn't realize arik already suggested it...

kravets-levko:
Sorry, cannot add animation for code :grimacing:

ranbena:
> Sorry, cannot add animation for code 😬

Of course you can. Create an ATOM animated theme.

PR #3400, opened 2019-02-05
Show active and pending users separately (for admins)

kravets-levko:
- [x] API changes;
- [x] Users list page:
  - [x] sidebar: show sections for active users, pending (invited but not activated) users and disabled users;
  - [x] for non-admins: show sidebar, but hide disabled users section;
- [x] Tests

## Screenshots

image

image


client/app/pages/users/UsersList.jsx
```diff
@@ -256,6 +274,11 @@ export default function init(ngModule) {
       title: 'Users',
       key: 'active',
     },
+    {
+      path: '/users/pending',
+      title: 'Invited Users',
```
ranbena: Not sure what the standard naming is for this. Perhaps `Pending Invites`?

kravets-levko: Yes, it's a subject to discuss

redash/models/users.py
```diff
@@ -187,8 +187,14 @@ def get_by_api_key_and_org(cls, api_key, org):
         return cls.get_by_org(org).filter(cls.api_key == api_key).one()
 
     @classmethod
-    def all(cls, org):
-        return cls.get_by_org(org).filter(cls.disabled_at.is_(None))
+    def all(cls, org, is_invitation_pending=None):
```
rauchy: Some delicate stuff here. Could we avoid having this code by providing better hints from the frontend? If not, can we add a test for this?

redash/handlers/users.py
```diff
@@ -40,6 +40,18 @@ def invite_user(org, inviter, user):
     send_invite_email(inviter, user, invite_url, org)
 
 
+def bool_or_none(v):
```
rauchy: image

client/app/pages/users/UsersList.jsx
kravets-levko: There is the only user that is not invited :smile:

rauchy:
One small nit - I find that having the orange "Invitation Pending" tag on a "Pending Invitations" page redundant.

ranbena:
Forgot to say, approved on my part but I'm no authority on Python stuffs...

kravets-levko:
@rauchy When user is non-admin, that page looks a bit different - it shows all users, both active and pending. At least there we need this label. But for admin probably not (as for me - I'd keep it)

ranbena:
> But for admin probably not

I agree with @rauchy - when showing pending only, don't render the label.

kravets-levko:
> Forgot to say, approved on my part but I'm no authority on Python stuffs..

Therefore I requested review from @arikfr and (especially) @rauchy (he is our Python guru :nerd_face:)

rauchy:
re: label on the admin page - imagine an admin inviting 15 new members - they all see the tag. Kinda fugly.

re: Python - I wrote my "Hello, World!" in Python 3 months ago so, you know, don't expect too much :)

redash/models/users.py
kravets-levko: Of course I'll add tests for this. I'll try to explain the idea behind this, maybe I'm doing something wrong. We have a method to get all disabled users (`all_disabled`) and this one - to get all enabled users. But in some cases we need all enabled users, and sometimes - only active or only pending. So I added parameter `is_invitation_pending` which is a filter: 

- `is_invitation_pending=None` - fetch all
- `is_invitation_pending=False` - only active
- `is_invitation_pending=True` - only pending.

I don't know what is the right way to create API like this; this implementation works as I need, so ATM we have what we have. But I'll appreciate any suggestions for improving this.

redash/handlers/users.py
kravets-levko: Okay, I'll move this to handler

arikfr:
@kravets-levko 
> When user is non-admin, that page looks a bit different - it shows all users, both active and pending. 

Why?

redash/models/users.py
rauchy: Sorry, I'm a total SQLAlchemy n00b, but I'd expect something similar to an ActiveRecord scope to be used here.

redash/handlers/users.py
arikfr: If we show them on a different tab (makes sense), how about using a dedicated API call (`/api/users/invites`), so we don't need this at all?

kravets-levko:
@arikfr It already worked in this way, so I decided to keep it; but I think it's better to show sidebar for non-admins, but hide "Disabled Users" menu. Will do it

redash/models/users.py
arikfr: At first it seemed like a terrible API design, but on second thought maybe it's ok...

redash/models/users.py
```diff
@@ -187,8 +187,14 @@ def get_by_api_key_and_org(cls, api_key, org):
...
+        condition = True  # no filter
+        if is_invitation_pending is not None:
+            if is_invitation_pending:
+                condition = cls.is_invitation_pending.is_(True)  # pending
+            else:
+                condition = cls.is_invitation_pending.isnot(True)  # not pending (check for both `false`/`null`)
+        return cls.get_by_org(org).filter(cls.disabled_at.is_(None), condition)
```
arikfr: What you can do instead is:

```python
users = cls.get_by_org(org).filter(cls.disabled_at.is_(None))

if is_invitiation_pending is not None:
    users = users.filter(cls.is_invitation_pending.is_(is_invitation_pending)) 

return users
```

You should _not_ check for `null` value, because existing users have a null value (@rauchy correct me if I'm wrong).

redash/models/users.py
rauchy: Surely there must be a way to achieve something like `User.all.pending`?

redash/handlers/users.py
kravets-levko: I don't think it's a good idea - implementation of a new endpoint handler will be 99% the same (with search support, pagination, etc.)

redash/models/users.py
kravets-levko: @rauchy Yes, that's another possible option. Therefore I need your help to choose better one

redash/models/users.py
kravets-levko: Yes, yes, YES!! That's exactly what I need. Thanks!

redash/handlers/users.py
kravets-levko: @arikfr (cc @rauchy) What is I'll add `get` handler to [UserInviteResource](https://github.com/getredash/redash/blob/master/redash/handlers/users.py#L132)?

kravets-levko: Nevermind, I'll create a new resource for this

kravets-levko:
@arikfr @ranbena @rauchy:

1. show sidebar both for admins and non-admins; for non-admins "Disabled Users" section is hidden;
2. removed "Invitation pending" label;
3. refined API (handler and model) implementation.

tests/handlers/test_users.py
```diff
@@ -74,17 +74,72 @@ def test_returns_400_when_email_taken_case_insensitive(self):
...
+    def test_returns_users_by_filter(self):
+        now = models.db.func.now()
+        user_enabled_active1 = self.factory.create_user(disabled_at=None, is_invitation_pending=None)
+        user_enabled_active2 = self.factory.create_user(disabled_at=None, is_invitation_pending=False)
+        user_enabled_pending = self.factory.create_user(disabled_at=None, is_invitation_pending=True)
+        user_disabled_active1 = self.factory.create_user(disabled_at=now, is_invitation_pending=None)
+        user_disabled_active2 = self.factory.create_user(disabled_at=now, is_invitation_pending=False)
+        user_disabled_pending = self.factory.create_user(disabled_at=now, is_invitation_pending=True)
+
+        # get all enabled
```
rauchy: We should strive to have one logical assertion per test. Could you separate all these to different tests?

kravets-levko: I decided to put them together because they use the same set of users and are related to the same functionality. Wasn't sure how to do it right :man_shrugging:

rauchy: `make_request_and_check_users` is a step in the right direction. Next I'd move all those `create_user` calls to the fixture level and split each call to `make_request_and_check_users` to its own test, indicating what it is testing ("test_gets_all_disabled_and_pending" etc)

rauchy: Another thing that bothers me - I'd prefer to see a more idiomatic test which actually asserts in the test body. Consider transforming `make_request_and_check_users` so it makes the request and returns the users, then use a set to make sure that the return value equals your expected users. Something like:

```python
expected = Set([ ua1, ua2 ])
actual = Set(make_request("pending=false"))
self.assertEquals(expected, actual)
```

Also, this way you don't have to assert about what _doesn't_ come back.

tests/handlers/test_users.py
```diff
@@ -81,75 +81,85 @@ class PlainObject(object):
...
+    def make_request_and_return_ids(self, *args, **kwargs):
+        rv = self.make_request(*args, **kwargs)
+        return map(lambda u: u['id'], rv.json['results'])
+
+    def assertUsersListMatches(self, actual_ids, expected_ids, unexpected_ids):
+        actual_ids = set(actual_ids)
+        expected_ids = set(expected_ids)
+        unexpected_ids = set(unexpected_ids)
+        self.assertSetEqual(actual_ids.intersection(expected_ids), expected_ids)
+        self.assertSetEqual(actual_ids.intersection(unexpected_ids), set())
```
rauchy: I'm not sure why you care about unexpected ids. If you work with a set, you test complete, unordered equality. No need for intersection stuff as well.

```python
Set([1,2,3]) == Set([1,3,2])
Out: True

Set([1,2,3]) == Set([1,3,2,4])
Out: False
```

You can just get rid of `assertUsersListMatches` and assert for set equality in the tests. i.e:

```python
def test_gets_all_enabled(self, users): # using a fixture here
    actual = self.make_request_and_return_ids('get', '/api/users')
    self.assertSetEquals(actual, Set([users.enabled_active1, users.enabled_pending]))
```

kravets-levko: I'll try to explain. I'm creating all possible combinations of users for my tests. But there may be other users in DB that will match particular conditions. So for example I create users 1,2,3 and do some request. Users 1,2 will match filter, user 3 not. But it's possible that pre-existed users will also be there, and, say, they'll also match filter, so result will be 1,2,3 instead of 1,2. If I'll compare for equality, test will fail. Therefore I check exactly what I have to check: which of users I expect to see in output and which shouldn't be there for sure.

tests/handlers/test_users.py
kravets-levko: https://github.com/getredash/redash/pull/3400#discussion_r254285338

arikfr:
👍 

Congrats on a full stack feature :-)

arikfr:
image

You also managed to collect all the PR Triage labels 👌

PR #3403, opened 2019-02-06
Handle InterruptException in Athena query runnner

jezdez:
This is a follow-up to https://github.com/getredash/redash/pull/1753 where handling this exception was added to the Presto query runner. It turns out we have some [anecdotal evidence](https://github.com/mozilla/redash/issues/775) that the Athena query runner also triggers that exception and requires additional help to cancel the query.


arikfr:
👍

Issue #2891, opened 2018-10-03
Move to Python 3: dependencies status


arikfr:
Thanks to the work of @cclauss our code is (more) ready for Python 3. I also ran [caniusepython3](https://github.com/brettcannon/caniusepython3) on our requirements files and the results are very encouraging:

```
$ caniusepython3 --requirements requirements.txt
Finding and checking dependencies ...

You need 5 projects to transition to Python 3.
Of those 5 projects, 4 have no direct dependencies blocking their transition:

  chromelogger
  flask-sslify
  python-geoip (which is blocking python-geoip-geolite2)
  restrictedpython
```

* `python-geoip` has a Python 3 compatible alternative (`python-geoip-python3`)
* Latest master version of `flask-sslify` is marked as Python 3 compatible, but no new version was released for a long time.
* `chromelogger` is unknown but we can either drop it or update to support Python 3 (~140 lines of code).
* `restrictedpython` has an unreleased version 4 that supports Python 3.

```
$ caniusepython3 --requirements requirements_all_ds.txt
Finding and checking dependencies ...

You need 1 project to transition to Python 3.
Of that 1 project, 1 has no direct dependencies blocking its transition:

  pyhive
```

I'm not sure why `pyhive` is marked as not supported, because it is. Maybe we just need to update the version.

```
$ caniusepython3 --requirements requirements_dev.txt
Finding and checking dependencies ...

You need 1 project to transition to Python 3.
Of that 1 project, 1 has no direct dependencies blocking its transition:

  pytest-watch
```

I'm not sure about the state of `pytest-watch`, but we can always remove it.

arikfr:
@jezdez sounds good to me. 👍

truebit:
I saw many ImportError when trying to run on python3.
Is there any issue to fix those incompatible codes for Python3?

jezdez:
@arikfr Oh btw, we merged flask-talisman! Which means there shouldn't be any blocker to this?

arikfr:
@jezdez We should probably run `caniusepython3` again to review if anything changed.

But there are probably other things that are blocking switching over to Python 3 in our code or dependencies. I'm just not sure what they are. 😳

jezdez:
```
$ caniusepython3 -r requirements.txt
Finding and checking dependencies ...

You need 3 projects to transition to Python 3.
Of those 3 projects, 2 have no direct dependencies blocking their transition:

  chromelogger
  python-geoip (which is blocking python-geoip-geolite2)
[...]

PR #3404, opened 2019-02-06
Use flask-talisman for handling backend response headers

### LINKED ISSUE #3060: Adopt flask-talisman
### Issue Summary

Right now we don't really configure some of the more common HTTP response headers that are recommended for web applications. Flask-Talisman (from the Google Cloud Platform team) would help us do this. It improves various security related Flask issues (including but limited to mozilla#562) and move us closer to figure out #2891.

Related: #3044.

jezdez:
Fixes #3060.
Refs #3044.
Refs https://github.com/mozilla/redash/issues/562.
Helps #2891. 🎉

### DIFF
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -245,6 +245,7 @@ def init_app(app):
     login_manager.anonymous_user = models.AnonymousUser
 
     app.secret_key = settings.COOKIE_SECRET
+    app.config.update(REMEMBER_COOKIE_SECURE=True, REMEMBER_COOKIE_HTTPONLY=True)
     app.register_blueprint(google_oauth.blueprint)
     app.register_blueprint(saml_auth.blueprint)
     app.register_blueprint(remote_user_auth.blueprint)
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -244,8 +244,6 @@ def init_app(app):
     login_manager.init_app(app)
     login_manager.anonymous_user = models.AnonymousUser
 
-    app.secret_key = settings.COOKIE_SECRET
-    app.config.update(REMEMBER_COOKIE_SECURE=True, REMEMBER_COOKIE_HTTPONLY=True)
     app.register_blueprint(google_oauth.blueprint)
     app.register_blueprint(saml_auth.blueprint)
     app.register_blueprint(remote_user_auth.blueprint)
--- a/redash/security.py
+++ b/redash/security.py
@@ -1,0 +1,28 @@
+from flask_talisman import talisman
+
+from redash import settings
+
+talisman = talisman.Talisman()
+
+
+def init_app(app):
+    talisman.init_app(
+        app,
+        feature_policy=settings.FEATURE_POLICY,
+        force_https=settings.ENFORCE_HTTPS,
+        force_https_permanent=settings.ENFORCE_HTTPS_PERMANENT,
+        force_file_save=settings.ENFORCE_FILE_SAVE,
+        frame_options=settings.FRAME_OPTIONS,
+        frame_options_allow_from=settings.FRAME_OPTIONS_ALLOW_FROM,
+        strict_transport_security=settings.HSTS_ENABLED,
+        strict_transport_security_preload=settings.HSTS_PRELOAD,
+        strict_transport_security_max_age=settings.HSTS_MAX_AGE,
+        strict_transport_security_include_subdomains=settings.HSTS_INCLUDE_SUBDOMAINS,
+        content_security_policy=settings.CONTENT_SECURITY_POLICY,
+        content_security_policy_report_uri=settings.CONTENT_SECURITY_POLICY_REPORT_URI,
+        content_security_policy_report_only=settings.CONTENT_SECURITY_POLICY_REPORT_ONLY,
+        content_security_policy_nonce_in=settings.CONTENT_SECURITY_POLICY_NONCE_IN,
+        referrer_policy=settings.REFERRER_POLICY,
+        session_cookie_secure=settings.SESSION_COOKIE_SECURE,
+        session_cookie_http_only=settings.SESSION_COOKIE_HTTPONLY,
+    )
--- a/redash/security.py
+++ b/redash/security.py
@@ -1,8 +1,23 @@
+import functools
 from flask_talisman import talisman
 
 from redash import settings
 
+
 talisman = talisman.Talisman()
+
+
+def csp_allows_embeding(fn):
+
+    @functools.wraps(fn)
+    def decorated(*args, **kwargs):
+        return fn(*args, **kwargs)
+
+    embedable_csp = talisman.content_security_policy + "frame-ancestors *;"
+    return talisman(
+        content_security_policy=embedable_csp,
+        frame_options=None,
+    )(decorated)
 
 
 def init_app(app):
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -1,5 +1,6 @@
 import os
 from funcy import distinct, remove
+from flask_talisman import talisman
 
 from .helpers import fix_assets_path, array_from_string, parse_boolean, int_or_none, set_from_string
 from .organization import DATE_FORMAT
@@ -14,7 +15,6 @@ def all_settings():
             settings[name] = item
 
     return settings
-
 
 REDIS_URL = os.environ.get('REDASH_REDIS_URL', os.environ.get('REDIS_URL', "redis://localhost:6379/0"))
 PROXIES_COUNT = int(os.environ.get('REDASH_PROXIES_COUNT', "1"))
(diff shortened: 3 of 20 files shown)


rauchy:
Thanks @jezdez!

@arikfr if you merge this, we can discard #3214

jezdez:
@rauchy Oh dang, I had missed your PR, apologies for the overlap 😬

redash/security.py
```diff
@@ -0,0 +1,28 @@
...
+        strict_transport_security_preload=settings.HSTS_PRELOAD,
+        strict_transport_security_max_age=settings.HSTS_MAX_AGE,
+        strict_transport_security_include_subdomains=settings.HSTS_INCLUDE_SUBDOMAINS,
+        content_security_policy=settings.CONTENT_SECURITY_POLICY,
+        content_security_policy_report_uri=settings.CONTENT_SECURITY_POLICY_REPORT_URI,
+        content_security_policy_report_only=settings.CONTENT_SECURITY_POLICY_REPORT_ONLY,
+        content_security_policy_nonce_in=settings.CONTENT_SECURITY_POLICY_NONCE_IN,
+        referrer_policy=settings.REFERRER_POLICY,
+        session_cookie_secure=settings.SESSION_COOKIE_SECURE,
+        session_cookie_http_only=settings.SESSION_COOKIE_HTTPONLY,
```
arikfr: This plays along with Flask-Login?

redash/settings/__init__.py
```diff
@@ -48,9 +49,77 @@ def all_settings():
...
+    os.environ.get("REDASH_SESSION_COOKIE_SECURE") or str(ENFORCE_HTTPS))
+# Whether the session cookie is set HttpOnly.
+SESSION_COOKIE_HTTPONLY = parse_boolean(
+    os.environ.get("REDASH_SESSION_COOKIE_HTTPONLY", "true"))
+
+# Doesn't set X-Frame-Options by default since it's highly dependent
+# on the specific deployment.
+# See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options
+# for more information.
+FRAME_OPTIONS = os.environ.get("REDASH_FRAME_OPTIONS", None)
```
arikfr: Does Talisman support setting this on a per route basis? Because we should set this by default to `deny` for all routes except the embeddable ones.

redash/settings/__init__.py
```diff
@@ -48,9 +49,77 @@ def all_settings():
...
+FRAME_OPTIONS_ALLOW_FROM = os.environ.get(
+    "REDASH_FRAME_OPTIONS_ALLOW_FROM", "")
+
+# Whether and how to send Strict-Transport-Security response headers.
+# See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
+# for more information.
+HSTS_ENABLED = parse_boolean(os.environ.get("REDASH_HSTS_ENABLED", "false"))
```
arikfr: I'm pretty sure Flask-SSLify was setting it up by default, so maybe the default value should be the one of HTTPS_ENFORCED?

redash/settings/__init__.py
```diff
@@ -48,9 +49,77 @@ def all_settings():
...
+HSTS_MAX_AGE = int(
+    os.environ.get("REDASH_HSTS_MAX_AGE", talisman.ONE_YEAR_IN_SECS))
+HSTS_INCLUDE_SUBDOMAINS = parse_boolean(
+    os.environ.get("REDASH_HSTS_INCLUDE_SUBDOMAINS", "false")
+)
+
+# Whether and how to send Content-Security-Policy response headers.
+# See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
+# for more information.
+CONTENT_SECURITY_POLICY = os.environ.get(
```
arikfr: `os.environ.get` returns a string, while the default value is an object. You probably want json.loads/dumps here.

requirements.txt
```diff
@@ -56,7 +56,7 @@ user-agents==1.1.0
 python-geoip-geolite2==2015.303
 chromelogger==0.4.3
 pypd==1.1.0
-disposable-email-domains
+disposable-email-domains==0.0.34
```
arikfr: Why set a specific version?

redash/settings/__init__.py
jezdez: @arikfr Yep, it does, like I did for the [ping endpoint](https://github.com/getredash/redash/pull/3404/files#diff-5935f49d3a59e4651b814a952cd84919R12). More info: https://github.com/GoogleCloudPlatform/flask-talisman#per-view-options

redash/settings/__init__.py
jezdez: That makes sense. For the record, Flask-SSLify only did that when the `request.is_secure` was truthy, which is only the case when the [wsgi environment has `wsgi.url_scheme` set to "https"](https://github.com/pallets/werkzeug/blob/36ed800599f45b74c742b2d9bd3de0ecb93a8547/werkzeug/wrappers/base_request.py#L639-L640), which in turn only happens when `ENFORCE_HTTPS` was `True`.

redash/settings/__init__.py
jezdez: Ah yeah, so the values for policy-style config variables (CONTENT_SECURITY_POLICY and FEATURE_POLICY) can either be full blown dicts or strings that are parsed automatically by Talisman: https://github.com/GoogleCloudPlatform/flask-talisman/blob/3c1a0d3555b16e918cbdc9c0c769f37a0da24970/flask_talisman/talisman_test.py#L150-L168

That's why it's okay to define the values as dicts here and not have to use json for the env vars. If we were to set the CONTENT_SECURITY_POLICY or FEATURE_POLICY env vars, we'd have to use the string version basically (e.g. `CONTENT_SECURITY_POLICY="default-src 'self'"`)

Would you prefer json instead?

requirements.txt
jezdez: To reduce the chance for breakage on an update of that dependency, for the same reason we pin all other packages.

jezdez: I noticed since I've written this patch there was a new version, it's now at 0.0.45.

redash/security.py
jezdez: AFAIU since Flask-Login simply relies on the Flask session system for tracking login state and has [for a while](https://github.com/maxcountryman/flask-login/commit/312d0e4554d3e0dfb7cd3973d79a7f342d645bff) shipped with an own httpOnly functionality for its rememberme cookie.

requirements.txt
arikfr: I know why in general we pin dependencies. It just that in this case it's not pinned on purpose. This is a simple list of domains, so no real danger of breakage. Unless they change the list name (which they did, but kept the old names).

I would rather add a test that asserts the list exists and contains domains rather than pin the version and have the hassle of upgrading it.

As a side note, such changes should be in their own Pull Request.

redash/settings/__init__.py
arikfr: In this case, let's add a link to the documentation on what formats are accepted here. Because people usually look at the default value to understand what's expected.

redash/settings/__init__.py
arikfr: > which in turn only happens when `ENFORCE_HTTPS` was `True`.

Exactly what I meant and suggested. Sorry for not being clear, but when I said it was setting it up by default, I meant when HTTPS is enforced.

redash/settings/__init__.py
arikfr: Then how about we use a strict policy here by default, and use per-view options to allow embedding for: visualization embed, public dashboard and non public dashboard?

redash/settings/__init__.py
jezdez: Good point!

redash/settings/__init__.py
jezdez: No worries, I wasn't exactly sure about the behavior, so it's good we spelled it out.

redash/settings/__init__.py
jezdez: That works for me, I'll append it..

requirements.txt
jezdez: Right, I think the original feature that we had this commit in was for some reason attached to this, and we used pyup at the time that informed us about the unpinned dependency.
Also totally right to split this, I think you reminded me a few times now to not conflate different changes in one PR 😬 Rest assured this isn't on purpose but often a result of the mess of unbundling our fork changes :shakes fist at past:

redash/settings/__init__.py
arikfr: Thanks.

jezdez: Wait, we have a problem, according to the spec:

```
   Wildcards or lists to declare multiple domains in one ALLOW-FROM
   statement are not permitted (see Section 2.3.2.3).
```

So we can't specify a list of URLs or a wildcard for the embed endpoints when using `X-Frame-Options: allow-from https://example.com/`. I don't think it's practical for us to spell out the individual URLs that embedding is allowed either.

There is a better alternative though, the [CSP directive frame-ancestors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors). It's [in CSP 2](https://www.w3.org/TR/CSP2/#directive-frame-ancestors), which means we should simply use the CSP header. Browser support for it isn't great though, IE and Mobile Edge don't support it, while X-Frame-Options has broader support. My gut says we should go with it anyway  though. What do you think?

arikfr: I thought of allowing a wildcard there.

redash/authentication/__init__.py
```diff
@@ -244,7 +244,6 @@ def init_app(app):
     login_manager.init_app(app)
     login_manager.anonymous_user = models.AnonymousUser
 
-    app.secret_key = settings.COOKIE_SECRET
```
jezdez: This is just a cleanup change that was missing in #2970.

arikfr: See other comment about this ⬇️

redash/settings/__init__.py
```diff
@@ -106,9 +185,6 @@ def all_settings():
 STATIC_ASSETS_PATH = fix_assets_path(os.environ.get("REDASH_STATIC_ASSETS_PATH", "../client/dist/"))
 
 JOB_EXPIRY_TIME = int(os.environ.get("REDASH_JOB_EXPIRY_TIME", 3600 * 12))
-COOKIE_SECRET = os.environ.get("REDASH_COOKIE_SECRET", "c292a0a3aa32397cdb050e233733900f")
-SESSION_COOKIE_SECURE = parse_boolean(os.environ.get("REDASH_SESSION_COOKIE_SECURE") or str(ENFORCE_HTTPS))
-SECRET_KEY = os.environ.get('REDASH_SECRET_KEY', COOKIE_SECRET)
```
arikfr: `SECRET_KEY` is supposed to be used only for encrypting data source options. `COOKIE_SECRET` is the `COOKIE_SECRET`. While by default we use the same value for both, the idea was to allow setting a different one, if the user wants to.

From checking your PR, I realize that Flask calls this `SECRET_KEY` and hence the confusion here. 🤦‍♂️ 

This needs to be reverted to previous behavior.

jezdez: @arikfr Yeah, I understand the confusion, `SECRET_KEY` is automatically set to `app.secret_key` when we update `app.config` with the settings values after all. Plus we are currently using the value of `REDASH_COOKIE_SECRET` as the value for `SECRET_KEY`.

Since app.secret_key "will be used for securely signing the session cookie and can be used for **any other security related needs by extensions or your application**" [docs](http://flask.pocoo.org/docs/1.0/config/#SECRET_KEY) and it's often used in the extension community in that sense as well, shouldn't we introduce a `REDASH_DATASOURCE_SECRET_KEY` just for the encryption of the datasource secrets?

So basically shouldn't we have three things to be able to handle:

REDASH_SECRET_KEY = random string
REDASH_DATASOURCE_SECRET_KEY = random string
REDASH_COOKIE_SECRET_KEY = REDASH_SECRET_KEY or own string

Alternatively we could simply use REDASH_SECRET_KEY for the secure cookies and not have a separate setting REDASH_COOKIE_SECRET_KEY.

Either way, decoupling the encryption keys of the data source options from the session/cookie encryption keys seem like a good idea in case one of it is leaked and would reduce the impact.
Do you concur?

jezdez: @arikfr Quick ping, what I should do?

arikfr: > Yeah, I understand the confusion, SECRET_KEY is automatically set to app.secret_key when we update app.config with the settings values after all. 

But it was overridden back to `COOKIE_SECRET` value in `redash.authentication`.

While we do use the `COOKIE_SECRET` as the default value for the data source `SECRET_KEY`, it still should be decoupled to allow setting a different value.

So the updated code should be:

```python
SECRET_KEY = os.environ.get("REDASH_COOKIE_SECRET", "c292a0a3aa32397cdb050e233733900f")
DATASOURCE_SECRET_KEY = os.environ.get('REDASH_SECRET_KEY', SECRET_KEY)
```

And then update in `redash.models` to use `redash.settings.DATASOURCE_SECRET_KEY` instead of `redash.settings.SECRET_KEY`, while updating the Flask code to `redash.settings.SECRET_KEY`. There is no need to have `COOKIE_SECRET` anymore (and if we used it explicitly somewhere, just switch to SECRET_KEY).

This way everything stays backward compatible, while also consistent with Flask's naming.

jezdez: @arikfr Hm, from a user perspective I think that environment variable naming would be inconsistent with what they'd be used for.

With your proposal `REDASH_COOKIE_SECRET` is actually used as the Flask secret key, but `REDASH_SECRET_KEY` is used just for encrypting data source options. This seems like a guarantee for user confusion.

But we can easily prevent this by keeping `REDASH_SECRET_KEY` as the default setting for the one and only Flask secret key (and the one we promote in the docs) and treat `REDASH_COOKIE_SECRET` as deprecated.

```python
SECRET_KEY = os.environ.get(
    "REDASH_COOKIE_SECRET",
) or os.environ.get(
    "REDASH_SECRET_KEY",
    "c292a0a3aa32397cdb050e233733900f"
)  # using or here to cover for an empty REDASH_COOKIE_SECRET
```

And add a new `REDASH_DATASOURCE_SECRET_KEY` environment variable for the new feature to encrypt datasource options?

```python
DATASOURCE_SECRET_KEY = os.environ.get('REDASH_DATASOURCE_SECRET_KEY', SECRET_KEY)
```

Sorry for being a stickler here, I'm worried that we can't redo this later on :)

arikfr: I agree it's not ideal, and if I was aware of the setting in Flask before I would've picked a different name. But changing this now can break existing deployments, and therefore I don't see many options other than preserve it.

jezdez: The fallback above would still be backward compatible IIUC, but maybe I'm missing something. 🙈

arikfr: How will it support someone setting the `REDASH_SECRET_KEY` value to a different value than `REDASH_COOKIE_SECRET` to be used with data sources?

jezdez: In Redash v6 we only have `REDASH_COOKIE_SECRET` which they can keep using or switch to `REDASH_SECRET_KEY`.

And then optionally define `REDASH_DATASOURCE_SECRET_KEY` (which defaults to the value of the secret key above) when needed. For the setup scripts I suggest to also add `REDASH_DATASOURCE_SECRET_KEY` to `/opt/redash/env`.

So all those situations work:

1. Just using the Redash <= v6 name:
```
REDASH_COOKIE_SECRET="foo"
```

2. Just specifying the secret key with the new name:
```
REDASH_SECRET_KEY="foo"
```

3. Specifying both secret and encryptio key:
```
REDASH_SECRET_KEY="foo"
REDASH_DATASOURCE_SECRET_KEY="bar"
```

arikfr: > In Redash v6 we only have REDASH_COOKIE_SECRET

Yes, but there are people already using v7. I don't know if they used the new setting already or not, but prefer not to break their deployments.

jezdez: Aaaaah, that would break things, indeed. Sorry I was missing this detail :)

jezdez: Okay, updated this for the last time.

jezdez:
@arikfr Is there anything needed to merge this?

jezdez:
@arikfr Anything needed to merge this?

arikfr:
> Anything needed to merge this?

Time.

PR #3405, opened 2019-02-06
Upgrade jest & babel

koooge:
Hello,
upgrade jest 24.1.0 and babel to 7 👐


koooge:
It seems we need to upgrade babel frist 🤔

client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -206,7 +206,7 @@ export class ScheduleDialog extends React.Component {
             <h5>On time</h5>
             <div data-testid="time">
               <TimePicker
-                allowEmpty={false}
+                allowClear={false}
```
ranbena: How did you know this changed? Console error?

package.json
```diff
@@ -83,38 +83,35 @@
     "underscore.string": "^3.3.4"
   },
   "devDependencies": {
-    "babel-core": "^6.26.0",
+    "@babel/core": "^7.2.2",
+    "@babel/plugin-proposal-class-properties": "^7.3.0",
```
ranbena: Yay! (wanted this but thought of waiting till TS which allows this by default)

package.json
```diff
@@ -83,38 +83,35 @@
...
+    "@babel/plugin-transform-object-assign": "^7.2.0",
```
ranbena: IE support?

package.json
kravets-levko: @ranbena This syntax is already supported (at least I use it sometimes) - not sure which plugin does it, but...

ranbena: Show me a in the code!

kravets-levko: [For example](https://github.com/getredash/redash/blob/master/client/app/components/ParameterMappingInput.jsx#L119-L139)

kravets-levko: and BTW - it's even your code :smile:

ranbena: Right. I guess what confused me is that I'm used to defining these first in classes (props defined outside class, constructor ,method rarely used). [Like here](https://github.com/CityOfZion/neon-wallet/blob/dev/app/hocs/withCameraAvailability.js).

ranbena:
@koooge you didn't write any reasoning for why the upgrades are at all needed... ?

client/app/components/queries/ScheduleDialog.jsx
koooge: Yes, you can know it in test

```
    console.error node_modules/warning/warning.js:34
      Warning: `allowEmpty` in TimePicker is deprecated. Please use `allowClear` instead.
```

package.json
koooge: Ah, actually I didn't check on IE, so even IE11 might not work since preset-env transforms ES2015+ by default.
It looks we can use the specific target polyfill https://babeljs.io/docs/en/babel-preset-env#targets. Personally, I don't use IE, so I didn't really pay attention.

kravets-levko: @koooge Can you please test this configuration as carefully as possible? A lot of Redash users still use IE10/11/Edge, so it's crucial to keep everything working.

package.json
koooge: To be honest, I follow the instruction below 🤔 
```
 PASS  client/app/components/Footer.test.js
 PASS  client/app/components/UserShow.test.js
 FAIL  client/app/components/queries/ScheduleDialog.test.js
  œ Test suite failed to run                                                                                                                          [        
    SyntaxError: /home/vagrant/go/src/github.com/koooge/redash/client/app/components/queries/ScheduleDialog.jsx: Support for the experimental syntax 'classPrope
rties' isn't currently enabled (23:20):

      21 |
      22 | export class ScheduleDialog extends React.Component {
    > 23 |   static propTypes = {
         |                    ^
      24 |     show: PropTypes.bool.isRequired,
      25 |     schedule: RefreshScheduleType,
      26 |     refreshOptions: PropTypes.arrayOf(PropTypes.number).isRequired,

    Add @babel/plugin-proposal-class-properties (https://git.io/vb4SL) to the 'plugins' section of your Babel config to enable transformation.

      at Object.raise (node_modules/@babel/parser/lib/index.js:3831:17)
      at Object.expectPlugin (node_modules/@babel/parser/lib/index.js:5148:18)
      at Object.parseClassProperty (node_modules/@babel/parser/lib/index.js:8229:12)
      at Object.pushClassProperty (node_modules/@babel/parser/lib/index.js:8193:30)
      at Object.parseClassMemberWithIsStatic (node_modules/@babel/parser/lib/index.js:8132:14)
      at Object.parseClassMember (node_modules/@babel/parser/lib/index.js:8069:10)
      at withTopicForbiddingContext (node_modules/@babel/parser/lib/index.js:8024:14)
      at Object.withTopicForbiddingContext (node_modules/@babel/parser/lib/index.js:7150:14)
      at Object.parseClassBody (node_modules/@babel/parser/lib/index.js:8001:10)
      at Object.parseClass (node_modules/@babel/parser/lib/index.js:7975:10)
```

ranbena: That's weird. I don't get that error..
@kravets-levko any idea what's making this syntax valid without the plugin?

kravets-levko: nodejs version :slightly_smiling_face:

ranbena: So we should upgrade node from the currently defined v8.0.0?

kravets-levko: I currently use `v10.13.0`, but `v8.0.0` seems to work fine for everything except of tests. Not sure what we should change... :thinking: Is it possible to setup `jest` to use babel for loading test files?

koooge: The test failed with the same message in my environment with node 8.15.0 which is no @babel/plugin-proposal-class-properties.
My guess is the test passes because the master branch has babel-preset-stage2. 
https://github.com/tc39/proposal-class-public-fields

arikfr: I'm not sure what's the best practice here, but maybe we should setup Jest to run on code post webpack compilation? It might slow down tests, so maybe there are better solutions?

package.json
arikfr: We should definitely target Edge and maybe IE11 but not beyond that.

package.json
koooge: My understanding is that class-properties is still experimental stage https://github.com/tc39/proposal-class-public-fields.
After all, `npm test` and `npm build` need @babel/plugin-proposal-class-properties in babel7 instead of npm-preset-stage2 in babel6.

package.json
kravets-levko: @arikfr 

image

(IE7 there probably is IE11/Edge in quirks mode)

koooge: okay I just tested it on win10 ie11 and edge, works fine.

koooge:
@ranbena ping? What ahould I do? 

#3515

ranbena:
> @ranbena ping? What should I do? What do you concern?

Hey @koooge no concern, just not sure what the motivation is for upgrading.

koooge:
> > @ranbena ping? What should I do? What do you concern?
> 
> Hey @koooge no concern, just not sure what the motivation is for upgrading.

Oh, I see. There are some boring reasons.
- Expect improvement fo the performance/feature
- Expect bugfix
- Expect to fix vulnerabilities
- Update smoothly to prevent that next update becomes a big change

The last one is main reason in this PR.

ranbena:
> The last one is main reason in this PR.

You mean, you have an upcoming PR that relies on this?

koooge:
> > The last one is main reason in this PR.
> 
> You mean, you have an upcoming PR that relies on this?
I mean that upgrading major version of these dependencies is tough. it would be greater to upgrade them strategically.

ranbena:
Gonna merge this on Sunday 👍

@koooge we'll be halting upgraded till we finalize migration to React 👷, then we'll do a thorough audit of our packages - relevance and version bump.

PR #3407, opened 2019-02-06
Refine dialog wrapper and use it for all existing dialogs

kravets-levko:
See comment in DialogWrapper.jsx for wrapper and API details.

Also suggesting convention for naming dialogs: 

- use `Dialog` suffix instead of `Modal`:
  - `AddWidgetModal` - wrong
  - `AddWidgetDialog`- correct

- Name of method that opens modal should reflect what it is going to do:
  - `showAddWidgetModal()` - wrong
  - `addWidget()` - correct


ranbena:
@kravets-levko I looked at this a lot. I keep asking myself if there's a way to enforce less by swapping Modal for a wrapper component?
```jsx
import Dialog from '@/components/Dialog';

render() {
  return {
    <Dialog
      visible={this.state.showModal} // not directly mapped to <Modal visible>
      afterClose={this.props.onClose} // will trigger destroyDialog
      onOK={() => save()} 
      onCancel={this.close()} //  will trigger dismissDialog
  }
}
```

Not sure this would allow covering all requirements from the dialog (e.g. resolve/reject), but if this can work, I think it'll be a better more transparent approach.

kravets-levko:
@ranbena With this wrapper I'm trying to avoid two things:

1. rendering dialog component at all: first of all, you have to decide when to render it, also it will be added into arbitrary place in document which may cause z-index issues (especially when used with some 3rd-party libraries - note that I removed some ugly code from `ScheduleDialog` related to this);
2. state management: you have to add a variable to manage dialog's visibility, and use `setState` when you need to show dialog. And while it's not a problem when you have 1-2 dialog, it becomes annoying when you need a lot of them.

Also, in my opinion, it's better level of abstraction: dialog is a kind of another window which is shown out of current context. It follows two things:

1. it's more natural to a consumer of a dialog component (caller) to _show_ (or _open_) it, rather then _render_;
2. the calling code becomes more self-expressive: compare `this.setState({ isDialogVisible: true })` + dialog that is rendered somewhere else, vs. `SomeDialog.showModal(...)` - you show dialog, wait for results and process results all in the single place.

client/app/components/DialogWrapper.jsx
```diff
@@ -0,0 +1,172 @@
...
+  dialog.close = closeDialog;
+  dialog.dismiss = dismissDialog;
+
+  const result = {
+    close: closeDialog,
+    dismiss: dismissDialog,
+    result: new Promise((resolve, reject) => {
+      dialogResult.resolve = resolve;
+      dialogResult.reject = reject;
+    }),
```
ranbena: I believe there would be an "unhandled rejection" exception if the other end doesn't attach a `result.catch()`.
But you can add an empty `.finally(() => {})` here to suppress the error.

client/app/components/dashboards/AddWidgetDialog.jsx
```diff
@@ -40,14 +34,12 @@ class AddWidgetDialog extends React.Component {
       searchedQueries: [],
       selectedVis: null,
       parameterMappings: [],
-      showModal: false, // show only after recent queries populated to avoid height "jump"
```
ranbena: The visual problem that this fixed will now resurface :/

client/app/pages/queries/query.html
```diff
@@ -135,8 +135,7 @@ <h3>
                 <span class="zmdi zmdi-refresh"></span> Refresh Schedule</span>
             </td>
             <td class="p-t-15 text-right">
-              <schedule-dialog show="showScheduleForm" schedule="query.schedule" refresh-options="refreshOptions" update-query="updateQueryMetadata" on-close="closeScheduleForm"></schedule-dialog>
-              <schedule-phrase ng-click="openScheduleForm()" is-link="true" schedule="query.schedule" is-new="query.isNew()" />
```
ranbena: 👍

client/app/components/dashboards/AddWidgetDialog.jsx
kravets-levko: I'll check what I can do here

client/app/pages/queries/query.html
kravets-levko: yeah, this one was really ugly (I clearly understand that it wasn't possible to do it better at that moment)

client/app/components/DialogWrapper.jsx
kravets-levko: @ranbena  `.finally(() => {})` doesn't help :disappointed: and there is no way to suppress this error in browsers. Some libraries like Bluebird have dedicated API to suppress unhandled rejection error on particular promise instance. So I added an option to override support library used by this wrapper, and also added support for known features (ATM only Bluebird). Also I set up this wrapper to use AngularJS promise (`$q`) which does not show this error. When Angular will be finally removed we can return to this (probably will start to use Bluebird, or just ignore this kind of errors).

client/app/components/dashboards/AddWidgetDialog.jsx
kravets-levko: @ranbena I added loader which takes some space, so now user see some feedback on what's going on, and dialog does not "jump" so much.

client/app/components/dashboards/AddWidgetDialog.jsx
```diff
@@ -202,38 +206,27 @@ class AddWidgetDialog extends React.Component {
...
-                    onClick={() => this.selectQuery(query.id)}
-                  >
-                    {query.name}
-                    {' '}
-                    <QueryTagsControl tags={query.tags} className="inline-tags-control" />
-                  </a>
-                ))}
-              </div>
-            )}
+        {!isLoaded && (
```
kravets-levko: That was really a fragment that needed refactoring (99% duplicated code) - hey CodeClimate, where you was then?

client/app/components/DialogWrapper.jsx
```diff
@@ -0,0 +1,206 @@
...
+        return (
+          <Modal {...dialog.props} onOk={() => this.customOkHandler()}>
+        );
+    }
+
+
+  Settings
+  ========
+
+  You can setup this wrapper to use custom `Promise` library (for example, Bluebird):
```
ranbena: @kravets-levko can you add the reasoning here?

client/app/components/dashboards/AddWidgetDialog.jsx
ranbena: Yeah it looks better 👍

client/app/components/DialogWrapper.jsx
kravets-levko: Just because it was simple fix for unhandled rejections (I realized that we don't see them from `$uibModal` because it uses `$q` service (AngularJS promises) - then this idea came itself). Also I saw this approach in some (couldn't remember in which :sweat_smile:) libraries that allow to override some default dependency.

ranbena: I meant add the reasoning for using a custom Promise in the code :)

```suggestion
  To avoid "unhandledrejection" exception that would fire with native Promises, you can setup this wrapper to use custom `Promise` library (for example, Bluebird):
```

kravets-levko: Super :+1: Just changed text to be more generic :wink:

PR #3408, opened 2019-02-06
Add tag filters to backend-unit-tests to allow build-tarball to build

jasonthomas:
We ran into this issue in https://github.com/mozilla/redash/pull/807 and @jezdez mentioned this should be applied here as well.

Based on the CircleCI documentation for git tag workflows:

> if a job requires any other jobs (directly or indirectly), you must use regular expressions to specify tag filters for those jobs.

`build-tarball` job builds on tags and requires`backend-unit-tests` job however tag filters are not specified here. This change adds tag filters to the `backend-unit-tests` workflow so that `build-tarball` jobs are executed.

https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag to documentation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           name: Copy Test Results
           command: |
             mkdir -p /tmp/test-results/unit-tests
-            docker cp tests:/app/coverage.xml ./coverage.xml 
+            docker cp tests:/app/coverage.xml ./coverage.xml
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks, but we no longer use tags for triggering release builds on CircleCI, because of all the complexity around them in CircleCI. We just use the `release/*` branches instead. We should actually remove the tags filter from the tarball job.

PR #3409, opened 2019-02-06
Show query execution status at page load if query is running

washort:
On initial page load, check Redis to see if a job is running for this query. If so, immediately show it as executing.

### DIFF
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -446,6 +446,12 @@ function QueryResource(
       return new QueryResultError("Can't execute empty query.");
     }
 
+    if (this.job) {
+      const q = new QueryResult({ job: { id: this.job } });
+      q.refreshStatus(this);
+      return q;
+    }
+
     const parameters = this.getParameters();
     const missingParams = parameters.getMissing();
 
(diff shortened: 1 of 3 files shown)


washort:
The general idea is to let users know when a query has already been started so that multiple runs of the same query aren't started by different users.

I assumed the performance impact of a redis lookup would be relatively small. Our current use case is to show the execution banner in the query view but we also considered a future version that shows execution status in the query list view as well.

I see your point about not being able to see cached results. Not sure how this should be handled.

arikfr:
> The general idea is to let users know when a query has already been started so that multiple runs of the same query aren't started by different users.

Multiple users running the same query at the same time won't start multiple execution -- this is what the lock is for (the one you're using to lookup the job...).

Anyway. if that's the goal then maybe my suggestion of detecting when an execution of the current query is currently in progress, and showing a message about it, might be good enough.

> I assumed the performance impact of a redis lookup would be relatively small.

While a single lookup is negligible N lookups accumulate and can introduce non negligible latency.

arikfr:
Hi,

Thanks.

jezdez:
Hey @arikfr is this something you're still interested in? We're doing a review of our fork commits and this came up again.

PR #3410, opened 2019-02-06
Pre-commit hook to run on  modified files only

ranbena:
### Why
Following [this discussion](https://github.com/getredash/redash/pull/3390) I understood our pre-push was running lint on all js files even if none were modified in pushed commits. It's wasteful.

### How
Apparently, it's very common to use [lint-staged](https://www.npmjs.com/package/lint-staged) to lint only matching modified files with a pre-commit hook.

### Example
Here you can see this commit had no js files so pre-commit hook skipped.
Then pre-push ran tests only.

screen shot 2019-02-07 at 1 00 32

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "analyze": "npm run clean && BUNDLE_ANALYZER=on webpack",
     "analyze:build": "npm run clean && NODE_ENV=production BUNDLE_ANALYZER=on webpack",
     "lint": "eslint --config ./client/.eslintrc.js --ext .js --ext .jsx ./client/app",
+    "lint-staged": "eslint --config ./client/.eslintrc.js",
     "test": "TZ=Asia/Jerusalem jest",
     "test:watch": "jest --watch",
     "cypress:install": "npm install --no-save cypress@^3.1.5 @percy/cypress@^0.2.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -152,7 +154,11 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "echo 'running lint...' && npm run lint && echo 'running tests...' && npm run test"
+      "pre-commit": "lint-staged",
+      "pre-push": "echo 'running tests...' && npm run test"
     }
+  },
+  "lint-staged": {
+    "client/app/**/*.{js,jsx}": ["npm run lint-staged", "git add"]
```
kravets-levko: Not sure if `git add` is needed - `eslint` will not change anything without `--fix` option

ranbena: Right. Fixed.

ranbena:
@arikfr @kravets-levko wdyt of testing only related files on commit? ([reference](https://benmccormick.org/2017/02/26/running-jest-tests-before-each-git-commit/))

screen shot 2019-02-07 at 10 06 43

jezdez:
@ranbena Nice work around! 👍

ranbena:
@arikfr 🛎

arikfr:
I've merged this as it's better than what we currently do, but I think we should make a few changes. I will open an issue to track & discuss those.

PR #3411, opened 2019-02-07
Migrate Group List and Details pages to React

kravets-levko:
- [x] Groups List page
  - [x] list
  - [x] additional actions
  - [x] add group
- [x] Group Details page
  - [x] edit group name
  - [x] delete group
  - [x] Group members
    - [x] list and navigation
    - [x] add member
    - [x] remove member
  - [x] Group data sources
    - [x] list and navigation
    - [x] add data source
    - [x] remove data source
    - [x] change permissions
- [x] Cleanup

image

image

image

image

image

image

image

image


arikfr:
I think it shows spinner forever, when there is an empty list. For example:
https://deploy-preview-3411--redash-preview.netlify.com/groups/1/data_sources

client/app/pages/groups/GroupsList.jsx
```diff
@@ -0,0 +1,186 @@
...
+        {(group.type === 'builtin') && <span className="label label-default m-l-10">built-in</span>}
+      </div>
+    ), {
+      field: 'name',
+      width: null,
+    }),
+    Columns.custom((text, group) => (
+      <Button.Group>
+        <Button href={`groups/${group.id}`} onClick={e => e.stopPropagation()}>Members</Button>
+        <Button href={`groups/${group.id}/data_sources`} onClick={e => e.stopPropagation()}>Data Sources</Button>
```
gabrieldutra: Quick thing I found when testing features: This button should only appear to admins in order to avoid this:

403-datasources

kravets-levko: @gabrieldutra You're right, but IMO it feels broken: non-admin can see other group members, why they cannot see datasources they have access to (but cannot add/delete, of course)? I need to think about it :thinking:

ranbena:
Great initiative @kravets-levko. The UX makes much more sense.

A few comments:
1. Create new group -> "Sorry, we couldn't find anything.". Looks like you did something wrong. Perhaps Instead suggest do add a new member, send to [docs](https://redash.io/help/user-guide/users/creating-editing-groups) or if there's no datasource, suggest to add one first.
2. Add members -> search term without results -> "Search results will appear here". Confusing - user might think to wait for results to appear. Also, it's sounds like an unintentional unimplemented placeholder. Some kind of "no results" alternative would be better.
3. Add members -> Loading indicator has outline. I think it would be better without.
4. Already added members - I don't think there's a need for a tooltip + icon here. The convention usually is a disabled row with an "added" (checked) icon.
5. Perhaps rename "Users to be added" to "New members" and consider disabling the "save" button when list is empty.
6. Clicking "add data sources" opens dialog but closes immediately with the error in console "Cannot read property '5' of undefined"
7. Have you considered moving the "add members" and "add data sources" buttons to top of lists (like "add group")?
8. Apparently group name is not unique. Happens in prod. Do you think this is an issue?

kravets-levko:
@ranbena 

1. :+1: 
2. "> Also, it's sounds like an unintentional unimplemented placeholder" - because it is :smile: and that's one of reasons why PR is still WIP
3. can you show me a screenshot? not sure I understand what you mean.
4. :+1: 
5. :+1: 
6. it's backend issue. I'll add a check for this, but preview instance will not mark data sources that are already in group (you can see my changes to the only `.py`-file in this PR)
7. I'll try :+1: 
8. I think it may be a bit confusing for users, but it's not a subject of this PR.

ranbena:
> 3. can you show me a screenshot? not sure I understand what you mean.

screen shot 2019-02-20 at 17 46 03

ranbena:
I think it's because BigMessage gets a "tiled" classname by default. Need to override with an empty classname.

kravets-levko:
@ranbena Thanks! I'm going to get rid of it anyway

kravets-levko:
@ranbena 

> 7. Have you considered moving the "add members" and "add data sources" buttons to top of lists (like "add group")?

So, I tried it, and it doesn't feel good IMO. On Groups page we don't have a sidebar, but also we don't have a header - so "New group" button looks like it's on the right place. On the other hand, on Members/Datasources pages we have a sidebar, and that "actions area" below it (with both "Add ***" and "Remove Group") also looks more or less on its place. But also there is a title - group name - which is editable. So putting one button next to the header (and keep second under the sidebar) looks really weird. Also I tried to put both buttons next to the header, but on the far end - and it also doesn't feel right. So I decided to keep it as is.

client/app/pages/groups/GroupsList.jsx
kravets-levko: Decided to keep it as is (non-admins cannot see group datasources).

ranbena:
@kravets-levko I experimented with different locations as well but couldn't find a clear winner either.
Would you consider making it a primary type button with a "+" icon to keep the convention?
screen shot 2019-02-20 at 22 48 28

kravets-levko:
@ranbena 

> Would you consider making it a primary type button with a "+" icon to keep the convention?

Yes, that's great idea :+1:

ranbena:
One suggestion regarding ux - the "already selected" items are identical to newly selected, which can seem as if they're clickable. Wdyt of lowering their opacity to `0.4`?
screen shot 2019-02-21 at 14 26 08

kravets-levko:
@ranbena Done

client/app/pages/groups/GroupDataSources.jsx
```diff
@@ -0,0 +1,262 @@
...
+    SelectItemsDialog.showModal({
+      dialogTitle: 'Add Data Sources',
+      inputPlaceholder: 'Search data sources...',
+      selectedItemsTitle: 'New Data Sources',
+      searchItems: (searchTerm) => {
+        searchTerm = searchTerm.toLowerCase();
+        return allDataSources.then(items => filter(items, ds => ds.name.toLowerCase().includes(searchTerm)));
+      },
+      renderItem: (item, { isSelected }) => {
+        const alreadyInGroup = isBoolean((item.groups || {})[this.groupId]);
```
ranbena: This might have a bug. Already selected data sources are still selectable (results in Internal Server Error).

client/app/pages/groups/GroupDataSources.jsx
```diff
@@ -0,0 +1,262 @@
...
+import { DataSource } from '@/services/data-source';
+import navigateTo from '@/services/navigateTo';
+import { routesToAngularRoutes } from '@/lib/utils';
+
+class GroupDataSources extends React.Component {
+  static propTypes = {
+    controller: ControllerType.isRequired,
+  };
+
+  groupId = parseInt(this.props.controller.params.groupId, 10);
```
ranbena: This needs to be an integer for `Group.get`?

ranbena: Also, I didn't know it was safe to read props like this! 😮😅

client/app/pages/groups/GroupDataSources.jsx
```diff
@@ -0,0 +1,262 @@
...
+
+  onTableRowClick = (event, item) => navigateTo('data_sources/' + item.id);
+
+  removeGroupDataSource = (event, datasource) => {
+    // prevent default click action on table rows
+    event.preventDefault();
+    event.stopPropagation();
+
+    $http.delete(`api/groups/${this.group.id}/data_sources/${datasource.id}`).success(() => {
+      this.props.controller.update();
```
ranbena: Can use `services/group` here instead?
Also, add toastr.error on fail?

client/app/components/PreviewCard.jsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// PreviewCard
+
+export function PreviewCard({ imageUrl, title, body, children, className, ...props }) {
+  return (
+    <div {...props} className={className + ' w-100 d-flex align-items-center'}>
+      <img src={imageUrl} height="32" className="profile__image--settings m-r-5" alt="Logo/Avatar" />
```
ranbena: Add `width="32"` or there will be a horizontal "jump" when image has loaded (happened to me).

client/app/components/groups/CreateGroupDialog.jsx
```diff
@@ -0,0 +1,44 @@
...
+    const buttonTitle = isNewGroup ? 'Create' : 'Save';
+    return (
+      <Modal {...dialog.props} title={title} okText={buttonTitle} onOk={() => this.save()}>
+        <Input
+          className="form-control"
+          defaultValue={this.state.name}
+          onChange={event => this.setState({ name: event.target.value })}
+          placeholder="Group Name"
+          autoFocus
+        />
```
ranbena: Perhaps put in `<form>` so "enter" key works.

client/app/components/groups/CreateGroupDialog.jsx
```diff
@@ -0,0 +1,44 @@
...
+    const { name } = this.state;
+
+    group.name = name;
+    dialog.close(group);
+  }
+
+  render() {
+    const { dialog, group } = this.props;
+    const isNewGroup = !group.id;
+    const title = isNewGroup ? 'Create a New Group' : 'Edit Group';
```
ranbena: I assume this is leftover code?

client/app/components/groups/DeleteGroupButton.jsx
```diff
@@ -0,0 +1,46 @@
...
+  // prevent default click action on table rows
+  event.preventDefault();
+  event.stopPropagation();
+
+  Modal.confirm({
+    title: 'Delete Group',
+    content: 'Are you sure you want to delete this group?',
+    okText: 'Yes',
+    okType: 'danger',
+    cancelText: 'No',
```
ranbena: Looks like the ramblings of a mad man 😆
"Delete Group! Are you sure? Yes! Danger! No!"

client/app/components/items-list/components/ItemsTable.jsx
```diff
@@ -136,10 +139,13 @@ export default class ItemsTable extends React.Component {
       })
     ) : null;
 
+    const { showHeader } = this.props;
+
     return (
       <Table
-        className="table-data"
+        className={classNames('table-data', { 'ant-table-headerless': !showHeader })}
```
ranbena: Hmmm.. I think that's risky to rely on Ant's implementation like this.

client/app/pages/groups/GroupDataSources.jsx
```diff
@@ -0,0 +1,262 @@
...
+  render() {
+    const { controller } = this.props;
+    const sidebar = (
+      <Sidebar
+        controller={controller}
+        group={this.group}
+        items={this.sidebarMenu}
+        canAddDataSources={currentUser.isAdmin}
+        onAddDataSourcesClick={this.addDataSources}
+        onGroupDeleted={() => navigateTo('/groups', true)}
```
ranbena: Nice trick :)

redash/handlers/data_sources.py
```diff
@@ -84,13 +84,15 @@ def get(self):
...
+        all_fields = request.args.get('extended', None) is not None
+
         response = {}
         for ds in data_sources:
             if ds.id in response:
                 continue
 
             try:
-                d = ds.to_dict()
+                d = ds.to_dict(all=all_fields)
```
arikfr: This is to get the groups information?

client/app/pages/groups/GroupDataSources.jsx
kravets-levko: That's what I mentioned earlier - https://github.com/getredash/redash/pull/3411#issuecomment-465630332 (6)

client/app/pages/groups/GroupDataSources.jsx
kravets-levko: I needed it to be integer in another controller (https://github.com/getredash/redash/pull/3411/files#diff-466f2851617caa71f7d6f5088cc8eaefR101), and here it's also converted because of copy-paste. But I think it's not an issue.

> Also, I didn't know it was safe to read props like this!

No sure why, but it works even if there is a constructor :thinking:

redash/handlers/data_sources.py
kravets-levko: Yes. I decided to add a parameter so it doesn't break any existing code (if any).

client/app/components/items-list/components/ItemsTable.jsx
kravets-levko: @ranbena what do you mean?

ranbena: Use the built in `<Table showHeader={false} />` prop instead.

kravets-levko: @ranbena  I already use it (see few lines below). `ant-table-headerless` is a custom class - to add top border line for first table row when header is invisible (by default there is no border) - https://github.com/getredash/redash/pull/3411/files#diff-8afda93fc5d4b6826361faf186019212R187

ranbena: So is it an Ant bug? Sounds like "ant-table-headerless" should be applied by default when `showHeader={false}`.

The risk is that the ant classnames are internal for Ant, which means they can change with versions. We can't rely on them.

kravets-levko: That's not Ant class - it's custom (also defined in this PR - see the link). I decided to prefix it with `ant-table-` because it's applicable only to Ant's tables, that's it.

ranbena: I'm sorry man, my mistake🙇‍♂️ Good job!

kravets-levko: NP :+1:

kravets-levko:
@ranbena @arikfr Updated PR, could you please review it again? Thanks!

redash/handlers/data_sources.py
arikfr: You should be careful when doing such changes -- in this case, you're now exposing all data source settings to all users. Previously we were showing this only to admins. While in theory all secrets are masked, in practice for some data sources not all secrets are masked.

You can add another `include_groups` option, and use it here to add just the groups.

But why do you need this in the first place?

kravets-levko: @arikfr 

> But why do you need this in the first place?

I need it in "Add data sources" dialog - to mark items that are already added (similar as in "Add members". I'll re-implement it - will add just a list of groups as array.

arikfr: But you already have the list of data sources in this group on the page that opens this modal, can't you pass this information to it? 

While you can update the API to include groups, it's just extra queries that we don't really need.

kravets-levko: @arikfr How about this? https://github.com/getredash/redash/pull/3411/commits/92c57e9058894033af177099306e4c4d4857740e

kravets-levko: @arikfr Reverted all changes to backend, you're right - I totally forgot that I already have all items :man_facepalming: 

https://github.com/getredash/redash/blob/fd0659d991f05adb27a2253b8a654cabfdde3d2e/client/app/pages/groups/GroupDataSources.jsx#L126-L128

kravets-levko: Implemented it in the same way for Add members dialog too

gabrieldutra:
The resource `/api/groups/1/data_sources` is responding with 500 :thinking:

This causes [this page](https://deploy-preview-3411--redash-preview.netlify.com/groups/1/data_sources) to fail:

screenshot from 2019-02-21 19-19-16

My previous comment was probably related to the same thing, but with a wrong guess from me...

As it's a backend error, it also happens in https://redash-preview.netlify.com/groups/1/data_sources, the difference is that in there this is treated somehow...

kravets-levko:
Yeah, that's strange - because it works for other groups, fails only for group 1 :thinking: @arikfr - do you have any ideas?

arikfr:
> Yeah, that's strange - because it works for other groups, fails only for group 1 🤔 @arikfr - do you have any ideas?

The same data source was associated twice with the same group. This caused the API to error.

arikfr:
(fixed now)

kravets-levko:
@arikfr Thanks! Isn't it something that backend should check (probably, DB constraints, etc.)?

kravets-levko:
I reproduced it on https://redash-preview.netlify.com/ - there is a way to reproduce it with almost any UI restrictions. Should I create an issue for this bug?

client/app/components/groups/CreateGroupDialog.jsx
gabrieldutra: `onPressEnter` prop in Ant Input would another option too :) (just sharing the existence here haha)

client/app/pages/groups/GroupsList.jsx
```diff
@@ -0,0 +1,171 @@
...
+          className="w-100"
+          disabled={!canRemove}
+          group={group}
+          onClick={() => this.onGroupDeleted()}
+        >
+          Delete
+        </DeleteGroupButton>
+      );
+      return canRemove ? button : (
+        <Tooltip placement="top" title="Cannot delete built-in group" mouseLeaveDelay={0}>{button}</Tooltip>
```
gabrieldutra: Disabled buttons don't seem to capture `mouseLeave` events very well :confused:. The only way to close this tooltip is unhovering the tooltip itself. Perhaps wrap the button with something else (like a `<div>`) fixes it?

tooltip-disabled-button

kravets-levko: @gabrieldutra Thanks! It wasn't working because `Tooltip` expects a `Button` as direct child, but I created a custom component. Fixed now (just wait for build)

kravets-levko:
This picture perfectly describes my plans for this PR:

image

PR #3413, opened 2019-02-07
[Widget Params] Title edit fixes

ranbena:
Some improvements to UI and fixes to functionality after feedback from @susodapop on #3332.

### Changes
1. Limited title editing to 100 chars (arbitrary, lmk if should change)
2. Static value title editing disabled and a "hidden"  icon appears with an explanation in tooltip.
3. The same for a mapped dashboard param, with a "link" icon.

ezgif com-video-to-gif

### DIFF
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -484,6 +484,7 @@ class Title extends React.Component {
 
     return (
       <Tooltip
+        placement="right"
         title={content}
         getPopupContainer={this.props.getContainerElement}
       >
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -372,10 +372,15 @@ class EditMapping extends React.Component {
 
 class Title extends React.Component {
   static propTypes = {
+    existingParams: PropTypes.arrayOf(PropTypes.object),
     mapping: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
     onChange: PropTypes.func.isRequired,
     getContainerElement: PropTypes.func.isRequired,
-  };
+  }
+
+  static defaultProps = {
+    existingParams: [],
+  }
 
   state = {
     showPopup: false,
@@ -426,15 +431,26 @@ class Title extends React.Component {
     }
 
     // should not be the dashboard source
-    if (IS_DASHBOARD_PARAM_SOURCE in mapping.param) {
-      return false;
-    }
-
-    return true;
+    return !mapping.param[IS_DASHBOARD_PARAM_SOURCE];
   }
 
   get isTypeStatic() {
     return this.props.mapping.type === MappingType.StaticValue;
+  }
+
+  get titleValue() {
+    let { mapping } = this.props;
+    const { existingParams } = this.props;
+
+    // if mapped to dashboard, find source param and return it's title
+    if (this.isTypeMappedToOther) {
+      const source = find(existingParams, { name: mapping.mapTo });
+      if (source) {
+        mapping = source;
+      }
+    }
+
+    return mapping.title || mapping.param.title;
   }
 
   save = () => {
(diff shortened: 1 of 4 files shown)


client/app/components/ParameterMappingInput.jsx
```diff
@@ -414,24 +417,44 @@ class EditTitle extends React.Component {
...
+  get isTypeMappedToOther() {
+    const { mapping } = this.props;
+
+    // should be type dashboard
+    if (mapping.type !== MappingType.DashboardMapToExisting) {
+      return false;
+    }
+
+    // should not be the dashboard source
+    if (IS_DASHBOARD_PARAM_SOURCE in mapping.param) {
```
ranbena: This is how a dashboard source is determined.

client/app/services/dashboard.js
```diff
@@ -170,6 +172,7 @@ function DashboardService($resource, $http, $location, currentUser, Widget, dash
                 globalParams[mapping.mapTo].name = mapping.mapTo;
                 globalParams[mapping.mapTo].title = mapping.title || param.title;
                 globalParams[mapping.mapTo].locals = [];
+                param[IS_DASHBOARD_PARAM_SOURCE] = true;
```
ranbena: Not ideal. @kravets-levko lmk if there's an alternative.

client/app/services/dashboard.js
```diff
@@ -2,6 +2,8 @@ import _ from 'lodash';
 
 export let Dashboard = null; // eslint-disable-line import/no-mutable-exports
 
+export const IS_DASHBOARD_PARAM_SOURCE = Symbol('dash_source');
```
kravets-levko: Someone should definitely check how it works in IE: IE does not support symbols at all, and due to it's nature symbols cannot be polyfilled.

ranbena: Swapped for normal string

ranbena:
@kravets-levko did the required changes 👍 

I do want to point out that there is one unexpected behavior that makes sense with how parameters work.

**Title hijacking:**
1. Add widgets w1 and then w2 to dashboard.
2. Both define their own dashboard params p_w1 and p_w2.
3. Set unique titles to each param .
4. Change p_w1 to point to p_w2.
5. p_w2 now has p_w1's title 😵

This happens cause dashboard params are determined according to chronological order and since w1 came before w2 - it became the source of p_w2.

I don't think this is a big deal as long as users can still re-edit the name.

arikfr:
@ranbena ,
Sorry for joining late, but what I failed to understand is why the user can't edit the title when mapping to existing parameter? How exactly are they supposed to edit it? Hunt down the original one?

ranbena:
> Hunt down the original one?

Yes, it's not good UX, but it's better than the previous behavior which allowed editing but would "work" only if you were on the dashboard source.
From discussions with @kravets-levko, I understood that this limitation has to do with the virtual nature of the dashboard params.

arikfr:
Why aren't we letting the user update it as they would expect, and just update the needed object ourselves?

arikfr:
If we know that the current widget isn't the owner, it means we know which one is, right?

ranbena:
> If we know that the current widget isn't the owner, it means we know which one is, right?

Afraid not

kravets-levko:
@arikfr (cc @ranbena) It's possible to do this, if widget has access to other widgets on the same dashboard. We had a discussion with Ran, and we both stuck with ideas. But while reading you comment I realized that actually we have access to dashboard from widget! Not from model instance, but in a component - yes, and it's even passed to Edit Mapping dialog:

https://github.com/getredash/redash/blob/23f5dde4887eb6a9baa58214e3ac2c0d787b2e98/client/app/components/dashboards/widget.js#L81-L85

I missed this - my bad :confounded: But how @ranbena could miss that working on that dialog - it's a mystery :sweat_smile:

kravets-levko:
@ranbena The idea is that having all widgets, you can find ones that are mapped to the same dashboard parameter, take corresponding mapping and update it's title when editing it. And save all widgets where parameter mapping title changed (yes, dirty check will be really important here).

ranbena:
@kravets-levko It's not the reference to the dashboard that I was missing - it was a clear indication of which widget holds the parameter source. Also, I preferred not to alter the functionality so as not to risk breakage of code indicated as delicate.

arikfr:
@kravets-levko does the dashboard hold any information about parameters or it's all derived in runtime from widgets?

kravets-levko:
@arikfr - it's all derived in runtime from widgets

When we discussed that with @ranbena, we both concluded that there are two ways to implement this:

1. allow to edit only title of "origin" widgets parameter;
2. when editing title - update all widgets mapped to that parameter.

While #2 is a bit better from UX point, we decided that it's easier to implement #1. From my side, I was thinking that we don't have access to a dashboard in that dialog, and didn't want to add it (I'm trying to remember what was in my mind at that time, but cant :thinking:). Also, #1 was much better in terms of making it working right. At the end, @ranbena implemented a good UI - if I were a user I would use it without any issues.

kravets-levko:
@arikfr FYI getredash/redash#3417

PR #3415, opened 2019-02-08
Refine ItemsList base component (previously LiveItemsList)

kravets-levko:
Remark: please refer to the code in `DashboardList.jsx`/`QueriesList.jsx`/`UsersList.jsx` for usage examples.

What was done:

0. Rename "LiveItemsList" -> "ItemsList"
1. Split monolithic and very coupled logic in "ItemsList" for better maintainability and configurability:
   - pure and independent implementation of pagination and sorting (single responsibility: classes from `lib/pagination/*` do maintain pagination state, sorting, current set of items and fetching data logic, and therefore are uncomfortable to use);
   - extract logic of serializing state from/to URL (and make possible to disable it - see below);
   - improved data fetching logic (more OOP-way, also see below).
2. More structured way to configure "ItemsList": instead of passing a lot of parameters, now it expects two instances:
    - resource: class responsible for fetching data based on its state;
    - state storage: class that "knows" how to serialize/retrieve state.

    Implemented few classes that can be used to support almost everything: API that returns plain items list, API with server-side pagination, it's possible to use static array of items, whatever.

    Also there are two state storages: "null" and URL parameters. First one can be used when no need to sync state with URL. (spoiler: needed for groups list/details pages, and probably some else in future).
3. Removed some "hacky" code (`actions`/`context`);
4. Added error handling.


client/app/components/items-list/classes/ItemsSource.js
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  setState(state) {
+    this._paginator = new Paginator(state);
+    this._sorter = new Sorter(state);
+
+    this._searchTerm = state.searchTerm || '';
+    this._selectedTags = state.selectedTags || [];
+
+    this._savedOrderByField = this._sorter.field;
```
ranbena: I'm not sure, but wouldn't it be better to leave ItemResource out of it and have sorter manage (save/update/use) this property?

kravets-levko: It's the place where sorting logic was coupled with search. This field (and related functionality) are not directly related to sorting, that's some custom logic for a case when user searches something, and we should show results ordered by relevance. That's not something that Sorter should know about. Similar logic - when user selected tag - reset current page: Paginator shouldn't be coupled with tags, etc.

kravets-levko: And another possible question about this - why is this logic here and not in `ItemsList` component: because this class is framework-agnostic, it just encapsulates logic of fetching data and managing related state, but does not know how to render it; `ItemsList` just wraps both this "data source" together with target component (and adds some useful things, like adding route parameters) - we could do it in every list page we have, but this wrapper makes things easier.

client/app/components/items-list/classes/StateStorage.js
```diff
@@ -0,0 +1,60 @@
...
+
+export class UrlStateStorage extends StateStorage {
+  getState() {
+    const defaultState = super.getState();
+    const params = $location.search();
+
+    const searchTerm = params.q || '';
+
+    // in search mode order by should be explicitly specified in url, otherwise use default
+    const defaultOrderBy = searchTerm !== '' ? '' : compileOrderBy(defaultState.orderByField, defaultState.orderByReverse);
```
ranbena: If `searchTerm !== ''` it's gonna anyway be taken from `params.order` so this is redundant, no?

If not, consider running this only when necessary (within the `param.order` condition).

client/app/components/items-list/classes/StateStorage.js
```diff
@@ -0,0 +1,60 @@
...
+
+    const {
+      field: orderByField,
+      reverse: orderByReverse,
+    } = parseOrderBy(params.order || defaultOrderBy);
+
+    return {
+      page: parseInt(params.page, 10) || defaultState.page,
```
ranbena: @kravets-levko wdyt of generally using `Number(str)` instead?

client/app/components/items-list/classes/Sorter.js
```diff
@@ -0,0 +1,71 @@
...
+    compiled = compiled.substring(1);
+  }
+  const field = compiled !== '' ? compiled : null;
+  return { field, reverse };
+}
+
+export default class Sorter {
+  field = null;
+
+  reverse = false;
```
ranbena: I like this syntax. Must we separate with line break? If it's a lint rule I'd remove it..

client/app/components/items-list/classes/Sorter.js
```diff
@@ -0,0 +1,71 @@
...
+    this.reverse = !!value; // cast to boolean
+  }
+
+  constructor({ orderByField, orderByReverse } = {}) {
+    this.setField(orderByField);
+    this.setReverse(orderByReverse);
+  }
+
+  toggleField(field) {
+    if (!isString(field) || (field === '')) {
```
ranbena: TS, WE AWAIT THEE!

client/app/components/items-list/classes/Sorter.js
kravets-levko: It's eslint rule, and sometimes it's useful, sometimes not. I think it's better to disable it locally in this file

client/app/components/items-list/classes/Sorter.js
kravets-levko: we'll anyway need this check (at least `field === ''), even with TS

client/app/components/items-list/classes/ItemsSource.js
```diff
@@ -0,0 +1,247 @@
...
+      processItem = isFunction(processItem) ? processItem : item => item;
+
+      return resource(request).$promise.then(data => processResults(data, processItem, context));
+    };
+  }
+
+  _processResults(data) {
+    if (this._isPlainList) {
+      return super._processResults(data); // from PlainItemsSource
+    }
```
ranbena: Hmm.. something about this doesn't sit well with me.
If the `ResourceItemsSource` extends `PlainItemsSource`, why would it need to be told if it should use inherited methods or not.

Could this be avoided by the implementation using the appropriate `ItemsSource`?
```js
itemsList(
  getRequest(request, { currentPage }) {
    new PlainItemsSource({
      ...
```

client/app/components/items-list/classes/Sorter.js
ranbena: Cause it's being taken from the url I suppose?

client/app/components/items-list/classes/StateStorage.js
kravets-levko: 1. it's not redundant - it's used as default value for `params.order` few lines below;
2. it's running only when necessary (also, see few lines below).

If you mean adding condition on `if (params.order)` - IMO this code looks better (more linear); adding a condition makes sense, but this function is called very rarely (only on page load), so I see no profit from it.

client/app/components/items-list/classes/Sorter.js
kravets-levko: Cause it comes from "outside", and TS will check only data type; but for this class field cannot be empty string as well (does not make sense). Probably it's not a big deal to omit this check at all with TS, but at least now we need it.

client/app/components/items-list/classes/ItemsSource.js
kravets-levko: Good point. I'll re-think this :+1:

client/app/components/items-list/classes/StateStorage.js
kravets-levko: Can't decide right now :thinking: I used to use `parseInt`/`parseFloat` (no particular reason why I prefer it).

ranbena: For our usage it has no real difference in output (except `Number('')` -> `0`, while `parseInt('', 10)` -> `NaN`).
`parseInt` comes with extra features that are irrelevant.

client/app/components/items-list/classes/Sorter.js
ranbena: How about `"exceptAfterSingleLine": true`. It'll enforce it on methods (which will most likely be multiline and leave class props alone.

kravets-levko: Yeah, I think it's perfect :+1: I'll try it here

client/app/components/items-list/classes/StateStorage.js
kravets-levko: Ah, I realized why `parseInt` here :slightly_smiling_face: Because `Number()` constructor is in fact `parseFloat`, so `Number('3.14') === 3.14`, but `parseInt('3.14') === 3`

kravets-levko: And probably that's also the answer why I prefer `parseInt`/`parseFloat` - it's easy to forget that `Number()` will parse string as floating point value, and when using functions you explicitly see what you're doing.

client/app/components/items-list/classes/ItemsSource.js
kravets-levko: @ranbena I was thinking about it for a long time and ended up with merging `ItemsSource` and `PlainItemsSource` classes - there were really minor differences between them (technically they're different enough, but I decided just to add flag). Also it allowed to simplify `ResourceItemsSource` a lot.

kravets-levko: @ranbena Finally decided to implement pattern "strategy" for fetching items and I'm pretty happy with the results :slightly_smiling_face:

ranbena: On it

ranbena:
Btw, Archived Queries [yield error 500](https://deploy-preview-3415--redash-preview.netlify.com/queries/archive?page=1&page_size=20&order=-created_at) (also in `master` but not in `app.redash`)

kravets-levko:
@ranbena 

> Btw, Archived Queries [yield error 500](https://deploy-preview-3415--redash-preview.netlify.com/queries/archive?page=1&page_size=20&order=-created_at) (also in `master` but not in `app.redash`)

It's okay. Preview instances use a bit outdated backend, so there it no API endpoint for archived queries.

PR #3416, opened 2019-02-08
Use Ant's Button component on users list page

kravets-levko:
image

image


client/app/assets/less/ant.less
```diff
@@ -47,6 +47,12 @@
   z-index: 1050;
 }
 
+// Button overrides
+
+.@{btn-prefix-cls} {
+  transition: none;
```
ranbena: What's this for?

Also, might be redundant since this rule overrides it:
```css
.ant-table * {
    transition: none!important;
}
```

ranbena: Yeah the transition is a bit slow.
I propose this instead:
```css
transition-duration: 150ms;
```

kravets-levko: @ranbena  80-90ms feels more comfortable imo. wdyt?

ranbena: For me 100 was too fast, 200ms too slow, 150 sweet spot

ranbena: "too fast" as in meaningless

kravets-levko: Done. With `80ms` animation is still noticeable, but not annoying

kravets-levko: 100ms feels to slow for me :smile:

ranbena:
Oh, about `htmlType="button"` - I think that's the default.

kravets-levko:
@ranbena there was a warning, let me check

kravets-levko:
Ah, no, it's my IDE's inspection. Will remove that. Thanks!

client/app/assets/less/ant.less
kravets-levko: @ranbena  Tried once more, and seems you're right - 150ms is better :thinking: Changed it again

ranbena: perfection1

arikfr: Is it possible there is a difference between MacOS and Linux?

I remember feeling Ant's animation are slow, while for Levko they were fine. Maybe something similar happens here?

kravets-levko: I think it's just different perception

susodapop: There might be something to @arikfr's idea though. I had a similar though about some animations that @ranbena used in his new dashboard parameters UI. Might be interesting to double check.

PR #3417, opened 2019-02-09
Alternative implementation of dashboard param title editing

kravets-levko:
Synchronize titles on editing. 

Pros: 

- more obvious and clear for user.

Cons: 

- O(n^2) algorithm (but with a a lot of checks, so if no a lot of widgets mapped to the same param - it will be very fast);
- saving more that one widget at once (but we do the same on editing layout too, so maybe not a big deal).


client/app/components/ParameterMappingInput.jsx
```diff
@@ -510,13 +537,10 @@ class Title extends React.Component {
   }
 
   render() {
-    // static value and mapped types are non-editable hence disabled
-    const disabled = this.isTypeMappedToOther || this.isTypeStatic;
-
     return (
-      <div className={classNames('title', { disabled })}>
-        <span className="text">{this.titleValue}</span>
-        {disabled ? this.renderTooltip() : this.renderEditButton()}
```
ranbena: Actually, static values should still have the disabled + tooltip look.

ranbena:
@kravets-levko works perfectly. Code looks great.
One comment about static value handling.

client/app/components/ParameterMappingInput.jsx
```diff
@@ -16,8 +16,8 @@ import Form from 'antd/lib/form';
 import Tooltip from 'antd/lib/tooltip';
 import { ParameterValueInput } from '@/components/ParameterValueInput';
 import { ParameterMappingType } from '@/services/widget';
-import { Parameter } from '@/services/query';
-import { IS_DASHBOARD_PARAM_SOURCE } from '@/services/dashboard';
+import { clientConfig } from '@/services/auth';
+import { Query, Parameter } from '@/services/query';
```
ranbena: Oh this is great

client/app/components/ParameterMappingInput.jsx
```diff
@@ -301,15 +316,14 @@ class EditMapping extends React.Component {
     }
 
     this.setState({ mapping, inputError });
-  }
+  };
 
-  get content() {
+  getContent() {
```
ranbena: Why?

kravets-levko: Because when it's property getter - it's less clear to understand that it renders something. Probably I even had to call it `renderContent` (as I did below :thinking: )

client/app/components/ParameterMappingInput.less
```diff
@@ -77,7 +77,7 @@
   }
```
ranbena: This is now unused

PR #3421, opened 2019-02-11
Fix: saving a new query removes reference to last query result

### LINKED ISSUE #3372: When creating a visualisation after running a query the query results are deleted
We are testing the latest stable version of redash (6.0.0+b10818 (7fa66654)
Found a regression when creating a new visualization, to reproduce:
0. Open a new browser tab and log into redash
1. Run a query
2. Create a new visualization
3. The X & Y drop downs are empty
4. When you cancel the visualization creation you see that the query results disappeared

Note - the issue occurs only when you run redash for the first time on a new session

arikfr:
Closes #3372.

### DIFF
--- a/client/app/pages/queries/view.js
+++ b/client/app/pages/queries/view.js
@@ -400,7 +400,7 @@ function QueryViewCtrl(
     }
 
     if (status === 'done') {
-      const ranSelectedQuery = $scope.query.query !== $scope.queryResult.query;
+      const ranSelectedQuery = $scope.query.query !== $scope.queryResult.query_result.query;
       if (!ranSelectedQuery) {
         $scope.query.latest_query_data_id = $scope.queryResult.getId();
         $scope.query.queryResult = $scope.queryResult;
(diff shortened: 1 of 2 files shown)


client/app/pages/queries/view.js
```diff
@@ -400,7 +400,7 @@ function QueryViewCtrl(
     }
 
     if (status === 'done') {
-      const ranSelectedQuery = $scope.query.query !== $scope.queryResult.query;
+      const ranSelectedQuery = $scope.query.query !== $scope.queryResult.query_result.query;
```
arikfr: This is the actual fix.

client/app/pages/queries/view.js
```diff
@@ -445,6 +445,9 @@ function QueryViewCtrl(
       $scope.saveQuery().then((query) => {
         // Because we have a path change, we need to "signal" the next page to
         // open the visualization editor.
+        // TODO: we don't really need this. Just need to assign query to $scope.query
+        // and maybe a few more small changes. Not worth handling this now, but also
+        // we shouldn't copy this bizzare method to the React codebase.
```
arikfr: This is for future generations.

redash/handlers/queries.py
```diff
@@ -228,7 +228,7 @@ def post(self):
             'object_type': 'query'
         })
 
-        return QuerySerializer(query).serialize()
+        return QuerySerializer(query, with_visualizations=True).serialize()
```
arikfr: This will be needed when we implement new query save and open visualization editor the right way.

PR #3424, opened 2019-02-11
Converted Share modal to Ant

ranbena:
Ant modal, switch, InputWithCopy (taken from #3354 😬)

screen shot 2019-02-11 at 22 51 53

Also, added notification on server error.

screen shot 2019-02-11 at 22 41 01

Feedback welcome on UX.


gabrieldutra:
> (taken from #3354 )

:eyes: 

I almost created a separate PR for this `InputWithCopy`, but no problem, what matters it to have it in the code, I'll rebase when this gets merged :stuck_out_tongue_closed_eyes:

ranbena:
@arikfr I played around with the UX of desc info inside a modal.
Wdyt of this?

screen shot 2019-02-12 at 11 25 17

arikfr:
@ranbena couldn't ask for better.

ranbena:
@kravets-levko @gabrieldutra ready for review

client/app/components/InputWithCopy.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+  copy = () => {
+    // select text
+    this.ref.current.select();
+
+    // copy
+    try {
+      const success = document.execCommand('copy');
+      if (!success) {
+        throw new Error();
```
kravets-levko: Why try/catch? :-)

```javascript
const success = document.execCommand('copy');
this.setState({ copied: success ? 'Copied!' : 'Copy failed' });
```

client/app/pages/dashboards/ShareDashboardDialog.jsx
```diff
@@ -0,0 +1,122 @@
...
+    labelCol: { span: 8 },
+    wrapperCol: { span: 16 },
+    style: { marginBottom: 7 },
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      saving: false,
+      apiUrl: replace(API_SHARE_URL, '{id}', props.dashboard.id),
```
kravets-levko: After `supes(props)` call you can use `this.props`

client/app/components/InputWithCopy.jsx
gabrieldutra: My guessing was that `document.execCommand()` could throw Errors on its own, but I'm not sure and I didn't find ath related :thinking:

kravets-levko: [According to docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) it does not throw errors, just returns `boolean` value (which also is meaningless if `document.execCommand()` not called as a part of user interaction, but in this case it's ok)

client/app/pages/dashboards/ShareDashboardDialog.jsx
```diff
@@ -0,0 +1,122 @@
...
+        </div>
+      </React.Fragment>
+    );
+  }
+
+  enable = () => {
+    const { dashboard } = this.props;
+    this.setState({ saving: true });
+
+    $http
```
gabrieldutra: This is more like a question, "how" is it ok to use `$http` inside components?

client/app/pages/dashboards/ShareDashboardDialog.jsx
gabrieldutra: Considering `apiUrl` does not change in the component, wouldn't it be better for it to be a `getter` or a variable instead of a state?

client/app/pages/dashboards/ShareDashboardDialog.jsx
kravets-levko: Not really ok. @ranbena you can add some methods to `services/dashboard`:
```javascript
const resource = $resource(
  // ...
  share: {
    method: 'post',
    url: 'api/dashboards/:id/share',
  },
  unshare: {
    method: 'delete',
    url: 'api/dashboards/:id/share',
  },
);
```
Usage:
```javascript
this.props.dashboard.share()
  .then(...)
  .catch(...)
  .finally(...);
```

client/app/pages/dashboards/ShareDashboardDialog.jsx
kravets-levko: It's even better if it will be in a service (see https://github.com/getredash/redash/pull/3424#discussion_r255914090)

client/app/pages/dashboards/ShareDashboardDialog.jsx
arikfr: While it make sense to have a layer where we implement the interaction with the server and expose a simple API to the rest of application, I don't think we should invest in implementing any more code with $resource...

Let's leave as is, and refactor when we decide on the $resource/$http successor.

client/app/components/InputWithCopy.jsx
ranbena: Yeah a `SecurityError` https://dvcs.w3.org/hg/editing/raw-file/tip/editing.html#the-copy-command

client/app/pages/dashboards/ShareDashboardDialog.jsx
ranbena: @gabrieldutra a getter would execute the replace function each time it is accessed so I opted for either `this.state.apiUrl` or `this.apiUrl`. I'll swap it.

client/app/pages/dashboards/ShareDashboardDialog.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+  static get headerContent() {
+    return (
+      <React.Fragment>
+        Share Dashboard
+        <div className="modal-header-desc">
+          Allow public access to this dashboard with a secret address.{' '}
+          <Tooltip title="Guide: Sharing and Embedding Dashboards">
+            { /* eslint-disable-next-line react/jsx-no-target-blank */}
+            <a href={HELP_URL} target="_blank" rel="noopener">Learn more</a>
```
gabrieldutra: One last question I forgot :sweat_smile:: why not `rel="noopener noreferrer"` to avoid `eslint-disable`? ([ref](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md))

client/app/components/InputWithCopy.jsx
kravets-levko: Hm. There is some conflict in that docs:

image

and then:

image

Cant say who tells the truth, so let's keep try/catch block.

client/app/pages/dashboards/ShareDashboardDialog.jsx
ranbena: Heh heh :) Cause I think it might be valuable for us to track the origin of visits to help pages.
That's what the referrer is for so I don't want to mess up the stats.

ranbena: @kravets-levko @gabrieldutra legit?

gabrieldutra: Yes, the vulnerability seems to be with the lack of `noopener`. Anyway, I think we can trust redash.io website :slightly_smiling_face:

kravets-levko:
:star2:

client/app/pages/dashboards/ShareDashboardDialog.jsx
ranbena: Exactly :)

PR #3426, opened 2019-02-12
Notifications randomly shown with visible page or not shown with inactive page

kravets-levko:
Document visibility wasn't tracking properly


client/app/services/notifications.js
```diff
@@ -11,7 +11,7 @@ if (!Notification) {
 
 const hidden = find(
   ['hidden', 'webkitHidden', 'mozHidden', 'msHidden'],
-  prop => prop in document,
+  prop => ('on' + prop.toLowerCase()) in document,
```
ranbena: I think `hidden` is ok but `visibilitychange` should be prefixed with `on`. No?

ranbena: Oh scratch that - it's used with addEventListener so no need for `on`.
Still, don't see why `onhidden`.

kravets-levko: Because `document.onhidden`, but `document.addEventListener('hidden', ...)` - `addEventListener` expects event name without `on` prefix

ranbena: But `const hidden` is not used in an event listener.

kravets-levko: that's also bug :man_facepalming: wrong place

PR #3427, opened 2019-02-12
Query Parameters: don't save urlPrefix

kravets-levko:
subj.


client/app/pages/queries/view.js
```diff
@@ -249,6 +249,9 @@ function QueryViewCtrl(
       $scope.saveQuery(options, data);
     }
 
+    request.options = extend({}, request.options);
+    request.options.parameters = map(request.options.parameters, param => omit(param, 'urlPrefix'));
```
arikfr: Won't this add parameters where it wasn't there before?

kravets-levko: @arikfr Added new commit - with another approach

PR #3428, opened 2019-02-13
[Widget Params] Updated help url and tooltip

ranbena:
An update to the exact help section url once https://github.com/getredash/website/pull/195 lands.

### What changed
* Help url updated to query-parameters#Value-Source-Options ([preview on Netlify](https://deploy-preview-195--redashio.netlify.com/help/user-guide/querying/query-parameters#Value-Source-Options))
* Tooltip text updated to a conventional `Guide: {help title}`.

### DIFF
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/no-multi-comp */
 
-import { isString, extend, each, map, includes, findIndex, find, fromPairs, clone, isEmpty, replace } from 'lodash';
+import { isString, extend, each, map, includes, findIndex, find, fromPairs, clone, isEmpty } from 'lodash';
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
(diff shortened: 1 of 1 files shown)


client/app/components/ParameterMappingInput.jsx
```diff
@@ -336,15 +339,15 @@ class MappingEditor extends React.Component {
...
+    const [helpUrl, tooltip] = HELP_URL;
 
     return (
       <div className="parameter-mapping-editor">
         <header>
           Edit Source and Value
           {/* eslint-disable-next-line react/jsx-no-target-blank */}
           <a href={helpUrl} target="_blank" rel="noopener">
-            <Tooltip title="Learn more about editing query paramaters (opens in a new window)">
+            <Tooltip title={tooltip}>
```
arikfr: Just a thought: maybe we should have a `<HelpTrigger>` component, that currently will: render circle icon with a tooltip. We will use it in all the places we add help content to. But later on allow us in one single swoop change the implementation to open a [drawer](https://ant.design/components/drawer/) with the help content inside the app?

(not a blocker of course, I'm waiting for the relevant docs PR to merge)

ranbena: I like the drawer idea. Will the content be retrieved with github api by help url?

arikfr: No, I thought we will update the Gatsby configuration to create a "drawer friendly" view of the help content alongside the regular one. The drawer pages will have the same URL structure, so we can deduce the new URL based on the regular one (so Cmd/Shift+Click will still work).

PR #3430, opened 2019-02-13
send ip and user id to sentry

rauchy:


jezdez:
@arikfr @rauchy Could this also be made a setting?

jezdez:
(I'm happy to write the patch)

arikfr:
@jezdez ,

> Could this also be made a setting?

Sure. Although I think that sending the user id shouldn't be a setting. Just the IP address.

jezdez:
> @jezdez ,
> 
> > Could this also be made a setting?
> 
> Sure. Although I think that sending the user id shouldn't be a setting. Just the IP address.

Yeah, the IP is critical in this case, the user ID less so.

PR #3431, opened 2019-02-13
HelpTrigger

ranbena:
Followup on https://github.com/getredash/redash/pull/3428#pullrequestreview-203050033

Usage:
```jsx
<HelpTrigger type="VALUE_SOURCE_OPTIONS" />
```

Btw, the tooltip "Guide: " prefix is one that GitHub uses and I like it.

### DIFF
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -18,15 +18,11 @@ import { ParameterValueInput } from '@/components/ParameterValueInput';
 import { ParameterMappingType } from '@/services/widget';
 import { clientConfig } from '@/services/auth';
 import { Query, Parameter } from '@/services/query';
+import HelpTrigger from '@/services/HelpTrigger';
 
 import './ParameterMappingInput.less';
 
 const { Option } = Select;
-
-const HELP_URL = [
-  'https://redash.io/help/user-guide/querying/query-parameters?source=dialog#Value-Source-Options',
-  'Guide: Value Source Options',
-];
 
 export const MappingType = {
   DashboardAddNew: 'dashboard-add-new',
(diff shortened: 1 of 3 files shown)


client/app/components/ParameterMappingInput.jsx
```diff
@@ -339,18 +335,11 @@ class MappingEditor extends React.Component {
...
       <div className="parameter-mapping-editor">
         <header>
-          Edit Source and Value
-          {/* eslint-disable-next-line react/jsx-no-target-blank */}
-          <a href={helpUrl} target="_blank" rel="noopener">
-            <Tooltip title={tooltip}>
-              <Icon type="question-circle" />
-            </Tooltip>
-          </a>
+          Edit Source and Value <HelpTrigger type="VALUE_SOURCE_OPTIONS" />
```
arikfr: Why not pass the title and URL here? I'm pretty sure we will rarely reuse the link helps, so having them in a single place and then referenced using some random string is just recipe for trouble.

ranbena: Not meant for reuse, but so that if help urls change, there's a single source to edit.
This is a "preparation for A/C" for when TS is integrated as it would enforce the string to be of type.

👍  if you're cool with it, 👎 if you'd rather go with your suggestion

arikfr: You mean that TS will enforce that the string is one of possible values?

ranbena: Yep

arikfr: TS support can't come any sooner... 😓 

I wonder if we should go through the effort of introducing it now, or wait to get rid of Angular and then switch to `react-scripts` and let them take care of the plumbing...

ranbena: I really dunno to tell you the amount of effort involved.
If it's truly possible to gradually adopt - let's do this now.
If not - it can wait.

The focus for the coming 4 weeks should be Angular->React migration completion.

PR #3436, opened 2019-02-14
HelpTrigger to open in drawer

ranbena:
HelpTrigger opened docs in new window - now opens in drawer.

ezgif com-video-to-gif 3

The catch is that for the help docs to look fine in the drawer - must add "drawer" classname to it's body. This is solved with parent<->iframe communication.

### The flow:
1. Iframe starts with "about: blank" to avoid animation jank while drawer opens.
2. Iframe is set to the help url, and communication starts.
  * Listens to `drawer_ready, then allows the iframe to show, and removes all listeners.
  * Waits for iframe to load, then messages the iframe with `drawer_request`.
3. If 5 seconds went by and `drawer_request` didn't arrive, show error msg (allows user to navigate to doc in new window, the url has special query param so we could know sth wrong).

screen shot 2019-02-14 at 13 46 58

### Notice
The page loads with a CSS flicker. That's not related to the iframe. It happens in prod standalone.
I can make the page show only after `iframe.onload` - but I think this should be addressed within Gatsby instead.

### DIFF
--- a/client/app/services/HelpTrigger.jsx
+++ b/client/app/services/HelpTrigger.jsx
@@ -1,9 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
 import Tooltip from 'antd/lib/tooltip';
 import Icon from 'antd/lib/icon';
+import Drawer from 'antd/lib/drawer';
+import { BigMessage } from '@/components/BigMessage';
 
-const BASE_URL = 'https://redash.io/help/user-guide/';
+import './HelpTrigger.less';
+
+const DOMAIN = 'https://redash.io';
+const HELP_PATH = '/help/user-guide/';
+const MESSAGES = {
+  HELP_DRAWER_READY: 'drawer_ready',
+  HELP_DRAWER_REQUEST: 'drawer_request',
+};
+const IFRAME_TIMEOUT = 5000;
 const TYPES = {
   VALUE_SOURCE_OPTIONS: [
     'querying/query-parameters#Value-Source-Options',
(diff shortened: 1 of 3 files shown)


client/app/services/HelpTrigger.jsx
```diff
@@ -20,16 +31,121 @@ export default class HelpTrigger extends React.PureComponent {
...
+
+  onPostMessageReceived = (event) => {
+    if (event.origin === DOMAIN && event.data === MESSAGES.HELP_DRAWER_READY) {
+      this.onIframeReady();
+    }
+  }
+
+  openDrawer = () => {
+    this.setState({ visible: true, loading: true, error: false });
+    setTimeout(this.onDrawerOpened, 300); // to prevent animation jank
```
ranbena: Yes, this is not ideal - but the simplest (and as risky as any other simple solution I could conceive). Open to suggestions.

client/app/services/HelpTrigger.jsx
```diff
@@ -20,16 +31,121 @@ export default class HelpTrigger extends React.PureComponent {
...
+  }
+
+  get helpUrl() {
+    const [pagePath] = TYPES[this.props.type];
+    return DOMAIN + HELP_PATH + pagePath;
+  }
+
+  onIframeLoaded = () => {
+    const { contentWindow } = this.iframeRef.current;
+    contentWindow.postMessage(MESSAGES.HELP_DRAWER_REQUEST, DOMAIN);
```
ranbena: The weird thing about this is that if the window doesn't match the domain (try loading a different url in the iframe), it'll fire an UNCATCHABLE error!

client/app/services/HelpTrigger.jsx
```diff
@@ -20,16 +31,121 @@ export default class HelpTrigger extends React.PureComponent {
...
+  }
+
+  state = {
+    visible: false,
+    loading: false,
+    error: false,
+  }
+
+  componentWillUnmount() {
+    this.onDrawerClosed();
```
gabrieldutra: This method is supposed to exist? :thinking:

ranbena: lol used to. Fixed now. TS WE AWAIT THEE!!

client/app/services/HelpTrigger.jsx
ranbena: I'll create a PR for Ant with `onAfterOpen` perhaps.

client/app/services/HelpTrigger.jsx
```diff
@@ -20,16 +31,121 @@ export default class HelpTrigger extends React.PureComponent {
...
+  removeIframeListeners = () => {
+    window.removeEventListener('message', this.onPostMessageReceived);
+    this.iframeRef.current.removeEventListener('load', this.onIframeLoaded);
+    window.clearTimeout(this.iframeLoadingTimeout);
+  }
+
   render() {
-    const [path, tooltip] = TYPES[this.props.type];
-    const href = BASE_URL + path;
+    const [, tooltip] = TYPES[this.props.type];
```
gabrieldutra: ES6 has some pretty cool tricks :+1:

ranbena:
> I don't remember if you've already commented about it somewhere (I got the feeling you have 🤔), but the behavior for `Esc` button is closing both the modal and the drawer (the modal first). Perhaps giving focus to the drawer solves it.

Yeah, in the latest convo's I had with AntD, there's no solution to this. I can elaborate offline if you want.

client/app/services/HelpTrigger.jsx
```diff
@@ -20,16 +31,121 @@ export default class HelpTrigger extends React.PureComponent {
...
+
+  addIframeListeners = () => {
+    window.addEventListener('message', this.onPostMessageReceived, DOMAIN);
+    this.iframeRef.current.addEventListener('load', this.onIframeLoaded);
+    this.iframeLoadingTimeout = setTimeout(this.onIframeError, IFRAME_TIMEOUT);
+  }
+
+  removeIframeListeners = () => {
+    window.removeEventListener('message', this.onPostMessageReceived);
+    this.iframeRef.current.removeEventListener('load', this.onIframeLoaded);
```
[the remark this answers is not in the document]
ranbena: Fixed in latest update 👍

ranbena:
Reverted the postMessaging. No need for it really. Worst-case if the header flickers, will put a minor wait before unhiding the iframe.

gabrieldutra:
> Yeah, in the latest convos I had with AntD, I realized there's no solution to this. I can elaborate offline if you want.

Is there a way to give focus to the Drawer? This way at least `Esc` does nothing instead of closing everything in case I didn't focus it myself...

ranbena:
> > Yeah, in the latest convos I had with AntD, I realized there's no solution to this. I can elaborate offline if you want.
> 
> Is there a way to give focus to the Drawer? This way at least `Esc` does nothing instead of closing everything in case I didn't focus it myself...

Might work if you add a `tabindex` attribute..

client/app/services/HelpTrigger.jsx
```diff
@@ -1,9 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
```
arikfr: Why `cx`?

client/app/services/HelpTrigger.jsx
```diff
@@ -20,16 +27,93 @@ export default class HelpTrigger extends React.PureComponent {
...
+            <BigMessage icon="fa-spinner fa-2x fa-pulse" message="Loading..." className="help-message" />
+          )}
+
+          {/* error message */}
+          {this.state.error && (
+            <BigMessage icon="fa-exclamation-circle" className="help-message">
+              Something went wrong.<br />
+              {/* eslint-disable-next-line react/jsx-no-target-blank */}
+              <a href={this.helpUrl + '?s=help_error'} target="_blank" rel="noopener">Click here</a>{' '}
+              to open the document in a new window.
```
arikfr: Maybe "article" or "page" instead of "document"?

(cc: @susodapop)

client/app/services/HelpTrigger.jsx
ranbena: I don't know actually but it's what we used in FB and also what everyone else uses - including the official npm library. I like it cause it's short.

arikfr: 👍

arikfr:
Should I hit the Merge button?

PR #3439, opened 2019-02-14
Sharing disabled if dashboard has query params

ranbena:
Following discussion in https://discuss.redash.io/t/ux-for-dashboard-sharing-constraints/3212, disabling sharing if dashboard has query params.

screen shot 2019-02-14 at 19 51 12

For offending dashboards that are already shared, the alert would appear but the access would still be enabled both in backend and UI. Once switched off and closed, the UI won't allow re-enabling until dashboard clear of query param widgets.

### DIFF
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -392,7 +392,11 @@ function DashboardCtrl(
   }
 
   this.openShareForm = () => {
-    const hasQueryParams = _.some(this.dashboard.widgets, w => !_.isEmpty(w.options.parameterMappings));
+    // check if any of the wigets have query parameters
+    const hasQueryParams = _.some(
+      this.dashboard.widgets,
+      w => !_.isEmpty(w.getQuery() && w.getQuery().getParametersDefs()),
+    );
 
     ShareDashboardDialog.showModal({
       dashboard: this.dashboard,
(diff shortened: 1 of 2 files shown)


client/app/pages/dashboards/dashboard.js
```diff
@@ -392,7 +392,12 @@ function DashboardCtrl(
   }
 
   this.openShareForm = () => {
-    ShareDashboardDialog.showModal({ dashboard: this.dashboard });
+    const hasQueryParams = _.some(this.dashboard.widgets, w => !_.isEmpty(w.options.parameterMappings));
```
kravets-levko: TL;DR [A proper way to get widget's query parameters](https://github.com/getredash/redash/blob/master/client/app/services/widget.js#L182)

Explanation: parameter mappings were introduced recently, and on "old" setups this check will fail (widgets there does not have that field, and will not have until someone will edit it). `widget.getParametersDefs()` returns only local parameters for the widget, `dashboard.getParametersDefs()` - only dashboard-level parameters. So the proper way to know if there are any parameters - to get it from widgets query (if exists - textboxes does not have query attached).

ranbena: Fixed @kravets-levko

kravets-levko: Yeah, I see :+1:

PR #3440, opened 2019-02-14
Fixes to Percy

gabrieldutra:
## Description
The general idea here is to give Percy a few fixes. I'm creating a PR so some of it can be discussed and since I needed a fork PR to validate.

## WIP Issues
- [x] a few seconds ago sometimes is a minute ago
- [x] replace `cy.wait()` with sth more reliable


gabrieldutra:
I've tried a bunch of different approaches to fix both issues (a few seconds ago and non-reliability of `cy.wait`). Here are some notes from my attempts:
- For the fix for "a few seconds ago ~ a minute ago" my first try was with `cy.clock(Date.now() + (60 * 60 * 1000))`. However, [weird unexpected result](https://percy.io/Redash/redash/builds/1439995).
- For the `cy.wait` thing I've tried to create a general waiting for all pages, such as checking completion of `pace-progress` and waiting for api resources to be finished. The problems: `pace-progress` somehow doesn't bother to show up when running tests and the api resources were not enough to ensure everything was loaded.

I eventually concluded that separate all page screenshots with render assertions for each page was worth it, despite the cons (+~25 sec in CI, some new spec files with one assertion only). Those cons actually don't bother me that much since we tend to create more tests and that would happen eventually.

I do hope now it's a lot more reliable, but we'll see how this goes.

LMK your thoughts :slightly_smiling_face:

gabrieldutra:
Let's see how this goes :shipit:

PR #3442, opened 2019-02-15
Add custom JSON encoder for PostgreSQL

### LINKED ISSUE #1764: PostgreSQL queries returning a tsrange fail
### Issue Summary

Running a query which includes a tsrange column fail with a JSON error like this:

```
DateTimeRange(datetime.datetime(2016, 8, 25, 15, 21, 45, 760521), datetime.datetime(2016, 10, 14, 5, 56, 52, 452297), '[)') is not JSON serializable
```

### Steps to Reproduce

1. Create a table with a tsrange column
2. Insert some data into it
3. Create a `SELECT * FROM table` query
4. Run the query

This is turning out to be confusing for our users - I'm getting regular complaints about queries failing with this error because some of our tables have range columns.

### Technical details:

* Redash Version: 1.0.1+b2845
* Browser/OS: Linux
* How did you install Redash: Docker image

s22su:
To handle columns with [range types][1] and display them as a
string custom JSON encoder for PostgreSQL was added.

Fixes #1764.

Example of querying table with `tstzrange` field:

redash-tstzrange

[1]:https://www.postgresql.org/docs/9.3/rangetypes.html


redash/query_runner/pg.py
```diff
@@ -28,6 +29,26 @@
...
+
+            items = [
+                o._bounds[0],
+                str(o._lower),
+                ', ',
+                str(o._upper),
+                o._bounds[1]
+            ]
+
+            return ''.join(items)
```
arikfr: Why not call `str(o)`?

s22su: @arikfr Calling `str(o)` doesn't give proper result. My proposed implementation formats range type column values the way they can be used in an SQL query.

Also there is a comment in code that references [psycopg2#779][1], this PR is still not in the master branch of psycopg2. When [psycopg2#779][1] will make it to master then we could update psycopg2 and call `str(o)` and it would give the same result.

[1]: https://github.com/psycopg/psycopg2/pull/779

arikfr: I missed the fact it's not released yet. 👍

arikfr:
Thanks!

Issue #3379, opened 2019-01-31
[Widget Params] Bug: widget oblivious to updated static value


ranbena:
Originally reported by @arikfr in https://github.com/getredash/redash/pull/3334#issuecomment-459269626.

STR:
1. Go to widget with static parameter, click "edit parameters".
2. Change value of parameter, click "save".

**Expected** - Widget auto refreshes with new value.
**Actual** - Widget remains with previous value.

Clicking the widget refresh button updates it. Just need to make it automatic on change.

ranbena:
@kravets-levko is this something you have time to fix?

ranbena:
@arikfr @kravets-levko do you think this would do it? https://github.com/getredash/redash/commit/a843b924cd67fe86dc9d0169559bcf5949f4b1ff

kravets-levko:
Indeed it will work. I'm just not sure if it solves described issue - it will refresh widget on any parameter mappings change, not only static values :thinking:

kravets-levko:
> Makes sense anyway, no?

Probably yes :slightly_smiling_face:

kravets-levko:
@arikfr WDYT?

arikfr:
I think it makes sense -- like the case where you change a mapping from one dashboard param to another.

So as long as it runs when mapping changed, I'm good with this.

ranbena:
Right now it refreshes on save button click. Which means, also param label renaming. Would you like to filter that case out?

Is refresh such an expensive action that we'd rather avoid any needless calls to it?

arikfr:
> Is refresh such an expensive action that we'd rather avoid any needless calls to it?

Sometimes it is, sometimes it isn't...

If it's easy to call it only when really needed, let's do it. Otherwise, let's fix the bug.

susodapop:
> In this pr it refreshes on save button click.

In my experience this could be a real dealbreaker for trying out the functionality. Some of my dashboards can take north of a minute to update. It feels weird to make changes to the **title** of a parameter and have it refresh the relevant widgets. Refresh should only happen when absolutely necessary IMO.

ranbena:
Made it so only parameter value changes trigger an automatic refresh.
Title changes won't, even param type change won't.
https://github.com/getredash/redash/pull/3445

PR #3445, opened 2019-02-16
Fix: widget oblivious to updated parameter values

ranbena:
Fixes https://github.com/getredash/redash/issues/3379.

### The solution
Comparing a signature made from parameter values - if not equal before and after edit - refresh widget.

This means that title changes won't trigger refresh. Even param type change won't if it's value stayed the same.

### In action

ezgif com-video-to-gif 1


client/app/components/dashboards/widget.js
```diff
@@ -82,8 +82,13 @@ function DashboardWidgetCtrl($scope, $location, $uibModal, $window, $rootScope,
     EditParameterMappingsDialog.showModal({
       dashboard: this.dashboard,
       widget: this.widget,
-    }).result.then(() => {
+    }).result.then((valuesChanged) => {
       this.localParameters = null;
+
+      // refresh widget if any parameter value has been updated
+      if (valuesChanged) {
+        setTimeout(() => this.refresh(), 0); // must tick for refresh to work
```
ranbena: @kravets-levko do you know why this is needed here?

kravets-levko: @ranbena nope. @arikfr - do you know something about this?

kravets-levko: @ranbena (I removed my previous comment because I realized why you need it). You need to wait for Angular `$digest`  to apply new parameters. And you have to use `$timeout` service here (just replace `setTimeout` with it) - it will enforce `$digest` cycle before executing callback (`setTimeout` will not, and sometimes widget will not refresh as well)

ranbena: Done

client/app/components/dashboards/EditParameterMappingsDialog.jsx
```diff
@@ -44,7 +59,9 @@ class EditParameterMappingsDialog extends React.Component {
 
     Promise.all(map(widgetsToSave, w => w.save()))
       .then(() => {
-        this.props.dialog.close();
+        const paramValuesSig = this.constructor.getParamValuesSignature(this.state.parameterMappings);
```
kravets-levko: I was thinking about that `getParamValuesSignature` function, and IMO it's not self-explaining. Look here: before we saved a widget (in the beginning of this function) we already have two snapshots: `widget.options.parameterMappings` and `this.state.parameterMappings`. Let's just create a function that will take two that snapshots and compare them, and call it here. You can detect if any parameter changed (instead of serializing all). WDYT?

ranbena: The was my first attempt.
But it mandates the following steps:
1. Diff objects
2. Omit irrelevant changes - type change, title change

The signature solution is shorter, simpler and more effective imo.

kravets-levko: No need to diff objects. I mean something like this:
```javascript
function mapingsChanged(prev, next) {
  const s = JSON.stringify;
  const c = v => v.mapTo;
  
  // not sure if `m.value || m.param.value` will work, maybe 
  // need `MappingType.StaticValue` check as in `getParamValuesSignature`
  const prevSnapshot = sortBy(prev, c)
    .map(m => [ m.mapTo, m.value || m.param.value ]);
  const nextSnapshot = sortBy(next, c)
    .map(m => [ m.mapTo, m.value || m.param.value ]);

  return s(prev) !== s(next);
```

Note that I also did two things to make that check as stable as possible: added names of parameters alongside to values, and sorted them by param name.

ranbena: That'll work but now you'll have to store a clone of original `paramterMappings` which was also a consideration of mine. But if you feel this has value, I'll change it.

kravets-levko: P.S. `m.mapTo` will not work, should use `m.name`

kravets-levko: @ranbena You already have that snapshot: 
https://github.com/getredash/redash/blob/2c1400d3235950989fcf9937a19e7af9cab051ac/client/app/components/dashboards/EditParameterMappingsDialog.jsx#L38
On the left side of this assignment you have previous mappings, on the right - new ones :slightly_smiling_face:

ranbena: You're sure about this? I checked it and it's not the same output.

kravets-levko: What do you mean?

ranbena: The object structure of `this.state.parameterMappings`  and `widget.options.parameterMappings` is not the same.

kravets-levko: `editableMappingsToParameterMappings` returns the same structure as in `widget.options.parameterMappings` (that function exists to create structure that could be assigned to `widget.options.parameterMappings`)

kravets-levko: You should turn that line into something like this:
```javascript
const prevMappings = widget.options.parameterMappings;
const newMappings = editableMappingsToParameterMappings(this.state.parameterMappings);

widget.options.parameterMappings = newMappings;

const somethingChanged = mapingsChanged(prevMappings, newMappings); 
// return `somethingChanged` from modal
```

arikfr: Not too excited about using `JSON.stringify` to comparison. While it works, it feels like a hack that can break in the future for unexpected reasons.

Maybe use [`_.isMatch`](https://lodash.com/docs/4.17.11#isMatch) instead?

(assuming we want to go this path)

ranbena: It's bad practice to do this:
```js
var a;
var b = process(a)
var diff = compare(a, b);
```
Since `process` has a different purpose, it might change one day and unintentionally mess the comparison up. (Unless we have a unit test for this of-course)

kravets-levko: In this case we can use [`_.some`](https://lodash.com/docs/4.17.11#some) to reduce amount of checks

kravets-levko: @ranbena You're not doing what is in you example. You do:
```javascript
var a;
var b = process(c); // <-- here
var diff = compare(a, b);
```
:wink:

kravets-levko: And once more: the only purpose of `editableMappingsToParameterMappings` is to create structure for `widget.options.parameterMappings`; it's the only function that creates that structure, it was intended for that. It has no side effects and by it's definition it returns new structure (and does not modify existing).

ranbena: I think it's a dangerous way to go about it, but it's not a big deal to me.
@kravets-levko, if you have a clear idea of how to implement this, I would appreciate if you could push a fix commit to this PR.

ranbena: > The object structure of `this.state.parameterMappings` and `widget.options.parameterMappings` is not the same.

@kravets-levko What I meant was that `editableMappingsToParameterMappings()` exposes only values of static params so it can't be used as input for comparison.

kravets-levko:
@arikfr @ranbena I updated this PR - feel free to discard my changes if you're not fine with them.

ranbena:
I can't technically approve my own PR - so @kravets-levko can you approve your own changes plz?😅

PR #3446, opened 2019-02-17
Add support for Amazon ES service with IAM authentication

arikfr:
This adds a dedicated query runner for Amazon Elasticsearch service which supports IAM based authentication.


redash/query_runner/amazon_elasticsearch.py
```diff
@@ -0,0 +1,56 @@
...
+    def configuration_schema(cls):
+        return {
+            'type': 'object',
+            'properties': {
+                'server': {
+                    'type': 'string',
+                    'title': 'Endpoint'
+                },
+                'region': {
+                    'type': 'string',
```
rauchy: No title here?

arikfr: No, because the default would be `Region`.

tilfin:
@arikfr It's a good feature but I made PR #3005 4 months ago. We need that EC2 can connect Amazon ES by the credentials of the IAM profile without specified AWS Access Key/Secret Key. To connect Amazon Athena should be supported as well.

arikfr:
@tilfin oops. 😳 I remembered seeing your PR, but when I came to implementing this I couldn't find it while searching GitHub. 🤦‍♂️ 

Would you mind updating your PR to use the new Amazon Elasticsearch data source? Also we probably can remove one of the requests/AWS dependencies?

tilfin:
@arikfr Of course not 😃 We have already used redash from the docker image applying my PR for 4 months so I want to select **requests_aws_sign**.

PR #3447, opened 2019-02-17
Fix: ScheduleDialog won't render for "30 days" interval with no time value

arikfr:
This happens for interval schedules migrated from previous format.

My first frontend PR with a test reproducing a bug 👌


ranbena:
@arikfr can you split the commits to "prettier --fix" and "actual fix"?

client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -192,7 +190,13 @@ class ScheduleDialog extends React.Component {
...
               <TimePicker
                 allowEmpty={false}
-                defaultValue={moment().hour(hour).minute(minute)}
+                defaultValue={
+                  hour
+                    ? moment()
+                      .hour(hour)
+                      .minute(minute)
+                    : null
+                }
```
arikfr: @ranbena,

This is the actual fix. Before it was:

```javascript
defaultValue = {  moment().hour(hour).minute(minute) }
```

And now it's:

```javascript
defaultValue = { hour ? moment().hour(hour).minute(minute) : null }
```

(`moment().hour(null)` returns current hour as integer)

client/app/components/queries/ScheduleDialog.test.js
```diff
@@ -126,6 +133,14 @@ describe('ScheduleDialog', () => {
...
       });
     });
+
+    describe('Supports 30 days interval with no time value', () => {
+      test('Time is none', () => {
+        const [wrapper] = getWrapper({ interval: 30 * 24 * 3600 });
+        const el = findByTestID(wrapper, 'time');
+        expect(el).toMatchSnapshot();
+      });
+    });
```
arikfr: This is the test for the bug.

PR #3448, opened 2019-02-17
Set isDirty to true when changing parameter types

rauchy:
feb-17-2019 12-09-40


client/app/components/parameter-settings.html
```diff
@@ -23,7 +23,7 @@ <h4 class="modal-title">{{ $ctrl.isNewParameter ? 'Add Parameter' : $ctrl.parame
       </div>
       <div class="form-group">
         <label>Type</label>
-        <select ng-model="$ctrl.parameter.type" class="form-control">
+        <select ng-model="$ctrl.parameter.type" class="form-control" ng-change="$ctrl.updateType()">
```
arikfr: We need to set isDirty for any value change, not only type. While the type change is important for validation, the rest are important for consistent UI/UX.

rauchy:
> An alternative would be to only use the OK to apply the changes, signal to query that it's dirty but not save (send HTTP request).

I tend to favor this alternative. The whole "parameters dialog has a mind of its own and decides when to save" feels unreliable.

@arik @ranbena should we take advantage of the fact that we need a major change in this component and rewrite it in React?

arikfr:
> should we take advantage of the fact that we need a major change in this component and rewrite it in React?

Of course :-)

ranbena:
Yeah no doubt should add an "ok/cancel" which auto-saves.
@rauchy feel free to assign me and I'll reactify it.

rauchy:
@ranbena Why do you think auto-save makes more sense here? I feel that having things that persist from different places only adds confusion and it's better to have one Save button.

ranbena:
> @ranbena Why do you think auto-save makes more sense here? I feel that having things that persist from different places only adds confusion and it's better to have one Save button.

Because of the visual connotation.
In a conventional form, for instance, it is obvious that a change to any field would require a "save" at the bottom of the form. There's a visual beginning (top of form) that flows downward to an end (submit button).

Here though, the same convention suggests a correlation between the query and the "save" button, but any other field in the page is unassociated.

ranbena:
Discussion in https://discuss.redash.io/t/query-unsaved-changes-indication/3302
Implementation in #3488.

PR #3450, opened 2019-02-17
Manage user groups in UserEdit

gabrieldutra:
## Description
The idea here is to improve UX for editing user groups by adding an option to manage this in the user edit page.

[Some discussion](https://discuss.redash.io/t/manage-user-groups-in-user-edit-page/3234)

## Changes
- [x] Update user handler to modify groups (`groups` seemed not be be a user attr, `group_ids` was instead)
- [x] Introduce SelectField in DynamicForm (it may be useful in the future as well)
- [x] Manage User Groups in UserEdit - :ok_hand: 

**Possible stuff to change depending on comments**
- Refactor Field proptypes (it seems to work, but it's not creating much better check than an Object propType)
- Separate DynamicForm Fields into smaller components (`DynamicFormUploadField`, `DynamicFormSelectField`) instead of render methods

## Preview (Admin editing other user)
multiselect-preview


client/app/components/proptypes.js
```diff
@@ -34,8 +34,15 @@ export const RefreshScheduleDefault = {
...
-  initialValue: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.bool]),
+  type: PropTypes.oneOf(['text', 'email', 'password', 'number', 'checkbox', 'file', 'select']).isRequired,
+  initialValue: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+    PropTypes.bool,
+    PropTypes.arrayOf(PropTypes.string),
+    PropTypes.arrayOf(PropTypes.number),
+  ]),
+  mode: PropTypes.string,
```
gabrieldutra: Some refactor here will probably be good. Currently this PropType is not much better than an Object assertion.

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -130,6 +131,25 @@ export const DynamicForm = Form.create()(class DynamicForm extends React.Compone
...
+    };
+
+    return getFieldDecorator(name, decoratorOptions)(
+      <Select {...props} mode={mode}>
+        {options && options.map(({ value, title }) => (
+          <Option key={`${value}`} value={value} disabled={readOnly}>{ title || value }</Option>
+        ))}
+      </Select>,
+    );
+  }
```
gabrieldutra: Perhaps transform this `renderSelect` and `renderUpload` into smaller components such as `DynamicFormSelectField` & `DynamicFormUploadField`? (this would reduce some of the logic around)

client/app/components/users/UserEdit.jsx
```diff
@@ -101,22 +108,34 @@ export class UserEdit extends React.Component {
...
+      (currentUser.isAdmin && currentUser.id !== user.id) ? {
+        name: 'group_ids',
+        title: 'Groups',
+        type: 'select',
+        mode: 'multiple',
+        options: this.state.groups,
+        initialValue: this.state.groups.filter(group => includes(user.groupIds, group.value))
+          .map(group => group.value),
+      } : null,
+    ], isNull).map(field => ({ ...field, readOnly: user.isDisabled }));
```
arikfr: Considering that it might take some time to load the groups, maybe have some loading state for the data? I know Ant's Select components have support for this, so it's only a matter of communicating this between the components.

gabrieldutra: Cool, locally it's already done, I also added a "Loading..." placeholder. (better than a whole blank select imo)

redash/handlers/users.py
```diff
@@ -216,7 +216,7 @@ def post(self, user_id):
             user.hash_password(params.pop('password'))
             params.pop('old_password')
 
-        if 'groups' in params and not self.current_user.has_permission('admin'):
+        if 'group_ids' in params and not self.current_user.has_permission('admin'):
```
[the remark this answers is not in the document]
gabrieldutra: We should rather abort with an error message or simply remove the non-belonging groups?

> * We shouldn't allow removing all groups from a user, a user should always have at least 1 group association.

I guess I should add the rule here in the backend as well (frontend it was as easy as putting the field required)

gabrieldutra:
> * When we don't render the select field, we should render the list of groups the user belongs to.

I don't think this should be in the form, but I'll try to come up with a place to put it. (perhaps below or beside the heading as tags).

> * How hard will it be to add this to the create user form?

In case the route for creating a user is pretty similar to this one I would say this will be easy once we migrate the page. I would separate the `renderUserInfoForm` method into a new component that could be used in both places. (Even without this new component, the use of `DynamicForm` brings it only a bit more of effort in case there will be different fields)

gabrieldutra:
> 1. Save not working
   Add group -> save -> refresh page - changes not saved.

I suppose this won't work on netlify since there are some backend changes on the user handler.

> 2. Autocomplete not working
    Focus on select ->"admin" group shows in suggestions -> type "a" -> no suggestions (should show "admin")

Good catch, I didn't notice this one, by default Ant Select filter works with Option values (in this case they are the group ids). Changing `optionFilterProp="children"` fixed that though.

Once I fix Arik's comments I'll comment here :slightly_smiling_face:

redash/handlers/users.py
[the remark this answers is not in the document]
gabrieldutra: @arikfr, please check and consider me inexperienced in those parts :joy:

gabrieldutra:
> * When we don't render the select field, we should render the list of groups the user belongs to.

@arikfr, my ideas for that in the current layout were:

**UserEdit** (User viewing his own profile)

header-tags

**UserShow**

user-show

From that, I'm not very satisfied with the result for **UserEdit** (I got the feeling that it is a lot of information in the header). So how about we render this only for **UserShow** until we figure out a new layout?

arikfr:
> > * When we don't render the select field, we should render the list of groups the user belongs to.

> I don't think this should be in the form, but I'll try to come up with a place to put it. (perhaps below or beside the heading as tags).

Why not put them in the form? It's like a read only field.

gabrieldutra:
> Why not put them in the form? It's like a read only field.

I don't think there is an Antd read only for Select. What I'm doing to bypass this is disabling the options:

readonlyselect

ranbena:
Idk @gabrieldutra the ux is not great with this solution.
A disabled select looks bad, but why not just a tag list?

arikfr:
> > Why not put them in the form? It's like a read only field.
> 
> I don't think there is an Antd read only for Select. What I'm doing to bypass this is disabling the options:

Can we just render the labels there like you do on the show component?

gabrieldutra:
> A disabled select looks bad, but why not just a tag list?

> Can we just render the labels there like you do on the show component?

I had assumed this would be in the form (and considering I'm currently using `DynamicForm`, that would require a bigger change) :sweat_smile: .

Percy view for that:
percy

arikfr:
> I had assumed this would be in the form (and considering I'm currently using `DynamicForm`, that would require a bigger change) 😅 .

Two options: 
1. Add a simple `content` type field, which takes a render function, and renders whatever the function returns.
2. Just skip `DynamicForm` here and implement the form directly. Although option 1 sounds simple enough.

> BTW, I don't know exactly how to handle **UserShow** jest tests now (after this, it now requires Angular 🤔).

Why?

gabrieldutra:
> 1\. Add a simple `content` type field, which takes a render function, and renders whatever the function returns.

Indeed not a bad idea, I totally forgot about those tricks... Thanks! :slightly_smiling_face: 

> Why?

Because of [this](https://circleci.com/gh/getredash/redash/14714). I'm not sure if I'm missing something, but my first guess was that the `Group` was null because the `ngModule` didn't init.

arikfr:
> Because of [this](https://circleci.com/gh/getredash/redash/14714). I'm not sure if I'm missing something, but my first guess was that the `Group` was null because the `ngModule` didn't init.

Oh, right. Maybe stub it until we have a better solution?

arikfr:
Is this ready?

gabrieldutra:
> Is this ready?

Yes, I've just updated with master :)

gabrieldutra:
> One thing perhaps is to swap [...]

I'm currently using the `Tag` component for that, so I guess in this case I'd have to stop using it or use css to style it. The first option doesn't seem good as we would use an Antd component class without the component wrapping it. The css option sounds better to me, so how about I add an override to `ant.less`?

```less
.ant-tag {
  a:hover {
    color: @somecolor
  }
}
```

ranbena:
@gabrieldutra in that case never mind. Leave as is. Not worth the complication. Feel free to merge.

PR #3452, opened 2019-02-17
Fail on superfluous parameters

rauchy:
When running a `ParameterizedQuery` with a schema (as in the case of saved queries), `is_safe` should fail if there are any superfluous parameters which are not specified in the schema.


redash/utils/parameterized_query.py
```diff
@@ -96,10 +96,13 @@ def apply(self, parameters):
         return self
 
     def _valid(self, name, value):
+        if not self.schema:
+            return True
```
arikfr: Why is this needed?

arikfr: For the non-textless API?

rauchy: Precisely.

arikfr: Feels like it might bite us in the future, but can't think of any reason why.

PR #3453, opened 2019-02-17
Handle dropdown queries which are detached from data source

rauchy:
This PR handles an edge case where dropdown queries point to a data source that no longer exists.


redash/utils/parameterized_query.py
```diff
@@ -21,10 +23,13 @@ def _load_result(query_id):
     from redash import models
 
     query = models.Query.get_by_id_and_org(query_id, current_org)
-    require_access(query.data_source.groups, current_user, view_only)
-    query_result = models.QueryResult.get_by_id_and_org(query.latest_query_data_id, current_org)
 
-    return json_loads(query_result.data)
+    if query.data_source:
+        require_access(query.data_source.groups, current_user, view_only)
+        query_result = models.QueryResult.get_by_id_and_org(query.latest_query_data_id, current_org)
```
arikfr: I'm assuming the tests fail because of the cyclic import issue. Maybe use `query.org` here instead of `current_org`?

rauchy: Actually that import was already handled elsewhere. Fixed.

PR #3454, opened 2019-02-17
Avoid reporting bad requests to Sentry

rauchy:


arikfr:
Does this include 404s?

redash/utils/sentry.py
```diff
@@ -4,7 +4,7 @@
 from redash import settings, __version__
 
 
-NON_REPORTED_EXCEPTIONS = ['QueryExecutionError']
+NON_REPORTED_EXCEPTIONS = ['BadRequest', 'QueryExecutionError']
```
rauchy: @arikfr nope, only 400s. If you want 404s as well, smash this:

```suggestion
NON_REPORTED_EXCEPTIONS = ['NotFound', 'BadRequest', 'QueryExecutionError']
```

arikfr: Only the author of the PR, can accept suggestions... Do you think we should keep reporting 404s there?

rauchy: The only thing I can think that we should keep 404s for is any manual construction of links in the frontend that might not pop up as a bug during development and testing.

arikfr: Two things about this:

1. In dev/test we don't send Sentry reports anyway.
2. This kind of issues should be tracked from the frontend.

rauchy:
So where are we with regards to this PR? Do we still want to keep 400s after all?

arikfr:
> So where are we with regards to this PR? Do we still want to keep 400s after all?

I'm conflicted. On the one hand it feels to me that Sentry should receive only actionable issues. On the otherhand, it proved useful debugging a production issue 🤔 

I will open a thread in Sentry's forum to see what they recommend.

arikfr:
https://forum.sentry.io/t/what-to-do-about-inactionable-errors/6059

justinclift:
Looking over that message on the Sentry forums, it seems like sending the extra data does have uses:

```
We were about to [filter those errors 3](https://github.com/getredash/redash/pull/3454), but then we had a production issue resulting in a BadRequest exception. Having all the data Sentry includes with each event was a god send and allowed to resolve this issue in minutes.
```

So, lets not merge this PR.  We can revisit this topic later on if this turns out to be a bad move. :grin:

PR #3456, opened 2019-02-17
Extend support to Elasticsearch nested field queries

fernandoprocy:
Hey, everyone,

In my first PR I mentioned this other issue with Elasticsearch support. Luckily I got some free time and could work on that this weekend.

Currently, Redash doesn't parse nested field queries results properly.

As illustration, there is the logs sample dataset from Elasticsearch, in which there are a few nested structs, like `machine` and `geo`.
```
logs
├── machine
│   ├── os
│   └── ram
├── geo
│   └── coordinates
│       ├── lat
│       ├── lon
│       └── ...
```

Currently, Redash doesn't parse the json result and send it as is, even if you query for a single nested field.

This PR adds support to that, parsing the dict using a dict flatten generic function. If the user queries the parent structure, it will return every nested field. If they only query the nested field, only the nested field will get returned.

screen shot 2019-02-17 at 20 01 55
screen shot 2019-02-17 at 20 00 24

Any criticism will be welcome. Thanks!

### DIFF
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -228,6 +228,16 @@ class BaseElasticSearch(BaseQueryRunner):
 
             return None
 
+        def get_column_name_and_value(column_name, value):
+            if isinstance(value, list) and len(value) == 1:
+                return column_name, value[0]
+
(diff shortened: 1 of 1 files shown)


arikfr:

PR #3458, opened 2019-02-18
Version data - footer -> header menu

ranbena:
In attempt to have `app-header` parity between Redash products, moved version data from the footer to the header dropdown menu.

### Features
1.  Backend version
2. Frontend version (if not equal to backend)
3. Unobtrusive upgrade notification.

screen shot 2019-02-18 at 21 07 43

### Tech notes:
Footer test now irrelevant, can't port test to header cause not yet react, so deleted it.


ranbena:
Btw, the "upgrade" icon does not look good in this size. I tried other FontAwesome icons but not all are available apparently (e.g. [this one](https://fontawesome.com/icons/arrow-alt-circle-down?style=solid). Should we update the lib?

arikfr:
> Btw, the "upgrade" icon does not look good in this size. I tried other FontAwesome icons but not all are available apparently (e.g. [this one](https://fontawesome.com/icons/arrow-alt-circle-down?style=solid). Should we update the lib?

We're on FA4 and that's from FA5. They still have docs for FA4 somewhere.

Of course we need to upgrade, just need the time for it.

client/app/components/Footer.jsx
```diff
@@ -1,27 +1,12 @@
 import React from 'react';
```
arikfr: Maybe delete the whole footer?

client/app/components/app-header/app-header.html
```diff
@@ -124,6 +124,19 @@
             <li>
               <a ng-click="$ctrl.logout()">Log out</a>
             </li>
+
+            <li ng-if="$ctrl.currentUser.hasPermission('super_admin')" class="divider"></li>
+            <li ng-if="$ctrl.currentUser.hasPermission('super_admin')" class="dropdown-menu__version">
```
arikfr: Why not show version to everyone?

ranbena: Just to avoid visual clutter.
Maybe just hide the upgrade button from non-admins?

client/app/components/Footer.jsx
ranbena: Out of this scope :)

client/app/components/app-header/app-header.html
arikfr: We should definitely show the version to users.

About the upgrade button: while non admins have no use for it (they can't upgrade), they can nag the admin to upgrade :) But if you feel it's a distraction, we can hide it for non admins.

client/app/components/Footer.jsx
arikfr: It's just deleting it 🤷‍♂️

Between the help button and version in dropdown, there is very little use for the footer. We can do this in a follow up, if you prefer.

client/app/components/app-header/app-header.html
ranbena: If it's green - it's a distraction.
If it's not green - it's purpose isn't clear (I experimented with many colors).
So, I'm for hiding it from non-admins.

ranbena: Done

client/app/components/Footer.jsx
ranbena: No risk of course but I need to feel confident that it's justified by bringing it to the team.

ranbena:
@arikfr it's ready as far as I'm concerned.

PR #3459, opened 2019-02-19
monitor "schemas" queue to run refresh_schema

koron:
`refresh_schema` tasks won't run because "schemas" queue isn't consumed
with default settings.
and it cause leaking redis storage, a "schemas" list is growing with time.

this PR fix it, adds "schemas" queue to monitor by celery.

----

Where adding `refresh_schema` tasks to "schemas" queue.
https://github.com/getredash/redash/blob/8fc2ecf55c2d9ecd94b391d6d80d58c33fed9373/redash/tasks/queries.py#L412

### DIFF
--- a/setup/docker-compose.yml
+++ b/setup/docker-compose.yml
@@ -24,13 +24,13 @@ services:
     <<: *redash-service
     command: worker
     environment:
-      QUEUES: "scheduled_queries"
+      QUEUES: "scheduled_queries,schemas"
       WORKERS_COUNT: 1
   adhoc_worker:
     <<: *redash-service
     command: worker
     environment:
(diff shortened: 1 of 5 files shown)


setup/docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
     <<: *redash-service
     command: worker
     environment:
-      QUEUES: "queries"
+      QUEUES: "queries,schemas"
```
arikfr: It should be added to the scheduled worker and not the adhoc ones (no reason to block execution of adhoc queries for a background task).

koron: thx! i have fixed it.

PR #3463, opened 2019-02-19
Fix: timedelta values were not JSON serialized properly

arikfr:
.


redash/utils/__init__.py
```diff
@@ -82,7 +82,7 @@ def default(self, o):
         elif isinstance(o, (datetime.timedelta, uuid.UUID)):
             result = str(o)
         # See "Date Time String Format" in the ECMA-262 specification.
-        if isinstance(o, datetime.datetime):
+        elif isinstance(o, datetime.datetime):
```
arikfr: @jezdez I assume this was accidental, right?

jezdez: Looks like it, yeah!

PR #3465, opened 2019-02-19
Make sure Flask app created in Celery's worker process

arikfr:
For the past week or so we experienced very strange issues where BigQuery and Kylin queries would hang on a futex for no apparent reason. Luckily we could trace the issue to the time we deployed a new code version, so we could limit the search permitter. 

Because there were no changes to BigQuery's query runner, I assumed the issue was with one of the dependencies that got an upgrade between the releases. But there was no luck there: none of the changed dependencies were related.

Following some other ideas and more closely looking at code changes, I've noticed that our previous deployment had a modified `redash.worker` module where we had commented out `add_periodic_tasks`. Looking closer, I realized that `add_periodic_tasks` changed the behavior we tried to create with `init_celery_flask_app`: the Flask app was now created in the master Celery process and not the forked workers.

In turn this resulted in memory sharing between processes and fun dead locks...

I commented out `add_periodic_tasks` until we rewrite the extensions registry to not rely on a Flask app (#3466).

### DIFF
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -138,11 +138,3 @@ def create_app(load_admin=True):
     users.init_app(app)
 
     return app
-
-
-def safe_create_app():
-    """Return current_app or create a new one."""
-    if current_app:
-        return current_app
-
-    return create_app()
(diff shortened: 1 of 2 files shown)


arikfr:
FYI @emtwo, @jezdez.

jezdez:
Interesting, I'll follow up in #3466.

PR #3470, opened 2019-02-19
Migrate Data Sources and Alert Destinations pages to React

gabrieldutra:
## Description
Migrate Data Sources and Alert Destinations pages to React

## Changes
**General**
- [x] Migrate TypePicker to React
- [x] Review TypePicker style

**List Pages**
- [x] Data Sources
- [x] Destinations
- [x] Improve behavior for empty set

**Create Pages**
- [x] Data Sources
  - [x] Select Type
  - [x] Configure Type
  - [x] Done
- [x] Destinations
  - [x] Select Type
  - [x] Configure Type
  - [x] Done

**Edit Pages**
- [x] Data Sources
- [x] Destinations

## To Do
- [x] Routes `/data_sources/new` & `/destinations/new` auto-opening the modal
- [x] Help using `HelpTrigger`
- [x] Refine Edit forms
- [x] Update Cypress tests

**Possible Extra**
- Add edit & delete actions to the cards on List Pages ([ref](https://ant.design/components/card/#components-card-demo-meta))

## Notes
Data Sources and Destinations are very similar pages, however I'm not sure how much logic I should share. They are sharing `TypePicker` (renamed to `CardsList`) and `DynamicForm`, but to me it doesn't feel that they should share much more. I'll keep the idea of refactoring parts of it to avoid code duplication anyway.

### Overview
overview-dialog


client/app/pages/data-sources/CreateDataSource.jsx
```diff
@@ -3,26 +3,61 @@ import Steps from 'antd/lib/steps';
...
   constructor(props) {
     super(props);
-    this.state = { dataSourceTypes: [], selectedType: null };
+    this.state = { dataSourceTypes: [], selectedType: null, currentStep: StepEnum.SELECT_TYPE };
     DataSource.types(dataSourceTypes => this.setState({ dataSourceTypes }));
+    this.topRef = React.createRef();
   }
 
+  scrollToTop = () => {
+    window.scrollTo({ top: this.topRef.current.offsetTop, behavior: 'smooth' });
```
kravets-levko: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo), this way to use `window.scrollTo` not supported in IE/Safari. Can you check it?

gabrieldutra: I was aware that sth like this could show up :thinking:. Thanks, I'll check this.

client/app/components/data-sources/EditDataSourceForm.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const handleSubmit = (values, successCallback, errorCallback) => {
+    helper.updateTargetWithValues(dataSource, values);
+    dataSource.$save(
+      (data) => {
+        successCallback('Saved.');
+        onSuccess(data);
+      },
+      (error) => {
+        if (error.status === 400 && 'message' in error.data) {
```
kravets-levko: You have to check if `error.data` exists - this callback is called on any error (it's actually `dataSource.$save(...).$promise.catch()`), and `error` may be an `Error` object which does not have `data` property - in this case `in` will throw an exception.

client/app/components/TypePicker.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+  constructor(props) {
+    super(props);
+    this.state = { searchText: '' };
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  renderListItem(item) {
+    const titleStyle = {
+      fontSize: '13px',
```
kravets-levko: Maybe, move it somewhere to `.less` file? IMO css-in-js is acceptable when it represent some structural stuff, and if it's unique to specific element; but this one is completely "design" and it's related to a class of elements - no need to set this as inline style for each element - make it a CSS class, and let browser to do it's job.

client/app/components/TypePicker.jsx
```diff
@@ -0,0 +1,85 @@
...
+    return (
+      <div className="text-center" data-test="TypePicker">
+        <Search
+          className="m-b-20"
+          placeholder="Search..."
+          onChange={e => this.setState({ searchText: e.target.value })}
+          style={{ maxWidth: 300 }}
+          autoFocus
+        />
+        {isEmpty(filteredTypes) ? (<EmptyState />) : (
```
kravets-levko: By default `EmptyState` and `LoadingState` have an outline ('.tiled' class set) - check if it works as expected, maybe you'll need to override this.

client/app/components/TypePicker.jsx
gabrieldutra: `css-in-js` have always messed up with my concepts :joy:. My parameter so far has been as simple as accepting it for changes that were isolated to the component, but I think I'll adopt yours. Makes total sense for it as a class. I'm thinking of rewriting all styling from here into a `type-picker.less`, however it may get tricky when it comes to handling it over Ant's components.

gabrieldutra:
I believe this can already get some review on code and UX improvements. A few notes:
- **Code Climate:** Data Sources and Destinations share the same logic, I tried to keep them sharing the same components, but I also wanted to keep some logic separate, as they are different pages. Ideas to reduce duplicity are accepted.
- The long name behavior in the `TypePicker` still need some improvement, it's now just cutting texts with more than 3 lines, but I think the best would be to make row cards height flexible.

gabrieldutra:
@kravets-levko do you think it's worthy to Create an `ItemsCard` visualization (similar to `ItemsTable`) to render those pages using `ItemsList` instead of `TypePicker`?

kravets-levko:
@gabrieldutra If you use (or think it's good idea to use) `ItemsList` wrapper for that pages - then yes, definitely.

arikfr:
It's just a simple list of cards with search that filters it. I doubt it's worth adding an abstraction on top of it while also making the ItemsList more complicated.

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -218,36 +217,6 @@ export const DynamicForm = Form.create()(class DynamicForm extends React.Compone
       </Form>
     );
   }
-});
-
-export default function init(ngModule) {
```
kravets-levko: :tada:

ranbena:
A few UX observations:
#### `/data_sources`
* The search functionality in Data Sources and Alert Destinations pages seems unnecessary. It makes sense when adding a new item since it's an overwhelming list, but irrelevant when it's a list of own items.
#### `/data_sources/id`
* The "Help setting up BigQuery" got stuck to the right with no spacing.
* The form got narrower  

#### `/new`
* I like the steps addition 👍 But it looks really stretched out and the layout "title (narrow) > steps (wide) > search (narrow) > items (wide)" does not look good. How about making both "Add new" open in a modal?
  * Pros: Much more visual appeal, more consistent with other "add new" flows in the site.
  * Cons: Data source items would take up too much space. Would require conversion to ItemsList.
@arikfr @kravets-levko @gabrieldutra Wdyt?
* In step 2, there's not enough bottom spacing under the "save" button. To be consistent with the top title spacing should add another 18px.

arikfr:
> Cons: Data source items would take up too much space. Would require conversion to ItemsList.

Can you explain why?

ranbena:
> > Cons: Data source items would take up too much space. Would require conversion to ItemsList.
> 
> Can you explain why?

The modal would be unbearably tall with the big "card" items. `ItemsList` would optimize the height, allow vertical scrolling (or pagination if we want) and search built in. Not a must but an option.

arikfr:
> The modal would be unbearably tall with the big "card" items. ItemsList would optimize the height, allow vertical scrolling (or pagination if we want) and search built in. Not a must but an option.

👍 Ok, got it. I thought you were referring to the list page. Yeah, we can show a list instead. I don't think we should use our `ItemsList` though. ItemsList is more about the data fetching than the visual representation. In this case we would probably just use Ant's [`<List>`](https://ant.design/components/list/) or just implement something simpler ourselves (but no reason to use a table either, right?).

What happens when someone navigates to `data_sources/new` directly: should we still show the list page in the "background"?

ranbena:
> AntD/list or just implement something simpler ourselves (but no reason to use a table either, right?).

Right!

> What happens when someone navigates to `data_sources/new` directly: should we still show the list page in the "background"?

Perhaps remove this route?

arikfr:
> Perhaps remove this route?

We send people to it from the Getting Started guide and Getting Started email.

ranbena:
> > Perhaps remove this route?
> 
> We send people to it from the Getting Started guide and Getting Started email.

Let's make the route auto-open the modal then. It's not a good idea to have parallel implementations (one in modal, one in page).

client/app/pages/data-sources/show.js
```diff
@@ -1,132 +0,0 @@
...
-      name: 'Test Connection', pullRight: true, callback: testConnection, disableWhenDirty: true,
-    },
-  ];
-}
-
-export default function init(ngModule) {
-  ngModule.controller('DataSourceCtrl', DataSourceCtrl);
-
-  return {
-    '/data_sources/new': {
```
kravets-levko: @gabrieldutra Probably you should handle this similar to New user page/dialog - we use `data_sources/new` in some other places (do search by `client/app` folder - you'll find them). For alert destinations seems it could be removed, but not sure if it's not used elsewhere - say in docs :thinking:

client/app/components/CreateSourceDialog.jsx
```diff
@@ -0,0 +1,179 @@
...
+        ] : [
+          (<Button key="previous" onClick={this.resetType}>Previous</Button>),
+          (
+            <Button key="submit" htmlType="submit" form="sourceForm" type="primary" loading={savingSource}>
+              Create
+            </Button>
+          ),
+        ]}
+      >
+        <Steps className="hidden-xs" size="small" current={currentStep} progressDot>
```
gabrieldutra: I had to hide this for mobiles because it was overflowing the dialog, I've opened ant-design/ant-design#15215 for this. I should add a comment to track this in case we want this removed when fixed, but perhaps this is better hidden for mobiles as it increases height that we already don't have much.

client/app/pages/data-sources/show.js
gabrieldutra: You're a few minutes ahead of me :stuck_out_tongue_closed_eyes:. It's my intention to do it, just wanted a general overview first :)

arikfr:
> IMO it's not that good that the form has a scroll in most cases, but I did like the Dialog's flow.

I agree. The scroll is a deal breaker. Maybe have it as a page but keep the steps widget the same width as the form?

ranbena:
> but I did like the Dialog's flow.

Me too. Especially the "Select type" list. Very appealing.
@gabrieldutra I see you implemented my "done" hack :P

> The scroll is a deal breaker.

Let's weigh some options:
1. Remove max height, let the dialog be as long as it needs to be. (Doesn't break, still scrollable just as in page format)
2. Split long forms into separate steps. (Can add `<steps>` dynamically)

I played around with option 1, it doesn't look bad and I +1 it.

> Maybe have it as a page but keep the steps widget the same width as the form?

Plan B 😒

arikfr:
> Remove max height, let the dialog be as long as it needs to be. (Doesn't break, still scrollable just as in page format)

This crossed my mind, but I thought you wouldn't like it :) If it looks good, I'm cool with this option.

gabrieldutra:
I've updated the Form not to have a `max-height`. I let the List with it because it looked better this way and this seemed to offer a better experience for it, but LMK in case you think the opposite.

> 2. Split long forms into separate steps. (Can add <steps> dynamically)

It would require some cool methods to be written, but that's certainly possible. I thought about it as well, but I'm not that sure about UX for this.

> I see you implemented my "done" hack :P

Yep, I remembered you had some work with Steps and dialogs, so I checked it to steal some ideas :grin:

> This crossed my mind, but I thought you wouldn't like it :) If it looks good, I'm cool with this option.

Haha, actually it would fall me apart to undo the Dialog, so I definitely wanted Ran's opinion on that first.

ranbena:
I think it's great.
I'll now continue reviewing 👨‍💻📝

ranbena:
Btw, noticed Google Analytics and Google Hangouts show no icon.

gabrieldutra:
> I'll now continue reviewing

Take note that I still have stuff to do in the code, I pushed it when I realized that the Dialog could have a blocker, hence it needed the discussion.

BTW, do you think it's a good option to have the helpers in the `HelpTrigger` instead of links? If so, there could be some kind of pattern for this `DATA_SOURCE_[TYPE]` or sth.

> Btw, noticed Google Analytics and Google Hangouts show no icon.

Analytics in here is showing fine, the Hangouts is expected, it seems the backend is updated, but I didn't update with master yet (so, no image in the folder) :slightly_smiling_face: .

ranbena:
> Take note that I still have stuff to do in the code

Oh ok I'll tend to it when ur ready.

> BTW, do you think it's a good option to have the helpers in the `HelpTrigger` instead of links? 

No doubt yes, all help docs should open in drawer.

> If so, there could be some kind of pattern for this `DATA_SOURCE_[TYPE]` or sth.

Sure.

> Analytics in here is showing fine, the Hangouts is expected, it seems the backend is updated, but I didn't update with master yet (so, no image in the folder) 🙂 .

👍

gabrieldutra:
@ranbena @kravets-levko, I'm a bit delayed, but this is ready for your comments again. Notes:
- I could not make a better behavior for Data Source long names using Ant Cards (currently height is fixed for 3 lines and overflow is hidden). I'll appreciate if you have ideas for this :grin:. The options I have in mind: - use javascript to do some better trick; - use the old `visual-card` view.
- For the `HelpTrigger` I prioritized the least changes to the component to make it work with Data Sources.
- `cy.wait` was necessary for Dialogs to appear on Percy. I need to investigate this better, but this won't affect review :)

No big changes in UI after the last gif, so I won't create another one :slightly_smiling_face:

client/app/pages/data-sources/DataSourcesList.jsx
```diff
@@ -0,0 +1,142 @@
...
+import { routesToAngularRoutes } from '@/lib/utils';
+import CardsList from '@/components/cards-list/CardsList';
+import LoadingState from '@/components/items-list/components/LoadingState';
+import CreateSourceDialog from '@/components/CreateSourceDialog';
+import helper from '@/components/dynamic-form/dynamicFormHelper';
+
+class DataSourcesList extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { dataSourceTypes: [], dataSources: [], loading: true };
```
ranbena: My personal pref is that if state has no initial values that rely on props, use the "out of constructor" syntax cause it's cleaner.

client/app/pages/data-sources/DataSourcesList.jsx
```diff
@@ -0,0 +1,142 @@
...
+      helpTriggerPrefix: 'DS_',
+      onCreate: this.createDataSource,
+    }).result.then((success) => {
+      if (success) {
+        this.setState({ loading: true });
+        DataSource.query(dataSources => this.setState({ dataSources }));
+      }
+    }).finally(() => {
+      if ($route.current.locals.isNewDataSourcePage) {
+        navigateTo('data_sources');
```
ranbena: What's this for?

client/app/pages/data-sources/DataSourcesList.jsx
```diff
@@ -0,0 +1,142 @@
...
+      title: dataSource.name,
+      imgSrc: `${IMG_ROOT}/${dataSource.type}.png`,
+      href: `data_sources/${dataSource.id}`,
+    }));
+
+    return isEmpty(dataSources) ? (
+      <div className="text-center">
+        There are no data sources yet.
+        <div className="m-t-5">
+          <a className="clickable" onClick={this.showCreateSourceDialog}>Click here</a> to add one.
```
ranbena: Should add `policy.isCreateDataSourceEnabled` check here?

client/app/pages/data-sources/DataSourcesList.jsx
```diff
@@ -0,0 +1,142 @@
...
+    {
+      path: '/data_sources',
+      title: 'Data Sources',
+      key: 'data_sources',
+    },
+    {
+      path: '/data_sources/new',
+      title: 'Data Sources',
+      key: 'data_sources',
+      isNewDataSourcePage: true,
```
ranbena:
```suggestion
      isNewDataSourcePage: policy.isCreateDataSourceEnabled(),
```

client/app/pages/destinations/DestinationsList.jsx
```diff
@@ -0,0 +1,134 @@
...
+    {
+      path: '/destinations',
+      title: 'Alert Destinations',
+      key: 'destinations',
+    },
+    {
+      path: '/destinations/new',
+      title: 'Alert Destinations',
+      key: 'destinations',
+      isNewDestinationPage: true,
```
ranbena: Add `policy.canCreateAlert()` where relevant.

client/app/pages/data-sources/DataSourcesList.jsx
```diff
@@ -0,0 +1,142 @@
...
+    }).result.then((success) => {
+      if (success) {
+        this.setState({ loading: true });
+        DataSource.query(dataSources => this.setState({ dataSources }));
+      }
+    }).finally(() => {
+      if ($route.current.locals.isNewDataSourcePage) {
+        navigateTo('data_sources');
+      }
+    });
```
ranbena: No need for state `{ loading: false }` ?

client/app/components/CreateSourceDialog.jsx
```diff
@@ -0,0 +1,191 @@
...
+        <div className="d-flex justify-content-center align-items-center">
+          <img
+            className="profile__image--settings"
+            src={`${imageFolder}/${selectedType}.png`}
+            alt={type.name}
+            width="48"
+          />
+          <h4 className="m-0">{type.name}</h4>
+          {(helpTriggerPrefix && HELP_TRIGGER_TYPES[helpTriggerType]) && (
+            <HelpTrigger className="p-l-5" type={helpTriggerType} />
```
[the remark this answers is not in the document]
arikfr: Having the setup instructions aligned this way feels like it's related to the `Name` field.

Also, in the past (before moving to React) we used to have a separator (`<hr>`) between the name and the other properties. The reason was that there is a conceptual separation between them: name is something you choose while the other are the configuration.

I think we should bring back this separator (or some other visual distinction). Also, worth adding some placeholder text there, as people often get confused about what the name should be (i.e. is it something predefined by the database or something they can choose).

client/app/components/CreateSourceDialog.jsx
```diff
@@ -0,0 +1,191 @@
...
+    const helpTriggerType = `${helpTriggerPrefix}${toUpper(selectedType)}`;
+    return (
+      <div className="p-5">
+        <div className="d-flex justify-content-center align-items-center">
+          <img
+            className="profile__image--settings"
+            src={`${imageFolder}/${selectedType}.png`}
+            alt={type.name}
+            width="48"
+          />
```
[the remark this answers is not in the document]
gabrieldutra: This one was tricky because ElasticSearch logo is an exception, all the other images already have some spacing. I tested with 5 px of padding in the image here, it looked acceptable to me.

client/app/components/CreateSourceDialog.jsx
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  renderItem(item) {
+    const { imageFolder } = this.props;
+    return (
+      <List.Item
+        className="p-l-10 p-r-10 clickable"
+        onClick={() => this.selectType(item.type)}
+      >
+        <PreviewCard title={item.name} imageUrl={`${imageFolder}/${item.type}.png`}>
```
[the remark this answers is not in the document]
gabrieldutra: > in the next step (in the title) they are not.

They actually are, but the round effect is only noticeable with ElasticSearch. I enabled it exactly due to this white spacing issue. ElasticSearch currently looks better without rounding, so I did :slightly_smiling_face:

client/app/pages/data-sources/DataSourcesList.jsx
gabrieldutra: Adopted :+1:

client/app/pages/data-sources/DataSourcesList.jsx
gabrieldutra: remove `/new` from the URL when the Create Dialog is closed

client/app/pages/data-sources/DataSourcesList.jsx
gabrieldutra: I was actually trying to avoid `/new` being in the url without the dialog open. Perhaps not open the modal and redirect to `/data_sources` in case the user can't create data sources?

client/app/pages/destinations/DestinationsList.jsx
gabrieldutra: `Alert !== AlertDestination` :stuck_out_tongue_closed_eyes: . Though I can add a new policy function for Alert Destinations if you think this is relevant

client/app/pages/data-sources/DataSourcesList.jsx
gabrieldutra: I'm currently setting `loading: false` only for success cases. Error cases should be caught by angular, but I realized they are not :thinking:.

client/app/components/CreateSourceDialog.jsx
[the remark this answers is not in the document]
ranbena: Since the ES logo needs updating anyway (it's an orange square now) we can just make sure it's got spacing like all the other kids.

client/app/pages/data-sources/DataSourcesList.jsx
ranbena: Refresh page protection?

client/app/pages/data-sources/DataSourcesList.jsx
ranbena: Go for it

client/app/pages/destinations/DestinationsList.jsx
ranbena: My thinking was that if you can't create an alert you probably can't create a destination as well.
It's only for future-proofing anyway. Whatever you choose.

client/app/pages/data-sources/DataSourcesList.jsx
gabrieldutra: Actually it just seemed the proper behavior for it :grin:. I didn't think about this before, but it ends up working as a refresh page protection as well.

cypress/integration/destination/create_destination_spec.js
```diff
@@ -5,7 +5,8 @@ describe('Create Destination', () => {
   });
 
   it('renders the page and takes a screenshot', () => {
-    cy.getByTestId('TypePicker').should('contain', 'Email');
+    cy.getByTestId('CreateSourceDialog').should('contain', 'Email');
+    cy.wait(1000);
```
ranbena: If you don't wait, it snaps before the animation completes?

gabrieldutra: > If you don't wait, it snaps before the animation completes?

Exactly :confused:

gabrieldutra: https://percy.io/Redash/redash/builds/1539798

ranbena: The best way to tackle this I think, is to disable animations when testing.
Is there a way to insert sth like this for tests?
```css
* {
    animation-duration: 0s !important;
}
```

ranbena: Or perhaps have a less file that overrides:
```less
@animation-duration-slow: 0;
```

client/app/components/CreateSourceDialog.jsx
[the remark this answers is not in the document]
gabrieldutra: **Edit Page**
edit-page

**Create Source Modal**
create-source-modal
This one may look a bit buggy due to image white spacing, but to me it was where it looked better :thinking:. If you prefer it right aligned (or somewhere else) lmk :+1: 

> I think we should bring back this separator (or some other visual distinction). Also, worth adding some placeholder text there, as people often get confused about what the name should be (i.e. is it something predefined by the database or something they can choose).

The placeholder is actually easy to add, but I'm out of ideas :sweat_smile:. I left it with `My [Source Type Name]`, tell me if you have anything better in mind.

For `<hr />` I created a `contentAfter` prop in DynamicForm field (can either be a Node or a function `(fieldValue) => Node`)

cypress/integration/destination/create_destination_spec.js
gabrieldutra: I was passing through their docs today and I saw [this](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Interacting-With-Elements):

"For example, when writing a `.click()` command, Cypress ensures that the element is able to be interacted with (like a real user would). It will automatically wait until the element reaches an “actionable” state by:
- Not being hidden
- Not being covered
- Not being disabled
- Not animating"

If this one is really due to the animation, it's an issue with Percy snapshot command, not with Cypress, so perhaps workaround this with `.click()` on, let's say, the Search input will be much easier than remove animations. WDYT?

ranbena: Interesting workaround. Try it and see if it works?

Btw, Cypress screenshots address this problem directly:
```js
Cypress.Screenshot.defaults({
  disableTimersAndAnimations: false
})
```

gabrieldutra:
> All things considered I favor positioning the "Setup Instructions (?)" at the top right.

Perhaps right-aligned but not aside to the Name label? (I don't see a simple way to put it aside to it, also it makes some sense for it not to be there :sweat_smile:)

right-aligned-help-trigger

> I think the step titles should be either "Select type > Configure > Done" or "Type selection > Configuration > Done".

:+1:

gabrieldutra:
Thanks, @ranbena, my last concern about this is:

> I could not make a better behavior for Data Source long names using Ant Cards (currently height is fixed for 3 lines and overflow is hidden). I'll appreciate if you have ideas for this :grin:. The options I have in mind: - use javascript to do some better trick; - use the old `visual-card` view.

If you've already checked this and actually think it's an acceptable behavior, LMK :sweat_smile:.

BTW, I had to revert `cy.click` to `cy.wait` as it didn't work. If this is really due to animation I think it won't make weird diffing, so it's safe to keep it until a better solution appears. (I've tried to use [percy specific CSS](https://docs.percy.io/docs/percy-specific-css) with your idea as well)

ranbena:
> I'll appreciate if you have ideas for this

Overflow hidden is good enough cause it prevents breakage and long names are probably an edge case (pun intended).
In addition you can add this, which will benefit the majority of our users:
```css
.cards-list .cards-list-item h3 {
 ...
  display: -webkit-box;
  -webkit-line-clamp: 3;
  -webkit-box-orient: vertical;
}
```

> BTW, I had to revert `cy.click` to `cy.wait` as it didn't work.

👍

gabrieldutra:
Thanks a lot @ranbena! :grin: 

I actually had to add `/* autoprefixer: off */` for it to work (the `box-orient` was being removed even thought it's necessary to do the `line-clamp`). It improved the experience anyway.

If there's no further comments, this is ready to go when convenient. :+1:

ranbena:
Go ahead and merge whenever _you're_ ready.

gabrieldutra:
Conflicts resolved, gonna merge this big guy tomorrow :rocket:

ranbena:
> When saving a data source it should open the data source page, so the user can do a `Test Connection`.

But, if possible, without the autofocusing in that specific scenario, cause it's confusing ("I just created this, why should I need to edit it??").

ranbena:
Just a ux thought I want to share (out of scope):

Screen Shot 2019-03-27 at 18 37 08

gabrieldutra:
> When saving a data source it should open the data source page, so the user can do a Test Connection.

:+1: 

> But, if possible, without the autofocusing in that specific scenario, cause it's confusing ("I just created this, why should I need to edit it??").

:thinking:, perhaps disable the autofocus for Edit Pages? It makes a lot of sense when it's a form for a new resource, but when editing I don't think it's that relevant (as edit pages also perform as a view page in this case)...

gabrieldutra:
@arikfr & @ranbena,

I've updated it to redirect to the created resource.

As for the autofocus, used to set the first input with it DynamicForm by default, I removed this and turned it into a field property. With this, I also removed autofocus for User Edit.

Please check my last commits :slightly_smiling_face:

gabrieldutra:
> Just a ux thought I want to share (out of scope):
> 
> Screen Shot 2019-03-27 at 18 37 08

I've been thinking about it and I actually don't like that much having an action button in a Notification, but I think an option would be to automatically run the connection test when creating a data source, then lead to 2 scenarios:
- Connection success: Inform the user with a notification and save it;
- Connection failure: Show the error with a dialog and offer to either Save it anyway or go back to edit setting data;

This could be in the Creation Dialog as a new step as well.

ranbena:
> * Connection success: Inform the user with a notification and save it;
> * Connection failure: Show the error with a dialog and offer to either Save it anyway or go back to edit setting data;

YES

> This could be in the Creation Dialog as a new step as well.

YESSSS

@gabrieldutra plz open a forum topic and I'll submit a design.

arikfr:
Good to go.

I have to admit that I'm not 100% sold on the new dialog interface: it feels cramped and makes the different types less discoverable/visible. Let's move on, but keep this in mind as something to revisit.

PR #3471, opened 2019-02-20
Fix prometheus query runner: get_schema and query range

veerkat:
1. Prometheus schema missing 'columns' field caused frontend QueryEditor loading failed
2. Prometheus query api accepts <rfc3339 | unix_timestamp> type for query range, but frontend QueryEditor provides datetime range parameter like 'YYYY-MM-DD HH:mm:ss', so convert datetime string to unix_timestamp

### DIFF
--- a/redash/query_runner/prometheus.py
+++ b/redash/query_runner/prometheus.py
@@ -1,5 +1,7 @@
 import requests
+import time
 from datetime import datetime
+from dateutil import parser
 from urlparse import parse_qs
 from redash.query_runner import BaseQueryRunner, register, TYPE_DATETIME, TYPE_STRING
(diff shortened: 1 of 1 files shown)


redash/query_runner/prometheus.py
```diff
@@ -37,6 +39,30 @@ def get_range_rows(metrics_data):
     return rows
 
 
+# Convert datetime string to timestamp
+def convert_query_range(payload):
+    range = {}
```
arikfr: range is a built-in function in Python, better use a different name.

veerkat: I resolved and pushed

arikfr:
Thanks!

PR #3473, opened 2019-02-21
Scrollbars styling

kravets-levko:
Webkit-based: 

image

image

Firefox:

image

image

WDYT?

### DIFF
--- a/client/app/assets/less/redash/redash-newstyle.less
+++ b/client/app/assets/less/redash/redash-newstyle.less
@@ -975,8 +975,17 @@ text.slicetext {
 // Scrollbars
 
 body {
-  @size: 12px;
+  @size: 10px;
 
+  // Firefox
+  @-moz-document url-prefix() {
+    * {
+      scrollbar-color: @gray transparent;
+      scrollbar-width: thin;
+    }
+  }
+
+  // Webkit-based
   ::-webkit-scrollbar {
     width: @size;
   }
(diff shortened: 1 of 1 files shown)


ranbena:
I think it looks SO MUCH BETTER without the track.
But perhaps the `thumb` should be left as is, to be consistent with OS styling. Wdyt?

kravets-levko:
@ranbena On most Linux desktops scrollbars are so ugly...

ranbena:
> @ranbena On most Linux desktops scrollbars are so ugly...

lol I know 😆 
I think removing the track gives a substantial benefit to the user, but other than that, it's considered a no-no to change native OS styling.

kravets-levko:
@ranbena Hah! When you apply custom style to one of scrollbar elements - browser stops using default for others:

image

image

We should keep all styles there :slightly_smiling_face:

kravets-levko:
@ranbena 

> Chrome thumb has no rounded corners but it's not consistent with FF.

It's square and exactly the same color (I checked). They have different width because I don't know the real width of scrollbar in FF (and it may be different in different OS - i.e. 1/2 from default scrollbar width).

> Chrome thumb doesn't appear/disappear when scrolling - it's now always displayed

Cannot test it on mac :-( in all Windows and 99.9% Linux versions scrollbar is always visible. I'll check if it's possible to do something with it, and if not - I'll better just close this PR.

ranbena:
> It's square and exactly the same color (I checked).

On a mac it's round in FF and square in Chrome :/

kravets-levko:
Okay. It was a nice try, but not this time

ranbena:
😆 Browsers need to work out their differences

war

PR #3480, opened 2019-02-21
Treat repeated BigQuery fields as arrays

washort:
The BigQuery runner doesn't check if fields are repeated and so produces errors trying to coerce them with `int()`, `float()` etc.

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -31,29 +31,27 @@ types_map = {
     'TIMESTAMP': TYPE_DATETIME,
 }
 
+def transform_cell(field_type, cell_value):
+    if field_type == 'INTEGER':
+        return int(cell_value)
+    elif field_type == 'FLOAT':
+        return float(cell_value)
+    elif field_type == 'BOOLEAN':
+        return cell_value.lower() == "true"
+    elif field_type == 'TIMESTAMP':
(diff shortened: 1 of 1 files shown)


redash/query_runner/big_query.py
```diff
@@ -31,29 +31,27 @@
...
-        elif field['type'] == 'FLOAT':
-            cell_value = float(cell_value)
-        elif field['type'] == 'BOOLEAN':
-            cell_value = cell_value.lower() == "true"
-        elif field['type'] == 'TIMESTAMP':
-            cell_value = datetime.datetime.fromtimestamp(float(cell_value))
+        if field['mode'] == 'REPEATED':
+            cell_value = [transform_cell(field['type'], item['v']) for item in cell['v']]
+        else:
+            cell_value = transform_cell(field['type'], cell['v'])
```
arikfr: If I'm not mistaken, `transform_cell` will now return None for any cell that has a type different than INTEGER/FLOAT/BOOLEAN/TIMESTAMP... You need a default return in `transform_cell`.

arikfr: Also you need to handle the case of `cell_value is None` there (many of the conversion will raise an exception if they get a `None` value).

arikfr:
Thanks.

PR #3482, opened 2019-02-22
Improve list pages layout

kravets-levko:
List pages (Dashboard, Queries, Users, Group Members, Group Data Sources) has a layout with sidebar. On large screens content (table with items) is on left and takes 3/4 of width, and sidebar is next to it and takes 1/4 of width. On small screens sidebar is placed above the table, and both of them are full width.

Previously it was implemented by rendering sidebar twice: sidebar, table, sidebar again. On large screens first sidebar was hidden and second one visible, on small screens they swapped - to keep visual order of blocks. That was bugging me for a while, and then I realized that CSS Flexbox can to do what we need - using `order` property: on large screens we set `order: 0` to table and `order: 1` to sidebar, on smaller screens we swap them - and actual order in DOM does not matter.

Also added some wrappers to do all the dirty job: `Layout`, `Layout.Content` and `Layout.Sidebar` component.

Visually nothing changed (at most - few pixels difference in width comparing to Bootstrap classes).

### DIFF
--- a/client/app/assets/less/redash/redash-newstyle.less
+++ b/client/app/assets/less/redash/redash-newstyle.less
@@ -58,24 +58,6 @@ body {
   border-left-color: #1b809e;
 }
 
-.list-content {
-  @media (min-width: 992px) {
-    padding-right: 0;
-  }
-}
-
-.list-control-r-b {
-  @media (max-width: 992px) {
-    display: none;
-  }
-}
-
-.list-control-t {
-  @media (min-width: 992px) {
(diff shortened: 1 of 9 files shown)


kravets-levko:
@ranbena I'm not a big UI expert, so just kept it as is. And i prefer not to change it for now - the idea of this PR is to remove that smelly code and replace it with new and shiny. Let's discuss your suggestion elsewhere, and do it as new PR (if needed).

PR #3484, opened 2019-02-23
Use series name as pie chart label

washort:
It was unconditionally using the column name; this uses the series name instead if specified in the chart editor.


washort:
Added to dashboard: https://redash-preview.netlify.com/queries/69#145

PR #3487, opened 2019-02-24
Fix DynamicForm ignoring default value for checkboxes

### LINKED ISSUE #3486: DynamicForm ignores default value for checkboxes
### Issue Summary

Checkbox field might have a default value of true, it seems that DyanmicForm ignores this.

### Steps to Reproduce

1. Open new Data Source form of type BigQuery
2. "Use Standard SQL" should  be checked by default, but it isn't.

### Technical details:

* Redash Version: master

gabrieldutra:
## Description
Closes #3486.

During DynamicForm migration the `setDefaults` method was removed for better UX (using placeholders instead). However, checkboxes don't have placeholders :man_shrugging:. This brings back the method, but applying it only to checkboxes.

## Preview
checkboxes-dynamic-form

### DIFF
--- a/client/app/components/dynamic-form/dynamicFormHelper.js
+++ b/client/app/components/dynamic-form/dynamicFormHelper.js
@@ -1,4 +1,4 @@
-import { each, includes } from 'lodash';
+import { each, includes, isUndefined } from 'lodash';
 
 function orderedInputs(properties, order, targetOptions) {
   const inputs = new Array(order.length);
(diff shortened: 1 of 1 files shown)


ranbena:
@gabrieldutra I'm having trouble understanding this code fix.
Can you elaborate on what's going on? Mainly, what kind of data can be expected for each argument.

gabrieldutra:
> Can you elaborate on what's going on? Mainly, what kind of data can be expected for each argument.

I'll try :grin:.

First, most of these methods are just like they were before the migration and the general idea is to transform data source configuration schema from the backend into fields to be rendered in the form. It takes basically two arguments to that:
- **configurationSchema** ([api](https://redash-preview.netlify.com/api/data_sources/types)): data about the fields (properties, order, required, etc)
- **Target**: the Resource object (can be DataSourceResource or DestinationResource). Most important from it is the `options` attribute, which has the values for the target. This is important when you are editing a resource because that's where you get the current values.

Part of legacy behavior was to set target options from the `default` in the configurationSchema and render filled fields. I removed this instruction to use placeholders instead and that's where this happened :laughing:. And now I just took the method back with checkbox validation.

LMK if you need more details

arikfr:
👍

PR #3488, opened 2019-02-24
Migrated query edit/add param dialog to React/AntD/Hooks

ranbena:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix

## Description
Swapped UI modal and elements to AntD.
Swapped query selector for `<QuerySelector type="select">`.

Depends on #3494 and #3501. 

Before | After
------------ | -------------
screen shot 2019-02-27 at 15 28 32 | screen shot 2019-02-27 at 11 37 08
screen shot 2019-02-27 at 15 29 04 | screen shot 2019-02-27 at 11 39 30
screen shot 2019-02-27 at 15 29 16 | screen shot 2019-02-27 at 11 38 26
screen shot 2019-02-27 at 15 29 40 | screen shot 2019-02-27 at 11 38 49
screen shot 2019-02-27 at 15 29 53 | screen shot 2019-02-27 at 11 39 08
screen shot 2019-02-27 at 15 58 00 | screen shot 2019-02-27 at 15 59 59


client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -0,0 +1,204 @@
...
+            <Input.TextArea
+              rows={3}
+              value={param.enumOptions}
+              onChange={e => setParam({ ...param, enumOptions: e.target.value })}
+            />
+          </Form.Item>
+        )}
+        {param.type === 'query' && (
+          <Form.Item label="Query" help="Select query to load dropdown values from" {...formItemProps}>
+            <Select
```
arikfr: We have query search here, in "Add Widget Dialog" and in new alert page. Maybe time to have a <QuerySearch> component?

In the "Add Widget Dialog" we use a bit of a different UI for searching queries (input + list of results), but I don't see a reason not to switch to a Select element there as well.

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -0,0 +1,204 @@
...
+
+  // keep name in sync
+  useEffect(() => {
+    setParam({ ...param, name: newName });
+  }, [newName]);
+
+  const onSaveClicked = () => {
+    // update title to default
+    if (!param.title) {
+      param.title = getDefaultTitle(param.name);
```
arikfr: Why not preset this when constructing the `param` object?

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -0,0 +1,204 @@
...
+      setFeedback({ isError: true, helpText: 'Parameter with this name already exists' });
+    } else if (!name) {
+      setFeedback({ isError: false, helpText: 'Choose a keyword for this parameter' });
+    } else {
+      setFeedback({ isError: false, helpText: `This is what will be added to your query editor {{ ${name} }}` });
+    }
+  }, [name]);
+
+  return [feedback, name, setName];
+}
```
arikfr: While this is an interesting use of `useEffect` and a custom hook, I think a small component will result in a more readable code:

https://codesandbox.io/s/3yj38pv68q

In the example above `App` is your `EditParameterSettingsDialog`.

client/app/components/EditParameterSettingsDialog.jsx
arikfr: (btw, no need to introduce it to the other places now, just use it here for starters)

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -0,0 +1,204 @@
...
+      Query.get({ id: props.parameter.queryId }, (query) => {
+        setQuerySearchResults([query]);
+      });
+    }
+  }, []);
+
+  // keep name in sync
+  useEffect(() => {
+    setParam({ ...param, name: newName });
+  }, [newName]);
```
arikfr: In the CodeSandBox example above, I also removed the need for a separate state for the name, so not sure this is needed.

package.json
```diff
@@ -76,9 +76,9 @@
     "pivottable": "^2.15.0",
     "plotly.js": "1.41.3",
     "prop-types": "^15.6.1",
-    "react": "^16.3.2",
+    "react": "^16.8.3",
     "react-ace": "^6.1.0",
-    "react-dom": "^16.3.2",
+    "react-dom": "^16.8.3",
```
arikfr: Worth extracting the package changes to a separate PR, no?

client/app/components/EditParameterSettingsDialog.jsx
ranbena: This is the required behavior.


I determine "touched" by differentiating `param.title`'s value  between `null` (untouched) and string (touched). That's why I have to keep it as is until the very end.

package.json
ranbena: Sure. Must decide if we actually wanna tho.

arikfr: Is there a reason not to upgrade regardless of hooks?

client/app/components/EditParameterSettingsDialog.jsx
arikfr: 🤔 doesn't Ant have a facility of their own to disable on errors?

package.json
ranbena: No

client/app/components/EditParameterSettingsDialog.jsx
arikfr: 👍

client/app/components/EditParameterSettingsDialog.jsx
ranbena: Yes but I don't want to use. It's code mess like you won't believe.

arikfr: I guess we can introduce some callback from the input to the parent component, although this starts to become messy too.

The code mess is what you discussed with Gabriel?

arikfr: It's beyond the scope of this PR, so here let's resort to less ideal workarounds. But for the long term we should either embrace Ant's mess, implement some abstraction on top of it and implement our own abstraction. One way or another, no point in repeating this in every form, right?

ranbena: I built this form in a way that has only one "feedback" field to avoid complexity.
I do favor your component approach, much cleaner and more readable. I'll see if I can work around it.

arikfr: 👍 I updated the CodeSandbox to include a possible solution for this. It's not ideal, but...

ranbena: Ha! I'm working on the same solution.
Problem is that it works only if one component is allowed to set validity.

arikfr: Oh, right. New option, that actually cleaner but introduces mild code duplication:

```javascript
function okDisabled () {
  return isNew && (includes(existingsParams, param.name) || !param.name) || param.title === '';
}
```

We dup the existing params check, but we don't need to pass around onError.

kravets-levko: @arikfr You can create a function to validate name, use it in dialog and pass as prop to the name input component

ranbena: @arikfr this works for me
https://codesandbox.io/s/ry3q97l83p

ranbena: (updated PR)

arikfr: @kravets-levko that's an option as well.

@ranbena 👍

ranbena:
@arikfr the discussion I'd like us to have is - do we favor hooks for this component?
My take - while I love the functionality, the file content is unreadable, even as the author.
The predictable structure of classes (props, state, constructor, methods) and descriptive function names (looking at you "useEffect"..) make the whole difference. I would love to have these with hooks.

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -0,0 +1,214 @@
...
+
+function EditParameterSettingsDialog(props) {
+  const [param, setParam] = useState(clone(props.parameter));
+  const [querySearchResults, setQuerySearchResults, setQuerySearchTerm] = useQuerySearch();
+  const [saving, save] = useParamSave(props.onConfirm, props.dialog.close);
+  const [isNameValid, setIsNameValid] = useState(true);
+  const isNew = !props.parameter.name;
+
+  // get query name by id
+  useEffect(() => {
```
arikfr: I think this should be above `isNew`. I recall reading that all hooks should be defined first in the function.

arikfr: (there is an ESLint plugin for this)

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -0,0 +1,214 @@
...
+    if (param.queryId) {
+      Query.get({ id: props.parameter.queryId }, (query) => {
+        setQuerySearchResults([query]);
+      });
+    }
+  }, []);
```
arikfr: Shouldn't you pass `param.queryId` here? Otherwise this will be called _everytime_ when there is a queryId defined.

ranbena: Not sending a param ensures this runs only once, which is what we want.

ranbena: But I'll abolish it once I incorporate a `<QuerySearch defaultValue={param.queryId}>` component.

arikfr:
@ranbena 

I'm not sure this PR is a good one to judge hooks. There are some other issues here that make it messier than it should be and in turn makes you over use hooks for things we probably shouldn't. Can look again when you get rid of the search logic.

> descriptive function names (looking at you "useEffect"..) make the whole difference. I would love to have these with hooks.

I agree on this one. But we should keep in mind that we have a bias here as we were using cDU and cDM for ~5 years.

client/app/components/EditParameterSettingsDialog.jsx
arikfr: Apparently there is a difference between: `useEffect(() => {})` and `useEffect(() => {}, [])`. The first one will be called every time.

🤔

arikfr:
@ranbena we can implement our own `cDM` and `cDU` hooks as sugar on top of `useEffect` if we want to :-) (but we probably should not)

kravets-levko:
@arikfr I think @ranbena want to say that hooks are not a silver bullet, and they can make some code better, but at the same something looks better with classes. And we should just find a criteria when it's better to use hooks, and when it's better to stick with OOP-style components.

arikfr:
@kravets-levko there are no silver bullets. Everything has its trade offs. React itself has its cons and pros compared to the alternatives.

But we need to be pragmatic and good citizens of the ecosystem we're in. This means sometimes letting go of our personal preferences in favor of community standards.

kravets-levko:
@arikfr But hooks are not a community standard yet, as well as class components/HOCs are not deprecated yet. I think using a common sense is essential: as my granny said - if everyone will go to jump from the roof - will you join them too? :smile:

client/app/components/EditParameterSettingsDialog.jsx
gabrieldutra: > just implement our own abstraction

:+1: 

I'm not a very fan of Ant Form coding style either and I wouldn't say it's worth it when you have only one field with validation.

gabrieldutra:
> The predictable structure of classes (props, state, constructor, methods, render) and descriptive function names (looking at you "useEffect"..) make the whole difference. I would love to have these with hooks.

At first glance it did take me sometime to understand what's happening, the missing class methods hierarchy is noticed :sweat_smile:. However, @arikfr is right, you need to consider the bias, it may get sometime to have a quick look with hooks too.

> We should use tools appropriately to the task we solve: it hooks make the code unreadable, but classes can do it better - why should we fight with hooks? Also it works an opposite way: if we need a small component, say, with a single state field, - no need to introduce a class for it, hooks will be perfect. And so on.

That's it!  :+1: Classes will keep being a part of React, we will just have another option to choose :slightly_smiling_face:.

In the specific case of this component, I agree that it would be better as a Class

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -0,0 +1,214 @@
...
+
+
+  const onSaveClicked = () => {
+    // update title to default
+    if (!param.title) {
+      param.title = getDefaultTitle(param.name);
```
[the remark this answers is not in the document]
ranbena: That's not the reason :)
It's cause `param` doesn't yet update with `title` the moment save executed 😕
Need to either force update (not sure how) or just mutate directly which is what I did eventually.

ranbena:
@arikfr @kravets-levko this is the same PR from the start of week that you already reviewed and discussed only that I've taken parts out to #3494 and #3501.

One thing to notice is that since `<QuerySelector>` now expects an object in `selectedQuery` but this component has only the id, there is a cDM effect to fetch the query and dedicated state for it.
https://github.com/getredash/redash/blob/94af2d0d341eab0044e1fe36ac6b46916d4d467a/client/app/components/EditParameterSettingsDialog.jsx#L63-L69

ranbena:
Also, not saving the query on modal confirm. I'll implement the dirty indication in https://github.com/getredash/redash/pull/3448.

client/app/components/parameters.js
```diff
@@ -77,21 +34,26 @@ function ParametersDirective($location, $uibModal) {
...
-          resolve: {
-            parameter: param,
-          },
-        });
+      scope.showParameterSettings = (parameter, index) => {
+        EditParameterSettingsDialog
+          .showModal({ parameter })
+          .result.then((updated) => {
+            // update parameters
+            scope.parameters[index] = extend(parameter, updated);
```
arikfr: Why `extend` and not just override? Extending has the issue of leaving irrelevant settings from previous parameter type.

client/app/components/parameters.js
```diff
@@ -77,21 +34,26 @@ function ParametersDirective($location, $uibModal) {
...
+
+            // save if already dirty
+            // https://discuss.redash.io/t/query-unsaved-changes-indication/3302/5
+            if (!scope.queryIsDirty) {
```
arikfr: Maybe this component should receive an `onUpdate` callback and handle the logic of saving the query in the query component?

ranbena: Thought of that, but then figured that this code:
```js
.result.then((updated) => {
  // update parameters
  scope.parameters[index] = extend(parameter, updated);
  scope.updateQuery();
}
```
might mislead developers, who won't understand why sometimes it saves and sometimes doesn't. wdyt?

arikfr: Call it `onUpdate` and then it's clear that there is more logic in the handler, no?

client/app/components/parameters.js
ranbena: Hmm.. cause of this
```js
const [param, setParam] = useState(clone(props.parameter));`
```
`param` doesn't then hold the non-enumerable "urlParams" and it gets lost...

I'll try to fix this.

ranbena: Actually, it's not the clone. It's this the object spreading that loses it.
```js
setParm({ ...param, prop: newValue });
```

ranbena: @kravets-levko got any good idea to overcome this?

ranbena: > leaving irrelevant settings from previous parameter type

@arikfr That's happening either way since we don't reset irrelevant params anyhow.

arikfr: You're right :-(

kravets-levko: @ranbena You need to wrap raw object with `Parameter` constructor
https://github.com/getredash/redash/blob/83668a6840075eb979c9cb0aeae9a680153416ee/client/app/services/query.js#L51-L53
and then manually update `urlPrefix` prop using "old" parameter object. That's the only way to do it properly (`clone` will probably do it, except of updating `urlPrefix`; `extend`/spread operator will lose object type).

client/app/components/parameters.js
ranbena: Update in https://github.com/getredash/redash/pull/3488/commits/c9ab9e173b93a736285a29ec915d27083d69a888

client/app/components/parameters.js
ranbena: I say let's leave it as is - at least this way the component stays oblivious to `urlPrefix`.

arikfr: 👍 let's just open an issue.

PR #3494, opened 2019-02-25
 Added <QuerySelector /> component

ranbena:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix

## Description
Extracted the query selector functionality from AddWidgetDialog into an independent component, in order to use later somewhere else (https://github.com/getredash/redash/pull/3488).

### Changes
* Built with Hooks - looks good.
* Moved loading indicator into searchbar - less jumpy, more pleasant.
* Omitted the highlighting feature - feels unnecessary. More lines deleted :P


client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,103 @@
...
+  return Query.recent().$promise
+    .then((results) => {
+      const filteredResults = results.filter(item => !item.is_draft); // TODO: filter at the fetch level
+      return Promise.resolve(filteredResults);
+    });
+}
+
+function search(q) {
+  return new Promise((resolve) => {
+    Query.query({ q }, ({ results }) => {
```
arikfr: Why not use `$promise` here as well?

client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,103 @@
...
+        />
+      ) : (
+        <Input
+          placeholder="Search a query by name"
+          value={searchTerm}
+          onChange={e => setSearchTerm(e.target.value)}
+          suffix={<i className={cx('fa fa-spinner fa-pulse', { hidden: !searching })} />}
+        />
+      )}
+      <div className="scrollbox" style={{ maxHeight: '50vh', marginTop: 15 }}>
```
arikfr: Why not use `<Select>` here and show the results inside the select dropdown?

client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,103 @@
...
+    return () => {
+      clearTimeout(debounce);
+    };
+  }, [searchTerm]);
+
+  // on query selected
+  useEffect(() => {
+    const id = selectedQuery ? selectedQuery.id : null;
+    props.onChange(id);
+  }, [selectedQuery]);
```
arikfr: Feels like this should be in the `onClick` handler.

Another way to look at it: would you use `componentWillUpdate` for similar purpose?

client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,103 @@
...
+  const [searchResults, setSearchResults] = useState([]);
+  const [selectedQuery, setSelectedQuery] = useState();
+
+  // on search term changed, debounced
+  useEffect(() => {
+    setSearching(true);
+    const debounce = setTimeout(() => {
+      const promise = searchTerm ? () => search(searchTerm) : fetchRecent;
+      promise()
+        .then(setSearchResults)
```
arikfr: It's possible you will receive older search results before newer ones, so you need to check if this is the result of the current `searchTerm`.

client/app/components/QuerySelector.jsx
ranbena: lol didn't know it was avail 😅

client/app/components/QuerySelector.jsx
ranbena: That's the next PR:
```jsx
<QuerySelector type="select" />
```

client/app/components/QuerySelector.jsx
ranbena: Right, I put that part aside but forgot to re-introduce it.
I'd like to do this in a more standard way - `response.term` that would come from the API response. Wdyt?

ranbena: Another way to go about it is to disallow concurrent fetches.

ranbena: Or promise.cancel of course.

client/app/components/QuerySelector.jsx
arikfr: Why not implement only the select type? I'm not sure we need in AddWidgetDialog the other format, no?

client/app/components/QuerySelector.jsx
arikfr: promise.cancel: see the discussion I have with @kravets-levko about this on Slack.

`response.term`: 👍

client/app/components/QuerySelector.jsx
ranbena: Are there any other components intended for QuerySelector?

arikfr: You mean components that will use it?

The only ones I remember are:
1. New alert screen
2. Query parameter dialog
3. Add Widget dialog

client/.eslintrc.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
     'no-param-reassign': 0,
     'no-mixed-operators': 0,
     'no-underscore-dangle': 0,
+    "no-use-before-define": ["error", "nofunc"],
```
ranbena: Function hoisting can be trusted..
It's a must for useEffects to be at the top of functions.

client/app/components/QuerySelector.jsx
ranbena: @arikfr this PR's scope is only to extract functionality from AddWidgetDialog, not to change it.
I suggest we revisit this UX in the upcoming AddWidgetDialog STEPS PR.

client/.eslintrc.js
kravets-levko: I think `useEffect` should be not at the top of function - it could anywhere, you just need to keep an order of hooks.

ranbena: I generally like the hooked components to have this structure:
```
import X from 'x';

const VAR1 = ...

function utilityFunc1() {
}

function utilityFunc2() {
}

function Component() {
  const [state1, setState1] = userState();
  const [state2, setState2] = userState();

  const var1 = ...
  const var2 = ...

  useEffect(() => { // init
  }, []);

  useEffect(() => {
  }, [state1]);
  
  function method1() {
  }

  function method2() {
  }

  function renderPart1() {
  }

  function renderPart2() {
  }

  return (
    {renderPart1()}
    {renderPart2()}
    ...
  );
}

Comment (kravets-levko) [reply client/.eslintrc.js]: In this case I'd prefer to take as much as possible of that `method1`, `method2`, `renderPart1` function outside of the component (adding more arguments if needed).

Comment (kravets-levko) [reply client/.eslintrc.js]: BTW - with class components that functions were class methods, and probably entire component structure were cleaner (IMHO :slightly_smiling_face:)

Comment (ranbena) [reply client/.eslintrc.js]: > In this case I'd prefer to take as much as possible of that `method1`, `method2`

Yeah I keep in only what uses state in a way that would take too many arguments if it was out.

> component structure were cleaner

It's also my opinion but I wanna try to get the hang of it and keep discussing and perfecting.

**Comment by ranbena** [comment]
@arikfr @kravets-levko attended to comments. Ready for approval.

File: client/app/components/QuerySelector.jsx [hunk 6]
```diff
@@ -0,0 +1,149 @@
...
+  const debouncedSearch = debounce(_search, SEARCH_DEBOUNCE_DURATION);
+  const placeholder = 'Search a query by name';
+  const clearIcon = <i className="fa fa-times" onClick={() => setSelectedQuery(null)} />;
+  const spinIcon = <i className={cx('fa fa-spinner fa-pulse', { hidden: !searching })} />;
+
+  // set selected from prop
+  useEffect(() => {
+    if (props.selectedQuery) {
+      setSelectedQuery(props.selectedQuery);
+    }
```
arikfr: Why this is an effect and not just in the function body? Won't the function be called whenever the prop changes?

client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,149 @@
...
+    props.onChange(selectedQuery);
+  }, [selectedQuery]);
+
+  function _search(term) {
+    setSearching(true);
+    search(term)
+      .then(rejectStale)
+      .then(setSearchResults)
+      .finally(() => {
+        setSearching(false);
```
arikfr: This will be called even if the results were rejected, but the new search might still be running... I think `setSearching` should be called as a `then` and not `finally`.

client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,149 @@
...
+  function rejectStale(results) {
+    return isStaleSearch
+      ? Promise.reject(new Error('stale'))
+      : Promise.resolve(results);
+  }
+
+  function selectQuery(queryId) {
+    const query = find(searchResults, { id: queryId });
+    if (!query) { // shouldn't happen
+      toastr.error('Something went wrong.. Couldn\'t select query');
```
kravets-levko: Seems you didn't decide what to use - single dot or three dots, so used something between :smile:

client/app/components/QuerySelector.jsx
ranbena: Both Alert page and Query Param Edit set this prop async, so it initializes with `null` then updates to actual value.

Putting it in the body would reset state on every component change, which is not the desired behavior. As an effect you limit it only to when the prop changes.

client/app/components/QuerySelector.jsx
ranbena: Thought about that and it's true for rejection but not for any other error that might be thrown (e.g fetch error) - then it would keep going. I guess sth like this would cover it?
```js
search(term)
  .then(rejectStale)
  .then((results) => {
    setSearchResults(results);
    setSearching(false);
  })
  .catch((err) => {
    if (!(err instanceof StaleError)) {
      setSearching(false);
    }
  })
```

client/app/components/QuerySelector.jsx
kravets-levko: @arikfr `componentWillUpdate` is deprecated, but `useEffect` is a replacement for `componentDidUpdate` - so if you doubt you'll use cDU for this - keep in mind that `useEffect` _is_ cDU

ranbena: Since `<Select>` doesn't allow an object as an option value, the query.id is the value which is sent to `selectQuery`

```jsx
<select onChange={selectQuery}>
  <option value={query.id}>{query.name}</option>
</select>
```
which uses the id to find the query object and updates the state.
```js
function selectQuery(queryId) {
  const query = find(searchResults, { id: queryId });
  if (!query) { // shouldn't happen
    toastr.error('Something went wrong.. Couldn\'t select query');
  }
  setSelectedQuery(query);
}
```

An effect then reacts to the state change:
```js
useEffect(() => {
  setSearchTerm(selectedQuery ? null : ''); // empty string forces recent fetch
  props.onChange(selectedQuery);
}, [selectedQuery]);
```

ranbena: (The above is code from #3501)

arikfr: I probably miss something, but why not call `setSearchTerm` and `props.onChange` in `selectQuery`?

client/app/components/QuerySelector.jsx
arikfr: I guess...

client/app/components/QuerySelector.jsx
ranbena: Cause I wanted clearing it to use only `setSelectedQuery`.
https://github.com/getredash/redash/blob/0d209cb6dbd44195ceaf33bac56c684e061f6def/client/app/components/QuerySelector.jsx#L36

client/app/components/QuerySelector.jsx
ranbena: That's exactly how I feel about it. I tend to go for a simple approach if the UX doesn't suffer that much. Whatever you choose.

arikfr: It can't stay the way it is now, because in some cases it will set it to not searching while in fact it does.

The other option is not to reject stale results (but just ignore them). I don't know what the implications of leaving pending promises though.

client/app/components/QuerySelector.jsx
arikfr: Putting a condition in `selectQuery` feels like a cleaner approach, because what this handler in fact does is to select a query. Now you have this logic split into places: one in the handlers (selectQuery and this anonymous one) and some watcher.

client/app/components/QuerySelector.jsx
ranbena: Done.
https://github.com/getredash/redash/blob/cd9f1e0d92fe92d4c6d09176e3fd497f44bff4c6/client/app/components/QuerySelector.jsx#L70-L80

client/app/components/QuerySelector.jsx
ranbena: Done.
https://github.com/getredash/redash/blob/cd9f1e0d92fe92d4c6d09176e3fd497f44bff4c6/client/app/components/QuerySelector.jsx#L89-L101

client/app/components/QuerySelector.jsx
ranbena: Fixed!

client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,160 @@
...
+    return (
+      <div className="list-group">
+        {searchResults.map(q => (
+          <a
+            href="javascript:void(0)"
+            className={cx('list-group-item', { inactive: q.is_draft })}
+            key={q.id}
+            onClick={() => selectQuery(q.id)}
+          >
+            {q.name}
```
kravets-levko: BTW - how about highlighting search phrase (https://github.com/getredash/redash/pull/3494/files#diff-9487bf74e60479e11fa384f7e6ca248cL243) ?

client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,160 @@
...
+  function selectQuery(queryId) {
+    let query = null;
+    if (queryId) {
+      query = find(searchResults, { id: queryId });
+      if (!query) { // shouldn't happen
+        toastr.error('Something went wrong... Couldn\'t select query');
+      }
+    }
+
+    setSearchTerm(query ? null : ''); // empty string triggers recent fetch
```
kravets-levko: It took some time for me to understand how it works, and then I realized that actually it's AngularJS-way to handle data flow: you set a state field, but setter just updates value; and there is watcher (`$scope.$watch`) that does something when that field changes. That's not something I personally expect to see in React code :thinking:

client/app/components/QuerySelector.jsx
ranbena: I mentioned in the PR desc that I omitted it intentionally cause it just seems pointless to me.

kravets-levko: Then, I think, you can remove `@/lib/highlight` helper - as I remember, it's the only place where it's used (more lines to delete! :smile:)

kravets-levko: _(but anyway check before removing - I may be wrong with "the only place")_

client/app/components/QuerySelector.jsx
ranbena: I do it rarely when it makes sense. But now reconsidering more since useEffect is so convenient. Will stay away from it if you feel it's an anti-pattern.

client/app/components/QuerySelector.jsx
ranbena: Looks safe. Removed.

client/app/components/QuerySelector.jsx
kravets-levko: > Will stay away from it if you feel it's an anti-pattern.

Yes, I think so. That's what makes AngularJS code hard to understand; it's much easier to understand what's going on when you see: "ah, this function updates state variable, but also it loads queries", etc. With such side effects you see: "this function updates state. okay." - and then you see that oh magic! it somehow loads data! whaaat?? And you need to find another place in the code which load query results. And BTW - in AngularJS this mechanism is a bit better because you describe dependency first, and then write handler. With `useEffect` if's really easy to miss dependencies (what I did, actually). Compare:
```javascript
$scope.$watch('field', () => { 
  ...
  // a lot of code
  ...
});

useEffect(() => {
  ...
  // a lot of code
  ...
}, [field]); // <- this line left unattended
```

I really this is anti-pattern; probably one of the things I most liked in React - strict data flows, without implicit dependencies.

arikfr: Interesting observation.

kravets-levko: _(Read that code once more and have something to add)_

Another con of this approach is harder data flow control. Did you notice that effect that loads search results could be optimized? When search term is `null` - you already have search results (empty list). If you have setter for search term - you can handle it immediately:

1. user clicks "clear" button - search term set to `null`, and search results to `null`;
2. component renders because state changed.

 but with `useEffect` it will:

1. user clicks "clear" button - search term set to null;
2. component renders because state changed;
3. side effect executed, it sees that search term is `null` and updates search results to `null`
4. component renders because state changed.

ranbena:
> @ranbena I approve this, but please take into consideration my comment about Angular-style code. It's very easy to start writing spaghetti-code again, but this time in react - so why needed to migrate codebase at all? 🙂

I think I'm missing something here - is this code still problematic?

client/app/components/QuerySelector.jsx
```diff
@@ -0,0 +1,160 @@
...
+      return () => {};
+    }
+
+    // search
+    debouncedSearch(searchTerm);
+    return () => {
+      debouncedSearch.cancel();
+      isStaleSearch = true;
+    };
+  }, [searchTerm]);
```
kravets-levko: @ranbena Look: you still load search results as a side-effect of changing search term. It's implicit dependency, that's what makes code hard to read and understand:
```javascript
$scope.setSearchTerm = (searchTerm) => {
  $scope.searchTerm = searchTerm;
};

// somewhere else

$scope.$watch('searchTerm', (searchTerm) => {
  if (searchTerm == '') {
    $scope.results = [];
  } else {
    doSearch(searchTerm);
  }
});
```
in opposite to explicit reaction, where you immediately see that if you change search term - it will do some extra stuff:
```javascript
$scope.setSearchTerm = (searchTerm) => {
  $scope.searchTerm = searchTerm;
  if (searchTerm == '') {
    $scope.results = [];
  } else {
    doSearch(searchTerm);
  }
};
```
I intentionally provide examples in Angular because this pattern is really common there, and one of best practices in React for a long time was to avoid such code.

ranbena: Since this useEffect also allows easy stale/debounce management (the alternative I assume is having dedicated refs), I'd rather keep it and keep moving 🏃

PR #3495, opened 2019-02-26
Sharing embeds with safe parameters

rauchy:

- [x] Feature

## Description
This PR points embeds are the textless endpoint, resulting in the ability to execute queries with safe parameters in embedded visualizations.

Towards #3284 

feb-27-2019 14-15-05


client/app/components/queries/visualization-embed.js
```diff
@@ -37,7 +37,7 @@ export default function init(ngModule) {
     return session($http, $route, Auth).then(() => {
       const queryId = $route.current.params.queryId;
       const query = $http.get(`api/queries/${queryId}`).then(response => response.data);
-      const queryResult = $http.get(`api/queries/${queryId}/results.json${location.search}`).then(response => response.data);
+      const queryResult = $http.post(`api/queries/${queryId}/results${location.search}`).then(response => response.data);
```
kravets-levko: It looks like semantic flaw: HTTP POST should be used to update something, and GET - to retrieve data. But here you use POST to fetch query results.

rauchy: Well, it has a side-effect - it modifies system state by creating a new record in the query_results table, so you can look at it as updating something.

kravets-levko: Then makes sense. But looks weird anyway :slightly_smiling_face:

redash/handlers/query_results.py
```diff
@@ -89,8 +89,15 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
     if query_result:
         return {'query_result': query_result.to_dict()}
     else:
-        job = enqueue_query(query.text, data_source, current_user.id, metadata={
-            "Username": current_user.email,
+        if current_user.is_api_user:
```
rauchy: @arikfr any thoughts on this?

redash/handlers/query_results.py
```diff
@@ -89,8 +43,15 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
...
+            id = None
+            name = repr(current_user)
+        else:
+            id = current_user.id
+            name = current_user.email
+
+        job = enqueue_query(query.text, data_source, id, metadata={
```
arikfr: 1. `id` is a reserved word.
2. We use the user id for the Query Results query runner. I think we need to pass in information about the API key for everything to work smoothly.

client/app/components/queries/embed-code-dialog.js
```diff
@@ -9,7 +9,7 @@ const EmbedCodeDialog = {
 
     this.embedUrl = `${clientConfig.basePath}embed/query/${this.query.id}/visualization/${
       this.visualization.id
-    }?api_key=${this.query.api_key}`;
+    }?api_key=${this.query.api_key}&${location.search.substring(1)}`;
```
arikfr: I think parameters have a method for serializing themselves into a query string. Feels like a better option than this.

rauchy: Are you referring to [this](https://github.com/getredash/redash/blob/master/client/app/components/parameters.js#L19)? If so, feels too coupled to the directive's scope, no?

arikfr: Nope, this:

https://github.com/getredash/redash/blob/ade3cc72a70d2c65b6f8e393304f4a394f61e926/client/app/services/query.js#L165

kravets-levko: @arikfr It will return an object, not a string. Of course, it's not an issue to merge that params as we do in `Parameters` component (see the link @rauchy posted above), but we don't have any library to serialize them into query string (`$location.search(params)` does that job, but it does not expose any API).

redash/handlers/query_results.py
rauchy: 1. 2692aa1
2. That's solid. 6ceac2e should address this. I'd appreciate if you could give it a close look as it feels like a sensitive area. Also, are there any other query runners who need `user`?

client/app/pages/home/home.html
```diff
@@ -1,5 +1,8 @@
 <div class="container">
-  <div ng-if="!$ctrl.isEmailVerified" class="alert alert-warning">
+  <div ng-if="$ctrl.messages.includes('using-deprecated-embed-feature')" class="alert alert-warning">
+      You have enabled <code>ALLOW_PARAMETERS_IN_EMBEDS</code> in <code>redash/settings/__init__.py</code>. This setting is now deprecated and supported by default and can be turned off. <a href="https://discuss.redash.io/t/support-for-parameters-in-embedded-visualizations/3337" target="_blank">Read more</a>.
```
arikfr:
```suggestion
      You have enabled <code>ALLOW_PARAMETERS_IN_EMBEDS</code>. This setting is now deprecated as parameters in embeds supported by default and should be turned off. <a href="https://discuss.redash.io/t/support-for-parameters-in-embedded-visualizations/3337" target="_blank">Read more</a>.
```

Maybe we should mention something about string parameters?

rauchy: A bit of a problematic way to phrase it. 
> This setting is now deprecated as parameters in embeds *are* supported by default and should be turned off.

Sounds a bit like "parameters in embeds" should be turned off.

rauchy: > Maybe we should mention something about string parameters?

Too detailed and subject to change IMO. Also that's mentioned in the forum post.

ranbena:
@rauchy this should affect the feature introduced in https://github.com/getredash/redash/pull/3439, yeah?

rauchy:
@ranbena not quite, this one's for embedded visualizations only. The next PR will, though.

HenroRitchie:
@rauchy I just tested this and I am struggling. Are there specific steps to implement to enable the parameters to be available in embeds?

rauchy:
The only restriction is that the query you are trying to share cannot have
parameters of type “text”. All other types are allowed.

client/app/components/queries/embed-code-dialog.js
arikfr: Ping :)

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -92,54 +130,162 @@ describe('Dashboard', () => {
...
   });
 
-
   describe('Textbox', () => {
-    before(function () {
-      cy.login();
-      createNewDashboardByAPI('Foo Bar')
-        .then(slug => `/dashboard/${slug}`)
-        .as('dashboardUrl');
+    beforeEach(function () {
```
arikfr: I think you have some changes in this PR not related to it like this one, right?

cypress/integration/embed/share_embed_spec.js
```diff
@@ -0,0 +1,22 @@
+describe('Embedded Queries', () => {
+  beforeEach(() => {
+    cy.login();
+    cy.visit('/queries/new');
+  });
+
+  it('are shared with text parameters', () => {
```
arikfr: I'm not sure what this is testing?

redash/handlers/queries.py
```diff
@@ -179,7 +180,12 @@ def require_access_to_dropdown_queries(user, query_def):
         groups = models.Query.all_groups_for_query_ids(dropdown_query_ids)
 
         if len(groups) < len(dropdown_query_ids):
-            abort(400, message='You are trying to associate a dropdown query that does not have a matching group. Please verify the dropdown query id you are trying to associate with this query.')
+            abort(400, message="You are trying to associate a dropdown query that does not have a matching group."
+                               "Please verify the dropdown query id you are trying to associate with this query.")
+
+        object = lambda: None
+        object.groups = dict(groups)
+        require_access(object, user, view_only)
```
arikfr: What's this?

redash/handlers/query_results.py
```diff
@@ -89,8 +43,8 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
     if query_result:
         return {'query_result': query_result.to_dict()}
     else:
-        job = enqueue_query(query.text, data_source, current_user.id, metadata={
-            "Username": current_user.email,
+        job = enqueue_query(query.text, data_source, current_user.id, current_user.is_api_user(), metadata={
+            "Username": repr(current_user) if current_user.is_api_user() else current_user.email,
```
arikfr: What `repr(current_user)` returns? As this is used in logging we need to make sure we don't pass the actual API key there.

client/app/components/queries/visualization-embed.js
arikfr: Why are we passing parameters in the query string and not in the request body?

redash/handlers/query_results.py
```diff
@@ -201,8 +157,13 @@ def post(self, query_id):
...
                                 always execute.
         """
-        params = request.get_json(force=True)
-        parameters = params.get('parameters', {})
+        if request.is_json:
+            params = request.get_json(force=True)
+            parameter_values = params.get('parameters')
+        else:
+            params = request.args
+            parameter_values = collect_parameters_from_request(params)
```
arikfr: Are you sure we _have_ to support this for this API?

redash/models/parameterized_query.py
```diff
@@ -142,5 +153,6 @@ def text(self):
 
 class InvalidParameterError(Exception):
     def __init__(self, parameters):
-        message = u"The following parameter values are incompatible with their definitions: {}".format(", ".join(parameters))
+        parameter_names = ", ".join(parameters)
```
arikfr:
```suggestion
        parameter_names = u", ".join(parameters)
```

redash/tasks/queries.py
```diff
@@ -130,14 +130,15 @@ def enqueue_query(query, data_source, user_id, scheduled_query=None, metadata={}
...
+                args = (query, data_source.id, metadata, user_id, is_api_key, scheduled_query_id)
                 argsrepr = json_dumps({
                     'org_id': data_source.org_id,
                     'data_source_id': data_source.id,
                     'enqueue_time': time.time(),
                     'scheduled': scheduled_query_id is not None,
                     'query_id': metadata.get('Query ID'),
-                    'user_id': user_id
+                    'user_id': user_id,
+                    'is_api_key': is_api_key
```
arikfr: We don't need this here, as we didn't update the monitoring code to use it (and probably won't need to).

cypress/integration/dashboard/dashboard_spec.js
ranbena: Should rebase.

cypress/integration/embed/share_embed_spec.js
ranbena: Yeah it's missing an `expect` after `cy.visit(embedUrl);`

HenroRitchie:
@rauchy I've updated to 7.0.0.b18042 and attempted to embed both a query and a dashboard with this query into an external site. The users are not logged into redash and the query uses a dropdown and a date range selector. In the embed neither of the two parameters are showing. Firstly, am I running the right version and secondly are their any specific configuration I need to do to enable this?

rauchy:
@HenroRitchie this isn't part of the v7.0.0 release. In order to use this feature you should checkout this branch (`allow-parameters-on-embeds`), or wait for it to be merged to master, hopefully in the next few days.

Second - this will currently only work for embedded queries. Dashboards will be addressed next.

Other than that - the parameters you describe are good to go for sharing.

redash/handlers/queries.py
rauchy: 😱

cypress/integration/embed/share_embed_spec.js
rauchy: I guess I was treating it more like a smoke test because it has previously exploded with a 500 when shared. Since there's nothing specific I want to expect here, but merely a "I got here and it shows data", I guess I should snap a Percy?

HenroRitchie:
@rauchy Thank you. I knew something obvious was a miss. I will wait for it to be merged with the Master Branch.

cypress/integration/embed/share_embed_spec.js
arikfr: Shouldn't this explode because it uses a text parameter?

ranbena: > I got here and it shows data

So the test is fine as it would fail by default upon status code 500.

**But** if it _should_ fail, you might need to do something like this:
```js
cy.server()
cy.route(embedUrl).as('embedRequest'); // listen to redirect
cy.visit(embedUrl, { failOnStatusCode: false }); // prevent 500 from failing test
cy.wait('@embedRequest').then((xhr) => {
  // expect xhr status code to be 500
});
```

redash/handlers/query_results.py
rauchy: It's [indeed the API key](https://github.com/getredash/redash/blob/master/redash/models/users.py#L375). What other way do suggest we identify an api user otherwise in the log?

client/app/components/queries/visualization-embed.js
rauchy: [No apparent reason](https://github.com/getredash/redash/pull/3495/commits/818fc27bf028992d488156fdf1178e6a6e915a78).

redash/handlers/query_results.py
rauchy: No, it's gone.

client/app/components/queries/visualization-embed.js
```diff
@@ -22,6 +22,9 @@ const VisualizationEmbed = {
   },
 };
 
+const queryStringAsObject = () => location.search.slice(1).split('&').map(p => p.split('='))
+  .reduce((obj, [key, value]) => ({ ...obj, [key.replace(/^p_/, '')]: value }), {});
```
arikfr: 😱 
We already have code in that takes parameters from the query string, we should reuse it.

redash/handlers/query_results.py
arikfr: Actually it's not. It's either the object id in case of the `ApiKey` model:

https://github.com/getredash/redash/blob/ec4f77c8b7fbcb4b04821e98bc12a0f994a6b16e/redash/models/users.py#L375

(notice it's `api_key.id` and not `api_key.api_key`)

... or the query id in case of a query API key:

https://github.com/getredash/redash/blob/ec4f77c8b7fbcb4b04821e98bc12a0f994a6b16e/redash/authentication/__init__.py#L137

So all good here.

redash/models/parameterized_query.py
arikfr: It doesn't look like you applied the suggestion?

redash/tasks/queries.py
```diff
@@ -278,20 +278,30 @@ class QueryExecutionError(Exception):
     pass
 
 
+def _resolve_user(user_id, is_api_key):
+    if user_id is not None:
+        if is_api_key:
+            api_key = user_id
+            q = models.Query.query.filter(models.Query.api_key == api_key).one()
```
arikfr: Let's try to avoid using `query.filter` outside of `redash.models`. This way we can easily see what's the public surface of a model.

redash/tasks/queries.py
```diff
@@ -278,20 +278,30 @@ class QueryExecutionError(Exception):
...
+            return models.ApiUser(api_key, q.org, q.groups)
+        else:
+            return models.User.query.get(user_id)
```
arikfr: Same here (we probably have a `#get_by_id` or `#get_by_id_and_org` on this model anyway).

redash/tasks/queries.py
```diff
@@ -380,11 +390,11 @@ def _load_data_source(self):
 # user_id is added last as a keyword argument for backward compatability -- to support executing previously submitted
 # jobs before the upgrade to this version.
 @celery.task(name="redash.tasks.execute_query", bind=True, track_started=True)
-def execute_query(self, query, data_source_id, metadata, user_id=None,
+def execute_query(self, query, data_source_id, metadata, user_id=None, is_api_key=False,
```
arikfr: Need to make sure that this is backward compatible with old messages that don't have `is_api_key` (i.e. is Celery really passing those as keyword args, so the order won't matter).

Also it probably makes sense to convert this thing into a dictionary so we don't have to update the sig everytime, but this is for a future refactor.

client/app/components/queries/visualization-embed.js
rauchy: Could you help me locate the piece of code you are mentioning? I did find similar behavior when constructing a [Parameters](https://github.com/getredash/redash/blob/master/client/app/services/query.js#L214) instance, but that is also bound to the actual query text, and doesn't seem to have the built-in ability to convert to a simple key-value JSON.

redash/tasks/queries.py
rauchy: 161d06a2

redash/tasks/queries.py
rauchy: 34404b44

redash/tasks/queries.py
rauchy: 5770888

cypress/integration/embed/share_embed_spec.js
rauchy: Not sure I follow on the xhr stuff here. I'm actually visiting a page here.
Anyway, I changed the existing test to snap a Percy and added another one that doesn't fail on 403 and snaps an additional Percy.

rauchy:
> Why is { force: true } needed?
@ranbena - the anchor tag is not actionable at that point (it's hidden away), so `{ force: true }` clicks it anyway.

client/app/components/queries/embed-code-dialog.js
rauchy: @arikfr @kravets-levko Pardon [my NIH](https://github.com/getredash/redash/pull/3495/commits/17636ec6965336ce5af6f6620ca4f02bbcbe341d). I couldn't find a way to use $location on _other_ URLs so extending it with `Parameter.toUrlParams()` was of no help. LMK what you think of this.

redash/tasks/queries.py
rauchy: It actually doesn't 👎 
Is 16a8a404 a sufficient remedy?

redash/models/parameterized_query.py
arikfr: Ping.

rauchy: Somehow missed that one!

arikfr:
🚀

HenroRitchie:
Hi Guys - thank you for the merge. What is the quickest way to get this into production? I usually pull the docker images from docker hub, but I guess this will not be there yet. Should I pull the preview tag?

arikfr:
@HenroRitchie you can use the `redash/preview` Docker repo and use a specific tag from there (once there is a master build with this PR).

HenroRitchie:
@arikfr Thank you. Will just wait for the master build with this PR.

HenroRitchie:
Hi Guys - I have attempted to embed a query into another page but I get an error response `It seems like we encountered an error. Try refreshing this page or contact your administrator.` 

When embedding a dashboard it still works fine, without the parameters being available. Looking at the developer console in chrome the response from Redash is a 400 Bad Request. Should I continue the discussion here, or somewhere else?

rauchy:
@HenroRitchie generally we'd prefer support discussions at [the forums](https://discuss.redash.io) but since this is a new feature it may be related directly to this ticket.

A 400 here could probably mean an incorrect parameter type passed in - could you check and also send over the response message that came along with that 400?

HenroRitchie:
@rauchy I will continue here until we are sure my problem and this PR are unrelated. 

I used the embed `Embed elsewhere` link to generate the iframe code. When I try to embed a query without any parameters this is the iframe code generated:
```
<iframe src="{{ $ctrl.embedUrl }}" width="720" height="391"></iframe>
```
This is clearly wrong.  When I try to embed a query with parameters and use the `Embed elsewhere` link this is the result:
```
<iframe src="https://url/embed/query/1/visualization/2?api_key=xd8wfBdBxsW8rB1jGxIgDySQhkQGBvLKfpiMPx9J&p_fridge_tag=fridge_1_top&p_date_range.start=2019-02-21%2000%3A00%3A00&p_date_range.end=2019-02-26%2000%3A00%3A00" width="720" height="391"></iframe>
```
If I just use the generated URI I get the same error response posted earlier.

The response message is: `"message": "The following parameter values are incompatible with their definitions: date_range.start, date_range.end"}`. I thus removed the date range parameter and now the embed works. However, when I select a different option in the embed the chart is not updated correctly. The chart defaults back to the initial setup, i.e. x-axis from -1 to 6 and y-axis from -1 to 4 without any data. Is the query automatically refreshed/run when the parameter value changed? Is looks like this is not the case. I first must run the query directly in Redash with the corresponding parameter selected before in will show correctly in the embed.

client/app/components/queries/visualization-embed.js
rauchy: Yeah, I guess we need to consolidate this. As a quick fix there's [this](https://github.com/getredash/redash/pull/3681/files), but we should really have this logic in one place. Could you let me know which piece of code you were mentioning in this comment?

rauchy:
@HenroRitchie thanks for the details. You've helped me find 2 bugs which are fixed [here](https://github.com/getredash/redash/pull/3680) and [here](https://github.com/getredash/redash/pull/3681). I'm assuming they will get merged today and I'll be looking to hear if they solved your problem.

HenroRitchie:
@rauchy Thank you for the updates and commits. I am happy to report that I can now embed data parameters successfully. However, the queries are still not refreshed when any of the parameters change. In my test, I have one drop-down parameter and one date range parameter. I followed the steps below to test.

In Redash select the first parameter and the date range and run the query. I manually run the query every time. Check the results in Redash. Embed the query in the secondary portal as an Iframe. This works and provides the same results as Redash directly. If I now select a different parameter from the dropdown or change the date range the chart display defaults to the generic chart. Furthermore, if I attempt to download the CSV file it fails with a server error. 

I monitored the Celery Status page - and it does seem that something is happening as soon as I change the parameter, but the page and chart are never updated. I don't know if the page is not updated with the latest info or if takes too long. I have a large number of data to display, and changing the first parameter might trigger the query but it takes such a long time that it never gets updated on the display. Visual feedback to show the query is being refreshed might help here.

Changing the refresh schedule to every 1 minute does not make a difference.

Lastly, am I approaching this correctly? Should the parameter change a variable in the SQL query, or should the variable filter the results? The difference is the amount of data sent from Postgres to Redash. Running the same query everytime might be better for Redash, or will it not make a difference?

HenroRitchie:
@rauchy **Update**

If parameter 1 option 1 is selected with a specific date range nothing happens.
If parameter 1 option 2 is selected with the same date range nothing happens.
If parameter 1 option 1 is selected with the same date range the correct chart is now available.
If parameter 1 option 2 is selected with the same date range the correct chart is available. 
If parameter 1 option 3 is selected with the same date range nothing happens.
If parameter 1 option 1 is selected, and then option 3 again the chart is available.

ghoshsanjoy78:
Does this PR also cover embedding dashboards with parameters. I am now able to embed visualizations with params but dashboards with params are still not supported?

Issue #3280, opened 2019-01-14
TypeError raised on Celery exception 


jezdez:
### Issue Summary

As part of the API to get the [status of a job](https://github.com/getredash/redash/blob/4c9326a9da392b6b1e0650f532552c7c77147b1c/redash/handlers/query_results.py#L284-L290) we've seen increased number of TypeError raised when the query for the job in question fails (for an unrelated reason) with an sqlalchemy.exc.IntegrityError exception. The cause for that exception is not clear (possibly fork related), but the purpose of this ticket is to investigate the way errors are handled in general and the cause of the error in not relevant per se.

https://github.com/celery/celery/issues/5057 describes the result pretty well, but here's our traceback: 

```exception
TypeError: __init__() takes at least 4 arguments (2 given)
  File "flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "flask_restful/__init__.py", line 477, in wrapper
    resp = resource(*args, **kwargs)
  File "flask_login/utils.py", line 228, in decorated_view
    return func(*args, **kwargs)
  File "flask/views.py", line 84, in view
    return self.dispatch_request(*args, **kwargs)
  File "redash/handlers/base.py", line 31, in dispatch_request
    return super(BaseResource, self).dispatch_request(*args, **kwargs)
  File "flask_restful/__init__.py", line 587, in dispatch_request
    resp = meth(*args, **kwargs)
  File "redash/handlers/query_results.py", line 297, in get
    return {'job': job.to_dict()}
  File "redash/tasks/queries.py", line 159, in to_dict
    task_info = self._async_result._get_task_meta()
  File "celery/result.py", line 410, in _get_task_meta
    return self._maybe_set_cache(self.backend.get_task_meta(self.id))
  File "celery/backends/base.py", line 365, in get_task_meta
    meta = self._get_task_meta_for(task_id)
  File "celery/backends/base.py", line 680, in _get_task_meta_for
    return self.decode_result(meta)
  File "celery/backends/base.py", line 284, in decode_result
    return self.meta_from_decoded(self.decode(payload))
  File "celery/backends/base.py", line 280, in meta_from_decoded
    meta['result'] = self.exception_to_python(meta['result'])
  File "celery/backends/base.py", line 260, in exception_to_python
[...]

PR #3498, opened 2019-02-26
Unwrap SQLAlchemy errors in worker

washort:
Due to poor interaction between Celery and SQLAlchemy's exception objects, an unwrapped version of DB errors is raised here to properly signal failure of the task.

Refs: #3280, https://github.com/celery/celery/issues/5057


washort:
This is unnecessary, there's a simpler way to prevent this problem.

PR #3499, opened 2019-02-26
Fail query task properly even if error message is empty

washort:
Query runners may fail with empty error messages; this was not handled as a failure and so `store_result` was invoked with `data=None`. This makes the check for error conditions explicit. (This was causing https://github.com/getredash/redash/issues/3280 )

### DIFF
--- a/redash/tasks/queries.py
+++ b/redash/tasks/queries.py
@@ -323,7 +323,7 @@ class QueryExecutor(object):
 
         _unlock(self.query_hash, self.data_source.id)
 
-        if error:
+        if error is not None and data is None:
             result = QueryExecutionError(error)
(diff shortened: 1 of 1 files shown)


redash/tasks/queries.py
```diff
@@ -479,7 +479,7 @@ def run(self):
 
         _unlock(self.query_hash, self.data_source.id)
 
-        if error:
+        if error is not None:
```
arikfr:
```suggestion
        if error is not None and data is None:
```

Maybe also make sure `data` is `None`?

washort: done

washort:
Rebased on master.

arikfr:
Thanks.

PR #3501, opened 2019-02-27
QuerySelector in Alert page

ranbena:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix

## Description
Swapped query selector in Alerts page for `<QuerySelector>`.
Added QuerySelector feature `type="select"` which renders a dropdown. 

Depends on #3494


client/app/pages/alert/index.js
```diff
@@ -11,6 +11,7 @@ function AlertCtrl($routeParams, $location, $sce, toastr, currentUser, Query, Ev
   this.trustAsHtml = html => $sce.trustAsHtml(html);
 
   this.onQuerySelected = (item) => {
+    this.alert.query = item;
```
ranbena: Right way to do it?

client/app/components/QuerySelector.jsx
```diff
@@ -116,6 +119,30 @@ export function QuerySelector(props) {
...
+        value={value || undefined} // undefined for the placeholder to show
+        onSearch={setSearchTerm}
+        onChange={selectQuery}
+        suffixIcon={searching ? spinIcon : suffixIcon}
+        notFoundContent={null}
+        filterOption={false}
+        defaultActiveFirstOption={false}
+      >
+        {searchResults && searchResults.map(q => (
+          <Option value={q.id} key={q.id}>{q.name}</Option>
```
arikfr: Can option render "richer" content?

kravets-levko: @arikfr Yes, it can render anything

kravets-levko: _(even `<iframe>` I think :smile: )_

client/app/pages/alert/index.js
kravets-levko: It's Angular code, so it's okay. But if this function will be called from React component - ensure you call `$scope.$applyAsync()`

ranbena: It does but everything works ok 🤷‍♂️

client/app/components/QuerySelector.jsx
ranbena: I'll try the tags

arikfr: Then we probably should show the same info as in the non `select` version?

kravets-levko: Also in Add widget dialog there was a feature (IIRC) when it shows unpublished queries too, but does not allow to select them - I think `<Option disabled>` is perfect for this.

ranbena: Yeah I missed that completely. On it.

client/app/pages/alert/index.js
kravets-levko: I think it works because you use Angular services later, and they trigger $digest at some point (I think, when `$q` promise resolves). But IMO to be 100% sure - it's better to explicitly call $applyAsync (because if that's promise - then data update will be delayed, and sometimes it may be noticeable).

client/app/components/QuerySelector.jsx
ranbena: Fixed.
screen shot 2019-02-27 at 12 44 02

kravets-levko: Looks really nice :+1:

client/app/pages/alert/index.js
ranbena: Done.

PR #3502, opened 2019-02-27
Fix verification_email endpoint when in MULTI_ORG mode

### LINKED ISSUE #3475: Show an indication when email verification sending fails
When clicking on "Resend email" in the verification email notification, there is no feedback to the user that anything happened when there is a failure.

rauchy:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix

Fixes #3475


client/app/pages/home/index.js
```diff
@@ -18,7 +18,8 @@ function HomeCtrl(Events, Dashboard, Query, $http, currentUser, toastr) {
   });
 
   this.verifyEmail = () => {
-    $http.post('/verification_email/').success(({ message }) => {
+    const slug = currentUser.org_slug ? `/${currentUser.org_slug}` : '';
+    $http.post(`${slug}/verification_email/`).success(({ message }) => {
```
arikfr: Why not just use `verification_email/` and let the `<base href>` meta tag do its thing?

rauchy: 🤦‍♂

Issue #3071, opened 2018-11-13
React migration


arikfr:
## Status

Type | Description | Status | Comment
-|-|-|-
Page | Query View | #4429
Page | Query Editor | #4429
Component | `AppView` | #4525 | (layout component, contains logic of redirecting to login, rendering the header and some more)
lib | Routing | #4159 | 
lib | State Management | | Not 100% sure we need this from the get go
lib | Model Classes | | Mostly the `$http`/`$resource` based objects

---

[As announced in July](https://discuss.redash.io/t/react-migration-an-experiment-with-less-di/2077), we started the migration of our frontend code to React:

> It’s long overdue that we migrate from Angular 1. Yet an actual migration was never prioritized, because it doesn’t directly benefit the end user so we preferred to focus on things that actually benefit the user.
>
> But opportunity presented itself, when the good folks from Mozilla decided to experiment with migration Redash from Angular to React. While personally, for a new project, I wouldn’t necessarily pick React (but rather Vue), for an existing code base React presented the following advantages:
>
> * it had a clearer path to doing a gradual migration using the [react2angular](https://github.com/coatue-oss/react2angular) library (and later the [angular2react](https://github.com/coatue-oss/angular2react) one).
> * Both the team at Mozilla and other contributors have past experience with React.
>
> Now that #2546 is merged we have a hybrid code base, where we can introduce React component into our code. We will start with gradually rewriting existing components and writing new code with React.

 This is a gradual process where until finished, our codebase will use both Angular.js and React. Since version 4 we were shipping this hybrid codebase. The code editor, date parameters and edit in place component use React now.

arikfr:
https://github.com/STRML/react-grid-layout - can be relevant for our dashboard grid.

ice2038:
[Libraries for React i18n](https://phraseapp.com/blog/posts/react-i18n-best-libraries/) for [**Add i18n support** #933](https://github.com/getredash/redash/issues/933).

brylie:
>  I wouldn’t necessarily pick React (but rather Vue)

I don't want to diminish the work that the Mozilla team are doing, nor discount their wisdom and experience. So, thanks to all the people who are updating the frontend framework.

My concern is that the overarching direction in the JavaScript ecosystem is somewhat dismissive of web standards, opting to re-invent CSS and HTML inside of JavaScript.

[...]

PR #3505, opened 2019-02-27
Migrate Alerts List Page to React

gabrieldutra:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix

## Description
React migration for Alerts List Page. Mostly just migrated the same page idea as an ItemsListPage. I didn't add a sidebar because I thought it could be premature for this page (LMK if you have a different idea).

#3071

screenshot from 2019-02-27 16-36-00


client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -60,23 +60,23 @@ body {
 
 // Fixed width layout for specific pages
 @media (min-width: 768px) {
-  settings-screen, home-page, page-dashboard-list, page-queries-list, alerts-list-page, alert-page, queries-search-results-page, .fixed-container {
+  settings-screen, home-page, page-dashboard-list, page-queries-list, page-alerts-list, alert-page, queries-search-results-page, .fixed-container {
```
arikfr: Out of the scope of this PR, but all this tag name based rules will no longer work once we move to React. Worth switching to a (single) class name soon.

gabrieldutra:
> Not sure but I think the default sorting should be by date instead of name.

Yep, I guess this one makes sense

> Also, "last updated" is misleading as it holds "last triggered" data, not when the alert was last updated. Might be confusing to users.

This was misleading before haha, if you do any sort of update in the Alert (e.g.: name) it will update this value, hence why I updated the column name. I was actually gonna mention that in the PR, but I forgot :man_facepalming:. Thanks for pointing this out! :grin:

PR #3506, opened 2019-02-27
Migrate User Pages to React

### LINKED ISSUE #3507: No invitation link after adding a new user when email server is not configured
### Issue Summary

After #3267 and #3229 admins no longer receive a link to invite users when the server doesn't have an email server configured.

### Steps to Reproduce

1. Open `/users/new` in a server without email configured (e.g.: [preview](https://redash-preview.netlify.com/users/new))
2. Create a new user and check that no link is provided.
3. Open the pending invitation User Profile and notice there's only the "Resend" option (again it leads to nowhere)

The admin should receive a link to send to the user manually.

### Technical details:

* Redash Version: Latest (Redash Preview)
* Browser/OS: --
* How did you install Redash: --

gabrieldutra:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix

## Description
Migrating what's left for the User:
- [x] Create User - Now it's a Dialog
- [x] User Profile - Migrated as is
- [x] Cleanups

Closes #3507 (feat @rauchy)

**Create User (Now it's a Dialog)**
new-user

**Create User - Email taken (The Dialog doesn't close)**
email-taken-alert

**Resend Invitation (Now a link is shown, feat. @rauchy)**
resend-invitation


arikfr:
> Create User - Email taken (The Dialog doesn't close)

Wouldn't it be better to show it in the dialog as an input validation text? Feels like more standard UX.

gabrieldutra:
> Wouldn't it be better to show it in the dialog as an input validation text? Feels like more standard UX.

Yes, I thought about this as well, this may require some tweaks since I would have to treat the custom error message and also add a custom validation. I'll give it sometime to try, but in any case I kept this in a working state with toastr.

arikfr:
Maybe we can change the response from server to include a message you can directly show to the user + reference to the invalid field?

-- 
Arik Fraimovich

gabrieldutra:
> Maybe we can change the response from server to include a message you can directly show to the user + reference to the invalid field?

Yep, that would reduce half of the tweaks. I edited the previous message to say this and as you read it by email you probably didn't read :sweat_smile:: Perhaps it's worthy to deliver as is in this PR and work in a following one. My reasons are:
- this PR is currently holding Omer's Fix;
- It'll make related changes clearer (need to think if I'll keep using DynamicForm for that, if not, perhaps it's time to work on our Form abstraction)

arikfr:
> Perhaps it's worthy to deliver as is in this PR and work in a following one. My reasons are:

You're right 👍 

But will it require a lot of changes to just show this as a message below the form (not "attached" to some specific input) instead of in a notification?

gabrieldutra:
> But will it require a lot of changes to just show this as a message below the form (not "attached" to some specific input) instead of in a notification?

No, it's done :slightly_smiling_face:. Just rendered an Alert instead of using toastr (could be ath instead)

email-taken-alert

client/app/components/users/UserShow.jsx
```diff
@@ -1,8 +1,7 @@
 import React from 'react';
-import { react2angular } from 'react2angular';
 import { UserProfile } from '../proptypes';
 
-export const UserShow = ({ user: { name, email, profileImageUrl } }) => (
+const UserShow = ({ user: { name, email, profileImageUrl } }) => (
   <div className="col-md-4 col-md-offset-4 profile__container">
```
kravets-levko: Turn it into regular function and `export default` right here

client/app/pages/users/UsersList.jsx
```diff
@@ -116,6 +120,45 @@ class UsersList extends React.Component {
...
   ];
 
+  componentDidMount() {
+    if ($location.path() === '/users/new') {
+      this.showCreateUserDialog();
+    }
+  }
+
+  createUser = values => new Promise((resolve, reject) => {
+    User.create(values, (user) => {
```
kravets-levko:
```javascript
createUser = values => User.create(values).$promise
  .then((user) => {
    // code from success callback
  })
  .catch((error) => {
    if (!(error instanceof Error)) {
      error = new Error(get(error, 'data.message', 'Failed saving.'));
    }
    throw error; // Promise should be rejected with Error instance
  });
```

gabrieldutra:
> JS part looks pretty good, just few notices. Python part also looks good to me, but probably @arikfr or @rauchy could provide better feedback

I guess Python part was already reviewd in #3519 (@rauchy was the one who did it :stuck_out_tongue_closed_eyes:)

client/app/pages/users/UsersList.jsx
gabrieldutra: I'm not sure if I did that wrong (I think not), but the ErrorHandler seems to log the thrown error in any case (Even though I catch the error :man_shrugging: and the rest working properly). I'm using `return Promise.reject(error)` instead of `throw error` to avoid this.

client/app/components/users/CreateUserDialog.jsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = { savingUser: false, errorMessage: null };
+    this.form = React.createRef();
+  }
+
+  componentDidMount() {
+    recordEvent('view', 'page', 'users/new');
```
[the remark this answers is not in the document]
kravets-levko: I think it's okay to keep it

kravets-levko:
> I guess Python part was already reviewd in #3519 (@rauchy was the one who did the changes stuck_out_tongue_closed_eyes)

I think you should catch up with master - then that changes should disappear from this PR.

gabrieldutra:
> I think you should catch up with master - then that changes should disappear from this PR.

They won't :laughing:, he actually target my branch so we wouldn't need to modify the frontend twice.

kravets-levko:
> he actually target my branch so we wouldn't need to modify the frontend twice.

Then that changes will disappear from his PR :grin: Anyways - nice job :+1:

gabrieldutra:
🤟

PR #3515, opened 2019-03-01
Add @babel/polyfill to npm dependencies

gabrieldutra:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix

## Description
This adds `@babel/polyfill` to npm dependencies in order to support IE11 and some Edge versions.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -28,6 +28,7 @@ const config = {
   mode: isProduction ? "production" : "development",
   entry: {
     app: [
+      "@babel/polyfill",
       "./client/app/index.js",
       "./client/app/assets/less/main.less",
       "./client/app/assets/less/ant.less"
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -28,6 +28,7 @@ const config = {
   mode: isProduction ? "production" : "development",
   entry: {
     app: [
+      "@babel/polyfill",
```
gabrieldutra: Not sure if this is the best place to put this (could be in `index.js` as well)

kravets-levko: I think it's a good place for such things as well.

ranbena:
So I think we're unanimous that Redash should no longer support IE11.
Perhaps polyfill promise only (https://stackoverflow.com/a/41062991/259289)

arikfr:
I opened #3579 about showing a message to IE users.
Closing this now.

PR #3517, opened 2019-03-02
Added UTC display to scheduler dialog

### LINKED ISSUE #3509: Show UTC time for scheduled queries in schedule dialog
When scheduling a query to run at a specific hour, we translate user's timezone to UTC. This is a really nice feature, except that it's not obvious. To make it easier to for the user, we should show the UTC time in the dialog:

Ugly Mockup

We already translate it in place in the dialog, so it's basically just a matter of displaying the data.

ranbena:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix

## Description
Added dimmed UTC time next to picker.

Fixes #3509.


client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -139,6 +139,29 @@ class ScheduleDialog extends React.Component {
     this.setScheduleUntil(null, date);
   };
 
+  getTimeSelector = () => {
+    const { hour, minute, newSchedule: { time: utc } } = this.state;
```
arikfr: What this `newSchedule: { time: utc }` is about?

ranbena: `newSchedule` is the actual data saved in DB, in which `time` is kept in utc.
https://github.com/getredash/redash/blob/ade3cc72a70d2c65b6f8e393304f4a394f61e926/client/app/components/queries/ScheduleDialog.jsx#L74-L80

arikfr: 👍 I got confused by the object deconstruction here.

client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -139,6 +139,29 @@ class ScheduleDialog extends React.Component {
     this.setScheduleUntil(null, date);
   };
 
+  getTimeSelector = () => {
```
arikfr: If not for the need to define PropTypes, I think this could be better served by a component than a function.

TypeScript can't come fast enough 😩 I wonder if we should just use an eslint comment to disable the PropTypes check for "inner" components? (it's not like we're checking them in this case anyway)

(this is not a blocker)

ranbena: Done 4156776926ff6510cbed00328d3d6c4f8bfa06ea

client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -19,6 +19,42 @@ const DATE_FORMAT = 'YYYY-MM-DD';
...
+      <TimePicker
+        allowEmpty={false}
+        value={time}
+        format={HOUR_FORMAT}
+        minuteStep={5}
+        onChange={onChange}
+      />
+      {showUtc && (
+        <span className="utc">
+          ({ moment.utc(time).format(HOUR_FORMAT) } UTC)
```
arikfr: Why not to keep taking it from `newSchedule`?

client/app/components/queries/ScheduleDialog.jsx
```diff
@@ -19,6 +19,42 @@ const DATE_FORMAT = 'YYYY-MM-DD';
 const HOUR_FORMAT = 'HH:mm';
 const { Option, OptGroup } = Select;
 
+function TimeEditor(props) {
+  const [time, setTime] = useState(props.defaultValue);
```
arikfr: Why store state and not show the time passed in props?

client/app/components/queries/ScheduleDialog.jsx
ranbena: Cause I think it's wrong for a component to be fed this kind of data from a prop when it's actually just a different presentation derived from it's own state.

When it was a method `getTimeSelector()` it was ok to be coupled with it's own `newSchedule` implementation but now that it's an independent component, this coupling should be avoided.

That's how I see it.

client/app/components/queries/ScheduleDialog.jsx
ranbena: Getting the updated value from prop is also a wrong approach imo, since the value not set externally but from the component itself. Also, value from prop assumes that a certain flow happens outside of the component, which introduces risk of breakage.

client/app/components/queries/ScheduleDialog.jsx
arikfr: While the logic for converting to utc is short and trivial, it has the risk of introducing a bug where we will save one thing and show a different one. We should either have a function that used in both places or have a test that ensures the value we show is the value we store (the function is probably better and easier). 

But I'm not sure you should apply different rules to this component from the ones you used to the `getTimeSelector` function. They serve the same purpose: a helper to render the time selector and encapsulate some small logic about it. Both not supposed to be used beyond the context of `SchedueDialog`.

ranbena: Then I think we both agree any implementation has risk so testing is a must.
You ok I leave as is and add tests?

arikfr: Sure.

ranbena: Added following tests:
1. UTC set correctly on init (2h below set time)
2. UTC time should not render (if already UTC)
3. onChange correct result - simulates time change then asserts UTC and callback value

🏄🏄‍♀️

client/app/components/queries/ScheduleDialog.test.js
```diff
@@ -97,6 +98,57 @@ describe('ScheduleDialog', () => {
       });
     });
 
+    describe('TimeEditor', () => {
```
arikfr: This is awesome 👌 

But has one small issue: you assume 2 hours difference between the timezone and UTC. But what happens when DST kicks in (or the other way around)?

kravets-levko: I think it could be solved setting timezone for tests not using name, but explicitly `TZ=GMT-2:00` or something

https://github.com/getredash/redash/blob/549f878c98dcf515968c4f5b94b0fda59f50cc7c/package.json#L16

arikfr:
Good idea.

-- 
Arik Fraimovich

client/app/components/queries/ScheduleDialog.test.js
ranbena: That's a good save @arikfr :)
Now that I think of it, our other test snapshots might also fail when DST changes on March 29th! 😵

But I couldn't figure out an appropriate `TZ` value. Couldn't find docs, seems like it accepts only location strings and UTC. No offset. Plz help.

If no solution is found, might need to fallback to `TZ=UTC` and mock offset in test itself.

kravets-levko: @ranbena `TZ` is just a system variable and all programs should respect it:
image

ranbena: 10x @kravets-levko but it just doesn't work.
I changed to a non-dst timezone instead, so we're good.
```
"test": "TZ=Africa/Khartoum jest",
```

kravets-levko: > but it just doesn't work.

Jest - :man_facepalming:

PR #3519, opened 2019-03-02
Show invitation link if email server not setup

rauchy:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix

## Description
If e-mail is not configured, there is no way to retrieve invitation links. This PR returns and shows invitation links when e-mail is not configured.

Solves #3507 

screen shot 2019-03-02 at 23 59 59
screen shot 2019-03-03 at 0 01 45

### DIFF
--- a/redash/handlers/users.py
+++ b/redash/handlers/users.py
@@ -37,10 +37,17 @@ order_results = partial(
 )
 
 
-def invite_user(org, inviter, user):
+def invite_user(org, inviter, user, send_email=True):
+    email_configured = settings.MAIL_DEFAULT_SENDER is not None
+    d = user.to_dict()
+
     invite_url = invite_link_for_user(user)
-    send_invite_email(inviter, user, invite_url, org)
-    return invite_url
+    if email_configured and send_email:
+        send_invite_email(inviter, user, invite_url, org)
+    else:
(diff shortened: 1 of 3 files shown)


rauchy:
Done (2239b72b). Take it away @gabrieldutra.

redash/handlers/users.py
```diff
@@ -37,10 +37,17 @@
...
 
-def invite_user(org, inviter, user):
+def invite_user(org, inviter, user, send_email=True):
+    email_configured = settings.MAIL_DEFAULT_SENDER is not None
+    d = user.to_dict()
+
     invite_url = invite_link_for_user(user)
-    send_invite_email(inviter, user, invite_url, org)
-    return invite_url
+    if email_configured and send_email:
```
arikfr: We should return `invite_link` _only_ if `email_configured=False`. In this case, if I pass `send_email=False`, I can get an invite link even if email server is configured.

This means that in a secure environment like the hosted service, I could create a user with any email address I want and get an invite link to it without receiving the actual email address.

rauchy: Good catch. For some reason I thought that the no-invite flag returns the invite link, but it actually doesn't.

PR #3522, opened 2019-03-03
Moved widget and dashboard save logic out of dialog

ranbena:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix

## Description
Was always bothered that AddWidgetDialog did the actual widget saving.
My reasoning is that dialogs should be treated as functional parts - input > output and nothing more.

In this PR I just moved the widget save logic into dashboard.js where I think it belongs.
Also, renamed `addWidget` -> showAddWidgetDialog` and `addTextBox` > `showAddTextboxDialog`.

### DIFF
--- a/client/app/components/dashboards/AddWidgetDialog.jsx
+++ b/client/app/components/dashboards/AddWidgetDialog.jsx
@@ -7,13 +7,11 @@ import { wrap as wrapDialog, DialogPropType } from '@/components/DialogWrapper';
 import {
   MappingType,
   ParameterMappingListInput,
-  editableMappingsToParameterMappings,
-  synchronizeWidgetTitles,
 } from '@/components/ParameterMappingInput';
 import { QuerySelector } from '@/components/QuerySelector';
 
 import { toastr } from '@/services/ng';
-import { Widget } from '@/services/widget';
+
 import { Query } from '@/services/query';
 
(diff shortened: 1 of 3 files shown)


gabrieldutra:
:thinking:
I didn't think of Dialog components this way before. In fact I was thinking the opposite... I had this in mind when I was dealing with them:
1. It was a way to reduce complexity for the component that would render the Dialog (the dialog had 50 ~ 70 LOC against +200 in the other component). Also, the logic was totally related to the Dialog behavior;
2. In case the Dialog would be reused there'd be no need to repeat the logic, it was a "full working component";

But yep, if you look into the concept of Dialogs it does seem to make sense for it to be only input > output

ranbena:
Thanks @gabrieldutra the way I see it, a dialog is a user prompt for missing information.
You click "Add Widget" in the dashboard, but dashboard needs more info, dialog pops up, on confirm missing info is fulfilled and now dashboard can continue with original procedure.

arikfr:
@ranbena's description of a dialog does make sense.

> In case the Dialog would be reused there'd be no need to repeat the logic, it was a "full working component";

This is a good point as well. I guess that for reusable dialogs we can have a function that contains the logic of: Open Dialog -> Process Dialog Result, which will be reused (instead of directly reusing the dialog).

PR #3525, opened 2019-03-04
Add Hangouts Chat as alert destination

pieter-venter:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix

## Description

Added support for [Google Hangouts Chat](https://gsuite.google.com/products/chat/) as an Alert Destination. Hangouts Chat is Google's version of Slack and integrated into GSuite. 

This allows organizations who use GSuite to have a webhook post an alert into a chat room. This is very useful in operational scenarios, where you may be monitoring key metrics and need to be alerted once some threshold criteria is met.

Like Slack, Hangouts Chat has a Web client and mobile client. This integration automatically works with both.

### The alerts get posted into chat.google.com rooms like this
5-alert-triggered

### When the alert state returns to normal
6-alert-normal

### Configuration steps inside hangouts chat

1. Click on the room name drowdown, select Configure Webhooks

0-configure-webhooks

2. Give it a name, and optionally a link to an icon

1-webhook-in-room

3. Click Save. You will see the URL for the webhook.
Copy this to the clipboard

2-webhook-url

4. In Redash, setup a new Alert Destination.
Give it a name (can be anything)
Paste the webhook URL
Optionally provide a link to an image that will appear in the post

3-setup-alert-destination

5. Setup an Alert for a Query in Redash

4-setup-alert


ranbena:
@pieter-venter quick q - what is that under the green icon?
screen shot 2019-03-04 at 22 07 25

pieter-venter:
@ranbena Yes, password manager. LastPass offering to fill in the field because it is called "name".

redash/destinations/hangoutschat.py
```diff
@@ -0,0 +1,86 @@
+import logging
+import requests
+
+from redash.destinations import *
+from redash.utils import json_dumps
+
+
+class HangoutsChat(BaseDestination):
```
arikfr:
```suggestion
class HangoutsChat(BaseDestination):
    @classmethod
    def name(cls):
        return "Google Hangouts Chat"
    
    @classmethod
    def type(cls):
        return "hangouts_chat"
```

Small change to show a friendlier name and split the two words in the type identifier.

redash/destinations/hangoutschat.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def icon(cls):
+        return 'fa-bolt'
+
+    def notify(self, alert, query, user, new_state, app, host, options):
+        try:
+            if new_state == "triggered":
+                message = "<b><font color=\"#c0392b\">Triggered</font></b>"
+            else:
```
arikfr: Alerts have 3 states: unknown, normal and triggered. You should probably add a `elif` for the unknown case, otherwise it will be treated as normal.

redash/destinations/hangoutschat.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class HangoutsChat(BaseDestination):
+    @classmethod
+    def name(cls):
+        return "Google Hangouts Chat"
+    
+    @classmethod
+    def type(cls):
+        return "hangouts_chat"
```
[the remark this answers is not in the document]
arikfr: @pieter-venter yes, your current database records reference the old type (`hangoutschat`) which can no longer be found. The following query should fix this:

```sql
UPDATE notification_destinations SET type = 'hangouts_chat' where type = 'hangoutschat';
```

But if it creates too much hassle for you, we can revert back to `hangoutschat`. It just triggered my OCD :D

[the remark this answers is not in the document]
pieter-venter: OK great thanks. I did coincidentally found this myself too and fixed it using PGAdmin. Still learning the internals of Redash.

Honestly I don't mind. I went with one word, because I saw Mattermost, Pagerduty etc all used one word - but then again - their names are actually one word.

Let's go with your suggestion. I'll change the icon name to match the underscore and submit.

redash/destinations/hangoutschat.py
pieter-venter: OK I based this off the logic in Slack, but I made the enhancement as you suggested.

If triggered - text will be red. If ok - text will be green. Otherwise text will be neutral.

My understanding is "unknown" will happen for example if the alert is unable to determine the state. For example, you're referencing a column name that does not exist or your query returned 0 rows. Instead of just staying "Unknown" - I tried to be a bit more friendly.

Open to suggestions though

image

pieter-venter:
@arikfr How would you like me to handle the codeclimate suggestions. Both the number of parameters and the suggested refactoring sounds like it should be handled by someone else (or accepted as-is).

redash/destinations/hangoutschat.py
arikfr: This is perfect 👌 Your understanding is correct. The alert also starts in unknown state, but we don't send a message in this case.

And you're right -- we have the same issue with Slack. And while I was bitten by it myself, I was just lazy to fix it until now 😞

arikfr:
@pieter-venter 

> How would you like me to handle the codeclimate suggestions.

You can safely ignore those. CodeClimate is a bit too nit picking sometimes :-)

arikfr:
Thanks, @pieter-venter for this great feature and for quickly following up on our feedback.

Btw, are you using Redash at Geotab? If you do, would love to hear more. Feel free to ping me via email -- arik at redash.io. Thanks!

kravets-levko:
@arikfr click hijacker :smile:

PR #3528, opened 2019-03-05
Param fix

### LINKED ISSUE #3523: Parameter Dialog UX Issues
* When opening the New Parameter dialog it should auto focus on the keyword field.
* Pressing `Enter` should submit the dialog.

The two above are to support keyboard only workflow, where you would hit Cmd+P -> Enter keyword -> Hit Enter, and can continue writing your query.

* "This is the keyword that will be added..." string always being split into two rows: maybe we should just show the keyword (the `{{ ... }}` part) on the second row?

* There should be some indication that we saved the query after hitting OK (when we save).

ranbena:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix

## Description
Added features:
* auto focus on name input
* submit on enter
* name help only on date range. Changed help text:

screen shot 2019-03-05 at 11 49 33

Fixes #3523

### DIFF
--- a/client/app/components/EditParameterSettingsDialog.jsx
+++ b/client/app/components/EditParameterSettingsDialog.jsx
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 import Modal from 'antd/lib/modal';
 import Form from 'antd/lib/form';
 import Checkbox from 'antd/lib/checkbox';
+import Button from 'antd/lib/button';
 import Select from 'antd/lib/select';
 import Input from 'antd/lib/input';
 import Divider from 'antd/lib/divider';
@@ -109,7 +110,7 @@ function EditParameterSettingsDialog(props) {
     return true;
   }
 
-  function onConfirm() {
+  function onConfirm(e) {
     // update title to default
     if (!param.title) {
       // forced to do this cause param won't update in time for save
@@ -118,18 +119,24 @@ function EditParameterSettingsDialog(props) {
     }
 
     props.dialog.close(param);
+
+    e.preventDefault(); // stops form redirect
   }
 
   return (
     <Modal
       {...props.dialog.props}
       title={isNew ? 'Add Parameter' : param.name}
-      onOk={onConfirm}
-      okText={isNew ? 'Add Parameter' : null}
-      okButtonProps={{ disabled: !isFulfilled() }}
       width={600}
+      footer={[(
+        <Button key="cancel" onClick={props.dialog.dismiss}>Cancel</Button>
+      ), (
+        <Button key="submit" htmlType="submit" disabled={!isFulfilled()} type="primary" form="paramForm">
+          {isNew ? 'Add Parameter' : 'OK'}
+        </Button>
+      )]}
     >
-      <Form layout="horizontal">
+      <Form layout="horizontal" onSubmit={onConfirm} id="paramForm">
         {isNew && (
           <NameInput
             name={param.name}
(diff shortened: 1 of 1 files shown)


client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -98,23 +119,31 @@ function EditParameterSettingsDialog(props) {
...
       {...props.dialog.props}
       title={isNew ? 'Add Parameter' : param.name}
-      onOk={onConfirm}
-      okText={isNew ? 'Add Parameter' : null}
-      okButtonProps={{ disabled: !isFulfilled() }}
+      width={600}
+      footer={[(
+        <Button key="cancel" onClick={props.dialog.dismiss}>Cancel</Button>
+      ), (
+        <Button key="submit" htmlType="submit" disabled={!isFulfilled()} type="primary" form="paramForm">
```
gabrieldutra: This is way better to handle Forms in Modals :rocket:

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -18,8 +20,16 @@ function getDefaultTitle(text) {
...
 }
 
-function NameInput({ name, onChange, existingNames, setValidation }) {
-  let helpText = `This is what will be added to your query editor {{ ${name} }}`;
+function isTypeDate(type) {
+  return startsWith('date', type) && !isTypeDateRange(type);
+}
+
+function isTypeDateRange(type) {
+  return startsWith('date-range', type);
```
arikfr: What about date/time range?

ranbena: 🤦‍♂️ Fixed with `/-range/.test(type);`

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -143,7 +172,7 @@ function EditParameterSettingsDialog(props) {
             <Option value="datetime-range-with-seconds">Date and Time Range (with seconds)</Option>
           </Select>
         </Form.Item>
-        {includes(['date', 'datetime-local', 'datetime-with-seconds'], param.type) && (
+        {isTypeDate(param.type) && (
```
arikfr: I don't know why, but it works only for `date` type:

image

(notice that there is no checkbox)

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -18,8 +20,16 @@ function getDefaultTitle(text) {
...
 }
 
-function NameInput({ name, onChange, existingNames, setValidation }) {
-  let helpText = `This is what will be added to your query editor {{ ${name} }}`;
+function isTypeDate(type) {
+  return startsWith('date', type) && !isTypeDateRange(type);
+}
+
+function isTypeDateRange(type) {
+  return /-range/.test(type);
```
arikfr: Doesn't seem to work at all 😳 

image

client/app/components/EditParameterSettingsDialog.jsx
ranbena: 😅 Mercy fire me plz.

Fixed https://github.com/getredash/redash/pull/3528/commits/7d6e0d8f8c86265aa9e06c481a3be4cbf5cd6f04

client/app/components/EditParameterSettingsDialog.jsx
ranbena: It's only for keyword - not title

arikfr: 🤦‍♂️

PR #3529, opened 2019-03-05
Refine New user modal

kravets-levko:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix

## Description

Add a flag to show New user dialog (instead of checking url path). UI/behavior didn't change.

### DIFF
--- a/client/app/pages/users/index.js
+++ b/client/app/pages/users/index.js
@@ -12,6 +12,7 @@ export default function init() {
       path: '/users/new',
       title: 'Users',
       key: 'active',
+      isNewUserPage: true,
     },
     {
       path: '/users/pending',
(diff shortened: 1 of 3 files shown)


client/app/pages/users/index.js
```diff
@@ -12,6 +12,7 @@ export default function init() {
       path: '/users/new',
       title: 'Users',
       key: 'active',
+      isNewUserPage: true,
```
gabrieldutra: Perhaps `openCreateUserDialog: true` will be more accurate on what this will do

kravets-levko: Yeah, I also called it something similar, but that name looks weird in context of line 157 (handle modal close). Let it stay `isNewUserPage` for now (if you don't mind of course).

gabrieldutra: That'll do it for now

PR #3530, opened 2019-03-05
Render LDAP and remote auth login links correctly when multi org mode is enabled.

jezdez:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix

## Description

When the multi org feature and remote or LDAP auth is enabled at the same time, the login links on the login landing page aren't correctly rendered since they are not passed the current org slug.

As an aside the LDAP auth handler isn't org scoped for some reason, and it's not clear if this was an oversight or deliberate decision.

https://github.com/getredash/redash/commit/3044c77309c864362dae69ffdeb9775654cdf0e5 added the URL reversal to login.html and seems to be the culprit of the missing org parameter.

### DIFF
--- a/redash/authentication/ldap_auth.py
+++ b/redash/authentication/ldap_auth.py
@@ -1,26 +1,28 @@
 import logging
-logger = logging.getLogger('ldap_auth')
+import sys
 
 from redash import settings
 
 from flask import flash, redirect, render_template, request, url_for, Blueprint
-from flask_login import current_user, login_required, login_user, logout_user
+from flask_login import current_user
 
 try:
-    from ldap3 import Server, Connection, SIMPLE, ANONYMOUS, NTLM
+    from ldap3 import Server, Connection
 except ImportError:
     if settings.LDAP_LOGIN_ENABLED:
-        logger.error("The ldap3 library was not found. This is required to use LDAP authentication (see requirements.txt).")
-        exit()
+        sys.exit("The ldap3 library was not found. This is required to use LDAP authentication (see requirements.txt).")
 
 from redash.authentication import create_and_login_user, logout_and_redirect_to_index, get_next_path
 from redash.authentication.org_resolving import current_org
+from redash.handlers.base import org_scoped_rule
+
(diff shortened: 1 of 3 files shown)


redash/authentication/ldap_auth.py
```diff
@@ -1,26 +1,29 @@
...
 try:
-    from ldap3 import Server, Connection, SIMPLE, ANONYMOUS, NTLM
+    from ldap3 import Server, Connection
 except ImportError:
     if settings.LDAP_LOGIN_ENABLED:
-        logger.error("The ldap3 library was not found. This is required to use LDAP authentication (see requirements.txt).")
-        exit()
+        raise RuntimeError(
+            "The ldap3 library was not found. This is required to use LDAP authentication (see requirements.txt)."
+        )
```
jezdez: Just a tiny correction to follow Python best practices, instead of using the logging system for something that Python is able to handle more graceful via the RuntimeError.

tests/test_handlers.py
```diff
@@ -91,18 +90,26 @@ def setUp(self):
         super(TestLogin, self).setUp()
         self.factory.org.set_setting('auth_password_login_enabled', True)
 
-    @classmethod
```
jezdez: I *think* those methods are leftovers from a previous iteration of the multi org feature.

redash/authentication/ldap_auth.py
jezdez: Fair enough, let's use `sys.exit` straight away since that solves the immediate problem of the logger being configurable (and possibly hiding the message). For the record, the `exit` function isn't a good practice since it's [only supposed to be used for interactive sessions](https://docs.python.org/3/library/constants.html#exit), not general purpose Python scripts. `sys.exit` on the other hand will [return an error code of 1 with the message printed to stderr](https://docs.python.org/3/library/sys.html#sys.exit).

jezdez:
@arikfr Ready for another look.

jezdez:
@arikfr Ready when you're ready.

Issue #3536, opened 2019-03-05
I'm unable to view a late binding view's columns in the Redash schema explorer


paulgraff:
### Issue Summary

Since https://github.com/getredash/redash/pull/2174 was merged in 2017, the [`svv_columns`](https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_COLUMNS.html) system view has been updated to include late-binding views. This results in **duplicate** columns being returned by [Redash's column aggregation query](https://github.com/getredash/redash/blob/b68051d3c50df9a727df4c6ca5c6f70750d387d3/redash/query_runner/pg.py#L271-L296), which is creating front-end errors when I go to view a table's columns using the schema browser:

```
Error: "[ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: column in table.columns track by column, Duplicate key: teamid, Duplicate value: teamid
http://errors.angularjs.org/1.5.11/ngRepeat/dupes?p0=column%20in%20table.columns%20track%20by%20column&p1=teamid&p2=teamid"
```

### Steps to Reproduce

1. Create a late-binding view in Redshift
2. Try to view its columns using the schema explorer in Redash
3. Note the JS errors in the browser console

### Technical details:

* Redash Version: 5.0.0
* Browser/OS: Firefox (Quantum) 65.0.1/MacOS 10.14
* How did you install Redash: Setup script

PR #3537, opened 2019-03-05
Remove duplicate column information for late-binding views

paulgraff:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix

## Description
Fixes https://github.com/getredash/redash/issues/3536 by removing the second half of the `UNION` that was introducing duplicate columns. Given the circumstances, I don't feel like the front-end code needs adjustment here.

Looks like AWS added support for late-binding views to `svv_columns` around Dec 2017: https://forums.aws.amazon.com/thread.jspa?messageID=793893&#793893:
> Late binding views improvements. Added support for svv_columns, and new functions PG_GET_LATE_BINDING_VIEW_COLS and PG_GET_COL, to return column information for late binding views

Does Redash operate under the assumption of regular Redshift maintenance? Or should this fix really be simply using a `UNION` instead of `UNION ALL` for backwards compatibility?

https://github.com/getredash/redash/issues/3536

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -261,8 +261,6 @@ class Redshift(PostgreSQL):
     def _get_tables(self, schema):
         # Use svv_columns to include internal & external (Spectrum) tables and views data for Redshift
         # https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_COLUMNS.html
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks!

I think that by this point it's safe to assume that all (most?) Redshift users are on a new enough version.

paulgraff:
Awesome, thanks for the quick response!

Issue #3539, opened 2019-03-06
List pages: Cmd+Click opens the link in current tab and new tab


arikfr:
Expected behavior: open only in new tab.

Issue #3535, opened 2019-03-05
Clicking on a group on users list takes you to the user page


arikfr:
From the users list, when clicking on a group name it opens the user page instead of the group page (but if you press back, it goes to the group page).

kravets-levko:
@arikfr I'll fix it tomorrow morning

PR #3540, opened 2019-03-06
Fix list pages

kravets-levko:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix

## Description

Fixes getredash/redash#3535
Fixes getredash/redash#3539


ranbena:
Another approach is to revoke the ability to click anywhere on the row and just have the row title clickable - which it already is! Wdyt?

(I don't think this feature is important enough to have our code unnecessarily complex and risky)

client/app/services/navigateTo.js
```diff
@@ -10,3 +10,13 @@ export default function navigateTo(url, replace = false) {
     $rootScope.$applyAsync();
   }
 }
+
+export function handleNavigationEvent(event, url, replace = false) {
+  event.preventDefault();
+  event.stopPropagation();
+  if (event.ctrlKey || event.metaKey) {
```
gabrieldutra: I've never liked this sort of workaround to adapt `onClick` events to work as `href`. I.e.: I mostly use the middle mouse button for this purpose. However, I also don't have a better option for such cases :man_shrugging:

kravets-levko: Probably better solution would be to turn table rows into links somehow, but I have no idea how to do that...

kravets-levko: _(well, I know how, but it requires significant changes to markup)_

kravets-levko:
@gabrieldutra @ranbena I changed this to use browser's native mechanism of handling `<a>` tags - at least in all my browsers it works perfectly (100% the same as using regular links). I'll test it in IE, can somebody test it on Mac?

kravets-levko:
Okay, cannot test in IE because "Promise is undefined" :thinking:

gabrieldutra:
> Okay, cannot test in IE because "Promise is undefined" thinking

Edge? (I believe it's even more important than IE :slightly_smiling_face:)

arikfr:
I'm with @ranbena on this one: it's not worth the complexity. Let's have the different elements clickable (the name, groups, etc) using <a> tags and the disable full row clickability.

kravets-levko:
@arikfr can you clarify what you did mean saying "Let's have the different elements clickable"? For example, should be something else clickable (in addition to name) on dashboard list page?

arikfr:
@kravets-levko no, only name (and favorite control). On users page, the groups should be clickable.

gabrieldutra:
> it's not worth the complexity

Perhaps a simpler helper method to track new tab events could be worthy, because iirc there is another part in Redash that does something similar (Fork query in new tab).

kravets-levko:
@gabrieldutra In a case with "Fork query" we always need to open a new tab; but here we should mimic default browser's behavior which differs in different browsers/OS

kravets-levko:
@arikfr It's very easy to remove this behavior; but could you try current solution first (in a case you'd like it :sweat_smile:)?

kravets-levko:
@arikfr @ranbena @gabrieldutra Removed onClick behavior for table rows - my friend tried it in Edge and FF on Window - my fix with native events doesn't work there :disappointed:

ranbena:
@kravets-levko perhaps also remove the `event.preventDefault(); event.stopPropagation();` in actions, like in UsersList?

kravets-levko:
@ranbena Thanks a lot! Done

Issue #3521, opened 2019-03-03
Users list: Disable button pushed beyond scroll when many groups are used


arikfr:
![image](https://user-images.githubusercontent.com/71468/53695932-35fd6d80-3dca-11e9-8d01-fc5b429891f8.png)

(you can see a live version on https://redash-preview.netlify.com)

ranbena:
Maybe this can be just tackled by a few css overrides for labels in table:
```css
.items-list .label-tag {
  display: inline-block;
  max-width: 135px;
}
```
Which results in:
![screen shot 2019-03-07 at 10 13 10](https://user-images.githubusercontent.com/486954/53941783-b3262c80-40c1-11e9-8325-d2d32712b076.png)

PR #3541, opened 2019-03-06
Make actions column fixed

kravets-levko:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix

## Description

Possible fix for getredash/redash#3521: make columns with action buttons fixed (see screenshots).

image

image

### DIFF
--- a/client/app/components/items-list/components/ItemsTable.jsx
+++ b/client/app/components/items-list/components/ItemsTable.jsx
@@ -149,6 +149,7 @@ export default class ItemsTable extends React.Component {
         dataSource={rows}
         rowKey={row => row.key}
         pagination={false}
+        scroll={{ x: true }}
         onRow={onTableRow}
       />
     );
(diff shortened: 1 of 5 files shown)


ranbena:
@kravets-levko Even-though #3545, this could function as a fail-safe in case sth gets messed up. Wdyt?

gabrieldutra:
> @kravets-levko Even-though #3545, this PR could function as a fail-safe in case sth gets messed up. Wdyt?

This is currently being applied to mobiles, even though #3545:
mobile-users

kravets-levko:
@gabrieldutra because, as I mentioned, fixing group labels is not enough - on narrow screens there will be scroll anyway.

kravets-levko:
@ranbena 

> Even-though #3545, this PR could function as a fail-safe in case sth gets messed up. Wdyt?

While it fixes the bug perfectly, design-wise it's unappealing. :grin:

ranbena:
Mobile needs a a different layout. This doesn't work with or without the fixed col.
IMO, if it can be easily turned off for mobile - let's use it, if not - let's withdraw the idea.

kravets-levko:
While this fix was one of possible options, I don't like it that much. Therefore I decided to close this PR. We should find a better solution for lists on small screens.

PR #3557, opened 2019-03-08
Prevent blank edit-in-place value

ranbena:
## What type of PR is this?

- [x] Bug Fiix

## Description
Edit-in-place component allows inserting empty spaces which can cause display problems across the app.

#### STR:
1. Edit query/dashboard/group name.
2. Insert "    ".
3. Click "enter".

Expected: Name unchanged
Actual: Control disappears.

### Fix
Since it already deals with an empty string, it's just a matter of trimming the set value.

### DIFF
--- a/client/app/components/EditInPlace.jsx
+++ b/client/app/components/EditInPlace.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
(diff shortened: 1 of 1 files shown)


ranbena:
I thought of perhaps doing this at the service level with resource transformResponse, but this is way lighter.

arikfr:
I think it makes sense for the component to take care of this.

Issue #3554, opened 2019-03-07
CSV download: filename does not include the name of the query


susodapop:
### Issue Summary

When I download a CSV of query results, the downloaded file reflects the `query_id` and the current date instead of the query name. This is different than what I expect: it should be the query name and today's date.

### Steps to Reproduce

1. Run a query
2. Download results as CSV

### Technical details:

* Redash Version: 7.0.0
* Browser/OS: Safari
* How did you install Redash: SaaS

ranbena:
Basically, we specify the name of the file in the `download` attribute of the link:
```html
<a target="_self" href="api/queries/67/results/466.csv" download="New_Query_2019_03_07.csv">
  Download as CSV File
</a>
```
It doesn't work though, if there's also a `Content-Disposition` header, which in fact was added in https://github.com/getredash/redash/pull/3359.

@arikfr ok to just change the server formatting to include `query.name` instead of `query.id`?
```py
str_key = None
if query is not None:
  if query.name is not None:
    str_key = query.name
  else
    str_key = str(query.id)
else:
  str_key = str(query_result.id)
filename = "{}_{}.{}".format(str_key, str_date, filetype,)
```

kravets-levko:
@ranbena +1. Also, query name should also be sanitized: filename cannot include some characters, like `/`, `\`, `:`, `#`, etc. (need to check which are forbidden in each OS and replace all of them with `_` or `-`)

kravets-levko:
The forbidden printable ASCII characters are:

- Linux/Unix:
```
/ (forward slash)
```
- Windows:
```
< (less than)
> (greater than)
: (colon - sometimes works, but is actually NTFS Alternate Data Streams)
" (double quote)
/ (forward slash)
\ (backslash)
| (vertical bar or pipe)
? (question mark)
* (asterisk)
```

Also, all non-alphanumeric/punctuation characters should be removed (most of them user cannot type in query name, but _can_ copy/paste from somewhere).

ranbena:
https://github.com/un33k/python-slugify seems to do the job:
```py
>>> slugify("1/2<3>4:5\"6\7|8?9*10 11.12,13!14'15=16+17_18#19--20é")
'1-2-3-4-5-6-8-9-10-11-1213-14-15-16-17-18-19-20e'
```

kravets-levko:
[...]

PR #3559, opened 2019-03-08
Download Query Result links: use query name for downloaded filename

kravets-levko:

- [x] Bug Fix

## Description

Fixes getredash/redash#3554. Set filename of downloaded file using query name (if possible). Also, previously used "slugify" algorithm replaced with more permissive one (removes only characters forbidden in Windows/Linux/Unix filesystems, and optimizes whitespaces a bit) - because downloaded filename is not a part of URL, therefore it can contain wider characters range (in fact - almost everything).

image

### DIFF
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -96,13 +96,13 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
         return {'job': job.to_dict()}
 
 
-def get_download_filename(query_result, query=None, filetype='json'):
-    str_date = query_result.retrieved_at.strftime("%Y-%m-%d")
-    if query is not None:
-        str_filename = to_filename(query.name) if query.name != '' else str(query.id)
+def get_download_filename(query_result, query, filetype):
+    retrieved_at = query_result.retrieved_at.strftime("%Y_%m_%d")
+    if query:
+        filename = to_filename(query.name) if query.name != '' else str(query.id)
     else:
-        str_filename = str(query_result.id)
-    return u"{}.{}.{}".format(str_filename, str_date, filetype)
+        filename = str(query_result.id)
+    return u"{}_{}.{}".format(filename, retrieved_at, filetype)
 
 
 class QueryResultListResource(BaseResource):
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -10,7 +10,7 @@ from redash.permissions import (has_access, not_view_only, require_access,
                                 require_permission, view_only)
 from redash.tasks import QueryTask
 from redash.tasks.queries import enqueue_query
-from redash.utils import (collect_parameters_from_request, gen_query_hash, json_dumps, utcnow)
+from redash.utils import (collect_parameters_from_request, gen_query_hash, json_dumps, utcnow, to_filename)
 from redash.utils.parameterized_query import ParameterizedQuery, InvalidParameterError, dropdown_values
 
 
@@ -94,6 +94,15 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
             "Query ID": query_id
         })
         return {'job': job.to_dict()}
+
+
+def get_download_filename(query_result, query=None, filetype='json'):
+    str_date = query_result.retrieved_at.strftime("%Y-%m-%d")
+    if query is not None:
+        str_filename = to_filename(query.name) if query.name != '' else str(query.id)
+    else:
+        str_filename = str(query_result.id)
+    return u"{}.{}.{}".format(str_filename, str_date, filetype)
 
 
 class QueryResultListResource(BaseResource):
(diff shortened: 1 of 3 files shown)


redash/handlers/query_results.py
```diff
@@ -96,6 +96,15 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
         return {'job': job.to_dict()}
 
 
+def get_download_filename(query_result, query=None, filetype='json'):
+    str_date = query_result.retrieved_at.strftime("%Y-%m-%d")
+    if query is not None:
+        str_filename = to_filename(query.name) if query.name != '' else str(query.id)
+    else:
+        str_filename = str(query_result.id)
+    return u"{}.{}.{}".format(str_filename, str_date, filetype)
```
gabrieldutra: Isn't it better to use underscores (or sth else) instead of dots to separate the filename? Windows may have issues with it, not sure :thinking:

kravets-levko: @gabrieldutra Windows doesn't have issues with dots since time when started using Fat32 :slightly_smiling_face:  Also, Windows started using Unicode for filenames approx. from the same time - I think it was Win95 or Win98.

ranbena: Well I'm still using Norton Commander and this is really not ok

kravets-levko: @ranbena NC was developed for MS DOS, and it's miracle that it worked in XP/Vista/7. FAR Manager is a modern replacement for NC, and it correctly works with Unicode filenames (I was using it few years ago on Win7)

gabrieldutra: I also consider this a matter of standard, a dot is commonly used to separate filename and extension. Windows can work fine with it today, but what about programs?

kravets-levko: @gabrieldutra All programs that were developed at times of WinXP can work with it smoothly. Some very old apps may not handle it properly, but, well, they cannot handle extensions like `.json` or `.xlsx` (more that 3 characters) as well. I think we shouldn't focus on so old software.

kravets-levko: BTW - WinXP (an newer versions) have a fallback for such old applications - DOS-alike filenames that exists simultaneously to regular ones: https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#short-vs-long-names

arikfr: I'm with @gabrieldutra: I think it's better to use underscores. It's also consistent with what we were doing until now. Also we were replacing spaces with underscores (in query name).

gabrieldutra: I was actually thinking about recent programs/scripts that could have issues with it by treating this  naively in the code. Personally, I've always tried to avoid it. Anyway, as this seems to be a really small context nowadays, I won't make big deal with it :slightly_smiling_face:

kravets-levko: @arikfr But what if query name itself will contain a dot?

arikfr: 🤷‍♂️ 

Anyway, the dot vs. underscore isn't a big deal, but I do prefer we replace whitespace with underscores (makes it easier to use with CLI tools).

redash/utils/__init__.py
```diff
@@ -186,3 +186,12 @@ def base_url(org):
 
 def filter_none(d):
     return select_values(lambda v: v is not None, d)
+
+
+def to_filename(s):
```
rauchy: I'd get rid of all comments here. Code is self explanatory and comments will get stale. They are already lying 😝

redash/handlers/query_results.py
```diff
@@ -96,6 +96,15 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
         return {'job': job.to_dict()}
 
 
+def get_download_filename(query_result, query=None, filetype='json'):
```
rauchy: Why are you defaulting `filetype` to `json`? That's a handler concern and is already done there.

redash/handlers/query_results.py
```diff
@@ -96,6 +96,15 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
...
+    str_date = query_result.retrieved_at.strftime("%Y-%m-%d")
+    if query is not None:
```
rauchy:
```suggestion
    if query:
```

redash/handlers/query_results.py
```diff
@@ -96,6 +96,15 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
         return {'job': job.to_dict()}
 
 
+def get_download_filename(query_result, query=None, filetype='json'):
+    str_date = query_result.retrieved_at.strftime("%Y-%m-%d")
```
rauchy: could you rename this to `date` or `retrieved_at`? Same for `str_filename` (`filename`)

redash/handlers/query_results.py
kravets-levko: @arikfr Done.

arikfr:
May the unicode gods be in our favor 🙏

PR #3563, opened 2019-03-10
Convert all dropdown values to strings to support parameter lookup

### LINKED ISSUE #3562: Issues with Query Based Dropdown when the value is numeric
When the values the query returns for the dropdown are numeric, we have two issues:

1. We don’t load the correct value because the value from the URL passed as string, while the actual value is a number.
2. We don’t load results at all, because they fail parameter schema validation (the string doesn't match the values the dropdown has, which are numbers).

rauchy:
Fixes #3562

- [x] Bug Fix

Solves #3562

### DIFF
--- a/redash/utils/parameterized_query.py
+++ b/redash/utils/parameterized_query.py
@@ -13,7 +13,7 @@ def _pluck_name_and_value(default_column, row):
     name_column = "name" if "name" in row.keys() else default_column.lower()
     value_column = "value" if "value" in row.keys() else default_column.lower()
 
-    return {"name": row[name_column], "value": str(row[value_column])}
+    return {"name": row[name_column], "value": unicode(row[value_column])}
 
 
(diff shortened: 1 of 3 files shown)


redash/utils/parameterized_query.py
```diff
@@ -13,7 +13,7 @@ def _pluck_name_and_value(default_column, row):
     name_column = "name" if "name" in row.keys() else default_column.lower()
     value_column = "value" if "value" in row.keys() else default_column.lower()
 
-    return {"name": row[name_column], "value": row[value_column]}
+    return {"name": row[name_column], "value": str(row[value_column])}
```
arikfr: `unicode(row[...])`?

rauchy: Y U NO SUGGEST CHANGE

PR #3565, opened 2019-03-10
Cypress test - dashboard create/archive

ranbena:

- [x] Test

## Description
E2E test
* Create new dashboard
* Archive it

ezgif com-video-to-gif


client/app/components/dashboards/edit-dashboard-dialog.html
```diff
@@ -1,17 +1,19 @@
...
+  <div class="modal-footer" ng-if="$ctrl.policy.isCreateDashboardEnabled()">
+    <button type="button" class="btn btn-default" ng-disabled="$ctrl.saveInProgress" ng-click="$ctrl.dismiss()">Close</button>
+    <button type="button" class="btn btn-primary" ng-disabled="$ctrl.saveInProgress || !$ctrl.isFormValid()" ng-click="$ctrl.saveDashboard()" data-test="DashboardSaveButton">Save</button>
+  </div>
 
-<div class="modal-body" ng-if="!$ctrl.policy.isCreateDashboardEnabled()">
-  <edit-dashboard-dialog-disabled></edit-dashboard-dialog-disabled>
+  <div class="modal-body" ng-if="!$ctrl.policy.isCreateDashboardEnabled()">
+    <edit-dashboard-dialog-disabled></edit-dashboard-dialog-disabled>
+  </div>
```
ranbena: Just wrapped all these elements with `<div data-test="EditDashboardDialog">`

cypress/integration/dashboard/create_dashboard_spec.js
```diff
@@ -0,0 +1,36 @@
+describe('Create Dashboard', () => {
+  beforeEach(() => {
+    cy.login();
+    cy.visit('/dashboards');
+  });
+
+  it('create new', () => {
```
gabrieldutra:
```suggestion
  it('creates a new dashboard', () => {
```

IMO test names should have readable names, often more verbose, but the idea is that when you read it you get exactly what it asserts only from its name. For E2E tests this is a bit complicated because you actually assert a lot of things in a block to improve performance, but you can read "it creates a new dashboard" :grin:.

cypress/integration/dashboard/create_dashboard_spec.js
```diff
@@ -0,0 +1,36 @@
...
+      .click()
+      .within(() => {
+        cy.get('li').contains('Archive').click();
+      });
+
+    cy.get('.btn-warning').contains('Archive').click();
+    cy.get('.label-tag-archived').should('exist');
+    
+    cy.visit('/dashboards');
+    cy.get('@listItem').should('not.exist');
```
gabrieldutra: The `.as` seems to yield the same object as before ([ref](https://docs.cypress.io/api/commands/as.html#Yields)). If so, in this case is probably better to:
```javascript
const listItem = '.table-main-title[href*="dashboard/foo-bar"]';
cy.get(listItem).click();
// ...
cy.get(listItem).should('not.exist');
```

ranbena: If I understand correctly, `as` is similar to wrapping a sequence of Cypress commands in a function and when called, it runs the function against the current context.
Why do you think it's better to save the selector instead?

ranbena: I updated the test. Doesn't use this anymore. I do the assertion with snapshot instead.

ranbena:
Hmm the way I did the snapshots doesn't work. Lemme revise.

cypress/integration/dashboard/create_dashboard_spec.js
gabrieldutra: I initially thought this one would hold the same reference as before, therefore after a `cy.visit` this second `cy.get('@listItem')` would never exist. However they actually have this as [an example](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Using-Aliases-to-Refer-to-Previous-Subjects). It seems it holds the same reference as before, but will run the function again if needed. Cypress has a lot of magic behind those commands :laughing:. With this in mind your approach is much better :rocket:

client/app/pages/dashboards/DashboardList.jsx
```diff
@@ -88,35 +88,35 @@ class DashboardList extends React.Component {
...
+                      itemsPerPage={controller.itemsPerPage}
+                      page={controller.page}
+                      onChange={page => controller.updatePagination({ page })}
+                    />
+                  </div>
+                )}
+              </div>
+            ) : (
+              <LoadingState />
+            )}
```
ranbena: I just moved some elements to a wrapper `<div data-test="DashboardLayoutContent">`.

ranbena:
@gabrieldutra I added a DOM snapshot plugin.
The test now snaps before adding a new dashboard and after archiving it and makes sure it went back to original state. It's much simpler than searching for an element to _not_ be there.

It would be ideal, though, to overwrite the snapshot on each run so that any changes to the original state (done by previous tests or changes to UI) won't require a snapshot update cause it would be unnecessary. https://git.io/fhjRW

ranbena:
Looks like build failed cause perhaps the plugin didn't get installed.
@gabrieldutra I added it to `cypress:install` script - did I miss sth?

gabrieldutra:
> did I miss sth?

https://github.com/getredash/redash/blob/63d3f22c93bd88ddac75e10d40983ea94c3c85df/.circleci/Dockerfile.cypress#L6

:man_facepalming:. I actually don't remember if there was an specific reason not to use `npm run cypress:install` in here.

ranbena:
> 🤦‍♂️. I actually don't remember if there was an specific reason not to use `npm run cypress:install` in here.

`npm ERR! enoent ENOENT: no such file or directory, open '/usr/src/app/package.json'`

cypress/integration/dashboard/create_dashboard_spec.js
```diff
@@ -0,0 +1,51 @@
...
+      urlPath = pathname;
+      expect(pathname).to.include('dashboard/foo-bar');
+    });
+
+    cy.visit('/dashboards');
+    cy.get('@content').within(() => {
+      cy.get('.table-main-title')
+      .should('exist')
+      .and('have.attr', 'href', urlPath.substring(1))
+      .click();
```
gabrieldutra: Eslint again :sweat_smile: (I tried to updated the eslint PR, but I'll have to continue with that tomorrow)

cypress/integration/dashboard/create_dashboard_spec.js
```diff
@@ -0,0 +1,51 @@
+function createNewDashboard(dashboardName) {
+  cy.getByTestId('CreateButton').click();
+  cy.get('li[role="menuitem"').contains('Dashboard').click();
```
gabrieldutra:
```suggestion
  cy.get('li[role="menuitem]"').contains('Dashboard').click();
```

arikfr:
> The test now snaps before adding a new dashboard and after archiving it and makes sure it went back to original state. It's much simpler than searching for an element to not be there.

You could also verify this by querying the API to check the dashboard state or list of dashboards to make sure it's not there.

I wonder what's better?

ranbena:
> I wonder what's better?

Since DOM is the end result, it implies that the API responded as expected but not the other way around. So, I'd favor snapshots over API inspection.

arikfr:
> Since DOM is the end result, it implies that the API responded as expected but not the other way around. So, I'd favor snapshots over API inspection.

The reason I didn't like this is because it depends on an unrelated behavior, that might change unrelated to this test. Currently we show the default empty state when there are no dashboards. It's possible that at some point we will change this to show a different empty state after a single dashboard was created. In this case this test will fail for no good reason.

Considering that this tests' goal is not to test the empty state rendering (this should have tests of its own), I'm not sure what value we get from inspecting the DOM.

I'm mainly sharing this as an observation. You don't have to change anything here.

ranbena:
> It's possible that at some point we will change this to show a different empty state after a single dashboard was created. In this case this test will fail for no good reason.

We're on the same page on this. Hence I wrote this:
> It would be ideal, though, to overwrite the snapshot on each run so that any changes to the original state (done by previous tests or changes to UI) won't require a snapshot update cause it would be unnecessary. https://git.io/fhjRW

arikfr:
> We're on the same page on this. Hence I wrote this:

We are not. You assume a change to the design of the default empty state. I'm talking about a _different_ empty state showing up after you archive this dashboard.

Basically the current empty state is designed for the first time you see the dashboards list and not familiar with dashboards. Once you created a single dashboard, it's possible we might want to show you something else when the list is empty.

We won't do this any time soon of course. So imagine this fails in a year from now after we changed the empty state logic. Now the person who implemented the new logic needs to figure if you wanted to assert something about the empty state or just used it as a way to validate something else.

ranbena:
> I'm talking about a _different_ empty state showing up after you archive this dashboard.

Gotcha. In that case, I think it would be better to have separate screenshots (not snapshots) for before and after, and when changed in the future, the image diff would make easy sense.

Either that, or I'll revert back to testing that the new dashboard list item doesn't exist. Wdyt @arikfr?

arikfr:
I'm not sure why screenshots would make it easier to understand in this case.

But anyway, I think you can move on and keep it as is :) Maybe just add a comment explaining what you're validating with this assertion.

ranbena:
Gonna revert to previous assertion method - snapshots are not ideal here.

ranbena:
Test is now bulletproof:
1. Reverted snapshots. Back to checking element existence.
2. Dashboards cleanup before and after.

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -0,0 +1,74 @@
...
+  cy.get('.label-tag-archived').should('exist');
+}
+
+describe('Dashboard', () => {
+  beforeEach(() => {
+    cy.login();
+    archiveAllDashboards();
+  });
+
+  afterEach(archiveAllDashboards);
```
gabrieldutra: Cypress has some [good points](https://docs.cypress.io/guides/references/best-practices.html#Using-after-or-afterEach-hooks) not to use the `afterEach` to reset state. You actually solved half of that by also using the `beforeHook`, but I wonder if it's really worth it to have this here.

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -0,0 +1,74 @@
+function archiveAllDashboards() {
+  cy.visit('/dashboards');
+
+  // archive all dashboards
+  cy.getByTestId('DashboardLayoutContent').then($wrapper => {
+    $wrapper.find('.table-main-title').each((_, { href }) => {
+      archiveDashboard(href);
+    });
+  });
+}
```
gabrieldutra: Personally I would not reset dashboards, neither before nor after. But if you want to guarantee a clean state here it will be better to use requests in this case:
```javascript
  cy.request('/api/dashboards').then((response) => {
    response.body.results.forEach((dashboard) => {
      cy.request('DELETE', `/api/dashboards/${dashboard.slug}`)
    });
  });
```

cypress/integration/dashboard/dashboard_spec.js
ranbena: Makes great sense 👍

cypress/integration/dashboard/dashboard_spec.js
ranbena: Totally perfect.

But you're right, a better approach is just not to depend on clean state. I'll fix the code. 🙇

ranbena: OMG I'm in love with the code, now doesn't require clean slate:
1. When creating a new dashboard, listening to api response and extracting slug.
2. Slug then used as test hook.

💪

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -0,0 +1,59 @@
...
+  cy.getByTestId('EditDashboardDialog').within(() => {
+    cy.getByTestId('DashboardSaveButton').should('be.disabled');
+    cy.get('input').type(dashboardName);
+    cy.getByTestId('DashboardSaveButton').click();
+  });
+
+  return cy.wait('@NewDashboard').then((xhr) => {
+    const slug = Cypress._.get(xhr, 'response.body.slug');
+    assert.isDefined(slug, 'Dashboard api call returns slug');
+    return Promise.resolve(slug);
```
kravets-levko: I think you can just return `slug` from here

ranbena: You're right!

cypress/integration/dashboard/dashboard_spec.js
gabrieldutra: > When creating a new dashboard, listening to api response and extracting slug.

That was really cool :ok_hand:

PR #3566, opened 2019-03-10
Move cypress to client folder

gabrieldutra:

- [x] Other

## Description
Moved Cypress to client folder, so eslint works for it by default.
Nope

Nope


ranbena:
@arikfr we spoke about this a few weeks ago. I think we should consider moving `.eslintrc.js` to the root folder of the repo.

gabrieldutra:
It seems I missed some config for CodeClimate (it only caught my test for cypress). It also made eslint required for build. I'll need to check this more calmly.

arikfr:
> How about instead git mv cypress client/?

This crossed my mind :-) But I thought that moving eslint configuration will be easier.

arikfr:
>  It also made eslint required for build.

It should be.

ranbena:
> > How about instead git mv cypress client/?

Would also make lint-staged config easier with "ignore":
```json
  "lint-staged": {
    "linter": {
      "client/**/*.{js,jsx}": [
        "npm run lint:base",
        "npm run test -- --bail --findRelatedTests"
      ]
    },
    "ignore": ["client/dist/*"]
  }
```
rather than having to set two linter paths and repeating the config:
```json
  "lint-staged": {
    "client/app/**/*.{js,jsx}": [
      "npm run lint:base",
      "npm run test -- --bail --findRelatedTests"
    ],
  "cypress/**/*.{js,jsx}": [
      "npm run lint:base",
      "npm run test -- --bail --findRelatedTests"
    ]
  }
```

arikfr:
@ranbena 👍

gabrieldutra:
> rather than having to set two linter paths and repeating the config:

Not sure this would be a good approach, but it did work :sweat_smile:
```json
"lint-staged": {
  "{client/app,cypress}/**/*.{js,jsx}": [
    "npm run lint:base",
    "npm run test -- --bail --findRelatedTests"
  ]
}
```

But I'll go with the client folder option :laughing:

ranbena:
@arikfr this lgtm. You cool with this?

cypress.json
```diff
@@ -1,4 +1,10 @@
 {
   "baseUrl": "http://localhost:5000",
-  "video": false
+  "video": false,
+  "fixturesFolder": "client/cypress/fixtures",
+  "integrationFolder": "client/cypress/integration",
+  "pluginsFile": "client/cypress/plugins/index.js",
+  "screenshotsFolder": "client/cypress/screenshots",
+  "videosFolder": "client/cypress/videos",
+  "supportFile": "client/cypress/support/index.js"
```
arikfr: Cypress assumes that it's in the root folder? Maybe there is a way to set its path/context once to avoid having to maintain all these folders/files mapping?

ranbena: Perhaps `fileServerFolder: "client/"`?

gabrieldutra: > Cypress assumes that it's in the root folder? Maybe there is a way to set its path/context once to avoid having to maintain all these folders/files mapping?

Yes :confused:, I didn't like that I had to change all of those either.

The way I see that could probably set the path/context is changing `npm run cypress open` and `npm run cypress run` to run from another path. However, the current option seems better than that. With cypress knowing it's not in the root folder, you can run still run it using other commands: `cypress open` (in case you have it global) or `npx cypress open`.

LMK if you had something simpler in mind

> Perhaps fileServerFolder: "client/"?

I thought about this one too. I actually didn't get what's its purpose, but changing it doesn't affect what we want. Cypress keeps thinking it's in the root folder and, since there's no cypress folder in there, it creates it in an empty state.

ranbena:
@arikfr I'd really like this to land asap.
Any suggestions on how to improve the config?

arikfr:
I didn't realize this was waiting for me. I just asked if there is another option :)
Feel free to merge when you & @gabrieldutra feel it's ready.

gabrieldutra:
I'm going to merge this once CI is done, ongoing cypress' PRs will be affected @ranbena & @arikfr. There are some eslint issues in `search_query_spec.js`, I chose not to fix them in order to avoid diff issues :)

arikfr:
> There are some eslint issues in `search_query_spec.js`, I chose not to fix them in order to avoid diff issues :)

Once merged, I'll rebase #3635 and fix the lint errors.

PR #3569, opened 2019-03-11
Decouple extensions from Flask app.

### LINKED ISSUE #3466: Rewrite Extensions registry not to depend on Flask app
- [ ] Rewrite Extensions registry not to depend on Flask app
- [ ] Bring back support for `add_periodic_tasks` (disabled in #3465)

jezdez:

- [x] Refactor
- [x] Feature
- [X] Bug Fix

## Description

This separates the extension registry from the Flask app and also introduces a separate registry for preriodic tasks.

Fixes #3466.
Example change in redash-stmo: https://github.com/mozilla/redash-stmo/pull/25

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,14 +107,14 @@ jobs:
           path: /tmp/artifacts/
   build-docker-image:
     docker:
-      - image: circleci/buildpack-deps:xenial
+      - image: circleci/node:8
     steps:
       - setup_remote_docker
       - checkout
       - run: sudo apt install python-pip
       - run: sudo pip install -r requirements_bundles.txt
-      - run: npm run bundle
       - run: .circleci/update_version
+      - run: npm run bundle
       - run: .circleci/docker_build
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,6 +111,9 @@ jobs:
     steps:
       - setup_remote_docker
       - checkout
+      - run: sudo apt install python-pip
+      - run: sudo pip install -r requirements_bundles.txt
+      - run: npm run bundle
       - run: .circleci/update_version
       - run: .circleci/docker_build
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           name: Copy Test Results
           command: |
             mkdir -p /tmp/test-results/unit-tests
-            docker cp tests:/app/coverage.xml ./coverage.xml 
+            docker cp tests:/app/coverage.xml ./coverage.xml
             docker cp tests:/app/junit.xml /tmp/test-results/unit-tests/results.xml
       - store_test_results:
           path: /tmp/test-results
@@ -61,6 +61,7 @@ jobs:
     steps:
       - checkout
       - run: sudo apt install python-pip
+      - run: sudo pip install -r requirements_bundles.txt
       - run: npm install
       - run: npm run bundle
       - run: npm test
@@ -95,6 +96,7 @@ jobs:
     steps:
       - checkout
       - run: sudo apt install python-pip
+      - run: sudo pip install -r requirements_bundles.txt
       - run: npm install
       - run: .circleci/update_version
       - run: npm run bundle
--- a/bin/bundle-extensions
+++ b/bin/bundle-extensions
@@ -1,11 +1,10 @@
 #!/usr/bin/env python
 
 import os
-from subprocess import call
 from distutils.dir_util import copy_tree
 
-from pkg_resources import iter_entry_points, resource_filename, resource_isdir
-
+import importlib_metadata
+import importlib_resources
 
 
 # Make a directory for extensions and set it as an environment variable
@@ -19,21 +18,28 @@ if not os.path.exists(EXTENSIONS_DIRECTORY):
     os.makedirs(EXTENSIONS_DIRECTORY)
 os.environ["EXTENSIONS_DIRECTORY"] = EXTENSIONS_RELATIVE_PATH
 
-for entry_point in iter_entry_points('redash.extensions'):
-    # This is where the frontend code for an extension lives
-    # inside of its package.
-    content_folder_relative = os.path.join(
-        entry_point.name, 'bundle')
-    (root_module, _) = os.path.splitext(entry_point.module_name)
 
-    if not resource_isdir(root_module, content_folder_relative):
+def is_resource_dir(module, resource):
+    try:
+        return (
+            resource in importlib_resources.contents(module)
+            and not importlib_resources.is_resource(module, resource)
+        )
+    except TypeError:
+        # module isn't a package, so can't have a subdirectory/-package
+        return False
+
+
+for entry_point in importlib_metadata.entry_points().get('redash.extensions', []):
+
+    # Check if there is a "bundle" subdirectory/-package in the
+    # entrypoint's module and ignoring the entrypoint if not.
+    module_name = entry_point.pattern.match(entry_point.value).group('module')
+    if not is_resource_dir(module_name, 'bundle'):
         continue
 
-    content_folder = resource_filename(root_module, content_folder_relative)
-
-    # This is where we place our extensions folder.
-    destination = os.path.join(
-        EXTENSIONS_DIRECTORY,
-        entry_point.name)
-
-    copy_tree(content_folder, destination)
+    with importlib_resources.path(module_name, 'bundle') as bundle_dir:
+        # Copy content of extension bundle into extensions directory
+        destination = os.path.join(EXTENSIONS_DIRECTORY, entry_point.name)
+        print("Copying {} to {}".format(bundle_dir, destination))
+        copy_tree(str(bundle_dir), destination)
--- a/bin/bundle-extensions
+++ b/bin/bundle-extensions
@@ -1,35 +1,43 @@
 #!/usr/bin/env python
+"""Copy bundle extension files to the client/app/extension directory"""
 import os
-from distutils.dir_util import copy_tree
+from pathlib2 import Path
+from shutil import copy
 
-import importlib_metadata
-import importlib_resources
-
-from redash.extensions import BUNDLE_DIRECTORY, resource_isdir, entry_point_module
-
+from redash import create_app, extensions
 
 # Make a directory for extensions and set it as an environment variable
 # to be picked up by webpack.
-EXTENSIONS_RELATIVE_PATH = os.path.join('client', 'app', 'extensions')
-EXTENSIONS_DIRECTORY = os.path.join(
-    os.path.dirname(os.path.dirname(__file__)),
-    EXTENSIONS_RELATIVE_PATH)
+extensions_relative_path = Path('client', 'app', 'extensions')
+extensions_directory = Path(__file__).parent.parent / extensions_relative_path
 
-if not os.path.exists(EXTENSIONS_DIRECTORY):
-    os.makedirs(EXTENSIONS_DIRECTORY)
-os.environ["EXTENSIONS_DIRECTORY"] = EXTENSIONS_RELATIVE_PATH
+if not extensions_directory.exists():
+    extensions_directory.mkdir()
+os.environ["EXTENSIONS_DIRECTORY"] = str(extensions_relative_path)
 
+# We need the app here for logging and to make sure the bundles have loaded.
+app = create_app()
 
-for entry_point in importlib_metadata.entry_points().get('redash.extensions', []):
-    # Check if there is a "bundle" subdirectory/-package in the
-    # entrypoint's module and ignoring the entrypoint if not.
-    module = entry_point_module(entry_point)
-    if not resource_isdir(module, BUNDLE_DIRECTORY):
+bundles = extensions.bundles.items()
+if bundles:
+    app.logger.info('Number of extension bundles found: %s', len(bundles))
+else:
+    app.logger.info('No extension bundles found.')
+
+for bundle_name, paths in extensions.bundles.items():
+    # Shortcut in case not paths were found for the bundle
+    if not paths:
+        app.logger.info('No paths found for bundle "%s".', bundle_name)
         continue
 
     # The destination for the bundle files with the entry point name as the subdirectory
-    destination = os.path.join(EXTENSIONS_DIRECTORY, entry_point.name)
+    destination = Path(extensions_directory, bundle_name)
+    if not destination.exists():
+        destination.mkdir()
+
     # Copy the bundle directory from the module to its destination.
-    with importlib_resources.path(module, BUNDLE_DIRECTORY) as bundle_dir:
-        print("Copying {} to {}".format(bundle_dir, destination))
-        copy_tree(str(bundle_dir), destination)
+    app.logger.info('Copying "%s" bundle to %s:', bundle_name, destination.resolve())
+    for src_path in paths:
+        dest_path = destination / src_path.name
+        app.logger.info(" - {} -> {}".format(src_path, dest_path))
+        copy(str(src_path), str(dest_path))
--- a/bin/bundle-extensions
+++ b/bin/bundle-extensions
@@ -4,7 +4,7 @@ import os
 from pathlib2 import Path
 from shutil import copy
 
-from redash import create_app, extensions
+from redash import extensions
 
 # Make a directory for extensions and set it as an environment variable
 # to be picked up by webpack.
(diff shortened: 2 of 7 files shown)


bin/bundle-extensions
```diff
@@ -1,11 +1,10 @@
 #!/usr/bin/env python
 
 import os
-from subprocess import call
 from distutils.dir_util import copy_tree
 
-from pkg_resources import iter_entry_points, resource_filename, resource_isdir
-
+import importlib_metadata
+import importlib_resources
```
jezdez: I've moved to those two libraries since they are the backports of importlib packages that are in Python 3's stdlib while setuptools' pkg_resources is considered being deprecated.

bin/bundle-extensions
```diff
@@ -19,21 +18,28 @@ if not os.path.exists(EXTENSIONS_DIRECTORY):
...
+    module_name = entry_point.pattern.match(entry_point.value).group('module')
+    if not is_resource_dir(module_name, 'bundle'):
+        continue
 
-    copy_tree(content_folder, destination)
+    with importlib_resources.path(module_name, 'bundle') as bundle_dir:
+        # Copy content of extension bundle into extensions directory
+        destination = os.path.join(EXTENSIONS_DIRECTORY, entry_point.name)
+        print("Copying {} to {}".format(bundle_dir, destination))
+        copy_tree(str(bundle_dir), destination)
```
jezdez: The `str` is needed here since `bundle_dir` is a pathlib object.

redash/extensions.py
```diff
@@ -1,30 +1,50 @@
...
-                app.redash_extensions[entry_point.name] = {
-                    "entry_function": None,
-                    "resources_list": resource_listdir(root_module, content_folder_relative)
-                }
+
+    """
+    for entry_point in entry_points().get('redash.periodic_tasks', []):
+        app.logger.info('Loading Redash periodic tasks %s.', entry_point.name)
+        init_periodic_task = entry_point.load()
+        periodic_tasks[entry_point.name] = init_periodic_task()
```
jezdez: Note that this doesn't pass in the Flask app to not accidentally create tight coupling between the entry point code and the Flask app again. Unless you can think of a case where the app would be useful to pass in when creating periodic tasks?

redash/worker.py
```diff
@@ -74,11 +74,8 @@ def init_celery_flask_app(**kwargs):
...
-# @celery.on_after_configure.connect
-# def add_periodic_tasks(sender, **kwargs):
-#     app = safe_create_app()
-#     periodic_tasks = getattr(app, 'periodic_tasks', {})
-#     for params in periodic_tasks.values():
-#         sender.add_periodic_task(**params)
+@celery.on_after_configure.connect
+def add_periodic_tasks(sender, **kwargs):
+    for params in extensions.periodic_tasks.values():
+        sender.add_periodic_task(**params)
```
jezdez: I guess it wouldn't hurt if we'd pring to the Celery logger that we're adding those tasks here?

redash/extensions.py
```diff
@@ -1,30 +1,50 @@
...
 def init_extensions(app):
-    """
-    Load the Redash extensions for the given Redash Flask app.
-    """
-    if not hasattr(app, 'redash_extensions'):
-        app.redash_extensions = {}
+    """Load the Redash extensions for the given Redash Flask app.
 
-    for entry_point in iter_entry_points('redash.extensions'):
+    The extension entry pooint can return any type of value but
```
rauchy:
```suggestion
    The extension entry point can return any type of value but
```

requirements.txt
```diff
@@ -61,3 +61,5 @@ disposable-email-domains
 # It is not included by default because of the GPL license conflict.
 # ldap3==2.2.4
 gevent==1.4.0
+importlib-metadata==0.8
```
arikfr: Maybe we should add a comment that this can be removed when upgrading to latest Python?

redash/extensions.py
```diff
@@ -1,30 +1,50 @@
...
-        try:
-            extension = entry_point.load()
-            app.redash_extensions[entry_point.name] = {
-                "entry_function": extension(app),
-                "resources_list": []
+        init_extension = entry_point.load()
+        extensions[entry_point.name] = init_extension(app)
+
+
+def init_periodic_tasks(app):
```
arikfr: This is never called?

jezdez: Yeah, I first wanted to get a nod-off from you on the changed API, this would be called from the create_app function, too.

jezdez:
> See some comments, but there is a higher level thing I want to discuss --
> 
> Wouldn't it be better to have a single entrypoint (`redash_extension`?) which returns a dictionary. This dictionary can have multiple keys, each defines a different type of possible extension:
> 
> ```python
> {
>     'celery_periodic_tasks': [...],
>     'flask_extensions': [...],
> }
> ```

Yeah, admittedly this PR is a proposal, thank you for looking it through. I forgot to use the PR draft feature 😬

I wondered about how to structure the extension metadata, too. I think it boils down to whether we want to use "entry point groups" ("redash.extensions", "redash.periodic_tasks", [like](https://github.com/cdent/paste/blob/master/setup.py#L65-L110), [some](https://github.com/Pylons/pyramid/blob/eeea366a2c5cb56fdc7e8bdb8eb098ea1eff3705/setup.py#L103-L119), [others](https://github.com/python-babel/babel/blob/69cda1d2082444ed99836d92135f175d20555463/setup.py#L76-L97)) as a format to separate extension use cases or whether we want to create an own metadata schema for extensions to treat as an Redash-internal artifact.

Since it's one of the [community standards](https://packaging.python.org/specifications/entry-points/) that is being adopted by the Python stdlib via the importlib.metadata, it'll be around for a long time.

Anyway, for this round of the PR I leaned to the first format of using entry point groups since I was worried that we'd reinvent the wheel when we have to parse and handle the returning dicts that the single extension return.

While the [entry point system](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins) doesn't specifically say anything about the return type of entry point callbacks, in my experience the callback are usually only catering to one specific use case, instead of being an entrypoint for another layer of configuration. E.g. Other than the examples above Pygments for example uses [multiple entry points](https://bitbucket.org/birkenfeld/pygments-main/src/e7a5451c2c53fc1427c29ce647b5b732fa09b28b/pygments/plugin.py?at=default&fileviewer=file-view-default) for various extension features. So far I have not seen versioned extensions based on entry points out there, so it seems this is not something people have used it for.

> The registry loads this once, and then exposes an API to query it for the different extensions based on the type of functionality you need.
> 
> Wdyt?

I'm fairly open to doing it differently of course, but I'm interested in you looking into the examples out there and consider the downsides of maintaining part of an extension system on the long run. I know we don't have many (any?) extensions out there yet, so it's a low risk decision right now.

Lastly -- to look from an even higher perspective -- none of these solution solve the Python/NPM parity problem yet, meaning how should we deal with client extensions that also need backend code, and vice-versa. My gut feel says we can't easily combine them (since we'd have to either have an integration layer in Python to install additional NPM modules, and hook them into webpack, or one in the client code, which seems weird, too).

redash/extensions.py
arikfr: It can't be called from `create_app`, because create_app is called after the point periodic tasks should be configured. This is something that should be explicitly called in `redash.worker`.

arikfr:
Thank you for the references. Let's go the way you implemented it: having separate entry points for different types of extensions. But let's make sure that only the "Flask" extensions (not sure how to call them) depend on a Flask app instance.

redash/extensions.py
jezdez: Oh right, that's what I meant. Sorry

redash/extensions.py
```diff
@@ -1,30 +1,50 @@
...
+
+    """
+    for entry_point in entry_points().get('redash.extensions', []):
         app.logger.info('Loading Redash extension %s.', entry_point.name)
-        try:
-            extension = entry_point.load()
-            app.redash_extensions[entry_point.name] = {
-                "entry_function": extension(app),
-                "resources_list": []
+        init_extension = entry_point.load()
```
emtwo: If I recall correctly, `entry_point.load()` throws an exception when there is a frontend-only extension (since it doesn't have a python entry function), but it seems you've removed that exception handling. Is this handled another way?

jezdez: Yeah, I'll re-add that again.

jezdez:
@arikfr Is this something you'd be interested in merging before v7? We'd lose the ability to add those periodic tasks if not..

redash/worker.py
```diff
@@ -67,18 +80,56 @@ def __call__(self, *args, **kwargs):
...
+            entry_points={
+                "redash.periodic_tasks": [
+                    "add_two_and_two = calculus.addition:add_two_and_two",
+                ]
+                # ...
+            },
+            # ...
+        )
+    """
+    for entry_point in entry_points().get('redash.periodic_tasks', []):
```
arikfr: I think that all the code that uses entry_points directly should be in `redash.extensions` and here we should use an API exposed in `redash.extensions`.

redash/worker.py
```diff
@@ -67,18 +80,56 @@ def __call__(self, *args, **kwargs):
 celery.Task = ContextTask
 
 
-# Create Flask app after forking a new worker, to make sure no resources are shared between processes.
 @worker_process_init.connect
 def init_celery_flask_app(**kwargs):
+    """Create the Flask app after forking a new worker.
+
+    This is to make sure no resources are shared between processes.
+    """
```
arikfr: 👌

redash/worker.py
```diff
@@ -67,18 +80,56 @@ def __call__(self, *args, **kwargs):
...
+                # ...
+            },
+            # ...
+        )
+    """
+    for entry_point in entry_points().get('redash.periodic_tasks', []):
+        logger.info('Loading periodic Redash tasks %s from %s.', entry_point.name, entry_point.value)
+        try:
+            params = entry_point.load()
+            # Keep a record of our periodic tasks
```
arikfr: Why?

redash/extensions.py
```diff
@@ -1,30 +1,72 @@
...
+        def extension(app):
+            app.logger.info("Loading the Foobar extenions")
+            Foobar(app)
+
+    """
+    for entry_point in entry_points().get('redash.extensions', []):
         app.logger.info('Loading Redash extension %s.', entry_point.name)
+        module = entry_point_module(entry_point)
+        # First of all, try to get a list of bundle files
+        extensions[entry_point.name]['resource_list'] = module_bundle_files(module)
```
arikfr: Why not have a different entry point for the files (`redash.bundle` or something) to avoid all this logic?

redash/extensions.py
```diff
@@ -1,30 +1,72 @@
...
+            continue
+
+        # Otherwise check if the loaded entry point is a module
+        if isinstance(obj, ModuleType):
+            app.logger.info('Extension %s is a module.', entry_point.name)
+            extensions[entry_point.name]['extension'] = obj
+        # or simply call the loaded entry point instead.
+        else:
+            app.logger.info('Extension %s is a callable.', entry_point.name)
+            extensions[entry_point.name]['extension'] = obj(app)
```
arikfr: Maybe we should always expect a callable? Is there a use for having a module supported?

redash/worker.py
jezdez: I was trying to decouple the Redash code and Celery worker initialization, but I'll move it to redash.extensions if you'd like.

redash/worker.py
jezdez: To be able to introspect in case something goes wrong, once we've added it to the Celery schedule it's a lot harder to get to that information and separate it programmatically from the Redash internal periodic tasks.

redash/extensions.py
jezdez: Hmm, I don't think there was a good reason to do it like this other than being conservative with the API design back in the day. Now that we've settled on multiple entry points it's a good idea to separate it, true. Thanks!

redash/extensions.py
jezdez: Yeah, technically entry points are allowed to refer to modules, so we'll need to handle it somehow, but we can instead check if the return value is a callable and otherwise throw an exception. 👍

redash/worker.py
jezdez: Okay, this is done.

redash/extensions.py
jezdez: This is done.

redash/extensions.py
jezdez: Yep, done.

redash/worker.py
jezdez: I've moved this into `redash.extensions` like the other extension records for now.

bin/bundle-extensions
```diff
@@ -1,39 +1,43 @@
...
 import os
-from subprocess import call
-from distutils.dir_util import copy_tree
-
-from pkg_resources import iter_entry_points, resource_filename, resource_isdir
-
+from pathlib2 import Path
+from shutil import copy
 
+from redash import create_app, extensions
```
arikfr: Importing from `redash` currently breaks when no Redis server is available. We should probably fix this regardless, but just note that this code won't work in a Docker build execution, which is where it was previously used.

bin/bundle-extensions
```diff
@@ -1,39 +1,43 @@
...
-    if not resource_isdir(root_module, content_folder_relative):
+extensions_relative_path = Path('client', 'app', 'extensions')
+extensions_directory = Path(__file__).parent.parent / extensions_relative_path
+
+if not extensions_directory.exists():
+    extensions_directory.mkdir()
+os.environ["EXTENSIONS_DIRECTORY"] = str(extensions_relative_path)
+
+# We need the app here for logging and to make sure the bundles have loaded.
+app = create_app()
```
arikfr: Why bundles need the `app` instance?

jezdez: This needs the app to trigger the `redash.extensions.load_bundles` function, alternatively we can call it manually without the app here.

bin/bundle-extensions
jezdez: Hmm, yeah, that'll be a problem. I think it would make sense to somehow make sure the importing the redash package doesn't have import time side-effects.

bin/bundle-extensions
arikfr: But why `load_bundles` needs `app`? Except for using the logger it doesn't seem to need it.

bin/bundle-extensions
arikfr: I agree, I just didn't want this to be a dependency of your work here.

bin/bundle-extensions
jezdez: Ah, no good reason other than the app logger. I can detach that.

bin/bundle-extensions
jezdez: No problem, I think this is something that we need to do anyway, so there is #3601 :)

redash/extensions.py
```diff
@@ -1,30 +1,157 @@
...
+        try:
+            periodic_tasks[entry_point.name] = entry_point.load()
+        except (ImportError, AttributeError):
+            # and move on if it couldn't load it
+            logger.error('Periodic Redash task "%s" could not be found at "%s".', entry_point.name, entry_point.value)
+
+
+def init_app(app):
+    load_extensions(app)
+    load_bundles(app)
```
arikfr: Why do we need to load_bundles when we init the app?

bin/bundle-extensions
```diff
@@ -1,39 +1,41 @@
...
+    extensions_directory.mkdir()
+os.environ["EXTENSIONS_DIRECTORY"] = str(extensions_relative_path)
 
-    # This is where we place our extensions folder.
-    destination = os.path.join(
-        EXTENSIONS_DIRECTORY,
-        entry_point.name)
 
-    copy_tree(content_folder, destination)
+bundles = extensions.bundles.items()
```
arikfr: It seems that nothing triggers `load_bundles`, won't this be empty then?

redash/extensions.py
jezdez: Mostly for consistency with the bundle script and so that we can inspect `redash.extensions.bundles` when needed.

bin/bundle-extensions
jezdez: Yeah, weird, this got lost in refactoring. I just fixed that.

jezdez:
@arikfr This would be ready for another round of reviews 😬

jezdez:
@arikfr Is there anything I can do to move this ahead?

requirements_bundles.txt
```diff
@@ -0,0 +1,9 @@
+# These are the requirements that the extension bundle
+# loading mechanism need on Python 2 and can be removed
+# when moved to Python 3.
+# It's automatically installed when running npm run bundle
+
+# These can be removed when upgrading to Python 3.x
+importlib-metadata==0.9  # remove when on 3.8
+importlib_resources==1.0.2  # remove when on 3.7
+pathlib2==2.3.3  # remove when on 3.x
```
arikfr: Why the separate requirements file? Considering that these requirements are mandatory for running Redash, feels like this will only create confusion/issues.

jezdez: Oh I missed this comment. In the CI steps that run frontend related tests we want to run the bundle script before, so we'll need to install the requirements. I thought it would be less expensive to simply only those that are needed for the bundle script.

arikfr: 🤔 how about we add to `requirements.txt` a directive to install this one (`-r requirements_bundles.txt`), to minimize the affect on the rest of the configuration?

jezdez: Sure, that works for me!

jezdez: Done in https://github.com/getredash/redash/pull/3569/commits/eb9eb3be48c97d8b8184dcfd0917b595a55dbd93

.circleci/config.yml
```diff
@@ -105,11 +107,14 @@ jobs:
           path: /tmp/artifacts/
   build-docker-image:
     docker:
-      - image: circleci/buildpack-deps:xenial
+      - image: circleci/node:8
```
jezdez: This is to make sure npm is available in this step and is consistent with the other steps where bundles are collected.

arikfr:
Thanks!

641210 (commented):
****

PR #3576, opened 2019-03-13
Reuse handlers base require_fields

Feeeenng:

- [x] Other

## Description
reuse handlers.base  function " require_fields "


Feeeenng:
pep8 quetions

PR #3583, opened 2019-03-13
Show queue position while waiting for query to execute

washort:
This puts the number of queries currently waiting to be executed in the status for a job reported in the query execution page.


client/app/pages/queries/query.html
```diff
@@ -207,7 +207,7 @@ <h3>
                   <rd-timer timestamp="queryResult.getUpdatedAt()"></rd-timer>
                 </div>
                 <div class="alert alert-info m-t-15" ng-show="queryResult.getStatus() == 'waiting'">
-                  Query in queue&hellip;
+                  Query in queue (waiting on {{queryResult.job.queue_length}} quer<span ng-if="queryResult.job.queue_length === 1">y</span><span ng-if="queryResult.job.queueLength !== 1">ies</span>)&hellip;
```
kravets-levko: [ngPluralize](https://docs.angularjs.org/api/ng/directive/ngPluralize)

redash/tasks/queries.py
```diff
@@ -71,12 +72,21 @@ def to_dict(self):
         else:
             query_result_id = None
 
+        queriesAhead = 0
+        if task_status == 'PENDING':
+            waiting = get_waiting_in_queue('queries')
+            # find our job's index, return number of queries after it
```
kravets-levko: Seems this comment lies :grin: (function returns number of queries _before_ requested one)

redash/tasks/queries.py
```diff
@@ -71,12 +72,21 @@ def to_dict(self):
         else:
             query_result_id = None
 
+        queriesAhead = 0
```
arikfr: camelCase?

rafrombrc:
@arikfr Maybe it would help if we stepped back to explain what we're trying to accomplish here. It's not necessarily important that we have an exactly correct count of how many queries are in front of the current one in the queue. What *is* important is that folks get an idea about the relative state of things. Is everything flowing smoothly and the queues are relatively empty, or are the queues relatively gummed up so the user can expect that the user should expect much higher latency than usual.

Currently users click "execute" and experience wait times ranging from a few seconds to 5-10 minutes for the same query, depending on the overall load on the entire queue / data source system. We're hoping to make it so that, when wait times are long due to heavy load, the user at least gets an indication this is so.

Given this requirement, how would you suggest we tackle the problem?

arikfr:
> Maybe it would help if we stepped back to explain what we're trying to accomplish here.

Good idea! :-) 

> What is important is that folks get an idea about the relative state of things. Is everything flowing smoothly and the queues are relatively empty, or are the queues relatively gummed up so the user can expect that the user should expect much higher latency than usual.

Based on this description, I would suggest a different solution. We recently added new screens for admins to see Celery status (#3057). Using the same raw data, we can create a less technical view for end users that will reflect the queues status.

There is the question of what to show exactly, where to show it and then how to implement (for example, I would go for a separate API call that is no combined with job status polling). Can you open a new issue so we can discuss this?

client/app/pages/queries/query.html
washort: thanks

washort:
I've moved the queue status check into a separate request and made it check the data source for the right queue to query. It checks the status every 10 seconds that it's in "waiting" state.


example

client/app/pages/queries/query.html
```diff
@@ -207,7 +207,7 @@ <h3>
                   <rd-timer timestamp="queryResult.getUpdatedAt()"></rd-timer>
                 </div>
                 <div class="alert alert-info m-t-15" ng-show="queryResult.getStatus() == 'waiting'">
-                  Query in queue&hellip;
+                  Query in queue <span ng-if="queryResult && queryResult.queueStatus">(waiting on {{queryResult.queueStatus}} <ng-pluralize count=queryResult.queueStatus when="{'0': 'queries', 'one': 'query', 'other': 'queries'}"></ng-pluralize>)</span>&hellip;
```
kravets-levko: `0` case could be omitted (`other` covers it)

client/app/services/query-result.js
```diff
@@ -519,6 +520,18 @@ function QueryResultService($resource, $timeout, $q, QueryResultError) {
       );
     }
 
+    refreshQueueStatus(dataSourceId) {
+      if (this.getStatus() === 'waiting') {
+        const p = $http.get(`/api/queue_status/${this.job.id}?data_source=${dataSourceId}`);
+        p.then((response) => {
+          $timeout(() => this.refreshQueueStatus(dataSourceId), 10000);
```
ranbena: When query page is left, does this interval halt?

PR #3568, opened 2019-03-11
Migrate Admin pages to React

kravets-levko:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [x] Feature

## Description

- [x] Migrate System Status page
- [x] Refine Celery Status page
- [x] Migrate Outdated Queries page

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

image

image

image

PR #3584, opened 2019-03-14
(WIP) SystemStatus query runner (get Redash system status using SQL)

kravets-levko:

- [x] Feature

## Description

I haven't been sleeping and decided to try myself in creating query runners. This is just an experiment inspired by https://github.com/getredash/redash/pull/3568#issuecomment-471999196

image

image


ranbena:
So awesome! 🤩

arikfr:
@ranbena 

> Conceptually, not sure this should be a new query runner cause it contains actual data.

Interesting definition, but I'm not sure the fact it provides its own data disqualifies it from being a query runner on its own.

Having said that, it will be useful if this data source could connect to remote instances as well. The only change is to load the data from the `/status.json` API instead of reading directly. Then in the data source configuration it can take a URL and API key.

kravets-levko:
Another +1 for using data source for system status: it will obey common ACL rules. Current `/status.json` is hard-locked to admins only and cannot be shared with other users; if making that URL available to all - it couldn't be restricted. With data source it becomes really flexible: we can add it by default when setting up redash instance, but make it available only for admin group; if somebody wants - he can share it with other users.

kravets-levko:
Added an example how it could be used to retrieve celery workers status:

image

PR #3586, opened 2019-03-14
Widget title link not updated when parameter value changes

kravets-levko:

- [x] Bug Fix

## Description

Regression: when changing parameter value, link in widget title contains value set on page load.


kravets-levko:
@ranbena There is another "View query" link in widget's dropdown menu - it works in the same way. The idea is that refresh button updates _data_, but links are updated immediately.

PR #3587, opened 2019-03-14
Revert "Add SAML scheme override env var"

arikfr:
Reverts getredash/redash#2947: it isn't working (uses `settings` which isn't imported).


leosunmo:
https://github.com/getredash/redash/pull/3614

arikfr:

PR #3589, opened 2019-03-14
Textbox testing - add, remove, edit

ranbena:
## What type of PR is this?

- [x] Other - Tests

## Description
Added Cypress tests for Textbox in dashboard:
1. Add and remove (from button)
2. Add and remove (from menu)
3. Edit content

ezgif com-video-to-gif

### DIFF
--- a/cypress/integration/dashboard/dashboard_spec.js
+++ b/cypress/integration/dashboard/dashboard_spec.js
@@ -98,36 +98,36 @@ describe('Dashboard', () => {
       cy.login();
       createNewDashboardByAPI('Foo Bar')
         .then(slug => `/dashboard/${slug}`)
-        .as('DashboardUrl');
+        .as('dashboardUrl');
     });
 
     beforeEach(function () {
-      cy.visit(this.DashboardUrl);
-      addTextbox('Hello World!').as('TextboxEl');
+      cy.visit(this.dashboardUrl);
+      addTextbox('Hello World!').as('textboxEl');
     });
 
     it('removes textbox from X button', function () {
       editDashboard();
 
-      cy.get('@TextboxEl').within(() => {
+      cy.get('@textboxEl').within(() => {
         cy.get('.widget-menu-remove').click();
       });
 
-      cy.get('@TextboxEl').should('not.exist');
+      cy.get('@textboxEl').should('not.exist');
     });
 
     it('removes textbox from menu', function () {
-      cy.get('@TextboxEl').within(() => {
+      cy.get('@textboxEl').within(() => {
         cy.get('.widget-menu-regular').click({ force: true }).within(() => {
           cy.get('li a').contains('Remove From Dashboard').click({ force: true });
         });
       });
 
-      cy.get('@TextboxEl').should('not.exist');
+      cy.get('@textboxEl').should('not.exist');
     });
 
     it('edits textbox', function () {
-      cy.get('@TextboxEl').within(() => {
+      cy.get('@textboxEl').within(() => {
         cy.get('.widget-menu-regular').click({ force: true }).within(() => {
           cy.get('li a').contains('Edit').click({ force: true });
         });
@@ -139,7 +139,7 @@ describe('Dashboard', () => {
         cy.contains('button', 'Save').click();
       });
 
-      cy.get('@TextboxEl').should('contain', newContent);
+      cy.get('@textboxEl').should('contain', newContent);
     });
   });
 });
--- a/cypress/integration/dashboard/dashboard_spec.js
+++ b/cypress/integration/dashboard/dashboard_spec.js
@@ -1,47 +1,41 @@
-function createNewDashboard(dashboardName) {
-  cy.visit('/dashboards');
-  cy.getByTestId('CreateButton').click();
-  cy.get('li[role="menuitem"]')
-    .contains('Dashboard')
-    .click();
-
+function createNewDashboardByAPI(name) {
   cy.server();
-  cy.route('POST', 'api/dashboards').as('NewDashboard');
-
-  cy.getByTestId('EditDashboardDialog').within(() => {
-    cy.getByTestId('DashboardSaveButton').should('be.disabled');
-    cy.get('input').type(dashboardName);
-    cy.getByTestId('DashboardSaveButton').click();
-  });
-
-  return cy.wait('@NewDashboard').then((xhr) => {
-    const slug = Cypress._.get(xhr, 'response.body.slug');
-    assert.isDefined(slug, 'Dashboard api call returns slug');
+  return cy.request('POST', 'api/dashboards', { name }).then((response) => {
+    const slug = Cypress._.get(response, 'body.slug');
+    assert.isDefined(slug, 'Widget api call returns widget slug');
     return slug;
   });
 }
 
-function archiveCurrentDashboard() {
+function editDashboard() {
   cy.getByTestId('DashboardMoreMenu')
     .click()
     .within(() => {
       cy.get('li')
-        .contains('Archive')
+        .contains('Edit')
         .click();
     });
-
-  cy.get('.btn-warning')
-    .contains('Archive')
-    .click();
-  cy.get('.label-tag-archived').should('exist');
 }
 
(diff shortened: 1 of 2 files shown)


cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -56,4 +64,66 @@ describe('Dashboard', () => {
       });
     });
   });
+
+  describe('Textbox', () => {
```
gabrieldutra:
```suggestion
  context('Editing a dashboard', () => {
```

This way the overview: We are inside the Dashboard tests -> We are editing a dashboard
Editing a dashboard requires a dashboard, so we create one in the `before` hook.

I haven't tested this syntax yet (actually just discovered it when searching stuff in docs), but we need to use sth like this:

```javascript
context('editing a dashboard', () => {
  before(() => {
    createDashboardWithApiOrWhatever().as('dashboard');
  });
    
  beforeEach(() => {
    cy.visit(`/dashboards/${this.dashboard}`);
    // commands to open the edit state
  });
  
  // ...
});
```

Then, inside the test blocks we'll be able to use `this.dashboard` when needed

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -56,4 +64,66 @@ describe('Dashboard', () => {
...
+    beforeEach(() => {
+      createNewDashboard('Foo Bar');
```
gabrieldutra: In case my comment above actually works it'll be better to use the API here (dashboard creation has already been tested) and to have this in the `before` hook.

cypress/integration/dashboard/dashboard_spec.js
gabrieldutra: Tested, it works with `this`. A few notes from my tests:
- it's necessary to use the `function(){}` syntax for it to work
- `cy.get('@dashboard')` also works, but it will have an issue with the `before` hook. All aliases are reset before each test block, so you'd only have access to it in the first test block. Anyway, with `this.dashboard` I didn't have this issue :man_shrugging:

ranbena: Omg @gabrieldutra the code is much more awesome now! 10x 🤘
It even shaved off 8 secs 😁

Notice I also refactored the create/archive tests - separated to individual tests, where the archive uses the API ;)

ranbena: I wish I could also utilize a `AddTextboxByAPI` but it requires some initial data (widget position calculation) which would complicate things. I'm aiming to change that when refactoring the widgets.

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -1,59 +1,147 @@
...
 
       cy.visit('/dashboards');
       cy.getByTestId('DashboardLayoutContent').within(() => {
         cy.getByTestId(slug).should('not.exist');
       });
     });
   });
+
+
+  describe('Editing a dashboard', () => {
```
gabrieldutra: :thinking:. I actually thought about the `context` where you'd be "editing a dashboard". But thinking better you can also use `describe` in this case. A good pattern for this:
- Use `context` when you are testing the same thing, but within a context - `context('editing a dashboard', () =>{})`;
- Use `describe` when you are testing a different thing :slightly_smiling_face: - `describe('Textbox', () =>{})` 

In the `context` case I'd also add the `editDashboard()` to the `beforeEach` (we are in this context, so it makes sense).

For this it's really open, you can either consider one of this options, or even aggregate both :man_shrugging:

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -1,59 +1,147 @@
...
+      return createNewDashboardByAPI('Foo Bar').then((slug) => {
+        this.dashboardUrl = `/dashboard/${slug}`;
+      });
+    });
+
+    beforeEach(function() {
+      cy.visit(this.dashboardUrl);
+      return addTextbox('Hello World!').then(($el) => {
+        this.$textboxEl = $el; 
+      });
```
gabrieldutra: Same comment above :+1:

cypress/integration/dashboard/dashboard_spec.js
gabrieldutra: Sorry about suggesting the exact thing you used before :sweat_smile:, but if you choose the `describe` option:
```suggestion
  describe('Textbox', () => {
```

In my defense I haven't thought we'd ever use aggregated `describe`'s, but turns out it makes sense :stuck_out_tongue_closed_eyes:

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -1,59 +1,147 @@
...
+
+      cy.wrap(this.$textboxEl).within(() => {
+        cy.get('.widget-menu-remove').click();
+      });
+
+      cy.wrap(this.$textboxEl).should('not.exist');
+    });
+
+    it('removes textbox from menu', function() {
+      cy.wrap(this.$textboxEl).within(() => {
```
gabrieldutra: If you change to `.as` you can use `cy.get('textboxEl')` instead of `wrap`

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -1,59 +1,147 @@
...
-    cy.get('input').type(dashboardName);
-    cy.getByTestId('DashboardSaveButton').click();
-  });
-
-  return cy.wait('@NewDashboard').then((xhr) => {
-    const slug = Cypress._.get(xhr, 'response.body.slug');
-    assert.isDefined(slug, 'Dashboard api call returns slug');
+  return cy.request('POST', 'api/dashboards', { name }).then((response) => {
+    const slug = Cypress._.get(response, 'body.slug');
+    assert.isDefined(slug, 'Widget api call returns widget slug');
```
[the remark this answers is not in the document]
ranbena: 🤦‍♂️

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -1,59 +1,147 @@
...
     });
   });
+
+
+  describe('Editing a dashboard', () => {
+    before(function() {
+      cy.login();
+      return createNewDashboardByAPI('Foo Bar').then((slug) => {
+        this.dashboardUrl = `/dashboard/${slug}`;
+      });
```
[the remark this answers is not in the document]
ranbena: The return is cause of [working with promises](https://mochajs.org/#working-with-promises) but yeah this ain't a promise. Quite confusing.

Anywho, I'll change to this syntax 👍

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -1,59 +1,145 @@
...
-    cy.get('input').type(dashboardName);
-    cy.getByTestId('DashboardSaveButton').click();
-  });
-
-  return cy.wait('@NewDashboard').then((xhr) => {
-    const slug = Cypress._.get(xhr, 'response.body.slug');
-    assert.isDefined(slug, 'Dashboard api call returns slug');
+  return cy.request('POST', 'api/dashboards', { name }).then((response) => {
+    const slug = Cypress._.get(response, 'body.slug');
+    assert.isDefined(slug, 'Dashboard api call returns widget slug');
```
gabrieldutra:
```suggestion
    assert.isDefined(slug, 'Dashboard api call returns dashboard slug');
```
almost sure the slug is for the dashboard too

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -1,59 +1,145 @@
...
     });
   });
+
+
+  describe('Textbox', () => {
+    before(function () {
+      cy.login();
+      createNewDashboardByAPI('Foo Bar')
+        .then(slug => `/dashboard/${slug}`)
+        .as('DashboardUrl');
```
gabrieldutra: Why Upper? it looks weird when you do `this.DashboardUrl` :joy:

ranbena: For some reason I thought it's the cypress convention.. I'll change back.

PR #3590, opened 2019-03-14
Widget tests - add, remove, auto height

ranbena:
- [x] Other - Tests

## Description
Added Cypress tests for widgets. Creates query on the fly and tests:
1. Adds widget.
2. Removes widget.
3. Auto height feature test.

ezgif com-video-to-gif

### DIFF
--- a/cypress/integration/dashboard/dashboard_spec.js
+++ b/cypress/integration/dashboard/dashboard_spec.js
@@ -184,11 +184,11 @@ describe('Dashboard', () => {
       cy.login();
       createNewDashboardByAPI('Foo Bar')
         .then(slug => `/dashboard/${slug}`)
-        .as('DashboardUrl');
+        .as('dashboardUrl');
     });
 
     beforeEach(function () {
-      cy.visit(this.DashboardUrl);
+      cy.visit(this.dashboardUrl);
     });
 
     it('adds widget', () => {
--- a/cypress/integration/dashboard/dashboard_spec.js
+++ b/cypress/integration/dashboard/dashboard_spec.js
@@ -215,7 +215,7 @@ describe('Dashboard', () => {
         });
       });
 
-      it('render correct height for 2 table rows', () => {
+      it('renders correct height for 2 table rows', () => {
         const queryData = {
           query: 'select s.a FROM generate_series(1,2) AS s(a)',
         };
--- a/cypress/integration/dashboard/dashboard_spec.js
+++ b/cypress/integration/dashboard/dashboard_spec.js
@@ -227,7 +227,7 @@ describe('Dashboard', () => {
         });
       });
 
-      it('render correct height for 5 table rows', () => {
+      it('renders correct height for 5 table rows', () => {
         const queryData = {
           query: 'select s.a FROM generate_series(1,5) AS s(a)',
         };
--- a/cypress/integration/dashboard/dashboard_spec.js
+++ b/cypress/integration/dashboard/dashboard_spec.js
@@ -38,13 +38,21 @@ function addTextbox(text) {
   });
 }
 
-function addWidget(queryData) {
-  const defaultQueryData = { data_source_id: 1, options: { parameters: [] }, schedule: null };
-  const merged = Object.assign(defaultQueryData, queryData);
+const defaultQueryData = {
+  name: 'Test Query',
+  query: 'select 1',
+  data_source_id: 1,
+  options: {
+    parameters: []
+  },
+  schedule: null,
+};
 
-  cy.server();
+function addWidget(queryData = {}) {
+  const merged = Object.assign({}, defaultQueryData, queryData);
 
   // create query
+  cy.server();
   return cy.request('POST', '/api/queries', merged)
     .then(({ body }) => {
       // publish it so it's avail for widget
@@ -52,6 +60,7 @@ function addWidget(queryData) {
     })
     .then(({ body }) => {
       // create widget
+      editDashboard();
       cy.contains('a', 'Add Widget').click();
       cy.getByTestId('AddWidgetDialog').within(() => {
         cy.get(`.query-selector-result[data-test="QueryId${body.id}"]`).click();
@@ -63,7 +72,7 @@ function addWidget(queryData) {
     })
     .then((xhr) => {
       const body = Cypress._.get(xhr, 'response.body');
-      assert.isDefined(body, 'Widget api call returns body');
+      assert.isDefined(body, 'Widget api call returns response body');
       return body;
     });
 }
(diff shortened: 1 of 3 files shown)


ranbena:
@gabrieldutra this is ready for your scrutiny 🙇

A few notes:
1. I have MANY more widget tests planned (dimensions, resizing, dragging, responsiveness, etc.)
2. Now that I add a query - [Percy](https://percy.io/Redash/redash/builds/1584092) 😁
3. After we move the cypress folder we should create some lint overrides.

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -142,4 +178,66 @@ describe('Dashboard', () => {
...
+      });
+    });
+
+    describe('Auto height for table visualization', () => {
+      it('has correct visualization and position config', () => {
+        addWidget().then(({ options, visualization }) => {
+          expect(visualization.type).to.eq('TABLE');
+          expect(options.position.autoHeight).to.be.true;
+        });
+      });
```
ranbena: This is unnecessary, right? 🤔

gabrieldutra: I think the cost for it now isn't worth it (currently 2 sec). I wonder how long it would be if you created the widget by API.

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -142,4 +178,66 @@ describe('Dashboard', () => {
       cy.get('@textboxEl').should('contain', newContent);
     });
   });
+
+  describe('Widget', () => {
+    before(function () {
+      cy.login();
+      createNewDashboardByAPI('Foo Bar')
+        .then(slug => `/dashboard/${slug}`)
+        .as('DashboardUrl');
```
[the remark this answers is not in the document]
ranbena: I need to be more attentive to my code 😩

cypress/integration/dashboard/dashboard_spec.js
ranbena: The problem with creating an API call for Widget/Textbox is that it requires some dynamically generated position data to be sent with it:
```js
data = {
  ...
  options: {
    ...
    position: {
      autoHeight: true,
      col: 3,
      maxSizeX: 6,
      maxSizeY: 1000,
      minSizeX: 1,
      minSizeY: 1,
      row: 0,
      sizeX: 3,
      sizeY: 3,
    },
  },
}
```

that I'd rather not have to mimic. We could get away with it if the dashboard is a clean slate, though that has a cost of its own.

ranbena:
> I think the cost for it now isn't worth it (currently 2 sec). I wonder how long it would be if you created the widget by API.

@gabrieldutra thing is that these autoHeight checks should be done against a widget with an autoHeight feature, therefore I check these parameters as part of the tests.

Without these checks, I would be assuming that:
* a new query returns with a table vis
* a table vis is configured to have `{ autoHeight: true }`

Which if one day changed, would break the test illegitimately.
Wdyt?

gabrieldutra:
> Which if one day changed, would break the test illegitimately.
Wdyt?

I see, this is a test for the test :laughing:. The "problem" with it is that you're only asserting this for that specific case of `addWidget`. I think in this case it won't be an issue, but just to show Cypress cool possibilities:

```javascript
describe('Auto height for table visualization', () => {
  const haveCorrectParams = ({options, visualization}) => {
    expect(visualization.type).to.eq('TABLE');
    expect(options.position.autoHeight).to.be.true;
  };

  it('renders correct height for 2 table rows', () => {
    const queryData = {
      query: 'select s.a FROM generate_series(1,2) AS s(a)',
    };

    addWidget(queryData)
      .should(haveCorrectParams) // use whenever you want to guarantee the parameters
      .then(({ id }) => {
        cy.getByTestId(`WidgetId${id}`)
          .its('0.offsetHeight')
          .should('eq', 235);
      });
  });
});
```

I'm not sure I liked this way of coding, but it's a possibility indeed.

cypress/integration/dashboard/dashboard_spec.js
kravets-levko: @ranbena (cc @gabrieldutra) Actually not all that fields are required (some of them are saved accidentally, one day I'm going to fix that):
```javascript
position: {
  autoHeight: true,
  col: 3,
  row: 0,
  sizeX: 3,
  sizeY: 3,
},
```
All that values could be set manually, just check constraints for selected widget type (e.g. not all widgets can have `autoHeight=true`, some have min width/height; position can be arbitrary at all, just keep in mind that widgets "flow" to top if there is free space)

ranbena: Ok I'll go for it.

ranbena:
> I'm not sure I liked this way of coding, but it's a possibility indeed.

I like it better than my way, but I decided to ditch it entirely for sake of simplicity.

cypress/integration/dashboard/dashboard_spec.js
ranbena: Implemented it. Hope the diff isn't too heavy to review..
Here's my summary:
**Pros**
* -10 seconds run time

**Cons** (imho):
* Since the ui doesn't update after the "add widget" api request was executed , must reload the page for the widget to appear. Had to move the `cy.visit(this.dashboardUrl)` from `beforeEach` to inside the test.
* For the same reason, `addTextboxByAPI` and `addWidgetByAPI` can't return an element reference to be aliased, since the element doesn't yet exist in the dom. I tried to work around that limitation but eventually fell back to returning `elTestId` which will then require the test to `cy.getByTestId(elTestId)`. 
* Currently, settling on hardcoded "first widget" positioning config `{ col: 0, row: 0, sizeX: 3, sizeY: 3 }` but if future test requires multiple widgets, must do calculation (or opt for widget add from ui).

Overall I think this added more "magic" to the tests which I favor for production code and less for testing code as readability and scalability are more important than line count and abstraction. 

Lmk what you think @gabrieldutra and @levko and if you can improve the code.

cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -17,25 +14,70 @@ function editDashboard() {
...
+    width: 1,
+    dashboard_id: dashId,
+    visualization_id: null,
+    text: 'text',
+    options: {
+      position: { col: 0, row: 0, sizeX: 3, sizeY: 3 },
+    },
+  };
+
   cy.server();
```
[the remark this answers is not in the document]
ranbena: Removed 👍

ranbena:
@gabrieldutra thanks for the detailed feedback. Seems we can abstain from any changes to the code and just land this. Can you "approve"?

PR #3592, opened 2019-03-15
allowing to specify a custom work group for AWS Athena queries

ialeinikov:

- [X] Feature

## Description
This change allows specifying a custom AWS Athena work group.

https://docs.aws.amazon.com/athena/latest/ug/workgroups.html

Screen Shot 2019-03-15 at 1 20 57 PM

### DIFF
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -80,12 +80,12 @@ class Athena(BaseQueryRunner):
                 },
                 'work_group': {
                     'type': 'string',
-                    'title': 'Athena work group',
+                    'title': 'Athena Work Group',
                     'default': 'primary'
                 },
             },
             'required': ['region', 's3_staging_dir'],
-            'order': ['region', 'aws_access_key', 'aws_secret_key', 's3_staging_dir', 'schema'],
+            'order': ['region', 'aws_access_key', 'aws_secret_key', 's3_staging_dir', 'schema', 'work_group'],
             'secret': ['aws_secret_key']
         }
 
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -78,6 +78,11 @@ class Athena(BaseQueryRunner):
                     'type': 'boolean',
                     'title': 'Use Glue Data Catalog',
                 },
+                'work_group': {
+                    'type': 'string',
+                    'title': 'Athena work group',
+                    'default': 'primary'
+                },
             },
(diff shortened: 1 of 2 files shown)


redash/query_runner/athena.py
```diff
@@ -78,6 +78,11 @@ def configuration_schema(cls):
                     'type': 'boolean',
                     'title': 'Use Glue Data Catalog',
                 },
+                'work_group': {
```
rauchy: Could you add this to the correct place in `order`? (it happens now automatically due to the alphabetical order, but who knows what will come next)

redash/query_runner/athena.py
```diff
@@ -78,6 +78,11 @@ def configuration_schema(cls):
...
+                    'type': 'string',
+                    'title': 'Athena work group',
+                    'default': 'primary'
```
arikfr: Wouldn't it be safer/more backward compatible to just not pass anything if no value provided?

redash/query_runner/athena.py
```diff
@@ -78,6 +78,11 @@ def configuration_schema(cls):
                     'type': 'boolean',
                     'title': 'Use Glue Data Catalog',
                 },
+                'work_group': {
+                    'type': 'string',
+                    'title': 'Athena work group',
```
arikfr:
```suggestion
                    'title': 'Athena Work Group',
```

redash/query_runner/athena.py
ialeinikov: According to the AWS [documentation](https://docs.aws.amazon.com/athena/latest/ug/user-created-workgroups.html):

> By default, each account has a primary workgroup and the default permissions allow all authenticated users access to this workgroup. The primary workgroup cannot be deleted.

ialeinikov:
@rauchy, this was tested on our STG instance of Redash with Athena connections.
I've tested it with providing custom values to the "work group" field (based on our configuration) and without any custom values in that field.

rauchy:
@ialeinikov great! Have you given PyAthena a sanity check to make sure nothing breaks? (I'm referring to [this change](https://github.com/getredash/redash/pull/3592/files#diff-aab4aeb938d015cca5784d2c3bcea65dR23))

ialeinikov:
@rauchy, yes, I ran queries internally. Using both specific work group datasource and the one with the default value. Haven't noticed any issues.

arikfr:
Thanks!

PR #3593, opened 2019-03-15
Add an optional footer-extra component.

jezdez:

- [x] Feature

## Description

This is similar to the `<home-extra/>` and `<datasource-list-extra/>` placeholders, which would allow adding additional infos and links to the footer via Redash extensions.

### DIFF
--- a/client/app/components/app-view/template.html
+++ b/client/app/components/app-view/template.html
@@ -15,3 +15,4 @@
 </div>
 <div ng-if="!$ctrl.handler.error" ng-view></div>
 <footer ng-if="$ctrl.layout.showFooter"></footer>
+<footer-extra ng-if="$ctrl.layout.showFooter"></footer-extra>


arikfr:
Now that we have a help link in the navbar and version in the menu, the plan is to remove the footer...

kravets-levko:
@arikfr +1. Aside of that, it should be implemented as [`DynamicComponent`](https://github.com/getredash/redash/blob/master/client/app/components/DynamicComponent.jsx)

jezdez:
> Now that we have a help link in the navbar and version in the menu, the plan is to remove the footer...

Aha, that's fair, I'm simply wondering how to add link to our bugtracker and thought this would we an acceptable location. Do you know where it would fit better?

> @arikfr +1. Aside of that, it should be implemented as [`DynamicComponent`](https://github.com/getredash/redash/blob/master/client/app/components/DynamicComponent.jsx)

Nice, that totally flew past me! Looking at the code, I'd basically call `registerComponent` with an existing React component which is then rendered instead?

kravets-levko:
@jezdez 

1. register: `registerComponent('SomeComponent', MyComponent)`
2. use: `<DynamicComponent name="SomeComponent" />`

jezdez:
@arikfr To clarify, since the footer hasn't been remove in 7.0.0, should this be closed?

arikfr:
> Do you know where it would fit better?

Maybe inside the help drawer? It already has a DynamicComponent point which we use to have a link to open chat support/show status page:

image

ranbena:
@jezdez here's some sample code:

```jsx
// HelpDrawerExtraContent.jsx

import React from 'react';
import { registerComponent } from '@/components/DynamicComponent';
import Button from 'antd/lib/button';

import './HelpDrawerExtraContent.less';

class HelpDrawerExtraContent extends React.Component {
  static propTypes = {
    onLeave: PropTypes.func.isRequired,
    openPageUrl: PropTypes.func.isRequired,
  }

  render() {
    const { openPageUrl, onLeave } = this.props;
    return (
      <div className="help-drawer-extra-content">
        <a href="#" onClick={() => openPageUrl("https://www.bugzilla.org")}>Bugzilla</a>
        <Button onClick={onLeave}>Close drawer</Button>
      </div>
    );
  }
}

export default function init() {
  registerComponent('HelpDrawerExtraContent', HelpDrawerExtraContent);
}

init.init = true;
```

```less
// HelpDrawerExtraContent.less

.help-drawer-extra-content {
  background-color: #F6F8F9;
  width: 100%;
  height: 50px;
  padding: 10px 10px 10px 20px;
  line-height: 30px;
  display: flex;
  border-top: 1px solid #e0e0e0;
 
  a {
    flex: 1
  }

  button {
    height: 30px;
  }
}
```

jezdez:
@ranbena Super super exciting, thank you for sharing this snippet. Cheers!

@arikfr That works for us I think, thanks for the help!

PR #3599, opened 2019-03-18
Show accessible tables only in New Query view for PostgreSQL

shinsuke-nara:

- [x] Bug Fix

#### Background

Left pane of `New Query` view shows table list. The list includes tables which user can not access.
Tables which user can not access should be hidden.

#### The bug

Function `redash.query_runner.pg.PostgreSQL._get_tables` does not consider privileges of a user for tables.

#### The Fix

Get tables for which `current_user` of PostgreSQL have privileges. 

--

--

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -126,38 +126,12 @@ class PostgreSQL(BaseSQLQueryRunner):
             schema[table_name]['columns'].append(row['column_name'])
 
     def _get_tables(self, schema):
-        '''
-        relkind constants per https://www.postgresql.org/docs/10/static/catalog-pg-class.html
-        r = regular table
-        v = view
-        m = materialized view
-        f = foreign table
-        p = partitioned table (new in 10)
-        ---
-        i = index
-        S = sequence
-        t = TOAST table
-        c = composite type
-        '''
-
         query = """
-        SELECT s.nspname as table_schema,
-               c.relname as table_name,
-               a.attname as column_name
-        FROM pg_class c
-        JOIN pg_namespace s
-        ON c.relnamespace = s.oid
-        AND s.nspname IN (
-            SELECT table_schema
-            FROM information_schema.table_privileges
-            WHERE grantee = (SELECT current_user)
-            AND table_schema NOT IN ('pg_catalog', 'information_schema')
-            GROUP BY table_schema)
-        JOIN pg_attribute a
-        ON a.attrelid = c.oid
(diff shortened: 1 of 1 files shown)


ranbena:
@arikfr is omitting inaccessible tables the desired user experience?
An alternative would be to show these in disabled mode with a tooltip explanation, but I'm not knowledgable enough to know if this has any value.

Also, should this type of filtering be replicated across all query runners?

arikfr:
@ranbena 

> is omitting inaccessible tables the desired user experience?

Yes.

> Also, should this type of filtering be replicated across all query runners?

In theory query runners should be implemented in this way already. But there is no single recipe that will work for all of them anyway.

shinsuke-nara:
@arikfr 

Thank you for checking my PR.
I agree with you. I will fix this PR with option 3.

shinsuke-nara:
@arikfr 

I fixed this PR according to your advice.
Could you check it if you have time?

redash/query_runner/pg.py
```diff
@@ -126,33 +126,12 @@ def _get_definitions(self, schema, query):
...
-        JOIN pg_attribute a
-        ON a.attrelid = c.oid
-        AND a.attnum > 0
-        AND NOT a.attisdropped
-        WHERE c.relkind IN ('r', 'v', 'm', 'f', 'p')
+        SELECT table_schema,
+               table_name,
+               column_name
+        FROM information_schema.columns
+        WHERE table_schema NOT IN ('pg_catalog', 'information_schema')
```
arikfr: Thank you for the update, but we still need the old query for the "special types", so it should be something like:

```python
query = """
        SELECT s.nspname as table_schema,
               c.relname as table_name,
               a.attname as column_name
        FROM pg_class c
        JOIN pg_namespace s
        ON c.relnamespace = s.oid
        AND s.nspname NOT IN ('pg_catalog', 'information_schema')
        JOIN pg_attribute a
        ON a.attrelid = c.oid
        AND a.attnum > 0
        AND NOT a.attisdropped
        WHERE c.relkind IN ('m', 'f', 'p')

        UNION 

        SELECT table_schema,
               table_name,
               column_name
"""

Thanks :)

shinsuke-nara: @akariv 

Thank you for the advice. I have fixed again.

arikfr:
Thanks!

shinsuke-nara:
@arikfr 

Thank you for accepting my PR :)

PR #3601, opened 2019-03-18
Split redash/__init__.py to prevent import time side-effects.

jezdez:

- [x] Refactor
- [x] Bug Fix

## Description

This basically makes sure that when import the redash package we don't accidentally trigger import-time side-effects such as requiring Redis.

Refs #3569 and #3466.


redash/settings/__init__.py
```diff
@@ -2,19 +2,7 @@
 from funcy import distinct, remove
 
 from .helpers import fix_assets_path, array_from_string, parse_boolean, int_or_none, set_from_string
-from .organization import DATE_FORMAT
-
-
-def all_settings():
```
jezdez: This isn't needed since Flask has `app.config.from_object`

redash/app.py
```diff
@@ -0,0 +1,62 @@
+from flask import Flask
+from flask_sslify import SSLify
+from werkzeug.contrib.fixers import ProxyFix
+
+from . import mail, settings
+from .utils import routes
+
+
+class Redash(Flask):
```
jezdez: I basically tried to separate Flask app specific code in the flask.Flask subclass and all Redash specific code in the `create_app` function.

redash/__init__.py
```diff
@@ -68,71 +60,13 @@ def create_redis_connection():
...
-
-mail = Mail()
-migrate = Migrate()
-mail.init_mail(settings.all_settings())
-statsd_client = StatsClient(host=settings.STATSD_HOST, port=settings.STATSD_PORT, prefix=settings.STATSD_PREFIX)
-limiter = Limiter(key_func=get_ipaddr, storage_uri=settings.LIMITER_STORAGE)
 
-import_query_runners(settings.QUERY_RUNNERS)
-import_destinations(settings.DESTINATIONS)
+redis_connection = LocalProxy(create_redis_connection)
```
jezdez: LocalProxy objects aren't instantiated until they are accessed, so this would prevent creating a connection until it's actually called. As long as there is no module level call with the redis_connection object, we should be fine.

redash/__init__.py
```diff
@@ -68,71 +60,13 @@ def create_redis_connection():
...
-    migrate.init_app(app, db)
-    mail.init_app(app)
-    authentication.init_app(app)
-    limiter.init_app(app)
-    handlers.init_app(app)
-    configure_webpack(app)
-    extensions.init_extensions(app)
-    chrome_logger.init_app(app)
-    users.init_app(app)
+statsd_client = StatsClient(host=settings.STATSD_HOST, port=settings.STATSD_PORT, prefix=settings.STATSD_PREFIX)
```
jezdez: Those don't need to be proxies since the settings module doesn't import from anywhere else, it's safe to import and use it here.

redash/app.py
```diff
@@ -0,0 +1,62 @@
...
+
+    sentry.init()
+    app = Redash()
+
+    # Check and update the cached version for use by the client
+    app.before_first_request(reset_new_version_status)
+
+    # Load query runners and destinations
+    import_query_runners(settings.QUERY_RUNNERS)
+    import_destinations(settings.DESTINATIONS)
```
jezdez: This moves importing the query runners and destinations into a bit later during startup, and ties them to the app creation process. I think this is more in line what we expect since both can trigger all kinds of imports in the various backend code.

redash/cli/__init__.py
```diff
@@ -15,9 +16,11 @@ def create(group):
 
     @app.shell_context_processor
     def shell_context():
-        from redash import models
-        return dict(models=models)
-
+        from redash import models, settings
+        return {
+            'models': models,
+            'settings': settings,
```
jezdez: Just a tiny helper to get to the settings faster in shells.

redash/cli/__init__.py
```diff
@@ -48,7 +51,7 @@ def status():
 @manager.command()
 def check_settings():
     """Show the settings as Redash sees them (useful for debugging)."""
-    for name, item in settings.all_settings().iteritems():
+    for name, item in current_app.config.iteritems():
```
arikfr: Won't this show settings only Flask recognizes?

redash/settings/__init__.py
```diff
@@ -2,19 +2,7 @@
...
-    from types import ModuleType
-
-    settings = {}
-    for name, item in globals().iteritems():
-        if not callable(item) and not name.startswith("__") and not isinstance(item, ModuleType):
-            settings[name] = item
-
-    return settings
-
+from .organization import DATE_FORMAT  # noqa
```
arikfr: ?

redash/utils/routes.py
```diff
@@ -0,0 +1,15 @@
+from werkzeug.routing import BaseConverter
+
+
+class SlugConverter(BaseConverter):
+    def to_python(self, value):
+        # This is ay workaround for when we enable multi-org and some files are being called by the index rule:
+        # for path in settings.STATIC_ASSETS_PATHS:
+        #     full_path = safe_join(path, value)
+        #     if os.path.isfile(full_path):
+        #         raise ValidationError()
```
arikfr: Considering this is commented out, we might not need this converter at all? 🤔

redash/metrics/celery.py
```diff
@@ -1,17 +1,13 @@
...
 import time
 
-from celery.signals import task_postrun, task_prerun
 from redash import settings, statsd_client
 from redash.utils import json_dumps
 
 tasks_start_time = {}
 
 
-@task_prerun.connect
```
arikfr: Why move this? It's better to have this module self contained.

redash/__init__.py
arikfr: I don't think you need a proxy here. When creating the redis connection, it only creates a connection pool but doesn't make a connection.

redash/__init__.py
```diff
@@ -68,71 +60,13 @@ def create_redis_connection():
...
 
-import_query_runners(settings.QUERY_RUNNERS)
-import_destinations(settings.DESTINATIONS)
+redis_connection = LocalProxy(create_redis_connection)
 
-from redash.version_check import reset_new_version_status
-reset_new_version_status()
+mail = LocalProxy(Mail)
 
+migrate = LocalProxy(Migrate)
```
arikfr: Why Migrate and Mail need a proxy?

redash/__init__.py
```diff
@@ -68,71 +60,13 @@ def create_redis_connection():
...
-        #     if os.path.isfile(full_path):
-        #         raise ValidationError()
-
-        return value
-
-    def to_url(self, value):
-        return value
-
-
-def create_app():
```
arikfr: Can we import `create_app` to keep things that were importing it from `redash` working?

redash/cli/__init__.py
jezdez: Nope, it'll do a [`dir(redash.settings)`](https://github.com/pallets/flask/blob/1c2a2cd0014693f1e1f894e498c2dc50dbb44f66/flask/config.py#L139-L171) and check if the variables name is all upper case. We'd lose the check for `__` at the start that Redash's code does, but I couldn't find any settings that uses those.

redash/settings/__init__.py
jezdez: The `# noqa` comment prevents flake8 to mark this as an error since the value isn't used.  My editor is using flake8 for inline checks, do you want me to revert it?

redash/utils/routes.py
jezdez: I checked and it's actually the [`to_url`](https://github.com/pallets/werkzeug/blob/03c81ae0e7efbc83b7ea245fb42614795daca0cd/src/werkzeug/routing.py#L1251-L1264) method that would differ. I think it should be fine getting rid of it since it deals with encoding and all other parameters use the default converter anyway. Just didn't mean to break anything.

redash/metrics/celery.py
jezdez: Mostly because I wanted to prevent import time side-effects. The decorators were called by importing the `redash.metrics.celery` module in `redash/worker.py` which felt like bad practices especially since we're connecting another signal handler in `worker.py` already. In other words, I think it's cleaner in this case to keep the handlers separated, but all Celery initialization in `worker.py`.

redash/__init__.py
jezdez: This is mostly defensive code, e.g. flask-migrate imports alembic. I'll revert.

redash/__init__.py
jezdez: Ah, yes, I meant to do that. Good idea.

redash/utils/routes.py
jezdez: Removed it.

jezdez:
@arikfr Is there anything else needed to merge this?

redash/metrics/celery.py
arikfr: But now if we want to remove this module we need to do cleanup in two places, instead of just removing the module import. I never checked what connecting signals in Celery does, but I'm not sure I would qualify this as side effects.

redash/settings/__init__.py
arikfr: I was asking about the import itself, but after the fact noticed that it was there already just moved...

PR #3604, opened 2019-03-18
Fix time parameter and add time units for prometheus query runner

veerkat:

- [x] Feature
- [x] Bug Fix

## Description
Improve the prometheus query runner:
1. [Fix] Convert the time parameter to unix timestamp in the instant query.
2. Add the time units of grafana, it's very useful for range query.
3. Add the JSON query support, because it's more explicit than the Prometheus HTTP API query string, and also retain the API query string compatibility.

about the time units of grafana http://docs.grafana.org/reference/timerange/#time-units


arikfr:
Looks like you need to rebase?

PR #3609, opened 2019-03-19
Browser support config

### LINKED ISSUE #3579: Show message when trying to load Redash from an unsupported browser
Unsupported browsers: IE or old versions of Edge.

Need to update `index.html` to show a message instead of loading broken code.

ranbena:
## What type of PR is this?
- [x] Feature

Fixes #3579

## Description
Using [browserslist](https://github.com/browserslist/browserslist) to define which browser versions Redash supports.
List established from minimum "Async function" browser feature support ([caniuse](https://caniuse.com/#feat=async-functions)).
```json
// package.json
"browserslist": [
  "Edge >= 15",
  "Firefox >= 52",
  "Chrome >= 55",
  "Safari >= 10.1",
  "iOS >= 10.3",
  "Opera >= 42",
  "op_mob >= 46",
  "android >= 67",
  "and_chr >= 71",
  "and_ff >= 64",
  "and_uc >= 11.8",
  "samsung >= 6.2"
]
```

This list allows us to utilize some powerful tools (all adhering to list automatically):
- [x] **Babel transpiler** 🚯
  * js reduced by 2k (comparing app.js on netlify preview)
  * css no change (autoprefix still must use `-ms` for Edge browser)
  * although babel should drop "async func" and "arrow func" polyfills [according to this list](https://github.com/kangax/compat-table/blob/7572692166b1e75717c725aed8c8d83ca31493e0/data-es2016plus.js#L473-L488) - it doesn't. So I excluded these explicitly in `.babelrc`.
- [x] **Eslint rule** - error on unsupported browser feature usage 👮
  * Also available with stylelint but doesn't support less file so whatever man.
- [x] **Outdated browser detection** ~~with browserslist-useragent~~ only IE for now🚦
  * redirecting to a dedicated "browser not supported" static page (TBD).

### DIFF
--- a/client/app/components/app-view/index.js
+++ b/client/app/components/app-view/index.js
@@ -1,7 +1,13 @@
 import debug from 'debug';
+import { matchesUA, resolveUserAgent } from 'browserslist-useragent';
 import PromiseRejectionError from '@/lib/promise-rejection-error';
 import { ErrorHandler } from './error-handler';
 import template from './template.html';
+
+if (!matchesUA(navigator.userAgent, { allowHigherVersions: true })) {
+  // TODO: create "browser outdated" page and redirect to it
+  console.log('Your browser is outdated', resolveUserAgent(navigator.userAgent));
+}
 
 const logger = debug('redash:app-view');
 
--- a/client/app/components/app-view/index.js
+++ b/client/app/components/app-view/index.js
@@ -1,13 +1,7 @@
 import debug from 'debug';
-import { matchesUA, resolveUserAgent } from 'browserslist-useragent';
 import PromiseRejectionError from '@/lib/promise-rejection-error';
 import { ErrorHandler } from './error-handler';
 import template from './template.html';
-
-if (!matchesUA(navigator.userAgent, { allowHigherVersions: true })) {
-  // TODO: create "browser outdated" page and redirect to it
-  console.log('Your browser is outdated', resolveUserAgent(navigator.userAgent));
-}
 
 const logger = debug('redash:app-view');
 
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -10,7 +10,7 @@
         window.location.href = '/unsupported';
       }
       if (
-        navigator.userAgent.match('/MSIE') || // IE10
+        navigator.userAgent.match('MSIE') || // IE10
         navigator.appVersion.match('Trident/') // IE11
       ){
         onIE();
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -5,6 +5,18 @@
     <meta charset="UTF-8">
     <base href="/">
     <title>Redash</title>
+    <script>
+      function onIE() {
+        window.location.href = '/unsupported';
+      }
+      if (
+        navigator.userAgent.match('/MSIE') || // IE10
+        navigator.appVersion.match('Trident/') // IE11
+      ){
+        onIE();
+      }
+    </script>
+    <!--[if IE]><script>onIE()</script><![endif]-->
 
     <link rel="icon" type="image/png" sizes="32x32" href="/static/images/favicon-32x32.png">
     <link rel="icon" type="image/png" sizes="96x96" href="/static/images/favicon-96x96.png">
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -7,7 +7,7 @@
     <title>Redash</title>
     <script>
       function onIE() {
-        window.location.href = '/unsupported';
+        window.location.href = '/static/unsupported';
       }
       if (
         navigator.userAgent.match('MSIE') || // IE10
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "antd": "^3.12.3",
     "bootstrap": "^3.3.7",
     "brace": "^0.11.0",
-    "browserslist-useragent": "^3.0.0",
     "chroma-js": "^1.3.6",
     "classnames": "^2.2.6",
     "core-js": "^2.6.4",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "antd": "^3.12.3",
     "bootstrap": "^3.3.7",
     "brace": "^0.11.0",
+    "browserslist-useragent": "^3.0.0",
     "chroma-js": "^1.3.6",
     "classnames": "^2.2.6",
     "core-js": "^2.6.4",
@@ -104,6 +105,7 @@
     "eslint-import-resolver-webpack": "^0.8.3",
     "eslint-loader": "^2.1.1",
     "eslint-plugin-chai-friendly": "^0.4.1",
+    "eslint-plugin-compat": "^3.0.1",
     "eslint-plugin-cypress": "^2.0.1",
     "eslint-plugin-import": "^2.14.0",
     "eslint-plugin-jest": "^22.2.2",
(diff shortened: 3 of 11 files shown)


client/app/components/app-view/index.js
```diff
@@ -1,8 +1,14 @@
 import debug from 'debug';
+import { matchesUA, resolveUserAgent } from 'browserslist-useragent';
 import PromiseRejectionError from '@/lib/promise-rejection-error';
 import { ErrorHandler } from './error-handler';
 import template from './template.html';
 
+if (!matchesUA(navigator.userAgent, { allowHigherVersions: true })) {
+  // TODO: create "browser outdated" page and redirect to it
+  console.log('Your browser is outdated', resolveUserAgent(navigator.userAgent));
```
ranbena: Is this the place (and method) to put the redirect?

kravets-levko: Yes, it's the right place. And probably you don't need a page redirect: create a component with message or whatever, and conditionally show it instead of `<div ng-view>`

ranbena: But then we're still running potentially unsupported js for that page.
If we early redirect to a static html page, we don't have to worry about it.

kravets-levko: If you detect UA on client-side - you always have a chance to run unsupported code. If you'll change `app-view/template.html` to something like this - no change it will render header/footer/page:
```html
<div ng-if="$ctrl.isBrowserSupported">
  <!-- original template here -->
</div>
<div ng-if="!$ctrl.isBrowserSupported">
  Error! Error!!!! Unsupported browser!!!!111
</div>
```

kravets-levko: It's important that `app-view` component itself shouldn't use any unsupported features, as well as all code it uses via imports.

ranbena: Sounds easy and simple.
I'll do that and also add a cypress test to make sure it doesn't break.

ranbena: @kravets-levko does this make sense? (I'd rather it have it's own template)
```js
  ngModule.factory(
    '$exceptionHandler',
    () => function exceptionHandler(exception) {
      handler.process(exception);
    },
  );

  const isBrowserSupported = someCheck(...);
  ngModule.component('appView', {
    template: isBrowserSupported ? template : notSupportedTemplate,
    controller: isBrowserSupported ? AppViewComponent : NotSupportedCtrl,
  });
}
```

kravets-levko: @ranbena Yes, it's probably even better :+1:

arikfr:
I think the initial approach of redirect is the correct one. We should assume that an unsupported feature can break the whole bundle and cause the app not to load in which case we won’t show any message to the user.

To avoid possible issues we should do the detection in a separate file we will include in index.html. Actually we can avoid redirect and just render a message. But point is: it should not depend on the bigger code base or on any of it working.

-- 
Arik Fraimovich

ranbena:
> I think the initial approach of redirect is the correct one.

Best practice would be to determine and redirect in `static.py`.
```py
@routes.route(org_scoped_rule('/<path:path>'))
@routes.route(org_scoped_rule('/'))
@supported_browser_required  <--
@login_required
def index(**kwargs):
    return render_index()
```
but that would require re-implementing `browserlist-useragent` in Python (currently available only in js and ruby).

But going with JS has a clear problem - it's not a standalone file, but a js module.

Options:
1. Back down and import it into `app-view.js`.
2. Implement it in python (and perhaps contribute to the open source project)
3. Reimplement as standalone js file to be loaded in `index.html`.

None are perfect, my personal pref is #2 (although not my comfort zone).
Wdyt @arikfr?

arikfr:
>  Eslint rule - error on unsupported browser feature usage 👮

If we only use browser supported features, what does Babel transpile?

ranbena:
> If we only use browser supported features, what does Babel transpile?

It's a good question though and I'd like to know as well. My understanding is that it does not polyfill all features and if a feature is polyfilled (either with dedicated babel plugin / npm package / own util) we need to be explicitly define it to be excluded from the linter.

ranbena:
#### What I want to do
Generate `browserslist-useragent.js` file that suits old browsers to be loaded in `client/app/index.html`.

#### The plan
Define an npm script `unsupported:build` which transpiles the `browserslist-useragent` npm package with a dedicated babel config into a single file.

@arikfr
1. Does this make sense?
2. Can babel cli transpile a node module with its dependencies? Or can webpack be used to so that?
3. How do you provide babel with a preset override? (e.g. `targets: [ "ie": 8 ]`)

arikfr:
This is what webpack is for. Not sure about the overrides but we can have a dedicated configuration for it (maybe in a separate project?).

But I feel like we’re reinventing the wheel here. Is there some prior art on how this is done?

-- 
Arik Fraimovich

ranbena:
Tested the redirect method in IE10 and 11 but the rest later this week (with windows machine).

client/app/index.html
```diff
@@ -5,6 +5,18 @@
...
+        window.location.href = '/unsupported';
+      }
+      if (
+        navigator.userAgent.match('/MSIE') || // IE10
+        navigator.appVersion.match('Trident/') // IE11
+      ){
+        onIE();
+      }
+    </script>
+    <!--[if IE]><script>onIE()</script><![endif]-->
```
ranbena: IE only for now, consider using https://github.com/browserslist/browserslist-useragent in the future.

ranbena:
@arikfr made html page static, no template. Lmk if this is what you meant.

ranbena:
Damnit, another codeclimate eslint plugin problem..
Might have to do with eslint channel compatibility 😒
@arikfr any reason not to switch to CircleCI step instead (assuming it's problem free)?

ranbena:
Also, @arikfr I put the redirect script in `index.html` and `multi_org.html` (dunno what that is).
Should I put it in `signed_out.html` (a main template) as well?

ranbena:
@arikfr I don't get the [pep8 error here](https://codeclimate.com/github/getredash/redash/pull/3609) - there are actually 2 lines..

package.json
```diff
@@ -162,5 +163,20 @@
         "npm run test -- --bail --findRelatedTests"
       ]
     }
-  }
+  },
+  "//": "browserslist set to 'Async functions' compatibility",
+  "browserslist": [
```
arikfr: How about using configuration like we had before (`> 0.5%, last 2 versions, Firefox ESR, ie 11, not dead`) so we don't have to update this list everytime browser versions update? I mean not necessarily the same one, but along the lines of `> 0.5%, last 2 versions` etc instead of hardcoded versions.

redash/templates/unsupported.html
```diff
@@ -0,0 +1,61 @@
+<!DOCTYPE html>
```
arikfr: This file should be in `client/app` and we need to make sure webpack copies it into dist.

package.json
ranbena: > so we don't have to update this list every time browser versions update

Why would you need to? We should revisit this list when:
1. There are lower browser versions we would like to support. Won't happen cause we've looked at the browser stats.
2. As a performance optimization to lower `app.js` file size. That can be done periodically.

> instead of hardcoded version

The advantage of hardcoded versions is that they allow to easily and reliably determine if a user's browser is supported or not. But if we define sth like "last 2 versions" you would have no clear cutoff point (cause under last 2 versions doesn't mean you can't use Redash).

redash/templates/unsupported.html
ranbena: Done.

Just realized `index.html` and `multi_org.html` are exactly the same. Intentional?

arikfr: You still need a Flask route serving it.

ranbena: The page is configured as a static asset `https://domain.com/static/unsupported.html` - so doesn't need it, nay?

arikfr: You're right. I didn't notice you updated the URL 👍

arikfr:
👍

PR #3610, opened 2019-03-19
Replace toastr with Ant Notification

### LINKED ISSUE #3547: Replace toastr with Ant component
- [ ] Pick Ant component to use ([`<Notification>`](https://ant.design/components/notification/) or [`<Message>`](https://ant.design/components/message/), maybe both?).
- [ ] Replace calls to `toastr` with one of the above.
- [ ] Remove `angular-toastr` from dependencies.

gabrieldutra:

- [x] Refactor

## Description
I've considered using the `Message` component for some cases and I've even given it a try, however as @kravets-levko mentioned in #3547, it's annoying sometimes (even to small messages). Notification seemed a better fit anyway.

What this does:
- Creates a `notification.js` to set our defaults to Ant `Notification` and replace function calls to be more similar to toastr (made life easier to change stuff + I think it will be easier to use);
- Replaces all `toastr` calls to use `notification`;
- Removes `toastr` from our dependencies :slightly_smiling_face:  

Closes #3547 

replace-toastr


client/app/pages/data-sources/show.js
```diff
@@ -1,20 +1,21 @@
 import { find } from 'lodash';
 import debug from 'debug';
 import template from './show.html';
+import notification from '@/services/notification';
 
 const logger = debug('redash:http');
 export const deleteConfirm = { class: 'btn-warning', title: 'Delete' };
-export function logAndToastrError(deleteObject, httpResponse, toastr) {
+export function logAndNotificateError(deleteObject, httpResponse) {
```
ranbena:
```js
notification.error('Notificate isn\'t a word', 'Consider logAndNotifyError', { duration: null });
```

client/app/pages/queries/view.js
```diff
@@ -267,13 +267,14 @@ function QueryViewCtrl(
 
             AlertDialog.open(title, message, confirm).then(overwrite);
           } else {
-            toastr.error(
-              errorMessage + ' Please copy/backup your changes and reload this page.',
-              { autoDismiss: false },
+            notification.error(
+              errorMessage,
+              ' Please copy/backup your changes and reload this page.',
```
ranbena: `errorMessage` doesn't feel like title material. Perhaps sth like:

```js
notification.error(
  'Changes not saved',
  errorMessage + ' Please copy/backup your changes and reload this page.',
  { duration: null },
);
```

client/app/services/notification.js
```diff
@@ -0,0 +1,22 @@
...
+  'warn',
+].reduce((acc, action) => ({
+  ...acc,
+  [action]: (message, description = null, props = null) => notification[action]({ ...props, message, description }),
+}), null);
+
+export default { // export Ant's notification and replace actions
+  ...notification,
+  ...updatedActions,
+};
```
ranbena: For notification consistency, consider some constants to be used across components:
```js
{
  SAVE_SUCCESS: 'Saved',
  SAVE_ERROR: 'Failed saving',
  ...
}
```
```js
notification.error(notification.SAVE_ERROR, 'Sth went wrong');
```

client/app/services/notification.js
```diff
@@ -0,0 +1,22 @@
...
+const updatedActions = [
+  'success',
+  'error',
+  'info',
+  'warning',
+  'warn',
+].reduce((acc, action) => ({
+  ...acc,
+  [action]: (message, description = null, props = null) => notification[action]({ ...props, message, description }),
+}), null);
```
arikfr: This is really clever, but I have to admit that I had to read this carefully and really think about what's going on here. I kind of knew what it should do, but still had to unpack all the logic in my head.

How about the following as a simplified solution?

```javascript
const simpleNotification = {};

['success', 'error', ...].forEach(action => {
  simpleNotification[action] = (message, description = null, props = null) => notification[action]({ ...props, message, description });
});

export default {...notification, ...simpleNotification};
```

client/app/services/user.js
```diff
@@ -33,18 +34,14 @@ function disableUser(user) {
   return $http
     .post(disableResource(user))
     .then((data) => {
-      toastr.warning(`User <b>${userName}</b> is now disabled.`, { allowHtml: true });
+      notification.warning(`User ${userName} is now disabled.`);
```
ranbena: I dunno if this should be a warning.. maybe `notification.info`?

client/app/services/user.js
```diff
@@ -53,16 +50,12 @@ function deleteUser(user) {
   return $http
     .delete(`api/users/${user.id}`)
     .then((data) => {
-      toastr.warning(`User <b>${userName}</b> has been deleted.`, { allowHtml: true });
+      notification.warning(`User ${userName} has been deleted.`);
```
ranbena: Same here. `info` is better I think.

client/app/services/notification.js
ranbena: +1

ranbena:
I think the default fixed width is not great.

#### Short messages are too distracting and needlessly bulky
Screen Shot 2019-03-20 at 9 33 49
"Go away! I'm trying to click on a button behind you!"

####  But with flexible width, a bit less
Screen Shot 2019-03-20 at 9 33 56

####  Long texts will break line
Screen Shot 2019-03-20 at 9 37 42

#### But not with flexible width
Screen Shot 2019-03-20 at 9 37 20

🎁
```css
.ant-notification {
  width: auto;
}
.ant-notification-notice {
  padding-right: 48px;
}
```

client/app/services/user.js
kravets-levko: @ranbena in this case it's okay - we show message that user was disabled, and admin should see that they probably did something unsafe

client/app/services/notification.js
kravets-levko: BTW - how about code completion? Does your IDE recognize this as object with methods?

client/app/services/notification.js
kravets-levko: This is an interesting idea, but not sure if we really need it. At least, we should check if we have two or more messages with the same titles.

ranbena: We do. Many.
And some others are variations which I think is worth consolidating.

ranbena: But, it's definitely out-of-scope here. Future thinking.

client/app/pages/data-sources/show.js
gabrieldutra: It seems I've started to create new words :sweat_smile: 

oops

client/app/services/user.js
gabrieldutra: I thought it annoying with toastr, but with notification it turns out to be okay (that's why I've opted not to change it during the replacement)

IMO both options are suitable here, but if you still think it's better with `info` I can simply change that @ranbena  :)

client/app/services/notification.js
kravets-levko: @gabrieldutra  So when you type `notification.` somewhere in code, and then press `Ctrl+Space` - it shows `success`, `error`, ... in dropdown? Did I understand you right? That's not something crucial, but usually nice to have.

gabrieldutra: Sorry @kravets-levko, I actually didn't write all the stuff that were on my mind :sweat_smile:. But what I forgot to mention is that code completion (Either with `ctrl+space` or just pressing a dot after `notification`) works for VSCode.

However, I had another issue... it only recognized the code and showed suggestions when I used the Relative Path to import (`import notification from '../../services/notification'`). Not sure about the reason, but I may be missing some extension here.

client/app/services/notification.js
gabrieldutra: We need this for sure :+1:. I think for now we can add both `SAVE_SUCCESS` and `SAVE_ERROR` as they are a direct change in code and wouldn't be out-of-scope. This will at least define a future pattern and we can later modify existing notifications, WDYT?

ranbena: I think it has value for ux, but I'd rather land this as is so we could move on to the next angular2react component.

client/app/services/user.js
ranbena: Let's leave as is.

PR #3611, opened 2019-03-20
Change has_access and require_access signatures

rauchy:
This was actually part of #3495, but it's spread across many files and is a simple refactor so I just wanted to take it out to it's own PR in order to clean up #3495.

- [x] Refactor

#3495

### DIFF
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -181,7 +181,9 @@ def require_access_to_dropdown_queries(user, query_def):
         if len(groups) < len(dropdown_query_ids):
             abort(400, message='You are trying to associate a dropdown query that does not have a matching group. Please verify the dropdown query id you are trying to associate with this query.')
 
-        require_access(dict(groups), user, view_only)
+        object = lambda: None
+        object.groups = dict(groups)
+        require_access(object, user, view_only)
 
 class QueryListResource(BaseQueryListResource):
     @require_permission('create_query')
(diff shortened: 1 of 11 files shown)


redash/permissions.py
```diff
@@ -14,24 +14,24 @@
 ACCESS_TYPES = (ACCESS_TYPE_VIEW, ACCESS_TYPE_MODIFY, ACCESS_TYPE_DELETE)
 
 
-def has_access(object_groups, user, need_view_only):
+def has_access(object, user, need_view_only):
```
arikfr: The purpose of switching to passing the `object` instead of `object.groups` is to be able to use a different logic when needed. For example: when using a query API key just assert that it "belongs" to this query.

Only changing where we pass the groups, means we still have the need to set pseudo groups on the API Key object, which is what we wanted to resolve. If I'm not mistaken, in this form we don't prevent query API key from running _any_ parameter.

redash/handlers/queries.py
```diff
@@ -181,7 +181,9 @@ def require_access_to_dropdown_queries(user, query_def):
         if len(groups) < len(dropdown_query_ids):
             abort(400, message='You are trying to associate a dropdown query that does not have a matching group. Please verify the dropdown query id you are trying to associate with this query.')
 
-        require_access(dict(groups), user, view_only)
+        object = lambda: None
+        object.groups = dict(groups)
+        require_access(object, user, view_only)
```
arikfr: This feels wrong. Let's add `require_access_to_groups` that takes groups?

Btw, `object` is a reserved word.

rauchy: > Btw, object is a reserved word.
You sure? I would expect things to not work if it were a reserved keyword.

arikfr:
```
>>> object
<type 'object'>
```

Python is funny in this way. `id` is reserved too and often we override it locally.

arikfr: I guess that reserved word is not the correct term here. Just a built-in object.

rauchy: Yeah, while not reserved, it might cause confusion. I'll rename.

arikfr:
:shipit:

PR #3618, opened 2019-03-21
Query Results fails to use query which has double quotes in column names

kravets-levko:

- [x] Bug Fix

## Description

If source query contains column with double quote (`"`) - Query Results fails to build temporary table with "Syntax error" message.


kravets-levko:
@rauchy I tested manually :smile: But being serious - you're right, I'll add some.

PR #3620, opened 2019-03-21
Widget resize tests

ranbena:
- [x] Other - Tests

## Description
Screen Shot 2019-03-23 at 15 46 54

### DIFF
--- a/client/app/pages/dashboards/dashboard.html
+++ b/client/app/pages/dashboards/dashboard.html
@@ -96,8 +96,8 @@
       ng-class="{'preview-mode': !$ctrl.layoutEditing, 'editing-mode': $ctrl.layoutEditing}">
       <div class="dashboard-widget-wrapper"
         ng-repeat="widget in $ctrl.dashboard.widgets track by widget.id"
-        gridstack-item="widget.options.position" gridstack-item-id="{{ widget.id }}">
-        <div class="grid-stack-item-content" data-test="WidgetId{{ widget.id }}">
+        gridstack-item="widget.options.position" gridstack-item-id="{{ widget.id }}"  data-test="WidgetId{{ widget.id }}">
+        <div class="grid-stack-item-content">
           <dashboard-widget widget="widget" dashboard="$ctrl.dashboard" on-delete="$ctrl.removeWidget(widget.id)"></dashboard-widget>
         </div>
       </div>
--- a/client/cypress/integration/dashboard/dashboard_spec.js
+++ b/client/cypress/integration/dashboard/dashboard_spec.js
@@ -1,4 +1,6 @@
 const DRAG_PLACEHOLDER_SELECTOR = '.grid-stack-placeholder';
+const RESIZE_HANDLE_SELECTOR = '.ui-resizable-se';
+
 
 function createNewDashboardByAPI(name) {
   return cy.request('POST', 'api/dashboards', { name })
@@ -98,6 +100,34 @@ function dragBy(wrapper, offsetTop = 0, offsetLeft = 0) {
     .then(() => ({
       left: end.left - start.left,
       top: end.top - start.top,
+    }));
+}
+
+function resizeBy(wrapper, offsetTop = 0, offsetLeft = 0) {
+  let start;
+  let end;
+  let from;
+  const getSize = $el => ({ height: $el.height(), width: $el.width() });
+
+  return wrapper
+    .then(($el) => {
+      start = getSize($el);
+    })
+    .within(() => cy.get(RESIZE_HANDLE_SELECTOR))
+    .then(($handle) => {
+      from = $handle.show().offset(); // turn on handle and get it's position
+      return wrapper
+        .trigger('mouseover')
+        .trigger('mousedown', { pageX: from.left, pageY: from.top, which: 1 })
+        .trigger('mousemove', { pageX: from.left + offsetLeft, pageY: from.top + offsetTop, which: 1 });
+    })
+    .then(() => {
+      end = getSize(Cypress.$(DRAG_PLACEHOLDER_SELECTOR)); // see comment in dragBy ^^
+      return wrapper.trigger('mouseup');
+    })
+    .then(() => ({
+      height: end.height - start.height,
+      width: end.width - start.width,
     }));
 }
 
@@ -268,7 +298,7 @@ describe('Dashboard', () => {
 
     describe('Draggable', () => {
       describe('Grid snap', () => {
-        beforeEach(function () {
+        beforeEach(() => {
           editDashboard();
         });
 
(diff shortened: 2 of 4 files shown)


client/app/pages/dashboards/dashboard.html
```diff
@@ -96,8 +96,8 @@ <h3>
       ng-class="{'preview-mode': !$ctrl.layoutEditing, 'editing-mode': $ctrl.layoutEditing}">
       <div class="dashboard-widget-wrapper"
         ng-repeat="widget in $ctrl.dashboard.widgets track by widget.id"
-        gridstack-item="widget.options.position" gridstack-item-id="{{ widget.id }}">
-        <div class="grid-stack-item-content" data-test="WidgetId{{ widget.id }}">
+        gridstack-item="widget.options.position" gridstack-item-id="{{ widget.id }}"  data-test="WidgetId{{ widget.id }}">
+        <div class="grid-stack-item-content">
```
ranbena: Moved to parent node in order to locate the handle (which is a sibling)

https://github.com/getredash/redash/blob/689e1a63bff22e5e7cdc354890e5c576cd307ea1/cypress/integration/dashboard/dashboard_spec.js#L112-L116

ranbena:
wip - I'm adding another resize related test

ranbena:
Added 2 tests for auto height cause resizing disables it.

ranbena:
@gabrieldutra this is ready for review

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -333,6 +363,104 @@ describe('Dashboard', () => {
...
+          editDashboard();
+        });
+
+        it('stays put when dragged under snap threshold', () => {
+          resizeBy(cy.get('@textboxEl'), 0, 90).then((delta) => {
+            expect(delta.width).to.eq(0);
+          });
+        });
+
+        it('moves one column dragged over snap threshold', () => {
```
gabrieldutra:
```suggestion
        it('moves one column when dragged over snap threshold', () => {
```

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -333,6 +363,104 @@ describe('Dashboard', () => {
...
+          resizeBy(cy.get('@textboxEl'), 0, 110).then((delta) => {
+            expect(delta.width).to.eq(200);
+          });
+        });
+
+        it('moves two column dragged over snap threshold', () => {
```
gabrieldutra:
```suggestion
        it('moves two columns when dragged over snap threshold', () => {
```

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -333,6 +363,104 @@ describe('Dashboard', () => {
...
+          editDashboard();
+        });
+
+        it('stays put when dragged under snap threshold', () => {
+          resizeBy(cy.get('@textboxEl'), 10, 0).then((delta) => {
+            expect(delta.height).to.eq(0);
+          });
+        });
+
+        it('moves one row dragged over snap threshold', () => {
```
gabrieldutra:
```suggestion
        it('moves one row when dragged over snap threshold', () => {
```

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -101,6 +103,34 @@ function dragBy(wrapper, offsetTop = 0, offsetLeft = 0) {
     }));
 }
 
+function resizeBy(wrapper, offsetTop = 0, offsetLeft = 0) {
```
gabrieldutra: I didn't think about it before in the Drag tests, but have you considered having those offsets inverted? My intuition says that the order is "width - height" (Not a good argument, but lmk what you think :laughing:)

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -333,6 +363,104 @@ describe('Dashboard', () => {
...
+          });
+        });
+
+        it('moves one row dragged over snap threshold', () => {
+          resizeBy(cy.get('@textboxEl'), 30, 0).then((delta) => {
+            expect(delta.height).to.eq(50);
+          });
+        });
+
+        it('shrink to minimum', () => {
```
gabrieldutra:
```suggestion
        it('shrinks to minimum', () => {
```

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -395,6 +523,71 @@ describe('Dashboard', () => {
...
+          // add 4 table rows
+          cy.get('@paramInput').clear().type('5');
+          cy.get('@refreshButton').click();
+          cy.wait('@FreshResults', { timeout: 10000 });
+
+          // expect to height to grow by 1 grid grow
+          cy.get('@widget').invoke('height').should('eq', 435);
+        });
+
+        it('auto height revoked after manual height adjustment', () => {
```
gabrieldutra:
```suggestion
        it('revokes auto height after manual height adjustment', () => {
```

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -395,6 +523,71 @@ describe('Dashboard', () => {
             .should('eq', 335);
         });
       });
+
+      describe('adjusts height on refresh', () => {
```
gabrieldutra: This one went a bit away from the use of `describe` so far, how about this:
```suggestion
      describe('Height behavior on refresh', () => {
```

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -395,6 +523,71 @@ describe('Dashboard', () => {
...
+          // resize height by 1 grid row
+          resizeBy(cy.get('@widget'), 5);
+          cy.get('@widget').invoke('height').should('eq', 335);
+
+          // add 4 table rows
+          cy.get('@paramInput').clear().type('5');
+          cy.get('@refreshButton').click();
+          cy.wait('@FreshResults');
+
+          // expect height to stay unchanged (would have been 435)
```
gabrieldutra: Just to show a different way to assert the same thing:
```javascript
expect(() => {
  // add 4 table rows
  cy.get('@paramInput').clear().type('5');
  cy.get('@refreshButton').click();
  cy.wait('@FreshResults');
}).not.to.change(this.widget, 'height');
```

client/cypress/integration/dashboard/dashboard_spec.js
kravets-levko: @gabrieldutra How do you deal with CSS (top-right-bottom-left)? :grin:

ranbena: Yeah I don't know why I did that :) I'll switch em.

client/cypress/integration/dashboard/dashboard_spec.js
ranbena: Oooh that's nice.

ranbena: But I don't think this specifically would work cause
1. `height` is a method, not a property.
2. `this.widget` is a cy object, not jquery.

Right?

ranbena: Ideally the `dragBy` and `resizeBy` functions would not return a delta value but instead:
```js
it('moves one column', () => {
  const getLeft = () => $el.offset().left;
  const drag = () => dragBy($el, 200);

  expect(drag).to.increase(getLeft).by(200);
});
```
Then `resizeBy` would simply be a `dragBy` of the element's handle. So nice.

BUT, disconnecting the drag from the delta calculation is currently problematic since the dragged element animates to its new position on mouseup (compensated by calculating the drag placeholder position instead).
Any workaround suggestion? (wait for animation end, disable animation for test, ...)

gabrieldutra: Really cool, once I get to my laptop I'll run a few tests with it. For the questions:

> But I don't think this specifically would work cause

It worked for success, need to try on failure haha. The assertion is essentially the same, they just run it before the function and then guarantee it's the same after.

> height is a method, not a property

I hope not to have seen the above wrongly, but this is not supposed to be a problem

> this.widget is a cy object, not jquery.

Need to confirm, but I think it is a jQuery object, when I saw it in docs it seemed the same as getting the result of `cy.get('@widget')`

ranbena: If you can make it work that would be awesome 👍

gabrieldutra: I could not :sweat_smile:. That works well with javascript objects, not with elements (I could use a helper object to represent the element, but I don't think it is worth it for my example).

Yours is much better, it would make tests a lot more readable.

> Any workaround suggestion?

Can we use the placeholder position for the `getLeft` method? If not, NP, we can move on.

ranbena: > Can we use the placeholder position for the getLeft method? If not, NP, we can move on.

It doesn't get rendered in the dom until mousedown (or mousemove, I didn't check) so it's not ideal.

PR #3622, opened 2019-03-21
Add env var to skip Flask rate limits

gabrieldutra:

- [x] Other

## Description
_This was supposed to be a Draft PR, though I pressed the enter key accidentally :confused:_ 

Well, the idea here is to suggest a fix to "Too Many Requests" error in Cypress ([this](https://circleci.com/gh/getredash/redash/17784)).

This adds `REDASH_SKIP_LOGIN_RATE_LIMIT` env var (not sure if this will be the best name here) and uses it in the `docker-compose`'s for dev and for cypress.


tests/__init__.py
```diff
@@ -13,6 +13,9 @@
 os.environ['REDASH_GOOGLE_CLIENT_SECRET'] = "dummy"
 os.environ['REDASH_MULTI_ORG'] = "true"
 
+# Disable login rate limit skip
+os.environ['REDASH_SKIP_LOGIN_RATE_LIMIT'] = "false"
```
gabrieldutra: This one is for the `test_throttle_login` not to fail during tests, but LMK if you think it's better to remove the test or sth else.

redash/settings/__init__.py
```diff
@@ -150,6 +150,7 @@ def all_settings():
 # See https://flask-limiter.readthedocs.io/en/stable/#rate-limit-string-notation
 
 THROTTLE_LOGIN_PATTERN = os.environ.get('REDASH_THROTTLE_LOGIN_PATTERN', '50/hour')
+SKIP_LOGIN_RATE_LIMIT = parse_boolean(os.environ.get('REDASH_SKIP_LOGIN_RATE_LIMIT', 'false'))
```
ranbena: For consistent terminology I'd name it to `REDASH_THROTTLE_LOGIN_ENABLED`

tests/__init__.py
ranbena: `test_throttle_login` tests the feature but assumes configuration.
Imo it should be split into 3 tests:
1. Check that feature is configured on the server in a certain way (true, 50/hour).
2. Expect 429 error when feature is enabled.
3. Expect no error when feature is disabled.

Thus, setting the env var won't be located here, but inside the tests 2 and 3.

docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
       REDASH_LOG_LEVEL: "INFO"
       REDASH_REDIS_URL: "redis://redis:6379/0"
       REDASH_DATABASE_URL: "postgresql://postgres@postgres/postgres"
+      REDASH_THROTTLE_LOGIN_ENABLED: "false"
```
ranbena: Not `"true"`?

tests/__init__.py
```diff
@@ -13,6 +13,9 @@
 os.environ['REDASH_GOOGLE_CLIENT_SECRET'] = "dummy"
 os.environ['REDASH_MULTI_ORG'] = "true"
 
+# Disable login rate limit skip
+os.environ['REDASH_THROTTLE_LOGIN_ENABLED'] = "true"
```
ranbena: If the purpose of the current test is to:
a) test the feature, then I think this should be set in/before the test.
b) make sure prod has this feature turned on, then this should not be set at all.

docker-compose.yml
gabrieldutra: This one is for dev :). Almost sure you don't want a login limit here

ranbena: So this would be the prod setting?
https://github.com/getredash/redash/blob/5760ef13eb2c9d5bbaf9eb26f7f34892aafce722/redash/settings/__init__.py#L153

tests/__init__.py
gabrieldutra: > arikfr marked this conversation as resolved.

@arikfr ? :man_shrugging:

ranbena: Arik: "I'll click resolve and run away before anyone notices" 💨

docker-compose.yml
gabrieldutra: > So this would be the prod setting?

That's the default setting, thus prod setting as prod doesn't have this set

ranbena: Got it 👍
The syntax here is confusing. I thought it works like `lodash.get`.

gabrieldutra:
Thanks @arikfr, I've modified this to disable all rate limits instead of only login. Two questions to proceed:
- "The other change is to override this setting in the rate limit change.". I'm not sure I understood this one, you mean force `RATELIMIT_ENABLED` to be true when `THROTTLE_LOGIN_PATTERN` env var is set?
- Bringing up Ran's comment about tests, I'm currently just setting rate limits prior to all tests. Should I cover all cases? (My concerns are: - existing test purpose; - it's a bigger context now that all rate limits can be disabled)

ranbena:
@gabrieldutra @arik plz let's give this one priority as some of my pending PRs depend on it.

arikfr:
Just resolve conflicts and it's good to go.

gabrieldutra:
> Just resolve conflicts and it's good to go.

Will do in a min :slightly_smiling_face:

gabrieldutra:
Done, just review the name (almost sure this is what I did :sweat_smile:) and once CI is finished it's good to go

PR #3625, opened 2019-03-22
Fix Celery worker --max-tasks-per-child for Celery 4.x.

jezdez:

- [x] Feature
- [x] Bug Fix

## Description

This fixes the name of the `--max-tasks-per-child` parameter whose name was changed in Celery 4.x and adds a `--max-memory-per-child` of 1/4 of total memory.

Port of https://github.com/mozilla/redash/pull/364

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -4,7 +4,7 @@ set -e
 worker() {
   WORKERS_COUNT=${WORKERS_COUNT:-2}
   QUEUES=${QUEUES:-queries,scheduled_queries,celery,schemas}
-  WORKER_EXTRA_OPTIONS=${WORKER_EXTRA_OPTIONS:}
+  WORKER_EXTRA_OPTIONS=${WORKER_EXTRA_OPTIONS:-}
 
   echo "Starting $WORKERS_COUNT workers for queues: $QUEUES..."
   exec /usr/local/bin/celery worker --app=redash.worker -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 -Ofair $WORKER_EXTRA_OPTIONS
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -6,7 +6,7 @@ worker() {
   QUEUES=${QUEUES:-queries,scheduled_queries,celery,schemas}
 
   echo "Starting $WORKERS_COUNT workers for queues: $QUEUES..."
-  exec /usr/local/bin/celery worker --app=redash.worker -c$WORKERS_COUNT -Q$QUEUES -linfo --maxtasksperchild=10 -Ofair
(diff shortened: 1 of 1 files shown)


bin/docker-entrypoint
```diff
@@ -4,9 +4,10 @@ set -e
 worker() {
   WORKERS_COUNT=${WORKERS_COUNT:-2}
   QUEUES=${QUEUES:-queries,scheduled_queries,celery,schemas}
+  MAX_MEMORY=$(($(/usr/bin/awk '/MemTotal/ {print $2}' /proc/meminfo)/4))
```
arikfr: I'm not sure this is a good default to have. 

Maybe just add the option to pass additional options to `celery worker` command? Then you can pass this one or whatever else you might need.

bin/docker-entrypoint
```diff
@@ -4,9 +4,10 @@ set -e
...
 
   echo "Starting $WORKERS_COUNT workers for queues: $QUEUES..."
-  exec /usr/local/bin/celery worker --app=redash.worker -c$WORKERS_COUNT -Q$QUEUES -linfo --maxtasksperchild=10 -Ofair
+  exec /usr/local/bin/celery worker --app=redash.worker -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 --max-memory-per-child=$MAX_MEMORY -Ofair
```
arikfr: So `--maxtasksperchild` was simply ignored or they had some backward compatibility support in place?

jezdez: I don't know, I'm just porting a change @jasonthomas has applied to our fork.

bin/docker-entrypoint
jezdez: Sure.

bin/docker-entrypoint
jasonthomas: There is backward compatibility in place as per [1].

[1] https://github.com/celery/celery/blob/8047ca9e818db8151dc6c80814600797ab94f5bc/docs/history/whatsnew-4.0.rst#worker

jezdez: Thanks @jasonthomas!

bin/docker-entrypoint
```diff
@@ -4,9 +4,10 @@ set -e
 worker() {
   WORKERS_COUNT=${WORKERS_COUNT:-2}
   QUEUES=${QUEUES:-queries,scheduled_queries,celery,schemas}
+  WORKER_EXTRA_OPTIONS=${WORKER_EXTRA_OPTIONS:}
```
jezdez: @jasonthomas Is this a legitimate shell script?

jasonthomas: No, it is throwing a bad substitution error. It should be `WORKER_EXTRA_OPTIONS=${WORKER_EXTRA_OPTIONS:-}` instead.

arikfr:
Thanks.

PR #3626, opened 2019-03-22
Fix a few more inconsistencies when loading and dumping JSON.

jezdez:

- [x] Bug Fix

## Description

These change have been added since c2429e92d2a5dd58cbe378b9cc06f3be969f747e.

Refs #2807. Original work in: #2817

### DIFF
--- a/redash/authentication/jwt_auth.py
+++ b/redash/authentication/jwt_auth.py
@@ -1,8 +1,7 @@
 import logging
 import jwt
 import requests
-
-from redash.utils import json_dumps
+import simplejson
 
 logger = logging.getLogger('jwt_auth')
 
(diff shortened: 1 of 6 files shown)


redash/authentication/jwt_auth.py
```diff
@@ -21,7 +22,7 @@ def get_public_keys(url):
         if 'keys' in data:
             public_keys = []
             for key_dict in data['keys']:
-                public_key = jwt.algorithms.RSAAlgorithm.from_jwk(json.dumps(key_dict))
+                public_key = jwt.algorithms.RSAAlgorithm.from_jwk(json_dumps(key_dict))
```
arikfr: Not sure we should use `json_dumps` here.

jezdez: Why? `json_dumps` is just a pass through for with our JSONEncoder which won't get triggered with the payload here.

This is just a step to make sure we use `simplejson` everywhere and not the outdated Python 2.x stdlib library. Once we're on Python 3 we can revert the use of simplejson to the stdlib version of course.

jezdez: Just changing that to simplejson.dump instead.

arikfr:
Thanks.

Issue #3602, opened 2019-03-18
Queries search is not working on release/7.0.x


veerkat:
### Issue Summary

I try building docker image with the redash dockerfile and start by docker-compose. But queries search is not working

### Steps to Reproduce

![Untitled](https://user-images.githubusercontent.com/6704470/54521288-93d2ad80-49a5-11e9-9794-94572b6698d0.gif)

Check out the queries table and find that search_vector is null.
![1](https://user-images.githubusercontent.com/6704470/54521423-de542a00-49a5-11e9-9f37-5b5923042973.png)

### Technical details:

* Redash Version: release/7.0.x
* Browser/OS: Chrome Mac
* How did you install Redash: redash setup

arikfr:
How did you build the database?

veerkat:
> How did you build the database?

use redash/setup/docker-compose.yml

```
postgres:
  image: postgres:9.5.6-alpine
  env_file: /opt/redash/env
  volumes:
    - /opt/redash/postgres-data:/var/lib/postgresql/data
  restart: always
```

arikfr:
Yes, but how did you create the database tables?

veerkat:
> Yes, but how did you create the database tables?

Replace the original image with the built image in redash/setup/docker-compose.yml,
and run in redash/setup
```
docker-compose -f docker-compose.yml -p redash run --rm server create_db
```

arikfr:
Just to make sure: this is a new install not an upgrade, correct?

veerkat:
> Just to make sure: this is a new install not an upgrade, correct?

Yes, just to test the new version

arikfr:
Thanks. I managed to confirm this :-(

@jezdez maybe we need to setup something for the trigger to be created when creating DB from scratch (using the `create_db` command)?

jezdez:
Ah, right, we simply have to run the migrations in the `create_tables` command so this gets picked up. I don't think there is a better way to cover ongoing schema changes than the migration system, it's built for that exact purpose. I don't recall why `create_tables` was written like this to be honest.

arikfr:
Because from my experience migrations don't age well. Usually over time they tend to break. 

When creating a database from scratch the ORM usually does a better job by creating the tables from their definitions than trying to mutate the DB over all the different versions we ever had.

jezdez:
[...]

PR #3633, opened 2019-03-24
E2E test for query search

arikfr:

- [x] E2E test

## Description

This is a very basic search test to reproduce #3602.

#3631 should have this test passing.

### DIFF
--- a/cypress/integration/query/search_query_spec.js
+++ b/cypress/integration/query/search_query_spec.js
@@ -1,0 +1,24 @@
+describe("Search Query", () => {
+  beforeEach(() => {
+    cy.login();
+
+    cy.request("POST", "api/queries", {
+      name: "Users Count",
+      data_source_id: 1,
+      query: "SELECT 1"
+    });
+
+    cy.request("POST", "api/queries", {
+      name: "Dashboards Count",
+      data_source_id: 1,
+      query: "SELECT 1"
+    });
+
+    cy.visit("/");
+  });
+
+  it("finds queries by name", () => {
+    cy.getByTestId("AppHeaderSearch").type("Users{enter}");
+    cy.contains("Users Count");
+  });
+});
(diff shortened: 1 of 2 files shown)


cypress/integration/query/search_query_spec.js
```diff
@@ -0,0 +1,24 @@
+describe("Search Query", () => {
```
arikfr: Once I see a _failing_ build for this PR, I'm going to merge this so I can test #3631. But please do take a moment to review it and suggest improvements if needed.

cypress/integration/query/search_query_spec.js
```diff
@@ -0,0 +1,24 @@
...
+    cy.login();
+
+    cy.request("POST", "api/queries", {
+      name: "Users Count",
+      data_source_id: 1,
+      query: "SELECT 1"
+    });
+
+    cy.request("POST", "api/queries", {
+      name: "Dashboards Count",
```
arikfr: I wanted to add a `not contains` assertion in my test, but didn't see how to do this. 🤔

ranbena:
> arikfr merged commit 

I was about to reject this!

cypress/integration/query/search_query_spec.js
gabrieldutra: I think the way to do it is to wrap the result table and:
```javascript
cy.getByTestIdOrAth('QuerySearchResult').should(($resultTable) => {
  expect($resultTable).to.contain('Expected Result');
  expect($resultTable).not.to.contain('Anything else');
});
```

ranbena: Exactly.
@arikfr your test will return ANY element in the page that contains that text. So if this query showed up in the "favorites" dropdown, this would always be true...

arikfr: 👍 Thanks. 

I'll update the tests to do this form now.

arikfr:
@ranbena as I'm on schedule here, I wanted to see #3631 passing with this test and then to create #3634. Now they all green and in progress I can apply your reviews to this test.

PR #3636, opened 2019-03-24
Cypress eslint fixes and config

ranbena:
## What type of PR is this?

- [x] Feature

## Description
1. Added cypress folder to lint script.
   * updated `npm run lint` to run on `./client` instead of `./client/app`.
   * added `**/dist` to .eslintignore
   * apparently eslint never detected the ignore file cause it's not in project root 😵 so added `--ignore-path ./client/.eslintignore`. Now there's also no need for `lint:staged` specific folder ignoring.
   * removed `node_modules` from ignore file cause it's ignored by default.
2. Cypress specific eslint config
   * Created `eslintrc` in `cypress` folder (eslint cascading feature).
   * Removed cypress stuff from main `eslintrc`.
   * Added cypress lint plugin that enforces [cypress best practices](https://docs.cypress.io/guides/references/best-practices.html).
   * Allowing [chai assertions](https://github.com/cypress-io/eslint-plugin-cypress#chai-and-no-unused-expressions) like `expect(element).to.be(true)`.
   * Moved jest plugin to `./client/app` cause it collided with chai syntax.
3. Fixed lint errors and warnings
   * fixed some code
   * disabled lint where necessary
   * disabled `func-name` lint to allow anonymous functions `it('test', function () {`.

Tested on `npm run lint`, `lint-staged` and vscode.


ranbena:
Looks like codeclimate complies with the root `eslintrc` but ignores the rest 😩
Could it be that this line is revoking the cascading feature?:
https://github.com/getredash/redash/blob/25910e7655bb377e70efbcdce7270986d8d1747c/.codeclimate.yml#L27
Might have been said [here](https://github.com/codeclimate/codeclimate-eslint/issues/110#issuecomment-236442478).

@arikfr @gabrieldutra wdyt?

ranbena:
Perhaps we can do linting from a CircleCI step instead of CodeClimate?

arikfr:
The issue you linked to mentions:

> We do support cascading .eslintrc* files in effectively the same way that vanilla ESLint supports them. Does that clear it up a bit?

Maybe we need to remove the explicit eslint setting and let it find on its own?

ranbena:
lol it worked

ranbena:
I'm gonna test it by pushing a lint error.

ranbena:
No good. There should be 2 errors showing up, but CodeClimate is oblivious @arikfr

arikfr:
It seems to ignore some plugins.

ranbena:
I see 3 ways to go about this:
1. Understand and fix CodeClimate config.
2. Switch to CircleCi step for linting (not tested).
3. Merge all rules into the main eslintrc.

Gonna try #3 now.

ranbena:
@arikfr so I narrowed it down to a working config:
1. Back to one `./client/eslintrc.js`.
2. Can't use `plugin:cypress/recommended` - causes CodeClimate error. Helpful lints but not a must.

Good to go?

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -43,6 +43,7 @@ function addQueryByAPI(data, shouldPublish = true) {
     schedule: null,
   }, data);
 
+  // eslint-disable-next-line cypress/no-assigning-return-values
```
gabrieldutra: `cypress/recommended` seemed nice :confused:. As we can't use it, should those cypress `eslint-disable` be removed?

ranbena: Maybe it's cause the cypress package isn't a required installation.

arikfr: Might be 🤔

arikfr:
> Good to go?

While I liked the fact you split the configuration, the current version is better than what we had until now so good to go. 👍

PR #3638, opened 2019-03-25
Update docker compose with the stable Redash v7 tag

justinclift:

- [x] 1-liner trivial update

### DIFF
--- a/setup/docker-compose.yml
+++ b/setup/docker-compose.yml
@@ -1,6 +1,6 @@
 version: '2'
 x-redash-service: &redash-service
-  image: redash/redash:6.0.0.b8537
+  image: redash/redash:7.0.0.b18042
   depends_on:
     - postgres
     - redis


justinclift:
Hmmm, the hard coded PostgreSQL version in the file is a cause for concern too.  It's literally behind by 10 patch revisions, including security updates.

I'll submit an updated PR with that fixed shortly.

arikfr:
Thanks, @justinclift ! About Postgres: you're right. We just need to keep the major version.

justinclift:
No worries.  Just submitted a PR with the change.  In theory (!) it shouldn't have any negative effect, and in quick testing locally here with a newly built VM for Redash using it, it was ok.  But, lets see if the tests for the PR pass.

The Redis version (v3.0) may also be a concern.  Looking at the Docker Hub, the oldest release series listed is v4:

&nbsp; &nbsp; https://hub.docker.com/_/redis

I'll experiment with it a bit here, and see how badly things break with that. :wink:

justinclift:
Redis 5 seems to work as a drop-in replacement.  Just updated the semi-production box here with it, and everything is working fine.

arikfr:
> Redis 5 seems to work as a drop-in replacement. Just updated the semi-production box here with it, and everything is working fine.

My concern with using a newer version of Redis in our Docker Compose setup is that people might use newer features that are not supported in Redis 3, which is what most deployments have.

How about we update the setup docker compose but keep the dev one as is?

justinclift:
Makes sense to me. :smile:

PR #3639, opened 2019-03-25
Update the PostgreSQL version to latest available in the 9.5 series

justinclift:

- [x] Security update

In theory (!), this updated PG version should require no corresponding changes made to anything else.

Updating to a more modern PG release would have benefits ([parallel query](https://www.postgresql.org/docs/11/parallel-query.html) stands out), but may require some code/SQL changes to match.  So, keeping this simple for now. :smile:

### DIFF
--- a/setup/docker-compose.yml
+++ b/setup/docker-compose.yml
@@ -36,7 +36,7 @@ services:
     image: redis:3.0-alpine
     restart: always
   postgres:
-    image: postgres:9.5.6-alpine
+    image: postgres:9.5-alpine
     env_file: /opt/redash/env
     volumes:
       - /opt/redash/postgres-data:/var/lib/postgresql/data


setup/docker-compose.yml
```diff
@@ -36,7 +36,7 @@ services:
     image: redis:3.0-alpine
     restart: always
   postgres:
-    image: postgres:9.5.6-alpine
+    image: postgres:9.5.16-alpine
```
arikfr: Does Postgres publish `postgres:9.5` image that tracks latest patch version?

justinclift: Good point.  Yes it does.  1 sec, will update. :smile:

justinclift:
As a data point, Redash seems to work happily with the PostgreSQL `11-alpine` image as well.

Testing it here now (only lightly though).

arikfr:
Thanks and thanks for testing w/ Pg11!

PR #3648, opened 2019-03-28
Migrate Timer component to React; update TimeAgo component

kravets-levko:

- [x] Refactor
- [x] Feature
- [x] Bug Fix

## Description

- Migrate `<rd-timer>` (now `Timer`) component to React (+ use React hooks).
- Update `TimeAgo` component: rewrite using hooks, fix bug (component not destroyed - memory leak).


client/app/components/Timer.jsx
```diff
@@ -0,0 +1,40 @@
...
+import PropTypes from 'prop-types';
+import { react2angular } from 'react2angular';
+import { Moment } from '@/components/proptypes';
+
+function timeFrom(startTime) {
+  return moment.utc(moment.now() - startTime).format('HH:mm:ss');
+}
+
+export function Timer({ from }) {
+  const startTime = useMemo(() => moment(from).valueOf(), [from]);
```
ranbena: Using `useMemo` as a sort of `componentDidMount` alternative?

kravets-levko: @ranbena  no, it's rather `getDerivedStateFromProps` - I need to prepare start time using `from` prop, but don't want to do it each time when component is re-rendered.

PR #3655, opened 2019-03-29
REST API to create user without invitation

shinsuke-nara:

- [x] Feature

## Description

I want to create a user without user invitation.

This is useful for systems using Redash as a part of it.
User verification is done in the systems.
Users must think that it is bother to do verification twice in a system.

-- 

--

### DIFF
--- a/redash/handlers/api.py
+++ b/redash/handlers/api.py
@@ -47,7 +47,8 @@ from redash.handlers.settings import OrganizationSettings
 from redash.handlers.users import (UserDisableResource, UserInviteResource,
                                    UserListResource,
                                    UserRegenerateApiKeyResource,
-                                   UserResetPasswordResource, UserResource)
+                                   UserResetPasswordResource, UserResource,
+                                   UserCreateResource)
 from redash.handlers.visualizations import (VisualizationListResource,
                                             VisualizationResource)
 from redash.handlers.widgets import WidgetListResource, WidgetResource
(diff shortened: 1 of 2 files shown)


rauchy:
Thanks for contributing, @shinsuke-nara!

Could you elaborate on your use case that requires this?

Would POSTing to `/users` with [the no_invite flag](https://github.com/getredash/redash/blob/master/redash/handlers/users.py#L148) or using an SSO provider be sufficient?

shinsuke-nara:
@rauchy 

Thank you for the comment.
I have been considering your advice.

I want to create an enabled users with easy way.

For `no_invite` flag, even though I create a user with `no_invite` flag, setting password via invitation page is required to enable the user. We can set password for the user with a path of invitation page included in response of user creation but two step user creation is troublesome. Recover process for two step user creation is more complex than one step user creation. For example, we may need to store failure reason to DB to recover it.

Using SSO is good but it is too much for me now.

Please tell me if you know an easy way to create enabled user.

arikfr:
> For no_invite flag, even though I create a user with no_invite flag, setting password via invitation page is required to enable the user. 

You can also set the password using an API call.

> Using SSO is good but it is too much for me now.

There is the option of passing the user via a request header. I don't know how your setup looks like, but it might be an easy option to auto login your users.

This functionality is not going to be merged. If you still prefer to have a single API call, you're welcome to make a pull request which:

1. Supports passing the password when calling `POST /api/users`.
2. Now that a user can directly login, you need to mark the "invitation pending" flag when the user logins (if it's set to true).

This way you have a single step. Make sure to add tests to cover this new code.

Thanks!

Issue #1081, opened 2016-05-30
Public dashboard won't load if query doesn't have results


arikfr:
If one of the queries in a public dashboard doesn't have a reference to results the dashboard won't load.

qwalker:
any update on this bug? We are seeing this issue and expecting a fix too.

arikfr:
@qwalker #3659 addresses this issue.

PR #3659, opened 2019-03-31
Allow Parameters on Public Dashboards

rauchy:

- [x] Feature

## Description
This PR enables dashboards with safe parameters to be shared publicly.

Towards #3284, #1081.


client/app/pages/dashboards/ShareDashboardDialog.jsx
```diff
@@ -35,7 +35,7 @@ class ShareDashboardDialog extends React.Component {
     };
 
     this.apiUrl = replace(API_SHARE_URL, '{id}', dashboard.id);
-    this.disabled = this.props.hasQueryParams && !dashboard.publicAccessEnabled;
+    this.disabled = !(this.props.hasOnlySafeQueries || dashboard.publicAccessEnabled);
```
kravets-levko: If I understand this correctly, "Allow public access" switch should be enabled if:

- it is already On;
- it is Off and there are only safe parameters.

My opinion is that it's better to avoid using negated condition (rewrite it as `this.isEnabled = dashboard.publicAccessEnabled || this.props.hasOnlySafeQueries;`) and negate it where it's used - it will be much easier to understand it.

client/app/pages/dashboards/dashboard.js
```diff
@@ -448,14 +448,14 @@ function DashboardCtrl(
 
   this.openShareForm = () => {
     // check if any of the wigets have query parameters
```
kravets-levko: This comment should be removed or updated.

client/app/pages/dashboards/ShareDashboardDialog.jsx
```diff
@@ -104,10 +104,10 @@ class ShareDashboardDialog extends React.Component {
         footer={null}
       >
         <Form layout="horizontal">
-          {this.props.hasQueryParams && (
+          {!this.props.hasOnlySafeQueries && (
             <Form.Item>
               <Alert
-                message="Sharing is currently not supported for dashboards containing queries with parameters."
+                message="For your security, sharing is currently not supported for dashboards containing queries with text parameters. Consider changing the text parameters in your query to a different type."
```
kravets-levko: This message shouldn't be visible if dashboard does not have parameters at all.

client/app/pages/dashboards/ShareDashboardDialog.jsx
rauchy: Yeah, makes much more sense 👍

client/app/pages/dashboards/dashboard.js
rauchy: https://twitter.com/nzkoz/status/538892801941848064?s=20

client/app/pages/dashboards/ShareDashboardDialog.jsx
rauchy: If a dashboard doesn't have parameters at all - that means all its queries are safe, so `hasOnlySafeQueries` is true and `!hasOnlySafeQueries` is false, so it won't show.

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -225,7 +225,7 @@ describe('Dashboard', () => {
       });
     });
 
-    it.only('is available to unauthenticated users', function () {
+    it('is available to unauthenticated users', function () {
```
ranbena: lol I did that SO MANY TIMES!

redash/models/__init__.py
```diff
@@ -690,6 +690,20 @@ def parameters(self):
...
+        query = """SELECT api_keys.api_key
+                   FROM api_keys
+                   JOIN dashboards ON object_id = dashboards.id
+                   JOIN widgets ON dashboards.id = widgets.dashboard_id
+                   JOIN visualizations ON widgets.visualization_id = visualizations.id
+                   JOIN queries ON visualizations.query_id = queries.id
+                   WHERE object_type='dashboards'
+                     AND queries.api_key = :api_key"""
+
+        api_keys = db.session.execute(query, {'api_key': self.api_key}).fetchall()
```
arikfr: Why do you filter the queries with `api_key` and not `id`? 🤔 

If you use `id` you can even skip the join with the queries table and just look up by `visualizations.query_id = :id`.

arikfr: Also, when thinking of it it's strange that you ask a _query_ for API keys of _dashboards_.

It probably makes more sense to ask the dashboard for all the query ids it holds. But I can see how it might complicate the code, so we can skip this for now.

redash/serializers.py
```diff
@@ -23,20 +23,27 @@ def public_widget(widget):
         'created_at': widget.created_at
     }
 
-    if widget.visualization and widget.visualization.id:
-        query_data = models.QueryResult.query.get(widget.visualization.query_rel.latest_query_data_id).to_dict()
+    v = widget.visualization
+    if v and v.id:
+        if v.query_rel.latest_query_data_id:
```
arikfr: We no longer need to do this -- public dashboard should use the results API as the rest of the pages. 👌

redash/serializers.py
```diff
@@ -23,20 +23,27 @@ def public_widget(widget):
...
+            'updated_at': v.updated_at,
+            'created_at': v.created_at,
             'query': {
+                'id': v.query_rel.id,
                 'query': ' ',  # workaround, as otherwise the query data won't be loaded.
-                'name': widget.visualization.query_rel.name,
-                'description': widget.visualization.query_rel.description,
-                'options': {},
+                'name': v.query_rel.name,
+                'data_source_id': v.query_rel.data_source_id,
```
arikfr: Why this is needed?

redash/permissions.py
```diff
@@ -16,13 +16,17 @@
...
 
 
-def has_access_to_object(obj, user, need_view_only):
-    return (obj.api_key == user.id) and need_view_only
+def has_access_to_object(obj, api_key, need_view_only):
+    if obj.api_key == api_key:
+        return need_view_only
+    else:
+        # check if api_key belongs to a dashboard containing this query
+        return api_key in obj.dashboard_api_keys and need_view_only
```
arikfr: This needs tests.

redash/models/__init__.py
rauchy: > Why do you filter the queries with api_key and not id? 🤔

cfni

must be a refactoring leftover or something 🤦‍♂

rauchy: f8380301

redash/serializers.py
rauchy: 🎉 fc7323de

redash/serializers.py
rauchy: Because of [this](https://github.com/getredash/redash/blob/master/client/app/services/query.js#L521).

redash/permissions.py
rauchy: 4f742fb

ghoshsanjoy78:
Is this feature done?

rauchy:
> I'm pretty sure you need to adjust this as well:
> 
> https://github.com/getredash/redash/blob/97492d7aa0f0eded339adf960ffc08db60796807/redash/tasks/queries.py#L280-L289

I'm not a big fan of [grabbing the query ID from the metadata](https://github.com/getredash/redash/commit/e0dc262d89f5d98b4cc4a229abab5e317183ef24). The alternative is passing the query ID as an argument down the `enqueue_query` chain, and I don't like that either. Any other ideas @arikfr?

arikfr:
> I'm not a big fan of grabbing the query ID from the metadata. The alternative is passing the query ID as an argument down the enqueue_query chain, and I don't like that either. Any other ideas @arikfr?

Why do we need the query object?

client/app/pages/dashboards/dashboard.js
```diff
@@ -428,15 +428,14 @@ function DashboardCtrl(
   }
 
   this.openShareForm = () => {
-    // check if any of the wigets have query parameters
-    const hasQueryParams = _.some(
+    const hasOnlySafeQueries = _.every(
       this.dashboard.widgets,
-      w => !_.isEmpty(w.getQuery() && w.getQuery().getParametersDefs()),
+      w => w.getQuery() && w.getQuery().is_safe,
```
truebit: when no param in any query, this will return false, then `hasOnlySafeQueries` would return false.
Change to below would work:
`w => !w.getQuery() || w.getQuery().is_safe,`

rauchy: I believe `is_safe` should return `true` for queries without any param (at least this happens in the backend - I'll have to verify that this is the same in the frontend)

truebit: I hardcoded `is_safe` in `ParameterizedQuery` to `True`, still not working. After changing above mentioned code, it works now.
That's what I did in my redash instance.

truebit: Besides, `is_safe` in `ParameterizedQuery` could be simplified as follow:
```
return all(i['type'] != 'text' for i in self.schema)
```

truebit: > I believe `is_safe` should return `true` for queries without any param (at least this happens in the backend - I'll have to verify that this is the same in the frontend)

IMHO It's not about `is_safe`, it's about `getQuery()` may return null as the widget does not contain any param.

truebit:
Another suggestion, we need to add `refresh` button in public dashboards with params. Otherwise user needs to refresh the page to get the new results after reselected params.

rauchy:
> Another suggestion, we need to add `refresh` button in public dashboards with params. Otherwise user needs to reload the page to get the new results after reselected params.

@truebit this was recently implemented by @gabrieldutra, and following f41efb8 works in this branch too.

redash/models/__init__.py
```diff
@@ -651,6 +651,20 @@ def parameters(self):
     def parameterized(self):
         return ParameterizedQuery(self.query_text, self.parameters)
 
+    @property
+    def dashboard_api_keys(self):
+        query = """SELECT api_keys.api_key
```
```suggestion
        query = """SELECT distinct api_keys.api_key
```

redash/permissions.py
arikfr: We might use `has_access_to_object` for other objects, in which case calling `dashboard_api_keys` will fail. Maybe add some more ifs?

redash/serializers/__init__.py
```diff
@@ -25,21 +25,22 @@ def public_widget(widget):
...
+            'created_at': v.created_at,
             'query': {
+                'id': v.query_rel.id,
                 'query': ' ',  # workaround, as otherwise the query data won't be loaded.
-                'name': widget.visualization.query_rel.name,
-                'description': widget.visualization.query_rel.description,
-                'options': {},
-                'latest_query_data': query_data
+                'name': v.query_rel.name,
+                'data_source_id': v.query_rel.data_source_id,
```
arikfr: Is this really needed?

redash/serializers/__init__.py
```diff
@@ -25,21 +25,22 @@ def public_widget(widget):
...
-            'created_at': widget.visualization.created_at,
+            'type': v.type,
+            'name': v.name,
+            'description': v.description,
+            'options': json_loads(v.options),
+            'updated_at': v.updated_at,
+            'created_at': v.created_at,
             'query': {
+                'id': v.query_rel.id,
                 'query': ' ',  # workaround, as otherwise the query data won't be loaded.
```
arikfr: Still needed?

client/app/pages/dashboards/public-dashboard-page.js
```diff
@@ -25,40 +25,44 @@ const PublicDashboardPage = {
...
+        this.filtersOnChange = (allFilters) => {
+          this.filters = allFilters;
+          $scope.$applyAsync();
+        };
+
+        this.extractGlobalParameters();
+      });
+
+      if (refreshRate) {
+        $timeout(this.refreshDashboard, refreshRate * 1000.0);
```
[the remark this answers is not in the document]
truebit: I tried, This does not work.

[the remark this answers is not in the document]
rauchy: @arikfr @truebit could you elaborate on why you think this does not work? Because when I'm adding a `refresh=30` querystring parameter, I'm seeing public dashboards update with new results just fine.

redash/permissions.py
rauchy: Not sure if I see this as premature or just general defensive programming, but anyway, 4098cf2.

client/app/pages/dashboards/dashboard.js
rauchy: @truebit you're right, I didn't interpret your comment correctly. This is fixed in c00c720. Thanks @arikfr for clarifying.

client/app/pages/dashboards/public-dashboard-page.js
```diff
@@ -25,40 +25,44 @@ const PublicDashboardPage = {
...
+    this.extractGlobalParameters = () => {
+      this.globalParameters = this.dashboard.getParametersDefs();
+    };
 
-          this.filters = []; // TODO: implement (@/services/dashboard.js:collectDashboardFilters)
-          this.filtersOnChange = (allFilters) => {
-            this.filters = allFilters;
-            $scope.$applyAsync();
-          };
+    $scope.$on('dashboard.update-parameters', () => {
```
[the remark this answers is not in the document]
rauchy: f259fde

redash/serializers/__init__.py
rauchy: You've [asked this before](https://github.com/getredash/redash/pull/3659#discussion_r275154351). Are you implying that you've read my answer and think it _isn't_ needed?

redash/serializers/__init__.py
rauchy: Yes, otherwise dropdown won't work.

client/app/pages/dashboards/public-dashboard-page.js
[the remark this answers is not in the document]
arikfr: @rauchy the queries you were testing with were scheduled?

redash/serializers/__init__.py
arikfr: Why?

redash/serializers/__init__.py
arikfr: I had a feeling of dejavu, but somehow lost track of the previous time I asked about it. 🤦‍♂ 

Anyway, it feels wrong to expose this just because of how the code works. The only reason we make this check there is for cases when the query has no data source associated with it:

https://github.com/getredash/redash/blob/92391e3cbc3fedac3380a44c57c9d8addeeb3238/client/app/services/query.js#L520-L524

We can move this logic to the backend and return the appropriate error there in such cases, making this `if` unnecessary.

client/app/pages/dashboards/public-dashboard-page.js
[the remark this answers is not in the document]
rauchy: @arikfr no they weren't. Why? Are public dashboard refreshes not supposed to trigger new query execution?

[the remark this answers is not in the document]
arikfr: @rauchy `refreshDashboard` calls `loadDashboard`, which just loads the dashboard definition. How is it supposed to trigger a refresh this way?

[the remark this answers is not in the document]
rauchy: @arikfr this seems like a (desired?) side-effect of [widget.load()](https://github.com/getredash/redash/blob/master/client/app/pages/dashboards/public-dashboard-page.js#L37).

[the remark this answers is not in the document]
arikfr: The logic there is a bit hard to follow, but if I'm not mistaken it will bring the last result and not trigger new execution (which is what the user expects in case of a refresh).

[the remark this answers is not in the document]
rauchy: Yeah, you're right. Public dashboards don't refresh data automatically, but they do update with last results according to the `refresh` querystring parameter. I think the reason I thought they do is that I had the same dashboard open in a standard (non-shared) view, and it was triggering refreshes (with `maxAge=0`).

So regarding

> We probably need to switch to same refresh as the regular dashboard?

Should I add this to public dashboards? (with same behavior as regular dashboards)

image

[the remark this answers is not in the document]
arikfr: > Should I add this to public dashboards? (with same behavior as regular dashboards)

No, for now let's not add more features but focus on finishing this. So just need to make it work, but no UI is needed. Later we can introduce the UI.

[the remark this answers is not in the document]
arikfr: Although now I wonder if all the changes make it possible now to reuse the same dashboard page for public dashboards and reduce code duplication?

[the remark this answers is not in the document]
rauchy: > Although now I wonder if all the changes make it possible now to reuse the same dashboard page for public dashboards and reduce code duplication?

I feel that the regular dashboard page is still an order of magnitude more complex than the shared dashboard page, i.e. it has a lot more concerns - saving, sharing, archiving, layout etc), while the shared dashboard page is a simpler, read-only version.

Definitely something to think about for future refactoring, but I think we can safely live with this minimal duplication, as it feels less costly than a grand refactoring ATM.

> No, for now let's not add more features but focus on finishing this. So just need to make it work, but no UI is needed. Later we can introduce the UI.

So just allow data refreshes using the `refresh` querystring parameter?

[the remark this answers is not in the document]
arikfr: > So just allow data refreshes using the refresh querystring parameter?

Yes.

redash/serializers/__init__.py
rauchy: 17fa5e0

client/app/pages/dashboards/public-dashboard-page.js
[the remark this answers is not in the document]
rauchy: 500c155

redash/serializers/__init__.py
rauchy: Ohhh, you were referring to the empty query text workaround? I thought you were referring to the query id. Anyway, the workaround can totally be removed - done in 16be2f6

rauchy:
@arikfr

> Aside from the code comments, based on our experience with visualization embeds it's worth adding the following tests:
> * Shared dashboard without parameters rendering
> * Shared dashboard with safe parameters rendering
> * Shared dashboard with un-safe parameters rendering

I've added your suggested tests in 4ad3787. The third test is a bit tricky, because you aren't supposed to be able to share a dashboard with unsafe parameters, so the test creates a dashboard without parameters, shares it, adds an unsafe parameter, logs out and visits the share link.

Currently - the result is that the dashboard can be viewed by an unauthenticated user with the link, and the widget with safe (or without any) parameter works, while the widget with an unsafe parameter errors. This raises a question - what should be the experience for this case? Halt the entire page or just the individual unsafe queries?

arikfr:
> This raises a question - what should be the experience for this case? Halt the entire page or just the individual unsafe queries?

Current behavior sounds reasonable (as long as the error message is clear). We can consider hiding these widgets and showing only the safe ones. This also means that we can change the share dialog behavior to just warn about the fact that these widgets will not be shown. (no need to implement in this PR , though)

rauchy:
> as long as the error message is clear

Overridden the message in fd874ca, it is now:

> This query contains potentially unsafe parameters and cannot be executed on a publicly shared dashboard.

justinclift:
As a generalisation, that wording sounds fine. :smile:

Kind of wondering if it's feasible to be more specific with the `potentially unsafe parameters` wording fragment though.  Something to convery `"foo" is a text parameter, which we don't yet allow`.  But, not sure of better wording. :wink:

rauchy:
@justinclift I agree we should be more explanatory, but I’d focus the attention on the part which creates and shares the dashboard. Anyway this is an edge case where someone shares a safe dashboard and then makes it unsafe.

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -109,6 +109,154 @@ describe('Dashboard', () => {
...
+          `OpenShareForm
+          PublicAccessEnabled`);
+
+          cy.getByTestId('SecretAddress').invoke('val').then((secretAddress) => {
+            callback(secretAddress);
+          });
+        });
+      };
+
+      it.only('when there are no parameters', function () {
```
ranbena: @rauchy a mistake?

gabrieldutra: Shouldn't CircleCI frontend-lint be warning those things? :thinking:

gabrieldutra: Nah, [cypress-io/eslint-plugin-cypress](https://github.com/cypress-io/eslint-plugin-cypress) doesn't have that :/

ranbena: Yep. Thanks @gabrieldutra!

ranbena: @rauchy do you wanna enable the other tests?

rauchy: > a mistake?

Totally 🤦‍♂️

> Shouldn't CircleCI frontend-lint be warning those things? 

I remember we had this discussion a while ago :)

rauchy: See #3995

PR #3681, opened 2019-04-06
Fix: support date ranges for parameterized embeds

rauchy:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Parsing of date ranges was unhandled in #3495, and is now fixed here, though I guess we should really toss this piece of code and consolidate.

Reported by @HenroRitchie [here](https://github.com/getredash/redash/pull/3495#issuecomment-480208986), thanks!

### DIFF
--- a/client/app/services/query-string.js
+++ b/client/app/services/query-string.js
@@ -1,0 +1,11 @@
+import qs from 'qs';
+
+const parse = queryString => qs.parse(queryString, { allowDots: true, ignoreQueryPrefix: true });
+const onlyParameters = ([k]) => k.startsWith('p_');
+const removePrefix = ([k, v]) => [k.slice(2), v];
+const toObject = (obj, [k, v]) => Object.assign(obj, { [k]: v });
+
+export default () => Object.entries(parse(location.search))
+  .filter(onlyParameters)
+  .map(removePrefix)
+  .reduce(toObject, {});
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "pivottable": "^2.15.0",
     "plotly.js": "1.41.3",
     "prop-types": "^15.6.1",
+    "qs": "^6.7.0",
     "react": "^16.8.3",
     "react-ace": "^6.1.0",
     "react-dom": "^16.8.3",
--- a/client/app/components/queries/visualization-embed.js
+++ b/client/app/components/queries/visualization-embed.js
@@ -22,8 +22,12 @@ const VisualizationEmbed = {
   },
 };
 
-const queryStringAsObject = () => location.search.slice(1).split('&').map(p => p.split('='))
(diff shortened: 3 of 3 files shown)


client/app/services/query-string.js
```diff
@@ -0,0 +1,11 @@
+import qs from 'qs';
+
+const parse = queryString => qs.parse(queryString, { allowDots: true, ignoreQueryPrefix: true });
+const onlyParameters = ([k]) => k.startsWith('p_');
+const removePrefix = ([k, v]) => [k.slice(2), v];
+const toObject = (obj, [k, v]) => Object.assign(obj, { [k]: v });
+
+export default () => Object.entries(parse(location.search))
```
ranbena: Since this parsing function is really specific to redash query parameters, I would:
```js
export default parseLocationSearch = () => Object.entries(parse(location.search));

export parseRedashQueryParameters = () => parseLocationSearch()
  .filter(onlyParameters)
  .map(removePrefix)
  .reduce(toObject, {});
```

Wdyt?

rauchy: I think that most of the (non-lib) code in the project is Redash specific but we don't explicitly add "Redash" to function and class names, so I tend to disagree.

package.json
```diff
@@ -76,6 +76,7 @@
     "pivottable": "^2.15.0",
     "plotly.js": "1.41.3",
     "prop-types": "^15.6.1",
+    "qs": "^6.7.0",
```
arikfr: You had to commit an updated `package-lock.json` file as well.

client/app/services/query-string.js
```diff
@@ -0,0 +1,11 @@
+import qs from 'qs';
```
arikfr: The file name doesn't follow the convention of `filename` = `name of default export`.

client/app/services/query-string.js
arikfr: While I agree that prefixing stuff with "Redash" in our codebase doesn't make sense, @ranbena  has a point that this is not some generic query string parsing function but a specific one for parameters. It makes sense to name it accordingly, like "parametersFromQueryString" or something along this lines.

package.json
rauchy: Hmm... I wonder how that slipped. Should I do this in a new PR?

client/app/services/query-string.js
rauchy: I was unaware of that convention. This will be handled in #3687

rauchy:
@arikfr 

> like "parametersFromQueryString" or something along this lines

A query is a model in Redash, a query can have multiple parameters, a query-string is a part of the url, parameters can be represented inside the query-string, sometimes the query-string is represented in our code as `query`, which has parameters. THIS IS NUTS.

package.json
arikfr: Yes.

PR #3682, opened 2019-04-07
Cypress Dashboard Service

ranbena:
## What type of PR is this?

- [x] Feature

## Description
Setting up Cypress Dashboard Service, this is required in the config file.


ranbena:
Gonna use env var instead.

PR #3683, opened 2019-04-07
Cypress Dashboard Service

ranbena:
## What type of PR is this?

- [x] Feature

## Description
Setting up Cypress Dashboard Service, this is required in the config file.


.circleci/config.yml
```diff
@@ -70,6 +70,7 @@ jobs:
       COMPOSE_FILE: .circleci/docker-compose.cypress.yml
       COMPOSE_PROJECT_NAME: cypress
       PERCY_TOKEN_ENCODED: ZGRiY2ZmZDQ0OTdjMzM5ZWE0ZGQzNTZiOWNkMDRjOTk4Zjg0ZjMxMWRmMDZiM2RjOTYxNDZhOGExMjI4ZDE3MA==
+      CYPRESS_PROJECT_ID_ENCODED: dTZ1MnZk
```
ranbena: Decodes to "u6u2vd"

client/cypress/cypress.js
```diff
@@ -33,11 +33,17 @@ function stopServer() {
...
+    if (PERCY_TOKEN_ENCODED) {
+      process.env.PERCY_TOKEN = atob(`${PERCY_TOKEN_ENCODED}`);
+    }
+    if (CYPRESS_PROJECT_ID_ENCODED) {
+      process.env.CYPRESS_PROJECT_ID = atob(`${CYPRESS_PROJECT_ID_ENCODED}`);
+    }
   }
+
   execSync(
     'docker-compose run cypress ./node_modules/.bin/percy exec -- ./node_modules/.bin/cypress run --browser chrome',
```
gabrieldutra: Don't you have to add a `--record` here? ([ref](https://docs.cypress.io/guides/core-concepts/dashboard-service.html#Setup))

ranbena: Right.

@gabrieldutra I need to also have a record key env var - should I also encode it in circleci config? (cypress docs advise to hide it..)

gabrieldutra: I'm afraid this is the way. It seems Cypress offers unlimited public records, right? If so, I think that's not much of a deal.

ranbena:
@gabrieldutra I'm getting an error as if the env vars aren't set (though the console log says they're aok). Any idea what wrong?

gabrieldutra:
> Any idea what wrong?

I think you have to add them here in order to pass the vars to the container
https://github.com/getredash/redash/blob/47bf91e15012ada1e8ad2cb12b07d1f8ce0e858d/.circleci/docker-compose.cypress.yml#L36-L38

ranbena:
Works :)
https://dashboard.cypress.io/#/projects/924cka/runs/1/specs

ranbena:
Looks like under the free plan we got max 3 users.
I sent an organization invite to @arikfr and @gabrieldutra (sent to @getredash.com emails but perhaps should to github account emails?)

ranbena:
I tested a fail scenario.

Cypress outputs record url

Screen Shot 2019-04-08 at 9 47 07


Screenshot is being taken.

Screen Shot 2019-04-08 at 9 49 36

client/cypress/cypress.js
```diff
@@ -33,13 +33,27 @@ function stopServer() {
...
+      process.env.CYPRESS_PROJECT_ID = atob(`${CYPRESS_PROJECT_ID_ENCODED}`);
+    }
+    if (CYPRESS_RECORD_KEY_ENCODED) {
+      process.env.CYPRESS_RECORD_KEY = atob(`${CYPRESS_RECORD_KEY_ENCODED}`);
+    }
   }
+
   execSync(
-    'docker-compose run cypress ./node_modules/.bin/percy exec -- ./node_modules/.bin/cypress run --browser chrome',
+    'docker-compose run cypress ./node_modules/.bin/percy exec -- ./node_modules/.bin/cypress run --record --browser electron',
```
gabrieldutra: Electron is the default one, should we
```suggestion
    'docker-compose run cypress ./node_modules/.bin/percy exec -- ./node_modules/.bin/cypress run --record',
```
?

I see this is required for the recorded videos, previously I used chrome because one of the tests didn't run with Electron (it seems the issue is gone? if so, NP with this change)

ranbena:
Added video capture

Screen Shot 2019-04-08 at 15 55 41
Screen Shot 2019-04-08 at 15 58 34

ranbena:
> I wonder if we can change github Details link to Cypress or add a link to it somewhere else 🤔

You mean that the cypress dashboard link would be appear in the failed build details (and not only in the terminal output)? I searched for it but couldn't find anything..

ranbena:
@gabrieldutra skipped the test and it looks like it's ready to go

gabrieldutra:
> skipped the test and it looks like it's ready to go

Sound good. One last thing, local runs are still with Chrome, right? Do you think we get any value from it or should we determine Electron as a standard?

ranbena:
> Do you think we get any value from it or should we determine Electron as a standard?

Totally right - Electron should be the standard.

PR #3687, opened 2019-04-08
Encapsulate query strings

rauchy:

- [x] Refactor
- [x] Feature


client/app/services/query.js
```diff
@@ -285,11 +285,7 @@ class Parameters {
   }
 
   toUrlParams() {
-    if (this.get().length === 0) {
-      return '';
-    }
-
-    return qs.toString(Object.assign(...this.get().map(p => p.toUrlParams())));
+    return qs.toString(this.getValues());
```
kravets-levko: I think it's better to use `param.toUrlParams()` instead of mapping `name => value` directly - it provides some layer of abstraction for a case when param will need some different serialization logic in future.

rauchy: What I'm trying to achieve with this refactoring is moving out the serialization logic from Parameter/Parameters and into the query string module. The approach here is that a query string is itself the serialization "view", thus the logic resides there and the model simply reveals its data.

arikfr:
This is no longer needed, right?

rauchy:
@arikfr it was never _needed_, but I guess at this point it'll be better to revisit this at `Datetime.max`.

PR #3689, opened 2019-04-09
Percy: Introduce hide-in-percy and hide diff problematic elements

gabrieldutra:

- [x] Other

## Description
This introduces the `.hide-in-percy` class to hide elements in the Page on Percy Screenshots and hide the following elements:
- User Edit Page: API Key
- All Pages: Pace progress bar

Check Percy diff :slightly_smiling_face:


.circleci/config.yml
```diff
@@ -84,6 +84,8 @@ jobs:
           command: |
             npm run cypress start
             docker-compose run cypress npm run cypress db-seed
+            # Make sure the API key is the same so Percy snapshots are consistent
+            docker-compose -p cypress run postgres psql -U postgres -h postgres -c "update users set api_key = 'secret' where email ='admin@redash.io';"
```
ranbena: Can you instead make this be able to be set from db seeding?
```js
{
    route: '/setup',
    type: 'form',
    data: {
      name: 'Example Admin',
      email: 'admin@redash.io',
      password: 'password',
      org_name: 'Redash',
      api_key: 'secret', // <- HERE
    },
  },
```

client/cypress/integration/user/edit_profile_spec.js
```diff
@@ -36,14 +41,6 @@ describe('Edit Profile', () => {
     });
```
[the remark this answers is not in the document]
gabrieldutra: IMO setting the state after the test is not a good approach as well. My ideas to improve it are:
- try to intercept API response and modify "api_key" from it (it was not as easy as I thought to do it ^^);
- modify the `edit_profile_spec` to do something I should've done at first, but leaned later: create a user to be tested prior the tests instead of asserting on the Admin user (this should fix this one).

The goal here is to get some time to do the above without annoyance in Percy. LMK what you think :slightly_smiling_face:

.circleci/config.yml
gabrieldutra: I don't like very much the idea of changing the API to this, in any case see my ideas in the other comment and LMK what you think.

client/cypress/integration/user/edit_profile_spec.js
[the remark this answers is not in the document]
ranbena: The best approach IMO is to actually "blackout" the element so it doesn't get included in screenshots.
```css
@media only percy {
  .hide-in-percy {
    visibility: hidden;
  }
}
```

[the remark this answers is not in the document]
ranbena: @gabrieldutra if you go this route, can you piggyback [pace](https://github.com/getredash/pace) as well (it's causing Percy false negatives)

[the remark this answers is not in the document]
gabrieldutra: Not sure this is the best solution to those cases as we are not capturing changes to that part of the UI. However this seems the best for now and I wanted to introduce Percy specific css anyway :slightly_smiling_face:.

@ranbena please leave your comments, when you think this is ok I'll rename the PR and merge it.

client/app/assets/less/inc/misc.less
```diff
@@ -225,4 +225,13 @@
...
     border-radius: 2px;
     width: 37px;
+}
+
+/* --------------------------------------------------------
+    Percy
+-----------------------------------------------------------*/
+@media only percy {
+    .hide-in-percy, .pace {
+        display: none;
```
gabrieldutra: `display: none;` seemed better, but tell me if you think we get some value on having a white space

gabrieldutra:
Next step is certainly discover a fix to those square icons. They are even more annoying than Pace :laughing:

client/app/assets/less/inc/misc.less
ranbena: There's no difference to Percy comparison.
But visibility keeps the layout unaltered while display might mess things up, which might confuse a human looking at the images. So I favor `visibility: hidden`.

You can even go as far as this, to have full clarity 😬:
```css
@media only percy {
  .hide-in-percy, .pace {
    visibility: hidden;
    position: relative;

    &:after {
      content: 'Hidden from Percy';
      font-size: 10px;
      color: #000;
      opacity: 0.5;
      visibility: visible;
      position: absolute;
      left: 2px;
      top: 2px;
    }
  }
}
```

client/app/components/users/UserEdit.jsx
```diff
@@ -176,7 +176,7 @@ export default class UserEdit extends React.Component {
     const { user, regeneratingApiKey } = this.state;
 
     return (
-      <Form layout="vertical">
+      <Form className="hide-in-percy" layout="vertical">
```
ranbena: I think it's important to hide only the precise offending content `<InputWithCopy>`.

client/app/assets/less/inc/misc.less
gabrieldutra: > You can even go as far as this, to have full clarity grimacing (only if you want to)

Nah, this may create issues when we want to hide small things.

PR #3656, opened 2019-03-29
Dashboard grid markings

ranbena:
## What type of PR is this?
- [x] Feature

## Description
Added grid tile markings to dashboard edit mode.
Discussion in https://discuss.redash.io/t/dashboard-grid-markings/3515.

* Auto adjusts to fit 6 cols in viewport width
* Turns off when in 1 column mode

### Reviewer 👋
1. Commit 1 - Added the grid bg.
2. Commit 2 - Made the grid container go all the way to the bottom (or else it would be only as high as the widgets 👎)

## Mobile & Desktop Screenshots

Screen Shot 2019-04-10 at 14 26 33

PR #3691, opened 2019-04-09
Migrated Sticky footer technique to flexbox

ranbena:
## What type of PR is this? 

- [x] Refactor

## Description
Redash's sticky footer implementation is problematic for https://github.com/getredash/redash/pull/3656.
Since currently `body` has `padding-bottom` to the height of the footer, using `height: 100vh` in the app will create a needless vertical scroll.

This is a more sensible flexbox implementation .

There should be no regression, nor visual change. Sticky footer still works.

### DIFF
--- a/client/app/assets/less/inc/base.less
+++ b/client/app/assets/less/inc/base.less
@@ -14,17 +14,11 @@ html {
     -ms-overflow-style: auto;
 }
 
-html, body {
-    min-height: 100vh;
-}
-
 body {
     padding-top: @header-height;
     position: relative;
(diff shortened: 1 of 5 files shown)


arikfr:
Remember we discussed removing the footer? I guess it's the right time :)

ranbena:
> Remember we discussed removing the footer? I guess it's the right time :)

https://discuss.redash.io/t/footer-advocacy/3599

ranbena:
Abandoned due to #3697

Issue #3570, opened 2019-03-12
ElasticSearch: schema refresh failed


LuisDeHaro:
### Issue Summary

When I try to create a new query using a ES data source I get an error: schema refresh failed.

### Steps to Reproduce

1. Created a ES datasource successfully 
2. When creating a new query and selecting the ES data source a message appears: Schema refresh failed. Please try again later.

### Technical details:

* Redash Version: 6.0.0+b8537 (4780bd9c)
* Browser/OS: MacOS 10.14.1, Chrome 72.0.3626.121
* How did you install Redash: Docker
* ElasticSearch version: 5.6.4

rauchy:
Thanks for reporting, @LuisDeHaro.

Given the details you provided, it works fine for me and I can't seem to reproduce the problem. Since this is probably not a general issue with the project, could you please submit this question in [our forums](https://discuss.redash.io/)? Make sure you include server logs that show this error. Also, indicate if you are able to execute queries, regardless of the schema refresh failure.

truebit:
I use Redash 7.0.0 and Elasticsearch 5.6.3
Here below is the debug log from redash:
the elasticsearch _mappings request got success response. but redash UI rejected the response for some reason.
> 
> [2019-04-09 22:07:07,466][PID:724][INFO][sqlalchemy.engine.base.Engine] BEGIN (implicit)
> [2019-04-09 22:07:07,467][PID:724][INFO][sqlalchemy.engine.base.Engine] SELECT organizations.updated_at AS organizations_updated_at, organizations.created_at AS organizations_created_at, organizations.id AS organizations_id, organizations.name AS organizations_name, organizations.slug AS organizations_slug, organizations.settings AS organizations_settings
> FROM organizations
> WHERE organizations.slug = %(slug_1)s
>  LIMIT %(param_1)s
> [2019-04-09 22:07:07,467][PID:724][INFO][sqlalchemy.engine.base.Engine] {'slug_1': 'default', 'param_1': 1}
> [2019-04-09 22:07:07,468][PID:724][DEBUG][metrics] table=organizations query=select duration=2.07
> [2019-04-09 22:07:07,469][PID:724][DEBUG][root] Current organization: TESTORG (1) (slug: default)
[...]

PR #3692, opened 2019-04-10
fix elasticsearch get_schema error when mapping contains no 'properties' key

truebit:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
fix elasticsearch get_schema error when index has no 'properties' key.
This bug could be reproduced by trying to use a elasticsearch instance with `.kibana` index in it.

The version of Elasticsearch and Kibana I use is `5.6.3`. Sample `.kibana` mappings or any other index which contains [`_default_`](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/default-mapping.html) in mappings.

[kibana_mappings.txt](https://github.com/getredash/redash/files/3061882/kibana_mappings.txt)

I also added exception traceback print for future debugging purposes.

#3570 
no changes


truebit:
This fix just fixes `_default_` mappings in elasticsearch. for `dynamic_templates` in `_default_`, now I cannot find a good way to achieve that.

truebit:
`_default_` mappings is deprecated by elasticsearch from version 6 and above. So this bug only occurs on v5 or earlier:

https://www.elastic.co/guide/en/elasticsearch/reference/current/default-mapping.html

lomkju:
Thanks this helped us.

truebit:
@axelsean That's odd. Was that because `doc` variable is not a dict but a list in your scenario?

axelsean:
I *think* so, but I'm a real Python novice (written 3 lines in my life !)

Dave

truebit:
Could you debug it by printing`doc` content to see what it is?

redash/query_runner/elasticsearch.py
```diff
@@ -148,7 +148,11 @@ def parse_doc(doc, path=None):
             '''
             path = path or []
             result = []
-            for field, description in doc['properties'].items():
+            # index mappings contains "_default_" do not have "properties"
+            properties = doc.get('properties')
+            if not properties:
+                return result
+            for field, description in properties.items():
```
aidarbek: Hi! What about making it a bit shorter and just writing this? 

`for field, description in doc.get('properties', {}).items():`

Thus, there will be no need for previous new lines

truebit: Making python code short is a style thing, I am OK with that.
Besides, as latter comments pointed out, the `doc` may not be a dict. So maybe we still need code below to avoid the "not dict" issue.
```
if not isinstance(doc, dict):
    return result
```

truebit: Or
```           
if not isinstance(doc, dict) or 'properties' not in doc:
    return result
```
and keep original code intact.

arikfr:

PR #3695, opened 2019-04-10
Add package-lock.json after including qs

rauchy:

- [x] Bug Fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2485,7 +2485,7 @@
     },
     "bl": {
       "version": "1.2.2",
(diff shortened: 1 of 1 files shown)


kravets-levko:
@rauchy I see that `npm` installed `qs` of the same version few times. Usually it happens when you add dependency to the project, but one of existing dependencies already have it. Try to remove `node_modules`, run `npm install` and see if it helps to remove duplicates.

rauchy:
@kravets-levko that's interesting. I tried doing it and it had no effect though :(

Issue #3700, opened 2019-04-11
Add Time Limit to Scheduled Queries


arikfr:
We already have a global time limit for adhoc queries (disabled by default). Now we want to have another time limit for scheduled queries, while also properly communicating the user what happened (rather than showing the raw exception message).

There are two things to implement here:

1\. 
Convert the TimeLimit exceptions (there are two: hard and soft, although I think we only use soft) into user friendly messages. Basically catch the exception and raise a new one with a friendly message.

2\. 
Adding a TimeLimit to scheduled queries. But here’s the tricky part: it needs to be variable based on the org’s plan. Alas there is no plan concept in the OSS Redash. To this end I thought of having Policy object, similar to the one we have in the frontend code. Right now it will have a single method to decide on the time limit, but later we can add more methods to it. This method will take a user object and whether it's scheduled as parameters and return the time limit value.

The default implementation will just read from configuration. But we need a way to extend/replace implementation in the SaaS codebase. One thought I had for this is to have the name of the object defined in `redash.settings` and dynamically loaded during startup. But open to other ideas as well.

rauchy:
Seems like #1 is already implemented :)

![image](https://user-images.githubusercontent.com/289488/55950657-16295700-5c5e-11e9-9acd-352de4033860.png)

PR #3702, opened 2019-04-11
Dynamic query time limits

rauchy:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
This PR allows users to define their own custom query time limit decisions by supplying their own function to determine the limit for a adhoc/scheduled query run by a specific user.

#3700


redash/settings/dynamic_settings.py
```diff
@@ -0,0 +1,6 @@
+import os
+from .helpers import int_or_none
+
+# Replace this method with your own implementation in case you want to limit the time limit on certain queries or users.
+def get_query_time_limit(is_scheduled, user_id):
+    return None if is_scheduled else int_or_none(os.environ.get('REDASH_ADHOC_QUERY_TIME_LIMIT', None))
```
arikfr: Let's take this opportunity to introduce a value for scheduled queries as well.

redash/settings/dynamic_settings.py
```diff
@@ -0,0 +1,6 @@
+import os
+from .helpers import int_or_none
+
+# Replace this method with your own implementation in case you want to limit the time limit on certain queries or users.
+def get_query_time_limit(is_scheduled, user_id):
```
arikfr: I thought we have a `User` object at hand when calling this. Considering we don't, let's pass in `org_id` as well (or only `org_id`?).

redash/settings/dynamic_settings.py
rauchy: 👍 2885e346

redash/settings/dynamic_settings.py
rauchy: If the purpose of extracting this to a dynamic class was to allow users to customize this according to their needs, I'd say a `user_id` would allow more cases to be handled. I say we provide both.

PR #3706, opened 2019-04-12
Fix search ordered by best match

Jakdaw:

- [X] Bug Fix

## Description

Make searches order by best-match (unless something else is explicitly requested). 

See discussion at:

https://discuss.redash.io/t/searching-for-queries-after-upgrade-to-v5/2590/6


jezdez:
I'm not really following this code anymore (and I wrote it), so maybe it would make sense to refactor the `order_results` function into something that is easier to understand?

PR #3707, opened 2019-04-13
Fix query based parameter has value null when created

gabrieldutra:

- [x] Bug Fix

## Description
Whenever a new Query Based Parameter was created it started with a `null` value.

For this I noticed that the prop value starts with `null` and is updated after some time, the problem is that Antd's `defaultValue` only sets the value on the first time.

I also took the opportunity to replace `componentWillReceiveProps` to `componentDidUpdate` as the first one is deprecated.

--

**Before**
query-based-parameter-null

**After**
query-based-parameter-fixed


arikfr:
👍

Issue #3038, opened 2018-11-01
Replace Makefile with a simple script


arikfr:
How about we replace the new Makefile with a simple script? 

In the hosted version repo I started sometime ago to experiment with a helper script `bin/dev`:

```bash
logs() {
    docker-compose logs --tail=30 -f $1
}

case "$1" in
  init)
    docker-compose up -d
    docker-compose run --rm server create_db
    npm i 
    npm run build
    ;;
  logs)
    shift
    logs $*
    ;;
  reload)
    shift
    docker-compose up -d --no-deps --force-recreate $1
    ;;
  restart)
    shift
    docker-compose restart $1
    ;;
  ps)
    docker-compose ps
    ;;
  psql)
    docker-compose run --rm postgres psql -h postgres -U postgres
    ;;
  migrate)
    docker-compose run --rm server migrate
    ;;
  test)
    shift
    docker-compose run --rm --no-deps server tests $@
    ;;
  *)
    echo "I have no idea what to do with that."
    ;;
esac
```

I wonder if we should adopt something similar instead of the Makefile?

The main reasons I think it makes sense to drop the Makefile:

* It's not really a build process, but rather a collection of useful shortcuts.
* Commands like the `bin/dev logs` command take parameters, which iirc are not as elegant with make.

WDYT?

jezdez:
The advantages of Make (and its Makefile format) in my experience are:

- simplistic syntax by design, less prone to become a mess of shell scripts
- auto-complete by default on most systems
- task dependencies, composability of tasks
- available on all major platforms by default

I guess I admit I'm not great at writing shell scripts, and even less so writing bash scripts and have used other tools like Fabric, or even simple one-off Python scripts, in the past to work around that.

aidarbek:
Hi! I've added PR that could solve this issue. I know, long time passed since issue creation, so let me know if it is still relevant :) Or maybe some additional discussion is needed on pros/cons of Makefile vs Bash?

arikfr:
Yes, this is still relevant. Thanks, @aidarbek ! I'll continue conversation in your PR.

justinclift:
Closing this, as (at least for now) the Makefile is serving us decently well. :smile:

PR #3708, opened 2019-04-14
Bash file with the same as functionality Makefile was added

aidarbek:

- [ ] Refactor

## Description
Replacing Makefile with bash script
Solves issue in #3038


guidopetri:
@aidarbek , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
Interesting idea.  We may revisit the concept at some point, but for now the Makefile is simpler and pretty easy to work with.

PR #3709, opened 2019-04-15
Add rate limits to user creation/update

arikfr:

- [x] Feature

## Description

This PR adds rate limits to `POST /api/users` and `POST /api/users/:user_id`. It's currently static, but we can make it configurable if needed.

Also changed tests not to use rate limits (except for the tests that need them).

### DIFF
--- a/tests/handlers/test_authentication.py
+++ b/tests/handlers/test_authentication.py
@@ -3,7 +3,7 @@ import time
 import mock
 from tests import BaseTestCase
 
-from redash import settings
+from redash import settings, limiter
 from redash.authentication.account import invite_token
 from redash.models import User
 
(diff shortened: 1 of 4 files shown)


tests/models/test_queries.py
```diff
@@ -21,13 +21,13 @@ def create_tagged_query(self, tags):
         return query
 
     def test_all_tags(self):
-        self.create_tagged_query(tags=['tag1'])
```
arikfr: All the changes here are to convert strings to unicode to avoid SQLA warnings (unrelated to the PR).

tests/handlers/test_authentication.py
```diff
@@ -82,6 +83,7 @@ def test_valid_password(self):
 
 class TestLogin(BaseTestCase):
     def test_throttle_login(self):
+        limiter.enabled = True
         # Extract the limit from settings (ex: '50/day')
         limit = settings.THROTTLE_LOGIN_PATTERN.split('/')[0]
         for _ in range(0, int(limit)):
```
rauchy:
```ruby
(limit + 1).times { do_something }
```
I miss Ruby 😢

PR #3712, opened 2019-04-16
add get_by_id to Organization

rauchy:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

### DIFF
--- a/redash/models/organizations.py
+++ b/redash/models/organizations.py
@@ -32,6 +32,10 @@ class Organization(TimestampMixin, db.Model):
     def get_by_slug(cls, slug):
         return cls.query.filter(cls.slug == slug).first()
 
+    @classmethod
+    def get_by_id(cls, _id):
+        return cls.query.filter(cls.id == _id).one()
+
     @property
     def default_group(self):
         return self.groups.filter(Group.name == 'default', Group.type == Group.BUILTIN_GROUP).first()


redash/models/organizations.py
```diff
@@ -32,6 +32,10 @@ def __str__(self):
     def get_by_slug(cls, slug):
         return cls.query.filter(cls.slug == slug).first()
 
+    @classmethod
+    def get_by_id(cls, _id):
+        return cls.query.filter(cls.id == _id).one()
+
```
jezdez: What if we add a custom base model class that has this method (and maybe be other duplicates)?

http://flask-sqlalchemy.pocoo.org/2.3/customizing/#model-class

```python
# in redash/models/base.py
from flask_sqlalchemy import Model

class RedashModel(Model):
     @classmethod
    def get_by_id(cls, _id):
        return cls.query.filter(cls.id == _id).one()

    def save(self, commit=True):
        postgres.session.add(self)
        if commit:
            postgres.session.commit()
        return self

    def delete(self, commit=True):
        postgres.session.delete(self)
        if commit:
            postgres.session.commit()
        return self

# ...

db = RedashSQLAlchemy(
    model_class=RedashModel,
    session_options={
        'expire_on_commit': False
    },
)
```

arikfr: That's a good idea, but a bit out of scope for this PR. Let's keep it in mind for future refactors.

jezdez: Work for me 😀

PR #3653, opened 2019-03-28
Dashboard auto-saving

ranbena:
## What type of PR is this?

- [x] Feature

## Description
UX discussion https://discuss.redash.io/t/auto-saving-dashboard-layout/3506
Demo https://deploy-preview-3653--redash-preview.netlify.com/dashboard/artists

### Motivation
Currently when in dashboard edit mode, there are “Apply Changes” and “Cancel” buttons which are inconsistent (only apply to resize and drag - not to added / deleted widgets) and buggy (there is a scenario causing widget overlap).

### The Change
Ditch cancellability and embrace auto-save, ala-Google Docs.
One button to switch off edit mode, next to it a save status indicator.

### How it works
Each layout change marks a dirty state, showing a "Saving..." indication (although technically not yet saving) and triggers a server save to each changed widgets, debounced by 2 seconds.

Any error saving yields the appropriate notification and dirty indication persists.

### Updated tests

Screen Shot 2019-04-12 at 11 00 18

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
ezgif com-video-to-gif

PR #3715, opened 2019-04-17
Dashboard save fail indication

ranbena:
## What type of PR is this?

- [x] Feature

## Description
Since dashboard auto-save #3653, in case of a failed save, there is a notification but it still says "Saved" at the top as @gabrieldutra [rightfully indicated](https://github.com/getredash/redash/pull/3653#discussion_r275966837).

This PR adds another boolean flag `saveDelay` to differentiate 3 states: Saved, Saving, Error.

In this recording I block my server url (with developer tools), get error and then turn back on, to demonstrate the flow.

### DIFF
--- a/client/app/assets/less/redash/redash-newstyle.less
+++ b/client/app/assets/less/redash/redash-newstyle.less
@@ -938,7 +938,3 @@ text.slicetext {
 .markdown strong {
   font-weight: bold;
 }
-
-.disabled-silent {
-  pointer-events: none;
-}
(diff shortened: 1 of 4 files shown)


ranbena:
I added an additional feature to the "done editing" button:
When dashboard is in saving state and users clicks the button - it turns into a loading state till save completes.

ranbena:
> Shouldn't it auto close edit mode when save completes with success?

Oh mannnnnn.. I've been working on the dashboard too long 🙃😖Fixed.
ezgif com-video-to-gif (2)

PR #3716, opened 2019-04-17
Fix Ability to Add Query-based Parameters to Existing Queries

### LINKED ISSUE #3711: Cannot Add Query-based Dropdown Parameter to a Saved Query
### Issue Summary

On a new query, query-based dropdown parameters get their values from `/api/queries/:id/dropdown` and on saved queries they get them by association from their parent query id (`/api/queries/:parent_id/dropdowns/:id`).

When adding a query-based dropdown parameter to a saved query, the association is not yet created which results in a 403.

### Steps to Reproduce

1. Create a new query without a parameter
2. Save it
3. Add a query-based parameter to the query.
4. The dropdown is not populated with values (a 403 occurs)

rauchy:

- [x] Bug Fix

## Description
In order to solve #3711, dropdown values would be fetched directly from the `/api/:id/dropdown` endpoint while a query is edited. Once it is saved, an association is created and dropdown values could be fetched from `/api/:parent_id/dropdowns/:id`.

If a user has edit privileges to the parent query but cannot access the dropdown query directly, he will be prompted with this message:

image

Fixes #3711


client/app/components/QueryBasedParameterInput.jsx
```diff
@@ -42,10 +45,23 @@ export class QueryBasedParameterInput extends React.Component {
     }
   }
 
+  async _fetchOptions() {
+    try {
+      return await this.props.parameter.loadDropdownValues(this.props.isDirty);
+    } catch {
+      if (this.props.isDirty) {
+        notification.error('Cannot Fetch Dropdown Parameter Values', 'This query has other queries associated with it as dropdown parameters. ' +
+        'In order to edit this query, you must have access to the associated queries.', { duration: 10 });
```
kravets-levko:
```js
notification.error(
  'Cannot Fetch Dropdown Parameter Values', 
  `
    This query has other queries associated with it as dropdown parameters. 
    In order to edit this query, you must have access to the associated queries.
  `, 
  { duration: 10 },
);
```

rauchy: Wouldn't that add a line break to the formatted text?

redash/handlers/query_results.py
```diff
@@ -108,18 +108,23 @@ def post(self):
...
     def get(self, query_id, dropdown_query_id):
         query = get_object_or_404(models.Query.get_by_id_and_org, query_id, self.current_org)
+        require_access(query.data_source, current_user, view_only)
 
         related_queries_ids = [p['queryId'] for p in query.parameters if p['type'] == 'query']
         if int(dropdown_query_id) not in related_queries_ids:
-            abort(403)
+            dropdown_query = get_object_or_404(models.Query.get_by_id_and_org, dropdown_query_id, self.current_org)
+            if not has_access(dropdown_query.data_source, current_user, view_only):
+              abort(403)
```
arikfr: Check the CodeClimate error -- it's actually relevant.

client/app/components/QueryBasedParameterInput.jsx
kravets-levko: It's html - browser will replace newline with whitespace, and then collapse all whitespaces.

redash/handlers/query_results.py
rauchy: I'm surprised Flask didn't crash on this 😝

rauchy:
> No frontend changes were needed? It will always use the API with parent query once the query is saved?

Just like it used to be, but only now it will fallback on the backend and still allow dropdown query results, even when they are unassociated (assuming user has direct access to the dropdown query).

redash/handlers/query_results.py
```diff
@@ -108,18 +108,23 @@ def post(self):
...
 class QueryDropdownsResource(BaseResource):
     def get(self, query_id, dropdown_query_id):
         query = get_object_or_404(models.Query.get_by_id_and_org, query_id, self.current_org)
+        require_access(query, current_user, view_only)
 
         related_queries_ids = [p['queryId'] for p in query.parameters if p['type'] == 'query']
         if int(dropdown_query_id) not in related_queries_ids:
-            abort(403)
+            dropdown_query = get_object_or_404(models.Query.get_by_id_and_org, dropdown_query_id, self.current_org)
+            if not has_access(dropdown_query.data_source, current_user, view_only):
```
arikfr: Can't we use `require_access` here as well?

rauchy: Totally. 13e398b

PR #3720, opened 2019-04-18
Feature/oryan/dark theme and pie values

oryanmoshe:

- [ ] Feature

## Description
Added support for dark theme in public dashboards (by adding /dark to the route)
Added support for displaying values in pie chart segments instead of just percents

image

image


ranbena:
Having a theme feature is really great. But:
1. Ideally it'd be controlled by a single `theme/dark.less` file which overrides color variables.
(might need to make sure all colors in the project are taken from variables for that to work)
I think Plotly's API only allows color settings in js, but that too can be done more conventionally.
2. If you're introducing themes, make it ready for multiple themes - not just dark/light.

arikfr:
@oryanmoshe 👋

arikfr:
It seems to be abandoned so I'm closing this, but you're welcome to reopen after addressing the comments.

PR #3722, opened 2019-04-18
Dashboard grid React migration #1

### LINKED ISSUE #3202: remove widget in dashboard works only once
### Issue Summary

When removing multiple widgets in a dashboard, I can only perform the action once. The second time I click on the 'cross' of a widget the 'are you sure' dialog doesn't appear and the widget isn't removed. When I refresh the page, I can remove one widget again. 

### Steps to Reproduce

1. Create a dashboard
2. add two widgets 
3. remove the first widget
4. remove the second widget, this doesn't work

I expect that I can remove multiple widgets, without refreshing the page in between. 

### Technical details:
* Redash Version: 6.0.0+b8537
* Browser/OS: chromium 66.3359.139 Ubuntu
* How did you install Redash: Ami docker version

### LINKED ISSUE #3285: Migrate the dashboard grid to React
https://github.com/STRML/react-grid-layout might be relevant here.

Part of the React migration (#3071).

Blocked by #3301.

### LINKED ISSUE #3721: Widgets obscured by "Add Widget" bar
### Issue Summary

Screen Shot 2019-04-18 at 12 51 00

### Steps to Reproduce

1. Enter editing mode in dashboard
2. Make sure there are enough widgets to fill your browser viewport.
3. Scroll to bottom.

Expected: Bottom widget unobscured and reachable.
Actual: "Add Widget" bar obscures bottom widgets.

### Technical details:

* Redash Version: 7
* Browser/OS: Chrome Mac

ranbena:
- [x] Refactor

## Description
Since `gridstack` has no React implementation, migrated to [`react-grid-layout`](https://github.com/STRML/react-grid-layout).

Closes #3285.

**Included**
- [x] Draggable/Resizable widgets
- [x] Added / Removed widgets
- [x] [Dashboard config](https://github.com/getredash/redash/blob/master/client/app/config/dashboard-grid-options.js) compliance
- [x] Public dashboard
- [x] Layout auto-save
- [x] Edit/Preview modes
- [x] Tests passing (wip #3726)
- [x] Auto Save/Fail/Retry
- [x] Grid markings

**Not included (tests skipped)**
1. Single column breakpoint (upcoming PR)
2. Widget auto-height (wip)

**Changed**
1. Only one resize handle (bottom-right).
2. Fixes #3202 (test `allows opening menu after removal` now unskipped)
3. Fixes #3721.

**Known Issues**
1. ~~Removing widget in preview mode, generates js error.~~
2. ~~Dragging/Resizing but no change - still a saving indication is triggered~~
3. Widgets draggable beyond page [as reported here](https://github.com/getredash/redash/pull/3722#issuecomment-484705986). [Leaving unattended](https://github.com/getredash/redash/pull/3722#issuecomment-484866721).


ranbena:
@gabrieldutra  widget drag tests fail cause the placeholder now animates and also doesn't react as expected. Fiddling with it I realized there are 2 better ways to go about determining widget position delta:
1. Disable animations when testing.
2. Utilize the built-in retry mechanism of `cyElement.should` to wait until animation settles.
I like this better. This would require sth like this:
```js
// original
it('moves one column when dragged over snap threshold', () => {
  dragBy(cy.get('@textboxEl'), 110).then((delta) => {
    expect(delta.left).to.eq(200);
  });
});

// suggested
it('moves one column when dragged over snap threshold', () => {
  const drag = () => cy.get('@textboxEl').dragBy(110);
  const getLeft = () => cy.get('@textboxEl').invoke('offset').its('left');

  cy.get('@textboxEl').should(($el) => {
    expect(drag).to.increase(getLeft).by(200);
  });
});
```

But, it's not possible cause for some reason Cypress doesn't expect `increase()` argument to be a function, only an object, although [Chai API does allow it](https://www.chaijs.com/api/bdd/#method_increase) (introduced [long ago](https://github.com/chaijs/chai/pull/607)). 

What am I doing wrong?

gabrieldutra:
> What am I doing wrong?

Isn't this `getLeft` async? (as it uses `cy.get`) Perhaps if you turn this spec into a jQuery style by getting the element first `cy.get().then(($el) => {...})`. Then you should be able to run functions and assert with `$el` (there is a chance that the element in the variable doesn't get updated, if so, assertions won't work).

> Also, did we already talk about this?

Kind of, I wanted before to use Chai with elements directly (it doesn't work at all), this is a bit different :slightly_smiling_face:.

gabrieldutra:
I've noticed this one while playing:
to-the-infinite

ranbena:
> I've noticed this one while playing:
> to-the-infinite

10x @gabrieldutra, that's a grid feature I don't care for either.
Fixed in https://github.com/STRML/react-grid-layout/pull/549 but not merged yet :/

client/app/components/dashboards/dashboard-grid.jsx
```diff
@@ -0,0 +1,121 @@
...
+      i: id.toString(),
+      x: pos.col,
+      y: pos.row,
+      w: pos.sizeX,
+      h: pos.sizeY,
+      minW: pos.minSizeX,
+      maxW: pos.maxSizeX,
+      minH: pos.minSizeY,
+      maxH: pos.maxSizeY,
+      __proto__: {
```
kravets-levko: I think no need to use `__proto__` here - just add `toString` property to object.

client/app/components/dashboards/lazyInjector.js
```diff
@@ -0,0 +1,16 @@
...
+        return $injector.get;
+      },
+    };
+  },
+  set $injector(_$injector) {
+    $injector = _$injector;
+  },
+};
+
+export default lazyInjector;
```
kravets-levko: I think there is a room for improvements:
1. move `export default function init(ngModule)` + `ngModule.run(...)` to this file and init `$injector` global here as well;
2. remove setter from `lazyInjector`;
3. simplify it:
```js
export const $injector = {
  get get() {
     return $injector.get;
  };
}

// usage:
import { $injector } from './lazyInjector';
export ... = anrgular2react(..., ..., $injector);
```
4. move this code to `@/services/ng`

client/app/config/dashboard-grid-options.js
```diff
@@ -13,7 +13,7 @@ const dashboardGridOptions = {
 };
 
 export default function init(ngModule) {
-  ngModule.constant('dashboardGridOptions', dashboardGridOptions);
+  ngModule.constant('dashboardGridOptions', gridOptions);
```
kravets-levko: Not sure if it should stay injectable - just remove it and import directly where needed (I think only some services need it)

client/app/components/dashboards/dashboard-grid.jsx
ranbena: You're right 👍
Do you think I should keep it? It's good for debugging but probably bad for perf.

kravets-levko: ¯\_(ツ)_/¯ I think it's no _too_ bad for perf until you start using `toString` too frequently.

client/app/config/dashboard-grid-options.js
ranbena: Updated.

client/app/components/dashboards/dashboard-grid.jsx
ranbena: Updated.

client/app/components/dashboards/lazyInjector.js
ranbena: I tried.. but couldn't make it work. Don't wanna invest more time in it (especially knowing this will be soon removed).
But feel free to push an update if it comes easy to you.

ranbena:
> I played around a bit with the migrated features, didn't find any new issues 🙂.

10x!
 
> +367 −727 LOC? 😮

There's still 2 upcoming PRs so don't believe it yet!

> BTW, Why `dashboard-grid.jsx` and not `DashboardGrid.jsx`?

Right - I started work on this PR quite long ago before we had definitive standards.

arikfr:
While trying it with the deploy preview [Viz demo dashboard](https://deploy-preview-3722--redash-preview.netlify.com/dashboard/demo) I noticed there is considerable lag when going into edit mode and when exiting edit mode (even when not doing any changes).

Any idea why is this?

ranbena:
> While trying it with the deploy preview [Viz demo dashboard](https://deploy-preview-3722--redash-preview.netlify.com/dashboard/demo) I noticed there is considerable lag when going into edit mode and when exiting edit mode (even when not doing any changes).
> 
> Any idea why is this?

@arikfr when clicking "Edit", all grid item's `isDraggable` and `isResizable` props are set to `true`, which renders each with a `react-draggable` wrapper and a `react-resizable` wrapper. It makes sense that the more grid items, the more time it would take to init them.

Tested these props with persistent `false` and lag disappeared.
Didn't find related open issue in repo.
I have no simple solution in mind.

One way is to introduce lazy loaded grid items according to viewport (using the Intersection Observer API).
@kravets-levko @gabrieldutra any suggestions?

kravets-levko:
> While trying it with the deploy preview Viz demo dashboard I noticed there is considerable lag when going into edit mode and when exiting edit mode (even when not doing any changes).

On my Core i7 / 8Gb RAM there is also noticeable lag when dragging widgets (seems it happens when library is computing new positions for widgets).

Except of this I didn't notice any bugs in dashboard behavior - I can move/resize widgets, everything gets saved, and seems there is no more that annoying bug with disabled buttons in widget header.

kravets-levko:
@ranbena https://github.com/STRML/react-grid-layout/issues/626, https://github.com/STRML/react-grid-layout/issues/704

arikfr:
In our case these issues might be caused by the React and Angular hybrid,
where it can't decide if something should be redrawn or not.

kravets-levko:
@arikfr Probably yes, I'll check if it's possible to do something with that.

Also, noticed a bug - widgets in the last column are not properly aligned (it gets fixed if I resize browser's window):
image

arikfr:
Worth checking this after we merge your viz work. It will either make it
better or worse :)

kravets-levko:
@arikfr I don't think it will become better - my PR will introduce one more `react - angular - react` layer (currently everything inside widget is Angular, with my PR widget will be in Angular, then visualization wrapper - in React, and each visualization in Angular again)

kravets-levko:
@arikfr @ranbena So, issue is not with our code, but because `react-grid-layout` re-renders resizable/draggable wrappers when corresponding props change. Also, I have a fix for that, probably will do a PR to `react-grid-layout` repo (it will also fix their issues I mentioned above). Probably we'll need to use a fork until they approve/merge/publish that.

kravets-levko:
Okay, someone already did it: https://github.com/STRML/react-grid-layout/pull/892

arikfr:
I just opened the last version on Netlify, and:

1. The delete ('X') button is always visible.
2. Considering you can only resize from the bottom-right corner, maybe the indicator there should be more prominent?

(2 is NTH, but 1 is a blocker of course)

arikfr:
Also the performance on huge dashboards isn't great, but I guess there isn't much we can do aside from pushing forward with converting widgets and visualizations to React?

arikfr:
Few more things:


1. The widget dropdown menu isn't seen if the widget is too short, see the text box here: https://redash-preview.netlify.com/dashboard/short-widget.
2. When the widget is too high, it overlaps with the bottom help box and you can't resize it. See the Albums widget on the dashboard above.
3. The trigger to open the widget menu is vertical on widgets and horizontal on text boxes. We should pick one :)

kravets-levko:
@arikfr I tested it on our React migration dashboard (which I think is pretty large): it is a bit slower when moving/resizing widgets comparing to `gridster`-based variant, but not _too_ slow. I think we should migrate everything to React first, and only then test it again to decide what to do with it. Issue with lag when entering/leaving editing mode is completely fixed.

For "X" button - yeah, don't know how did I miss that. Will fix now.

arikfr:
Ignore 2 in my last comment -- it's actually fixed in this PR :)

ranbena:
@kravets-levko can you rebase with master?

kravets-levko:
@ranbena 

```bash
$ git merge master
Already up to date.
$
```

ranbena:
@kravets-levko I thought it wasn't cause it's missing the grid markings https://github.com/getredash/redash/pull/3656 but it's actually a bug.
I can work on it and push to this branch.

ranbena:
@kravets-levko I pushed the fix. Tested on Chrome/FF/Safari on Mac. Tested public dashboard.
Added one change: markings now appear also when there are no widgets in dashboard yet 🤟

ranbena:
@arikfr 

Before | After
-------|-------
Screen Shot 2019-05-13 at 17 27 47 | Screen Shot 2019-05-13 at 17 28 23

arikfr:
Merged 🚀

PR #3726, opened 2019-04-19
Refactored dashboard drag/resize testing

ranbena:
- [x] Refactor

## Description
As preparation for #3722 I leveraged the retrying nature of `should()` to "wait" till drag/resize animations are over, therefore making these methods much simpler and more generic.

This is a no brainer - all test pass and that's all that matters!

---------------

@gabrieldutra I couldn't make `increase(func).by(value)` work, so instead I just assert the end value (width/height/top/left). It's less pretty but works just as well.

### DIFF
--- a/client/cypress/integration/dashboard/dashboard_spec.js
+++ b/client/cypress/integration/dashboard/dashboard_spec.js
@@ -310,21 +310,21 @@ describe('Dashboard', () => {
         });
 
         it('stays put when dragged under snap threshold', () => {
-          dragBy(cy.get('@textboxEl'), 90).should(($el) => {
-            expect($el.offset().left).to.eq(15); // no change, 15 -> 15
-          });
+          dragBy(cy.get('@textboxEl'), 90)
+            .invoke('offset')
+            .should('have.property', 'left', 15); // no change, 15 -> 15
         });
 
         it('moves one column when dragged over snap threshold', () => {
-          dragBy(cy.get('@textboxEl'), 110).should(($el) => {
-            expect($el.offset().left).to.eq(215); //  moved by 200, 15 -> 215
-          });
+          dragBy(cy.get('@textboxEl'), 110)
+            .invoke('offset')
+            .should('have.property', 'left', 215); //  moved by 200, 15 -> 215
         });
 
         it('moves two columns when dragged over snap threshold', () => {
-          dragBy(cy.get('@textboxEl'), 330).should(($el) => {
-            expect($el.offset().left).to.eq(415); //  moved by 400, 15 -> 415
-          });
+          dragBy(cy.get('@textboxEl'), 330)
+            .invoke('offset')
+            .should('have.property', 'left', 415); //  moved by 400, 15 -> 415
         });
       });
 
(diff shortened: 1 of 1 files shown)


client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -402,26 +376,28 @@ describe('Dashboard', () => {
         });
 
         it('stays put when dragged under snap threshold', () => {
-          resizeBy(cy.get('@textboxEl'), 0, 10).then((delta) => {
-            expect(delta.height).to.eq(0);
-          });
+          resizeBy(cy.get('@textboxEl'), 0, 10)
+            .then(() => cy.get('@textboxEl'))
```
gabrieldutra: I think this being necessary is a sign that passing a `cy.get()` as an argument is an anti-pattern (similar to [this one](https://docs.cypress.io/guides/references/best-practices.html#Assigning-Return-Values)). We should rather use it as a Cypress command or pass something else as argument.

For now I think it's ok to leave as is, but let's keep in mind this to when we discover a better way to treat those helper commands.

client/cypress/integration/dashboard/dashboard_spec.js
```diff
@@ -342,20 +310,20 @@ describe('Dashboard', () => {
...
-            expect(delta.left).to.eq(200);
+          dragBy(cy.get('@textboxEl'), 110).should(($el) => {
+            expect($el.offset().left).to.eq(215); //  moved by 200, 15 -> 215
           });
         });
 
         it('moves two columns when dragged over snap threshold', () => {
-          dragBy(cy.get('@textboxEl'), 330).then((delta) => {
-            expect(delta.left).to.eq(400);
+          dragBy(cy.get('@textboxEl'), 330).should(($el) => {
```
gabrieldutra: Here it's also possible to have Cypress' syntax :)
```javascript
dragBy(cy.get('@textboxEl'), 330)
  .invoke('offset')
  .its('left')
  .should('eq', 415); // moved by 400, 15 -> 415
```

ranbena: Actually, this won't work cause the `invoke+its` would execute only once (which would surely yield a fail), while the `except` within the `should` re-executes till passed or timed-out.

gabrieldutra: I had had a similar thought, but ever since this worked (at least to track the change) I started to believe more in Cypress' magic :slightly_smiling_face: :
```javascript
cy.document()
    .its('fonts')
    .invoke('check', '14px FontAwesome')
    .should('eq', true);
```

dashboard_spec

ranbena: I tried it locally and always get a fail 🤔
I need to watch your video slowed down. Can you send me the vid?

gabrieldutra: That's weird :thinking: 

Another way to do it that is even safer as the command that precedes `should` is what should be repeated:
```javascript
dragBy(cy.get('@textboxEl'), 330)
  .invoke('offset')
  .should('have.property', 'left', 415); // moved by 400, 15 -> 415
```

> Can you send me the vid?

This one was created as a gif haha, I pushed to a temporary repo so we could test it over CI -> https://circleci.com/gh/getredash/redash/20305 (Succeeds)

ranbena: This works well 👍Updated.
```js
.invoke('offset')
.should('have.property', 'left', val);
```

client/cypress/integration/dashboard/dashboard_spec.js
ranbena: I worked on making `dragBy` and `resizeBy` into commands but it didn't go smooth so I'm leaving as is to keep moving forward.

Issue #2316, opened 2018-02-14
Refresh query when parameters update


arikfr:
Currently when you change parameter values the query doesn't update until you execute it again. This can be somewhat confusing, specially with dropdown parameters, where nothing happens after you pick a value.

The main reason for this was to avoid running a query before the user updated all the parameter values (in case there is more than one).

Some ideas for solutions:

* Execute the query when the parameter value is updated. Add some `debounce` (that resets when the user "touches" another parameter?) to let the user update other parameters before running the query.
* Add some sort of "Apply" button next to the params, so it's clear you need to click on it. Maybe even add some indication on it when the params are changed (like we do for the save button).
* Add a "Pause execution" toggle, that allows you to pause query refresh in case you want to update multiple parameters.

Any other ideas?

kocsmy:
I like the second option, I think that could work well.

kocsmy:
So here's my solution:

Add an execute button (similar to Show Data Only mode bottom right "Play" icon), that stays disabled until you edit a parameter. Once you change any values, the button becomes active indicating that you need to click to re-execute the query.

I think this would work nice because we don't need to add any 'debounce' or 'Pause execution' and should be quite intuitive.

![query parameter](https://user-images.githubusercontent.com/2378022/42137272-c8ad7bba-7d69-11e8-8938-31f2ec5b230c.png)

arikfr:
> Add an execute button (similar to Show Data Only mode bottom right "Play" icon), that stays disabled until you edit a parameter.

I think that once we add it, people will get used to using it to refresh the query. It will be confusing to have it disabled sometimes and enabled at other times.

Maybe we can make it "glow" to indicate you need to click it?

Also, how will the button be placed when you have multiple parameters? Maybe even having two lines of them?

kocsmy:
![query parameter](https://user-images.githubusercontent.com/2378022/42137482-57b53a34-7d6d-11e8-9be8-e484a7c5263f.png)

arikfr:
> I'm wondering why don't we just leave it there always enabled?

I think you had a good point about indicating to the user that they need to press it.

> We'll just add it to the right of the last item. If there are multiple rows, it'll go after the last one.

It might look confusing/ugly -- can you create a mockup of multiple parameters and another one with multiple rows?

kocsmy:
I think should use the * like we are using with the Save. Button Glow would be too foreign to for the current design.
[...]

PR #3737, opened 2019-04-24
Refresh query when parameters update

gabrieldutra:

- [x] Feature

## Description
This is a work on #2316.

So far, I've added a touch state to Query Parameters and it now refreshes query results whenever a parameter is modified (debounce already added). Next step will be to add an "Apply" button for input fields while refining UX.

- [x]  Refresh query when parameters are modified
- [x] Add "Apply" button for input fields and let auto-refresh for the rest (Dropdowns, Time pickers)
- [x] Remove button from #3752

#2316 

Not yet :sleeping:


gabrieldutra:
Thanks for the comments @arikfr, I've imagined this would be applied to other places where Parameters is used. I thought about doing as you mentioned before, the main reason I went with more logic in the `view` was thinking about the need of an "Apply All" functionality in the future (e.g: when clicking Execute Query button). I though about 2 ways to make it work:
- Having an exposed method in `Parameters` component that would force update, so I could use refs and when needed just run `parametersComponent.applyAll()` - however, I'm not so sure how to handle it with both Angular and React coexisting
- Lift state up - things end up in the `view` :laughing: (that's why I had changed `getQueryResults`)

Well, I agree that the second one is not ideal as we would repeat a lot of logic across components, so I'm moving to 1. Gonna make it work first without this functionality, then find a way to handle it. In any case, this is sort of what I had in mind.

Regarding debounce, NP, I'm more using it for testing, later we figure out where it should be.

arikfr:
@gabrieldutra I don't think the component need to expose a method, but rather it should use the same data model object that the parent component will use.

Also, not sure that executing a query should apply the new inputs. It's probably simpler to have the parent components see the old value until it's applied by the input component. This should make implementation simpler and makes sense UX wise.

gabrieldutra:
> Also, not sure that executing a query should apply the new inputs. It's probably simpler to have the parent components see the old value until it's applied by the input component. This should make implementation simpler and makes sense UX wise.

I thought there could be a case of pressing, for example, `Ctrl+Enter` all inputs would be applied. In any case, I'm with: "It's probably simpler to have the parent components see the old value until it's applied by the input component".

For the sake of moving forward, I've added the "Apply Button" and played a little bit with it:

**Current implementation**
width-fixed

**With auto width** - it's kind of cool, but there's the issue of wrapping as you increase the final width
width-auto


There was an issue with Spinners in Number Inputs and the parameters, so I've removed them. LMK if you prefer a different solution :slightly_smiling_face: 

number-input

@arikfr, @ranbena suggestions for UX/UI here?

ranbena:
@gabrieldutra these are very cool suggestions.
* I would rather avoid a button within the input element cause it may cause unforeseen problems.
* Should there be a cancel button along with the apply? 

My tendency is to do something similar to the UI in param editor:
Screen Shot 2019-04-30 at 22 24 38

So perhaps something like this:



That when clicked turns to this:

arikfr:
@ranbena there is no need for edit button, as this is an input field... Having the buttons outside of the control does indeed solve some issues, but is it clear what it will do?

Although I wonder if we should implement this UI:

image

(including static label for value until edited) 🤔 

I'll sleep on it.

@kravets-levko @rauchy wdyt?

ranbena:
> @ranbena there is no need for edit button, as this is an input field... Having the buttons outside of the control does indeed solve some issues, but is it clear what it will do?

I meant for this "edit" button to swap the "cog" (should keep it cog tho), and when `input.onChange` to swoosh the apply/cancel buttons in over it.

arikfr:
> I meant for this "edit" button to swap the "cog" (should keep it cog tho), and when input.onChange to swoosh the apply/cancel buttons in over it.

Oh, I see now. Well, the cog only visible for the query editor in edit mode, so not sure if I would bother with moving it or changing it.

My concern that it won't be intuitive what this confirm/cancel button do. But it seems like the simplest solution for now, so maybe worth starting with it and iterating from there.

@gabrieldutra wdyt?

rauchy:
I still find the word "Apply" more descriptive and easy to understand than an icon.

gabrieldutra:
> I would rather avoid a button within the input element cause it may cause unforeseen problems.

I did validate this for different resolutions, but I share the same concern. 

> Should there be a cancel button along with the apply?

If the button will be added in the "same place" as the input (not in a popover) I think it's better not to have a cancel button. Seems simpler without it + I don't think it adds much in this case.

Regarding Static label (or Popover): the idea is interesting, but it adds a new step to the process. I mean, when editing 1 parameter it's fine, but when editing 5 this may get annoying. Also, it wouldn't be possible to use only the keyboard when editing more than 1 parameter.

I agree with @rauchy that the word "Apply" is more intuitive. Perhaps explore it without it being within the input?

arikfr:
My thoughts after reviewing this again and playing with the current version:

* I agree that `Apply` is more intuitive. But placing it outside of the Input will most likely make it less intuitive, so I feel that this is not worth exploring.
* Auto width doesn't feel necessary, because the input is large enough as it is even after adding the button.

It seems that the only issue are the spinners for numbers -- I wasn't that attached to them, so don't mind them going away. If we get feedback they were useful, we can explore other options. But for now I think that your current implementation is good 👌 

Let's move on ⏩ and get this implemented :)

gabrieldutra:
I'll have to work on a different approach to it anyway, following @ranbena's comment, I've just tested it on Firefox:

firefox-apply

Remove the border from the input and put it in a `div` that wraps both the input and the button should be more reliable and give the same effect. I think I'll leave as is until I finish the core functionality, just in case other ideas come :)

arikfr:
>  I think I'll leave as is until I finish the core functionality, just in case other ideas come :)

Good call. Let's finish the functionality, add debounce where needed and do the FF fix in a follow up PR so we can release this already. While the bouncing inputs are annoying, it's still better than current behavior.

gabrieldutra:
I've updated this to include the Apply Button in the other places. @arikfr, is anything missing for this to be done in terms of functionality? If not, this can be open for review :slightly_smiling_face: 

The best place to put debounce seemed to be the `Parameters` component in the `onChange` method. Ideally this should debounce the parameter's `setValue` as well, but this will be easier when we migrate the `Parameters` component (by using states and so).

arikfr:
I think you can open this to review already. Let's finalize this before moving on to implementing the better rendering.

Two issues I did notice when testing it:

- [ ] When you hit "Cmd + Enter" inside an input dialog it will execute the query twice (because both the page & input dialog catch the Enter key event).
- [ ] DateTime Parameter: it should apply only when you click on `Ok` (or close the dialog?), but currently it applies immediately when picking some date. This is specially an issue with date/time input, because it doesn't give you the chance to pick the time.

gabrieldutra:
> When you hit "Cmd + Enter" inside an input dialog it will execute the query twice (because both the page & input dialog catch the Enter key event).

I've excluded both Meta (cmd) and Ctrl key, if someone can validate this for MacOS :).

> DateTime Parameter: it should apply only when you click on Ok (or close the dialog?), but currently it applies immediately when picking some date. This is specially an issue with date/time input, because it doesn't give you the chance to pick the time.

Antd's `DatePickers` don't have Ok buttons while `DateTimePickers` do, I've updated both `DateTime` and `DateTimeRange` to update values when you close the dialog (it was the one with apparently less issues). I used some hooks to manage state in the component :slightly_smiling_face:. There's an `onOk` callback which could be an alternative, but it keeps the selected values when you close the dialog regardless of the ok button, so it could be confusing.

client/app/components/parameters.html
```diff
@@ -18,14 +18,6 @@
...
+    <parameter-value-input param="param" on-change="onValuesChangeDebounced" apply-button="true"></parameter-value-input>
   </div>
-  <button
-    class="m-t-20 btn btn-primary hidden-print"
-    ng-if="onRefresh"
-    ng-click="onRefresh()"
-    title="Refresh Dataset"
-  >
-    <span class="zmdi zmdi-play"></span>
-  </button>
```
gabrieldutra: @rauchy :)

client/app/components/ParameterValueInput.less
```diff
@@ -0,0 +1,19 @@
...
+.parameter-apply-button {
+  margin-left: -55px;
+  width: 50px;
+}
+
+.parameter-input {
+  width: 195px !important;
+
+  &--apply-button {
+    padding-right: 60px !important;
```
ranbena: To make it equal the 4px from the right edge.

```suggestion
    padding-right: 58px !important;
```

ranbena:
> It seems that the only issue are the spinners for numbers -- I wasn't that attached to them, so don't mind them going away. If we get feedback they were useful, we can explore other options. But for now I think that your current implementation is good 👌

I'm not sold on it but we can avoid the functionality regression by insetting the apply button (and persisting spinner opacity) for number inputs:

Screen Shot 2019-05-09 at 17 07 18

client/app/components/ParameterValueInput.less
```diff
@@ -0,0 +1,19 @@
+@import '~antd/lib/input-number/style/index'; // for ant @vars
+
+.parameter-apply-button {
+  margin-left: -55px;
```
ranbena: This is the cause of the FF display issue.
It's better to take the button "out of the flow" so it doesn't affect neighboring elements.

```css
parameter-value-input {
  position: relative;
  display: inline-block;
}

.parameter-apply-button {
  width: 50px;
  position: absolute !important;
  right: 4px;
  top: 6px; /* 5px after button height set to 25px */
}
```

client/app/components/ParameterValueInput.less
```diff
@@ -1,19 +1,50 @@
...
-    display: none;
+    opacity: 1 !important;
+  }
+
+  .parameter-apply-button {
+    position: absolute !important;
+    height: 25px !important;
+    width: 50px;
+    top: 5px;
+    right: 26px;
```
ranbena: 😬
```css
right: 27px;
```

gabrieldutra:
@ranbena, I've updated the styling after considering your comments :). What did you base `right: 4px;` in?

apply-button-4px

Should it be 5px instead to fit top & bottom of Apply Button's position?

> When text is long and starting to edit, the edited area is suddenly out of bounds, so user has to stop and scroll it into view to keep editing..

This is sort of an edge case if you consider that when you go on typing it automatically goes to the caret position, I think if there is an easy solution we should do. I'll see what I find for this.

> Since all input param types now apply only on selection / apply click - no reason to debounce at all, right?

I think the idea is to have some time to apply more parameters without executing the query multiple times (although the 1 sec I let for tests seems not to be enough ^^)

ranbena:
> @ranbena, I've updated the styling after considering your comments :). What did you base `right: 4px;` in?

Looking good. 4px comes from the button's distance from top and bottom. For visual consistency it should be 4px from all sides.

> Should it be 5px instead to fit top & bottom of Apply Button's position?

With absolute positioning it should be `top: 5px; right: 5px` 👍 

> > When text is long and starting to edit, the edited area is suddenly out of bounds, so user has to stop and scroll it into view to keep editing..
> 
> This is sort of an edge case if you consider that when you go on typing it automatically goes to the caret position, I think if there is an easy solution we should do. I'll see what I find for this.

👍 

> > Since all input param types now apply only on selection / apply click - no reason to debounce at all, right?
> 
> I think the idea is to have some time to apply more parameters without executing the query multiple times (although the 1 sec I let for tests seems not to be enough ^^)

Why is it needed though? Doesn't the previous execution get cancelled on re-execution?
The debounce has a bad UX effect in which you click "apply" and nothing happens for a sec. It's annoying and we should avoid it.

ranbena:
> This is sort of an edge case if you consider that when you go on typing it automatically goes to the caret position, I think if there is an easy solution we should do. I'll see what I find for this.

@gabrieldutra just noticed that if you keep typing, Chrome will automatically scroll to caret. So can leave as is.

gabrieldutra:
> Why is it needed though? Doesn't the previous execution get cancelled on re-execution?
The debounce has a bad UX effect in which you click "apply" and nothing happens for a sec. It's annoying and we should avoid it.

Totally agreed, I'm not really sure if the previous get cancelled, for MySQL iirc it doesn't work ^^. I guess in such cases it would queue the query executions (which can be an issue for slow queries). For the sake of testing I've removed debounce:

execution-queued
(BTW, interesting way to test slow queries ^^)

It seems that: - when there's one worker available, the triggered execution goes directly to it; - when the execution is queued (not running yet), it is replaced by the new one(?)

ranbena:
> Totally agreed, I'm not really sure if the previous get cancelled, for MySQL iirc it doesn't work ^^. I guess in such cases it would queue the query executions (which can be an issue for slow queries). For the sake of testing I've removed debounce:

Well I can tell you the user experience now is soooo much better :)
> (BTW, interesting way to test slow queries ^^)

Nice!

> It seems that: - when there's one worker available, the triggered execution goes directly to it; - when the execution is queued (not running yet), it is replaced by the new one(?)

@arikfr how does it work?

arikfr:
> It seems that: - when there's one worker available, the triggered execution goes directly to it; - when the execution is queued (not running yet), it is replaced by the new one(?)

We deduplicate executions of the same query multiple times, i.e. if the query already executes and you execute it again, it will "join" the previous execution. So in the above case, because you didn't change the values, it was all one execution.

Indeed the idea here was to avoid multiple executions when you want to update several parameters. But thinking about it, I see that debounce is not enough -- what needs to happen is more complex: if you start editing another parameter, we should wait with execution.

Let's roll out without debounce, and improve this based on feedback from users and our experience using this.

arikfr:
Is this ready in terms of implementation?

ranbena:
A sidenote on debounce and UX:
The problem is not the debounce itself, but the delay in "responsiveness". There must be a visual indication that the "apply" click has been registered and acted upon - even if in actuality it's idle.

Similar to the dashboard [save debounce implementation](https://github.com/getredash/redash/blob/master/client/app/pages/dashboards/dashboard.js#L90).

gabrieldutra:
> Is this ready in terms of implementation?

Considering the last comments, yes. I do have a concern about slow queries, but I think it's safe to go through "Let's roll out without debounce, and improve this based on feedback from users and our experience using this.".

> There must be a visual indication that the "apply" click has been registered and acted upon - even if in actuality it's idle.

This is sort of why I was keeping with 1 sec of debounce only, but definitively better without debounce or with a visual indication.

client/app/components/ParameterValueInput.jsx
```diff
@@ -132,25 +159,64 @@ export class ParameterValueInput extends React.Component {
...
+    };
+
     return (
-      <InputNumber
-        className={'form-control ' + className}
-        defaultValue={!isNaN(value) && value || 0}
-        onChange={onSelect}
-      />
+      <div
+        className={classNames('parameter-input-number', { 'parameter-input-number--apply-button': showApplyButton })}
```
ranbena: When it comes to element state, I tend to favor describing the state rather than it's impact (cause that usually changes as app progresses) so instead of classname `parameter-input-number--apply-button` I'd go with `input-dirty`. And if it's a bool, I'd even ditch classnames and do a `data-dirty="true"` attribute on the element.

For your consideration 👨‍💻

ranbena: Oh, this is just to have the apply button on/off 🤦‍♂️

ranbena: Hmm so I would have these as two separate attributes actually.

Where does `$ctrl.applyButton` determined? When is it `false`?

ranbena: Found it - hardcoded to `true`. Then perhaps this prop can be removed and code simplified?

ranbena:
Another sidenote - the reason I prefer an icon to textual label in this case ("apply") is that some day we'll incorporate i18n, and this won't do 🔮

client/app/components/ParameterValueInput.jsx
gabrieldutra: > When is it `false`?

`ParameterMappingInput` :slightly_smiling_face: (e.g: in the Add Widget Dialog). Also I thought it could be interesting to support it without the Apply Button.

Regarding the css, yes, it's to turn on/off the appearing of the apply button on the input. I tried to follow [BEM naming convention](http://getbem.com/naming/) on that, but yes, I can use a data attribute instead. It will actually look better than this `classNames` syntax.

gabrieldutra: BTW, it was Cypress that made me realize that the Apply Button would affect the Add Widget Dialog :grin:

arikfr: > I tried to follow BEM naming convention on that

Maybe we should have a conversation on whether we want to keep using BEM.

arikfr:
> Another sidenote - the reason I prefer an icon to textual label in this case ("apply") is that some day we'll incorporate i18n, and this won't do 🔮

🤔 fair point. Let's keep it as is for now with Apply and revisit in the future. I feel like the whole UI for parameters is due for deeper UX work. One option is to have it in a non editable form when presenting and have a proper UI for editing. Here's an example from Intercom:

image

And when editing:

image

---

About an alternative to debounce, we could have a dropdown next to the Apply. 



~Not sure it's appropriate considering the limited space we have for it.~ We can't do it with current UI, because it will work only with `<Input>`s but not dropdowns or dates. But this is something we can do once we switch to rendering a dedicated UI for editing.

arikfr:
Date Range parameters don't seem to update on dashboards. See example here: https://deploy-preview-3737--redash-preview.netlify.com/dashboard/legacy-parameters-dashboard?p_dateparam=2019-01-21&p_AlbumId=2001-01-20&p_time_frame.start=2019-01-03&p_time_frame.end=2019-01-11&p_param1=title1&p_date%20range.start=2019-05-24&p_date%20range.end=2019-05-27.

client/app/components/ParameterValueInput.jsx
gabrieldutra: Updated :+1:

> Maybe we should have a conversation on whether we want to keep using BEM.

This was actually the only case where I was applying BEM (modifiers like `--apply-button`). But I preferred as @ranbena suggested `data-dirty`. I think BEM is sort of deprecated for Redash already, just need to update the docs if that's really the case.

gabrieldutra:
> Date Range parameters don't seem to update on dashboards

Actually none of the parameters are updating in there, it seems that `debounce` had an extra job for the order of functions. I'll dig into it and try to come up with sth more reliable for this part.

client/app/components/ParameterValueInput.jsx
```diff
@@ -181,15 +244,21 @@ export default function init(ngModule) {
...
     bindings: {
       param: '<',
+      applyButton: '=?',
+      onChange: '=',
     },
     controller($scope) {
       this.setValue = (value) => {
         this.param.setValue(value);
-        $scope.$applyAsync();
+        $scope.$apply();
```
gabrieldutra: @arikfr changing `$applyAsync` to `$apply` here did the trick for the parameters not working on dashboards. What happens is that dashboards depends on Url Parameters to be updated (and it's a `watch` that does it) and the refresh was being triggered before.

arikfr: Luckily soon we will get rid of all these crap 😎

arikfr:
🤩

PR #3740, opened 2019-04-28
Use REDASH_BASE_PATH everywhere instead of hardcoded path

### LINKED ISSUE #3727: setup/*.sh scripts inconsistant use of /opt/redash
### Issue Summary

The docker setup scripts in setup/*.sh make inconsistent use of /opt/redash.
At the beginning of the script there is an env variable set, implying it could be another path (say a path that doesnt require sudo /root ) -- but no.  
the path /opt/redash is hard coded in many places.

Docker installs should not require local host hard coded paths in root-only-accessable directories.

arikfr:

- [x] Other

## Description

Uses `REDASH_BASE_PATH` everywhere instead of a hard coded path.

Closes #3727.


arikfr:
Thanks, @jezdez !

Issue #3719, opened 2019-04-17
Embedded visualizations with parameters: data loading issues


arikfr:
* When one of the parameters is missing a value --
  Current behavior: a generic error message is shown.
  Expected behavior: ~~either~~ show input for missing parameter ~~or show a meaningful error message~~.

* When changing parameter value, it tries to load new results but if they don't exist, it doesn't seem to poll for results.
  Expected behavior: show spinner and poll until there is a result.

~~Unlike parameters in in-app context, here when you change parameter value it tries to load a result. I'm not sure if we should implement this until we have implemented #2316. 🤔 WDYT? (cc: @ranbena)~~ On second thought, let's move on with this and hopefully #2316 will be implemented soon.

rauchy:
@arikfr regarding the second part - are you sure it's wise to proceed with this with #2316 in pretty advanced stages? Perhaps I should branch out of #2316?

PR #3741, opened 2019-04-28
Show input for missing parameters in embedded visualizations

rauchy:
- [x] Bug Fix

## Description
In embedded visualizations, if a parameter is missing - the `/results` endpoint errors out. This fix redirects to the default value of the parameter, resulting in the ability to enter a value in the parameter input fields.

Fixes the first part of #3719


client/app/components/queries/visualization-embed.js
```diff
@@ -38,7 +38,8 @@ export default function init(ngModule) {
     return session($http, $route, Auth).then(() => {
       const queryId = $route.current.params.queryId;
       const query = $http.get(`api/queries/${queryId}`).then(response => response.data);
-      const queryResult = $http.post(`api/queries/${queryId}/results`, { parameters: queryStringParameters() }).then(response => response.data);
+      // eslint-disable-next-line no-unused-vars
+      const queryResult = $http.post(`api/queries/${queryId}/results`, { parameters: queryStringParameters() }).then(response => response.data, _ => ({}));
```
rauchy: A polite way to solve this would have been to instantiate a `Query` right here, and see if all parameters are supplied, but since this is happening before Query is injected, I'm not sure if that's even possible.
So basically I'm ignoring the error, knowing that in the case of an error, the same request will be launched from `QueryResult` later on.

arikfr: Why not load the results after we have the query? 

It used to work this way, because we wanted to load everything and then render, assuming there will always be something to load. Now that we're making embed more interactive, it makes sense to just show a loading UI and load things properly.

rauchy: That's good context. Thanks, I'll modify this.

rauchy: I'm thinking it might be worth considering to keep it _as is_ - loading the results after we have the query will result in 2 consecutive roundtrips for _all_ embed page loads. If we keep the change suggested in this PR, it'll result in an extra roundtrip only for the rare case a parameter is missing. WDYT?

arikfr:
How when we call the results API changes number of requests we make in this
case?

rauchy:
It won't affect the number of requests, but it'll enforce them to become sequential and not parallel.

I'll try to explain:

* **Loading the query** -> checking all parameters are present -> **loading query results** (2 consecutive requests, 100% of the cases)
* **Loading the query** and simultaneously **loading the query results** -> possibly failing to get query results due to missing parameters -> **loading query results with default values found in the query** (most of the time 2 simultaneous requests, 1 extra roundtrip for the missing parameter edge case)

arikfr:
Ok, I understand now. Well, in this case we can still make both requests in
parallel, but not block the UI but rather show a loading screen.

For this you will probably need to move these calls away from the route
dependencies.

client/app/components/queries/visualization-embed.js
```diff
@@ -8,47 +8,46 @@ const VisualizationEmbed = {
...
+
+      const query = $http.get(`api/queries/${queryId}`).then(response => response.data);
+      const queryResult = $http.post(`api/queries/${queryId}/results`, {
+        parameters: queryStringParameters(),
+      }).then(response => response.data, (error) => {
+        if (error.status === 400) {
+          return {};
+        }
+
+        throw error;
```
rauchy: Previously, an unhandled promise rejection here would render a full-page error message.
Now, after the data-loading code has moved into the controller, the page layout is already loaded and re-throwing this error does not render an error layout. @kravets-levko / @ranbena could you please give me a little guidance on how to render a full-page error message here?

arikfr:
In what cases do we want a full page error here?

rauchy:
@arikfr for example, previously any attempt to embed a non-safe query would show an error.

client/app/components/queries/visualization-embed.js
kravets-levko: @rauchy here is a good example how to do it: https://github.com/getredash/redash/blob/f9cc230227b42799886decdf315dea5c273773f3/client/app/pages/data-sources/EditDataSource.jsx
1. inject Angular's `$exceptionHandler` service;
2. if it's HTTP error - you need to convert it to `PromiseRejectionError`
3. pass the error to `$exceptionHandler`

It should be something like this:
```js
// ANGULAR_REMOVE_ME This code is related to Angular's HTTP services
if (error.status && error.data) {
  error = new PromiseRejectionError(error);
}
$exceptionHandler(error);
```

rauchy:
We're back to this:

image

arikfr:
It's a bid hard to test this separately from #3752 (there is complete view refresh when changing values here). Also, some of the changes from #3752 are needed here and vice-versa (think the flow of loading the result the first time).

How about you move the changes from #3752 over here?

rauchy:
Why is it hard to test this? Just remove parameters from the query string.

arikfr:
Because you don't see things refresh.

Issue #3739, opened 2019-04-26
Monospace font in query output log


otsaloma:
I use the Python query runner and `print` calls to do debugging. The output shows up in normal paragraphs of sans-serif text, which doesn't really work for certain kinds of output, such as `pandas.DataFrame.head`. Below is what I mean, that's how I hacked it in my copy, but there is likely to be a more elegant fix.

```diff
diff --git a/client/app/pages/queries/query.html b/client/app/pages/queries/query.html
index d95b0b76..1eca38a2 100644
--- a/client/app/pages/queries/query.html
+++ b/client/app/pages/queries/query.html
@@ -224,7 +224,7 @@
                 <div class="d-flex flex-column p-absolute static-position__mobile" style="left: 0; top: 0; right: 0; bottom: 0;">
                   <div class="p-10" ng-show="showLog">
                     <p>Log Information:</p>
-                    <p ng-repeat="l in queryResult.getLog()">{{l}}</p>
+                    <p ng-repeat="l in queryResult.getLog()" style="font-family: monospace; white-space: pre; margin: 0;">{{l}}</p>
                   </div>
                   <ul class="tab-nav">
                     <rd-tab ng-if="!query.visualizations.length" tab-id="table" name="Table" base-path="query.getUrl(sourceMode)"></rd-tab>

```

arikfr:
This makes sense. You're welcome to make a pull request which implements this, but please move the style definition to a style file instead of inline styling.

PR #3743, opened 2019-04-29
Use monospace font in query output log

otsaloma:

- [x] Bug Fix

## Description

Query output log currently uses a regular sans-serif font. That is unreadable for some types of output, such as `pandas.DataFrame.head` (useful when debugging Python queries).

#3739


arikfr:
Thanks, @otsaloma ! :)

PR #3744, opened 2019-04-29
Fix query page height

ranbena:
- [x] Bug Fix

## Description
Fixed non-to-bottom height bug in query page introduced in #3656.


ranbena:
@arikfr any other pages that have an "all the way to the bottom" feature?

kravets-levko:
@ranbena No, query page is the only one with such layout. Other pages have a "regular" layout with vertical scrollbar.

PR #3748, opened 2019-04-30
Cypress: Fix Stuck E2E test - create_query_spec

### LINKED ISSUE #3688: Stuck E2E test - create_query_spec
### Issue Summary
When running Cypress tests with Electron, it halts on `create_query_spec.js`.
This happens in CI. Doesn't happen with Chrome but Electron is a must in order to have video capture (https://github.com/getredash/redash/pull/3683).

Screen Shot 2019-04-08 at 16 59 05

### Steps to Reproduce

1. `npm run cypress run` with `--browser electron`.

### Workaround

Marked as skipped.

gabrieldutra:

- [x] Other

## Description
After a bug that could be avoided by Percy (#3744) I took a time to investigate #3688. It turns out this was getting stuck due to the `onbeforeunload` confirm dialog :sleepy:.

This fix is a workaround for cypress-io/cypress#2118

Closes #3688

--

### DIFF
--- a/client/cypress/integration/query/create_query_spec.js
+++ b/client/cypress/integration/query/create_query_spec.js
@@ -4,8 +4,7 @@ describe('Create Query', () => {
     cy.visit('/queries/new');
   });
 
-  // https://github.com/getredash/redash/issues/3688
-  it.skip('executes the query', () => {
+  it('executes the query', () => {
     cy.getByTestId('SelectDataSource')
       .click()
       .contains('Test PostgreSQL').click();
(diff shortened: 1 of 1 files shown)


ranbena:
@gabrieldutra is this relevant cause the query is left unsaved?

gabrieldutra:
> @gabrieldutra is this relevant cause the query is left unsaved?

I did think about saving the query instead, though I considered having a clean state afterwards. However from a second thought, it can be interesting to turn it into a "executes and saves a query" :thinking:. I will change it here.

ranbena:
> I considered having a clean state afterwards.

Better to make sure tests are absolutely independent.

> it can be interesting to turn it into a "executes and saves a query" 🤔. I will change it here.

Perfecto.
Too bad `cy.route()` doesn't work with redirects 😒

gabrieldutra:
> Better to make sure tests are absolutely independent.

Yep, but so far there are no apparent issues for Query pages :)

We should probably create a few more tests prior to the Query Page migration, so until then we can stand with this.

PR #3749, opened 2019-04-30
Dashboard Grid #3 - Auto Height

ranbena:
- [x] Refactor

## Description
Ported the auto height feature

Included:
- [x]  Auto height param taken from widget service info
- [x] Triggered by change to height of widget header / body / footer
- [x] Feature available in single column mode
- [x]  Related tests unskipped and passing

## Mobile & Desktop Screenshots/Recordings

🔔 Red stripe indicator on top is for local testing purposes only. Not in prod. 🔔

### DIFF
--- a/client/app/components/dashboards/AutoHeightController.js
+++ b/client/app/components/dashboards/AutoHeightController.js
@@ -9,14 +9,14 @@ const WIDGET_CONTENT_SELECTOR = [
   '.spinner-container', // loading state
   '.tile__bottom-control', // footer
 ].join(',');
-const INTERVAL = 500;
+const INTERVAL = 200;
 
 export default class AutoHeightController {
-  widgets = {}
+  widgets = {};
 
-  interval = null
+  interval = null;
 
-  onHeightChange = null
+  onHeightChange = null;
 
   constructor(handler) {
     this.onHeightChange = handler;
@@ -61,7 +61,7 @@ export default class AutoHeightController {
         }, 0);
       },
     ];
-  }
+  };
 
   remove = (id) => {
     // not actually deleting from this.widgets to prevent case of unwanted re-adding
@@ -70,11 +70,11 @@ export default class AutoHeightController {
     if (this.isEmpty()) {
       this.stop();
     }
-  }
+  };
 
-  exists = id => !!this.widgets[id.toString()]
+  exists = id => !!this.widgets[id.toString()];
 
-  isEmpty = () => !some(this.widgets)
+  isEmpty = () => !some(this.widgets);
 
   checkHeightChanges = () => {
     Object.keys(this.widgets).forEach((id) => {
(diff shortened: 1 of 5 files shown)


client/app/components/dashboards/AutoHeightController.js
```diff
@@ -0,0 +1,109 @@
...
+      if (height && height !== prevHeight) {
+        this.widgets[id][1] = height; // save
+        this.onHeightChange(id, height); // dispatch
+      }
+    });
+  }
+
+  start = () => {
+    this.stop();
+    this.interval = setInterval(this.checkHeightChanges, INTERVAL);
```
ranbena: @kravets-levko Eventually implemented with polling cause `ResizeObserver` turned out to be not a good fit  - complex when observing multiple elements per widget.

client/app/components/dashboards/AutoHeightController.js
```diff
@@ -0,0 +1,109 @@
...
+
+const WIDGET_SELECTOR = '[data-widgetid="{0}"]';
+const WIDGET_CONTENT_SELECTOR = [
+  '.widget-header', // header
+  'visualization-renderer', // visualization
+  '.scrollbox .alert', // error state
+  '.spinner-container', // loading state
+  '.tile__bottom-control', // footer
+].join(',');
+const INTERVAL = 500;
```
ranbena: Seems ok to me. Lmk if you think it should be 50 like the original.

client/app/components/dashboards/dashboard-grid.jsx
```diff
@@ -126,6 +167,8 @@ class DashboardGrid extends React.Component {
           margin={[cfg.margins, cfg.margins]}
           isDraggable={this.props.isEditing}
           isResizable={this.props.isEditing}
+          onResizeStart={this.autoHeightCtrl.stop}
```
ranbena: When widget is manually resized, interval is paused and then resumed to avoid "height fights" and also better perf.

client/app/components/dashboards/AutoHeightController.js
kravets-levko: I think it's fine. We can adjust it later if needed.

client/app/components/dashboards/widget.less
```diff
@@ -105,4 +105,9 @@ visualization-name {
   &.resizing {
     z-index: 3;
   }
+
+  // autoHeight animation
+  &.cssTransforms:not(.resizing) {
```
kravets-levko: `kebab-case` please :pray:

client/app/components/dashboards/AutoHeightController.js
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        // get all content elements
+        const els = widgetEl.querySelectorAll(WIDGET_CONTENT_SELECTOR);
+
+        // calculate accumulated height
+        return reduce(els, (acc, el) => {
+          const height = el ? el.getBoundingClientRect().height : 0;
+          return acc + height;
+        }, 0);
```
ranbena: PS I would like to eventually move away from checking multiple element heights - with the right HTML/CSS changes, the same look and functionality can be implemented with observation of a single element.

PR #3619, opened 2019-03-21
Support for Presto password

aidarbek:
## What type of PR is this? (check all applicable)

- [x] Feature

## Description
Adds support for password in Presto data source.

## Related Tickets & Documents
Fix #2925 

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
Снимок экрана 2019-03-21 в 18 27 46

PR #3757, opened 2019-05-02
Fix: default value for Presto password should be None

arikfr:

- [x] Bug Fix

## Description

If we pass empty string as a password, PyHive will complain if we don't set the protocol to `https` as well. This change fixes this by using `None` as the default value for the password.

#3619

### DIFF
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -1,10 +1,10 @@
+from collections import defaultdict
 from redash.query_runner import *
 from redash.utils import json_dumps, json_loads
 
 import logging
 logger = logging.getLogger(__name__)
 
-from collections import defaultdict
 
 try:
     from pyhive import presto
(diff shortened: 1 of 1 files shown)


redash/query_runner/presto.py
```diff
@@ -99,30 +99,33 @@ def get_schema(self, get_stats=False):
...
-                protocol=self.configuration.get('protocol', 'http'),
-                username=self.configuration.get('username', 'redash'),
-                password=self.configuration.get('password', ''),
-                catalog=self.configuration.get('catalog', 'hive'),
-                schema=self.configuration.get('schema', 'default'))
+            host=self.configuration.get('host', ''),
+            port=self.configuration.get('port', 8080),
+            protocol=self.configuration.get('protocol', 'http'),
+            username=self.configuration.get('username', 'redash'),
+            password=self.configuration.get('password'),
```
arikfr: This is the actual change. The rest is just autopep8... 🤦‍♂

Issue #3074, opened 2018-11-13
Hide widget level filters when dashboard filters are active


s1moe2:
### Issue Summary

When dashboard level filters are enabled, all widgets still keep their individual filters, which produces a lot of visual noise and redundancy (I also found this mentioned on [discuss](https://discuss.redash.io/t/hide-widget-level-filters-when-dashboard-level-filters-are-on/1571)).
There may be use cases when both dashboard and widget level filters are needed but generally I see this as a "switch".

### Technical details:

I've tried locally and came up with a simple solution to hide the widget filters according to the flag affected by the setting by basically adding this `ng-if` in [client/app/visualizations/index.js] (and a couple more lines, ofc):
```template: `<filters ng-if="!dashboardFiltersEnabled" filters="filters"></filters>\n${Visualization.renderVisualizationsTemplate}`,```
If I get some positive feedback on this and this behaviour makes sense, I'll create the PR.

But... As I mentioned (and I'm looking for feedback here), there may be situations when both dashboard and widget level filters are needed. For this case I suggest adding an extra setting along with "Use Dashboard Level Filters", such as "Show Widget Level Filters", for example.

PR #3760, opened 2019-05-04
Feature: Add option to hide widget level filters

yoavbls:

- [x] Feature

## Description
When dashboard filters are enabled all the widgets still keep their filters and make 
a lot of visual noise and redundancy.
In this PR I Added another flag that shows only when dashboard filters are checked
and let you choose to hide the filters in the widget level.

#3074

Before:
image

After:
image


rauchy:
Thanks @yoavbls, LGTM.

I'm wondering if two checkboxes are even required - maybe selecting dashboard level filters should automatically hide widget filters? WDYT @arikfr?

arikfr:
> I'm wondering if two checkboxes are even required - maybe selecting dashboard level filters should automatically hide widget filters? WDYT @arikfr?

This makes sense. I just wonder if it's worth having both options for the sake of backward compatibility? 🤔

rauchy:
> I just wonder if it's worth having both options for the sake of backward compatibility? 🤔

I think we should bite the bullet and allow toggling between dashboard-level and widget-level. Having both displayed feels confusing to me.

arikfr:
Hi,

Thanks.

PR #3773, opened 2019-05-06
Show non relative timestamp when printing an embed

arikfr:

- [x] Bug Fix

## Description

Show non relative timestamp when printing an embed. (Also used for the Slack snapshots)

This brings back a functionality accidentally removed in #3752.

### DIFF
--- a/client/app/components/parameters.html
+++ b/client/app/components/parameters.html
@@ -1,16 +1,31 @@
-<div class="parameter-container form-inline bg-white"
+<div
+  class="parameter-container form-inline bg-white"
   ng-if="parameters | notEmpty"
   ui-sortable="{ 'ui-floating': true, 'disabled': !editable }"
   ng-model="parameters"
 >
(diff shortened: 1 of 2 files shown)


client/app/components/parameters.html
```diff
@@ -1,16 +1,31 @@
...
+      ng-click="showParameterSettings(param, $index)"
+      data-test="ParameterSettings-{{ param.name }}"
+    >
       <i class="zmdi zmdi-settings"></i>
     </button>
     <parameter-value-input param="param"></parameter-value-input>
   </div>
-  <button class="m-t-20 btn btn-primary" ng-if="onRefresh" ng-click="onRefresh()" title="Refresh Dataset">
+  <button
+    class="m-t-20 btn btn-primary hidden-print"
```
arikfr: Actual change: add `hidden-print` here.

The rest is Prettier...

PR #3777, opened 2019-05-07
Avoid error with duplicate log lines

otsaloma:

- [x] Bug Fix

## Description

Displaying log output from a query runner fails on an error if the output contains duplicate lines and as a result of that error no log output is shown in such cases. This is due to how `ng-repeat` works. Fixed in this PR as shown in the Angular documentation.

https://docs.angularjs.org/error/ngRepeat/dupes


arikfr:
Thanks 👍

PR #3778, opened 2019-05-07
Visualization: details view

arikfr:

- [x] Feature

## Description

Details visualization -- shows each row as a table (similar to the expanded view in psql):

image

Paginator only shows if you return more than one row.

Ideally it would use Table's column renderers (or maybe just be a mode of the table visualization), but this can wait until the Table visualization is rewritten with React.

Todo:

- [x] Add PropTypes.
- [x] Fix issue with rendering when there is no column information.
 
https://discuss.redash.io/t/single-row-view/311/6?u=arikfr

### DIFF
--- a/client/app/visualizations/details/DetailsRenderer.jsx
+++ b/client/app/visualizations/details/DetailsRenderer.jsx
@@ -1,0 +1,37 @@
+import React, { useState } from 'react';
+import map from 'lodash/map';
+import Pagination from 'antd/lib/pagination';
+
+import './details.less';
+
+export default function DetailsRenderer({ data }) {
+  const [page, setPage] = useState(0);
+
+  if (!data || !data.rows) {
+    return null;
+  }
+
+  // We use columsn to maintain order of columns in the view.
+  const columns = data.columns.map(column => column.name);
+  const row = data.rows[page];
+
+  return (
+    <div>
+      <table className="table table-bordered details-viz">
+        <tbody>
+          {map(columns, key => (
+            <tr key={key}>
+              <th>{key}</th>
+              <td>{'' + row[key]}</td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+      {data.rows.length > 1 && (
+        <div className="paginator-container">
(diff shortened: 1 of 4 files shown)


client/app/visualizations/details/DetailsRenderer.jsx
```diff
@@ -0,0 +1,37 @@
...
+    return null;
+  }
+
+  // We use columsn to maintain order of columns in the view.
+  const columns = data.columns.map(column => column.name);
+  const row = data.rows[page];
+
+  return (
+    <div>
+      <table className="table table-bordered details-viz">
```
ranbena: @arikfr I assume you're not using Ant Table cause it has no "vertical header" feature.
I think I can implement this feature with CSS only - [`write-mode: lr;`](https://caniuse.com/#feat=css-writing-mode). Lmk if you see value in this attempt.

ranbena:
@arikfr perhaps consider allowing to choose different viz themes:
1. Table
Screen Shot 2019-05-09 at 16 18 43

2. List
Screen Shot 2019-05-09 at 16 23 23

3. Card grid
Screen Shot 2019-05-09 at 16 25 48

etc.

client/app/visualizations/details/DetailsRenderer.jsx
arikfr: Well, it was this and the fact it felt too much for this use case. But if we can make it work with just CSS, it can go along with my plan of merging this into the table visualization later on.

Not sure it's worth the effort at the moment though.

arikfr:
I'm not sure when the list one will be useful, but the card grid is interesting when you want to show multiple counters without having the create multiple ones 🤔

client/app/visualizations/details/DetailsRenderer.jsx
ranbena: Then let's leave it be. I never used this css method before and it might have unforeseen implications. The best way would be to PR Ant with `<Table orientation="vertical" />`.

arikfr:
All green ✅

arikfr:
Our first 100% React viz :)

Issue #3734, opened 2019-04-24
Tag editor should not accept empty strings as tags


arikfr:
Related to #3733.

1. We should change the editor to filter such values.
2. We should change the backend to filter them as well.

PR #3780, opened 2019-05-08
tag editor to filter such values

taminif:
- [x] Bug Fix

## Description
Tag editor to filter such values on display availableTags and change text.
I confirmed that not saved empty values. But I don't fix display empty tag at changing time yet.

#3734
`1. We should change the editor to filter such values.`


ranbena:
@taminif how about lodash's `compact` instead? It's more performant and less code :P

taminif:
Thanks for the reviews @ranbena . Updated code. Please review once again!

ranbena:
10x @taminif

Issue #3009, opened 2018-10-28
Add more predefined options to Date and Date Range parameters


arikfr:
The [Date Picker](https://ant.design/components/date-picker/) we use has support for Presetted Ranges, so we can add some useful short hands like: last month, last week, last 7 days, last 30 days, etc. 

- [ ] Add support for picking presetted ranges.
- [ ] Make it possible to pick such range as the default value of this parameter (a more general case of the current "Use today as default value" option)

ssozonoff:
Surprised their is not more demand for this. To me its a fundamental requirement of any Dashboarding solution.

arikfr:
The first part (more predefined options) is easy. The tricky part is how to persist those selections as defaults. 

One idea I had is to save them as values for start/end, i.e. when you pick "Last 7 Days" the value we pass for param.start/param.end will be something like `last_7_days`, so we know to convert it to the appropriate timestamps.

A more generic solution will be to pass datetime deltas as values, i.e. when you pick "Last 7 Days" what we pass for start will be "-7 days" and for end "-0 days".

gabrieldutra:
**UX/UI**
My ideas so far for the second part of it:
- Add a "Set dynamic" toggle on the parameter. This should make a good discoverability for it, I considered just modify the Parameter Setting dialog, but this would limit the scope to "Editing a query". An example ~probably with not the best icon~ that worked as a start:
![date-range-1](https://user-images.githubusercontent.com/3356951/58911208-bf844a80-86ed-11e9-9701-2da3cb5a88ed.png)
- When clicking on it a dialog will open showing usual options and a Custom (with 'from' and 'to'):
![daterange-dialog](https://user-images.githubusercontent.com/3356951/58912042-906ed880-86ef-11e9-8815-80c25d3b6779.png)
- When in Dynamic mode it will no longer appear as a DatePicker, but it will show a fixed message containing the selected date range and an option to make it static again.

This is sort of the draft of what I thought for UX/UI, @ranbena can give magic ideas here :)

**Implementantion**
For implementation, I loved the generic idea of delta values and it's here that the solution for #3053 may be similar. Should the parameter "translation" be backend-wise?

I think it's sort of a workaround doing in the frontend as deep down it loads a static value. Also there's the case for Scheduled Queries.

From the backend I guess my concerns are that I haven't seen any of these "translations" in there and depending on how repeated query execution is identified for repeated parameters the delta values could cause an issue.

arikfr:
[...]

PR #3781, opened 2019-05-08
Add presetted ranges to DateRange and DateTimeRange inputs

gabrieldutra:

- [x] Feature

## Description
This is related to the first part of #3009, it adds a few presetted options (from [Antd presetted ranges](https://ant.design/components/date-picker/#components-date-picker-demo-presetted-ranges)) to DateRange and DateTimeRange inputs.

I tried to think the ranges that could be useful (and the ones suggested on the issue :)), thought add suggestions if you have any :slightly_smiling_face:.

#3009 

date-ranges-presetted-ranges

date-ranges-predefined-ranges


client/app/components/DateRangeInput.jsx
```diff
@@ -1,23 +1,43 @@
...
 import PropTypes from 'prop-types';
+import moment from 'moment';
 import { react2angular } from 'react2angular';
 import DatePicker from 'antd/lib/date-picker';
 import { clientConfig } from '@/services/auth';
 import { Moment } from '@/components/proptypes';
 
 const { RangePicker } = DatePicker;
 
+export const PRESETTED_RANGES = {
```
gabrieldutra: It seemed interesting to export here since Date Time Ranges will basically have the same Presetted ranges + its specific ones (with time included in the ranges).

ranbena:
@gabrieldutra this is really great!

A few thoughts:
1. In order to minimize preset UI bloat, perhaps limit to one row along with a "more..." button that opens up all suggestions. Not a must, not so simple to implement.
2. "This/Last week" mandates that the calendar be set to the correct "workweek" (starting Mon or Sun). Is that automatic? Is it configurable?
3. "Today" might be confused for a "dynamic" setting (day of query run) instead of the actual "static" (day of preset click). Same goes for "Last X" presets. Wdyt?

gabrieldutra:
> In order to minimize preset UI bloat, perhaps limit to one row along with a "more..." button that opens up all suggestions. Not a must, not so simple to implement.

I tried to keep it with less than 2 rows for this reason. Unfortunately there's no easy way to implement the "more..." :confused:. It would require to use Antd's `renderExtraFooter` and basically do again the `ranges` feature... Let's just keep it without many options for now (It's possible that this feature come up from Ant meanwhile).

> "This/Last week" mandates that the calendar be set to the correct "workweek" (starting Mon or Sun). Is that automatic? Is it configurable?

According to moment's docs: "As of version 2.1.0, moment#startOf('week') uses the locale aware week start day.". The default seems to be `en` (which leads to Sun-Sat), but this can be changed using `moment.locale`. Wdyt?

> "Today" might be confused for a "dynamic" setting (day of query run) instead of the actual "static" (day of preset click). Same goes for "Last X" presets. Wdyt?

There's indeed a chance it is confusing at first, but then it shows the specific days in the calendar and when you set to it, specific dates will show up. Dynamic settings are the second part of #3009, which I do have an idea to have a "Dynamic toggle" (should discuss this later in the issue).

The general idea here is to first give value without much, if this turns out to be confusing we can just remove those and let them only in the "Dynamic" options.

ranbena:
1. 👍 
2. Sounds good. Maybe you can [adjust the calendar itself](https://ant.design/components/date-picker/#Localization) while you're at it? 😬(Doesn't look so easy 😔)
Btw, why is `en-US` set to Sun-Sat when should actually be Mon-Fri?
3. 👍

client/app/components/DateRangeInput.jsx
ranbena: With this implementation the presets could go "stale" since they're calculated on page load (right?). Ideally should be calculated on preset click / calendar open. But it's unlikely so let's just keep it in mind.

gabrieldutra: I can use the function notation instead to avoid this :)

gabrieldutra:
> Btw, why is en-US set to Sun-Sat when should actually be Mon-Fri?

It's not considering work weekdays, but the whole week ^^ (I think this is what you expect when considering "This/Last week")

arikfr:
Actually in this context, I think people expect a work week.

Maybe we will add a setting for "week starts on"? Taking this from local
computer might not always work: think of our team, for me and Ran it starts
on Sunday but for you it starts on Monday :)

gabrieldutra:
> Actually in this context, I think people expect a work week.

Can this be a point of conflict too :thinking: ? I've reached some friends that use Redash and this can be more confusing than I initially thought...

- Future dates may not be expected (so I just removed the 'This' ones)
- The people from Brazil expected "Last week" to be the whole week, but had the following points of conflict: - It could be interpreted the same as 'Last 7 days'; - One actually preferred Mon-Sun over Sun-Mon (even though the second was expected);

A conclusion is that this will depend a lot from each business model, but for this part this is not that relevant as this will be only a shorthand and people can learn from the dates that appear in there.

So for this part the options I see are:
1. Remove 'Last week' as it's the most conflicting one;
2. Let it there with a defined value from us (which should fit most people);
3. Add the setting option, so users can define what is their "week";

To this part I'm supporting 1 or 2. In case the setting option becomes relevant when doing Part 2 of #3009, we can extend it here as well.

arikfr:
anything stopping us from merging this?

ranbena:
> anything stopping us from merging this?

I suggest to actually bundle these presets along with #3009 instead.

gabrieldutra:
Closing in favor of #3904

Issue #3212, opened 2018-12-21
Query source can be displayed from "View Query" on the dashboard


Blufe:
### Issue Summary

Even without the view_source privilege, a query source can be displayed from "View Query" on the dashboard.

### Expected Behavior

The "View Query" link on the dashboard should generate a URL to the screen of the query, not the query source.

### Technical details:

* Redash Version: 5.0.1.b4850

PR #3782, opened 2019-05-09
enable view_source permission

muddydixon:
## Permission for viewing query source 
- [x] Feature

## Description
Now users who do not have `view_source` permission can view query source, so this permission does not work for the intent that is.
In this commit, the users are not able to view query sources.

#3212

image
image

### DIFF
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -372,11 +372,12 @@ class QueryResource(BaseResource):
             'object_type': 'query',
         })
 
-        if not self.current_user.has_permissions('view_source'):
-            result['query'] = '\n'.join(
-                ['-- Query Source requires \'view_source\' permission.'] \
-                + list(map(lambda p: '-- {{{{{}}}}}'.format(p['name']), result['options']['parameters']))
-            )
+        if not self.current_user.has_permission('view_source'):
+            result['query'] = '-- Query Source requires \'view_source\' permission.'
+            if 'parameters' in result['options']:
+                result['query'] += '\n' + '\n'.join(
+                    [] + list(map(lambda p: '-- {{{{{}}}}}'.format(p['name']), result['options']['parameters']))
+                )
         return result
 
     # TODO: move to resource of its own? (POST /queries/{id}/archive)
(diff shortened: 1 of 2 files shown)


muddydixon:
@arikfr  Thank you for your comment.

I changed to show `-- Query Source requires 'view_source' permission.`, If user who has not `view_source` permission accesses query source.

This commit enables to display embed visualization and to hide query source.

muddydixon:
@arikfr Hi

I fixed that the chart is not displayed in embedded environment.
How about this

redash/handlers/queries.py
```diff
@@ -372,6 +372,12 @@ def get(self, query_id):
             'object_type': 'query',
         })
 
+        if not self.current_user.has_permission('view_source'):
+            result['query'] = '-- Query Source requires \'view_source\' permission.'
```
arikfr: Why not just remove this field?

redash/handlers/queries.py
```diff
@@ -372,6 +372,12 @@ def get(self, query_id):
...
+            if 'parameters' in result['options']:
+                result['query'] += '\n' + '\n'.join(
+                    [] + list(map(lambda p: '-- {{{{{}}}}}'.format(p['name']), result['options']['parameters']))
+                )
```
arikfr: What is this?

redash/handlers/queries.py
arikfr: Also this logic belongs in the `QuerySerializer`.

redash/handlers/query_results.py
```diff
@@ -171,11 +171,15 @@ def post(self, query_id):
         max_age = int(max_age)
 
         query = get_object_or_404(models.Query.get_by_id_and_org, query_id, self.current_org)
-
         allow_executing_with_view_only_permissions = query.parameterized.is_safe
 
         if has_access(query, self.current_user, allow_executing_with_view_only_permissions):
-            return run_query(query.parameterized, parameter_values, query.data_source, query_id, max_age)
+            query_result = run_query(query.parameterized, parameter_values, query.data_source, query_id, max_age)
+            if isinstance(query_result, tuple) and query_result[1] != 200:
```
arikfr: As this is not the only place we return a `QueryResult` instance, this behavior needs to be in a serializer. I don't think we created one for a `QueryResult` yet and we still use a `to_dict` method, but we can as part of this change.

arikfr:
Seems like this is abandoned. I'm closing this now, but you're welcome to re-open after addressing the comments. Thanks!

PR #3783, opened 2019-05-09
Migrate Embed Query Dialog to React

gabrieldutra:

- [x] Refactor

## Description
It was a pretty simple migration, so it seemed a good opportunity to explore Antd's [Typography](https://ant.design/components/typography/).

To make Typography's `<code>` better fit the current in Redash we can remove the gray background.

embed-query-dialog

**Alert when embedding query containing text parameters**
embed-query-alert (1)


client/app/components/queries/EmbedQueryDialog.jsx
```diff
@@ -0,0 +1,45 @@
...
+import { clientConfig } from '@/services/auth';
+
+const { Text } = Typography;
+
+function EmbedQueryDialog({ dialog, query, visualization }) {
+  const embedUrl = `${clientConfig.basePath}embed/query/${query.id}/visualization/${
+    visualization.id}?api_key=${query.api_key}&${query.getParameters().toUrlParams()}`;
+
+  let snapshotUrl;
+  if (window.snapshotUrlBuilder) {
```
gabrieldutra: I'm not sure what this is for, but I just followed what was in the Angular version.

client/app/components/queries/EmbedQueryDialog.jsx
```diff
@@ -0,0 +1,45 @@
...
+    snapshotUrl = window.snapshotUrlBuilder(query, visualization);
+  }
+
+  return (
+    <Modal {...dialog.props} title="Embed Code" footer={(<Button onClick={dialog.dismiss}>Close</Button>)}>
+      <h5 className="m-t-0">IFrame Embed</h5>
+      <div className="m-b-10">
+        <Text data-test="EmbedIframe" copyable code>
+          {`<iframe src="${embedUrl}" width="720" height="391"></iframe>`}
```
kravets-levko: @gabrieldutra I think it would be better to replace this with `<textarea>` - so user will be able to copy it manually and it will not require ellipsis (I personally use this dialog just to get embed URL when I need to test something, so it's a case for having all the code available).

gabrieldutra: I've considered using a `TextArea` too, but I thought the copy button could be a good feature. I'll check it with the `TextArea` :+1:

ranbena:
Wdyt of the following design?

1. Code block.
2. Copy button appears in corner on hover.
3. Allow setting iframe width / height.

Screen Shot 2019-05-09 at 10 58 45

gabrieldutra:
> Wdyt of the following design?

Seems nice, perhaps without the Checkbox (not really necessary to make width and height disabled since it doesn't really change the behavior) or if you think it has a reason to be there, use a switch instead?

client/app/components/queries/EmbedQueryDialog.jsx
arikfr: I agree with @kravets-levko that we shouldn't use an ellipsis here, but why a textarea?

ranbena:
> Seems nice, perhaps without the Checkbox (not really necessary to make width and height disabled since it doesn't really change the behavior) or if you think it has a reason to be there, use a switch instead?

I agree, it's not necessary - it serves as a hint for "normal" users that the current setting is default and doesn't require adjustment - ignorable if you will. Not a must.

(A switch I think would be too distracting tho)

arikfr:
My suggestions to address some common use cases with this dialog:

1. Show a copy-able URL for the visualization (the one we use for the iframe). 99% of the times I used this dialog, I just needed to copy this URL :) We can call this "Public URL". 
2. As the next section, show the iframe.
3. If the query has unsafe parameters (query.is_safe = true), then show a message saying you can't currently embed queries with parameters of type text. This is instead the embed URLs.

gabrieldutra:
I've updated this based on your comments :), just need some final UI adjustments. @ranbena welcome to give your magic look to it already :laughing:

react-embed-query-dialog-code-block

My thoughts: now that it has two sections the default look for Form Labels (Width and Height) may be confusing, so it's better with a smaller `font-weight` (I think I'll add an `<hr />` to separate sections too). Regarding the checkbox, after playing with the dialog I think it's better with it.

Those are my ideas, I'll update this tomorrow, but you can already give your comments.

arikfr:
Last version looks good 👌 One thing I would change: have the copy button always visible instead of only on hover.

Also, can you add a screenshot of the dialog when you try to embed with unsafe params?

ranbena:
Screen Shot 2019-05-12 at 5 16 42

ranbena:
I think it looks good as well 👌👌

Here's an updated version:
Screen Shot 2019-05-12 at 5 36 49

It's got:
* persistent copy buttons - give the codebox `padding-left: 27px` so button won't overlap code.
* checkbox - has the effect of indenting the row which makes it clearer that it's iframe related
* normal weighted labels
* small input-numbers + width 72px

client/app/components/CodeBlock.less
```diff
@@ -0,0 +1,23 @@
+@import '~antd/lib/button/style/index';
+
+.code-block {
+  background: rgba(0, 0, 0, 0.06);
+  border: 1px solid rgba(0, 0, 0, 0.06);
+  border-radius: 3px;
```
gabrieldutra:
```suggestion
  border-radius: 2px;
```

ranbena:
> Alternatively with `<hr>` (it's sort of cool to separate stuff)

With the right spacing it could be nice. `<Divider dashed />` is also an option.

I personally though would instead go with `m-b-30` for the top code block. Looks nice and clean.
```html
<div class="m-b-30">
  <div class="code-block">
    …
  </div>
</div>
```


> I've updated the dialog's title: "Embed Code" -> "Embed Query", looks more related to what it is

👍

arikfr:
👌

PR #3788, opened 2019-05-12
Loading indicator till app inits

ranbena:
- [x] Feature

## Description
Before app loads there might be a substantial wait for `<app-view>` to populate. Instead of looking at a blank screen, here's a nice hovering logo.

This is a css+html only change 😛
Even determining indicator hide/show is done with css, according to `<app-view>` being empty/full.

### Some notes
I created a new beautiful sleek animation with the bars but for some reason it's animation got stuck on the main thread. Don't know how since I used only gpu accelerated transition props. Anywho, this is good enough.

I only wish I didn't have to stick the entire html elements into `index.html` and it's twin brother.
Lmk if there's a way around that (independent of Angular/React).

## Mobile & Desktop Screenshots/Recordings

Before:

After:

### DIFF
--- a/client/app/assets/less/redash/css-logo.less
+++ b/client/app/assets/less/redash/css-logo.less
@@ -1,8 +1,10 @@
 // based on https://github.com/outbrain/tech-companies-logos-in-css/pull/28
+
+@primary: #ff7964;
+@shadow: #ef6c58;
+@bar: white;
+
 #css-logo {
-  --primary: #ff7964;
-  --shadow: #ef6c58;
-  --bar: white;
   width: 100px;
   height: 100px;
   position: relative;
@@ -10,7 +12,7 @@
   #circle {
     width: 79px;
     height: 79px;
-    background-color: var(--shadow);
+    background-color: @shadow;
     border-radius: 50%;
     margin: auto;
     overflow: hidden;
@@ -19,7 +21,7 @@
     & > div {
       width: 79px;
       height: 73px;
-      background-color: var(--primary);
+      background-color: @primary;
       border-radius: 50%;
       position: absolute;
       top: 0;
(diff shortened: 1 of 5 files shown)


client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -936,4 +936,56 @@ text.slicetext {
...
+      opacity: 0.2;
+    }
+  }
+}
+
+// hide indicator when app-view has content
+app-view:not(:empty) ~ .loading-indicator {
+  opacity: 0;
+  transform: scale(0.9);
+  pointer-events: none;
```
ranbena: Unnecessary but it helps me sleep at night.
Better make sure transparent elements are absolutely non-interactable.

arikfr:
Is the E2E fail due to the CSS changes?

ranbena:
> Is the E2E fail due to the CSS changes?

Nope. I'm looking into it.

client/app/assets/less/redash/css-logo.less
```diff
@@ -0,0 +1,86 @@
+// based on https://github.com/outbrain/tech-companies-logos-in-css/pull/28
+#css-logo {
+  --primary: #ff7964;
+  --shadow: #ef6c58;
```
kravets-levko: I think it's better to use LESS variables (just for 100% compatibility)

client/app/assets/less/redash/css-logo.less
```diff
@@ -0,0 +1,86 @@
...
+    transform-origin: left;
+  }
+  
+  #point > div {
+    bottom: -12px;
+    border-color: var(--primary);
+    transform: scaleX(1.04);
+    left: -17px;
+  }
+}
```
kravets-levko: Please add empty line here

client/app/index.html
```diff
@@ -26,6 +26,23 @@
...
+            <div></div>
+          </div>
+          <div id="bars">
+            <div class="bar"></div>
+            <div class="bar"></div>
+            <div class="bar"></div>
+            <div class="bar"></div>
+          </div>
+        </div>
+        <shadow></shadow>
```
kravets-levko: Not sure if it was intentional: `<shadow>` tag is an obsolete tag for shadow DOM. Probably here it should be just `<div>`. Also, it's not recommended to use single-word names for _custom_ components (because such names may be used for new HTML tags in future) - custom components should have names with few words separated by `-`

PR #3793, opened 2019-05-13
Get start development without npm

shunsuke-excite:

- [x] Other

## Description
*  docker-compose can `up`  without  installing `npm`  
* `Postgres`  can save persistently


docker-compose.yml
```diff
@@ -13,6 +13,7 @@ services:
       - "5678:5678"
     volumes:
       - ".:/app"
+      - "/app/client/dist"
```
arikfr: Won't this break using npm on the host?

shunsuke-excite: With using `npm` for development on host,  `npm start` on host for development.

arikfr: I meant, wouldn't setting up the volume this way make the changes on the host not seen in the container?

shunsuke-excite: I see. 
I'll update the changes made on the host can appear  in the container.

shunsuke-excite: I'm sorry, I misunderstood.
This Pull Request close.

Issue #3753, opened 2019-05-01
Visualization rendering getting cut in editor on first load


arikfr:
1. Open this URL: https://redash-preview.netlify.com/queries/36#62
2. Click on "Edit Visualization".

You will see that the visualization in the editor is cut on the left side:

![image](https://user-images.githubusercontent.com/71468/57020673-b1489780-6c32-11e9-820b-baf0510d77a0.png)

If you switch to a different visualization tab and then back to this one, it gets resolved.

kravets-levko:
Thanks! I think it's because of dialog animation :thinking:

arikfr:
@kravets-levko let's try to solve this this week.

PR #3794, opened 2019-05-13
Visualization rendering getting cut in editor on first load

kravets-levko:

- [x] Bug Fix

## Description

Ant Modal has open/close animation (Bootstrap modal didn't), which affects `resize-event` directive: it uses `offsetWidth`/`offsetHeight` properties and does not take into account any sort of CSS transformations. Solution: use `getBoundingClientRect` which returns real element dimensions.

Fixes getredash/redash#3753

### DIFF
--- a/client/app/directives/resize-event.js
+++ b/client/app/directives/resize-event.js
@@ -4,20 +4,22 @@ const items = new Map();
 
 function checkItems() {
   items.forEach((item, node) => {
-    const offsetWidth = node.offsetWidth;
-    const offsetHeight = node.offsetHeight;
+    const bounds = node.getBoundingClientRect();
+    // convert to int (because these numbers needed for comparisons), but preserve 1 decimal point
+    const width = Math.round(bounds.width * 10);
+    const height = Math.round(bounds.height * 10);
 
     if (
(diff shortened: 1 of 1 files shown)


client/app/directives/resize-event.js
```diff
@@ -4,20 +4,22 @@ const items = new Map();
 
 function checkItems() {
   items.forEach((item, node) => {
-    const offsetWidth = node.offsetWidth;
-    const offsetHeight = node.offsetHeight;
+    const bounds = node.getBoundingClientRect();
+    // convert to int (because these numbers needed for comparisons), but preserve 1 decimal point
+    const width = Math.round(bounds.width * 10);
+    const height = Math.round(bounds.height * 10);
```
ranbena: This definitely works but kinda confusing cause it's not storing the actual dimensions.
Perhaps `bounds.width.toFixed(1)` can work?

kravets-levko: Well, it's actually the same - just I store `int`, but this will store `string`. But in fact we just need some normalized values that we can compare. Since your variant will do exactly the same as mine, and we'll remove this directive after React migration - I prefer to keep this as is.

Issue #3797, opened 2019-05-14
Send an e-mail report for failed scheduled queries


rauchy:
Currently, when scheduled queries fail (due to error or timeout), users don't receive any notification about it. We should send out an aggregated e-mail to the organization admin with information on failures, which should include for each failure:
1. Which query failed
2. When it failed
3. What is the failure

Notes:

- This feature should be turned off by default and can be turned on with a setting, in order to avoid surprises when upgrading.
- We should default to an hourly e-mail, and the frequency should be configurable.
- If a query failed over `threshold` times, don't report it.
- If a query failed over 75% of `threshold` times, mention in the e-mail that this query will soon be unreported.
- `threshold` should be configurable.

PR #3798, opened 2019-05-14
Failed Scheduled Queries Report

rauchy:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
This PR sends out an aggregated e-mail report for failing scheduled queries as suggested in #3797. 

#3797 

image


ranbena:
@rauchy if you wish to collaborate on the layout and design of the report, plz open a discuss topic.

rauchy:
Updated code and PR screenshot to @ranbena's design ❤️

redash/tasks/failure_report.py
```diff
@@ -14,15 +14,21 @@ def send_aggregated_errors(email_address):
...
-    html = '<hr>'.join(["""
-            <p>
-              <h3><a href="{base_url}/queries/{id}">{name}</a></h3>
-              Last failed at: {failed_at} (failed {failure_count} times since last report)<br>
-              Error message: <pre style="border: 1px solid black; padding: 10px; background: #fbfbfb">
-                  {failure_reason}
-              </pre>
-              <b>{comment}</b>
-            </p>""".format(
+    failures_html = ''.join(["""
```
rauchy: It's definitely time to use a templating engine here.

rauchy: a2072b6

redash/templates/emails/failures.html
```diff
@@ -0,0 +1,45 @@
...
+                            <br />
+                            <small style="color: #8c8c8c;font-size: 13px;">&nbsp; + {{failure.failure_count - 1}} more failures since last report</small>
+                        {% endif %}
+		    		</p>
+		    		<h3 style="font-size: 14px; font-weight:normal; margin-bottom: 6px">Exception</h3>
+		    		<div style="background: #ececec;padding: 8px;border-radius: 3px;font-family: monospace; overflow: scroll;">{{failure.failure_reason}}</div>
+                </div>
+
+                {% if failure.comment %}
+                    <div style="margin-top: 25px;font-size: 13px;border-top: 1px solid #ececec;padding-top: 19px;">{{failure.comment}}</div>
```
ranbena:
```suggestion
                    <div style="margin-top: 25px;font-size: 13px;border-top: 1px solid #ececec;padding-top: 19px;">NOTICE: {{failure.comment}}</div>
```

This is from my original design, it's meant to draw attention from the multitude of information but feel free to reject this if you think it's stupid.

redash/templates/emails/failures.html
```diff
@@ -0,0 +1,45 @@
...
+		    	<div>
+		    		<p>
+		    			Last failed: {{failure.failed_at}}
+                        {% if failure.failure_count > 1 %}
+                            <br />
+                            <small style="color: #8c8c8c;font-size: 13px;">&nbsp; + {{failure.failure_count - 1}} more failures since last report</small>
+                        {% endif %}
+		    		</p>
+		    		<h3 style="font-size: 14px; font-weight:normal; margin-bottom: 6px">Exception</h3>
+		    		<div style="background: #ececec;padding: 8px;border-radius: 3px;font-family: monospace; overflow: scroll;">{{failure.failure_reason}}</div>
```
ranbena:
```suggestion
		    		<div style="background: #ececec;padding: 8px;border-radius: 3px;font-family: monospace; word-break: break-all;">{{failure.failure_reason}}</div>
```

This makes sure any long word would line break.

redash/templates/emails/failures.html
```diff
@@ -0,0 +1,45 @@
...
+        </div>
+        <h1 style="font-size: 14px; font-weight: normal;">Redash failed to run the following queries:</h1>
+
+        {% for failure in failures %}
+		    <div style="background: white;padding: 5px 20px 20px 20px; border-radius: 3px; box-shadow: 0 4px 9px -3px rgba(102,136,153,.15); margin-top: 20px">
+		    	<h2 style="font-size: 16px; position: relative">{{failure.name}} <a href="{{failure.base_url}}/queries/{{failure.id}}" style="background-color: #ff7964;font-size: 13px;color:  white;text-decoration: none;padding: 0 5px;display: inline-block;border-radius: 3px;font-weight: normal;position: absolute;right: 0;top:-1px">Go
```
ranbena:
```suggestion
		    	<h2 style="font-size: 16px; position: relative; padding-right: 90px;">{{failure.name}} <a href="{{failure.base_url}}/queries/{{failure.id}}" style="background-color: #ff7964;font-size: 13px;color:  white;text-decoration: none;padding: 0 5px;display: inline-block;border-radius: 3px;font-weight: normal;position: absolute;right: 0;top:-1px">Go
```
Whoops. Preventing the button from obscuring the title.

redash/tasks/failure_report.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def notify_of_failure(message, query):
+    if not settings.SEND_EMAIL_ON_FAILED_SCHEDULED_QUERIES:
+        return
+
+    if query.schedule_failures < settings.MAX_FAILURE_REPORTS_PER_QUERY:
+        key = 'aggregated_failures:{}'.format(query.user.email)
+        reporting_will_soon_stop = query.schedule_failures > settings.MAX_FAILURE_REPORTS_PER_QUERY * 0.75
+        comment = """This query has failed a total of {failure_count} times.
```
rauchy:
```suggestion
        comment = """NOTICE: This query has failed a total of {failure_count} times.
```

redash/templates/emails/failures.html
rauchy: I don't mind the rude, all-caps "NOTICE", but I just prefer to build the string [at one place](https://github.com/getredash/redash/pull/3798#pullrequestreview-240426364).

ranbena:
Regarding the logo:
1. Transparent bg is better. Np.
2. Gmail strips the embedded image data so need to have a url instead. @arikfr is there a static asset server I can locate this file in?

arikfr:
Static asset: 🤔I wonder if we should host one for everyone or try to generate a URL that points at current Redash?

Arik

ranbena:
> Static asset: 🤔I wonder if we should host one for everyone or try to generate a URL that points at current Redash?

I think one for everyone. If old emails break it has no significance.

@arikfr I placed the up-to-date image here https://imgur.com/a/WtbIAqZ.
We can alternatively create a GH project with the template and image in it and serve with github-pages.

arikfr:
@ranbena how about we go with no logo on this email?

susodapop:
I added a note about these pending change to a PR on the website repo (4a38685461). I'd like to add more details and a screenshot with the finalized notification email format. The initial issue mentioned that these notifications would have a control in the settings menu. Is that still expected before this merges? Or will the failure notification emails be sent by default?

rauchy:
@susodapop I'll comment in https://github.com/getredash/website/pull/245 with details. Regarding setting this on or off - the issue mentions there should be a setting, as in, an entry in the settings file, not the settings menu. That's not a bad idea though. As of now, it is turned off by default to avoid surprises when upgrading. @arikfr do you think this should go to the org settings menu?

susodapop:
> Regarding a screenshot - is it really necessary?

IMO these emails are part of the UI and deserve a screenshot. When we add the in-app notification we'll **certainly** screencap it too (maybe we can remove the screencap of the email at that time?). In the meanwhile, these emails are the predecessor to that behavior, so a cap seems appropriate.

rauchy:
>  these emails are the predecessor to that behavior, so a cap seems appropriate.

In that case we'll have to include an error that makes sense, and maybe the comment that indicates that reporting will soon reach the reported limit.

rauchy:
Well, there's this now:

image

@susodapop comments on wording?

rauchy:
@arikfr any plans to merge this?

redash/settings/__init__.py
```diff
@@ -227,6 +227,9 @@ def email_server_is_configured():
 
 HOST = os.environ.get('REDASH_HOST', '')
 
+SEND_FAILURE_EMAIL_INTERVAL = int(os.environ.get('REDASH_SEND_FAILURE_EMAIL_INTERVAL', 3600))
+MAX_FAILURE_REPORTS_PER_QUERY = int(os.environ.get('REDASH_MAX_FAILURE_REPORTS_PER_QUERY', 100))
```
arikfr: Was there a reason that we set it so high? I would expect 10 or so to be enough, no?

redash/tasks/failure_report.py
```diff
@@ -0,0 +1,65 @@
...
+from flask import render_template
+from redash.tasks.general import send_mail
+from redash.worker import celery
+from redash import redis_connection, settings
+from redash.utils import json_dumps, json_loads, base_url
+
+
+@celery.task(name="redash.tasks.send_aggregated_errors")
+def send_aggregated_errors(email_address):
+    key = 'aggregated_failures:{}'.format(email_address)
```
arikfr: Using email address as a key is a really bad idea, considering that the same email address can be in multiple organizations. Why not use the user id?

redash/tasks/failure_report.py
```diff
@@ -0,0 +1,65 @@
...
+        text = render_template('emails/failures.txt', **context)
+        subject = "Redash failed to execute {} of your queries".format(len(unique_errors.keys()))
+        send_mail.delay([email_address], subject, html, text)
+
+    redis_connection.delete(key)
+
+
+def notify_of_failure(message, query):
+    if not settings.organization.SEND_EMAIL_ON_FAILED_SCHEDULED_QUERIES:
+        return
```
arikfr: Another check to perform is if the query owner is disabled. If they are, then no need to send anything.

redash/tasks/failure_report.py
```diff
@@ -0,0 +1,65 @@
...
+                    'failed_at': v.get('failed_at'),
+                    'failure_reason': v.get('message'),
+                    'failure_count': occurrences[k],
+                    'comment': v.get('comment')
+            } for k, v in unique_errors.iteritems()]
+        }
+
+        html = render_template('emails/failures.html', **context)
+        text = render_template('emails/failures.txt', **context)
+        subject = "Redash failed to execute {} of your queries".format(len(unique_errors.keys()))
```
arikfr:
```suggestion
        subject = "Redash failed to execute {} of your scheduled queries".format(len(unique_errors.keys()))
```

redash/tasks/failure_report.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def notify_of_failure(message, query):
+    if not settings.organization.SEND_EMAIL_ON_FAILED_SCHEDULED_QUERIES:
+        return
+
+    if query.schedule_failures < settings.MAX_FAILURE_REPORTS_PER_QUERY:
+        key = 'aggregated_failures:{}'.format(query.user.email)
+        reporting_will_soon_stop = query.schedule_failures > settings.MAX_FAILURE_REPORTS_PER_QUERY * 0.75
+        comment = """NOTICE: This query has failed a total of {failure_count} times.
```
arikfr: Why this comment is not part of the email template generation instead of putting it inside the object?

redash/tasks/failure_report.py
```diff
@@ -0,0 +1,65 @@
...
+                     Reporting may stop when the query exceeds {max_failure_reports} overall failures.""".format(
+            failure_count=query.schedule_failures,
+            max_failure_reports=settings.MAX_FAILURE_REPORTS_PER_QUERY
+        ) if reporting_will_soon_stop else ''
+
+        redis_connection.lpush(key, json_dumps({
+            'id': query.id,
+            'name': query.name,
+            'base_url': base_url(query.org),
```
arikfr: Why not generate this when sending the report?

redash/tasks/failure_report.py
```diff
@@ -0,0 +1,65 @@
...
+            'message': message,
+            'comment': comment,
+            'failed_at': datetime.datetime.utcnow().strftime("%B %d, %Y %I:%M%p UTC")
+        }))
+
+        if not redis_connection.exists('{}:pending'.format(key)):
+            send_aggregated_errors.apply_async(args=(query.user.email,), countdown=settings.SEND_FAILURE_EMAIL_INTERVAL)
```
arikfr: I would avoid using Celery features like `countdown` as it can result in multiple executions of the same task 😞 Iirc, if `SEND_FAILURE_EMAIL_INTERVAL` > Celery's `VISIBILITY_TIMEOUT` (1 hour by default), you will get this task to execute more than once.

We can have a periodic task run every X minutes, that will check if there are any reports to send.

redash/tasks/failure_report.py
```diff
@@ -0,0 +1,65 @@
...
+            redis_connection.set('{}:pending'.format(key), 1)
+            redis_connection.expire('{}:pending'.format(key), settings.SEND_FAILURE_EMAIL_INTERVAL)
```
arikfr: If there is a delay in processing the queue, the reports might get lost. Why do we need the expire if we delete it anyway after sending?

redash/tasks/failure_report.py
rauchy: Do disabled users' scheduled queries continue to run? Shouldn't they be disabled or at least [skipped](https://github.com/getredash/redash/commit/b5b56430904a17819916d64f5d3b950d0a82d756#diff-5ca0dc869c9c3218d8a95ba7f99f5f4cR278)?

redash/tasks/failure_report.py
rauchy: Yeah that makes sense. Fixed in 0bfe9f4.

redash/tasks/failure_report.py
rauchy: No specific reason. 8182365.

redash/tasks/failure_report.py
rauchy: Heh, I didn't have the query back then, but since I'm not loading it anyway (to get the user's e-mail by id), I can just figure it out when creating the context. 4e80124

redash/tasks/failure_report.py
rauchy: giphy

8333e25

redash/tasks/failure_report.py
rauchy: This makes things simpler! I guess I needed an outside POV to realize that it really doesn't matter if the report is sent 60 minutes _after_ the failure, since these are scheduled queries...

Anyway, this is changed in 2fb4882 and it allowed me to remove the pending keys 6f0066e.

redash/settings/__init__.py
arikfr: > Does it make more sense to cap the number of emails sent per query instead of the number of failures per query?

Yep. But how will we track # of emails?

redash/tasks/failure_report.py
arikfr: Many times although the user left the company (or was disabled for whatever reason) their content (queries, dashboards) are still used.

redash/tasks/failure_report.py
arikfr: I think the benefit of sending it after 60 minutes is that it's more likely to pick up more failing queries in a single email, no?

rauchy: Yeah, but only for the first report. IMO it simplifies the code and less likely to cause duplication issues so I'm fine with it as it is now.

redash/tasks/failure_report.py
rauchy: 👍  7b232be

redash/tasks/failure_report.py
arikfr: I think that with a 60 seconds window the chances of it actually aggregating anything are slim. Maybe we should increase the default to 5 minutes

redash/settings/__init__.py
arikfr: Maybe the frequency should be once per day? We can rate limit it by creating an expiring key in Redis for a query id that received a report. Once the key exists -- we stop reporting, when the query expires we resume reporting.

rauchy: Yeah that's tricky because we need some sort of long-term storage (one that doesn't evaporate after an hour) like `Query.scheduled_failures`. We _could_ add an extra column or dump it in `Query.options`, but I'm not sure if we want to invest in that. WDYT?

arikfr: > Yeah that's tricky because we need some sort of long-term storage (one that doesn't evaporate after an hour)

What's wrong with using Redis for this?

rauchy: Keys will be left dangling. We could handle this by setting a weekly/monthly expiry - that is, up to 10 e-mails per query (each e-mail could represent multiple failures), and then silence for the rest of the week/month. Otherwise, the collection of dangling keys will grow and grow.

redash/tasks/failure_report.py
rauchy: It's [60 minutes](https://github.com/getredash/redash/pull/3798/files#diff-ad93995bb355251c02f87fc02db69351R44).

redash/settings/__init__.py
arikfr: > Keys will be left dangling.

Why? When the query fails we set a key with expiration = 24 hours. Next time if it fails, we check for the key existence and either skip reporting or create a new one. If the query stops failing the key will be removed in 24 hours.

redash/tasks/failure_report.py
arikfr: :man_facepalming:

redash/settings/__init__.py
rauchy: Which is what I wrote, but instead I offered a weekly/monthly expiry :)

The meaning of a daily expiry is that silenced reports (those who past the reporting threshold) will renew daily. This means that you may receive, for example, 10 e-mails a day, with the last few e-mails warning you that this will soon be silenced, and then the next morning you will start receiving them again.

Don't you think a weekly expiry is less annoying?

arikfr: I suggested the rate limit/silencing as an alternative to the counter. This way the user will receive only a single report per day for each query.

Issue #3785, opened 2019-05-11
z-Index of ant-modal-mask needs to increase


deecay:
### Issue Summary

In some cases, plotly modebar appears in Choropleth and other visualization editors because of z-Index misalignment.

### Steps to Reproduce

1. Have plotly visualization shown in query editor.
2. Click New Visualization and select Map (Choropleth).
3. You can make the plotly modebar of the underlying visualization to appear over choropleth map when you hover over where the plotly modebar exsists.

Plotly `modebar` class has fixed z-Index of `1001`. Our `ant-modal-mask` and `ant-modal-wrap` both has z-Index of `1000`. Maybe we need to increase these to 1002 or more?

kravets-levko:
Changing z-index for Ant modal will require changing it for other Ant components as well (otherwise it may break that components if using them within modals: https://github.com/ant-design/ant-design/blob/896011f5861eb4062abe7272cf657f2a644adcce/components/style/themes/default.less#L266-L273). It could be easier to fix Plotly instead (that's just another option).

deecay:
Maybe you're right. Plotly also has other components like tooltips close to z-Index of 1000, but seems to be easier to change than Ant components.

PR #3799, opened 2019-05-14
Plotly modebar appears above modals

kravets-levko:

- [x] Bug Fix

## Description

Increased z-index base (from `1xxx` to `2xxx`) for all Antd components.

Fixes getredash/redash#3785

### DIFF
--- a/client/app/assets/less/ant.less
+++ b/client/app/assets/less/ant.less
@@ -35,6 +35,16 @@
 @import "~antd/lib/notification/style/index";
 @import 'inc/ant-variables';
 
+// Increase z-indexes to avoid conflicts with some other libraries (e.g. Plotly)
+@zindex-modal: 2000;
+@zindex-modal-mask: 2000;
+@zindex-message: 2010;
+@zindex-notification: 2010;
+@zindex-popover: 2030;
+@zindex-dropdown: 2050;
+@zindex-picker: 2050;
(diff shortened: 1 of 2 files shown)


client/app/assets/less/ant.less
```diff
@@ -289,7 +300,7 @@
 }
 
 .ant-popover {
-  z-index: 1000; // make sure it doesn't cover drawer
+  z-index: 2000; // make sure it doesn't cover drawer
```
ranbena: Not `@zindex-popover`?

ranbena: Actually, this can be removed altogether?

kravets-levko: Well, I actually don't know where it's used (tried to find any place but failed). Can you please tell me how to check this?

ranbena: Screen Shot 2019-05-14 at 7 29 48

Above modal but under drawer (click the "?" icon in the popover to make sure)

kravets-levko: I was so close :sweat_smile: Will check in a moment. Thanks for pointing on this!

kravets-levko: @ranbena Fixed. I increased z-index only for help drawer - to place it above all other Antd components.

PR #3800, opened 2019-05-14
Edit parameter mapping: error when trying to change mapping type to Static; cannot change static value

kravets-levko:

- [x] Bug Fix

## Description

1. When editing parameter mapping for query-based parameter, changing it's type to static shows error in console: image Solution: pass required property to component.
2. Cannot change parameter mapping's static value (bug in data flow between components).
3. Some cleanup (removed unused properties).


kravets-levko:
@ranbena Thanks! I think it's another bug, I'll check.

kravets-levko:
@ranbena Looks that bug wasn't related to mentioned error message, but it was definitely a bug. Now fixed.

PR #3805, opened 2019-05-15
JSON Data Source

arikfr:

## Description

JSON data source, which allows loading data from (almost) any JSON URL ([Documentation](https://redash.io/help/data-sources/querying/urls#JSON)).

TODO:

~- [ ] Rename it to `HTTP` (?).~
~- [ ] Migration to update the type of existing URL data sources to use this one.~
~- [ ] Migration to update existing queries to support the new format.~
~- [ ] Remove URL data source.~

I figured there is no point in waiting to deprecate the URL data source and delaying the release of this one. We will do it in a separate task.

https://discuss.redash.io/t/missing-json-data-source/3448/22


AlanHodgkinson1:
Hello.

When is this likely to be available in the preview docker containers?

Thanks.

morsedl:
Fantastic, and thanks for your efforts on this! We're just now switching over to a more heavily REST-oriented approach, so this will be most helpful. Please keep us apprised, and if I can help with some testing or the like, I'd be happy to do so (at least as best I can; we're swamped laying down an entirely new software infrastructure).

rwatts3:
I understand you all are busy building, would it be possible to provide instructions for how to build this PR into a docker image so that it can be tested?

I attempted to deploy a version of redash from the docs with docker, however I ran into a few issues getting things to run nicely. I.E webpack, etc.

rwatts3:
Just stopping by to see if there is an update on the docker preview build. 
I see the new previews are starting to tag 8.0 beta, however, I'm still noticing there is no JSON data source.

flolivaud:
Hi everyone,

Tried to build docker image locally, but i can't view/use the current json datasources.
Tried to test it locally without docker too, still the same problem.

Is there something i miss about it, or maybe it's not functionnal now ?

Thanks.

redash/query_runner/json_ds.py
```diff
@@ -0,0 +1,222 @@
...
+            query = parse_query(query)
+
+            if not isinstance(query, dict):
+                raise QueryParseError(
+                    "Query should be a YAML object describing the URL to query.")
+
+            if 'url' not in query:
+                raise QueryParseError("Query must include 'url' option.")
+
+            if is_private_address(query['url']):
```
flolivaud: Why should we avoid querying private url/ip ?

arikfr: Because in some environments being able to query private addresses might allow you to poke around things you shouldn't (like AWS metadata of the instance hosting Redash). We can make this optional if needed.

flolivaud: I think it's a good idea to have an option to disable it

PR #3808, opened 2019-05-16
Fix "Now" in DateTime parameter not working

### LINKED ISSUE #3803: "Now" in DateTime parameter isn't working
The option to click on "Now" doesn't work:

* In a dashboard, it doesn't update the value at all.
* In a query, the value updates in the parameter, but the query executes with the previous value.

Examples:

https://redash-preview.netlify.com/dashboard/now-param?p_param=4444&p_dateparam=2019-01-25%2000:03#table

https://redash-preview.netlify.com/queries/54

gabrieldutra:

- [x] Bug Fix

## Description
I did find the cause for this bug, but with DatePicker methods and our current needs the only way I could think as a fix was to create an "unnecessary" state. Leaving this as WIP so @ranbena and @kravets-levko can join here (also I have to extend the fix to `DateTimeRangeInput` as a guarantee).

Antd's DatePicker provides the following callbacks:
- `onChange`: Triggered whenever you change the value on the dialog
- `onOk`: Triggered only when you click "ok"
- `onOpenChange`: Triggered when you open/close the dialog

`onOpenChange` seemed to be the most reliable event to trigger the update for the Query, so I used it by adding a state to track the value.

The cause for the issue: when you click "Now" it invokes `onChange` -> `onOpenChange` fast enough that the component doesn't update.
The workaround in this PR: Considering that React guarantees the order for `setState` I added a new state to use its callback to make sure `currentValue` would be updated.

BTW, `onOk` may seem a good solution, but it's not triggered when you click on "Now".

An alternative is to add the "Apply" button for this too, so the update event could be `onChange` with no problems. The con is that it's not the best to have both "Ok" and "Apply" to set the parameter.

Edit:
- Updated to use `useRef` instead.
Closes #3803
 
**Dashboard**
now-dashboard

**Query**
now-query

### DIFF
--- a/client/app/components/DateTimeInput.jsx
+++ b/client/app/components/DateTimeInput.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useRef } from 'react';
 import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
 import DatePicker from 'antd/lib/date-picker';
@@ -13,20 +13,21 @@ export function DateTimeInput({
 }) {
   const format = (clientConfig.dateFormat || 'YYYY-MM-DD') +
     (withSeconds ? ' HH:mm:ss' : ' HH:mm');
-  let defaultValue;
+  const additionalAttributes = {};
   if (value && value.isValid()) {
-    defaultValue = value;
+    additionalAttributes.defaultValue = value;
   }
-  const [currentValue, setCurrentValue] = useState(defaultValue);
+  const currentValueRef = useRef(additionalAttributes.defaultValue);
   return (
     <DatePicker
       className={className}
       showTime
-      value={currentValue}
+      {...additionalAttributes}
       format={format}
       placeholder="Select Date and Time"
-      onChange={newValue => setCurrentValue(newValue)}
+      onChange={(newValue) => { currentValueRef.current = newValue; }}
       onOpenChange={(status) => {
+        const currentValue = currentValueRef.current;
         if (!status) { // on close picker
           if (currentValue && currentValue.isValid()) {
             onSelect(currentValue);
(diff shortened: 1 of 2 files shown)


client/app/components/DateTimeInput.jsx
```diff
@@ -1,55 +1,58 @@
...
+    return (
+      <DatePicker
+        className={className}
+        showTime
+        value={this.state.currentValue}
+        format={format}
+        placeholder="Select Date and Time"
+        onChange={newValue => this.setState({ currentValue: newValue })}
+        onOpenChange={(status) => {
+          this.setState({ open: status }, () => {
```
gabrieldutra: unnecessary state just to use the callback

kravets-levko:
@gabrieldutra I analyzed the code a little bit and I found the bug. `DateTimeInput` uses `useState` to keep selected value while user changes selection to fire `onChange` event only when dropdown is closed. It works when user selects value and then presses "Ok" button: `onChange -> setCurrentValue -> render -> "Ok" -> onOpenChange -> onChange (uses new value of currentValue)` - keep in mind that functional components will "capture" props and state values until next render. So when user uses "Now" button - sequence is a bit different: `onChange -> setCurrentValue -> onOpenChange (immediately!) -> onChange (uses captured previous value)`. So your fix works just because you switched from functional to class component, new state variable can be removed. Also, you can try to use `useRef` instead of `useState` - I tried it, it works as well.

gabrieldutra:
Thanks @kravets-levko :grin:

> So your fix works just because you switched from functional to class component, new state variable can be removed.

I'd thought so, however this was one of my tests (the values differ):
date-time

> Also, you can try to use `useRef` instead of `useState`

State seemed more the "React way", but `useRef` will be the best here indeed. I'll move to it :slightly_smiling_face:.

gabrieldutra:
@kravets-levko, updated to use `useRef`. Seems to be working fine and code much less polluted :ok_hand:

Issue #3575, opened 2019-03-12
Elasticsearch aggretations not working


LuisDeHaro:
### Issue Summary

When I try to run an aggregation query to the ES cluster directly it works, in Redash it doesn't.

Query:

```
{
    "query": {
        "match": {
            "user.email": "some@email.com"
        }
    },
    "aggs" : {
        "subtotal" : { "sum" : { "field" : "subtotal" } }
    },
    "size": 0
}

```
Expected output (when executed directly to the cluster):

```
{
  "took" : 6,
  "timed_out" : false,
  "_shards" : {
    "total" : 77,
    "successful" : 56,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : 5,
    "max_score" : 0.0,
    "hits" : [ ]
  },
  "aggregations" : {
    "subtotal" : {
      "value" : 592.0
    }
  }
}

```
### Steps to Reproduce

1. Created a ES datasource successfully 
2. Running the above query 
3.- 0 rows shown

### Technical details:

* Redash Version: 6.0.0+b8537 (4780bd9c)
* Browser/OS: MacOS 10.14.1, Chrome 72.0.3626.121
* How did you install Redash: Docker
* ElasticSearch version: 5.6.4

arikfr:
Aggregations have very limited support at the moment. There is a community member currently looking into improving the support and we plan to allocate resources to this too when we finish other ongoing projects.

LuisDeHaro:
@arikfr, thanks for your response. 

I've been using Redash connected with PGSQL and it works just perfect, amazing and beautifully. 

However this week creating a new data source pointing to ElasticSearch I would say: it's not useful... needs a lot of improvement in order to just do the job. 

I hope this get's fixed soon, I would like to use Redash for every data source I have now and in the future. 

God bless Israel my friend.

lkm:
@arikfr I wouldn't mind lending a hand to code and test this. Could you please point me to the community members code so I don't repeat anything?

Issue #2789, opened 2018-09-06
Elasticsearch Nested Aggregations


shivam1892:
### Issue Summary

I am working on creating the table from elasticsearch with nested aggregation query.
Below is my example Query

{
“index”:“data”,
“size”: 0,
“aggs”: {
“date_aggr”: {
“date_histogram”: {
“field”: “timestamp”,
“interval”: “1d”
},
“aggs”: {
“agg_abc”: {
“terms”: {
“field”: “abc.keyword”,
“order”: {
“price”: “desc”
}
},
“aggs”: {
“price”: {
“sum”: {
“field”: “price”
}
}
}
}
}
}
}
}

The above Nested aggregated query does not give all the columns in the table as the response of elasticsearch. Also “agg_abc” have more than 5 categories in it but in table it shows only for 2 categories which is also not complete.
According to me the nested aggregated query is not being parsed properly as some of the columns are missing.
But when I hit the above query to elastic search It gives the proper response with all categories.

### Technical details:
* Redash version 4.0.1
* Ubuntu 16.04
* Installed with docker setup guide.

arikfr:
Please share an example of the results you do get for this query in Redash and the results you get when querying directly. Thanks.

moghira10:
@arikfr Any update on this?

ivanli1990:
I met this issue too.
```
My result from elasticsearch:
{
  "took": 1,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": 19,
    "max_score": 0,
    "hits": []
  },
  "aggregations": {
    "moduleId": {
      "doc_count_error_upper_bound": 0,
      "sum_other_doc_count": 0,
      "buckets": [
        {
          "key": "数据",
          "doc_count": 11,
          "labels": {
            "doc_count_error_upper_bound": 0,
            "sum_other_doc_count": 0,
            "buckets": [
              {
                "key": "数据自身问题",
                "doc_count": 10,
                "onesId_num": {
                  "value": 10
                }
              },
              {
                "key": "未填写原因",
                "doc_count": 1,
                "onesId_num": {
                  "value": 1
                }
              }
            ]
          }
        },
        {
          "key": "多码后端",
          "doc_count": 8,
          "labels": {
            "doc_count_error_upper_bound": 0,
            "sum_other_doc_count": 0,
            "buckets": [
              {
                "key": "未填写原因",
[...]

PR #3809, opened 2019-05-16
Support for nested aggregations and single value aggregations

lkm:
Signed-off-by: Lasse Mammen <lkm@bookboon.com>

- [x] Refactor
- [x] Bug Fix

## Description
This will add support for sub-aggregations in Elasticsearch query runner. During my initial pass at this I couldn't find in "out of the box" way of parsing Elasticsearch queries into a tabular output format. 

In the results it will use a `>` character to separate aggregation results in column headers as this is the way in Elasticsearch to reference aggregations. 

I'm not the most routined python coder, so please let me know if there's anything that can be done in order to improve the code quality. Also this includes a tests for query parser. 

#3575 and #2789

### DIFF
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -197,34 +197,53 @@ class BaseElasticSearch(BaseQueryRunner):
             add_column_if_needed(mappings, key, key, result_columns, result_columns_index)
             row[key] = value
 
-        def collect_aggregations(mappings, rows, parent_key, data, row, result_columns, result_columns_index):
-            if isinstance(data, dict):
-                for key, value in data.iteritems():
-                    val = collect_aggregations(mappings, rows, parent_key if key == 'buckets' else key, value, row, result_columns, result_columns_index)
-                    if val:
-                        row = get_row(rows, row)
-                        collect_value(mappings, row, key, val, 'long')
-
-                for data_key in ['value', 'doc_count']:
-                    if data_key not in data:
-                        continue
-                    if 'key' in data and len(data.keys()) == 2:
-                        key_is_string = 'key_as_string' in data
-                        collect_value(mappings, row, data['key'] if not key_is_string else data['key_as_string'], data[data_key], 'long' if not key_is_string else 'string')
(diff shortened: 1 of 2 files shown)


redash/query_runner/elasticsearch.py
```diff
@@ -197,34 +197,53 @@ def collect_value(mappings, row, key, value, type):
...
+                if key == 'key_as_string':
+                    continue
+                if key == 'key':
+                    if 'key_as_string' in data:
+                        collect_value(mappings, row, agg_key, data['key_as_string'], 'string')
+                    else:
+                        collect_value(mappings, row, agg_key, data['key'], 'string')
+                    continue
+
+                if type(item) is str:
```
truebit: using `isinstance(item, str)` may be better; as also in all below type checks

truebit:
I think we should apply to kibana's logic to deal with rows

redash/query_runner/elasticsearch.py
lkm: Done

lkm:
@truebit If that comment is meant for me I'm not sure what you'd suggest doing. Is there some code in Kibana you were thinking of specifically? I don't know that codebase well, but I would anticipate that there are substantial architectural differences given ES output schemas etc.

truebit:
@lkm
You could copy the kibana datatable visualization `Request`([instructions](https://github.com/elastic/kibana/issues/12218#issuecomment-307489746)), and then paste them in Redash query. 
Although the column name is different, but the column order indicates the readability.
IMHO, the column list order in this PR should be following `Inorder Traversal (LDR) `. and `doc_count` related column should be in the last. That would be more clear for user. Just my two cents:)

lkm:
Thanks for the feedback @truebit, I appreciate it. It is actually transversing tree using pre-order, unless I'm misunderstanding something. The column order returned from test `test_get_sub_aggregation` is:

`agg_name, agg_name>doc_count, sub_agg, sub_agg>doc_count`

Am I understanding correctly that you would expect to see the following instead?
`agg_name, sub_agg, agg_name>doc_count, sub_agg>doc_count`

arikfr:

PR #3810, opened 2019-05-17
Cypress tests for query parameters

gabrieldutra:

- [x] Other

## Description
Following #3737 (and the bug #3803) I wanted to add a few tests to another important part of Redash: Query parameters :grin:

Parameters:
- [x] Text
- [x] Number
- [x] Date
- [x] Date and time
  - [x] Date selection
  - [x] "Now" button
- [x] Dropdown
- [x] Use requests instead of UI to improve speed

--


client/cypress/integration/query/parameter_spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    cy.getByTestId('QueryEditor')
+      .get('.ace_text-input')
+      .type("SELECT '{{}{{}test-parameter}}' AS parameter{esc}", { force: true });
+  });
+
+  afterEach(() => {
+    cy.getByTestId('SaveButton').click();
+    cy.url().should('match', /\/queries\/\d+\/source/);
+  });
```
ranbena: You sure this is needed? Can leave new query unsaved.

ranbena:
Suggesting to add tests in Dashboard context.

client/cypress/integration/query/parameter_spec.js
gabrieldutra: It's either save the query or [this](https://github.com/getredash/redash/pull/3748/commits/c3d521fb9cc6a787d4d7791fe696204daacb2bb6) workaround :grimacing:

gabrieldutra:
> Suggesting to add tests in Dashboard context.

Mostly I want to test the `Parameters` component (the shared part), which seemed easier from the Query context. But I think it's good to add the "extra" part that dashboards have (global, local and static parameters).

--

Just to share some points I was thinking about:
- It can be interesting to save the queries and name then according to the parameter type - so it's possible to access the server and its state well organized after the tests;
- We could introduce Faker or any other library with the same purpose if we want some dynamic data (thought this after thinking about random numbers to put on tests :sweat_smile:). I think this one is out of scope here, but I wanted to share it anyway (can be addressed as a GH issue).

client/cypress/integration/query/parameter_spec.js
ranbena: Oh right right!

ranbena:
> Mostly I want to test the `Parameters` component (the shared part), which seemed easier from the Query context. But I think it's good to add the "extra" part that dashboards have (global, local and static parameters).

I mean specifically test the "apply" functionality in the dashboard context, that it triggers widget refresh.

> * It can be interesting to save the queries and name then according to the parameter type - so it's possible to access the server and its state well organized after the tests;

For what purpose? To db-seed these queries?

arikfr:
Great initiative!

gabrieldutra:
It seems that Cypress has an issue when it's necessary to click more than once on Antd's Date Calendar, the first click works, the following ones don't ^^. This affected creating specs for DateTime and DateRanges. I think my last test will be to trigger DateTime by clicking enter or clicking outside the panel.

> I mean specifically test the "apply" functionality in the dashboard context, that it triggers widget refresh.

I thought about adding a smaller test just to make sure the behavior extends to the dashboard context, but this is sort of already happening with specs that rely on that behavior (e.g: "Height behavior on refresh"). Do you think we gain any value by having that separate?

> For what purpose? To db-seed these queries?

:thinking:. I though more about development and unautomated tests. After running Cypress you'd get Redash with a basic set of organized data (in this case, queries that you could test each parameter). I'll think better about it, perhaps sth not for now as I don't really use Cypress data after it runs (and guessing nobody actually does it here).

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -188,6 +188,7 @@ function EditParameterSettingsDialog(props) {
               rows={3}
               value={param.enumOptions}
               onChange={e => setParam({ ...param, enumOptions: e.target.value })}
+              data-test="DropdownValuesInput"
```
gabrieldutra:
```suggestion
```

client/cypress/support/redash-api/index.js
```diff
@@ -19,11 +19,11 @@ export function createQuery(data, shouldPublish = true) {
   }, data);
 
   // eslint-disable-next-line cypress/no-assigning-return-values
-  let request = cy.request('POST', '/api/queries', merged);
+  let request = cy.request('POST', '/api/queries', merged).then(({ body }) => body);
```
gabrieldutra: To be consistent with the `shouldPublish = true` changed it to return the body directly.

PR #3813, opened 2019-05-19
delete variable

taminif:

- [x] Refactor

## Description
I found variable don't used and re-declarate. So it deleted

### DIFF
--- a/redash/query_runner/mongodb.py
+++ b/redash/query_runner/mongodb.py
@@ -250,7 +250,6 @@ class MongoDB(BaseQueryRunner):
 
                     step["$sort"] = SON(sort_list)
 
-
         if "fields" in query_data:
             f = query_data["fields"]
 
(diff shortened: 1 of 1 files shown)


redash/query_runner/mongodb.py
```diff
@@ -251,7 +251,6 @@ def run_query(self, query, user):
                     step["$sort"] = SON(sort_list)
 
         if not aggregate:
-            s = None
```
arikfr: It looks like this whole if block can be removed, as it being overridden below.

taminif: that's exactly. thanks!

taminif:
Thanks for the reviews @arikfr . I updated code. Please review once again!

arikfr:
So it took me a while to get back to this, but it's merged now. Thanks.

PR #3819, opened 2019-05-22
Plug custom Celery tasks via dynamic settings

rauchy:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
This would allow users to provide their own custom tasks to run periodically.

### DIFF
--- a/redash/settings/dynamic_settings.py
+++ b/redash/settings/dynamic_settings.py
@@ -8,3 +8,12 @@ def query_time_limit(is_scheduled, user_id, org_id):
     adhoc_time_limit = int_or_none(os.environ.get('REDASH_ADHOC_QUERY_TIME_LIMIT', None))
 
     return scheduled_time_limit if is_scheduled else adhoc_time_limit
+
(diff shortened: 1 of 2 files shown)


jezdez:
Huh, that's a new twist on the extension problem, we have had extensible periodic tasks for a while until they were removed in #3465. Would you say the overlap is okay keeping given dynamic settings are more deployment specific and not particularly tied to extensions?

That reminds me @arikfr I just saw your last comment in #3569 about the requirements file added there. Merging this would go a long way to a better separation of App and Worker code.

jezdez:
IOW, I'm very much interested in solving #3466 still.

arikfr:
@jezdez just to clarify: this is not instead of the extension point for periodic tasks as implemented in #3569. This is just another option which is easier to use in some scenarios.

As for the requirements file, I will follow up in #3569.

jezdez:
Okay, understood, thanks for clarification!

PR #3822, opened 2019-05-22
Include BigQueryGCE into default query runners

rueian:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

include big_query_gce into default_query_runners
include bigquery_gce.png into db-logos

https://github.com/getredash/redash/pull/3356

before:

image

after:

image


rueian:
Thanks, I will create another PR for the missing icon.

And may I ask the reason why BigQuery GCE is not enabled by default? Will it be removed soon?

It was just suddenly disappeared after we upgraded redash to v7.0.0, and took me a while to find out how it was disappeared.

Maybe we can add some documentation about this?

arikfr:
> And may I ask the reason why BigQuery GCE is not enabled by default? Will it be removed soon?

Its `enabled` check was sometimes causing issues when trying to start Redash and because it's not commonly used, it was simpler to remove from the default list.

There are no plans to remove it at the moment.

> Maybe we can add some documentation about this?

Good idea! The documentation is in the [website repo](https://github.com/getredash/website).

PR #3826, opened 2019-05-23
Migrate CreateDashboardDialog to React

kravets-levko:

- [x] Feature

## Description

Migrate CreateDashboardDialog to React. Restore some `DynamicComponent` usages (missed during React migration). 

image
image


client/app/components/dashboards/CreateDashboardDialog.jsx
```diff
@@ -0,0 +1,86 @@
...
+  const isCreateDashboardEnabled = policy.isCreateDashboardEnabled();
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      if (inputRef.current) {
+        inputRef.current.focus();
+      }
+    }, 100);
+    return () => clearTimeout(timer);
+  }, []);
```
ranbena: I think you can substitute this for `<Input autofocus />`, no?

client/app/components/dashboards/CreateDashboardDialog.jsx
```diff
@@ -0,0 +1,86 @@
...
+    const value = trim(event.target.value);
+    setName(value);
+    setIsValid(value !== '');
+  }
+
+  function save() {
+    if (name !== '') {
+      setSaveInProgress(true);
+
+      $http.post('api/dashboards', { name })
```
ranbena: Wdyt of `_.capitalize(name)`?

client/app/components/dashboards/CreateDashboardDialog.jsx
kravets-levko: It does not work in modals. This fix was suggested in [related discussion](https://github.com/ant-design/ant-design/issues/8668)

client/app/components/dashboards/CreateDashboardDialog.jsx
kravets-levko: Let's keep it as is (at least, for now). Automatic capitalization is not so brilliant idea in most cases (just because algorithm is too generic), especially if we do it on save - so user will have no chance to fix typos introduced by `_.capitalize` (I think that's bad UX?)

client/app/components/dashboards/CreateDashboardDialog.jsx
gabrieldutra: Weird, it seems to work on `CreateSourceDialog` and `ChangePasswordDialog` :thinking:

kravets-levko: @gabrieldutra I'll check once more

kravets-levko: @gabrieldutra I see that `autoFocus` work in other modals, but it doesn't work in this one - I really have no idea what's wrong :thinking: I tried to simplify that dialog as mu as i can - removed almost all attributes, `<DynamicComponent>`, etc. - just `<Modal ...><Input autoFocus /></Modal>` - input does not get focused.

ranbena: There's one significant difference - Class component vs Functional component 🤔

kravets-levko: @ranbena Interesting idea. I'll try to implement it as class component - will see what happens.

kravets-levko: @ranbena Tried class component - the same issue (no autofocus).

Upd.: When this dialog is opened from React code - Input gets autofocus, when from Angular - no. I'll add a comment with `ANGULAR_REMOVE_ME`

gabrieldutra: > When this dialog is opened from React code - Input gets autofocus, when from Angular - no.

Makes sense and also good to know this sort of thing happens :thinking:

kravets-levko: Yeah... But usually applications are built using only one framework, so such shit does not happen :slightly_smiling_face:

PR #3830, opened 2019-05-25
Fix for Mac browser’s scrollbar obscuring content

ranbena:
- [x] Bug Fix

## Description
In table visualizations two visual bugs occur on Mac browsers (FF, SF, CH):
1. Scrollbar obscures table content. Solved in this PR.
2. After scroll the bar doesn't disappear in CH, SF. There's an [open bug for it](https://bugs.chromium.org/p/chromium/issues/detail?id=914844#c36) and very recent discussions.

Before:

Screen Shot 2019-05-24 at 18 48 52

After:

Screen Shot 2019-05-24 at 18 48 35

### DIFF
--- a/client/app/components/app-view/index.js
+++ b/client/app/components/app-view/index.js
@@ -37,6 +37,10 @@ class AppViewComponent {
     this.layout = layouts.defaultSignedOut;
     this.handler = handler;
 
+    // remove when fix lands
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=914844#c36
+    $rootScope.platform = navigator.platform;
+
     $rootScope.$on('$routeChangeStart', (event, route) => {
       this.handler.reset();
 
--- a/client/app/components/dynamic-table/dynamic-table.less
+++ b/client/app/components/dynamic-table/dynamic-table.less
@@ -1,6 +1,13 @@
 .dynamic-table-container {
   overflow-x: auto;
 
+  // remove when fix lands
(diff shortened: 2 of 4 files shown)


kravets-levko:
@ranbena There are issues on Linux (Chrome and FF, didn't have a chance to test on Windows) - before & after:
image
image

arikfr:
The same happens on Windows/Chrome. Is there a way to target a specific OS?

gabrieldutra:
Perhaps remove the `margin-bottom: -10px; // compensation` will make this acceptable?

ranbena:
10x @kravets-levko.

New approach:

```less
.table.table-condensed {
  tbody > tr:last-child > td {
    padding-bottom: 15px; // <-- make room for scrollbar within the last row
  }
}
```

ranbena:
> The same happens on Windows/Chrome. Is there a way to target a specific OS?

Not natively I believe but it's possible though I would avoid it unless last resort.

ranbena:
> Perhaps remove the `margin-bottom: -10px; // compensation` will make this acceptable?

It's a viable option. Though it would be a tradeoff in both Mac and Win/Lin.

ranbena:
Looking into this further I suggest:
1. 1st solution (dynamic-table bottom padding)
2. Only for mac - with js and css.
3. Revert commit completely when [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=914844) is fixed and published.

ranbena:
@kravets-levko ready for review

arikfr:
👍

PR #3832, opened 2019-05-27
Unsupported browser redirect script outputted to file

### LINKED ISSUE #3816: Content Security Policy error because of inline script
### Issue Summary

@ranbena @arikfr The change in #3609 leads to a Content Security Policy error when loading any page:

Chrome:

```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-lUaehHefE2UfaxjnDzUj5HBFcQ3z+oaNbFFBqOJn9Ck='), or a nonce ('nonce-...') is required to enable inline execution.
```

Firefox:
```
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).
```

I don't think we should add `unsafe-inline` to the `script-src` directive but instead indeed provide a per-request nonce in the format

```
<script nonce="<random nonce>">
    //...
</script>
```

Flask-Talisman supports rendering a random nonce like [shown in the example here ](https://github.com/GoogleCloudPlatform/flask-talisman#example-6). Sadly we can't just not apply CSP since the templates are used on every page, if I understand correctly.

Right now `multi_org.html` is the only one that is rendered via Jinja2, but I think it's safe to do that for `index.html`, right?

Alternatively, would it suffice to move the `<script>` block into an own script that we can load instead?

Can you think of a way to fix this?

### Steps to Reproduce

1. Open any page.
2. See the browser console.

### Technical details:

* Redash Version: master
* Browser/OS: Firefox/macOS

ranbena:
- [x] Refactor
- [x] Bug Fix

## Description
Fixes #3816.

* Moved browser detection/redirect script to file in order to comply with csp directive.
* Made script [async](https://caniuse.com/#feat=script-async) for optimal perf.
* Used alternative method to detect IE<12 to allow above method.

Script now executing asynchronously. Console error gone.

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -5,18 +5,7 @@
     <meta charset="UTF-8">
     <base href="/">
     <title>Redash</title>
-    <script>
-      function onIE() {
-        window.location.href = '/static/unsupported';
-      }
-      if (
-        navigator.userAgent.match('MSIE') || // IE10
-        navigator.appVersion.match('Trident/') // IE11
-      ){
-        onIE();
-      }
-    </script>
-    <!--[if IE]><script>onIE()</script><![endif]-->
+    <script src="/static/unsupportedRedirect.js" async></script>
 
     <link rel="icon" type="image/png" sizes="32x32" href="/static/images/favicon-32x32.png">
     <link rel="icon" type="image/png" sizes="96x96" href="/static/images/favicon-96x96.png">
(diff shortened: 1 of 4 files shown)


ranbena:
Was not tested in IE.
@kravets-levko can you help out?

kravets-levko:
@ranbena I have only IE11 right now, and everything works fine there:
image

P.S. I also have IE8, but it does not support TLS1.2 :man_shrugging:

PR #3840, opened 2019-05-28
Custom Map Markers

deecay:

- Feature

## Description
Adds the feature to customize the looks of the map markers. You can change marker colors according to visualizations/ColorPalette, and marker icons with Font Awesome.

Before
58472199-b2f97280-8180-11e9-8755-3eeadf6b485c-min

After (Showing train stations with 'train' icon)
58472220-c0166180-8180-11e9-9dcf-6de1f2a4fc52-min

Library used: https://github.com/marslan390/BeautifyMarker
Available icons: https://fontawesome.com/v4.7.0/icons/

editor (rectangle shape with 'ambulance' icon)
editor1

(dot shape, similar look to current circleMarker)
editor2-min

(circle with icon)
editor3-min

## WIP Because...
* Is Font Awesome okay? Upgrade from FontAwesome v4 to v5 may become difficult because of this feature. Icon names have changed from v4 to v5, so if we save the icon name as setting, we might need to migrate.
* How to incorporate this with Groups? Icon per group?
* ~Should we create a list of possible icon names as a list (+600)?~

### DIFF
--- a/client/app/visualizations/map/index.js
+++ b/client/app/visualizations/map/index.js
@@ -347,12 +347,12 @@ const MapEditor = {
     this.mapTiles = MAP_TILES;
 
     this.iconShapes = {
-      marker: 'Marker',
-      circle: 'Circle',
-      rectangle: 'Rectangle',
+      marker: 'Marker + Icon',
+      doughnut: 'Circle',
       'circle-dot': 'Circle Dot',
-      'rectangle-dot': 'Rectangle Dot',
-      doughnut: 'Doughnut',
+      circle: 'Circle + Icon',
+      'rectangle-dot': 'Square Dot',
+      rectangle: 'Square + Icon',
     };
 
     this.colors = {
--- a/client/app/visualizations/map/index.js
+++ b/client/app/visualizations/map/index.js
@@ -93,7 +93,7 @@ const DEFAULT_OPTIONS = {
   clusterMarkers: true,
   iconShape: 'marker',
   iconFont: 'circle',
-  textColor: '#ffffff',
+  foregroundColor: '#ffffff',
   backgroundColor: '#356AFF',
   borderColor: '#356AFF',
 };
@@ -254,7 +254,7 @@ const MapRenderer = {
               iconSize: this.options.iconShape === 'rectangle' ? [22, 22] : false,
               iconAnchor: iconAnchors[this.options.iconShape],
               prefix: 'fa',
-              textColor: this.options.textColor,
+              textColor: this.options.foregroundColor,
               backgroundColor: this.options.backgroundColor,
               borderColor: this.options.borderColor,
             });
--- a/client/app/visualizations/map/index.js
+++ b/client/app/visualizations/map/index.js
@@ -79,6 +79,15 @@ const MAP_TILES = [
   },
 ];
 
+const iconAnchors = {
+  marker: [14, 32],
+  circle: [10, 10],
+  rectangle: [11, 11],
+  'circle-dot': [1, 2],
+  'rectangle-dot': [1, 2],
+  doughnut: [8, 8],
+};
+
 const DEFAULT_OPTIONS = {
   classify: 'none',
   clusterMarkers: true,
@@ -242,6 +251,8 @@ const MapRenderer = {
             const icon = L.BeautifyIcon.icon({
               iconShape: this.options.iconShape,
               icon: this.options.iconFont,
+              iconSize: this.options.iconShape === 'rectangle' ? [22, 22] : false,
+              iconAnchor: iconAnchors[this.options.iconShape],
               prefix: 'fa',
               textColor: this.options.textColor,
               backgroundColor: this.options.backgroundColor,
@@ -331,8 +342,9 @@ const MapEditor = {
 
     this.iconShapes = {
       marker: 'Marker',
-      'circle-dot': 'Circle Dot',
+      circle: 'Circle',
       rectangle: 'Rectangle',
+      'circle-dot': 'Circle Dot',
       'rectangle-dot': 'Rectangle Dot',
       doughnut: 'Doughnut',
     };
--- a/client/app/visualizations/map/index.js
+++ b/client/app/visualizations/map/index.js
@@ -5,6 +5,8 @@ import 'leaflet.markercluster';
 import 'leaflet/dist/leaflet.css';
 import 'leaflet.markercluster/dist/MarkerCluster.css';
 import 'leaflet.markercluster/dist/MarkerCluster.Default.css';
+import 'beautifymarker';
+import 'beautifymarker/leaflet-beautify-marker-icon.css';
 import markerIcon from 'leaflet/dist/images/marker-icon.png';
 import markerIconRetina from 'leaflet/dist/images/marker-icon-2x.png';
 import markerShadow from 'leaflet/dist/images/marker-shadow.png';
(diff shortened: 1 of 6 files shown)


deecay:
@ranbena, thanks for your evaluation and input.

> 1. The different markers types do not indicate the same exact location.
I didn't expect this kind of behavior. Thanks for examining. This should be adjustable, so I will look into this first.

For all other suggestions, I will make additional commits after fixing #1.

deecay:
@ranbena 
Where do you think the rectangle marker should align? 

The 'x' should come at the center point of the bottom edge of the rectangle? Right on, or do we want some margin between the 'x' and the box?

Another option could be that 'x' comes at the rectangle's center of gravity. Icon will hide what's at the 'x' point, but it is arguably more accurate.

WDYT?

deecay:
In addition to the feedback, I have
1. Added new shape type 'Circle', which was missing. Screenshot added to the original comment.

2. Made Rectangle a square shape. And it will be positioned so that the specified coordinates comes to the center of its gravity. Screenshot updated.

3. Added Color Name to color dropdown list.

ranbena:
> @ranbena
> Where do you think the rectangle marker should align?
> 
> The 'x' should come at the center point of the bottom edge of the rectangle? Right on, or do we want some margin between the 'x' and the box?
> 
> Another option could be that 'x' comes at the rectangle's center of gravity. Icon will hide what's at the 'x' point, but it is arguably more accurate.
> 
> WDYT?

Sorry for disappearing @deecay. Was on vacation 🏝
I think having all shapes (except for the "Marker") be in exact center makes sense.

client/app/visualizations/map/index.js
```diff
@@ -229,7 +247,21 @@ const MapRenderer = {
           const groupColor = this.options.groups[name].color;
           marker = heatpoint(lat, lon, groupColor);
         } else {
-          marker = createMarker(lat, lon);
+          if (this.options.customizeMarkers) {
+            const icon = L.BeautifyIcon.icon({
+              iconShape: this.options.iconShape,
+              icon: this.options.iconFont,
+              iconSize: this.options.iconShape === 'rectangle' ? [22, 22] : false,
```
ranbena: @deecay can you elaborate on why rectangle is a special case?

client/app/visualizations/map/map-editor.html
```diff
@@ -70,6 +72,91 @@
...
+          <label class="control-label">Foreground Color</label>
+          <ui-select ng-model="$ctrl.options.foregroundColor"
+            ng-disabled="!$ctrl.options.customizeMarkers">
+            <ui-select-match>
+              <color-box color="$select.selected.value"></color-box>
+              <span ng-bind-html="$select.selected.key | capitalize"></span>
+            </ui-select-match>
+            <ui-select-choices repeat="color.value as (key, color) in $ctrl.colors">
+              <color-box color="color.value"></color-box>
+              <span ng-bind-html="color.key | capitalize | highlight: $select.search"></span>
```
ranbena: This is great (I assume it's due to the white color) but notice the text and color are a bit vertically misaligned.

client/app/visualizations/map/map-editor.html
```diff
@@ -70,6 +72,91 @@
...
+          <label>Shape</label>
+          <select ng-options="key as value for (key, value) in $ctrl.iconShapes"
+            ng-model="$ctrl.options.iconShape" class="form-control"
+            ng-disabled="!$ctrl.options.customizeMarkers"></select>
+        </div>
+      </div>
+      <div class="col-xs-6">
+        <div class="form-group">
+          <label>Icon Font
+            <span class="m-1-5" uib-popover-html="'Available <a href=&quot;https://fontawesome.com/v4.7.0/icons/&quot; target=&quot;_blank&quot;>icons.</a>'"
```
ranbena: I'd like to suggest a bit more verbosity. Sth like:
Screen Shot 2019-06-15 at 17 31 06

client/app/visualizations/map/index.js
ranbena: Also, it seems `[22,22]` is the BeautifyMarker default so perhaps `this.options.iconShape === 'rectangle' ? null : false` could be better?

client/app/visualizations/map/index.js
```diff
@@ -308,6 +340,20 @@ const MapEditor = {
...
     this.mapTiles = MAP_TILES;
 
+    this.iconShapes = {
+      marker: 'Marker',
+      circle: 'Circle',
+      rectangle: 'Rectangle',
+      'circle-dot': 'Circle Dot',
+      'rectangle-dot': 'Rectangle Dot',
+      doughnut: 'Doughnut',
+    };
```
ranbena: I know these are the original labels, but they're not very accurately descriptive.

@deecay How do you feel about these?
```js
{
  marker: 'Marker + Icon',
  doughnut: 'Circle',
  'circle-dot': 'Circle Dot',
  circle: 'Circle + Icon',
  'rectangle-dot': 'Rectangle Dot',
  rectangle: 'Rectangle + Icon',
}
```
Screen Shot 2019-06-15 at 17 43 30

client/app/visualizations/map/index.js
deecay: `[22, 22]` is there to make the marker a square. Comparison:
image

Another example at the PR comment above, with ambulance icon.

I thought square looked better and also consistent with other marker types.

client/app/visualizations/map/map-editor.html
deecay: Much better 😄

client/app/visualizations/map/index.js
deecay: Strangely, I couldn't make the default value work, so I had to specify the size.

client/app/visualizations/map/index.js
deecay: I think it's now clearer. The remaining thing is if we want "Rectangle" or "Square".

Because:
* `Rectangle Dot` is actually square dot (you can see that in your screenshot's orange square)
* As discussed, I would like to make `Rectangle` actually a square.

client/app/visualizations/map/map-editor.html
deecay: Help needed here. I played around a little, but I couldn't make the color box aligned for both selected-item and listed-item.

maybe two pixels up for selected-item and one pixel down for listed-item?
image

client/app/visualizations/map/index.js
deecay: Seems like default of `[22, 22]` only applies to circles.

https://github.com/marslan390/BeautifyMarker/blob/b650118f45a683437818278d1df7cfa778a9e9d8/leaflet-beautify-marker-icon.js#L48

client/app/visualizations/map/map-editor.html
deecay: I've tried my best. Please review.

ranbena: Here ya go https://github.com/deecay/redash/pull/2

client/app/visualizations/map/map-editor.html
```diff
@@ -70,6 +72,91 @@
...
+        <input type="checkbox" ng-model="$ctrl.options.customizeMarkers">
+        <i class="input-helper"></i>Override default style
+      </label>
+    </div>
+
+    <div class="row">
+      <div class="col-xs-6">
+        <div class="form-group">
+          <label>Shape</label>
+          <select ng-options="key as value for (key, value) in $ctrl.iconShapes"
```
ranbena: For visual consistency, can you convert this to a `ui-select` like the rest?

deecay:
@ranbena, ready for review again.

deecay:
Thank you @ranbena for your support.

ranbena:
Thanks @deecay for putting in the time to make this a feature available to all Redash users 🍾

PR #3841, opened 2019-05-28
CSV: correctly serialize booleans and dates.

### LINKED ISSUE #2751: Datetime format on CSV is different.
Re-opening from issue #2557 since it otherwise is the same situation we are in:

---
Issue Summary
When I download CSV file on the queries page,
the datetime format that CSV has is not as same as an original data.
is this a bug or a feature?

Original data on MySQL is "2018-01-23 14:00:30"
Redash CSV is "2018-01-23T14:00:30"

I'd like to get rid of "T".

Redash Version: redash:4.0.0-rc.1
Browser/OS: Chrome
How did you install Redash: Docker

---

Now I did see the workaround for this:

`SELECT DATE_FORMAT(NOW(), '%Y-%m-%d %H:%i:%s') now;`

However as we use Redash across many queries with many people it would seem a bit silly to have everyone use the workaround since it otherwise is not an issue if used directly with the database.  Is there a global environment setting or something we can change to make ReDash not put the T in the date format?

Thanks!

### LINKED ISSUE #3736: Exporting data should returns SQL types and not python types
### Issue Summary

When I export the result of a select, the primitive types returned are those of Python and not those of sql, booleans usually come as False/True rather than false/true.

A summary of the issue and the browser/OS environment in which it occurs.

### Steps to Reproduce

1. Select data from any table
2. Export the data.

### Technical details:

* Redash Version: 7.0.0+b17535 (15c815fb)
* Browser/OS: Chrome/Ubuntu
* How did you install Redash:

arikfr:

- [x] Bug Fix

## Description

When downloading CSV version of a query result, it will:

1. Serialize booleans as `true`/`false` (instead of `True`/`False`).
2. Serialize date/time using the organization format for date/time.

Closes #3736, closes #2751.

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -1,12 +1,9 @@
-import cStringIO
-import csv
 import datetime
 import calendar
 import logging
 import time
 import pytz
 
-import xlsxwriter
 from six import python_2_unicode_compatible, text_type
 from sqlalchemy import distinct, or_, and_, UniqueConstraint
 from sqlalchemy.dialects import postgresql
@@ -25,7 +22,7 @@ from redash.destinations import (get_configuration_schema_for_destination_type,
                                  get_destination)
 from redash.metrics import database  # noqa: F401
 from redash.query_runner import (get_configuration_schema_for_query_runner_type,
-                                 get_query_runner)
+                                 get_query_runner, TYPE_BOOLEAN, TYPE_DATE, TYPE_DATETIME)
 from redash.utils import generate_token, json_dumps, json_loads
 from redash.utils.configuration import ConfigurationContainer
 from redash.models.parameterized_query import ParameterizedQuery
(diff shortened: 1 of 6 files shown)


redash/models/__init__.py
```diff
@@ -322,41 +319,6 @@ def store_result(cls, org, data_source, query_hash, query, data, run_time, retri
     def groups(self):
         return self.data_source.groups
 
-    def make_csv_content(self):
```
arikfr: Used the opportunity to move this serialization logic from `redash.models` to `redash.serializers`.

redash/serializers/query_result.py
```diff
@@ -0,0 +1,92 @@
...
+    return fmt.replace('DD', '%d').replace('MM', '%m').replace('YYYY', '%Y').replace('YY', '%y').replace('HH', '%H').replace('mm', '%M').replace('ss', '%s')
+
+
+def serialize_query_result_to_csv(query_result):
+    s = cStringIO.StringIO()
+
+    query_data = json_loads(query_result.data)
+
+    fieldnames = []
+    bool_columns = []
```
rauchy: This feels a tad too scripty for me. I'd prefer to have less details and noise here.
Perhaps this function could be split to several chunks?

1. Calling a function to slice `query_data['columns']` to three lists:
```python
bool_columns, date_columns, datetime_columns = split_columns_by_type(query_data['columns'])
```
2. Handling `csv.DictWriter` stuff and for each row:
2a. Calling a function to format booleans
3b. Calling a function to format dates
4c. Calling a function to format date times

redash/serializers/query_result.py
```diff
@@ -0,0 +1,92 @@
...
+    date_columns = []
+    datetime_columns = []
+
+    for col in query_data['columns']:
+        fieldnames.append(col['name'])
+        if col['type'] == TYPE_BOOLEAN:
```
rauchy: These feel pretty mutually exclusive, so an `elif` might be in place.

tests/serializers/test_query_results.py
```diff
@@ -0,0 +1,50 @@
...
+        {"datetime": None, "bool": None, "date": None},
+    ], 
+    "columns": [
+        {"friendly_name": "bool", "type": "boolean", "name": "bool"}, 
+        {"friendly_name": "date", "type": "datetime", "name": "datetime"},
+        {"friendly_name": "date", "type": "date", "name": "date"}
+    ]
+}
+
+def get_csv_content(factory):
```
rauchy: Why not stuff this inside `CsvSerializationTest` and avoid the passing of `factory`?

redash/models/__init__.py
rauchy: Great! and +1 for tests!

redash/serializers/query_result.py
arikfr: https://github.com/getredash/redash/pull/3841/commits/f26adedb9431c18e18bf1ac901eacc89489775d1

redash/serializers/query_result.py
arikfr: I've started with implementing 1 (https://github.com/getredash/redash/pull/3841/commits/5e0ccb992cba67e18c31c3ec34e08e1118feb99a), but then when I tried to split it further I realized I will need to pass around the lists 🤢or convert this into a class 🤔. So ended up doing this instead: https://github.com/getredash/redash/pull/3841/commits/feda3c9a163ce871e28918855fff4eaf912a6893.

I didn't do this initially as it felt too much, but it does clean the code and makes it easy to add support for new types later.

tests/serializers/test_query_results.py
arikfr: No good reason :) I just didn't realize at first it will need the factory, and then I moved it out it was easier to pass it the factory -> c6ff1388.

PR #3842, opened 2019-05-28
Fix table isn't displayed with date filter

### LINKED ISSUE #3836: Table isn't displayed with Date filter
### Issue Summary

The table breaks if there is a date filter

### Steps to Reproduce

1) with Python executor:
```
add_result_column(result, 'name', 'name', TYPE_STRING)
add_result_column(result, 'month__filter', 'month__filter', TYPE_DATE)
add_result_column(result, 'section__filter', section__filter', TYPE_STRING)


add_result_row(result, {
    'section__filter': 'c', 
    'month__filter': '2019-01-01', 
    'name': 'a'
})
```
2) comment `month__filter` column both in add column and row sections and the table will be shown.

### Technical details:

* Redash Version: master

gabrieldutra:

- [x] Bug Fix

## Description
Fix for #3836 

Fixes #3836 

**Filter**
filter-example

**Multifilter**
multifilter-example

### DIFF
--- a/client/app/components/Filters.jsx
+++ b/client/app/components/Filters.jsx
@@ -90,8 +90,8 @@ export function Filters({ filters, onChange }) {
       <div className="parameter-container container bg-white">
         <div className="row">
           {map(filters, (filter) => {
-            const options = map(filter.values, value => (
-              <Select.Option key={indexOf(filter.values, value)}>{formatValue(value)}</Select.Option>
+            const options = map(filter.values, (value, index) => (
+              <Select.Option key={index}>{formatValue(value)}</Select.Option>
             ));
 
             return (
@@ -106,6 +106,7 @@ export function Filters({ filters, onChange }) {
                       value => ({ key: `${indexOf(filter.values, value)}`, label: formatValue(value) })) :
                     ({ key: `${indexOf(filter.values, filter.current)}`, label: formatValue(filter.current) })}
                   allowClear={filter.multiple}
+                  filterOption={(searchText, option) => includes(option.props.children, searchText)}
                   showSearch
                   onChange={values => onChange(filter, values)}
                 >
--- a/client/app/components/Filters.jsx
+++ b/client/app/components/Filters.jsx
@@ -4,6 +4,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
 import Select from 'antd/lib/select';
+import { clientConfig } from '@/services/auth';
+import { formatDateTime } from '@/filters/datetime';
 
 const ALL_VALUES = '###Redash::Filters::SelectAll###';
 const NONE_VALUES = '###Redash::Filters::Clear###';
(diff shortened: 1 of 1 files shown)


gabrieldutra:
@arikfr there are two commits of mine:
- First one - fix the issues I found using String to compare objects
- Second one - implements my idea of using the keys

Comparing keys approach with the string one, the good part is that it uses `Moment` values for comparison, thus it's more reliable. Also if another type of value appear in the future the string approach may not cover it. The con I see is that I had to introduce the `{ key, label }` syntax from Antd select's `labelInValue` because I couldn't define the `value` otherwise, this made this fix a bit more complex.

So, lmk what you prefer.

kravets-levko:
@gabrieldutra There is some issue with searching values in filter's dropdown - when I open dropdown and try to type something in the input - it sometimes shows matching filter values, and sometimes not (but I know that there are values matching my input).

kravets-levko:
@gabrieldutra  BTW - `Select.Option` has two attributes: `key` and `value`. `key` can be only string or number, but `value` can be anything. If `value` is present, `Select` will use it instead of `key`. I think you can use it for "raw" values (and put formatted as `Option`'s content) - it should simplify your code and fix search.

gabrieldutra:
Thanks for point this out @kravets-levko.

The issue with the search is that it works on the `value || key` attr by default, it will be necessary to modify the `filterOption` prop with a search function :sleepy:

> Select.Option has two attributes: key and value. key can be only string or number, but value can be anything

I'd tried that with `value`, it does "allow" you to send raw values, but it has a `propType` warning:

propTypes.

I wouldn't trust Ant internal API to handle those values without errors internally (e.g: on search)

arikfr:
LGTM. @kravets-levko wdyt?

client/app/components/Filters.jsx
```diff
@@ -75,20 +90,25 @@ export function Filters({ filters, onChange }) {
...
+                  labelInValue
                   className="w-100"
                   mode={filter.multiple ? 'multiple' : 'default'}
-                  value={filter.current}
+                  value={isArray(filter.current) ?
+                    map(filter.current,
+                      value => ({ key: `${indexOf(filter.values, value)}`, label: formatValue(value) })) :
+                    ({ key: `${indexOf(filter.values, filter.current)}`, label: formatValue(filter.current) })}
                   allowClear={filter.multiple}
+                  filterOption={(searchText, option) => includes(option.props.children, searchText)}
```
kravets-levko: One more thing I notices after review: please make search case-insensitive

gabrieldutra:
It seems that Percy still has some weird diffing, at least it became rare. Well, that one looks normal both local and on Preview :slightly_smiling_face:

arikfr:
Good to go? :)

denisov-vlad:
Thanks.

But I think that DATE column should be without hours and minutes. It looks a little bit noisy in multifilter.

arikfr:
@denisov-vlad while I agree, it felt like a reasonable compromise for now. Would you mind opening an issue for this, so we don't forget to address it?

PR #2357, opened 2018-03-01
Add support for TLS connections to redis

myoung34:
The current method does not allow secure redis `rediss://`

```
redash@1dfd88a3e15e:/app$ bin/docker-entrypoint shell                                                                                                
[2018-03-01 18:20:54,568][PID:160][INFO][root] Generating grammar tables from /usr/lib/python2.7/lib2to3/Grammar.txt                                                                                                                     
[2018-03-01 18:20:54,593][PID:160][INFO][root] Generating grammar tables from /usr/lib/python2.7/lib2to3/PatternGrammar.txt                                                                                                                   Traceback (most recent call last):                                                                                                                                                                                                              File "/app/manage.py", line 6, in <module>                                                                                                                                                                                                      from redash.cli import manager                                                                                                                                                                                                              File "/app/redash/__init__.py", line 74, in <module>                                                                                                                                                                                            reset_new_version_status()                                                                                                                                                                                                               
  File "/app/redash/version_check.py", line 34, in reset_new_version_status
    latest_version = get_latest_version()                                                                  
  File "/app/redash/version_check.py", line 40, in get_latest_version                                                                                                                                                                    
    return redis_connection.get(REDIS_KEY)                                                                                                                                                                                               
  File "/usr/local/lib/python2.7/dist-packages/redis/client.py", line 880, in get
    return self.execute_command('GET', name)                                                                                                                                                                                                 
  File "/usr/local/lib/python2.7/dist-packages/redis/client.py", line 578, in execute_command
    connection.send_command(*args)                             
  File "/usr/local/lib/python2.7/dist-packages/redis/connection.py", line 563, in send_command
    self.send_packed_command(self.pack_command(*args))                     
  File "/usr/local/lib/python2.7/dist-packages/redis/connection.py", line 538, in send_packed_command
    self.connect()                                              
  File "/usr/local/lib/python2.7/dist-packages/redis/connection.py", line 446, in connect                                                                                                                                                         self.on_connect()                                                                                                                                                                                                                        
  File "/usr/local/lib/python2.7/dist-packages/redis/connection.py", line 514, in on_connect                                                                                                                                                      if nativestr(self.read_response()) != 'OK':                                                                             
  File "/usr/local/lib/python2.7/dist-packages/redis/connection.py", line 577, in read_response                                                                                                                                              
    response = self._parser.read_response()                                  
  File "/usr/local/lib/python2.7/dist-packages/redis/connection.py", line 238, in read_response                            
[...]

PR #3848, opened 2019-05-29
Add support ssl connections to redis

nason:

- [x] Feature

## Description

Allows TLS connection to Redis via `rediss` scheme https://www.iana.org/assignments/uri-schemes/prov/rediss

This is achieved mostly via dependency upgrades, and some config tweaks. I've been running this on our self-hosted for a few weeks and its been working great.

This follows #2357

I'd appreciate any thoughts or pointers on:
 - how can `ssl_config` be passed in?
 - testing this
 - docs

### DIFF
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -48,7 +48,7 @@ def create_redis_connection():
 
         client = redis.StrictRedis(unix_socket_path=redis_url.path, db=db)
     else:
-        use_ssl = False
+        use_ssl = redis_url.scheme == 'rediss'
 
         if redis_url.scheme == 'rediss':
             use_ssl = True
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -50,9 +50,6 @@ def create_redis_connection():
     else:
         use_ssl = redis_url.scheme == 'rediss'
 
-        if redis_url.scheme == 'rediss':
-            use_ssl = True
-
         if redis_url.path:
             redis_db = redis_url.path[1]
         else:
--- a/redash/__init__.py
+++ b/redash/__init__.py
@@ -48,13 +48,18 @@ def create_redis_connection():
 
         client = redis.StrictRedis(unix_socket_path=redis_url.path, db=db)
     else:
+        use_ssl = False
+
+        if redis_url.scheme == 'rediss':
+            use_ssl = True
+
         if redis_url.path:
             redis_db = redis_url.path[1]
         else:
             redis_db = 0
         # Redis passwords might be quoted with special characters
         redis_password = redis_url.password and urllib.unquote(redis_url.password)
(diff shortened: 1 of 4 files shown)


redash/__init__.py
```diff
@@ -48,13 +48,20 @@ def create_redis_connection():
 
         client = redis.StrictRedis(unix_socket_path=redis_url.path, db=db)
     else:
+        use_ssl = False
```
rauchy:
```suggestion
        use_ssl = redis_url.scheme == 'rediss'
```

redash/worker.py
```diff
@@ -12,9 +12,22 @@
 from redash import create_app, settings
 from redash.metrics import celery as celery_metrics  # noqa
 
+import ssl
+
+redis_ssl_config = None
+if settings.CELERY_BROKER.startswith('rediss'):
+    # TODO make these values configurable
```
rauchy: Kindly do :)

nason: Happy to! Do you have any suggestions on how? Should each of these properties be wired up to new settings? What should the defaults be?

rauchy: We configure pretty much everything in `redash/settings/__init__.py`. Since these configuration bits are tightly coupled, it might be a good idea to move the entire config dict over there, and hydrate it from `os.environ`.

Regarding defaults - I'm really only guessing here, but perhaps a good idea is to default to optional cert and include guidance (documentation or links) on how to setup proper certificates for the connection? (/cc @arikfr?)

redash/settings/__init__.py
```diff
@@ -30,6 +31,12 @@
 CELERY_RESULT_EXPIRES = int(os.environ.get(
     "REDASH_CELERY_RESULT_EXPIRES",
     os.environ.get("REDASH_CELERY_TASK_RESULT_EXPIRES", 3600 * 4)))
+CELERY_SSL_CONFIG = {
```
nason: @rauchy how does this look as far as settings?

I'll try this change out on our self-hosted setup later today

rauchy: Settings look fine to me. However, the build is failing and requires your attention.

arikfr:
Thank you for opening this Pull Request, @nason. Based on the failing tests, it seems that the new code does not work when using no SSL configuration. Can you please check this out?

nason:
On it! I saw the failing tests once I made this configurable but wasn't sure how to read them. I will look into this later today and update.

nason:
I’ve updated with config changes to ensure that ssl config is not passed to celery unless ssl is enabled.

The build and tests seem ok again, and our self hosted deploys with this branch and SSL enabled Redis is still working correctly

arikfr:
👍 thanks!

PR #3850, opened 2019-05-29
Cancel previous execution on parameters value update

gabrieldutra:

- [x] Other

## Description
Hi :slightly_smiling_face: 

I was a bit concerned about slow queries after #3737 (when updating many parameters at once). One thing that can improve this a lot (@ranbena mentioned in the PR) is to cancel the previous execution before running the new one. This idea was simple enough to open a PR directly for discussion from it. The concerns I have regarding it:
- It's necessary that the cancel operation works properly. If it doesn't, the behavior will be no different than it is now. I remember there were issues with MySQL cancelling (afaik a recent PR fixed that), but for PostgreSQL it works nicely.
- Can there be any performance issues with it, like, is it a function that is not supposed to be used that often?
- Are there datasources without cancel operation?

I guess this can be also done backend-wise, but I opted to update it as is to change it only where it could be needed (on parameters value update).

--
--

### DIFF
--- a/client/app/components/dashboards/widget.html
+++ b/client/app/components/dashboards/widget.html
@@ -38,7 +38,7 @@
         </div>
       </div>
       <div class="m-b-10" ng-if="$ctrl.localParametersDefs().length > 0">
-        <parameters parameters="$ctrl.localParametersDefs()" on-values-change="$ctrl.refresh"></parameters>
+        <parameters parameters="$ctrl.localParametersDefs()" on-values-change="$ctrl.cancelPreviousAndRefresh"></parameters>
       </div>
     </div>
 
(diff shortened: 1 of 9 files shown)


arikfr:
> It's necessary that the cancel operation works properly. If it doesn't, the behavior will be no different than it is now.

This is exactly my concern here. I'm not entirely sure about the situation with cancel, but I do know some people reported it causing some trouble. While some of it might be related to the issue that was fixed with MySQL, it's possible there are other data sources that have this issue.

Also, with some data sources the cancel just cancels the Redash job, but not the query execution in the database.

For these reasons, I prefer to try and minimize unneeded executions instead of cancelling them -> https://discuss.redash.io/t/what-to-do-when-you-want-to-update-several-parameters-before-running-a-query/3906

Let's keep this PR around for now, but wait with following up with it until we explore other ideas.

gabrieldutra:
Closing in favor of #3907 :slightly_smiling_face:

Issue #3851, opened 2019-05-30
When a filter has no values it shows -1


arikfr:
### Issue Summary

When a filter has no values it shows -1.

Should probably just say "No Values" or something of that sort.

### Steps to Reproduce

https://redash-preview.netlify.com/queries/151/source

PR #3852, opened 2019-05-30
Disable filter which does not have values

kravets-levko:

- [x] Bug Fix

## Description

If filter does not have options, disable it and show "No values" (see screenshot).

Fixes getredash/redash#3851

image
image


arikfr:
👍 good idea on disabling it.

Issue #3849, opened 2019-05-29
When opening a dashboard, each widget uses its own default value


arikfr:
### Issue Summary

Current behavior: when opening a dashboard, each widget uses its own default value.

Expected behavior: a single value will be picked and used for all widgets.

### Steps to Reproduce

Visit: https://redash-preview.netlify.com/dashboard/wrong-parameter-example. Note that the parameter value is "value1", but the second widget is actually using "value2":

![image](https://user-images.githubusercontent.com/71468/58582289-13101780-825a-11e9-8e1b-6fba73ddea5f.png)

kravets-levko:
@arikfr I'll take it tomorrow

ranbena:
I fiddled with it and can't reproduce it now :/
@arikfr what's the STR?

arikfr:
@ranbena just open the link in the issue, it will reproduce immediately. Once you opened it, it will update the value in the url, so if you refresh the page it will work properly.

PR #3853, opened 2019-05-30
Fix initialization of dashboard-level parameters

kravets-levko:

- [x] Bug Fix

## Description

When creating dashboard-level parameter object, propagate its value to all underlying parameters, and only then try to init from URL.

Fixes getredash/redash#3849


arikfr:
👍

PR #3866, opened 2019-06-02
Remove legacy session identifier support

rauchy:

- [x] Other

## Description
Users who logged in prior to https://github.com/getredash/redash/pull/3174 going live have their (integer) user_id as their session user identifier.
This no longer needs to be supported.

#3174

### DIFF
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -43,15 +43,14 @@ def sign(key, path, expires):
 def load_user(user_id_with_identity):
     org = current_org._get_current_object()
 
-    user_id, _ = user_id_with_identity.split("-")
-
     try:
+        user_id, _ = user_id_with_identity.split("-")
         user = models.User.get_by_id_and_org(user_id, org)
         if user.is_disabled or user.get_id() != user_id_with_identity:
             return None
 
         return user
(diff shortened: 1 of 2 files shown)


redash/authentication/__init__.py
```diff
@@ -43,15 +43,14 @@ def sign(key, path, expires):
...
-
     try:
+        user_id, _ = user_id_with_identity.split("-")
         user = models.User.get_by_id_and_org(user_id, org)
         if user.is_disabled or user.get_id() != user_id_with_identity:
             return None
 
         return user
-    except models.NoResultFound:
+    except (models.NoResultFound, Exception):
```
arikfr: Maybe we should only catch `SyntaxError` instead of `Exception` here?

rauchy: Well it could be either a `ValueError` when `user_id_with_identity` isn't hyphenated or a `AttributeError` when `user_id_with_identity` is an integer, so I thought I'd just catch all.

arikfr: My concern here is that it might swallow some unrelated exceptions and make debugging future issues harder. Aside from this there is not much harm in catching `Exception` here.

arikfr:
👍

PR #3873, opened 2019-06-02
Cypress: Fix date parameters false positive

gabrieldutra:

- [x] Other

## Description
[Example test results with this issue](https://dashboard.cypress.io/#/projects/924cka/runs/444/specs)

I did run this fix a few times and got no fail on tests, but I guess I'll just know it really works after merge. 

My guess is that the Save action to close the dialog didn't occur due to a sudden change in the dialog (the checkbox rendering). The solution I could think: click on the checkbox as well, this way it will wait for it to be rendered.

@ranbena I tried the `cy.wait` to see if the false positive would happen, didn't get it in my tests as well. In any case, if the fix without `cy.wait` works, I prefer it.

--

--


gabrieldutra:
I wanna check if this will actually solve the issue and also Cypress is probably annoying a few people at this point, so I'm merging it :slightly_smiling_face:

PR #3875, opened 2019-06-03
Restrict dynamic-table internal scroll only when pagination appears

### LINKED ISSUE #3859: Autoheight widgets miscalculating height
### Issue Summary

Comparing version 7 with current master, it's apparent that autoheight table visualization widgets are higher now by 1 row.

Version 7 | Master branch
----------|---------------
Screen Shot 2019-05-30 at 19 22 17 | Screen Shot 2019-05-30 at 19 22 02

### How come?
#3830 added a 10px bottom padding to tables but with -10px margin to compensate*.
Thing is, autoheight is calculated with `getBoundingClientRect()` which understandably includes padding but not margin, thus adding 10px to the visualization height - resulting in an additional grid row.

\* Mac only - therefore not detected by e2e tests 😓

ranbena:
- [x] Bug Fix

## Description
Fixes #3859.
1. Reverted #3830 - no more autoheight miscalculation.
2. Restricted internal scrolling only when pagination is available - so the Mac browser scroll bug won't manifest in this case

Before|After
-------|--------
Screen Shot 2019-06-03 at 16 33 57 | Screen Shot 2019-06-03 at 16 34 00

### DIFF
--- a/client/app/components/app-view/index.js
+++ b/client/app/components/app-view/index.js
@@ -37,10 +37,6 @@ class AppViewComponent {
     this.layout = layouts.defaultSignedOut;
     this.handler = handler;
 
-    // remove when fix lands
-    // https://bugs.chromium.org/p/chromium/issues/detail?id=914844#c36
-    $rootScope.platform = navigator.platform;
-
     $rootScope.$on('$routeChangeStart', (event, route) => {
       this.handler.reset();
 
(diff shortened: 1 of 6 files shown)


arikfr:
Are you sure about the logic?

This happens when there is _no pagination_. Maybe even only when there is no pagination.

arikfr:
What if the containing div for the viz wasn’t limited to the size of the table: would the scroll still overlap?

I noticed it happening in the "Show API Key" dialog:

image

But on a different page, I saw it only happen with the bottom block but not the upper one. Any idea why this behavior will not be consistent?

Another observation: when the scrollbar overlaps the content, it doesn't not automatically disappear. While when it doesn't overlap, when you move the mouse away, it does disappear (as supposed to).

ranbena:
> This happens when there is _no pagination_. Maybe even only when there is no pagination.

ranbena:
@arikfr @kravets-levko I updated the PR

Issue #3820, opened 2019-05-22
Query results auto round numbers


flolivaud:
### Issue Summary

Query results auto round numbers

### Steps to Reproduce
With Query results datasource :
`SELECT 2.4 as value;`
Will return `2`

`SELECT 2.5 as value;`
Will return `3`

arikfr:
What type of data source did you run this query with?

Arik

flolivaud:
@arikfr I use the url data sources. But to reproduce the bug you don't need any other datasource.

arikfr:
@flolivaud it does not happen with every data source. It's a data source specific issue rather than a global one, therefore important to know which one is it.

Here's an example where it does not happen: https://redash-preview.netlify.com/queries/156/source

In your case: can you share the `columns` part of your URL response? (and mention which column included the number you had problem with)

arikfr:
Resolved in #3913.

PR #3876, opened 2019-06-03
check float if scale > 0 in snowflake query result

wudi-ayuan:

- [ ] Bug Fix

## Description
The simple `select 2.5` will return result 3 when the datasourc is snowflake. Redahs directly use the inferred datatype given by datasource, but it's not totally accurate. However, snowflake also provide precision and scale of query result. If the data is numerical, we check whether scale is greater than 0 to determine Float or Integer. 

https://github.com/getredash/redash/issues/3820
This rounding issues appeared in snowflake data source as well.


arikfr:
Thanks!

PR #3877, opened 2019-06-04
Authorize according to API key (if given) over cookies

### LINKED ISSUE #3770: API Key authentication should take precedence over cookies
Currently when you're logged in as a user, you can't use embeds or shared dashboards for which you do not have access -- even if you provide the API key. This is because the current user is being defined by the session cookie and the API key is being ignored.

We need to change this behavior so the API key takes precedence. From a quick look at [Flask-Login](https://flask-login.readthedocs.io)'s documentation, it seems that it calls `user_loader` first and only if this fails will try the `requets_loader`.

Unless this is configurable, we might need to change `user_loader` to always return `None` and implement our own logic in `request_loader`.

rauchy:

- [x] Bug Fix

Fixes #3770

### DIFF
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -41,6 +41,10 @@ def sign(key, path, expires):
 
 @login_manager.user_loader
 def load_user(user_id_with_identity):
+    user = api_key_load_user_from_request(request)
+    if user:
+        return user
+
     org = current_org._get_current_object()
 
     try:
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -43,15 +43,14 @@ def sign(key, path, expires):
 def load_user(user_id_with_identity):
     org = current_org._get_current_object()
 
-    user_id, _ = user_id_with_identity.split("-")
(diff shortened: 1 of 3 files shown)


redash/authentication/__init__.py
```diff
@@ -41,6 +41,10 @@ def sign(key, path, expires):
 
 @login_manager.user_loader
 def load_user(user_id_with_identity):
+    user = api_key_load_user_from_request(request)
+    if user:
+        return user
```
arikfr: This feels almost as too simple 😆

rauchy: I've battled flask_login for 8 hours, trying to get it to cooperate with `user_loader` and `request_loader` until this hit me 🤦‍♂️

arikfr:
`login_user` doesn't have any side effects like setting a cookie or anything like this, right?

rauchy:
@arikfr I CURLed this around, and there are no `Set-Cookie` headers in the response when using an apikey querystring parameters, so I believe we're safe.

PR #3884, opened 2019-06-06
Force a Choose account step for Google OAuth

kravets-levko:

- [x] Other

## Description

Force show a Choose account step for Google OAuth (previously this feature didn't work for me - I have two accounts, and it was using only active (default? most recently used?) one). Small side-effect: with this fix it will show Choose account step even if there is the only account available (probably not a big deal - at least, users will be 100% sure which account will be used).


rauchy:
I don't really know if most users have 1 account or more, but I guess adding another click for everyone is a reasonably fair price to pay in order to have multiple-account users being able to log in. 👍

PR #3886, opened 2019-06-06
Fix empty values sent in dynamic form

### LINKED ISSUE #3885: Datasource is created with empty SSL mode
### Issue Summary

When you create a new PG datasource, modify the default SSL mode and delete it before submission, the SSL mode value sent over is "". It should be absent from the request. This is most likely not specific to this scenario, but since I didn't investigate yet, I'm leaving this issue as concrete as possible and not generalizing.

### Steps to Reproduce

1. Create a new Postgres data source
2. Fill in whatever you'd like as SSL mode
3. Clear the text box
4. Submit the form - the SSL mode sent is "".

### Technical details:

* Redash Version:
* Browser/OS:
* How did you install Redash:

rauchy:

- [x] Bug Fix

## Description
Fixes #3885

### DIFF
--- a/client/app/components/dynamic-form/DynamicForm.jsx
+++ b/client/app/components/dynamic-form/DynamicForm.jsx
@@ -77,10 +77,12 @@ class DynamicForm extends React.Component {
   handleSubmit = (e) => {
     this.setState({ isSubmitting: true });
     e.preventDefault();
+    const notEmpty = (value, key) => !(this.props.fields.find(f => f.name === key).initialValue === undefined && value === ''); // rejects empty values, unless they are intentionally set to empty
+
     this.props.form.validateFieldsAndScroll((err, values) => {
       if (!err) {
         this.props.onSubmit(
-          pickBy(values, v => v !== ''),
+          pickBy(values, notEmpty),
           (msg) => {
             const { setFieldsValue, getFieldsValue } = this.props.form;
             this.setState({ isSubmitting: false });
--- a/client/app/components/dynamic-form/DynamicForm.jsx
+++ b/client/app/components/dynamic-form/DynamicForm.jsx
@@ -81,7 +81,7 @@ class DynamicForm extends React.Component {
     this.props.form.validateFieldsAndScroll((err, values) => {
       Object.entries(values).forEach(([key, value]) => {
         const initialValue = this.props.fields.find(f => f.name === key).initialValue;
-        if ((initialValue === undefined || initialValue === '') && value === '') {
+        if ((initialValue === null || initialValue === undefined || initialValue === '') && value === '') {
           values[key] = null;
         }
       });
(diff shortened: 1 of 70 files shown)


gabrieldutra:
I've searched here, and if this is relevant for OrgSettings as well, the same fix can be applied here:

https://github.com/getredash/redash/blob/5b780ac46013feef23ede7356532f892655d35c1/client/app/pages/settings/OrganizationSettings.jsx#L45

rauchy:
@gabrieldutra good catch! 042ef09 should remedy that. I... think?

gabrieldutra:
> @gabrieldutra good catch! [042ef09](https://github.com/getredash/redash/commit/042ef09fcc2f7a226ac117c011d6262fc038b238) should remedy that. I... think?

It will indeed remedy that, but the empty values will still exist when you delete an existing input data. BTW, what actually happens with the empty strings? Is it more for data consistency or there are issues caused by it?

rauchy:
> the empty values will still exist when you delete existing input data

That's exactly the point. If you delete existing data, you _want_ an empty string to be submitted.

gabrieldutra:
That's probably the easiest option to handle it then.

Deep down I think that ideally empty strings should be treated as "set an empty value" in any case. If that's easy to do and has no apparent problems, I'd add a check for this in the backend just in case.

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -77,10 +77,12 @@ class DynamicForm extends React.Component {
   handleSubmit = (e) => {
     this.setState({ isSubmitting: true });
     e.preventDefault();
+    const notEmpty = (value, key) => !(this.props.fields.find(f => f.name === key).initialValue === undefined && value === ''); // rejects empty values, unless they are intentionally set to empty
```
arikfr: So if some field had a value before, but now I clear it out, we will still send an empty value?

rauchy: Yes, an empty string will be sent.

arikfr: This is a problem, because some values will work with a null but not an empty string. In most cases I guess we handled it (like the Presto data source), but not sure if in all of them.

Maybe set it to null in such case?

Another option, is to remove these values on the backend when merging with old params. Or maybe not merge at all?

rauchy: > Maybe set it to null in such case?
Seems to work well.

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -77,7 +77,15 @@ class DynamicForm extends React.Component {
   handleSubmit = (e) => {
     this.setState({ isSubmitting: true });
     e.preventDefault();
+
     this.props.form.validateFieldsAndScroll((err, values) => {
+      Object.entries(values).forEach(([key, value]) => {
+        const initialValue = this.props.fields.find(f => f.name === key).initialValue;
+        if ((initialValue === undefined || initialValue === '') && value === '') {
```
arikfr: What if `initialValue` was `null`? Maybe some `isEmpty` might work better here?

rauchy: I decided to be a bit verbose and check for `null` explicitly because `isEmpty` has issues with numbers.

PR #3894, opened 2019-06-10
Remove schema after deleting data source

### LINKED ISSUE #3893: When deleting a data source, it should remove the cached schema from Redis
When deleting a data source ([DataSource#delete](https://github.com/getredash/redash/blob/dda75cce24437d5c308b2ca8c42acaa164a62667/redash/models/__init__.py#L138-L143)), we need to remove the cached schema from Redis as well.

(The schema is being in the [DataSource#get_schema](https://github.com/getredash/redash/blob/dda75cce24437d5c308b2ca8c42acaa164a62667/redash/models/__init__.py#L145-L160) method.)

rauchy:

- [x] Bug Fix

Fixes #3893

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -164,22 +164,23 @@ class DataSource(BelongsToOrgMixin, db.Model):
     def _schema_key(self):
         return "data_source:schema:{}".format(self.id)
 
+    @property
     def _pause_key(self):
         return 'ds:{}:pause'.format(self.id)
 
     @property
     def paused(self):
-        return redis_connection.exists(self._pause_key())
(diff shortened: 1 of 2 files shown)


redash/models/__init__.py
```diff
@@ -140,41 +140,47 @@ def delete(self):
...
         else:
             schema = json_loads(cache)
 
         return schema
 
+    @property
+    def _schema_key(self):
+        return "data_source:schema:{}".format(self.id)
+
+    @property
```
rauchy: Unrelated to this PR, but the similarity forced me.

arikfr:
👍

PR #3887, opened 2019-06-11
Query Results: querying a column with a dictionary or array fails

rauchy:
### Issue Summary

When using the Query Results data source to query from a query result that one of the columns if of type dictionary of array, it fails with a cryptic message.

### Steps to Reproduce

Example query: https://redash-preview.netlify.com/queries/154/source. When you execute it you get:

> Error binding parameter 0 - probably unsupported type.

For starters, we should show a friendlier message, explaining which column had this issue. 

As an improved behavior we can convert the value into JSON, so the user can further query it using SQLite's JSON functions.


arikfr:
👍

PR #3898, opened 2019-06-12
Celery doesn't auto reload in development

rauchy:
### Issue Summary

Using the standard Docker development setup, it seems that code changes are not picked up by `redash-worker`.

### Steps to Reproduce

1. Start `docker-compose up`
2. Wreak havoc on some Python file.
3. Everything works!

### Technical details:

* Redash Version:
* Browser/OS:
* How did you install Redash:


docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration file is for the **development** setup.
 # For a production example please refer to setup/docker-compose.yml.
-version: '3'
+version: '3.7'
```
arikfr: Was this necessary?

bin/docker-entrypoint
```diff
@@ -14,10 +14,14 @@ scheduler() {
   WORKERS_COUNT=${WORKERS_COUNT:-1}
   QUEUES=${QUEUES:-celery}
   SCHEDULE_DB=${SCHEDULE_DB:-celerybeat-schedule}
+  if [[ -n "$AUTO_RELOAD" ]]; then
+    WATCHER="watchmedo auto-restart --directory=./ --pattern=*.py --recursive --"
```
arikfr: Maybe only watch the code library (`redash/`) so it doesn't reload on tests?

arikfr: Maybe we should add a new entrypoint for dev-celery to avoid any confusion or issues that might result here?

docker-compose.yml
rauchy: The [long volume syntax](https://docs.docker.com/compose/compose-file/#long-syntax-3), required in order to specify the volume type which was required to get watchdog to pick up host changes, was introduced in version 3.2. We could downgrade it to 3.2 if you prefer to support docker-compose 17.0.4, I was just following docs.

rauchy:
> did you check that Flask still reloads after adding watchdog? :)

Of course :) It does.

bin/docker-entrypoint
rauchy: See efef063 and df2eb1d.

bin/docker-entrypoint
```diff
@@ -17,7 +17,17 @@ scheduler() {
 
   echo "Starting scheduler and $WORKERS_COUNT workers for queues: $QUEUES..."
 
-  exec /usr/local/bin/celery worker --app=redash.worker --beat -s$SCHEDULE_DB -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 -Ofair
+  /usr/local/bin/celery worker --app=redash.worker --beat -s$SCHEDULE_DB -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 -Ofair
```
jezdez: @rauchy Is this intentionally not using exec to run the worker in production?

jezdez: To be clear, this change leads to the celery worker not receiving signals as explained in https://hynek.me/articles/docker-signals/

rauchy: No, this is a mistake. I initially used the `scheduler` command for both dev and production, but @arikfr suggested I split them and I left this out. Thanks for spotting this!

rauchy:
```suggestion
exec /usr/local/bin/celery worker --app=redash.worker --beat -s$SCHEDULE_DB -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 -Ofair
```

jezdez: Woot!

bin/docker-entrypoint
```diff
@@ -17,7 +17,17 @@ scheduler() {
 
   echo "Starting scheduler and $WORKERS_COUNT workers for queues: $QUEUES..."
 
-  exec /usr/local/bin/celery worker --app=redash.worker --beat -s$SCHEDULE_DB -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 -Ofair
+exec /usr/local/bin/celery worker --app=redash.worker --beat -s$SCHEDULE_DB -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 -Ofair
```
rauchy:
```suggestion
  exec /usr/local/bin/celery worker --app=redash.worker --beat -s$SCHEDULE_DB -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 -Ofair
```

bin/docker-entrypoint
```diff
@@ -41,6 +51,7 @@ help() {
   echo "server -- start Redash server (with gunicorn)"
   echo "worker -- start Celery worker"
   echo "scheduler -- start Celery worker with a beat (scheduler) process"
+  echo "dev_scheduler -- start Celery worker with a beat (scheduler) process which picks up code changes and reloads"
```
arikfr: Let's name this `dev_worker` as it better describes what it does.

bin/docker-entrypoint
```diff
@@ -20,6 +20,16 @@ scheduler() {
   exec /usr/local/bin/celery worker --app=redash.worker --beat -s$SCHEDULE_DB -c$WORKERS_COUNT -Q$QUEUES -linfo --max-tasks-per-child=10 -Ofair
 }
 
+dev_scheduler() {
+  WORKERS_COUNT=${WORKERS_COUNT:-1}
+  QUEUES=${QUEUES:-celery}
```
arikfr: As we have a dedicated dev worker entrypoint, we can have defaults that are inline with it:

WORKERS_COUNT = 2
QUEUES = "queries,scheduled_queries,celery,schemas"

bin/docker-entrypoint
rauchy: f5f707e

bin/docker-entrypoint
rauchy: 57e51a5

PR #3899, opened 2019-06-12
Fix OverflowError on Celery worker

guyco33:

- [x] Bug Fix

## Description
`next_iteration += datetime.timedelta(minutes=2**failures)` start failing when number of failures is larger then 31 and celery worker stops refreshing queries

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -354,7 +354,10 @@ def should_schedule_next(previous_iteration, now, interval, time=None, day_of_we
         next_iteration = (previous_iteration + datetime.timedelta(days=days_delay) +
                           datetime.timedelta(days=days_to_add)).replace(hour=hour, minute=minute)
     if failures:
-        next_iteration += datetime.timedelta(minutes=2**failures)
+        try:
+            next_iteration += datetime.timedelta(minutes=2**failures)
+        except OverflowError:
+            return False
     return now > next_iteration
 
 
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -131,6 +131,11 @@ class ShouldScheduleNextTest(TestCase):
         self.assertFalse(models.should_schedule_next(two_hours_ago, now,
                                                      "3600", failures=10))
 
+    def test_next_iteration_overflow(self):
+        now = utcnow()
+        two_hours_ago = now - datetime.timedelta(hours=2)
+        self.assertFalse(models.should_schedule_next(two_hours_ago, now, "3600", failures=32))
+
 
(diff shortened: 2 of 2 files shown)


redash/models/__init__.py
```diff
@@ -354,7 +354,10 @@ def should_schedule_next(previous_iteration, now, interval, time=None, day_of_we
         next_iteration = (previous_iteration + datetime.timedelta(days=days_delay) +
                           datetime.timedelta(days=days_to_add)).replace(hour=hour, minute=minute)
     if failures:
-        next_iteration += datetime.timedelta(minutes=2**failures)
+        try:
```
rauchy: What you're saying is basically equivalent to "after failure 31, the next scheduled time is soooo far ahead in the future that there's really no point in scheduling it, meaning that the `should_schedule_next` function should simply return `False`, instead of a magic date in the post apocalyptic world. 😄 

How about this?

```suggestion
        try:
            next_iteration += datetime.timedelta(minutes=2**failures)
            return now > next_iteration
        except OverflowError:
            return False            
```

guyco33: Thanks. Make sense

rauchy: Seems like the suggestion wasn't applied correctly. Not sure Github is supporting multi-line suggestions ATM.

image

guyco33: `return now > next_iteration` should be outside the `if failures` to cover also the non failure cases.
I fixed it accordingly

tests/test_models.py
```diff
@@ -130,7 +130,7 @@ def test_backoff(self):
                                                     failures=5))
         self.assertFalse(models.should_schedule_next(two_hours_ago, now,
                                                      "3600", failures=10))
-
+        self.assertFalse(models.should_schedule_next(two_hours_ago, now, "3600", failures=32))
```
rauchy: @guyco33 sorry for being a pain in the ass about this, but I'm a firm believer in one assert per test, and specifically, one logical assert per test.

The two assertions above the one you added are kind of related - they test the flip side of the same conditional, and belong in the same test (although this is debatable). The assertion you added feels like it tests a different concept, and I believe it should be separated.

Also - in the absence self-explanatory code, it's the tests' job to provide documentation, and just asserting "false" when failures=32 feels like a magic number with magic behavior. A dedicated test, with an explanatory test name, would really help here.

guyco33: @rauchy I totally agree with you, especially when the tests cover special behaviour of magic numbers and unique corner cases.  
I will pick a proper name for a new assertion and will update the PR accordingly

rauchy:
Thanks @guyco33!

PR #3903, opened 2019-06-16
Increase celery job timeout

ranbena:
- [x] Bug Fix

## Description
With the default 4 sec timeout, I'm experiencing endless repeated job restarts that kill my cpu 🥵
Increased to 10 sec and it doesn't happen no more.

### DIFF
--- a/redash/metrics/celery.py
+++ b/redash/metrics/celery.py
@@ -4,6 +4,9 @@ import logging
 import socket
 import time
 
+from celery.concurrency import asynpool
+asynpool.PROC_ALIVE_TIMEOUT = 10.0
+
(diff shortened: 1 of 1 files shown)


deecay:
How aboit exposing this conf to env? The necessary time depends heavily on the machine you have. I usually set this value to 30 in my 2gb ram ec2.

arikfr:
@deecay I didn't realize it can be so slow :-\ 

In this case, then yes -- we should make it configurable. Do you want to take this?

deecay:
@arikfr sure I can.

arikfr:
@deecay great, thanks! I merged this anyway, as it's an improvement over current state and so you have what to base your version on.

PR #3904, opened 2019-06-16
Add Dynamic Values to Date and Date Range Parameters

### LINKED ISSUE #3009: Add more predefined options to Date and Date Range parameters

gabrieldutra:

- [x] Feature

## Description
Allow Dynamic Values to Date and Date Range Parameters :tada:.

See preview below for details :)

Closes #3009

dynamic-values-apply-all


gabrieldutra:
Idea of implementation updated considering [Arik's comment](https://github.com/getredash/redash/issues/3009#issuecomment-502448951).

- Dynamic Values are just like regular values, but with a standard syntax (current strings with "d_" prefix);
- There will be **Date Range** Dynamic options and **Date** Dynamic options. Date Ranges can be composed by a static or dynamic **date** values or have a Dynamic Date Range. This will make some more specific cases possible;
- Dynamic Date Values can hold a mapped string (e.g: "d_yesterday") or a "delta" datetime number (e.g: "d_-[7days in ms]"). This will cover cases where you want custom values and cases where the date cannot be expressed as deltas (e.g: last month);

A possible refactor that I was thinking is that at some point parameters can be rewritten taking advantage of inheritance (should reduce complexity for the methods like getValue, setValue, toUrlParams, etc);

ranbena:
Great to see it in action :)

gabrieldutra:
@ranbena @arikfr finally I could sit down and work on this :slightly_smiling_face:, the fist part of the feature already has its shape:

dynamic-dates-1 ([preview](https://deploy-preview-3904--redash-preview.netlify.com/queries/157))

I'm pretty happy with the UX/UI result :)

Currently the Dynamic information is saved in the Parameter `value` as a string (e.g: "d_last_week"), can there be an issue backend-wise with queries saved with it? (it seemed right to put it there, but I can move it elsewhere if so).

The next is to implement a Custom dialog (should make this reach all of its potential) and then it's a matter of final touches.

One note is that I reverted the [DateTime Parameter apply only onOk or close thing](https://github.com/getredash/redash/pull/3737#issuecomment-489608357) so this one would be easier (I think it won't be a problem considering #3907).

ranbena:
> I'm pretty happy with the UX/UI result :)

Me too 👍 

1. Some UI refinements for size and pos:
   ```less
   .dynamic-button {
     height: 100%;
  
     &:after {
       height: 19px;
     }
 
    .ant-dropdown-trigger {
       height: 100%;
     }
   }
   ```

2. I think, if it's possible, to get the "clear" button in there for both static and dynamic selected values (requiring an x offset).
   
   Screen Shot 2019-06-29 at 17 08 49

   * If so, the "Static value" option can be omitted.
   * If not, I think moving it to last is better. And some other suggested changes (selected state, divider, back button, lighter text color)

   `   Screen Shot 2019-06-29 at 16 30 24

> Currently the Dynamic information is saved in the Parameter value as a string.

Probably need to make scheduled queries convert dynamic to static values upon execution (backend).

> The next is to implement a "Custom option" dialog (should make this reach all of its potential) and then it's a matter of final touches.

You sure it's a must? I would abstain from it at this point.

client/app/components/dynamic-parameters/DynamicButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+        <Menu.Item key={index}>
+          {option.name} {option.label && (
+            <em>{isFunction(option.label) ? option.label() : option.label}</em>
+          )}
+        </Menu.Item>
+      ))}
+      {enabled && <Menu.Divider />}
+      {enabled && (
+        <Menu.Item>
+          <i className="fa fa-arrow-left" /> <Text type="secondary">Back to Static Value</Text>
```
ranbena: @gabrieldutra no luck with `allowClear`?

client/app/components/dynamic-parameters/DynamicParameters.less
```diff
@@ -0,0 +1,18 @@
+.dynamic-value {
+  & ::placeholder, & .dynamic-icon {
+    color: #1890ff !important
+  }
+
+  .ant-calendar-range-picker-separator,
+  .ant-calendar-range-picker-input:not(:first-child) {
+    display: none;
+  }
```
ranbena: To prevent text clipping
Screen Shot 2019-07-01 at 8 31 47


add this:
```less
.dynamic-value {
  .ant-calendar-range-picker-input {
    width: auto;
    text-align: left;
  }
}
```

client/app/components/dynamic-parameters/DateRangeParameter.jsx
```diff
@@ -0,0 +1,111 @@
...
+  { name: 'Last week',
+    value: 'd_last_week',
+    label: () => DYNAMIC_DATE_RANGES.last_week.value()[0].format('MMM D') + ' - ' +
+                 DYNAMIC_DATE_RANGES.last_week.value()[1].format('MMM D') },
+  { name: 'Last month', value: 'd_last_month', label: () => DYNAMIC_DATE_RANGES.last_month.value()[0].format('MMMM') },
+  { name: 'Last year', value: 'd_last_year', label: () => DYNAMIC_DATE_RANGES.last_year.value()[0].format('YYYY') },
+  { name: 'Last 7 days',
+    value: 'd_last_7_days',
+    label: () => DYNAMIC_DATE_RANGES.last_7_days.value()[0].format('MMM D') + ' - Today' },
+];
```
ranbena: Wdyt of adding "This X" as well?

Screen Shot 2019-07-01 at 8 43 03

client/app/components/dynamic-parameters/DynamicParameters.less
gabrieldutra: `width: 100%` keeps some flexibility while fixing this too :)

ranbena:
> The clear button was actually fine in this implementation, but there was a different issue: when there's no value the clear button is not rendered and since we need the placeholder there... There was probably a way to workaround that with css, but I preferred to go with option 2.

But that does mean that there's no way to clear a static value, no?
Perhaps a compromise with `allowClear={!!this.state.dynamicValue}`?

gabrieldutra:
> But that does mean that there's no way to clear a static value, no?

I wasn't considering this as required :sweat_smile:, anyway I updated implementation and put the clear button back in there :)

gabrieldutra:
To give an update on this one. My initial idea for Custom Option was to allow options like this:

composed-custom-option
Date Ranges could be expressed as composed Dates, so a Dynamic date and a Static date could be selected for example.

Turns out it would make it more complex without much gain. So I'm thinking about adding a simple "Last [date interval]" handler for Custom Option and later we can improve it based on feedback. So what's missing to wrap this up:
- [ ] Add this simple Custom Option (shouldn't be that hard)
- [ ] Treat weird values (can happen when changing Parameter type)
- [ ] Update tests

@ranbena I'm also waiting for #3907 as it simplifies this one and Tests will need to be updated, I'll ping you when I'm about to be done, so we can architect this delivery :)

arikfr:
>  Treat weird values (can happen when changing Parameter type)

Maybe we should clear values when changing parameter type? It might be annoying in some cases, so maybe we should clear them if they don't pass some validation?

arikfr:
@gabrieldutra I just interacted with the deploy preview and I think we don't need to do any other changes to the functionality as it is right now 💯  It's perfect as is and we better ship this improvement and not delay it in favor of adding more abilities. Specially when we don't know yet what further abilities users might want.

Of course if there are known bugs _with current feature_, then let's address them. But otherwise let's prepare to ship this 🚀 :)

gabrieldutra:
> Maybe we should clear values when changing parameter type? It might be annoying in some cases, so maybe we should clear them if they don't pass some validation?

The weird error I was getting specifically for Dynamic Values was actually the Moment PropType not handling `undefined` as a value.

I do see some probably annoying cases when changing the parameter type, but not associated to this feature.

To ship this there's one thing left: I've changed DateTime Inputs to work with `onSelect` instead of the `OK` click or `onClose` (basically reverted [this](https://github.com/getredash/redash/pull/3737#issuecomment-489608357)), this made implementation easier, but need to see #3907 status.

Regarding bugs, looks pretty stable, but I'll run another series of tests.

gabrieldutra:
@arikfr I probably should remove the "Today/Now" option from Parameter Settings, should I remove its support on code too? The new feature will be quite easy to be noticed :thinking:

arikfr:
> I probably should remove the "Today/Now" option from Parameter Settings

Yes.

> , should I remove its support on code too? The new feature will be quite easy to be noticed 🤔

How much will it complicate the code to trigger dynamic value of today, when we have this option enabled?

gabrieldutra:
> How much will it complicate the code to trigger dynamic value of today, when we have this option enabled?

About more ~ 5 LOC, I think we can keep it for now with a comment and remove in another iteration, this way we don't remove the support right away.

ranbena:
> About more ~ 5 LOC, I think we can keep it for now with a comment and remove in another iteration, this way we don't remove the support right away.

In many cases the temporary becomes the permanent 😇
I might be going overboard with this, but perhaps sth like this?

Screen Shot 2019-07-17 at 20 53 42

(alert close probably implemented with localStorage)

gabrieldutra:
> In many cases the temporary becomes the permanent 
I might be going overboard with this, but perhaps sth like this?

That's true :thinking:, I actually don't think the missing option on Parameter Settings will be an issue, the new option feels easy to be discovered and I think not many people were using that. My concern is more related not to break existing Dashboards/Queries that use that.

client/cypress/integration/query/parameter_spec.js
```diff
@@ -265,14 +277,11 @@ describe('Parameter', () => {
       cy.getByTestId('ParameterApplyButton')
         .click();
 
-      cy.get('@Input').then(($input) => {
-        const now = Cypress.moment($input.val(), 'DD/MM/YY HH:mm');
-        cy.getByTestId('DynamicTable')
-          .should('contain', now.format('YYYY-MM-15 HH:mm'));
-      });
+      cy.getByTestId('DynamicTable')
+        .should('contain', Cypress.moment(this.now).format('YYYY-MM-15 HH:mm'));
```
gabrieldutra: @ranbena and @rauchy I updated a bit the Cypress date assertion, `Cypress.moment()` is not affected by `cy.clock()`, so I put the time selected as an alias and took it from `this.now`.

gabrieldutra:
Updated with Apply Changes & Cypress tests added.

dynamic-values-apply-all

---

@arikfr, @ranbena, @kravets-levko this is ready for review.

You can explore the functionality here -> https://deploy-preview-3904--redash-preview.netlify.com/queries/157 (or create another query :grin:).

Please LMK if you find any issues.

client/app/components/ParameterValueInput.jsx
```diff
@@ -189,12 +147,12 @@ export class ParameterValueInput extends React.Component {
...
-      case 'date': return this.renderDateInput();
-      case 'datetime-range-with-seconds': return this.renderDateTimeRangeWithSecondsInput();
-      case 'datetime-range': return this.renderDateTimeRangeInput();
-      case 'date-range': return this.renderDateRangeInput();
+      case 'datetime-with-seconds': return this.renderDateParameter();
+      case 'datetime-local': return this.renderDateParameter();
+      case 'date': return this.renderDateParameter();
+      case 'datetime-range-with-seconds': return this.renderDateRangeParameter();
+      case 'datetime-range': return this.renderDateRangeParameter();
+      case 'date-range': return this.renderDateRangeParameter();
```
ranbena: Nicer syntax?

```js
switch (type) {
  case 'date': 
  case 'datetime-local':
  case 'datetime-with-seconds': return this.renderDateParameter();
  case 'date-range':
  case 'datetime-range':
  case 'datetime-range-with-seconds': return this.renderDateRangeParameter();
```

client/app/components/dynamic-parameters/DateRangeParameter.jsx
```diff
@@ -0,0 +1,115 @@
...
+    } else {
+      onSelect(dynamicValue.value);
+    }
+  };
+
+  render() {
+    const { type, value, parameter, onSelect, className } = this.props;
+    const isDateTimeRange = includes(type, 'datetime-range');
+    const hasDynamicValue = isDynamicDateRange(value);
+    const options = isDateTimeRange ? DYNAMIC_DATETIME_OPTIONS : DYNAMIC_DATE_OPTIONS;
```
ranbena: Is there a difference between "Today / Yesterday" in Date and that in DateRange? (in terms of sql query perhaps?)

client/app/components/dynamic-parameters/DynamicParameters.less
```diff
@@ -0,0 +1,25 @@
+.dynamic-value {
+  & ::placeholder, & .dynamic-icon {
+    color: #1890ff !important
+  }
```
ranbena: @arikfr @gabrieldutra Distractingly blue?
Screen Shot 2019-07-18 at 8 19 41

Some other options:

Option 1
Screen Shot 2019-07-18 at 8 16 39

Option 2
Screen Shot 2019-07-18 at 8 17 54

Option 3
Screen Shot 2019-07-18 at 8 25 33

I think Option 3. Wdyt?

package.json
```diff
@@ -48,7 +48,7 @@
     "angular-ui-bootstrap": "^2.5.0",
     "angular-vs-repeat": "^1.1.7",
     "angular2react": "^3.0.2",
-    "antd": "^3.12.3",
+    "antd": "^3.19.7",
```
ranbena: Woohoo!

client/app/services/query.js
```diff
@@ -83,14 +164,51 @@ export class Parameter {
...
+  }
+
+  get hasDynamicDateRange() {
+    return isDateRangeParameter(this.type) && isDynamicDateRange(this.value);
+  }
+
+  get dynamicDateRange() {
+    return getDynamicDateRange(this.value);
+  }
+
```
ranbena: Is there a way around having type specific code in the generic param object?
Can this be reduced to `get hasDynamicValue()` and `get dynamicValue()`? (leaving out data type)

client/cypress/integration/query/parameter_spec.js
```diff
@@ -170,46 +181,47 @@ describe('Parameter', () => {
...
-        .contains('.ant-calendar-date', '15')
+      cy.getByTestId('DynamicTable')
+        .should('contain', Cypress.moment(this.now).format('15/MM/YY'));
+    });
+
+    it('allows picking a dynamic date', function () {
+      cy.getByTestId('DynamicButton')
         .click()
-        .click(); // workaround for datepicker display bug
+        .click();
```
ranbena: Assuming the double click is the picker workaround.

```suggestion
        .click(); // workaround for datepicker display bug
```

client/cypress/integration/query/parameter_spec.js
```diff
@@ -286,11 +295,24 @@ describe('Parameter', () => {
...
-          .should('contain', now.format('YYYY-MM-01 HH:mm'));
-      });
+      cy.getByTestId('DynamicTable')
+        .should('contain', Cypress.moment(this.now).format('YYYY-MM-DD HH:mm'));
+    });
+
+    it('allows picking a dynamic date', function () {
+      cy.getByTestId('DynamicButton')
+        .click()
+        .click();
```
ranbena: Assuming the double click is the picker workaround.

```suggestion
        .click(); // workaround for datepicker display bug
```

package.json
ranbena: Plz do a site-wide sweep to see that no ant component broke.

ranbena:
> My concern is more related not to break existing Dashboards/Queries that use that.

So maybe we need to do a db migration where if `param.useCurrentDateTime` is true and `param.value` is empty, then set dynamic value to `Today`?

arikfr:
> I actually don't think the missing option on Parameter Settings will be an issue,

I agree.

arikfr:
> So maybe we need to do a db migration where if param.useCurrentDateTime is true and param.value is empty, then set dynamic value to Today?

That's definitely an option but these migrations always carry some risk and that's why I asked what's the alternative cost.

client/app/components/dynamic-parameters/DynamicParameters.less
arikfr: +1 for Option 2.

arikfr: (corrected typo in previous reply)

arikfr:
*Not a blocker*, but I noticed that date range changes size when switching between dynamic and static. How about we keep it same size and show the dates along with the range name when it's dynamic?

client/app/components/ParameterMappingInput.jsx
```diff
@@ -536,7 +536,16 @@ export class ParameterMappingListInput extends React.Component {
...
-    const value = Parameter.getValue(param);
+    let value = Parameter.getValue(param);
+
+    // in case of dynamic value display the name instead of value
+    if (param.hasDynamicDate) {
+      value = param.dynamicDate.name;
+    }
+    if (param.hasDynamicDateRange) {
+      value = param.dynamicDateRange.name;
+    }
```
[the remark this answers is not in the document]
gabrieldutra: I'll run more tests on it, but it seems to me this is not directly related with Dynamic Values, but with the Values keeping in the URL when you change the mapping. (Happens in preview as well).

client/app/components/dynamic-parameters/DateRangeParameter.jsx
gabrieldutra: Yes, for DateRanges you'll have two values so in SQL you'd have:

```sql
WHERE column BETWEEN '{{range.start}}' AND '{{range.end}}'
```

That's why I haven't enabled Today/Yesterday for Date Ranges without time, it seemed to have no sense as you wouldn't have the "Range".

client/app/services/query.js
gabrieldutra: > Is there a way around having type specific code in the generic param object?

Inheritance! :tada: 

> Can this be reduced to get hasDynamicValue() and get dynamicValue()? (leaving out data type)

That was my first approach, I'm not so sure why I changed this part too... I'll move it back to it in this part and in case it makes sense I separate it back in the future when parameters are destructured .

client/app/services/dashboard.js
```diff
@@ -210,7 +210,7 @@ function DashboardService($resource, $http, $location, currentUser) {
       }
     });
     return _.values(_.each(globalParams, (param) => {
-      param.setValue(param.getValue()); // apply global param value to all locals
+      param.setValue(param.value); // apply global param value to all locals
```
gabrieldutra: @kravets-levko I guess this won't affect the solution for #3853 :slightly_smiling_face:

client/app/components/dynamic-parameters/DynamicParameters.less
gabrieldutra: Updated styling to Option 3 for tests. The interesting thing about the prominent blue was that it made me think "there's some 'magic' happening in that parameter".

gabrieldutra:
> That's definitely an option but these migrations always carry some risk and that's why I asked what's the alternative cost.

In this case I think it's safe to just remove the logic for that, parameters will get the "empty" error and the new option should be discovered while trying to solve that.

package.json
gabrieldutra: Biggest change seems to be the Table sorting arrows that are now blue and aligned with the text ([Percy](https://percy.io/Redash/redash/builds/2222641/view/144507643/1280?mode=diff&browser=chrome&snapshot=144507643)):
image

ranbena: Awesome. I also did a quick run through all pages and all is well.

client/app/components/dynamic-parameters/DynamicParameters.less
ranbena: > Updated styling to Option 3 for tests.

👍 

> The interesting thing about the prominent blue was that it made me think "there's some 'magic' happening in that parameter".

I definitely agree but getting the feel for it, I realized it draws too much attention.

client/app/components/dynamic-parameters/DateRangeParameter.jsx
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    return (
+      <DateRangeComponent
+        className={classNames('redash-datepicker', { 'dynamic-value hide-end-value': hasDynamicValue }, className)}
+        onSelect={onSelect}
+        suffixIcon={(
+          <DynamicButton
+            options={options}
+            selectedDynamicValue={(hasDynamicValue && isString(parameter.value)) ? parameter.value : null}
```
ranbena: Doesn't seem to work in date range:

Screen Shot 2019-07-20 at 18 34 48

ranbena: Actually, what's happening is that the selected value is `param.value` but should be the "pending" value if there is one.

ranbena: (I just realized the arrow is pointing to the wrong option but you know what I mean 😄)

gabrieldutra: After the merge with Apply All `parameter.value` and `value` became a different thing ^^, so I had to change a lot on the implementation, this was a missing point :grimacing:.

client/app/components/ParameterMappingInput.jsx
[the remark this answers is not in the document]
ranbena: If you could isolate this bug to a non-dynamic value STR, then we could open a separate issue for it.

ranbena:
As far as I'm concerned, this is complete apart for https://github.com/getredash/redash/pull/3904/files#r304734935

client/app/components/ParameterMappingInput.jsx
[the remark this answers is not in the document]
gabrieldutra: #4004

gabrieldutra:
> @gabrieldutra one thing I didn't test is "Refresh Schedule". If you're confident about it - go ahead and merge

Well reminded. I guess so far the pending discussion points in this PR are the "Refresh Schedule" and the size changing for Dynamic Date Ranges.

**Refresh Schedule**
@arikfr mentioned that in the issue, it's a point where we will decide whether it's valuable to replicate the frontend logic in the backend (which is a bit different as in frontend we can use the user's time). From what I understood, Refresh Schedule for Parameterized queries can't be used on Alerts, so it's purpose is to keep results updated when you open the page.

Also, when trying to work on that part in #3952, I couldn't get the Refresh Schedule to work (used Python debugger to track).

**Changing size for Date Ranges**
> _Not a blocker_, but I noticed that date range changes size when switching between dynamic and static. How about we keep it same size and show the dates along with the range name when it's dynamic?

It's possible, different styling can be done by using the two inputs placeholders. One con is that it will differ from Date Parameters (also Date Ranges are very wide :sweat_smile:). @ranbena any quick ideas for this?

image

ranbena:
> any quick ideas for this? So far I'm ok with the changing size thing for styling purposes

I don't think we can justify this huge component with dynamic values.
How about a transition to soften it up?

arikfr:
> I don't think we can justify this huge component with dynamic values.

In your example it's huge because it's a date range with time. But I wonder if date range with time needs dynamic values? All the dynamic values are date based, so maybe it should be only for date range and date picker? Then the size delta isn't that big.

> Refresh Schedule

The timezone issue is indeed tricky in this case :-| With dates only it's less an issue, but still can manifest on day changes. Not sure what to do about this, we can save the current timezone when saving the dynamic values but that's not great.

We can decide that there is no scheduling for such queries. Although this opens its own Pandora's box -- what if the query is already scheduled, etc.

gabrieldutra:
> so maybe it should be only for date range and date picker

I don't think we should limit the scope of the feature, from a "fixed" value standpoint, yes, it makes no sense for Time pickers (except for the "Now" one). However I think it's convenient to have the Dynamic Values available: you can have a dashboard with a Date Range Time specified and if you want to check out the result for "Last week" you are just a few clicks away.

arikfr:
@gabrieldutra fair points about keeping it for all types :+1: 

So what do you think about the option of disabling scheduled queries?

ranbena:
> > so maybe it should be only for date range and date picker
> 
> I don't think we should limit the scope of the feature, from a "fixed" value standpoint, yes, it makes no sense for Time pickers (except for the "Now" one). However I think it's convenient to have the Dynamic Values available: you can have a dashboard with a Date Range Time specified and if you want to check out the result for "Last week" you are just a few clicks away.

That's my thinking as well.
Moreover, I'm not certain users would necessarily pick up on the logic behind the differentiation, which might lead to frustration.

gabrieldutra:
> So what do you think about the option of disabling scheduled queries?

I got a feeling that scheduled queries are not working for queries with parameters, which makes "what if the query is already scheduled" not an issue.

But I agree with the "disabling scheduled" for such queries, in backend we can actually use the existing validation to skip execution (`d_values` are not valid dates, thus the query is not valid) and in the UI we can disable the option with a tooltip.

arikfr:
> I got a feeling that scheduled queries are not working for queries with parameters, which makes "what if the query is already scheduled" not an issue.

:thinking: I just checked and it does seem like it has no effect / not working. I will check this out.

> But I agree with the "disabling scheduled" for such queries

Let's do it in a separate PR then.

gabrieldutra:
> Let's do it in a separate PR then.

I'll wrap this with the transition and fix Cypress tests then

gabrieldutra:
@ranbena, I've updated this with the transition + a fix to a bug (Date Ranges with no previous value + click on "Back to static value" was resulting in `moment(null.start)`). If you wanna give a check on those :)

client/app/components/dynamic-parameters/DynamicParameters.less
```diff
@@ -0,0 +1,36 @@
...
+.redash-datepicker {
+  width: auto !important;
+
+  .ant-calendar-picker-clear {
+    right: 35px;
+  }
+
+  &.date-range-input .ant-calendar-picker-input {
+    width: 376px;
+    transition: width 150ms ease-in-out;
```
ranbena: I like this speed better.
Feel free to reject if you disagree.

```suggestion
    transition: width 100ms ease-in-out;
```

client/app/components/dynamic-parameters/DynamicParameters.less
```diff
@@ -0,0 +1,36 @@
...
+
+.redash-datepicker {
+  width: auto !important;
+
+  .ant-calendar-picker-clear {
+    right: 35px;
+  }
+
+  &.date-range-input .ant-calendar-picker-input {
+    width: 376px;
```
ranbena: Too wide for other date range types I think.
Perhaps set the width at the component level?
```jsx
const widthByType = {
  'date-range': 250,
  'datetime-range': 320,
  'datetime-range-with-seconds': 370,
};

return <DateTimeRangeInput
  width={widthByType(this.props.type)}
}
```

gabrieldutra: I went with 270 for date-range, 250 I felt the clear button was too close.

client/app/components/dynamic-parameters/DateRangeParameter.jsx
```diff
@@ -38,6 +38,12 @@ const DYNAMIC_DATETIME_OPTIONS = [
   ...DYNAMIC_DATE_OPTIONS,
 ];
 
+const widthByType = {
+  'date-range': 270,
+  'datetime-range': 320,
+  'datetime-range-with-seconds': 370,
+};
```
ranbena: You're right about the clear button position.
Here's a new set of widths with a perfect layout:
```js
const widthByType = {
  'date-range': 294,
  'datetime-range': 352,
  'datetime-range-with-seconds': 382,
};
```
And also:
```css
.redash-datepicker .ant-calendar-picker-clear {
  background: transparent;
}
```

gabrieldutra:
@arikfr I guess this one is on you now, whenever you think is convenient :)

arikfr:
:+1:

Issue #608, opened 2015-10-13
Add all in the filter


vishesh92:
Option to add all the options in the table filter as well as to remove all of them in a single click.

arikfr:
you mean when using a "multi-filter"?

vishesh92:
yeah in a multi filter.
Sorry for being less descriptive.

rakeshsingh:
+1

ChiragKParmar:
+1

matbott:
+1

Olaktal:
+1

yairraz:
+1

maer89:
+1

funyapu:
I rather want no filtered result if multi-filter has no value...

fakob:
+1 @funyapu comment

jenslukas:
+1

iel19:
+1

ultymike:
+1

tsavner:
+1

isaakd:
+1

akichatov:
+1

heristop:
+1

Woffendm:
+1

bgroff:
I think this would be nice to have on a single filter as well. Another way I could see this working is an optional parameter where clause. The input would then be a select that was backed by an additional SQL query.

I have some data that I would like to be able to filter by say a US state, but sometimes I want data filtered by all states (aka not filtered at all). This seemed hard to do at the moment, but it could be that I am misunderstanding how redash works.

malajul:
+1

luiscbf:
+1

amygriffis:
+1

JosephDevaney:
+1

ykensuke:
+1

harmy:
+1

eschultze:
+1

fatayri:
+1

Winangel:
+1 This is really a must have feature...

g-fei-song:
+1

TheScoobyBoy:
Any news? We really need this..

mfulleratlassian:
+1

dimar7:
+1

taciomedeiros:
+1

arikfr:
I'm not sure anyone even remembers how to create multi filters at this point 😅 And parameters have support for this.

PR #3905, opened 2019-06-16
Provide settings to control multifilter initial selection

kewisch:

- [X] Feature

## Description
Back in #608 there was a hidden feature to pre-select all entries in a multi-filter, which got removed when migrating the filters to react. I heavily depend on this and would like to see it come back. I think it would make sense to fix this in a less hacky way, using column metadata.

In below screenshot, the priority and status columns are multi-filters, therefore they have an extra dropdown to select one of "first", "all", or "none". Selecting "first" uses the current behavior to select the first value, "all" and "none" are self-explanatory.

### To-Do

- [ ] Possibly refactor filter current value
- [ ] Find sensible defaults for dashboard-level filters
- [ ] Tests

### Discussion items

I'm not particularly happy about how `setFilterDefaults` runs after the filters are gathered in the query result service and after the options are retrieved. Changing this may require some refactoring, but before I start with that I'd like to get some feedback

The second thing that will require some discussion, how to do dashboard level filters. These don't have access to a specific visualization, so if two visualizations have different defaults for the default state for a multi-filter column then there is a conflict. Also, it doesn't seem like the specific visualizations are available at the time the filters are determined, similar to the first discussion point.

Looking forward to your thoughts.

TBD

image

### DIFF
--- a/client/app/components/Filters.jsx
+++ b/client/app/components/Filters.jsx
@@ -41,6 +41,31 @@ function createFilterChangeHandler(filters, onChange) {
   };
 }
 
+export function setFilterDefaults(filters, columnOptions) {
+  const multiFilterDefaults = {};
+  columnOptions.forEach((column) => {
+    multiFilterDefaults[column.name] = column.multiFilterDefault;
+  });
+
+  filters.forEach((filter) => {
+    if (filter.current) {
+      return;
+    }
+
+    if (filter.multiple) {
+      if (multiFilterDefaults[filter.name] === 'all') {
+        filter.current = filter.values;
+      } else if (multiFilterDefaults[filter.name] === 'none') {
+        filter.current = [];
+      } else {
+        filter.current = [filter.values[0]];
+      }
+    } else {
+      filter.current = filter.values[0];
+    }
+  });
+}
+
 export function filterData(rows, filters = []) {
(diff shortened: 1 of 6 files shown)


arikfr:
Thank you for taking a stab at this, @kewisch. It seems that this is local to a specific table visualization, but filters are at the query level (i.e. same filters apply to all the visualizations of the same query). It's also unclear how this will work on dashboards.

We been doing a lot of work on improving parameters and I wonder if you could replace your use of filters with parameters once they support multi-select? (#3952)

arikfr:
@kewisch did you see my previous comment?

kewisch:
Hi @arikfr sorry for the delay. I think a lot of this can be covered by the multi-value query filter, possibly with some small features in addition. I've checked the gifs and it seems possible to add multiple values either based on a string list or a query. This is a great start and would allow me to replicate what filters can do now.

What I am missing, apologies if I missed this, is a feature that would auto-select all values in the multi-select when the query is executed. Is this something possible with #3952 ?

arikfr:
> What I am missing, apologies if I missed this, is a feature that would auto-select all values in the multi-select when the query is executed. Is this something possible with #3952 ?

Not at the moment.

PR #3907, opened 2019-06-17
Parameter “Apply Changes” button

ranbena:
- [x] Feature

## Description
Following ux discussion in [what-to-do-when-you-want-to-update-several-parameters-before-running-a-query](https://discuss.redash.io/t/what-to-do-when-you-want-to-update-several-parameters-before-running-a-query/3906)

### Applied for:
1. Dashboard level parameters
2. Widget level params
3. Query page
4. Visualization Embed

(And not for param mapping edit modal, as this is now a feature of `<parameters>`)

### Mobile
Bigger button, not "floating"


ranbena:
@gabrieldutra I'm at a loss on why applying changes to dashboard params specifically, doesn't then update the url (resulting in changes reverted). Any idea?

gabrieldutra:
A quick one before I forget: ~ParamaterApplyButton.jsx~ -> ParameterApplyButton.jsx :)

> I'm at a loss on why applying changes to dashboard params specifically, doesn't then update the url (resulting in changes reverted). Any idea?

I haven't checked the code yet to confirm, though I faced the same issue and `$scope.$apply();` before triggering the update event fixed it to me. ([happened here](https://github.com/getredash/redash/pull/3737#discussion_r283543036))

ranbena:
@arikfr Enter keypress functionality retained.
But @susodapop would you like it to work differently?

ranbena:
Found bug: applying empty text reverts change.

ranbena:
Found bug: clearing date input removes the component 😅

client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -472,7 +472,7 @@ body {
 
 .widget-wrapper {
   .parameter-container {
-    padding: 0 15px;
+    margin: 0 15px;
```
ranbena: So that the apply button is positioned correctly.

gabrieldutra:
> Enter keypress functionality retained.

Actually the "Enter" keypress is triggering all parameters stack and only works in text or number inputs.

I wonder if CTRL + Enter should force parameters update from anywhere (implementation may not be that simple).

ranbena:
> Actually the "Enter" keypress is triggering all parameters stack and only works in text or number inputs.

@arikfr worked with the version before and concluded that it's a feature he'd like to retain.

> I wonder if CTRL + Enter should force parameters update from anywhere (implementation may not be that simple).

I think so too. It would be great to allow site-wide query execution with ctr+ent but yep, it's risky business - requires certainty to which dashboard context the execution should be applied to and the code would scream "bad practice".

arikfr:
Just to clarify: `Cmd/Ctrl` + `Enter` should trigger query execution/applying of parameters in a query page.

As for dashboards:

> requires certainty to which dashboard context the execution should be applied to 

All the currently unapplied parameters?

> and the code would scream "bad practice".

why?

ranbena:
> Just to clarify: `Cmd/Ctrl` + `Enter` should trigger query execution/applying of parameters in a query page.

Ok I'll make sure.

> As for dashboards:
> > requires certainty to which dashboard context the execution should be applied to
> All the currently unapplied parameters?

To me feels like an unexpected behavior. What would be expected is that the context I'm currently focused on (dashboard params or specific widget) would be the one to react to ctr+ent.

> > and the code would scream "bad practice".
> why?

Afaik this would require determining focus on non-focusable elements (e.g. <div>) or keeping tabs on focus, which usually means elaborate code leading to unstable UX.

arikfr:
> I would expect only the context I'm currently focused on (dashboard params or specific widget) to react to ctr+ent.

Agree, but because of the issue you mentioned later, I felt it's ok to implement a simpler version. Anyway, better to finish the non keyboard UX + keyboard UX on query page and then consider our options on a dashboard.

ranbena:
> Just to clarify: Cmd/Ctrl + Enter should trigger query execution/applying of parameters in a query page.

@arikfr any execution should apply current edits to params - be it keyboard shortcut or **clicking execute button**?

arikfr:
> @arikfr any execution should apply current edits to params - be it keyboard shortcut or clicking execute button?

I think so.

ranbena:
> Cmd/Ctrl + Enter should trigger query execution/applying of parameters in a query page.

I'm not sure how to implement this.
The way I did it, `<parameters>` is the one that saves the pending changes and then applies them on click. With this architecture, is there a way for other elements (execute buttons and keyboard shortcut) to have access to executing the apply?

@gabrieldutra @kravets-levko

arikfr:
@ranbena this could be easy if we were using a state management library :) 

But even without it, we have the Parameters model -- maybe we can have it hold the interim state, and then when executing a query, we will trigger "apply changes" on the Parameters model and execute query?

ranbena:
> @ranbena this could be easy if we were using a state management library :)

Yep.

> But even without it, we have the Parameters model -- maybe we can have it hold the interim state, and then when executing a query, we will trigger "apply changes" on the Parameters model and execute query?

How would you access the Parameters instance from another component?

arikfr:
The parameters instance available from the query/dashboard object which is
available to the code that executes the query.

gabrieldutra:
@ranbena what I had thought before was to expose a `forceUpdate` method from the component, considering parameters was already migrated to React, this way we could access it through refs. It was my best idea considering I think Parameters should be the one holding the state.

The problem is that the Query Page is still in Angular, so one workaround to make it work and simulate the above (definitively not pretty code-looking) is to define an object that will pass as a prop to the Parameters component, the Parameters component will then point the `applyChanges` function in this object. The apply method would then be available to be executed from the Query/Dashboard.

Arik's Parameters instance approach is way better in terms of code, but for a React Code I guess it wouldn't trigger re-render, for Angular I don't know.

ranbena:
Some changes:
1. `ctrl+enter` also works when focused on param input, `enter`-only functionality removed. Code is simpler and UX is flowing. @arikfr @gabrieldutra ok with you?
2. Query execution with button and shortcut - now applies dirty params before execution 🤟

I managed to do this by moving the "dirty" code from `<parameters>` into `Parameters`. The code is simpler and neater. It also allows a future feature of "undo changes" 🎈.

The only downside is that the "Apply Changes" button doesn't know that applying is being run if triggered from shortcut/execute button, so it won't display an "applying" state. This seems negligible to me since this state is usually too fast to even notice. Lmk if you think or experience differently.

ranbena:
I noticed that `DateTimeInput` and `DateTimeRangeInput` don't trigger an `onSelect` when clicking the "clear" button (resulting in not showing a "dirty" indication). @gabrieldutra Do you know why this is?

client/app/components/ParameterValueInput.less
```diff
@@ -1,50 +1,80 @@
...
   display: inline-block;
   position: relative;
-  width: 195px;
 
-  &[data-dirty] {
-    .@{ant-prefix}-input {
-      padding-right: 58px !important;
-    }
+  .@{ant-prefix}-input {
+    width: 195px;
```
gabrieldutra: This is also being applied to DateRanges :sweat_smile: 

date-time-range

client/app/components/DateTimeInput.jsx
```diff
@@ -17,23 +17,14 @@ export function DateTimeInput({
...
       showTime
       {...additionalAttributes}
       format={format}
       placeholder="Select Date and Time"
-      onChange={(newValue) => { currentValueRef.current = newValue; }}
-      onOpenChange={(status) => {
-        const currentValue = currentValueRef.current;
-        if (!status) { // on close picker
-          if (currentValue && currentValue.isValid()) {
-            onSelect(currentValue);
```
gabrieldutra: @ranbena this logic was used to trigger changes only when the DatePicker was closed, with nis implementation is no longer needed. This was the cause for the clear button issue.

client/app/components/ParameterValueInput.less
ranbena: Whoops! https://github.com/getredash/redash/pull/3907/commits/bb197311ea2c1d1091225e90b0949e405fc183f2

client/app/services/query.js
```diff
@@ -483,6 +507,7 @@ function QueryResource(
     }
 
     const parameters = this.getParameters();
+    parameters.applyPendingValues();
```
ranbena: This is the line that allows "Execute" button and kb shortcut to apply before running the query.

client/app/components/queries/visualization-embed.html
```diff
@@ -12,7 +12,7 @@ <h3>
   </div>
 
   <div class="col-md-12 query__vis">
-    <div class="p-t-15 p-b-5" ng-if="$ctrl.query.hasParameters() && !$ctrl.hideParametersUI">
+    <div class="p-t-15 p-b-10" ng-if="$ctrl.query.hasParameters() && !$ctrl.hideParametersUI">
```
ranbena: So that the apply button is positioned correctly.

client/app/components/ParameterValueInput.jsx
```diff
@@ -145,79 +143,49 @@ export class ParameterValueInput extends React.Component {
...
-    };
+    const normalize = val => !isNaN(val) && val || 0;
 
     return (
-      <div className="parameter-input-number" data-dirty={showApplyButton || null}>
-        <InputNumber
-          className={className}
-          value={!isNaN(value) && value || 0}
-          onChange={onChange}
-          onKeyUp={showApplyButton ? (e) => {
```
ranbena: @arikfr I removed the "enter" listener. Only "ctrl+enter" works. Lmk if you're ok with that.

client/app/components/parameters.html
```diff
@@ -18,6 +18,7 @@
     >
       <i class="zmdi zmdi-settings"></i>
     </button>
-    <parameter-value-input param="param" on-change="onValuesChange" apply-button="true"></parameter-value-input>
+    <parameter-value-input param="param" />
```
gabrieldutra: I saw some [inputs](https://github.com/angular/angular.js/issues/1953#issuecomment-13135021) against self-closing tags in angular, but we do have them in other parts of the code and no problems reported so far :man_shrugging:

client/app/components/ParameterValueInput.less
```diff
@@ -1,50 +1,80 @@
...
+    .@{ant-prefix}-input-number,
+    .@{ant-prefix}-select-selection {
+      background-color: @input-dirty;
+    }
   }
 }
 
-.parameter-input-number {
-  display: inline-block;
+.parameter-container {
```
gabrieldutra: This CSS seems more coupled to the `Parameters` component than to `ParameterValueInput`, guess it should be moved when we migrate the first, right?

client/cypress/integration/query/parameter_spec.js
```diff
@@ -1,6 +1,22 @@
 import { createQuery } from '../../support/redash-api';
 
 describe('Parameter', () => {
+  const expectDirtyStateChange = (edit) => {
```
gabrieldutra: Cool idea :+1:. Only thing is that you're actually expecting "the dirty state", not it to change :sweat_smile:. Anyway, no big deal on this one.

client/cypress/integration/query/parameter_spec.js
```diff
@@ -21,25 +37,35 @@ describe('Parameter', () => {
...
       cy.getByTestId('DynamicTable')
         .should('contain', 'Redash');
 
       cy.getByTestId('ParameterName-test-parameter')
         .find('input')
-        .type('{selectall}New value{enter}');
+        .type('{selectall}New value');
+
+      cy.getByTestId('ParameterApplyButton')
+        .click();
```
gabrieldutra: Those tests had two parts: One clicking in the apply an another pressing enter. If you want you can remove this second part.

Alternatively you can test CTRL+Enter, but this one can also be one separate test for all types.

client/app/components/parameters.js
```diff
@@ -42,6 +42,29 @@ function ParametersDirective($location) {
...
+      scope.applyChanges = () => {
+        forEach(scope.parameters, p => p.applyPendingValue());
+        scope.isApplying = false;
+      };
+
+      scope.onApply = () => {
+        scope.isApplying = true;
+        scope.$apply();
+
+        scope.$apply(scope.applyChanges);
```
gabrieldutra: The two `$apply` (one only for `scope.isApplying = true`) are to trigger the "updating" state?

client/app/components/parameters.html
ranbena: Oh ok it's actually not an Angular thing but a browser thing. I got used to it with jsx... I'll revert.

client/app/components/ParameterValueInput.jsx
```diff
@@ -34,108 +32,108 @@ export class ParameterValueInput extends React.Component {
...
-        size="small"
-        onClick={() => onSelect(value)}
-      >
-        Apply
-      </Button>
-    );
+  componentDidUpdate = (prevProps) => {
+    // if value prop updated, reset dirty state
+    if (prevProps.value !== this.props.value) {
+      this.setState({ isDirty: false });
```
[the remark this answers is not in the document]
ranbena: > I wonder if the state value should also be updated here. It doesn't seem a good React practice, but there are some components that use that state as value and not the prop.

My code is assuming that value changes can only come from direct user interaction, so in that case there won't be prop/state syncing issues. But in case the prop is changed from outside (for instance, another user changed the param and it auto-updated) this would cause such an issue. I just thought it would be highly unlikely so I left it like this. I can add it just to be sure.

> A second thought: Is the value state really necessary now? (it was for the individual Apply button)

I think since the "apply button" changes the `prop.value` indirectly and not the actual component value update, we have to use `state.value`. Or do you mean, using `defaultValue={this.props.value}` instead?

client/cypress/integration/query/parameter_spec.js
ranbena: My intention was that it changes from false to true :) I'll leave it though.
I wish I could use `expect(paramEdit).to.change(getDirtyState);` 😭

client/app/components/ParameterValueInput.jsx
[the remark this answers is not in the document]
gabrieldutra: > My code is assuming that value changes can only come from direct user interaction, so in that case there won't be prop/state syncing issues.

:+1: 

> Or do you mean, using defaultValue={this.props.value} instead?

Exactly that.

client/cypress/integration/query/parameter_spec.js
ranbena: Yes! I'll use `{meta}{enter}` 👍

client/app/components/parameters.js
ranbena: Yes. This part I'm not sure of since a) I don't know much Angular b) it applies so fast it's hard to notice the state change.

client/app/components/ParameterValueInput.jsx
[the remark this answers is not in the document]
ranbena: BUT I want to leave it open for an "undo" feature, therefore would need state. Wdyt?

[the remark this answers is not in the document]
gabrieldutra: > BUT I want to leave it open for an "undo" feature, therefore would need state. Wdyt?

Yeah, makes sense then :grin:

client/app/components/parameters.js
gabrieldutra: :thinking:, this seems to provide the same result:

```javascript
scope.applyChanges = () => {
  scope.isApplying = true;
  forEach(scope.parameters, p => p.applyPendingValue());
  scope.isApplying = false;
};

scope.onApply = () => {
  scope.$apply(scope.applyChanges);
  scope.onValuesChange();
};
```

ranbena: Done 👍

client/cypress/integration/query/parameter_spec.js
ranbena: There are now 3 separate apply tests:
* applies changes from "Apply Changes" button
* applies changes from "Execute" button
* applies changes from "meta+enter" keyboard shortcut

client/app/components/parameters.html
ranbena: Fixed

client/app/components/ParameterValueInput.jsx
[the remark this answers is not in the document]
ranbena: Added `value` to the state update in `componentDidUpdate`

arikfr:
> ctrl+enter also works when focused on param input, enter-only functionality removed. Code is simpler and UX is flowing

I think it's a reasonable tradeoff.

arikfr:
`meta`+`enter` works only on a query page?

arikfr:
Not a blocker, just a nice to have to add in the future: show some hint about meta+enter when clicking the button.

ranbena:
> `meta`+`enter` works only on a query page?

Yes

arikfr:
@ranbena because of the focus on the un-focusable things issue? We can't have the listener on the `<parameters/>` component and let the browser do its thing?

ranbena:
> @ranbena because of the focus on the un-focusable things issue? We can't have the listener on the `<parameters/>` component and let the browser do its thing?

I think so yeah. I'll make sure.

arikfr:
@ranbena now that we have the ability to trigger parameters application from the outside (like you do on the query page), how about Ctrl+Enter on a dashboard applies all unapplied parameters, be it in a widget or dashboard? I think it's a reasonable tradeoff and should be simple to implement?

ranbena:
> @ranbena now that we have the ability to trigger parameters application from the outside (like you do on the query page), how about Ctrl+Enter on a dashboard applies all unapplied parameters, be it in a widget or dashboard? I think it's a reasonable tradeoff and should be simple to implement?

I'll give it a try.

ranbena:
@arikfr Keyboard shortcut PR https://github.com/getredash/redash/pull/3971

client/cypress/integration/query/parameter_spec.js
```diff
@@ -364,26 +375,22 @@ describe('Parameter', () => {
...
+      expectAppliedChanges(() => {
+        cy.getByTestId('ExecuteButton').click();
+      });
+    });
 
-      cy.location('search').should('contain', 'Redash');
-      cy.wait('@Results');
+    it('applies changes from "meta+enter" keyboard shortcut', () => {
+      expectAppliedChanges((input) => {
+        input.type('{meta}{enter}');
```
gabrieldutra: I'm not so sure what's the `{meta}` key on Linux :thinking:, will check that as Cypress is failing :sweat_smile: .

client/cypress/integration/query/parameter_spec.js
```diff
@@ -225,11 +283,114 @@ describe('Parameter', () => {
...
+          .click();
+
+        cy.get('.ant-calendar-date-panel')
+          .contains('Now')
+          .click();
+      });
+    });
+  });
+
+  describe.only('Apply Changes', () => {
```
gabrieldutra: BTW, don't forget to remove the `.only` before merging ^^
```suggestion
  describe('Apply Changes', () => {
```

client/cypress/integration/query/parameter_spec.js
gabrieldutra: For Linux (thus CI) it needs to be `{ctrl}` or `{alt}`

client/cypress/integration/query/parameter_spec.js
```diff
@@ -225,11 +283,114 @@ describe('Parameter', () => {
...
+      });
+    });
+
+    it('applies changes from "Execute" button', () => {
+      expectAppliedChanges(() => {
+        cy.getByTestId('ExecuteButton').click();
+      });
+    });
+
+    it('applies changes from "meta+enter" keyboard shortcut', () => {
```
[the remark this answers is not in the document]
ranbena: > Is `ctrl` the one that works also for Mac?

`ctrl` is for all, `cmd` is for Mac.
I changed it to "alt" so it works for all OSs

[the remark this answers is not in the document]
arikfr: It should be Ctrl+Enter on PC and Cmd+Enter on Mac, as this is the shortcut we use today. It's very common to have a shortcut that uses Cmd on Mac switch to Ctrl on PC.

...

I see that the code supports both options, but I don't see a reason why we need to support Alt+Enter here? Cypress supposed to support {meta}, so the test works on both PC & Mac?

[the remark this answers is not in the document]
ranbena: {alt+enter} is for tablets.
Cypress's {meta} is only an alias to {cmd} and doesn't adjust according to platform.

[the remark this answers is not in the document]
arikfr: > {alt+enter} is for tablets.

oh. forgot about this one 🤦‍♂ We should document this -- both in code and in the docs.

> Cypress's {meta} is only an alias to {cmd} and doesn't adjust according to platform.

 Bug in their platform?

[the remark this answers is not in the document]
ranbena: > Bug in their platform?

I don't think it's meant to be auto-adjusting.
We can do it ourselves using node.platform but is "darwin" a definite MacOS identifier?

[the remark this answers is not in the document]
arikfr: Never mind, I misread their docs...

ranbena:
@arikfr I want to make sure you're cool with the shortcut implementation before merging

arikfr:
@ranbena I'm going to test out the behavior tomorrow and then merge 🤩

ranbena:
@arikfr 🔔

ranbena:
Fixed "refresh/auto-refresh" revoking dashboard param dirty state.

@kravets-levko can you review this specific change https://github.com/getredash/redash/pull/3907/commits/0d1f28801d5d5a0ded9d2eb803af4b8952899203?

client/cypress/integration/query/parameter_spec.js
```diff
@@ -225,11 +283,114 @@ describe('Parameter', () => {
...
+      apply(cy.get('@Input'));
+
+      cy.location('search').should('contain', 'Redash');
+      cy.wait('@Results');
+    };
+
+    beforeEach(() => {
+      const queryData = {
+        name: 'Testing Apply Button',
+        query: "SELECT '{{test-parameter-1}} {{ test-parameter-2 }}'",
```
gabrieldutra: It seems you also need to provide the `options.parameters`. I've seen something related, not sure where :thinking:

ranbena: Yeah, I think @rauchy had a similar solution lately. Thanks!

gabrieldutra:
:tada:

arikfr:
:tada:

ranbena:
Screen Shot 2019-07-17 at 17 39 19

PR #3908, opened 2019-06-18
Support multi-byte search for query names and descriptions

### LINKED ISSUE #2622: Can't search query correctly with non-ASCII chars
### Issue Summary

Can't search query correctly with non-ASCII chars.

### Steps to Reproduce

1. Make query which has non-ASCII chars name or description
2. Search with non-ASCII chars

e.g.

There is a query which has non-ASCII chars `ユーザ`.

all_queries

Search with `ユーザ`, then no quries in the result.

search_query1

When I search with `ユー`, then hit correctly.

search_query2

I guess that `Query.search`(in #2041) has changed this behavior. But I have no idea what we should fix it with keeping full text search feature.

### Technical details:

* Redash Version: master
* Browser/OS: Version 67.0.3396.87 (Official Build) (64-bit)

sekiyama58:
* add multi_byte_support_enabled option to organization settings
* add `ilike %...%` to query search conditions when the option is enabled

- [x] Bug Fix

## Description

Since postgres tsvector does not work well with multi-byte characters in CJK- (and some others) launguages , currently queries named in multi-byte characters are often fails to be searched as described #2622.

This fixes the issue by adding an option to enable multi-byte compliant search into organization settings, and if it is enabled, search for queries using `<name or description> ilike %<search_term>%` pattern in addition to tsvector.
This search method is slower than tsvector search, so it is recommended only when multi-byte search is more important than speed.

This closes #2622.

* New organization settings for multi-byte search

スクリーンショット 2019-06-18 16 16 13
スクリーンショット 2019-06-18 16 17 03

* The result of search result when multi-byte search is enabled

スクリーンショット 2019-06-18 16 18 06

(If the option is disabled, nothing is hit in `ユーザー` search results.)

### DIFF
--- a/client/app/pages/settings/OrganizationSettings.jsx
+++ b/client/app/pages/settings/OrganizationSettings.jsx
@@ -161,7 +161,7 @@ class OrganizationSettings extends React.Component {
             checked={formValues.multi_byte_search_enabled}
             onChange={e => this.handleChange('multi_byte_search_enabled', e.target.checked)}
           >
-          Enable multi-byte search for query names and descriptions (slower)
+          Enable multi-byte (Chinese, Japanese, and Korean) search for query names and descriptions (slower)
           </Checkbox>
         </Form.Item>
         <Form.Item label="Feature Flags">
--- a/client/app/pages/settings/OrganizationSettings.jsx
+++ b/client/app/pages/settings/OrganizationSettings.jsx
@@ -155,6 +155,15 @@ class OrganizationSettings extends React.Component {
             ))}
           </Select>
         </Form.Item>
+        <Form.Item label="Multi-byte Search">
+          <Checkbox
+            name="multi_byte_search_enabled"
+            checked={formValues.multi_byte_search_enabled}
+            onChange={e => this.handleChange('multi_byte_search_enabled', e.target.checked)}
+          >
+          Enable multi-byte search for query names and descriptions (slower)
+          </Checkbox>
+        </Form.Item>
         <Form.Item label="Feature Flags">
           <Checkbox
             name="feature_show_permissions_control"
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -578,8 +578,7 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
 
     @classmethod
     def search(cls, term, group_ids, user_id=None, include_drafts=False,
-               limit=None, include_archived=False):
-        from redash.authentication.org_resolving import current_org
+               limit=None, include_archived=False, multi_byte_search=False):
         all_queries = cls.all_queries(
             group_ids,
             user_id=user_id,
@@ -587,7 +586,7 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
             include_archived=include_archived,
         )
 
-        if current_org.get_setting("multi_byte_search_enabled"):
+        if multi_byte_search:
             # Since tsvector doesn't work well with CJK languages, use `ilike` too
             pattern = u'%{}%'.format(term)
             return all_queries.filter(
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -579,12 +579,25 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
     @classmethod
     def search(cls, term, group_ids, user_id=None, include_drafts=False,
                limit=None, include_archived=False):
+        from redash.authentication.org_resolving import current_org
         all_queries = cls.all_queries(
             group_ids,
             user_id=user_id,
             include_drafts=include_drafts,
             include_archived=include_archived,
         )
+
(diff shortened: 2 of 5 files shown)


redash/models/__init__.py
```diff
@@ -579,12 +579,25 @@ def outdated_queries(cls):
...
             include_drafts=include_drafts,
             include_archived=include_archived,
         )
+
+        if current_org.get_setting("multi_byte_search_enabled"):
+            # Since tsvector doesn't work well with CJK languages, use `ilike` too
+            pattern = u'%{}%'.format(term)
+            return all_queries.filter(
+                or_(
+                    cls.search_vector.match(term),
```
arikfr: Isn't it enough to match only `name` and `description`?

client/app/pages/settings/OrganizationSettings.jsx
```diff
@@ -155,6 +155,15 @@ class OrganizationSettings extends React.Component {
             ))}
           </Select>
         </Form.Item>
+        <Form.Item label="Multi-byte Search">
+          <Checkbox
+            name="multi_byte_search_enabled"
+            checked={formValues.multi_byte_search_enabled}
+            onChange={e => this.handleChange('multi_byte_search_enabled', e.target.checked)}
+          >
+          Enable multi-byte search for query names and descriptions (slower)
```
arikfr: How about the following description?

```suggestion
          Enable multi-byte (Chinese, Japanese, and Korean) search for query names and descriptions (slower)
```

It might not be 100% technically accurate, but will be more familiar to users.

sekiyama58: Thank you, will update this soon!

redash/models/__init__.py
sekiyama58: I'm not sure it is popular, but `search_vector` contains the query's SQL too, so we can find by some terms in SQL with this (only with small overhead). How do you think?

jezdez:
@arikfr So I'm not sure if this should block this PR, I just wondered whether the "search backends" idea would still useful for this?

Looking back, I'm not super convinced the tsvector based search delivered what we hoped to get -- smooth deployment and improved results. Especially given the fact that tsvector is limited in many ways and excludes a huge part of the market for non-latin languages.

My suggestion would be to:

1. adding a backend system (like the query runner registry) that can be extended with additional Python packages. It would need to have a CRUD interface, so could be pretty simplistic.
2. ship the old ILIKE based search as the default backend
3. implement more complex (tsvector or Elasticsearch) based backends as extensions (either by the Redash team/contributors or 3rd parties?)

jezdez:
Of course the additional search backends (e.g. tsvector or Elasticsearch) could even be shipped and just made optional (e.g. when a ES cluster is available).

redash/models/__init__.py
arikfr: In this case, you can add the `query_text` column to the search instead. But I think it's better to leave it out, because for some search terms it can yield really bad results (like when you search for something that is also a table name). When using the regular search it has weights in place, so those are ranked lower.

sekiyama58: Thank you for your advise. I have removed `search_vector` for multi-byte search. It still provides reasonable search results for me.

redash/models/__init__.py
```diff
@@ -579,12 +579,24 @@ def outdated_queries(cls):
...
                limit=None, include_archived=False):
+        from redash.authentication.org_resolving import current_org
         all_queries = cls.all_queries(
             group_ids,
             user_id=user_id,
             include_drafts=include_drafts,
             include_archived=include_archived,
         )
+
+        if current_org.get_setting("multi_byte_search_enabled"):
```
arikfr: I just noticed that all the search tests are failing, because now this code requires a request context to use `current_org`.

Let's make this an argument this method takes from the calling code (which has access to `current_org`).

arikfr:
@jezdez this is good feedback, but will be lost once this is merged. Let's move it to https://discuss.redash.io/c/development?

redash/models/__init__.py
sekiyama58: Fixed the tests by making `multi_byte_search` an argumeent.
Also added the model test for it.

jezdez:
> @jezdez this is good feedback, but will be lost once this is merged. Let's move it to https://discuss.redash.io/c/development?

Good idea, expanded on it in https://discuss.redash.io/t/search-backend-proposal/4080

arikfr:
Merged 👍

redash/models/__init__.py
```diff
@@ -578,13 +578,24 @@ def outdated_queries(cls):
...
+
+        if multi_byte_search:
+            # Since tsvector doesn't work well with CJK languages, use `ilike` too
+            pattern = u'%{}%'.format(term)
+            return all_queries.filter(
+                or_(
+                    cls.name.ilike(pattern),
+                    cls.description.ilike(pattern)
+                )
+            ).order_by(Query.id).limit(limit)
```
arikfr: Just noticed -- maybe we should order by `created_at` or `updated_at` timestamps?

PR #3909, opened 2019-06-18
Add an `extras` field for Presto connection

ktmud:

- [x ] Feature

## Description

This added a new "Extras" fields for Presto connection,
 mostly for allowing self-assigned SSL certificate.

For example, services like [Qubole](https://docs.qubole.com/en/latest/user-guide/presto/connect-to-presto-when-ssl-enabled.html)
only allows SSL connection via self-assigned CA certificate.

Example usage is to include `requests_kwargs` as JSON string:

```json
{"requests_kwargs":  {"verify": "~/qubole.cert"} }
```

### DIFF
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -59,8 +59,12 @@ class Presto(BaseQueryRunner):
                 'password': {
                     'type': 'string'
                 },
+                'extras': {
+                    'type': 'string',
+                    'default': '{"requests_kwargs": null}'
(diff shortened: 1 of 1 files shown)


arikfr:
> For example, services like Qubole only allows SSL connection via self-assigned CA certificate.

Did you know that we have a Qubole query runner?

ktmud:
Oops, didn't see that... But can this PR be of any use for other scenarios still?

PR #3911, opened 2019-06-18
Add keyboard shortcut for format query (Ctrl/Cmd+Shift+F)

arikfr:

- [x] Feature

## Description

`Ctrl/Cmd` + `Shift` + `F` will trigger format query. Currently works when you're in the query editor.


gabrieldutra:
I could swear this shortcut already existed :thinking: 

> How can users find out about this?

Yep, perhaps add in the tooltip simular to "Add new parameter".

arikfr:
> Yep, perhaps add in the tooltip simular to "Add new parameter".

I forgot about this 🤦‍♂ 

Thanks, @ranbena for taking care of this :)

PR #3912, opened 2019-06-18
Expose celery job timeout setting to env var

deecay:

- [x] Feature

## Description
This PR allows users to define celery job timeout from environment variable. Environment variable is `REDASH_CELERY_INIT_TIMEOUT` and the default value is 10 seconds.

It should help people with limited computing resource seeing `[ERROR][MainProcess] Process 'ForkPoolWorker-19' pid:63 exited with 'signal 9 (SIGKILL)'` in their logs.

This PR is an follow up of #3903.

### DIFF
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -31,8 +31,8 @@ CELERY_RESULT_BACKEND = os.environ.get(
 CELERY_RESULT_EXPIRES = int(os.environ.get(
     "REDASH_CELERY_RESULT_EXPIRES",
     os.environ.get("REDASH_CELERY_TASK_RESULT_EXPIRES", 3600 * 4)))
-CELERY_JOB_TIMEOUT = int(os.environ.get(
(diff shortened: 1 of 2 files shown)


redash/settings/__init__.py
```diff
@@ -31,6 +31,8 @@
 CELERY_RESULT_EXPIRES = int(os.environ.get(
     "REDASH_CELERY_RESULT_EXPIRES",
     os.environ.get("REDASH_CELERY_TASK_RESULT_EXPIRES", 3600 * 4)))
+CELERY_JOB_TIMEOUT = int(os.environ.get(
```
arikfr: `CELERY_JOB_TIMEOUT` is a bit misleading here. How about `CELERY_INIT_TIMEOUT`?

arikfr:
Thanks!

jezdez:
Just for record-keeping, this is considered a potential bug in Celery and may be fixed with an upstream fix: https://github.com/celery/celery/issues/4323

PR #3924, opened 2019-06-21
Update Dockerfile to copy only client side files at the front end stage

ktmud:

- [x] Feature

## Description

Update Dockerfile to copy only client side files at the front end stage
So that changing other files will not trigger the
very expensive npm build process.


arikfr:
👍 thanks.

Issue #3730, opened 2019-04-22
WordCloud hides words that don't fit in the container


kravets-levko:
### Issue Summary

WordCloud hides words that don't fit in the container. 

Issue comes from library used for Word Cloud layout: https://github.com/jasondavies/d3-cloud/issues/36

### Steps to Reproduce

1. Prepare a data for Word Cloud visualization. Include some very long words with high frequencies (they should be large).
2. Create a Word cloud visualization: some words will be missing.

### Technical details:

* Redash Version: current [`master`](https://github.com/getredash/redash/tree/fb48bc374a362f058edcc43e28a698aa2d44e737)
* Browser/OS: any
* How did you install Redash: docker

PR #3668, opened 2019-04-01
[#1414] Word Cloud Word Frequency Selection

mohamagdy:
## What type of PR is this?

- [x] Feature

## Description

This PR addresses part of https://github.com/getredash/redash/issues/1414 which is the ability to pass the frequency when creating "Word Cloud"

## Related Tickets & Documents

Related GitHub issue https://github.com/getredash/redash/issues/1414

Issue #3301, opened 2019-01-17
Migrate Visualizations to React


arikfr:
This is an issue to track the migration of visualizations code to React:

- [x] Migrate visualizations wrapper (#3493)
- [x] Migrate visualizations:
    - [x] Counter (#4106)
    - [x] Chart
        - [x] Renderer (#4130)
        - [x] Editor (getredash/redash#4139)
    - [x] Table
        - [x] Renderer (#3963)
        - [x] Editor (getredash/redash#4175)
    - [x] Pivot Table
        - [x] Renderer (#4133)
        - [x] Editor
    - [x] Choropleth (getredash/redash#4313)
    - [x] Cohort (getredash/redash#4270)
    - [x] Funnel (getredash/redash#4267)
    - [x] Map (getredash/redash#4278)
    - [x] Sankey
        - [x] Renderer (getredash/redash#4255)
        - [x] Editor
    - [x] Sunburst
        - [x] Renderer (getredash/redash#4259)
        - [x] Editor
    - [x] Word Cloud (#3930)
    - [x] Box Plot (#3948)

arikfr:
We should consider using https://github.com/plotly/react-pivottable#readme when switching the Pivot Table visualization to React.

arikfr:
🎉 All the visualizations are now in React. 🎉

Issue #1414, opened 2016-11-21
Word Cloud visualization improvements


arikfr:
Additional configuration options:

* Limit on words
* Ability to pass the word frequency

dermatou:
Not quite sure how the word cloud is supposed to be working:
here, with 20 results, word cloud is not properly counting or showing the results: https://demo.redash.io/queries/4263
Same query, with 15 results, all fine: https://demo.redash.io/queries/4262

Is this a current limitation or would this be a bug?

satheeshravir:
Any plan to implement this feature?

diegodorgam:
hey, I'm currently trying to get keywords and frequency count from a mongodb database, returning a dataset with a [count , tag] columns, is that correct?

mohamagdy:
@arikfr one of the configuration options is implemented as part of https://github.com/getredash/redash/pull/3668. 

The bug @dermatou addressed most probably fixed also as part of https://github.com/getredash/redash/pull/3668 by setting a max value for the size of the texts in the Word Cloud.

mosheka:
> Not quite sure how the word cloud is supposed to be working:
> here, with 20 results, word cloud is not properly counting or showing the results: https://demo.redash.io/queries/4263
> Same query, with 15 results, all fine: https://demo.redash.io/queries/4262
> 
> Is this a current limitation or would this be a bug?

We face the same issues. Word count works with 75 lines (5 unique values).
When we have 100 lines (same 5 unique values) only a single random word is presented.

mosheka:
Was accepted as a Redash issue: https://github.com/getredash/redash/issues/3730

PR #3930, opened 2019-06-25
Migrate Word Cloud visualization to React

kravets-levko:

- [x] Refactor
- [x] Feature
- [x] Bug Fix

## Description

- [x] Make visualization responsive (scale contents to fit container) - this should also fix getredash/redash#3730:
  - [x] fit visualization to its container;
  - [x] watch container size changes and update visualization;
  - [x] handle very long words (with high frequencies).
- [x] Implement getredash/redash#1414 (replace getredash/redash#3668):
  - [x] pass word frequencies in another column;
  - [x] limits on min/max word frequency and min/max word length.
- [x] Tests

getredash/redash#3301 (Migrate Visualizations to React -> Word Cloud)

Query page:
image

New editor options:
image
image

Dashboard:
image

### DIFF
--- a/client/app/lib/hooks/useQueryResult.js
+++ b/client/app/lib/hooks/useQueryResult.js
@@ -2,9 +2,9 @@ import { useState, useEffect } from 'react';
 
 function getQueryResultData(queryResult) {
   return {
-    columns: queryResult ? queryResult.getColumns() : [],
-    rows: queryResult ? queryResult.getData() : [],
-    filters: queryResult ? queryResult.getFilters() : [],
+    columns: queryResult ? queryResult.getColumns() || [] : [],
+    rows: queryResult ? queryResult.getData() || [] : [],
+    filters: queryResult ? queryResult.getFilters() || [] : [],
   };
 }
 
--- a/client/app/lib/hooks/useQueryResult.js
+++ b/client/app/lib/hooks/useQueryResult.js
@@ -2,9 +2,9 @@ import { useState, useEffect } from 'react';
 
 function getQueryResultData(queryResult) {
   return {
-    columns: queryResult ? queryResult.getColumns() || [] : [],
-    rows: queryResult ? queryResult.getData() || [] : [],
-    filters: queryResult ? queryResult.getFilters() || [] : [],
+    columns: (queryResult && queryResult.getColumns()) || [],
+    rows: (queryResult && queryResult.getData()) || [],
+    filters: (queryResult && queryResult.getFilters()) || [],
   };
 }
 
--- a/client/app/services/resizeObserver.js
+++ b/client/app/services/resizeObserver.js
@@ -1,0 +1,31 @@
+const items = new Map();
+
+function checkItems() {
+  items.forEach((item, node) => {
+    const bounds = node.getBoundingClientRect();
+    // convert to int (because these numbers needed for comparisons), but preserve 1 decimal point
+    const width = Math.round(bounds.width * 10);
+    const height = Math.round(bounds.height * 10);
+
+    if (
+      (item.width !== width) ||
+      (item.height !== height)
+    ) {
+      item.width = width;
+      item.height = height;
+      item.callback(node);
+    }
+  });
+
+  setTimeout(checkItems, 100);
+}
+
+checkItems(); // ensure it was called only once!
+
+export default function observe(node, callback) {
+  if (!items.has(node)) {
+    items.set(node, { callback });
+    return () => items.delete(node);
+  }
+  return () => {};
+}
--- a/client/app/services/resizeObserver.js
+++ b/client/app/services/resizeObserver.js
@@ -1,31 +1,50 @@
+/* global ResizeObserver */
+
+function observeNative(node, callback) {
+  if ((typeof ResizeObserver === 'function') && node) {
+    const observer = new ResizeObserver(() => callback()); // eslint-disable-line compat/compat
+    observer.observe(node);
+    return () => observer.disconnect();
+  }
+  return null;
+}
+
 const items = new Map();
 
 function checkItems() {
-  items.forEach((item, node) => {
-    const bounds = node.getBoundingClientRect();
-    // convert to int (because these numbers needed for comparisons), but preserve 1 decimal point
-    const width = Math.round(bounds.width * 10);
-    const height = Math.round(bounds.height * 10);
+  if (items.size > 0) {
+    items.forEach((item, node) => {
+      const bounds = node.getBoundingClientRect();
+      // convert to int (because these numbers needed for comparisons), but preserve 1 decimal point
+      const width = Math.round(bounds.width * 10);
+      const height = Math.round(bounds.height * 10);
 
-    if (
-      (item.width !== width) ||
-      (item.height !== height)
-    ) {
-      item.width = width;
-      item.height = height;
-      item.callback(node);
-    }
-  });
+      if (
+        (item.width !== width) ||
+        (item.height !== height)
+      ) {
+        item.width = width;
+        item.height = height;
+        item.callback(node);
+      }
+    });
 
-  setTimeout(checkItems, 100);
+    setTimeout(checkItems, 100);
+  }
 }
 
-checkItems(); // ensure it was called only once!
-
-export default function observe(node, callback) {
-  if (!items.has(node)) {
+function observeFallback(node, callback) {
+  if (node && !items.has(node)) {
+    const shouldTrigger = items.size === 0;
     items.set(node, { callback });
+    if (shouldTrigger) {
+      checkItems();
+    }
     return () => items.delete(node);
   }
-  return () => {};
+  return null;
+}
+
+export default function observe(node, callback) {
+  return observeNative(node, callback) || observeFallback(node, callback) || (() => {});
 }
--- a/client/app/visualizations/word-cloud/Editor.jsx
+++ b/client/app/visualizations/word-cloud/Editor.jsx
@@ -15,19 +15,21 @@ export default function Editor({ options, data, onOptionsChange }) {
       <div className="form-group">
         <label className="control-label" htmlFor="word-cloud-words-column">Words Column</label>
         <Select
+          data-test="WordCloud.WordsColumn"
           id="word-cloud-words-column"
           className="w-100"
           value={options.column}
           onChange={column => optionsChanged({ column })}
         >
           {map(data.columns, ({ name }) => (
-            <Select.Option key={name}>{name}</Select.Option>
+            <Select.Option key={name} data-test={'WordCloud.WordsColumn.' + name}>{name}</Select.Option>
           ))}
         </Select>
       </div>
       <div className="form-group">
         <label className="control-label" htmlFor="word-cloud-frequencies-column">Frequencies Column</label>
         <Select
+          data-test="WordCloud.FrequenciesColumn"
           id="word-cloud-frequencies-column"
           className="w-100"
           value={options.frequenciesColumn}
@@ -35,7 +37,7 @@ export default function Editor({ options, data, onOptionsChange }) {
         >
           <Select.Option key="none" value=""><i>(count word frequencies automatically)</i></Select.Option>
           {map(data.columns, ({ name }) => (
-            <Select.Option key={'column-' + name} value={name}>{name}</Select.Option>
+            <Select.Option key={'column-' + name} value={name} data-test={'WordCloud.FrequenciesColumn.' + name}>{name}</Select.Option>
           ))}
         </Select>
       </div>
@@ -46,6 +48,7 @@ export default function Editor({ options, data, onOptionsChange }) {
         <Grid.Row gutter={15} type="flex">
           <Grid.Col span={12}>
             <InputNumber
+              data-test="WordCloud.WordLengthLimit.Min"
               className="w-100"
               placeholder="Min"
               min={0}
@@ -55,6 +58,7 @@ export default function Editor({ options, data, onOptionsChange }) {
           </Grid.Col>
           <Grid.Col span={12}>
             <InputNumber
+              data-test="WordCloud.WordLengthLimit.Max"
               className="w-100"
               placeholder="Max"
               min={0}
@@ -71,6 +75,7 @@ export default function Editor({ options, data, onOptionsChange }) {
         <Grid.Row gutter={15} type="flex">
           <Grid.Col span={12}>
             <InputNumber
+              data-test="WordCloud.WordCountLimit.Min"
               className="w-100"
               placeholder="Min"
               min={0}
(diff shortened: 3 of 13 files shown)


kravets-levko:
@arikfr @ranbena @gabrieldutra This PR is ready for review - I'll appreciate any feedback. I'm still testing it, and will remove WIP label once I'm finally happy with everything.

kravets-levko:
@mohamagdy Since this PR is going to replace yours (getredash/redash#3668) I'll appreciate your feedback as well (sorry, cannot request your review from GitHub UI)

client/app/visualizations/word-cloud/Renderer.jsx
```diff
@@ -0,0 +1,187 @@
...
+import resizeObserver from '@/services/resizeObserver';
+import { RendererPropTypes } from '@/visualizations';
+
+import './renderer.less';
+
+function computeWordFrequencies(rows, column) {
+  const result = {};
+
+  each(rows, (row) => {
+    const wordsList = row[column].toString().split(/\s/g);
```
gabrieldutra: Better to use Lodash's `toString` and `split`  here, otherwise:

toString
(null column values)

This actually happens in the current version as well :point_up_2:

client/app/visualizations/chart/index.js
```diff
@@ -311,6 +311,7 @@ export default function init(ngModule) {
     registerVisualization({
       type: 'CHART',
       name: 'Chart',
+      isDefault: true,
```
arikfr: This is to make sure Chart is always used first?

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
-        onChange={onColumnChanged}
-      >
-        {map(data.columns, ({ name }) => (
-          <Option key={name}>{name}</Option>
-        ))}
-      </Select>
-    </div>
+    <React.Fragment>
+      <div className="form-group">
+        <label className="control-label" htmlFor="word-cloud-words-column">Words column</label>
```
arikfr:
```suggestion
        <label className="control-label" htmlFor="word-cloud-words-column">Words Column</label>
```

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
+          value={options.column}
+          onChange={column => optionsChanged({ column })}
+        >
+          {map(data.columns, ({ name }) => (
+            <Select.Option key={name}>{name}</Select.Option>
+          ))}
+        </Select>
+      </div>
+      <div className="form-group">
+        <label className="control-label" htmlFor="word-cloud-frequencies-column">Frequencies column</label>
```
arikfr:
```suggestion
        <label className="control-label" htmlFor="word-cloud-frequencies-column">Frequencies Column</label>
```

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
+        >
+          <Select.Option key="none" value=""><i>(count word frequencies automatically)</i></Select.Option>
+          {map(data.columns, ({ name }) => (
+            <Select.Option key={'column-' + name} value={name}>{name}</Select.Option>
+          ))}
+        </Select>
+      </div>
+      <div className="form-group">
+        <label className="control-label" htmlFor="word-cloud-word-length-limit">
+          Show only words with length between
```
arikfr:
```suggestion
          Words Length Limit
```

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
+        </label>
+        <Grid.Row gutter={15} type="flex">
+          <Grid.Col span={12}>
+            <InputNumber
+              className="w-100"
+              placeholder="no limit"
```
arikfr:
```suggestion
              placeholder="No Limit"
```

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
+              min={0}
+              value={options.wordLengthLimit.min}
+              onChange={value => optionsChanged({ wordLengthLimit: { min: value > 0 ? value : null } })}
+            />
+          </Grid.Col>
+          <Grid.Col span={12}>
+            <InputNumber
+              className="w-100"
+              placeholder="no limit"
```
arikfr:
```suggestion
              placeholder="No Limit"
```

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
+              min={0}
+              value={options.wordLengthLimit.max}
+              onChange={value => optionsChanged({ wordLengthLimit: { max: value > 0 ? value : null } })}
+            />
+          </Grid.Col>
+        </Grid.Row>
+      </div>
+      <div className="form-group">
+        <label className="control-label" htmlFor="word-cloud-word-length-limit">
+          Show only words with count between
```
arikfr:
```suggestion
          Frequencies Limit
```

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
+        </label>
+        <Grid.Row gutter={15} type="flex">
+          <Grid.Col span={12}>
+            <InputNumber
+              className="w-100"
+              placeholder="no limit"
```
arikfr:
```suggestion
              placeholder="No Limit"
```

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
+              min={0}
+              value={options.wordCountLimit.min}
+              onChange={value => optionsChanged({ wordCountLimit: { min: value > 0 ? value : null } })}
+            />
+          </Grid.Col>
+          <Grid.Col span={12}>
+            <InputNumber
+              className="w-100"
+              placeholder="no limit"
```
arikfr:
```suggestion
              placeholder="No Limit"
```

client/app/visualizations/chart/index.js
kravets-levko: Yes. For some reason Word Cloud became first and tests started failing. I think it's useful anyway.

arikfr: I think the same happened in the details viz pull request. In the future it's better to push such changes in a separate PR as they are easier to merge and useful beyond this one (no need to do this now though).

kravets-levko: NP, I'll pick this fix to another PR

client/app/visualizations/word-cloud/Renderer.jsx
kravets-levko: @gabrieldutra Done :+1: Thanks!

client/app/visualizations/chart/index.js
kravets-levko: @arikfr getredash/redash#3944

client/app/lib/hooks/useQueryResult.js
```diff
@@ -2,9 +2,9 @@ import { useState, useEffect } from 'react';
 
 function getQueryResultData(queryResult) {
   return {
-    columns: queryResult ? queryResult.getColumns() : [],
-    rows: queryResult ? queryResult.getData() : [],
-    filters: queryResult ? queryResult.getFilters() : [],
+    columns: queryResult ? queryResult.getColumns() || [] : [],
```
ranbena: A bit prettier:
```js
columns: queryResult && queryResult.getColumns() || [],
```

client/app/services/resizeObserver.js
```diff
@@ -0,0 +1,31 @@
...
+      item.width = width;
+      item.height = height;
+      item.callback(node);
+    }
+  });
+
+  setTimeout(checkItems, 100);
+}
+
+checkItems(); // ensure it was called only once!
```
ranbena: @kravets-levko are you sure having this here won't cause any trouble? (e.g. won't run if not needed, would run only once)

client/app/visualizations/word-cloud/Editor.jsx
```diff
@@ -1,30 +1,95 @@
...
+        >
+          <Select.Option key="none" value=""><i>(count word frequencies automatically)</i></Select.Option>
+          {map(data.columns, ({ name }) => (
+            <Select.Option key={'column-' + name} value={name}>{name}</Select.Option>
+          ))}
+        </Select>
+      </div>
+      <div className="form-group">
+        <label className="control-label" htmlFor="word-cloud-word-length-limit">
+          Words Length Limit
```
ranbena: So if I understand correctly, it's a limit range and the fields are "from" and "to"?

kravets-levko: Yes, correct

client/app/services/resizeObserver.js
kravets-levko: We discussed it some time ago with @arikfr (he had the same questions). Currently I left this code almost unchanged (just moved it to service to be able to use from React), maybe later I'll fix it to be more clear.

Answering your questions: this function runs only once (because `require` will cache module) and then it will re-schedule itself every 100ms. There are no performance issues (when elements list is empty - this function returns in a millisecond).

client/app/visualizations/word-cloud/Editor.jsx
ranbena: In that case, the UI is a bit unclear. How about a slider instead? https://ant.design/components/slider/

ranbena: (Or descriptive labels next to the numberinputs)

kravets-levko: Not sure if slider is a good choice because technically there is no upper limit (user can use any positive number which even may be larger than any value from dataset); also, lower limit is `1` and there is "null" value (no limit). Maybe, instead of slider and labels I can set placeholders to `min`/`max` (instead of current `no limit`)?

kravets-levko: @ranbena Done (this and previous suggestion)

ranbena:
I think the only thing missing now is a simple Percy test for it.

kravets-levko:
You're right :+1: Will implement some

kravets-levko:
@ranbena Added some tests

client/cypress/integration/visualizations/word_cloud_spec.js
```diff
@@ -0,0 +1,83 @@
...
+      cy.getByTestId('ExecuteButton').click();
+    });
+  });
+
+  it('creates visualization with automatic word frequencies', () => {
+    const visualizationName = 'Word Cloud (auto)';
+
+    cy.getByTestId('NewVisualization').click();
+    cy.getByTestId('VisualizationType').click();
+    cy.getByTestId('VisualizationType.WORD_CLOUD').click();
```
ranbena: Can swap for `cy.clickThrough()`

client/cypress/integration/visualizations/word_cloud_spec.js
```diff
@@ -0,0 +1,83 @@
...
+    cy.getByTestId('WordCloud.WordLengthLimit.Min').clear().type('4');
+    cy.getByTestId('WordCloud.WordLengthLimit.Max').clear().type('5');
+    cy.getByTestId('WordCloud.WordCountLimit.Min').clear().type('1');
+    cy.getByTestId('WordCloud.WordCountLimit.Max').clear().type('3');
+
+    cy.getByTestId('VisualizationPreview').find('svg text').should('have.length', 2);
+
+    cy.getByTestId('EditVisualizationDialog').contains('button', 'Save').click();
+    cy.getByTestId('QueryPageVisualizationTabs').contains('li', visualizationName).should('exist');
+  });
```
ranbena: Tests are great. A few things I would do different.
1. Swap some code blocks with `cy.clickThrough()` unless you think it's unnecessary.
2. Extract the `vis name -> save -> tab should exist` to a single test - bc it's testing a feature independent of the test's focus (which is the wordcloud outcome based on config). Also, might not even be a wordcloud feature but a visualization feature so might need to be extracted further.
3. Testing word length is good, but since size, color and position are also factors we'd like to protect from regression, I suggest a percy snapshot instead.

@kravets-levko wdyt?

kravets-levko: @ranbena Updated test according to your suggestions. Can you please take another look?

kravets-levko:
@ranbena Thanks! There are some issues with Percy (visualization is messed up on screenshots), I'm trying to figure this out. After that I'll merge this PR.

kravets-levko:
@arikfr @ranbena @gabrieldutra 

There is an issue with this visualization and Percy, which was quite hard to figure out (even required a help from Percy support). Word Cloud used an "Impact" font to compute layout, and later the same font was assigned to all words. But "Impact" is not available on all machines, so browser uses some fallback. To take a screenshot of a page, Percy takes a complete DOM snapshot with all loaded resources and inline CSS, copies it to own machines and then loads in own browsers. On CI machines word cloud was computed using some default font, and after copying DOM to Percy - restored with another default font available on that machine:

image
_Words are messed up because Percy's browser uses different font with a different size of letter_

The only solution which 100% worked was to use some fixed font, which will be exactly the same on all machines. I tried Roboto because we already have it (however, don't use since #2081), but before I merge it - I decided to start a discussion to be sure that I'm not missing anything important.

image
_Using Roboto font_

arikfr:
@kravets-levko good catch! How about we define a Percy only CSS rule that uses Roboto, but keep things for users unchanged?

kravets-levko:
@arikfr We need to detect Cypress env as well - because Roboto should be loaded there for proper layout calculation. Not sure if there is any "legal" way to target Cypress and Percy :thinking:

kravets-levko:
I think I can add some env to Cypress dockerfile and then use it to conditionally add fonts during webpack build: https://github.com/getredash/redash/blob/d7b03bac02d946134b816bcf815f3a2bbb76bd98/.circleci/docker-compose.cypress.yml#L12-L16

@gabrieldutra Do you have any better ideas?

arikfr:
> We need to detect Cypress env as well - because Roboto should be loaded there for proper layout calculation. Not sure if there is any "legal" way to target Cypress and Percy 🤔

In the test code we have full control of the browser, why not change the CSS from code? Feels safer and easier than passing around from the Dockerfile.

kravets-levko:
That's interesting idea. If Percy will pick that injected font - then it should work. Need to try

kravets-levko:
@arikfr Yay, it works! I needed to update webpack config to copy Roboto files to `dist/` without referencing them anywhere, but they won't load anywhere except of CI. Can you please take a last look before I merge it? Thanks!

arikfr:
Looks good 👍

PR #3931, opened 2019-06-26
[WIP] Add support for Vega and Vega-lite

ktmud:

- [x] Feature

## Description

Vega and Vega-Lite allow a user to specify a JSON configuration
to declaratively define a complex interactive visualization.

It could theoretically become the foundation of many visualizations
we have in Redash.

This PR adds Vega and Vega-Lite and a new visualization type,
and allows users to directly edit a Vega or Vega-Lite spec in either JSON or YAML. A lot of code is borrowed from the official [Vega Editor](https://vega.github.io/editor/).

While most of the functionalities are working, this is not ready for shipment yet. Some known issues:

- [x] Vega allows only fixed container size. We need to figure out how to do autoresize.
- [x] Not all themes are useful to the majority of the users.
- [x] Haven't tested tooltips.
- [x] <strike>Vega-Embed allows downloading PNG files, which can be very useful.</strike> Will add support in future PR.
- [x] <strike>Was trying to add [monaco-yaml](https://github.com/pengx17/monaco-yaml) so the YAML editor can also have the very nice IntelliSense, including autocomplete and linting. But that Monaco plugin isn't working well with ESM yet.</strike> Will consider migrating to Monaco in future work.

https://discuss.redash.io/t/add-vega-grammar-support-for-advanced-visualizations/3020

### Visualization editor:

Snip20190625_1

### Visualization renderer (with a different Vega theme)
Snip20190625_3


client/app/visualizations/vega/theme.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * Vega Theme tweaked for Redash
+ */
```
ktmud: Currently this is just a Copy & Paste of https://github.com/vega/vega-themes/blob/master/src/theme-quartz.ts
still needs more tweaking.

kravets-levko:
Hi @ktmud! Thanks for your incredible work! I know that this PR is still WIP, but I have some notices that would be better to fix on early stage:

- don't pass `query` object to each visualization; you need only URLs there - so pass them; probably, `data` object is a good place;
- please try to reuse existing dependencies; I like Monaco as well, but Redash already has an editor library (Ace), so please use it; also, do you really need to update all that dependencies?

Also, please see my comments to the code. Thanks!

client/app/services/query.js
```diff
@@ -115,7 +115,7 @@ export class Parameter {
       }
 
       if (isArray(value) && (value.length === 2)) {
-        value = [moment(value[0]), moment(value[1])];
+        value = [moment(value[0], 'YYYY-MM-DD'), moment(value[1], 'YYYY-MM-DD')];
```
kravets-levko: What does it fix?

client/app/visualizations/VisualizationRenderer.jsx
```diff
@@ -8,7 +8,7 @@ import { registeredVisualizations, VisualizationType } from './index';
 
 function combineFilters(localFilters, globalFilters) {
   // tiny optimization - to avoid unnecessary updates
-  if ((localFilters.length === 0) || (globalFilters.length === 0)) {
+  if (localFilters.length === 0 || globalFilters.length === 0) {
```
kravets-levko: Why is it needed?

client/app/visualizations/index.js
```diff
@@ -80,11 +86,16 @@ export function registerVisualization(config) {
 -----------------------------------------------------------*/
 
 export function getDefaultVisualization() {
-  return find(registeredVisualizations, visualization => !visualization.isDeprecated);
+  return find(
+    registeredVisualizations,
+    viz => !viz.isDeprecated && viz.isDefault,
+  );
```
kravets-levko: Now this code assumes that at least one visualization has `isDefault` attribute; not sure if it's right or not, but I'd add a fallback for this case (if no `isDefault` visualization - return first non-deprecated). That's just a suggestion

redash/handlers/admin.py
```diff
@@ -33,7 +35,7 @@ def outdated_queries():
 
     response = {
         'queries': QuerySerializer(outdated_queries, with_stats=True, with_last_modified_by=False).serialize(),
-        'updated_at': manager_status['last_refresh_at'],
+        'updated_at': manager_status.get('last_refresh_at', datetime.now()),
```
kravets-levko: What does it fix?

redash/handlers/query_results.py
```diff
@@ -262,10 +262,13 @@ def get(self, query_id=None, query_result_id=None, filetype='json'):
 
             filename = get_download_filename(query_result, query, filetype)
 
-            response.headers.add_header(
-                "Content-Disposition",
-                'attachment; filename="{}"'.format(filename.encode("utf-8"))
-            )
+            if filetype in ('csv', 'json') and request.args.get('download') in ['false', '0']:
+                response.headers.set('Content-Type', 'text/plain; charset=UTF-8')
```
kravets-levko: Can you please explain these changes (what are they needed for?). Thanks!

ktmud:
Hi Levko, thanks for reviewing!

> * please try to reuse existing dependencies; I like Monaco as well, but Redash already has an editor library (Ace), so please use it; also, do you really need to update all that dependencies?

Can we actually replace Ace Editor with Monaco? Monaco comes with some highly desirable advanced features such as inline validation of JSON schemas, which is quite essential for a Vega editor.

I had to upgrade Webpack because the latest version of Monaco wouldn't build.

client/app/services/query.js
ktmud: It fixes a warning message from Moment (it now requires you to supply the date format without relying on the defaults).

client/app/visualizations/VisualizationRenderer.jsx
ktmud: Auto fixed by Prettier.

client/app/visualizations/index.js
ktmud: Makes sense. Maybe we should pre-sort the registered visualizations and make the default visualization a constant instead of calling `find()` every time?

This is just a quick hack to avoid Vega showing up as the default editor. If there's any other way of doing it, I'd be happy to avoid adding a new config field.

redash/handlers/admin.py
ktmud: It throws 500 errors when Redis is restarted and no manager status was found. (Not relevant to this PR, but fixed as it is minor.)

redash/handlers/query_results.py
ktmud: This is for allowing users to click on the data source url that populated in Vega schema and review the data directly in browser without automatically downloading a file.

kravets-levko:
> Can we actually replace Ace Editor with Monaco?

Well, yes, we can, of course. But definitely not within this PR - we use a lot of Ace features in query editor and all of them should be kept with Monaco.

redash/handlers/query_results.py
kravets-levko: I think in this case you have to set a proper MIME type for each extension. E.g. I have a chrome extensions that pretty-print both CSV and JSON (see screenshot with example), and both that extensions detect their targets by MIME type, so changes like this will break them.

image

client/app/visualizations/index.js
kravets-levko: Adding new config field is a good solution, let's keep it. Calling `find` every time is not a problem because ATM it is used only in Edit Visualization dialog and is called there only once (on dialog open).

ktmud:
> > Can we actually replace Ace Editor with Monaco?
> 
> Well, yes, we can, of course. But definitely not within this PR - we use a lot of Ace features in query editor and all of them should be kept with Monaco.

Sure! I think I'll remove Monaco dependency for now and replace it with Ace Editor. There was some building issues with Monaco as well (it takes forever to build client/dist for Docker image).

redash/handlers/query_results.py
ktmud: I also have a JSON prettier extension installed, but couldn't find a good one for CSV. Which one are you using?

Since this change added a new URL parameter "download=false", it shouldn't break any existing links. I'm inclined to keep it. What do you think?

client/app/visualizations/index.js
kravets-levko: @ktmud I created a separate PR for this feature (getredash/redash#3944) because we needed it right now in few other PRs to fix tests.

client/app/visualizations/vega/Editor.jsx
```diff
@@ -0,0 +1,269 @@
...
+// };
+
+/**
+ * Add additional language support for the text editor
+ */
+// function setupEditor() {
+//   monaco.languages.json.jsonDefaults.setDiagnosticsOptions(monacoDiagnostics);
+//   monaco.languages.registerDocumentFormattingEditProvider('json', jsonFormatter);
+//   monaco.languages.registerDocumentFormattingEditProvider('yaml', yamlFormatter);
+// }
```
ktmud: Keeping the code for Monaco here for now, so later migration will be easier (if we are going to do it)

client/app/visualizations/vega/Renderer.jsx
```diff
@@ -0,0 +1,229 @@
...
+    let vPadding = this.props.fromEditor ? 40 : 5;
+    if (typeof spec.padding === 'number') {
+      hPadding += 2 * spec.padding;
+      vPadding += 2 * spec.padding;
+    } else if (isObject(spec.padding)) {
+      hPadding += (spec.padding.left || 0) + (spec.padding.right || 0);
+      vPadding += (spec.padding.top || 0) + (spec.padding.bottom || 0);
+    }
+    width = width || specWidth || Math.max(parentSize.width - hPadding, 100);
+    height = height || specHeight || Math.min(400, Math.max(parentSize.height - vPadding, 300));
```
ktmud: By default, a Vega visualization has min-height 300px and max-height 400px. This is inline with what Plot.ly currently has, although I was not able to find where we set the height for Plot.ly...

kravets-levko: We don't set height for Plotly charts: on Dashboard page chart fits it's container in widget, and on Query page it has fixed height (IIRC, `500px`) and width `100%`.

ktmud:
@kravets-levko @arikfr I think this PR is more or less ready now. Can you please take another look? It's actually quite fun to play with Vega specs.

client/app/visualizations/vega/Renderer.jsx
```diff
@@ -0,0 +1,249 @@
...
+import ResizeObserver from 'resize-observer-polyfill';
+import { isObject } from 'lodash';
+import React from 'react';
+import Vega from 'react-vega';
+import * as vl from 'vega-lite';
+// import * as YAML from 'js-yaml';
+import { Handler } from 'vega-tooltip';
+import { Alert, Icon } from 'antd';
+import LZString from 'lz-string';
+import memoize from 'memoize-one';
```
kravets-levko: [`_.memoize`](https://lodash.com/docs/4.17.11#memoize)? [`useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo)?

ktmud: `_.memoize` does not cache only one return value, having higher risk of memory leaks.
`useMemo` does not work with class components. I don't see the added benefits of rewriting my class to a function component.

`memoize-one` only has 45 LOC.

client/app/visualizations/vega/Renderer.jsx
ktmud: So the default height `500px` comes from Plot.ly library itself?

I implemented `100%` height for Vega charts everywhere.

bazhenov:
Hi.

We are very interested in this functionality, as standard visualisations types are quite limiting for us. What is the status of this MR and do you need any help? Maybe some functionality should be implemented before it considered "done"?

kravets-levko:
Hi @ktmud! I'd like to review your PR once more, so can you please resolve conflicts? Also, I'd ask you to remove changes that are not related to subject of this PR and commented out fragments. Thank you! 

P.S. Please ping me when it's ready for review :bowing_man:

ktmud:
Hi all, I just pushed a quick fix to resolve the conflicts. Things might still not work. Will pick this up again during the weekend.

ktmud:
I haven't worked on Redash for a while, so things might be a little slow.

itssimon:
This looks amazing @ktmud! Do you think you'll get a chance to get this across the line? Happy to help!

ktmud:
Just updated the branch and it is now 90% done.

It depends on #5104 , but the dependency can be easily removed if that PR isn't going anywhere.

vadimcn:
Hi!   What happened to this PR?   Is it abandoned?

guidopetri:
@ktmud , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

PR #3932, opened 2019-06-26
Parameter spec fixes

rauchy:

- [x] Bug Fix

## Description
I've noticed that `parameter_spec.js` is failing due to a couple of reasons:
- Parameter not showing up for text, date and datetime - this is due to the query options not being set and the parameter does not load up when hitting `/source`.
- Date and datetime were pointing to the 15 of the _current_ month, which made the assertions fail.

This makes the test pass on my machine, but I suspect it might still fail on CI due to locale issues so I want to see it pass first. Also I might have missed some context here, so please feel free to reject this PR entirely.

### DIFF
--- a/client/cypress/integration/query/parameter_spec.js
+++ b/client/cypress/integration/query/parameter_spec.js
@@ -210,12 +210,13 @@ describe('Parameter', () => {
         .find('input')
         .click();
 
+      const now = Cypress.moment();
       cy.get('.ant-calendar-date-panel')
         .contains('Now')
         .click();
 
       cy.getByTestId('DynamicTable')
-        .should('contain', Cypress.moment().format('YYYY-MM-01 HH:mm'));
+        .should('contain', now.format('YYYY-MM-01 HH:mm'));
     });
   });
 });
--- a/client/cypress/integration/query/parameter_spec.js
+++ b/client/cypress/integration/query/parameter_spec.js
@@ -201,8 +201,11 @@ describe('Parameter', () => {
       cy.get('.ant-calendar-ok-btn')
         .click();
 
-      cy.getByTestId('DynamicTable')
-        .should('contain', Cypress.moment().format('YYYY-MM-15 HH:mm'));
+      cy.get('.ant-calendar-input').then(($input) => {
+        const now = Cypress.moment($input.val(), 'DD/MM/YY HH:mm');
+        cy.getByTestId('DynamicTable')
+          .should('contain', now.format('YYYY-MM-15 HH:mm'));
+      });
     });
 
     it('shows the current datetime after clicking in Now', () => {
--- a/client/cypress/integration/query/parameter_spec.js
+++ b/client/cypress/integration/query/parameter_spec.js
@@ -116,6 +116,11 @@ describe('Parameter', () => {
       const queryData = {
         name: 'Date Parameter',
         query: "SELECT '{{test-parameter}}' AS parameter",
+        options: {
+          parameters: [
+            { name: 'test-parameter', title: 'Test Parameter', type: 'date' },
+          ],
+        },
       };
 
       createQuery(queryData, false)
@@ -156,6 +161,11 @@ describe('Parameter', () => {
       const queryData = {
         name: 'Date and Time Parameter',
         query: "SELECT '{{test-parameter}}' AS parameter",
+        options: {
+          parameters: [
+            { name: 'test-parameter', title: 'Test Parameter', type: 'datetime-local' },
+          ],
+        },
       };
 
       createQuery(queryData, false)
--- a/client/cypress/integration/query/parameter_spec.js
+++ b/client/cypress/integration/query/parameter_spec.js
@@ -1,4 +1,3 @@
-import moment from 'moment';
 import { createQuery } from '../../support/redash-api';
 
 describe('Parameter', () => {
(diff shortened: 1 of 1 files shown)


client/cypress/integration/query/parameter_spec.js
```diff
@@ -142,7 +154,7 @@ describe('Parameter', () => {
         .click();
 
       cy.getByTestId('DynamicTable')
-        .should('contain', '15/01/19');
+        .should('contain', moment().format('15/MM/YY'));
```
ranbena: Can't import moment unless it's preinstalled (I assume) but you can use `Cypress.moment` which is a built in util.

client/cypress/integration/query/parameter_spec.js
```diff
@@ -124,7 +135,8 @@ describe('Parameter', () => {
         SaveParameterSettings 
       `);
 
-      const now = new Date(2019, 0, 1).getTime(); // January 1, 2019 timestamp
+      const now = new Date();
+      now.setDate(1);
```
ranbena: Why wouldn't the original code work though?
Perhaps cause `cy.clock()` had better be called before `cy.visit()`? (https://docs.cypress.io/api/commands/clock.html#Behavior)

gabrieldutra: I have no idea why, but `cy.clock` has a weird behavior, when writing the test if I put it before `cy.visit` the page wouldn't open :grimacing:.

I think it's a good option to check the format for now as it's not really necessary to have the date frozen.

client/cypress/integration/query/parameter_spec.js
rauchy: ya learn somethin' every day

client/cypress/integration/query/parameter_spec.js
rauchy: Yeah I couldn't find a way to fix the `cy.clock` behavior, so basically just made sure that a date change *within the current month and year* is applied correctly.

ranbena: As long as the test does what it should 👍

ranbena:
@rauchy I got this error locally. It failed on last test because if `mm` discrepancy.

Screen Shot 2019-06-27 at 9 28 22

rauchy:
@ranbena 4a70913 makes sense?

ranbena:
> @ranbena [4a70913](https://github.com/getredash/redash/commit/4a709138c0849182881f880315ae99681420d2f4) makes sense?

Probably, but I think it would be better to just compare the input value and query result, no?

ranbena:
This should work:

```js
    it.('shows the current datetime after clicking in Now', () => {
      cy.getByTestId('ParameterName-test-parameter')
        .find('input')
        .as('Input') // <-- save for later
        .click();

      cy.get('.ant-calendar-date-panel')
        .contains('Now')
        .click();

      cy.get('@Input').then(($input) => {
        const value = Cypress.moment($input.val());  // <-- get input value

        cy.getByTestId('DynamicTable')
          .should('contain', value.format('YYYY-DD-MM HH:mm')); 
      });
    });
```

rauchy:
@ranbena I was just about to commit something very similar and ask which solution you think is better. Your solution will almost work - one thing needs to change:

```js
const value = Cypress.moment($input.val());
```

should be

```js
const value = Cypress.moment($input.val(), 'DD/MM/YY HH:mm');
```

Which do you think is better? 4a70913 or 9a912ce? If they both work, I tend to think 4a70913 is simpler.

gabrieldutra:
https://github.com/getredash/redash/commit/9a912ced7699f05f80fd3fb05415e612505e0dc0 sems more reliable as you'll have the exact value you triggered query execution

ranbena:
@rauchy A few requests before signing this off:
1. Implement the same solution to the test above it (same likelihood of flakiness).
2. I guess there's no more reason to do mocking for this test (remove `cy.clock` usage)
3. And therefore, no more need to hardcode the formatting (`YYYY-MM-01` -> `YYYY-MM-DD`).

rauchy:
> Implement the same solution to the test above it (same likelihood of flakiness).

Sure, tha'ts done in 81cf07a.

> I guess there's no more reason to do mocking for this test (remove cy.clock usage)
> And therefore, no more need to hardcode the formatting (YYYY-MM-01 -> YYYY-MM-DD).

Well, except on the 15th of each month, in which the current date will equal the selected date, and then you could have a false positive. I tend to think that |head scratching due to mocking time| < |head scratching due to portential temporary false positive| 😝but if you think otherwise I'll gladly change this.

ranbena:
> @ranbena
> 
> > Implement the same solution to the test above it (same likelihood of flakiness).
> 
> Sure, tha'ts done in [81cf07a](https://github.com/getredash/redash/commit/81cf07a3ad68298de61f13f22f19b78837747a7b).
> 
> > I guess there's no more reason to do mocking for this test (remove cy.clock usage)
> > And therefore, no more need to hardcode the formatting (YYYY-MM-01 -> YYYY-MM-DD).
> 
> Well, except on the 15th of each month, in which the current date will equal the selected date, and then you could have a false positive. I tend to think that |head scratching due to mocking time| < |head scratching due to portential temporary false positive| 😝but if you think otherwise I'll gladly change this.

In that case, mocking is required, but hardcoding ain't, right?

rauchy:
> In that case, mocking is required, but hardcoding ain't, right?

I think both are required - mocking is required to force the date to be one that isn't the hardcoded one, so even if the test runs on the 15th, mocking the date would force to system to behave as if it is the 1st, then the test would choose the 15th, which will trigger a value change and verify it with a hardcoded value.

PR #3940, opened 2019-06-28
fix: SparkSQL query refresh in the hive datasource

QuantamHD:
- [x] Bug Fix

## Description

When using SparkSQL and the Hive Datasource Redash fails to refresh the query because SparkSQL has different column names than Hive. Here is the map below.

Hive has,
Database Name => 'database_name'
Tablename => 'tab_name'
Column Name => 'field',

While SparkSQL has,
Database Name => 'databaseName'
Tablename => 'tableName'
Column Name => 'col_name',

In this PR I detect which one the user is connecting to and then change the names dynamically.

image

### DIFF
--- a/redash/query_runner/hive_ds.py
+++ b/redash/query_runner/hive_ds.py
@@ -120,7 +120,7 @@ class Hive(BaseSQLQueryRunner):
 
         if isinstance(query, str):
             query_result = self._run_query_internal(query)
-        elif isinstance():
+        elif isinstance(query, list):
             query_result = query
 
         return filter(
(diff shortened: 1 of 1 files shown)


alexisdondon:
Hi, what's the process if we want to make this PR integrated to next version?

I've also the same problem i have made another response by adding a custom sparkSql query runner that just exactly the same that the hive one but changing column names ttps://github.com/InseeFrLab/redash/blob/master.

susodapop:
Go ahead and open a new PR. We are cleaning up old ones that we never got to. We're relaunching this effort and reviewing much faster than in the past.

guidopetri:
@QuantamHD , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

PR #3942, opened 2019-06-29
Fix: waiting tasks are not shown in admin view

arikfr:

- [x] Bug Fix

## Description

Waiting tasks were not showing in admin view, because it was expecting them to have a different name for the state. To avoid breaking existing users of this API, we're just translating the state name in the UI.

### DIFF
--- a/client/app/pages/admin/Tasks.jsx
+++ b/client/app/pages/admin/Tasks.jsx
@@ -13,6 +13,8 @@ import { $http } from '@/services/ng';
 import recordEvent from '@/services/recordEvent';
 import { routesToAngularRoutes } from '@/lib/utils';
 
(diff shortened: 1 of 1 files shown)


client/app/pages/admin/Tasks.jsx
```diff
@@ -13,6 +13,13 @@ import { $http } from '@/services/ng';
 import recordEvent from '@/services/recordEvent';
 import { routesToAngularRoutes } from '@/lib/utils';
 
+function stateName(state) {
+  if (state === 'waiting_in_queue') {
```
rauchy: A comment on backward compatibility feels appropriate here.

arikfr: Done.

PR #3943, opened 2019-06-30
Add option to hide Pivot Table totals

arikfr:

- [x] Feature

## Description

Added toggles to hide/show the totals row/column in a pivot table.

https://discuss.redash.io/t/cohort-heatmap-displayed-incorrectly/4075/6

### DIFF
--- a/client/app/visualizations/pivot/Editor.jsx
+++ b/client/app/visualizations/pivot/Editor.jsx
@@ -14,10 +14,10 @@ export default function Editor({ options, onOptionsChange }) {
         <Switch
           data-test="PivotEditor.HideControls"
           id="pivot-show-controls"
-          checked={options.controls.enabled}
-          onChange={enabled => updateOptions({ controls: { enabled } })}
+          checked={!options.controls.enabled}
+          onChange={enabled => updateOptions({ controls: { enabled: !enabled } })}
         />
-        <span className="m-l-10">Hide Pivot Controls</span>
+        <span className="m-l-10">Show Pivot Controls</span>
       </label>
       <label className="d-flex align-items-center" htmlFor="pivot-show-row-totals">
         <Switch
--- a/client/app/visualizations/pivot/Editor.jsx
+++ b/client/app/visualizations/pivot/Editor.jsx
@@ -9,7 +9,7 @@ export default function Editor({ options, onOptionsChange }) {
   };
 
   return (
-    <div className="form-group">
+    <div className="form-group m-t-30">
       <label className="d-flex align-items-center" htmlFor="pivot-show-controls">
         <Switch
           data-test="PivotEditor.HideControls"
--- a/client/app/visualizations/pivot/Editor.jsx
+++ b/client/app/visualizations/pivot/Editor.jsx
@@ -1,9 +1,21 @@
 import { merge } from 'lodash';
-import React from 'react';
+import React, { useState } from 'react';
 import Switch from 'antd/lib/switch';
 import { EditorPropTypes } from '@/visualizations';
 
 export default function Editor({ options, onOptionsChange }) {
+  const [hideRowTotals, setHideRowTotals] = useState(
+    options.rendererOptions &&
+      options.rendererOptions.table &&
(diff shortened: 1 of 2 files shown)


client/app/visualizations/pivot/Editor.jsx
```diff
@@ -1,23 +1,57 @@
...
+import React, { useState } from 'react';
 import Switch from 'antd/lib/switch';
 import { EditorPropTypes } from '@/visualizations';
 
 export default function Editor({ options, onOptionsChange }) {
+  const [hideRowTotals, setHideRowTotals] = useState(
+    options.rendererOptions &&
+      options.rendererOptions.table &&
+      options.rendererOptions.table.rowTotals === false,
+  );
```
kravets-levko: You can add defaults to visualization's default options and then omit all that tests: https://github.com/getredash/redash/blob/9a4433bf685eec1f3fb24639ee11ab3e9fd8b115/client/app/visualizations/pivot/index.js#L13-L17

Also, no need to use state here: just use `const hideRowTotals = options.rendererOptions.table.rowTotals === false;` or pass `checked={options.rendererOptions.table.rowTotals === false}` where needed (the same for second option).

arikfr: > You can add defaults to visualization's default options and then omit all that tests:

Two questions:

1. Will this work with existing visualizations?
2. `rendererOptions` might have other properties set by PivotTable.js so I don't want to override it. Will it get merged or overriden?

> Also, no need to use state here: just use...

This is what I did at first, but then it wouldn't update then I change it...

kravets-levko: 1. yes
2. it's merged, so everything should be fine.

> it wouldn't update then I change it...

That's strange :thinking: Try to add defaults to `DEFAULT_OPTIONS` and check will it work or not (it may be related to `_.merge`). If it will not work - leave everything as is, I'll check check it when migrating this to React.

ranbena:
From a UX stand point, since the switch element has a connotation of "on means more, off means less", I would consider flipping the bool - "Show Pivot controls" instead of "Hide".

ranbena:
Also, give some spacing between the vis name text field and the switches.

kravets-levko:
@ranbena Maybe, change "Hide pivot controls" as well? :thinking:

ranbena:
> @ranbena Maybe, change "Hide pivot controls" as well? 🤔

Not sure what you mean - that's what I suggested, no? 🤔

arikfr:
> From a UX stand point, since the switch element has a connotation of "on means more, off means less", I would consider flipping the boo

I did just that when implementing @kravets-levko feedback.

I did leave the old switch as is, but will see if it's not too much effort to change it as well.

kravets-levko:
> Not sure what you mean - that's what I suggested, no?

For some reason I decided that you mean only two new options :sweat_smile:

client/app/visualizations/pivot/Editor.jsx
arikfr: Done. https://github.com/getredash/redash/pull/3943/commits/e313dcbab1b4a8673174c257ea23ac1bad470603

arikfr:
Flipped them all: https://github.com/getredash/redash/pull/3943/commits/d602f0dbfb53b84a328238f26f31d06fcc734732.

image

client/app/visualizations/pivot/Editor.jsx
```diff
@@ -11,13 +11,29 @@ export default function Editor({ options, onOptionsChange }) {
   return (
     <div className="form-group">
```
ranbena:
```suggestion
    <div className="form-group m-t-30">
```

arikfr:
> Give it a top margin

Given. Although it something that we should probably handle at the container component instead of each editor on its own.

ranbena:
> > Give it a top margin
> 
> Given. Although it something that we should probably handle at the container component instead of each editor on its own.

Considered it. But most of the visualization editors have tabs which I'd rather not affect.

PR #3948, opened 2019-07-03
Migrate Box Plot visualization to React

kravets-levko:

- [x] Refactor
- [x] Feature

## Description

- [x] Make visualization responsive (scale contents to fit container) + watch container resize
- [x] Tests

getredash/redash#3301 (Migrate Visualizations to React -> Box Plot)

Query page:

image

Dashboard:

image


ranbena:
Ok I gotta ask - if this visualization is marked as `(deprecated)`, wouldn't this be the opportunity to actually drop it?

kravets-levko:
I think someone may still use it, therefore we don't drop it. Anyway, it's already migrated (and it didn't took much time)...

client/app/visualizations/box-plot/renderer.less
```diff
@@ -0,0 +1,4 @@
+.box-plot-deprecated-visualization-container {
+  overflow: hidden;
+  height: 400px;
```
ranbena: Looks like the original is `500px` so this renders smaller.

kravets-levko: I decided to set it to `400px` because other visualizations use this size as well. It's only for query page, on dashboard pages now it will fit container (like chart, counter, word cloud, etc. - see screenshots). I think it's not an issue at all.

client/app/visualizations/box-plot/Renderer.jsx
```diff
@@ -0,0 +1,175 @@
...
+    .call(chart);
+}
+
+export default function Renderer({ data, options }) {
+  const [container, setContainer] = useState(null);
+
+  useEffect(() => {
+    if (container) {
+      render(container, data, options);
+      return resizeObserver(container, () => {
```
ranbena: For clarity better add the `const unwatch =`;

kravets-levko: My IDE says that there is unnecessary variable :smile: But yeah, I'll add it, and for word cloud as well :+1:

client/app/visualizations/box-plot/renderer.less
ranbena: Don't you think it looks a bit too small when in query view mode.
Screen Shot 2019-07-03 at 20 20 42

ranbena: (Actually, the best option is actually to make this query view resize observable)

ranbena: (different scope)

client/cypress/integration/visualizations/box_plot_spec.js
```diff
@@ -0,0 +1,68 @@
...
+      'BoxPlot.XAxisLabel': 'X Axis',
+      'BoxPlot.YAxisLabel': 'Y Axis',
+    });
+
+    // Wait for proper initialization of visualization
+    cy.wait(500); // eslint-disable-line cypress/no-unnecessary-waiting
+
+    cy.getByTestId('VisualizationPreview').find('svg').should('exist');
+
+    cy.percySnapshot('Visualizations - Box Plot');
```
ranbena: Snapshot doesn't show the full width of the viz?

percy

kravets-levko: It's a problem of Percy: we run our tests with viewport width 1000px or something like that, and visualization fits this size; then Percy takes a complete DOM snapshot and restored on their machines, using different dimensions (1280px and 375px, the same DOM snapshot for both); there is no JS that can update visualization, so it uses coordinates generated on Cypress machines. It's actually the same issue that with Word Cloud fonts.

kravets-levko: Actually, an idea crossed my mind - I can set Cypress viewport, and then ask Percy to take snapshot - one by one for each width. Need to try, it may work

ranbena: So weird.
And `cy.viewport(1280...)` is no help?

ranbena: > Actually, an idea crossed my mind - I can set Cypress viewport, and then ask Percy to take snapshot - one by one for each width. Need to try, it may work

If it works, can global override in percy.yml https://docs.percy.io/v1/docs/configuration

client/cypress/integration/visualizations/box_plot_spec.js
```diff
@@ -58,11 +58,11 @@ describe('Box Plot', () => {
...
-    // Wait for proper initialization of visualization
-    cy.wait(500); // eslint-disable-line cypress/no-unnecessary-waiting
-
     cy.getByTestId('VisualizationPreview').find('svg').should('exist');
 
-    cy.percySnapshot('Visualizations - Box Plot');
+    cy.viewport(1280, 1024);
+    cy.wait(500); // eslint-disable-line cypress/no-unnecessary-waiting
+    cy.screenshot();
+    cy.percySnapshot('Visualizations - Box Plot', { widths: [1280] });
```
ranbena: Oh I thought you were gonna go for:
`cy.percySnapshot('Visualizations - Box Plot', { widths: [1000] });`

kravets-levko: There is `cy.viewport` few lines above - I want to take snapshots of the same sizes that we have for other tests (1280 and 375px). But ATM I'm just checking if it will work - if yes, I'll improve it

ranbena: You could also set `cypress.json` to `viewportWidth: 1000` so no percy shots get invalidated.

gabrieldutra: `1280px` seems a better number :slightly_smiling_face: (since the Cypress viewport is defined prior to it, should make no difference). In case it works it's probably better to set it globally for Cypress.

kravets-levko:
@ranbena @gabrieldutra Updating viewport definitely helped:

image

So now few questions.

1. Do we need other width except of 1280? _(my opinion - yes)_
2. Should I change Cypress viewport (probably for all tests), and leave Percy defaults, or to tell Percy to use Cypress browser size (1000px in our case)? _(my opinion - change Cypress to match Percy)_
3. Do not touch anything, but create a helper to take few Percy screenshots using different viewports.

WDYT?

arikfr:
> I think someone may still use it, therefore we don't drop it. Anyway, it's already migrated (and it didn't took much time)...

It's actually almost unused. Considering you already did the work to migrate it, then let's keep it. But how about we update the list not to show deprecated types in the list when creating a new visualization but render it if it exists?

kravets-levko:
@arikfr It's good idea, but I'll do it in another PR

gabrieldutra:
> Do we need other width except of 1280?

In this case I'd say only 1280px is enough. 

For the default we should think if Mobile Percy diff is that useful in most cases as it doubles the number of screenshots taken on each build. Imo we could leave only the 1280px as a default and trigger mobile ones only for specific pages.

> Should I change Cypress viewport (probably for all tests), and leave Percy defaults, or to tell Percy to use Cypress browser size (1000px in our case)?

I prefer `1280px` too.

> Do not touch anything, but create a helper to take few Percy screenshots using different viewports 

:thinking: I would do it in case there are Snapshots that requires more than only the `1280px` with the same issue. With only the that resolution, the global cypress option should do fine.

arikfr:
>  It's good idea, but I'll do it in another PR

Definitely.

kravets-levko:
It had to be pretty small and quick refactor/React migration PR of a not-so-important component, but we have what we have :sweat_smile: 

image

client/cypress/support/commands.js
```diff
@@ -30,3 +30,16 @@ Cypress.Commands.add('fillInputs', (elements) => {
...
+
+Cypress.Commands.add(
+  'percyMultipleSnapshots',
+  (name, { widths = [1280, 375], minHeight = 1024, takeSnapshotDelay = null } = {}) => {
+    each(widths, (width) => {
+      cy.viewport(width, minHeight);
+      if (takeSnapshotDelay > 0) {
+        cy.wait(takeSnapshotDelay); // eslint-disable-line cypress/no-unnecessary-waiting
+      }
+      cy.percySnapshot(name, { widths: [width], minHeight });
```
gabrieldutra: I'm curious to see if Percy will recognize this as the same image with different widths

kravets-levko: he-he, I'm curious as well :slightly_smiling_face: we will know it in few minutes :slightly_smiling_face:

kravets-levko: huh, no, it does recognize only first screenshot. crap :thinking:

cypress.json
```diff
@@ -9,5 +9,7 @@
   "videosFolder": "client/cypress/videos",
   "supportFile": "client/cypress/support/index.js",
   "defaultCommandTimeout": 20000,
-  "requestTimeout": 15000
+  "requestTimeout": 15000,
+  "viewportWidth": 1280,
```
kravets-levko: @gabrieldutra Look: I changed default viewport dimensions for Cypress, and I expected to see a lot of diffs, but see only few ones (https://percy.io/Redash/redash/builds/2137256). Does that mean that pages that we already captured were properly resized by CSS? (sorry if it's a dumb question, I'm just trying to not miss anything)

gabrieldutra: > Does that mean that pages that we already captured were properly resized by CSS? 

Exactly, the only pages that showed a diff are ones that rely on JavaScript to determine size, the others should not change as in Percy the width is the same.

kravets-levko: > ...regarding the two other visualizations that rely on JS...

I hope so :thinking:

kravets-levko: Upd: yes, seems they where scaled using vertical dimensions (because their height > width), and since container has fixed height they're unchanged. The first one has width > height and therefore was scaled using horizontal dimensions which depend on viewport size.

kravets-levko:
Okay, I thinks it's ready for review and (if everyone is happy with code) to merge.

client/cypress/integration/visualizations/box_plot_spec.js
gabrieldutra: One thing: in this case due to the width difference between Cypress and Percy the mobile resolution is not that reliable, how about we keep only the desktop one?

```suggestion
    cy.percySnapshot('Visualizations - Box Plot', { widths: [1280] });
```

client/cypress/integration/visualizations/box_plot_spec.js
```diff
@@ -38,6 +38,8 @@ const SQL = `
 `;
 
 describe('Box Plot', () => {
+  const viewportWidth = Cypress.config('viewportWidth');
```
gabrieldutra: Cool :+1:

kravets-levko: I really didn't want to hardcode it :grin:

PR #3951, opened 2019-07-04
Fix: Details Visualization: render date/time values as strings and not epoch time

arikfr:

- [x] Bug Fix

## Description

Because we were just converting values to strings (`'' + value`) date/time values were rendered as epoch times:

image

This Pull Request fixes this by properly converting them using the system settings date/time format.

There is some logic duplication here with the Table visualization, which we should remove in the future.

### DIFF
--- a/client/app/visualizations/details/DetailsRenderer.jsx
+++ b/client/app/visualizations/details/DetailsRenderer.jsx
@@ -1,9 +1,26 @@
 import React, { useState } from 'react';
-import map from 'lodash/map';
+import { map, mapValues, keyBy } from 'lodash';
+import moment from 'moment';
 import { RendererPropTypes } from '@/visualizations';
+import { clientConfig } from '@/services/auth';
 import Pagination from 'antd/lib/pagination';
 
 import './details.less';
+
+function renderValue(value, type) {
+  const formats = {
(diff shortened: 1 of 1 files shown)


client/app/visualizations/details/DetailsRenderer.jsx
```diff
@@ -1,17 +1,36 @@
...
 import './details.less';
 
+function renderValue(value, type) {
+  const formats = {
+    date: clientConfig.dateFormat,
+    datetime: clientConfig.dateTimeFormat,
+  };
+
+  if (type === 'date' || type === 'datetime') {
+    if (moment.isMoment(value)) {
```
kravets-levko: I think `if (moment.isMoment(value)) {` is enough, previous `if` is almost useless.

arikfr: It's relevant to make sure we have a known format type. But it's true that it will only be a moment if it's one of those (in normal situations).

arikfr:
> I think, the next level will be an editor similar to Table's one - where user can manually choose how to display each column

I agree. I think it will make sense to reuse parts of the Table editor and rendering here, we can look into this when migrating the Table editor. Although wouldn't want to delay it because of this.

PR #3952, opened 2019-07-04
Support multi-select in parameters

### LINKED ISSUE #3053: Support for multi select in parameters
When using a Dropdown List type of parameter, the user should be able to define the parameter to take multiple values.

This will require some additional configuration of how to serialize the values. We can probably do away with adding a checkbox for "quote each value".

gabrieldutra:

- [x] Feature

## Description
This one is to add an option to allow multiple values in Dropdown parameters (see preview below :)).

- [x] Make 'quote each value' an option (and let the user customize the separator?)
- [x] Apply this to Query based Dropdown parameters
- [x] Styling updates (select `maxWidth`, etc)
- [x] Add tests 
- [x] Refresh Schedule

Closes #3053

**Dropdown**
multi-select-enum

**Query Based Dropdown**
multi-select-query


gabrieldutra:
There are currently two places that invoke `mustache_render`, one is the ParametizedQuery, and the other is this task:

https://github.com/getredash/redash/blob/45a3b727301c9039f437b243de5916b1b4cfd865/redash/tasks/queries.py#L192-L197

The List logic is in the ParametizedQuery, so I was thinking about turn the above to use it instead of a direct call to `mustache_render`. This way the ParametizedQuery would centralize logic related to Queries with parameters :grin:.

As I'm not that familiar with the backend, where would you put this `resolve_parameter_values` logic @arikfr or @rauchy, is the Parametized Query model a good place?

ranbena:
Since this component's dimensions could be as high and wide as the content it holds, we should limit it so layout doesn't break. Here are some suggestions:

CSS limits
```css
{
  max-width: 350px;
  max-height: 69px;
  overflow: auto;
}
```
or Ant options (I think it's better):
```jsx
<Select
  maxTagCount="3"
  maxTagTextLength="10"
  maxTagPlaceholder={num => `+${num.length} more`}
/>
```

arikfr:
> This way the ParametizedQuery would centralize logic related to Queries with parameters 😁.

Sounds like a good idea 👍 But probably worth waiting to hear what @rauchy thinks.

> or Ant options (I think it's better):

Sounds like the better option. Not sure we should change the default, btw. And we should reuse this for multi-filters as well (but in a separate PR).

redash/models/parameterized_query.py
```diff
@@ -92,6 +99,12 @@ def _is_date_range(obj):
         return False
 
 
+def _is_dropdown_value_valid(value, dropdown_options):
+    if isinstance(value, list):
```
rauchy: This function is basically `issubset`, no?

```python
return set(map(unicode, value)).issubset(set(dropdown_options))
```

redash/models/parameterized_query.py
```diff
@@ -36,6 +36,13 @@ def dropdown_values(query_id):
     return map(pluck, data["rows"])
 
 
+def resolve_parameter_values(parameters):
```
rauchy: Again, I feel that this function name is too generic for what it does. There's no "resolving" here, but rather flattening of multiple values.

client/app/components/ParameterValueInput.jsx
```diff
@@ -128,31 +130,41 @@ export class ParameterValueInput extends React.Component {
   }
 
   renderEnumInput() {
-    const { value, onSelect, enumOptions } = this.props;
+    const { value, onSelect, enumOptions, allowMultipleValues } = this.props;
     const enumOptionsArray = enumOptions.split('\n').filter(v => v !== '');
     return (
       <Select
         className={this.props.className}
+        mode={allowMultipleValues ? 'multiple' : 'default'}
```
rauchy: Why is the boolean2string conversion needed here? Why not simply pass `allowMultipleValues`?

rauchy: My bad, I didn't notice it was an Antd component restriction.

gabrieldutra:
Taking another step and considering Serialization Settings I wanted to make it custom (separator, prefix and suffix to each value options at least backend-wise for future support to any UI). While exploring options I realized there is another case that will need to be handled:

When running adhoc queries (or running a modified unsaved existing one) there is no information about parameters (only their values), thus no information about the Serialization options. As it's a "trusted" route for queries I think I'll just replicate serialization logic in the frontend and send the value as string for such cases.

An alternative is to always serialize this on frontend and have a "reverse operation" to validate the values in backend based on settings.

arikfr:
> Taking another step and considering Serialization Settings I wanted to make it custom (separator, prefix and suffix to each value options at least backend-wise for future support to any UI)

In what cases one would want different prefix and suffix?

I'm not sure if at this stage we should let the user control the separator anyway, but we can have a setting for it just in case without visible UI.

Also, if the user picks quotes they need to pick the escaping form. Most SQL implementation escape a single quote by double it (`select ''''` returns `'`). But in other cases you might want to prefix with a slash.

> An alternative is to always serialize this on frontend and have a "reverse operation" to validate the values in backend based on settings.

The regex seems to work, but this is tricky and can lead to various issues down the road.

Maybe instead we can just skip validation in such case (after confirming the user has access to the data source), like we do with dropdowns in general?

gabrieldutra:
> In what cases one would want different prefix and suffix?

I have no idea, but could be a use case at some point, my plan was to have it only on backend for now :sweat_smile:. The same for the separator.

For UI I was thinking about showing quote options in a Select.

> The regex seems to work, but this is tricky and can lead to various issues down the road.

That was sort of my concern too, a Regex that takes strings on it doesn't seem to be something that reliable.

> Maybe instead we can just skip validation in such case (after confirming the user has access to the data source), like we do with dropdowns in general?

For such cases that you only have the parameter values the validation is already skipped the problem is that you'd only have the list of values, not information on how to serialize data.

As the validation is already skipped I think the best option will be to replicate the serialization logic on frontend and just send it as String.

gabrieldutra:
@arikfr, @ranbena this is what I was thinking for the Parameter Settings Dialog, you can give your comments :slightly_smiling_face:

multi-enum

---

> Also, if the user picks quotes they need to pick the escaping form.

Why is that :thinking:? Isn't the purpose of the quotes to allow scenarios like this one:
multi-query
?

ranbena:
Very nice @gabrieldutra. I have a few wording suggestions.

* None (default)
* Single quotation mark
* Double quotation mark

Screen Shot 2019-07-16 at 12 22 00

The hint at the bottom is essential to understanding the feature and its result.
A small change to the sentence:

Screen Shot 2019-07-16 at 12 25 01

client/app/components/ParameterValueInput.jsx
```diff
@@ -128,31 +131,41 @@ export class ParameterValueInput extends React.Component {
...
     const enumOptionsArray = enumOptions.split('\n').filter(v => v !== '');
     return (
       <Select
-        className={this.props.className}
+        className={classNames({ 'parameter-multi-select': allowMultipleValues }, this.props.className)}
+        mode={allowMultipleValues ? 'multiple' : 'default'}
+        optionFilterProp="children"
         disabled={enumOptionsArray.length === 0}
-        defaultValue={value}
+        value={value}
```
gabrieldutra: Otherwise the changes from `setValue` wouldn't affect here.

Other inputs remain the same as is, one need is to see the behavior after #3907.

redash/models/parameterized_query.py
```diff
@@ -105,7 +126,7 @@ def apply(self, parameters):
             raise InvalidParameterError(invalid_parameter_names)
         else:
             self.parameters.update(parameters)
-            self.query = mustache_render(self.template, self.parameters)
+            self.query = mustache_render(self.template, join_parameter_list_values(parameters, self.schema))
```
gabrieldutra: Needed to provide the `schema` to get the prefix, suffix and separator options.

arikfr:
> > Also, if the user picks quotes they need to pick the escaping form.

> Why is that 🤔? Isn't the purpose of the quotes to allow scenarios like this one:

Think of a case where you quote with `'` but the content has `'` in it (like `Arik's Book`). It's not very common, but can happen.

gabrieldutra:
> Think of a case where you quote with ' but the content has ' in it (like
Arik's Book). It's not very common, but can happen.

I'll look for some easy options around it, but isn't that a case that can
already happen with the current single valued Dropdown?

arikfr:
> but isn't that a case that can already happen with the current single valued Dropdown?

Indeed it is. Let's table it then and move on. :+1:

gabrieldutra:
@ranbena good thing to have let the `value` state (updated on `componentDidUpdate`) on `ParameterValueInput`, I actually needed it here :sweat_smile:.

> Indeed it is. Let's table it then and move on.

And no easy solution, for each data source is a different thing. This is probably together with "make parameters safe".

ranbena:
Screen Shot 2019-07-17 at 19 39 40

arikfr:
> And no easy solution, for each data source is a different thing. This is probably together with "make parameters safe".

The easy solution in this case is just to let the user select from a few options. For some it might be a trial and error process to find the right setting, but it's a reasonable thing.

client/cypress/integration/query/parameter_spec.js
```diff
@@ -129,14 +129,14 @@ describe('Parameter', () => {
         .find('.ant-select')
         .click();
 
-      cy.contains('li.ant-select-dropdown-menu-item', 'value1')
+      cy.contains('li.ant-select-dropdown-menu-item', 'value2')
```
gabrieldutra: Needed to change to `value2` since now I make sure the first dropdown option is selected when value is invalid or empty.

client/app/components/QueryBasedParameterInput.jsx
```diff
@@ -50,27 +52,37 @@ export class QueryBasedParameterInput extends React.Component {
       if (this.props.queryId === queryId) {
         this.setState({ options, loading: false });
 
-        const found = find(options, option => option.value === this.props.value) !== undefined;
-        if (!found && isFunction(this.props.onSelect)) {
-          this.props.onSelect(options[0].value);
+        if (this.props.mode === 'multiple' && isArray(this.props.value)) {
+          const optionValues = map(options, option => option.value);
+          const validValues = intersection(this.props.value, optionValues);
+          if (validValues.length !== this.props.value.length) {
```
ranbena: Not sure what this means, can you explain or add a comment?

client/app/components/parameters.js
```diff
@@ -38,7 +38,7 @@ function ParametersDirective($location) {
         EditParameterSettingsDialog
           .showModal({ parameter })
           .result.then((updated) => {
-            scope.parameters[index] = extend(parameter, updated);
+            scope.parameters[index] = extend(parameter, updated).setValue(updated.value);
```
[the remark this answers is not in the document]
ranbena: Cool cool cool

client/app/services/query.js
```diff
@@ -83,13 +84,13 @@ export class Parameter {
...
+  getValue(joinListValue = false) {
+    return this.constructor.getValue(this, joinListValue);
   }
 
-  static getValue(param) {
-    const { value, type, useCurrentDateTime } = param;
-    const isEmptyValue = isNull(value) || isUndefined(value) || (value === '');
+  static getValue(param, joinListValue = false) {
+    const { value, type, useCurrentDateTime, multiValuesOptions } = param;
+    const isEmptyValue = isNull(value) || isUndefined(value) || (value === '') || (isArray(value) && value.length === 0);
```
ranbena: I think this is now exactly how `_.isEmpty()` works 😆

client/app/services/query.js
```diff
@@ -83,13 +84,13 @@ export class Parameter {
     return isNull(this.getValue());
   }
 
-  getValue() {
-    return this.constructor.getValue(this);
+  getValue(joinListValue = false) {
```
ranbena: How about this, so that things remain generic?
```js
getValue(extra = {}) {
```

client/cypress/integration/query/parameter_spec.js
```diff
@@ -145,7 +145,7 @@ describe('Parameter', () => {
           .find('.ant-select')
           .click();
 
-        cy.contains('li.ant-select-dropdown-menu-item', 'value1')
+        cy.contains('li.ant-select-dropdown-menu-item', 'value2')
           .click();
       });
     });
```
ranbena: No new tests?

client/app/services/query.js
gabrieldutra: Almost there :joy:, `_.isEmpty` considers Numbers as empty. It may be interesting to have the "Empty Parameter" rule explicit, though.

gabrieldutra:
Important note I forgot :grimacing:: I didn't code the "Refresh Schedule" part, because I couldn't get Parameterized Queries to be scheduled (Python debugger + logs to track), I suspect this is a bug, but I'm not 100% aware of how it's supposed to work ^^ (if you schedule a parameterized query on preview, it won't update the results).

shenoyroopesh:
Would this work for selecting multiple dates as well? Or only for non-date related parameters?

gabrieldutra:
> How could this work for selecting multiple dates as well?

Hi @shenoyroopesh :slightly_smiling_face:, this will work for the existing Dropdown and Query Based Dropdown parameters. Dates won't be supported directly, but Query Based Dropdown parameters make any type of data possible with [this](https://discuss.redash.io/t/time-intervals-as-date-parameters-last-day-last-week-last-month/4120) workaround (with actual dates and not intervals). Unfortunately, it would still not be that convenient.

arikfr:
@shenoyroopesh can you share where you might want to select multiple dates but a date range won't be applicable? (just trying to understand the use case)

redash/tasks/queries.py
```diff
@@ -193,7 +193,7 @@ def refresh_queries():
                 if query.options and len(query.options.get('parameters', [])) > 0:
                     query_params = {p['name']: p.get('value')
                                     for p in query.options['parameters']}
-                    query_text = mustache_render(query.query_text, query_params)
+                    query_text = query.parameterized.apply(query_params).query
```
gabrieldutra: Refresh schedule updated. Tested that by changing the `outdated_queries` method locally so all queries were outdated :slightly_smiling_face:.

gabrieldutra:
@arikfr this one is ready to me, but it's worth to give a check on the backend part :)

arikfr:
:tada:

PR #3955, opened 2019-07-07
Add options to hide different parts of embed UI (parameters, title, link to query)

arikfr:

- [x] Feature

## Description

Add options to hide different parts of embed UI (parameters, title, link to query).

### DIFF
--- a/client/app/components/queries/visualization-embed.html
+++ b/client/app/components/queries/visualization-embed.html
@@ -1,5 +1,5 @@
-<div class="tile m-l-10 m-r-10 embed__vis" data-test="VisualizationEmbed">
-  <div class="embed-heading p-t-10 p-b-10 p-r-15 p-l-15" ng-if="!$ctrl.hideHeader">
+<div class="tile m-l-10 m-r-10 p-t-10 embed__vis" data-test="VisualizationEmbed">
+  <div class="embed-heading p-b-10 p-r-15 p-l-15" ng-if="!$ctrl.hideHeader">
     <h3>
       <img ng-src="{{$ctrl.logoUrl}}" style="height: 24px;"/>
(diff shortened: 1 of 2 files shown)


ranbena:
@arikfr This is for snaps?

arikfr:
@ranbena we might use it for snaps in the future, but this is actually for a customer who wanted cleaner embeds.

I think that later on we will add this as switches to the embed dialog to make it discoverable. But this was an easy way to deliver this.

client/app/components/queries/visualization-embed.html
```diff
@@ -32,9 +32,9 @@ <h3>
           <span am-time-ago="$ctrl.queryResult.getUpdatedAt()" ng-if="!$ctrl.loading"></span>
           <rd-timer from="$ctrl.refreshStartedAt" ng-if="$ctrl.loading"></rd-timer>
         </a>
+        <span class="small visible-print"><i class="zmdi zmdi-time-restore"></i> {{$ctrl.queryResult.getUpdatedAt() | dateTime}} UTC</span>
```
ranbena: Please give this `margin-left: 5px` so it's not so close to the timeago text.

arikfr: What element should receive the extra margin?

ranbena:
```suggestion
        <span class="small visible-print m-l-5"><i class="zmdi zmdi-time-restore"></i> {{$ctrl.queryResult.getUpdatedAt() | dateTime}} UTC</span>
```

arikfr: But this shows up only when the other is hidden or I am misunderstanding something?

ranbena: Nope it's me who misunderstood how this works.

PR #3958, opened 2019-07-07
Allow calling query results endpoint without parameters.

arikfr:

- [x] Bug Fix

## Description

Allow calling `POST api/queries/<id>/results` with no parameters.


arikfr:
@rauchy I don't remember if it was when I tried using the API and got frustrated by having to pass an empty body or found it in Sentry.

PR #3961, opened 2019-07-07
Add additional Celery config options

arikfr:

- [x] Feature
- [x] Other

## Description

1. Change default Celery worker prefetch multiplier to 1.
2. Add additional Celery configuration options.

### DIFF
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -41,6 +41,11 @@ CELERY_SSL_CONFIG = {
     'ssl_keyfile': os.environ.get("REDASH_CELERY_BROKER_SSL_KEYFILE"),
 } if CELERY_BROKER_USE_SSL else None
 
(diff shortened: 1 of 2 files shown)


redash/settings/__init__.py
```diff
@@ -41,6 +41,11 @@
     'ssl_keyfile': os.environ.get("REDASH_CELERY_BROKER_SSL_KEYFILE"),
 } if CELERY_BROKER_USE_SSL else None
 
+CELERY_WORKER_PREFETCH_MULTIPLIER = int(os.environ.get("REDASH_CELERY_WORKER_PREFETCH_MULTIPLIER", 1))
```
rauchy: Changing this from the default of 4 to 1 would disable prefetching. Could this lead to any unexpected side-effects for anyone?

arikfr: It doesn't really disable prefetching, because Celery still prefetches one job for every worker available. So if the worker capacity is 4, there will be 4 jobs executing and 4 jobs waiting.

Regardless it will have only positive impact on anyone :-)

PR #3938, opened 2019-06-28
Add 'connection_source' property in presto query runner

iamhongri:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Query Runner (Data Source)
- [ ] New Alert Destination
- [x] Other

## Description
 - Add 'connection_source' property in presto query runner

PR #3962, opened 2019-07-08
Add an `extras` field for Presto connector

ktmud:

- [x] Feature

## Description

This adds a new "Extras" field for Presto connection, which is simply a JSON string and will be parsed as a dictionary and passed to [PyHive's `presto.connect` API](https://github.com/dropbox/PyHive/blob/6925cd74721cc474a02a6759490b2b4c8bef3831/pyhive/presto.py#L83).

The most useful utilization of this feature is to allow self-assigned SSL certificate. For example, services like [Qubole](https://docs.qubole.com/en/latest/user-guide/presto/connect-to-presto-when-ssl-enabled.html) only allows SSL connection via self-assigned CA certificate. In which case we can then add `requests_kwargs` in the extras:

```json
{"requests_kwargs":  {"verify": "~/qubole.cert"} }
```

We prefer this approach instead of the Qubole connector because Qubole charges you for every API call and the roundtrip from local EMR cluster to Qubole servers is not ideal.

This change also solves the need of adding extra fields for other connection arguments such as `Connection Source` implemented in #3938 .

One can simply set the "Extras" field to:
```json
{"source": "redash"}
```

This is an iteration of #3909. I've added type validation and help text.

Snip20190620_6

Snip20190707_6

### DIFF
--- a/client/app/components/dynamic-form/DynamicForm.jsx
+++ b/client/app/components/dynamic-form/DynamicForm.jsx
@@ -28,7 +28,7 @@ const fieldRules = ({ type, required, minLength }) => {
     jsonRule && {
       type: 'object',
       transform(x) {
-        if (x.trim()) {
+        if (x && x.trim()) {
           try {
             JSON.parse(x);
           } catch {
@@ -93,7 +93,12 @@ class DynamicForm extends React.Component {
   parseValues = (values) => {
     this.props.fields.forEach((field) => {
       if (field.type === 'json' && field.name in values) {
-        values[field.name] = JSON.parse(values[field.name]);
+        try {
+          values[field.name] = JSON.parse(values[field.name]);
+        } catch {
+          // Invalid JSON must be discarded
+          values[field.name] = null;
+        }
       }
     });
     return values;
--- a/client/app/components/dynamic-form/DynamicForm.jsx
+++ b/client/app/components/dynamic-form/DynamicForm.jsx
@@ -7,22 +7,37 @@ import Checkbox from 'antd/lib/checkbox';
 import Button from 'antd/lib/button';
 import Upload from 'antd/lib/upload';
 import Icon from 'antd/lib/icon';
-import { includes, isFunction } from 'lodash';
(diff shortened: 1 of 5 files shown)


ranbena:
Just a quick observation about the json input - since we already have ACE editor bundled (as the query source editor) perhaps leverage it for syntax highlighting?

arikfr:
@ranbena I'm not sure DynamicForm supports it and whether it worth the extra effort.

gabrieldutra:
> @ranbena I'm not sure DynamicForm supports it and whether it worth the extra effort.

It doesn't, though it's actually quite easy to change `TextArea` to `AceEditor` in here. However this will happen:

react-ace-value

As I tried this some time ago:

It seems that the AceEditor changes its value, which conflicts with Antd Form. A way I've just found to handle that is to create a "useless class component" and use it instead (doesn't work with functional :sweat_smile:).

```javascript
class AceTest extends React.Component {
  render() {
    return (<AceEditor {...this.props} />);
  }
}
```

After that would still need to align the validation mark.

HimanshPal:
@arikfr - Adding options one by one is also not suitable option since over time lot of configurations have added or deprecated/removed in Presto, Also the configuration is version dependent. Internally we done similar and so thereby easily able to upgrade presto versions with single change in config on Redash when needed.

ktmud:
@arikfr Airflow and Superset both have something similar to most connection types:
https://superset.incubator.apache.org/installation.html#deeper-sqlalchemy-integration

This is an advanced feature and will be used by advanced users only anyway. They should know what they are doing and how to upgrade. The only way I see this causes backward capability issue was when the `pyhive` API changes and we decide to upgrade it for Redash. In that case, there can always be some kind of upgrade/downgrade script with Alembic.

guidopetri:
@ktmud , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

PR #3963, opened 2019-07-08
Migrate Table visualization to React Part 1: Renderer

kravets-levko:

- [x] Refactor

## Description

- [x] Replace `<dynamic-table>` with Ant's `<Table>` component
- [x] Implement custom features:
  - [x] number, date/time, boolean formatting;
  - [x] HTML features (simple formatting, highlight links);
  - [x] Image column;
  - [x] Link column;
  - [x] JSON column;
  - [x] search in data;
  - [x] sort data by column click (multi-column sort);
- [x] Cleanup
- [x] Tests

getredash/redash#3301 (Migrate Visualizations to React -> Table)

Mostly the same as before:

image
image
image
image


kravets-levko:
@arikfr @ranbena @gabrieldutra I still need to create some tests for Table visualization, but this code is ready for review. Also, please someone who had that issue with horizontal scroll - please test it again with new component (I removed that CSS because entire `<dynamic-table>` was replaced, but I may need to restore it if that bug reproduces with Ant's Table as well). Thanks!

arikfr:
image

When the column is very wide the sort indicators are placed far from the column name. Should we place them always next to the text or to the right side of the text?

(cc: @ranbena )

client/app/visualizations/VisualizationRenderer.jsx
```diff
@@ -45,13 +46,26 @@ export function VisualizationRenderer(props) {
 
   const { showFilters, visualization } = props;
   const { Renderer, getOptions } = registeredVisualizations[visualization.type];
-  const options = getOptions(visualization.options, data);
+
+  // Avoid unnecessary updates (which may be expensive or cause issues with
+  // internal state of some visualizations like Table) - compare options deeply
+  // and use saved reference if nothing changed
+  let options = getOptions(visualization.options, data);
```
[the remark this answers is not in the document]
kravets-levko: Particular case - on dashboard page, when auto-height feature is enabled. When auto-height controller updates widget height, it triggers event handled by dashboard controller, which updates all widgets (sets new dimensions) and saves them. As a result, widget object gets updated with a new entire `visualization` field. For widget it looks like something has changed to this visualization, and React completely re-renders entire widget. In most cases it's not an issue (probably except of performance), but when table has a JSON field - you cannot expand it: when you expand JSON, height of the table changes, auto-height widget triggers event (see the rest of this trip above), at the end widget receives a new visualization object and re-renders it - this resets internal state of JSON field component.

One of possible solution was to disable that callback in non-edit mode, but then this problem stays available in editing mode. So I decided to mimic Angular's behavior here: compare options deeply and not by a reference. But there is no way to tell React that "nothing changed - stop updating", so I need to keep previous reference to options and use it when component shouldn't update. It looks a bit tricky, but works perfectly. Maybe, I just had to implement this as a hook :thinking:

kravets-levko:
@arikfr 

> When the column is very wide the sort indicators are placed far from the column name. Should we place them always next to the text or to the right side of the text?

It's default position for sort indicators in Ant (you can see that table looks the same, say, on Query or Dashboard list page). If we gonna change this - probably we should change it everywhere.

gabrieldutra:
> It's default position for sort indicators in Ant (you can see that table looks the same, say, on Query or Dashboard list page). If we gonna change this - probably we should change it everywhere.

It was changed in some recent version of Antd, in #3904 I updated Antd, you can see in [this snapshot](https://percy.io/Redash/redash/builds/2244669/view/145957390/1280?mode=diff&browser=chrome&snapshot=145957390).

kravets-levko:
@gabrieldutra Yeah, I like new ones much more :+1: But anyway, when your PR will be merged - new sort indicators will become available everywhere (and in Table visualization as well).

ranbena:
>someone who had that issue with horizontal scroll - please test it again with new component

Yes it's still an issue:

Screen Shot 2019-07-25 at 10 22 48

client/app/visualizations/VisualizationRenderer.jsx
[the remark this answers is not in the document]
arikfr: While I understand why it happens, I wouldn't expect the widget to resize when you expand a JSON field. It's probably low priority, but I wonder if there is a way to tell it that "stop auto resizing" :grimacing:

I don't think we need to "hide it" in a hook. Let's keep as is + add a link to this conversation in the comment, so we have full reference.

arikfr:
@ranbena when you move the cursor out, does it hide the scrollbar?

arikfr:
@kravets-levko considering that Ant takes care of the indicators in new version, let's keep as is. :+1:

client/app/visualizations/table/renderer.less
```diff
@@ -0,0 +1,53 @@
+.table-visualization-container {
+  .ant-pagination.ant-table-pagination {
+    float: none;
+    display: block;
+    text-align: center;
```
ranbena: Perhaps also add `margin-bottom: 0` so there's no needless spacing in the widget and full screen view.

client/app/visualizations/table/renderer.less
```diff
@@ -0,0 +1,53 @@
...
+        border-radius: @size / 2;
+        background: #c0c0c0;
+        text-align: center;
+        line-height: @size;
+        color: #fff;
+        padding: 0 @size * 1/4;
+        margin: 0 5px 0 0;
+      }
+    }
+  }
```
ranbena: A few differences in table style I noticed:

1. Cell height a pixel higher than original. This causes autoheight calculation to add a needless row ([example](https://redash-preview.netlify.com/dashboard/shared-dashboard)).
This should fix it:
```css
td {
  line-height: 1.4;
}
```
2. Ant table has a bottom border and no top border. I think the original looks better. Should change:
```less
td {
  border-bottom: 0;
  border-top: @the-variable;
}
```
3. The rules in `redash-table.less` should be revised. It's loaded but unused by table viz now.
One thing I'd like to apply is this, so the content aligns with the widget:
https://github.com/getredash/redash/blob/ed8b022e786b994e05a8087b9ae0db3a97e4da57/client/app/assets/less/redash/redash-table.less#L31-L37
4. Having the sort icons constantly visible makes for too much visual noise in widgets and dashboards.
This is my suggestion:
   1. Hidden by default `.ant-table-column-sorter .anticon.off { opacity: 0 }`
   2. Displayed on hover `thead:hover .off { opacity: 1 }`
   <br />
   

   Bonus: Find a way to display the `.off` sibling of the `.on` as well. 

5. The table header cell [in this viz](https://deploy-preview-3963--redash-preview.netlify.com/queries/106#195) for some reason breaks line.

ranbena:
> @ranbena when you move the cursor out, does it hide the scrollbar?

@arikfr usually yes, but after numerous attempts it eventually sticks.

ranbena:
Hold up! This might have been fixed final in Chrome 76 https://bugs.chromium.org/p/chromium/issues/detail?id=914844#c77

client/app/components/json-view-interactive/json-view-interactive.less
```diff
@@ -3,15 +3,15 @@
 @jvi-gutter: 20px;
 @jvi-spacing: 2px;
 
+.jvi-root {
+  display: block;
+  font-family: @font-family-monospace;
+}
```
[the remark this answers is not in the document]
kravets-levko: But without it top-most block will have a different font which is quite noticeable when you expand object/array:
image
image
(ellipsis on a top level looks differently from nested ones; square braces look differently; icon is a bit unaligned because of wrong font)

client/app/visualizations/table/renderer.less
kravets-levko: 1 and 2: should it be an app-wide fix or only for this visualization?
3, 4 - :+1: 
5: Previously headers had `white-space: no-wrap` CSS, but I noticed that it does not look good with long column names (for some column types like Number even 10-15 characters was too long). So I removed this CSS rule. But some columns (line Number, Boolean, etc.) have `width: 1%` - otherwise table will distribute available space accordingly to column's contents, and such columns will take too much space. With `width: 1%` they will be as narrow as possible, but wide enough to show the data. It works perfectly for numbers, booleans, date/time, etc. But It means that browser will wrap headers if needed to make that columns as narrow as possible.

ranbena: > 1 and 2: should it be an app-wide fix or only for this visualization?

Only for viz (tested the change app-wide and it's no good)

> I noticed that it does not look good with long column names (for some column types like Number even 10-15 characters was too long).

Can you show me an example of this?

kravets-levko: https://redash-preview.netlify.com/queries/187/source

I checked old CSS rules, and seems that previously idea was that long column titles should be truncated, but seems it didn't work.I can try to implements this: truncate long titles (show `...`), and add a tooltip with full text. WDYT?

ranbena: That would be perfect

ranbena: Actually, perhaps enable this as a viz setting?
Full text (nowrap) / shortened (ellipsis) / line break (normal).

kravets-levko: Hm, interesting idea :thinking: I'll do it if it will not take too much effort

client/app/components/json-view-interactive/json-view-interactive.less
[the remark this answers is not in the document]
ranbena: Oh. Then I guess we'll have to reapply the normal font onto it?

```css

.jvi-root {
  display: block;
  font-family: @font-family-monospace;
}

.jvi-comment {
  font-family: @font-family;
}
```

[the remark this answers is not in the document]
kravets-levko: :thinking: I'll try and see how it looks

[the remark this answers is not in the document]
arikfr: If this is the only issue left in this PR, how about we address it in a separate PR? 

This feature, awesome as it is, is the least important thing in this PR and releasing this PR sooner means less frustrated people who see partial tables...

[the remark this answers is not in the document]
kravets-levko: @arikfr It's a one-minute fix, so I'll give it a try. I plan to push all other fixes today, and then we can give another look to it and merge :+1:

client/app/visualizations/table/renderer.less
kravets-levko: @ranbena I cannot align tooltip better, but it looks much better as before (at least, it shows ellipsis as intended). I think we can do another iteration on it later:
image
image

About cell height and auto-height calculation: almost always there will be some free space below the table. The only condition when there will be no extra space - if table height + height of widget header and footer will be a multiple of 50 (height of dashboard grid row). And there are no chance to impact on it: if we change line height, for some tables that space will decrease, but at the same time for other tables it will increase. Also, keep in mind that line height depends on font chosen by browser, and for different browsers it will be different. So for now I decided to keep everything as is to avoid adding extra code that will not improve the situation.

client/app/components/json-view-interactive/json-view-interactive.less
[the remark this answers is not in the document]
kravets-levko: It looks better when comment has non-monospace font: 
image

client/app/visualizations/table/renderer.less
kravets-levko: About rules from `redash-table.less`: I restored some rules that were applied to table visualization, but this files cannot be removed ATM because it's used by some other components (visualization editors and also I saw it somewhere in `components` folder IIRC). Also, I decided to keep Ant's table styles (like cell paddings) - I think such things should be applied app-wide (e.g. on Queries/Dashboards list pages).

client/app/visualizations/VisualizationRenderer.jsx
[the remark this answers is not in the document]
kravets-levko: @arikfr 

> I wouldn't expect the widget to resize when you expand a JSON field.

Here is an example: https://redash-preview.netlify.com/dashboard/test3 IMHO it's quite natural that table (and widget) change its height when you expand JSON field. Anyway, I'm open to discussion.

client/app/visualizations/table/renderer.less
```diff
@@ -0,0 +1,101 @@
...
+
+        .ant-input-suffix .anticon {
+          cursor: auto;
+        }
+      }
+
+      .table-visualization-heading {
+        display: inline-block;
+        white-space: nowrap;
+        position: absolute;
```
ranbena: This means that the column would always be as wide as it's content - never it's title and imho it makes the th ellipsised too frequently.

I was thinking of simply giving it a max-width. Sth like:
```less
.table-visualization-heading {
  // nothing

  span {
    display: inline-block;
    max-width: 100%;
    vertical-align: middle;
  }
}

.ant-table-column-sorter {
  right: -3px;
}
.ant-table-column-has-sorters {
  padding-right: 12px !important;
}
```

### Before (lots of ellipsis):

Screen Shot 2019-07-31 at 7 45 11

### After (not as much):

Screen Shot 2019-07-31 at 7 44 21


Wdyt?

client/app/visualizations/table/renderer.less
ranbena: > if we change line height, for some tables that space will decrease, but at the same time for other tables it will increase.

> Also, keep in mind that line height depends on font chosen by browser, and for different browsers it will be different.

All true. My reasoning is to just keep it the same as before, so autoheight widgets don't become larger and mess up ppl's dashboard alignments.

client/app/visualizations/table/renderer.less
kravets-levko: Yeah, it's way much better:
image

ranbena:
@kravets-levko perhaps upgrade Ant so the new table header design kicks in?

ranbena:
> Hold up! This might have been fixed finally for Chrome 76 https://bugs.chromium.org/p/chromium/issues/detail?id=914844#c77

Just upgraded and can't repro bug 👍

ranbena:
> Just upgraded and can't repro bug 👍

Nope, I just did 👎

kravets-levko:
> perhaps upgrade Ant so the new table header design kicks in?

Already doing (need to resolve conflicts as well).

>> Just upgraded and can't repro bug :+1:

> Nope, I just did :-1:

You reproduced the bug in Chrome 76?

ranbena:
> You reproduced the bug in Chrome 76?

Yes but I just realized I got v76.0.3809 and the fix is on v76.0.3901..

kravets-levko:
@ranbena 
image

P.S. Ant's Table is a bit stupid: it has built-in sorting feature, which supports single-column sort. It's also possible to disable it (`sorter: true` and then set `sortOrder` per column where needed) - table will just show sorter icons and will not actually sort data. But it does not allow to display sort order on multiple columns - always only one column can have active (blue) arrow, others will be ignored :man_facepalming:

client/app/visualizations/table/renderer.less
```diff
@@ -42,11 +42,14 @@
         opacity: 0;
       }
 
-      &:hover .anticon.off {
+      &:hover .anticon.off,
+      .table-visualization-column-is-sorted .anticon.off {
```
ranbena: 👍

kravets-levko:
@ranbena 

> Regarding the cell height, I prefer the original 33px only for the sake of backwards compatibility, if you think it's important.

I'm not sure it's so important. Auto-height feature exists to deal with variable-height tables, and it may mess up dashboard layout any time (not only after CSS changes, but also after re-running query, changing visualization options, etc.). Considering this, I prefer to not override `line-height` for the visualization.

1. Totally agree. What could be a better value for this - how do you think? I think 200px should be more than enough, but who knows :thinking: 
2. I think it's better to create an issue (you can even assign it to me, or label as "Help wanted").

ranbena:
> 1. Totally agree. What could be a better value for this - how do you think? I think 200px should be more than enough, but who knows 🤔

200 is better than 100 in this case 👍 and having the setting later will make sure everyone's happy.

> 2. I think it's better to create an issue (you can even assign it to me, or label as "Help wanted").

Will do.

kravets-levko:
@arikfr @gabrieldutra If everyone is happy with this PR - I'll merge it

arikfr:
:+1:

Issue #3878, opened 2019-06-05
More visible indication that a widget *and dashboard* are refreshing


arikfr:
Currently when a widget is refreshing we show a small spinner in the bottom left corner of the widget, which is easy to miss.

Also, when refreshing a dashboard it's not clear when the whole dashboard is refreshed.

arikfr:
@ranbena as a follow up to yesterday's discussion, see my comment about the dashboard -- maybe we should have a dashboard level indication that a refresh is in progress? This way, when the user is refreshing the whole dashboard, they don't need to hunt for individual widget indications.

ranbena:
Yes, that makes sense 👍

ranbena:
I played around with a few versions (badges, refresh button text) but following our dashboard "Saving..." indication here's my suggestion:

<img src="https://user-images.githubusercontent.com/486954/58975772-ff494180-8782-11e9-9b28-cfa27e4c77d1.gif" width=400 />

This indication will appear in all cases in which widgets are refreshing:
1. Bulk refresh - manual and scheduled.
2. Widget refresh - manual and on parameter change.

The number indicator shows how many widgets are being refreshed at the moment, and the indicator disappears altogether when none remain.

arikfr:
I would change the counter to show the total too to make it easier to understand it's counting widgets (i.e. `Refreshing (2/7...`).

But are you sure it's not too simple visually? :) 
Btw, what about showing a [Notification](https://ant.design/components/notification/) or a [Message](https://ant.design/components/message/) with this content?

arikfr:
Also, having a dashboard level indicator doesn't remove the need to make the spinner a bit more noticeable.

ranbena:
> I would change the counter to show the total too to make it easier to understand it's counting widgets (i.e. `Refreshing (2/7...`).

I experimented with the following:

1. Feels too verbose
   <img width="356" alt="Screen Shot 2019-06-05 at 11 27 28" src="https://user-images.githubusercontent.com/486954/58976531-ee99cb00-8784-11e9-8045-3f3a853df662.png">

2. Minimal but might give the false notion that it's refreshing 7 widgets one by one 😓
   <img width="330" alt="Screen Shot 2019-06-05 at 11 26 54" src="https://user-images.githubusercontent.com/486954/58976532-ef326180-8784-11e9-97e3-c64cc525045a.png">
[...]

PR #3970, opened 2019-07-10
Added widget header refresh indicator

ranbena:
- [x] Feature

## Description
Testing indicator design.
Play with it [in the preview](https://deploy-preview-3970--redash-preview.netlify.com/dashboard/artists?p_AlbumId=4&p_test=MrCirca&p_text=Wor&p_number=4&p_date=2019-06-05&p_range.start=2019-06-07&p_range.end=2019-06-21&p_w191_AlbumId=4&p_w210_new=test).

ezgif com-video-to-gif

https://github.com/getredash/redash/issues/3878


arikfr:
It looks good, but why in the right top corner and not in the bottom left where we already have a spinner? Just replace the existing spinner with this one when the query executes?

ranbena:
> It looks good, but why in the right top corner and not in the bottom left where we already have a spinner? Just replace the existing spinner with this one when the query executes?

Our assumption is that it's not clear when a widget is refreshing. It's obvious that the text and spinner are inconspicuous, so that's what I went for with my first experimentations.
But I concluded it's still not clear enough since lower/tall widgets had their bottom part obscured, hiding the indication - so I added the indication up top.
Also, conceptually this type of information feels to me like the "front and center" kind.

I can make the bottom only version to get the feel and help decide.

arikfr:
> But I concluded it's still not clear enough since lower/tall widgets had their bottom part obscured, hiding the indication - so I added the indication up top.

Good point.

It just bothers me that we have multiple spinners and counters. Maybe we should move the counter next to the spinner at the top?

ranbena:
> It just bothers me that we have multiple spinners and counters.

Yeah, it's now 3 corners with refresh spinners :D

> Maybe we should move the counter next to the spinner at the top?

Depends what this info is for - to gain a rough estimate of how much longer it's gonna take? To indicate a stuck job?

arikfr:
> Depends what this info is for - to gain a rough estimate of how much longer it's gonna take? To indicate a stuck job?

Very deep questions for 11PM ... :-)

Another option is to drop the spinner next to the counter, but keep the counter where it currently is.

ranbena:
> Another option is to drop the spinner next to the counter, but keep the counter where it currently is.

Why not show the timer next to the spinner?
1. It's not valuable information unless job stuck.
2. It'll take up widget title real-estate.

Proposed UX:
1. Remove the bottom spinner and timer.
2. Show timer in tooltip when hovering the spinner.
3. When job stuck, change spinner color to red and link to troubleshooting doc from tooltip.

arikfr:
Ok, let's give a try to:

> Remove the bottom spinner and timer.
> Show timer in tooltip when hovering the spinner.

(we have no concept of "stuck", so we can skip this for now)

ranbena:
> Ok, let's give a try to:
> 
> > Remove the bottom spinner and timer.
> > Show timer in tooltip when hovering the spinner.
> 
> (we have no concept of "stuck", so we can skip this for now)

Actually, it's a bit more complex than that. The bottom refresh indicator has two more features that can't be easily ported to the top:
1. Time ago info - important to have at a glance. So can't be put it in a tooltip, nor persist in the widget header.
2. Acts as a refresh button - but top indicator was designed to be an indication only, not an interactive component (moves away and resizes on hover).

I might need to make some more changes.

arikfr:
> Time ago info - important to have at a glance. So can't be put it in a tooltip, nor persist in the widget header.

I thought of keeping only this, and hiding it when the query executes.

ranbena:
> I thought of keeping only this, and hiding it when the query executes.

It doesn't make sense to me. You press a button, it disappears, and an indication shows up at the opposite corner (which might be off-screen).

ranbena:
I'm trying a different approach - a candy stripe animation atop but my current implementation is cpu intensive and doesn't work well.

ezgif com-video-to-gif (1)

But lmk if you think this has potential.

ranbena:
Oh and also, if the run time is under an hour I display `mm:ss` instead of `HH:mm:ss`.

arikfr:
> It doesn't make sense to me. You press a button, it disappears, and an indication shows up at the opposite corner (which might be off-screen).

It shows when the query was executed, so in a way it makes sense that you won't see it when you're currently executing the query.

> I'm trying a different approach - a candy stripe animation atop but my current implementation is cpu intensive and doesn't work well.

I think it has potential. Mainly because it works well with the existing elements. The downside is that it feels a bit out of touch with the rest of the visual language of the application.
Maybe we can use an image animation to reduce the CPU overhead?

ranbena:
Reverted candy stripe and updated the following changes:

1. Bottom refresh button now shows "last updated" persistently and refresh runtime counter moved to top on hover. Also, removed `HH` when runtime is under 1 hour.
   
   

3. It was agreed that the bottom spinner adds too much visual noise and that one spinning icon would suffice. So I halted it.
But this resulted in  no clear indication that the refresh has started when clicking the two bottom widget refresh buttons. 
My solution is to have the clicked icon spin till the refresh cycle is done, or the other button gets clicked. All other scenarios they stay still.
I think it strikes the right balance between usability and visual noise.

   

4. Since the timer indication could potentially overlap a long query title, added a semi-transparent underlay.

   Screen Shot 2019-07-27 at 10 25 30

ranbena:
Also, added a simple dashboard "🔄Refresh" button indication.

arikfr:
> My solution is to have the clicked icon spin till the refresh cycle is done, or the other button gets clicked. All other scenarios they stay still.

The inconsistency feels confusing. Maybe they shouldn't animate at all? In most cases they will see the top indicator, so it's clear refresh started.

Also on Ubuntu/Chrome the refresh indicator looks unaligned:
image

ranbena:
> The inconsistency feels confusing. Maybe they shouldn't animate at all? In most cases they will see the top indicator, so it's clear refresh started.

If you played around with it and feel it's inconsistent and visually cluttered - I'm convinced.
But if not - plz do, the experience is the true test of this feature.

> Also on Ubuntu/Chrome the refresh indicator looks unaligned:

👍Thanks.

arikfr:
> If you played around with it and feel it's inconsistent and visually cluttered - I'm convinced.

I did play with it, but don't forget I'm biased: I already know what should happen...

ranbena:
> I did play with it, but don't forget I'm biased: I already know what should happen...

Good enough for me.
I revoked the button click spin in ab5494a. But keeping it on the dashboard refresh button.

@arikfr lmk if this feature is complete as far as you're concerned and I'll release it from draft.

client/app/components/dashboards/widget.html
```diff
@@ -29,6 +29,12 @@
             <li ng-if="$ctrl.dashboard.canEdit()"><a ng-click="$ctrl.deleteWidget()">Remove from Dashboard</a></li>
           </ul>
         </div>
+        <div id="topRefreshIndicator" ng-if="$ctrl.widget.loading">
```
arikfr: Isn't `id` supposed to be singular on a page?

ranbena: Only when retrieving elements by id but in terms of css hooks, it has no significance.

I'll change it though 👍🏻

client/app/components/Timer.jsx
```diff
@@ -14,7 +14,10 @@ export function Timer({ from }) {
     return () => clearInterval(timer);
   }, []);
 
-  return moment.utc(moment.now() - startTime).format('HH:mm:ss');
+  const diff = moment.now() - startTime;
+  const format = diff > 1000 * 60 * 60 ? 'HH:mm:ss' : 'mm:ss'; // no HH under an hour
```
kravets-levko: :+1:

client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -480,6 +490,86 @@ body {
   background-color: fade(@redash-gray, 12%) !important;
 }
 
+#topRefreshIndicator {
```
kravets-levko: Kebab-case please :pray: Also, don't use `id` for elements that may occur multiple times on page - they're not intended for this. Also, ids are very hard to override in CSS because they have highest priority and support the only value in attribute.

ranbena: 40ec0c0

client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -480,6 +490,86 @@ body {
   background-color: fade(@redash-gray, 12%) !important;
 }
 
+.refresh-indicator {
```
arikfr: Something that I missed when doing the review: we need to stop adding styles to `redash-newstyle.less`. The refresh indicator should've been a component and its style should've been next to it.

Not worth revisiting at the moment, just wanted to point out so we keep this in mind for future stuff.

ranbena: Had to be done! https://github.com/getredash/redash/pull/4017

arikfr:
@ranbena ,

> I did play with it, but don't forget I'm biased: I already know what should happen...

Now that I had real life experience with it, I can say that I was wrong. When you click on a widget to refresh, it's in the more subtle mode of the refresh indicator. Therefore it's easy to miss, and because the button doesn't animate it feels like nothing happens.

I guess that we should bring back the animation on the button you click you had there...

ranbena:
I had an alternative I tried - making the buttons have a hover+press+pressed state.
Still, it's enough without the icon spinning.

ranbena:
Reverted in https://github.com/getredash/redash/pull/4027

PR #3971, opened 2019-07-11
Added apply on keyboard shortcut to all parameters

ranbena:
- [x] Feature

## Description
Added meta+enter (and alt+enter) key bindings to `parameters`.
Activated when focusing on input elements when there are changes to be applied.

#### How
Put focus/blur listeners on the `<parameters>` element which adds/removes the key bindings accordingly.

#### Caveat
Since the query page already has these key bindings, it would result in double execution ☠
I tried a few approaches but eventually I went for disabling it for the query page with an attribute.
I don't like this solution cause it's coupled (the tooltip promises an ability that's fulfilled by another component) but it's the simplest 😣

### DIFF
--- a/client/app/components/ParameterApplyButton.jsx
+++ b/client/app/components/ParameterApplyButton.jsx
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
 import Button from 'antd/lib/button';
 import Badge from 'antd/lib/badge';
+import Tooltip from 'antd/lib/tooltip';
+import { KeyboardShortcuts } from '@/services/keyboard-shortcuts';
 
 function ParameterApplyButton({ paramCount, onClick, isApplying }) {
   // show spinner when applying (also when count is empty so the fade out is consistent)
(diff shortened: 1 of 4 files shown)


ranbena:
@arikfr can you test that this works as expected?

ranbena:
> It seems that netlify-preview is not available for non-`master` target.

Yeah it's a shame and there's no setting or info on the web about it.

> Locally I noticed that the Apply Changes button click was not triggering update... And when I went to see Cypress had already [pointed that out](https://dashboard.cypress.io/#/projects/924cka/runs/753/specs) 😅.

Darn. Fixing!

> I wonder if considering a React `Parameters` component this would be as simple as monitoring the key press and trigger update function with `event.stopPropagation()`.

Sure. I tried to make it happen with mousetrap but just gave up.

gabrieldutra:
If this becomes too complicated you can consider postpone to the `<Parameters>` migration, I believe it's not a critical thing that can't wait a bit :slightly_smiling_face:.

gabrieldutra:
@ranbena should this be merged so Netlify preview becomes available?

PR #3973, opened 2019-07-11
Use AceEditor for Query Snippets

gabrieldutra:

- [x] Other

## Description
Following my idea on #3962 this makes `AceEditor` available on `DynamicForm` and applies it to Query Snippets.

I didn't find a simple way to align the validation mark (but didn't spend much time on it), for Query Snippets there's no need anyway.

Notes:
- It was necessary to create an almost useless Class component `AceEditorInput` only for the sake of avoiding [this](https://github.com/getredash/redash/pull/3962#issuecomment-509316423) issue.
- Not sure why, but `className` didn't work with `<AceEditor>`

--

query_snippets_ace


kravets-levko:
> It was necessary to create an almost useless Class component AceEditorInput only for the sake of avoiding this issue.

@gabrieldutra Can you please explain me what's the issue there? It looks really interesting that just wrapping a component fixes it :thinking:

gabrieldutra:
> @gabrieldutra Can you please explain me what's the issue there? It looks really interesting that just wrapping a component fixes it

Sure, my assumption of the issue was that the AceEditor was changing the value internally, thus triggering Antd Form errors. By wrapping it in a Class Component I force the value to be the provided one, making the only changeable through the upper component, in this case Antd Form :slightly_smiling_face:.

"I force the value to be the provided one" -> this seemed safe since AceEditor supports it and the effect is basically the same as having the AceEditor on any Class Component with its value on a state.

ranbena:
Any way to hide the selected line and ghost cursor when editor field isn't focused?
It's really confusing cause it looks like it's already focused :/

ranbena:
@gabrieldutra I can help with the validation mark alignment. What should I do to reproduce it?

kravets-levko:
> Sure, my assumption of the issue was that the AceEditor was changing the value internally, thus triggering Antd Form errors. By wrapping it in a Class Component I force the value to be the provided one, making it only changeable through the upper component, in this case Antd Form 

I don't understand. Your wrapper passes all props to editor, and does not handle any changes, so if editor calls any callbacks/triggers events - your wrapper just passes them up. If editor modifies value in place - then wrapper also does not do anything: non-objects are always copied (therefore cannot be modified in place), and objects are passed by reference - wrapper does not create a copy, so if editor modifies an object - it should work exactly the same with and without a wrapper.

gabrieldutra:
@ranbena I've pushed the furthest I got :slightly_smiling_face: and enabled the feedback icons for testing purposes, I sort of gave up when I realized it was still not good for Scrollbars. Mind that disabling it for the Ace Editor is also an option (we do that for checkbox and file inputs).

@kravets-levko to be honest I don't 100% understand why this happened and why this works, I actually saw the errors and figured that wrapping it would somehow "isolate" the Editor in a situation where the component `value` wouldn't get updated, as you mentioned.

It's probably something to do with the way `ReactAce` was implemented that conflicts with `AntdForm`. From a quick look on their code they use a `ref` for the editor and set/get value from it. I guess with a `ref` it's possible to change the component's `value`.

kravets-levko:
> From a quick look on their code they use a ref for the editor and set/get value from it. I guess with a ref it's possible to change the component's value.

Yeah, that's possible. Probably AntdForm takes a ref to each element and expects to see some Ant control there, but to ReactAce it gets something else but tries to use it. Wrapping it with a component isolated ReactAce's ref from Ant. It's interesting why functional component does not work in this case, but I think there may be some difference in how React passes ref for class-based and functional components. Can you do an experiment to confirm this theory? Rewrite wrapper component in this way:

```javascript
import React, { useImperativeHandle } from 'react';

function AceEditorInput(props, ref) {
  useImperativeHandle(ref, () => null); // expose `null` as this component's ref
  return (
    // all the same as now
  );
}
```

gabrieldutra:
The issue of this component as a Functional Component is this (Ant Control needs the component Ref):
functional-ref

The `useImperativeHandle` actually breaks it :sweat_smile:. Surprisingly, 68ae87d5015cb89d8484537605302ecc471a86ff works.

client/app/components/AceEditorInput.jsx
```diff
@@ -0,0 +1,21 @@
...
+function AceEditorInput(props, ref) {
+  return (
+    <div className="ace-editor-input">
+      <AceEditor
+        ref={ref}
+        mode="sql"
+        theme="textmate"
+        height="100px"
+        editorProps={{ $blockScrolling: Infinity }}
+        {...props}
```
ranbena:
```suggestion
        showPrintMargin={false}
        {...props}
```

So this unnecessary line doesn't show.
Screen Shot 2019-07-12 at 10 13 09

client/app/components/query-snippets/QuerySnippetDialog.jsx
```diff
@@ -79,6 +78,7 @@ class QuerySnippetDialog extends React.Component {
           fields={formFields}
           onSubmit={this.handleSubmit}
           hideSubmitButton
+          feedbackIcons
```
ranbena: An alternative that I think makes better sense https://github.com/getredash/redash/pull/3974

gabrieldutra: Should we keep the icons for the Query Snippets context or...
```suggestion
```

ranbena: It's a really nice and fun UX, but not a must. I have no strong opinion.

PR #3974, opened 2019-07-12
Feedback icon adjustment for tall components

ranbena:
- [x] Refactor
- [x] Bug Fix

## Description
A generic implementation for feedback icon positioning.
Reverted the original fix.

Screen Shot 2019-07-12 at 10 06 42


gabrieldutra:
The Hide cursor thing was in that commit :sweat_smile:. I'll put it back after merge, one last thing, the Scrollbar issue (images below) seems to happen due to a `z-index: 6` that ace has, I can replace that to default, wdyt? (It is like this for TextAreas already)

Before
scrollbar-feedback-icon

After
scrollbar-after

ranbena:
> The Hide cursor thing was in that commit 😅. I'll put it back after merge

Oh man. So glad there's someone sharper than me in the team 🤦‍♂️

> one last thing, the Scrollbar issue (images below) seems to happen due to a `z-index: 6` that ace has for scrollbars, I can replace that to default, wdyt? (It is like this for TextAreas already)

Hmmm the result does not look appealing. I tried a few alternatives but nothing wins over it.
So go for it 👍

Issue #3975, opened 2019-07-12
Legend overlaps plot on small screens


justinclift:
## Issue Summary

When a visualisation is displayed in a very small screen size, then clicked upon, the legend moves to overlap the main chart content.

### Chrome 75 (Win 7)
![redash-legend_overlaps_content-chrome 1](https://user-images.githubusercontent.com/406299/61122604-940f2f80-a4e5-11e9-9fa6-519cb63334fa.png)
### Firefox 68 (Linux)
![redash-legend_overlaps_content-firefox 1](https://user-images.githubusercontent.com/406299/61122605-940f2f80-a4e5-11e9-80d1-e0dc2c1dcfd4.png)

Url to try with:  https://app.redash.io/default/embed/query/226021/visualization/384887?api_key=BHrjoUJrFxMCyfRe6M6dZXA1hO5qy7wSeQEVadjo&p_Date=2012-12-14

### Steps to Reproduce

1. Go to https://app.redash.io/default/embed/query/226021/visualization/384887?api_key=BHrjoUJrFxMCyfRe6M6dZXA1hO5qy7wSeQEVadjo&p_Date=2012-12-14
2. Resize the browser window to be extremely small (as per screenshots)
3. Click on the chart.  The legend now moves position, covering the chart.

Covering the chart like this should not happen.

### Technical details:

* Redash Version: Our hosted SaaS version
* Browser/OS: Firefox 68 on Linux, and Chrome 75 on Win 7
* How did you install Redash: n/a

justinclift:
As a data point, this may be related to https://github.com/getredash/redash/pull/3882

kravets-levko:
@justinclift Thanks a lot! Yes, it's the same bug, but it appears under other conditions. I'll try to fix it ASAP.

PR #3976, opened 2019-07-12
Plotly legent overlaps plot on small screens (when legend clicked)

kravets-levko:

- [x] Bug Fix

## Description

Revert changes from getredash/redash#3882 and apply a new fix. Previous time I misunderstood the problem and applied a wrong fix, but the real problem was with the code that detects CSS support.

Fixes getredash/redash#3975

### DIFF
--- a/client/app/visualizations/chart/plotly/utils.js
+++ b/client/app/visualizations/chart/plotly/utils.js
@@ -747,6 +747,14 @@ export function updateLayout(plotlyElement, layout, updatePlot) {
   layout.width = Math.floor(plotlyElement.offsetWidth);
   layout.height = Math.floor(plotlyElement.offsetHeight);
 
+  const transformName = find([
+    'transform',
+    'WebkitTransform',
+    'MozTransform',
+    'MsTransform',
+    'OTransform',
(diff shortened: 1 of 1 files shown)


justinclift:
Awesome @kravets-levko.  I'll this out in a test instance later today (~10 hrs away), to confirm it solves the problem. :smile:

arikfr:
Justin, are you familiar with the preview deployments?

justinclift:
This change fixes the problem when testing just now in Firefox.

justinclift:
> Justin, are you familiar with the preview deployments?

Not yet.  Sounds like there's a better way then running things up in a local VM?

justinclift:
> Justin, are you familiar with the preview deployments?

Ahhh.  Thus the `deploy/netlify - Deploy preview ready` thing in the list above.  That'll save time in future. :smile:

arikfr:
@justinclift 

> Ahhh. Thus the deploy/netlify - Deploy preview ready thing in the list below. That'll save time in future. 😄

Exactly. For every PR, Netlify creates a deployment. This allows testing only the frontend (they all use the same backend), but in many cases it's enough.

PR #3983, opened 2019-07-15
Show error in case of failing to load a shared dashboard

arikfr:

- [x] Other

image

### DIFF
--- a/client/app/pages/dashboards/public-dashboard-page.js
+++ b/client/app/pages/dashboards/public-dashboard-page.js
@@ -1,3 +1,4 @@
+import PromiseRejectionError from '@/lib/promise-rejection-error';
 import logoUrl from '@/assets/images/redash_icon_small.png';
 import template from './public-dashboard-page.html';
 import dashboardGridOptions from '@/config/dashboard-grid-options';
@@ -52,6 +53,8 @@ const PublicDashboardPage = {
         };
 
         this.extractGlobalParameters();
+      }).catch((error) => {
+        throw new PromiseRejectionError(error);
       });
 
       if (refreshRate) {


client/app/pages/dashboards/public-dashboard-page.js
```diff
@@ -52,6 +53,8 @@ const PublicDashboardPage = {
         };
 
         this.extractGlobalParameters();
+      }).catch((error) => {
+        throw new PromiseRejectionError(error);
```
ranbena: Do we have deliberate relevant exceptions where we could insert meaningful error messages to?

arikfr: `PromiseRejectionError` is actually such an exception: https://github.com/getredash/redash/blob/master/client/app/lib/promise-rejection-error.js

ranbena: I mean the `error` we catch - make it's message be meaningful.
"...page cannot be found" is just confusing in this case.

arikfr: What would you expect here?

ranbena: Oh I see there's a relevant error message a few lines above:
https://github.com/getredash/redash/blob/2ca763747a787d8a9af90189953440745fcfa238/client/app/pages/dashboards/public-dashboard-page.js#L45

 BUT it seems to be doing it wrong, as PromiseRejectionError requires
1. `rejection.status !== undefined/404/403`
2. `rejection.data.error = 'the message'`

arikfr: But this is not what this change about. This change about showing a 404 when you use a bad token (either non existing or just invalid).

ranbena: Ok 😭

PR #3964, opened 2019-07-08
[Data Sources] Initial commit for adding Dgraph support

The-Alchemist:
## What type of PR is this? (check all applicable)

- [x] Feature
- [x] New Query Runner (Data Source)

## Description
Adding support for querying [Dgraph](https://dgraph.io).

Feedback welcome!

PR #3987, opened 2019-07-15
[Data Sources] Initial commit for adding Dgraph support

The-Alchemist:

- [x] Feature

## Description
Adding support for querying [Dgraph](https://dgraph.io).

Original PR was #3964


redash/query_runner/dgraph.py
```diff
@@ -0,0 +1,109 @@
...
+            data = json.loads(response_raw.json)
+
+            first_key = next(iter(data.keys()))
+            first_node = data[first_key]
+
+            # grab all the column names
+            column_names = set().union(*first_node)
+            # create a dict for column names in the format that the API wants
+            # the type as 'string' is just a hack for now
+            columns = [{'name': c, 'friendly_name': c, 'type': 'string'} for c in column_names]
```
arikfr: All the values are untyped?

tsl-karlp: > How do queries look like in Dgraph? I mean, what type of query language does it use?

Dgraph uses a variant of GraphQL they call GraphQL+- (https://docs.dgraph.io/query-language/).  The output is JSON.

There might be a syntax highlighter for GraphQL+-.    Should that be part of this PR, or a separate one?


> All the values are untyped?

Unfortunately, there's no type metadata attached to the results, like in most SQL databases.  The only way is check the JSON itself (`typeof`).

tsl-karlp: > All the values are untyped?


Unfortunately, there's no type metadata attached to the results, like in most SQL databases.  The only way is check the JSON itself (`typeof`).

The-Alchemist:
> How do queries look like in Dgraph? I mean, what type of query language does it use?

Dgraph uses a variant of GraphQL they call GraphQL+- (https://docs.dgraph.io/query-language/). The output is JSON.

There might be a syntax highlighter for GraphQL+-. Should that be part of this PR, or a separate one?

redash/query_runner/dgraph.py
The-Alchemist: > All the values are untyped?

Basically.  

Unfortunately, there's no type metadata attached to the results, like in most SQL databases. The only way is check the JSON itself (`typeof`).

redash/query_runner/dgraph.py
```diff
@@ -0,0 +1,160 @@
...
+        # Change the encoding type if needed
+        return s.encode('utf-8')
+
+
+def reduce_item(reduced_item, key, value):
+    """From https://github.com/vinay20045/json-to-csv"""
+    # Reduction Condition 1
+    if type(value) is list:
+        i = 0
+        for sub_item in value:
```
arikfr: You can use `enumerate` here to avoid the need to manually keep track of `i`.

```suggestion
        for i, sub_item in enumerate(value):
```

redash/query_runner/dgraph.py
```diff
@@ -0,0 +1,160 @@
...
+            reduce_item(reduced_item, key+'/'+to_string(i), sub_item)
+            i = i+1
+
+    # Reduction Condition 2
+    elif type(value) is dict:
+        sub_keys = value.keys()
+        for sub_key in sub_keys:
+            reduce_item(reduced_item, key+'/'+to_string(sub_key), value[sub_key])
```
arikfr: To be consistent with other query runners let's use `.` as delimiter here. Also, using unicode object and the `format` function will simplify the code here:

```suggestion
            reduce_item(reduced_item, u'{}.{}'.format(key, sub_key), value[sub_key])
```

redash/query_runner/dgraph.py
```diff
@@ -0,0 +1,160 @@
...
+        return s.encode('utf-8')
+
+
+def reduce_item(reduced_item, key, value):
+    """From https://github.com/vinay20045/json-to-csv"""
+    # Reduction Condition 1
+    if type(value) is list:
+        i = 0
+        for sub_item in value:
+            reduce_item(reduced_item, key+'/'+to_string(i), sub_item)
```
arikfr: As we know what type of `i` is, we can simplify it by:

```suggestion
            reduce_item(reduced_item, u'{}.{}'.format(key, i), sub_item)
```

redash/query_runner/dgraph.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # Reduction Condition 2
+    elif type(value) is dict:
+        sub_keys = value.keys()
+        for sub_key in sub_keys:
+            reduce_item(reduced_item, key+'/'+to_string(sub_key), value[sub_key])
+
+    # Base Condition
+    else:
+        reduced_item[to_string(key)] = to_string(value)
```
arikfr: I would just go with `reduced_item[key] = value` unless `key` might be non string, then I would consider casting it. 

Why convert all the values to strings?

redash/query_runner/dgraph.py
The-Alchemist: done

redash/query_runner/dgraph.py
The-Alchemist: probably just a copy/paste thing.  will test and verify it still works

arikfr:
Merged. Thanks!

PR #3988, opened 2019-07-16
A couple of parameters-on-public-dashboards loose ends

rauchy:
- [x] Bug Fix

## Description
* Replace "unsafe widget" error messages only for those that have a visualization
* Rethrow the error

Follow up for #3659

### DIFF
--- a/client/app/pages/dashboards/public-dashboard-page.js
+++ b/client/app/pages/dashboards/public-dashboard-page.js
@@ -39,17 +39,14 @@ const PublicDashboardPage = {
         this.dashboard = new Dashboard(data);
         this.dashboard.widgets = Dashboard.prepareDashboardWidgets(this.dashboard.widgets);
         this.dashboard.widgets.forEach((widget) => {
-          const promise = widget.load(!!refreshRate);
-          if (promise) {
-            promise.catch((error) => {
-              const isSafe = widget.getQuery() ? widget.getQuery().is_safe : true;
-              if (!isSafe) {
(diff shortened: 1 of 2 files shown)


ranbena:
@rauchy I think a better approach is to make sure `load()` consistently returns a promise.

We can return an already resolved promise:
```js
load(force, maxAge) {
  if (!this.visualization) {
     return Promise.resolve();
  }
```

rauchy:
> I think a better approach is to make sure load() consistently returns a promise.

Totally a better approach. Was just afraid of breaking other usages, but everything seems to work.

client/app/pages/dashboards/public-dashboard-page.js
```diff
@@ -44,6 +44,8 @@ const PublicDashboardPage = {
             if (!isSafe) {
               error.errorMessage = 'This query contains potentially unsafe parameters and cannot be executed on a publicly shared dashboard.';
             }
+
+            throw error;
```
ranbena: Makes sense but does it have any affect on the flow?

rauchy: It doesn't right now, but as a good practice it's best to re-throw. (You never know if someone would want to handle this in the future)

arikfr: This won't blow up the dashboard load?

ranbena: Nope - gets caught by original promise.
https://github.com/getredash/redash/blob/c3cc65a21d74408da490ee076ae34bfd0391f5f6/client/app/services/widget.js#L139-L142

PR #3985, opened 2019-07-16
Query Result API response shouldn't include query information for non authenticated users

rauchy:
Query Results API response should include the bare minimum fields if this is an API call (dashboard, query) and not an authenticated user.

We need to move the serialization logic into `redash.serializers` and apply the relevant checks.

### DIFF
--- a/client/app/pages/dashboards/public-dashboard-page.js
+++ b/client/app/pages/dashboards/public-dashboard-page.js
@@ -39,17 +39,14 @@ const PublicDashboardPage = {
         this.dashboard = new Dashboard(data);
         this.dashboard.widgets = Dashboard.prepareDashboardWidgets(this.dashboard.widgets);
         this.dashboard.widgets.forEach((widget) => {
-          const promise = widget.load(!!refreshRate);
-          if (promise) {
-            promise.catch((error) => {
-              const isSafe = widget.getQuery() ? widget.getQuery().is_safe : true;
-              if (!isSafe) {
-                error.errorMessage = 'This query contains potentially unsafe parameters and cannot be executed on a publicly shared dashboard.';
-              }
+          widget.load(!!refreshRate).catch((error) => {
+            const isSafe = widget.getQuery() ? widget.getQuery().is_safe : true;
+            if (!isSafe) {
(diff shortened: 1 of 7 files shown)


redash/serializers/query_result.py
```diff
@@ -57,6 +57,14 @@ def _get_column_lists(columns):
     return fieldnames, special_columns
 
 
+def serialize_query_result(query_result, is_api_user):
+    if is_api_user:
```
arikfr: It's safe to use `current_user` in the serialization code instead of passing `is_api_user`. It's not a big deal for this PR, but #3782 will extend this method to apply other checks anyway.

client/app/pages/dashboards/public-dashboard-page.js
```diff
@@ -44,6 +44,8 @@ const PublicDashboardPage = {
             if (!isSafe) {
               error.errorMessage = 'This query contains potentially unsafe parameters and cannot be executed on a publicly shared dashboard.';
             }
+
+            throw error;
```
arikfr: Will this make the whole dashboard load blow up?

redash/serializers/query_result.py
```diff
@@ -57,6 +57,14 @@ def _get_column_lists(columns):
     return fieldnames, special_columns
 
 
+def serialize_query_result(query_result, is_api_user):
+    if is_api_user:
+        publicly_needed_keys = ['data', 'retrieved_at']
+        return {key: query_result.to_dict()[key] for key in publicly_needed_keys}
```
arikfr: Current implementation will call `to_dict` multiple times, which will result in calling `json_loads` multiple times.

You should make sure to call `to_dict` only once, and you can also use [`project`](https://funcy.readthedocs.io/en/stable/colls.html?highlight=project#project) instead of "manual" extraction.

client/app/pages/dashboards/public-dashboard-page.js
rauchy: Nope, and this is not really a part of this PR :)

Anyway, this is removed in #3990

redash/serializers/query_result.py
rauchy: You've shown me `project` before, and I was looking for it with the wrong search term (looked for dict masking in Python, but all examples were soooo complicated). Thanks!

PR #3990, opened 2019-07-16
Return unsafe sharing error from backend

rauchy:
- [x] Refactor

## Description
Create a message explaining that unsafe queries can't be shared in the backend, and use that message in public dashboards and embeds.

Replaces the behavior introduced in #3659 and #3988 (/cc @ranbena)

### DIFF
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -20,7 +20,7 @@ def error_response(message, http_status=400):
 
 
 error_messages = {
-    'unsafe_when_shared': error_response('This query contains potentially unsafe parameters and cannot be executed on a shared dashboard or visualization embed.', 403),
+    'unsafe_when_shared': error_response('This query contains potentially unsafe parameters and cannot be executed on a shared dashboard or an embedded visualization.', 403),
     'unsafe_on_view_only': error_response('This query contains potentially unsafe parameters and cannot be executed on this data source.', 403),
     'no_permission': error_response('You do not have permission to run queries with this data source.', 403),
     'select_data_source': error_response('Please select data source to run this query.', 401)
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -15,8 +15,16 @@ from redash.models.parameterized_query import ParameterizedQuery, InvalidParamet
 from redash.serializers import serialize_query_result_to_csv, serialize_query_result_to_xlsx
 
 
-def error_response(message):
-    return {'job': {'status': 4, 'error': message}}, 400
+def error_response(message, http_status=400):
+    return {'job': {'status': 4, 'error': message}}, http_status
+
+
+error_messages = {
+    'unsafe_when_shared': error_response('This query contains potentially unsafe parameters and cannot be executed on a shared dashboard or visualization embed.', 403),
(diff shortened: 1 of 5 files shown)


redash/handlers/query_results.py
```diff
@@ -182,7 +182,12 @@ def post(self, query_id):
         if has_access(query, self.current_user, allow_executing_with_view_only_permissions):
             return run_query(query.parameterized, parameter_values, query.data_source, query_id, max_age)
         else:
-            return {'job': {'status': 4, 'error': 'You do not have permission to run queries with this data source.'}}, 403
+            if current_user.is_api_user() and not query.parameterized.is_safe:
```
arikfr: Tests?

redash/handlers/query_results.py
```diff
@@ -182,7 +182,12 @@ def post(self, query_id):
...
+                message = 'This query contains potentially unsafe parameters and cannot be executed on a publicly shared resource.'
```
arikfr:
```suggestion
                message = 'This query contains potentially unsafe parameters and cannot be executed on a shared dashboard or visualization embed.'
```

rauchy: I deliberately used "resource" to avoid the word "dashboard" appearing on embeds and vice versa, but 🤷🏻‍♂️

arikfr: I know, but "publicly shared resource" sounds like something taken from a programming manual and not very friendly. :)

redash/handlers/query_results.py
rauchy: 👍done in 474dae1

redash/handlers/query_results.py
rauchy: image

redash/handlers/query_results.py
```diff
@@ -15,8 +15,16 @@
...
 
 
-def error_response(message):
-    return {'job': {'status': 4, 'error': message}}, 400
+def error_response(message, http_status=400):
+    return {'job': {'status': 4, 'error': message}}, http_status
+
+
+error_messages = {
+    'unsafe_when_shared': error_response('This query contains potentially unsafe parameters and cannot be executed on a shared dashboard or visualization embed.', 403),
```
ranbena:
```suggestion
    'unsafe_when_shared': error_response('This query contains potentially unsafe parameters and cannot be executed on a shared dashboard or an embedded visualization.', 403),
```

redash/handlers/query_results.py
```diff
@@ -15,8 +15,16 @@
...
 
-def error_response(message):
-    return {'job': {'status': 4, 'error': message}}, 400
+def error_response(message, http_status=400):
+    return {'job': {'status': 4, 'error': message}}, http_status
+
+
+error_messages = {
+    'unsafe_when_shared': error_response('This query contains potentially unsafe parameters and cannot be executed on a shared dashboard or an embedded visualization.', 403),
+    'unsafe_on_view_only': error_response('This query contains potentially unsafe parameters and cannot be executed on this data source.', 403),
```
arikfr:
```suggestion
    'unsafe_on_view_only': error_response('This query contains potentially unsafe parameters and cannot be executed with read-only access to this data source.', 403),
```

Maybe something like this?

PR #3993, opened 2019-07-17
Move query runners/destinations import from redash.app to redash.

arikfr:

- [x] Refactor

## Description

While it was nice to delay the loading of query runners to the Flask app creation, it was creating issues on systems with less resources. Specifically when Celery was creating multiple workers at the same time, it was stressing the system.

Considering that due to how Flask works you need the app object anyway for everything, we were not gaining much by this delay.

I moved the alert destinations import to the same location for consistency.


arikfr:
> What exactly are we losing by moving them back?

:man_shrugging: not a lot I guess, because we create the app instance anyway. I assume that in some cases it saves us from loading all the query runners, but those are rare.

PR #3995, opened 2019-07-18
Avoid committing `it.only`

rauchy:
- [x] Refactor

## Description
Committing `it.only` really shouldn't be a thing :)


client/.eslintrc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     "no-control-regex": "off",
     "no-multiple-empty-lines": "warn",
     "no-script-url": "off", // some <a> tags should have href="javascript:void(0)"
+    "no-only-tests/no-only-tests": process.env.NODE_ENV === "production" ? "off" : "error",
```
ranbena: Why 'off' in prod?

rauchy: I was under the impression that webpack would run lint on these files as well. Fixed in 247a1d4

ranbena: I don't know what the implication is 🤷‍♂️

Issue #3217, opened 2018-12-26
Unpublished queries show up in search results for other users


arikfr:
Looks like that in one of the changes to the queries API, we started showing unpublished queries to all the users. This is not the intended behavior, although I'm no longer sure if we should revert back.

We have the option to allow editing queries/dashboards by other users in a feature flag, that soon will become enabled for all. If we keep the old logic (hide unpublished queries from non owners) it will make it really hard to load the queries a user show see if we also need to filter unpublished queries.

Would love to hear opinions.

kravets-levko:
I think with new permissions settings for queries/dashboards neither approach is correct. We should show items that user has access to - own items, published items or items with granted access. It will be the most correct behavior from the user's point of view.

kravets-levko:
Okay, got it :thinking:

susodapop:
I think this change should be reverted to make application behavior consistent. It seems to me that the whole justification for a "Published Queries" mechanic was privacy. Don't publish a query until you're ready for others to see it. This is how Dashboard publishing still works. But with this behavior removed, why would a user choose to publish a query? Or phrased another way, why **wouldn't** a user want to publish a query? If it's to allow the query on dashboards, then publishing seems like an irrelevant extra step. If it's to allow embedding the visualization outside of Redash then "publish" seems like the wrong verb. It could probably just be a checkbox instead.

Perhaps a compromise on the first question above would be to change the UI for Publishing. Suppose a user wants to add a visualization from an unpublished query to their dashboard. The query would appear in the "Add Widget" modal search on the dashboard screen. And when they select the query it would pop a message box explaining "This query hasn't been published. Would you like to publish it now?". If they choose later to unpublish, they can visit the query screen and toggle the button.

justinclift:
> Don't publish a query until you're ready for others to see it.

On that note, noticed something interesting while attempting to replicate the documented behaviour today.

With a new Query, if you change the title from the default text string (at least in v7), the query automatically changes to Published.  It's possible to Unpublish it afterwards, but it did catch me out initially.

susodapop:
> On that note, noticed something interesting while attempting to replicate the documented behaviour today.

What you describe is the documented behavior. 

> To publish a query, give it a name or click the `Publish` button.

justinclift:
Oh.  I guess that just seems weird to me.

Literally the first thing I do with any new query is name it something appropriate, to match it's purpose.

[...]

PR #3998, opened 2019-07-19
Added publish notification to query rename

ranbena:
- [x] Feature

## Description
Following up on https://github.com/getredash/redash/issues/3217#issuecomment-512783137, added "Query published" notification when query rename auto-publishes.

### DIFF
--- a/client/app/pages/queries/view.js
+++ b/client/app/pages/queries/view.js
@@ -303,12 +303,16 @@ function QueryViewCtrl(
   $scope.saveName = (name) => {
     $scope.query.name = name;
     Events.record('edit_name', 'query', $scope.query.id);
+
+    let customOptions;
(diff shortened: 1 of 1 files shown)


client/app/pages/queries/view.js
```diff
@@ -305,6 +305,7 @@ function QueryViewCtrl(
     Events.record('edit_name', 'query', $scope.query.id);
     if ($scope.query.is_draft && clientConfig.autoPublishNamedQueries && $scope.query.name !== 'New Query') {
       $scope.query.is_draft = false;
+      notification.success('Query published');
```
arikfr: This will result in multiple notifications + might show a success notification while the query save actually fails. 

How about we update the save notification to include a message about publishing the query? We can do a check in the `saveQuery` function to see if the publish state changed and update the message accordingly.

ranbena: Done 997182b 👍 

Screen Shot 2019-07-20 at 22 23 13

PR #4000, opened 2019-07-21
Fixed last test

ranbena:
- [x] Bug Fix

## Description
Added height test to 2nd widget in test.

PR #4002, opened 2019-07-22
Percy: Fix shared dashboard inconsistent snapshots

gabrieldutra:

- [x] Other

## Description
Some updates on the code to use the existent `createQueryAndAddWidget` as it easily provides the widget element, therefore it allows us to assert the widget height before it be defined.

**Caveat**: [this](https://github.com/getredash/redash/pull/4000#issuecomment-513555255)

--

--


rauchy:
@gabrieldutra Looks good. I'm just wondering why you chose to remove `addWidgetAndShareDashboard` and repeat the code in all 3 tests?

gabrieldutra:
> @gabrieldutra Looks good. I'm just wondering why you chose to remove `addWidgetAndShareDashboard` and repeat the code in all 3 tests?

I initially used that method, but the widgets have different expected `height`s, I didn't like that much the idea of passing that as a parameter (both the `addWidget` and the `shareDashboard` parts were requiring specific code), so I went only with `createQueryAndAddWidget` to see the result.

From current implementation I think an option is to add a `shareDashboard` to improve readability.

PR #3972, opened 2019-07-22
Add "deprecated" flag to query runners (and alert destinations)

rauchy:
Deprecated Query Runners/Alert Destinations will not show up in the list of available types when creating a new one, but will remain there for those data sources/destinations that use them already.

This is useful to deprecate the URL data source and HipChat alert destinations.


redash/destinations/hipchat.py
```diff
@@ -13,6 +13,7 @@
 }
 
 
+@deprecated()
```
rauchy: @arikfr not sure you actually wanted to deprecate this (and the URL data source) as part of this PR, so let me know if you want me to revert this commit.

redash/destinations/__init__.py
```diff
@@ -34,6 +34,10 @@ def enabled(cls):
     def configuration_schema(cls):
         return {}
 
+    @classmethod
```
arikfr: I made a mistake when designing these classes by using class methods for these attributes. How about we use a class property:

```python
class BaseQueryRunner(object):
    deprecated = False
```

and then in the deprecated classes:

```python
class Url(BaseQueryRunner):
    deprecated = True
```

rauchy: I feel the same. I went for a decorator in order to remain compatible with the existing `classmethod` style and still allow a succinct way to declare deprecations at the top of the object being deprecated. 

I can drop the decorator and stick with class properties, however we can choose to keep using decorators as they will always stay on top and feel "more meta". WDYT?

redash/destinations/hipchat.py
arikfr: No, it's cool. We do need to update the docs about the URL data source though.

redash/destinations/__init__.py
arikfr: image

How about we move to class properties, but also keep the decorator to make sure it's visible that this object is deprecated?

rauchy: yeah that's what I meant :) 5a9e13c

redash/destinations/hipchat.py
rauchy: The page currently mixes `JSON` and `URL` and feels like it might need a slight rewrite (or some good chunks related to `URL` removed). WDYT @susodapop?

Issue #4004, opened 2019-07-22
Parameter values get lost when changing the mapping on dashboards


gabrieldutra:
### Issue Summary
Parameter values are not being kept when you change the parameter mapping:

![dashboard-mapping](https://user-images.githubusercontent.com/3356951/61603205-151ba480-ac13-11e9-8224-1eb8c6835b40.gif)

A summary of the issue and the browser/OS environment in which it occurs.

### Steps to Reproduce

1. Open a dashboard with widgets containing parameters (e.g: [this](https://redash-preview.netlify.com/dashboard/now-param))
2. Change the parameter mapping (widget -> dashboard or dashboard -> widget)
3. Change the parameter value
4. Change the parameter mapping back
5. Notice that the value set got lost

This seems to be happening due to the Parameter value being kept on the URL after you change the mapping (e.g: `p_param` is still in the url aside with `p_w190_param`).

### Technical details:

* Redash Version: Latest 
* Browser/OS: --
* How did you install Redash: --

ranbena:
@kravets-levko perhaps you can best tackle this?

arikfr:
I guess a simple solution is to make sure to remove from the URL parameters that are no longer relevant?

PR #4006, opened 2019-07-22
Migrate Parameters component to React

gabrieldutra:

- [x] Refactor

## Description
Migration of the Parameters Component to React.

--

parameters


gabrieldutra:
@ranbena it seems my worries became something to be figured out :joy:, since `parameters` is an array and React does a shallow comparison it doesn't trigger rerender on the component. `shouldComponentUpdate` won't do it, it seems it exists only to improve performance (`shouldComponentUpdate = () => true;` by default).

I'm thinking about other options (had a workaround for the dragging feature). Need to either create a new array for changes or use an external update trigger. Can use an external angular `$watch` to help, but I don't want to rely on that... I'll think better about this tomorrow.

arikfr:
> Can use an external angular `$watch` to help, but I don't want to rely on that...

You definitely don't :-)

ranbena:
> @ranbena it seems my worries became something to be figured out 😂, since `parameters` is an array and React does a shallow comparison it doesn't trigger rerender on the component. 

Why the need to re-render it?

gabrieldutra:
> Why the need to re-render it?

To the UI get updated when parameters get changed externally (e.g: parameter value changed).

ranbena:
> To the UI get updated when parameters get changed externally (e.g: parameter value changed).

You mean if someone else has modified the query / dashboard while another is viewing / editing it? I'm not sure that's a feature we should necessarily support.

gabrieldutra:
> You mean if someone else has modified the query / dashboard while another is viewing / editing it? I'm not sure that's a feature we should necessarily support.

Yep, I'm actually handling with state + an `onUpdate` function and will see where this goes. In any case, this is a bit trickier, there are cases where it's necessary to support (e.g: when you add a parameter in a query). So far, it's going well ^^.

---

BTW, I found a bug with the Apply Changes (initially thought it was an issue with this migration):

state-lost

When you edit the query you lose the "Apply Changes" state. The cause for this is in here: https://github.com/getredash/redash/blob/c518c7a4bccec3bde4aa2c13d274ce69edb26070/client/app/services/query.js#L290

Parameter is recreated and the `pendingValue` is not initiated in the constructor.

I'll write a fix for it today :)

gabrieldutra:
Yep, overall seems to be working with the workaround I mentioned:

parameters

I'm using Antd Button as a Link instead of the button from the [discussion](https://discuss.redash.io/t/migration-of-parameters-component-to-react/4161/4):
image

I'll officially open this for review tomorrow with Cypress tests + attempt to do the Drag Placeholder (just remembered of it)

ranbena:
Seems to work well 👍
What's the sort save logic? (seems to be saved on "Save" click with no dirty indication)

> I'm using Antd Button as a Link instead of the button from the [discussion](https://discuss.redash.io/t/migration-of-parameters-component-to-react/4161/4) (lmk if you prefer the button):

I've weighed both options a lot while making the mockup till it drove me crazy! 🤪
I'd like to see the icon encased in a button, to get a sense of the look and feel.
(Notice I went for a bootstrap button style cause ant ain't a good fit for this page yet)

> attempt to do the Drag Placeholder (just remembered of it)

I think it's just a matter of giving `.parameter-container { background-color: #f6f8f9; }`.
But make sure to give each parameter proper padding.
(don't forget to take the opportunity to move the relevant css from `ParameterValueInput.less`)

gabrieldutra:
> What's the sort save logic? (seems to be saved on "Save" click with no dirty indication)

Exactly that, guess it should follow the save-query-if-not-dirty rule?

> I've weighed both options a lot while making the mockup till it drove me crazy! 🤪
> I'd like to see the icon encased in a button, to get a sense of the look and feel. I'm still unsure about it. (Notice I went for a bootstrap button style cause ant ain't a good fit for this page yet)

Cool, will update and we choose later.

> I think it's just a matter of giving `.parameter-container { background-color: #f6f8f9; }`.

I initially thought it was easy like that, but this only works when the `.parameter-container` has only one row of parameters :grimacing:

> (don't forget to take the opportunity to move the relevant css from `ParameterValueInput.less`)

:+1:

ranbena:
> > What's the sort save logic? (seems to be saved on "Save" click with no dirty indication)
> 
> Exactly that, guess it should follow the save-query-if-not-dirty rule?

Honestly I think any change to params (value, type, position) should be reflected with the "apply changes" button and not the "Save". Do you think you can swing that (just the position)?

> > I think it's just a matter of giving `.parameter-container { background-color: #f6f8f9; }`.
> 
> I initially thought it was easy like that, but this only works when the `.parameter-container` has only one row of parameters 😬

Try this:
```css
.parameter-container {
  background-color: #f6f8f9;
  display: inline;
  padding: 35px 0 11px;
}
.parameter {
  display: inline-block;
  background: white;
  padding: 6px;
}
```

But perhaps styling the sortable ghost can work as well.

PS. Wdyt about `lockToContainerEdges: true`?

arikfr:
> Honestly I think any change to params (value, type, position) should be reflected with the "apply changes" button and not the "Save".

This doesn't feel right: "Apply Changes" is an end-user functionality that doesn't (necessarily) change the query object itself. Changing position, type, default value is a query editor functionality. Merging the two into "Apply Changes" feels confusing.

ranbena:
> This doesn't feel right: "Apply Changes" is an end-user functionality that doesn't (necessarily) change the query object itself. Changing position, type, default value is a query editor functionality. Merging the two into "Apply Changes" feels confusing.

I disagree. Save button is directly connected to the sql query content, while parameter content (apart for param key which appears in the sql query) has no representation in it, therefore no correlation. Save button for sql query, apply button for params. And they'll pay for the wall.

arikfr:
> Save button for sql query

This I can agree with, but I still don't think that "Apply Changes" is the right thing to use for applying parameter configuration changes.

I would retain the same behavior we have for other parameter configuration changes (i.e. auto save unless dirty) and revisit this when we redo this whole page.

gabrieldutra:
@ranbena the drag placeholder worked for most cases, but it is not correct sometimes (mainly with 3+ parameters with different sizes). I think styling the ghost wouldn't work because you would be styling the ghost of the parameter you are moving, not the position it will be afterwards (Default styling for that is to be hidden). 

Updated the button to the Old styling, so we can get the feel of it.

Created 2 new Cypress tests: - Parameter dragging; - Change parameter title from settings (so we make sure the settings dialog is opening and saving the parameter).

The behavior for the Sorting feature for now is "auto save unless dirty".

Guess from now it's a matter of final UI touches.

client/app/components/ParameterApplyButton.jsx
```diff
@@ -1,14 +1,13 @@
...
-import { react2angular } from 'react2angular';
 import Button from 'antd/lib/button';
 import Badge from 'antd/lib/badge';
 import Tooltip from 'antd/lib/tooltip';
 import { KeyboardShortcuts } from '@/services/keyboard-shortcuts';
 
-function ParameterApplyButton({ paramCount, onClick, isApplying }) {
-  // show spinner when applying (also when count is empty so the fade out is consistent)
-  const icon = isApplying || !paramCount ? 'spinner fa-pulse' : 'check';
+function ParameterApplyButton({ paramCount, onClick }) {
```
gabrieldutra: Removed `isApplying` as it was not affecting anything.

client/app/components/ParameterValueInput.less
```diff
@@ -18,65 +18,3 @@
     }
   }
 }
-
-.parameter-container {
```
gabrieldutra: Moved to `Parameters.less`.

client/app/components/Parameters.jsx
```diff
@@ -0,0 +1,166 @@
...
+      this.setState({ parameters });
+    }
+  };
+
+  handleKeyDown = (e) => {
+    const { parameters } = this.props;
+    const dirtyParamCount = size(filter(parameters, 'hasPendingValue'));
+
+    // Cmd/Ctrl/Alt + Enter
+    if (dirtyParamCount > 0 && e.keyCode === 13 && (e.ctrlKey || e.metaKey || e.altKey)) {
```
gabrieldutra: The easiest was to make it independent of the OS, so any of the modifier keys + enter will trigger Apply. LMK if there are issues with this approach.

client/app/components/Parameters.jsx
```diff
@@ -0,0 +1,166 @@
...
+  };
+
+  showParameterSettings = (parameter, index) => {
+    const { onParametersEdit } = this.props;
+    EditParameterSettingsDialog
+      .showModal({ parameter })
+      .result.then((updated) => {
+        this.setState(({ parameters }) => {
+          const updatedParameter = extend(parameter, updated);
+          parameters[index] = new Parameter(updatedParameter, updatedParameter.parentQueryId);
```
gabrieldutra: Same idea [here](https://github.com/getredash/redash/pull/3952/files#r303924717), but reconstructing the Parameter instead.

client/app/pages/queries/view.js
```diff
@@ -523,6 +523,21 @@ function QueryViewCtrl(
...
+  if (!$scope.query.isNew()) {
+    $scope.$watch(
+      'query.options && query.options.parameters',
+      (parameters) => {
+        const params = extend({}, $location.search());
+        parameters.forEach((param) => {
+          extend(params, param.toUrlParams());
+        });
+        Object.keys(params).forEach(key => params[key] == null && delete params[key]);
+        $location.search(params);
```
gabrieldutra: I initially thought this logic was only for the query page, but it's actually to all pages that contain parameters.

Will need to find a replacement to that. The fix for #4004 can probably be done while doing it.

ranbena:
Regarding title width, how about this?
1. Title max width 200 (or whatever you think) - so long titles get cut.
2. Title min width of the input width - so if the input is wide, the title doesn't get cut needlessly.
3. Input min width the title - so the input grows if the title is longer (looks better).

```less
.parameter-heading label {
  margin-bottom: 1px;
  overflow: hidden;
  text-overflow: ellipsis;
  min-width: 100%;
  max-width: 200px;
  white-space: nowrap;

  .parameter-container[data-editable] & {
    min-width: calc(100% - 27px); // make room for settings button
  }
}

.parameter-input,
.ant-input { // <- need to make it for all ant input components
  width: 100%;
}
```

#### Before:

Screen Shot 2019-07-31 at 8 36 30


#### After:

Screen Shot 2019-07-31 at 8 36 45

gabrieldutra:
> Regarding title width, how about this?

I loved the flexibility :grin:, not sure why, but it didn't work for Number inputs (will check that better). I let the Label width to be the same as the Date Range when it holds a Dynamic Value as transformation works better with an specific `width` value.

ranbena:
> I loved the flexibility 😁

CSS magic! 🎩✨🐇

> Not sure why, but it didn't work for Number inputs (will check that better).

Hmm. That's weird. Lmk if you need assistance.

> I let the Label width to be the same as the Date Range when it holds a Dynamic Value as transformation works better with an specific `width` value.

Sounds good.

client/app/components/ParameterValueInput.less
```diff
@@ -5,9 +5,12 @@
 .parameter-input {
   display: inline-block;
   position: relative;
+  width: 100%;
 
-  .@{ant-prefix}-input[type="text"] {
-    width: 195px;
+  .@{ant-prefix}-input,
+  .@{ant-prefix}-input-number,
+  .@{ant-prefix}-select {
```
ranbena: Suggestion for future, groups these selectors into a less variable `@ant-inputs` (if that's an actual less feature).

gabrieldutra:
> Hmm. That's weird. Lmk if you need assistance.

Yep that would be helpful :)

I made a test by setting the max label width to a small size, the result was this:
image

Notice that Inputs keep their sizes at `195px`. After spending some time looking for that in css and not finding anything I discovered that there's the `size=20` (20 is the browser default even though that attr isn't there) html attribute that seems to be taking place in such circumstance :sleepy:. If I add `size=10` in devtools the issue seems to go away. Problem: I guess I cannot replace that on Antd ^^

Should we work an specific styling for the Number input?

client/app/components/ParameterValueInput.less
```diff
@@ -5,9 +5,12 @@
...
+    width: 100%;
```
ranbena:
```suggestion
    min-width: 100% !important;
```
I think this fixes the all width problems 🤟

client/app/components/Parameters.jsx
ranbena: No issues with this approach.

No need to re-evaluate `dirtyParamCount` but instead:
```jsx
<div className="parameter-container" onKeyDown={dirtyParamCount ? this.handleKeyDown : null} ...>...</div>
```

client/app/components/Parameters.jsx
```diff
@@ -0,0 +1,199 @@
...
+    }
+  };
+
+  handleKeyDown = (e) => {
+    const { parameters } = this.props;
+    const dirtyParamCount = size(filter(parameters, 'hasPendingValue'));
+
+    // Cmd/Ctrl/Alt + Enter
+    if (dirtyParamCount > 0 && e.keyCode === 13 && (e.ctrlKey || e.metaKey || e.altKey)) {
+      e.stopPropagation();
```
ranbena: This does not work for Date components https://github.com/getredash/redash/issues/4025.
I fiddled with it but could not find a solution yet.

client/app/components/Parameters.less
```diff
@@ -0,0 +1,121 @@
...
+  background: white;
+  padding: 3px 6px 6px;
+
+  .parameter-container[data-draggable] & {
+    margin: 4px 0 0 4px;
+  }
+
+  &.parameter-dragged {
+    box-shadow:  0 4px 9px -3px rgba(102, 136, 153, 0.15);
+  }
```
ranbena:
```suggestion
    width: auto !important;
  }
```

Or else this happens:


That's cause in this case the element computes to `width: 93.2px` but the sortable rounds it down to `93px`. So keeping it untouched with `width: auto !important` fixes it .

client/app/components/ParameterValueInput.less
gabrieldutra: Thanks! :slightly_smiling_face:

client/app/components/Parameters.jsx
gabrieldutra: I'll check the issue :+1:

client/app/components/ParameterValueInput.less
gabrieldutra: Note: Just tried that, doesn't seem to work ^^. Also, in this case I don't want Date Picker parameters to be included (the transformation effect thing).

client/cypress/support/commands.js
```diff
@@ -50,3 +50,18 @@ Cypress.Commands.add('fillInputs', (elements) => {
     cy.getByTestId(testId).clear().type(value);
   });
 });
+
+Cypress.Commands.add('dragBy', { prevSubject: true }, (subject, offsetLeft, offsetTop, force = false) => {
```
gabrieldutra: This is actually optional as I've already changed and eventually needed to create a new drag helper for `Parameters`. LMK if you prefer it as a Cypress command or a dashboard api function.

gabrieldutra:
@ranbena (CC @arikfr) I again feel very happy with the result, I see 2 improvements within this migration: The updated Drag feature and the flexible width label naming. Overall I feel this is ready, will run some tests tomorrow to make sure it's solid, but code shouldn't change.

react-parameters-1
react-parameters-2

client/app/components/ParameterValueInput.less
```diff
@@ -5,9 +5,15 @@
...
 
-  .@{ant-prefix}-input[type="text"] {
-    width: 195px;
+  .@{ant-prefix}-input,
+  .@{ant-prefix}-input-number {
+    min-width: 100% !important;
+  }
+
+  .@{ant-prefix}-select {
+    width: 100%;
```
gabrieldutra: Fix this :beetle: (Select needs a `min-width`):
min-width-search

ranbena: Select doesn't have minWidth 60px?

client/app/components/Parameters.less
```diff
@@ -0,0 +1,124 @@
...
+    background-color: #f6f8f9;
+  }
+
+  .parameter-apply-button {
+    display: none; // default for mobile
+    
+    // "floating" on desktop
+    @media (min-width: 768px) {
+      position: absolute;
+      bottom: -42px;
```
ranbena:
```suggestion
      bottom: -36px;
```
Sits in a more accurate position.

client/app/components/ParameterValueInput.less
gabrieldutra: Yes, previously we were replacing that with the `!important` flag, so now it's flexible and has minWidth 60px

gabrieldutra:
@arikfr if you fine with it, this is good to go :slightly_smiling_face:

client/app/components/Parameters.jsx
```diff
@@ -138,7 +139,7 @@ export class Parameters extends React.Component {
         data-test={`ParameterName-${param.name}`}
       >
         <div className="parameter-heading">
-          <label>{param.title || param.name}</label>
+          <label>{param.title || toHuman(param.name)}</label>
```
ranbena: 👍

gabrieldutra:
Cypress making a good job: Looks like the new pending value set is not triggering Angular update, thus the execute button doesn't change disabled status unless you change something else like the query text... I'm looking into that.

ranbena:
@gabrieldutra let's land this!

arikfr:
Are you sure it's stable enough to land it a moment before cutting a release branch?

ranbena:
> Are you sure it's stable enough to land it a moment before cutting a release branch?

No. Better wait for it. Plz cut ✂️

arikfr:
:+1:

PR #4010, opened 2019-07-26
Fix Apply Changes is lost when query is edited

gabrieldutra:

- [x] Bug Fix

## Description
When editing a query the "Apply Changes" state was lost:
state-lost

--

apply-all-edit-query

### DIFF
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -39,7 +39,7 @@ export class ParameterValueInput extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      value: props.parameter.pendingValue || props.value,
+      value: props.parameter.hasPendingValue ? props.parameter.pendingValue : props.value,
       isDirty: props.parameter.hasPendingValue,
     };
   }
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -49,7 +49,7 @@ export class ParameterValueInput extends React.Component {
     // if value prop updated, reset dirty state
     if (prevProps.value !== value || prevProps.parameter !== parameter) {
       this.setState({
-        value: parameter.pendingValue || value,
+        value: parameter.hasPendingValue ? parameter.pendingValue : value,
         isDirty: parameter.hasPendingValue,
       });
     }
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -39,16 +39,19 @@ export class ParameterValueInput extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      value: props.value,
-      isDirty: false,
+      value: props.parameter.pendingValue || props.value,
+      isDirty: props.parameter.hasPendingValue,
     };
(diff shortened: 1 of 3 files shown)


client/app/components/ParameterValueInput.jsx
```diff
@@ -39,16 +39,19 @@ export class ParameterValueInput extends React.Component {
...
     };
   }
 
   componentDidUpdate = (prevProps) => {
-    const { value } = this.props;
+    const { value, parameter } = this.props;
     // if value prop updated, reset dirty state
-    if (prevProps.value !== value) {
-      this.setState({ value, isDirty: false });
+    if (prevProps.value !== value || prevProps.parameter !== parameter) {
```
ranbena: I think that now there's no actual need for any condition, right?

client/app/components/ParameterValueInput.jsx
```diff
@@ -39,16 +39,19 @@ export class ParameterValueInput extends React.Component {
...
+      this.setState({
+        value: parameter.pendingValue || value,
```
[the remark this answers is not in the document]
gabrieldutra: Forgot those values existed :dizzy_face:

client/app/components/ParameterValueInput.jsx
gabrieldutra: It is also called when state is update, removing it would cause an infinite loop :slightly_smiling_face:

client/app/components/ParameterValueInput.jsx
```diff
@@ -39,16 +39,19 @@ export class ParameterValueInput extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      value: props.value,
-      isDirty: false,
+      value: props.parameter.pendingValue || props.value,
```
gabrieldutra:
```suggestion
      value: props.parameter.hasPendingValue ? props.parameter.pendingValue : props.value,
```

gabrieldutra:
> I've noticed that this happens to Query Based Dropdown Lists after typing in the query editor:

This also happens in master, the issue is that the `parameters` DOM is reconstructed every time the objects are changed (from my tests the constructor of `ParameterValueInput` was called, not `componendDidUpdate`). I changed the approach to avoid this in f81650dd728d739784ec1a5c25d26b3af7051ed1. 

I also removed the initialization of the pendingValue (although I think it should be there), because parameters could have this prop saved before when saving a query and with the initialization now they would be rendered with the state as well.

Will run some tests to make sure nothing breaks with it and comment back here.

client/app/services/query.js
```diff
@@ -429,7 +429,8 @@ class Parameters {
 
     const parameterExists = p => includes(parameterNames, p.name);
     const parameters = this.query.options.parameters;
-    this.query.options.parameters = parameters.filter(parameterExists).map(p => new Parameter(p, this.query.id));
+    this.query.options.parameters = parameters.filter(parameterExists)
+      .map(p => (p instanceof Parameter ? p : new Parameter(p, this.query.id)));
```
gabrieldutra: This was introduced a long ago in #1649, the change seems not be affecting the fix :stuck_out_tongue_closed_eyes:

Issue #3806, opened 2019-05-15
Adopt Black for formatting Python code


arikfr:
Black is being adopted by Django  and other projects as the default formatter, and it makes sense for us to adopt it too. Only blocker is the fact it needs Python 3 to run, but our codebase is still Python 2.

While Black can format Python 2, without us making the switch it will be harder for developers to run the tool.

Blocked by: #4181.

jezdez:
🏴

PR #4011, opened 2019-07-26
Fix according to pycodestyle format

yoshiken:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description
Corrected the code style to pep8.
A correction that does not affect the operation.

before
```
$pycodestyle redash/|wc -l                                                                                                                                                     181
```

after
```
$pycodestyle redash/|wc -l                                                                                                                                                     111
```


arikfr:
Thanks, but in the future please avoid opening style only changing pull requests. Until we adopt a single formatting tool (#3806) we prefer to address these things on an adhoc basis, when editing the code in a specific file.

PR #4012, opened 2019-07-27
Updated timeago strings

ranbena:
- [x] Refactor

## Description
1. "a few seconds ago" is too common to be so long. "just now" is better.
2. Sometimes differences in server/client clocks cause the confusing "**in** a few seconds". Now will simply say "just now".
3. Updated the tooltip to ant design.

This also manifests to "Outdated Queries", "Users" list and "Alerts" list.

Before|After
-------|-------
Screen Shot 2019-07-27 at 16 42 27 | Screen Shot 2019-07-27 at 16 40 41

### DIFF
--- a/client/app/components/TimeAgo.jsx
+++ b/client/app/components/TimeAgo.jsx
@@ -6,6 +6,25 @@ import PropTypes from 'prop-types';
 import { Moment } from '@/components/proptypes';
 import { clientConfig } from '@/services/auth';
 import useForceUpdate from '@/lib/hooks/useForceUpdate';
+import Tooltip from 'antd/lib/tooltip';
+
+moment.updateLocale('en', {
+  relativeTime: {
+    future: '%s',
+    past: '%s',
+    s: 'just now',
+    m: 'a minute ago',
+    mm: '%d minutes ago',
+    h: 'an hour ago',
+    hh: '%d hours ago',
+    d: 'a day ago',
+    dd: '%d days ago',
+    M: 'a month ago',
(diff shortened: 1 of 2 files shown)


client/app/components/TimeAgo.jsx
```diff
@@ -6,6 +6,25 @@ import PropTypes from 'prop-types';
 import { Moment } from '@/components/proptypes';
 import { clientConfig } from '@/services/auth';
 import useForceUpdate from '@/lib/hooks/useForceUpdate';
+import Tooltip from 'antd/lib/tooltip';
+
+moment.updateLocale('en', {
```
arikfr: As this is something that applies globally it probably should be somewhere else (`app/config`?).

client/app/components/TimeAgo.jsx
```diff
@@ -6,6 +6,25 @@ import PropTypes from 'prop-types';
...
+  relativeTime: {
+    future: '%s',
+    past: '%s',
+    s: 'just now',
```
arikfr: Did we change any of the other strings or only this one?

ranbena: Nay.
I experimented with short-handing ("2 minutes ago" to "2m ago") but there's no straightforward way with `moment` to apply it for widgets only.

client/app/components/TimeAgo.jsx
ranbena: 👍 56b6571cca7da6b39192413069615885be957b8b

PR #4007, opened 2019-07-27
Dropdown endpoint is called on every keystroke

rauchy:
### Issue Summary

When editing the source of a query with a dropdown parameter, the `/api/queries/<query_id>/dropdown` endpoint is hit on every keystroke.

### Steps to Reproduce

1. Create a query with a dropdown parameter
2. Type in something in the source editor
3. Watch the Network tab for requests

### Technical details:

* Redash Version: master
* Browser/OS: Chrome

### DIFF
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -569,11 +569,13 @@ function QueryResource(
       asDropdown: {
         method: 'get',
         isArray: true,
+        cache: true,
         url: 'api/queries/:id/dropdown',
       },
       associatedDropdown: {
         method: 'get',
         isArray: true,
+        cache: true,
         url: 'api/queries/:queryId/dropdowns/:dropdownQueryId',
       },
       favorites: {


rauchy:
Looking into this, I realized that the parameter inputs are mounted from scratch on each re-render, and that the parameter objects (and query objects) are actually fresh objects, so there's no way to apply any memoization of these values into our object model.

A different approach to this is caching the Angular resource, effectively returning the same values within the same browser session (a page reload will invalidate this).

While it does save a significant amount of resource fetching (think someone typing in a large query), it increases the possibility of side-effects (for example, fresh values added to the dropdown won't show up until a page reload).

Which approach do you think is better? Perhaps you can think of another solution?

gabrieldutra:
@rauchy I've just realized I may have this issue fixed on #4010 while fixing another issue :sweat_smile:

rauchy:
@gabrieldutra awesome stuff! 👍

PR #4013, opened 2019-07-28
Recycle gunicorn workers

rauchy:
- [x] Feature

## Description
In order to avoid some issues that pop up on long-standing workers (faulty connections etc), it could be useful to refresh workers every n-th request.


HughWarrington:
There's clearly a memory leak in the code, shouldn't we at least try to fix it rather than applying a performance-sapping band aid?

PR #4015, opened 2019-07-29
Fix: support for unicode in DynamoDB queries

arikfr:

- [x] Bug Fix

## Description

It was trying to cast query into a string which will fail if it's unicode.

### DIFF
--- a/redash/query_runner/dynamodb_sql.py
+++ b/redash/query_runner/dynamodb_sql.py
@@ -102,8 +102,10 @@ class DynamoDBSQL(BaseSQLQueryRunner):
         try:
             engine = self._connect()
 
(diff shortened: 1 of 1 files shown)


redash/query_runner/dynamodb_sql.py
```diff
@@ -102,8 +102,10 @@ def run_query(self, query, user):
         try:
             engine = self._connect()
 
-            result = engine.execute(query if str(
-                query).endswith(';') else str(query) + ';')
+            if not query.endswith(';'):
```
arikfr: This logic isn't perfect but it's the same as we had before just without the unicode issue.

PR #4016, opened 2019-07-29
Cypress: Force readonly inputs click

gabrieldutra:

- [x] Other

## Description
After cypress update to v3.4.1 readonly inputs are not reachable by default, so I added `{ force: true }` where needed to make existing CI tests pass again.

--

--


gabrieldutra:
This error felt a bit weird to me, so I reached Cypress and it is actually supposed to only affect `cy.type`. Issue: cypress-io/cypress#4874

We can revert this once they solve it :)

ranbena:
I knew it!
I read that it was for type and not click so i opened their code and then said to myself - forget it, you don't have to know everything!

PR #4019, opened 2019-07-30
Styling changes

ranbena:
- [x] Feature

## Description
ezgif com-video-to-gif

Added:
1. Transitioning bg color when dragging.
2. Box shadow to dragged element.
3. A few minor styling changes.

### DIFF
--- a/client/app/components/Parameters.less
+++ b/client/app/components/Parameters.less
@@ -22,8 +22,7 @@
     margin: 4px 0 0 4px;
   }
 
-  // dragged element
-  body > & {
+  &.parameter-dragged {
     box-shadow:  0 4px 9px -3px rgba(102, 136, 153, 0.15);
   }
 }
--- a/client/app/components/Parameters.less
+++ b/client/app/components/Parameters.less
@@ -43,7 +43,7 @@
   &[data-draggable] {
     padding: 0 4px 4px 0;
     transition: background-color 200ms ease-out;
-    transition-delay: 500ms; // short pause before returning to original bgcolor
+    transition-delay: 300ms; // short pause before returning to original bgcolor
   }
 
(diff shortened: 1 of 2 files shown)


client/app/components/Parameters.less
```diff
@@ -7,29 +7,50 @@
...
   background: white;
-  padding: 6px;
+  padding: 3px 6px 6px;
+
+  .parameter-container[data-draggable] & {
+    margin: 4px 0 0 4px;
+  }
+
+  // dragged element
+  body > & {
```
ranbena: No other hook 😅

gabrieldutra: It is possible to use `helperContainer` attr to append it somewhere else than the body, but I don't think it's worth it :)

gabrieldutra: I wonder if `helperClass` is targeted on that element, I'll check that and edit this comment with my conclusion.

ranbena: aaad44e

client/app/components/Parameters.less
```diff
@@ -7,29 +7,49 @@
...
 .parameter-container {
   position: relative;
-  background-color: #f6f8f9;
-  display: inline;
-  padding: 35px 0 11px;
+
+  &[data-draggable] {
+    padding: 0 4px 4px 0;
+    transition: background-color 200ms ease-out;
+    transition-delay: 500ms; // short pause before returning to original bgcolor
```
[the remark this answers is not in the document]
ranbena: You're right. Decreased fc54399

ranbena:
> One last thing to be discussed: How should we handle long parameter titles (longer than the parameter width actually)?

I'm gonna reply to this in your PR.

PR #4026, opened 2019-07-31
Cannot edit dashboard after adding widget

kravets-levko:

- [x] Bug Fix

## Description

After adding a new widget user cannot edit dashboard properties: name, tags, enable dashboard-level filters. It happens because after adding a new widget dashboard version is increased but not sent to frontend which continues using old version number.

### DIFF
--- a/redash/handlers/widgets.py
+++ b/redash/handlers/widgets.py
@@ -24,12 +24,11 @@ class WidgetListResource(BaseResource):
         :>json object widget: The created widget
         """
         widget_properties = request.get_json(force=True)
-        dashboard = models.Dashboard.get_by_id_and_org(widget_properties.pop('dashboard_id'), self.current_org)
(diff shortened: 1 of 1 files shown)


kravets-levko:
`models.db.session.add(widget)` will stage `widget.dashboard` as well, and then `models.db.session.commit()` will run all hooks on staged instances, increase version number of the dashboard and save it as well. So it's the most straightforward solution: not to stage dashboard object. But widget needs `dashboard_id`, and as it was available within passed data (and later validated through `require_object_modify_permission(dashboard, self.current_user)`) - I decided just to use it.

kravets-levko:
BTW - Why do we need two `models.db.session.commit()` calls there?

arikfr:
> BTW - Why do we need two `models.db.session.commit()` calls there?

Probably some merge issue :-)

PR #4028, opened 2019-08-01
Add assume role as a credential source for AWS Athena Query runner 

ialeinikov:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
If enabled (via environment variable flag) exposes alternative assume role to be used instead of an IAM user.

Screen Shot 2019-08-01 at 11 39 33 AM


redash/query_runner/athena.py
```diff
@@ -118,13 +140,31 @@ def annotate_query(cls):
...
+                'aws_access_key_id': creds['Credentials']['AccessKeyId'],
+                'aws_secret_access_key': creds['Credentials']['SecretAccessKey'],
+                'aws_session_token': creds['Credentials']['SessionToken'],
+                'region_name': self.configuration['region']
+            }
+        else:
+            return {
+                'aws_access_key_id': self.configuration.get('aws_access_key', None),
+                'aws_secret_access_key': self.configuration.get('aws_secret_key', None),
+                'aws_session_token': None,
```
arikfr: Considering we weren't passing this before, maybe it's safer to omit this option instead of passing `None`?

```suggestion
```

redash/query_runner/athena.py
```diff
@@ -118,13 +140,31 @@ def annotate_query(cls):
...
 
-    def __get_schema_from_glue(self):
-        client = boto3.client(
-                'glue',
-                aws_access_key_id=self.configuration.get('aws_access_key', None),
-                aws_secret_access_key=self.configuration.get('aws_secret_key', None),
-                region_name=self.configuration['region']
+    def _get_iam_credentials(self, user=None):
+        if ASSUME_ROLE:
+            role_session_name = 'redash' if user is None else str(user.email)
```
arikfr: Why do you need to call `str()` on `user.email`?

ialeinikov: I wasn't sure if it's a string, removed.

redash/query_runner/athena.py
ialeinikov: yep, fair point, removed.

arikfr:
Thanks!

Issue #4025, opened 2019-07-31
Ctrl/Cmd+Enter not working with date parameters


arikfr:
### Steps to Reproduce

1. Change the value of a date or date range parameter.
2. Hit Ctrl+Enter.

Expected: it would apply changes.
Actual: nothing happens. :cry: 

### Technical details:

* Redash Version: latest master

ranbena:
Looks like the calendar hijacks the keyboard, so ctrl+ent closes it then another applies.
I'll look into it.

ranbena:
Oh I think it's just a matter of disabling the need for "ok" click as now it's not needed.

gabrieldutra:
The behavior seems a bit different with me, am I missing something?

![ctrl-enter-date-params](https://user-images.githubusercontent.com/3356951/62310105-2a889e00-b45f-11e9-88a6-079f5946137f.gif)

@ranbena for the React version it's basically the same, with the difference that `e.preventDefault()` doesn't allow it to trigger the calendar opening.

I tested when the Calendar was closed, when it's open "the calendar hijacks the keyboard" seems to happen. The same doesn't happen in #4006.

gabrieldutra:
It is indeed not working when you select a Dynamic value to it, but in this case seems to be a matter of giving focus to the parameter after the selection.

ranbena:
> It is indeed not working when you select a Dynamic value to it, but in this case seems to be a matter of giving focus to the parameter after the selection.

That's right.

> I tested when the Calendar was closed, when it's open "the calendar hijacks the keyboard" seems to happen. The same doesn't happen in #4006.

Also right. Still, the calendar opens up on apply with ctrl+enter.
I struggled with it a bit but didn't get anywhere.

gabrieldutra:
> Still, the calendar opens up on apply with ctrl+enter.

It doesn't on the React version :stuck_out_tongue_closed_eyes:. 

I'll start by working on the focus on the Date Picker and then we see where we can get :)

ranbena:
> > Still, the calendar opens up on apply with ctrl+enter. 

> It doesn't on the React version 😝.

I meant this: "(Ctrl+enter applies the value and opens the calendar again as enter key press event is passed to Antd DatePicker)"

gabrieldutra:
> I meant this: "(Ctrl+enter applies the value and opens the calendar again as enter key press event is passed to Antd DatePicker)"

Me too, the preventDefault worked to me.

[...]

PR #4033, opened 2019-08-02
Focus DatePicker after selecting dynamic values

gabrieldutra:

- [x] Bug Fix

## Description
Didn't have much time today, so I will continue tomorrow. So far it holds some changes to give Date and Date Range focus after selecting a value.

@ranbena you were right, the issue still happens in the React version, I was testing with Date Range parameters and it only happens with Date Parameters :sweat_smile:.

#4025 

--

### DIFF
--- a/client/app/components/DateInput.jsx
+++ b/client/app/components/DateInput.jsx
@@ -1,17 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { react2angular } from 'react2angular';
 import DatePicker from 'antd/lib/date-picker';
 import { clientConfig } from '@/services/auth';
 import { Moment } from '@/components/proptypes';
 
-export function DateInput({
+const DateInput = React.forwardRef(({
   defaultValue,
   value,
   onSelect,
   className,
   ...props
-}) {
+}, ref) => {
   const format = clientConfig.dateFormat || 'YYYY-MM-DD';
   const additionalAttributes = {};
   if (defaultValue && defaultValue.isValid()) {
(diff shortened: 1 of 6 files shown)


gabrieldutra:
@ranbena I'm not sure we will be able to handle the issue properly now. If I correctly understood what's happening is that the Date Parameter has an `Enter` key that's being triggered first and affecting propagation to `Parameters`. The issue is that Antd still don't have key handling on DatePicker's api ([ref](https://github.com/ant-design/ant-design/issues/15828)).

Something I was thinking is that it makes some sense for the Calendar to "steal" the Enter key when it's open. One thing we can do is to make sure it correctly steals it to avoid possible inconsistencies (testing on Date params in 5ded531f5eae47eb3a95ceb56d430e51251c13a4).

When the Calendar is closed I'm trying to see a workaround to focus some upper element to avoid it triggering Antd's default enter behavior. I don't think the issue is worth it the extra complexity on that, but lmk your opinion.

---

Just to illustrate what happened to me in the react version:
ctrl-enter-react-version
(selecting each parameter value and pressing Ctrl+Enter after each one)

ranbena:
> I don't think the issue is worth it the extra complexity on that, but lmk your opinion.

@gabrieldutra thanks for diving in to this. I too have not found a simple way of mitigating this. I think the issue is not worth the code complexity and our time at this point.

gabrieldutra:
We can deliver the focus after clicking on Dynamic Values anyway (c21ad9cb3ad1edd5780df253e73b3e728934935f).

Issue #4031, opened 2019-08-01
Counter Target Value formatting doesn't match Counter Value


susodapop:
### Issue Summary

Potentially related to #2558. When I configure a number format for a counter, the Target Value doesn't follow the formatting rule.

### Steps to Reproduce

1. Create a counter
2. Format the counter value (two decimal places, `$` prefix
3. Add a Target Value

![image](https://user-images.githubusercontent.com/17067911/62317593-a2ef5f00-b45f-11e9-9e5d-b1cb9e01cfad.png)

### Technical details:

* Redash Version: SaaS
* Browser/OS: Firefox
* How did you install Redash: SaaS

djabhe:
HI ALL,
how to remove brackets in target value?

PR #4035, opened 2019-08-02
Counter visualization: formatting not applied to target value

kravets-levko:

- [x] Refactor
- [x] Bug Fix

## Description

Refactor:
- extracted formatting logic to a function (and revisited it) - so it will be easier later move it to db migration (added TODO to explain it);
- fixed some edge case handling (e.g. when target value column selected but it's not a number - previously counter highlighted "trend negative" value and showed an empty target value);
- replaced `underscore.string/numberFormat` with `numeral` and removed `underscore.string` dependency - it was the only file that used it;
- replaced AngularJS `number` filter with `numeral` (for consistency and to simplify future React migration); added `tooltipFormat` options that currently just matches `number` filter behavior; added TODO to implement this option in editor.

Fixes getredash/redash#4031

Before:
image

After:
image

### DIFF
--- a/client/app/visualizations/counter/index.js
+++ b/client/app/visualizations/counter/index.js
@@ -1,5 +1,5 @@
-import numberFormat from 'underscore.string/numberFormat';
-import { isNumber } from 'lodash';
+import { isNumber, toString } from 'lodash';
+import numeral from 'numeral';
 import { angular2react } from 'angular2react';
 import { registerVisualization } from '@/visualizations';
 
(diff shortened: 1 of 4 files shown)


kravets-levko:
@susodapop You're welcome to test the fix as well :bowing_man:

susodapop:
image
Yeah it works properly on my example!

client/app/visualizations/counter/index.js
```diff
@@ -1,5 +1,5 @@
-import numberFormat from 'underscore.string/numberFormat';
```
ranbena: Nice!

ranbena:
> image
> Yeah it works properly on my example!

Damn @susodapop you don't got Roboto on your system. Until we fix it, can you install it on your OS (for the vids)?

susodapop:
@ranbena You are way better at spotting fonts than I am! I downloaded Roboto to my system. Previous to that the counter was rendered in San Francisco.

image

PR #4044, opened 2019-08-06
Added arrow to multi-select component

### LINKED ISSUE #4042: Dropdown Parameter Missing Arrow
Previous versions:

image

Current version:

image

Was this intentional?

ranbena:
- [x] Feature

## Description
Added arrow to Select.
Fixes #4042.

### DIFF
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -111,7 +111,7 @@ export class ParameterValueInput extends React.Component {
         dropdownClassName="ant-dropdown-in-bootstrap-modal"
         showSearch
         showArrow
-        style={{ minWidth: allowMultipleValues ? 195 : 60 }}
+        style={{ minWidth: 60 }}
         notFoundContent={null}
         {...multipleValuesProps}
       >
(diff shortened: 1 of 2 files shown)


gabrieldutra:
Also needs to add the arrow to the Query Based Dropdown Parameter.

ranbena:
@gabrieldutra I would really like to also make it not `min-width: 195px`. Don't you think it's a waste of space?I appreciate your opinion 🙏

gabrieldutra:
> @gabrieldutra I would really like to also make it not `min-width: 195px`. Don't you think it's a waste of space?I appreciate your opinion

Go ahead, It is a waste of space indeed, my only concern is when it's empty-valued, but we can experience that with the arrow. One option is to add a placeholder in case it turns out to be weird looking.

ranbena:
Now that there's an arrow, it's obvious that the remaining space is clickable.

Screen Shot 2019-08-06 at 16 10 46

ranbena:
> my only concern is when it's empty-valued

It'll look exactly like an empty single value dropdown having minimum `60px` width.

Issue #2721, opened 2018-08-05
Make "missing parameter value" message friendlier


arikfr:
Currently we piggy back the query execution error mechanism to notify the user that they need to enter a value for a parameter when a value is missing.

On a query screen:

![image](https://user-images.githubusercontent.com/71468/43687881-737da8fc-98e6-11e8-8c94-2aed32dc2680.png)

On a dashboard:

![image](https://user-images.githubusercontent.com/71468/43687884-7c4967fa-98e6-11e8-8987-145cbd1978e5.png)

The big red message feels very intimidating and I wonder if there is a better way to convey this message?

kocsmy:
How about changing the design to warning or info so it'll be less intimidating? We can also refine the text to be friendlier: Please set a value for example parameter.

kocsmy:
We should add the error message next to the input fields.

arikfr:
We shouldn't treat this as an error message. It's just a state of the UI. One possible solution is to mark the parameters which miss value with some indication.

kocsmy:
I meant like an input validation error, Bootstrap has this supported by default:
<img width="374" alt="screenshot 2018-08-27 17 59 38" src="https://user-images.githubusercontent.com/2378022/44670497-19e7d080-aa23-11e8-93a3-7f8f4c7534fe.png">

thoechsmann:
Would also be good to allow for empty parameters. Sometimes that is intended and I have to make a workaround like enter a text like 'none' when string parameter should be empty.

kocsmy:
@arikfr I'm wondering if we should move those parameters inputs to Ant. Their error message is similar but placed below the form (which makes more sense to me). Furthermore, the date and time pickers are already in Ant.

![image](https://user-images.githubusercontent.com/2378022/48271892-a78d6680-e43d-11e8-819b-9ce62fb284ef.png)

arikfr:
I think that @kravets-levko's implementation of dashboard parameters already switches everything to React, so using Ant won't be a problem.

But --

1. I'm not sure we should show any message beyond marking the field.
2. I'm looking for a more holistic definition of the behavior here when trying to run a query (or refresh a dashboard) and some parameter is missing value.

kocsmy:
1. I think we should show error message. Traditionally, the UX of error message should 1. show where the error is, 2. tell what the error is, 3. recommend a solution. For example: Missing parameter, please fill. (or something similar)

[...]

PR #4049, opened 2019-08-07
Introduce inheritance to the Parameter structure

### LINKED ISSUE #3769: parentQueryId gets persisted in parameter options
`parentQueryId` gets persisted in parameter options. This is a frontend state and there is no reason to store it.

gabrieldutra:

- [x] Refactor

## Description
This is a start of a draft of what I was thinking of how parameters should be in code as we are increasing complexity for them. So far basically started with inheritance, the "abstract" Parameter holds the default methods that will be replaced as needed for the parameters.

This way it's easier to control Parameter valid values and introduce new features to existing parameters or add new types.

Opening this already to gather comments from your ideas.

Fixes #3769


client/app/services/parameters/Parameter.js
```diff
@@ -0,0 +1,77 @@
...
+  static create(param) {
+    switch (param.type) {
+      case 'text':
+        return new TextParameter(param);
+      case 'number':
+        return new NumberParameter(param);
+      default:
+        return null;
+    }
+  }
```
ranbena: This seems like an anti-pattern, no?
The super class should not be aware of its heirs.

client/app/services/parameters/TextParameter.js
```diff
@@ -0,0 +1,16 @@
...
+  constructor(parameter, parentQueryId) {
+    super(parameter, parentQueryId);
+    this.setValue(parameter.value);
+  }
+
+  getValue() {
+    const value = toString(this.value);
+    return !isEmpty(value) ? value : null;
+  }
+}
```
ranbena: Perhaps can also have a `normalize()` override which acknowledges 0 for NumberParameter and '' for TextParameter etc.

client/app/services/parameters/Parameter.js
gabrieldutra: :thinking: seemed ok as it was `static`, the goal is to have a factory method, can move it to an external function in any case :)

client/app/services/parameters/TextParameter.js
gabrieldutra: So far it was ok to have the logic in `getValue` as the method was simple. Do you see a case where we will need the `normalize` outside that method?

ranbena: Perhaps the param's `isDirty()` could use it too

client/app/services/parameters/Parameter.js
ranbena: It's a matter of taste :) Whatever you see fit 👍

client/app/services/parameters/Parameter.js
```diff
@@ -0,0 +1,130 @@
...
+  get isEmpty() {
+    return isNull(this.getValue());
+  }
+
+  get hasPendingValue() {
+    return this.pendingValue !== undefined && !isEqual(this.pendingValue, this.normalizedValue);
+  }
+
+  get normalizedValue() {
+    return this.$$value;
```
gabrieldutra: Basically from what I understood `$$value` will be the value used internally in Inputs and `value` is the stored value in the backend (also is the one that goes to the query in Refresh schedule cases). I'm not 100% happy with how this is, but not the biggest issue here and it worked so far, perhaps just gonna see different name to the attribute.

gabrieldutra:
So far moved all code to their separate files, I still want to revise the methods names (there are several "value" attributes and it would be nice if they were a bit more descriptive on what they do, or at least have comments). Also need to see if something else can be cleaned up. Will continue tomorrow. @kravets-levko and @ranbena lmk if you have inputs here :)

client/app/services/parameters/Parameter.js
```diff
@@ -0,0 +1,146 @@
...
+    this.name = parameter.name;
+    this.type = parameter.type;
+    this.global = parameter.global; // backward compatibility in Widget service
+    this.parentQueryId = parentQueryId;
+
+    // Used for meta-parameters (i.e. dashboard-level params)
+    this.locals = [];
+
+    // Used for URL serialization
+    Object.defineProperty(this, 'urlPrefix', {
```
gabrieldutra: As Cypress pointed out, this doesn't work with inheritance :slightly_smiling_face:

client/app/services/parameters/DateRangeParameter.js
```diff
@@ -0,0 +1,157 @@
...
+      if (this.dynamicValue) {
+        const dateRange = this.dynamicValue.value();
+        return {
+          start: dateRange[0].format(DATETIME_FORMATS[this.type]),
+          end: dateRange[1].format(DATETIME_FORMATS[this.type]),
+        };
+      }
+    }
+    return this.value;
+  }
```
ranbena: Tidy nit:

```js
getValue() {
  if (this.hasDynamicValue && this.dynamicValue) {
    const format = x => x.format(DATETIME_FORMATS[this.type]);
    const [start, end] = this.dynamicValue.value().map(format);
    return { start, end };
  }
  return this.value;
}
```

client/app/services/parameters/Parameter.js
ranbena: Totally! Can you provide a breakdown of the different value types and what format they're in?
1) param.value
2) param.$$value
3) param.getValue()
4) param.normalizedValue
5) param.pendingValue
etc.

gabrieldutra:
Review of existent values in the Parameter Structure:
- **param.value**: Value stored in DB. It's also used as Parameter Value for Scheduled queries that contain parameters. Can be String (dates are formatted), integer or a List in case of Multi-select parameters.
- **param.$$value**: Value used internally for Inputs. Guess it should not be stored in DB (not sure if it is now). Can be anything, dates are moment instances in this case.
- **param.getValue()**: Execution Value. It's the function that's is called to get Parameter value before execution. Usually the same as param.value, with some differences for example for Dynamic Values (they are solved into actual dates here). Thinking about renaming it to `getExecutionValue` or sth else that will make more sense.
- **param.normalizedValue**: Getter for $$value.
- **param.pendingValue**: Pending value for apply changes. Same type of value as $$value (except for Dynamic Values, but I'll review that too)
- **param.dynamicValue**: Returns the Dynamic Date or Date Range, which is basically an object with a Label and a function that generates moment values. I'm thinking about merging it into $$value.
- **param.setValue(value)**: Takes anything as a value, normalizes and modify both the value stored and the internal value. 

From those need to make sure only `param.value` is stored

gabrieldutra:
Dynamic Values merged in `$$value` (also added `DynamicDate[Range]Type` PropType).

Regarding values, this is what I've been thinking: (once defined I would also put as comment in the "abstract" Parameter class)
- **this.value** (attr): DB value
- **this.$$value** / **this.normalizedValue** (attr): Input value, can always be obtained through the `this.normalizeValue(value)` method;
- **this.normalizeValue(value)** (func): Normalizes input values. Should turn anything that makes sense into a predefined type (e.g.: turn an integer into a String for TextParameters) or to `null` in case it can't. In the basic form, it needs to have instructions to normalize both `this.value` and `this.$$value` types.
- **this.setValue(newValue)** (func): Defines both `this.value` and `this.$$value` as the normalizeValue form, should be overridden in the cases where `this.value` is different.
- **this.getExecutionValue()** (func): Execution value. Basic form is just returns `this.value`. Overridden when it should have any other special treat (e.g.: dynamic values or multi-select).

ranbena:
I like it a lot. So glad you're doing this :)

Thinking about this, in order to make things _unmistakable_, wdyt of having **one value only** which can be converted into different formats via static method `param.formatValue`?:
* `param.value` - for storage
* `param.formatValue(param.value, FORMAT.DISPLAY)` - for humans
* `param.formatValue(param.value, FORMAT.SQL_QUERY)` - for sql server

It's long, might not be performant, but I'd like to find ways to minimize confusion above all.

gabrieldutra:
> wdyt of having **one value only**

Definitively more readable, the `FORMAT.TYPE` isn't usual unless you consider `formatValue` as a "private" method. My other concern was on the various possible executions of the function to get the Input Value (that's why I left `$$value` in there). However I liked it, I'll maturate this idea and follow my conclusions :)

ranbena:
> the `FORMAT.TYPE` isn't usual unless you consider `formatValue` as a "private" method.

Yeah I guess it's meaningless.

> My other concern was on the various possible executions of the function to get the Input Value (that's why I left `$$value` in there).

You mean performance, yeah? Maybe memoize?

gabrieldutra:
@ranbena I'm actually thinking about keeping them as variables (updated on `setValue`) and just rename them to something that will make more sense (or add helpful comments). I think the memoize thing could work, but also seems to me we would need to reset at every `setValue` (as it can mess up for the values that depend on more than the function parameters).

@kravets-levko lmk if you have other suggestions for the parameter structure.

client/app/services/parameters/Parameter.js
```diff
@@ -146,6 +151,7 @@ class Parameter {
     return `{{ ${this.name} }}`;
   }
 
+  /** Get a saveable version of the Parameter by omitting unnecessary props */
```
gabrieldutra: @ranbena I'm thinking about using JSDoc syntax for the most confusing methods (can either be as simple as that or hold more info). Considering #4135 it would be helpful in VSCode for autocomplete :)

client/app/services/parameters/DateParameter.js
```diff
@@ -0,0 +1,95 @@
...
+  'datetime-local': 'YYYY-MM-DD HH:mm',
+  'datetime-with-seconds': 'YYYY-MM-DD HH:mm:ss',
+};
+
+const DYNAMIC_PREFIX = 'd_';
+
+const DYNAMIC_DATES = {
+  now: {
+    name: 'Today/Now',
+    value: () => moment(),
```
kravets-levko: Just curious: shouldn't it be `moment.utc()`?

client/app/services/parameters/Parameter.js
```diff
@@ -0,0 +1,160 @@
...
+
+  /** Get normalized value to be used in inputs */
+  get normalizedValue() {
+    return this.$$value;
+  }
+
+  // TODO: Remove this property when finally moved to React
+  get ngModel() {
+    return this.normalizedValue;
+  }
```
kravets-levko: Do we use this property anywhere? I mean - can it be removed right now?

client/app/services/parameters/DateParameter.js
gabrieldutra: In this case we can and want to use the user's TZ (I believe it's less confusing and it reaches more use cases).

client/app/services/parameters/Parameter.js
gabrieldutra: I thought it could be something used by Angular internally :sweat_smile:, but from a quick search on when it was introduced it does seem we don't use it anymore, so yeah, we can get rid of that :scissors: 

Thanks!

client/app/services/parameters/DateParameter.js
kravets-levko: When user selects a dynamic value for parameter - what do we pass to a server when query is executed: "magic" constant or computed value?

client/app/services/parameters/DateRangeParameter.js
```diff
@@ -0,0 +1,163 @@
...
+
+  toUrlParams() {
+    const prefix = this.urlPrefix;
+    if (isObject(this.value) && this.value.start && this.value.end) {
+      return {
+        [`${prefix}${this.name}`]: null,
+        [`${prefix}${this.name}.start`]: this.value.start,
+        [`${prefix}${this.name}.end`]: this.value.end,
```
ranbena: This `start`/`end` implementation affects the superclass and subclasses by forcing them to be aware of reseting these. Personally, I think that's not good practice cause it doesn't adhere to agnosticism, therefore making code maintenance tricky and complex.

Any way around it?

Maybe move these keys to the value, allowing all classes to implement nothing beyond `[${prefix}${this.name}]`?
```js
toUrlParams() {
    const prefix = this.urlPrefix;
    if (isObject(this.value) && this.value.start && this.value.end) {
      return {
        [`${prefix}${this.name}`]: `start::${this.value.start}--end::${this.value.end}`,
      }
  }
}
```

client/app/services/parameters/DateParameter.js
gabrieldutra: Computed value (actual "dates/date ranges")

client/app/services/parameters/DateRangeParameter.js
```diff
@@ -0,0 +1,163 @@
...
+  constructor(parameter, parentQueryId) {
+    super(parameter, parentQueryId);
+    this.setValue(parameter.value);
+  }
+
+  get hasDynamicValue() {
+    return isDynamicDateRange(this.normalizedValue);
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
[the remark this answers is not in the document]
gabrieldutra: :thinking: I need the Parameter instance for two cases: Enum and QueryBased parameters. I initially went with a static method, but eventually changed to that due to this. An alternative is to have a static method, but pass the parameter as an argument (sth like `this.constructor.normalizeValue(this, value)`), though I still prefer the current syntax.

client/app/services/parameters/Parameter.js
```diff
@@ -0,0 +1,151 @@
...
+    this.name = parameter.name;
+    this.type = parameter.type;
+    this.global = parameter.global; // backward compatibility in Widget service
+    this.parentQueryId = parentQueryId;
+
+    // Used for meta-parameters (i.e. dashboard-level params)
+    this.locals = [];
+
+    // Used for URL serialization
+    this.urlPrefix = 'p_';
```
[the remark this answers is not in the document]
gabrieldutra: That doesn't concern me that much, but do you have any other idea in mind?

client/app/services/parameters/DateRangeParameter.js
gabrieldutra: I totally agree with you on this one, a different option is to use a different logic to remove unused parameters from the url. I'll explore those alternatives :slightly_smiling_face:

client/app/services/parameters/DateParameter.js
kravets-levko: Okay. I think, it may be an issue. Imagine that there is a query like `SELECT * FROM some_table WHERE created_at >= '{{ param }}'`, and that DB server is configured to use UTC, so `created_at` column contains UTC values. Now, imagine that I want to pass `yesterday` as param value. My local time zone is `GMT+3`, and if I'll try to use that param with `yesterday`, server will return results not for last 24 hours, but for last 27 hours - because client will pass my local time, but server will treat it as UTC - so there will be 3 hours difference. Also, this example can be extrapolated for a case with few users in different time zones - then each of them will get results for different period.

So shouldn't we always pass date/time/range params in UTC to make them predictable and not dependent on DB config and user's locale settings?

client/app/services/parameters/Parameter.js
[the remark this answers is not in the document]
ranbena: Well nothing concrete. My thinking is that `Parameter` should be agnostic to this property because it isn't the one using it. Dashboards, Widgets and Query pages use it with their own logic - therefore they should be the ones executing that logic.

But yeah I'd leave it as is for now :)

client/app/services/parameters/DateParameter.js
gabrieldutra: :thinking: you have a point, from my experience I've always used `CONVERT_TZ` in the queries so it would be less confusing for the people that were only going to view it and get the results. Though everyone had the same TZ in there. 

I still have the feeling that the user TZ is the less confusing when considering the user, so I believe it's not a bug, it's one acceptable behavior for this. But we could eventually offer more options in there. 

This is certainly not in the scope of this PR, so I say we should move this discussion to somewhere else with more visibility (forum or create an issue).

kravets-levko: I see your point about `CONVERT_TZ`, but I described a bit different case: when we send param value to DB server - it does not know in which TZ that value actually is, and what's the difference between server's TZ and provided value. Even more: values that come from different users may be in different TZ, which is totally impossible to handle in a single query. So actually there are two solutions: 1) pass TZ in param value (ISO8601 allows this), or 2) pass UTC (we can show value in any TZ in UI, but when executing query - convert it to UTC).

It's definitely not in the scope of this PR, but we have to discuss it somewhere.

arikfr: Let's have it as `moment.utc()` in this PR (I assume this is the current behavior?) and start a discussion about how to handle timezones apart from this PR. It's a broad enough topic that requires deeper discussion.

gabrieldutra: The current behaviour is `moment()`, so I'm assuming that's the one you meant we should keep in this scope.

client/app/services/parameters/DateRangeParameter.js
gabrieldutra: I like the idea of going even simpler on this, like using only a small separator `p_date_range=[start]--[end]`. Beside gaining more organization on code we also reduce the size of the url :)

On the other hand, one downside of changing is for people that have saved url links with parameter values. It doesn't seem a big issue, but if it turns out to be, it's possible to support both url syntaxes.

ranbena: Yeah I think it would be easy and to add a short backwards compatibility code 👍

ranbena:
@gabrieldutra I'm working on #2721 and I wanna base it on this PR.
Though I'm missing a param feature - `isEmpty` for the `pendingValue`.
If that's straight-forward, awesome.

If not, maybe consider making `pendingValue` be a `Parameter` instance (instead of `normalizedValue`), which will allow its `isEmpty` out of the box.

gabrieldutra:
Cool, @ranbena, I actually want to move forward and merge this soon anyway (will sync it today).

For the `isEmpty` thing I believe we can have it as a static method (`constructor.isEmpty(value)`) and so we would have it available at any given moment?

ranbena:
I’m with you on that 👍

Regarding isEmpty - don’t you think it should be allowed to be set by each param type?
Or is it always gonna be ‘not null’ ‘not empty string’?

client/app/services/parameters/Parameter.js
```diff
@@ -0,0 +1,153 @@
...
+  get normalizedValue() {
+    return this.$$value;
+  }
+
+  clone() {
+    return Parameter.create(this, this.parentQueryId);
+  }
+
+  isEmptyValue(value) {
+    return isNull(this.normalizeValue(value));
```
gabrieldutra: @ranbena I've realized I could generalize the "empty" concept like this (ideally invalid values are turned into "null")

client/app/services/parameters/NumberParameter.js
```diff
@@ -0,0 +1,20 @@
...
+    super(parameter, parentQueryId);
+    this.setValue(parameter.value);
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  normalizeValue(value) {
+    if (isNull(value)) {
+      return null;
+    }
+    const normalizedValue = toNumber(value);
```
ranbena: `toNumber` converts empty strings to 0, but I think it should be normalized to `null` instead.
How about:
```js
import { toNumber, trim } from 'lodash';

normalizeValue(value) {
    if (!trim(value)) {
      return null;
    }
    ...
}
```

client/app/services/parameters/TextParameter.js
```diff
@@ -0,0 +1,20 @@
...
+  // eslint-disable-next-line class-methods-use-this
+  normalizeValue(value) {
+    const normalizedValue = toString(value);
+    if (isEmpty(normalizedValue)) {
+      return null;
+    }
+    return normalizedValue;
+  }
+}
+
```
ranbena: Suggesting addition:
```js
isEmptyValue(value) {
  return !trim(value);
}
```

arikfr:
:ship:

PR #4045, opened 2019-08-07
refresh_queries requires Request Context

rauchy:
[Now that `refresh_queries` uses `query.parameterized.apply`](https://github.com/getredash/redash/commit/f0576a362352bd63ae3f96d70bd2d0da1d0646da#diff-5ca0dc869c9c3218d8a95ba7f99f5f4cR196), it requires Flask's Request Context to work.

As this code is used from a background job, it should work without a Request Context. In general we should avoid using Request Context code in models.

### DIFF
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -451,7 +451,7 @@ class QueryRefreshResource(BaseResource):
         require_access(query, self.current_user, not_view_only)
 
         parameter_values = collect_parameters_from_request(request.args)
-        parameterized_query = ParameterizedQuery(query.query_text)
+        parameterized_query = ParameterizedQuery(query.query_text, org=self.current_org)
 
         return run_query(parameterized_query, parameter_values, query.data_source, query.id)
 
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -263,10 +263,6 @@ class QueryResult(db.Model, BelongsToOrgMixin):
         }
 
     @classmethod
-    def get_by_id(cls, _id):
-        return cls.query.filter(cls.id == _id).one()
-
-    @classmethod
     def unused(cls, days=7):
         age_threshold = datetime.datetime.now() - datetime.timedelta(days=days)
         return (
@@ -692,7 +688,7 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
 
     @property
     def parameterized(self):
-        return ParameterizedQuery(self.query_text, self.parameters)
+        return ParameterizedQuery(self.query_text, self.parameters, self.org)
 
     @property
     def dashboard_api_keys(self):
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -263,6 +263,10 @@ class QueryResult(db.Model, BelongsToOrgMixin):
         }
 
     @classmethod
+    def get_by_id(cls, _id):
+        return cls.query.filter(cls.id == _id).one()
+
+    @classmethod
     def unused(cls, days=7):
         age_threshold = datetime.datetime.now() - datetime.timedelta(days=days)
         return (
--- a/redash/models/parameterized_query.py
+++ b/redash/models/parameterized_query.py
@@ -15,20 +15,20 @@ def _pluck_name_and_value(default_column, row):
     return {"name": row[name_column], "value": unicode(row[value_column])}
 
 
-def _load_result(query_id):
+def _load_result(query_id, org):
     from redash import models
 
-    query = models.Query.get_by_id(query_id)
+    query = models.Query.get_by_id_and_org(query_id, org)
 
     if query.data_source:
-        query_result = models.QueryResult.get_by_id(query.latest_query_data_id)
+        query_result = models.QueryResult.get_by_id_and_org(query.latest_query_data_id, org)
         return json_loads(query_result.data)
     else:
         raise QueryDetachedFromDataSourceError(query_id)
 
 
-def dropdown_values(query_id):
-    data = _load_result(query_id)
+def dropdown_values(query_id, org):
+    data = _load_result(query_id, org)
     first_column = data["columns"][0]["name"]
     pluck = partial(_pluck_name_and_value, first_column)
     return map(pluck, data["rows"])
@@ -112,8 +112,9 @@ def _is_value_within_options(value, dropdown_options, allow_list=False):
 
 
 class ParameterizedQuery(object):
-    def __init__(self, template, schema=None):
+    def __init__(self, template, schema=None, org=None):
         self.schema = schema or []
+        self.org = org
         self.template = template
         self.query = template
         self.parameters = {}
@@ -151,7 +152,7 @@ class ParameterizedQuery(object):
                                                            enum_options,
                                                            allow_multiple_values),
             "query": lambda value: _is_value_within_options(value,
-                                                            [v["value"] for v in dropdown_values(query_id)],
+                                                            [v["value"] for v in dropdown_values(query_id, self.org)],
                                                             allow_multiple_values),
             "date": _is_date,
             "datetime-local": _is_date,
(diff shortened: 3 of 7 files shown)


redash/tasks/queries.py
```diff
@@ -190,12 +191,18 @@ def refresh_queries():
...
-                    query_text = query.parameterized.apply(query_params).query
-                else:
-                    query_text = query.query_text
+                query_text = query.query_text
+
+                parameters = {p['name']: p.get('value') for p in query.options.get('parameters', [])}
+                if any(parameters):
+                    try:
+                        query_text = query.parameterized.apply(parameters).query
+                    except InvalidParameterError as e:
```
rauchy: I tend to think these should actually not be skipped but rather executed (with an unparameterized `query_text`) and fail - so they will show up in failure reports, in order to give users the chance to rectify them.

rauchy:
Tests coming up.

redash/models/__init__.py
```diff
@@ -262,6 +262,10 @@ def to_dict(self):
             'retrieved_at': self.retrieved_at
         }
 
+    @classmethod
+    def get_by_id(cls, _id):
```
rauchy: @arikfr how bad is this? What was the motivation behind adding an `org_id` filter on these? The way I understand it, query results identifiers are unique.

arikfr: It's an extra safety feature to enforce security in a multi tenant context. Even if someone managed to workaround our other safe guards, he will not be able to load a query from a different account. 

TL;DR very bad :)

redash/models/parameterized_query.py
```diff
@@ -17,16 +16,15 @@ def _pluck_name_and_value(default_column, row):
 
 
 def _load_result(query_id):
-    from redash.authentication.org_resolving import current_org
     from redash import models
 
-    query = models.Query.get_by_id_and_org(query_id, current_org)
+    query = models.Query.get_by_id(query_id)
```
arikfr: Instead of removing the `org_id` argument or relaying on `current_org`, I would pass it to `_load_result`.

arikfr: I think it should be ok to use the `org_id` value from the Query object we're rendering.

redash/tasks/queries.py
arikfr: We do the reporting at a different stage and hence why we can't report on them. If we move the parameter merging to the query task, then we can report on this issue.

redash/models/parameterized_query.py
rauchy: > Instead of removing the org_id argument or relaying on current_org, I would pass it to _load_result.

Figured we'd end up with that but thought I'd see if it's avoidable first.

redash/tasks/queries.py
rauchy: Heh, I was actually thinking of a far less elegant solution - when unable to do parameter merging - just let a broken unapplied query run and have the DB report a failure, which would end up in the report :)

redash/models/__init__.py
rauchy: 9cfadd8 😖

redash/models/parameterized_query.py
rauchy: Some `ParameterizedQuery` instances do not represent a persisted `Query` so we can't rely on having it all the time. Instead I just pass in the actual org 😖

redash/tasks/queries.py
rauchy: One caveat about moving the parameter merging to the query task - `enqueue_query` creates job locks according to the hashed query. If we don't apply the parameters before calling `enqueue_query`, we will lock according to hashed versions of strings like `SELECT {{foo}}`, which could prevent different queries from running if they hash to the same value.

This is probably a non-issue because scheduled parameterized queries use their default values, so they hash to the same value regardless. It's just disturbing to see.

arikfr: We should revise the hashing logic anyway, as it's currently lower cases the query. We can hash the query text + parameter values to de-duplicate based on actual query without merging in the values.

redash/tasks/queries.py
```diff
@@ -190,12 +191,20 @@ def refresh_queries():
...
             else:
-                if query.options and len(query.options.get('parameters', [])) > 0:
-                    query_params = {p['name']: p.get('value')
-                                    for p in query.options['parameters']}
-                    query_text = query.parameterized.apply(query_params).query
-                else:
-                    query_text = query.query_text
+                query_text = query.query_text
+
+                parameters = {p['name']: p.get('value') for p in query.options.get('parameters', [])}
```
arikfr:
```suggestion
                parameters = {p['name']: p.get('value') for p in query.parameters}
```

redash/tasks/queries.py
arikfr: Another option is to report from here directly to the error report. We should also increment the schedule failures counter to avoid having to reprocess these queries all the time.

Out of scope for this PR, but worth keeping in mind.

redash/tasks/queries.py
```diff
@@ -190,12 +191,20 @@ def refresh_queries():
...
+                if any(parameters):
+                    try:
+                        query_text = query.parameterized.apply(parameters).query
+                    except InvalidParameterError as e:
+                        logging.info("Skipping refresh of %s because of invalid parameters: %s", query.id, e.message)
```
arikfr: This and the other log message have the potential to significantly spam the log (in a busy system). Let's move them to `debug` for now.

redash/handlers/queries.py
```diff
@@ -451,7 +451,7 @@ def post(self, query_id):
         require_access(query, self.current_user, not_view_only)
 
         parameter_values = collect_parameters_from_request(request.args)
-        parameterized_query = ParameterizedQuery(query.query_text)
+        parameterized_query = ParameterizedQuery(query.query_text, org=self.current_org)
```
arikfr: Why aren't we using query.paremterized_query here?

rauchy: Using `query.parameterized` would create a `ParameterizedQuery` with the query schema baked into it which would, in turn, force parameter schema validation.

It might be desirable to verify the schema here as well, I just didn't want to change any behavior in that part of the code as part of this PR. Let me know if you think it's safe.

arikfr: Got it. Keep as is.

redash/tasks/queries.py
rauchy: Went ahead with the latter. Let me know if you want to keep cb82a53 or discard it.

redash/tasks/queries.py
```diff
@@ -366,10 +374,7 @@ def run(self):
             result = QueryExecutionError(error)
             if self.scheduled_query is not None:
                 self.scheduled_query = models.db.session.merge(self.scheduled_query, load=False)
-                self.scheduled_query.schedule_failures += 1
-                models.db.session.add(self.scheduled_query)
-                notify_of_failure(error, self.scheduled_query)
-            models.db.session.commit()
+                track_failure(self.scheduled_query, error)
```
arikfr: Previously, `models.db.session.commit` was happening in all cases. Now it happens only on failure.

rauchy: First - I think you mean it was happening for all queries and now it only happens with scheduled queries?

Second - I have a hunch this was a mistake with indentation, as there is no `db.session` work prior to it.

arikfr: I thought it's for saving the result, but that actually happens in a different branch of the code. So I guess you're right.

redash/tasks/queries.py
```diff
@@ -182,20 +183,28 @@ def refresh_queries():
     with statsd_client.timer('manager.outdated_queries_lookup'):
         for query in models.Query.outdated_queries():
             if settings.FEATURE_DISABLE_REFRESH_QUERIES:
-                logging.info("Disabled refresh queries.")
+                logging.debug("Disabled refresh queries.")
```
arikfr: This shouldn't be a debug message, because it low frequency and important.

```suggestion
                logging.info("Disabled refresh queries.")
```

arikfr:
:+1:

PR #4055, opened 2019-08-11
Add more flake8 tests and fail build if any test fails

cclauss:
Run all flake8 E9xx + F63x + F7xx + F82x tests.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other: tests

### DIFF
--- a/redash/query_runner/couchbase.py
+++ b/redash/query_runner/couchbase.py
@@ -2,9 +2,11 @@ import datetime
 import logging
 
 from dateutil.parser import parse
+from six import text_type
 
 from redash.query_runner import *
 from redash.utils import JSONEncoder, json_dumps, json_loads, parse_human_time
+from redash.utils.compat import long
 import json
 
 logger = logging.getLogger(__name__)
@@ -17,7 +19,7 @@ except ImportError as e:
 
 TYPES_MAP = {
     str: TYPE_STRING,
-    unicode: TYPE_STRING,
+    text_type: TYPE_STRING,
     int: TYPE_INTEGER,
     long: TYPE_INTEGER,
     float: TYPE_FLOAT,
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -4,9 +4,11 @@ import urllib
 
 import requests
 from requests.auth import HTTPBasicAuth
+from six import string_types, text_type
 
 from redash.query_runner import *
 from redash.utils import json_dumps, json_loads
+from redash.utils.compat import long
 
 try:
     import http.client as http_client
@@ -35,7 +37,7 @@ ELASTICSEARCH_BUILTIN_FIELDS_MAPPING = {
 
 PYTHON_TYPES_MAPPING = {
     str: TYPE_STRING,
-    unicode: TYPE_STRING,
+    text_type: TYPE_STRING,
     bool: TYPE_BOOLEAN,
     int: TYPE_INTEGER,
     long: TYPE_INTEGER,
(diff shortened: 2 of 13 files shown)


redash/query_runner/json_ds.py
```diff
@@ -5,10 +5,15 @@
...
 from funcy import compact, project
+from six import text_type
 from redash.utils import json_dumps
 from redash.query_runner import (BaseHTTPQueryRunner, register,
                                  TYPE_BOOLEAN, TYPE_DATETIME, TYPE_FLOAT,
                                  TYPE_INTEGER, TYPE_STRING)
+try:
+    long
+except NameError:
+    long = int
```
jezdez: I would suggest to add a `redash/utils/compat.py` file for this stuff.

cclauss: Given that the only place that __long__ appears is in types_maps in
* redash/query_runner/couchbase.py
* redash/query_runner/elasticsearch.py
* redash/query_runner/json_ds.py
* redash/query_runner/mongodb.py

Perhaps it would be a better choice to create a single types_map that these query runners can import.

cclauss: Given that __couchbase.py__ is conflicted about what datatype a __datetime.datetime__ is, I just followed @jezdez recommended solution above.

jezdez: Yeah, let's not overthink this, this will be code to be removed when Redash moves to Python 3 after all.

redash/query_runner/elasticsearch.py
```diff
@@ -35,7 +37,7 @@
 
 PYTHON_TYPES_MAPPING = {
     str: TYPE_STRING,
-    unicode: TYPE_STRING,
+    text_type: TYPE_STRING,
```
jezdez: BTW, we should leave a note to remove this duplicate-on-Python-3 thing when we switch to Python 3.

redash/query_runner/json_ds.py
```diff
@@ -25,19 +27,19 @@ def parse_query(query):
...
     hostname = urlparse(url).hostname
     ip_address = socket.gethostbyname(hostname)
-    return ipaddress.ip_address(unicode(ip_address)).is_private
+    return ipaddress.ip_address(text_type(ip_address)).is_private
 
 
 TYPES_MAP = {
     str: TYPE_STRING,
-    unicode: TYPE_STRING,
+    text_type: TYPE_STRING,
```
jezdez: Same with the duplicate here.

redash/query_runner/elasticsearch.py
cclauss: This is super simple...  Just do a global find on __six__.  I have done this now on many projects and it is quite simple to do even across complex repos.  In fact, this is exactly why __six__ did not just reuse the names __unicode__ and __basestring__ on Python 3 but instead used different symbols that are easy to search on.  When __six__ is removed from the dependencies list these just pop right out on the automated testing.

jezdez: Makes sense, you're right. I've done it as well, and just tried to optimize for our future selves. Moving on.. :)

arikfr:
Thanks!

One step closer to Python 3 :-)

redash/query_runner/couchbase.py
```diff
@@ -17,7 +19,7 @@
 
 TYPES_MAP = {
     str: TYPE_STRING,
-    unicode: TYPE_STRING,
+    text_type: TYPE_STRING,
```
jezdez: @arikfr @cclauss I think this and other cases of the `TYPES_MAP` is incorrect, since we're now on Python 3 and `str` is Unicode, while `text_type` is as well. While we were on Python 2 that wasn't the case. I'll send a PR.

cclauss: https://pypi.org/project/pyupgrade is pretty useful for finding and fixing these things.

jezdez: https://github.com/getredash/redash/pull/4533

PR #4060, opened 2019-08-11
Cypress: Fix cy.clock not freezing time

gabrieldutra:

- [x] Other

## Description
With [this](https://dashboard.cypress.io/#/projects/924cka/runs/1042/failures) I noticed that time wasn't being frozen after using `cy.clock` (it was set, but some second differences were happening). In order to properly set that it must come before `cy.visit`, but this was causing page loading issues. Solved those by using `cy.clock(timestamp, ['Date'])` so the `Date` element is affected and other Timer functions are not.

Note: Could remove all workarounds after that :slightly_smiling_face:

--
--

### DIFF
--- a/client/cypress/integration/query/parameter_spec.js
+++ b/client/cypress/integration/query/parameter_spec.js
@@ -188,8 +188,7 @@ describe('Parameter', () => {
 
       cy.get('.ant-calendar-date-panel')
         .contains('.ant-calendar-date', date)
-        .click()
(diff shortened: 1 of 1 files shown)


gabrieldutra:
> How come there's no need for workaround? I thought it was a render/display issue.

I initially thought that as well, I was surprised that Cypress pointed me to remove the workarounds. My guess is that `cy.clock` was affecting other internal functions (such as Timeout ones).

PR #4061, opened 2019-08-12
Switch to mysqlclient from Python-MySQL

arikfr:

- [x] Other

## Description

* Newer versions of Debian use MariaDB clients by default, which aren't compatible with Python-MySQL (it's unclear to me how the CI build passes - maybe it uses a cached redash/base:debian image?).
* Python-MySQL isn't compatible with Python 3 and unmaintained, so needs to be replaced anyway.
* As older version of `memsql` were using Python-MySQL I had to upgrade it as well. This version of memsql might not support Python 2, but as memsql isn't very popular in Redash, it seemed like a reasonable trade off.

### DIFF
--- a/requirements_all_ds.txt
+++ b/requirements_all_ds.txt
@@ -2,7 +2,7 @@ google-api-python-client==1.5.1
 gspread==3.1.0
 impyla==0.10.0
 influxdb==2.7.1
-MySQL-python==1.2.5
+mysqlclient==1.4.4
 oauth2client==3.0.0
 pyhive==0.5.1
 pymongo[tls,srv]==3.6.1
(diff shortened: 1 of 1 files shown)


arikfr:
@Jakdaw this doesn't affect your work in #3003, right?

jezdez:
@arikfr Yeah, we've also noticed tests passing, but it only fails when we tried building the Docker image to do a deploy. I assume that something in the Circle config does that, maybe the fact that the tests are run via docker-compose and the Docker images are built by the docker CLI? Have you not encountered this when deploying v8beta to the SASS?

Jakdaw:
Can confirm no impact to the #3003  stuff from this.

Issue #4039, opened 2019-08-05
CSP error when clicking on "Create your first Dashboard"


jezdez:
### Issue Summary

When clicking on the "Create your first Dashboard" link after setup I get a CSP error in the browser console.

### Steps to Reproduce

1. Setup Redash following the Docker dev docs.
2. Setup a data source (e.g. the postgres container)
3. Add a query
4. Try to click on the "Create your first Dashboard"

I believe the change in https://github.com/getredash/redash/commit/9480d89e4c3d87e092e64e2d6c634413e42cf0a0#diff-90812e4fcbcd1df5f887a43bda5953f5R125 resulted in the code that is run when clicking on the link to be considered inline, instead of the separate function before.

### Technical details:

* Redash Version: 8.0.0beta
* Browser/OS: Firefox/macOS 10.14
* How did you install Redash: Docker

arikfr:
What is the message you receive?

jezdez:
In Chrome (😬):

"Refused to run the JavaScript URL because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution."

<img width="1098" alt="Screen-Shot-2019-08-06-14-49-45 93" src="https://user-images.githubusercontent.com/1610/62541163-81c0b100-b859-11e9-8fb6-5b6e6343df9d.png">

jezdez:
I think this is about the difference between:

```diff
-            onClick={createDashboard}
+            onClick={() => CreateDashboardDialog.showModal()}
```

kravets-levko:
@jezdez I think it may happen because of `javascript:void(0)` in `Step` component :thinking:

jezdez:
Ah, that could be it?

kravets-levko:
I think so. If you can experiment with it - try to remove `javascript:void(0)` there (use `null` or something like that instead), and check if the issue is still present or fixed.

arikfr:
@kravets-levko is correct.  I will fix it.

PR #4017, opened 2019-07-30
Removed redash-newstyle.less

ranbena:
- [x] Refactor

## Description
1. Merged overrides
2. Moved component specific rules to component style files.
3. Deleted deprecated rules.
4. The rest moved to base.less.

Tested manually and Percy passes.

PR #4062, opened 2019-08-12
Prevent CSP violations by not having script URLs 

arikfr:

- [x] Bug Fix

## Description

Following the discovery in #4039 I searched the code for further locations of where we use this pattern and removed them all. I also restored the lint rule to prevent future cases.

We should probably use `<Button type="link" ...>` in the future instead of `<a>` tags in such cases, but didn't want to work around the styling issues so kept things as is.

### DIFF
--- a/client/app/assets/less/redash/redash-newstyle.less
+++ b/client/app/assets/less/redash/redash-newstyle.less
@@ -173,6 +173,7 @@ body {
 
   &:hover, &:focus {
     color: @yellow-darker;
+    cursor: pointer;
   }
 
   .fa-star {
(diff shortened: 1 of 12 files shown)


client/app/assets/less/redash/redash-newstyle.less
```diff
@@ -173,6 +173,7 @@ body {
 
   &:hover, &:focus {
     color: @yellow-darker;
+    cursor: pointer;
```
arikfr: @ranbena this adds some work for #4017... Sorry, but @gabrieldutra is right: we can YOLO it, but only after V8 branch is cut.

jezdez:
Thank you @arikfr!

PR #4064, opened 2019-08-12
Fix loading of periodic tasks and clean up extension loading.

jezdez:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

This does a few things:

- add tests for extension loading
- actually call the periodic tasks entry point callback instead of using the function to pass to Celery (regression!)
- refactor the extension and periodic task loading
- better handle assertions raised by extensions (e.g. when an extension tries to override an already registered view)
- attach exception traceback to error log during loading for improved debugging


jezdez:
@arikfr Hey, just wanted to say that this is something we really need to ship in v8 since the periodic task loading mechanism was simply broken last time we touched the code. I forgot to not only load the entrypoints but also call the them to get the schedule parameters to be passed to Celery.

jezdez:
Hm, the tests pass locally, it's weird. Must be a difference in environment since it calls pip to actually install the dummy package. Let me see if I can figure out a different way to ship the dummy package.

arikfr:
Did the local tests work inside a container?

jezdez:
@arikfr Oh yeah, totally. I've been running them locally in the regular Docker dev environment.

I just pushed a change to use `site.addsitedir` instead of calling pip, let's see if this fixes it.

jezdez:
Update: Yep, using `site.addsitedir` fixed it. I haven't a clue why calling pip didn't work on Circle.

jezdez:
For the record, I created the `egg-info` directory (which is what is created in the site-packages directory when running `pip install <packagename>`) by running:

- `make bash` to create a shell inside the container
- `cd tests/extensions`
- `pip install --user --editable redash-dummy/`
- `pip uninstall --yes redash-dummy`

tests/extensions/test_extensions.py
```diff
@@ -0,0 +1,42 @@
...
+logger = logging.getLogger(__name__)
+here = os.path.dirname(__file__)
+dummy_extension = "redash-dummy"
+
+
+class TestExtensions(BaseTestCase):
+    @classmethod
+    def setUpClass(cls):
+        dummy_path = os.path.join(here, dummy_extension)
+        site.addsitedir(dummy_path)
```
arikfr: Probably worth removing in the cleanup step?

jezdez: Yep, done!

arikfr:
:+1:

Issue #4065, opened 2019-08-13
[Qubole] Add support to process Quantum query type. 


SandeepBelagavi:
Qubole query runner now supports processing **presto** and **hive** queries. 
Quantum is a serverless interactive service that offers direct SQL access to user's data lake.
Quantum is needs to integrated with query runner.

For more info visit: 
https://docs.qubole.com/en/latest/user-guide/quantum/

arikfr:
Implemented in #4066.

PR #4066, opened 2019-08-13
[Qubole] - Adding support to process Quantum query types.

sandeepV2:
Quantum is a server-less interactive service that offers
direct SQL access to user's data lake. Changes are made
to accept `quantum` query type from user which makes
`Cluster Label` as optional.

issue: https://github.com/getredash/redash/issues/4065


sandeepV2:
Merging qubole_quantum to master

sandeepV2:
@arikfr Can you please review this.

redash/query_runner/qubole.py
```diff
@@ -11,7 +11,8 @@
 try:
     import qds_sdk
     from qds_sdk.qubole import Qubole as qbol
-    from qds_sdk.commands import Command, HiveCommand, PrestoCommand
+    from qds_sdk.commands import Command, HiveCommand
+    from qds_sdk.commands import SqlCommand, PrestoCommand
```
arikfr: Do we need to update the Qubole package version we have for SqlCommand?

redash/query_runner/qubole.py
```diff
@@ -67,8 +68,12 @@ def run_query(self, query, user):
                        api_url='%s/api' % self.configuration['endpoint'])
 
         try:
-            cls = PrestoCommand if(self.configuration['query_type'] == 'presto') else HiveCommand
-            cmd = cls.create(query=query, label=self.configuration['cluster'])
+            if self.configuration['query_type'] == 'quantum':
```
arikfr: This is an issue we had even before, but worth addressing here: `self.configuration['query_type']` might not have a value at all. The default value we mention is up to Query Runner implementation to enforce.

I would change this part of the code to something like:

```python
query_type = self.configuration.get('query_type', 'hive')

if query_type == 'quantum':
    cmd = SqlCommand.create(query=query)
elif query_type == 'hive':
   cmd = HiveCommand(query=query, label=self.configuration['cluster'])
elif query_type == 'presto':
   cmd = PrestoCommand(query=query, label=self.configuration['cluster'])
else:
   raise Exception("Invalid value for Query Type: {}. Should be one of: hive, presto, quantum.")
```

(this also adds handling of the case we get a bad value for query_type).

redash/query_runner/qubole.py
arikfr: I didn't notice we use `>=` there... All good then :+1:

redash/query_runner/qubole.py
sandeepV2: I see none of the fields in `self.configuration` are getting populated with input, except for secrete property (token).
@arikfr i am witnessing following error, on executing first query using registered data source.Any suggestions?
Code:
```
query_type = self.configuration.get('query_type')
 if query_type == 'quantum':
     cmd = SqlCommand.create(query=query)
 elif query_type == 'hive':
     cmd = HiveCommand.create(query=query, label=self.configuration.get('cluster'))
 elif query_type == 'presto':
     cmd = PrestoCommand.create(query=query, label=self.configuration.get('cluster'))
 else:
     raise Exception("query_type:%s\n end_point:%s\n cluster:%s\n token:%s\n"%(self.configuration.get('query_type'),self.configuration.get('end_point'), self.configuration.get('cluster'), self.configuration.get('token')))
```

Error:

>  Traceback (most recent call last):
worker_1    |   File "/app/redash/tasks/queries.py", line 340, in run
worker_1    |     data, error = query_runner.run_query(annotated_query, self.user)
worker_1    |   File "/app/redash/query_runner/qubole.py", line 82, in run_query
worker_1    |     self.configuration.get('end_point'), self.configuration.get('cluster'), self.configuration.get('token')))
worker_1    | Exception: query_type:None
worker_1    |  end_point:None
worker_1    |  cluster:None
worker_1    |  token:3b2febd544exxxxxxxxx>

arikfr: You didn't provide a default value to `query_type`, change the first line to:

```python
query_type = self.configuration.get('query_type', 'hive')
```

sandeepV2: Done, qubole.py module was missing class method for `type` and `name`. This lead to failure of picking right object name. Things worked for me after adding them.

sandeepV2:
I have updated `requirements_all_ds.txt` with recent release version of SDK (v1.13.0). Checks with failure (build: importlib-metadata>=0.18 and percy) does not seem to relate with my changes. 

@arikfr Can we merge it?

sandeepV2:
@arikfr can i get write access to re-run failed checks? and any hints to get over build issue?

Thanks in advance,
Sandeep

redash/query_runner/qubole.py
```diff
@@ -59,16 +68,26 @@ def annotate_query(cls):
...
     def run_query(self, query, user):
-        qbol.configure(api_token=self.configuration['token'],
-                       api_url='%s/api' % self.configuration['endpoint'])
+        qbol.configure(api_token=self.configuration.get('token'),
+                       api_url='%s/api' % self.configuration.get('endpoint'))
 
         try:
-            cls = PrestoCommand if(self.configuration['query_type'] == 'presto') else HiveCommand
-            cmd = cls.create(query=query, label=self.configuration['cluster'])
+            query_type = self.configuration.get('query_type')
```
arikfr:
```suggestion
            query_type = self.configuration.get('query_type', 'hive')
```

You need to provide a default value. Using `hive` here to be consistent with previous version and not break existing deployments (see previous comment).

redash/query_runner/qubole.py
```diff
@@ -24,6 +25,11 @@ def configuration_schema(cls):
         return {
             "type": "object",
             "properties": {
+                "query_type": {
+                    "type": "string",
+                    "title": "Query Type (quantum / presto / hive)",
+                    "default": "quantum"
```
arikfr:
```suggestion
                    "default": "hive"
```

To prevent existing deployments breaking we need to keep this the same as it was in previous versions :-(

If you think it makes sense, you can create another query runner (just a subclass of this one) which is only for Quantum and will use this setting hard coded.

sandeepV2:
> The `importlib-metadata` is something we fixed. Did you rebase with latest master?

Yes i did, It is up to date with master.
```
$git rebase master
Current branch qubole_quantum is up to date.
```

redash/query_runner/qubole.py
sandeepV2: Done. Made hive as default query type.

redash/query_runner/qubole.py
sandeepV2: Done

arikfr:
Did you update your local `master` branch with the remote one? Otherwise it's probably the same master you branched out from. Looking at your tree I can see that `requirements_bundle.txt` is not the same as the current master, which explains this issue.

While it's probably safe to merge regardless, it's better to be on the safe side.

sandeepV2:
> Did you update your local `master` branch with the remote one? Otherwise it's probably the same master you branched out from. Looking at your tree I can see that `requirements_bundle.txt` is not the same as the current master, which explains this issue.
> 
> While it's probably safe to merge regardless, it's better to be on the safe side.

Done ! When can we expect it to be merged to master?

arikfr:
Merged. Thanks :)

PR #4067, opened 2019-08-13
Fork keep visualizations order

Jakdaw:
- [X] Bug Fix

## Description
At present when a query is forked the new query might not have the visualisations in the same order as in the original query. There's no way to reorder them in the UI, so let's make sure we maintain the same order.


rauchy:
LGTM. Wondering in what scenarios are forked visualizations ordered differently though?

kravets-levko:
> Wondering in what scenarios are forked visualizations ordered differently though?

Generally, if you omit `ORDER BY` clause from SQL query, order of results is undefined (DB engine can return records in any order). Usually (but not a must) DB will return records ordered by primary key **or** any other key used to process the query.

arikfr:
Thanks!

You might want to associate the email address in your Git commit with your GitHub account so this is probably represented in the contributors tab here.

PR #4032, opened 2019-08-13
Add parameter dialog doesn't work when query has selected text

rauchy:
### Issue Summary
Adding a parameter through the Parameter settings dialog won't work when you have a selected text in the query (parameter is added without any settings): 

add-param

This is definitively not critical, but I thought it was worth to report in any case.

### Steps to Reproduce

1. Open a query or create one
2. Select part of the text
3. Add a Parameter using `Ctrl/Cmd+P`
4. Notice that the parameter will be created without any settings

### Technical details:

* Redash Version: Latest
* Browser/OS: --
* How did you install Redash:

### DIFF
--- a/client/app/pages/queries/source-view.js
+++ b/client/app/pages/queries/source-view.js
@@ -1,4 +1,4 @@
-import { map, defer } from 'lodash';
+import { map, debounce, defer } from 'lodash';
 import template from './query.html';
 import EditParameterSettingsDialog from '@/components/EditParameterSettingsDialog';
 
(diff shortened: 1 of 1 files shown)


client/app/pages/queries/source-view.js
```diff
@@ -52,7 +52,9 @@ function QuerySourceCtrl(
 
   $scope.canForkQuery = () => currentUser.hasPermission('edit_query') && !$scope.dataSource.view_only;
 
-  $scope.updateQuery = newQueryText => defer(() => $scope.$apply(() => { $scope.query.query = newQueryText; }));
+  $scope.updateQuery = debounce(
+    newQueryText => defer(() => $scope.$apply(() => { $scope.query.query = newQueryText; })),
```
kravets-levko: Not a blocker, but - when you use `debounce` - `defer` is no longer needed (`debounce` includes its behavior).

rauchy: That makes sense :)

gabrieldutra:
I can't approve the PR :sweat_smile: (I guess it's because I initially created the issue and now the issue is a PR).

In any case, LGTM! :)

PR #1009, opened 2019-08-15
Dashboard URL does not show new name when dashboard name is updated

rauchy:
Before dashboard title change:
before_title_change

After dashboard title change: 
after_title_change


redash/handlers/dashboards.py
```diff
@@ -146,7 +146,14 @@ def get(self, dashboard_slug=None):
         :>json string widget.created_at: ISO format timestamp for widget creation
         :>json string widget.updated_at: ISO format timestamp for last widget modification
         """
-        dashboard = get_object_or_404(models.Dashboard.get_by_slug_and_org, dashboard_slug, self.current_org)
+        if dashboard_slug.split('-')[0].isdigit():
+            identifier = int(dashboard_slug.split('-')[0])
+            fn = models.Dashboard.get_by_id_and_org
```
arikfr: We have 185 dashboards in our database that match this pattern already ("28 Days Growth Metrics"), so old URLs will stop working. We need a fallback here.

Some ideas:

* Try with `get_by_slug_and_org` if `get_by_id_and_org` fails.
* Test if the slug matches the dashboards slug.

client/app/services/dashboard.js
```diff
@@ -4,6 +4,15 @@ import { Widget } from './widget';
...
 export let Dashboard = null; // eslint-disable-line import/no-mutable-exports
 
+export const urlForDashboard = ({ id, name }) => `dashboard/${id}-${name.toString()
+  .trim()
+  .toLowerCase()
+  .replace(/\s+/g, '-')
+  .replace(/[^\w-]+/g, '')
+  .replace(/--+/g, '-')
+  .replace(/^-+/, '')
+  .replace(/-+$/, '')}`;
```
arikfr: We should still create the slug in the backend, it will just update when you update the dashboard name.

rauchy: Yeah, but that might break existing links, if anyone kept them.

My intention was to have existing slugs continue to work, while new links would be 123-pretty-much-the-name-of-the-dashboard-here-but-could-be-anything-youd-like-cause-we-only-care-about-the-id-from-now.

arikfr: It doesn't have to:

1. We will keep the current `slug` value in the database as is, so we can do lookups for existing dashboards.
2. We will no longer update it, but compute it when returning the API response based on the current name.

redash/handlers/dashboards.py
rauchy: Preferred and implemented the first, but then realized that it breaks when you have a dashboard called "28 Days Growth Metrics" (available through `/dashboards/28-days-growth-metrics`) and then you create another dashboard called "I like trains" which gets assigned with id 28 (available through `/dashboards/28-i-like-trains`).

In that case, `get_by_id_and_org` wouldn't fail, and your previous links would break.

arikfr: Why?

When you request `28-days...` it will work with `get_by_slug_and_org`, and when you request `28-i-like...` it will fail `get_by_slug_and_org` and then try `get_by_id_and_org` which will work.

rauchy: Hehe, that's the opposite direction of what you suggested in the first option :)
Yeah, that might work, but it would add an extra query for _most_ dashboard loads. I went ahead and implemented the second option which seems to work fine.

rauchy: Rethinking it, there's no possible way I can think that we can remain agnostic to the slug if we don't do `get_by_slug_and_org` and fallback to `get_by_id_and_org` (i.e. if we don't go down this path, renaming dashboards would always break existing links). aba6e3b

client/app/services/dashboard.js
rauchy: [name_as_slug](https://github.com/getredash/redash/pull/1009/files#diff-8897279477368a0297f53278e8fdcdd5R871) 😬😬😬

redash/handlers/dashboards.py
arikfr: > Hehe, that's the opposite direction of what you suggested in the first option :)

:flushed: when I first read your reply I thought that's what happened, *read my suggestion again* but still parsed it in reverse order. :man_facepalming: 

> Yeah, that might work, but it would add an extra query for _most_ dashboard loads. I went ahead and implemented the second option which seems to work fine.

:thinking: how about the following:

Currently the dashboard URLs are `/dashboard/<slug>`. But they should actually be `/dashboards/<id>-<whatever>`. So what if as part of this change we will change the URLs and keep the old ones for backward compatibility. 

This way if we get a request for `/dashboard/...` we know it's an old URL and we should pass a slug but if it's `/dashboards/...` we use the new logic?

rauchy:
> 1. You should probably move this into redash.utils.slugify.
> 2. What does it do differently?
> 3. How does it handle unicode?
> 4. Tests?

I realized this after a few minutes 🤦‍♂ 

image

redash/handlers/dashboards.py
rauchy: 👍 a851402 😈

redash/serializers/__init__.py
```diff
@@ -245,7 +245,7 @@ def serialize_dashboard(obj, with_widgets=False, user=None, with_favorite_state=
 
     d = {
         "id": obj.id,
-        "slug": obj.slug,
+        "slug": obj.name_as_slug,
```
gabrieldutra, line 248: If we want to update the "slug" from the backend we also need to update other Dashboard resource urls to use the Id instead of the slug, e.g.: the Archive, favorite and unfavorite ones.

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -144,36 +144,48 @@ DashboardComponent.propTypes = {
...
   routeWithUserSession({
     path: "/dashboard/:dashboardSlug",
     render: pageProps => <DashboardPage {...pageProps} />,
   })
 );
+
+routes.register(
+  "Dashboards.ViewOrEditWithId",
+  routeWithUserSession({
+    path: "/dashboards/:dashboardId-(.*)?",
```
kravets-levko, line 188: If `- ` is optional - probably it should be inside of brackets as well

gabrieldutra: I thought so, I tried it in many ways, but this is the only way I got it to work as I wanted 😅

redash/handlers/dashboards.py
```diff
@@ -150,7 +150,7 @@ def get(self, dashboard_slug=None):
         :>json string widget.updated_at: ISO format timestamp for last widget modification
         """
         dashboard = get_object_or_404(
-            models.Dashboard.get_by_slug_and_org, dashboard_slug, self.current_org
+            models.Dashboard.get_by_id_or_slug_and_org, dashboard_slug, self.current_org
```
gabrieldutra, line 153: On a second thought, it's actually safer to have a different route for this 🤔, it can cause problems with existing `/dashboard/:slug` urls.

client/app/pages/dashboards/DashboardPage.jsx
kravets-levko: :thinking:

gabrieldutra:
I decided to keep the `-` as a separator here.

I believe this can receive an initial review, there's one thing left I'd like to discuss that is related to https://github.com/getredash/redash/pull/1009#discussion_r447927944 (also https://github.com/getredash/redash/pull/1009#discussion_r447978962) - about "how" we want to keep existing slug url working. The first for archive, favorite and unfavorite routes (currently use "slug") and the later for that "hack" to make them work on both. It works with priority on "id" > "slug", but still there is a "rare" context where the id is the same as the slug.

The "safe" way through I see would be to create a new Dashboard resource to operate with id's, while keeping the existing one to work with slugs for backward compatibility.

client/app/pages/dashboards/DashboardPage.jsx
arikfr: Worth adding a Cypress test that tries to load dashboard with `id`, `id-`, `id-slug`, `id-random string`.

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -144,36 +144,48 @@ DashboardComponent.propTypes = {
...
   onError: PropTypes.func,
 };
 
+// route kept for backward compatibility
 routes.register(
   "Dashboards.ViewOrEdit",
   routeWithUserSession({
     path: "/dashboard/:dashboardSlug",
     render: pageProps => <DashboardPage {...pageProps} />,
   })
```
arikfr, line 182: Let's make this route redirect to the new one, to make sure people stop using them. If it's not easy with the frontend router, we can consider doing this on the backend.

client/app/components/dashboards/CreateDashboardDialog.jsx
```diff
@@ -8,6 +8,7 @@ import { wrap as wrapDialog, DialogPropType } from "@/components/DialogWrapper";
 import navigateTo from "@/components/ApplicationArea/navigateTo";
 import recordEvent from "@/services/recordEvent";
 import { policy } from "@/services/policy";
+import { urlForDashboard } from "@/services/dashboard";
```
arikfr, line 11: How about we transform the response we receive and add a URL field to it, so we don't have to import (or remember to use) `urlForDashboard` everywhere?

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -69,9 +70,12 @@ function useDashboard(dashboardData) {
...
       }
       return Dashboard.save(data)
-        .then(updatedDashboard =>
-          setDashboard(currentDashboard => extend({}, currentDashboard, pick(updatedDashboard, keys(data))))
-        )
+        .then(updatedDashboard => {
+          setDashboard(currentDashboard => extend({}, currentDashboard, pick(updatedDashboard, keys(data))));
+          if (has(data, "name")) {
+            navigateTo(urlForDashboard(updatedDashboard), true);
+          }
```
arikfr, line 77: ~What's this about?~ This is to update the URL when the dashboard is renamed?

redash/serializers/__init__.py
```diff
@@ -246,6 +246,7 @@ def serialize_dashboard(obj, with_widgets=False, user=None, with_favorite_state=
     d = {
         "id": obj.id,
         "slug": obj.slug,
+        "name_as_slug": obj.name_as_slug,
```
arikfr, line 249: Why not return it as the slug value?

redash/models/__init__.py
```diff
@@ -1168,6 +1172,15 @@ def favorites(cls, user, base_query=None):
     def get_by_slug_and_org(cls, slug, org):
         return cls.query.filter(cls.slug == slug, cls.org == org).one()
 
+    @classmethod
+    def get_by_id_or_slug_and_org(cls, id_or_slug, org):
+        return (
+            cls.query.filter(
+                cast(cls.id, db.String) == id_or_slug, cls.org == org
+            ).one_or_none()
+            or cls.query.filter(cls.slug == id_or_slug, cls.org == org).one()
```
arikfr, line 1181: This is likely to mess up the index performance (because of the casting to string).

If we redirect from the old /dashboard/ URL to the new /dashboards/ ones, it means we only need to support a GET request by slug and we know when it's just a slug vs. id (let's add a flag to the GET request to note when it's a slug, maybe `?legacy`?). And we can convert all the other API calls to use IDs.

Wdyt?

redash/handlers/dashboards.py
arikfr: yeah, see comment.

client/app/pages/dashboards/DashboardPage.jsx
kravets-levko: When you move `-` to brackets it decides that `(-.*)?` is a pattern for `dashboardId`. This should work: `/dashboards/:dashboardId([^-]+)(-.*)?`

gabrieldutra: > Worth adding a Cypress test that tries to load dashboard with id, id-, id-slug, id-random string.

Yep, that's actually on my plans :)

gabrieldutra: > When you move - to brackets it decides that (-.*)? is a pattern for dashboardId.

Oh, I see, it's the same pattern for both indicating I want ath and deciding the attr pattern 🤦‍♂️... Thanks! :)

tests/handlers/test_dashboards.py
```diff
@@ -76,13 +86,13 @@ def test_get_dashboard_filters_unauthorized_widgets(self):
...
 
-        rv = self.make_request("get", "/api/dashboards/{0}".format(dashboard.slug))
+        rv = self.make_request("get", "/api/dashboards/{0}".format(dashboard.id))
         self.assertEqual(rv.status_code, 200)
         self.assertTrue(rv.json["widgets"][0]["restricted"])
         self.assertNotIn("restricted", rv.json["widgets"][1])
 
     def test_get_non_existing_dashboard(self):
-        rv = self.make_request("get", "/api/dashboards/not_existing")
+        rv = self.make_request("get", "/api/dashboards/-1")
```
gabrieldutra, line 95: This actually doesn't happen only to the dashboard, but anything that's `/api/resource/not_a_number` causes a sqlalchemy error. This PR didn't feel like a scope to fix it and at least now I don't have an idea on how to solve this now as it turned to be a dynamic thing.

redash/handlers/dashboards.py
```diff
@@ -149,9 +149,12 @@ def get(self, dashboard_slug=None):
         :>json string widget.created_at: ISO format timestamp for widget creation
         :>json string widget.updated_at: ISO format timestamp for last widget modification
         """
-        dashboard = get_object_or_404(
-            models.Dashboard.get_by_slug_and_org, dashboard_slug, self.current_org
+        fn = (
+            models.Dashboard.get_by_slug_and_org
+            if request.args.get("legacy") is not None
+            else models.Dashboard.get_by_id_and_org
         )
```
arikfr, line 156: Ternary operators aren't that readable in Python. Might be better to write this as:

```python
if request.args.get("legacy") is not None:
    fn = models.Dashboard.get_by_slug_and_org
else:
    fn = models.get_by_id_and_org
```

(`fn` will be available in the outer context)

redash/serializers/__init__.py
arikfr: Now that we migrated all API routes to use an id, we can drop this?

redash/handlers/dashboards.py
gabrieldutra: Done 👍

redash/serializers/__init__.py
gabrieldutra: Done 👍

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -145,35 +147,52 @@ DashboardComponent.propTypes = {
...
 
 DashboardPage.defaultProps = {
+  dashboardSlug: null,
+  dashboardId: null,
   onError: PropTypes.func,
 };
 
+// route kept for backward compatibility
 routes.register(
   "Dashboards.ViewOrEdit",
```
kravets-levko, line 185: Let's name a new route `"Dashboards.ViewOrEdit"` and rename this one to something like `"Dashboards.LegacyViewOrEdit"`

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -63,15 +64,17 @@ function useDashboard(dashboardData) {
...
         data = { ...data, version: dashboard.version };
       }
       return Dashboard.save(data)
-        .then(updatedDashboard =>
-          setDashboard(currentDashboard => extend({}, currentDashboard, pick(updatedDashboard, keys(data))))
-        )
+        .then(updatedDashboard => {
+          setDashboard(currentDashboard => extend({}, currentDashboard, pick(updatedDashboard, keys(data))));
+          if (has(data, "name")) {
+            location.setPath(url.parse(updatedDashboard.url).pathname, true);
```
kravets-levko, line 75: You can use `navigateTo(..., true)`: second argument is to replace URL state - router will not reload controller in this case

gabrieldutra: IIRC `navigateTo` replaces the url params, in this case I don't want them updated.

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -145,35 +147,52 @@ DashboardComponent.propTypes = {
...
   routeWithUserSession({
     path: "/dashboard/:dashboardSlug",
     render: pageProps => <DashboardPage {...pageProps} />,
   })
 );
+
+routes.register(
+  "Dashboards.ViewOrEditWithId",
```
gabrieldutra, line 193:
```suggestion
  "Dashboards.ViewOrEdit",
```

client/app/pages/dashboards/hooks/useDashboard.js
kravets-levko: Got it. Thanks for explanation!

arikfr:
👍

PR #4072, opened 2019-08-15
ParameterizedQuery: handle the case where a value is null

arikfr:

- [x] Bug Fix


arikfr:
@rauchy I guess it's sounds reasonable.

PR #4076, opened 2019-08-16
Rockset data source improvements

ari-e:

- [x] Feature
- [x] Bug Fix

## Description

This updates the Rockset data source in 2 important ways

1. Bug fix– The list of tables and schemas is now loaded for all workspaces whereas previously only `commons` workspace was loaded and others were ignored
2. Feature– The DESCRIBE for all the tables is now run in parallel instead of sequentially for much faster load times of the schema

N/A

N/A

### DIFF
--- a/redash/query_runner/rockset.py
+++ b/redash/query_runner/rockset.py
@@ -1,3 +1,5 @@
+from multiprocessing.pool import ThreadPool
+
 import requests
 from redash.query_runner import *
 from redash.utils import json_dumps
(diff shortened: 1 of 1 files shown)


redash/query_runner/rockset.py
```diff
@@ -76,12 +87,23 @@ def __init__(self, configuration):
             'api_server', "https://api.rs2.usw2.rockset.com"))
 
     def _get_tables(self, schema):
-        for col in self.api.list():
-            table_name = col['name']
-            describe = self.api.query('DESCRIBE "{}"'.format(table_name))
-            columns = list(set(map(lambda x: x['field'][0], describe['results'])))
-            schema[table_name] = {'name': table_name, 'columns': columns}
-        return schema.values()
+        pool = ThreadPool(processes=10)
```
arikfr: In some configurations (and this is going to become the default) we run Redash's API server with gevent's monkeypatching. I'm not sure how "healthy" it is to mix threads with gevent. 

How about we land this PR without the threads optimization and then revisit this in a follow up?

ari-e: Hey @arikfr, thanks for the feedback. Given that future direction, would it make sense to replace this multithreaded approach with [grequests](https://github.com/spyoungtech/grequests)? I was originally leaning in that direction actually but figured this would be a less invasive change.

ari-e:
@arikfr thoughts on ^?

ari-e:
@arikfr would appreciate some direction here, thanks.

redash/query_runner/rockset.py
arikfr: At the moment the default deployment doesn't use gevent and we don't have a timeline for the switch. So for now I suggest we don't use anything that depends on it.

ari-e:
Abandoning in favor of a simpler PR coming later today

PR #4078, opened 2019-08-17
Fix clickhouse password leak

vv-p:

- [x] Bug Fix

## Description
Clickhouse credentials appear in web-interface after unsuccessfully http connection while query running:

`Error running query: HTTPConnectionPool(host='10.1.2.3', port=8123): Max retries exceeded with url: /?password=123qweasd&user=redashreadonly&database=default (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7ff36739bd90>: Failed to establish a new connection: [Errno 111] Connection refused',))
`

So every user with access to redash web-interface can get it. In this PR I catch all requests exceptions and generate new, without private data.

### DIFF
--- a/redash/query_runner/clickhouse.py
+++ b/redash/query_runner/clickhouse.py
@@ -67,21 +67,25 @@ class ClickHouse(BaseSQLQueryRunner):
         return schema.values()
 
     def _send_query(self, data, stream=False):
-        r = requests.post(
-            self.configuration.get('url', "http://127.0.0.1:8123"),
-            data=data.encode("utf-8"),
-            stream=stream,
-            timeout=self.configuration.get('timeout', 30),
-            params={
-                'user': self.configuration.get('user', "default"),
-                'password':  self.configuration.get('password', ""),
-                'database': self.configuration['dbname']
-            }
-        )
-        if r.status_code != 200:
-            raise Exception(r.text)
-        # logging.warning(r.json())
-        return r.json()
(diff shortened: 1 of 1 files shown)


redash/query_runner/clickhouse.py
```diff
@@ -67,21 +67,25 @@ def _get_tables(self, schema):
...
+                    'password':  self.configuration.get('password', ""),
+                    'database': self.configuration['dbname']
+                }
+            )
+            if r.status_code != 200:
+                raise Exception(r.text)
+            # logging.warning(r.json())
+            return r.json()
+        except requests.RequestException:
+            raise Exception('Connection error to %s' % url)
```
arikfr: The problem now is that the error message is very light on details. How about something like the following (assuming `e` is the exception):

```
if e.response:
    details = "({}, Status Code: {})".format(e.__class__.__name__, e.response.status_code)
else:
    details = "({})".format(e.__class__.__name__)

message = "Connection error to: {} {}.".format(url, details)
```

vv-p: It sounds good, fixed, thank you. Now exceptions look like:
`Exception: Connection error to: http://127.0.0.1:8123 (ReadTimeout).`
or
`Exception: Connection error to: http://127.0.0.1:8123 (ConnectionError).`

arikfr:
Thanks!

PR #4083, opened 2019-08-19
Fix netlify preview redirect loop in Safari

ranbena:
- [x] Bug Fix

## Description
In Safari, accessing Netlify Preview logged-out results in redirect loop.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -10,7 +10,7 @@
 
 [[redirects]]
   from = "/login"
-  to = "http://preview-login.redashapp.com/login"
+  to = "https://preview-login.redashapp.com/login"
   status = 200
 
 [[redirects]]


netlify.toml
```diff
@@ -10,7 +10,7 @@
 
 [[redirects]]
   from = "/login"
-  to = "http://preview-login.redashapp.com/login"
+  to = "https://preview-login.redashapp.com/login"
```
arikfr: I don't remember exactly why (was it necessary or just convenient), but `preview-login` doesn't support https. 

But why do you think it's causing the issue you see? It redirects to a different page. Also, as @susodapop mentioned, calling `/login` directly works for him.

ranbena:
lol this won't work cause these urls aren't actually accessible with https.
@arikfr can you do sth about it?

netlify.toml
ranbena: > But why do you think it's causing the issue you see? It redirects to a different page.

I think it might be a browser security restriction that Safari carries out.

> Also, as @susodapop mentioned, calling /login directly works for him.

Calling directly is ok, but redirecting from https to http might be disallowed.

I dunno. I'll look into it a bit but I don't really care 😬

ranbena: @arikfr maybe the api/session retries simply don't let the redirect happen. Are the retries and redirect coded explicitly or are they abstracted in some login mechanism?

ranbena:
Dunno how to fix. Abandoning.

PR #4084, opened 2019-08-19
Convert unexpected parameter validation errors to InvalidParameterError

rauchy:
- [x] Refactor

## Description
A quick follow up to #4072


package-lock.json
```diff
@@ -1945,11 +1945,6 @@
       "integrity": "sha512-jp/uFnooOiO+L211eZOoSyzpOITMXx1rBITauYykG3BRYPu8h0UcxsPNB04RR5vo4Tyz3+ay17tR6JVf9qzYWg==",
```
arikfr: Probably worth not committing this here.

arikfr (commented):
Looks fine, but I wonder if we won't swallow some edge cases that we should handle differently this way? Maybe worth keeping existing error handling and expend it over time as we discover new cases? 

This way we can add logging for exceptions in the case of really unexpected ones (and properly handle them if needed).

Wdyt?

guidopetri:

guidopetri (approved):

guidopetri (approved):

justinclift:
This one kind of looks like it could have more test coverage added, to get the Codecov pieces to pass?

guidopetri:
Yeah, it was passing before I rebased, heh

guidopetri:
Alright, that's the best I'm getting this PR. :sweat_smile: there was a coverage change because _all_ error types were being converted to `InvalidParameterError`, when in reality there was a specific kind of error (`QueryDetachedFromDataSourceError`) that needed to be raised separately.

Fixing that, there's no indirect coverage changes, but the coverage is going down anyway because we are removing lines that were fully covered - i.e. the percentage of fully covered lines goes down because the absolute number of fully covered lines is going down.

justinclift:
Cool, good stuff @guidopetri. :grin:

guidopetri:
It took me like two hours to figure out the fix lol. Lesson learned: don't use bare exceptions

DEZwak:
Hey @guidopetri maybe you can help me with https://github.com/getredash/redash/issues/7710

On queries with dynamic date range parameters like (example using "Last 7 Days"):

`{"parameters": [{"name": "date_range", "type": "date-range", "title": "date_range", "value": "d_last_7_days", "global": false}], "apply_auto_limit": false}`

We get Invalid param errors `The following parameter values are incompatible with their definitions`. Trying to hack at your code a bit to add those `d_*` (where they are strings, not objects) definitions didnt really work, do you have any ideas?

PR #3804, opened 2019-08-20
Dashboard: when updating parameters, run only relevant queries

rauchy:
When dashboard-level parameters are updated, all widgets in that dashboard are updated, even if they are unaffected by the new parameter values.

For example - in this dashboard, the bottom widget is completely unrelated to `param1`, yet updating its value refreshes the widget.

dashboard

This PR changes this behavior by only refreshing the widgets that have at least parameter which maps to one of the dashboard-level parameter which were updated.

### DIFF
--- a/client/app/components/Parameters.jsx
+++ b/client/app/components/Parameters.jsx
@@ -111,8 +111,9 @@ export class Parameters extends React.Component {
   applyChanges = () => {
     const { onValuesChange, disableUrlUpdate } = this.props;
     this.setState(({ parameters }) => {
+      const parametersWithPendingValues = parameters.filter(p => p.hasPendingValue);
       forEach(parameters, p => p.applyPendingValue());
-      onValuesChange();
+      onValuesChange(parametersWithPendingValues);
       if (!disableUrlUpdate) {
         updateUrl(parameters);
       }
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -142,7 +142,7 @@ function DashboardCtrl(
       widget => Object.values(widget.getParameterMappings()).filter(
         ({ type }) => type === 'dashboard-level',
       ).some(
-        ({ mapTo }) => updatedParameters.map(p => p.name).includes(mapTo),
+        ({ mapTo }) => _.includes(updatedParameters.map(p => p.name), mapTo),
       ),
     ) : this.dashboard.widgets;
 
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -139,8 +139,10 @@ function DashboardCtrl(
 
   const collectFilters = (dashboard, forceRefresh, updatedParameters = []) => {
     const affectedWidgets = updatedParameters.length > 0 ? this.dashboard.widgets.filter(
-      widget => widget.getQuery() && widget.getQuery().getParametersDefs().some(
-        ({ name }) => updatedParameters.map(p => p.name).includes(name),
+      widget => Object.values(widget.getParameterMappings()).filter(
+        ({ type }) => type === 'dashboard-level',
+      ).some(
+        ({ mapTo }) => updatedParameters.map(p => p.name).includes(mapTo),
       ),
     ) : this.dashboard.widgets;
 
--- a/client/app/pages/dashboards/dashboard.js
+++ b/client/app/pages/dashboards/dashboard.js
@@ -137,10 +137,10 @@ function DashboardCtrl(
     this.extractGlobalParameters();
   });
 
-  const collectFilters = (dashboard, forceRefresh, pendingParameters = []) => {
-    const affectedWidgets = pendingParameters.length > 0 ? this.dashboard.widgets.filter(
+  const collectFilters = (dashboard, forceRefresh, updatedParameters = []) => {
+    const affectedWidgets = updatedParameters.length > 0 ? this.dashboard.widgets.filter(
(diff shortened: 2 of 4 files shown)


client/app/components/Parameters.jsx
```diff
@@ -111,8 +111,9 @@ export class Parameters extends React.Component {
   applyChanges = () => {
     const { onValuesChange, disableUrlUpdate } = this.props;
     this.setState(({ parameters }) => {
+      const parametersWithPendingValues = parameters.filter(p => p.hasPendingValue);
       forEach(parameters, p => p.applyPendingValue());
```
gabrieldutra: Now we can use `parametersWithPendingValues` here too :)

client/app/pages/dashboards/dashboard.js
```diff
@@ -137,8 +137,14 @@ function DashboardCtrl(
     this.extractGlobalParameters();
   });
 
-  const collectFilters = (dashboard, forceRefresh) => {
-    const queryResultPromises = _.compact(this.dashboard.widgets.map((widget) => {
+  const collectFilters = (dashboard, forceRefresh, pendingParameters = []) => {
```
gabrieldutra: In here the parameters have already been updated, so perhaps `updatedParameters` instead?

client/app/components/Parameters.jsx
rauchy: Not sure I understand. Are you referring to using `parametersWithPendingValues` in the following line? As in
```js
forEach(parametersWithPendingValues, p => p.applyPendingValue());
```

If that's what you mean, I deliberately avoided that in order to future-proof someone removing the `hasPendingValue` check in `applyPendingValue` without knowing of this external concern.

gabrieldutra: I wouldn't be that concerned about it (doesn't seem a likely thing or something that would cause a big issue here). But not a big deal in any case :)

client/app/pages/dashboards/dashboard.js
```diff
@@ -137,8 +137,14 @@ function DashboardCtrl(
...
   });
 
-  const collectFilters = (dashboard, forceRefresh) => {
-    const queryResultPromises = _.compact(this.dashboard.widgets.map((widget) => {
+  const collectFilters = (dashboard, forceRefresh, updatedParameters = []) => {
+    const affectedWidgets = updatedParameters.length > 0 ? this.dashboard.widgets.filter(
+      widget => widget.getQuery() && widget.getQuery().getParametersDefs().some(
+        ({ name }) => updatedParameters.map(p => p.name).includes(name),
+      ),
+    ) : this.dashboard.widgets;
```
ranbena: I'm uncertain `Array.includes` is supported in Edge 11. Better use lodash.

ranbena: Filtering by `name` won't do as they are unique only within a widget scope.
The same widget can show up twice in a dashboard - one with a dashboard-level param, the other with a widget-level param, and with this implementation, both will get refreshed instead of only one. (Same goes for different widgets that have the same param names).

Here's a demo https://deploy-preview-3804--redash-preview.netlify.com/dashboard/param-update-tester. Change the dashboard parameter, and all widgets refresh even though the middle one shouldn't, since it only has a widget-level param.

rauchy: Good catch. I switched it to use the [proper parameter mapping](https://github.com/getredash/redash/pull/3804/commits/283498aebee72397cefefa15ab3c6675b15e2f2d).

rauchy:
> your PR description is lacking - plz add a detailed problem->solution, a demo dashboard with STR so we can compare master to.

@ranbena hehe, I used `hub pull-request` so I guess the original issue was lacking that info. I can go ahead and update it, but it feels like you already created a demo dashboard :)

client/app/pages/dashboards/dashboard.js
ranbena: Works perfectly 👍 

I know you must have worked on making the filter logic readable but I think it's still a bit complex to comprehend. Here's my suggestion:

```js
let affectedWidgets = dashboard.widgets;
if (updatedParameters.length) {
  const updateParamNames = updatedParameters.map(p => p.name);
  affectedWidgets = dashboard.widgets.filter(
    widget => _.chain(widget.getParameterMappings())
      .filter({ type: 'dashboard-level' })
      .some(({ mapTo }) => _.includes(updateParamNames, mapTo))
      .value(),
  );
}
```

arikfr: Chaining prevents webpack from including only the functions we use from lodash :grimacing:

rauchy: @ranbena thanks for introducing me to lodash chain sequences!

I don't think it would be significant to lose tree shaking on lodash, but I'm still not entirely convinced that the change from 
```js
filter({ type }) => type === 'dashboard-level')
```
to
```js
filter({ type: 'dashboard-level' })
```
makes things simpler, especially given that some simpletons such as myself would possibly have to scratch their heads around `_.chain` and `value()`. The main disadvantage I see here is transforming this from an expression (`const affectedWidgets = stuff`) to a declarative block.

I personally find expressions a primary goal for achieving simplicity.

gabrieldutra: > Chaining prevents webpack from including only the functions we use from lodash 

As @rauchy mentioned it does makes things a lot more readable, I wonder if it's really that costly when considering Lodash :grimacing:

ranbena: Forget it! Let's land this already!

PR #4090, opened 2019-08-21
Fix: MySQL connections without SSL are failing

arikfr:

- [x] Bug Fix

## Description

I don't know if it's a change in `mysqlclient` or the new MariaDB C connector, but now when you pass a value to the `ssl` argument of `MySQLdb.connect` it will try to start an SSL connection (where before it was just ignored).

This PR changes the connect function to not pass `ssl` when not needed and some other cleanup.

### DIFF
--- a/redash/query_runner/mysql.py
+++ b/redash/query_runner/mysql.py
@@ -2,7 +2,7 @@ import logging
 import os
 import threading
 
-from redash.query_runner import *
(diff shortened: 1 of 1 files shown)


redash/query_runner/mysql.py
```diff
@@ -95,12 +104,26 @@ def name(cls):
...
+                      port=self.configuration.get('port', 3306),
+                      charset='utf8',
+                      use_unicode=True,
+                      connect_timeout=60)
+
+        ssl_options = self._get_ssl_parameters()
 
-        return True
+        if ssl_options:
+            params['ssl'] = ssl_options
```
arikfr: This is the actual fix...

Issue #4092, opened 2019-08-21
Replace Celery with RQ


rauchy:
[Celery](http://www.celeryproject.org) has been our task runner for a while, but we’ve been experiencing operational issues with it in our SaaS instance. While Celery is a respected and standard approach to perform asynchronous tasks, it has optimization benefits that simply aren’t important to us, and overall I feel that its complexity is greater than its offering, at least for our (mostly trivial) use cases.

[RQ](https://python-rq.org) has a lot in common with Celery, but it is way simpler, and could vastly reduce our project’s complexity.

In order to get a feel of RQ’s design concepts and operational shenanigans, we could start by migrating all the scheduled tasks from Celery to RQ. They can easily run side by side with the query execution tasks, which can still run in Celery for now.

jezdez:
FWIW I'm in favor of moving away from Celery. But I wanted to also add [Spinach](https://spinach.readthedocs.io/) as an option on the table since I think it has some interesting [design choices](https://spinach.readthedocs.io/en/stable/user/design.html) compared to RQ.

It comes with:
- [support](https://spinach.readthedocs.io/en/stable/user/integrations.html) for Flask (e.g. access to Flask app inside tasks), Sentry and task batching
- BSD-licensed
- built for [Redis](https://spinach.readthedocs.io/en/stable/user/design.html#only-two-brokers)
- native retries for [tasks](https://spinach.readthedocs.io/en/stable/user/tasks.html)
- has easy optional in-memory workers for tests
- is thread-safe
- [scheduled tasks](https://spinach.readthedocs.io/en/stable/user/design.html#jobs-scheduled-for-the-future) already already built-in
- [periodic tasks](https://spinach.readthedocs.io/en/stable/user/tasks.html#periodic-tasks) don't require a separate scheduler/beat process which could fail if the container dies
- the performance behavior in my experience running [Jazzband.co](https://jazzband.co) with it is very predictable since it simply uses standard threads to execute tasks
- extensible via Blinker based [signals](https://spinach.readthedocs.io/en/stable/user/signals.html) like Flask
- [worker setup](https://spinach.readthedocs.io/en/stable/user/design.html#no-worker-entrypoint) is  left to users
- supports namespaces for running separate Spinach instances in the same interpreter (e.g. queues)
- is Python 3-only 🥇 

[...]

PR #4093, opened 2019-08-21
Replace Celery with RQ (except for execute_query tasks)

rauchy:
- [x] Refactor

## Description
As a first probe towards migrating to RQ, I'm interested in trying to swap out Celery beat and all non-execute-query tasks and replace them with RQ.

Towards #4092


docker-compose.yml
```diff
@@ -19,6 +19,34 @@ services:
...
+        target: /app
+    depends_on:
+      - server
+    environment:
+      REDASH_REDIS_URL: "redis://redis:6379/0"
+  rq_worker:
+    build: .
+    command: dev_rq_worker
+    deploy:
+      replicas: 2
```
rauchy: Left this here for discussion - if one would want to allow multiple workers running simultaneously using `docker-compose`, one would have to run `docker-compose --compatibility up`

redash/cli/rq.py
```diff
@@ -0,0 +1,56 @@
...
+
+    scheduler.run()
+
+
+@manager.command()
+@argument('queues')
+def worker(queues='default'):
+    with Connection(redis_connection):
+        w = Worker(queues)
+        w.work()
```
jezdez: I should note that there is [flask-rq2](https://github.com/rq/Flask-RQ2) which provides a Flask friendly API for this.

jezdez: (I'm its author)

redash/__init__.py
```diff
@@ -45,5 +47,9 @@ def setup_logging():
 statsd_client = StatsClient(host=settings.STATSD_HOST, port=settings.STATSD_PORT, prefix=settings.STATSD_PREFIX)
 limiter = Limiter(key_func=get_ipaddr, storage_uri=settings.LIMITER_STORAGE)
 
+def enqueue(queue, func, *args, **kwargs):
```
jezdez: Can't we move this into `redash.queues` or something? Putting more into `redash/__init__.py` is kind of backwards in terms of maintainability regrading possible import-time-side-effects.

rauchy: Hehe, I'm guessing this is the downside of a draft PR - I created the `enqueue` function temporarily to reduce duplication and pushed. Planned on finding an appropriate place later :)

redash/cli/rq.py
```diff
@@ -0,0 +1,56 @@
...
+        scheduler.schedule(scheduled_time=datetime.utcnow(), func=func, **kwargs)
+
+    schedule(refresh_queries, interval=30)
+    schedule(empty_schedules, interval=timedelta(minutes=60))
+    schedule(refresh_schemas, interval=timedelta(minutes=settings.SCHEMAS_REFRESH_SCHEDULE))
+    schedule(sync_user_details, timeout=60, ttl=45, interval=timedelta(minutes=1))
+    schedule(send_aggregated_errors, interval=timedelta(minutes=settings.SEND_FAILURE_EMAIL_INTERVAL))
+    
+    for (func, kwargs) in settings.dynamic_settings.custom_tasks().iteritems():
+        schedule(func, **kwargs)
```
arikfr: If `Scheduler#schedule` can be called at any time, maybe we don't need `dynamic_settings` for custom scheduled tasks. We will just import `scheduler` when needed and call `schedule`.

redash/cli/rq.py
```diff
@@ -0,0 +1,56 @@
...
+from rq_scheduler import Scheduler
+
+manager = AppGroup(help="RQ management commands.")
+
+
+@manager.command()
+def scheduler():
+    scheduler = Scheduler(connection=redis_connection,
+                          queue_name="periodic",
+                          interval=5)
```
arikfr: Let's move this (along with the task definitions) to `redash.worker` and have the scheduler object defined at the module level.

rauchy: Eventually went for `redash.schedule` to (a) avoid bloat (b) avoid circular dependencies (c) worker/scheduler is a mesh-up which originated in Celery and is represented by separate concepts in RQ.

redash/cli/rq.py
rauchy: That's a good idea! 5e71aa7

redash/schedule.py
```diff
@@ -0,0 +1,49 @@
...
+                          cleanup_query_results,
+                          version_check, send_aggregated_errors)
+
+rq_scheduler = Scheduler(connection=redis_connection,
+                         queue_name="periodic",
+                         interval=5)
+
+
+def schedule(func, **kwargs):
+    previously_scheduled_jobs = filter(lambda job: job.func == func, rq_scheduler.get_jobs())
```
rauchy: @jezdez is it wise to do this?

rauchy:
@arikfr added some monitoring in 06576f8.

image

image

This PR is starting to grow larger than what I initially hoped, so I think this is a good time to review it.

redash/cli/rq.py
```diff
@@ -22,3 +25,17 @@ def worker(queues='default'):
     with Connection(redis_connection):
         w = Worker(queues)
         w.work()
+
+
+@manager.command()
+def healthcheck():
```
rauchy: @arikfr thoughts on this? This command will exit with 0 if all workers running on this machine have had a heartbeat in the last 60 seconds.

bin/docker-entrypoint
```diff
@@ -52,9 +65,12 @@ help() {
...
-  echo "worker -- start Celery worker"
-  echo "scheduler -- start Celery worker with a beat (scheduler) process"
-  echo "dev_worker  -- start Celery worker with a beat (scheduler) process which picks up code changes and reloads"
+  echo "celery_worker -- start Celery worker"
+  echo "dev_celery_worker -- start Celery worker process which picks up code changes and reloads"
+  echo "worker -- start a single RQ worker"
+  echo "dev_worker -- start a single RQ worker with code reloading"
+  echo "scheduler -- start an rq-scheduler instance"
+  echo "dev_scheduler -- start an rq-scheduler instance with code reloading"
   echo "celery_healthcheck -- runs a Celery healthcheck. Useful for Docker's HEALTHCHECK mechanism."
```
arikfr: Need to update this...

rauchy: Does it make sense to have an entrypoint for `rq_healthcheck`? Running the command will start a new container, which will obviously not have any workers running within it. The CLI command needs to run within an existing container with a running (or at least registered) worker process.

arikfr: The HEALTHCHECK Docker command uses this entrypoint. It runs it in the context of the same container.

redash/cli/rq.py
arikfr: I wonder if this logic makes sense or whether we should have an option to specify a single worker :thinking:

rauchy: Assuming you would typically run a single worker per container - it makes sense, because a non-zero exit code would probably indicate that you need to restart the container.

Given multiple workers in a container, you would probably want to return the PIDs of the misbehaving ones so you could kill and restart them, but I'm not sure how well that plugs into things like ECS's (or other services) healthchecks.

bin/docker-entrypoint
rauchy: 👍 76040ae

redash/monitor.py
```diff
@@ -134,3 +138,43 @@ def celery_tasks():
         tasks += get_waiting_in_queue(queue_name)
 
     return tasks
+
+
+def fetch_jobs(queue, job_ids):
+        return [{
```
jezdez: The indentation seems off here.

redash/schedule.py
jezdez: Hmm, not sure if you can compare the job functions, but they should be singletons, right?

redash/worker.py
```diff
@@ -1,68 +1,56 @@
...
-        'schedule': crontab(minute=randint(0, 59), hour=randint(0, 23))
-    }
-
-if settings.QUERY_RESULTS_CLEANUP_ENABLED:
-    celery_schedule['cleanup_query_results'] = {
-        'task': 'redash.tasks.cleanup_query_results',
-        'schedule': timedelta(minutes=5)
-    }
-
-celery_schedule.update(settings.dynamic_settings.custom_tasks())
```
jezdez: Since we're removing so much of the Celery stuff, I think it makes sense to also move the call to 
`celery.on_after_configure.connect` which adds the periodic tasks that were installed by Redash extensions to using RQ. WDYT  @arikfr and @rauchy?

redash/schedule.py
rauchy: I think unbound methods are equals, and I've debugged this and watched previously scheduled jobs get cancelled successfully. I'm more curious if you think that the concept of cancelling the previously scheduled periodic jobs is idiomatic here.

redash/worker.py
rauchy: Sounds right. Can you take it?

jezdez: Sure can tomorrow, if not I can deliver post-merge.

jezdez: Hm, I forgot that there is a thing about scheduled jobs not being deduplicated automatically, see my [ticket from 3,5 years ago](https://github.com/rq/rq-scheduler/issues/121), have we covered that for other places?

I think it may make sense to use `scheduler.delete()` rather than `scheduler.cancel()` in our schedule function, @rauchy. See https://stackoverflow.com/questions/17537771/where-do-i-register-an-rq-scheduler-job-in-a-django-app for the idea. Alternatively use the job id to explicitly checking if the job has already been scheduled as commented on in https://github.com/rq/rq-scheduler/issues/121#issuecomment-489357064.

rauchy: @jezdez you're right, `job.delete()` (I'm assuming you didn't mean `scheduler.delete()`) makes more sense than `scheduler.cancel()`. This is fixed in 9c73f0a. Thanks!

client/cypress/cypress.js
```diff
@@ -28,6 +28,7 @@ function startServer() {
 }
 
 function stopServer() {
+  execSync('docker-compose -p cypress logs', { stdio: 'inherit' });
```
arikfr: This is still needed?

docker-compose.yml
arikfr: Isn't `deploy` only for Docker Swarm deployments? Have you considered using `scale` option?

docker-compose.yml
```diff
@@ -19,9 +19,38 @@ services:
...
     command: dev_worker
+    deploy:
+      replicas: 2
+    volumes:
+      - type: bind
+        source: .
+        target: /app
+    depends_on:
+      - server
+    tty: true
```
arikfr: ?

redash/__init__.py
```diff
@@ -1,10 +1,12 @@
+from __future__ import absolute_import
 import logging
 import os
 import sys
 import urllib
 import urlparse
 
 import redis
+from rq import Queue
```
arikfr: unused?

redash/authentication/account.py
```diff
@@ -1,7 +1,7 @@
 import logging
 from flask import render_template
 
-from redash import settings
+from redash import settings, redis_connection
```
arikfr: unused?

redash/settings/dynamic_settings.py
```diff
@@ -10,11 +10,10 @@ def query_time_limit(is_scheduled, user_id, org_id):
...
 
-# Provide any custom tasks you'd like to run periodically
-def custom_tasks():
-    return {
-        # 'key': {
-        #     'task': 'redash.task.example',
-        #     'schedule': timedelta(minutes=5)
-        # }
-    }
+"""You can provide any custom jobs you'd like to run periodically here. For example:
```
arikfr: Probably better to move it into `redash.scheduler` as a docstring for the `schedule` function?

docker-compose.yml
rauchy: If I understand correctly, the `scale` option is deprecated. One could use the `--scale` CLI option, and that would have the same consequence as using `--compatibility`. I just saw the `deploy.replicas` option as a better source of documentation for anyone who would want to scale through docker-compose.

docker-compose.yml
rauchy: 🌈 Colorful output from `docker-compose` 🌈

redash/settings/dynamic_settings.py
rauchy: Not sure about this. `dynamic_settings.py` is sort of a template for users to follow when they want dynamic settings. We're sort of expecting people to create a copy of this module, modify it and point their `settings` to the modified module. Moving this piece of documentation to `redash.schedule` would make it hard find.

docker-compose.yml
arikfr: You're right. They deprecated it in v3 :unamused: I fail to see the logic behind this move, specially when they recommend _not_ using `--compatibility` in production.

Considering we have no intention of using Swarm -- maybe we should switch to v2?

redash/settings/dynamic_settings.py
arikfr: The chances of someone finding it here or elsewhere are very low and therefore similar. When we would like people to find this (when it's stable enough), we will just add it to the Knowledge Base and document properly.

docker-compose.yml
rauchy: We have at least one thing we need from v3.2 - bind volumes (for proper code reloading), so v2 is not an option. I guess it'll be best to recommend the `--scale` option outside this file. Where do you think would be the best place to do so?

arikfr: Do you remember what's the difference between how we define the volume for `server` and `worker`? I know you added it when working on auto restart, but from looking at the docs I fail to understand the difference.

If it is required, we can consider downgrading the production compose file to v2 as there the scale configuration is more important.

redash/settings/dynamic_settings.py
rauchy: c27ec3d8

docker-compose.yml
rauchy: `server` is using the default `volume` volume type, because Flask's auto reloading works fine with it. For `worker` (or anything that we use `watchmedo` for) we need a `bind` volume type for FSEvents to play nicely with macOS.

We could definitely use v2 for production compose files though.

arikfr: Ok, as this is all interim -- let's move on and revisit before migrating everything to RQ.

rauchy: OK, for now I've removed the `deploy.replicas` key.

redash/settings/dynamic_settings.py
rauchy: Unfortunately I had to revert this in 7a59113 (but with a double-dispatch approach) to allow periodic jobs to be scheduled when app is ready.

arikfr:
@NicolasLM We're planning on merging this soon into master. Anything we can do to make your life easier post the merge?

rauchy:
> If we can't just directly use the function, why not return that definitions dict in this function and call schedule in redash.schedule?

@arikfr yeah that makes sense. I've used the opportunity to schedule standard jobs in the same way to make both invocations appear the same in 4fb49c0.

arikfr:
When we merge this we should post an update on Development mentioning that master is in a hybrid state at the moment and not recommended for general consumption.

redash/schedule.py
```diff
@@ -0,0 +1,56 @@
...
+
+from rq_scheduler import Scheduler
+
+from redash import settings, redis_connection
+from redash.tasks import (sync_user_details, refresh_queries,
+                          empty_schedules, refresh_schemas,
+                          cleanup_query_results,
+                          version_check, send_aggregated_errors)
+
+rq_scheduler = Scheduler(connection=redis_connection,
```
arikfr: In Python 3 we use `decode_responses=True`, while `rq-scheduler` expects it to be `False` and uses `decode` on values it receives:

https://github.com/rq/rq-scheduler/blob/061488e79b82af215f7f98fa1554a0aa84dfd62f/rq_scheduler/scheduler.py#L321

When porting this to Python 3, we should probably create a Redis connection for it without the `decode_responses` change :thinking: We can create a "factory" function for creating the Redis connection and move the logic of mangling the Redis URL there (and make it optional).

Another option to consider is to revert to not decoding responses and do it where needed. It just felt redundant, so I was hoping to avoid it.

@rauchy @NicolasLM

PR #4101, opened 2019-08-28
Widget table scroll-x visible

ranbena:
- [x] Feature

## Description
In wide table visualizations, the horizontal scrollbar doesn't appear unless scrolled to bottom and is therefore inconvenient for those using a mouse pointer. 

This PR introduces two changes:
1. Horizontal scrollbar is always accessible.
2. Pagination as well. (it's possible to keep it at the bottom, but that would cause a double scrollbar)

## About the implementation

There are two reason I'm not a fan of this implementation.
1. It relies on `height: 100%` which means all of the elements between `.table-visualization-container` and `.ant-spin-container` must all inherit this property or else the chain is broken.
2. It relies on ant elements `.ant-spin-container` and `.ant-table` which might change over time and the feature would break.

So I made tests with Percy shots in which any breakage in the two conditions above would either push the pagination out of sight (cause height won't be 100%) or h-scrollbar won't appear 😉

Screen Shot 2019-08-29 at 21 59 23

### DIFF
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -208,13 +208,6 @@ edit-in-place p.editable:hover {
   }
 }
 
-.visualization-renderer {
-  .pagination,
-  .ant-pagination {
-    margin-top: 10px;
-  }
-}
-
 .embed__vis {
   display: flex;
   flex-flow: column;
--- a/client/app/visualizations/VisualizationRenderer.jsx
+++ b/client/app/visualizations/VisualizationRenderer.jsx
@@ -60,14 +60,13 @@ export function VisualizationRenderer(props) {
   return (
     <React.Fragment>
       {showFilters && <Filters filters={filters} onChange={setFilters} />}
-      <div>
-        <Renderer
-          key={`visualization${visualization.id}`}
-          options={options}
-          data={filteredData}
-          visualizationName={visualization.name}
-        />
-      </div>
+      <Renderer
+        key={`visualization${visualization.id}`}
+        options={options}
+        data={filteredData}
+        visualizationName={visualization.name}
+        context={props.context}
+      />
     </React.Fragment>
   );
 }
--- a/client/app/visualizations/table/renderer.less
+++ b/client/app/visualizations/table/renderer.less
@@ -8,7 +8,7 @@
     margin-bottom: 0;
   }
 
-  .ant-table-body {
+  .ant-table {
     overflow-x: auto;
   }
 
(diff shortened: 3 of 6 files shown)


ranbena:
@arikfr give this another try.
https://deploy-preview-4101--redash-preview.netlify.com/dashboard/test-scroll-x?p_count=30&p_w287_count=5

1. AutoHeight should be ok now.
2. Widgets now display pagination mini version.

arikfr:
Seems to work correctly now, but the size of the widget is a bit too wide now:

image

ranbena:
> Seems to work correctly now, but the size of the widget is a bit too wide now:

I can't believe the scroll is not inset! Happens in Chrome as well?

arikfr:
@ranbena Chrome is fine.. Only Firefox :-(

kravets-levko:
@arikfr Seems it's a bug in `react-grid-layout` (or maybe it's related to how we use it, don't know) - it happens sometimes in different browsers (Chrome, FF, IE) and grid gets fixed when you resize browser window (try to open devtools or whatever). IMHO it should be fixed, but definitely not in this PR.

arikfr:
@kravets-levko the reason it feels part of this PR is because before this PR I never experienced it. So even if it's not directly a bug in this PR, it does trigger it.

Did you experience it before?

kravets-levko:
> Did you experience it before?

Yes, from the first day we use `react-grid-layout`. But it didn't happen very often to me, even on the same dashboard this bug does not reproduce stably. Therefore I cannot guess what caused this bug.

ranbena:
@arikfr does it consistently reproduce for you?
If so:
1. Does it get fixed on window resize?
2. What is the inline style width of `.react-grid-item` when it's ok and when it's not?

arikfr:
:thinking: it was reproducing when I commented, but now it no longer reproduces... I guess we can merge this then?

ranbena:
Then I'll prepare this for review

client/app/assets/less/redash/query.less
```diff
@@ -208,13 +208,6 @@ edit-in-place p.editable:hover {
   }
 }
 
-.visualization-renderer {
-  .pagination,
-  .ant-pagination {
-    margin-top: 10px;
-  }
-}
-
```
ranbena: There's no such classname

client/app/visualizations/VisualizationRenderer.jsx
```diff
@@ -60,14 +60,13 @@ export function VisualizationRenderer(props) {
...
-          visualizationName={visualization.name}
-        />
-      </div>
+      <Renderer
+        key={`visualization${visualization.id}`}
+        options={options}
+        data={filteredData}
+        visualizationName={visualization.name}
+        context={props.context}
+      />
```
ranbena: Removed unused `<div>`

client/app/visualizations/table/renderer.less
```diff
@@ -100,4 +100,23 @@
       }
     }
   }
+
+  /* START table x scroll */
+  .dashboard-widget-wrapper:not(.widget-auto-height-enabled) & {
+    height: 100%;
+
+    & div {
+      height: inherit;
```
ranbena: A must so `.ant-spin-container` gets the actual `height: 100%`.

PR #4105, opened 2019-08-30
Upgrading from 3.0.0 to 5.0.0 or upper is failed

shinsuke-nara:

- [x] Bug Fix

## Description

`tags` field is added to `models.Dashboard` from 5.0.0.
As a result, `upgrade` from 3.0.0 or lower fails now.

I fixed this by using [session.execute](https://docs.sqlalchemy.org/en/12/orm/session_api.html#sqlalchemy.orm.session.Session.execute) method.

https://discuss.redash.io/t/upgrade-from-3-0-0-b3134-to-v5/2384


arikfr:
Thanks :+1:

Issue #3551, opened 2019-03-07
Replace `numeral` with `numbro` (or another alternative)


kravets-levko:
We use [`numeral`](http://numeraljs.com/) package to format numbers in visualizations (and probably in some other places), but it wasn't updated for a quite long time, and seems it's owner don't plan to improve it. But sometimes we need features not supported by `numeral` - say, using escape sequences in format string.

There are other libraries similar to `numeral` which are more actively maintained, like [`numbro`](http://numbrojs.com/) - which supports similar format syntax and can use escape sequences. `numbro` is my suggestion, however, some another library could be used as well (if it supports the same format syntax and escape sequences - in fact we don't use all that number manipulation stuff, only formatting).

kravets-levko:
Added "Feature request" label because technically we need to replace the library in order to use new features (escape sequences in format string)

morsedl:
+1

rcoup:
https://discuss.redash.io/t/other-currency-types-in-tables/3356/4 links here to point out '£' doesn't work as a currency symbol.

Numbro's [approach for currency formatting](http://numbrojs.com/format.html#currency) seems weird to me: if you call `.formatCurrency()` and your language is set to Japanese, it'll show ¥, but that makes the _enormous_ assumption people only ever want to use their "local" currency in queries? Simpler: allow locale-based number formatting, but let the user put the appropriate currency symbol in the correct place? (eg. `'£ 0.00'`)

kravets-levko:
@rcoup The idea is not to use currency formatting abilities provided by `numeral`/`numbro` at all, instead letting user to just add arbitrary characters in format string(e.g. `0[.00] {USD}`). This will allow to use different currency signs at any position, which is more flexible.

vvohra87:
Any update on when this will be sorted out?

basilvetas:
Is there a way to use the custom Data Labels field in combination with numbers format? I'd like to be able to do something like {{ currency_col }} 0,0[.].00 to format my and apply the correct currency based on a query column.

beetsandroots:
+1

zhongjiajie:
I'm not familiar with our codebase, but I find `numeral` key word in repo, https://github.com/getredash/redash/search?q=numeral&type=Code and only 5 match. So the point is `numbro ` not 100% compatible with `numeral`?

Issue #2558, opened 2018-05-29
Option to set number format in counters


kravets-levko:
Add an option to counter visualization so user can set format for values like in other visualizations (tables, charts, etc).

PR #4106, opened 2019-08-30
Migrate Counter visualization to React

kravets-levko:

- [x] Feature
- [x] Refactor

## Description

  - [x] Renderer
  - [x] Editor
- [x] Fix things mentioned in comments (handle edge cases, etc.)
- [x] Cleanup
- [x] Test

getredash/redash#3301 (Migrate Visualizations to React -> Counter)

image
image


client/app/visualizations/counter/utils.js
```diff
@@ -0,0 +1,140 @@
...
+  const result = numeral(value).format(formatString);
+
+  locale.delimiters = savedDelimiters;
+  return result;
+}
+
+// 0 - special case, use first record
+// 1..N - 1-based record number from beginning (wraps if greater than dataset size)
+// -1..-N - 1-based record number from end (wraps if greater than dataset size)
+function getRowNumber(index, rowsCount) {
```
kravets-levko: @arikfr I may need your opinion about this ^

client/app/visualizations/counter/Renderer.jsx
```diff
@@ -0,0 +1,74 @@
...
+import { RendererPropTypes } from '@/visualizations';
+
+import { getCounterData } from './utils';
+
+import './render.less';
+
+function getCounterStyles(scale) {
+  return {
+    OTransform: `scale(${scale})`,
+    MsTransform: `scale(${scale})`,
```
ranbena: I believe this should be `msTransform`.

In any case, wdyt of using `modernizr-prefixed` or a simple custom util here instead?

client/app/visualizations/counter/Renderer.jsx
```diff
@@ -0,0 +1,74 @@
...
+    counterLabel,
+  } = getCounterData(data.rows, options, visualizationName);
+  return (
+    <div
+      className={cx(
+        'counter-visualization-container',
+        { 'trend-positive': showTrend && trendPositive, 'trend-negative': showTrend && !trendPositive },
+      )}
+    >
+      <div className="counter-visualization-content" ref={setContainer}>
```
ranbena: Why not use the dedicated `useRef`?

client/app/visualizations/counter/Renderer.jsx
```diff
@@ -0,0 +1,74 @@
...
+  const [container, setContainer] = useState(null);
+
+  useEffect(() => {
+    if (container) {
+      const unwatch = resizeObserver(container, () => {
+        setScale(getCounterScale(container));
+      });
+      return unwatch;
+    }
+  }, [container, setScale]);
```
ranbena: I'm probably not understanding how this works, but are you sure `setScale` is correct in the 2nd argument?

client/app/visualizations/counter/Editor.jsx
```diff
@@ -0,0 +1,236 @@
...
+        <GeneralSettings {...props} onOptionsChange={optionsChanged} />
+      </Tabs.TabPane>
+      <Tabs.TabPane key="format" tab={<span data-test="Counter.EditorTabs.Formatting">Format</span>}>
+        <FormatSettings {...props} onOptionsChange={optionsChanged} />
+      </Tabs.TabPane>
+    </Tabs>
+  );
+}
+
+Editor.propTypes = EditorPropTypes;
```
ranbena: @arikfr wdyt of having a link to the counter video from here?
(Actually, linking to a help doc or blog post with the video in it would be better)

client/cypress/integration/visualizations/counter_spec.js
```diff
@@ -0,0 +1,191 @@
...
+    cy.clickThrough(`
+      NewVisualization
+      VisualizationType
+      VisualizationType.COUNTER
+
+      Counter.General.ValueColumn
+      Counter.General.ValueColumn.a
+    `);
+
+    cy.wait(200); // eslint-disable-line cypress/no-unnecessary-waiting
```
ranbena: What's the wait for?

client/app/visualizations/counter/Editor.jsx
arikfr: I agree that having a help trigger here makes sense. :+1: But I guess it should be a global thing for all visualizations, which we enable when we have a relevant page?

client/app/visualizations/counter/utils.js
```diff
@@ -0,0 +1,141 @@
+import { isNumber, isFinite, toString } from 'lodash';
+import numeral from 'numeral';
+
+// TODO: allow user to specify number format string instead of delimiters only
+// It will allow to remove this function (move all that weird formatting logic to a migration
+// that will set number format for all existing counter visualization)
```
arikfr: Can't we do this already without a migration by just translating the different settings into a format string for existing visualizations and use a string for new ones?

client/app/visualizations/counter/utils.js
arikfr: The description sounds right :)

kravets-levko: Previous implementation worked in a similar way (I believe), so I first described how it should behave to be clear for users, and then implemented according to that.

client/app/visualizations/counter/Renderer.jsx
kravets-levko: Yeah, you're right - `-ms-` is the only exception :man_facepalming: Also, probably I'll remove `-o-` rule (Opera 12 is in fact dead)

client/app/visualizations/counter/Renderer.jsx
kravets-levko: Previously I had some issues passing a regular ref as a dependency to `useEffect` (it just ignored ref changes). This variant works perfectly

client/app/visualizations/counter/Renderer.jsx
kravets-levko: Yes it's correct - when you use `useEffect`/`useMemo`/`useCallback` you should pass every dependency you'll use in a hook function because the function will capture them and then you'll need to create new function instance that will capture new dependencies values (you may omit some of dependencies or pass extra ones if you know what you do)

client/cypress/integration/visualizations/counter_spec.js
kravets-levko: There were some issues similar to other visualizations tests - visualization needed some delay to properly update itself before taking a snapshot.

client/app/visualizations/counter/utils.js
kravets-levko: It's the next thing I'd like to do. But I think it's better to do it as a separate feature (as implementation of getredash/redash#2558), but for a proper implementation in a single pass we need to do something with getredash/redash#3551 - it will allow to get rid of "Formatting String Prefix/Suffix" settings as well.

arikfr: :+1:

client/app/visualizations/counter/Renderer.jsx
kravets-levko: `modernizr` seems to be an overkill here :thinking: I'll check if we're using something similar elsewhere - I'll create a custom utility

client/app/visualizations/counter/Renderer.jsx
gabrieldutra: It still feels to me like you indeed should pass dependencies that you use in the hook function, but only the ones that get changed over time. Also in [docs](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) they don't seem to include the setters in the dependency list

kravets-levko: @gabrieldutra You're right. It's also mentioned in hooks API reference. Thanks!

client/app/visualizations/counter/Editor.jsx
ranbena: We can embed the video here then https://redash.io/help/user-guide/visualizations/visualization-types#Counter

client/cypress/integration/visualizations/counter_spec.js
ranbena: Then perhaps it makes more sense to place it before the snapshot (instead of before the element assertion) 👆

client/app/visualizations/counter/Renderer.jsx
kravets-levko: Fixed

client/cypress/integration/visualizations/counter_spec.js
kravets-levko: @ranbena Done

client/app/visualizations/counter/Renderer.jsx
gabrieldutra: [This](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) seems relevant to cases where the DOM changes and you need to track it in the ref.

kravets-levko:
@arikfr @ranbena @gabrieldutra If you are happy with this PR and have no more comments to add - I'd like to merge it.

client/cypress/integration/visualizations/counter_spec.js
```diff
@@ -0,0 +1,207 @@
...
+
+      Counter.General.TargetValueColumn
+      Counter.General.TargetValueColumn.b
+    `);
+
+    cy.getByTestId('VisualizationPreview').find('.counter-visualization-container').should('exist');
+
+    // wait a bit before taking snapshot
+    cy.wait(500); // eslint-disable-line cypress/no-unnecessary-waiting
+    cy.percySnapshot('Visualizations - Counter (format target value)', { widths: [viewportWidth] });
```
gabrieldutra: It seems all the Percy Snapshots have the same name :sweat_smile:, guess it will mess with Percy...

kravets-levko: Uh-oh... Thanks! Probably a copy-paste leftover...

kravets-levko:
> This is probably part of another bug

Not necessarily, it may be introduced in this PR - I'll check

client/cypress/integration/visualizations/counter_spec.js
kravets-levko: Fixed :+1:

kravets-levko:
> This is probably part of another bug, but as a note, when you have an empty filter the Counter now disappears.

@gabrieldutra There is a condition - counter will not render anything if there are no rows in result - and this is intended behavior. I think it's kinda correct (probably we can show `0` if `Count rows` option enabled), but we can discuss it and change something in another PR.

gabrieldutra:
It makes sense for it to show 0 when count rows is enabled, but not in this scope indeed.

arikfr:
:+1:

PR #4109, opened 2019-09-01
Fix Dropdown parameter options appearing behind Dialog

gabrieldutra:

- [x] Bug Fix

## Description
This adds `getPopupContainer` to Select parameters to fix z-index issues with it.

Extra: Fix typo :slightly_smiling_face: 

--

**Before**
dropdown-before

**After**
dropdown-after

### DIFF
--- a/client/app/assets/less/ant.less
+++ b/client/app/assets/less/ant.less
@@ -77,12 +77,6 @@
   }
 }
 
-// Fix for Ant dropdowns when they are used in Boootstrap modals
-// ANGULAR_REMOVE_ME Remove when all dialogs will be migrated to React (also search and remove usages)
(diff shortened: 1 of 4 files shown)


ranbena:
I think the reason is actually this:
https://github.com/getredash/redash/blob/cab011def96158adb89c4a00290e1a20ef3f4bd0/client/app/components/QueryBasedParameterInput.jsx#L84

It's an old method of dealing with ant-design z-index which I don't think is relevant anymore.
And if you remove it, the Select gets the right z-index.

gabrieldutra:
Good one @ranbena, it also works and it really seams we don't need it anymore :slightly_smiling_face:.

PR #4110, opened 2019-09-01
Fix widget bottom element alignment

ranbena:
- [x] Bug Fix

## Description
@arikfr noticed a few misalignments in dashboard widgets. This is the fix.

Before|After
------|------
Screen-Shot-2019-09-01-at-10 29 59 | Screen-Shot-2019-09-01-at-10 29 25

### DIFF
--- a/client/app/components/dashboards/widget.html
+++ b/client/app/components/dashboards/widget.html
@@ -64,22 +64,26 @@
       </div>
     </div>
 
-    <div class="body-row clearfix tile__bottom-control">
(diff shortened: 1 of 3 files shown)


ranbena:
Actually, had to change the whole thing to flexbox, that insets only outlaying transparent buttons.

Tested on:
* private dashboard
* public dashboard
* embeded viz
* snapshot.

ranbena:
@gabrieldutra making sure you're up-to-date about this cause it'll require rebasing your branch.

gabrieldutra:
I actually made some changes on the widget structure/styling, but not on its bottom, will update in there :+1:

Issue #4111, opened 2019-09-01
User can't refresh some Query Results queries


arikfr:
### Issue Summary

User A has access to Data Source A and the Query Results data source.
User B has only access to the Query Results data source.

User A creates a Query Results query which uses Data Source A.

Current behavior: User B can't refresh this query, because they don't have access to Data Source A.

Expected behavior: User B should be able to refresh this query, just like an external user can using the query API key.

arikfr:
Possible solution:

• "Compute" all the queries this query accesses before applying parameters.
• Make sure query creator has access to all of them.
• Allow query viewer to access them.

rauchy:
Is there anything left to be done here or did #4112 solve this entirely?

arikfr:
Currently we support the case where you have Read Only access to all the data sources involved. We should probably support the case where you have Read Only access only to the Query Results data source.

PR #4112, opened 2019-09-01
Fix: allow users with view only acces to use the queries in Query Results

arikfr:

- [x] Bug Fix

## Description

Allow users who have View Only access to a data source using its queries in a Query Results query.

This does not solve the use case described in #4111, but a limited scope of it (where you have View Only access vs. Full Access).

### DIFF
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -38,7 +38,8 @@ def _load_query(user, query_id):
     # TODO: this duplicates some of the logic we already have in the redash.handlers.query_results.
     # We should merge it so it's consistent.
     if not has_access(query.data_source, user, view_only):
-        raise PermissionError("Query id {} not found.".format(query.id))
+        raise PermissionError(u"You do not have access to query id {}.".format(
+            query.id))
 
     return query
 
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -38,9 +38,7 @@ def _load_query(user, query_id):
     # TODO: this duplicates some of the logic we already have in the redash.handlers.query_results.
     # We should merge it so it's consistent.
     if not has_access(query.data_source, user, view_only):
-        raise PermissionError(
-            u"You are not allowed to execute queries on {} data source (used for query id {})."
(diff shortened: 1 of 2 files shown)


redash/query_runner/query_results.py
```diff
@@ -34,9 +35,12 @@ def _load_query(user, query_id):
     if user.org_id != query.org_id:
         raise PermissionError("Query id {} not found.".format(query.id))
 
-    if not has_access(query.data_source, user, not_view_only):
-        raise PermissionError(u"You are not allowed to execute queries on {} data source (used for query id {}).".format(
-            query.data_source.name, query.id))
+    # TODO: this duplicates some of the logic we already have in the redash.handlers.query_results.
+    # We should merge it so it's consistent.
+    if not has_access(query.data_source, user, view_only):
```
arikfr: This is the actual change. The rest is my auto formatter :grimacing:

rauchy:
LGTM. It's a very subtle assumption, so I believe it needs a very descriptive test.

tests/query_runner/test_query_results.py
```diff
@@ -135,6 +211,15 @@ def test_returns_query(self):
         loaded = _load_query(user, query.id)
         self.assertEquals(query, loaded)
 
+    def test_returns_query_when_user_has_view_only_access(self):
```
arikfr: New test.

redash/query_runner/query_results.py
```diff
@@ -34,9 +35,12 @@ def _load_query(user, query_id):
...
+        raise PermissionError(
+            u"You are not allowed to execute queries on {} data source (used for query id {})."
```
rauchy: Is this error message still accurate after this change?

arikfr: https://github.com/getredash/redash/pull/4112/commits/32ba5d76790871175934db016a4f34c08d192a4e

arikfr: Oops: https://github.com/getredash/redash/pull/4112/commits/bb4f76808e206549d5397bc7e14fa852583255a9.

PR #4116, opened 2019-09-02
Fix number param value normlization

### LINKED ISSUE #4115: Number parameter initial value yields "missing value" error
### Issue Summary

Adding a new number parameter shows `0` as initial value but sends `null` to the query execution, which results in an error.

Screen Shot 2019-09-02 at 8 44 12

ranbena:
Fixes #4115.

- [x] Bug Fix

## Description

The initial value is empty.
Non-number values are normalized to zero but only for show.

### The fix
Changed normalization to be `undefined` if not a number value, which displays empty, so there's no more confusion.

Screen Shot 2019-09-02 at 9 03 44

### DIFF
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -140,7 +140,7 @@ class ParameterValueInput extends React.Component {
     const { className } = this.props;
     const { value } = this.state;
 
-    const normalize = val => !isNaN(val) && val || 0;
+    const normalize = val => (isNaN(val) ? undefined : val);
 
     return (
       <InputNumber


ranbena:
> When we have a string value, will it send to the server the original value or the undefined one?

Neither. An undefined value would trigger the front-end "missing param value" error.
https://github.com/getredash/redash/blob/1f1f85329718772f2d92f87c0ad08382154c7e24/client/app/services/query.js#L711-L716

gabrieldutra:
The `0` default value was actually annoying me sometimes... Well I believe this kind of logic should better be handled in the Parameter Structure (parameter value normalization), but it's something I'll continue on #4049.

PR #4121, opened 2019-09-03
hive_ds: show a user friendly error message when possible

sphenlee:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Catch DatabaseErrors from pyhive and extract the user friendly error message to display to the user. This will suppress the full Java traceback which is usually not relevant.

Before:
before

After:
after


arikfr:
Thanks!

PR #4123, opened 2019-09-03
Use ng-src instead of src for data source icons

gabrieldutra:

- [x] Other

## Description
This doesn't lead to a bug, but it will avoid this:
ng-src-errors

#4119 

--


arikfr:
:+1:

PR #4130, opened 2019-09-09
Migrate Chart visualization to React Part 1: Renderer

kravets-levko:

- [x] Refactor

## Description

  - [x] "normal" chart
  - [x] custom chart
- [x] Cleanup
- [ ] More tests (?)

getredash/redash#3301 (Migrate Visualizations to React -> Chart)

No visual changes.

### DIFF
--- a/client/app/visualizations/chart/Renderer/PlotlyChart.jsx
+++ b/client/app/visualizations/chart/Renderer/PlotlyChart.jsx
@@ -1,42 +1,21 @@
 import { isArray, isObject } from 'lodash';
-import React, { useState, useEffect, useMemo } from 'react';
+import React, { useState, useEffect } from 'react';
 import { RendererPropTypes } from '@/visualizations';
 import resizeObserver from '@/services/resizeObserver';
 
 import getChartData from '../getChartData';
 import { Plotly, prepareData, prepareLayout, updateData, applyLayoutFixes } from '../plotly';
 
-export default function PlotlyChart(props) {
+export default function PlotlyChart({ options, data }) {
   const [container, setContainer] = useState(null);
 
-  const options = useMemo(() => {
-    const result = { ...props.options };
-    if (['normal', 'percent'].indexOf(result.series.stacking) >= 0) {
-      // Backward compatibility
-      result.series = {
-        ...result.series,
-        percentValues: result.series.stacking === 'percent',
-        stacking: 'stack',
-      };
-    }
-    return result;
-  }, [props.options]);
-
-  const plotlyData = useMemo(() => {
-    const series = getChartData(props.data.rows, options);
-    return prepareData(series, options);
-  }, [props.data, options]);
-
-  const plotlyLayout = useMemo(() => {
-    if (container) {
-      return prepareLayout(container, options, plotlyData);
-    }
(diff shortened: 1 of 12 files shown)


arikfr:
We won't gain anything from using their React wrapper?

kravets-levko:
@arikfr Not really - we don't use almost anything from the features it offers

client/app/visualizations/chart/Renderer/PlotlyChart.jsx
```diff
@@ -0,0 +1,51 @@
...
+      const plotlyOptions = { showLink: false, displaylogo: false };
+
+      const chartData = getChartData(data.rows, options);
+      const plotlyData = prepareData(chartData, options);
+      const plotlyLayout = prepareLayout(container, options, plotlyData);
+
+      // It will auto-purge previous graph
+      Plotly.newPlot(container, plotlyData, plotlyLayout, plotlyOptions).then(() => {
+        applyLayoutFixes(container, plotlyLayout, (e, u) => Plotly.relayout(e, u));
+      });
```
kravets-levko: @arikfr That's actually all we need from Plotly here + few `Plotly.relayout` calls below. That's why I think we don't need extra dependency of `react-plotly.js` - it will not allow us to get rid of `plotly.js`anyway because React component is just a wrapper around it.

client/app/visualizations/chart/Renderer/PlotlyChart.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+      container.on('plotly_restyle', (updates) => {
```
ranbena: Shouldn't this be moved to the below `useEffect()`?

kravets-levko: I'm not sure I understand your point. It uses some variables from the same scope, so cannot be moved to another `useEffect` without moving that variables to upper scope. If you mean that both `useEffect` calls should be merged - probably yes, but current variant is a sort of optimization: first `useEffect` will re-build chart when container, options and/or data changes - `Plotly.newPlot` will purge old chart if needed. And we need to call `Plotly.purge` only when container changes.

ranbena: > It uses some variables from the same scope

Perhaps one of us is misunderstanding useEffect's 2nd argument usage.
AFAIK, it's used as "componentshouldupdate". You can use any variables in the useEffect scope without adding them to the 2nd argument. Am I wrong about this?

My concern here is that `container.on` is being re-declared needlessly every time `options` and `data` change, but perhaps needs to only when `container` does.

kravets-levko: `container.on` should be called after each `Plotly.newPlot` because Plotly does a sort of shitty thing: it extends DOM element which is a plot container and adds new properties and methods to it, like `on`. So when you call `Plotly.newPlot` - it will destroy previous plot and unbind all events, and then you have to re-bind everything again. If that was a regular DOM events (`addEventListener`) - then yes, I totally agree with you

arikfr:
:+1:

PR #4133, opened 2019-09-11
Migrate Pivot Table visualization to React

gabrieldutra:

- [x] Refactor

## Description
Migration for the Pivot Table Visualization

- [x]  Render updated Pivot Table
- [x]  Make Editor settings work
- [x] Clean up old code
- [x] Add tests
- [x] Review changes and manual test

--

react-pivot-visualization


client/app/components/Filters.jsx
```diff
@@ -1,10 +1,10 @@
-import { isArray, indexOf, get, map, includes, every, some, toNumber, toLower } from 'lodash';
+import { isArray, indexOf, get, map, includes, every, some, toNumber } from 'lodash';
```
kravets-levko: Is this related to this PR?

gabrieldutra: Yes, I faced an issue with the react version, it doesn't handle Moment values as the jQuery version does. The jQuery tries to stringfy it, while the react version just breaks the whole thing.

So I was thinking about moving the logic we have to normalize row values for Filters somewhere shared and use it here as well.

client/app/visualizations/pivot/index.js
```diff
@@ -67,12 +67,12 @@ const PivotTableRenderer = {
 export default function init(ngModule) {
   ngModule.component('pivotTableRenderer', PivotTableRenderer);
 
-  ngModule.run(($injector) => {
+  ngModule.run(() => {
```
kravets-levko: You don't need `ngModule.run()` (and `ngModule` at all) now - it was needed only for two things: 1) register angular component (removed now); 2) retrieve `$injector` service - not needed as well. So just remove all this shit :slightly_smiling_face:

client/app/visualizations/pivot/Renderer.jsx
```diff
@@ -0,0 +1,66 @@
...
+function formatRows({ rows, columns }) {
+  return map(rows, row => mapValues(row, (value, key) => formatColumnValue(value, find(columns, { name: key }).type)));
+}
+
+export default function Renderer({ data, options, onOptionsChange }) {
+  const [config, setConfig] = useState({ data: formatRows(data), ...options });
+
+  useEffect(() => {
+    setConfig({ data: formatRows(data), ...options });
+  }, [data, options]);
```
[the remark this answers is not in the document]
gabrieldutra: My reasoning for using a state here is that from my understanding `onOptionsChange` will only "exist" (something else than `() => {}`) when in Editor context. However, for the Pivot Table it has the possibility of keeping the controls in the visualization, so I wanted to keep them updating internally even though it's not persisted.

In a more practical way, this is what happens when I use `useMemo`:
pivottable-usememo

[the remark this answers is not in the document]
kravets-levko: Got it :+1: Then let's keep as is

arikfr:
From the screenshots it looks like the Pivot table has a different font than the rest of the application. Is it like this for the regular pivot table for you as well?

arikfr:
:+1:

Issue #4132, opened 2019-09-10
Enhancements to Alert feature


ranbena:
Following #4081 the following feature should be implemented:
- [ ] **Disable/Enable alert**
Allow toggling an alert on/off if it gets temporarily irrelevant or noisy.
- [ ] **Top value / Any value**
Allow alert condition evaluation, not only on the query result top row, but also on each result row 
   <img width="508" alt="Any row Top row" src="https://user-images.githubusercontent.com/486954/64640544-e18b1880-d412-11e9-9c2d-be95e4acfb17.png">

- [ ] **Additional condition comparison options**
<img width="412" alt="Screen Shot 2019-09-10 at 21 44 30" src="https://user-images.githubusercontent.com/486954/64641134-31b6aa80-d414-11e9-95ee-89ca16d5bcda.png">

- [ ] **Trigger on save**
Whenever criteria changes are saved:
 1. Show loading indicator on "Save" button.
 2. Trigger a backend alert evaluation (which can result in sending alert notifications).
 3. Stop loading indicator on fresh data.

kravets-levko:
Adding more comparison options probably is the easiest one:
https://github.com/getredash/redash/blob/2c98f0425d58ec2d127cc55cf24e08e1e99e9f98/redash/models/__init__.py#L790-L808

PR #4134, opened 2019-09-11
Alerts: Add more condition comparison options

kravets-levko:

- [x] Refactor
- [x] Feature

Related to getredash/redash#4132

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -791,14 +791,25 @@ class Alert(TimestampMixin, BelongsToOrgMixin, db.Model):
         data = json_loads(self.query_rel.latest_query_data.data)
 
         if data['rows'] and self.options['column'] in data['rows'][0]:
-            value = data['rows'][0][self.options['column']]
-            op = self.options['op']
+            operators = {
+                '>': lambda v, t: v > t,
+                '>=': lambda v, t: v >= t,
+                '<': lambda v, t: v < t,
+                '<=': lambda v, t: v <= t,
+                '==': lambda v, t: v == t,
+                '!=': lambda v, t: v != t,
 
-            if op == 'greater than' and value > self.options['value']:
-                new_state = self.TRIGGERED_STATE
-            elif op == 'less than' and value < self.options['value']:
-                new_state = self.TRIGGERED_STATE
-            elif op == 'equals' and value == self.options['value']:
+                # backward compatibility
+                'greater than': lambda v, t: v > t,
+                'less than': lambda v, t: v < t,
+                'equals': lambda v, t: v == t,
+            }
+            should_trigger = operators.get(self.options['op'], lambda: False)
+
+            value = data['rows'][0][self.options['column']]
+            threshold = self.options['value']
+
+            if should_trigger(value, threshold):
                 new_state = self.TRIGGERED_STATE
             else:
                 new_state = self.OK_STATE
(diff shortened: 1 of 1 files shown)


redash/models/__init__.py
```diff
@@ -791,14 +791,25 @@ def evaluate(self):
...
+                '<=': lambda v, t: v <= t,
+                '==': lambda v, t: v == t,
+                '!=': lambda v, t: v != t,
+
+                # backward compatibility
+                'greater than': lambda v, t: v > t,
+                'less than': lambda v, t: v < t,
+                'equals': lambda v, t: v == t,
+            }
+            should_trigger = operators.get(self.options['op'], lambda: False)
```
arikfr: The fallback lambda should take arguments as well.

redash/models/__init__.py
```diff
@@ -791,14 +791,25 @@ def evaluate(self):
         data = json_loads(self.query_rel.latest_query_data.data)
 
         if data['rows'] and self.options['column'] in data['rows'][0]:
+            operators = {
+                '>': lambda v, t: v > t,
+                '>=': lambda v, t: v >= t,
+                '<': lambda v, t: v < t,
+                '<=': lambda v, t: v <= t,
+                '==': lambda v, t: v == t,
+                '!=': lambda v, t: v != t,
```
kravets-levko: WDYT about replacing that "long" operator names with these? (old ones kept for backward compatibility: already created alerts can use them, but new ones will use new operators)

ranbena: I'd also consider the dedicated chars like `≥`.

kravets-levko: User does not see them; in dropdown we can show whatever we want, but here I'd prefer to use ANSI characters like we do for variable names.

arikfr: @ranbena why? I don't even know how to type those without looking up some unicode characters map.

redash/models/__init__.py
kravets-levko: Ooops... Done :+1:

redash/models/__init__.py
ranbena: > use ANSI characters like we do for variable names.

Cool.

> why?

Cleaner when implementing on the Select component.

kravets-levko: > Cleaner when implementing on the Select component.

Not sure how it will help. Imagine that we'll add some new operation that will not have a dedicated Unicode symbol, e.g. `contains` or `match regexp`. IMHO these identifiers are for internal use and shouldn't be used for dropdown option names.

ranbena: I agree. I meant sth else. Doesn't matter. Let's land it.

kravets-levko:
BTW - seems that all comparison operations are not limited to numbers - they can compare strings as well :thinking: Is this somehow restricted in UI?

arikfr:
> BTW - seems that all comparison operations are not limited to numbers - they can compare strings as well  Is this somehow restricted in UI?

I'm not sure how is it in Ran's PR, but current UI has `<input type="number">` which prevents strings.

ranbena:
> I'm not sure how is it in Ran's PR, but current UI has `<input type="number">` which prevents strings.

Same. Let's continue discussion here https://discuss.redash.io/t/alerts-feature-breakdown-aug-19/4312

PR #4147, opened 2019-09-16
Add interface to abstract query result persistence

arikfr:

- [x] Refactor

## Description

This is a step towards supporting different backends (S3, filesystem, etc) for storing query results (the actual data). In this PR we're abstracting the persistence from the model. The default implementation keeps storing it in the DB as it was until now.

We DO NOT recommend creating extensions based on this, as the interface will probably change. Current interface has limitations like preventing streaming data, which we would like to eventually support.

#78

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -252,7 +252,7 @@ QueryResultPersistence = settings.dynamic_settings.QueryResultPersistence or DBP
 
 @python_2_unicode_compatible
 @generic_repr('id', 'org_id', 'data_source_id', 'query_hash', 'runtime', 'retrieved_at')
-class QueryResult(db.Model, BelongsToOrgMixin, QueryResultPersistence):
+class QueryResult(db.Model, QueryResultPersistence, BelongsToOrgMixin):
     id = Column(db.Integer, primary_key=True)
     org_id = Column(db.Integer, db.ForeignKey('organizations.id'))
     org = db.relationship(Organization)
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -232,9 +232,21 @@ class DataSourceGroup(db.Model):
     __tablename__ = "data_source_groups"
 
 
+class DBPersistence(object):
+    @property
+    def data(self):
+        return self._data
+
+    @data.setter
+    def data(self, data):
+        self._data = data
+
+
+QueryResultPersistence = settings.dynamic_settings.QueryResultPersistence or DBPersistence
+
 @python_2_unicode_compatible
 @generic_repr('id', 'org_id', 'data_source_id', 'query_hash', 'runtime', 'retrieved_at')
-class QueryResult(db.Model, BelongsToOrgMixin):
+class QueryResult(db.Model, BelongsToOrgMixin, QueryResultPersistence):
     id = Column(db.Integer, primary_key=True)
     org_id = Column(db.Integer, db.ForeignKey('organizations.id'))
     org = db.relationship(Organization)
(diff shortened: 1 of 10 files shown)


redash/models/__init__.py
```diff
@@ -304,22 +322,12 @@ def store_result(cls, org, data_source, query_hash, query, data, run_time, retri
...
-            Query.query_hash == query_hash,
-            Query.data_source == data_source
-        )
-        for q in queries:
-            q.latest_query_data = query_result
-            # don't auto-update the updated_at timestamp
-            q.skip_updated_at = True
-            db.session.add(q)
-        query_ids = [q.id for q in queries]
-        logging.info("Updated %s queries with result (%s).", len(query_ids), query_hash)
```
arikfr: We moved this to a separate class method on the `Query` model. Eventually wasn't needed for this refactor, but it's a good separation regardless so we kept it here.

redash/models/__init__.py
```diff
@@ -256,7 +274,7 @@ def to_dict(self):
             'id': self.id,
             'query_hash': self.query_hash,
             'query': self.query_text,
-            'data': json_loads(self.data),
+            'data': self.data,
```
arikfr: We now deserialize it in a single location instead of having every user call `json_loads`.

redash/models/__init__.py
```diff
@@ -639,6 +647,26 @@ def all_groups_for_query_ids(cls, query_ids):
...
 
         return db.session.execute(query, {'ids': tuple(query_ids)}).fetchall()
+    
+    @classmethod
+    def update_latest_result(cls, query_result):
+        # TODO: Investigate how big an impact this select-before-update makes.
+        queries = Query.query.filter(
+            Query.query_hash == query_result.query_hash,
+            Query.data_source == query_result.data_source
+        )
```
jezdez: In case there are many similar queries (e.g. forked but never changed) it may make sense to use `with_for_update` here to generate a `FOR UPDATE` SQL statement and then use `update()` here instead of iterating over the list of queries to update the query_result.

redash/models/__init__.py
```diff
@@ -232,17 +232,35 @@ class DataSourceGroup(db.Model):
     __tablename__ = "data_source_groups"
 
 
+class DBPersistence(object):
+    @property
+    def data(self):
+        if self._data is None:
```
jezdez: Shouldn't this also check if `self._deserialized_data` has already been set and return it if yes (and also unset it in the property setter below)?

redash/models/__init__.py
arikfr: So far this wasn't an issue, but anyway it's the same behavior we had until now -- I just moved it. So changing it is out of scope for this PR.

redash/models/__init__.py
arikfr: In what case `_data` will be `None` and `_deserialized_data` has some value?

redash/models/__init__.py
jezdez: Makes sense.

redash/models/__init__.py
jezdez: This is basically a race condition when `_data` is set to `None` via the data property setter below when it was previously not `None`. If `data` should just be used as a read-only property I would suggest to remove the property setter.

arikfr: I don't know. It feels like this complicates the code for no good reason, as this class has a very defined use :thinking:

arikfr: Read your comment again and realized that actually making `_deserialized_data` in sync with the setter isn't that complex. Implemented in 16fbc5e2940fc6b98d2bfdc53653479347ba403d + added tests.

PR #4152, opened 2019-09-18
WIP: Use UTC for Dynamic Values

gabrieldutra:

- [x] Bug Fix

## Description
Following [this](https://github.com/getredash/redash/pull/4049#discussion_r325121340) comment, this PR updates Dynamic Values to use UTC format to avoid internal inconsistencies.

--

--

### DIFF
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -27,50 +27,50 @@ const DYNAMIC_PREFIX = 'd_';
 const DYNAMIC_DATE_RANGES = {
   today: {
     name: 'Today',
-    value: () => [moment().startOf('day'), moment().endOf('day')],
+    value: () => [moment.utc().startOf('day'), moment.utc().endOf('day')],
   },
   yesterday: {
     name: 'Yesterday',
-    value: () => [moment().subtract(1, 'day').startOf('day'), moment().subtract(1, 'day').endOf('day')],
+    value: () => [moment.utc().subtract(1, 'day').startOf('day'), moment.utc().subtract(1, 'day').endOf('day')],
   },
   this_week: {
     name: 'This week',
-    value: () => [moment().startOf('week'), moment().endOf('week')],
+    value: () => [moment.utc().startOf('week'), moment.utc().endOf('week')],
   },
   this_month: {
     name: 'This month',
-    value: () => [moment().startOf('month'), moment().endOf('month')],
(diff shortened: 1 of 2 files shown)


kravets-levko:
@gabrieldutra Your fix for dynamic values works fine, however I found another related issue - with regular datetime/range values - they are in local timezone as well.

I ran this query at `16:03` local time (+3.00) and it should return events for last 1 hour (approx.). But since server's local time was `13:03` - it returned nothing because the time passed in parameter didn't come yet.
image

So now I think - maybe we should show values in local timezone (as it was), but convert them to utc right before passing to backend (e.g. `param: value.utc().format(...)`). WDYT?

gabrieldutra:
> but convert them to utc right before passing to backend (e.g. param: value.utc().format(...)). 

This one is trickier to change, it's been like that for a while. What I meant when I mentioned that I've always used `CONVERT_TZ` within the queries is that when I wrote them I knew that when a user selected a datetime, they would expect it to be in a Brazilian TZ, so I always wrote the queries converting it to make it friendlier to whoever would use those parameters or visualize date results.

I wonder if the aforementioned behavior we had to write queries is common among other companies :confused:. In case it is, it's sort of a risk to invalidate the existing queries that take that behavior into account.

I'm again concerned about using UTC in here (for dynamic values) as it's not synced with Date parameters behavior.

I'll think better about this as this comment is just an immediate thought that came to my mind and I wanted to share.

gabrieldutra:
Yeap, making a different behavior for Dynamic Values than it is for picking the values as static does concern me to merge this. 

UTC values for Dynamic Values only make sense in case we really assume the user that'll run the query will input UTC values in the Parameter (what was not the context, at least for my case).

I think we should leave it as the users TZ for now to be consistent with the behavior static date parameters had so far and move the discussion to the forum (along with other TZ issues).

To be clearer about what's going on in the UI, we can perhaps add a tooltip with the actual values that'll be placed. WDYT?

kravets-levko:
@gabrieldutra Please don't close this PR - I want to try some ideas I have around it

guidopetri:
@kravets-levko , thanks for the interest in this PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

PR #4154, opened 2019-09-19
Adjust botocore dependency, so we don't need to update it as often

justinclift:

- [x] Bug Fix

## Description

Updated botocore again, as boto3 now requires the latest release (v1.12.239).

With this PR, in theory we shouldn't need to update it manually in future all that much.  Hopefully. :smile:

Without this change, installation gives the following error:

```
boto3 1.9.231 has requirement botocore<1.13.0,>=1.12.239, but you'll have
botocore 1.12.220 which is incompatible.
```

### DIFF
--- a/requirements_all_ds.txt
+++ b/requirements_all_ds.txt
@@ -12,7 +12,7 @@ pymssql==2.1.3
 dql==0.5.24
 dynamo3==0.4.7
 boto3==1.9.115
-botocore==1.12.220
+botocore>=1.12.239,<1.13.0
 sasl>=0.1.3
 thrift>=0.8.0
 thrift_sasl>=0.1.0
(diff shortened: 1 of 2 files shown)


requirements_all_ds.txt
```diff
@@ -12,7 +12,7 @@ pymssql==2.1.3
 dql==0.5.24
 dynamo3==0.4.7
 boto3==1.9.115
-botocore==1.12.220
+botocore>=1.12.220
```
arikfr:
```suggestion
botocore>=1.12.239,<1.13.0
```

requirements_dev.txt
```diff
@@ -6,7 +6,7 @@ mock==2.0.0
 # PyMongo and Athena dependencies are needed for some of the unit tests:
 # (this is not perfect and we should resolve this in a different way)
 pymongo[tls,srv]==3.6.1
-botocore==1.12.220
+botocore>=1.12.220
```
arikfr:
```suggestion
botocore>=1.12.239,<1.13.0
```

requirements_all_ds.txt
justinclift: No worries, done for both. :smile:

justinclift:
Updated.  Should be good to go now. :smile:

arikfr:
:+1:

Issue #3771, opened 2019-05-06
Split Data Sources (and Alert Destinations) setup to regular and advanced


arikfr:
Some data sources have a very long list of options, where many times only a handful of them are used on a regular basis.

We should add a boolean option to the field definitions, which notes if the option is regular or advanced. If it's advanced it will most to a collapsed by default advanced settings section in the `DynamicForm`.

PR #4160, opened 2019-09-20
Split setup in advanced and regular for data sources and destinations

gabrieldutra:

- [x] Feature

## Description
Following #3771 to simplify the setup for DS, this will allow selecting settings to be considered "advanced".

#3771 

advanced-settings

## Notes
- I've randomly selected a few options to be considered advanced, just so we get the feel of it on Preview;
- Default values for advanced options are filled in the form when you are creating a DS/Destination (regular options are turned into placeholders). This allows the user to fill in the form without even looking into the advanced options;
- So far I'm using Antd's Collapse default for the job, but lmk if you have other suggestions :slightly_smiling_face:


ranbena:
Awesome @gabrieldutra.

A few thoughts:
1. I think the collapse style is good 👍
2. Give the collapsed element `m-t-30, m-b-10` for some spacing.
3. I'd rename to "More options" since it's not necessarily "advanced".
4. Consider keeping this a front-end-only feature, by putting all non-required fields in the collapsed section, or by max field count.

Sidenote - I don't see a point to this extra padding:
https://github.com/getredash/redash/blob/cb654b3f2117507e2fcf2f866e8da5543665c928/client/app/components/CreateSourceDialog.jsx#L103

arikfr:
@ranbena it can't be a front end only feature because each data source
needs to pick which configuration should be at front and which should be
collapsed.

Naming it "More Options" makes sense.

Also, is the collapsed state visible enough?

kravets-levko:
We already have a bit customized Collapse in SaaS, maybe we can use it here as well?

image
image

gabrieldutra:
> Consider keeping this a front-end-only feature, by putting all non-required fields in the collapsed section

As Arik said, it's better to pick the fields that will be collapsed. There are cases where we want non-requierd fields to be visible (e.g: password field, it can be left empty, but it should be visible)

> Also, is the collapsed state visible enough?

I was a bit concerned at the beginning, but it turned out to be ok to me. One other option is the SaaS style Levko mentioned.

Also, I think it makes sense to use the customized Collapse we have, so we keep consistent in the UI.

---

I've just realized the randomly selected options were done in the backend, so it's not possible to try this in preview :sweat_smile:

ranbena:
I made a kind of a hybrid between the two and I quite like it:

Screen Shot 2019-09-21 at 16 40 12

```less
{
  height: 40px;
  font-weight: 500;
  background-color: @btn-danger-bg;
  border-color: @btn-danger-border;
}
```

client/app/components/dynamic-form/dynamicFormHelper.js
```diff
@@ -42,26 +43,32 @@ function normalizeSchema(configurationSchema) {
...
-  if (Object.keys(options).length === 0) {
-    const properties = configurationSchema.properties;
-    Object.keys(properties).forEach((property) => {
-      if (!isUndefined(properties[property].default) && properties[property].type === 'checkbox') {
+function setDefaultValueToFields(configurationSchema, options = {}) {
+  const properties = configurationSchema.properties;
+  Object.keys(properties).forEach((property) => {
+    if (!isUndefined(properties[property].default)) {
+      // set default value for checkboxes and for advanced options
+      if (properties[property].type === 'checkbox' || properties[property].advanced) {
```
arikfr: Why the advanced options need special handling here?

gabrieldutra: This is related to the second note 🙂:

- Default values for advanced options are filled in the form when you are creating a DS/Destination (regular options are turned into placeholders). This allows the user to fill in the form without even looking into the advanced options;

gabrieldutra:
> consolidate naming in codeadvanced options / extra options / additional settings

Yes, that's on my plans as soon as I figure what makes more sense (it's been changing since I opened, so I haven't handled that yet)

client/app/components/dynamic-form/dynamicFormHelper.js
arikfr: We show as a placeholder the default value and we use it as well. So there is no need to have a different behavior for the advanced options: they can skip them and the default value will still be used.

Although this made me think of why aren't we just filling in the default value for regular options as well? The only reason I can think of is to be able to change the default, but not sure if this is desired :thinking: 

Anyway, it's out of scope, so worth keeping the same behavior until we decide if we want to change.

gabrieldutra: I was thinking about required values also being able to be additional settings, but considering what you mentioned I shouldn't be worried about them at all (also not a very important context anyway)

arikfr: A required setting will never be in advanced settings. The idea is to hide there all the things people won't really bother configure in normal cases. This doesn't apply to required settings.

ranbena:
@arikfr @gabrieldutra I question the need to have the "Additional Settings" functionality in datasource/destination _view mode_. It ain't a modal so let it be as long as it's settings, no?

If not, I think at least we should have the additional settings _open by default_ if a setting there has been modified.

arikfr:
> @arikfr @gabrieldutra I question the need to have the "Additional Settings" functionality in datasource/destination _view mode_. It ain't a modal so let it be as long as it's settings, no?

In general you're right. Let's address it when we actually have a view mode for Data Sources ;-)

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -56,10 +56,11 @@ class DynamicForm extends React.Component {
   constructor(props) {
     super(props);
 
+    const filledExtraFields = filter(props.fields, field => field.extra && field.initialValue !== undefined);
```
ranbena: Perhaps use `some`?

ranbena: Notice that if a user sets an extra field but then clear the field - it won't be undefined but '' hence the section would open needlessly.

Maybe check `!isEmpty(field.initialValue)` instead?

gabrieldutra: It is a bit trickier to determine whether a field is empty :confused:, `undefined` was the easiest I thought about (but there is the issue you mentioned). `isEmpty`won't work neither for number nor for booleans. I'm thinking about an alternative option as I would like to keep it as a DynamicForm behavior :grimacing:

client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -51,9 +56,14 @@ class DynamicForm extends React.Component {
   constructor(props) {
     super(props);
 
+    const hasFilledExtraField = some(props.fields, (field) => {
+      const { extra, initialValue } = field;
+      return extra && (!isEmpty(initialValue) || isNumber(initialValue) || isBoolean(initialValue) && initialValue);
```
gabrieldutra: @ranbena this seemed the simplest without separating this and deep evaluating the values with their defaults. I've also tried the last one and my opinion is that it's an unnecessary extra complexity.

gabrieldutra:
@arikfr, I can go around and try to classify some DS/Destination fields should be classified as Additional (what of course should be reviewed afterwards), should I?

arikfr:
> @arikfr, I can go around and try to classify some DS/Destination fields as Extra Option (what of course should be reviewed afterwards), should I?

No need. We will do it in a separate PR.

arikfr:
:-)

PR #4161, opened 2019-09-21
Support for dropdown of predefined options in data sources setup

gabrieldutra:

- [x] Feature

## Description
This adds support to use `Select` in Data sources/Destinations setup.

- [x] Rendering in frontend
- [x] Check if ath is missing backend-wise

--

ds-predefined-options


redash/query_runner/impala_ds.py
```diff
@@ -47,8 +47,12 @@ def configuration_schema(cls):
                     "type": "number"
                 },
                 "protocol": {
-                    "type": "string",
-                    "title": "Please specify beeswax or hiveserver2"
+                    "type": "enum",
+                    "options": [
+                        {"value": "beeswax"},
+                        {"value": "hiveserver2"}
+                    ],
```
arikfr: This won't pass schema validation. A valid enum is:

```json
{
    "protocol": {
        "type": "string",
        "enum": ["beeswax", "hiveserver2"]
    }
}
```

We might want to add our own extension to provide friendly names for options:

```
{
    "protocol": {
        "type": "string",
        "extendedEnum": [{"value", "beeswax", "name": "Beeswax"}, {"value": "hiveserver2", "name": "Hive Server 2"}]
    }
}
```

And we will transform `extendedEnum` into `enum` before passing it into `jsonschema.validate` call.

gabrieldutra: >We might want to add our own extension to provide friendly names for options

That's exactly why I went with the list of objects on that, but yeah, it does make sense to permit a simplified version.

> And we will transform extendedEnum into enum before passing it into jsonschema.validate call.

That's the only thing missing in terms of backend code?

gabrieldutra:
> Will it pick some value by default? I think think it should pick either the default value as default and if no default is specified, then it should pick the first value.

For now only if you specify a default value, for other cases I went with the placeholder. First value is an option too, I'll update that.

arikfr:
> For now only if you specify a default value, for other cases I went with the placeholder. First value is an option too, I'll update that.

Because it will fail validation if you don't specify a value from the list it's best if we just pick one for you.

redash/query_runner/impala_ds.py
arikfr: > That's the only thing missing in terms of backend code?

I think so.

redash/utils/configuration.py
```diff
@@ -24,6 +24,11 @@ def __init__(self, config, schema=None):
         self.set_schema(schema)
 
     def set_schema(self, schema):
+        if isinstance(schema, dict):
+            for prop in schema.get('properties', {}).values():
+                if 'extendedEnum' in prop:
+                    prop['enum'] = map(lambda v : v['value'], prop['extendedEnum'])
+                    del prop['extendedEnum']
```
gabrieldutra: @arikfr (or @rauchy)

This seemed the best place to put the transformation extendedEnum -> enum logic (so `ConfigurationContainer` stays in sync with what it's expected from jsonschema).

One note is that this modifies the schema in the upper scope (for the DS/Destination handler it didn't seem that big of a deal), so lmk if there is a dict "deep copy" that I should be using in such circumstances.

redash/utils/configuration.py
```diff
@@ -24,6 +24,11 @@ def __init__(self, config, schema=None):
         self.set_schema(schema)
 
     def set_schema(self, schema):
+        if isinstance(schema, dict):
```
arikfr: This is not needed: `schema` is always a dict.

redash/utils/configuration.py
arikfr: > One note is that this modifies the schema in the upper scope

:thinking: I wonder if we might end up in a situation where it returns the modified version in an API call. Several options:

1. Deep copy.
2. Keep `extendedEnum` and change the UI to use it first and ignore `enum` if `extendedEnum` exists.

redash/utils/configuration.py
gabrieldutra: I was getting a "`NoneType` has no attribute..." And that's why I put it 😬

redash/utils/configuration.py
gabrieldutra: There is already a preference to `extendedEnum` in UI, just need to keep it there then 😉

gabrieldutra: Deep copy seemed more future-proof, so I went with it in the end :+1:

arikfr:
:+1:

PR #4166, opened 2019-09-22
List enabled Query Runner types during build 

arikfr:

- [x] Feature

## Description

Add CLI command to list enabled Query Runner types (`manage ds list_types`) + update build configuration to list them during build.

### DIFF
--- a/redash/cli/data_sources.py
+++ b/redash/cli/data_sources.py
@@ -35,6 +35,15 @@ def list(organization=None):
             ds.id, ds.name, ds.type, ds.options.to_json()))
 
 
+@manager.command()
+def list_types():
+    print ("Enabled Query Runners:")
+    types = sorted(query_runners.keys())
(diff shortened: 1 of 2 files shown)


redash/cli/data_sources.py
```diff
@@ -35,6 +35,15 @@ def list(organization=None):
             ds.id, ds.name, ds.type, ds.options.to_json()))
 
 
+@manager.command()
+def list_types():
+    print ("Enabled Query Runners:")
```
rauchy: The. Space. Before. The. Parentheses. Is. Killing. Me.

arikfr: Fixed. e9892b6

arikfr:
It's not very useful because we don't build the image with data source dependencies at this stage :man_facepalming: Might revise it a bit in a future PR.

PR #4170, opened 2019-09-22
Alert redesign #4 - custom notification template

ranbena:
- [x] Refactor
- [x] Feature

## Description
Adds the previously hidden feature of custom subject/body (https://github.com/getredash/redash/pull/3137) for alert notifications.

Now added to Alert edit mode:

Screen Shot 2019-09-23 at 7 17 33

Screen Shot 2019-09-23 at 7 19 26

Switched to preview mode:

Screen Shot 2019-09-23 at 7 19 42

And in Alert page view mode:

Screen Shot 2019-09-23 at 7 17 48

How it looks on webmail:

Screen Shot 2019-09-24 at 11 36 41

How it looks on slack:

Screen Shot 2019-09-24 at 11 50 06


ranbena:
@arikfr @rauchy I'd like to test the python changes to templating in this PR - for instance, get an email alert notification to mail or slack. Is that possible?

redash/models/__init__.py
```diff
@@ -821,20 +821,33 @@ def evaluate(self):
...
+            QUERY_RESULT_ROWS: data['rows'],
+            QUERY_RESULT_COLS: data['columns'],
+        }
+        return mustache_render(template, context)
 
     @property
-    def template(self):
-        return self.options.get('template', '')
+    def custom_body(self):
+        return self.render_template(self.options.custom_body);
```
```suggestion
        return self.render_template(self.options.custom_body)
```

redash/models/__init__.py
```diff
@@ -821,20 +821,33 @@ def evaluate(self):
...
 
     @property
     def custom_subject(self):
-        return self.options.get('subject', '')
+        return self.render_template(self.options.custom_subject);
```
```suggestion
        return self.render_template(self.options.custom_subject)
```

redash/models/__init__.py
```diff
@@ -821,20 +821,33 @@ def evaluate(self):
...
+        context = {
+            ALERT_NAME: self.name,
+            ALERT_URL: '{host}/alerts/{alert_id}'.format(host=host, alert_id=self.id),
+            ALERT_STATUS: self.state,
+            ALERT_EVALUATION_TIME: datetime.datetime.now(), # could be problematic cuz utc?
+            ALERT_CONDITION: self.options['op'],
+            ALERT_THRESHOLD: self.options['value'],
+            QUERY_NAME: self.query_rel.name,
+            QUERY_URL: '{host}/queries/{query_id}'.format(host=host, query_id=self.query_rel.id),
+            QUERY_RESULT_VALUE: data['rows'][0][self.options['value']],
```
arikfr: 1. This is unsafe as `rows` might be empty or it might not have the the specified column. 
2. `self.options['value']` is not the column name.

arikfr:
> Is that possible?

Locally?

redash/models/__init__.py
ranbena: Fixed in https://github.com/getredash/redash/pull/4170/commits/7bbe27f45345398fae7f27165134f75be32e0e59

ranbena:
> Locally?

Yes. If not, maybe in deploy preview somehow?

client/app/pages/alert/components/NotificationTemplate.jsx
```diff
@@ -0,0 +1,123 @@
...
+    ALERT_THRESHOLD: alert.options.value,
+    ALERT_NAME: alert.name,
+    ALERT_URL: `${window.location.origin}/alerts/${alert.id}`,
+    QUERY_NAME: query.name,
+    QUERY_URL: `${window.location.origin}/queries/${query.id}`,
+    QUERY_RESULT_VALUE: topValue,
+    QUERY_RESULT_ROWS: resultValues,
+    QUERY_RESULT_COLS: columnNames,
+  };
+}
```
ranbena: @arikfr this is a duplicate of the server side implementation and that's not good practice.
Wdyt of making an api call that sends a template string and returns the rendered string.

arikfr: I noticed this too, but decided not to burden you with more work :) 

Ironically, in the original implementation there was an API call, but then we switched to Mustache so we can render it in the client... But I agree that we should have an endpoint for it.

rauchy:
> Is that possible?

@ranbena #4173 will help you with this.

client/app/pages/alert/components/NotificationTemplate.jsx
ranbena: I'll leave as is and open an issue once this is merged.

redash/models/__init__.py
```diff
@@ -821,20 +821,40 @@ def evaluate(self):
...
+            'QUERY_RESULT_ROWS': data['rows'],
+            'QUERY_RESULT_COLS': data['columns'],
+        }
+        return mustache_render(template, context)
 
     @property
-    def template(self):
-        return self.options.get('template', '')
+    def custom_body(self):
+        return self.render_template(self.options['custom_body'])
```
kravets-levko: Could it be an issue in terms of backward compatibility (replacing `template` option with `custom_body` - and for subject as well - what if somebody already has alerts with custom )?

Also - shouldn't it be `self.options.get('custom_body', '')` as before?

arikfr: > Could it be an issue in terms of backward compatibility

There are several backward incompatible changes in here (like the template variables). Because it's hidden behind a feature flag, I doubt many people use it. 

But to be friendlier to them I suggest we keep the option name as `template`? As for the variables, we will document the change in the release notes.

ranbena: > Could it be an issue in terms of backward compatibility

This feature was off since it landed, so I'm assuming that chances are slim that someone changed `clientConfig.extendedAlertOptions` apart for the original author @k-tomoyasu.

Maybe I can do this?:
```py
# backwards compatibility
def template(self):
  return self.custom_body;
```

> Also - shouldn't it be self.options.get('custom_body', '') as before?

`render_template()` now returns `''` if `custom_body` us empty. Don't think it's enough?

kravets-levko: `custom_body` feels a better name for this option than `template` - so if you think it's not an issue - I'd prefer new names as well.

About template variables: do we introduce them (or some new ones), or rename/remove some existing?

ranbena: > About template variables: do we introduce them (or some new ones), or rename/remove some existing?

I renamed the original 3:
https://github.com/getredash/redash/blob/cbfd25f9d7cc7c159170f11b81b984d663641f1b/client/app/services/alert-template.js#L6-L10

arikfr: > `render_template()` now returns `''` if `custom_body` us empty. Don't think it's enough?

I think @kravets-levko was referring to the case when `self.options` doesn't have the `custom_body` key (every existing alert). I address this in my example below. If you access `self.options['custom_body']` (or `custom_subject`) elsewhere in the code, you need to add similar treatment.

> Maybe I can do this?:

This won't help. What might happen is that a user will have an existing alert object with the `template` key in `options`, while you look for `custom_body`. To make it backward compatible you can change the `custom_body` property method into something like:

```python
 def custom_body(self):
        template = self.options.get('custom_body', self.options.get('template'))
        return self.render_template(template)
```

This will look in the `custom_body` key but if not found will give a try at looking in `template`. If not found it will return `None`, so you need to update `render_template` as well (to return empty if it's `None`).

Btw, shouldn't the configuration key be `custom_body_template`?

client/app/components/proptypes.js
```diff
@@ -93,10 +93,45 @@ export const Destination = PropTypes.shape({
...
 });
 
+export const Query = PropTypes.shape({
+  id: PropTypes.number.isRequired,
+  name: PropTypes.string.isRequired,
+  description: PropTypes.string,
+  data_source_id: PropTypes.number.isRequired,
+  created_at: PropTypes.string.isRequired,
+  updated_at: PropTypes.string,
+  user: UserProfile,
```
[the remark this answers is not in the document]
ranbena: You think we could just revert it back to snake case?

redash/models/__init__.py
ranbena: @arikfr does this suffice? https://github.com/getredash/redash/pull/4170/commits/5f8634acc92d73cb5d35b8c38e9154c4d17a8f0e

client/app/components/proptypes.js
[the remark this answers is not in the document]
gabrieldutra: Yes, it's way better to have it unified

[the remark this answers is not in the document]
ranbena: It's trickier than I hoped. Not sure how to get around the linter without conversion..
Since I'm using it only for `user.id` in this Alert feature, it doesn't matter. Ok to leave as is?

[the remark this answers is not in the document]
gabrieldutra: No problem, I also thought it would be easier :+1:

redash/models/__init__.py
```diff
@@ -821,20 +821,42 @@ def evaluate(self):
...
         data = json_loads(self.query_rel.latest_query_data.data)
-        context = {'rows': data['rows'], 'cols': data['columns'], 'state': self.state}
-        return mustache_render(self.template, context)
+        host = base_url(self.query_rel.org)
+
+        col_name = self.options['column']
+        if data['rows'] and col_name in data['rows'][0]:
+            result_value = data['rows'][0][col_name]
+        else:
+            result_value = ''
```
```suggestion
            result_value = None
```

`None` feels like a better fallback value in such case.

redash/models/__init__.py
```diff
@@ -821,20 +821,42 @@ def evaluate(self):
     def subscribers(self):
         return User.query.join(AlertSubscription).filter(AlertSubscription.alert == self)
 
-    def render_template(self):
-        if not self.template:
+    def render_template(self, template=None):
```
```suggestion
    def render_template(self, template):
```
Why do we need a default value for `template`?

PR #4174, opened 2019-09-23
Fix Pivot Visualization should not be saving data

gabrieldutra:

- [x] Bug Fix

## Description
Pivot visualization should not be saving data, otherwise it stays unchanged

Pending:
- [x] Test this again on Netlify Preview
- [x] See if a simple test can be created (can be in another PR depending on how quickly this needs to go)

--

**Issue**
pivot-issue

**After**
pivot-fixed

### DIFF
--- a/client/cypress/integration/visualizations/pivot_spec.js
+++ b/client/cypress/integration/visualizations/pivot_spec.js
@@ -19,6 +19,22 @@ const SQL = `
   SELECT 'c' AS stage1, 'c4' AS stage2, 44 AS value
 `;
 
+function createPivotThroughUI(visualizationName, options = {}) {
+  cy.getByTestId('NewVisualization').click();
+  cy.getByTestId('VisualizationType').click();
+  cy.getByTestId('VisualizationType.PIVOT').click();
+  cy.getByTestId('VisualizationName').clear().type(visualizationName);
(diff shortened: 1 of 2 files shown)


client/cypress/integration/visualizations/pivot_spec.js
```diff
@@ -19,6 +19,22 @@ const SQL = `
   SELECT 'c' AS stage1, 'c4' AS stage2, 44 AS value
 `;
 
+function createPivotThroughUI(visualizationName, options = {}) {
```
gabrieldutra: I was going to use UI, but eventually realized that the API was needed. This function is a leftover from it and it's here because it seemed more organized than before.

client/cypress/integration/visualizations/pivot_spec.js
```diff
@@ -73,6 +74,36 @@ describe('Pivot', () => {
     });
   });
 
+  it('updates the visualization when results change', function () {
+    const options = {
+      aggregatorName: 'Count',
+      data: [], // force it to have a data object, although it shouldn't
```
gabrieldutra: That was the way I found to test what will happen in UI when the object has a 'data'.
It's impracticable to use the UI to configure Pivot Table (the entire thing is drag&drop).

PR #4176, opened 2019-09-23
Fix Execute Selected not working for dirty queries

gabrieldutra:

- [x] Bug Fix

## Description
The Execute Selected was not working when the query wasn't in a "dirty state".

Pending:
- [x] Test on Netlify (and make gifs)
- [x] ~Cypress test(?)~ Not that easy to select a text inside Ace Editor in Cypress :confused:

--

**Before**
execute-selected

**After**
execute-selected-fix

### DIFF
--- a/client/app/pages/queries/view.js
+++ b/client/app/pages/queries/view.js
@@ -1,4 +1,4 @@
-import { pick, some, find, minBy, map, intersection, isArray, omit } from 'lodash';
+import { pick, some, find, minBy, map, intersection, isEmpty, isArray, omit } from 'lodash';
 import { SCHEMA_NOT_SUPPORTED, SCHEMA_LOAD_ERROR } from '@/services/data-source';
 import getTags from '@/services/getTags';
 import { policy } from '@/services/policy';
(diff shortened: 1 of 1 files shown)


client/app/pages/queries/view.js
```diff
@@ -42,7 +42,7 @@ function QueryViewCtrl(
     }
 
     $scope.showLog = false;
-    if ($scope.isDirty) {
+    if ($scope.isDirty || !isEmpty(selectedQueryText)) {
```
ranbena: Why is `isDirty` being checked here actually?

arikfr: If the query text wasn't changed we use the results API that takes just the query id (the safe API which requires only read access). If the text was changed, we use the results API that takes SQL (requires full access).

ranbena:
@gabrieldutra maybe this could help with ace/cypress https://github.com/cypress-io/cypress/issues/1818#issuecomment-500792891

gabrieldutra:
@ranbena thanks, that helps with Ace Editor usages in code (avoids having to to use `{ force: true }`), but the bigger issue is to select the text in Ace. I think I'll give a last try with mouse events :+1:

ranbena:
@gabrieldutra hmm right. Not sure it's worth the effort.

PR #4179, opened 2019-09-24
Fix: Alert trigger condition

ranbena:
- [x] Bug Fix

## Description
PR https://github.com/getredash/redash/pull/4134 introduced new alert condition operator logic.
But it seems to have reversed the trigger condition.


ranbena:
I got confused. Unit tests proved worthy. Forget it  😔

PR #4153, opened 2019-09-19
Alert page - migrate to React and redesign

ranbena:
This is the first part of a handful of PRs.

- [x] Refactor
- [x] Feature

## Description
1. Migration to React
2. Front end redesign status, criteria, rearm, edit/view modes and more.

## Feature list

Create alert
Screen Shot 2019-09-19 at 19 36 20

And select query
Screen Shot 2019-09-19 at 19 36 46

View mode
Screen Shot 2019-09-19 at 19 37 58

&nbsp;|&nbsp;
------|------
Statuses | Screen Shot 2019-09-19 at 19 38 23 Screen Shot 2019-09-19 at 19 39 20 Screen Shot 2019-09-19 at 19 39 00
Query edit | Screen Shot 2019-09-19 at 19 41 23
Query view | Screen Shot 2019-09-19 at 19 40 01
No sched | Screen Shot 2019-09-19 at 19 40 18
Criteria edit | Screen Shot 2019-09-19 at 19 41 57
Data errors | Screen Shot 2019-09-19 at 19 43 08 Screen Shot 2019-09-19 at 19 43 21
Criteria view | Screen Shot 2019-09-19 at 19 40 39
Rearm edit | Screen Shot 2019-09-19 at 19 43 46
At most every ... | Screen Shot 2019-09-19 at 19 44 05
Rearm view | Screen Shot 2019-09-19 at 19 40 46
Save / cancel / Delete | Screen Shot 2019-09-19 at 19 44 46

More features:
* Users with no edit permissions on specific alert get the view version always.
* Criteria allows comparing strings with "equals".

PR #4183, opened 2019-09-24
Separated Alert to View / Edit / New page components

ranbena:
- [x] Refactor

## Description
By request in [comment](https://github.com/getredash/redash/pull/4153/files?file-filters%5B%5D=.js&file-filters%5B%5D=.jsx&file-filters%5B%5D=.less#r327473471) for separation of mega file, now Alert.jsx renders:

```jsx
<div className="container alert-page">
  {mode === MODES.NEW && <AlertNew {...commonProps} />}
  {mode === MODES.VIEW && <AlertView canEdit={canEdit} {...commonProps} />}
  {mode === MODES.EDIT && <AlertEdit {...commonProps} />}
</div>
```

Also, made abstracted the alert name component into:
```jsx
<Title alert={alert} name={name} onChange={onNameChange} editMode />
```

### DIFF
--- a/client/app/pages/alert/Alert.jsx
+++ b/client/app/pages/alert/Alert.jsx
@@ -68,6 +68,11 @@ class AlertPage extends React.Component {
           // force view mode if can't edit
           if (!canEdit) {
             this.setState({ mode: MODES.VIEW });
+            notification.warn(
+              'You cannot edit this alert',
+              'You do not have sufficient permissions to edit this alert, and have been redirected to the view-only page.',
+              { duration: 0 },
+            );
           }
 
           this.setState({ alert, canEdit, pendingRearm: alert.rearm });
--- a/client/app/pages/alert/Alert.jsx
+++ b/client/app/pages/alert/Alert.jsx
@@ -1,8 +1,6 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import { react2angular } from 'react2angular';
-import { head, includes, template as templateBuilder, trim } from 'lodash';
-import cx from 'classnames';
+import { head, includes, trim, template } from 'lodash';
 
 import { $route } from '@/services/ng';
 import { currentUser } from '@/services/auth';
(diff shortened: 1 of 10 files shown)


client/app/pages/alert/AlertEdit.jsx
```diff
@@ -0,0 +1,139 @@
...
+      </>
+    );
+  }
+}
+
+AlertEdit.propTypes = {
+  alert: AlertType.isRequired,
+  queryResult: PropTypes.object, // eslint-disable-line react/forbid-prop-types,
+  pendingRearm: PropTypes.number,
+  delete: PropTypes.func.isRequired,
```
ranbena: I keep breaking Github @kravets-levko :)

client/app/pages/alert/Alert.jsx
```diff
@@ -383,14 +194,20 @@ export default function init(ngModule) {
   ngModule.component('alertPage', react2angular(AlertPage));
 
   return routesToAngularRoutes([
+    {
+      path: '/alerts/new',
+      title: 'New Alert',
+      mode: MODES.NEW,
```
arikfr: Why not have each path render its own component? What's the purpose of having a single top level page component? To share the callback functions?

ranbena: Yes. Less code duplication
Also allows `alert`, `query` and `queryResult` to be props instead of state, simplifying the page components.

ranbena: Also allows this, which imo is more elegant than redirection to view page:
https://github.com/getredash/redash/blob/9aeccb4d8fc251bc6110441539c7a7415ab038ae/client/app/pages/alert/Alert.jsx#L68-L71

arikfr: You could use composition instead:

```
<AlertPage ...>
    <EditAlertPage ...>
</AlertPage>
```

*But* the ability to change states without making API request is a win. In the long run, we should find a way to do this without as this pattern isn't always applicable. But here it's fine.

ranbena: > _But_ the ability to change states without making API request is a win.

Are you sure that's happening here? Changing Alert states mandates url changes which causes page to reload (`navigateTo`).

arikfr:
Oops, I misunderstood your previous comment. I thought this was what
happens when you switch between view and edit.

I don't think that just rendering the view page when you hit the edit url
is the thing to do. At the very least we should show a message saying you
don't have permission to edit.

But we could do all that with composition as well. No need for a single
component. No need to change it now though, it's not important enough and
better to focus on other aspects of the alert PRs.

ranbena:
> I don't think that just rendering the view page when you hit the edit url is the thing to do.

That's not the case. The edit button doesn't get rendered if the user can't edit the alert.
It's for the off-chance that an editing user shares an alert url with a non-editing user.

client/app/pages/alert/Alert.jsx
ranbena: @arikfr it was an easy modification https://github.com/getredash/redash/pull/4183/commits/e8654600daff0c4f6e9a08af017315a482130bb4

arikfr:
>  It's for the off-chance that an editing user shares an alert url with a non-editing user.

I know and still I think it has the potential to confuse people who did expect to edit the alert using this URL. Even showing a notification saying something like "You do not have permission to edit this Alert." is probably enough. Not a huge priority.

ranbena:
Was also an ez change https://github.com/getredash/redash/pull/4183/commits/eecb6d397938b77ee5b4eaf2034720a1d2057f6d

As long as you're ok with the prompt being a notification (cause it's easy):

Screen Shot 2019-09-26 at 15 11 23

And instead of hiding action buttons they're disabled with a tooltip:

Screen Shot 2019-09-26 at 15 32 21

arikfr:
Merged.

PR #4186, opened 2019-09-25
Fix query based dropdown not adding quote marks correctly

### LINKED ISSUE #4184: Query with a drop down list based on query (Allow multiple values with quotation mark) not adding quotation mark correctly
### Issue Summary

Query with a drop down list based on query (Allow multiple values with quotation mark) not adding quotation mark correctly.

### Steps to Reproduce

1. Create a query with a parameter with query based dropdown list.
2. Select allow multiple values, double quotation mark.
3. Execute query, when the parameter select only 1 choice and the parameter is text, it didn't add double quotation mark and the query returns an error.

### Technical details:

* Redash Version: 8.0 Beta 2
* Browser/OS: CentOS 7

gabrieldutra:

- [x] Bug Fix

## Description
This is not perfect yet, I still wanna explore a few more options without having the Dropdown Options in the Parameter structure context (that's the only difference from the regular Dropdown code - what handles if the value is valid is the Input).

Fixes #4184

--

### DIFF
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -254,7 +254,7 @@ export class Parameter {
     }
 
     if (this.type === 'query' && !isEmptyValue(value)) {
-      if (this.multiValuesOptions && !isArray(value) && value !== null) {
+      if (this.multiValuesOptions && !isArray(value)) {
         value = [value];
       }
     }
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -253,6 +253,12 @@ export class Parameter {
       }
     }
 
+    if (this.type === 'query' && !isEmptyValue(value)) {
+      if (this.multiValuesOptions && !isArray(value) && value !== null) {
+        value = [value];
+      }
+    }
+
     if (isDateRangeParameter(this.type)) {
       this.value = null;
       this.$$value = null;
@@ -387,7 +393,8 @@ export class Parameter {
       if (has(query, key)) {
         if (this.multiValuesOptions) {
           try {
-            this.setValue(JSON.parse(query[key]));
+            const valueFromJson = JSON.parse(query[key]);
+            this.setValue(isArray(valueFromJson) ? valueFromJson : query[key]);
           } catch (e) {
             this.setValue(query[key]);
           }
--- a/client/cypress/integration/query/parameter_spec.js
+++ b/client/cypress/integration/query/parameter_spec.js
@@ -108,7 +108,7 @@ describe('Parameter', () => {
   describe('Dropdown Parameter', () => {
     beforeEach(() => {
       const queryData = {
-        name: 'Number Parameter',
+        name: 'Dropdown Parameter',
         query: "SELECT '{{test-parameter}}' AS parameter",
         options: {
           parameters: [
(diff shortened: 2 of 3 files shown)


client/app/services/query.js
```diff
@@ -387,7 +393,8 @@ export class Parameter {
       if (has(query, key)) {
         if (this.multiValuesOptions) {
           try {
-            this.setValue(JSON.parse(query[key]));
+            const valueFromJson = JSON.parse(query[key]);
+            this.setValue(isArray(valueFromJson) ? valueFromJson : query[key]);
```
gabrieldutra: Turns out numbers are also parsed and converted :sweat_smile:, so this makes sure we don't convert it unless it's an array.

client/app/services/query.js
```diff
@@ -253,6 +253,12 @@ export class Parameter {
       }
     }
 
+    if (this.type === 'query' && !isEmptyValue(value)) {
+      if (this.multiValuesOptions && !isArray(value) && value !== null) {
+        value = [value];
```
gabrieldutra: QueryBasedDropdown with multi-values should always hold array values in its structure.

client/cypress/integration/query/parameter_spec.js
```diff
@@ -180,6 +180,69 @@ describe('Parameter', () => {
...
+            ],
+          },
+        };
+
+        cy.visit(`/queries/${dropdownQuery.id}`);
+        cy.getByTestId('ExecuteButton').click();
+        cy.getByTestId('TableVisualization')
+          .should('contain', 'value1')
+          .and('contain', 'value2')
+          .and('contain', 'value3');
```
ranbena: Better move this to a test of its own? (I don't think setup is a right place for assertions)

client/app/services/query.js
```diff
@@ -253,6 +253,12 @@ export class Parameter {
       }
     }
 
+    if (this.type === 'query' && !isEmptyValue(value)) {
+      if (this.multiValuesOptions && !isArray(value) && value !== null) {
```
ranbena: Don't think you need to check `value !== null` cause it's checked the line before.

client/cypress/integration/query/parameter_spec.js
```diff
@@ -180,6 +180,69 @@ describe('Parameter', () => {
...
+      cy.getByTestId('QueryEditor').click(); // just to close the select menu
+
+      cy.getByTestId('ParameterApplyButton')
+        .click();
+
+      cy.getByTestId('TableVisualization')
+        .should('contain', '"1","2","3"');
+    });
+  });
+
```
ranbena: Will you add a #4184 specific test?

client/cypress/integration/query/parameter_spec.js
gabrieldutra: Considering Cypress style I don't see any problems on assertions in the setup (they are quite fast, new tests are not). But in this case I wanted to make sure the query was executed and had results (needed for the Dropdown), so the assertion is indeed part of the setup.

client/cypress/integration/query/parameter_spec.js
gabrieldutra: I thought about it, but a general functionality test seemed more efficient as we cover more cases. In any case, this test fails in master and passes in here.

client/app/services/query.js
gabrieldutra: Yeah, I was a bit paranoid on this one :sweat_smile:

arikfr:
:bug:

PR #4194, opened 2019-09-27
WIP - Add automatic self signed SSL by default to setup

amunchet:

- [X] Feature

## Description
Redash's main strength is sharing of queries/reports between parties, technical and non-technical.  If self-hosted, a better security practice is to use SSL (even on an internal network).  Redash should support self-signed certificates out of the box as the bare minimum security.


arikfr:
Thank you for opening this, but I don't see a reason to have self signed
setup created by default when it's so simple[1] to use LetsEncrypt to get a
recognized certificate.

[1] https://gist.github.com/arikfr/64c9ff8d2f2b703d4e44fe9e45a7730e

amunchet:
Letsencrypt is wonderful when the Redash instance is accessible externally; however for my use case (and I imagine other environments), the internal database I wish to query and current network configuration do not allow external routes, making certificate generation/renewal impossible.  I suppose you could generate a certificate externally, then copy it internally, but that doesn't seem very friendly out of the box.  (DDNS isn't allowed for us via general policy, but that would probably work as well).

amunchet:
I guess the use case this solves is for a semi-local installation (by that I mean, installation on an internal network with more clients than just the installer's personal machine).  My situation was, I installed redash on one of our local servers, was quite excited to get up and querying, then realized I almost sent all our database information over cleartext, which would have landed me in hot water with the security folks.  This PR is trying to address that particular "gotcha" and make initial installation have at least that minimum of security.

arikfr:
Thanks for the explanation. I see why LetsEncrypt won't work in your case, but I believe it's the less common case. 

How about we add a page that documents how to create self signed certificates to the documentation instead of having it as the default setup?

amunchet:
Sounds good - I think the scope of the PR has changed significantly, so I'll go ahead and close this one.  Thanks!

PR #4175, opened 2019-09-23
Migrate Table visualization to React Part 2: Editor

kravets-levko:
## What type of PR is this? (check all applicable)

- [x] Refactor

## Description

- [x] Convert component to React
  - [x] Columns tab
    - [x] common options
    - [x] options specific to each column type
    - [x] drag-and-drop column sorting
  - [x] Grid tab
- [x] Debounce inputs
- [x] Cleanup
- [x] Tests

## Related Tickets & Documents

getredash/redash#3301 (Migrate Visualizations to React -> Table)

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

image
image

PR #4139, opened 2019-09-13
Migrate Chart visualization to React Part 2: Editor

kravets-levko:
## What type of PR is this? (check all applicable)

- [x] Refactor

## Description

- [x] Convert component to React
  - [x] General tab
  - [x] X-Axis tab
  - [x] Y-Axis tab
  - [x] Series tab
  - [x] Colors tab
    - [x] for Pie
    - [x] for Heatmap
    - [x] for other types
  - [x] Data Labels tab
  - [x] Custom chart settings
- [x] Refine & performance
  - [x] Debounce inputs
- [x] Cleanup
- [x] Tests

### To discuss:

- [x] Custom chart: `Auto update graph` options (seems it never worked)
  - Keep as is for now, fix (re-implement) later.
- [x] Are X and Y Axis tabs needed for Pie chart type?
  - Not needed, now are hidden when global series type is `pie`.
- [x] Use Antd icons for chart types (seems there are all icons we need)
  - Keep as is for now.

## Related Tickets & Documents

getredash/redash#3301 (Migrate Visualizations to React -> Chart)

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

There are some screenshots to show general look&feel. I didn't attach screenshots for all possible settings combinations.

<details>
  <summary>Click to expand</summary>

### General:

image

### X Axis:

image

#### All except of heatmap:

image

#### Heatmap:

image

### Series:

image
image

### Colors:

image

### Data Labels:

image

</details>

PR #4199, opened 2019-09-30
Sortable component

kravets-levko:

- [x] Refactor

## Description

Currently the only place where we used drag'n'drop sorting was `Parameters` component: we used `react-sortable-hoc` for DnD and also added some nice animation effects to it. Now we need the same for Chart (getredash/redash#4139) and Table (getredash/redash#4175) editors. This PR decouples that custom features from `Parameters` components and adds a basic implementation for "sortable" components which should help with repeatable operations (like maintaining "isDragged" state and corresponding classes, defining a default container, or ability to disable drag'n'drop).

Also, this PR:
- fixes minor UI bug on Parameters component (see screenshots).
- updates `react-sortable-hoc` (from `1.9.1` to `1.10.1`) - previous version had a bug (`sortableContainer` HOC was passing wrong properties to wrapped component) which is fixed in new version.

Before:
image

After:
image


client/app/components/sortable/index.js
```diff
@@ -0,0 +1,80 @@
...
+      'sortable-container',
+      { 'sortable-container-dragging': isDragging },
+      containerProps.className,
+    );
+    containerProps.ref = containerRef;
+  }
+
+  // order of props matters - we override some of them
+  return (
+    <SortableContainerWrapper {...wrapperProps}>
```
gabrieldutra: What usually annoys me is that GH doesn't recognize this when the extension is not jsx :grimacing:

client/app/components/sortable/index.js
```diff
@@ -0,0 +1,80 @@
...
+      setIsDragging(false);
+      if (isFunction(onSortEnd)) {
+        onSortEnd(...args);
+      }
+    });
+
+    // - update container element: add classes and take a ref
+    containerProps.className = cx(
+      'sortable-container',
+      { 'sortable-container-dragging': isDragging },
```
gabrieldutra: Why not a `data-dragging` modifier instead? I thought it was becoming a standard :slightly_smiling_face:

client/app/components/Parameters.less
```diff
@@ -1,32 +1,24 @@
...
-  vertical-align: bottom;
-  margin-right: 5px;
-  cursor: move;
-}
-
 .parameter-block {
   display: inline-block;
   background: white;
   padding: 0 12px 6px 0;
   vertical-align: top;
```
ranbena: Add `z-index: 1` so this don't happen:

Screen Shot 2019-09-30 at 17 33 25

client/app/components/sortable/index.js
ranbena: @gabrieldutra +1

client/app/components/sortable/index.js
kravets-levko: oh crap :grimacing: sorry, I'll fix this in a moment

client/app/components/sortable/index.js
kravets-levko: I think because we don't use `data-` attributes anywhere except of `data-test`. Also, IMHO class-based selectors are easier to read in CSS files

client/app/components/Parameters.less
kravets-levko: Done :+1:

client/app/components/sortable/index.js
kravets-levko: Done :+1:

client/app/components/sortable/index.js
ranbena: I use it all over the place  😬

Imo, as long as `data-prop1={condition || null}`, the selector is as coherent in the CSS

```css
[data-prop1] {
  /* rule */
}
```

kravets-levko: Well, I think we need some convention for this :grin: I personally don't like (and avoid to use) attribute-based selectors for few reasons:
- performance (all browsers have some optimizations for id/class-based selectors, but don't optimize tag/attribute based selectors)
- attributes selectors have a least specificity, so they're very easy to override (especially accidentally) by classes, but almost no way to override other rules by attribute-based selector. - (as a part of previous one) `class` attribute is a list and may contain multiple values (the next class overrides rules of previous. Data attributes selectors usually are used by 1) presense 2) exact value - so classes are more flexible.

I feel comfortable to use attributes for things that are standalone and "boolean" in their nature (like `disabled` property). In this case, we actually have two flags: `draggable` which defines some styles for draggable _container_ (another thing that feels wrong to me - `draggable` flag should be set for element which is draggable itself), and `dragged` which depends on and overrides some styles of `draggable` - so actually these two attributes were used as classes - when second one extends/overrides previous one.

Actually it's quite a good topic for discussion :thinking:

ranbena: I think these are all good points.
I personally have not encountered these potential downsides, but I appreciate the thoughtfulness and am ok with the approach.

gabrieldutra: :thinking: personally I prefer the data attributes, but I haven't taken the opportunity to answer your arguments yet.

We should find a convention, but this shouldn't block this PR.

PR #4201, opened 2019-09-30
Vertica: prevent overwriting row data when duplicated column names exist

rz269:
This pull request fixes issues with duplicated column names for Vertica query runner.

[fetch_columns](https://github.com/getredash/redash/blob/master/redash/query_runner/__init__.py#L101) method is used for this fix.

**Problem Description:**
If we have columns with the same name in the select statement then the query result is overridden by the value of first column. e.g. 
`select count(*)
    , count(distinct user_id)
    , count(user_id)
from example_table`

**Example:** 
Duplicated column names -- 2nd and 3rd columns' values were overwritten by the 1st column:
vertica_dup_column_name
A simple fix is to give each column a unique name:
vertica_dup_column_name2

### DIFF
--- a/redash/query_runner/vertica.py
+++ b/redash/query_runner/vertica.py
@@ -132,7 +132,7 @@ class Vertica(BaseSQLQueryRunner):
 
                 columns = self.fetch_columns(column_data)
                 rows = [dict(zip(([c['name'] for c in columns]), r))
-                        for i, r in enumerate(cursor.fetchall())]
+                        for r in cursor.fetchall()]
 
                 data = {'columns': columns, 'rows': rows}
                 json_data = json_dumps(data)
--- a/redash/query_runner/vertica.py
+++ b/redash/query_runner/vertica.py
@@ -128,12 +128,11 @@ class Vertica(BaseSQLQueryRunner):
 
             # TODO - very similar to pg.py
             if cursor.description is not None:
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks :+1:

redash/query_runner/vertica.py
```diff
@@ -126,16 +126,12 @@ def run_query(self, query, user):
...
-                
-                # bug -- same name columns will always be assigned single last value
-                # corrected with fetch_columns in pull request https://github.com/getredash/redash/pull/4201
-                rows = [dict(zip((c[0] for c in columns_data), row)) for row in cursor.fetchall()]
-                columns = [{'name': col[0],
-                            'friendly_name': col[0],
-                            'type': types_map.get(col[1], None)} for col in columns_data]
+                columns_data = [(i[0], types_map.get(i[1], None)) for i in cursor.description]
+
+                columns = self.fetch_columns(column_data)
```
NicolasLM: Hi, I stumbled upon this commit while rebasing my work on master.
Tell me if I'm missing something but it looks like `column_data` is not defined anywhere. Shouldn't it be `columns_data`?

arikfr: :flushed: of course you're right.

arikfr: #4210

PR #4211, opened 2019-10-03
Fix dashboard parameter mapping issues

gabrieldutra:

- [x] Bug Fix

## Description
Bugs:
1. When changing the mapping type to Static without changing the value, the Field was set to `null` (even though there was a value in the parameter)
2. Field mapping sometimes does not update after changing it

To-do:
- [x] Test bugs on netlify + gifs
- [x] Write Cypress tests :tada:

--
### Issue 1
**Before**
mapping-static-issue
**After**
mapping-static-solved

### Issue 2
**Before**
mapping-update-issue

**After**
mapping-update-solved

### DIFF
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -1,6 +1,7 @@
 /* eslint-disable react/no-multi-comp */
 
-import { isString, extend, each, map, includes, findIndex, find, fromPairs, clone, isEmpty } from 'lodash';
+import { isString, extend, each, has, map, includes, findIndex, find,
+  fromPairs, clone, isEmpty } from 'lodash';
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
@@ -157,6 +158,13 @@ export class ParameterMappingInput extends React.Component {
     if (newMapping.value !== mapping.value) {
       newMapping.param = newMapping.param.clone();
       newMapping.param.setValue(newMapping.value);
+    }
+    if (has(update, 'type')) {
+      if (update.type === MappingType.StaticValue) {
+        newMapping.value = newMapping.param.value;
+      } else {
+        newMapping.value = null;
+      }
     }
     onChange(newMapping);
   };
--- a/client/app/components/ParameterMappingInput.jsx
+++ b/client/app/components/ParameterMappingInput.jsx
@@ -176,7 +176,7 @@ export class ParameterMappingInput extends React.Component {
         value={this.props.mapping.type}
         onChange={e => this.updateSourceType(e.target.value)}
       >
-        <Radio className="radio" value={MappingType.DashboardAddNew}>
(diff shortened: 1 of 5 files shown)


client/app/components/ParameterMappingInput.jsx
```diff
@@ -158,6 +159,13 @@ export class ParameterMappingInput extends React.Component {
       newMapping.param = newMapping.param.clone();
       newMapping.param.setValue(newMapping.value);
```
ranbena: Why doesn't this line already do the trick?
Cause in `StaticValue` `newMapping.value` is null?

Will #4049 fix this?

client/app/components/ParameterMappingInput.jsx
```diff
@@ -158,6 +159,13 @@ export class ParameterMappingInput extends React.Component {
...
     }
+    if (has(update, 'type')) {
```
ranbena: Is this an indication that the code is bad? When will there not be a `type`?

client/app/components/ParameterMappingInput.jsx
gabrieldutra: That like wasn't doing the trick for this case because the `value` wasn't actually changing, but the `type` was. From the little I understood in the code, only type that should have a `value` is the Static. I think #4049 won't solve it because it has to do with the "mapping" instructions (and those are placed in the widget structure?).

client/app/components/ParameterMappingInput.jsx
gabrieldutra: I think so it is an indication that code is bad, but because it's missing a treatment to the values when the type changes. It's like I mentioned, from my understanding the "value" is coupled to the Static type, and there's no function to handle such change.

There can be no type because I'm checking the updates, not the actual value :slightly_smiling_face:.

gabrieldutra: That's actually similar to changing the Parameter type without updating relevant properties.

ranbena: I misunderstood it. 10x for the clarification 👍

client/cypress/integration/dashboard/parameter_mapping_spec.js
```diff
@@ -0,0 +1,136 @@
...
+      .within(() => {
+        cy.contains('button', 'OK')
+          .click();
+      });
+
+    cy.contains('button', 'OK')
+      .click();
+  };
+
+  it('has widget mapping as the default', function () {
```
ranbena: The default mapping by API is widget level param (cause api-redash:addWidget doesn't send parameterMapping data), but in the UI (AddWidgetDialog) it's dashboard level param.

Dunno if this is an actual problem or not, but it did confuse me when I read the test.

gabrieldutra: Well noticed, I've never realized that :joy:. I'll rename the test for now.

PR #4224, opened 2019-10-07
Postgres: make sure table from the public schema doesn't get merged with table from other schemas

arikfr:

- [x] Bug Fix

## Description

This pull request addresses an edge case in Postgres' schema handling: 

By default we omit the public schema name from the table name. But there are edge cases, where this might cause conflicts. For example:

* We have a schema named "main" with table "users".
* We have a table named "main.users" in the public schema. 

(while this feels unlikely, this actually happened)

In this case if we omit the schema name for the public table, we will have a conflict. This pull reuqest makes sure that if such collision exists, we will name the public schema table as `public."main.users"` (quotes are needed because of the dot).

This is probably relevant to other data sources, but for now limiting the change to the Postgres data source to make the change smaller and simpler.

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -63,24 +63,29 @@ def _wait(conn, timeout=None):
             raise psycopg2.OperationalError("select.error received")
 
 
+def full_table_name(schema, name):
+    if '.' in name: 
+        name = u'"{}"'.format(name)
+    
+    return u'{}.{}'.format(schema, name)
+
+
 def build_schema(query_result, schema):
-    # By default we omit the public schema name from the table name. But there are 
+    # By default we omit the public schema name from the table name. But there are
     # edge cases, where this might cause conflicts. For example:
     # * We have a schema named "main" with table "users".
-    # * We have a table named "main.users" in the public schema. 
+    # * We have a table named "main.users" in the public schema.
     # (while this feels unlikely, this actually happened)
     # In this case if we omit the schema name for the public table, we will have
     # a conflict.
-
-    full_table_name = lambda r: u'{}.{}'.format(r['table_schema'], r['table_name'])
-    table_names = set(map(full_table_name, query_result['rows']))
+    table_names = set(map(lambda r: full_table_name(r['table_schema'], r['table_name']), query_result['rows']))
 
     for row in query_result['rows']:
         if row['table_schema'] != 'public':
-            table_name = full_table_name(row)
+            table_name = full_table_name(row['table_schema'], row['table_name'])
         else:
             if row['table_name'] in table_names:
-                table_name = u'{}."{}"'.format(row['table_schema'], row['table_name'])
+                table_name = full_table_name(row['table_schema'], row['table_name'])
             else:
                 table_name = row['table_name']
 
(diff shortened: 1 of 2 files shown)


redash/query_runner/pg.py
```diff
@@ -63,6 +63,38 @@ def _wait(conn, timeout=None):
...
 
+def full_table_name(schema, name):
+    if '.' in name: 
+        name = u'"{}"'.format(name)
+    
+    return u'{}.{}'.format(schema, name)
+
+
+def build_schema(query_result, schema):
+    # By default we omit the public schema name from the table name. But there are
```
rauchy: Can you convert this to a heredoc or simply link to this PR's description?

redash/query_runner/pg.py
```diff
@@ -63,6 +63,38 @@ def _wait(conn, timeout=None):
...
+    # edge cases, where this might cause conflicts. For example:
+    # * We have a schema named "main" with table "users".
+    # * We have a table named "main.users" in the public schema.
+    # (while this feels unlikely, this actually happened)
+    # In this case if we omit the schema name for the public table, we will have
+    # a conflict.
+    table_names = set(map(lambda r: full_table_name(r['table_schema'], r['table_name']), query_result['rows']))
```
rauchy: IIRC, as part of the Python 3 PR, there's a massive conversion of maps to comprehensions. While we can argue on which is nicer (hey, I'm a map guy!), I do feel should consider a consolidation of styles.

arikfr: @NicolasLM is this something `2to3` does on its own? 

Until `python-3` hits master branch and as we merge new code, I would like to reduce the amount of manual changes you need to do. So if there are anything like this that you need to manually update, please let us know so we are mindful of it.

redash/query_runner/pg.py
```diff
@@ -63,6 +63,38 @@ def _wait(conn, timeout=None):
...
+
+    for row in query_result['rows']:
+        if row['table_schema'] != 'public':
+            table_name = full_table_name(row['table_schema'], row['table_name'])
+        else:
+            if row['table_name'] in table_names:
+                table_name = full_table_name(row['table_schema'], row['table_name'])
+            else:
+                table_name = row['table_name']
```
rauchy: Since this is an edge case, maybe it'll be best if it has less distraction?

```suggestion
        table_name = row['table_name']
                
        if row['table_schema'] != 'public' or table_name in table_names:
            table_name = full_table_name(row['table_schema'], row['table_name'])                
```

arikfr: I remember having some issue with doing it this way, but looking at it now it seems legit. :man_shrugging:

redash/query_runner/pg.py
NicolasLM: Yes, 2to3 replaces maps with list comprehensions (but only when the first argument given to map is a lambda). This is because in Python 3 `map` returns a lazy iterator while in Python 2.7 it returns a list, so 2to3 tries to preserve the return type. A similar thing happens with `dict.keys()` or `dict.values()`.

In this specific example it does not matter since everything is wrapped in a set just after, so it will work in both Python 2.7 and Python 3.

I don't think you need to bother too much about this until the `python-3` branch is merged as these incompatibilities are usually caught early on in tests after a rebase.

PR #4173, opened 2019-09-23
Add maildev to the dev stack

rauchy:
## What type of PR is this? (check all applicable)
- [x] Feature

## Description
[Maildev](https://github.com/djfarrelly/MailDev) adds a simple interface for receiving e-mail in dev mode. This PR adds maildev to the development Docker stack and configures the app to use it. E-mails can be found at `http://localhost:1080`.

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
image

PR #4225, opened 2019-10-07
Fix: Failing alert test

ranbena:
- [x] Bug Fix

## Description
Fixes alert [test failure](https://dashboard.cypress.io/#/projects/924cka/runs/1542/failures).

### The bug
When email server is configured in Redash, user creation api omits `invite_link` which the test was dependent on.
I assume this succeeded in the feature branch tests cause it wasn't up to date with https://github.com/getredash/redash/pull/4173.

### The fix
Added `no_invite` to api request and now the `invite_link` does appear and test continues.
Also added clear error if this prop doesn't show up in the future.

### DIFF
--- a/client/cypress/support/redash-api/index.js
+++ b/client/cypress/support/redash-api/index.js
@@ -106,6 +106,7 @@ export function createUser({ name, email, password }) {
     method: 'POST',
     url: 'api/users',
     body: { name, email },
+    qs: { no_invite: 1 },
     failOnStatusCode: false,
   })
     .then((xhr) => {
--- a/client/cypress/support/redash-api/index.js
+++ b/client/cypress/support/redash-api/index.js
@@ -123,12 +123,17 @@ export function createUser({ name, email, password }) {
       const id = get(body, 'id');
       assert.isDefined(id, 'User api call returns user id');
 
-      return cy.request({
-        url: body.invite_link,
-        method: 'POST',
-        form: true,
-        body: { password },
-      });
+      if (body.is_invitation_pending) {
+        const url = get(body, 'invite_link');
+        assert.isDefined(url, 'User api call returns user invite link');
+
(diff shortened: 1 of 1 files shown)


arikfr:
> I assume this succeeded in the feature branch tests cause it wasn't up to date with #4173.

This is correct, but not for the reason you think: the default setting in #4173 broke the "is email server configured" test :face_with_head_bandage: I will fix this.

ranbena:
@arikfr #4226 is important but not a replacement - this commit makes the test independent of the environment setting, therefore I think a must.

arikfr:
This commit is using a bug that will be patched in a follow up pull
request. Redash should only send the invite url if there is no email server
configured.

If you want to make the test independent of the env we have two options:

1. Make the behavior of whether to send invite_url have its own setting so
we can turn it on explicitly in E2E tests.
2. Make users creation in E2E tests not rely on this (by directly updating
the db?).

ranbena:
How about returning `invite_link` in both cases?

https://github.com/getredash/redash/blob/3b7efb8c1fc2cb8ca16a623300062f73bbb3bbf3/redash/handlers/users.py#L44-L50

arikfr:
What both cases?

ranbena:
If email was sent or not.

```python
invite_url = invite_link_for_user(user) 
if settings.email_server_is_configured() and send_email: 
  send_invite_email(inviter, user, invite_url, org) 
 
d['invite_link'] = invite_url   
return d 
```

arikfr:
You missed the point: with the way things implemented right now (using an invite link validates email address), we shouldn't be sharing the invite URL if email server is enabled. 

But I thought about it some more and I think that what we can do is:

1. Stop validating email address when using an invite URL and always send a validation email after creating the user.
2. Once 1 is implemented, we can always return the invite URL.

Issue #982, opened 2016-04-12
Add an "unarchive" button on archived dashboards


jeffwidman:
If someone accidentally clicks "archive", it'd be nice to be able to undo that from the UI

benmanns:
For people who find this by Googling, here's how I un-archive via updating the DB:

```sql
UPDATE dashboards
SET is_archived = false
WHERE slug = 'mycooldashboard';
```

gclsoft:
@benmanns How to update the postgress DB in the docker?
```
root@ubuntu:/home/g/redash# docker exec -it -u 0 b1a2f4fb93e8  bash
bash-4.3# top
Mem: 2909540K used, 140736K free, 22432K shrd, 350008K buff, 936752K cached
CPU:   0% usr   0% sys   0% nic 100% idle   0% io   0% irq   0% sirq
Load average: 0.08 0.11 0.24 2/463 66
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
   29     1 postgres S     165m   5%   1   0% postgres: postgres postgres 172.19.0.5(33074) idle
   26     1 postgres S     165m   5%   0   0% postgres: postgres postgres 172.19.0.5(33056) idle
   28     1 postgres S     162m   5%   1   0% postgres: postgres postgres 172.19.0.5(33070) idle
   27     1 postgres S     162m   5%   0   0% postgres: postgres postgres 172.19.0.5(33066) idle
   55     1 postgres S     161m   5%   0   0% postgres: postgres postgres 172.19.0.4(33380) idle
   56     1 postgres S     161m   5%   1   0% postgres: postgres postgres 172.19.0.4(33390) idle
   21     1 postgres S     161m   5%   0   0% postgres: autovacuum launcher process   
    1     0 postgres S     160m   5%   0   0% postgres
   19     1 postgres S     160m   5%   0   0% postgres: writer process   
   18     1 postgres S     160m   5%   0   0% postgres: checkpointer process   
   20     1 postgres S     160m   5%   1   0% postgres: wal writer process   
   22     1 postgres S    19664   1%   1   0% postgres: stats collector process   
   61     0 root     S     6212   0%   0   0% bash
   66    61 root     R     1532   0%   0   0% top
bash-4.3# psql
psql: FATAL:  role "root" does not exist
bash-4.3# exit
exit
root@ubuntu:/home/g/redash# docker exec -it -u 0 b1a2f4fb93e8  bash
bash-4.3# sudo psql
[...]

PR #4228, opened 2019-10-08
Migrate Dashboard and Public Dashboard to React

### LINKED ISSUE #2076: Refactoring: merge `dashboard` and `public-dashboard`

### LINKED ISSUE #4204: Dashboard filters disappear after changing dashboard parameter values
### Issue Summary

My guess is that this was introduced with #3804 

### Steps to Reproduce

1. Create a dashboard with a widget containing a filter and another widget containing a parameter (or use an [existing one](http://localhost:5000/dashboard/test));
2. Map the Widget Parameter to be a Dashboard Parameter and enable Dashboard level filters;
3. Change the Parameter value and verify the filter options disappear

dashboard-filters-parameter

### Technical details:

* Redash Version: Latest
* Browser/OS: --
* How did you install Redash: --

### LINKED ISSUE #4341: Widgets API is called when loading a dashboard
### Issue Summary

On some dashboards it happens more often than other, but I noticed that in some cases a POST request is made to some of the widgets on a dashboard when loading it.

Here's an example: https://redash-preview.netlify.com/dashboard/now-param

### Technical details:

* Redash Version: v9
* Browser/OS: Chrome

gabrieldutra:

- [x] Refactor

## Description
Migration of the Dashboard to React

**Refresh and Refresh rate**
Same presets kept, now updates URL accordingly with `refresh=[refreshRate]`.
No button on the Public Dashboard, but can be updated through the URL (as it was before).

refresh

**Fullscreen**
Now updates URL accordingly.
The button is still only visible for Published queries, lmk if you want that changed.

fullscreen

**Publish/Unpublish**
publish-unpublish

**Mobile**
- Dashboard Header is no longer Sticky for Mobile (there was an issue with it now that the Navbar is fixed). Only con I see is in case you want to refresh the dashboard you'll have to go all the way up.
- Refresh is now the only Button that appears in mobile (previously the "Publish" button would appear too)

  

**Edit Mode**

edit-mode

**Public Dashboard**
react-public-dashboard

**Features**
- [ ] Should it be possible to unarchive a dashboard? (#982 + #2180 :sweat_smile:)

**Bugs Found**
- [x] Bug :beetle:: Dashboard Level Filters not updating when it's changed.

Closes #2076 
Closes #4204
Closes #4341


client/app/pages/dashboards/useDashboard.js
```diff
@@ -18,11 +22,85 @@ function getRefreshRateFromUrl() {
...
+    document.querySelector('body').classList.toggle('headless', fullscreen);
+    updateUrlSearch('fullscreen', fullscreen ? true : null);
+  }, [fullscreen]);
+
+  const toggleFullscreen = () => setFullscreen(!fullscreen);
+  return [fullscreen, toggleFullscreen];
+}
+
+function useRefreshRateHandler(refreshDashboard, updateUrlSearch) {
+  const [refreshRate, setRefreshRate] = useState(getRefreshRateFromUrl());
```
ranbena: I like that this now allows custom refresh rates  🕺

ranbena:
I noticed if you're in edit mode and resize the viewport to a certain point - it won't disable the edit buttons but instead toggle back to non-edit mode. I think it's good, but is it on purpose?

gabrieldutra:
> I noticed if you're in edit mode and resize the viewport to a certain point - it won't disable the edit buttons but instead toggle back to non-edit mode. I think it's good, but is it on purpose?

It was intentional, but no special reason on my choice. Do you think it's better to keep the state saved?

kravets-levko:
@gabrieldutra `prepareWidgetsForDashboard` was needed for very old dashboards (in versions prior to 4 they were implemented as Bootstrap grid) - since there are no migration that updates dashboard grid between v3 and v4 - `prepareWidgetsForDashboard` was doing that job. Without it, users that will decide to upgrade from Redash v3 to latest (not sure if there are a lot of them) will get all their dashboards broken.

I know that it was an attempt to fix the issue with unnecessary API calls, but IMHO the solution should be different: there is a dirty-check mechanism for auto-save layout feature - this mechanism should be disabled in non-edit mode (as it was prior to Dashboard grid migration to React).

arikfr:
I agree with @kravets-levko -- we should keep the migration mechanism and implement the solution only for saves during non edits.

gabrieldutra:
Thanks @kravets-levko, I didn't know about that. Moved to option B to be closer to how this was done prior to the autosave: disabled layout saving when not in edit mode.

susodapop:
This passes the sanity check from my end-user perspective. I like the new button look.

gabrieldutra:
Thanks, @susodapop :slightly_smiling_face:

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -0,0 +1,383 @@
...
+  );
+}
+
+DashboardPageTitle.propTypes = {
+  dashboardOptions: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
+};
+
+function RefreshButton({ dashboardOptions }) {
+  const { refreshRate, setRefreshRate, refreshing, refreshDashboard } = dashboardOptions;
+  const refreshRateOptions = clientConfig.dashboardRefreshIntervals;
```
arikfr: It seems to be missing the use of a policy to define allowed refresh intervals:

https://github.com/getredash/redash/blob/263305214ebe8d3cca903579ae96d637154a6fd4/client/app/pages/dashboards/dashboard.js#L116-L121

gabrieldutra: :thinking:, currently it's still possible to put any interval you want through the url, should it be reflected in there?

arikfr: I think it was possible before as well, but we can also make sure that the value in the url is not lower than the minimum value the policy allows. Something like:

```
const refreshRate = max(urlValue, lowestPolicyValue);
```

gabrieldutra: Yeah, my currently was meant to be `master` version, but cool, will go with it :+1:

client/app/pages/dashboards/useDashboard.js
```diff
@@ -0,0 +1,367 @@
...
+  const editModeHandler = useEditModeHandler(!gridDisabled && canEditDashboard, dashboard.widgets);
+
+  useEffect(() => {
+    setDashboard(dashboardData);
+    loadDashboard();
+  }, [dashboardData]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  useEffect(() => {
+    document.title = dashboard.name;
+  }, [dashboard.name]);
```
kravets-levko: :+1:

arikfr:
:+1:

PR #4230, opened 2019-10-10
Fix test_outdated_queries_works_scheduled_queries_tracker (use utcnow)

arikfr:
This should fix the last failing test. 

I'm not sure why the other tests in this group used `utcnow` correctly, but this one wasn't. Probably because it didn't fail on this error in Python 2.


NicolasLM:
Thank you for trying, using `utcnow` fixes the comparison between TZ aware and TZ naive datetimes, but it fails later on the actual assert.

Also I rebased my branch on `master`, so this PR would need to be updated to reflect that.

arikfr:
I guess I was too optimistic :-) I will setup Python 3 env locally and properly review this now.

arikfr:
@NicolasLM see #4232 with an actually passing test :) (also based on latest version of `python-3`)

PR #4231, opened 2019-10-10
Avoid depending on app context when templating failure reports

rauchy:
- [x] Bug Fix

## Description
Failure reports use Jinja for templating, but they really don't need anything from the app's context but the templating directories, available in the jinja environment. This PR reduces the dependency, allowing the reports to run freely outside context.


redash/tasks/failure_report.py
```diff
@@ -50,9 +50,10 @@ def send_failure_report(user_id):
             'base_url': base_url(user.org)
         }
 
-        html = render_template('emails/failures.html', **context)
-        text = render_template('emails/failures.txt', **context)
         subject = "Redash failed to execute {} of your scheduled queries".format(len(unique_errors.keys()))
+        html, text = [
+            current_app.jinja_env.get_template('emails/failures.{}'.format(f)).render(**context) for f in ['html', 'txt']]
```
arikfr: Worth extracting a `render_template` function with some docstring that explains why we do it this way, just in case:

1. Someone decided to "clean" this up in the future.
2. Someone will need similar functionality in the future.

PR #4232, opened 2019-10-10
(Python 3) Use Redis' decode_responses=True option

arikfr:
This is replacing #4230 with an actual fix: the reason for the test failure was that the key wasn't a string representation of the query id, but a bytes one.

To avoid similar issues in the future, I changed the Redis connection creation to use the `decode_responses` option.

### DIFF
--- a/redash/settings/helpers.py
+++ b/redash/settings/helpers.py
@@ -1,4 +1,5 @@
 import os
+from urllib.parse import urlparse, urlunparse
 
 
 def fix_assets_path(path):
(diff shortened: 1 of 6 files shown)


redash/settings/helpers.py
```diff
@@ -34,3 +35,13 @@ def int_or_none(value):
         return value
 
     return int(value)
+
+
+def fix_redis_url(url):
```
arikfr: I should rename this to be more descriptive (`add_decode_responses_option`) and make sure it doesn't add the option twice (?).

arikfr: https://github.com/getredash/redash/pull/4232/commits/1ca1268b13efe714b1be42f6aa57902024a1ac59

NicolasLM:
That's great, thank you.

PR #4233, opened 2019-10-10
Multiprocess RQ workers

rauchy:
- [x] Refactor

## Description
[Created this as a separate PR from the RQ one, because that one is already too bloated to spot anything]

This is a naive way to launch multiple worker processes, thus utilizing containers more efficiently. It relies on a healthcheck being performed continuously - when the `rq healthcheck` command returns 1, it means that at least one of the workers in that host is not healthy* and the container should be replaced.

* A healthy worker is either currently busy (processing a job). If it is not busy, it is still considered healthy if it sent a heartbeat in the past 60 seconds. If it is neither busy or sent a heartbeat in the past 60 seconds, it is healthy if all of the queues it watches are empty (i.e. it has nothing to do).

I won't be surprised if we choose to replace this approach rather quickly, but this is the simplest way I can think of to get started and collect feedback.


bin/docker-entrypoint
```diff
@@ -23,9 +23,15 @@ dev_scheduler() {
...
+  WORKERS_COUNT=${WORKERS_COUNT:-2}
+  echo "Starting $WORKERS_COUNT RQ workers for queues $QUEUES..."
+
+  for i in `seq 1 $WORKERS_COUNT`;
+  do
+    exec /app/manage.py rq worker $QUEUES &
+  done    
 
-  exec /app/manage.py rq worker $QUEUES
+  sleep infinity
```
arikfr: I wonder if it will be more robust/cleaner to use something like [Honcho](https://github.com/nickstenning/honcho). This way one worker exiting/crashing won't kill the others.

And to clarify: I don't mean to use the Honcho CLI, but rather use its `Manager` and `Process` classes from our code directly, skipping the need to define create a `Procfile`. We can also implement our own logic, but I guess Honcho took care of many of the possible issues with managing multiple processes.

jezdez: +1 for Honcho!

rauchy: Using Honcho directly in 0d95d24. I like it because it will kill all workers and exit if one of the worker processes goes AWOL, but it feels a tad strange to specify a command to run within the code 🤷‍♂

arikfr: > it feels a tad strange to specify a command to run within the code

I agree. I sort of expected that we can provide a function to invoke as the subprocess, but it makes sense that their code assumes you provide them an executable path.

> I like it because it will kill all workers and exit if one of the worker processes goes AWOL

Better option would be to just restart it, no?

(btw, you need to rebase :grimacing:)

rauchy: > Better option would be to just restart it, no?
Yeah, it does that now.

> (btw, you need to rebase 😬)
That. Was. Fun.

arikfr: Now it restarts all of them instead of only the faulty one. Between the previous version (process exits when one or more workers misbehave) and current one (all workers restart in a loop when one of the workers misbehave), I think the safer one is the first.

While the end result is somewhat the same, it leaves more room for seeing what's going on. For example with ECS, we will see tasks restarting frequently.

rauchy: If we want to restart _only_ the faulty workers, we probably need to move towards something like supervisor (which was what we initially had in mind, but the purpose of this PR was to provide something lighter 😉)

rauchy: Anyway, reverted the auto-restart for now.

redash/cli/rq.py
```diff
@@ -19,9 +21,27 @@ def scheduler():
...
+    sys.exit(m.returncode)
+
+
 @manager.command()
 @argument('queues', nargs=-1)
 def worker(queues='default'):
+    # Configure any SQLAlchemy mappers loaded until now so that the mapping configuration 
+    # will already be available to the forked work horses and they won't need 
+    # to spend valuable time re-doing that on every fork.
+    configure_mappers()
```
arikfr: We should probably move this to its own PR, so it's not forgotten in case we abandon this one.

rauchy: 🙊https://github.com/getredash/redash/pull/4314

arikfr: Oops :-)

rauchy:
Running some load tests, it feels like #4371 is the direction we want to go with for multi-process workers, so I'm closing this PR for now.

PR #4235, opened 2019-10-10
Fixed alert destination hrefs

### LINKED ISSUE #4234: URLs on new alert page are broken when MULTI_ORG=True
### Issue Summary

All the URLs on the Alert page are missing the org_slug because they use the form `<a href="/destinations"...` instead of `<a href="destinations"...`. I've noticed this with the destinations and query links, but maybe there are others.

ranbena:
- [x] Bug Fix

## Description
Fixes #4234


arikfr:
A similar fix needed for the query link.

PR #4239, opened 2019-10-12
pagerduty.py: Change default summary text

### LINKED ISSUE #4223: Change PagerDuty's default summary text
Currently PagerDuty's Alert destination default summary text uses the query id and name. We should change it to use the alert name as it's usually better explains what the alert is.

While #4153 implements ability to customize the summary text, it's good to have a saner default regardless.

(If #4153 is not merged before implementing, should be implemented based on its branch)

frextrite:

- [x] Refactor

## Description
Change is made to use alert name in PagerDuty's alert
destination default summary text instead of
query id and name

Closes #4223

### DIFF
--- a/redash/destinations/pagerduty.py
+++ b/redash/destinations/pagerduty.py
@@ -12,7 +12,7 @@ except ImportError:
 class PagerDuty(BaseDestination):
 
     KEY_STRING = '{alert_id}_{query_id}'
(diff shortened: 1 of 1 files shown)


arikfr:
Thank you! :-)

frextrite:
Glad to be part of this! 🎉

PR #4243, opened 2019-10-14
Fix double-scrollbar when in fullscreen

stefanmaric:

- [x] Bug Fix

## Description

When using fullscreen mode inside an embedded report, a double scrollbar appears because padding is added to the `body` but the container is set to be `min-height: 100vh`. This PR fixes it.

### DIFF
--- a/client/app/assets/less/inc/base.less
+++ b/client/app/assets/less/inc/base.less
@@ -25,8 +25,10 @@ body {
     position: relative;
     
     &.headless {
-        padding-top: 10px;
-        
+        app-view {
+            padding-top: 10px;
+        }
+
         .nav.app-header, .navbar {
             display: none;
         }


client/app/assets/less/inc/base.less
```diff
@@ -26,7 +26,11 @@ body {
     
     &.headless {
         padding-top: 10px;
-        
+
+        app-view {
+          min-height: ~"calc(100vh - 10px)";
```
ranbena: I think this would be a simpler solution:

```less
&.headless {
  padding-top: 10px; // <- remove this line

  app-view {
     padding-top: 10px; // <- move it to here
  }
```

stefanmaric: Fixed.

ranbena:
Gracias @stefanmaric

stefanmaric:
> Gracias @stefanmaric

No hay de qué, @ranbena :smile:

PR #4245, opened 2019-10-15
App Header React migration

ranbena:
- [x] Refactor

### TODO 
- [x] React migration
- [x] Responsive layout
- [x] Mobile layout - now with fixed position and better UX
- [x] All tests pass

Screen Shot 2019-10-16 at 21 49 03

### Changes
1. No faves
Screen Shot 2019-10-15 at 9 32 44

2. Favorite lazy loading update

3. Create button options (added "New" text)

Screen Shot 2019-10-15 at 9 56 58

### Manually tested
- [x] Firefox, Safari, Chrome on Mac
- [x] Different viewport widths
- [x] Verified redundancy of deleted files
- [x] Verified permission defined rendering
- [x] Verified update available state


ranbena:
Thanks to @gabrieldutra realized that the mobile version is crappy.
Updated a very simple fixed position header with minimal dropdown menu.

Screen Shot 2019-10-15 at 17 38 02

arikfr:
Quick things I noticed:

1. The on-hover background color of menu items is hardly visible on my monitor.
2. When clicking on queries/dashboards in some cases it both navigates and open the dropdown menu. In the current version there is a clear distinction between the hit area for the dropdown and for the link.
3. In the current version when a menu open, the arrow points upwards.

client/app/components/app-header/AppHeader.jsx
```diff
@@ -0,0 +1,252 @@
...
+            </Menu>
+          )}
+        >
+          <Button type="primary" data-test="CreateButton">
+            Create <Icon type="down" />
+          </Button>
+        </Dropdown>
+      </div>
+      <div className="header-logo">
+        <a href={clientConfig.basePath}>
```
gabrieldutra:
```suggestion
        <a href="./">
```

It annoys me that this depends on a backend config and it doesn't fit our Redash-preview case, for example... I wonder if the above works for Multi-org :thinking:

client/app/components/app-header/components/FavoritesDropdown.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const noItems = isEmpty(items);
+  const urlCompiled = useMemo(() => template(urlTemplate), []);
+
+  const fetchItems = useCallback(() => {
+    setLoading(true);
+    fetch().$promise.then(({ results }) => {
+      setLoading(false);
+      setItems(results);
+    });
```
gabrieldutra: It worths using `.finally(() => setLoading(false))`, to avoid breaking in this case:

image

ranbena:
10x @arikfr 

> 1. The on-hover background color of menu items is hardly visible on my monitor.

(It's the default ant-design dropdown menu color that we have across the site)
I strengthened it now - lmk if suitable.

> 2. When clicking on queries/dashboards in some cases it both navigates and open the dropdown menu. In the current version there is a clear distinction between the hit area for the dropdown and for the link.

I had to do some cunning css to recreate the way it is in the current version.
Now made it more accurate - lmk if it still happens.

> 3. In the current version when a menu open, the arrow points upwards.

Personally, I don't see the point in it.
But recreated it now. Check it out.

client/app/components/app-header/AppHeader.jsx
ranbena: That actually works for multi-org!
Learned sth new today :)

client/app/components/app-header/components/FavoritesDropdown.jsx
ranbena: Very true. Fixed!

ranbena:
> For the desktop version I liked the blue bottom border and wdyt of having a black text color instead? (That gray feels de-emphasized to me)

@gabrieldutra I was waiting for @arikfr to say sth about it  😄
I kinda like the gray as it doesn't take the focus from the actual app content, but let's agree on #&#8203;333?

client/app/components/app-header/AppHeader.jsx
```diff
@@ -0,0 +1,252 @@
...
+          onSearch={onSearch}
+        />
+        <Menu mode="horizontal" selectable={false}>
+          <Menu.Item key="help">
+            <HelpTrigger type="HOME" className="menu-item-button" />
+          </Menu.Item>
+          {currentUser.isAdmin && (
+            <Menu.Item key="settings">
+              <a href="data_sources" className="menu-item-button">
+                <i className="fa fa-sliders" />
```
gabrieldutra: Considering the Help Trigger has a Tooltip on its own, wdyt about adding a Tooltip here as well? So will be a "standard" Icon Button behavior

gabrieldutra:
> I kinda like the gray as it doesn't take the focus from the actual app content, but let's agree on #​333?

It does look like something Arik would say :joy:. I prefer black, but `#333333` doesn't bother me.

client/app/components/app-header/AppHeader.jsx
ranbena: Done. Also converted it to `<Button>` so it has the cool click animation.

client/app/components/app-header/index.js
```diff
@@ -1,58 +0,0 @@
...
-import './app-header.css';
-
-const logger = debug('redash:appHeader');
-
-function controller($rootScope, $location, $route, $uibModal, Auth, currentUser, clientConfig, Dashboard, Query) {
-  this.logoUrl = logoUrl;
-  this.basePath = clientConfig.basePath;
-  this.currentUser = currentUser;
-  this.showQueriesMenu = currentUser.hasPermission('view_query');
-  this.showAlertsLink = currentUser.hasPermission('list_alerts');
```
arikfr: I noticed you didn't port this. Please make sure that you port the other permissions correctly.

client/app/components/app-header/index.js
```diff
@@ -1,58 +0,0 @@
...
-      this.dashboards = data.results;
-    });
-    Query.favorites().$promise.then((data) => {
-      this.queries = data.results;
-    });
-  };
-
-  this.reload();
-
-  $rootScope.$on('reloadFavorites', this.reload);
```
arikfr: I thought of suggesting of removing the `$broadcast` call for this event, but wasn't sure if it isn't use for other things. Maybe @kravets-levko knows.

ranbena:
> We might want to add some caching to favorites, so it doesn't fetch on every click.

@arikfr this was a way to disconnect from the `reloadFavorites` listener.
Can do caching but what would cache bust? (assuming time based is a bad idea)

client/app/components/app-header/index.js
kravets-levko: This event was needed to update favorites lists when user adds/removes favorite items. Also, `FavoritesControl` uses this event to force re-render - this is a ugly hack, hope we'll remove it soon after completing React migration.

arikfr:
> Can do caching but what would cache bust? (assuming time based is a bad idea)

Why? Some short time span (1 minute?) will help with making sure that repeated clicks on the menu don't require new data.

ranbena:
> Why? Some short time span (1 minute?) will help with making sure that repeated clicks on the menu don't require new data.

Cuz if a user adds or removes faves, the fave menu would be out of date until a minute has passed. I think that's no good.
Is the concern perf optimization? The UX is such that the fave list loads immediately and updates on server response (it's not optimal but it works).

arikfr:
> Is the concern perf optimization?

Of course. I didn't look too much into it, but in a single check I saw that one of the favs endpoint returned results in 300ms and the other in 700ms. This feels borderline acceptable, no? Specially for repeated presses.

The other option is a Favorties "service" class, that holds current state and has a simple method of `reset` which clears the cache. This can be called by the Favorites control where we currently invoke the event.

ranbena:
> Of course. I didn't look too much into it, but in a single check I saw that one of the favs endpoint returned results in 300ms and the other in 700ms. This feels borderline acceptable, no? Specially for repeated presses.

Borderline acceptable how? The user wait time is 0 on repeated presses.

> The other option is a Favorties "service" class, that holds current state and has a simple method of `reset` which clears the cache. This can be called by the Favorites control where we currently invoke the event.

Sure sounds good. Ok to implement this in a subsequent PR? (Don't want the PR to balloon 🎈)

ranbena:
Correction - there IS a wait, but only when there're no faves in the list.

client/app/components/app-header/AppHeader.jsx
```diff
@@ -0,0 +1,266 @@
...
+import { react2angular } from 'react2angular';
+
+import Dropdown from 'antd/lib/dropdown';
+import Button from 'antd/lib/button';
+import Icon from 'antd/lib/icon';
+import Menu from 'antd/lib/menu';
+import Input from 'antd/lib/input';
+import Tooltip from 'antd/lib/tooltip';
+
+import DropdownFavorites from './components/FavoritesDropdown';
```
arikfr: It should be `FavoritesDropdown`.

arikfr:
> Correction - there IS a wait, but only when there're no faves in the list.

Why?

If the values are memoized, then it's fine to leave as is.

ranbena:
> Why?

You're right I changed it now.

> If the values are memoized, then it's fine to leave as is.

👍

client/app/components/app-header/AppHeader.jsx
```diff
@@ -0,0 +1,266 @@
...
+import logoUrl from '@/assets/images/redash_icon_small.png';
+
+import './AppHeader.less';
+
+function onSearch(q) {
+  $location.path('/queries').search({ q });
+  $route.reload();
+}
+
+function Desktop() {
```
gabrieldutra: Although implicit (and it's not a blocker for this) I would prefer naming those `DesktopHeader` and `MobileHeader`.

client/app/components/app-header/AppHeader.jsx
```diff
@@ -0,0 +1,266 @@
...
+            <Menu.Item key="dashboards" className="dropdown-menu-item">
+              <Button onClick={() => navigateTo('dashboards')}>
+                Dashboards
+              </Button>
+              <FavoritesDropdown fetch={Dashboard.favorites} urlTemplate="dashboard/${slug}" />
+            </Menu.Item>
+          )}
+          {currentUser.hasPermission('view_query') && (
+            <Menu.Item key="queries" className="dropdown-menu-item">
+              <Button onClick={() => navigateTo('queries')}>
```
gabrieldutra: Why not `<Button href="queries">`? (Same for the other `navigateTo`'s), this way it allows opening in a new tab :)

arikfr: This is actually important.

ranbena: Done 👍

client/app/components/app-header/AppHeader.jsx
```diff
@@ -0,0 +1,259 @@
...
+                      <a href="admin/status">System Status</a>
+                    </Menu.Item>
+                  )}
+                  <Menu.Divider />
+                  <Menu.Item key="logout" onClick={() => Auth.logout()}>Log out</Menu.Item>
+                  <Menu.Divider />
+                  <Menu.Item key="version" disabled>
+                    Version: {clientConfig.version}
+                    {frontendVersion !== clientConfig.version && (
+                      <>{' '}({frontendVersion.substring(0, 8)})</>
```
kravets-levko:
```suggestion
                      {` (${frontendVersion.substring(0, 8)})`}
```

client/app/components/app-header/components/FavoritesDropdown.jsx
```diff
@@ -0,0 +1,69 @@
...
+import Menu from 'antd/lib/menu';
+
+import { HelpTrigger } from '@/components/HelpTrigger';
+
+export default function FavoritesDropdown({ fetch, urlTemplate }) {
+  const [items, setItems] = useState();
+  const [loading, setLoading] = useState(false);
+
+  const noItems = isEmpty(items);
+  const urlCompiled = useMemo(() => template(urlTemplate), []);
```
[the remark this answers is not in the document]
kravets-levko: Most likely `urlTemplate` and `fetch` will never change after app start, so it's not an issue. This fix will be just nice for code style. Also, probably similar needed for `fetch` (for the same purpose :slightly_smiling_face:)

client/app/components/app-header/AppHeader.jsx
kravets-levko: +1 for this :slightly_smiling_face:

client/app/components/app-header/app-header.html
```diff
@@ -1,234 +0,0 @@
...
-            >
-          </li>
-          <li role="menuitem"><a href="alerts/new">Alert</a></li>
-        </ul>
-      </div>
-
-      <!-- Profile -->
-      <ul class="nav navbar-nav navbar-right">
-        <li>
-          <help-trigger
```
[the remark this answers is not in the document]
ranbena: YES! Will do.

client/app/components/app-header/components/FavoritesDropdown.jsx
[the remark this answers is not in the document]
ranbena: That's exactly the reason I left it blank, cause it doesn't change but I'll commit :)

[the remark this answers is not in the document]
gabrieldutra: You never know what's gonna happen in the future, so it's always good to keep code "ready".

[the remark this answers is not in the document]
kravets-levko: Also, IMHO, when you use something in non-obvious way (like here when you omit a dependency for hook) - it's better to leave a comment for the future you reading that code :grin:

kravets-levko:
Linter crashes for a some reason :thinking:

PR #4250, opened 2019-10-16
Fix: Alert page breaks when target query returns null result

### LINKED ISSUE #4249: Alert page breaks when target query returns null result
### Issue Summary

I wrote a query that returns no data. If I click **Create** > **Alert** then choose this query, the page goes blank and the app requires a full reload to recover. My javascript console shows these errors immediately after the failure.

Screen Shot 2019-10-15 at 18 37 49

### Steps to Reproduce

1. Write a query that returns 0 rows of data. I queried the `users` table of Redash metadata `WHERE api_key = "this_is_not_a_real_api_key"`.
2. Click **Create** > **Alert** and then pick the query from step 1.

I understand why this doesn't work, but it should fail more gracefully.

### Technical details:

* Redash Version: Version 9
* Browser/OS: Firefox / Safari / Chrome

ranbena:
- [x] Bug Fix

## Description
Fixes #4249.

The new Alert doesn't assumes query result value.
Added condition in case result is an empty array.

Now empty data looks like this, instead of breaking:

Screen Shot 2019-10-16 at 10 27 04

### DIFF
--- a/client/app/pages/alert/components/Criteria.jsx
+++ b/client/app/pages/alert/components/Criteria.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { head, includes, toString } from 'lodash';
+import { head, includes, toString, isEmpty } from 'lodash';
 
 import Input from 'antd/lib/input';
 import Icon from 'antd/lib/icon';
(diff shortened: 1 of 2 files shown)


client/app/pages/alert/components/Criteria.jsx
```diff
@@ -34,7 +34,7 @@ DisabledInput.propTypes = {
 };
 
 export default function Criteria({ columnNames, resultValues, alertOptions, onChange, editMode }) {
-  const columnValue = resultValues && head(resultValues)[alertOptions.column];
+  const columnValue = !isEmpty(resultValues) ? head(resultValues)[alertOptions.column] : null;
```
arikfr: Why use `head` and not `resultValues[0]`?

arikfr: Why use `head` and not `resultValues[0]`?

susodapop:
When will this be merged into SaaS?

Issue #4181, opened 2019-09-24
Migrate to Python 3


arikfr:
As Python 2 reaches EOL, we're planning to have the next release of Redash to support Python 3.

Because Python 2 reaches EOL and the fact we don't have capacity to support both versions, once we switch to Python 3 *we will support Python 3 only*.

arikfr:
Related: #2891.
Previous attempt: https://github.com/getredash/redash/tree/py3-tests.

arikfr:
We will probably need to update some dependencies (either versions or replace) as outlined in #2891. But as we will be carefully testing everything, it's worth bumping the versions of other dependencies like moving to Flask 1.0+.

NicolasLM:
Hi, here is the summary of my progress on this issue for the week:

- The most obvious syntax changes have been addressed throughout the code base, mostly using 2to3
- The Dockerfile has been updated to use Python 3.7
- The frontend build in CircleCI has been changed to also use Python 3
- A few dependencies have been upgraded because they were not compatible with Python 3. I tried to avoid upgrading unnecessary dependencies since I want to isolate the issues related to Python 3 from more generic backward incompatibilities that are often introduced in new versions of a package.
- I am working my way through the tests, fixing the broken ones. The vast majority of them already pass.

Currently the web app and the tasks work fine for me on Python 3. I have not encountered any error using the application locally, but I have only tested the Postgres and MySQL datasources. Feel free to give it a try if you want, the code is in the `python-3` branch.

arikfr:
Thank you for the detailed update, @NicolasLM. :+1:

NicolasLM:
Only a single test remains broken: `test_outdated_queries_works_scheduled_queries_tracker`. After spending some time on it I feel that the migration to Python 3 uncovered a logic error in the test.
I am not sure which behavior is expected in the test as fixing the obvious comparison between TZ aware and naive datetime, makes the actual assert fail.

Besides that I also stopped using the CSV `UnicodeWriter`. I think that the clear distinction of `str` vs `bytes` in Python 3 makes this class redundant. If you could tell me why this class was needed in the first place I could verify that it is indeed not required anymore.

arikfr:
> Only a single test remains broken: `test_outdated_queries_works_scheduled_queries_tracker`. After spending some time on it I feel that the migration to Python 3 uncovered a logic error in the test.

I think you're right. I changed the test to use `redash.utils.utcnow` as the other tests in this group (#4230).

[...]

PR #4251, opened 2019-10-16
Migrate the application to Python 3

NicolasLM:

- [X] Refactor

## Description
Migrate the application to Python 3.

#4181


redash/query_runner/sqlite.py
```diff
@@ -81,12 +78,6 @@ def run_query(self, query, user):
             connection.cancel()
             error = "Query cancelled by user."
             json_data = None
-        except Exception as e:
-            # handle unicode error message
-            err_class = sys.exc_info()[1].__class__
-            err_args = [arg.decode('utf-8') for arg in sys.exc_info()[1].args]
-            unicode_err = err_class(*err_args)
-            reraise(unicode_err, None, sys.exc_info()[2])
```
jezdez: Previously this was catching all errors, even if they weren't UnicodeErrors, so I would suggest to keep this.

redash/utils/configuration.py
```diff
@@ -55,7 +55,7 @@ def to_json(self):
         return json_dumps(self._config, sort_keys=True)
 
     def iteritems(self):
-        return self._config.iteritems()
+        return iter(self._config.items())
```
jezdez: `self._config.items()` is already an iterator since it's a dict view object, so I think it may not needed.

requirements_bundles.txt
```diff
@@ -6,4 +6,3 @@
 # These can be removed when upgrading to Python 3.x
 importlib-metadata>=0.19  # remove when on 3.8
 importlib_resources==1.0.2  # remove when on 3.7
```
jezdez: I think his can be removed as well (and the import updated in `bin/bundle-extensions`) for the importlib.resources package.

requirements_dev.txt
```diff
@@ -1,12 +1,12 @@
-pytest==3.2.3
-pytest-cov==2.5.1
-coverage==4.0.3
+pytest==5.2.0
```
jezdez: ZOMG this makes me so happy! is it my birthday already!!?1! 🎉

requirements_dev.txt
```diff
@@ -1,12 +1,12 @@
...
+pytest-cov==2.7.1
+coverage==4.5.4
 mock==2.0.0
 
 # PyMongo and Athena dependencies are needed for some of the unit tests:
 # (this is not perfect and we should resolve this in a different way)
 pymongo[tls,srv]==3.6.1
-botocore>=1.12.239,<1.13.0
+botocore==1.12.241
```
jezdez: I thought we wanted to remove this hard pin?

redash/query_runner/sqlite.py
NicolasLM: I think this was not catching `UnicodeError` but rather catching everything, decoding content and reraising. I still think that this is not needed on Python 3.

requirements_bundles.txt
NicolasLM: This is actually used by the container building the frontend which uses the default Python version shipped with it (I think Python 3.5) so this is still needed for now.

requirements_dev.txt
NicolasLM: Done

redash/utils/configuration.py
NicolasLM: Right, done.

NicolasLM:
I upgraded most dependencies as requested in #4181, but the bump of Werkzeug from 0.11 to 0.16 breaks tests related to redirecting to the next url after login.

I've spent some time debugging the problem, but I didn't find the reason. Any idea?

requirements_bundles.txt
jezdez: Ah good point!

arikfr:
> I upgraded most dependencies as requested in #4181, but the bump of Werkzeug from 0.11 to 0.16 breaks tests related to redirecting to the next url after login.

See #4280, I included an explanation of what happened there.

redash/query_runner/sqlite.py
arikfr: I agree w/ @NicolasLM , this seems unnecessary anymore (it was supposed to handle error message *with* Unicode).

requirements_all_ds.txt
```diff
@@ -1,34 +1,34 @@
...
-snowflake_connector_python==1.8.6
+pydruid==0.5.5
+requests_aws_sign==0.1.5
+snowflake_connector_python==2.0.0
 phoenixdb==0.7
 # certifi is needed to support MongoDB and SSL:
-certifi
-pydgraph==1.2.0
-azure-kusto-data==0.0.32
+certifi==2019.9.11
```
arikfr:
```suggestion
certifi>=2019.9.11
```
This is the kind of dependencies that is probably safe to leave unversioned (added a lower boundary to appease linters).

requirements.txt
```diff
@@ -1,60 +1,59 @@
...
-simplejson==3.10.0
-ua-parser==0.7.3
-user-agents==1.1.0
+parsedatetime==2.4
+PyJWT==1.7.1
+cryptography==2.7
+simplejson==3.16.0
+ua-parser==0.8.0
+user-agents==2.0
 python-geoip-geolite2==2015.303
```
arikfr: Don't we need to replace this with the Python 3 version?

requirements.txt
```diff
@@ -1,60 +1,59 @@
...
+passlib==1.7.1
+aniso8601==8.0.0
+blinker==1.4
+psycopg2==2.8.3
 python-dateutil==2.8.0
-pytz==2016.7
-PyYAML==3.12
-redis==3.2.1
-requests==2.21.0
+pytz==2019.3
```
arikfr:
```suggestion
pytz>=2019.3
```

Probably safe to keep upgrading automatically.

redash/query_runner/dgraph.py
```diff
@@ -93,7 +93,7 @@ def run_query(self, query, user):
         try:
             data = self.run_dgraph_query_raw(query)
 
-            first_key = next(iter(data.keys()))
+            first_key = next(iter(list(data.keys())))
```
arikfr:
```suggestion
            first_key = next(data.keys())
```

Looks like the iter/list conversions are redundant.

arikfr: Apparently `iter` is needed.

requirements.txt
NicolasLM: Good catch, we should probably use `geoip2` instead.

NicolasLM: Done

jezdez:
@arikfr !!!! 🐍🚀

PR #3697, opened 2019-04-10
Bye footer

ranbena:
## What type of PR is this?

- [ ] Refactor
- [x] Feature ?
- [ ] Bug Fix
- [ ] New Query Runner (Data Source)
- [ ] New Alert Destination
- [ ] Other

## Description
Since https://github.com/getredash/redash/pull/3458 the footer has no purpose. So we're killing it.
Alternatives to "lost content" are [discussed here](https://discuss.redash.io/t/footer-advocacy/3599) for future reference.

## Mobile & Desktop Screenshots

It's not appropriate to take snapshots of the dead.

PR #4252, opened 2019-10-16
Restore <body> bottom padding

kravets-levko:

- [x] Bug Fix

## Description

Restore `<body>` bottom padding accidentally removed in getredash/redash#3697

### DIFF
--- a/client/app/assets/less/inc/base.less
+++ b/client/app/assets/less/inc/base.less
@@ -23,10 +23,18 @@ body {
     background: #F6F8F9;
     font-family: @redash-font;
     position: relative;
-    
+
+    app-view {
+      padding-bottom: 15px;
+    }
+
     &.headless {
         padding-top: 10px;
-        
+
+        app-view {
+          padding-bottom: 0;
+        }
+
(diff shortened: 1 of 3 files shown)


client/app/assets/less/inc/base.less
```diff
@@ -23,10 +23,18 @@ body {
...
+    app-view {
+      padding-bottom: 15px;
+    }
+
     &.headless {
         padding-top: 10px;
-        
+
+        app-view {
+          padding-bottom: 0;
```
ranbena: Why is it needed?

kravets-levko: Previously in headless mode padding was removed as well - I just restored that behavior

ranbena: Perhaps it has to do with the iframe version not having a needless vertical scroll or sth

kravets-levko:
Percy caught few more issues with this fix. Need to check once more

ranbena:
I sent you a PR with some fixes

kravets-levko:
@ranbena Thanks a lot!

PR #4254, opened 2019-10-16
Give query case sensitive treatment in query hash

### LINKED ISSUE #2137: Case insensitive parameters in Redash query result cache

osule:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Generating the query hash from the query text with no lowercasing of the query text
allows case-sensitive parameter values in the dashboard to have different cache entries.

Fixes #2137

### DIFF
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -51,14 +51,14 @@ def slugify(s):
 
 def gen_query_hash(sql):
     """Return hash of the given query after stripping all comments, line breaks
-    and multiple spaces, and lower casing all text.
+    and multiple spaces.
 
-    TODO: possible issue - the following queries will get the same id:
+    The following queries will get different ids:
         1. SELECT 1 FROM table WHERE column='Value';
         2. SELECT 1 FROM table where column='value';
     """
     sql = COMMENTS_REGEX.sub("", sql)
-    sql = "".join(sql.split()).lower()
+    sql = "".join(sql.split())
     return hashlib.md5(sql.encode("utf-8")).hexdigest()
 
 
--- a/migrations/versions/1038c2174f5d_make_case_insensitive_hash_of_query_text.py
+++ b/migrations/versions/1038c2174f5d_make_case_insensitive_hash_of_query_text.py
@@ -1,0 +1,51 @@
+"""Make case insensitive hash of query text
+
+Revision ID: 1038c2174f5d
+Revises: fd4fc850d7ea
+Create Date: 2023-07-16 23:10:12.885949
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import table
+
+from redash.utils import gen_query_hash
+
+# revision identifiers, used by Alembic.
+revision = '1038c2174f5d'
+down_revision = 'fd4fc850d7ea'
+branch_labels = None
+depends_on = None
+
+
+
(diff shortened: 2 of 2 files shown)


redash/utils/__init__.py
```diff
@@ -63,7 +63,6 @@ def gen_query_hash(sql):
         2. SELECT 1 FROM table where column='value';
     """
     sql = COMMENTS_REGEX.sub("", sql)
-    sql = "".join(sql.split()).lower()
```
arikfr: 1. We need to keep the join/split behavior as this takes care of removing multiple whitespaces. 
2. Need to update the docstring to actually reflect reality :)

migrations/versions/9aafdb8de771_give_query_case_sensitive_treatment_in_query_hash.py
```diff
@@ -0,0 +1,64 @@
...
+
+    query_results = table(
+        'query_results',
+        sa.Column('id', sa.Integer, primary_key=True),
+        sa.Column('query', sa.Text),
+        sa.Column('query_hash', sa.String(length=10)))
+
+    conn = op.get_bind()
+    change_query_hash(conn, queries, query_text_to=str)
+    change_query_hash(conn, query_results, query_text_to=str)
```
arikfr: Updating the query results table can be expensive on busy deployments. Let's skip this. It will mean that some of the cached results will become irrelevant but it's a reasonable trade off.

redash/utils/__init__.py
osule: @arikfr Thanks for catching that. I shall perform an update to correct these

guidopetri:
@osule , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

justinclift:
Looking over this PR, the requested changes were made so it should be ok to approve.

Looks like it's not passing our CI tests though, so some tweaking seems like it's still needed.

osule:
@justinclift Thanks for trying to get this. 
I'll take a look at the information you have shared. I'll regenerate the migration against the latest version.

justinclift:
Thanks heaps for fixing that CI failure @osule.  Merging this now. :smile:

PR #4255, opened 2019-10-17
Migrate Sankey renderer to React

kravets-levko:

- [x] Refactor

## Description

- [x] Revisit code, refine if needed
- [x] Cleanup
- [x] Tests

getredash/redash#3301 (Migrate Visualizations to React -> Sankey)

No visual changes


client/app/visualizations/sankey/initSankey.js
```diff
@@ -0,0 +1,237 @@
...
+
+  return nodes;
+}
+
+function graph(data) {
+  const nodesDict = {};
+  const links = {};
+  const nodes = [];
+
+  // ANGULAR_REMOVE_ME $$ check is for Angular's internal properties
```
ranbena: Not redundant now?

client/cypress/integration/visualizations/sankey_sunburst_spec.js
```diff
@@ -35,6 +37,12 @@ describe('Sankey and Sunburst', () => {
     cy.getByTestId('VisualizationPreview').find('svg').should('exist');
     cy.getByTestId('EditVisualizationDialog').contains('button', 'Save').click();
     cy.getByTestId('QueryPageVisualizationTabs').contains('li', visualizationName).should('exist');
+
+    cy.getByTestId('QueryPageShowDataOnly').click();
+
+    // wait a bit before taking snapshot
+    cy.wait(1000); // eslint-disable-line cypress/no-unnecessary-waiting
+    cy.percySnapshot('Visualizations - Sunburst', { widths: [viewportWidth] });
```
ranbena: Perhaps you can skip the wait by instead of clicking "Show Data Only", you can `cy.visit('queries/${id}')` and then snap?

client/app/visualizations/sankey/initSankey.js
kravets-levko: No - sankey may still receive data from Angular part, so these properties may be present there. It will be safe to remove when we'll completely remove Angular itself.

client/cypress/integration/visualizations/sankey_sunburst_spec.js
kravets-levko: I added this timeout to wait for chart to load a query results data. "Show Data Only" is actually a link, so it should be completely equal to `cy.visit(...)`. But I may give it a try

kravets-levko: Seems there is no difference - `cy.visit` need `cy.wait` as well. I prefer to keep current variant because 1) it already works 2) it's a bit simpler :slightly_smiling_face:

PR #4256, opened 2019-10-17
Monitor general RQ queues (default, periodic and schemas)

rauchy:
- [x] Feature

## Description
This PR adds queue sizes for the general RQ queues (`default`, `periodic` and `schemas`) to `/status.json`.

image


redash/monitor.py
```diff
@@ -31,7 +31,8 @@ def get_queues():
     scheduled_queue_names = db.session.query(DataSource.scheduled_queue_name).distinct()
     query = db.session.execute(union_all(queue_names, scheduled_queue_names))
 
-    return ['celery'] + [row[0] for row in query]
+    general_rq_queues = map(lambda q: 'rq:queue:' + q, ['default', 'periodic', 'schemas'])
```
arikfr: Why the `rq:queue:` prefix? 

Also, is there a way to ask `rq` for all the active queues?

arikfr: Oh, is it because that's the Redis key?

rauchy: Yupp

rauchy: > Also, is there a way to ask rq for all the active queues?

Yeah, I refactored it and now it reports all queues (on the way, it made sense to remove the `rq:queue:` prefix due to not poking directly in Redis anymore).

redash/monitor.py
```diff
@@ -35,14 +35,8 @@ def get_queues():
...
-    queues = {}
-
-    for queue in get_queues():
-        queues[queue] = {
-            'size': redis_connection.llen(queue)
-        }
-
-    return queues
+    return dict({queue: {'size': redis_connection.llen(queue)} for queue in get_celery_queues()}.items() +
+                {queue.name: {'size': len(queue)} for queue in Queue.all(connection=redis_connection)}.items())
```
arikfr:
```suggestion
                {queue.name: {'size': queue.count} for queue in Queue.all(connection=redis_connection)}.items())
```

Probably better to use `Queue.count` here.

arikfr: Just realized you're using Python's `len` and not `llen`. :man_facepalming: I guess it's my cue to go to sleep. :sleeping:

redash/monitor.py
```diff
@@ -26,7 +26,7 @@ def get_object_counts():
     return status
 
 
-def get_queues():
+def get_celery_queues():
```
arikfr: This function is also used in `celery_tasks()`, so need to update the name there as well.

rauchy: D'oh!

PR #4257, opened 2019-10-17
Reduce RQ status load time

rauchy:
- [x] Refactor

## Description
The RQ job status page was loading job information for all queued jobs, but for those, we only care about the amount. This PR fixes this.

Running a small benchmark - for 60k queued jobs, this PR has reduced the load time from approx 34 seconds to 2 seconds.


redash/monitor.py
```diff
@@ -155,7 +155,7 @@ def rq_queues():
         q.name: {
             'name': q.name,
             'started': fetch_jobs(q, StartedJobRegistry(queue=q).get_job_ids()),
-            'queued': fetch_jobs(q, q.job_ids)
+            'queued': len(q.job_ids)
```
arikfr: Does `q.job_ids` include `queued` and `started`?

rauchy: RQ does an `LPOP` when it starts executing a job so `q.job_ids` won't include started ones (instead, they are moved to `rq:wip:...`, which is observed by the `StartedJobRegistry`).

Issue #4187, opened 2019-09-26
Deduplicate Alert notification templating implemenation


ranbena:
Alert notification templating was introduced in #4170.
The feature itself is implemented in the backend
https://github.com/getredash/redash/blob/fbb46c9a7674303bff4168c1241c3f88ed0a61fd/redash/models/__init__.py#L837-L849

And the UI shows a preview in the frontend
https://github.com/getredash/redash/blob/fbb46c9a7674303bff4168c1241c3f88ed0a61fd/client/app/pages/alert/components/NotificationTemplate.jsx#L20-L31
https://github.com/getredash/redash/blob/fbb46c9a7674303bff4168c1241c3f88ed0a61fd/client/app/pages/alert/components/NotificationTemplate.jsx#L41

### Problem 
Results in duplicate code and risks them being out of sync.

### Proposal
Instead of the JS implementation, create an api endpoint that receives alertId + template string, and responds with the rendered template.

PR #4258, opened 2019-10-17
Fix: Alert condition rendering

ranbena:
- [x] Bug Fix

## Description
Since https://github.com/getredash/redash/pull/4240 alert conditions are no longer represented by a textual description but by chars. This caused templating (notification template and alert default name) to render differently.

This is a fix for that.

_|Before Fix|After Fix
------|------|------
Alert Edit|Screen Shot 2019-10-17 at 14 01 22|Screen Shot 2019-10-17 at 13 59 40
Email|Screen Shot 2019-10-17 at 14 03 45|Screen Shot 2019-10-17 at 14 03 52

(Yes, I know there's server-client code duplication https://github.com/getredash/redash/issues/4187)

### DIFF
--- a/client/app/pages/alert/Alert.jsx
+++ b/client/app/pages/alert/Alert.jsx
@@ -13,6 +13,7 @@ import LoadingState from '@/components/items-list/components/LoadingState';
 import AlertView from './AlertView';
 import AlertEdit from './AlertEdit';
 import AlertNew from './AlertNew';
+import { getConditionText } from './components/Criteria';
 
 import Modal from 'antd/lib/modal';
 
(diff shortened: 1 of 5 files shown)


arikfr:
I'm not sure this is worth the extra complexity. Does it feel that important?

ranbena:
> I'm not sure this is worth the extra complexity. Does it feel that important?

Was thinking that while working on it but I'm undecided. Up to you @arikfr.

ranbena:
Maybe we can settle for fixing only the alert name and leave out template rendering.

ranbena:
> I'm not sure this is worth the extra complexity. Does it feel that important?

Closing due to public disinterest.

PR #4266, opened 2019-10-18
Migrate PermissionsEditor to React

gabrieldutra:

- [x] Refactor

## Description
Migration of the PermissionsEditor to React.

- [x] Make it availble on the Dashboard Page
- [x] Get rid of the old code
- [x] Review UI/UX

--

react-permissions-editor


ranbena:
@gabrieldutra I took it for a spin and I don't think the two pane SelectItemsDialog is  optimal for this use case - it's confusing and unintuitive. In all other cases using SelectItemsDialog, the selected list and remove action are separate therefore more suitable.

I actually think the original layout is best and perhaps SelectItemsDialog can be fitted to be like this:

Screen Shot 2019-10-19 at 13 06 32

Notice I added:
* Title change suggestion
* Description under title
* Help trigger to [here](https://redash.io/help/user-guide/querying/writing-queries#Managing-Query-Permissions)

gabrieldutra:
Thanks @ranbena, I actually had the same thoughts and it was not easy to create a satisfactory flow with the SelectItemsDialog. I also thought about recreating the existing Dialog, but Antd's Select wouldn't support the Avatars.

I have another idea in mind to apply here that I'll explore later today (or tomorrow): Something similar to Google docs's Share Dialog, one dialog for View/Edit/Remove (at some point with a Select for the Permission Type, but for now we can use a Tag) and another SelectItemsDialog only for adding new users.

Regarding your suggestions the only one I have downside in mind is showing the Admins in the list, that can become confusing when there are many Admins, we better have that implicit :)

arikfr:
I like @ranbena's suggestion. I don't think SelectItemsDialog is a good fit here, because in some cases it's unfeasible to show a list of all users and it's better to let them search.

But I also agree about this one:

> Regarding your suggestions the only one I have downside in mind is showing the Admins in the list, that can become confusing and not objective when there are many Admins, we better have that implicit :)

We can add a note about admins having access instead.

ranbena:
> but Antd's Select wouldn't support the Avatars.

@gabrieldutra Not sure what you mean. The way `QuerySelector` is built, you can put any content in the `<option>`.

> We can add a note about admins having access instead.

Sure. Can put it in the dialog description.

gabrieldutra:
> Not sure what you mean. The way `QuerySelector` is built, you can put any content in the `<option>`.

My bad :sweat_smile:, I mixed up the `value` it can have with the Options. The Options really can have ath... So I've updated to be the UI you suggested and now will improve usability from there.

gabrieldutra:
@ranbena I didn't go much further from what the implementation is today, so lmk if you got any quick wins here :)

client/app/components/permissions-editor/PermissionsEditorDialog.jsx
```diff
@@ -0,0 +1,183 @@
...
+      <h5>Users with permissions</h5>
+      {!loadingGrantees ? (
+        <div className="scrollbox p-5 m-t-10" style={{ maxHeight: '40vh' }}>
+          <List
+            size="small"
+            dataSource={[owner, ...grantees]}
+            renderItem={user => (
+              <List.Item>
+                <UserPreviewCard key={user.id} user={user}>
+                  {user.id === owner.id ? (<Tag>Owner</Tag>) : (
```
ranbena: I'd rename to `author`

ranbena: Also, reset the `.ant-tag` right margin so it aligns to the edge.

client/app/components/permissions-editor/PermissionsEditorDialog.jsx
```diff
@@ -0,0 +1,183 @@
...
+  isUserDisabled: PropTypes.func,
+};
+UserSelect.defaultProps = { onSelect: () => {}, previewCardAddon: () => null, isUserDisabled: () => false };
+
+function PermissionsEditorDialog({ dialog, owner, context, aclUrl }) {
+  const [loadingGrantees, setLoadingGrantees] = useState(true);
+  const [grantees, setGrantees] = useState([]);
+  const { loadGrantees, addPermission, removePermission } = useGrantees(aclUrl);
+  const loadUsersWithPermissions = useCallback(() => {
+    setLoadingGrantees(true);
```
ranbena: Swapping the list for a loading indicator on each add and remove is quite jarring.
How about adding a small indicator to the top right of the scrollbox?

client/app/components/permissions-editor/PermissionsEditorDialog.jsx
```diff
@@ -0,0 +1,183 @@
...
+  return (
+    <Modal
+      {...dialog.props}
+      className="permissions-editor-dialog"
+      title={<PermissionsEditorDialogHeader context={context} />}
+      footer={(<Button onClick={dialog.dismiss}>Close</Button>)}
+    >
+      <UserSelect
+        onSelect={userId => addPermission(userId).then(loadUsersWithPermissions)}
+        previewCardAddon={user => (userHasPermission(user) ? '(already has permission)' : null)}
```
ranbena: Wdyt of omitting the already permitted from search results?

client/app/components/permissions-editor/PermissionsEditorDialog.jsx
gabrieldutra: That was bothering me too, I was waiting for you to say sth about it. I tried a few options like using an indicator in the Search bar, but ended up with more complexity and not satisfactory. Will try that one :+1:

client/app/components/permissions-editor/PermissionsEditorDialog.jsx
gabrieldutra: > I'd rename to `author`

Author does look better, but do you think it can be confusing in code, considering all the other places use `owner`? (including the backend)

ranbena: I do that stuff all the time and i think it’s ok as long as it says author only on the presentation layer.

arikfr: Currently there is no difference  between Owner and Author. But it's very possible we will allow changing the user reference here and in this case semantically it becomes more correct to call it Owner, no?

client/app/pages/dashboards/dashboard.js
```diff
@@ -250,12 +251,11 @@ function DashboardCtrl(
...
-    $uibModal.open({
-      component: 'permissionsEditor',
-      resolve: {
-        aclUrl: { url: `api/dashboards/${this.dashboard.id}/acl` },
-        owner: this.dashboard.user,
-      },
+    const aclUrl = `api/dashboards/${this.dashboard.id}/acl`;
+    PermissionsEditorDialog.showModal({
+      aclUrl,
+      context: 'query',
```
arikfr:
```suggestion
      context: 'dashboard',
```

gabrieldutra: Thanks! :slightly_smiling_face:

client/app/components/permissions-editor/PermissionsEditorDialog.jsx
ranbena: It's not that important. Just making sure users understand the label and I think `author` is clearer.

client/app/components/permissions-editor/PermissionsEditorDialog.jsx
```diff
@@ -0,0 +1,179 @@
...
+    >
+      <UserSelect
+        onSelect={userId => addPermission(userId).then(loadUsersWithPermissions)}
+        shouldShowUser={user => !userHasPermission(user)}
+      />
+      <div className="d-flex align-items-center">
+        <h5 className="flex-fill">Users with permissions</h5>
+        {loadingGrantees && <i className="fa fa-spinner fa-pulse" />}
+      </div>
+      <div className="scrollbox p-5 m-t-10" style={{ maxHeight: '40vh' }}>
```
ranbena: A small layout suggestion.
- Give the "d-flex align-items-center" element `m-t-5`.
- Remove the `m-t-10` from scrollbox.

gabrieldutra:
Merging, so I can add it to the Dashboard migration :ok_hand:

PR #4270, opened 2019-10-21
Migrate Cohort visualization to React

### LINKED ISSUE #4329: Cohort visualization isn't using the same font as the rest of the application
image

https://redash-preview.netlify.com/queries/234#427

kravets-levko:

- [x] Refactor

## Description

  - [x] Renderer
  - [x] Cornelius library
  - [x] Editor
- [x] Refine and improvements
  - [x] Cornelius: dynamic colors (using `chroma-js`)
  - [x] Cornelius: format numbers with `numeral` (no editor options for now)
  - [x] Cornelius: `moment` for date formatting and manipulations + remove hard-coded month names
  - [x] Cornelius: revisit styles
- [x] Cleanup
- [x] Tests

getredash/redash#3301 (Migrate Visualizations to React -> Cohort)
Fixes #4329

image
image


deecay:
Hi @kravets-levko,

Very nice changes. While you are at it, I have two suggestions.

1. Change Column title from "People" to "Population" to make it even more general?
2. When screen is wide, the chart would not occupy the whole width by expanding the "Time" column.

kravets-levko:
Hi @deecay! Thank you for a feedback.

> Change Column title from "People" to "Population" to make it even more general?

"People" is an accidental bug (was working late night yesterday :sleeping:) - restored "Users".

Actually, as you may notice, I prepare a ground for more configuration options for this visualizations (like changing titles, labels, value formats, colors) - they're now supported in rendering engine, but not available in UI yet because I don't want to introduce too much changes and new features in this PR.

> When screen is wide, the chart would not occupy the whole width by expanding the "Time" column.

That was also leftover after some experiments - restored previous behavior:
image

deecay:
@kravets-levko Nice. Thanks.

client/app/visualizations/cohort/Cornelius.jsx
```diff
@@ -0,0 +1,209 @@
...
+  });
+
+  return extend(options, {
+    timeLabelFormat: timeLabelFormats[options.timeInterval],
+    formatNumber: createNumberFormatter(options.numberFormat),
+    formatPercent: createNumberFormatter(options.percentFormat),
+    getColorForValue: chroma.scale([options.colors.min, options.colors.max])
+      .mode('hsl')
+      .domain([0, 100])
+      .classes(options.colors.steps),
```
ranbena: This is different from the original Cornelius - the color thresholds are divided **evenly** through the 7 steps while in the original it's:
```js
var repeatLevels = {
  'low': [0, 10],
  'medium-low': [10, 20],
  'medium': [20, 30],
  'medium-high': [30, 40],
  'high': [40, 60], // <- 20 diff
  'hot': [60, 70],
  'extra-hot': [70, 100] // <- 30 diff
}
```
This results in different color accentuation, which might not be an issue - I never worked with cohorts so I dunno if this matters.

kravets-levko: I think it's not an issue, especially considering that original Cornelius scale is very irregular (cannot choose right word for this) - it's not linear, nor logarithmic, not exponential - it's just hard-coded sequence which even does not always grow (10-10-10-10-**20**-10-**30**). I built a graph of that old pseudo-function and of linear and logarithmic functions on the same domain - linear was closest to original. It looks good enough, and visual changes are very subtle.

Maybe, later we could add a new setting to allow users to choose interpolation function from some list.

kravets-levko: I think it's not an issue, especially considering that original Cornelius scale is very irregular (cannot choose right word for this) - it's not linear, nor logarithmic, not exponential - it's just hard-coded sequence which even does not always grow (10-10-10-10-**20**-10-**30**). I built a graph of that old pseudo-function and of linear and logarithmic functions on the same domain - linear was closest to original. It looks good enough, and visual changes are very subtle.

Maybe, later we could add a new setting to allow users to choose interpolation function from some list.

ranbena: Whatever you see fit. I have no experience with such vizs.

(Btw maybe changing steps from 7 to 10 could be a good compromise)

kravets-levko: Let's keep 7 for now, later I planned add a setting for it :+1:

arikfr:
:+1:

Issue #2986, opened 2018-10-19
Impala and Hive connection issues


sabmalik:
### Issue Summary

With varying libraries in redash5 on ubuntu, I could only get one of the connections to work at a time. With my limited python knowledge, I debugged the code and figured out the combination that works.

**Scenarios**
**thrift = 0.11.0, pyhive = 0.3.0**
Hive: Connection option available and connects.
Impala: Connection option available but connection test says "TypeError: expecting list of size 2 for struct args". 

**thrift = 0.9.3, pyhive = 0.3.0**
Hive: Connection option disappears and trying "from pyhive import hive" gives you "ImportError: cannot import name TFrozenDict"
Impala: Connection option available and connects

**thrift = 0.9.3, pyhive = 0.2.1**
Hive: Connection option available and connects
Impala: Connection option available and connects

_Note that the requirements_all_ds.txt file dictates pyhive to be 0.3.0 so I am not sure if this impacts some other connection type_

### Steps to Reproduce

1. Create a connection for Impala/hive
2. Test connection

### Technical details:

* Redash Version: 5 (upgraded from 4 using "sudo bin/upgrade"
* Browser/OS: Chrome
* How did you install Redash: Script (from memory but def not docker)

BigFF:
I am using impala and find the same problem
I found there are some similar issues before
https://github.com/getredash/redash/pull/2410
https://github.com/getredash/redash/issues/1969

to solve impala we should change thrift>=0.8.0 to thrift==0.9.3
or maybe upgrade to pipenv to solve the env

I have no idea about pyhive but I think its the same problem we should keep the version of package?

justmiles:
Just tested this successfully. Setting `thrift==0.9.3` and `pyhive = 0.2.1` on top of v7.0.0 {4a978ba}. I can query CDH 5.11.0 Hive, CDH 5.11.0 Impala, and EMR 5.21.0 Hive

funyeah:
+1

Issue #3807, opened 2019-05-15
Integration tests for query runners (data sources)


arikfr:
We have plenty of open pull requests updating query runners, that we are hesitant to merge because there is no way to verify it doesn't introduce a regression. We also had our share of regressions introduced in changes that seemed safe and were merged...

I think it's due time to start looking at setting up integration tests for data sources. We will probably start with the most used once and move forward from there. The community's help here is instrumental.

As a first step, I want to document which data sources have a Docker image we can use:

Data Source | Docker Image | Notes
-- | -- | --
Amazon Elasticsearch | Y | We can create a test AWS account
Athena | X | We can create a test AWS account
Axibase | ? | 
BigQuery | X | We can create a test account
Cassandra | Y | Not sure how straightforward is the Docker image to use
Clickhouse | Y |
Couchbase | ? | 
Databricks | X | We can create a test account, but testing the Hive query runner should be enough.
IBM DB2 | ? |
Apache Drill | ? | It's HTTP based, so the very least we can have mocks.
Druid | Y | 
DynamoDB | X | We can create a test AWS account
Elasticsearch | Y |
Google Analytics | X | We can create a test account
Google Spreadsheets | X | We can create a test account
Graphite | ? | 
Hive | ? | 
Impala | ? | 
InfluxDB | Y |
JIRA | X | We can create a test account
Kylin | ? |
MapD | ? |
MemSQL | ? |
MongoDB | Y |
Microsoft SQL Server | ? |
MySQL | Y |
Oracle | ? |
PostgreSQL | Y |
Phoenix | ? |
Presto | ? |
Prometheus | ? |
Qubole | ? |
Rockset | X | We can create a test account
Salesforce | X | 
Snowflake | X | We can create a test account
SQLite | Y | 
TreasureData | X | 
Uptycs | ? |
Vertica | ? |
Yandex Metrica | ? |

Some notes:

* Because some (most?) tests will require some data for tests, it might make sense to have all the possible data sources setup on a remote server, and test against it. Although this introduces its own set of issues, so need to check.
* For each data source, we need to test against multiple versions. Which versions we test against/support should be documented.

koooge:
Hi. I'm with you and I worried about their stability. 

dynamodb-local or localstack might be useful as a dynamodb mock
https://hub.docker.com/r/amazon/dynamodb-local/
https://github.com/localstack/localstack

PR #4272, opened 2019-10-21
Fix impala connection

foxdalas:

- [x] Bug Fix

## Description

Fixes getredash/redash#2986


foxdalas:
@kravets-levko thanks

arikfr:
This is a duplicate of #3303. The reason it wasn't merged until now is that we don't have a way to know the impact of pinning the Thrift version on other data sources that use it. 

Until we have better test coverage (#3807) for our query runners or someone from the community can test it with the different ones that use Thrift, we won't be able to merge it :-(

PR #4274, opened 2019-10-22
Minor alert bug fixes

ranbena:
- [x] Refactor
- [x] Bug Fix

## Description
Looked at the code and found some non-crucial bugs/unnecessary code in the alert implementation.
1. Removed "canceling" state - since https://github.com/getredash/redash/pull/4183 clicking cancel doesn't actually redirect but instead immediately changes `state.mode` so the canceling state is redundant.
2. `HelpTrigger` was getting the flex column style 😬 Moved it into a `<div>`. 
3. `<Query>` shouldn't have an `onChange` in view mode so I removed the prop. It was `undefined` anyway 😬
4. The "Loading query data" indicator was showing up twice cause it's already part of `<Query>`, so I removed the duplicate.

### DIFF
--- a/client/app/pages/alert/AlertView.jsx
+++ b/client/app/pages/alert/AlertView.jsx
@@ -7,7 +7,6 @@ import { Alert as AlertType } from '@/components/proptypes';
 
 import Form from 'antd/lib/form';
 import Button from 'antd/lib/button';
-import Icon from 'antd/lib/icon';
 import Tooltip from 'antd/lib/tooltip';
 
 import Title from './components/Title';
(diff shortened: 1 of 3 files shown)


kravets-levko:
@ranbena Please synchronize this branch with `master` - it will resolve conflicts and hide duplicate changes

client/app/pages/alert/AlertView.jsx
```diff
@@ -46,6 +45,7 @@ AlertState.defaultProps = {
   lastTriggered: null,
 };
 
+// eslint-disable-next-line react/prefer-stateless-function
 export default class AlertView extends React.Component {
```
kravets-levko: Seems it could be a functional component (`this` is not used anywhere)

ranbena: Oh forgot to mention, I'm adding state to this component in #4276  so no point in changing it.

https://github.com/getredash/redash/blob/c01fbf90619d9053623690892c54bac827651089/client/app/pages/alert/AlertView.jsx#L51-L60

PR #4275, opened 2019-10-22
Reduce RQ status load time (Python 3)

rauchy:
- [x] Refactor

## Description
The RQ job status page was loading job information for all queued jobs, but for those, we only care about the amount. This PR fixes this.

Running a small benchmark - for 60k queued jobs, this PR has reduced the load time from approx 34 seconds to 2 seconds.


arikfr:
Let's finish reviewing/merging the RQ tasks and then create a single PR for all of them.

PR #4276, opened 2019-10-22
New feature - Alert muting

ranbena:
- [x] Feature

## Description
Added ability to mute alert.

UX discussion in https://discuss.redash.io/t/alert-feature-toggle-notifications-on-off/4704/4

Added "Mute Notifications" to the menu button, with an extra unmute button in the prompt:

Screen Shot 2019-10-22 at 14 05 16

Alerts list page also shows muted status:

Screen Shot 2019-10-22 at 14 09 40

In action:


redash/handlers/alerts.py
```diff
@@ -46,6 +46,33 @@ def delete(self, alert_id):
         models.db.session.delete(alert)
         models.db.session.commit()
 
+class AlertMuteResource(BaseResource):
+    def post(self, alert_id):
+        alert = get_object_or_404(models.Alert.get_by_id_and_org, alert_id, self.current_org)
+        require_admin_or_owner(alert.user.id)
+
+        setattr(alert.options, 'muted', True)
+        models.db.session.commit()
```
ranbena: @arikfr @rauchy I'm attempting to update `alert.options.muted` but this doesn't save to db.
What's the right way to do this?

kravets-levko: Have you tried just set property - like this: `alert.options.muted = True` ?

ranbena: IT DOES  🎉 🙏

kravets-levko: Python is not like JS, `dict.key = sth` != `dict['key'] = sth` :smile:

kravets-levko: Python is not like JS, `dict.key = sth` != `dict['key'] = sth` :smile:

ranbena: So this would have worked? 
```py
setattr(alert["options"], 'muted', True)
```

kravets-levko: Nope. `setattr(obj, key, value)` is the same as `obj.key = value`

client/app/pages/alert/components/AlertDestinations.less
```diff
@@ -34,8 +36,8 @@
 
   .add-button {
       position: absolute;
-      right: 14px;
-      top: 9px;
+      right: 0;
+      top:-33px;
```
ranbena: This just locks the "+Add" button to its parent instead of the whole wrapper.
Needed cause this section gets pushed down by the "Notifications are muted" block.

PR #4278, opened 2019-10-22
Migrate Map visualization to React

kravets-levko:

- [x] Refactor

## Description

  - [x] Renderer
  - [x] Editor
    - [x] General tab
    - [x] Groups tab
    - [x] Style tab
- [x] Refine code
- [x] Cleanup
- [x] Tests

getredash/redash#3301 (Migrate Visualizations to React -> Map)

image
image
image


client/app/visualizations/map/initMap.js
```diff
@@ -0,0 +1,188 @@
...
+
+function createMarkerClusterGroup(classify, color) {
+  const layerOptions = {};
+
+  if (classify) {
+    layerOptions.iconCreateFunction = (cluster) => {
+      const childCount = cluster.getChildCount();
+      const style = `color: white; background-color: ${color};`;
+      return L.divIcon({
+        html: `<div style="${style}"><span>${childCount}</span></div>`,
```
ranbena: Seems like the style attribute is not being added.
Screen Shot 2019-10-27 at 8 28 04

(tested on https://deploy-preview-4278--redash-preview.netlify.com/queries/49#118)

client/app/visualizations/map/Editor/StyleSettings.jsx
```diff
@@ -0,0 +1,231 @@
...
+                  placement="topLeft"
+                  arrowPointAtCenter
+                  content={(
+                    <React.Fragment>
+                      <div className="m-b-5">
+                        Enter an icon name from{' '}
+                        <a href="https://fontawesome.com/v4.7.0/icons/" target="_blank" rel="noopener noreferrer">Font-Awesome 4.7</a>
+                      </div>
+                      <div className="m-b-5">
+                        Examples: <code>check</code>, <code>times-circle</code>, <code>flag</code>
```
ranbena: Awesome

client/app/visualizations/map/Editor/StyleSettings.jsx
```diff
@@ -0,0 +1,231 @@
...
+            data-test="Map.Editor.CustomizeMarkers"
+            defaultChecked={options.customizeMarkers}
+            onChange={event => onOptionsChange({ customizeMarkers: event.target.checked })}
+          />
+          <span>Override default style</span>
+        </label>
+      </div>
+
+      {options.customizeMarkers && (
+        <React.Fragment>
```
ranbena: This whole section looks much better 👍 

If it's not too much trouble, it could be great to omit irrelevant colors by shape.
For instance, for "Circle Dot" show only "Border color".

kravets-levko:
@ranbena noticed that marker cluster icon style is inconsistent depending on "Group by" option. I implemented a fix which IMO looks much better, but now marker group icons look slightly different (see screenshots). It's a kind of breakable change, so I need your opinion - @arikfr, @ranbena, @gabrieldutra - should we keep old version on new one?

### before:

Note: green markers have default style set by Leaflet - settings from "Groups" tab are completely ignored.

image
image

### after:

Note: on both screenshots markers use colors defined on "Groups" tab.

image
image

client/app/visualizations/map/initMap.js
kravets-levko: @ranbena It worked that way before, so I just kept it. But after your comment I decided that it should be fixed - for details see my comment below :point_down:

client/app/visualizations/map/Editor/StyleSettings.jsx
kravets-levko: @ranbena WDYT - currently custom marker styles are not applied when `Group by` option enabled. Maybe, I should hide entire settings block (including checkbox) when `Group by` enabled?

ranbena:
> @ranbena noticed that marker cluster icon style is inconsistent depending on "Group by" option. I implemented a fix which IMO looks much better, but now marker group icons look slightly different (see screenshots). It's a kind of breaking change, so I need your opinion - @arikfr, @ranbena, @gabrieldutra - should we keep old version on new one?

I played around with it and it behaves very sensibly 👍

client/app/visualizations/map/Editor/StyleSettings.jsx
ranbena: I think it would be best to disable the options and also hide the settings block underneath if selected.

Screen Shot 2019-10-28 at 17 44 56

kravets-levko: @ranbena (tooltip text might have to be improved though):

image

kravets-levko: Upd.: `?` icon and tooltip are shown only when checkbox is disabled

arikfr:
One thing I noticed is that the `Groups` tab won't show the current color unless you change it (you can see this with: https://deploy-preview-4278--redash-preview.netlify.com/queries/49#118).

kravets-levko:
> One thing I noticed is that the Groups tab won't show the current color unless you change it

We have something like this in other visualizations (e.g. Chart) - when you choose Auto color - it does not show actual color in editor.

arikfr:
> > One thing I noticed is that the Groups tab won't show the current color unless you change it
> 
> We have something like this in other visualizations (e.g. Chart) - when you choose Auto color - it does not show actual color in editor.

Good point. I think that in the chart editor it's less confusing, because in the current color picker UI for charts, at least when you click on it, it tells you it's automatic and if it's automatic the preview is empty. Here it shows the "transparent" color icon, which is a bit confusing imo.

How about we use an "A" to signal automatic color? And maybe show a tooltip saying the same when hovering over it?

kravets-levko:
@arikfr How about something like this?

image

arikfr:
> @arikfr How about something like this?

This is an optional behavior, right? Not sure it will "fit" everywhere where we use the color picker. 
But don't you like changing the icon of the automatic color to be an "A"?

kravets-levko:
@arikfr Yes, this is optional behavior. I tried to add a tooltip (does not work fine with color picker's popover), tried adding letter "A" (code looked a bit ugly and it still wasn't easy to understand what that mean). I ended up with this variant.

arikfr:
@kravets-levko ok, let's move on and maybe figure something else if needed. 👍

PR #4280, opened 2019-10-23
Fix get_next_path tests

arikfr:

- [x] Other

## Description

When passing a relative path for the `location` header (when calling `redirect`), before Werkzeug 0.15 it was joining it with the root path, but after version 0.15 they started using the original path (pallets/werkzeug#693).

This PR fixes the tests to conform with this behavior. This is a non issue in app code, because usually the next URL is crafted by the app itself. This whole code is a security measure.

Also changed the tests initialization a bit to avoid one of the new warnings.

#4251

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 Flask==1.1.1
-Werkzeug==0.16.0
 Jinja2==2.10.3
 itsdangerous==1.1.0
 click==6.7
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -271,4 +271,10 @@ def get_next_path(unsafe_next_path):
     parts[1] = ''  # clear netloc
     safe_next_path = urlunsplit(parts)
 
+    # If the original path was a URL, we might end up with an empty
+    # safe url, which will redirect to the login page. Changing to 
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,5 +1,4 @@
 Flask==1.1.1
-Werkzeug==0.16.0
```
arikfr: Also removed the explicit dependency, as it's probably better to let Flask set the version.

NicolasLM: I personally feel better when all dependencies are pinned to specific versions. I remember a specific version of itsdangerous (a Flask dependency) breaking compatibility overnight, as it was not pinned it broke the app on the next deployment.

NicolasLM:
Great catch, thank you for looking into it.

requirements.txt
arikfr: We had a similar issue with Kombu (Celery's dependency), but I'm not sure if those rare (?) issues are worth the extra work / risk of having incompatible dependency because of forgetting to upgrade one but not the other.

An approach of a lock file is probably a better one to handle such issues. This way you know what gets upgraded but need to define only top level dependencies (until an issue happens). But not sure if there is a stable/well adopted solution for Python that does this :|

PR #4281, opened 2019-10-23
Python 3 RQ modifications

rauchy:
- [x] Other

## Description
This PR merges commits from several modifications made to the RQ branch. It includes all commits from the following PRs:

* [Show worker's current job](https://github.com/getredash/redash/pull/4262)
* [Reduce RQ status load time](https://github.com/getredash/redash/pull/4257)
* [Monitor general RQ queues (default, periodic and schemas)](https://github.com/getredash/redash/pull/4256)
* [Extend result_ttl for refresh_queries](https://github.com/getredash/redash/pull/4283)

1. #4262 
2. #4257 
3. #4256
4. #4283

### DIFF
--- a/client/app/components/admin/RQStatus.jsx
+++ b/client/app/components/admin/RQStatus.jsx
@@ -24,9 +24,11 @@ const workersColumns = [Columns.custom(
     /> {value}
     </span>
   ), { title: 'State', dataIndex: 'state' },
-)].concat(map(['Hostname', 'PID', 'Name', 'Queues', 'Successful Job Count',
-  'Failed Job Count', 'Birth Date', 'Total Working Time'],
-c => ({ title: c, dataIndex: c.toLowerCase().replace(/\s/g, '_') })));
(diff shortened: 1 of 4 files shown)


arikfr:
I just noticed that #4251 doesn't have any merge conflicts with master. I also noticed, this fails on Python 3 because of:

> TypeError: unsupported operand type(s) for +: 'dict_items' and 'dict_items'

How about we just merge #4251 and then apply a fix for this issue?

rauchy:
Is that due to [this](https://github.com/getredash/redash/pull/4281/commits/f3c149398734b81a2659ba7eeb4114c07272e7cb#diff-c05f73ee31584a14aed26cf798192055R38)? Because in Python 3 we would probably want to do something like 

```python
return {***{queue: {'size': redis_connection.llen(queue)} for queue in get_celery_queues()},
        ***{queue.name: {'size': len(queue)} for queue in Queue.all(connection=redis_connection)}}
```

rauchy:
> How about we just merge #4251 and then apply a fix for this issue?

Given that these modifications are already on master, merging #4251 without resolving this issue would fail the build on master.

Issue #4217, opened 2019-10-06
Use SHOW_COLUMNS to fetch Snowflake's schema


arikfr:
### Issue Summary

The current method of refreshing the Snowflake schema is starting a cluster, which is undesired -- specially when doing a background refresh of the schema.

We can alternatively use [`SHOW_COLUMNS`](https://docs.snowflake.net/manuals/sql-reference/sql/show-columns.html) to load the schema without waking a up a cluster.

### Technical details:

* Redash Version: v8

monicagangwar:
@arikfr are there any tests for snowflake with which any change to snowflake query runner can be verified ?

arikfr:
@monicagangwar this is a great question. Sadly, the answer is no :-( Most of the query runners are untested. It's something we do want to address (see #3807), but will take us time to get there.

monicagangwar:
@arikfr alright. Although i've seen there is no pagination handled while refreshing schema. 
A/c to the documentation for SHOW COLUMNS
```
The command returns a maximum of 10K records for the specified object type, as dictated by the access privileges for the role used to execute the command; any records above the 10K limit are not returned, even with a filter applied.
```
Should I handle this in the PR ?

arikfr:
@monicagangwar let's ignore the 10K limit for now. If we get reports of this being an issue will try to think of how to address it.

Thanks!

bradleyhurley:
@arikfr - Can this issue be closed? I was looking for some low hanging fruit I could help out with, but it seems the update was already merged.

arikfr:
@bradleyhurley definitely. Thanks for noticing!

PR #4285, opened 2019-10-24
refreshing snowflake schema w/o waking cluster

monicagangwar:

- [x] Refactor

## Description

Refreshing snowflake schema w/o waking cluster. Used `show columns` for this which executes the command w/o having to require a warehouse.
Created a new internal function in the query runner `__run_query_without_warehouse` instead of modifying the existing `run_query` function so as to not change the definition as specified in the base class

#4217

### DIFF
--- a/redash/query_runner/snowflake.py
+++ b/redash/query_runner/snowflake.py
@@ -112,7 +112,7 @@ class Snowflake(BaseQueryRunner):
 
         return json_data, error
 
-    def _run_query_without_warehouse(self, query, user):
+    def _run_query_without_warehouse(self, query):
         connection = self._get_connection()
         cursor = connection.cursor()
 
@@ -134,7 +134,7 @@ class Snowflake(BaseQueryRunner):
         SHOW COLUMNS IN DATABASE {database}
         """.format(database=self.configuration['database'])
 
-        results, error = self._run_query_without_warehouse(query, None)
+        results, error = self._run_query_without_warehouse(query)
 
         if error is not None:
             raise Exception("Failed getting schema.")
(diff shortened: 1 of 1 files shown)


redash/query_runner/snowflake.py
```diff
@@ -104,16 +104,47 @@ def run_query(self, query, user):
...
-        SELECT col.table_schema,
-               col.table_name,
-               col.column_name
-        FROM {database}.information_schema.columns col
-        WHERE col.table_schema <> 'INFORMATION_SCHEMA'
+        SHOW COLUMNS IN DATABASE {database}
         """.format(database=self.configuration['database'])
 
-        results, error = self.run_query(query, None)
+        results, error = self.__run_query_without_warehouse(query, None)
```
arikfr:
```suggestion
        results, error = self._run_query_without_warehouse(query)
```

(changed to be in sync with previous comment)

redash/query_runner/snowflake.py
```diff
@@ -104,16 +104,47 @@ def run_query(self, query, user):
...
+            region = None
+
+        connection = snowflake.connector.connect(
+            user=self.configuration['user'],
+            password=self.configuration['password'],
+            account=self.configuration['account'],
+            region=region
+        )
+
+        cursor = connection.cursor()
```
arikfr: At this is shared between both run query methods, let's extract it into its own method (`get_cursor`?).

monicagangwar:
@arikfr addressed PR comments and rebased from master

arikfr:
Thank you!

PR #4287, opened 2019-10-24
Fix Redash 8 requirements...

Jakdaw:
Due to $REASONS I've found myself deploying Redash against fresh builds of Python 2.7.17 with nothing else installed and have run across a couple of problems.

This change bumps pymssql to a version that builds cleanly (with fix for pymssql/issues/520) and clamps the pandas package (a dependency of pymapd) to a version that doesn't require python 3.

### DIFF
--- a/requirements_all_ds.txt
+++ b/requirements_all_ds.txt
@@ -8,7 +8,7 @@ pyhive==0.5.1
 pymongo[tls,srv]==3.6.1
 vertica-python==0.8.0
 td-client==0.8.0
-pymssql==2.1.3
+pymssql==2.1.4
 dql==0.5.24
 dynamo3==0.4.7
 boto3==1.9.115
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks, but we just merged in Python 3 support into master. So I guess this is no longer relevant?

Jakdaw:
That's why it's a merge into the redash 8 branch, or are you saying that
redash 8 final will be python 3? (woohoo!).

I'd assumed it would be redash 9 as the first p3 release.

Chris

arikfr:
Sorry, didn't notice it's a merge into the v8 branch. No, v8 is finished and stays Python 2 :-)

Still not sure as to what to do about this PR, as I was hoping not to touch the v8 branch (unless some critical issues needs fixing). :thinking:

arikfr:
I updated the 8.0.x branch and had to force push into it. Can you please rebease on top of the latest version?

Thanks.

Jakdaw:
Done; I think!

PR #4292, opened 2019-10-26
Split Dockerfile for production

koooge:

- [x] Refactor

## Description
Hi there.
This PR splits Dockerfile for production(Dockerfile) from development(Dockerfile.dev) to reduce dev dependencies(requirements_dev) influence and size.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,20 +37,28 @@ jobs:
             docker-compose up -d
             sleep 10
       - run:
+          name: Build Front End Assets due to some test cases # TODO: Fix test cases and remove this step
+          command: |
+            curl -sSL https://deb.nodesource.com/setup_12.x | sudo -E bash -
+            sudo apt install -y nodejs
+            npm install
+            npm run build
+            docker cp ./client/dist ${COMPOSE_PROJECT_NAME}_redash_1:/app/client/
+      - run:
           name: Create Test Database
           command: docker-compose run --rm postgres psql -h postgres -U postgres -c "create database tests;"
       - run:
           name: List Enabled Query Runners
-          command: docker-compose run --rm redash manage ds list_types
+          command: docker-compose exec redash /app/bin/docker-entrypoint manage ds list_types
       - run:
           name: Run Tests
-          command: docker-compose run --name tests redash tests --junitxml=junit.xml --cov-report xml --cov=redash --cov-config .coveragerc tests/
+          command: docker-compose exec redash /app/bin/docker-entrypoint tests --junitxml=junit.xml --cov-report xml --cov=redash --cov-config .coveragerc tests/
       - run:
           name: Copy Test Results
           command: |
             mkdir -p /tmp/test-results/unit-tests
-            docker cp tests:/app/coverage.xml ./coverage.xml
-            docker cp tests:/app/junit.xml /tmp/test-results/unit-tests/results.xml
+            docker cp ${COMPOSE_PROJECT_NAME}_redash_1:/app/coverage.xml ./coverage.xml
+            docker cp ${COMPOSE_PROJECT_NAME}_redash_1:/app/junit.xml /tmp/test-results/unit-tests/results.xml
           when: always
       - store_test_results:
           path: /tmp/test-results
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,15 +24,11 @@ jobs:
     environment:
       COMPOSE_FILE: .circleci/docker-compose.circle.yml
       COMPOSE_PROJECT_NAME: redash
-    machine: true
+    docker:
+      - image: circleci/buildpack-deps:xenial
     steps:
+      - setup_remote_docker
       - checkout
-      - run:
-          name: Install Docker Compose
-          command: |
-            curl -L https://github.com/docker/compose/releases/download/1.24.1/docker-compose-`uname -s`-`uname -m` > ~/docker-compose
-            chmod +x ~/docker-compose
-            sudo mv ~/docker-compose /usr/local/bin/docker-compose
       - run:
           name: Build Docker Images
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,11 +24,15 @@ jobs:
     environment:
       COMPOSE_FILE: .circleci/docker-compose.circle.yml
       COMPOSE_PROJECT_NAME: redash
-    docker:
-      - image: circleci/buildpack-deps:xenial
+    machine: true
     steps:
-      - setup_remote_docker
       - checkout
+      - run:
+          name: Install Docker Compose
+          command: |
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -65,3 +65,9 @@ gevent==1.4.0
 # Uncomment the requirement for ldap3 if using ldap.
 # It is not included by default because of the GPL license conflict.
 # ldap3==2.2.4
+
+# PyMongo and Athena dependencies are needed for some of the unit tests:
+# (this is not perfect and we should resolve this in a different way)
+pymongo[srv,tls]==3.9.0
+botocore>=1.12.241,<1.13.0
+PyAthena>=1.5.0
```
arikfr: Why move them here?

Dockerfile.dev
```diff
@@ -0,0 +1,56 @@
+FROM node:10 as frontend-builder
+
+WORKDIR /frontend
+COPY package.json package-lock.json /frontend/
+RUN npm install
+
+COPY client /frontend/client
+COPY webpack.config.js /frontend/
+RUN npm run build
+
```
arikfr: Now that we have a dedicated Docker file for development we no longer need this here, as we build frontend assets on the host.

.dockerignore
```diff
@@ -12,3 +12,4 @@ venv/
 /.github/
 /netlify.toml
 /setup/
+/tests/
```
arikfr: Why? The size they occupy is really negligible.

.circleci/config.yml
```diff
@@ -12,11 +12,15 @@ jobs:
     environment:
       COMPOSE_FILE: .circleci/docker-compose.circle.yml
       COMPOSE_PROJECT_NAME: redash
-    docker:
-      - image: circleci/buildpack-deps:xenial
+    machine: true
```
arikfr: Why?

Dockerfile.dev
```diff
@@ -0,0 +1,56 @@
...
+FROM python:3.7-slim
```
jezdez: @arikfr I just noticed this, were we okay with not using the debian base image anymore? This was done in the Python 3 move and I noted it there but I missed if this was discussed somewhere.

arikfr: The main reason we created `redash/base` image is to trim down the time it takes to build the image. I noticed that it didn't affect it, so kept as is for now. If the need arrises, we can update `redash/base` to Python 3.

jezdez: That makes sense, thanks for the clarification!

requirements.txt
koooge: Because the dependencies were needed for athena and mongo data sources. https://percy.io/Redash/redash/builds/2935315
Frontend test uses Dockerfile(not Dockerfile.dev) since it no need dev dependencies for server side.

.circleci/config.yml
koooge: For including tests dir by mount /app/test/ due to .dockerignore.
https://circleci.com/docs/2.0/executor-types/#machine-executor-overview

.dockerignore
koooge: I both agree and disagree with you on that. It is because redash's docker image is too large, however, I will rebase it because .circleci/config became to need unnecessary change. 🤔

Dockerfile.dev
koooge: By deleted this build step, then the some tests which depend on client/dist/ became to fail. So I changed index handler to return 200 with error.html in that TemplateNotFound case https://github.com/getredash/redash/pull/4292/commits/6ab16ecb503dda35294c91c6dd4f5a69703b0695

requirements.txt
arikfr: It's ok -- the Percy snapshots will be updated not to include them.

redash/handlers/static.py
```diff
@@ -6,14 +8,21 @@
...
-        response = send_file(full_path, **dict(cache_timeout=0, conditional=True))
+    try:
+        if settings.MULTI_ORG:
+            response = render_template("multi_org.html", base_href=base_href())
+        else:
+            full_path = safe_join(settings.STATIC_ASSETS_PATH, 'index.html')
+            response = send_file(full_path, **dict(cache_timeout=0, conditional=True))
+    except TemplateNotFound as e:
+        logger.exception("%s is not found", e.name)
+        response = render_template("error.html", error_message='Index page is not found.')
```
arikfr: Let's make the message even friendlier:

```python
if current_app.debug: # need to make sure current_app is imported
    message = "Missing template file ({}). Did you build the frontend assets with npm?".format(e.name)
else:
    message = "Error Rendering Page."
```

(and use this for the `error_message`)

.circleci/docker-compose.circle.yml
```diff
@@ -1,7 +1,9 @@
 version: '3'
 services:
   redash:
-    build: ../
+    build:
+      context: ../
+      dockerfile: Dockerfile.dev
```
arikfr: In CircleCI we don't build assets outside of the Docker container, so I wonder how the the tests are passing? Is it because of the exception handling in `static.py`?

redash/handlers/static.py
```diff
@@ -1,19 +1,32 @@
...
+        else:
+            full_path = safe_join(settings.STATIC_ASSETS_PATH, 'index.html')
+            response = send_file(full_path, **dict(cache_timeout=0, conditional=True))
+    except TemplateNotFound as e:
+        logger.exception("%s is not found", e.name)
+        if current_app.debug:
+            message = "Missing template file ({}). Did you build the frontend assets with npm?".format(e.name)
+        else:
+            message = "Error Rendering Page."
+        response = render_template("error.html", error_message=message)
```
arikfr: I realized that this needs to be a response with status code 500.

.circleci/docker-compose.circle.yml
koooge: Right. Some test cases depend on the build assets.

redash/handlers/static.py
koooge: Yes, actually it should return 404 or 500. it returns 200 due to backend test 🤔 
https://github.com/getredash/redash/pull/4292#discussion_r340425274

.circleci/config.yml
```diff
@@ -35,21 +35,29 @@ jobs:
...
             docker-compose up -d
             sleep 10
+      - run:
+          name: Build Front End Assets due to some test cases # TODO: Fix test cases and remove this step
+          command: |
+            curl -sSL https://deb.nodesource.com/setup_12.x | sudo -E bash -
+            sudo apt install -y nodejs
+            npm install
+            npm run build
+            docker cp ./client/dist ${COMPOSE_PROJECT_NAME}_redash_1:/app/client/
```
koooge: Build Frontend For Backend Test!

koooge:
Supersed by #4879

PR #4295, opened 2019-10-27
Fix: number based alerts evaluation isn't working

### LINKED ISSUE #4279: Number-based alerts are broken
### Issue Summary
It looks like the alert always sees a query result as a string and not a number. So an alert configured to trigger when `field = 0` never triggers because `"0"` does not equal `0`.

The alerts seems to evaluate fine for targets that are obviously strings. If the alert should trigger if `field = boyhowdy` then the alert triggers when `field` is "boyhowdy".

### Steps to Reproduce

1. Save a query that returns 0: `SELECT 0 "value"`
2. Make an alert for the query that triggers when `value = 0`
3. Run the query
4. Observe the alert does not trigger.

The converse is also true, by the way. 

1. Same as above
2. Make an alert for the query that triggers when `value != 0`
3. Run the query
4. Observe the alert triggers

### Technical details:

* Redash Version: 9.0.0-alpha+b348b58b (348b58b) 
* Browser/OS: Firefox on MacOS

arikfr:

- [x] Bug Fix

## Description

Following the alerts new UI, all the threshold values are saved as strings. This change casts the value into a number, if the query result value is a number.

Closes #4279.

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -2,6 +2,7 @@ import datetime
 import calendar
 import logging
 import time
+import numbers
 import pytz
 
 from six import text_type
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -792,11 +792,18 @@ OPERATORS = {
 
 
 def next_state(op, value, threshold):
-    if isinstance(value, numbers.Number):
+    if isinstance(value, numbers.Number) and not isinstance(value, bool):
         try:
             threshold = float(threshold)
         except ValueError:
             return Alert.UNKNOWN_STATE
+    # If it's a boolean cast to string and lower case, because upper cased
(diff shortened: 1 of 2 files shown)


arikfr:
I updated the logic:

* Because boolean values are detected a numbers, I revised the test to make sure to exclude strings.
* Cast all non number values to strings to make it possible to compare with a string.
* In case of booleans, we cast to string and lower case (`True` -> `true`) because upper cased boolean values is Python specific and will be confusing for most users.

ranbena:
> * In case of booleans, we cast to string and lower case (`True` -> `true`) because upper cased boolean values is Python specific and will be confusing for most users.

To back it up, I believe a bool value will appear lowercase in the alert column value preview.

Screen Shot 2019-10-28 at 9 37 20

PR #4314, opened 2019-10-29
Optimize work horse forking

rauchy:
- [x] Bug Fix

## Description
We've noticed that worker horses are consuming a lot of CPU. After some profiling, I realized that `configure_mapping` is called every time a work horse is forked (i.e. for every job). This PR calls `configure_mapping` once in the parent worker process, which is then available to the forked horses. We've seen significant performance improvement with this.

### DIFF
--- a/redash/cli/rq.py
+++ b/redash/cli/rq.py
@@ -6,6 +6,7 @@ import datetime
 from click import argument
 from flask.cli import AppGroup
 from rq import Connection, Worker
+from sqlalchemy.orm import configure_mappers
 
 from redash import rq_redis_connection
 from redash.schedule import rq_scheduler, schedule_periodic_jobs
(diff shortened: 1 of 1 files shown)


redash/cli/rq.py
```diff
@@ -22,6 +23,11 @@ def scheduler():
 @manager.command()
 @argument('queues', nargs=-1)
 def worker(queues='default'):
+    # Configure any SQLAlchemy mappers loaded until now so that the mapping configuration 
+    # will already be available to the forked work horses and they won't need 
+    # to spend valuable time re-doing that on every fork.
+    configure_mappers()
```
arikfr: I've noticed that in `redash.models.base` we use the `flask_sqlalchemy.SQLAlchemy`'s instance `configure_mappers`. I wonder what's the difference?

rauchy: [It's the same](https://github.com/pallets/flask-sqlalchemy/blob/master/flask_sqlalchemy/__init__.py#L73).

arikfr: WTF. Why did they do that? :confused:

rauchy: So you can access everything through `flask_sqlalchemy.SQLAlchemy`, I guess.

jezdez: Yeah, the idea being that you define a db object like `db = SQLAlchemy()` when using flask-sqlalchemy and then can call anything inside the `sqlalchemy.orm` module. In other words, it's to simplify the Flask-like API of flask-sqlalchemy.

PR #4315, opened 2019-10-29
Add user/pass authentication support for Druid

Jakdaw:

- [X] Feature

## Description

Add support for user/password authentication when connecting to Druid.

### DIFF
--- a/redash/query_runner/druid.py
+++ b/redash/query_runner/druid.py
@@ -55,9 +55,9 @@ class Druid(BaseQueryRunner):
         connection = connect(host=self.configuration['host'],
                              port=self.configuration['port'],
                              path='/druid/v2/sql/',
-                             scheme=self.configuration['scheme'],
-                             user=self.configuration['user'],
-                             password=self.configuration['password'])
+                             scheme=(self.configuration.get('scheme') or 'http'),
+                             user=(self.configuration.get('user') or None),
+                             password=(self.configuration.get('password') or None))
 
         cursor = connection.cursor()
 
--- a/redash/query_runner/druid.py
+++ b/redash/query_runner/druid.py
@@ -34,10 +34,17 @@ class Druid(BaseQueryRunner):
                 "scheme": {
                     "type": "string",
                     "default": "http"
+                },
+                "user": {
+                    "type": "string"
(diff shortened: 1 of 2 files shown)


redash/query_runner/druid.py
```diff
@@ -48,7 +55,9 @@ def run_query(self, query, user):
         connection = connect(host=self.configuration['host'],
                              port=self.configuration['port'],
                              path='/druid/v2/sql/',
-                             scheme=self.configuration['scheme'])
+                             scheme=self.configuration['scheme'],
+                             user=self.configuration['user'],
```
arikfr:
```suggestion
                             user=self.configuration.get('user'),
```

`user` might be missing from the config dict (and same for password)

Also, do you know if the client library has no side effects if passing `None` to `user`? What about an empty string?

redash/query_runner/druid.py
```diff
@@ -48,7 +55,9 @@ def run_query(self, query, user):
...
+                             password=self.configuration['password'])
```
arikfr:
```suggestion
                             password=self.configuration.get('password'))
```

redash/query_runner/druid.py
```diff
@@ -48,7 +55,9 @@ def run_query(self, query, user):
         connection = connect(host=self.configuration['host'],
                              port=self.configuration['port'],
                              path='/druid/v2/sql/',
-                             scheme=self.configuration['scheme'])
+                             scheme=self.configuration['scheme'],
```
arikfr:
```suggestion
                             scheme=self.configuration.get('scheme', 'http'),
```

(this was an issue before your PR, but worth addressing)

Jakdaw:
Library treats None like no parameter being passed... empty string it doesn't.... so I guess I need to go look at how self.configuration is populated.

Jakdaw:
That's neat!

arikfr:
Thanks!

dlevison:
To overcome this issue in redash v8.0.0 you can add the username and password in the host field:
<user>:<pass>@druid-host
Not great security wise but works.

arikfr:
@dlevison clever :)

PR #4312, opened 2019-10-29
[WIP] Parameter feedback - #1 Server errors

ranbena:
- [x] Feature

PR #4319, opened 2019-10-30
Parameter feedback - #2 Client errors in query page

ranbena:
- [x] Feature

## Description
Now that #4312 brings query result `errorData`, we can show them inline in context.
**This PR is for the query page only**
Screen Shot 2019-10-30 at 19 44 58

Notice, since the client is oblivious to the error's meaning (it's evaluated in the server), it can't be responsible for it's hiding. For instance, if a parameter was left empty, it displays "Required parameter". Once the user fills the field in, the client doesn't know that the error has been fulfilled - so it can't hide the error.
This causes a jarring user experience. As a user you expect the error indication to disappear immediately when fixed. To mitigate this I implemented a "touched" flag on each parameter which removes the red indication but keeps the error text visible, until re-evaluated by the server. This works well.
Any change to parameter - be it value or settings - toggles on the "touched" flag. The flag resets when the error data from the server has changed.

To accommodate the error messages, more bottom spacing has been given to the parameters.

### DIFF
--- a/client/app/components/Parameters.jsx
+++ b/client/app/components/Parameters.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { size, filter, forEach, extend, get } from 'lodash';
+import { size, filter, forEach, extend, get, includes } from 'lodash';
 import { react2angular } from 'react2angular';
 import { SortableContainer, SortableElement, DragHandle } from '@/components/sortable';
 import { $location } from '@/services/ng';
@@ -34,6 +34,7 @@ export class Parameters extends React.Component {
     queryResultErrorData: PropTypes.shape({
       parameters: PropTypes.objectOf(PropTypes.string),
     }),
+    unsavedParameters: PropTypes.arrayOf(PropTypes.string),
   };
 
   static defaultProps = {
@@ -44,6 +45,7 @@ export class Parameters extends React.Component {
     onPendingValuesChange: () => {},
     onParametersEdit: () => {},
     queryResultErrorData: {},
+    unsavedParameters: null,
   };
 
   constructor(props) {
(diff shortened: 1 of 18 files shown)


client/app/components/Parameters.jsx
```diff
@@ -69,14 +84,15 @@ export class Parameters extends React.Component {
 
   setPendingValue = (param, value, isDirty) => {
     const { onPendingValuesChange } = this.props;
-    this.setState(({ parameters }) => {
+    this.setState(({ parameters, touched }) => {
       if (isDirty) {
         param.setPendingValue(value);
+        touched[param.name] = true;
```
kravets-levko: Shouldn't `touched` be cloned before mutating?

client/app/components/Parameters.jsx
```diff
@@ -109,17 +125,32 @@ export class Parameters extends React.Component {
     EditParameterSettingsDialog
       .showModal({ parameter })
       .result.then((updated) => {
-        this.setState(({ parameters }) => {
+        this.setState(({ parameters, touched }) => {
+          touched[parameter.name] = true;
```
kravets-levko: The same - shouldn't `touched` be cloned?

client/app/components/Parameters.jsx
ranbena: Seems needless to me.
I wouldn't consider this a side-effect since it's being done in the context of `setState` and since this flag isn't designed to "toggle off" but only "on", there's no risk of it being out-of-sync.

kravets-levko: It's not about side effects, rather about change detection. You return the same `touched` object - how React knows that something changed? Even if it works now - I think it's accidental; what if one day React will optimize `setState` and it will no longer update component if you return the same object from `setState` callback? IMHO it worth fixing, especially considering that fix is the single line of code: `touched = { ...touched, [param.name]: true };`

PR #4321, opened 2019-10-30
Parameter feedback - #4 Added in Dashboard params

ranbena:
- [x] Feature

## Description
Extended feature to Dashboard parameters.

This was a tricky one - needs aggregating all query results from widgets and filtering out the dashboard ones. But it works!

Screen Shot 2019-10-30 at 20 06 29

- [x] Basic Cypress tests

### DIFF
--- a/client/app/pages/dashboards/public-dashboard-page.js
+++ b/client/app/pages/dashboards/public-dashboard-page.js
@@ -2,6 +2,7 @@ import PromiseRejectionError from '@/lib/promise-rejection-error';
 import logoUrl from '@/assets/images/redash_icon_small.png';
 import template from './public-dashboard-page.html';
 import dashboardGridOptions from '@/config/dashboard-grid-options';
+import _ from 'lodash';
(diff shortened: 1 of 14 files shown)


client/app/pages/dashboards/public-dashboard-page.js
```diff
@@ -70,6 +72,38 @@ const PublicDashboardPage = {
     };
 
     this.refreshDashboard();
+
+    this.getQueryResultsErrorData = () => {
```
ranbena: Duplicate code!  😱
I didn't wanna do anything knowing #4228 is at hand.

gabrieldutra: Considering it doesn't need to update state during the execution of the function, wdyt of moving it  to the Dashboard service? (We can later compute it with `useMemo`)

ranbena: Done  🙏

PR #4322, opened 2019-10-30
Parameter feedback - #5 Unsaved indication

ranbena:
- [x] Feature

## Description
Riding on the new param feedback feature, here's an indication for unsaved parameters as they are added.

- [ ] Basic Cypress tests

### DIFF
--- a/client/app/components/Parameters.jsx
+++ b/client/app/components/Parameters.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { size, filter, forEach, extend, get } from 'lodash';
+import { size, filter, forEach, extend, get, includes } from 'lodash';
 import { react2angular } from 'react2angular';
 import { SortableContainer, SortableElement, DragHandle } from '@/components/sortable';
 import { $location } from '@/services/ng';
@@ -34,6 +34,7 @@ export class Parameters extends React.Component {
     queryResultErrorData: PropTypes.shape({
       parameters: PropTypes.objectOf(PropTypes.string),
     }),
+    unsavedParameters: PropTypes.arrayOf(PropTypes.string),
   };
 
   static defaultProps = {
@@ -44,6 +45,7 @@ export class Parameters extends React.Component {
     onPendingValuesChange: () => {},
     onParametersEdit: () => {},
     queryResultErrorData: {},
+    unsavedParameters: null,
   };
 
   constructor(props) {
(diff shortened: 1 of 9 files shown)


gabrieldutra:
Just one thing in case you already had thoughts about it: It makes sense to have the same indication in the future for modified parameters (not sure if that will be a quick thing, so no need for it now :sweat_smile:)

ranbena:
> Just one thing in case you already had thoughts about it: It makes sense to have the same indication in the future for modified parameters (not sure if that will be a quick thing, so no need for it now 😅)

Definitely. It's really not clear that changing a parameter setting requires saving the query.
But the implementation is not as simple as this one.

client/app/components/Parameters.jsx
```diff
@@ -140,7 +142,22 @@ export class Parameters extends React.Component {
...
+      return [feedback, 'error'];
+    }
+
+    // unsaved
+    const { unsavedParameters } = this.props;
+    if (includes(unsavedParameters, param.name)) {
+      const feedback = (
+        <>
+          Unsaved{' '}
+          <Tooltip title='Click the "Save" button to preserve this parameter.'>
```
ranbena: @susodapop any better suggestions for this text?

susodapop: 'Click the "Save" button to make this the default value for this parameter.'

ranbena: But it's actually for the parameter itself - not its value.
The case is when a new parameter is typed into the query but ain't saved to DB till the user hits the "save" button.

susodapop: @ranbena Forgot to tag you in the previous reply.

susodapop: In that case it's no different than any other change to a query. Is a special indication necessary?

client/app/services/query.js
```diff
@@ -35,13 +35,14 @@ class Parameters {
     this.initFromQueryString(queryString);
   }
 
-  parseQuery() {
+  parseQuery(queryText = null) {
```
kravets-levko:
```suggestion
  parseQuery(queryText = this.query.query) {
```

(and line 40 could be removed)

client/app/services/query.js
```diff
@@ -35,13 +35,14 @@ class Parameters {
...
     const fallback = () => map(this.query.options.parameters, i => i.name);
+    queryText = queryText || this.query.query;
 
     let parameters = [];
-    if (this.query.query !== undefined) {
+    if (queryText !== undefined) {
```
kravets-levko:
```suggestion
    if (!isNil(queryText)) {
```

(need to import `isNil` from lodash)

client/app/components/Parameters.jsx
ranbena: It's not a must, but the asterisk indication imo is not clear enough for changes that are represented below it (including parameter setting changes which is out of scope).
I encourage you to experience it and see for your self if it provides a better user experience.

client/app/services/query.js
ranbena: I thought this wasn't possible  😲😎

kravets-levko: babel compiles it to something like: 
```javascript
queryText = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : this.query.query
```

so no magic :slightly_smiling_face:

client/app/components/Parameters.jsx
susodapop: I agree the asterisk is sub-optimal. Could we replace it with an indicator that is more obvious?

If you keep the tooltip then the text you've written above is fine.

ranbena: One way to go about it is not show or disable the Save button when query isn't dirty.
So when it is, there would be a clear indication that there are changes to be saved.

Another reason I wanted this inline indication was so that it would be clear which parameter corresponds to the one added. So "Unsaved" could be substituted with "New parameter" with a green color instead.

@arikfr how do you feel about this feature?

PR #4189, opened 2019-09-27
Fixes #3766.

susodapop:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Query Runner (Data Source)
- [ ] New Alert Destination
- [ ] Other

## Description
Updated the `Password` prompt on JQL data source setup. It now reads `API Key`.

## Related Tickets & Documents
Fixes #3766.

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

0_before
1_after

PR #4324, opened 2019-10-31
Exposing setting for overriding template directory

blarghmatey:
#4189  What type of PR is this? (check all applicable)
- [x] Bug Fix

## Description
When using some of the customized login flows such as `REMOTE_USER` the deployed site breaks due to not finding template files. This change updated the app default to use the existing Flask templates directory rather than the compiled static assets directory which only contains an index.html file.

### DIFF
--- a/redash/app.py
+++ b/redash/app.py
@@ -10,7 +10,7 @@ class Redash(Flask):
     def __init__(self, *args, **kwargs):
         kwargs.update(
             {
-                "template_folder": settings.STATIC_ASSETS_PATH,
+                "template_folder": settings.FLASK_TEMPLATE_PATH,
                 "static_folder": settings.STATIC_ASSETS_PATH,
                 "static_url_path": "/static",
             }
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -229,7 +229,9 @@ LDAP_SEARCH_DN = os.environ.get(
 STATIC_ASSETS_PATH = fix_assets_path(
     os.environ.get("REDASH_STATIC_ASSETS_PATH", "../client/dist/")
 )
-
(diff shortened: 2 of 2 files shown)


arikfr:
Just to get the context for this:

> When using some of the customized login flows such as `REMOTE_USER` the deployed site breaks due to not finding template files

How using the `REMOTE_USER` feature requires additional template files?

arikfr:
@blarghmatey did you see my question above?

blarghmatey:
@arikfr sorry for the slow reply. I remember looking through the specifics of the logic flow when I first ran into this after first deploying Re:Dash ~2 years ago, but as I recall the short version is that if using the remote user login feature (https://github.com/getredash/redash/blob/master/redash/authentication/remote_user_auth.py) it is trying to load the `login` template, which only exists under the template directory inside the app. By default it is only able to find the `index.html` template which bootstraps the frontend code, but fails to allow for loading any of the other existing templates. I have been running in production with this configuration for quite some time and am happy to provide more details of our settings and environment.

The majority of the configuration can be seen in our Salt repository here: https://github.com/mitodl/salt-ops/blob/master/pillar/apps/redash.sls#L46-L74

Please let me know if you have any other questions.

blarghmatey:
Just a quick ping to keep this PR visible @arikfr

arikfr:
@blarghmatey I was about to merge this, but then noticed that it picked up a few unrelated commits. Can you create a branch that has only the settings updates and I will merge it?

Thanks and sorry for keeping it open for so long.

blarghmatey:
Thanks for following up on this @arikfr. I just rebased my branch off of master so it should be ready to merge.

arikfr:
Merged 👍

PR #4327, opened 2019-11-01
Parameter feedback - #6 Better value normalization

ranbena:
- [x] Bug Fix

## Description
Text and Number parameters need some tweaking in terms of value normalization.

### Text parameter
1. An empty text field value is converted to `null` but multiple spaces do not and should. Now trimming.
2. Adding spaces to value triggered `dirty` indication. Now it won't.

### Number parameter
1. Empty value sent `0` to server. Now sending `null` instead, displaying "Required parameter" error.
2. Typing non-number values would disappear on blur/apply. Now sending value as is, displaying "Invalid value" error.

Screenshots have been added inline in code.

### DIFF
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -5,7 +5,7 @@ import Input from 'antd/lib/input';
 import InputNumber from 'antd/lib/input-number';
 import DateParameter from '@/components/dynamic-parameters/DateParameter';
 import DateRangeParameter from '@/components/dynamic-parameters/DateRangeParameter';
-import { isEqual } from 'lodash';
+import { isEqual, trim } from 'lodash';
 import { QueryBasedParameterInput } from './QueryBasedParameterInput';
 
 import './ParameterValueInput.less';
@@ -59,7 +59,7 @@ class ParameterValueInput extends React.Component {
   }
 
   onSelect = (value) => {
-    const isDirty = !isEqual(value, this.props.value);
+    const isDirty = !isEqual(trim(value), trim(this.props.value));
     this.setState({ value, isDirty });
     this.props.onSelect(value, isDirty);
   }
@@ -140,13 +140,11 @@ class ParameterValueInput extends React.Component {
     const { className } = this.props;
     const { value } = this.state;
 
-    const normalize = val => (isNaN(val) ? undefined : val);
-
     return (
       <InputNumber
         className={className}
-        value={normalize(value)}
-        onChange={val => this.onSelect(normalize(val))}
+        value={value}
+        onChange={val => this.onSelect(val)}
       />
     );
   }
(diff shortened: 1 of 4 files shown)


client/app/components/ParameterValueInput.jsx
```diff
@@ -59,7 +59,7 @@ class ParameterValueInput extends React.Component {
   }
 
   onSelect = (value) => {
-    const isDirty = !isEqual(value, this.props.value);
+    const isDirty = !isEqual(trim(value), trim(this.props.value));
```
ranbena: Before:


After:

client/app/services/parameters/NumberParameter.js
```diff
@@ -9,11 +9,11 @@ class NumberParameter extends Parameter {
 
   // eslint-disable-next-line class-methods-use-this
   normalizeValue(value) {
-    if (isNull(value)) {
+    if (!trim(value)) {
       return null;
     }
     const normalizedValue = toNumber(value);
-    return !isNaN(normalizedValue) ? normalizedValue : null;
+    return !isNaN(normalizedValue) ? normalizedValue : value;
```
ranbena: Before:
Screen Shot 2019-11-01 at 9 04 05

After:
Screen Shot 2019-11-01 at 9 03 55

client/app/services/parameters/TextParameter.js
```diff
@@ -15,6 +15,13 @@ class TextParameter extends Parameter {
     }
     return normalizedValue;
   }
+
+  getExecutionValue() {
+    if (!trim(this.value)) {
+      return null;
+    }
+    return this.value;
+  }
```
ranbena: Before:
Screen Shot 2019-11-01 at 9 05 58


After:
Screen Shot 2019-11-01 at 9 06 10

client/app/services/parameters/NumberParameter.js
```diff
@@ -9,11 +9,11 @@ class NumberParameter extends Parameter {
 
   // eslint-disable-next-line class-methods-use-this
   normalizeValue(value) {
-    if (isNull(value)) {
+    if (!trim(value)) {
       return null;
     }
```
ranbena: Before:
Screen Shot 2019-11-01 at 9 07 51

After:
Screen Shot 2019-11-01 at 9 07 25

client/app/components/ParameterValueInput.jsx
```diff
@@ -140,13 +140,11 @@ class ParameterValueInput extends React.Component {
...
 
-    const normalize = val => (isNaN(val) ? undefined : val);
-
     return (
       <InputNumber
         className={className}
-        value={normalize(value)}
-        onChange={val => this.onSelect(normalize(val))}
+        value={value}
+        onChange={val => this.onSelect(val)}
```
gabrieldutra: :+1:

client/app/services/parameters/TextParameter.js
gabrieldutra: I didn't see a reason why someone would use a parameter only containing whitespaces, so this looks good to me :+1:.

client/app/components/ParameterValueInput.jsx
arikfr: I wonder if there is a legitimate use case for multiple spaces in a parameter :grimacing:

PR #4338, opened 2019-11-04
Update TagsList and Sidebar to use Ant components

gabrieldutra:

- [x] Refactor

## Description
Following [this](https://github.com/getredash/redash/pull/4323#discussion_r341839683) discussion, this is an adaptation using Ant components for the Sidebar menu and the TagsList.

Initially this is to discuss the result and possible modifications from the Ant Components.

--

**Queries List**
ant-tags-and-sidebar

**Users List**
users-page

### DIFF
--- a/client/app/assets/less/ant.less
+++ b/client/app/assets/less/ant.less
@@ -375,3 +375,15 @@
     margin-top: 9px;
   }
 }
+
+.@{menu-prefix-cls} {
+  // invert stripe position with class .invert-stripe-position
+  &-inline.invert-stripe-position {
+    .@{menu-prefix-cls}-item {
+      &::after {
+        right: auto;
+        left: 0;
+      }
+    }
+  }
+}
--- a/client/app/assets/less/inc/list-group.less
+++ b/client/app/assets/less/inc/list-group.less
@@ -17,31 +17,6 @@
     }
 }
 
-tags-list {
-  a {
-    line-height: 1.1;
-  }
-}
-
-.tags-list__name {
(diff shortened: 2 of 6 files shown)


kravets-levko:
Awesome @gabrieldutra! :tada: 

Is it possible to move that blue stripe (selected item indication) to left - as it was with Bootstrap styles?
image

arikfr:
@kravets-levko I agree that with the current position of the sidebar the indication should be on the left side, but what if we moved the sidebar to the left side? :open_mouth:

kravets-levko:
I like current placement of sidebar, but, I think, @gabrieldutra may try both variants and then decide what looks better :+1:

gabrieldutra:
**Inverted Stripe**
inverted-stripe

**Inverted Menu**
inverted-menu

In general perspective, left-to-write readers will face the sidebar first when it's on the left side.

I liked the feeling when it was on the left side, but lmk what you prefer :)

kravets-levko:
I like sidebar on right side more, but left side looks good as well :+1:

arikfr:
I think the thinking behind putting the sidebar on the right side was:

* It makes sense to show the main content (the list) first. 
* The sidebar isn't a menu but more like filters for the data shown (which is more common to have on the right).

Considering that both options look good, let's go with having it as is today and put the stripe in the right place (https://github.com/getredash/redash/pull/4338/commits/c21ecdfdb6e457f258c7ec8d5d35650ffb4d596d).

gabrieldutra:
Done!

PR #4306, opened 2019-11-06
RQ: periodically clear failed jobs

rauchy:
As per rq/rq#1143, failed jobs stay in Redis forever. If this is true, we should implement our own periodic cleanup of these jobs.

### DIFF
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -197,7 +197,7 @@ LDAP_SEARCH_DN = os.environ.get('REDASH_LDAP_SEARCH_DN', os.environ.get('REDASH_
 STATIC_ASSETS_PATH = fix_assets_path(os.environ.get("REDASH_STATIC_ASSETS_PATH", "../client/dist/"))
 
 JOB_EXPIRY_TIME = int(os.environ.get("REDASH_JOB_EXPIRY_TIME", 3600 * 12))
-JOB_DEFAULT_FAILURE_TTL = int(os.environ.get("REDASH_JOB_DEFAULT_FAILURE_TTL", 30 * 24 * 60 * 60))
+JOB_DEFAULT_FAILURE_TTL = int(os.environ.get("REDASH_JOB_DEFAULT_FAILURE_TTL", 7 * 24 * 60 * 60))
 
 LOG_LEVEL = os.environ.get("REDASH_LOG_LEVEL", "INFO")
 LOG_STDOUT = parse_boolean(os.environ.get('REDASH_LOG_STDOUT', 'false'))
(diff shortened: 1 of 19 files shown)


redash/tasks/general.py
```diff
@@ -60,3 +62,33 @@ def send_mail(to, subject, html, text):
 
 def sync_user_details():
     users.sync_last_active_at()
+
+
+def purge_failed_jobs():
+    jobs = rq_redis_connection.scan_iter('rq:job:*')
+
+    is_idle = lambda key: rq_redis_connection.object('idletime', key) > settings.JOB_DEFAULT_FAILURE_TTL
```
arikfr: >  This subcommand is available when maxmemory-policy is set to an LRU policy or noeviction. 

(This command being `idletime`)

Is this the default config for Redis?

redash/tasks/general.py
```diff
@@ -60,3 +62,33 @@ def send_mail(to, subject, html, text):
...
+        with Connection(rq_redis_connection):
+            failed_registries = [FailedJobRegistry(queue=q) for q in Queue.all()]
+
+        job_id = lambda key: key.decode().split(':').pop()
+        return all([job_id(key) not in registry for registry in failed_registries])
+
+    stale_jobs = [key for key in jobs if is_idle(key) and has_failed(key) and not_in_any_failed_registry(key)]
+
+    for key in stale_jobs:
+        rq_redis_connection.delete(key)
```
arikfr: Maybe worth removing it from the FailedRegistry while we at it?

redash/tasks/general.py
rauchy: The default is `noeviction`. On AWS, for example, it defaults to `volatile-lru`.

redash/tasks/general.py
rauchy: We _could_ do that, but the point is that we want to let the FailedJobRegistry handle its own state. From what I can tell, there aren't any dire consequences to have ghost job ids in the FailedJobRegistry (it is only used for requeueing and in that case - these jobs will simply not get requeued)

rauchy: If we delete from `FailedJobRegistry`, we might as well just do that and avoid checking for job inclusion (and avoid the whole comment+bypass at the top of the function).

🤔

arikfr: I'm just worried that over a year it can accumulate quite a lot of job ids there, which might have some consequences in performance or at least memory usage.

Re. avoid checking job inclusion: I guess we can skip this.

rauchy: Yeah 02555b9 makes things simpler.

redash/settings/__init__.py
```diff
@@ -197,6 +197,7 @@
 STATIC_ASSETS_PATH = fix_assets_path(os.environ.get("REDASH_STATIC_ASSETS_PATH", "../client/dist/"))
 
 JOB_EXPIRY_TIME = int(os.environ.get("REDASH_JOB_EXPIRY_TIME", 3600 * 12))
+JOB_DEFAULT_FAILURE_TTL = int(os.environ.get("REDASH_JOB_DEFAULT_FAILURE_TTL", 30 * 24 * 60 * 60))
```
arikfr:
```suggestion
JOB_DEFAULT_FAILURE_TTL = int(os.environ.get("REDASH_JOB_DEFAULT_FAILURE_TTL", 7 * 24 * 60 * 60))
```

I really don't see the point in keeping them for so long.

PR #4302, opened 2019-11-07
Restarting rq-scheduler reschedules all periodics

rauchy:
When `rq-scheduler` restarts, it deletes all existing scheduled jobs and reschedules them. The result of this is that, for example, tasks that should run daily may run more than once a day.

Instead of deleting all jobs and rescheduling them, the `scheduler` command should respect previously scheduled jobs and continue their original iterations.


redash/schedule.py
```diff
@@ -11,26 +15,36 @@
...
+
+    return hashlib.sha1(json.dumps(metadata, sort_keys=True).encode()).hexdigest()
 
-def schedule(**kwargs):
+
+def prep(kwargs):
     interval = kwargs['interval']
     if isinstance(interval, timedelta):
-        interval = interval.seconds
+        interval = int(interval.total_seconds())
```
arikfr: What's the diff between `total_seconds()` and `seconds`?

redash/schedule.py
```diff
@@ -11,26 +15,36 @@
...
+logger = logging.getLogger(__name__)
+
 rq_scheduler = Scheduler(connection=rq_redis_connection,
                          queue_name="periodic",
                          interval=5)
 
+def job_id(kwargs):
+    metadata = kwargs.copy()
+    if 'func' in metadata:
+        metadata['func'] = metadata['func'].__name__
```
arikfr: Can there be a job definition without `func`? Also, isn't the func name good enough as a job id?

redash/cli/rq.py
```diff
@@ -9,14 +9,15 @@
...
+from redash.schedule import rq_scheduler, schedule_periodic_jobs, periodic_job_definitions
 
 manager = AppGroup(help="RQ management commands.")
 
 
 @manager.command()
 def scheduler():
-    schedule_periodic_jobs()
+    jobs = periodic_job_definitions()
+    schedule_periodic_jobs(jobs)
```
arikfr: Why not call the function that returns job definitions in `schedule_periodic_jobs`?

redash/schedule.py
rauchy: `timedelta(days=1).seconds == 0`

redash/schedule.py
rauchy: Not sure what I was thinking with checking for existence in metadata.

if we use the function name as an identifier, then we would have to change the function name when we want to apply other things (for example, the interval or results TTL) in order to reschedule them.

redash/cli/rq.py
rauchy: Purely for testability.

redash/schedule.py
arikfr: :open_mouth:

redash/schedule.py
arikfr: How about the job id becomes `metadata['func'].__name__:hash`? This way when poking around and seeing a job id it's easy to tell what job it references?

PR #4344, opened 2019-11-11
Fix the DB migration so that the correct key is used for encrypting DS credentials

Jakdaw:
Without this upgrades from at least v5 (and earlier) won't work.

- [X] Bug Fix

## Description

For background, see: 

https://discuss.redash.io/t/database-migration-using-incorrect-key-for-encryption/4833/4


arikfr:
Thanks!

PR #4351, opened 2019-11-13
use  to set the hash instead of directly manipulating it

rauchy:
- [x] Refactor

## Description
This is a short refactoring of the hash manipulation done in #4299. Nothing too interesting.

#4299

### DIFF
--- a/client/app/pages/admin/Jobs.jsx
+++ b/client/app/pages/admin/Jobs.jsx
@@ -82,7 +82,6 @@ class Jobs extends React.Component {
     const changeTab = (newTab) => {
       $location.hash(newTab);
       $rootScope.$applyAsync();
-      location.replace(`${location.pathname}#${newTab}`);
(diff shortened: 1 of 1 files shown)


client/app/pages/admin/Jobs.jsx
```diff
@@ -80,6 +80,8 @@ class Jobs extends React.Component {
     const { isLoading, error, queueCounters, startedJobs, overallCounters, workers, activeTab } = this.state;
 
     const changeTab = (newTab) => {
+      $location.hash(newTab);
+      $rootScope.$applyAsync();
       location.replace(`${location.pathname}#${newTab}`);
```
kravets-levko: This line should be removed, no? :slightly_smiling_face:

rauchy: Which line? 😉

PR #4352, opened 2019-11-13
Use column types provided by server rather than trying to derive them from the values

Jakdaw:

- [X] Bug Fix

## Description
For background see:

https://discuss.redash.io/t/hex-for-binary-data/4900/11

Be warned, first time I've touched Javascript!

Start using the column types in the JSON document rather than trying to derive them from the values. Now a MySQL query like SELECT "2019-01-01", date("2019-01-01"); will return a string and a date.

I don't think the 'float' thing was ever actually being set in the old logic; and I can't see what our require the JSON.stringify() logic.

### DIFF
--- a/client/app/services/query-result.js
+++ b/client/app/services/query-result.js
@@ -1,6 +1,6 @@
 import debug from 'debug';
 import moment from 'moment';
-import { uniqBy, each, isNumber, isString, includes, extend, forOwn } from 'lodash';
+import { uniqBy, each, includes, extend, forOwn } from 'lodash';
 
 const logger = debug('redash:services:QueryResult');
 const filterTypes = ['filter', 'multi-filter', 'multiFilter'];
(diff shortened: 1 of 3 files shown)


Jakdaw:
Ahem! and I've only actually tested this on the Redash8 branch... tho I don't think that's relevant in practice!

gabrieldutra:
Hi @Jakdaw, saw your comment about the tests in the forum :slightly_smiling_face:.

Changing this line to use the `SELECT DATE('{{test-parameter}}') AS parameter` query should fix the failing tests:
https://github.com/getredash/redash/blob/aa06b32e17fa6377073f0cefa7fa95a7e9a21795/client/cypress/integration/query/parameter_spec.js#L260

There's also a change in a Percy snapshot that this should fix too:
https://github.com/getredash/redash/blob/aa06b32e17fa6377073f0cefa7fa95a7e9a21795/client/cypress/integration/visualizations/table/.mocks/all-cell-types.js#L8

using `TIMESTAMP '1995-09-03T07:45' AS "date",` instead.

LMK if you get any issues with it

Jakdaw:
Thanks @gabrieldutra looks like that sorted it!

guidopetri:
@Jakdaw , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

PR #4305, opened 2019-11-13
RQ: implement reliable timeout 

rauchy:
RQ's timeout implementation is based on a signal invoked in the "work horse" process, which might be blocked by the executing code. We should implement a more reliable timeout -- probably by handling it in the parent process.

Relevant RQ issues: rq/rq#323, rq/rq#1142


redash/worker.py
```diff
@@ -93,3 +97,68 @@ def add_periodic_tasks(sender, **kwargs):
     for params in extensions.periodic_tasks.values():
         # Add it to Celery's periodic task registry, too.
         sender.add_periodic_task(**params)
+
+
+class HardTimeLimitingWorker(Worker):
```
arikfr: How about we move this into its own module in the `redash.tasks` package? The `redash.schedule` module belongs there as well. 

I would move all the rq stuff from `redash.worker` into `redash.tasks` and eventually remove this module when we say goodbye to Celery.

arikfr: Also worth adding some documentation on why we added this class.

redash/tasks/__init__.py
```diff
@@ -3,3 +3,5 @@
                       refresh_schemas, cleanup_query_results, empty_schedules)
 from .alerts import check_alerts_for_query
 from .failure_report import send_aggregated_errors
+from .hard_limiting_worker import *
+from .schedule import *
```
arikfr: Nit: please don't do `*` imports.

arikfr:
One more thing for this PR: we need to make sure that all the tasks have sensible timeout setting.

rauchy:
> One more thing for this PR: we need to make sure that all the tasks have sensible timeout setting.

All jobs have the default timeout of 3 minutes at the moment, except for `sync_user_details` which maxes out at 60 seconds. These feel like sensible values for me. WDYT?

Obviously, when we get to query executions, these will be dynamic.

arikfr:
Do we need to merge this or #4413 covers this too?

rauchy:
I based #4413 on this because of the similarities in the custom worker. Given that #4413 will take some more time for testing, if you think this is ready to ship - let's merge it on its own.

redash/tasks/hard_limiting_worker.py
```diff
@@ -0,0 +1,84 @@
...
+    2. A hard time limit, enforced by the parent worker
+
+    The HardLimitingWorker class changes the default monitoring behavior of the default
+    RQ Worker by checking if the work horse is still busy with the job, even after
+    it should have timed out (+ a grace period of 15s). If it does, it kills the work horse.
+    """
+    grace_period = 15
+
+    def soft_limit_exceeded(self, job):
+        seconds_under_monitor = (utcnow() - self.monitor_started).seconds
```
mr-trouble: Small consideration. The timedelta docs say that .seconds is:
`0 <= seconds < 3600*24 (the number of seconds in one day)`
If jobs are supported that run longer than a day, probably should use .total_seconds() instead.
```suggestion
        seconds_under_monitor = (utcnow() - self.monitor_started).total_seconds()
```

arikfr:
#4413 is merged, so I guess we can close this one?

arikfr:
Ping, @rauchy.

rauchy:
Yes, this is all included in `master` by now.

PR #4354, opened 2019-11-13
Remove Chrome Logger and update Cypress and Percy

### LINKED ISSUE #4286: Make Cypress tests work with Cypress@3.6.0
Running our tests with Cypress@3.5.0 doesn't work. Need to figure out what happened, until then pinning the version to 3.4.1 (#4284).

gabrieldutra:

- [x] Other

## Description
This removes the Chrome Logger (was causing issues with Cypress requests).

Also took the opportunity to:
- Revert #3732 (no longer necessary without the chrome logger)
- Revert #4016 (no longer necessary with Cypress@3.5.0)
- Use the "server" command instead of "dev_server" for Redash in Cypress CI

Other than the chrome logger, solved some other issues that appeared in the specs.

Closes #4286.

--


client/app/components/ParameterApplyButton.jsx
```diff
@@ -12,7 +12,7 @@ function ParameterApplyButton({ paramCount, onClick }) {
   return (
     <div className="parameter-apply-button" data-show={!!paramCount} data-test="ParameterApplyButton">
       <Badge count={paramCount}>
-        <Tooltip title={`${KeyboardShortcuts.modKey} + Enter`}>
+        <Tooltip title={paramCount ? `${KeyboardShortcuts.modKey} + Enter` : null}>
```
gabrieldutra: The Tooltip was still up after click in front of some parameter inputs, blocking tests to continue

gabrieldutra:
Percy highlighted that the Tooltip issue was not isolated to the Apply All button, now all clicked objects show their tooltips and since there's no unhover action on the tooltip or in the wrapped object, it stays open.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "The frontend part of Redash.",
   "main": "index.js",
   "scripts": {
-    "start": "node --max-http-header-size=16385 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js",
```
arikfr: Good riddance :ok_hand:

client/cypress/plugins/index.js
```diff
@@ -1 +1,5 @@
-module.exports = () => {};
+const percyHealthCheck = require('@percy/cypress/task'); // eslint-disable-line import/no-extraneous-dependencies, import/no-unresolved
```
arikfr: What's this?

package.json
```diff
@@ -16,7 +16,7 @@
     "lint:ci": "npm run lint -- --format junit --output-file /tmp/test-results/eslint/results.xml",
     "test": "TZ=Africa/Khartoum jest",
     "test:watch": "jest --watch",
-    "cypress:install": "npm install --no-save cypress@3.4.1 @percy/cypress@^0.2.3 atob@2.1.2",
+    "cypress:install": "npm install --no-save cypress@3.6.1 @percy/cypress@^2.2.0 atob@2.1.2",
```
arikfr: Maybe use a less strict version specifier for cypress?

gabrieldutra: I usually don't pin it, but it seems to be often breaking something on those updates (already happen with a patch version as well). Perhaps just pin it with Patch version?

```suggestion
    "cypress:install": "npm install --no-save cypress@~3.6.1 @percy/cypress@^2.2.0 atob@2.1.2",
```

client/cypress/plugins/index.js
gabrieldutra: I forgot to mention, but percy-cypress lib from version 2 is requiring this. It's the Step 2 in their [guide](https://docs.percy.io/docs/cypress#section-setup).

package.json
arikfr: Whatever you feel is best :-)

gabrieldutra:
@kravets-levko I wonder how [this snapshot](https://percy.io/Redash/redash/builds/3133447/view/196518293/1280?mode=diff&browser=chrome&snapshot=196518293) could change just by updating Cypress/Percy :thinking:. You think it's a valid result too? (for me it's good, just need to be stable like that)

kravets-levko:
@gabrieldutra I may be caused by different visualization container width (but not sure what may cause this, though). Both results look valid (there are no partially visible elements, font looks the same and so on), but the most reliable way to check which snapshot is correct is to run this test locally with the same browser window size we use for Percy. Also, I'm not sure if it will be stable result.

gabrieldutra:
Thanks @kravets-levko  I actually just wanted to make sure it was still valid :sweat_smile:, the rest I don't think it's worth it to be testing manually, we will see what happens over time :)

PR #4355, opened 2019-11-14
applied a condition because some mappings don't contain 'properties' …

dextermapalya:
…field

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other


arikfr:
It's really unclear what this Pull Request supposed to address also it seems to require a proper rebase.

PR #4362, opened 2019-11-16
Fix unicode encode error by column duplicate in presto datasource

elonlo:
ex.
select '1' as "测试abc", '2' as "测试abc", '3' as "test_abc", '4' as "test_abc"
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)


arikfr:
The master branch is using Python 3, so the `u''` prefix is no longer needed there. You're welcome to add a test case for the `fetch_columns` method to verify this.

elonlo:
okay, i close pr

PR #4364, opened 2019-11-16
Migrate VisualizationEmbed to React

gabrieldutra:

- [x] Refactor

## Description
Migration of VisualizationEmbed to React

All as is, with one small difference in the "Open in Redash" icon, I changed the icon to "External Link" instead of "Link".

I also changed the Refresh button because it doesn't seem to work in the Visualization Embed context, lmk if I should handle that differently :slightly_smiling_face:

#3071 

visualization-embed


client/app/components/queries/VisualizationEmbed.jsx
```diff
@@ -0,0 +1,235 @@
...
+function VisualizationEmbedHeader({ queryName, queryDescription, visualization }) {
+  return (
+    <div className="embed-heading p-b-10 p-r-15 p-l-15">
+      <h3>
+        <img
+          src={logoUrl}
+          alt="Redash Logo"
+          style={{ height: '24px', verticalAlign: 'text-bottom' }}
+        />
+        <VisualizationName visualization={visualization} />{' '}
```
arikfr: I think we no longer need the `react2angular` definition of this component...

arikfr:
:+1:

PR #4371, opened 2019-11-19
Multiprocess RQ workers (using supervisor)

rauchy:
- [x] Refactor

## Description
This is an alternative implementation to the one suggested in #4233. Here we use `supervisor` instead of `Honcho` in order to spin up, monitor and restart multiple worker processes.

supervisor is installed in the Docker image and is launched inside the `worker` entrypoint. We use the non-daemonized mode of supervisor in order to allow the container to exit in case the supervisor process exits.

#4233


bin/docker-entrypoint
```diff
@@ -25,13 +25,22 @@ dev_scheduler() {
...
-  exec watchmedo auto-restart --directory=./redash/ --pattern=*.py --recursive -- ./manage.py rq worker $QUEUES
+  start_worker "watchmedo auto-restart --directory=./redash/ --pattern=*.py --recursive -- "
+}
+
+start_worker() {
+  export EXECUTION_PREFIX=$1
+  export WORKERS_COUNT=${WORKERS_COUNT:-2}
+  export QUEUES=${QUEUES:-}
+
+  touch /tmp/worker.log && tail -f /tmp/worker.log &
```
arikfr: Is there a reason you didn't configure supervisord to write to stdout directly?

rauchy: That's even simpler :) didn't think of that! 1fb1a7e handles this.

worker.conf
```diff
@@ -0,0 +1,17 @@
+[supervisord]
+logfile=/tmp/supervisord.log
```
arikfr: Need to update this to stdout as well.

rauchy: fdbbd13

rauchy:
> We still some healthcheck, right? Can Supervisor handle it?
In 092eb6c I've used [supervisor_checks](https://github.com/vovanec/supervisor_checks) to run a check every minute - the check runs for every monitored process and asserts that the worker is behaving as expected (on top of what you get from supervisor, which is basically testing if the process is up). If a specific worker process is misbehaving, it is replaced.

We can also add more checks for CPU / memory (provided OOTB by supervisor_checks)

Let me know what you think of this.

rauchy:
@arikfr can we merge this?

PR #4372, opened 2019-11-19
Amazon CloudWatch query runners

arikfr:

## Description

Query runner for Amazon CloudWatch and CloudWatch Insights.

- [x] Implement Insights query runner.
- [x] Update names & types.
- [x] Add logos.
- [x] Implement Test Connection.


arikfr:
> 1. How do you decide if a helper function is inside or outside a class scope? For example, I'm not sure why `CloudWatch._get_client` is inside the `CloudWatch` scope, but `parse_response` is outside.

`_get_client` uses values from the instance, hence it's a class method. `parse_response` doesn't so it's outside.

> 2\. How do you pick between list comprehensions and building with `for` loops?

Mostly habit thing. When the for loop is simple I use list comprehensions, but when it's multi line I usually do a for loop.

rauchy:
> `_get_client` uses values from the instance, hence it's a class method. parse_response doesn't so it's outside.

Yeah, but encapsulating is about _knowledge_, not access to memory. `parse_response` is tightly coupled to `CloudWatch`'s knowledge and not at all portable. It's as related to `CloudWatch` as `_get_client`...

arikfr:
I'm looking at it not at the class level, but the module level. `redash.query_runner.cloudwatch` is a module about accessing CloudWatch metrics. It has the `CloudWatch` class which is the Redash driver/interface and `parse_response` which is used to parse responses coming from the API.

There is also a benefit in testing: it makes it trivial to test `parse_response` without having to go through the whole `CloudWatch` class. Of course it could be a `staticmethod` on the `CloudWatch` class, but I don't see much benefit in doing so.

kangzj:
@arikfr Hi mate, good job & greate effort
I'm a newbie to redash. May I ask how to query from CloudWatch? It doesn't seem to work the normal way. Thanks
image

susodapop:
Hey @kangzj, the CloudWatch query runner documentation is underway. The associated issue is https://github.com/getredash/website/issues/453. This data source is part of V9 which is still in beta. The doc should be ready before the full release.

The short answer is that you don't use SQL to query CloudWatch. It uses a YAML query (similar to the JSON query source or MongoDB).

PR #4378, opened 2019-11-20
add pyexasol datasource, ensure that integer dont overflow in javascript

stefan-mees:

## Description
Adds Exasol as datasource to Redash, we are running our own fork with this connector since a while and would like to contribute it back to the mainline.


requirements.txt
```diff
@@ -38,7 +38,7 @@ gunicorn==19.9.0
 rq==1.1.0
 rq-scheduler==0.9.1
 celery==4.3.0
-kombu==4.6.5
+kombu==4.6.6
```
arikfr: If I remember right `4.6.6` had a bug. As this change is probably not relevant for this PR, let's keep as is.

redash/query_runner/exasol.py
```diff
@@ -0,0 +1,153 @@
...
+    enabled = False
+
+
+class Exasol(BaseQueryRunner):
+    noop_query = "SELECT 1 FROM DUAL"
+
+    def test_connection(self):
+        data, error = self.run_query(self.noop_query, None)
+        if error is not None:
+            raise Exception(error)
```
arikfr: You can remove this, as this is the default implementation.

redash/query_runner/exasol.py
```diff
@@ -0,0 +1,153 @@
...
+            "properties": {
+                "user": {
+                    "type": "string"
+                },
+                "password": {
+                    "type": "string"
+                },
+                "host": {
+                    "type": "string",
+                    "default": "127.0.0.1"
```
arikfr: As it's very unlikely to have an Exasol instance running on the same host as Redash, I would remove this default value. Also let's make `host` a required value.

redash/query_runner/exasol.py
```diff
@@ -0,0 +1,153 @@
...
+                    "type": "number",
+                    "default": 8563
+                },
+            },
+            "order": ['host', 'port', 'user', 'password'],
+            "secret": ["password"]
+        }
+
+    def _get_connection(self):
+        exahost = "%s:%s" % (self.configuration['host'], self.configuration['port'])
```
arikfr: You need to use `self.configuration.get('port', 8563)`, as default values are not automatically assigned.

redash/query_runner/exasol.py
```diff
@@ -0,0 +1,153 @@
...
+        return pyexasol.connect(
+            dsn=exahost,
+            user=self.configuration['user'],
+            password=self.configuration['password'],
```
arikfr: `user`/`password` might not exist, so you need to use `self.configuratino.get` here as well (if no default value provided, it iwll return `None`).

redash/query_runner/exasol.py
```diff
@@ -0,0 +1,153 @@
...
+        try:
+            statement = connection.execute(query)
+            columns = [{'name': n, 'friendly_name': n,'type': _type_mapper(t)} for (n, t) in statement.columns().items()]
+            cnames = statement.column_names()
+            
+            rows = [dict(zip(cnames, row)) for row in statement]
+
+            data = {'columns': columns, 'rows': rows}
+            error = None
+            json_data = json.dumps(data, default=self._serializer)
```
arikfr: Is there a reason you didn't use `json_dumps` from `redash.utils` here?

requirements_all_ds.txt
```diff
@@ -32,3 +32,5 @@ phoenixdb==0.7
 certifi>=2019.9.11
 pydgraph==2.0.2
 azure-kusto-data==0.0.35
+pyexasol==0.9.1
+python-rapidjson==0.8.0
```
arikfr: Did you notice any noticeable performance difference when using this?

stefan-mees: Its recommended in the best practices of pyexasol, i have not done any benchmarking on this by myself. 

https://github.com/exasol/pyexasol/blob/master/docs/BEST_PRACTICES.md#consider-faster-json-parsing-libraries

requirements.txt
stefan-mees: Sounds good,  i bumped it cause of a bug we where seeing in the admin status view of celery and bumping the version resolved the issue, but i agree this should be part of a different PR

redash/query_runner/exasol.py
stefan-mees: Changed it to use `json_dumps` from `redash.utils`

requirements.txt
arikfr: Btw, `4.6.5` has a bug. `4.6.3` is the version I had in mind. We accidentally upgraded as part of the Python 3 migration and dependencies update... It's possible that `4.6.6` is good too, but as we deprecating Celery anyway we moved back to the last version that was stable for us.

wildraid:
@stefan-mees , @arikfr , Hi guys,

PyEXASOL creator is here. Thank you for the pull request!

Let's see if we can improve this code a little bit:

- Regarding to `rapidjson` library. It improves performance of json decoding by ~10-30% in ideal case scenario. But in this query runner the response is encoded in JSON format again by `redash.utils.json_dumps`, so serialisation overhead is doubled or even tripled. In this case `rapidjson` won't really help much. Maybe it's not worth to add it as a mandatory dependency.

- Regarding to `_exasol_type_mapper`. Please note that creation of `datetime.date` and `datetime.datetime` objects is quite expensive. If you have a lot of DATE / TIMESTAMP columns in result set, it may take more CPU resources than all other actions. But you do not really need those objects. They are serialised into strings again before returning the result set.

- It might be worth adding more configurable parameters ([full reference](https://github.com/badoo/pyexasol/blob/master/docs/REFERENCE.md#connect)). At very least we need `compression` (it should be `False` in the cloud and in the same data center, it should be `True` for remote data center or WiFi connection), `encryption` (security), `socket_timeout`, `schema`, `http_proxy`.

- Please note that having separate `host` and `port` parameters might be a bad idea in the long run. Currently it works fine, but starting from Exasol 7.0+ we'll have an ability to create multiple read-only clusters, and each of them might have a different port. DSN might look like this: `cluster1exa1..10:8563,cluster2exa1..10:8564`. It might be a good idea to keep this parameter as pure `dsn` and pass it directly to `pyexasol.connect()` without concatenation.

- It is advised to add `/*snapshot execution*/` prefix to this SQL query:

```/*snapshot execution*/
        SELECT
            COLUMN_SCHEMA,
            COLUMN_TABLE,
            COLUMN_NAME
        FROM EXA_ALL_COLUMNS
```

This is a new feature described in https://www.exasol.com/support/browse/IDEA-476. It will help to prevent locks while retrieiving large amounts of meta data.

- Variable `error` does not seem to be populated by exceptions. It might be worth adding a basic `except` block catching exceptions and returning text message to ReDash. Please note that PyEXASOL exceptions are instances of `pyexasol.ExaException`, but sometimes we may get `OSException` from underlying WebSocket / socket networking layer. If default error messages are too verbose for ReDash, you may set `verbose_error=False` to get only test without extra information.

I would patch it myself, but we do not currently use ReDash in Badoo, so I cannot test those changes on prod long term.

Thank you! Have a nice day.

PR #4379, opened 2019-11-20
Migrate Home to React

gabrieldutra:

- [x] Refactor

## Description
This migrates the Homepage to react, everything was migrated except for one addition: added a small Loading indicator to Favorites list.

I'm not using Antd `List` because it doesn't support Links in `List.Item` and I didn't want to use an `onClick` event to mitigate that.

--

Alerts:
Screenshot from 2019-11-20 17-32-35

Loading:
favorites-loading

EmptyState:
Screenshot from 2019-11-20 17-51-50


client/app/components/BeaconConsent.jsx
```diff
@@ -76,8 +75,4 @@ export function BeaconConsent() {
   );
 }
 
-export default function init(ngModule) {
-  ngModule.component('beaconConsent', react2angular(BeaconConsent));
-}
-
-init.init = true;
+export default BeaconConsent;
```
kravets-levko: Just wondering: why do you prefer this pattern (declare component and export it at the end of file) instead of immediately exporting (`export default function BeaconConsent(...)`)?

gabrieldutra: I don't have a strong reasoning for that, both are fine to me. I usually go with this pattern because at some point I got used to see those exports in the end of files. I guess you prefer them immediately exported :laughing:?

kravets-levko: > I guess you prefer them immediately exported

Yes :smile: (but it's just because I used to use this variant, I think both are fine)

Issue #4195, opened 2019-09-29
Make sure docker-compose up works with default Docker setup


arikfr:
I don't know if this is always the case, but `docker-compose up -d` just failed for me on a clean install of Windows because of the memory issue with webpack (the need to have at least 4GB of memory). I have a feeling it only works when:

1) running on Linux (and the host has enough memory)
2) running on Windows/Mac with memory allocated to Docker already increased

This is a terrible experience for contributors and I think we should invest some effort in fixing this. I see two options here, but happy to hear suggestions:

1. Make sure webpack can work with default settings (2GB of memory). webpack/webpack#6389 has some relevant discussion.
2. Make development setup skip building frontend assets as they're not needed anyway (being overwritten by local ones).

arikfr:
`npm run build` also won't work on Windows because `rm` is not recognized command.

PR #4381, opened 2019-11-21
Remove --max-old-space-size=4096 from npm build command

arikfr:

- [x] Other

## Description

Looks like it's no longer needed.

Related: #4195.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10 as frontend-builder
+FROM node:12 as frontend-builder
 
 WORKDIR /frontend
 COPY package.json package-lock.json /frontend/
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "start": "webpack-dev-server",
     "bundle": "bin/bundle-extensions",
     "clean": "rm -rf ./client/dist/",
-    "build": "npm run clean && NODE_ENV=production node --max-old-space-size=4096 node_modules/.bin/webpack",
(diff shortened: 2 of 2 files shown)


kravets-levko:
Seems it does not work: build failed with out of memory error in node

arikfr:
> Seems it does not work: build failed with out of memory error in node

Strange it worked locally after the unused dependencies removal. Will verify again.

arikfr:
Looks like it's also Node version dependent. I wonder if changing this now will break existing environments for people? :-\

I did stumble at [Volta](https://volta.sh/) recently, which makes it easy to run the correct Node version per project so maybe it's not a huge deal.

@kravets-levko @gabrieldutra @rauchy what Node version you have locally?

rauchy:
Locally I'm on v11.9, but I don't do local nowadays :) so it's v8.10

arikfr:
@rauchy 

> Locally I'm on v11.9, but I don't do local nowadays :) so it's v8.10

v8.10 :scream: Just use Volta. Takes a minute and you have whatever version you want.

kravets-levko:
> what Node version you have locally?

v10.13.0

gabrieldutra:
> what Node version you have locally?

v12.13.0, I updated a few days ago :slightly_smiling_face:

rauchy:
> Just use Volta. Takes a minute and you have whatever version you want.

It really did!

arikfr:
I added another npm run command `build:old-node-version`, which has the old max memory setting for those who use an older Node version.

Issue #4366, opened 2019-11-17
Add additional aggregate functions to Counter visualization


arikfr:
Currently the Counter visualization can either show a value or count the rows. We should extend the functionality to allow additional aggregate functions:

* Sum
* Average
* Min, Max
* First/Last (not sure those are really needed, as can be done already with the index)

PR #4385, opened 2019-11-22
Fix counter vizualization

### LINKED ISSUE #4265: Counter Visualization: count rows never shows zero
### Issue Summary

If I toggle **Count Rows** and my query returns no results, the visualization disappears. It should show "0". It makes sense for the visualization to disappear if it points at a field and row index. But sometimes the number of rows is actually zero and this should be reflected by the counter.

### Steps to Reproduce

1. Write a query that returns a single row: `SELECT 'foo', 'bar'`
2. Add a counter visualization and toggle **Count Rows**. The counter will show 1.
3. Add a NULL `WHERE` clause to the query ... `where 1 =2`. The counter disappears.

### Technical details:

* Redash Version: 7, 8, 9alpha
* Browser/OS: Firefox / Chrome / Safari
* How did you install Redash: Saas / Docker

dmudro:

- [x] Add Unit Tests
- [x] Bug Fix

## Description
Small change in behaviour of the counter visualisation as we want to always display a value when count rows option is enabled; even when the value is 0 (i. e. 0 rows in the underlying data set). Introduces tests for logic contained in utils, though there is more to be done in this area (another PR coming as it touches a more general point).

Closes #4265.


client/app/visualizations/counter/utils.test.js
```diff
@@ -0,0 +1,168 @@
+import { getCounterData } from './utils';
```
dmudro: Not able to stub `getCounterData` and other private functions in utils at the moment. Will be addressed in another PR.

arikfr: Why would you want to stub it?

dmudro: I meant to say '`getRowNumber` and other private functions'. Stubbing (mocking) would depend on the testing strategy though.

Atm all these 'private' functions including `getRowNumber`, `numberFormat`, `formatValue` etc: 
1. Contain logic and are not covered. This is fixable by simply exporting and running cases on them.
2. From a puristic perspective, when executing tests on `getCounterData` one would want to mock the calls to these private functions to isolate the test subject. Otherwise it's more of an integration testing (which is not too bad anyway). This type of mocking is not an easy one in React as it does not have dependency injection and available solutions [are](https://stackoverflow.com/questions/45111198/how-to-mock-functions-in-the-same-module-using-jest) [cumbersome](https://medium.com/@DavideRama/mock-spy-exported-functions-within-a-single-module-in-jest-cdf2b61af642) or would require coding style.

I assume it's ok to test on integration level and gradually add tests for other functions.

client/app/visualizations/counter/utils.js
```diff
@@ -70,31 +70,27 @@ function formatTooltip(value, formatString) {
   return toString(value);
 }
 
-export function getCounterData(rows, options, visualizationName) {
+export function getCounterData(rows = [], options = {}, visualizationName = '') {
```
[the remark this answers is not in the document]
dmudro: I agree refactoring code base for the sole sake of testing is generally not ideal.

`rows = []` was meant to protect against calling `getCounterData()` at runtime which would trow TypeError on `rows.length` (and since there is no type checking at build time). Fair enough such case is not realistic and `rows` should be a mandatory param.

Same for `options = {}` although the nature of options may suggest they should be _optional_. In that sense keeping the default value is not just for tests and the same is true for `visualizationName = ''` which one may want to fallback to an empty string.

Would following amended function signature do the job here?:
`function getCounterData(rows, options = {}, visualizationName = '')`

dmudro:
> Thank you for fixing this! Taking the extra effort of adding the tests is appreciated 🙇

Am not a huge fan of putting an extreme emphasis on tests but here the code with logic has been begging for decent coverage 🤓. 

On related note `getCounterData()` seems to be slightly bloated and may benefit from splitting in more atomic parts, perhaps while addressing #4366. Wondering if values formatting could be done outside, i. e. in the React component.

client/app/visualizations/counter/utils.test.js
arikfr: > I meant to say '`getRowNumber` and other private functions'.

Oh, ok. This makes sense now :-)

> I assume it's ok to test on integration level and gradually add tests for other functions.

It's indeed ok. We're not purists here but rather pragmatists :-) So even having just a test for the `getCounterData` function is enough considering its scope/size.

client/app/visualizations/counter/utils.js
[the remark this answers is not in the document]
arikfr: There will always be options and a visualization name... I'm actually not sure about `rows` as I don't know what happens when we're still waiting for the results to load. But having a default value won't help in such case anyway, because we definitely pass some value, it just might be empty.

My suggestion is to revert the function definition to be as it was to make it clear that it needs these values, but for convenience in the test you can wrap it with a function that has default values.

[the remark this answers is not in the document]
dmudro: ok cool, default param values removed

client/app/visualizations/counter/utils.test.js
dmudro: > We're not purists here but rather pragmatists :-)

Nice one. Aligned.

client/app/visualizations/counter/utils.test.js
```diff
@@ -0,0 +1,192 @@
...
+describe('Visualizations -> Counter -> Utils', () => {
+  beforeEach(() => {
+    dummy = {
+      rows: [
+        { city: 'New York City', population: 18604000 },
+        { city: 'Shangai', population: 24484000 },
+        { city: 'Tokyo', population: 38140000 },
+      ],
+      options: {},
+      visualisationName: 'Visualisation Name',
```
gabrieldutra: `visualisation` -> `visualization` :slightly_smiling_face:

dmudro: Nice catch and both forms are correct in terms of grammar if it's what you meant. These are just test dummies; might as well be _foo name_. Please feel free to rename to whatever you prefer.

gabrieldutra: It was actually to keep it as is in the rest of the codebase, but NP, I'll revisit those tests in #4392.

arikfr:
Thank you @dmudro for the contribution and patience during the review :-)

Issue #4393, opened 2019-11-23
Table visualization with column named "children" renders +/- buttons


gabrieldutra:
### Issue Summary
Initially reported in [this forum post](https://discuss.redash.io/t/is-it-possible-to-hide-common-rows-under-a-spoiler/5060).

This is related to [this](https://ant.design/components/table/#components-table-demo-expand-children) Antd feature. Basically having a "children" column name in a query is triggering the +/- button to appear.

Can be an issue from Antd side, but creating the issue here so we have it mapped.

![children-column](https://user-images.githubusercontent.com/3356951/69486688-22e56e80-0e2d-11ea-9f71-38e2cdbea340.gif)

### Steps to Reproduce

1. Create a query with a column named `children` or use [this](https://redash-preview.netlify.com/queries/260)
2. Check it has an extra +/- button

### Technical details:

* Redash Version: v8
* Browser/OS: Any
* How did you install Redash: --

PR #4394, opened 2019-11-24
Table visualization with column named "children" renders +/- buttons

kravets-levko:

- [x] Bug Fix

Fixes getredash/redash#4393

Before:
image

After:
image

### DIFF
--- a/client/app/visualizations/table/Renderer.jsx
+++ b/client/app/visualizations/table/Renderer.jsx
@@ -4,12 +4,11 @@ import Table from 'antd/lib/table';
 import Input from 'antd/lib/input';
 import { RendererPropTypes } from '@/visualizations';
 
-import { prepareColumns, filterRows, sortRows } from './utils';
+import { prepareColumns, initRows, filterRows, sortRows } from './utils';
 
 import './renderer.less';
 
 export default function Renderer({ options, data, context }) {
-  const [rowKeyPrefix, setRowKeyPrefix] = useState(`row:1:${options.itemsPerPage}:`);
   const [searchTerm, setSearchTerm] = useState('');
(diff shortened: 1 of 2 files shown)


gabrieldutra:
Yeah, it is supposed to be a very simple fix, I missed the part of the docs that mentioned `childrenColumnName` when I first read it :sweat_smile:. But that surprisingly still holds the issue:

image

In the docs it seem that prop is an string array, so perhaps `childrenColumnName={[]}`

client/app/visualizations/table/Renderer.jsx
```diff
@@ -65,6 +65,7 @@ export default function Renderer({ options, data, context }) {
         data-test="TableVisualization"
         columns={tableColumns}
         dataSource={preparedRows}
+        childrenColumnName={null}
```
gabrieldutra:
```suggestion
        childrenColumnName={[]}
```

kravets-levko: @gabrieldutra 

image

gabrieldutra: :man_facepalming:

gabrieldutra: That worked after I update Antd, maybe they changed it to allow array recently?

kravets-levko: `childrenName` prop is [declared as optional/nullable](https://github.com/ant-design/ant-design/blob/master/components/table/interface.tsx#L212), but later there are no any conditions - it's just used as is - e.g. `item[this.props.childrenName]` - implicit cast to string. I'll try to find a new solution.

PR #4311, opened 2019-11-26
RQ: simplify logging of jobs

rauchy:
RQ logs the arguments the task was called with. This creates too verbose logs and might leak information where it not supposed to go (specially in the case of `execute_query`).

We should switch it log only the job name.

### DIFF
--- a/redash/cli/rq.py
+++ b/redash/cli/rq.py
@@ -33,7 +33,7 @@ def worker(queues):
         queues = ['periodic', 'emails', 'default', 'schemas']
 
     with Connection(rq_redis_connection):
-        w = Worker(queues)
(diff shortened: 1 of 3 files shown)


arikfr:
For some reason GitHub kept requesting me for a comment when I tried to leave an approval review, so I just merged...

rauchy:
Here are a few suggestions for next time:

> Oh boy, terrific job!
This PR has taught me so much, thank you!
At first I thought this was an impossible task, but you proved me wrong.
This is code craftsmanship at it's finest.
You make me want to be a better programmer.
Holy shit, you nailed it!

I can go on.

arikfr:
@rauchy thanks! This is very helpful.

PR #4408, opened 2019-11-27
Migrate AddToDashboard dialog

kravets-levko:

- [x] Refactor

image
image
image
image


client/app/lib/hooks/useSearchResults.js
```diff
@@ -0,0 +1,33 @@
+import { useState, useEffect } from 'react';
+import { useDebouncedCallback } from 'use-debounce';
+
+export default function useSearchResults(
```
gabrieldutra: Cool :+1:, I will keep it in mind for next search uses

kravets-levko:
@gabrieldutra 

image

client/app/components/queries/AddToDashboardDialog.jsx
```diff
@@ -0,0 +1,110 @@
...
+import notification from '@/services/notification';
+import useSearchResults from '@/lib/hooks/useSearchResults';
+
+import './add-to-dashboard-dialog.less';
+
+function AddToDashboardDialog({ dialog, visualization }) {
+  const [searchTerm, setSearchTerm] = useState('');
+
+  const [doSearch, dashboards, isLoading] = useSearchResults((term) => {
+    if (isString(term) && (term.length >= 3)) {
```
gabrieldutra: The loading indicator was not the only thing that caused me a bit of confusion :confused:, do we really need this `term.length >= 3`? (I mean, the results are paged, right?) There is the option of using Antd form to indicate that minimum length, but I don't think it's that worth it. Well, wdyt?

kravets-levko: I'm not sure why it was implemented this way; of course that API can search by a single character. Also, here we don't show all results, but only a most relevant ones (20 or 25 - don't remember exactly) - I think it makes sense.

arikfr: I think the idea was to reduce number of requests we send because it's less likely that a <3 characters term will be meaningful. We can drop this.

kravets-levko: @arikfr @gabrieldutra https://github.com/getredash/redash/pull/4408/commits/40d5aa91f5910de7117842a52406df98e480d52b

gabrieldutra:
(The dragged parameters spec is flaky, will fix that in a few minutes)

Issue #4307, opened 2019-10-28
Move query execution to RQ


arikfr:
Move the `execute_query` task from Celery to RQ. Once this is done, we're Celery free. :tada: 

Depends on #4305 and ability to cancel jobs (rq/rq#684, which we will probably implement in a custom Worker in our codebase and then try to port to RQ). 

(Opening this to track the work, but this is currently sparse on details)

rauchy:
Regarding query cancellation - how awful would it be if we piggy-back ride on the work horse monitoring loop? Effectively this means that queries that get cancelled might only stop after up to `job_monitoring_interval` (default is 30s, but we can tighten that up). 

The alternative would be a thread watching for cancellations in the work horse, but I'm trying to think of ways to avoid making our code complex.

arikfr:
It doesn't sound too bad if we can lower the interval to something like 5s. What impact it will have?

rauchy:
From what I can tell, just more heartbeat registrations.

arikfr:
No impact on CPU or other resources? If so let's lower it and move on.

rauchy:
I guess also more alarm signals will be registered, but at 5 second intervals, these are extremely negligible.

arikfr:
I don't know if it complicates things much, but we should consider having an optional handler function (in the query runner) that does the actual cancel and if it's not available then resort to current "kill process".

This will allow for implementing #4410 or [MySQL's cancel procedure](https://github.com/getredash/redash/blob/36ab8eae89f66de45c9e784e4020c50c7d704451/redash/query_runner/mysql.py#L171-L175) without the use of threads.

This is a nice to have of course and if it complicates the implementation let's move on with implementing the current behavior (of sending an interrupt signal and on second try a kill signal).

arikfr:
I guess we can close this?

PR #4413, opened 2019-11-28
Execute Queries in RQ

rauchy:

- [x] Refactor

## Description
Moves query executions from Celery to RQ.

- [x] Fix tests
- [x] Remove all Celery monitoring stuff
- [x] Try to stick to the term "job" wherever possible (remove "task"s)
- [x] Test on different data sources
- [x] Perform load tests

I was really tempted to clean up the code as I went along with this refactoring, but I decided to move things as they are and get them working on RQ, then simplify.

#4307


redash/tasks/queries/execution.py
```diff
@@ -111,10 +109,10 @@ def enqueue_query(query, data_source, user_id, is_api_key=False, scheduled_query
             if job_id:
```
rauchy: I know I said I’ll port `enqueue_query` from Celery to RQ as is and deal with refactoring later, but I’m really tempted to use RQ’s ability to set custom IDs to jobs and get rid of the whole lock mechanism using it (i.e. job IDs will be rq:job:<ds.id>:<query_hash> ). WDYT @arikfr?

arikfr: This is indeed tempting! But -- 

1. We lose ability to track down specific invocation (in logs and such), because every invocation has the same id.
2. Current hash implementation needs fixing (#2137).
3. Are we sure job IDs were meant to be used this way?

rauchy: 1. That could be easily solved by a custom description or meta attribute holding an invocation counter.
2. That needs fixing regardless, but does that affect whether we should drop locks?
3. It's not a hack, if that's what you mean. RQ docs specify that you can use a predetermined ID instead of the auto-generated one.

arikfr: > RQ docs specify that you can use a predetermined ID instead of the auto-generated one.

But do they expect the ID to be non unique/reusable? An execution's id feels like something that supposed to be unique, I'm just concerned about the can of worms we're opening if we assign a non unique one.

rauchy: All jobs scheduled from `rq-scheduler` reuse the same identifier :)

arikfr: You might want to check out: https://github.com/rq/rq/issues/793.

arikfr: TL;DR: 

> What sirex said, the job will be executed twice (if the job is still in Redis).

rauchy: I fail to see the issue. We still check for the job's existence in Redis. We get the same effect of a lock by checking if the job is queued.

arikfr: I thought you were suggesting reusing job ids to avoid having to implement our own lock. 

While we don't need to worry about another key, we do need to worry about lots of other things and fully understand how rq works. Considering that the current implementation works and we have reasonable processes for it, I don't think we will gain much here at this stage, until we're more familiar with RQ. The risk vs. gain is just not worth it.

redash/tasks/queries/execution.py
```diff
@@ -26,75 +27,6 @@ def _unlock(query_hash, data_source_id):
     redis_connection.delete(_job_lock_id(query_hash, data_source_id))
 
 
-class QueryTask(object):
```
arikfr: :wave: good bye old friend. This class survived since the very first iterations of Redash (it was actually named `Job` back then), when we had my half-assed background job implementation and Tornado as the web server.

rauchy:
> Also, can't we just call push_connection when starting the app or something?

After too much frolicking, I realized that setting this at the start of the app doesn't play nice with Flask's threads and RQ's LocalStack. I'll keep the pre-request & post-request for now unless you can see any fundamental issues with it.

arikfr:
> I'll keep the pre-request & post-request for now unless you can see any fundamental issues with it.

Do we really push a request context when starting the app and/or executing a job?

rauchy:
@arikfr which query runners do you think I should try this on?

PR #4420, opened 2019-12-04
Cypress: Make sure params are saved before reload

gabrieldutra:

- [x] Other

## Description
This makes sure Parameters were saved before reloading the page, to avoid flakyness in the Parameters spec.

--

--


kravets-levko:
So what's next? Will you re-run build 100500 times to be sure that fix works? :grin:

gabrieldutra:
> So what's next? Will you re-run build 100500 times to be sure that fix works?

I think only 20.234 times should be enough, so let's go with it

PR #4421, opened 2019-12-05
Fix  Google Apps SAML: missing FirstName / LastName

e271828-:

- [X] Bug Fix

## Description

Two issues with the "supported" bitnami images when enabling Google SAML:

1. missing xmlsec1

Solution:

`sudo apt-get install xmlsec1`

(not sure where to PR that fix)

2. This PR: line 82 of /opt/bitnami/apps/redash/htdocs/redash/authentication/saml_auth.py
(`name = "%s %s" % (authn_response.ava['FirstName'][0], authn_response.ava['LastName'][0])`)
required the following patch to replace it:

```
    try:
        name = "%s %s" % (authn_response.ava['FirstName'][0], authn_response.ava['LastName'][0])
    except Exception as e:
        name = email.split('@')[0]
```

Prior to that patch, error upon hitting SAML Login was as follows:

(server logs were in `/opt/bitnami/apps/redash/logs/redash_gunicorn.log`)
```
[2019-12-05 01:54:40,607] ERROR in app: Exception on /saml/callback [POST]
  File "/opt/bitnami/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/bitnami/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/bitnami/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask_restful/__init__.py", line 271, in error_router
    return original_handler(e)
  File "/opt/bitnami/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/bitnami/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/bitnami/apps/redash/htdocs/venv/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
  File "/opt/bitnami/apps/redash/htdocs/redash/authentication/saml_auth.py", line 82, in idp_initiated
    name = "%s %s" % (authn_response.ava['FirstName'][0], authn_response.ava['LastName'][0])
KeyError: 'FirstName'
```

With both changes, Google SAML now works.

Non-obvious steps to enable it:
1. Download the metadata file (option 2) from Google when setting up Custom SAML app, upload to URL your redash can read
2. use this URL in redash config
3. entity ID in redash: https://accounts.google.com/o/saml2?idpid=<id_from_google>
(make sure the whole URI is in the Entity ID field on Google)
4. "SAML NameID Format" in redash: `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`


arikfr:
> Two issues with the "supported" bitnami images when enabling Google SAML:
> 
> 1. missing xmlsec1
> 
> Solution:
> 
> `sudo apt-get install xmlsec1`
> 
> (not sure where to PR that fix)

No need to PR this: our recommended deployment uses a Docker image which already installs this. We're also advising against using Bitnami's images as they aren't compatible with our documentation or standard practices. 



> 1. This PR: line 82 of /opt/bitnami/apps/redash/htdocs/redash/authentication/saml_auth.py
>    (`name = "%s %s" % (authn_response.ava['FirstName'][0], authn_response.ava['LastName'][0])`)
>    required the following patch to replace it:

I would expect the response from Google to include the name somewhere. Maybe it's under a different field?

e271828-:
> > Two issues with the "supported" bitnami images when enabling Google SAML:
> > 
> > 1. missing xmlsec1
> > 
> > Solution:
> > `sudo apt-get install xmlsec1`
> > (not sure where to PR that fix)
> 
> No need to PR this: our recommended deployment uses a Docker image which already installs this. We're also advising against using Bitnami's images as they aren't compatible with our documentation or standard practices.

Good to know. Where is that recommendation? Perhaps it should be more prominent.
 
> > 1. This PR: line 82 of /opt/bitnami/apps/redash/htdocs/redash/authentication/saml_auth.py
> >    (`name = "%s %s" % (authn_response.ava['FirstName'][0], authn_response.ava['LastName'][0])`)
> >    required the following patch to replace it:
> 
> I would expect the response from Google to include the name somewhere. Maybe it's under a different field?

Could be, or may need to be explicitly exposed. Regardless, dying without "FirstName" and "LastName" in the SAML response is clearly not optimal. Presumably other providers also do not send both fields.

e271828-:
@arikfr is there something else you'd like to see here? Current master and latest release is totally broken for this use case, and presumably many other SAML implementations.

e271828-:
@arikfr bump

posix4e:
Hello?

arikfr:
When setting up the SAML application in Google's Admin console, you need to provide the following mapping:

Screen Shot 2020-02-16 at 2 15 04 PM

While it's not ideal that Redash blows up without much information, I think it's reasonable to expect a certain setup.

Happy to accept a PR that makes the SAML endpoint return sensible error messages for this case and others.

Closing this for now, but please feel free to follow up.

e271828-:
@arikfr Google is not the only SAML provider in the world, and without this particular set of mappings (where is it documented?) any other will also break. This patch is already a good generic solution. Unclear why you'd prefer the current behavior, which breaks in a difficult-to-identify way.

arikfr:
> Unclear why you'd prefer the current behavior, which breaks in a difficult-to-identify way.

As I said, I'm happy to accept a PR that makes the SAML endpoint return sensible error messages for this case and others (to make it easier to identify).

> (where is it documented?)

It will also be great to update [the documentation for SAML](https://redash.io/help/user-guide/users/authentication-options) ([source](https://github.com/getredash/website/blob/master/src/pages/kb/user-guide/users/authentication-options.md)) to be more generic and include this information. (cc: @susodapop)

e271828-:
> > Unclear why you'd prefer the current behavior, which breaks in a difficult-to-identify way.
> 
> As I said, I'm happy to accept a PR that makes the SAML endpoint return sensible error messages for this case and others (to make it easier to identify).

This adds a log message, but failing seems wrong: FirstName and LastName are not mandatory in underlying data sources, so you'd be forcing people to create a fake mapping just to get it working.

> > (where is it documented?)
> 
> It will also be great to update [the documentation for SAML](https://redash.io/help/user-guide/users/authentication-options) ([source](https://github.com/getredash/website/blob/master/src/pages/kb/user-guide/users/authentication-options.md)) to be more generic and include this information. (cc: @susodapop)

mchechik:
I work on a team that doesn't have access to the SAML server's code, but I was able to modify Redash's saml_auth.py to do a sensible thing imo - report the error and allow the user to log in.

I replaced 
`name = ...`
with
`name = "%s %s" % (authn_response.ava['FirstName'][0] if "FirstName" in authn_response.ava else "FirstName Not Found in SAML Response", authn_response.ava['LastName'][0] if "LastName" in authn_response.ava else "LastName Not Found in SAML Response")`

The profile data is messy since no one is named "FirstName Not Found", but the end result is that users have been able to login without issues due to this SAML field and then they report it to the team maintaining SAML for fixing on their end. I checked Redash's backend constraints and the `name` field does not enforce uniqueness. TLDR - seems like a sensible handling of the error in my experience.

arikfr:

PR #4425, opened 2019-12-07
A few updates post eslint

gabrieldutra:

- [x] Other

## Description
@arikfr here are a few updates I had in mind (was quicker to share in a PR than in comments). The ones @kravets-levko mentioned I'm not so sure about, so will confirm them first.

### DIFF
--- a/client/app/components/EditParameterSettingsDialog.jsx
+++ b/client/app/components/EditParameterSettingsDialog.jsx
@@ -86,13 +86,13 @@ function EditParameterSettingsDialog(props) {
 
   // fetch query by id
   useEffect(() => {
-    const { queryId } = props.parameter;
+    const queryId = props.parameter.queryId;
     if (queryId) {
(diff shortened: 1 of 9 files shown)


client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -86,13 +86,13 @@ function EditParameterSettingsDialog(props) {
 
   // fetch query by id
   useEffect(() => {
-    const { queryId } = props.parameter;
+    const queryId = props.parameter.queryId;
```
arikfr: Is this change needed for eslint to detect the correct value for the dependencies array?

gabrieldutra: Yes, using it like this it automatically corrects to `[props.parameter.queryId]`

gabrieldutra:
@kravets-levko ath else you think it's important here?

PR #4426, opened 2019-12-08
Histogram

deecay:

- [x] Feature

## Description
New visualization: Histogram  by Plotly
No tests ready yet.

Example1: https://deploy-preview-4426--redash-preview.netlify.com/queries/270#493
Example2: https://deploy-preview-4426--redash-preview.netlify.app/queries/224#955

image

image

image

image

### DIFF
--- a/viz-lib/src/visualizations/chart/Editor/AxisSettings.tsx
+++ b/viz-lib/src/visualizations/chart/Editor/AxisSettings.tsx
@@ -1,13 +1,9 @@
-import { isString, isObject, isFinite, isNumber, merge } from "lodash";
+import { isString, isObject, merge } from "lodash";
 import React from "react";
 import { useDebouncedCallback } from "use-debounce";
 import * as Grid from "antd/lib/grid";
 import { Section, Select, Input, InputNumber, ContextHelp } from "@/components/visualizations/editor";
-
-function toNumber(value: any) {
-  value = isNumber(value) ? value : parseFloat(value);
-  return isFinite(value) ? value : null;
-}
+import { toNumber } from "../plotly/utils";
 
 type OwnProps = {
   id: string;
--- a/viz-lib/src/visualizations/chart/Editor/ChartTypeSelect.tsx
+++ b/viz-lib/src/visualizations/chart/Editor/ChartTypeSelect.tsx
@@ -6,6 +6,7 @@ import { visualizationsSettings } from "@/visualizations/visualizationsSettings"
 const allChartTypes = [
   { type: "line", name: "Line", icon: "line-chart" },
   { type: "column", name: "Bar", icon: "bar-chart" },
+  { type: "histogram", name: "Histogram", icon: "bar-chart" },
   { type: "area", name: "Area", icon: "area-chart" },
   { type: "pie", name: "Pie", icon: "pie-chart" },
   { type: "scatter", name: "Scatter", icon: "circle-o" },
--- a/viz-lib/src/visualizations/chart/Editor/GeneralSettings.tsx
+++ b/viz-lib/src/visualizations/chart/Editor/GeneralSettings.tsx
@@ -293,7 +293,7 @@ export default function GeneralSettings({ options, data, onOptionsChange }: any)
             label="Stacking"
             data-test="Chart.Stacking"
             defaultValue={options.series.stacking}
-            disabled={!includes(["line", "area", "column"], options.globalSeriesType)}
+            disabled={!includes(["line", "area", "column", "histogram"], options.globalSeriesType)}
             onChange={(stacking: any) => onOptionsChange({ series: { stacking } })}>
             {/* @ts-expect-error ts-migrate(2339) FIXME: Property 'Option' does not exist on type '({ class... Remove this comment to see the full error message */}
             <Select.Option value={null} data-test="Chart.Stacking.Disabled">
@@ -309,7 +309,7 @@ export default function GeneralSettings({ options, data, onOptionsChange }: any)
         </Section>
       )}
 
-      {includes(["line", "area", "column"], options.globalSeriesType) && (
+      {includes(["line", "area", "column", "histogram"], options.globalSeriesType) && (
         // @ts-expect-error ts-migrate(2745) FIXME: This JSX tag's 'children' prop expects type 'never... Remove this comment to see the full error message
         <Section>
           <Checkbox
--- a/viz-lib/src/visualizations/chart/Editor/XAxisSettings.tsx
+++ b/viz-lib/src/visualizations/chart/Editor/XAxisSettings.tsx
@@ -1,6 +1,6 @@
 import { includes } from "lodash";
 import React from "react";
-import { Section, InputNumber, Switch } from "@/components/visualizations/editor";
+import { Section, InputNumber, Switch, Input } from "@/components/visualizations/editor";
 import { EditorPropTypes } from "@/visualizations/prop-types";
 import { toNumber } from "../plotly/utils";
 
@@ -21,13 +21,13 @@ export default function XAxisSettings({ options, onOptionsChange }: any) {
         <React.Fragment>
           {/* @ts-expect-error ts-migrate(2745) FIXME: This JSX tag's 'children' prop expects type 'never... Remove this comment to see the full error message */}
           <Section>
-            <InputNumber
+            <Input
               label="Bin Size"
               className="w-100"
               placeholder="Auto"
               data-test="Chart.XAxis.BinSize"
               defaultValue={options.binSize}
-              onChange={(binSize: any) => onOptionsChange({ binSize: toNumber(binSize) })}
+              onChange={(e: any) => onOptionsChange({ binSize: e.target.value })}
             />
           </Section>
 
--- a/viz-lib/src/visualizations/chart/Editor/XAxisSettings.tsx
+++ b/viz-lib/src/visualizations/chart/Editor/XAxisSettings.tsx
@@ -1,6 +1,8 @@
+import { includes } from "lodash";
 import React from "react";
-import { Section, Switch } from "@/components/visualizations/editor";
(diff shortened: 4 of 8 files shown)


arikfr (commented):
Thanks, @deecay! I'm sure this will make some people very happy :-)

Considering this is less likely to be used alongside other chart types, how about we create a dedicated visualization type for it with its own settings tab? I think it will reduce confusion on how to use it and what are the options that are available.

deecay:
Could you advise what level of separation from chart visualization is necessary?

For example, I have added `function prepareHistogramSeries(series, options)` to `chart/plotly/prepareDefaultData.js`. Should I leave this function and the related code in the file here? Should I copy the whole file to `histogram` folder and leave the `chart` folder alone? If so, should I erase everything unrelated to histogram once I copy the file to `histogram` folder? Would it be better to move things to some common library?

kravets-levko:
@arikfr Why do you think that histogram series cannot be used with other series types? Brief googling found pictures like this one:

image

arikfr:
@kravets-levko it looks like a line representation of the histogram. Can you think of a real use case? @deecay do you have a use case for mixing them?

deecay:
@arikfr The curved line over histograms are density curves, which is a fairly common companion of the bar-histogram. The curve is a different representation of the same data set. I can't think of any other chart type mix to be honest.

arikfr:
> Could you advise what level of separation from chart visualization is necessary?

The most important part is to have a dedicated interface for the Histogram visualization. The implementation can still share code with charts if it makes sense and/or makes code simpler.

client/app/visualizations/chart/Editor/XAxisSettings.jsx
```diff
@@ -1,9 +1,15 @@
...
+import { Section, InputNumber, Switch } from '@/components/visualizations/editor';
 import { EditorPropTypes } from '@/visualizations';
 
 import AxisSettings from './AxisSettings';
 
+function toNumber(value) {
+  value = isNumber(value) ? value : parseFloat(value);
+  return isFinite(value) ? value : null;
+}
+
```
kravets-levko, line 12: Consider using existing utility for this: https://github.com/getredash/redash/blob/a1255b41442da27b8bff70f70018265b7a2a1540/viz-lib/src/visualizations/chart/plotly/utils.js#L5-L7

kravets-levko (commented):
Hi @deecay! Can you please resolve conflicts - I'd like to review it once more, and I guess we'll merge it. Thanks!

deecay:
@kravets-levko 

Resolved conflict and incorporated your point on utility function.

One thing I could not resolve is the issue around automatic update of the BinSize and BinStart. When an input value for these two inputs gets deleted in the editor, it should redraw the chart with default value but it doesn't. I've played around with debounce but I couldn't get it to work. Any help would be welcomed.

kravets-levko:
Thank you @deecay! I'll check the issue with inputs, hope there's nothing serious. One more thing I'd like to revive in this discussion - having a new Histogram visualization vs. Histogram chart type (within existing Chart viz) - what's your point on this?

deecay:
Hi @kravets-levko 

On chart types:
About a year ago I have tried to implement plotly charts (treemap and radar) using "new Visualization" method.
It was difficult. What I realized was that the current structure of (/chart/ and /plotly/) functions was not easy to re-use in a new Visualization. I had to copy decent amount of code from the chart visualization, such as chart resizing, chart refreshing, Data Labels Editor interface, etc, and then tweak it. It made me want to just copy the chart visualization entirely and modify that copied code. But this method obviously decreases maintainability.

Ideally, we could refactor the code so that adding a new plotly-based visualization can be done with relatively small effort, avoiding duplicate code. This PR, the existing Chart viz method, has just +100 lines of code, with almost no duplicate code. So this is my preferred method for now.

If you could show me and example and/or mentor me, I might be able to see things differently.

Eager to hear your thoughts.

viz-lib/src/visualizations/chart/Editor/XAxisSettings.jsx
```diff
@@ -14,6 +16,32 @@ export default function XAxisSettings({ options, onOptionsChange }) {
...
+      {includes(["histogram"], options.globalSeriesType) && (
+        <React.Fragment>
+          <Section>
+            <InputNumber
+              label="Bin Size"
+              className="w-100"
+              placeholder="Auto"
+              data-test="Chart.XAxis.BinSize"
+              defaultValue={options.binSize}
+              onChange={binSize => onOptionsChange({ binSize: cleanNumber(binSize) })}
```
deecay, line 28: @kravets-levko using toNumber() from AxisSettings resolves the non-updating issue mentioned earlier. Should I move the toNumber function as a common function and use it here?

kravets-levko: yes, sounds good

deecay: Done in https://github.com/getredash/redash/pull/4426/commits/f1a8fe6e20df2a1b8210e6a225876ff83e3cb426

borsi:
This feature is so close to completion, I can feel it. :)

mmcdermott:
Any ETA on this?

susodapop:
Currently only merging fixes directly related to the V10 beta. We will pipeline this for merge after the V10 release next month.

gaganc-on:
Any updates on this

thrau:
would be great if this would make it into a release!

susodapop:
Thanks for the ping! Will aim to merge and include this in V11 this summer.

beknazar:
@susodapop Any news on this PR? This would be a great feature

nathanlmeyers:
@susodapop yeah just wanted to bump this too!

fibr:
Bump. This feature would be very useful.

guidopetri:
@deecay , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
Yeah, this looks like it'd be really useful.  Hopefully it gets rebased. :smile:

eradman:
@deecay do you have time to rebase this change? There are quite a few conflicts, but all of them appear to be easy to resolve

deecay:
@guidopetri @justinclift 
Rebased.

justinclift:
Awesome! :smile:

@eradman Any interest in reviewing this PR?

eradman:
> Awesome! 😄
> 
> @eradman Any interest in reviewing this PR?

Yes, I'll take this change for a test drive

eradman:
I ran into some trouble with larger data sets on Chrome and Firefox. With as few as 500 entries both browsers would spin at 90% CPU for several minutes.

Now that I spent some time with this I feel that the "binning" function should be more generic. In SQL we can bin values in may ways. For example this bins by powers of 2

```sql
WITH series AS (
    SELECT
      power(2, series-1)::int AS lo,
      power(2, series)::int AS hi
    FROM generate_series(0, 10) AS series
)
SELECT count(files), lo || '-' || hi AS range
FROM dir_stats
JOIN series ON (files >= series.lo AND files < series.hi)
GROUP BY series, lo, hi
ORDER BY lo
```

This could be easily adapted to date ranges, range categories, and so on.

A bar chart works fine, except that I do not believe that Redash we supports multiple ranges yet, so perhaps this is the feature we should focus on.

Secondarily, binning options could be added, but I feel that would be somewhat complicated to support multiple _scales_ and _data types_.

deecay:
@eradman 
> I ran into some trouble with larger data sets on Chrome and Firefox. With as few as 500 entries both browsers would spin at 90% CPU for several minutes.

Does the CPU spin for test deployed instance too? (2000 entries)
https://deploy-preview-4426--redash-preview.netlify.com/queries/270#493

eradman:
> > I ran into some trouble with larger data sets on Chrome and Firefox. With as few as 500 entries both browsers would spin at 90% CPU for several minutes.
> 
> Does the CPU spin for test deployed instance too? (2000 entries) https://deploy-preview-4426--redash-preview.netlify.com/queries/270#493

I seem to have no trouble with that demo.  I'll try rebuilding my dev environment to see if this is repeatable

eradman:
Tried cleaning my test env (`git clean`, `docker rm -f ...`) and rebuilding dependencies I still hit the same problem with CPU. Is the demo link on netlify.com up to date?

What do the rest of you think?

My thoughts:

1. The histogram feature should be a data filter, not a new graph type (Apologies in an advance! I know this contradicts early feedback)
2. Binning on the client side won't scale very well, so this transform should happen on the server-side (this is the current limitation of pivot tables)
3. The X axis should be able to handle dates and numbers

deecay:
@eradman 
> Is the demo link on netlify.com up to date?
I think it is. And I think I've found the reason for the high CPU usage.

This problem happens when the X-axis data is specified as, or looks like, a date. In these cases plotly treats x-axis as `type: date`. The binning function in histogram treats the value of `bin size` differently between number axis and dates axis. If the x-axis is numbers, size: 1 means 1.0. But if x-axis is date, then size: 1 means 1ms (!). Therefore, when you have a data with timeseries with range scale of months or years, plotly tries to bin that data with 1ms-bin and gets overloaded.

I will look into what I can do with date binning. (your thought #3)

deecay:
@eradman 

I have added some features.

1. Chart will not render if xAxis bin size is "too small". Here, "too small" means that the resulting bin count will be > 10000.
2. xAxis bin size now works for date axis, too. Possible values for date axis bin size are:  
   a. number in ms (86,400,000 will represent a day (1000 * 60 * 60 * 24) )
   b. "M1", "M2", etc. where M1 is one month, M2 is two months. 
   See Example2: https://deploy-preview-4426--redash-preview.netlify.app/queries/224#955

denisov-vlad:
@eradman @justinclift Could you review it please?

PR #4430, opened 2019-12-10
Sync QueryView and QuerySource

gabrieldutra:

- [x] Other

## Description
@kravets-levko this is a sync for the code, opening a PR since I was also working on some changes to simplify the Query Header component (moving style to `query-page-header.less`


gabrieldutra:
@kravets-levko although the QueryView page isn't done, please give it a check when possible to sync (mainly the changes in the `QueryPageHeader`) and share you opinion. [Link to View](https://deploy-preview-4430--redash-preview.netlify.com/queries-react/274) - [Link to Source](https://deploy-preview-4430--redash-preview.netlify.com/queries/274/source2)

client/app/pages/queries/components/QueryVisualizationTabs.jsx
```diff
@@ -0,0 +1,38 @@
...
+        </TabPane>
+      ))}
+    </Tabs>
+  );
+}
+
+QueryVisualizationTabs.propTypes = {
+  queryResult: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+  visualizations: PropTypes.arrayOf(PropTypes.object),
+  currentVisualizationId: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
kravets-levko: How could this prop be a string? Maybe, let's restrict it to number and cast type when passing props to this component?

gabrieldutra: The idea was to accept string id's and cast them in the component, but yeah, this can be done outside, so it's more consistent.

client/app/assets/less/inc/base.less
```diff
@@ -261,12 +261,6 @@ text.slicetext {
       favorites-control {
         float: left;
       }
-
-      h3 {
-        width: 100%;
-        margin-bottom: 5px !important;
-        display: block !important;
-      }
```
kravets-levko: I noticed some difference between old and new header version: when query does not have tags, tags control should stay on the same line with query name (see screenshots). Seems this style is related to that

image
image

kravets-levko: Seems it's better to keep markup from base branch, unless you have some better UX ideas for this component. Also, if you think that button group does not fit here - feel free to remove it.

gabrieldutra: It wraps in the older version too, just take more for it to happen :slightly_smiling_face:. That's due to the existing conflict with `.page-title { display: block }`, fixed in 37c9b469cad6e0b245e31fcc1be4fb5176bffeeb

kravets-levko: Now layout is good, but still there some differences:

existing
image

react
image

Notice: favorites control should be on the same line with query name, too much space between query name and tags, and tags should go under buttons on right side.

gabrieldutra: Yeah, in the original there are two `QueryTagsControl` components, one for the desktop position and another for the mobile one :sleepy:

gabrieldutra: Decided to come back with the old markup and move on for now

kravets-levko: Agreed. And I think - what if we'll use some HOC (?) to conditionally change a markup depending on viewport width (similar to what dashboard grid does)?

gabrieldutra: I usually feel css is more reliable than js when it can be used, do you have any advantages in mind?

client/app/pages/queries/components/QueryPageHeader.jsx
```diff
@@ -111,50 +114,46 @@ export default function QueryPageHeader({ query, sourceMode }) {
...
-            )}
-          </Button.Group>
+          {!query.isNew() && (
+            <Dropdown overlay={createMenu(moreActionsMenu)} trigger={['click']}>
+              <Button><Icon type="ellipsis" rotate={90} /></Button>
+            </Dropdown>
+          )}
         </div>
-        <span className={cx('query-tags__mobile', { 'query-tags__empty': query.tags.length === 0 })}>
+        <span className={cx('m-t-5 query-tags__mobile', { 'query-tags__empty': query.tags.length === 0 })}>
```
kravets-levko: Seems it's better to remove padding at all for this component (update `.query-tags__mobile` rules) instead of using `m-t-5` - tags and query name should be closer on small screens.

kravets-levko: Old:

image

This without fix:
image

This with fix:
image
image

client/app/assets/less/inc/base.less
kravets-levko: In this case the only advantage is that tags component will be rendered only once (+markup will simplified a bit). CSS Grid is the perfect solution, but it still is not supported good enough. For JS, there is a [`matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) API which is 100% supported in all our target browsers.

client/app/pages/queries/components/QueryPageHeader.jsx
gabrieldutra: :+1:

client/app/assets/less/inc/base.less
gabrieldutra: BTW you meant to use a hook, right? (like [this](https://usehooks.com/useMedia/) one)

It's an interesting hook to have anyway, I tried to use it once I wanted to check in JS if the browser was in Percy,

kravets-levko: Yes, something like that :slightly_smiling_face:

gabrieldutra: That hook is not hard to be applied in this context:
```jsx
const isMobile = useMedia(['(max-width: 880px)'], [true], false);
```

But I prefer keeping it in css for now as I don't want to break the angular page yet (we can reconsider when cleaning angular stuff)

PR #4432, opened 2019-12-10
Migrate schema browser to react

kravets-levko:

- [x] Refactor

## Description

Query Editor page: migrate data source select and schema browser.

Replaces getredash/redash#3140


client/app/assets/less/inc/schema-browser.less
```diff
@@ -72,8 +72,14 @@ div.table-name {
 
 .schema-control {
   display: flex;
+  flex-wrap: nowrap;
   padding: 0;
 
+  button {
```
gabrieldutra: Is there any other `<button>` or this refers to the Ant Button?

client/app/pages/queries/QuerySource.jsx
```diff
@@ -1,19 +1,123 @@
...
+  const [dataSources, setDataSources] = useState([]);
+  const dataSource = useMemo(() => (find(dataSources, { id: query.data_source_id }) || null), [query, dataSources]);
+  const [schema, setSchema] = useState([]);
+  const refreshSchemaTokenRef = useRef(null);
+
+  useEffect(() => {
+    recordEvent('view_source', 'query', query.id);
+
+    let isCancelled = false;
+    DataSource.query().$promise.then((data) => {
```
gabrieldutra: I was thinking if this could also go in the angular `resolve`. Universal Router supports [async routes](https://github.com/kriasoft/universal-router/blob/master/docs/api.md#async-routes) and that should be a fit.

My thoughts ended up in: if it's sth that is a "main" resource for the page that should be in there (basically when it's required for the page to work), otherwise it's better to have a "loading state". Thus, I eventually concluded that in this case it's better to render the rest of the page and have it as is, but wanted to share this anyway :stuck_out_tongue_closed_eyes: 

PS: It's better to have a loading state in the Data Source selector than showing the Data source id while it's loading.

client/app/pages/queries/QuerySource.jsx
```diff
@@ -1,19 +1,123 @@
...
       </div>
       <main className="query-fullscreen">
-        Content
+        <nav>
+          <div className="editor__left__data-source">
+            <Select
+              className="w-100"
+              value={query.data_source_id}
+              disabled={!query.can_edit}
+              onChange={handleDataSourceChange}
```
gabrieldutra:
```suggestion
              onChange={handleDataSourceChange}
              optionFilterProp="children"
              showSearch
```
(Hope that alone will allow the search)

client/app/pages/queries/QuerySource.jsx
kravets-levko: In this case, data source list is not a blocker for the page (there are some issues with QueryEditor component, but I'll address them later). Adding loader is a good idea :+1:

kravets-levko:
@arikfr Not yet. I'm going to do it today, and probably will need to add `react-virtualized` or `react-window` as a replacement for `angular-vs-repeat`

client/app/assets/less/inc/schema-browser.less
kravets-levko: Ant Button. Replaced this with `.ant-btn`

kravets-levko:
Added `react-virtualized` as a replacement for `angular-vs-repeat`. With 20k schema items (table + columns) it worked quite fast, but I tested on my desktop PC. With 200k it became noticeably slower, so I decided to add a virtual scroll library.

client/app/pages/queries/QuerySource.jsx
kravets-levko: Done

client/app/pages/queries/QuerySource.jsx
```diff
@@ -1,19 +1,128 @@
...
+      notification.error('Schema refresh failed.', 'Please try again later.');
+      return Promise.resolve([]);
+    });
+}
+
+function QuerySource(props) {
+  const [query, setQuery] = useState(props.query);
+  const [dataSources, setDataSources] = useState([]);
+  const [dataSourcesLoaded, setDataSourcesLoaded] = useState(false);
+  const dataSource = useMemo(() => (find(dataSources, { id: query.data_source_id }) || null), [query, dataSources]);
```
gabrieldutra: Need to find a way to handle this for the `new` page :slightly_smiling_face: 

Otherwise...
image

The behavior on the Angular page is to use `localStorage.lastSelectedDataSourceId || first(dataSources)`

kravets-levko: Thanks for pointing. ATM I'm testing both pages, so probably will find even more issues.

kravets-levko: Fixed :+1: Thank you once more!

client/app/pages/queries/QuerySource.jsx
```diff
@@ -76,11 +82,25 @@ function QuerySource(props) {
...
+      setQuery(newQuery);
+    }
   }, [query]);
 
+  useEffect(() => {
+    // choose data source id for new queries
+    if (dataSourcesLoaded && query.isNew()) {
+      const firstDataSourceId = dataSources.length > 0 ? dataSources[0].id : null;
+      handleDataSourceChange(chooseDataSourceId(
+        [query.data_source_id, localStorage.lastSelectedDataSourceId, firstDataSourceId],
```
gabrieldutra: I debugged it in here and there are two things happening that don't allow the localStorage to work: 1 - The localStorage values are always strings; 2 - When the value is `null` it replaces the local storage value, which shouldn't happen IMO as in the first `chooseDataSourceId` executions the DS's are not loaded, thus it will always return `null`.

image
-> illustrative image for the 2 above :)

kravets-levko: I replaced `useEffect` (which is async) with `useMemo` (sync) - this fixed the issue and also reduced amount of component updates. 

https://github.com/getredash/redash/pull/4432/commits/eb53220c1cc818e7cebaea78c8cc695bc575c1ee

client/app/pages/queries/QuerySource.jsx
```diff
@@ -1,19 +1,174 @@
...
+    [query]
+  );
+
+  useEffect(() => {
+    // choose data source id for new queries
+    if (dataSourcesLoaded && query.isNew()) {
+      const firstDataSourceId = dataSources.length > 0 ? dataSources[0].id : null;
+      handleDataSourceChange(
+        chooseDataSourceId(
+          [query.data_source_id, localStorage.getItem("lastSelectedDataSourceId"), firstDataSourceId],
```
gabrieldutra: Any practical difference using the getters/setters?

kravets-levko: [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API) recommends to use getter/setters, and I found that arguments strong enough. So I took the opportunity and replaced props with getters/setters.

PR #3776, opened 2019-05-07
WIP: Markdown template visualization

arikfr:
## What type of PR is this? (check all applicable)

- [x] Feature

## Description

Following @daldei's [suggestion in the forum](https://discuss.redash.io/t/single-row-view/311/6?u=arikfr), decided to take a stab at implementing a markdown/mustache template visualization.

This is the first 100% React visualization. :)

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

image

PR #4443, opened 2019-12-14
Map: add tooltip and popup templating

deecay:

- [x] Feature

## Description
Adds feature to define your own tooltip and popup format for Map Visualization

image

### DIFF
--- a/client/app/assets/less/inc/visualizations/map.less
+++ b/client/app/assets/less/inc/visualizations/map.less
@@ -7,3 +7,8 @@
     z-index: 0;
   }
 }
+
+.leaflet-popup-content img {
+  max-width: 100%;
+  height: auto;
+}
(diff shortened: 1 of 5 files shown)


deecay:
> 1. I think tooltip/popup inputs need some placeholder that will tell user that default template will be used. It may even use that default template as a placeholder. Also, the same should be mentioned in context help tooltip (e.g. "Leave this field empty to use default template").

tooltip is fine, but popup is an issue. Default value is something impossible to recreate in simple HTML. We would need some kind of template engine that can iterate over a list.

> 2. WDYT about using markdown instead of plain HTML for templates?

I assumed markdown cannot handle image layout very well, but I guess you can align images to the left etc. these days. There was a discussion of which library to use in https://github.com/getredash/redash/pull/3170. Are we going with mustache https://github.com/getredash/redash/pull/3776?

kravets-levko:
1 - let's add a "Leave this field empty to use default template" text to `TemplateFormatHint` component - I think it will be enough.

2 - noticed that Choropleth does not use markdown as well :sweat_smile: Let's keep it as is :+1:

deecay:
@kravets-levko, I've added placeholder, too.

image

50shashwat:
The format option is not showing up in the latest build. Did you guys removed it ?

image

PR #4445, opened 2019-12-14
Support download as TSV File

tsuyoshizawa:

- [x] Feature

## Description

Re:dash supports CSV/Excel formats to download.
CSV format is text type, but there are comma in the result.

It is sometimes noise, TSV format must be useful to see the raw result by pasting.

https://discuss.redash.io/t/can-i-download-result-as-text-format/5171

image
image
image

### DIFF
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -30,7 +30,7 @@ from redash.models.parameterized_query import (
 )
 from redash.serializers import (
     serialize_query_result,
-    serialize_query_result_to_csv,
+    serialize_query_result_to_dsv,
     serialize_query_result_to_xlsx,
 )
 
@@ -400,12 +400,12 @@ class QueryResultResource(BaseResource):
     @staticmethod
     def make_csv_response(query_result):
         headers = {"Content-Type": "text/csv; charset=UTF-8"}
-        return make_response(serialize_query_result_to_csv(query_result), 200, headers)
+        return make_response(serialize_query_result_to_dsv(query_result, ","), 200, headers)
 
     @staticmethod
     def make_tsv_response(query_result):
         headers = {"Content-Type": "text/tab-separated-values; charset=UTF-8"}
-        return make_response(serialize_query_result_to_csv(query_result, "\t"), 200, headers)
+        return make_response(serialize_query_result_to_dsv(query_result, "\t"), 200, headers)
 
     @staticmethod
     def make_excel_response(query_result):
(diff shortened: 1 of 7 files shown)


redash/handlers/query_results.py
```diff
@@ -368,6 +368,8 @@ def get(self, query_id=None, query_result_id=None, filetype="json"):
                 response = self.make_json_response(query_result)
             elif filetype == "xlsx":
                 response = self.make_excel_response(query_result)
+            elif filetype == "tsv":
```
rauchy: Reaching 4 conditional branches, it might be time to refactor this into a simpler assignment in the style of `response = response_builders[filetype](query_result)`

kravets-levko:
> Can you try to think of a more general name for that function?

Both CSV and TSV are subtypes of more general [DSV format](https://en.wikipedia.org/wiki/Delimiter-separated_values) - I think it's a good candidate for the function name.

rauchy:
@kravets-levko thank you 🙇‍♂️was looking for that :)

tsuyoshizawa:
Thank you for review.

I changed code your suggestion below.
1. rename to serialize_query_result_to_dsv
2. refactor filetype processing to be a simpler assignment

tsuyoshizawa:
> Also, I'd love to see some test coverage for this.

I missed this comment. I added some test cases.
I thought it was useless to copy all csv test cases, so I only added serialization tests for tsv.

kravets-levko:
@tsuyoshizawa could you please sync with master and resolve conflicts? Thanks!

tsuyoshizawa:
@kravets-levko Hi, I synced with master branch and fixed the conflict. 

```
 M client/app/components/queries/VisualizationEmbed.jsx
DU client/app/components/queries/visualization-embed.html
```

I only modified `VisualizationEmbed.jsx` file.
Could you check it?

tests/serializers/test_query_results.py
```diff
@@ -37,10 +37,14 @@ def test_doesnt_serialize_sensitive_keys_for_unauthenticated_users(self):
         self.assertSetEqual(set(["data", "retrieved_at"]), set(serialized.keys()))
 
 
-class CsvSerializationTest(BaseTestCase):
+class DsvSerializationTest(BaseTestCase):
     def get_csv_content(self):
```
rauchy: Could we avoid duplicating this function and just have something like `self.delimited_content("\t")`?

tsuyoshizawa: @rauchy Thank you for review. I replaced two functions that `get_csv_content` and `get_tsv_content` with one function names `delimited_content` following your suggestion.

Actually I was worried if I would replace one when I added the test code.

kravets-levko:
Thank you @tsuyoshizawa! :star2:

tsuyoshizawa:
Thank you for merge and cooperation to my PR!

PR #4429, opened 2019-12-10
Migrate Query pages to React

gabrieldutra:
## What type of PR is this? (check all applicable)

- [x] Refactor

## Description
Main PR for the Query Pages migration.

### Status

- [x] Query View page
- [x] Query Source page
  - [x] layout (resizable sections)
  - [x] overlays block (permissions and datasource alerts)
  - [x] unsaved changes alert when leaving page
  - [x] left section
    - [x] data source selector
    - [x] schema browser
      - [x] integration with query editor
    - [x] schedule
    - [x] description and metadata
  - [x] top-right section
    - [x] query editor
  - [x] bottom-right section
    - [x] visualizations
    - [x] actions and metadata
  - [x] Refine code
  - [x] Preserve page state when saving new query
- [x] Common components
  - [x] page header
  - [x] visualization tabs
  - [x] parameters
  - [x] query execution status
- [x] Desktop and mobile layout
- [x] Testing
- [x] Cleanup (PR getredash/redash#4512)

PR #4452, opened 2019-12-16
Migrate Query page to React: Save changes

kravets-levko:

- [x] Refactor

getredash/redash#4429


client/app/pages/queries/QuerySource.jsx
```diff
@@ -41,6 +42,98 @@ function chooseDataSourceId(dataSourceIds, availableDataSources) {
...
+      "description",
+      "name",
+      "data_source_id",
+      "options",
+      "latest_query_data_id",
+      "is_draft",
+    ]);
+  }
+
+  // omit pendingValue before saving
```
gabrieldutra:
```suggestion
  // omit parameter properties that don't need to be stored
```
(or whatever like this, I forgot to update this comment in the Parameter structure update PR :sweat_smile:)

client/app/pages/queries/QuerySource.jsx
```diff
@@ -41,6 +42,98 @@ function chooseDataSourceId(dataSourceIds, availableDataSources) {
...
+      onCancel: () => {
+        reject();
+      },
+      maskClosable: true,
+      autoFocusButton: null,
+    });
+  });
+}
+
+function saveQuery(query, data, options) {
```
gabrieldutra: These methods will be shared with the QueryView, so perhaps have them in a separate file?

kravets-levko: Of course :slightly_smiling_face: Already doing that :+1:

kravets-levko: Done

client/app/pages/queries/QuerySource.jsx
kravets-levko: Done :+1:

gabrieldutra:
@kravets-levko is there ath else you want to do in here, or :rocket:?

kravets-levko:
@gabrieldutra 

> is there ath else you want to do in here

Yes - implementing `saveQuery` function was a half of job; another half is to use it :slightly_smiling_face:  Also, I implemented fork/archive query actions. I think we could proceed with this :+1:

client/app/pages/queries/utils/archiveQuery.jsx
```diff
@@ -0,0 +1,44 @@
...
+      autoFocusButton: null,
+    });
+  });
+}
+
+function doArchiveQuery(query) {
+  // Prettier will put `.$promise` before `.catch` on next line :facepalm:
+  // prettier-ignore
+  return Query.delete({ id: query.id }).$promise
+    .then(() => Promise.resolve(extend(clone(query), { is_archived: true, schedule: null })))
```
gabrieldutra: I'd have preferred `Promise.resolve({ ...query, is_archived: true, schedule: null })`, but that works

kravets-levko: Spread operator will create a plain object and copy properties to it - so query will loose it's prototype with all methods (like `.isNew()`). Therefore I used `clone` + `extend` here

PR #4453, opened 2019-12-17
Migrate Query page to React: Visualization Tabs

gabrieldutra:

- [x] Refactor

## Description
This finishes changes to `VisualizationTabs` and adds a `useVisualizationTabHandler` that also handles url hash updates. (Didn't add the Visualization Dialog to Query View yet -- will add in a next PR along with other updates)

#4429

VIsualizationTabs

### DIFF
--- a/client/app/pages/queries/components/QueryVisualizationTabs.jsx
+++ b/client/app/pages/queries/components/QueryVisualizationTabs.jsx
@@ -7,8 +7,14 @@ import Button from "antd/lib/button";
 
 const { TabPane } = Tabs;
 
-export default function QueryVisualizationTabs({ visualizations, queryResult, selectedTab,
-  showNewVisualizationButton, onChangeTab, onClickNewVisualization }) {
+export default function QueryVisualizationTabs({
+  visualizations,
+  queryResult,
+  selectedTab,
+  showNewVisualizationButton,
+  onChangeTab,
+  onClickNewVisualization,
+}) {
   const tabsProps = {};
   if (find(visualizations, { id: selectedTab })) {
     tabsProps.activeKey = `${selectedTab}`;
--- a/client/app/pages/queries/components/QueryVisualizationTabs.jsx
+++ b/client/app/pages/queries/components/QueryVisualizationTabs.jsx
@@ -7,23 +7,26 @@ import Button from "antd/lib/button";
 
 const { TabPane } = Tabs;
 
-export default function QueryVisualizationTabs({ visualizations, queryResult, currentVisualizationId }) {
+export default function QueryVisualizationTabs({ visualizations, queryResult, selectedTab,
+  showNewVisualizationButton, onChangeTab, onClickNewVisualization }) {
   const tabsProps = {};
-  if (find(visualizations, { id: currentVisualizationId })) {
-    tabsProps.activeKey = `${currentVisualizationId}`;
+  if (find(visualizations, { id: selectedTab })) {
+    tabsProps.activeKey = `${selectedTab}`;
+  }
+
+  if (showNewVisualizationButton) {
+    tabsProps.tabBarExtraContent = (
+      <Button onClick={onClickNewVisualization}>
+        <i className="fa fa-plus" />
+        <span className="m-l-5 hidden-xs">New Visualization</span>
+      </Button>
+    );
   }
 
   const orderedVisualizations = orderBy(visualizations, ["id"]);
 
   return (
-    <Tabs
-      {...tabsProps}
-      tabBarExtraContent={
-        <Button>
-          <i className="fa fa-plus m-r-5" />
-          New Visualization
-        </Button>
-      }>
+    <Tabs {...tabsProps} onChange={activeKey => onChangeTab(+activeKey)}>
       {orderedVisualizations.map(visualization => (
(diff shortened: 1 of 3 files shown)


client/app/pages/queries/components/QueryVisualizationTabs.jsx
```diff
@@ -7,23 +7,26 @@ import Button from "antd/lib/button";
...
   return (
-    <Tabs
-      {...tabsProps}
-      tabBarExtraContent={
-        <Button>
-          <i className="fa fa-plus m-r-5" />
-          New Visualization
-        </Button>
-      }>
+    <Tabs {...tabsProps} onChange={activeKey => onChangeTab(+activeKey)}>
```
kravets-levko:
```suggestion
    <Tabs {...tabsProps} animated={false} tabBarGutter={0} onChange={activeKey => onChangeTab(+activeKey)}>
```

client/app/pages/queries/utils/useVisualizationTabHandler.js
```diff
@@ -0,0 +1,30 @@
...
+  const [selectedTab, setSelectedTab] = useState(+$location.hash() || firstVisualization.id);
+
+  useEffect(() => {
+    if ($location.hash() !== "add") {
+      updateUrlHash(selectedTab !== firstVisualization.id ? selectedTab : null);
+    }
+  }, [firstVisualization.id, selectedTab]);
+
+  useEffect(() => {
+    if ($location.hash() === "add") {
```
kravets-levko: This is a legacy which we should avoid in React code:

https://github.com/getredash/redash/blob/bc52b788892181f50d419048ce2879148feae6e7/client/app/pages/queries/view.js#L465-L470

Let's just drop it and I'll find another solution later

kravets-levko:
One more thing I noticed: `useVisualizationTabHandler` should watch url has changes and update current tab. ATM you could just use something like this:
```javascript
import { $rootScope, $location } from "@/services/ng";

useEffect(() => {
  const unwatch = $rootScope.$watch(
    () => $location.hash(),
    () => { /* update tab here */ },
  );
  return unwatch;
}, []);
```
Later, when migrating router, we'll replace it with `history` library

client/app/pages/queries/utils/useVisualizationTabHandler.js
gabrieldutra: I thought just the part below of it was the "TODO" part :sweat_smile:, my bad. Anyway, this makes my life easier as that `onAddInUrl` needed a different function in the View :grin:.

gabrieldutra:
> `useVisualizationTabHandler` should watch url has changes and update current tab

Why is that? Is there any other part in app that will change the URL?

So far I thought just the opposite way (and applied it): You update it on the App and the App updates the URL (no two-way).

kravets-levko:
> Is there any other part in app that will change the URL?

User _may_ change it - and app should reflect that change. For example, sometimes I modify URL directly :grin:

client/app/pages/queries/utils/useVisualizationTabHandler.js
kravets-levko: The only purpose of this logic was to open Add Visualization dialog for new queries: query is saved, url changes, Angular destroys old controller and creates new one - so it's the way to signal new controller that it should open Add Visualization dialog. This flow is related only to Query Source controller and it's impossible with Query View page (because it cannot open unsaved query)

gabrieldutra:
> > Is there any other part in app that will change the URL?
> 
> User _may_ change it - and app should reflect that change. For example, sometimes I modify URL directly

But in such case, it should update on reload, no?

kravets-levko:
Browser does not reload page when hash changes; Angular also does not do it for Query pages (because if we enable `reloadOnSearch` flag - Angular will destroy/create controller on each hash change).

gabrieldutra:
> Browser does not reload page when hash changes

Right, forgot this case was a hash and was thinking about the Dashboard page (I handled the same way in #4228). In there if you change the URL search you end up reloading the page (even though `reloadOnSearch` is false).

client/app/pages/queries/utils/useVisualizationTabHandler.js
```diff
@@ -0,0 +1,35 @@
...
+      () => $location.hash(),
+      () => {
+        const expectedHashValue = selectedTab !== firstVisualization.id ? `${selectedTab}` : null;
+        if ($location.hash() !== expectedHashValue) {
+          setSelectedTab(+$location.hash() || firstVisualization.id);
+        }
+      },
+    );
+    return unwatch;
+  }, [firstVisualization.id, selectedTab]);
```
kravets-levko: WDYT about merging these two `useEffect` calls? They have the same dependencies and their logic is quite related.

gabrieldutra: Done :+1:. LMK if you had more simplification in mind for that part

kravets-levko: I'd also inline `updateUrlHash` function, but it's up to you. Code looks good :+1:

yankeeguyu: Hi, this feature can be add to dashboard page? Thank you!

PR #4455, opened 2019-12-17
Migrate Query page to React: Query View

gabrieldutra:

- [x] Refactor

## Description
Ongoing work with the Query View (will rebase with #4453 once merged and then organize separate commits) 

TO-DO:
- [x] Sync with target branch
- [x] Make Ctrl+Enter trigger query execution

getredash/redash#4429


client/app/pages/queries/QueryView.jsx
```diff
@@ -78,9 +78,10 @@ function QueryView(props) {
             value={query.description}
             isEditable={query.can_edit}
             onDone={description => updateQueryDescription(query, description).then(setQuery)}
-            editor="textarea"
             placeholder="Add description"
             ignoreBlanks={false}
+            editorProps={{ autosize: { minRows: 2, maxRows: 4 } }}
```
kravets-levko: That's exactly what I had in mind :+1:

kravets-levko:
Technically everything is good: code is good, all controls work fine. The only thing that bothers be is "Execute" button and execution status. Execution status (blue alert) looks weird on its current place - I suggest to render it in the card instead of visualizations.

About Execute button (sorry for poor GIF quality) - when clicked, entire page "jumps" and then user has to scroll down to reach it again:
test

Feels that t needs to be placed above visualizations (maybe, even with entire actions+metadata block), or both at existing place and above visualizations.

gabrieldutra:
> I suggest to render it in the card instead of visualizations.

I'll explore this :+1: 

Regarding the Execute button: this one is complicated to find a place that is not below the results, it makes sense for it to be there (could also be above the visualizations, but it can be a bit polluted in that region).

I also had sometime thinking about what could create an annoying behavior in that page. What most concerned me was that whenever someone wants to trigger new results, it's necessary to go all the way down to do so (same happens for changing Parameter values, but it's all the way up) -- I tried to mitigate this by making the execute button and query results info sticky, but didn't like that much.

The fact that it jumps can be a bit annoying because it's not smooth, but it sort of makes sense to go to the beginning of the results when you trigger new ones. Also there should be very few cases in which an user will have to Execute a query multiple times before checking its results :sweat_smile:.

Anyway, if code and controls are fine, let's merge this to have it with the Query migration PR (so we have a full experience and avoid rebasing in any following change). We can then continue in an open discussion for it.

kravets-levko:
> Anyway, if code and controls are fine, let's merge this to have it with the Query migration PR (so we have a full experience and avoid rebasing in any following change). We can then continue in an open discussion for it.

Let's fix query execution status, and then merge

arikfr:
I think one possible solution here is to have an "internal" scroll for the visualization content while the bottom controls are always fixed to the bottom. Basically, like we had in the previous version. This has its downsides, so I was hoping to avoid it.

Let's see if Mohit will have any ideas (the results view is the first thing he's looking at) before we merge the parent PR.

PR #4461, opened 2019-12-17
Migrate Query page to React: Delete visualization button

gabrieldutra:

- [x] Refactor

## Description
It was missing the delete visualization button, just copied the existing styling:

vis-tabs-delete-btn

#4429


gabrieldutra:
Ant Tabs doesn't handle well size changes for the Tab Element (and the delete button is hidden for mobile):

tabs-width-change

Options:
- Leave it like this (such updates are not common)
- Use the `useMedia` hook to watch for width changes instead of css

client/app/pages/queries/components/QueryVisualizationTabs.jsx
```diff
@@ -29,12 +61,28 @@ export default function QueryVisualizationTabs({
...
 
   return (
-    <Tabs {...tabsProps} animated={false} tabBarGutter={0} onChange={activeKey => onChangeTab(+activeKey)}>
+    <Tabs
+      {...tabsProps}
+      className="query-visualization-tabs"
+      animated={false}
+      tabBarGutter={0}
+      onChange={activeKey => onChangeTab(+activeKey)}
+      destroyInactiveTabPane
```
gabrieldutra: Extra

client/app/pages/queries/components/QueryVisualizationTabs.jsx
```diff
@@ -29,12 +61,28 @@ export default function QueryVisualizationTabs({
...
+    >
       {orderedVisualizations.map(visualization => (
-        <TabPane key={`${visualization.id}`} tab={visualization.name}>
+        <TabPane
+          key={`${visualization.id}`}
+          tab={(canDeleteVisualizations && !isFirstVisualization(visualization.id)) ? (
+            <TabWithDeleteButton
+              visualizationName={visualization.name}
```
kravets-levko: Let's move this condition to `TabWithDeleteButton` component: add prop `canDelete: PropTypes.bool`, and use it as:
```javascript
tab={(
  <TabWithDeleteButton
    visualizationName={visualization.name}
    canDelete={canDeleteVisualizations && !isFirstVisualization(visualization.id)}
    onDelete={() => onDeleteVisualization(visualization)}
  />
)}
```

kravets-levko:
About tabs bug: I think it worth fixing now. We can either use suggested solution with `useMedia`, or just show remove button on small screens (not sure if this is a good idea though).

client/app/lib/hooks/useMedia.js
```diff
@@ -0,0 +1,42 @@
...
+  );
+
+  return value;
+}
+
+export function useMediaBreakpoint() {
+  return useMedia(
+    ['(min-width: 1200px)', '(min-width: 992px)', '(min-width: 768px)', '(min-width: 576px)'],
+    ['xl', 'lg', 'md', 'sm'],
+    'xs',
```
gabrieldutra: @kravets-levko there are probably better ways of providing this, but I wanted to give a "quick match" to the css selectors.

kravets-levko: I'd pass first two arrays as map (object: key => media query), and split it inside of `useMedia` hook (using `keys`/`values` or even updated code to use it as is). BTW - haven't you tried [this package](https://www.npmjs.com/package/use-media)?

gabrieldutra: > haven't you tried [this package](https://www.npmjs.com/package/use-media)?

I have now :slightly_smiling_face:

PR #4467, opened 2019-12-20
Migrate Query Source page to React: Visualization actions

gabrieldutra:

- [x] Other

## Description
@kravets-levko lmk your opinion about this option for those actions, since they are shared. Also took the opportunity to remove the `forceUpdate` usage by updating the query state.

(Code in QuerySource is more an example and not necessarily needs to go now)

--

--

### DIFF
--- a/client/app/pages/queries/utils/deleteQueryVisualization.js
+++ b/client/app/pages/queries/utils/deleteQueryVisualization.js
@@ -1,0 +1,15 @@
+import { clone, extend, filter } from "lodash";
+import { Visualization } from "@/services/visualization";
+import notification from "@/services/notification";
+
+export default function deleteQueryVisualization(query, visualization) {
+  return Visualization.delete({ id: visualization.id })
+    .$promise.then(() => {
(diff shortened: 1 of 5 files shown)


client/app/pages/queries/utils/deleteQueryVisualization.js
```diff
@@ -0,0 +1,15 @@
+import { clone, extend, filter } from "lodash";
+import { Visualization } from "@/services/visualization";
+import notification from "@/services/notification";
+
+export default function deleteQueryVisualization(query, visualization) {
+  return Visualization.delete({ id: visualization.id })
+    .$promise.then(() => {
+      const filteredVisualizations = filter(query.visualizations, v => v.id !== visualization.id);
+      return Promise.resolve(extend(clone(query), { visualizations: filteredVisualizations }));
```
kravets-levko: While working on getredash/redash#4468 I realized that `_.clone` is not enough to properly clone query instance, so I added a `.clone()` method to Query service. No need to fix it in this PR, but we should keep this in mind and fix in one of  subsequent PRs

PR #4470, opened 2019-12-20
Migrate Query pages to React: Execute query hook

gabrieldutra:

- [x] Other

### DIFF
--- a/client/app/pages/queries/utils/useQueryExecute.js
+++ b/client/app/pages/queries/utils/useQueryExecute.js
@@ -13,7 +13,9 @@ function getMaxAge() {
 export default function useQueryExecute(query) {
   const [queryResult, setQueryResult] = useState(query.getQueryResult(getMaxAge()));
   const queryResultData = useQueryResult(queryResult);
-  const isQueryExecuting = useMemo(() => !includes(["done", "failed"], queryResultData.status), [queryResultData.status]);
+  const isQueryExecuting = useMemo(() => !includes(["done", "failed"], queryResultData.status), [
+    queryResultData.status,
+  ]);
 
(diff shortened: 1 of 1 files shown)


client/app/pages/queries/utils/useQueryExecute.js
```diff
@@ -0,0 +1,33 @@
...
+    selectedQueryText => setQueryResult(query.getQueryResultByText(0, selectedQueryText)),
+    [query]
+  );
+
+  useEffect(() => {
+    if (!queryExecuting && queryResult.query_result.query === query.query) {
+      query.latest_query_data_id = queryResult.getId();
+      query.queryResult = queryResult;
+    }
+  }, [queryExecuting]); // eslint-disable-line react-hooks/exhaustive-deps
```
gabrieldutra: Not 100% sure why this logic exists, but it seems to keep the Query `latest_query_data` updated. (without having to GET the query again).

As for the `eslint-disable-line` I didn't want it to be triggering unnecessarily -- an alternative to such cases is creating a `usePrevious` hook.

client/app/pages/queries/utils/useQueryExecute.js
```diff
@@ -0,0 +1,33 @@
...
+
+  return { queryResult, queryResultData, queryExecuting, executeQuery, executeQueryWithText };
```
kravets-levko: Just a suggestion: `queryExecuting` => `isQueryExecuting`, `executeQueryWithText` => `executeAdHocQuery`

client/app/pages/queries/utils/useQueryExecute.js
kravets-levko: > it seems to keep the Query `latest_query_data` updated.

Yes, exactly.

PR #4491, opened 2019-12-25
Visualization Embed: Add option to hide timestamp

arikfr:
## What type of PR is this?

- [x] Feature

### DIFF
--- a/client/app/components/queries/VisualizationEmbed.jsx
+++ b/client/app/components/queries/VisualizationEmbed.jsx
@@ -47,7 +47,7 @@ VisualizationEmbedHeader.propTypes = {
 
 VisualizationEmbedHeader.defaultProps = { queryDescription: "" };
 
-function VisualizationEmbedFooter({ query, queryResults, updatedAt, refreshStartedAt, queryUrl }) {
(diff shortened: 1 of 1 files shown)


client/app/components/queries/VisualizationEmbed.jsx
```diff
@@ -76,15 +76,17 @@ function VisualizationEmbedFooter({ query, queryResults, updatedAt, refreshStart
...
     <div className="tile__bottom-control">
-      <span>
-        <a className="small hidden-print">
-          <i className="zmdi zmdi-time-restore" />{" "}
-          {refreshStartedAt ? <Timer from={refreshStartedAt} /> : <TimeAgo date={updatedAt} />}
-        </a>
-        <span className="small visible-print">
-          <i className="zmdi zmdi-time-restore" /> {formatDateTime(updatedAt)}
+      {!hideTimestamp && (
+        <span>
```
gabrieldutra: Just one note: currently this `<span>` is what moves the buttons to the right side, if you don't want them to be in the left side when there's no timestamp, just keep it empty :slightly_smiling_face:

arikfr: Thanks for noticing, but that's inline with the current implementation on SaaS, so safe to keep :)

PR #4493, opened 2019-12-26
Update EditInPlace to use Antd components

gabrieldutra:

- [x] Refactor

## Description
This updates EditInPlace to use Antd components instead of rd-form.

I considered using Antd's Typography, but it's not 100% the same thing and Typography doesn't seem to support multi-line with Shift+enter or Textarea as editor.

Anyway, most of the existing EditInPlace's didn't change, aside from the titles that now have a smaller font-size when being edited:

**Before**
before

**After**
after

Also took the opportunity to allow `className` in the component.

### DIFF
--- a/client/app/components/EditInPlace.jsx
+++ b/client/app/components/EditInPlace.jsx
@@ -2,15 +2,16 @@ import React from "react";
 import PropTypes from "prop-types";
 import { react2angular } from "react2angular";
 import { trim } from "lodash";
+import Input from "antd/lib/input";
 
 export class EditInPlace extends React.Component {
   static propTypes = {
     ignoreBlanks: PropTypes.bool,
     isEditable: PropTypes.bool,
-    editor: PropTypes.string.isRequired,
     placeholder: PropTypes.string,
     value: PropTypes.string,
     onDone: PropTypes.func.isRequired,
+    textArea: PropTypes.bool,
   };
 
   static defaultProps = {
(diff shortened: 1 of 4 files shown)


client/app/components/EditInPlace.jsx
```diff
@@ -1,23 +1,26 @@
...
 
 export class EditInPlace extends React.Component {
   static propTypes = {
     ignoreBlanks: PropTypes.bool,
     isEditable: PropTypes.bool,
-    editor: PropTypes.string.isRequired,
     placeholder: PropTypes.string,
     value: PropTypes.string,
     onDone: PropTypes.func.isRequired,
+    textArea: PropTypes.bool,
```
kravets-levko: How about calling it `multiline`? Also, let's add `editorProps: PropTypes.object` and pass it to editor component when in editing mode - it will allow to use features like `Textarea.autoSize` when needed.

gabrieldutra:
> WDYT about rewriting this component to function-based? Will it make code anyhow better/cleaner?

I don't think it's worth it, the component looks simple already :slightly_smiling_face:

client/app/components/EditInPlace.jsx
gabrieldutra: I don't care that much about the `textArea` toggle naming (though `textArea` itself is a bit weird), so updated it; `editorProps` is a good one.

Done :+1:

kravets-levko: I also don't mind about using multi-word/camel-case prop names. But in this case `textArea` was too much about internal implementation rather than compoonent's API.

PR #4494, opened 2019-12-26
Show error origin

rauchy:
- [x] Feature

## Description
When executing queries, some errors may occur from Redash's processing code, but most errors occur in the executing database. This PR gives clear indication of the error's origin.

* I've based this on `execute-query-in-rq` because it included the job serialization refactoring there's not much point in supporting this for Celery.

#4413 

Errors caused by Redash code:
image

Errors caused by query runner code:
image

### DIFF
--- a/redash/serializers/__init__.py
+++ b/redash/serializers/__init__.py
@@ -69,6 +69,7 @@ class Serializer(object):
 
 
 class QuerySerializer(Serializer):
+
     def __init__(self, object_or_list, **kwargs):
         self.object_or_list = object_or_list
         self.options = kwargs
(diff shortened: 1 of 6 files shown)


redash/serializers/__init__.py
```diff
@@ -249,20 +250,22 @@ def serialize_job(job):
         updated_at = 0
 
     status = STATUSES[job_status]
+    error_origin = ''
+    error = ''
 
     if isinstance(job.result, Exception):
+        error_origin = 'Data source' if isinstance(job.result, QueryExecutionError) else 'System'
```
arikfr: We use `QueryExecution` error in the catch all handler as well, so assuming that anything with `QueryExecutionError` is non system is probably very optimistic. Also some errors that we do handle in the query runners might be on our end. I think this is logic that should be implemented at the Query Runner level and each Query Runner will qualify the error messages.

guidopetri:

justinclift:
The feedback from @arikfr's review above sounds like this PR needs to be reworked differently.

We can probably just close it. :smile:

PR #4497, opened 2019-12-26
Replace $http with axios

gabrieldutra:

- [x] Refactor

## Description
Replace $http and other related angular dependencies with axios ($resource and $q)

- [x] Replace plain React usages of $http
- [x] Update Services (remove angular dependencies and refactor them in a non-angular way)
  - [x] AlertDestination
  - [x] DataSource
  - [x] AlertSubscription
  - [x] Alert
  - [x] recordEvent/Event
  - [x] getTags
  - [x] QuerySnippet
  - [x] Group
  - [x] User
  - [x] Query
  - [x] Widget
  - [x] QueryResult
  - [x] Visualization
  - [x] Dashboard
  - [x] Auth (can't be entirely updated yet - depends on router and app-view)
- [x] Update $http apiKey interceptor with axios interceptor
- [x] Update item.$favorite & item.$unfavorite calls
- [x] Update with React Query Pages
- [x] Cleanup and review own code

#3071 

--


client/app/services/query-snippet.js
```diff
@@ -1,8 +1,12 @@
-export let QuerySnippet = null; // eslint-disable-line import/no-mutable-exports
+import axios from "@/services/axios";
+import { extend, map } from "lodash";
 
-function QuerySnippetService($resource) {
-  const resource = $resource("api/query_snippets/:id", { id: "@id" });
-  resource.prototype.getSnippet = function getSnippet() {
+class QuerySnippet {
```
gabrieldutra: Example of simple class in a service (this one is read-only, so has no issues).

I could rewrite the `getSnippet` in a non-class way, but wanted to show a simple class in here (for other services it's the same base idea, but a lot more complex)

client/app/components/dashboards/CreateDashboardDialog.jsx
```diff
@@ -36,7 +37,7 @@ function CreateDashboardDialog({ dialog }) {
     if (name !== "") {
       setSaveInProgress(true);
 
-      $http.post("api/dashboards", { name }).then(({ data }) => {
+      axios.post("api/dashboards", { name }).then(({ data }) => {
```
kravets-levko: I think that things like this should be moved to services (e.g. have a method called `updateDashboard` of whatever) - there should be no direct HTTP calls in components. There are a lot of similar places in other components as well

client/app/services/query-snippet.js
```diff
@@ -13,17 +17,16 @@ function QuerySnippetService($resource) {
...
+const getQuerySnippet = querySnippet => new QuerySnippet(querySnippet);
 
-  ngModule.run($injector => {
-    QuerySnippet = $injector.get("QuerySnippet");
-  });
-}
+const QuerySnippetService = {
+  query: () => axios.get("api/query_snippets").then(data => map(data, getQuerySnippet)),
+  create: data => axios.post("api/query_snippets", data).then(getQuerySnippet),
+  save: data => axios.post(`api/query_snippets/${data.id}`, data).then(getQuerySnippet),
```
kravets-levko: Interesting solution :+1: But you should keep in mind that some API calls do not return a full updated instance (for performance reasons), but only updated fields. I believe, only query and dashboard API do so, but anyway. Also, in some places we started using a (sort of) pattern: we implement a method that updates some instance; it accepts instance and a list of updates, makes a request to backend, then creates a copy of instance and picks from server's response only fields passed in arguments (or few others, like `version` for queries and dashboard). I think we should not try to mimic Angular resource here and try to use this pattern where possible.

client/app/components/dashboards/CreateDashboardDialog.jsx
arikfr: You're right. It was wrong to use `$http` all over. But wherever things are just working, we should keep things as they were (except for adopting `axios` so we can drop `$http`). In parallel, let's decide on a pattern we want to use for our API calls/resources, that will be more inline with how React works (returning immutable objects for example). Any suggestions for such pattern? Let's discuss in the issue.

Wherever things are not working as is or we have to jump through hoops to make it work, we can start adopting the new pattern. The rest we will update when we are done with the React migration.

client/app/services/query-snippet.js
gabrieldutra: I think the simpler ones don't need much more than this.

My thoughts for this part was to try to keep every response returning a new valid object and minimizing the changes in code. The update pattern is interesting, but also depends on each resource... Wdyt of simplifying it here and postpone this update to when we define a standard for an ideal api resource layer?

It's not "exactly" an Angular resource mimic :sweat_smile:, the class in this case was more to show how it could be done, only the API requests change the object, so in this case it's not an issue :). Angular resources also had the api methods in its prototype (`$save`, `$delete`, etc), those are quickly changeable, I'm more concerned about the ones that have lots of extra stuff in the prototype (I'll share comments when I reach them :slightly_smiling_face:)

client/app/components/EditParameterSettingsDialog.jsx
```diff
@@ -77,7 +77,7 @@ function EditParameterSettingsDialog(props) {
   useEffect(() => {
     const queryId = props.parameter.queryId;
     if (queryId) {
-      Query.get({ id: queryId }, query => {
+      Query.get({ id: queryId }).then(query => {
         setInitialQuery(query);
       });
```
kravets-levko:
```javascript
Query.get({ id: queryId }).then(setInitialQuery);
```

client/app/components/QuerySelector.jsx
```diff
@@ -14,14 +14,14 @@ const { Option } = Select;
 function search(term) {
   // get recent
   if (!term) {
-    return Query.recent().$promise.then(results => {
+    return Query.recent().then(results => {
       const filteredResults = results.filter(item => !item.is_draft); // filter out draft
       return Promise.resolve(filteredResults);
     });
```
kravets-levko: In this and other places `return Promise.resolve(...)` was used because `$q` had some issues with promise chaining (ATM I don't remember which, but returning new promise solved them). Now, when `$q` is not used anymore, such code could be simplified:
```javascript
return Query.recent().then(results => results.filter(item => !item.is_draft)); // filter out draft
```

client/app/components/QuerySelector.jsx
```diff
@@ -14,14 +14,14 @@ const { Option } = Select;
...
   }
 
   // search by query
-  return Query.query({ q: term }).$promise.then(({ results }) => Promise.resolve(results));
+  return Query.query({ q: term }).then(({ results }) => Promise.resolve(results));
```
kravets-levko:
```js
return Query.query({ q: term }).then(({ results }) => results);
```

client/app/pages/alerts/AlertsList.jsx
```diff
@@ -121,7 +121,7 @@ export default function init(ngModule) {
             return Alert.query.bind(Alert);
           },
           getItemProcessor() {
-            return item => new Alert(item);
+            return item => item;
```
kravets-levko: This is default behavior, so entire `getItemProcessor` could be removed

client/app/pages/groups/GroupDataSources.jsx
```diff
@@ -244,7 +244,7 @@ export default function init(ngModule) {
               return Group.dataSources.bind(Group);
             },
             getItemProcessor() {
-              return item => new DataSource(item);
+              return item => item;
```
kravets-levko: This is default behavior, `getItemProcessor` could be removed

client/app/pages/groups/GroupMembers.jsx
```diff
@@ -207,7 +207,7 @@ export default function init(ngModule) {
               return Group.members.bind(Group);
             },
             getItemProcessor() {
-              return item => new User(item);
+              return item => item;
```
kravets-levko: This is default behavior, `getItemProcessor` could be removed

client/app/pages/groups/GroupsList.jsx
```diff
@@ -146,7 +146,7 @@ export default function init(ngModule) {
               return Group.query.bind(Group);
             },
             getItemProcessor() {
-              return item => new Group(item);
+              return item => item;
```
kravets-levko: This is default behavior, `getItemProcessor` could be removed

client/app/pages/query-snippets/QuerySnippetsList.jsx
```diff
@@ -210,7 +212,7 @@ export default function init(ngModule) {
               return QuerySnippet.query.bind(QuerySnippet);
             },
             getItemProcessor() {
-              return item => new QuerySnippet(item);
+              return item => item;
```
kravets-levko: This is default behavior, `getItemProcessor` could be removed

client/app/pages/users/UsersList.jsx
```diff
@@ -276,7 +276,7 @@ export default function init(ngModule) {
               return User.query.bind(User);
             },
             getItemProcessor() {
-              return item => new User(item);
+              return item => item;
```
kravets-levko: This is default behavior, `getItemProcessor` could be removed

client/app/services/events.js
```diff
@@ -1,5 +1,7 @@
 import recordEvent from "@/services/recordEvent";
 
+// ANGULAR_REMOVE_ME delete this file - it replaced with recordEvent
```
kravets-levko: I think it's time to remove this file (and replace usages with `recordEvent` of course) :slightly_smiling_face:

client/app/services/query-result.js
```diff
@@ -35,418 +39,398 @@ function getColumnFriendlyName(column) {
...
 
-  class QueryResult {
-    constructor(props) {
-      this.deferred = $q.defer();
-      this.job = {};
-      this.query_result = {};
-      this.status = "waiting";
+class QueryResult {
+  constructor(props) {
+    this.deferred = $q.defer(); // TODO: replace this defer
```
kravets-levko: `$q.defer()` is and old API that was later replaced with `Promise`. Replacement pattern is something like:
```js
function defer() {
  const result = {
    promise: new Promise((resolve, reject) => {
      result.resolve = resolve;
      result.reject = reject;
    });
  };
  return result;
}
```

client/app/services/widget.js
```diff
@@ -75,194 +76,185 @@ export const ParameterMappingType = {
...
-    constructor(data) {
-      // Copy properties
-      each(data, (v, k) => {
-        this[k] = v;
-      });
+  constructor(data) {
+    // Copy properties
+    each(data, (v, k) => {
+      this[k] = v;
+    });
```
kravets-levko:
```js
extend(this, data);
```

client/app/services/events.js
gabrieldutra: I think the only place that has those events without `recordEvent` is the Angular query pages, the comment seemed worth it, since the usages will be removed as soon as the react version gets merged :slightly_smiling_face:.

client/app/services/query-result.js
gabrieldutra: I was actually thinking if I was going to keep the defer pattern with something like that or replace it with something else

client/app/lib/defer.js
```diff
@@ -0,0 +1,8 @@
+export default function defer() {
```
kravets-levko: Query service is the only place where this is needed. I'd just add this function there

client/app/services/axios.js
```diff
@@ -0,0 +1,24 @@
+import axiosLib from "axios";
+
+export const axios = axiosLib.create();
+
+const getData = ({ data }) => data;
+const getResponse = ({ response }) => Promise.reject(response);
```
kravets-levko: Is this needed to get a response object instead of error when request fails? If yes - I don't think it's so good idea. Angular's $http had similar behavior, and that's why we needed that PromiseRejectionError - to convert response to a Error instance, because in terms of error handling it's much more convenient if all promises are rejected with Error objects - global error handler does not need to perform additional checks, and if you handle them locally in component - you alread know what type of error you have and can, say, get `error.response` there.

gabrieldutra: I mostly returned the response like this because it was the behavior $http had and I wanted to minimize the risks -- I suppose we hardly ever test errors in the application (unlike the success flow), it's a point to focus for future and add tests to alternative/error paths as well.

Regarding the global promise handler, I agree, it's easy to make sure it's indeed an error and show an error page in the end when the response is an Error instance. I think we can try to remove it when handling errors globally (not sure if you're doing this in #4525, I still need to sync myself to everything that's in there again).

In brief, I believe this is correct way to do it, but my fear is that changing this may cause issues like unhandled promises showing errors where you didn't want them and error notifications not showing what they should.

kravets-levko: Got it :+1: Let's merge it, and revisit it later if needed.

PR #4498, opened 2019-12-26
Fix: properly encode UTF-8 filenames in query results request

### LINKED ISSUE #4332: Queries with unicode in query name fails downloading result file
This just happened on deploy preview. Sentry error:
https://sentry.io/share/issue/b8268b7c97784a30b67c512332a8d779/

Some possible solutions:

1. Implement a solution similar to pallets/flask#2223.
2. Switch to using Flask's `send_file` method instead of handling this on our own (need to make sure that we always have a file-like object at hand and that the behavior consistent with our current one).
3. Try quoting the filename, as suggested here: https://github.com/benoitc/gunicorn/issues/1214#issuecomment-238193505.

3 is the simplest, 2 is probably the most future proof.

arikfr:

- [x] Bug Fix

## Description

Ended up copying the implementation from Flask's `send_file` helper function, because `send_file` doesn't really fit our use case.

Closes #4332.

### DIFF
--- a/tests/handlers/test_query_results.py
+++ b/tests/handlers/test_query_results.py
@@ -29,6 +29,16 @@ class TestQueryResultsCacheHeaders(BaseTestCase):
         self.assertEqual(404, rv.status_code)
 
 
+class TestQueryResultsContentDispositionHeaders(BaseTestCase):
+    def test_supports_unicode(self):
+        query_result = self.factory.create_query_result()
+        query = self.factory.create_query(name="עברית", latest_query_data=query_result)
+
+        rv = self.make_request("get", "/api/queries/{}/results.json".format(query.id))
+        # This is what gunicorn will do with it
+        rv.headers['Content-Disposition'].encode('ascii')
+
+
 class TestQueryResultListAPI(BaseTestCase):
     def test_get_existing_result(self):
         query_result = self.factory.create_query_result()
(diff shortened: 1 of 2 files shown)


tests/handlers/test_query_results.py
```diff
@@ -29,6 +29,16 @@ def test_returns_404_if_no_cached_result_found(self):
...
 
 
+class TestQueryResultsContentDispositionHeaders(BaseTestCase):
+    def test_supports_unicode(self):
+        query_result = self.factory.create_query_result()
+        query = self.factory.create_query(name="עברית", latest_query_data=query_result)
+
+        rv = self.make_request("get", "/api/queries/{}/results.json".format(query.id))
+        # This is what gunicorn will do with it
+        rv.headers['Content-Disposition'].encode('ascii')
```
rauchy: Is this an implicit assertion?

arikfr: Yes. I don't care to test the content, sort of assume it's fine. I just want to make sure it doesn't raise the exception.

rauchy: I see, in that case - how about:

```suggestion
        try:
            rv.headers['Content-Disposition'].encode('ascii')
        except Exception as e:
            self.fail(repr(e))            
```

PR #4499, opened 2019-12-27
Migrate Query Source page to React: refine code

kravets-levko:

- [x] Refactor

## Description

Refine Query Source implementation: extract some logic to hooks/utils, refine code, etc. Implement proper permission checks.

getredash/redash#4429


client/app/pages/queries/hooks/useQueryDataSources.js
```diff
@@ -0,0 +1,29 @@
...
+        setAllDataSources(data);
+      }
+    });
+
+    return () => {
+      cancelDataSourceLoading = true;
+    };
+  }, []);
+
+  return useMemo(() => ({ dataSourcesLoaded, dataSources, dataSource }), [dataSourcesLoaded, dataSources, dataSource]);
```
gabrieldutra: This `useMemo` was to avoid unnecessary updates when listing dependencies with the hook value, right?

client/app/components/queries/QueryEditor/index.jsx
```diff
@@ -15,7 +15,7 @@ const QueryEditor = React.forwardRef(function(
   ref
 ) {
   const [container, setContainer] = useState(null);
-  const editorRef = useRef(null);
+  const [editorRef, setEditorRef] = useState(null);
```
gabrieldutra: Why change it to state?

client/app/pages/queries/hooks/useQueryDataSources.js
gabrieldutra: I think it's possible to use only the attributes in the dependencies without having to disable eslint:
```js
[hook.value] /* instead of */ [hook]
```

LMK you already tried this :)

Even with the `useMemo`, it makes sense since it would avoid unnecessary updated when a non-related value changes.

client/app/components/queries/QueryEditor/index.jsx
kravets-levko: It was needed to implemented cleanup: `useEffect` does not see changes when ref value changes (e.g. when `someRef.current` changes - component will not re-render).

client/app/pages/queries/hooks/useAddNewParameterDialog.js
```diff
@@ -0,0 +1,24 @@
+import { isFunction, map } from "lodash";
+import { useCallback, useRef } from "react";
+import EditParameterSettingsDialog from "@/components/EditParameterSettingsDialog";
+
+export default function useAddNewParameterDialog(query, onParameterAdded) {
+  const onParameterAddedRef = useRef(null);
+  onParameterAddedRef.current = isFunction(onParameterAdded) ? onParameterAdded : () => {};
```
gabrieldutra: Confirming: is this ref to make sure the most recent `onParameterAdded` is used in the promise?

kravets-levko: Yes, exactly

client/app/pages/queries/components/QueryPageHeader.jsx
```diff
@@ -53,133 +59,126 @@ function createMenu(menu) {
...
+  const queryFlags = useQueryFlags(query, dataSource);
 
-  const saveTags = useCallback(
-    tags => {
-      updateQuery(query, { tags }).then(onChange);
-    },
-    [query, onChange]
-  );
+  const updateName = useRenameQuery(query, onChange);
+
```
gabrieldutra: I prefer it without that many `\n`, but up to you :slightly_smiling_face:

client/app/pages/queries/hooks/useDeleteVisualization.js
```diff
@@ -0,0 +1,22 @@
...
+export default function useDeleteVisualization(query, onChange) {
+  const onChangeRef = useRef(null);
+  onChangeRef.current = isFunction(onChange) ? onChange : () => {};
+
+  return useCallback(
+    visualizationId =>
+      Visualization.delete({ id: visualizationId })
+        .$promise.then(() => {
+          const filteredVisualizations = filter(query.visualizations, v => v.id !== visualizationId);
+          onChangeRef.current(extend(clone(query), { visualizations: filteredVisualizations }));
```
gabrieldutra:
```suggestion
          onChangeRef.current(extend(query.clone(), { visualizations: filteredVisualizations }));
```

There's another in here too:
https://github.com/getredash/redash/blob/b20046593f7e0e04dbf2413c8915dd0cde47e554/client/app/components/queries/ApiKeyDialog/index.jsx#L25

client/app/pages/queries/hooks/useDeleteVisualization.js
```diff
@@ -0,0 +1,22 @@
+import { clone, extend, filter, isFunction } from "lodash";
+import { useRef, useCallback } from "react";
+import { Visualization } from "@/services/visualization";
+import notification from "@/services/notification";
+
+export default function useDeleteVisualization(query, onChange) {
+  const onChangeRef = useRef(null);
+  onChangeRef.current = isFunction(onChange) ? onChange : () => {};
```
gabrieldutra: It would be interesting to create a custom hook to do this (should make it clearer on what's it for). But it's still only 2 lines and it has to return the ref itself as it needs to be an object, benefits are nothing, but the the naming.

client/app/pages/queries/hooks/useDeleteVisualization.js
kravets-levko: Fixed. Thank you!

client/app/pages/queries/components/QueryPageHeader.jsx
kravets-levko: Fixed :slightly_smiling_face:

PR #4502, opened 2019-12-28
Replace angular-sanitize with DOMPurify

gabrieldutra:

- [x] Refactor

## Description
Initially an experiment, if there are no issues, we can go with it


arikfr:
Are there any known differences between DOMPurify and $sanitize?

gabrieldutra:
> Are there any known differences between DOMPurify and $sanitize?

They both seem to work by parsing the html and only allowing a whitelist/blacklist of tags. One thing I noticed is that [DOMPurify is not meant to be used with AngularJS](https://github.com/cure53/DOMPurify/wiki/Security-Goals-&-Threat-Model) - probably due to its different directives. That shouldn't be an issue, considering we will not have it very soon.

arikfr:
:+1:

PR #4510, opened 2019-12-30
Run prettier on cypress folder

gabrieldutra:

- [x] Other

## Description
I was waiting for other PRs containing cypress tests to be merged before doing this (to avoid conflicts). Now it should be ok.


arikfr:
Don't you need to update `package.json` and maybe restyled as well?

gabrieldutra:
> restyled

This is actually just `npm run prettier` after #4458. Restyled seems to include Cypress folder, but I'll confirm.

PR #4518, opened 2020-01-02
Add error boundary to catch errors in visualizations

kravets-levko:

- [x] Feature
- [x] Bug Fix
- [x] Other

image
image

### DIFF
--- a/client/app/components/ErrorBoundary.jsx
+++ b/client/app/components/ErrorBoundary.jsx
@@ -1,0 +1,71 @@
+import { isFunction } from "lodash";
+import React from "react";
+import PropTypes from "prop-types";
+import debug from "debug";
+import Typography from "antd/lib/typography";
+import Alert from "antd/lib/alert";
+
+const logger = debug("redash:errors");
+
+export const ErrorBoundaryContext = React.createContext({
+  handleError: error => {
+    throw error;
+  },
+});
+
+export default class ErrorBoundary extends React.Component {
+  static propTypes = {
+    children: PropTypes.node,
+    renderError: PropTypes.func, // error => ReactNode
+  };
+
+  static defaultProps = {
+    children: null,
+    renderError: null,
+  };
+
+  state = { error: null };
+
+  handleError = error => {
+    this.setState(this.constructor.getDerivedStateFromError(error));
+    this.componentDidCatch(error, null);
+  };
+
+  reset = () => {
+    this.setState({ error: null });
+  };
+
+  static getDerivedStateFromError(error) {
+    return { error };
+  }
+
+  componentDidCatch(error, errorInfo) {
+    logger(error, errorInfo);
+  }
+
+  renderError(error) {
+    if (isFunction(this.props.renderError)) {
+      return this.props.renderError(error);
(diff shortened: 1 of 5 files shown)


client/app/components/ErrorBoundary.jsx
```diff
@@ -0,0 +1,71 @@
...
+  renderError(error) {
+    if (isFunction(this.props.renderError)) {
+      return this.props.renderError(error);
+    }
+    return (
+      <Alert
+        message="Something went wrong."
+        description={
+          <Typography.Text style={{ display: "block" }} ellipsis>
+            {error.message}
```
arikfr: Let's avoid showing users our dirty laundry and only log it (with `debug`). In the future it makes sense to send this into Sentry.

kravets-levko: image

kravets-levko: @arikfr Maybe, we could show some message instead - like "Please check visualization options"/"read docs"/link to knowledge base/"Your PC is tired, turn it off"/etc?

arikfr: I would go with either one of:

1. `Error while rendering visualization. Please check visualization options and/or dataset.`
2. `Error while rendering visualization.`

I'm not sure about suggesting checking the option as usually an error boundary will be triggered due to a bug, no?

kravets-levko: Yes, mostly that will be bugs. But I think there are situations when checking options may help - for example, when visualization expects some column to be, say, numeric, but it's date/time. Of course there should be sanity check in the visualization's code, but in this case user can just use another column or update query. Anyway, we could leave just a "Something went wrong" message

arikfr: "Error while rendering visualization." is probably a bit better than the generic "something went wrong" as it gives a direction to what went wrong :)

kravets-levko: Done https://github.com/getredash/redash/pull/4518/commits/7c5024b86438f51397ba30612c21fde0cf9e820f

image

client/app/visualizations/chart/Renderer/PlotlyChart.jsx
```diff
@@ -1,42 +1,65 @@
...
 
 import getChartData from "../getChartData";
 import { Plotly, prepareData, prepareLayout, updateData, applyLayoutFixes } from "../plotly";
 
+function catchErrors(func, errorHandler) {
+  return (...args) => {
+    try {
+      return func(...args);
+    } catch (error) {
+      errorHandler.handleError(error);
```
gabrieldutra: What is this part for? Overall it seemed it only needed a `<ErrorBoundary>` component as a fallback in the tree (no need to use context for example), so was there any issues or there are any gains when you add this part?

kravets-levko: Error boundary will catch only errors that happened during rendering; this part is needed to catch async errors as well

arikfr:
:+1:

PR #4524, opened 2020-01-07
Hide unavailable page links to non-admin users in settings and header

gabrieldutra:

- [x] Bug Fix

## Description
Found this while testing the $http -> axios PR, since it was quick and I had a few minutes, just pushed.

Since #4323, tab links in settings were appearing to all users (data sources and alert destinations for example). This PR hides them again.

I can add a Cypress test to it tomorrow if you think it's worth it :)

--

--


arikfr:
> I can add a Cypress test to it tomorrow if you think it's worth it :)

Probably worth it.

PR #4525, opened 2020-01-07
Migrate router and <app-view> to React

### LINKED ISSUE #4159: React Migration: Routing library
As we're dropping Angular, we need to pick a new routing library. Some of the options are:

* [react-router](https://github.com/ReactTraining/react-router)
* [Buttermilk](https://buttermilk.js.org/)
* [Universal Router](https://github.com/kriasoft/universal-router) ([blog post](https://medium.com/@ippei.tanaka/universal-router-history-react-97ec79464573))

kravets-levko:

- [x] Refactor

## Description

- [x] Replace `angular-router` with `universal-router` + `history`
  - [x] wrapper component for `universal-router` + `history`;
  - [x] handle `route.resolve` map in the same manner as `angular-router` does;
  - [x] handle `<a>` tags globally similarly to what `angular` does.
- [x] Convert `<app-view>` component to React
  - [x] convert component to React;
  - [x] update layout on route change;
  - [x] alternative to route change events _([history.block()](https://github.com/ReactTraining/history/blob/master/docs/Blocking.md))_;
  - [x] error handling;
  - [x] mount new component instead of `ng-app` directive.
- [x] Update routes definitions to work with new router.
- [x] Cleanup
  - [x] fix CSS (ones that are on custom tags of AngularJS components);
  - [x] replace `$route`/`$routeParams` services;
  - [x] replace `$location` service.
- [x] Testing
  - [x] multi-org mode
  - [x] check if jQuery is used somewhere on non-app pages (signup, login, etc.) and either (1) update code [VanillaJS](http://vanilla-js.com/) if it's easy or (2) restore jQuery (was removed in this PR) _(Upd.: there is a very small piece of jQuery code in SaaS repo; except of that it's safe to remove jQuery)_

React migration getredash/redash#3071, closes #4159.

No UI changes


arikfr:
>  handle route.resolve map in the same manner as angular-router does;

I'm not sure we should keep this pattern. It's probably better to handle the load lifecycle in the page component itself.

> handle <a> tags globally similarly to what angular does.

Also not sure this worth implementing. We will need to do some extra work if we don't, but will end up with a "cleaner" and more explicit solution.

_But_ we can start with both implemented (specially considering you already done the resolve) as a crutch to help with move the migration forward and once everything is working, make another PR that removes each of them and replaces it with better pattern.

kravets-levko:
@arikfr Implementing both features is not that hard (as you mentioned, I already implemented `route.resolve`; handling <a> tags globally is also another 15-20 LoC). So I'll start with that, and - as you mentioned - later we can discuss if we need that features and remove if necessary.

srs231198:
Does this issue still require any  assistance

client/app/services/url.js
```diff
@@ -0,0 +1,25 @@
+import { pick, extend } from "lodash";
+
+const link = document.createElement("a"); // the only way to get an instance of Location class
```
gabrieldutra: If that's what we have :man_shrugging:, at least only one `<a>` object goes to the DOM.

client/app/visualizations/EditVisualizationDialog.jsx
```diff
@@ -13,7 +13,6 @@ import Visualization from "@/services/visualization";
 import recordEvent from "@/services/recordEvent";
 
 // ANGULAR_REMOVE_ME Remove when all visualizations will be migrated to React
```
gabrieldutra: I think this comment should have been removed too, we can remove it during cleanup, but just to have in mind there's nothing else to be done.

client/app/components/admin/Layout.jsx
```diff
@@ -11,7 +11,7 @@ export default function Layout({ activeTab, children }) {
       <PageHeader title="Admin" />
 
       <div className="bg-white tiled">
-        <Tabs className="admin-page-layout-tabs" defaultActiveKey={activeTab} animated={false}>
+        <Tabs className="admin-page-layout-tabs" defaultActiveKey={activeTab} animated={false} tabBarGutter={0}>
```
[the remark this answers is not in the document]
kravets-levko: That's rather for consistency - we set that padding to zero everywhere, and when I noticed that here we still have that padding - I took the opportunity and updated it

client/app/visualizations/EditVisualizationDialog.jsx
kravets-levko: I wanted to remove it (and few other things) in this PR, but realized that it does not break the app, and decided to not add more changes to PR. I think it's better to do all the cleanup in own PR

client/app/services/url.js
kravets-levko: After your comment I realized that probably Wepack can polyfill Node's `url` module. But it will require more work - it does not "know" about `<base>` tag, so we'll need to get base href from the DOM, manually resolve relative URLs, etc. `<a>` tag already does all that things - you just assign _any_ URL to `a.href`, and browser properly resolves it according all its internal rules. It means that we will always have :100: correctly resolved URL

client/app/pages/home/Home.jsx
```diff
@@ -173,15 +173,12 @@ function Home() {
...
-    },
-  };
-}
-
-init.init = true;
+export default {
+  path: "/",
+  title: "Redash",
+  render: (routeParams, currentRoute, location) => (
+    <AuthenticatedPageWrapper key={location.path}>
```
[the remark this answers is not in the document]
kravets-levko: They all should have `key` attribute - I need to check it :thinking: In short - `key` allows React to know if component instance of the same type is still the same (and should be just updated), or new one (to destroy old and initialize new). In this case I set `key` to current URL path because we use the same component for different pages (e.g. there are four query lists (all/my/favorites/archive), two dashboard lists, two query editors (new/existing query), etc. All that components wasn't designed to handle updates - they do some stuff on construction and then live their life. Therefore if you try to navigate between, say, query lists (without `key` attribute) - React will decide that that's the same component and try to update it; component ignores that and shows previous data.

First I was thinking about setting that `key` in Router, but later realized that it may be useful (and even needed sometimes): for example, we have Query Snippets pages (list and edit) which use the same controller; that controller can handle all updates, therefore for all routes it should have the same key (or not have the key at all).

client/app/components/ApplicationArea/Router.jsx
```diff
@@ -0,0 +1,144 @@
...
+function resolveRouteDependencies(route) {
+  return Promise.all(
+    map(route.resolve, (value, key) => {
+      value = isFunction(value) ? value(route.routeParams, route, location) : value;
+      return Promise.resolve(value).then(result => [key, result]);
+    })
+  ).then(results => {
+    route.routeParams = extend(route.routeParams, fromPairs(results));
+    return route;
+  });
```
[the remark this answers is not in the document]
kravets-levko: Correct, will fix, thanks!

client/app/pages/home/Home.jsx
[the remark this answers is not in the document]
kravets-levko: Upd.: I forgot to move `key` attributes for some controllers when adding wrappers :man_facepalming:

kravets-levko:
Differences I noticed so far:

> Clicking in "Edit" in an Alert Page now reloads all the page and resources

It's easy fix: I just need to tweak `key` attribute (for explanation see my reply to your question about `key`s)

> Also regarding errors I'd be careful not to show the error messages on screen (avoid this), one example is the vis embed page when accessing a query you don't have access to (example - this one is happening because the page doesn't have an onError, but it's an example of an error that shouldn't be that descriptive).

I think I'll better update `ErrorMessage` component to show hard-coded messages for 4** statuses (move that logic from `PromiseRejectionError` so later we could remove it at all), and some generic message for other error types.

client/app/visualizations/EditVisualizationDialog.jsx
gabrieldutra: In this case I guess it's just the comment left :sweat_smile:, the "code part" that it was pointing to be removed seems to have been removed in this PR.

client/app/services/url.js
arikfr: > it does not "know" about <base> tag

It might not be relevant at the moment, but just wanted to note that we will probably stop relaying on `<base>` tag for the multi-org functionality.

kravets-levko: That will be cool :+1: But it's already implemented, and I think it's right to keep it working until something changes.

client/app/components/ApplicationArea/ApplicationHeader/index.jsx
```diff
@@ -9,22 +9,22 @@ import Menu from "antd/lib/menu";
...
 
-import "./AppHeader.less";
+import FavoritesDropdown from "./FavoritesDropdown";
+import "./index.less";
 
 function onSearch(q) {
-  $location.path("/queries").search({ q });
-  $route.reload();
+  // TODO: Does not work, needs to be fixed
+  navigateTo(`/queries?q=${encodeURIComponent(q)}`);
```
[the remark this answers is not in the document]
kravets-levko: No, it does not :smile: It should be relative url

client/app/services/url.js
arikfr: I agree, just something to keep in mind.

client/app/components/ApplicationArea/Router.jsx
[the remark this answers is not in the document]
kravets-levko: Done

client/app/components/ApplicationArea/ApplicationHeader/index.jsx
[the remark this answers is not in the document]
kravets-levko: Fixed :+1:

kravets-levko:
About error handling (and latest changes to it):

1. @gabrieldutra I passed `onError`prop to admin/Jobs page, but it does not really need it - it already handles errors on its own. For other pages I added `onError` prop and error handling (at least, where they load some data). I can remove it if it looks confusing.

2. @gabrieldutra to get rid of `PromiseRejectionError` we have to make axios return original error instead of `error.response` (also, I noticed that if turning off network - axios rejects requests with error without `response` field; not sure what Angular did in that case, but I think it always returned some error data - because we didn't handle it in `PromiseRejectionError`). Also, seems we really don't use that rejections anywhere except of global error handler, so we have a weird situation: first we unwrap axios error to get response object, and then wrap it everywhere with `PromiseRejectionError` to have error object back (and yes - I know that `PromiseRejectionError` has some logic, but it can be moved to `ErrorMessage` component). But I'm not going to change it in this PR - let's discuss it and maybe fix the things in following cleanup PR.

3. Route change errors are now displayed without header. It is a result of changing authorization handling mechanism. In old router, each route had `authorized` field and router requested auth on each page change. Pages that didn't need auth handled it themselves. Now we have two wrappers - for auth and signed out pages which handle auth in a different way (regular login process and API key). Authorized page wrapper also renders header if user is authorized (that's important). So now router does not know anything about authorization and does not render header - it cannot show header on route change errors. Also, we should keep in mind that route change error may occur by two reasons: 
    - route does not exist at all: that situation should not happen normally (I believe most cases should be typos in manually typed URL); in this case it's not that important - is header visible or not. Also, we could add a link to home screen to error message to allow user to escape;
    - route's resolve promises failed: currently not much pages use this mechanism and eventually it should be moved to pages components. When it will be done - such fails will be displayed as usual (with header if needed) because now they will be handled not by router, but by page wrappers, and authorized page wrapper will keep header visible in that case. Working example of such behavior are most of list pages (see screenshot). (P.S. Probably we should use Ant Icon for that error message because it does not require external resources :thinking:)

image

client/app/components/ApplicationArea/Router.jsx
```diff
@@ -0,0 +1,144 @@
...
+          .then(route => {
+            if (route) {
+              setCurrentRoute(route);
+            }
+          })
+          .catch(error => {
+            if (!isAbandoned && currentPathRef.current === pathname) {
+              if (error.status === 404) {
+                // just a rename, original message is "Route not found"
+                error = new Error("Page not found");
```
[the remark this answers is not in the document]
kravets-levko: @gabrieldutra I added an extended comment below :point_down:

client/app/pages/destinations/DestinationsList.jsx
```diff
@@ -60,12 +65,7 @@ class DestinationsList extends React.Component {
         Destination.query().then(destinations => this.setState({ destinations, loading: false }));
         return destination;
       })
-      .catch(error => {
-        if (!(error instanceof Error)) {
-          error = new Error(get(error, "data.message", "Failed saving."));
```
gabrieldutra: In this specific case this was intentional :grimacing:, the error message "Alert Destination with the name X already exists." comes from it

kravets-levko: Omg... Okay, I'll restore it, thank you!

gabrieldutra:
@kravets-levko, just remembered to check one thing again :sweat_smile:. #4549 started happening again since now we're back not reloading the url :grimacing:.

I don't see a big problem on reloading the page in that circumstance, it's not a "very executed" path in normal conditions, so it doesn't feel harm on the extra second(s) it'll take to reload. However, if you really want to keep it, we need to somehow tell the router the "path" has now changed, so it triggers a reload when you go back to `queries/new`.

kravets-levko:
@gabrieldutra Fixed :+1: I also refined code a bit, so please review it as well. Thank you!

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -412,10 +412,10 @@ DashboardPage.defaultProps = {
...
-  render: routeParams => (
-    <AuthenticatedPageWrapper key={`/dashboard/${routeParams.dashboardSlug}`}>
+  render: currentRoute => (
+    <AuthenticatedPageWrapper key={currentRoute.key}>
       <ErrorBoundaryContext.Consumer>
-        {({ handleError }) => <DashboardPage {...routeParams} onError={handleError} />}
+        {({ handleError }) => <DashboardPage {...currentRoute.routeParams} onError={handleError} />}
       </ErrorBoundaryContext.Consumer>
     </AuthenticatedPageWrapper>
   ),
```
gabrieldutra: Not sure if it's that worth it at this point, but it looks like pages follow the same pattern when being defined, if this is not intended to be changed (and every page should follow this notation), we perhaps could abstract it a bit more with a closure returning this function? (seems we can tweak it to receive only the page component and some extra routeParams):

```js
{
  path: "/dashboard/:dashboardSlug",
  render: authenticatedPage(DashboardPage, { extraData: "dashboard-page" }),
}
```

BTW, if I had to choose, I think it's better to pass `routeParams` as separate props (like you did to List Pages) and not directly to the props (I don't know, but it feels risky to me to leave props that open :grimacing:).

Well, lmk your opinion

kravets-levko: 1. Yes, indeed it makes sense. But let's do it in a follow-up PR - I don't want to add anything more to this one (except of bug fixes related to PR's subject);
2. With new router it's safe to spread route params - that object contains only values from URL (in this example - `dashboardSlug`) and resolved values (which will be removed soon), Router does not add anything to it (unlike Angular's router which was merging params from URL, resolved values and query string params).

client/app/pages/index.js
```diff
@@ -0,0 +1,53 @@
...
+  groupsMembersRoutes,
+  homeRoutes,
+  queriesListRoutes,
+  queryViewRoutes,
+  querySourceRoutes,
+  visualizationEmbedRoutes,
+  querySnippetsRoutes,
+  organizationSettingsRoutes,
+  usersListRoutes,
+  userProfileRoutes,
```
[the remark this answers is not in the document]
kravets-levko: Forgot to reply to this comment. Why I ended up using explicit imports: if you remember that issue with parameters classes (circular references) - it happened because we removed `require.resolve` for services and started importing them explicitly, and Webpack did a deeper optimization of that files which revealed that bug (the fact it worked before was just an accident). I googled about that a bit and found, that for dynamic and implicit imports Webpack creates so-called "context module", and for that module type some optimizations are unavailable because Webpack cannot guess if that module has any side effect. When using explicit imports, Weback can properly analyze code and perform better optimization of the bundle.

client/app/components/tags-control/TagsControl.jsx
```diff
@@ -24,7 +24,9 @@ export class TagsControl extends React.Component {
   };
 
   editTags = (tags, getAvailableTags) => {
-    EditTagsDialog.showModal({ tags, getAvailableTags }).result.then(this.props.onEdit);
+    EditTagsDialog.showModal({ tags, getAvailableTags })
+      .result.then(this.props.onEdit)
+      .catch(() => {}); // ignore dismiss
```
arikfr: Coudln't we handle this in `showModal` instead of having to add this everywhere?

client/app/pages/alert/Alert.jsx
```diff
@@ -229,47 +241,51 @@ class AlertPage extends React.Component {
...
+          {({ handleError }) => <AlertPage {...currentRoute.routeParams} mode={MODES.VIEW} onError={handleError} />}
+        </ErrorBoundaryContext.Consumer>
+      </AuthenticatedPageWrapper>
+    ),
+  },
+  {
+    path: "/alerts/:alertId([0-9]+)/edit",
+    title: "Alert",
+    render: currentRoute => (
+      <AuthenticatedPageWrapper key={currentRoute.key}>
```
arikfr: Why do we need to duplicate this piece of logic (AuthenticatedPage + ErrorBoundry) on every page?

client/app/pages/index.js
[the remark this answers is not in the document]
arikfr: :+1: good find.

Not something to spend on time now, but it's interesting to see how Next.js handles this. Because they have dynamic imports and produce optimized builds.

client/app/config/index.js
```diff
@@ -72,59 +60,11 @@ function registerExtensions() {
...
-    });
-  });
-}
-
 requireImages();
-registerServices();
-initAppView(ngModule);
-registerPages();
 registerExtensions();
 registerVisualizations();
```
arikfr: Probably worth switching to static imports like you did for the pages.

client/app/pages/alert/Alert.jsx
arikfr: Also, passing `handleError` to every page feels like too much. Can't we solve this with some global state/context?

client/app/components/ApplicationArea/AuthenticatedPageWrapper.jsx
```diff
@@ -0,0 +1,61 @@
...
+        }
+      });
+    return () => {
+      isCancelled = true;
+    };
+  }, []);
+
+  useEffect(() => {
+    if (bodyClass) {
+      document.body.classList.toggle(bodyClass, true);
```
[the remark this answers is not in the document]
kravets-levko: ATM - yes, because it stretches few wrapper components when that class is set. To avoid this, we have to revisit some layout styles which I didn't want to do in this PR.

client/app/components/tags-control/TagsControl.jsx
kravets-levko: Sadly, no :slightly_frowning_face:  After `.catch()` handler promise will go to next `.then()` handler, so if we catch that promises inside of dialog wrapper - the caller's `.catch()` will not be called. It actually is not a problem, but browser will show unhandled rejection errors in console. We can either:
1. don't catch that promises and ignore errors in console;
2. always add empty catch handler like I did here;
3. replace browser's Promise with something else that does not trigger unhandled rejection errors (e.g. "bluebird") - prior to this PR we used Angular's `$q` for this purpose;
4. change a way we handle dialog's close/dismiss events (for me current flow looks fine, but that annoying unhandled rejections...)

client/app/pages/alert/Alert.jsx
kravets-levko: As I mentioned in reply to @gabrieldutra, I'm going to revisit this and few other things in a follow-up PR; currently it's stable, and I prefer not to add any more changes to this PR.

client/app/config/index.js
kravets-levko: Indeed :slightly_smiling_face: But it's not related to a subject of this PR, so I'll do it separately

client/app/pages/alert/Alert.jsx
kravets-levko: > Also, passing handleError to every page feels like too much. Can't we solve this with some global state/context?

Yes, and that's what I plan to do. There is a context already though; it's easy to use it with function-based components and a bit trickier with class-based ones - I didn't want to deal with it now so just passed it as a prop everywhere.

client/app/components/tags-control/TagsControl.jsx
kravets-levko: I'll explore if we can suppress that error with handling `unhandledrejection` event if error comes from dialog wrapper

client/app/pages/index.js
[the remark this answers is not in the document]
kravets-levko: It does not mean that with dynamic imports bundle wasn't optimized. But, for example, with explicit imports Webpack can merge modules, mangle exports (because now it exactly knows where they're used), etc.

client/app/config/index.js
arikfr: https://github.com/getredash/redash/issues/4565

client/app/components/tags-control/TagsControl.jsx
arikfr: Ok, let's continue the conversation here: https://github.com/getredash/redash/issues/4566.

kravets-levko:
@arikfr Seems it was broken during Query pages migration (I reproduced it on `master`). I'll open separate PR for it as well, let's not block this one with this bug.

arikfr:
Here's another one:

1. Open homepage.
2. Click on the "Now Param" dashboard.

You get every parameter update added to the history so going back feels like it's not working (you need to click lots of time until you get back to the homepage).

arikfr:
I triggered another build.

kravets-levko:
@arikfr That test was still not stable even after my previous fix; new fix worked fine on my PC, need to try on CI now.

A short explanation about history bloating bug fix: the actual issue is in `useDashboard` hook and it's not fixed here, because it's not related to a router itself and also requires it's own proper fix and discussion (PR is coming soon). In short: every time dashboard it updated, `useDashboard` decides to update URL as well; when refreshing dashboard - each widget updates dashboard twice - therefore so much history entries. I'm not sure how it worked with `$location` service (probably it updates URL only if something really changes). But since that updates are not necessary - they should not happen at all.

Also, while reproducing this bug on `master`, I noticed that when changing param values manually (or update other query string params), and then hitting Back button in browser - URL gets changed, but page does not update. It happens on dashboard page, all list pages and on query pages. So I decided to just replace URL (not push new history entry) - it (a sort of) fixed both bugs: history is not flooded with entries that pages cannot handle anyway.

arikfr:
:raised_hands:

jezdez:
Congrats!!

Issue #4526, opened 2020-01-07
TypeError: ssl_wrap_socket_with_ocsp() got an unexpected keyword argument 'key_password'


maximecharron:
Using the preview docker image and deploying with the setup script, I'm hitting the following error when trying to execute a query against a Prometheus Data Source: 

```
[2020-01-07 18:30:31,789][PID:26][WARNING][root] Unexpected error while running query:
Traceback (most recent call last):
  File "/app/redash/tasks/queries/execution.py", line 160, in run
    data, error = query_runner.run_query(annotated_query, self.user)
  File "/app/redash/query_runner/prometheus.py", line 139, in run_query
    response = requests.get(api_endpoint, params=payload)
  File "/usr/local/lib/python3.7/site-packages/requests/api.py", line 75, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/requests/api.py", line 60, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python3.7/site-packages/requests/sessions.py", line 646, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/requests/adapters.py", line 449, in send
    timeout=timeout
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 672, in urlopen
    chunked=chunked,
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 376, in _make_request
    self._validate_conn(conn)
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 994, in _validate_conn
    conn.connect()
  File "/usr/local/lib/python3.7/site-packages/urllib3/connection.py", line 394, in connect
    ssl_context=context,
TypeError: ssl_wrap_socket_with_ocsp() got an unexpected keyword argument 'key_password'
```

Downgrading `requests` to 2.21.0 fixes the issue.

stevebuckingham:
I've fixed this my incrementing the snowflake connector and the pyopenssl library. I did this as part of extending Redshift functionality (as boto hit the SSL error) and this is in my open PR. https://github.com/getredash/redash/pull/4532

phdowling:
Seeing this for Kibana as well

phdowling:
[...]

PR #4532, opened 2020-01-09
Add redshift role use option

stevebuckingham:

- [x] Refactor
- [x] Feature

## Description

This is to add the ability to use IAM Roles and Users for Redshift access. It supports specifying a role and/or keys or using a role or keys already attached to the instance.


stevebuckingham:
Looking at this I think the failing test has nothing to do with the code in the commit - I assume there is another failure on master. Is that likely?

arikfr:
Yes, this failure is not related to your code. Sorry about this.

redash/query_runner/pg.py
```diff
@@ -284,6 +349,27 @@ def configuration_schema(cls):
             "required": ["dbname", "user", "password", "host", "port"],
             "secret": ["password"],
         }
+        if redshift_role_enabled:
```
arikfr: Having all the options under one data source will make the configuration a bit confusing. Also it makes it hard to enforce required fields.

How about we create a subclass for the IAM based authentication?

stevebuckingham: Sure - that's a sensible option. So we end up with something like `Redshift - User Login` and `Redshift - IAM Login`.

arikfr: I would go with `Redshift` and `Redshift (IAM Login)`.

stevebuckingham:
I'm now making the logo work for the new version. Then it should all work.

stevebuckingham:
This should now work. I've tested it using AWS Keys. I'll make sure with roles as well.

stevebuckingham:
I've tested this using roles as well now.

requirements.txt
```diff
@@ -60,6 +60,7 @@ disposable-email-domains>=0.0.52
 gevent==1.4.0
 supervisor==4.1.0
 supervisor_checks==0.8.1
+urllib3==1.25.7
```
arikfr: Are you sure we need this as an explicit dependency?

redash/query_runner/pg.py
```diff
@@ -331,6 +345,131 @@ def _get_tables(self, schema):
...
 
+class RedshiftIAM(Redshift):
+
+    @classmethod
+    def type(cls):
+        return "redshift_iam"
+
+    @classmethod
+    def name(cls):
+        return "Redshift IAM User/Role"
```
arikfr:
```suggestion
        return "Redshift (with IAM User/Role)"
```

redash/query_runner/pg.py
```diff
@@ -227,11 +235,17 @@ def run_query(self, query, user):
...
 
 class Redshift(PostgreSQL):
+
     @classmethod
     def type(cls):
         return "redshift"
 
+    @classmethod
+    def name(cls):
+        return "Redshift User"
```
arikfr:
```suggestion
        return "Redshift"
```

requirements.txt
stevebuckingham: Hmm - I think it addresses a bug raised with in a couple of places. I'll try it with just specification of the `pyopenssl` version.

stevebuckingham: OK tested it and specify of pyopenssl is enough so I have requirement requirement for specified urllib.

arikfr: If it's because of #4526 then #4567 should be enough. But I'll check release notes for the other libs to see if there is anything risky there.

arikfr: The `cryptography` upgrade looks safe, but `pyopenssl` has some deprecations which I'm not sure if they are supported by all the other libs we depend on.

stevebuckingham: I'll try rerunning with just `cryptography` tomorrow morning - Berlin time.

arikfr: Thanks! Appreciate the effort.

stevebuckingham: It worked with `cryptography` 2.8 and `pyopenssl` 19.0.0 so I've made that commit.

arikfr: Great. I fixed a merge conflict and once the build finishes will merge :+1:

arikfr:
Thanks!

PR #4533, opened 2020-01-10
Get rid of six and fix str/unicode types regression that became active on Python 3.

jezdez:
- [x] Bug Fix

## Description

Get rid of six and fix str/unicode types regression that became active on Python 3.

This was introduced in d38ca803c5dbf08d33f3c73327196db417b125a7.

Refs https://github.com/getredash/redash/pull/4055#pullrequestreview-341176574

### DIFF
--- a/redash/cli/data_sources.py
+++ b/redash/cli/data_sources.py
@@ -2,7 +2,6 @@ from sys import exit
 
 import click
 from flask.cli import AppGroup
-from six import text_type
 from sqlalchemy.orm.exc import NoResultFound
 
 from redash import models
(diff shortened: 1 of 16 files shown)


jezdez:
I'm working on a rebase with current Redash master and this just came up since one of our extensions wasn't working right and I stumbled over a similar code pieces. From there I wondered if that would be also wrong in Redash upstream.

I don't think we should consider this just to keep things clean, it's actually a big risk for query runners that may return bytes or unicode data since bytes aren't handled at all right now, e.g. https://github.com/getredash/redash/pull/4533/files#diff-cc1177c16e24f67a5b4d1b3ad412d773L21-R20. But the removal of six is certainly a cleanup.

arikfr:
:+1:

PR #4536, opened 2020-01-11
Query View redesign

gabrieldutra:

- [x] Refactor
- [x] Feature

## Description

To-do:
- [x] Initial preview for Desktop (mostly to see how data will fit)
- [ ] Add details
  - [x] Parameters
  - [x] EmptyState
  - [x] Better placement for QueryExecutionStatus (?)
  - [ ] ...
- [x] Mobile version 
- [ ] Organize styling (less variables, etc)

[Figma](https://www.figma.com/file/DS8f6bZUmwEoqKgFvJCkpg/Redash?node-id=113%3A195)

Not yet  :)


client/app/pages/queries/QueryView.jsx
```diff
@@ -125,51 +142,47 @@ function QueryView(props) {
                 onChangeTab={setSelectedVisualization}
                 onAddVisualization={addVisualization}
                 onDeleteVisualization={deleteVisualization}
+                cardStyle
```
arikfr: Let's not make this configurable but use this on both pages, to make sure it makes sense in both pages (as it has to be the same).

gabrieldutra:
Regarding the fullscreen toggle:
1. `fa-arrows-alt` looked weird imo (it's too 'bold'), so I went with `zmdi-fullscreen` (the same we have in the Dashboard page).
2. I added an Alt+F shortcut (lmk if you have another idea and thoughts regarding MacOS shortcut), it makes it handy to operate with the page, it should make the experience better for more experienced users.
3. I also tried to add a css animation for the height change, but it's not a simple thing because in this case the height is automatic. I can explore JS animation options if you think it's worth it.

I inverted the "Add tag" with "Add description" (considering Figma style) just because it was easier, LMK if it actually should be exactly as in Firgma (it seemed an ok change to me).

gabrieldutra:
I'm currently using a `query-fixed-layout` class to make the layout fixed or not (for mobiles I just make it non-fixed). Since it's triggered inside JS, I can make pretty much anything enable/disable it (in case we want to make a different trigger based on height, or the user's input).

In my 14'' laptop (1366x768) https://deploy-preview-4536--redash-preview.netlify.com/queries/307 doesn't look that good (only one row appearing at a time, but the Toogle Fullscreen with a shortcut makes it work once you get used to it.

arikfr:
I rebased this with master and force pushed.

arikfr:
Screen Shot 2020-02-10 at 10 51 59 AM

The visualization tabs need some padding from the border.

arikfr:
Also, need to update the header on mobile to be more like the Figma design. I would skip adding "add tag"/"add description" buttons on mobile.

gabrieldutra:
I think I'll just handle those comments along with #4584. It doesn't make that much sense to be in there, but it will spare the rebasing.

PR #4538, opened 2020-01-12
Google Analytics runner - iterate over keys the Python 3 way

rauchy:
- [x] Bug Fix

## Description
Following #4181, Google Analytics queries were broken since dict `.keys()` returns `dict_keys`, which would in turn yield wrong results when iterating over param values, resulting in errors such as
> Parameter "end_date" value "y,e,s,t,e,r,d,a,y" does not match the pattern "[0-9]{4}-[0-9]{2}-[0-9]{2}|today|yesterday|[0-9]+(daysAgo)" 

#4181

### DIFF
--- a/redash/query_runner/google_analytics.py
+++ b/redash/query_runner/google_analytics.py
@@ -163,11 +163,8 @@ class GoogleAnalytics(BaseSQLQueryRunner):
         try:
             params = json_loads(query)
         except:
-            params = parse_qs(urlparse(query).query, keep_blank_values=True)
-            for key in [*params]:
-                params[key] = ",".join(params[key])
-                if "-" in key:
-                    params[key.replace("-", "_")] = params.pop(key)
+            query_string = parse_qs(urlparse(query).query, keep_blank_values=True) 
+            params = {k.replace('-', '_'): ",".join(v) for k,v in query_string.items()}
 
         if "mcf:" in params["metrics"] and "ga:" in params["metrics"]:
             raise Exception("Can't mix mcf: and ga: metrics.")
--- a/redash/query_runner/google_analytics.py
+++ b/redash/query_runner/google_analytics.py
@@ -164,7 +164,7 @@ class GoogleAnalytics(BaseSQLQueryRunner):
             params = json_loads(query)
         except:
             params = parse_qs(urlparse(query).query, keep_blank_values=True)
-            for key in params.keys():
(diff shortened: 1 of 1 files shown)


redash/query_runner/google_analytics.py
```diff
@@ -164,7 +164,7 @@ def run_query(self, query, user):
             params = json_loads(query)
         except:
             params = parse_qs(urlparse(query).query, keep_blank_values=True)
-            for key in params.keys():
+            for key in [*params]:
```
arikfr: I'm not sure this is the issue:

```
>>> d = {'ab': 'yesterday', 'cd': 2}
>>> for k in d.keys():
...     print(k)
... 
ab
cd
```

(this shows that the keys are iterated properly)

*But*, check the following:

```
>>> ",".join("yesterday")
'y,e,s,t,e,r,d,a,y'
```

So the issue is actually line 168 which applies `",".join` on any value.

rauchy: I agree, but subsequently, using `[*params]` would resolve this.

rauchy: The real problem is line 170, which messes around with the `dict_keys` iterator, something that didn't happen in Python 2. Populating a new list of keys with `[*params]` solves this.

arikfr:
```suggestion
            # Because we change keys list on line 170, we need to create a copy of the keys
            for key in [*params]:
```

redash/query_runner/google_analytics.py
```diff
@@ -164,7 +164,7 @@ def run_query(self, query, user):
...
                 params[key] = ",".join(params[key])
```
arikfr:
```suggestion
                # parse_qs returns every value as a list
                params[key] = ",".join(params[key])
```

redash/query_runner/google_analytics.py
rauchy: Another option that would make fiddling with iterators (and the need to explain it) redundant is:

```python
 query_string = parse_qs(urlparse(query).query, keep_blank_values=True) 
 params = {k.replace('-', '_'): ",".join(v) for k,v in query_string.items() }
```
We'll still have to explain `",".join`, but maybe we don't have to.

arikfr: I like this :+1:

PR #4544, opened 2020-01-14
 Refactor docker-compose.yml for development

ariarijp:

- [x] Refactor

## Description

Simplify docker-compose.yml by [Extension fields](https://docs.docker.com/compose/compose-file/#extension-fields).


docker-compose.yml
```diff
@@ -1,55 +1,49 @@
 # This configuration file is for the **development** setup.
-version: '3.2'
 # For a production example please refer to getredash/setup repository on GitHub.
+version: '3.4'
```
arikfr: Is the version upgrade necessary? In the past we had issues with newer versions of compose configuration and usually there was little benefit in them.

ariarijp: @arikfr Extention fields added in v3.4 format in my understanding.

I'll check it again.

> In the past we had issues with newer versions of compose configuration and usually there was little benefit in them.
Could I have that issue URL?
I'd like to check that.

ariarijp: @arikfr I could not use Extention Fields on v3.2 format of docker-compose file.

However, we can use that if we can go back to v2 format like [docker-compose.yml in setup repository](https://github.com/getredash/setup/blob/master/data/docker-compose.yml#L1). 
Here is output of `git diff` between v3.4 style and v2 style.

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index bfeb38b7..904f89bb 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,12 +1,10 @@
 # This configuration file is for the **development** setup.
 # For a production example please refer to getredash/setup repository on GitHub.
-version: '3.4'
+version: "2"
 x-redash-service: &redash-service
   build: .
   volumes:
-    - type: bind
-      source: .
-      target: /app
+    - ".:/app"
 x-redash-environment: &redash-environment
   REDASH_REDIS_URL: "redis://redis:6379/0"
   REDASH_MAIL_DEFAULT_SENDER: "redash@example.com"
```

IMHO, I think we can stay in the old version, but we should upgrade to 3.4 or higher for the future development if we could not get big benefit.

arikfr: If we don't need any feature from 3.4, let's use 2.

ariarijp: @arikfr 
Sure, I changed this PR can be run on v2 configulation.
Could you review it again?

Thanks, and stay healthy.

arikfr:
Thanks!

PR #4545, opened 2020-01-14
Fix Cypress issues after React version of Query Pages

gabrieldutra:

- [x] Other

## Description
After the React version of the query pages, a couple of specs started to show some flakyness. I decided to investigate, since it could be related to a real issue, so this PR has two fixes:

- **Pivot Spec**: for some reason the `.pvtGrandTotal` was pointing a different value in Cypress, and in the screenshot it looked ok... I assumed it was just a different way of rendering that caused it, so I changed the assertion method;
- **Parmeter Spec**: this one is a real issue and deserves a small discussion: previously to the React Migration of the query pages we used to allow the user to Re-execute a query while it's running when parameters are changed. This can cause overflow on queries execution since the user can trigger a query multiple times. On the other hand, it can also be annoying when you trigger a query with wrong parameters and has to wait for it to finish.

--

--

### DIFF
--- a/client/app/pages/queries/QuerySource.jsx
+++ b/client/app/pages/queries/QuerySource.jsx
@@ -153,7 +153,7 @@ function QuerySource(props) {
 
   const doExecuteQuery = useCallback(
     (skipParametersDirtyFlag = false) => {
-      if (!queryFlags.canExecute || (!skipParametersDirtyFlag && areParametersDirty) || isQueryExecuting) {
+      if (!queryFlags.canExecute || (!skipParametersDirtyFlag && (areParametersDirty || isQueryExecuting))) {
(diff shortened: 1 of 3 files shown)


arikfr:
> * his one is a real issue and deserves a small discussion: previously to the React Migration of the query pages we used to allow the user to Re-execute a query while it's running when parameters are changed.

this is no longer possible?

gabrieldutra:
> this is no longer possible?

I forgot to mention what I did :sweat_smile:. This PR makes it possible as it was before the migration (I preferred to keep the behavior users have). Basically: if a query is running, you can't trigger it again with the Execute button, but you can with the Apply Changes button.

PR #4547, opened 2020-01-15
fix mysql compatible issue

exherb:

- [x] Bug Fix

## Description

some mysql compatible api always return upper case column name when query information_schema schema


arikfr:
> some MySQL compatible api

Can you give an example?

exherb:
> > some MySQL compatible api
> 
> Can you give an example?

A product named `AnalyticDB for MySQL`  from Aliyun (Chinese cloud provider). Here is the screenshot:

CleanShot 2020-01-16 at 14 04 54@2x

PR #4551, opened 2020-01-15
Add loading state to Query save button

### LINKED ISSUE #4548: Saving a new query creates multiple queries
### Issue Summary

In query editor, "Save" button is not locked when saving query, therefore it's possible to click it multiple times during sending previous save requests to backend. If frontend didn't receive a response from previous request(s), it will use the same query object to save. If that query object is new (does not have ID yet) - on each request backend will create a new record in database.

Suggested solution: disable "Save" button until save request completed; show some animation (e.g. "Saving..." with animated dots - like on Dashboard page).

### Steps to Reproduce

1. Create a new query, type something in editor;
2. Click "Save" button quickly few times;
3. Go to queries list and see few new queries, completely equal to created one.

### Technical details:

* Redash Version: master
* Browser/OS: any

gabrieldutra:

- [x] Bug Fix
- [x] Other

## Description
This adds a loading state to the Query Save button, the main reason is to avoid multiple query created when the connection is not that good.

Fixes #4548 

fix-multiple-query-saves

### DIFF
--- a/client/app/components/queries/QueryEditor/QueryEditorControls.jsx
+++ b/client/app/components/queries/QueryEditor/QueryEditorControls.jsx
@@ -106,7 +106,7 @@ export default function EditorControl({
             loading={saveButtonProps.loading}
             onClick={saveButtonProps.onClick}
             data-test="SaveButton">
-            <span className="fa fa-floppy-o" />
+            {!saveButtonProps.loading && <span className="fa fa-floppy-o" />}
             {saveButtonProps.text}
           </Button>
         </ButtonTooltip>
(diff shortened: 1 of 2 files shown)


kravets-levko:
A very first impression: two icons (animated spinner and save) look weird together. Probably makes sense to hide save icon when save in progress. Also, loading state does not clearly indicate action being performing - IMO it's better to change button text as well (to "Saving..."). Also, a dirty state indication (`*`) also should be hidden while saving.

arikfr:
+1 to everything @kravets-levko said, except that I would keep the text as “Save”. But yes, let’s hide the icon and the dirty (*) indicator.

PR #4553, opened 2020-01-16
Add padding to QueryExecutionStatus text

### LINKED ISSUE #4550: Query execution status “flickers”
The new query execution status (query in queue -> executing -> loading results) seems like it’s flickers (removed from view and then rendered again) on status changes. Also it seems like its height is being changed.

Need to make sure that as long as the query is executing, we render the same element and only change its text.

gabrieldutra:

- [x] Bug Fix

## Description
Add padding to QueryExecutionStatus text

Fixes #4550 

status-changes-height-fix

### DIFF
--- a/client/app/pages/queries/components/QueryExecutionStatus.jsx
+++ b/client/app/pages/queries/components/QueryExecutionStatus.jsx
@@ -40,7 +40,7 @@ export default function QueryExecutionStatus({ status, updatedAt, error, isCance
       type={alertType}
       message={
         <div className="d-flex align-items-center">
-          <div className="flex-fill">
+          <div className="flex-fill p-t-5 p-b-5">
             {message} {showTimer && <Timer from={updatedAt} />}
           </div>
           <div>


kravets-levko:
@gabrieldutra Maybe, just hide button with `visibility: hidden` instead of conditionally rendering it? Then it will be invisible, but browser will preserve space for it. Or render a placeholder <span> element of the same height as button has? (that's just ideas, I didn't try your solution yet :grimacing:)

gabrieldutra:
> Maybe, just hide button with `visibility: hidden` instead of conditionally rendering it? Then it will be invisible, but browser will preserve space for it. Or render a placeholder element of the same height as button has? (that's just ideas, I didn't try your solution yet )

`visibility: hidden` was one option I considered, one problem is that it takes width space too (for errors, or smaller screens it may be relevant). My conclusion was that the simplest was to either add some padding to the text to make sure it had more height than the button or add a `min-height` prop. The padding felt like a quick-win, so I went with it.

Issue #4557, opened 2020-01-17
Filtering by tags in query list is not working


nason:
### Issue Summary

A summary of the issue and the browser/OS environment in which it occurs.

### Steps to Reproduce

1. Go to /queries
2. Select a tag in the sidebar

I expect the queries list to filter by the selected tags, but this does not appear to be happening.

### Technical details:

* Redash Version: 9.0.0-alpha (b5bfad0547ba04ed1ba0ad7ea17631e0e2b9846)
* Browser/OS: Chrome / MacOS
* How did you install Redash: Docker

arikfr:
@kravets-levko @gabrieldutra at first I thought it's a backend issue, but this doesn't happen in the SaaS repo. Can it be related to one of he latest React changes? Maybe the Axios one?

kravets-levko:
@arikfr Yes, seems it's related to Axios: it serializes arrays in a different way comparing to Angular:
- Angular (works fine): `/api/queries?tags=Tag1&tags=Tag2` 
- Axios (doesn't work): `/api/queries?tags[]=Tag1&tags[]=Tag2`

arikfr:
@kravets-levko is this something we can easily control? Otherwise we can update the backend to support it.

kravets-levko:
Seems yes. I'll explore it a bit later (need to finish some work I already started)

gabrieldutra:
I checked in here, it seems we can change the serializer axios uses (we can pass it as a default config in the instance defined in `axios.js`):

```js
{
  paramsSerializer: params => qs.stringify(params),
}
```

For qs I tested the `query-string` lib that's already being used in the Router PR and looks like it translated the arrays without the brackets by default :slightly_smiling_face:.

LMK if you want me to push those changed or in case you already had another idea.

nason:
Thanks for the quick fix! Great work y'all

kravets-levko:
Quick bug - quick fix :sunglasses:

ptsilva:
I think that this fix causes a bug in multi tag filter. I tried here and it's easy to reproduce.

gabrieldutra:
> I think that this fix causes a bug in multi tag filter. I tried here and it's easy to reproduce.

Can you provide more details? (e.g: in which url you tried it and what exactly happened)

It looks like it's working in our [preview](https://redash-preview.netlify.com/) environment

ptsilva:
[...]

PR #4563, opened 2020-01-20
Tags filter doesn't work because of wrong query params format

kravets-levko:

- [x] Bug Fix

Fixes getredash/redash#4557


arikfr:
:+1:

PR #4569, opened 2020-01-20
Save new query before adding visualization

kravets-levko:

- [x] Bug Fix

## Description

Restore feature accidentally removed during React migration: when trying to add a first visualization to a new query, save the query first (otherwise visualization couldn't be saved).

### DIFF
--- a/client/app/pages/queries/hooks/useAddVisualizationDialog.js
+++ b/client/app/pages/queries/hooks/useAddVisualizationDialog.js
@@ -1,0 +1,25 @@
+import { useState, useCallback, useEffect } from "react";
+import useQueryFlags from "./useQueryFlags";
+import useEditVisualizationDialog from "./useEditVisualizationDialog";
+
+export default function useAddVisualizationDialog(query, queryResult, saveQuery, onChange) {
+  const queryFlags = useQueryFlags(query);
+  const editVisualization = useEditVisualizationDialog(query, queryResult, onChange);
+  const [shouldOpenDialog, setShouldOpenDialog] = useState(false);
+
+  useEffect(() => {
+    if (!queryFlags.isNew && shouldOpenDialog) {
+      setShouldOpenDialog(false);
+      editVisualization();
+    }
+  }, [queryFlags.isNew, shouldOpenDialog, editVisualization]);
+
+  return useCallback(() => {
+    if (queryFlags.isNew) {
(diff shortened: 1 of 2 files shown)


client/app/pages/queries/hooks/useAddVisualizationDialog.js
```diff
@@ -0,0 +1,25 @@
+import { useState, useCallback, useEffect } from "react";
+import useQueryFlags from "./useQueryFlags";
+import useEditVisualizationDialog from "./useEditVisualizationDialog";
+
+export default function useAddVisualizationDialog(query, queryResult, saveQuery, onChange) {
```
arikfr: Why the extensive use of hooks here, instead of something that looks more like:

```javascript
const addViz = () => {
  if (query.isNew) {
    saveQuery().then(() => openNewVizDialog());
  } else {
    openNewVizDialog();
  }
};
```

(obviously this more pseudo code than actual implementation, but I'm referring to the concept)

kravets-levko: Because of the major difference between function and class-based components: functions capture values.

With the simple workflow as in your pseudo code: `saveQuery` saves the query, calls `setQuery` and updates query object; then component is re-rendered and creates a new scope with a new query object; but `openNewVizDialog()` still uses captured value from the previous render - without query ID - and will fail to save visualization as well. Even more: when saving visualization it will return that old query instance to component and will overwrite previously saved query.

This flow ensures that query is saved before opening dialog and dialog is using latest query instance.

arikfr: It's pseudo code and wasn't meant to cover every case, but just to show the flow. In real implementation you could pass the new query object returned from `saveQuery` to `openNewVizDialog` to make sure they both use the same one?

kravets-levko: I was thinking about this. But query pages were implemented in a way when they have a single source of truth - `useQuery` hook which manages query state. And for me it looks right to use it everywhere just to be sure that I always have the "right" query object.

arikfr: I see. From a first look, it looks like we're over using hooks. But I will review it with full context and share comments if I have separately.

PR #4572, opened 2020-01-21
React migration cleanup

kravets-levko:

- [x] Refactor
- [x] Other

## Description

- [x] Revisit ANGULAR_REMOVE_ME things
- [x] Remove unused CSS (probably not 100% cleanup, but at least remove ones that are not used for sure)
- [x] Revisit error handling flow (how HTTP and promise rejection errors are handled)


client/app/assets/less/inc/form.less
```diff
@@ -29,285 +29,266 @@ textarea.v-resizable {
...
-  transition: border-color ease-in-out 0.15s, box-shadow ease-in-out 0.15s;
-  width: 90%;
-}
-
-
 /* --------------------------------------------------------
     Input Fields
 -----------------------------------------------------------*/
 .form-control {
-    .transition(all);
```
kravets-levko: This is the actual change, the rest is from Prettier

client/app/assets/less/inc/form.less
```diff
@@ -29,285 +29,266 @@ textarea.v-resizable {
   }
 }
 
-/* light version of bootstrap's form-control */
-.rd-form-control {
```
kravets-levko: This is the actual change, the rest is from Prettier

client/app/components/ApplicationArea/ErrorMessage.jsx
```diff
@@ -1,17 +1,43 @@
...
+    // Router errors
+    if (error.status) {
+      return getErrorMessageByStatus(error.status, message);
+    }
+  }
+  return message;
+}
+
+export default function ErrorMessage({ error }) {
+  const message = useMemo(() => getErrorMessage(error), [error]);
```
arikfr: Why bother with `useMemo` here? `getErrorMessage` doesn't feel that intensive.

kravets-levko: Fixed :+1:

kravets-levko:
@arikfr While working on this PR I briefly checked the usage of Bootstrap components - looks that there's really no much left as well. But I decided to not touch it for now.

arikfr:
@kravets-levko good call. It's probably better to do in a separate PR anyway. I have a feeling we mostly use different helper classes (`m-l-5` and friends), grid and some "component" classes (like panes and such), which probably have an alternative in Ant.

kravets-levko:
We mostly use helper classes (which are actually not a part of Bootstrap  - mostly), and somewhere - grid. I didn't see any Boostrap components usages (like alerts/panels/etc.) - probably I missed something because that wasn't my target at that time (I was reviewing a lot of files for other purpose and checked this just by the way), but there are really no much left. And a lot of icons of course :slightly_smiling_face:

arikfr:
I wonder if there is some utility we can use to analyze our code and tell us which styles are actually in use?

PR #4579, opened 2020-01-22
Refine routes definitions

kravets-levko:

- [x] Refactor
- [x] Other

# Description

Move repeatable code to wrappers.


client/app/components/ApplicationArea/withUserSession.jsx
```diff
@@ -0,0 +1,67 @@
...
+    if (!isAuthenticated) {
+      return null;
+    }
+
+    return (
+      <>
+        <ApplicationHeader />
+        <ErrorBoundary renderError={error => <ErrorMessage error={error} />}>
+          <ErrorBoundaryContext.Consumer>
+            {({ handleError }) => <WrappedComponent onError={handleError} {...props} />}
```
arikfr: This is definitely better, but I wonder: why a HOC instead of just a regular component that will receive the rendered page as children? Is it to support the ability to pass in the `handleError` function or there is some other reason?

client/app/pages/admin/Jobs.jsx
```diff
@@ -121,14 +120,10 @@ class Jobs extends React.Component {
   }
 }
 
+const JobsPage = withUserSession(Jobs);
```
arikfr: A follow up to previous question is: why wrap it here instead of in the route definition?

client/app/components/ApplicationArea/withUserSession.jsx
arikfr: I just realized another thing: this way the header will get re-rendered every time, including running its load (load favorites) every time.

Wouldn’t it be better to render a single header and only replace the page itself? I think we can achieve this if we have a single wrapper rendered by the Router and then only replace the page.

Not sure about the downsides though.

kravets-levko: I'll try to answer for both questions in this comment.

HoC is a compromise between flexibility and amount of code when defining a route. Of course we can make router accept a `component: SomePage` field instead of `render: () => { ... }`, but having a render function allows to more precisely manage how React will update page controller (using `key` props where needed), the way we pass props to page controller (in render function we can spread route params, or pass entire object, add or omit some params, etc.), do conditional rendering if needed, etc. That features are already used, say, for Alert page - where we have a single component for new/view/edit pages and just pass a different `mode` param - to achieve the same with passing just a component and rendering it in Router we have to keep `resolve` thing (that adds new values to route params) - but I want to get rid of it as well.

Another reason is what you mentioned - with HoC we can pass error handler to wrapper components. Also, I'm not sure if it's a good idea to combine it with auth stuff, maybe `withErrorHander` HoC will be better? Anyway, ATM I don't want to add more wrappers, that two auth wrappers are required for all pages, so I think we can keep error handlers there.

Also, HoC is a good choice because we already wrap _most_ of pages with other HoCs, so adding one more wrapper looks not that bad :slightly_smiling_face:

kravets-levko: > I just realized another thing: this way the header will get re-rendered every time, including running its load (load favorites) every time.

I'll check this, but I think it can be fixed much easier - just by setting `key` prop to header (but I need to try it to be sure :slightly_smiling_face:)

gabrieldutra: I was actually thinking about it returning the render function instead of a HoC, but it ends up with not that much of a difference. How about moving this `key` to it too (most or all pages seem to use it), in case you need it different, it's just matter of making it replaceable.

kravets-levko: @gabrieldutra That may be a good idea, but I'll explore it tomorrow (for me it's quite late time already)

kravets-levko: Actually it may be a really cool idea, especially in combination with what @arikfr suggested :thinking: I'll add one more commit tomorrow - will see how it goes

kravets-levko:
@arikfr (cc @gabrieldutra) I updated implementation based on our discussion - is it something you had in mind?

client/app/components/ErrorBoundary.jsx
```diff
@@ -8,7 +8,10 @@ const logger = debug("redash:errors");
 
 export const ErrorBoundaryContext = React.createContext({
   handleError: error => {
-    throw error;
+    // Allow calling chain to roll up, and then throw the error in global context
```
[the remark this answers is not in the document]
kravets-levko: I found this issue when worked on ItemsList component. In short - if you throw (in this case - re-throw) an exception during some component's lifecycle events - it may completely crash app (sadly, I cannot remember details now). Using `setTimeout` allows component/React to finish whatever it needs, and when JS engine goes idle - we re-throw the error. It's completely safe because Error constructor captures all stack and when we re-throw error - we'll see correct debug info.

kravets-levko:
`currentPage` is needed for some pages to show some state in UI (for example, highlight menu item on query/dashboard lists, etc.). Previously ItemsList was getting that value from `route.resolve`, but using component's prop looks more clear.

client/app/pages/data-sources/DataSourcesList.jsx
```diff
@@ -156,28 +155,14 @@ const DataSourcesListPage = wrapSettingsTab(
...
-    render: currentRoute => (
-      <AuthenticatedPageWrapper key={currentRoute.key}>
-        <ErrorBoundaryContext.Consumer>
-          {({ handleError }) => <DataSourcesListPage {...currentRoute.routeParams} onError={handleError} />}
-        </ErrorBoundaryContext.Consumer>
-      </AuthenticatedPageWrapper>
-    ),
-  },
-  {
+    render: pageProps => <DataSourcesListPage {...pageProps} />,
```
gabrieldutra: Looks cleaner now, one last thing that went through my mind: how about allow both a Component and a Function in the `render`? This way you can make `render: pageProps => <Component {...pageProps} />` be the default behavior and still can use a function when you need sth different.

kravets-levko: I don't want to make things too complicated. When having both `render` and `component` we should now define and document which takes precedence when both are specified. Considering that we don't define routes that often and that `render: pageProps => <Component {...pageProps} />` looks clean enough - I don't think we need one more abstraction (feature?) there.

gabrieldutra: I actually meant to just allow both options (a function or a component) to passed under the `render` prop. But it falls under the "one more abstraction" anyway.

kravets-levko: Ah, I see your idea now. Well, while it's possible, I think it's better if we'll define all routes in a single manner - as a function, to not confuse future us :grin:

kravets-levko:
@arikfr @gabrieldutra do you have more question or it's good to go?

arikfr:
Only comment I might have is that we should consider dropping `currentPage` prop in favor of parsing the URL, _but_ it's not important enough. Merging.

PR #4582, opened 2020-01-23
Fix treasuredata endpoint

### LINKED ISSUE #3185: TreasureData getSchema fails when setting non-default region
### Issue Summary

There are some regions in Treasure Data, but getSchema alsways fails when setting non-default region.

### Steps to Reproduce

1. Set datasource using non-default region (e.g. Tokyo region)
2. Push schema refresh then "Schema refresh failed" error occurs

### Technical details:

* Redash Version: confirmed v5.0.2
* Browser/OS: any Browsers/OSs
* How did you install Redash: from Amazon AMI

### Details

When accessing Treasure Data to get schema, always default region will be set because the parameter is not prepared.
https://github.com/getredash/redash/blob/6c364369bb0eb98e2191c2e502fed72abe5a74c7/redash/query_runner/treasuredata.py#L82

mickeey2525:

- [X] Other

## Description
So far, TreasureData Query runner could not identify endpoint correctly and uses only `https://api.treasuredata.com`

Closes #3185

### DIFF
--- a/redash/query_runner/treasuredata.py
+++ b/redash/query_runner/treasuredata.py
@@ -68,7 +68,7 @@ class TreasureData(BaseQueryRunner):
         schema = {}
         if self.configuration.get("get_schema", False):
             try:
-                with tdclient.Client(self.configuration.get("apikey")) as client:
+                with tdclient.Client(self.configuration.get("apikey"),endpoint=self.configuration.get("endpoint")) as client:
                     for table in client.tables(self.configuration.get("db")):
                         table_name = "{}.{}".format(
                             self.configuration.get("db"), table.name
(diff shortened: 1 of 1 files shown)


redash/query_runner/treasuredata.py
```diff
@@ -68,7 +68,7 @@ def get_schema(self, get_stats=False):
         schema = {}
         if self.configuration.get("get_schema", False):
             try:
-                with tdclient.Client(self.configuration.get("apikey")) as client:
+                with tdclient.Client(self.configuration.get("apikey"),endpoint=self.configuration.get("endpoint")) as client:
```
arikfr: If `endpoint` is`None` it will use the default value?

mickeey2525: Yes, default parameter(https://api.treasuredata.com) is used

arikfr:
Thanks.

PR #4583, opened 2020-01-23
Remove core-js and polyfill include as we don't use Phantom anymore

arikfr:

- [x] Other

## Description

We switched to Puppeteer a long ago, so this is no longer relevant.


package.json
```diff
@@ -47,7 +47,6 @@
     "brace": "^0.11.0",
     "chroma-js": "^1.3.6",
     "classnames": "^2.2.6",
-    "core-js": "^2.6.4",
```
arikfr: @kravets-levko I was expecting it to update `package-lock.json` but it didn't. Do you know why?

kravets-levko: Probably because some other dependency still use it. You can try to search for `core-js` in `package-lock.json` to check this. I guess Babel may need CoreJS, maybe some other libs.

arikfr: That's what I thought too, but then I expected it to pick up a newer versions. Oh, well :man_shrugging:

kravets-levko: To get a newer version you have to run `npm install` again; `npm uninstall` just removes packages and does not update others.

arikfr: I actually just edited `package.json` :cowboy_hat_face: But I did run `npm install` afterwards.

kravets-levko: I checked `package-lock.json` - Babel and React have `core-js` among their dependencies (React has it through `fbjs` dependency)

PR #4584, opened 2020-01-23
Refactor useQueryExecution & add inline query execution status

arikfr:
## What type of PR is this?

- [x] Refactor

## Description

While trying to add an inline Query Execution status that won't remove the current results in display, I realized that we need to refactor `useQueryExecution` to allow the view to retain current results.

While the `QueryResult` class needs significant refactoring (probably split into two: one for execution and one for data), I did one change: add a callback to receive current execution status. This calllback allows us to remove the polling logic from `useQueryResult` and actually remove `useQueryResult` entirely. I did keep the `getQueryResultData` helper, but we should probably remove it when we get to do said refactoring.

I will add some more implementation notes in comments.

It's still not complete:

- [x] When triggering new execution while previous still running it messes up the status (both flashing).
- [ ] Need better empty state for when no results are available (error or otherwise).
- [x] Empty state ("no results, try refresh") flashes on first load.
- [ ] Make sure both View/Editor have the same query result information.
- [x] Add space between text and Cancel button.
- [ ] Consider adding "inline" query execution status in Query Editor too.
- [ ] Move QueryExecutionStatus elsewhere in mobile view.

Peek 2020-01-23 12-46


client/app/pages/queries/QuerySource.jsx
```diff
@@ -154,20 +159,14 @@ function QuerySource(props) {
         return;
       }
       if (isDirty || !isEmpty(selectedText)) {
-        executeAdhocQuery(selectedText);
+        executeQuery(null, () => {
```
arikfr: Need to improve the API here so we don't have to pass `null` as first argument if we're passing in custom executor.

client/app/pages/queries/QueryView.jsx
```diff
@@ -35,18 +37,10 @@ import useDeleteVisualization from "./hooks/useDeleteVisualization";
...
-function updateUrlSearch(...params) {}
-
 function useFullscreenHandler(available) {
-  // TODO: bring back URL sync
   const [fullscreen, setFullscreen] = useState(false);
   const toggleFullscreen = useCallback(() => setFullscreen(!fullscreen), [fullscreen]);
 
-  useEffect(() => {
-    updateUrlSearch("fullscreen", fullscreen ? true : null);
-  }, [fullscreen]);
```
arikfr: I removed the sync w/ URL on purpose. I'm not sure we need it here.

client/app/pages/queries/QueryView.jsx
```diff
@@ -159,18 +157,18 @@ function QueryView(props) {
           </div>
         )}
         <div className="query-results m-t-15">
-          {queryResult && queryResultData.status !== "done" && (
-            <div className="query-alerts m-t-15 m-b-15">
+          {(executionError || isExecuting) && (
+            <div style={{ position: "absolute", bottom: "75px", right: "20px", zIndex: 1000 }}>
```
arikfr: Need to move this into the stylesheet.

client/app/pages/queries/QueryView.jsx
```diff
@@ -159,18 +157,18 @@ function QueryView(props) {
...
+                status={executionStatus}
+                error={executionError}
                 isCancelling={isExecutionCancelling}
                 onCancel={cancelExecution}
+                updatedAt={updatedAt}
               />
             </div>
           )}
-          {(!queryResult || queryResultData.status === "done") && (
+          {true && (
```
arikfr: If we indeed keep viz always visible, need to remove this line.

client/app/pages/queries/QueryView.jsx
```diff
@@ -209,10 +207,10 @@ function QueryView(props) {
               <QueryViewButton
                 className="icon-button m-r-5 hidden-xs"
                 title="Toggle Fullscreen"
-                type={fullscreen ? "primary" : "default"}
+                type="default"
                 shortcut="alt+f"
                 onClick={toggleFullscreen}>
-                <i className="zmdi zmdi-fullscreen" />
+                <Icon type={fullscreen ? "fullscreen-exit" : "fullscreen"} />
```
arikfr: Switched to Ant's icons. If we like it, we should use it in other places where we have "fullscreen" toggle.

client/app/pages/queries/hooks/useQueryExecute.js
```diff
@@ -1,79 +1,90 @@
...
+      };
 
-  const executeQuery = useCallback(() => {
-    recordEvent("execute", "query", query.id);
-    setQueryResult(query.getQueryResult(0));
-  }, [query]);
+      newQueryResult
+        .toPromise(onStatusChange)
+        .then(queryResult => {
+          // TODO: this should probably belong in the QueryEditor page.
```
kravets-levko: A quick note: here (and in other similar places) we need to handle a case when component was destroyed during async operation (ignore all `.then`/`.catch` and callback calls otherwise all sort of `setState`s will trigger error).

arikfr: :+1:

gabrieldutra:
This is definitely in the direction I was thinking about (always having a result showing up). From the gif, imo the alert will need a different styling, I considered either having a notification (any from the Antd options) or just having the text saying the status of the execution in that footer. At most a colored Icon to emphasize the status.

client/app/pages/queries/QueryView.jsx
gabrieldutra: Yeah, I actually kept it thinking about sharing the URL and to have the same behavior the Dashboard page have. But both cases are not that important anyway.

client/app/pages/queries/QueryView.jsx
gabrieldutra: I'm leaned towards the zmdi one, but I think  I'm biased anyway.

client/app/pages/queries/QueryView.jsx
gabrieldutra: I think this is probably the best option, the query view page just look so empty when loading and it makes the whole process of updating results look smoothier :)

client/app/pages/queries/hooks/useQueryExecute.js
```diff
@@ -1,79 +1,120 @@
...
+  }, [executeQuery, query]);
 
   useEffect(() => {
-    if (!isQueryExecuting) {
-      setIsExecutionCancelling(false);
-      if (queryResult && queryResult.query_result.query === query.query) {
-        query.latest_query_data_id = queryResult.getId();
-        query.queryResult = queryResult;
-      }
+    // TODO: this belongs on the query page?
```
gabrieldutra: I tend to prefer this logic in here mainly for 2 reasons: 1 - Reduce complexity from Pages; 2 - It's a shared piece of code between View and Source.

In terms of where this code "belongs", it feels to me it could be in both the query page or in here.

client/app/pages/queries/hooks/useQueryExecute.js
```diff
@@ -1,79 +1,120 @@
...
-      queryResult.cancelExecution();
-      setIsExecutionCancelling(true);
-    }
-  }, [query.id, queryResult]);
+  const queryRef = useRef(query);
+  const executeQueryRef = useRef(executeQuery);
+  useEffect(() => {
+    queryRef.current = query;
+    executeQueryRef.current = executeQuery;
+  }, [executeQuery, query]);
```
gabrieldutra: This is the safest way I could think to remove the dependencies from the `useEffect` hook and make it only execute once.

gabrieldutra:
For "Empty state ("no results, try refresh") flashes on first load." I added a state to control whether the first results were loaded or not. So far I just keep it hidden until first results are loaded and it looks ok, apart from 2 cases:
1. If you use "maxAge" it will trigger a whole new query execution, thus take more time before showing up;
2. If you have a huge set of results. In this case I don't think introduce a loading state will solve the problem as the UI freezes for a while (I'm assuming it's due to the heavy transform calculations it does);

For 2, it's not in the scope in here, but explore WebWorkers for heavy calculations is passing through my mind (I think this could help with Dashboard performance issues as well).

client/app/pages/queries/hooks/useQueryExecute.js
kravets-levko: Is `useEffect` really needed here? Seems you can leave only assignment - assigning value to `ref.current` does not re-render component.

client/app/pages/queries/hooks/useQueryExecute.js
```diff
@@ -1,97 +1,132 @@
...
+          }
+        })
+        .catch(queryResult => {
+          if (queryResultInExecution.current === newQueryResult) {
+            if (executionState.loadedInitialResults) {
+              notifications.showNotification("Redash", `${query.name} failed to run: ${queryResult.getError()}`);
+            }
+
+            setExecutionState({
+              queryResult,
```
gabrieldutra: There's also the option of keeping old queryResults when a query has an error, but you were thinking about an Error State?

gabrieldutra:
@arikfr, I think most of the functionality side of this feature is done (I checked from the checkboxes in the description). Now it's mostly a matter of iterating design for both desktop and mobile.

For those I didn't understand very well what you mean:
1. "Need better empty state for when no results are available (error or otherwise)." -> (https://github.com/getredash/redash/pull/4584#discussion_r375008432 should be related)
2. "Make sure both View/Editor have the same query result information." -> You mean to keep the query result information when you change between view/edit? (if it's what you meant, I can do it before iterating design)
3. "Consider adding "inline" query execution status in Query Editor too." -> Is it related to the "jumping" that happens on the existing position for the execution status?

client/app/pages/queries/hooks/useQueryExecute.js
arikfr: It's not for this PR, but eventually it makes sense to use `useQueryExecute` in Widgets, where this logic isn't needed.

client/app/pages/queries/hooks/useQueryExecute.js
arikfr: 🤔 not sure what I had in mind here. Will experiment with both options and decide.

arikfr:
> 1. "Need better empty state for when no results are available (error or otherwise)." -> ([#4584 (comment)](https://github.com/getredash/redash/pull/4584#discussion_r375008432) should be related)

I need to try it again, but I remember that there were cases when there are no results and things look a bit too empty.

> 2. "Make sure both View/Editor have the same query result information." -> You mean to keep the query result information when you change between view/edit? (if it's what you meant, I can do it before iterating design)

No, I meant show the same information (and potentially reuse the same component). Right now we show different things in view and edit.

> 3. "Consider adding "inline" query execution status in Query Editor too." -> Is it related to the "jumping" that happens on the existing position for the execution status?

No, just for consistency.

arikfr:
1. Query Editor: "No results found" empty state is shown on New Query page.
2. "No Results found" is shown when query returns 0 rows, which is fine I guess but we should update the text in this case. The original design/text was for cases when you open a query page and it has no cached results, but for just a regular case of 0 rows it's a bit too much. I'll take care of this.
2. The inline status should have consistent width.

arikfr:
4. Description is unaligned.

gabrieldutra:
> No, I meant show the same information (and potentially reuse the same component). Right now we show different things in view and edit.

:+1: I think I got it now, I updated the QueryExecutionMetadata to be a separate component. Now it's the same for both view and edit.

---

1. Updated :+1: 
2. I separated in two different cases and it's a matter of refining the message now. We'll also need to handle the case where the visualization should be rendered with 0 rows. I think the only existing case now is a Counter with `countRows`(?) If so, we can create an exception just for it now and open an issue to make it more general as it seems out of the scope here.
3. Added a `min-width` that covers the non-error messages :+1:. 
4. Updated :+1:

arikfr:
> We'll also need to handle the case where the visualization should be rendered with 0 rows. I think the only existing case now is a Counter with countRows(?) If so, we can create an exception just for it now and open an issue to make it more general as it seems out of the scope here.

It makes sense for the visualization to decide how it handles 0 rows. Why not render the viz in such case?

arikfr:
When clicking on "Add Description" it should auto focus the input field.

gabrieldutra:
> It makes sense for the visualization to decide how it handles 0 rows. Why not render the viz in such case?

:thinking:, the simplest should be to just pass an `EmptyState` node to the `VisualizationRenderer`. I'll see how easy it is to make the default behavior for visualizations, but also not be hard to replace it,

gabrieldutra:
:beetle: Just noticed there's some change specifically in this PR that breaks Filters with multiple React updated (only multi-filters perhaps). E.g.: https://deploy-preview-4584--redash-preview.netlify.com/queries/334 

Will investigate

arikfr:
What will pass the empty state node? The viz implementation?

gabrieldutra:
> What will pass the empty state node? The viz implementation?

QueryPage pass emptyState -> General Vis Render checks if visualization `isEmpty(data, options)` (default to be rows.lenght = 0, but changeable by visualizations). So by default what passes the emptyState is the context (query/widget), but it can be changed to the visualization by changing that isEmpty function. This is what I could think not to go on each visualization and add an empty state... Also as much as it makes sense for it to have a visualization controlling it's own empty state, it does for it to have a "Query" or "Widget" shared one... Anyway, lmk your opinion

arikfr:
Let's minimize changes to the code or behavior and delegate to the visualization to handle this (which at the moment means do nothing -- like we had before).

I'll try to think about this by considering the various visualizations & states (query page vs. widget) and come up with a definition for future improvement.

arikfr:
@gabrieldutra I did two small changes here:

1. Add label to the field Data Source field.
2. Hide navbar in full screen mode (and reuse dashboard's `useFullscreenHandler`).

There are still few small design fixes to do, but I will merge this (and the parent PR) now and open a separate issue detailing them.

PR #4586, opened 2020-01-23
Dynamic Form: boolean fields related fixes

arikfr:

- [x] Bug Fix

## Description

Two fixes:

* When a boolean field had default value of `false` we would set `false` as the default value instead of the string value, which fails prop types.
* When an extra boolean field had default value of true it would make the extra fields section to be open.

### DIFF
--- a/client/app/components/dynamic-form/DynamicForm.jsx
+++ b/client/app/components/dynamic-form/DynamicForm.jsx
@@ -57,8 +57,13 @@ class DynamicForm extends React.Component {
     super(props);
 
     const hasFilledExtraField = some(props.fields, field => {
-      const { extra, initialValue } = field;
-      return extra && (!isEmpty(initialValue) || isNumber(initialValue) || (isBoolean(initialValue) && initialValue));
+      const { extra, initialValue, placeholder } = field;
+      return (
+        extra &&
+        (!isEmpty(initialValue) ||
+          isNumber(initialValue) ||
+          (isBoolean(initialValue) && initialValue.toString() !== placeholder))
+      );
     });
 
     const inProgressActions = {};
--- a/client/app/components/dynamic-form/DynamicForm.jsx
+++ b/client/app/components/dynamic-form/DynamicForm.jsx
@@ -100,7 +100,7 @@ class DynamicForm extends React.Component {
     this.props.form.validateFieldsAndScroll((err, values) => {
       Object.entries(values).forEach(([key, value]) => {
         const initialValue = this.props.fields.find(f => f.name === key).initialValue;
-        if ((initialValue === null || initialValue === undefined || initialValue === '') && value === '') {
+        if ((initialValue === null || initialValue === undefined || initialValue === "") && value === "") {
           values[key] = null;
         }
       });
--- a/client/app/components/dynamic-form/dynamicFormHelper.js
+++ b/client/app/components/dynamic-form/dynamicFormHelper.js
@@ -9,7 +9,7 @@ function orderedInputs(properties, order, targetOptions) {
       name: key,
       title: properties[key].title,
       type: properties[key].type,
-      placeholder: properties[key].default && properties[key].default.toString(),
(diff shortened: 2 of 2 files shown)


client/app/components/dynamic-form/DynamicForm.jsx
```diff
@@ -57,8 +57,13 @@ class DynamicForm extends React.Component {
...
 
     const hasFilledExtraField = some(props.fields, field => {
-      const { extra, initialValue } = field;
-      return extra && (!isEmpty(initialValue) || isNumber(initialValue) || (isBoolean(initialValue) && initialValue));
+      const { extra, initialValue, placeholder } = field;
+      return (
+        extra &&
+        (!isEmpty(initialValue) ||
+          isNumber(initialValue) ||
+          (isBoolean(initialValue) && initialValue.toString() !== placeholder))
```
kravets-levko: Shouldn't `initialValue.toString() !== placeholder` check be applied to other field types?

client/app/components/dynamic-form/dynamicFormHelper.js
```diff
@@ -9,7 +9,7 @@ function orderedInputs(properties, order, targetOptions) {
       name: key,
       title: properties[key].title,
       type: properties[key].type,
-      placeholder: properties[key].default && properties[key].default.toString(),
+      placeholder: (properties[key].default === false || properties[key].default) && properties[key].default.toString(),
```
kravets-levko: If I understand this correctly, it should add placeholder for fields with default value `false`, right? If yes - isn't it better to use something like `isNil(properties[key].default && toString(properties[key].default)`?

client/app/components/dynamic-form/DynamicForm.jsx
arikfr: For the other fields we have the `isEmpty` check. But booleans don't have an empty state, so we need to compare with the default.

client/app/components/dynamic-form/dynamicFormHelper.js
arikfr: I'm not sure I understand your question, but basically if a default value exists, it takes the default value and converts it into a string. Previously the check would miss default value of `false` as it doesn't pass the check -- this change fixes this.

kravets-levko: My question is about `(properties[key].default === false || properties[key].default)`. I understand why there's a check for `properties[key].default === false`, but second part will not match, say, number `0` (it's evaluated to false), so fields with default numeric value `0` will not have placeholder as well. So I suggested to replace that entire thing with `isNil` which basically is a check for `value === undefined || value === null` - so all fields that have defined default value will have a placeholder. There is also a case when default value is empty string, but in this case it will receive an empty string placeholder - the same as with current code.

kravets-levko: I don't know if numeric fields with default `0` should have a placeholder at all, so maybe it's expected behaviour and my comment does not make any sense :man_shrugging:

arikfr: You're right the nil check will work better. I'll fix this. 

As for the question about numbers, I'm not sure. You're probably right but I would rather not make tto many changes here as I don't fully understand the code. 😬

gabrieldutra: Looking again at this, from what I remember this is just supposed to make any `default` value a placeholder in the inputs. It looks like Lodash's `toString` does exactly that (converts all to String, except `null` or `undefined`).

Having only this lodash function should make it a bit more readable (at list in this line :sweat_smile:).

arikfr: `_.toString` returns an empty string for `null`/`undefined`.

gabrieldutra: Yes, but it seemed ok to have `placeholder: ""` in such cases as the result should remain the same :slightly_smiling_face:

PR #4588, opened 2020-01-24
Replace some Bootstrap components/classes with Ant analogues

kravets-levko:

- [x] Refactor
- [x] Other

#### What's changed:

- replaced Bootstrap's `.list-group` with Antd `<List>` (`QuerySelector` component and `Home` page);
- removed Bootstrap form classes (`.form-group`, `.form-control`);
- replaced Bootstrap's `.badge` with Antd `<Badge>` (System Status page);
- replaced markup based on `.btn-favourite + .fa-star` with `<FavoritesControl>` component (and added a readonly mode to it) - for consistency;
- replace Bootstrap's buttons (`.btn`) - with `<Button>` in `Parameters` component (see screenshot) and Query List page's empty state block (looks almost the same). There was also usage in VisualizationWidget component but it didn't really need that `.btn` styles (most were overridden) - so I just simplified a markup and removed classes at all.

#### What's left (will be fixed in this PR only after discussion if we decide so):

- a lot of helper classes: `m-*`, `p-*`, `bg-*`, `c-*`, `f-*`, `text-*`, `w-*`, `h-*`, display (`d-*`) and flex markup. They are actually not a part of Bootstrap (maybe just some of `text-*` ones) - so probably they should be collected into a single file (or few files) and kept apart from our custom Bootstrap theme;
- `.table-responsive`, `.table-bordered` - there are few usages, need to check what could be used as a replacement;
- `.hidden-*`, `.visible-*` - same as for table classes above;
- Bootstrap grid: `.container`, `.row`, `.col-*` - not that much usages, _may_ be replaced as a part of this PR (after discussing);
- `.label-*` classes: part of Bootstrap, but currently used only as tags labels. _May_ be replaced as a part of this PR (after discussing).

Also, Bootstrap defines some styles for plain tags (e.g. headers, typography, etc.) - need to check if Ant has something similar and how it fits.

Parameters components with updated button:
image


client/app/assets/less/ant.less
```diff
@@ -250,6 +250,31 @@
   }
 }
 
+.@{list-prefix-cls}-bordered.@{list-prefix-cls}-sm {
```
kravets-levko: The actual change, other is from Prettier

client/app/assets/less/ant.less
```diff
@@ -394,9 +419,16 @@
...
+@checkbox-prefix-cls: ~"@{ant-prefix}-checkbox";
 
 .@{checkbox-prefix-cls}-wrapper + span,
 .@{checkbox-prefix-cls} + span {
   padding-right: 0;
 }
+
+// set some defaults for badge
+.@{badge-prefix-cls} {
+  &-count {
```
kravets-levko: The actual change, other is from Prettier

client/app/components/Parameters.less
```diff
@@ -20,27 +20,43 @@
...
     margin-bottom: 1px;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    min-width: 100%;
-    max-width: 195px;
-    white-space: nowrap;
-
-    .parameter-block[data-editable] & {
-      min-width: calc(100% - 27px); // make room for settings button
-      max-width: 195px - 27px;
```
gabrieldutra: It may be trickier to reach the same Label behavior for parameters:

**Before**
image
**After**
image

The idea is to keep it with a flexible width, if the label itself has more text and the width is not along it's limit, the whole parameter can take up more space :slightly_smiling_face:.  You couldn't get it to work with the previous css or you just wanted to move it to flex?

kravets-levko: I wanted to make it flex (and get rid of `calc`). Thank you for pointing, I'll check this once more :+1:

guidopetri:
@kravets-levko , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

Issue #4566, opened 2020-01-20
Refactor showModal to avoid having catch everywhere


arikfr:
(follow up to [this conversation](https://github.com/getredash/redash/pull/4525#discussion_r368512384))

https://github.com/getredash/redash/blob/2755e1275af400b9bca167602d33cfc9b8e52418/client/app/components/tags-control/TagsControl.jsx#L27-L30

@arikfr:

> > Coudln't we handle this in showModal instead of having to add this everywhere?

@kravets-levko 

> Sadly, no  After `.catch()` handler promise will go to next `.then()` handler, so if we catch that promises inside of dialog wrapper - the caller's `.catch()` will not be called. It actually is not a problem, but browser will show unhandled rejection errors in console. We can either:
> 
> 1. don't catch that promises and ignore errors in console;
> 2. always add empty catch handler like I did here;
> 3. replace browser's Promise with something else that does not trigger unhandled rejection errors (e.g. "bluebird") - prior to this PR we used Angular's `$q` for this purpose;
> 4. change a way we handle dialog's close/dismiss events (for me current flow looks fine, but that annoying unhandled rejections...)

arikfr:
@kravets-levko here's another suggestion:

When the dialog is closed or dismissed always call resolve the promise, but resolve it with values that represent whether it was dismissed and what the result was.

Is this possible?

kravets-levko:
Yes, sure - it's actually option 4 from my list. Also I posted one more comment in that thread:

> I'll explore if we can suppress that error with handling `unhandledrejection` event if error comes from dialog wrapper

arikfr:
I don't think the last suggestion is a good one. There just shouldn't be an error when there isn't one :)

kravets-levko:
Rejected promise is not an error, actually. I like current flow because it clearly describes two paths - when used accepted some action (then dialog is "resolved"), and when cancelled (then dialog is "rejected"). So when you need to handle only success path - you don't have to add any `if`s to your code.

arikfr:
We could avoid using promises here and just pass a callback for dialog closed and another (optional) for dialog cancelled. This way handler stays the same and we don't have to abuse a promise.

PR #4594, opened 2020-01-27
Refactor dialog wrapper component

kravets-levko:

- [x] Refactor
- [x] Feature
- [x] Other

## Description

- [x] Stop using promises to get dialog result; use callback instead.
- [x] Handle async operations on dialog close/dismiss: show spinner on corresponding button (depending on action - close or dismiss) and wait until operation completes; if operation fails - keep modal visible. This pattern was used in a lot of dialogs, so now this feature is "built-in" and makes dialogs UX more consistent.
- [x] Refactor: update some dialogs to be function-based components (reduced amount of code and simplified code a bit). Updated only dialogs that needed feature from [2]. 

Also fixed some minor bugs found during working on these changes.

Closes getredash/redash#4566


kravets-levko:
@gabrieldutra 

>  One single thing I would review: the onClose name can be a bit confusing, wdyt about using something else? (e.g: onOk - the same Antd uses)

I see your point, but there's one thing that may be confusing in this case: there are two methods to close dialog: `dialog.close(...)` to accept it and `dialog.dismiss(...)` to dismiss. And I tried to keep naming consistent. If we'll decide to rename `onClose` to something else (e.g. `onAccept`) - then we have to rename `dialog.close()` method as well (`dialog.accept(...)` - ?). So WDYT?

gabrieldutra:
For the dialog commands it does feels better to have `close` and `dismiss` (those look more like actual actions). So I think we can leave as is and just rename it in case it becomes confusing.

kravets-levko:
So you offer to keep `dialog.close()/.dismiss()` methods names, but rename handlers to `onAccept/onDismiss`, right? Looks a bit weird, but why not :slightly_smiling_face:

gabrieldutra:
> So you offer to keep `dialog.close()/.dismiss()` methods names, but rename handlers to `onAccept/onDismiss`, right? Looks a bit weird, but why not slightly_smiling_face

I actually meant to leave all names as they are :sweat_smile:, but I'm good with any of those options.

kravets-levko:
> I actually meant to leave all names as they are , but I'm good with any of those options.

So okay then, let's keep it as is (does this mean that I'm too lazy? :thinking:) I addressed other comments and resolved conflicts - can you please take one more look before I merge it?

PR #4595, opened 2020-01-27
Sort schema columns alphabetically

susodapop:

- [X] Feature

## Description
This PR sorts the column names shown in the schema viewer alphabetically at the API level. It also moves table name sorting from the Model code into the `get_schema()` method of the schema resource. If the sort raises an exception then the API returns the columns in the order sent by the query runner.

Users will need to refresh their schemas before the change is visible in the front-end.

https://github.com/redashlabs/product/issues/15. 

## Before
before-edited

## After
after-edited

### DIFF
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -200,12 +200,13 @@ class DataSourceSchemaResource(BaseResource):
             response["error"] = {"code": 2, "message": "Error retrieving schema."}
 
         try:
-            sorted_schema = [{"name": i['name'], "columns": sorted(i['columns'])} for i in schema]
+            sorted_schema = [{"name": i['name'], "columns": sorted(i['columns'])}
+                for i in sorted(schema, key=lambda x: x['name'])]
             response["schema"] = sorted_schema
         except Exception:
             logging.exception(
                 "Error sorting schema columns for data_source {}"\
-                .format(self.data_source.id))
+                .format(data_source_id))
             response['schema'] = schema
 
         return response
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -190,7 +190,7 @@ class DataSourceSchemaResource(BaseResource):
         response = {}
 
         try:
-            response["schema"] = data_source.get_schema(refresh)
+            schema = data_source.get_schema(refresh)
         except NotSupported:
             response["error"] = {
                 "code": 1,
@@ -198,6 +198,15 @@ class DataSourceSchemaResource(BaseResource):
             }
         except Exception:
             response["error"] = {"code": 2, "message": "Error retrieving schema."}
+
+        try:
+            sorted_schema = [{"name": i['name'], "columns": sorted(i['columns'])} for i in schema]
+            response["schema"] = sorted_schema
+        except Exception:
+            logging.exception(
+                "Error sorting schema columns for data_source {}"\
+                .format(self.data_source.id))
+            response['schema'] = schema
 
         return response
 
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -190,7 +190,7 @@ class DataSourceSchemaResource(BaseResource):
         response = {}
 
         try:
-            schema = data_source.get_schema(refresh)
+            response['schema'] = data_source.get_schema(refresh)
         except NotSupported:
             response["error"] = {
                 "code": 1,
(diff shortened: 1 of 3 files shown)


redash/handlers/data_sources.py
```diff
@@ -199,6 +199,15 @@ def get(self, data_source_id):
         except Exception:
             response["error"] = {"code": 2, "message": "Error retrieving schema."}
 
+        try:
+            sorted_schema = [{"name": i['name'], "columns": sorted(i['columns'])} for i in schema]
+            response["schema"] = sorted_schema
```
kravets-levko: Shouldn't it sort tables as well (or it already does)?

susodapop:
I'd like to add tests for this behavior. What's the best way to do so? I have a few ideas:

1) Inject a schema into the data source factory
2) `patch` the `get_schema` method to return an unordered list
3) Build an in-memory data-source and point SQLA there

redash/handlers/data_sources.py
susodapop: The tables are sorted in API handler. It makes sense to move it here, though.

https://github.com/getredash/redash/blob/19c6d331b6f2d22006bfcda3feab12c86c076afc/redash/models/__init__.py#L191

susodapop: I pushed a commit that moves table name sorting to the schema handler.

redash/models/__init__.py
```diff
@@ -187,9 +187,7 @@ def get_schema(self, refresh=False):
 
         if cache is None:
             query_runner = self.query_runner
-            schema = sorted(
-                query_runner.get_schema(get_stats=refresh), key=lambda t: t["name"]
-            )
+            schema = query_runner.get_schema(get_stats=refresh)
```
arikfr: Why not do all the sorting here and then store the sorted version in cache?

susodapop: I'm not picky. I put it in the Resource code because key sorting seems more like a "view concern" than a "model concern". But it really doesn't matter.

susodapop: I moved this logic all into the model.

redash/models/__init__.py
```diff
@@ -187,11 +187,18 @@ def get_schema(self, refresh=False):
...
+
+            try:
+                out_schema = [{"name": i['name'], "columns": sorted(i['columns'])}
+                for i in sorted(schema, key=lambda x: x['name'])]
```
arikfr: Are you using `black`? The indentation here is confusing :thinking:

susodapop: I wasn't before. Just added it to my env.

susodapop:
I used @arikfr's second testing option. Extracted the sort into its own function and tested with dummy data.

susodapop:
Pending further feedback I think this is ready to merge.

susodapop:
Not sure what's causing CircleCI e2e tests to fail. I didn't change any front-end files.

gabrieldutra:
> Not sure what's causing CircleCI e2e tests to fail. I didn't change any front-end files.

:sleepy:, looks like it kept flaky after #4545, for now it's a matter of re-trigger the build. I guess Cypress indeed keeps the chained element and we will need to make sure the results are fresh before asserting (cc @kravets-levko).

kravets-levko:
@gabrieldutra I found another thing that makes pivot test flaky: cypress input is too fast. When test adds a new line to query - it immediately clicks Execute button. But query editor is debounced, so 50/50 it will execute query with old text - therefore test fails. Here is my attempt to fix this - https://github.com/getredash/redash/commit/ec2ede4c414ffcd02b447cb24023031c55fdf79a I have no other idea for how to detect when query object received a new text from editor.

susodapop:
I'm continuing to test local. This isn't ready to merge yet. When connected to Redash's internal metadata database the schema sort doesn't work at all.

susodapop:
Found the issue. When I refactored to `_sort_schema()` I returned the wrong variable from the sort method. And I trusted my test instead of testing manually.

```
          _ __            
   __  __(_) /_____  _____
  / / / / / //_/ _ \/ ___/
 / /_/ / / ,< /  __(__  ) 
 \__, /_/_/|_|\___/____/  
/____/                    
                          
```

susodapop:
Alrighty. I added a second test case that would have caught this issue addressed in a31f901.

susodapop:
On my second unit test, would it be better to use `call_count` to ensure that the `sort_schema()` method is called? It feels strangely duplicative using identical assertions on separate tests.

The first test verifies the sort behavior. The second test verifies that Redash _uses_ that sort behavior. Seems like the `call_count` is the important bit.

Thoughts @arikfr or @levko?

MRdevX:
We're using Redash Version: 24.04.0-dev and the sorting functionality is not working, I saw there is a new MR from @susodapop 
 to resolve this issue, is there any plan for it? It would be beneficial to have this feature on Redash.

PR #4596, opened 2020-01-28
Counter: tooltip format setting

kravets-levko:

- [x] Feature

## Description

Counter visualization has a hidden setting for value tooltip format (applied to counter and target value if they're numbers). This PR adds ability to edit this setting in editor.

image


kravets-levko:
> What will it take to convert the configuration for the value/target to use the string format?

Do you mean value/target tooltips or value/target themselves?

arikfr:
> > What will it take to convert the configuration for the value/target to use the string format?
> 
> Do you mean value/target tooltips or value/target themselves?

Themselves.

kravets-levko:
I think one day at most, but it will require migration (and I may need some assistance with it)

arikfr:
Can't we just translate old options into new string in runtime?

kravets-levko:
> Can't we just translate old options into new string in runtime?

That's possible. I think I'll implement it and update this PR to include this feature.

kravets-levko:
Replaced by getredash/redash#4642

PR #4574, opened 2020-01-28
RQ: reuse same names as before for Docker entrypoints

rauchy:
Now that we no longer have Celery around, let's see if we can have the same Docker entrypoints (i.e. functions of the `docker_entrypoint` script) as before to make upgrades simpler for people.

The only one I'm not sure about is scheduler.

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -39,10 +39,6 @@ create_db() {
   exec /app/manage.py database create_tables
 }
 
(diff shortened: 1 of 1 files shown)


arikfr:
Previously you could use `scheduler` to handle the default queue. But that's no more. 

Maybe we should use Supervisor to execute two processes when running scheduler in "compatibility" mode? (one for processing jobs and one for scheduler)

PR #4600, opened 2020-01-30
Base duplicate index on column names

rauchy:
# What type of PR is this? (check all applicable)

- [x] Bug Fix

## Description

I've noticed that duplicate column names are indexed according to the amount of duplicates, and not based off the actual column name duplicates. For example: if you join the users table with the organizations table you would see `name, created_at, updated_at, name1, created_at2, updated_at3` where I would expect `name, created_at, updated_at, name1, created_at1, updated_at1`.

Not sure if this is intentional, and also it will probably break existing queries, so feel free to reject this :)

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -1,4 +1,5 @@
 import logging
+from collections import defaultdict
 from contextlib import ExitStack
 from functools import wraps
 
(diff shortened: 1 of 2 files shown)


rauchy:
Duplicate of #4353.

rauchy:
Following the edge cases provided in #4353, I believe 03d7836b should be a better solution.

guidopetri:

justinclift:
Attempted a manual merge just now.  No idea if it'll work, but it'll probably fail the format check.  But, lets see what happens... :smile:

justinclift:
Just fixed the formatting with `make format`. :smile:

guidopetri:
I think we should merge this one instead of #4353 , since this one fully encompasses the other PR. I also think we might want a test here before merging

justinclift:
Good thinking @guidopetri, yep we'll want some kind of test. :smile:

Issue #4605, opened 2020-02-04
Pages with "resolve" don't redirect guests to the login page


gabrieldutra:
### Issue Summary

Currently only the Query Pages use it.

Opening an issue more to discuss how to proceed, since there was discussion regarding not using 'resolve' in pages anymore, I think this is the way we'll go.

One thing to see is that without the "resolve", the Query Pages will now need a Loading State (or just have nothing rendered until finish with loading like the dashboard pages).

### Steps to Reproduce

1. Open any query page logged out (can just open in incognito), like [this one](https://redash-preview.netlify.com/queries/327)
2. Verify that the 404 page takes precedence and it does not redirect to the login page

![resolve-issue](https://user-images.githubusercontent.com/3356951/73704452-eb6b2c00-46d1-11ea-9969-f19afab44918.gif)

### Technical details:

* Redash Version: Latest
* Browser/OS: *
* How did you install Redash: *

kravets-levko:
Good catch @gabrieldutra! Luckily, yesterday I already started working on removing `resolve` feature :smile:

PR #4607, opened 2020-02-04
Remove `route.resolve` feature

kravets-levko:

- [x] Refactor
- [x] Bug Fix

## Description

During React migration of Router we decided to temporary keep `resolve` feature like in `angular-router` (this feature allows route to define dependencies that should be resolved before rendering page). Till now only Query View and Editor pages were using this feature, so it's time to remove it.

Now query pages will load query object themselves showing spinner while loading. It's a little bit different from previous behavior (with `resolve` feature previous page stays visible until new is completely ready), but considering that other pages already work in the same manner - it shouldn't be an issue.

Fixes getredash/redash#4605


kravets-levko:
> As for other pages, the dashboard currently doesn't have a LoadingState (it's just blank) and the Alert Page has this spinner, but with a `margin-top: 30px`. If the matter is a quick win I'd go with just adding this margin.

I also noticed that different pages have different spinners. I was planning to revisit all that stuff together (it will also affect ItemsList component), and for me it feels a bit out of scope of this PR.

client/app/pages/queries/components/wrapQueryPage.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { useState, useEffect, useRef } from "react";
+import PropTypes from "prop-types";
+import LoadingState from "@/components/items-list/components/LoadingState";
+import { Query } from "@/services/query";
+
+export default function wrapQueryPage(WrappedComponent) {
+  function QueryPageWrapper({ queryId, onError, ...props }) {
+    const [query, setQuery] = useState(null);
+    const onErrorRef = useRef();
+    onErrorRef.current = onError;
```
arikfr: Why do we need to persist the `onError` callback?

PR #4615, opened 2020-02-06
Handle stale jobs more carefully before purging them.

jezdez:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

In case of a dead worker, the ended_at value could be None, preventing
the purge task from running successfully and leading to the purge never
running successfully.

### DIFF
--- a/redash/tasks/general.py
+++ b/redash/tasks/general.py
@@ -72,16 +72,23 @@ def purge_failed_jobs():
         for queue in Queue.all():
             failed_job_ids = FailedJobRegistry(queue=queue).get_job_ids()
             failed_jobs = Job.fetch_many(failed_job_ids, rq_redis_connection)
-            stale_jobs = [
-                job
-                for job in failed_jobs
-                if job
-                and (datetime.utcnow() - job.ended_at).seconds
-                > settings.JOB_DEFAULT_FAILURE_TTL
-            ]
+            stale_jobs = []
+            for failed_job in failed_jobs:
+                # the job may not actually exist anymore in Redis
+                if not failed_job:
+                    continue
+                # the job could have an empty ended_at value in case
+                # of a worker dying before it can save the ended_at value,
+                # in which case we also consider them stale
+                if not failed_job.ended_at:
+                    stale_jobs.append(failed_job)
+                elif (
+                    datetime.utcnow() - failed_job.ended_at
+                ).seconds > settings.JOB_DEFAULT_FAILURE_TTL:
+                    stale_jobs.append(failed_job)
 
-            for job in stale_jobs:
-                job.delete()
+            for stale_job in stale_jobs:
+                stale_job.delete()
 
             if stale_jobs:
                 logger.info(


redash/tasks/general.py
```diff
@@ -72,16 +72,23 @@ def purge_failed_jobs():
...
-                and (datetime.utcnow() - job.ended_at).seconds
-                > settings.JOB_DEFAULT_FAILURE_TTL
-            ]
-
-            for job in stale_jobs:
-                job.delete()
+            stale_jobs = []
+            for failed_job in failed_jobs:
+                # the job may not actually exist anymore in Redis
+                if not failed_job:
```
rauchy: How about we just `compact` these?

redash/tasks/general.py
```diff
@@ -72,16 +72,23 @@ def purge_failed_jobs():
...
+                    continue
+                # the job could have an empty ended_at value in case
+                # of a worker dying before it can save the ended_at value,
+                # in which case we also consider them stale
+                if not failed_job.ended_at:
```
rauchy: This feels more like an `or` conditional and less like a multi-branch statement to me.

redash/tasks/general.py
```diff
@@ -72,16 +72,23 @@ def purge_failed_jobs():
...
-                for job in failed_jobs
-                if job
-                and (datetime.utcnow() - job.ended_at).seconds
-                > settings.JOB_DEFAULT_FAILURE_TTL
-            ]
-
-            for job in stale_jobs:
-                job.delete()
+            stale_jobs = []
+            for failed_job in failed_jobs:
```
rauchy: If you share my thoughts on the other couple of comments, this whole block might be better represented by a filter on `failed_jobs`

jezdez: I'm not sure I follow, what do you mean with "a filter on `failed_jobs`"?

rauchy: I just mean that it feels like `stale_jobs` is just a sub-list of `failed_jobs` that satisfies a predicate. Something like:

```python
is_stale = lambda job: job.ended_at is None or
                       (datetime.utcnow() - job.ended_at).seconds > settings.JOB_DEFAULT_FAILURE_TTL
stale_jobs = filter(is_stale, compact(failed_jobs))
```

jezdez: While you may be right that this is another way to write it, I don't see this as more readable. But it's up to you, feel free to change the patch the way you like it better.

arikfr: I'm fine with keeping @jezdez's implementation as is, as it leaves room for explaining the different steps.

rauchy: 👍 

Don't feel strongly either way, but generally I'm more in the camp of having descriptive variable / function / lambda names instead of comments (i.e. `is_stale = worker_died or too_old`). They just expire slower than comments.

PR #4559, opened 2020-02-08
RQ: Keep Job object of query execution for longer period of time

rauchy:
We need to keep jobs for longer, as people can execute a query, close the machine, come back and it shouldn't fail.

Also, in case the job isn't found, we should return proper message instead of currently the server blowing up.


arikfr:
Please rebase with master, so we can rule out failures due to #4617.

rauchy:
> Please rebase with master, so we can rule out failures due to #4617.

Why would #4617 make a difference here?

arikfr:
Because it prevents running tests.

PR #4626, opened 2020-02-09
Set default query execution time limit to unlimited

rauchy:
- [x] Bug Fix

## Description
After migrating query executions to RQ, time limits were implicitly set by RQ to 180 seconds. This PR changes the default to 1 hour and adds some docs on how to remove time limits.

### DIFF
--- a/redash/settings/__init__.py
+++ b/redash/settings/__init__.py
@@ -222,13 +222,13 @@ STATIC_ASSETS_PATH = fix_assets_path(
     os.environ.get("REDASH_STATIC_ASSETS_PATH", "../client/dist/")
 )
 
-# Time limit for scheduled queries. Set this to -1 to execute without a time limit.
+# Time limit (in seconds) for scheduled queries. Set this to -1 to execute without a time limit.
 SCHEDULED_QUERY_TIME_LIMIT = int(
-    os.environ.get("REDASH_SCHEDULED_QUERY_TIME_LIMIT", 3600)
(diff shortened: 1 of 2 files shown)


redash/settings/__init__.py
```diff
@@ -222,6 +222,14 @@
     os.environ.get("REDASH_STATIC_ASSETS_PATH", "../client/dist/")
 )
 
+# Time limit for scheduled queries. Set this to -1 to execute without a time limit.
+SCHEDULED_QUERY_TIME_LIMIT = int(
+    os.environ.get("REDASH_SCHEDULED_QUERY_TIME_LIMIT", 3600)
```
arikfr: Now that I look at the existing implementation, I wonder if we should use `-1` as the default to remain backward compatible?

jezdez: +1

arikfr: @jezdez you have queries running longer than an hour?

jezdez: I'm not sure if currently, but it happened in the past. Granted, we're probably not the average user group.

arikfr: I think that an hour default value is definitely reasonable for most cases. I just wonder if we should cause issues for any existing deployments :\

jezdez: Well, I think I change my initial response to +0, since this is a settings we can set it to a larger value before deploy. IOW don't consider our use case as a blocker.

rauchy: I don't see any value in imposing a default limit. As long as it's configurable, a `-1` sounds good.

PR #4627, opened 2020-02-09
Fix: no need to encode strings anymore

### LINKED ISSUE #4611: Alerts: Korean characters in default template misbehave
### Issue Summary

Alert titles that contain Korean or Japanese characters are badly encoded if the default alert email template is used. The custom alert templates do not appear so affected.

### Steps to Reproduce

1. Make a query that returns logical data: `SELECT 1 "foo";`
2. Point an alert at this query that asserts `foo == 2`
3. Name the alert `숨 쉬는 법을 모르겠다`
4. Save the alert and add your email as an alert destination
5. Trigger the query

You will receive an email with the following subject:

`(TRIGGERED) b'\xec\x88\xa8 \xec\x89\xac\xeb\x8a\x94 \xeb\xb2\x95\xec\x9d\x84 \xeb\xaa\xa8\xeb\xa5\xb4\xea\xb2\xa0\xeb\x8b\xa4'`

You can workaround this using the custom template. The custom template properly encodes Korean and Japanese characters.

### Technical details:

* Redash Version: V9
* Browser/OS: FF

arikfr:

- [x] Bug Fix

## Description

The `encode` call is a relic from Python 2 an no longer needed.

Closes #4611.

### DIFF
--- a/redash/destinations/email.py
+++ b/redash/destinations/email.py
@@ -46,7 +46,6 @@ class Email(BaseDestination):
         logging.debug("Notifying: %s", recipients)
 
         try:
-            alert_name = alert.name
             state = new_state.upper()
             if alert.custom_subject:
                 subject = alert.custom_subject
@@ -54,7 +53,7 @@ class Email(BaseDestination):
                 subject_template = options.get(
                     "subject_template", settings.ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE
                 )
-                subject = subject_template.format(alert_name=alert_name, state=state)
+                subject = subject_template.format(alert_name=alert.name, state=state)
 
             message = Message(recipients=recipients, subject=subject, html=html)
             mail.send(message)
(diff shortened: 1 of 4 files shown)


redash/destinations/email.py
```diff
@@ -46,7 +46,7 @@ def notify(self, alert, query, user, new_state, app, host, options):
         logging.debug("Notifying: %s", recipients)
 
         try:
-            alert_name = alert.name.encode("utf-8", "ignore")
+            alert_name = alert.name
```
rauchy: just drop this variable and use alert.name in line 57 :)

arikfr: https://github.com/getredash/redash/pull/4627/commits/0a8663402921647531bf813d46b308d0f73490a6

arikfr:
Grepped the code for additional places we use `.encode` in and removed some more.

PR #4638, opened 2020-02-11
Fix comparison error when scale is None

### LINKED ISSUE #4786: error running query : ** '>' is not supported between instance of NoneType and 'int'
Issue Summary:
Database = Oracle 12c

`select count(*) from table `

throwing the following error

`error running query : ** '>' is not supported between instance of NoneType and 'int'`

Redash v9.0.0-alpha(dev)

cemremengu:

- [x] Bug Fix

## Description

Prevents `'>' not supported between instances of 'NoneType' and 'int'` error when scale is `None`

Closes #4786.

### DIFF
--- a/redash/query_runner/oracle.py
+++ b/redash/query_runner/oracle.py
@@ -35,7 +35,7 @@ class Oracle(BaseSQLQueryRunner):
     @classmethod
     def get_col_type(cls, col_type, scale):
         if col_type == cx_Oracle.NUMBER:
-            return TYPE_FLOAT if (scale or 0) > 0 else TYPE_INTEGER
+            return TYPE_INTEGER if scale is False else TYPE_FLOAT
         else:
             return TYPES_MAP.get(col_type, None)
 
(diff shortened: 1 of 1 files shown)


redash/query_runner/oracle.py
```diff
@@ -35,7 +35,7 @@ class Oracle(BaseSQLQueryRunner):
     @classmethod
     def get_col_type(cls, col_type, scale):
         if col_type == cx_Oracle.NUMBER:
-            return TYPE_FLOAT if scale > 0 else TYPE_INTEGER
+            return TYPE_INTEGER if scale is False else TYPE_FLOAT
```
arikfr: If `scale` can be `None`, then in such cases it will be classified as a float. Is this correct? If not, you might want to change the if statement.

cemremengu: My understanding is that if scale is *not* `None` or is `> 0`  (so basically it is a truthy value) it should be classified as a Float, otherwise Integer.

However, here is the [definition ](https://cx-oracle.readthedocs.io/en/latest/user_guide/sql_execution.html#id10) from `cx_Oracle` for `NUMBER`:
```
If the column is unconstrained (no precision and scale specified), the value will be returned as a float or an int depending on whether the value itself is an integer. In all other cases the value is returned as a float.
```

Maybe we should look at the value it self whether it is an int or float to decide at this point?

arikfr: > Maybe we should look at the value itself whether it is an int or float to decide at this point?

Sounds like the safest option, no?

PR #4392, opened 2019-11-23
Counter visualization enhancements

gabrieldutra:
## What type of PR is this? (check all applicable)

- [x] Feature

## Description
This PR will allow more aggregation types for the Counter Visualization.

So far I just started exploring it, I wanted to have the types definition centered in one place and also it makes sense to merge the existing `countRows`. Currently I centered that logic in utils because it's where the current counter logic is, but I'll later see a better place to that.

One thing that I'll need to see is what's the best way of handling backward compatibility for the `countRows`, I was thinking a migration as it would just define Types for existing counters, though I haven't done a migration on my own yet so just confirming that :sweat_smile:.

To do:
- [x] Add Extra Counter Types
- [x] Add Migration 
- [x] Update Tests
- [x] Does it make sense to have the same Types choice for the Target Value?

## Related Tickets & Documents

Closes #4366 
Closes getredash/redash#2558

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
counter-types

PR #4642, opened 2020-02-12
Counter enhancements

kravets-levko:

- [x] Refactor
- [x] Feature

## Description

Several enhancements for Counter visualization.

- [x] Reorganize options:
  - [x] convert "old" options to new schema;
  - [x] rename "Counter Value" to "Primary Value" and "Target Value" to "Secondary Value";
  - [x] make aggregation functions work for both primary and secondary value;
  - [x] add display and tooltip format options for both primary and secondary value;
  - [x] option to hide secondary value;
- [x] Tests

getredash/redash#4392

image
image
image


gabrieldutra:
Overall it looks easier to define a target value (I guess mostly because it's now a "Secondary value"). Also it's good to have the counter types for the Secondary value as well :)

A few comments from my first experience:
1. The new formatting option with `{{ @@formatted }}` can be a bit confusing, mostly because the input already contains it and there's no helper to say it. There's also the option of having the same template styling that the Alert Template has (e.g.: `{{RAW_VALUE}}` and  `{{FORMATTED_VALUE}}`, or one of those just as `{{VALUE}}`) which I think can be better than the double @;
2. Disable Secondary Value options when "Show Secondary Value" is unchecked;

As for the schema version for visualizations I don't think it's worth to do now, but I was reflecting if it may be a pattern we may want to abstract for other visualizations.

kravets-levko:
Thank you for feedback @gabrieldutra!

> 1. The new formatting option with `{{ @@formatted }}` can be a bit confusing, mostly because the input already contains it and there's no helper to say it. There's also the option of having the same template styling that the Alert Template has (e.g.: `{{RAW_VALUE}}` and  `{{FORMATTED_VALUE}}`, or one of those just as `{{VALUE}}`) which I think can be better than the double @;

There is a reason why I use "@@" there. In other visualizations where we use such format strings, we usually allow to use 1) fields from query result records 2) some computed values. Query result fields are available under their names (`{{ some_field }}`), and computed values - with `@@` prefix (`{{ @@computed }}`). For this visualization I also plan to allow to use query result record when available (ATM when using `rowValue` function). But of course `{{ @@value }}` and `{{ @@value_formatted }}` looks much better.

> 2. Disable Secondary Value options when "Show Secondary Value" is unchecked;

:+1:

> As for the schema version for visualizations I don't think it's worth to do now, but I was reflecting if it may be a pattern we may want to abstract for other visualizations.

As you can see, options structure considerably changed, so I decided to implement this "migration" to keep code cleaner and errorproof. I don't say that this implementation is perfect and ready to use in future. It just shows the idea of implementing such migrations, and definitely the implementation can be improved - so I'm open to discussion.

kravets-levko:
Update to my previous comment (since I already implemented some of mentioned things):

> 1. ...

All input will have a help icon explaining what variables can be used there (as for other visualizations).

Also, some screenshots of what I mean by using record if available. I think this feature will be definitely useful for tooltips, but also may be useful to format values itself (e.g., if someone stores currency in database per record so value cannot be hard-coded, or similar examples):

image
image

> 2. Disable Secondary Value options when "Show Secondary Value" is unchecked;

After answering, I realized that actually that checkbox only controls _visibility_ of secondary value. It means that user can setup secondary value to mark primary value green/red, but don't show secondary value. It's a subject to discuss of course. Maybe, we can implement "Don't use" function for primary/secondary value to disable it, and move "Show ..." checkbox to the bottom of options to reflect that it only controls visibility :thinking: 

image

kravets-levko:
> 2. Disable Secondary Value options when "Show Secondary Value" is unchecked;

I added new option to Type select that allows to disable Secondary value. "Show Secondary Value" checkbox is now placed after other options to not confuse user:

image
image

For Primary Value no changes

kravets-levko:
Formatting hints:

image
image

gabrieldutra:
> "Show Secondary Value" checkbox is now placed after other options to not confuse user

Having two options that can disable the Secondary Value isn't potentially confusing on its own? How about just having the "Disabled"/"Unused" type?

kravets-levko:
> Having two options that can disable the Secondary Value isn't potentially confusing on its own? How about just having the "Disabled"/"Unused" type?

Second option ("Show ...") does not disable Secondary value, it just hides the line from visualization, but secondary value will still be available and can be used to compare against Primary value. When "Unused" type if selected (naming could be changed) - secondary value will be not available at all, not just invisible.

kravets-levko:
@gabrieldutra I cannot figure out what's wrong with frontend unit tests. On CI they just hang (and CircleCI kills the job) and in log there is only last passed test; locally all tests pass fine (I tried to reproduce the issue many times).

gabrieldutra:
> Do you have any ideas what may cause the issue, or how I can debug it?

No idea :grimacing:, I tried to debug with CircleCI ssh, but in ssh `npm test` also runs fine :man_shrugging:.

client/app/visualizations/counter/getCounterData.js
```diff
@@ -0,0 +1,72 @@
...
+function getCounterValue(rows, valueOptions, counterOptions) {
+  const [value, additionalFields] = invoke(counterTypes[valueOptions.type], "getValue", rows, valueOptions);
+
+  if (!valueOptions.show || isNil(value)) {
+    return { value: null, display: null, tooltip: null };
+  }
+
+  const formatData = extend({}, additionalFields, {
+    "@@value": toString(value),
+    "@@value_formatted": isFinite(value) ? formatValue(value, counterOptions) : toString(value),
```
gabrieldutra: @kravets-levko, I couldn't get the `{{ column_name }}` to work neither for counters, nor for Charts. I wonder if it's missing an entry here?

gabrieldutra: Nevermind, I was putting the wrong column name :sweat_smile:.

client/app/visualizations/counter/Editor/PrimaryValueSettings.jsx
```diff
@@ -0,0 +1,22 @@
+import { omit } from "lodash";
+import React, { useCallback } from "react";
+import { EditorPropTypes } from "@/visualizations/prop-types";
+import CounterValueOptions from "./CounterValueOptions";
+import counterTypes from "../counterTypes";
+
+export default function PrimaryValueSettings({ options, data, onOptionsChange }) {
+  const onChange = useCallback(primaryValue => onOptionsChange({ primaryValue }), [onOptionsChange]);
```
gabrieldutra: Since nothing really depends on `onChange` for rendering condition, I think we can remove this `useCallback` and just have that method directly on the `onChange={}` below.

client/app/visualizations/counter/Editor/SecondaryValueSettings.jsx
```diff
@@ -0,0 +1,34 @@
+import React, { useCallback } from "react";
+import { Section, Checkbox } from "@/components/visualizations/editor";
+import { EditorPropTypes } from "@/visualizations/prop-types";
+import CounterValueOptions from "./CounterValueOptions";
+import counterTypes from "../counterTypes";
+
+export default function SecondaryValueSettings({ options, data, onOptionsChange }) {
+  const onChange = useCallback(secondaryValue => onOptionsChange({ secondaryValue }), [onOptionsChange]);
```
gabrieldutra: Same here about the `useCallback`

client/app/visualizations/counter/Renderer.jsx
```diff
@@ -44,15 +56,11 @@ export default function Renderer({ data, options, visualizationName }) {
...
-  const {
-    showTrend,
-    trendPositive,
-    counterValue,
-    counterValueTooltip,
-    targetValue,
-    targetValueTooltip,
-    counterLabel,
-  } = getCounterData(data.rows, options, visualizationName);
+  const { counterLabel, showTrend, trendPositive, primaryValue, secondaryValue } = getCounterData(
```
gabrieldutra: How about a `useMemo` to avoid recalculating on each re-render?

kravets-levko: Done :+1:

client/app/visualizations/counter/Editor/PrimaryValueSettings.jsx
kravets-levko: Done (and the second one as well)

client/app/visualizations/counter/Editor/PrimaryValueSettings.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+  return (
+    <React.Fragment>
+      <CounterValueOptions
+        counterTypes={omit(counterTypes, ["unused"])}
+        options={options.primaryValue}
+        data={data}
+        onChange={onChange}
+      />
+    </React.Fragment>
```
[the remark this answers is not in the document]
kravets-levko: Done :+1:

client/app/visualizations/counter/counterTypes.js
```diff
@@ -0,0 +1,51 @@
...
+    return rowNumber;
+  }
+  const wrappedIndex = (Math.abs(rowNumber) - 1) % rowsCount;
+  return rowNumber > 0 ? wrappedIndex : rowsCount - wrappedIndex - 1;
+}
+
+export default {
+  unused: {
+    name: "Unused",
+    getValue: () => [undefined, null],
```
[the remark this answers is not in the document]
kravets-levko: I changed `returnsColumnNames` option name to `canReturnRow` - unfortunately I have no better idea for it :slightly_frowning_face: Also, added a comment that explains options for counter types.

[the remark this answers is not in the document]
gabrieldutra: IMO ideally it should have `columnName` in the name to seem related to that setting, but I also don't have a better idea. Let's go with that and if eventually we add more conditional format options, we review that. Thanks for the explanation comment! :slightly_smiling_face:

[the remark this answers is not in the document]
kravets-levko: Anything with `columnName` looks not right because `getValue` actually returns not just a column names, but entire row (associated with this counter value).

kravets-levko:
@gabrieldutra I'm merging this, and let's address getredash/redash#4727 directly in your PR getredash/redash#4392

gabrieldutra:
> I'm merging this, and let's address #4727 directly in your PR #4392

I was actually thinking about closing that PR and targeting this one in master, but that works 😅

PR #4163, opened 2020-02-12
refresh_queries shouldn't break because of a single query having a bad schedule object

rauchy:
### Issue Summary

We recently had an issue where a single query had a bad schedule object:

```json
{
    "interval": 604800, 
    "until": null,
    "day_of_week": "Sunday", 
    "time": "Invalid date"
}
```

And it caused `Query.outdated_queries` to blow up, which in turn stopped `refresh_queries` from working.

We should make sure there is enough exception handling that a single query won't stop the scheduler from working.

### Technical details:

* Redash Version: 8

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -11,7 +11,6 @@ from sqlalchemy.event import listens_for
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import backref, contains_eager, joinedload, subqueryload, load_only
 from sqlalchemy.orm.exc import NoResultFound  # noqa: F401
-from sqlalchemy.sql import expression
 from sqlalchemy import func
 from sqlalchemy_utils import generic_relationship
 from sqlalchemy_utils.types import TSVectorType
@@ -618,18 +617,11 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
 
     @classmethod
     def outdated_queries(cls):
-        with_valid_schedule = and_(
-            Query.schedule.isnot(None),
-            expression.func.coalesce(
-                expression.text("schedule::json->>'interval'"), None
-            ).isnot(None),
-        )
-
         queries = (
             Query.query.options(
                 joinedload(Query.latest_query_data).load_only("retrieved_at")
             )
-            .filter(with_valid_schedule)
+            .filter(Query.schedule.isnot(None))
             .order_by(Query.id)
         )
 
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -638,32 +638,35 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
         scheduled_queries_executions.refresh()
 
         for query in queries:
+            try:
+                if query.schedule["until"]:
+                    schedule_until = pytz.utc.localize(
+                        datetime.datetime.strptime(query.schedule["until"], "%Y-%m-%d")
+                    )
 
-            if query.schedule["until"] is not None:
-                schedule_until = pytz.utc.localize(
-                    datetime.datetime.strptime(query.schedule["until"], "%Y-%m-%d")
+                    if schedule_until <= now:
+                        continue
+
+                retrieved_at = scheduled_queries_executions.get(query.id) or (
+                    query.latest_query_data and query.latest_query_data.retrieved_at
                 )
 
-                if schedule_until <= now:
-                    continue
-
-            if query.latest_query_data:
-                retrieved_at = query.latest_query_data.retrieved_at
-            else:
-                retrieved_at = now
-
-            retrieved_at = scheduled_queries_executions.get(query.id) or retrieved_at
-
-            if should_schedule_next(
-                retrieved_at,
-                now,
-                query.schedule["interval"],
-                query.schedule["time"],
-                query.schedule["day_of_week"],
-                query.schedule_failures,
-            ):
-                key = "{}:{}".format(query.query_hash, query.data_source_id)
-                outdated_queries[key] = query
+                if should_schedule_next(
+                    retrieved_at or now,
+                    now,
+                    query.schedule["interval"],
+                    query.schedule["time"],
+                    query.schedule["day_of_week"],
+                    query.schedule_failures,
+                ):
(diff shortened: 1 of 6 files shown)


rauchy:
While wrapping it in `try/except` blocks, I felt that `refresh_queries` has too much going on and wasn't easy to read, so I've taken the opportunity to refactor it into something (_I see as_) more readable.

redash/models/__init__.py
```diff
@@ -630,34 +638,35 @@ def outdated_queries(cls):
...
+                    query.schedule_failures,
+                ):
+                    key = "{}:{}".format(query.query_hash, query.data_source_id)
+                    outdated_queries[key] = query
+            except Exception as e:
+                logging.info(
+                    "Could not determine if query %d is outdated due to %s",
+                    query.id,
+                    repr(e),
+                )
```
arikfr: 1. This should go into Sentry.
2. Considering the way our scheduler works this is going to be spammy.  Maybe we should report this and then unschedule?

redash/tasks/queries/execution.py
```diff
@@ -43,16 +43,24 @@ def enqueue_query(
             job_id = pipe.get(_job_lock_id(query_hash, data_source.id))
             if job_id:
                 logger.info("[%s] Found existing job: %s", query_hash, job_id)
+                job_exists = Job.exists(job_id)
+                job_complete = None
+
+                if job_exists:
+                    job = Job.fetch(job_id)
```
arikfr: While it's not very likely, it still can happen that between the call to `exists` to `fetch`, the job will be removed. Is there a version of `fetch` that returns `None` if job doesn't exist? Or maybe just catch exception?

redash/tasks/queries/maintenance.py
```diff
@@ -27,85 +27,86 @@ def empty_schedules():
...
+def refresh_queries():
+    logger.info("Refreshing queries...")
 
-    statsd_client.gauge(
-        "manager.seconds_since_refresh", now - float(status.get("last_refresh_at", now))
-    )
+    outdated = models.Query.outdated_queries()
+    refreshable = _skip_unrefreshable_queries(outdated)
+    queries = _apply_default_parameters(refreshable)
+    enqueued = list(_enqueue_queries(queries))
```
arikfr: Wouldn't it be easier to understand, more performant (less loops) and simpler if it was something like:

```python
for query in models.Query.outdated_queries():
    # logic from _skip_unrefreshable_queries
    if not should_refresh_query(query):
        continue
    try:
        enqueue_query(
            # logic from _apply_default_parameters
            apply_default_parameters(query.query_text),
            query.data_source,
            query.user_id,
            scheduled_query=query,
            metadata={"Query ID": query.id, "Username": "Scheduled"},
      )
    except Exception as e:
        logging.info("Could not enqueue query %d due to %s", query.id, repr(e))
```

arikfr: Also, similar comment about exception logging as with the one in `outdated_queries`. Except that here we should only report to Sentry, but not unschedule.

redash/models/__init__.py
```diff
@@ -617,11 +618,18 @@ def past_scheduled_queries(cls):
 
     @classmethod
     def outdated_queries(cls):
+        with_valid_schedule = and_(
+            Query.schedule.isnot(None),
+            expression.func.coalesce(
+                expression.text("schedule::json->>'interval'"), None
+            ).isnot(None),
```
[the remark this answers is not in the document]
rauchy: Yes, it was. I just found the fact that the filtering logic is divided between the query and the enqueuing loops more confusing. I prefer a (_slightly_) more complex query over mixed concerns.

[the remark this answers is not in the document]
arikfr: But it's split anyway, because later we do tests on the schedule data to see if a query is outdated. How's checking if `interval` is `None` any different from the other tests?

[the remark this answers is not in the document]
arikfr: Btw, I'm not sure if this test is even needed. I believe it's from an interim version of the code where there was always a `schedule` object, but later we switched to using `None` to signal no schedule.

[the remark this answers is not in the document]
rauchy: > How's checking if interval is None any different from the other tests?

The only difference is how much complexity it adds. Checking for `until` was way harder to understand in the query.

>  later we switched to using `None` to signal no schedule

Yeah, looks like there are no residues of that so it's safe to remove.

redash/models/__init__.py
rauchy: Regarding #2 - this function should be idempotent. It's kinda strange to call `outdated_queries` and have your schedule change as a result (perhaps it was a volatile error?).

I feel more comfortable with calling `track_failure` in these cases - this way users could find out that there was an error with running the query, and it would increase the `schedule_failures` counter, which would push the next reschedule down the exponential backoff track.

rauchy: Actually it would make more sense to use `track_failure` in `_enqueue_queries` instead.

arikfr: I don't think `schedule_failures` will work here, because it won't reach the part of the code that calculates the next iteration.

While usually you're right in the approach, we're talking about queries with malformed schedule object. Also, sending a user a notification about it doesn't feel useful as in most cases they won't know what to do about it.

Another approach we can take is to introduce a `disabled` boolean to the `schedule` object which we will trigger in such cases (and later can review these). The downside here is that it won't work in case it's completely malformed (not even a valid JSON).

rauchy: well, `disabled` could be set to `True` if `schedule` is broken or `schedule['disabled']` is `True`.

arikfr: 👍

redash/tasks/queries/maintenance.py
rauchy: Simplicity / understandability is a matter of preference. Personally I thought of the journey from `outdated_queries` to `enqueued` as a set of transformations and filters, so I took a FP-based approach. I've switched to your suggestion in ef2eb39.

Regarding more performant - not sure, those were generator loops and I think there's a built-in optimization there.

redash/models/__init__.py
rauchy: Reporting to Sentry is done in 57b1d4f.

redash/tasks/queries/execution.py
rauchy: 24ca387

Issue #4632, opened 2020-02-10
Add setting to hide Plotly mode bar


daniellangnet:
I find that the "mode bar" on Plotly charts can be confusing to people and doesn't add any value in our use of Redash. Hiding it is as simple as adding `displayModeBar: false` to the Plotly configuration. 

I'm happy to submit a PR and make it a setting on the organization ("Hide Plot.ly mode bar"), but before doing so wanted to check if there was support for this idea?

arikfr:
@daniellangnet definitely don't mind adding this setting. You're welcome to submit this PR 👍

PR #4644, opened 2020-02-13
Added setting to hide Plotly mode bar

daniellangnet:
## What type of PR is this?
- [x] Feature

## Description
This adds an organization setting that allows hiding the Plotly mode bar.

_Note: I wasn't sure if this should be an environmental variable for the entire instance or an organization setting. It "felt" right to make it an org-level setting but I'm not sure what rules you guys follow for this._

https://github.com/getredash/redash/issues/4632
https://discuss.redash.io/t/add-setting-to-hide-plotly-mode-bar/5529

image

### DIFF
--- a/client/app/pages/settings/OrganizationSettings.jsx
+++ b/client/app/pages/settings/OrganizationSettings.jsx
@@ -168,6 +168,14 @@ class OrganizationSettings extends React.Component {
             ))}
           </Select>
         </Form.Item>
+        <Form.Item label="Chart Visualization">
+          <Checkbox
+            name="hide_plotly_mode_bar"
+            checked={formValues.hide_plotly_mode_bar}
+            onChange={e => this.handleChange("hide_plotly_mode_bar", e.target.checked)}>
+            Hide Plotly mode bar
+          </Checkbox>
+        </Form.Item>
         <Form.Item label="Feature Flags">
           <Checkbox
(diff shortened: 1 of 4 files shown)


arikfr:
I just wonder, does it really make sense to disable it globally for all visualizations? I sort of assumed this will be a per visualization setting.

daniellangnet:
> I just wonder, does it really make sense to disable it globally for all visualizations? I sort of assumed this will be a per visualization setting.

I can only speak for our own situation in which I don’t see any value at all in the Plotly mode bar and I’d rather disable it globally instead of for each chart. I’m also not used to such a thing in other tools like Looker. You may want to ask more people though.

arikfr:
@daniellangnet do you have this already enabled in your deployment? If you do, can you still zoom in without the modebar?

daniellangnet:
@arikfr Yes, we've had this enabled for a while. You can still still zoom in and double click to zoom out like before, even without the mode bar.

arikfr:
Ok, in this case it makes senes :) Might even make it a default eventually. 

@gabrieldutra this adds another thing for you to move to configuration when extracting visualizations 😬

arikfr:
Thanks, @daniellangnet !

daniellangnet:
Great, thank you @arikfr!

Issue #3014, opened 2018-10-28
Search Query Based Parameters


arikfr:
While using a query as input for a dropdown parameter is very useful, sometimes the source data is too large to fit in a single dropdown. A common example might be users list. In many cases you will have thousands (if not more) users.

Today in such cases the user's only option is to revert to having a simple input parameter, where they need to lookup the value (id) in some other way.

To support this better we should introduce a new type of query based parameter: Search Query Based Parameter. This kind of parameter will render a select box with a search field. The search value will be passed to the query as a parameter and the results will be rendered as selection items. The query will be something like:

```sql
select id, name
from users
where name ilike '%{{search}}%'
limit 50; # a sensible limit is still needed
```

(`{{search}}` will be provided from the input to the search box)

arikfr:
In terms of configuring this, one way to go about this is to show new options when you pick a query for "Query Based Dropdown" which has parameters of its own. In this case it will show you a UI to map these parameters to:

1. Search value (for this feature).
2. Static value you can provide.
3. Propagate to parent query (this is basically Use Case 1 from #3010) - out of scope for this issue, but worth keeping in mind.

Static value option has its own complexity, specially when considering backend validation for values. So if it makes implementation more complex, we can skip it as well, and only show a message that saying: "Only single parameter can be used" or something like this. And later implement the new functionality.

PR #4648, opened 2020-02-13
Add options for Query Based Parameter for a query with parameters

gabrieldutra:

- [x] Feature

## Description

To-do
- [x] Start experiments
- [ ] Update backend param verification (so far only working with Query execution by text) - this one can be a bit tricky as verification is done on every execution
- [ ] Allow other options on mapping parameters

#3014

Not yet :)

### DIFF
--- a/client/app/components/QueryBasedParameterInput.jsx
+++ b/client/app/components/QueryBasedParameterInput.jsx
@@ -1,4 +1,4 @@
-import { find, isArray, map, intersection, isEqual } from "lodash";
+import { debounce, find, isArray, get, first, map, trim, intersection, isEqual } from "lodash";
 import React from "react";
 import PropTypes from "prop-types";
 import Select from "antd/lib/select";
@@ -30,6 +30,7 @@ export default class QueryBasedParameterInput extends React.Component {
       options: [],
       value: null,
       loading: false,
+      currentSearchTerm: null,
     };
   }
 
@@ -56,7 +57,7 @@ export default class QueryBasedParameterInput extends React.Component {
       return validValues;
     }
     const found = find(options, option => option.value === this.props.value) !== undefined;
-    value = found ? value : options[0].value;
+    value = found ? value : get(first(options), "value");
     this.setState({ value });
     return value;
   }
@@ -79,13 +80,26 @@ export default class QueryBasedParameterInput extends React.Component {
   }
 
   render() {
-    const { className, value, mode, onSelect, ...otherProps } = this.props;
+    const { parameter, className, value, mode, onSelect, ...otherProps } = this.props;
     const { loading, options } = this.state;
+    const selectProps = { ...otherProps };
+    if (parameter.searchFunction) {
+      selectProps.onSearch = debounce(searchTerm => {
+        if (trim(searchTerm)) {
+          this.setState({ loading: true, currentSearchTerm: searchTerm });
+          parameter.searchFunction(searchTerm).then(values => {
+            if (this.state.currentSearchTerm === searchTerm) {
+              this.setState({ options: values, loading: false });
+            }
+          });
+        }
+      }, 200);
+    }
     return (
       <span>
         <Select
           className={className}
-          disabled={loading || options.length === 0}
+          disabled={!parameter.searchFunction && (loading || options.length === 0)}
           loading={loading}
           mode={mode}
           value={this.state.value}
@@ -95,7 +109,7 @@ export default class QueryBasedParameterInput extends React.Component {
           showSearch
           showArrow
           notFoundContent={null}
-          {...otherProps}>
+          {...selectProps}>
           {options.map(option => (
             <Option value={option.value} key={option.value}>
               {option.name}
--- a/client/app/services/parameters/QueryBasedDropdownParameter.js
+++ b/client/app/services/parameters/QueryBasedDropdownParameter.js
@@ -1,5 +1,6 @@
 import { isNull, isUndefined, isArray, isEmpty, get, map, join, has } from "lodash";
 import { Query } from "@/services/query";
+import QueryResult from "@/services/query-result";
 import Parameter from "./Parameter";
 
 class QueryBasedDropdownParameter extends Parameter {
@@ -7,6 +8,7 @@ class QueryBasedDropdownParameter extends Parameter {
     super(parameter, parentQueryId);
     this.queryId = parameter.queryId;
     this.multiValuesOptions = parameter.multiValuesOptions;
+    this.searchTerm = parameter.searchTerm;
     this.setValue(parameter.value);
   }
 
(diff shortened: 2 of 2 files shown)


gabrieldutra:
(From @arikfr 's comment https://github.com/getredash/redash/issues/3014#issuecomment-584287570)

> Static value option has its own complexity, specially when considering backend validation for values. So if it makes implementation more complex, we can skip it as well, and only show a message that saying: "Only single parameter can be used" or something like this. And later implement the new functionality.

After my first iterations with the backend I wouldn't say the static value has its complexity for validation rules. Since they are configured in the Parameter Edit, the values should be reliable (it's just a matter of computing them backend-wise on the validation).

However, a bigger problem in the same manner is the Search parameter. At a moment I tried to handle the safety of Parameters with the real time Search enabled, but realized it falls back on being a Text Parameter (just with one more layer before it :sweat_smile:). I'm not sure if you already had that in mind before, but I guess those will have to be a new entry on the "Unsafe param list".

Based on this, I got a few options on how to proceed:
1. Make only Dropdowns with the Search feature enabled unsafe: more robust and allow ~not that many~ more cases;
2. Make all Dropdowns based on any query parameters unsafe: this safes us from digging into query results to validate such parameters. I already have the piece of code that evaluates this in this PR from my tests, but I think it's still a pain to have it running on every query execution.

Personally I prefer going with 2 as that logic only made sense to me for the Search feature. Anyway, I'm sharing this in case you had other thoughts on this and to make sure I'll go into the right direction :)

arikfr:
It's an overlook on my part :-( Unfortunately you're right: for the original use case of a search parameter which is text based, we will have to mark those as unsafe.

While option 1 is better, option 2 is also acceptable for first iteration as a text based parameter is the most common case anyway.

redash/models/parameterized_query.py
```diff
@@ -191,11 +191,14 @@ def is_safe(self):
...
-
-                if query.parameters:
-                    return False
+                try:
+                    query_id = param.get("queryId")
+                    query = models.Query.get_by_id_and_org(query_id, self.org)
+
+                    if query.parameters:
+                        return False
+                except (models.NoResultFound):
```
gabrieldutra: Should this be used in other `get_by_id_and_org` uses too?

gabrieldutra:
I guess the implementation here is pretty much done, so I'll brief it.

**UI**
Basically the same as the Dashboard Parameters mapping (this should cover adding the Propagate to parent query option in the future).
query-based-mapping

The Parameter Mapping is a required in the Form.

**Functionality**
I chose my option 2 to go with to have most part of the functionality in the frontend and avoid replicating code on the backend. So, validation is disabled for Parameterized Query based parameters and they are marked as unsafe (otherwise would need to rerun the query on the backend to confirm the dropdown values).

For the parameter value in the URL it was necessary to add the search term since the value is now always coupled to it. I added it as `p_value={searchTerm}--{paramValue}`, lmk if you have better ideas.

**Next steps**
- [ ] Add backend unit tests for this PR changes
- [ ] Add Cypress tests for the UI (perhaps in a following PR)
- [ ] Update messages about unsafe parameters (most of them just mention "text parameters")

redash/models/parameterized_query.py
```diff
@@ -29,6 +30,13 @@ def _load_result(query_id, org):
         raise QueryDetachedFromDataSourceError(query_id)
 
 
+def query_has_parameters(query_id, org):
+    try:
+        query = models.Query.get_by_id_and_org(query_id, org)
+        return bool(query.parameters)
+    except (models.NoResultFound):
```
rauchy: Wouldn't we prefer that this blows up with a 404 in such a case? Returning `False` implies that this query doesn't have parameters, but in-fact, it's a misconfigured parent query. It might make debugging harder.

tests/models/test_parameterized_query.py
```diff
@@ -166,7 +166,8 @@ def test_validates_enum_list_value_parameters(self):
         "redash.models.parameterized_query.dropdown_values",
         return_value=[{"value": "1"}],
     )
-    def test_validation_accepts_integer_values_for_dropdowns(self, _):
+    @patch("redash.models.parameterized_query.query_has_parameters", return_value=False)
```
rauchy: I'm not too happy about stubbing out implementation details such as `query_has_parameters`. Is there a way to setup the data to organically provide this behavior?

redash/models/parameterized_query.py
gabrieldutra: Returning `False` was just the easy option (because it will invalidate any option you provide to the parameter), I actually just wanted to avoid the error happening on the server and check what should be done in the rest of this file (currently a server error happens for misconfigured parent queries).

Rethinking about it, to better cover the misconfigured parameters at this point we can either create a new exception type or reuse `InvalidParameterError` and blow up with a 400. IMO there's probably a more robust way, but showing "Your query has misconfigured parameterers: {}" error on query execution should be helpful enough.

tests/models/test_parameterized_query.py
gabrieldutra: Unfortunately my experience with the backend unit tests is very limited :sweat_smile:, for this one I just cloned the behavior it had for `load_results` since it was returning a "There is no app context". 

Another option I have in mind is getting rid of that method and creating a `get_query_by_id_and_org` (so the stub would be a query, not a boolean)

redash/models/parameterized_query.py
gabrieldutra: I now wonder if it's worth it to add the extra logic to have a new Exception type, considering that #4707 will show no options and add a "No options available" empty state. Also I think any extra work here may end up on rework from #4312

gabrieldutra: Updated it to consider that `NoResultFound` as an invalid parameter, the UI will still show it as required when editing. This should be enough help for identifying a problem with a query based dropdown.

tests/models/test_parameterized_query.py
gabrieldutra: A few months later and I'm a little bit better at this now 🙂. So, turns out the `test_parameterized_query` is based in `TestCase`, thus it doesn't have an App Context, we then can:
1. Make it based in `BaseTestCase`, so that it will have an App Context for us to create an actual query. Downside: will make it slow;
2. Keep it like that, add a `_load_dropdown_query` that we can mock and keep mocking `dropdown_values` and a fast running test;

Any preference in this case?

gabrieldutra: This test was actually harder to set with the organic App Context than I thought, so I'm back with the mocks, but I've updated the `query_has_parameters` to be a mock on the Query object.

gabrieldutra:
I realized that although way it works with making options coupled to a search term is better in terms of validating the options, this will cause UX issues 😕. So I'll take a stab at storing both Value and Label (both as the query default value and in the search url).

jhult:
Looking forward to this.

justinclift:
@gabrieldutra This PR sounds useful.  Should we leave this open?

If you don't have time to look at it though, we can just close it. :smile:

konnectr:
I see this as an important feature )

justinclift:
@konnectr Cool. :smile:

Any interest in trying to finish the PR?

PR #4653, opened 2020-02-18
Avoid timing out when no timeout is set

rauchy:
- [x] Bug Fix

## Description
Following #4626, query execution jobs were set to never time out by default, but our `Worker` implementation would enforce a hard limit if soft limits + grace period (15s) are not handled. In this case, if a time limit is not set, jobs would time out after `-1 + 15 = 14` seconds.

### DIFF
--- a/redash/tasks/worker.py
+++ b/redash/tasks/worker.py
@@ -50,8 +50,11 @@ class HardLimitingWorker(BaseWorker):
         self.log.warning("Job %s has been cancelled.", job.id)
 
     def soft_limit_exceeded(self, job):
-        seconds_under_monitor = (utcnow() - self.monitor_started).seconds
-        return seconds_under_monitor > job.timeout + self.grace_period
+        if job.timeout < 0:
+            return False
+        else:
+            seconds_under_monitor = (utcnow() - self.monitor_started).seconds
+            return seconds_under_monitor > job.timeout + self.grace_period
 
     def enforce_hard_limit(self, job):
         self.log.warning(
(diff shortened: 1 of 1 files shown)


redash/tasks/worker.py
```diff
@@ -50,8 +50,11 @@ def stop_executing_job(self, job):
         self.log.warning("Job %s has been cancelled.", job.id)
 
     def soft_limit_exceeded(self, job):
-        seconds_under_monitor = (utcnow() - self.monitor_started).seconds
-        return seconds_under_monitor > job.timeout + self.grace_period
+        if job.timeout < 0:
```
kravets-levko: Shouldn't it be `<=` ? Or zero timeout has some special meaning?

arikfr: Maybe we should use explicit -1 here? So it's clear what is it about? I would also consider defining a constant for it, so it's clear where this magic number coming from.

rauchy: @kravets-levko only -1 means "run infinitely". While I agree that a 0 doesn't make sense, someone might want to choose to set their Redash instance to time out immediately (🤷🏻‍♂️), so `soft_limit_exceeded` should include those. I agree with @arikfr's approach regarding explicitness.

rauchy: > I would also consider defining a constant for it, so it's clear where this magic number coming from.

Decided to avoid constant since it wouldn't be re-used in settings. Went for a [descriptive variable name](https://github.com/getredash/redash/pull/4653/commits/3e375bb528542e9c68cdce9036c328c71cac8dc0) instead.

PR #4659, opened 2020-02-18
Oracle: update DSN construction to support special characters in user/password.

juanvasquezreyes:
The reason I propose this change is to fix an issue when oracle password has an @
example of connection string: user/p@ssword@host

- [X] Bug Fix

## Description

Currently oracle.py is building a query string, but if the DB password has an @ it doesn't work cause it build a connection string like: user/p@ssword@host

### DIFF
--- a/redash/query_runner/oracle.py
+++ b/redash/query_runner/oracle.py
@@ -63,17 +63,7 @@ class Oracle(BaseSQLQueryRunner):
         return "oracle"
 
     def __init__(self, configuration):
-        super(Oracle, self).__init__(configuration)
-
-        dsn = cx_Oracle.makedsn(
-            self.configuration["host"],
-            self.configuration["port"],
-            service_name=self.configuration["servicename"],
-        )
-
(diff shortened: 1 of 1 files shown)


redash/query_runner/oracle.py
```diff
@@ -130,7 +130,11 @@ def output_handler(cls, cursor, name, default_type, length, precision, scale):
                 )
 
     def run_query(self, query, user):
-        connection = cx_Oracle.connect(self.connection_string)
+        dsn_t = cx_Oracle.makedsn(
+            self.configuration["host"],
+            self.configuration["port"],
+            service_name=self.configuration["servicename"])
+        connection = cx_Oracle.connect(user=self.configuration["user"],password=self.configuration["password"], dsn=dsn_t)
```
arikfr: Thanks, but this is make the creation of `connection_string` in the constructor redundant. Let's remove the dead code there as well.

juanvasquezreyes: Just updated it, thanks

PR #4665, opened 2020-02-20
feat: Add ssl options for Cassandra data source

arihantsurana:
- [x] Feature

## Description
Cassandra is regularly used in production with Encryption enabled, causing the Client to handle SSL options like protocol, and certificate. This PR adds the option to:
* Enable/Disable SSL for the connection
* Provide an SSL Protocol 
* Upload a certificate file

https://discuss.redash.io/t/cassandra-does-not-support-ssl-with-certificates/5584

Changes in the UI for Cassandra connection.
Before:
image

After:
image

image


redash/query_runner/cass.py
```diff
@@ -45,14 +68,43 @@ def configuration_schema(cls):
                     "default": 3,
                 },
                 "timeout": {"type": "number", "title": "Timeout", "default": 10},
+                "useSsl": {"type": "boolean", "title": "Use SSL", "default": False},
+                "sslCertificateFile": {
+                    "type": "string",
+                    "title": "SSL Certificate File"
+                },
+                "sslProtocol": {
+                    "type": "string",
```
arikfr: You can use an enum here. You can see an example in the MongoDB query runner. Although it uses an extended enum type, which allows for setting a title for each option. There is also a simpler type, but I don't have an example for it :|

redash/query_runner/cass.py
```diff
@@ -22,6 +25,26 @@ def default(self, o):
         return super(CassandraJSONEncoder, self).default(o)
 
 
+def generate_cert_file(ssl_cert, host):
+    cert_dir = f"./tmp/cassandra_certificate/{host}"
```
arikfr: I think it will be safer to use a random file name + clean up after use. Maybe use [`NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)?

arihantsurana: ah cool, will use that.

redash/query_runner/cass.py
arihantsurana: Yeah, I wanted to find if there were enum like option for the configuartion but didn't know exactly which library or config parser you are using, and didn't dig deeper. Will look at the Mongo Example and set it up.

arihantsurana: I had trouble using the extended Enum, Throws errors when saving the form. I tried the MongoDb connection form, and it has the same issue. So I am using plain enum and it seems to work.

arihantsurana:
hey @arikfr , did you get a chance to review the cahnges?

arihantsurana:
Hey @arikfr ... just wanted to drop in and see if you had any timelines for this feature to be released?

 I have seen other PRs on this repo being rejected due to underlying refactor, do you think this PR might suffer the same fate?

Also, Is there some documentation around how you publish the docker images, so I can try and have an interim docker Image unit the next release?

redash/query_runner/cass.py
arikfr: Sorry about this, the issue with extendedEnum is fixed in https://github.com/getredash/redash/pull/4774.

arikfr:
Thanks, @arihantsurana ! This is now merged. When master branch builds it builds two images:

1. `redash/redash:preview` this tag follows latest master.
2. `redash/preview:$VERSION.b$CIRCLE_BUILD_NUM` - this is a versioned image that is specific to a build.

This is the build for master with this PR merged:
https://circleci.com/workflow-run/444e47d2-f0aa-4a67-95c5-d1ff87765bb0

arihantsurana:
Awesome this is great!!

SORC3r3r:
@arihantsurana @arikfr 
I am testing that beta release.
Do you have a hint how to format the certificate? 🤔 
Plain `.pem` or `.pfx`?
How do I order the leaf and the chain in the certificate?

arihantsurana:
> @arihantsurana @arikfr
> I am testing that beta release.
> Do you have a hint how to format the certificate? 🤔
> Plain `.pem` or `.pfx`?
> How do I order the leaf and the chain in the certificate?

IIRC I tested this with a .pem file.
The redash backend code uses the DataStax python driver for Cassandra, so you should be able to upload any Cert that fits the spec on that library. https://docs.datastax.com/en/developer/python-driver/3.20/security/#ssl

PR #4668, opened 2020-02-20
Fix Visualization scrollbar in query editor

gabrieldutra:

- [x] Bug Fix

## Description
Noticed it for [this visualization](https://redash-preview.netlify.com/queries/278/source#509), in a mobile resolution. The scrollbar does not exist:

image

So this PR removes an `overflow: visible` that was applied only for the Query Editor. since it didn't happen on the Query View page, it should be safe to have the same prop value as in there.

--

image

### DIFF
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -245,11 +245,6 @@ a.label-tag {
     align-content: space-around;
     padding: 0;
     overflow-x: hidden;
-
-    .pivot-table-visualization-container > table,
-    .visualization-renderer > .visualization-renderer-wrapper {
-      overflow: visible;
-    }
   }
   .row {
     background: #fff;


client/app/assets/less/redash/query.less
```diff
@@ -245,11 +245,6 @@ a.label-tag {
     align-content: space-around;
     padding: 0;
     overflow-x: hidden;
-
-    .pivot-table-visualization-container > table,
-    .visualization-renderer > .visualization-renderer-wrapper {
-      overflow: visible;
```
kravets-levko: How about rule for `.pivot-table-visualization-container`? Was there any issue as well, or this rule no longer needed?

gabrieldutra: For pivot I think this rule was outdated since its migration to React. I quickly tested it and didn't see any issues with it.

Issue #4663, opened 2020-02-19
Query View design follow ups


arikfr:
Follow ups for #4536:

- [x] The Data Source and Refresh Schedule fields are unaligned. Data Source has height of 30px (because of icon) while Refresh Schedule has height of 28px. 
- [x] The execution status is closer to the lower border than the upper one, need to center it.
- [x] On Mobile we need to make the header more like the design: while we decided to drop the "Add Tag"/"Add Description" buttons, we still should move the other action buttons to a separate row + have the tags on a separate row (if they exist).
- [ ] Update text of empty state @arikfr 

![image](https://user-images.githubusercontent.com/71468/74839845-a6303680-532e-11ea-92c4-34a6fd8c2c9b.png)

PR #4670, opened 2020-02-20
Query View page design adjustments

gabrieldutra:

- [x] Other

## Description
Addresses comments from #4663 

[Will write a detailed explanation of what I did while netlify runs]

#4663 

Soon

### DIFF
--- a/client/app/assets/less/inc/base.less
+++ b/client/app/assets/less/inc/base.less
@@ -244,10 +244,6 @@ text.slicetext {
   display: flex;
   align-items: center;
 
-  h3 {
-    margin-right: 5px !important;
-  }
-
   .label {
     margin-top: 3px;
     display: inline-block;
@@ -255,7 +251,7 @@ text.slicetext {
 
   .favorites-control {
     font-size: 19px;
-    margin-right: 5px;
+    margin-right: 10px;
   }
 }
 
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -416,12 +416,10 @@ nav .rg-bottom {
 .query-tags {
   display: inline-block;
   vertical-align: middle;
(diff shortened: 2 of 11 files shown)


client/app/assets/less/inc/base.less
```diff
@@ -244,18 +244,14 @@ text.slicetext {
   display: flex;
   align-items: center;
 
-  h3 {
-    margin-right: 5px !important;
```
gabrieldutra: This was more a DashboardPage thing, so moved to it.

client/app/assets/less/inc/base.less
```diff
@@ -244,18 +244,14 @@ text.slicetext {
...
-
   .label {
     margin-top: 3px;
     display: inline-block;
   }
 
   .favorites-control {
     font-size: 19px;
-    margin-right: 5px;
+    margin-right: 10px;
```
gabrieldutra: This also affects the DashboardPage now, we added those extra 5px on the Query Pages now, I thought it should also be in there for consistence, LMK if you think it differently.

arikfr:
One issue I noticed on mobile: charts have some min height which is probably too small:

Screen Shot 2020-02-24 at 11 33 21 AM

Another example:

image

Although to be honest, I'm not sure what the right solution here. Can min-height be calculated?

gabrieldutra:
@arikfr I think it can be an unrelated bug with Chart as it seems it happened in Angular too:
chart-height-bug
(Showing both React and Angular versions)

Perhaps it's calculating its size too early? (notice that slightly changing the width makes it better fit the space)

@kravets-levko, do you know ath related?

kravets-levko:
I didn't see this issue before. Looks like it's calculating it's size too early, but I have no idea why :thinking:

PR #4629, opened 2020-02-23
Job timeout doesn't kill the mysql query

rauchy:
I'm using the master branch. During refreshing one of my dashboards, a **mysql query** reached the timeout which raised the "**query execution time limit**".
I thought the timeout handler would also kill the related sql query in mysql server while it didn't. So there are lots of stuck sql queries in the mysql server.

So is that possible to also kill the related running query in the database server during handling the timeout?

Thanks.


rauchy:
@arikfr WDYT about this? We'll probably need to attach to this signal to other multi-threaded runners.

PR #4681, opened 2020-02-24
Perform cleanup on job timeouts

rauchy:
- [x] Refactor
- [x] Bug Fix

## Description
After moving query executions to RQ, job timeouts resulted in `JobTimeoutException` being raised. Since some query runners perform cleanup when executions are being cancelled by users, the same cleanup should be performed when jobs time out. (i.e. cancel the database query).

While going through the different query runners, I've noticed that query cancellation code is very repetitive, and sometimes just unnecessary and only returns a "Query cancelled by user." error message. I've modified these to raise the exception and have the serializer handle that automatically.

One caveat is that previously, when cancelling a query, the user would see "Query cancelled by user", and now it will also include an "Error running query" prefix.

image

This is due to me reusing the error status code for the serialized job, we could easily work around this by introducing a "cancelled" status call ("failed" doesn't feel right).

#4307 #4629

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -5,7 +5,7 @@ import requests
 
 from redash import settings
 from redash.utils import json_loads
-from redash.tasks.worker import JobTimeoutException
+from rq.timeouts import JobTimeoutException
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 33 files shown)


arikfr:
> One caveat is that previously, when cancelling a query, the user would see "Query cancelled by user", and now it will also include an "Error running query" prefix.

How come there was no prefix before? I thought we were always reporting this as an error?

rauchy:
My mistake, it did show with the prefix before.

PR #4682, opened 2020-02-25
Fixed Clickhouse column name encoding problem

### LINKED ISSUE #4677: Clickhouse column name encoding problem
### Issue Summary

Alias column name contains non 'latin-1' characters return encoding  error.
For example : select count(*) as 'כמות'…

**Error message**: 'latin-1' codec can't encode characters in position 285-288: Body ('כמות') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.

BTW, This works fine with other data source like mysql.

### Steps to Reproduce

1. This is the first step
Create a query with ClickHouse as the data source

2. This is the second step, etc.
Add a column alias in utf8 chars like : select colum1 as 'ש'

I expected to see the column alias like I do with mysql data source .
No problem with the 'utf-8' data,  so probably column names should support 'utf-8' charset as well.
This used to work with the older version (v8)
### Technical details:

* Redash Version:
9.0.0-alpha
* Browser/OS:
 Chrome
* How did you install Redash:
Docker Based Developer Installation

erels:

- [x] Bug Fix

## Description
Fix column name error from Clickhouse when alias column name contains non 'latin-1' characters.

Fixes #4677.

### DIFF
--- a/redash/query_runner/clickhouse.py
+++ b/redash/query_runner/clickhouse.py
@@ -68,7 +68,7 @@ class ClickHouse(BaseSQLQueryRunner):
             verify = self.configuration.get("verify", True)
             r = requests.post(
                 url,
-                data=data,
+                data=data.encode("utf-8","ignore"),
                 stream=stream,
                 timeout=self.configuration.get("timeout", 30),
                 params={


arikfr:
Thanks!

PR #4686, opened 2020-02-26
Visualization Editor: consistent styles for control labels

kravets-levko:

- [x] Refactor

Before/After (checkbox labels):

image

image

### DIFF
--- a/client/app/components/visualizations/editor/control-label.less
+++ b/client/app/components/visualizations/editor/control-label.less
@@ -1,4 +1,8 @@
 .visualization-editor-control-label {
+  label {
+    font-weight: 500;
+  }
+
   &.visualization-editor-control-label-horizontal {
     label {
       margin-bottom: 0;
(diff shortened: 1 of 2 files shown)


gabrieldutra:
I actually prefer it with `font-weight: normal` :sweat_smile:. It's been changed since [this](https://github.com/getredash/redash/pull/3296#discussion_r248618992). Anyway, I think it makes sense to be a change to the whole application (not only to the Visualization Editor), regardless of the choice.

kravets-levko:
Hm :thinking: For me it looks wrong if some labels have normal font weight and others are bolder (especially when there are both types on the same form). So what you suggest? Make all labels bolder? normal? keep everything as is?

arikfr:
The mental model I have: the non checkbox labels are more like titles while the label for a checkbox is more like regular text. Hence why one is bold and the other isn't.

I vote for keeping as is.

arikfr:
As we don't have a consensus here I'm closing it.

PR #4690, opened 2020-02-27
Cypress: Separate start command and update to v4.1.0

gabrieldutra:

- [x] Other

## Description
This separates `npm run cypress build` and `npm run cypress start` (also automatically seeds database in here as this is usually done anyway, skipable with `--skip-db-seed`).

--

--


kravets-levko:
Just an idea: `cypress:start` does some extra job, so it makes sense to keep it; but looks that other commands are just wrappers for `docker-compose` with specified project name (`-p cypress`). So what if there'll be just a single command that will run `docker-compose -p cypress ${args}`? Use case: for me it's quite annoying that Cypress container is re-built every time I `cypress:start` it (it's because `cypress:stop` destroys container). With such command I'll be able to choose the action - `stop` or `down`. WDYT?

gabrieldutra:
I don't know :sweat_smile:, then it starts to be just easier to run the `docker-compose -p cypress`. The reason to have a `down` command instead was mostly to clean test data afterwards and have a quick set of commands within this purpose. Also the time consuming part of the `start` command is in the db creation and data seed (the container creation seems quite fast).

gabrieldutra:
@kravets-levko I also updated Cypress to v4.1.0 (didn't need to change ath, just the version on package.json). Do you have any other quick wins in mind or let's merge it?

PR #4696, opened 2020-03-01
Snowflake: use different method of showing columns if no schema specified in db name

arikfr:

- [x] Bug Fix

## Description

The change done in #4634 breaks fetching schema when database name doesn't include schema name. This one fixes it.

### DIFF
--- a/redash/query_runner/snowflake.py
+++ b/redash/query_runner/snowflake.py
@@ -129,7 +129,7 @@ class Snowflake(BaseQueryRunner):
         return '.' in self.configuration.get('database')
 
     def get_schema(self, get_stats=False):
-        if self._database_name_with_schema():
+        if self._database_name_includes_schema():
             query = "SHOW COLUMNS"
         else:
             query = "SHOW COLUMNS IN DATABASE"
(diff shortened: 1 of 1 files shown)


redash/query_runner/snowflake.py
```diff
@@ -123,10 +123,18 @@ def _run_query_without_warehouse(self, query):
             cursor.close()
             connection.close()
 
-        return data, error
+        return data, error    
+    
+    def _database_name_with_schema(self):
```
rauchy: The name of this function sounds like it returns the database name with schema. How about

```suggestion
    def _database_name_includes_schema(self):
```

redash/query_runner/snowflake.py
```diff
@@ -123,10 +123,18 @@ def _run_query_without_warehouse(self, query):
...
+        return '.' in self.configuration.get('database')
 
     def get_schema(self, get_stats=False):
-        results, error = self._run_query_without_warehouse("SHOW COLUMNS")
+        if self._database_name_with_schema():
```
arikfr:
```suggestion
        if self._database_name_includes_schema():
```

redash/query_runner/snowflake.py
arikfr: 👍

PR #4701, opened 2020-03-01
Cancel BigQuery Queries

rauchy:
- [x] Bug Fix

## Description
When a user requests to cancel a BigQuery query, or when the Redash time limit is reached, the underlying job in BigQuery continues to run until completion. This PR cancels the underlying BigQuery job in these situations.

#4681

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -192,12 +192,13 @@ class BigQuery(BaseQueryRunner):
         project_id = self._get_project_id()
         job_data = self._get_job_data(query)
         insert_response = jobs.insert(projectId=project_id, body=job_data).execute()
+        self.current_job_id = insert_response["jobReference"]["jobId"]
         current_row = 0
         query_reply = _get_query_results(
             jobs,
             project_id=project_id,
             location=self._get_location(),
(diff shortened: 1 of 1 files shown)


redash/query_runner/big_query.py
```diff
@@ -332,6 +333,18 @@ def run_query(self, query, user):
                 error = json_loads(e.content)["error"]["message"]
             else:
                 error = e.content
+        except (KeyboardInterrupt, InterruptException, JobTimeoutException):
+            if self.current_job_id:
+                response_kind = None
+                while response_kind != "bigquery#jobCancelResponse":
+                    response_kind = jobs.cancel(
```
arikfr: We need to repeatedly execute the cancel request to get its status? No way to just get status after the first request?

rauchy: This bothered me as well. Turns out the client is executing previous requests when calling `execute()` in the cancellation block (the result kind is a `getQueryResultResponse` for a call to `/cancel` 🤨), so I'm guessing it's either some sort of optimization or a bug. Tried drilling into their source code to see if this is intentional, and if there's a flush method of some sort. Eventually gave up and realized it'll be easier to instantiate a new client. See 893c427.

PR #4706, opened 2020-03-03
Fix: show size of actually used Redash database

citrin:

- [x] Bug Fix

## Description

System status dashboard currently shows a size of `postgres` database, which is a default database name, but it can be changed in Redash settings. If non-default database name is used dashboard shows a wrong size. This change fixes the size shown in a dashboard for non-default database name by using `current_database()` instead of hard-coded database name.


arikfr:
Thanks :)

PR #4707, opened 2020-03-03
Fix query based param with no results crashing page

### LINKED ISSUE #4703: A query with a query based dropdown list parameter with empty results could cause page error
### Steps to Reproduce

1. Create a query with parameter which is a query based drop down list.
2. When the parameter returns empty results, the query page shows "It seems like we encountered an error. Try refreshing this page or contact your administrator".
image
3. When the parameter results is not empty, the query page works fine.

The same query worked fine in version 8.0.

There is an example in the deploy preview: https://redash-preview.netlify.com/queries/343/source

### Technical details:

* Redash Version: 9.0 alpha
* Browser/OS: Cent OS 7

gabrieldutra:

- [x] Bug Fix

## Description
Change `QueryBasedParameter` to use lodash functions and avoid crashing the Query Page.

Fixes #4703 

--


gabrieldutra:
WDYT about adding a `notFoundContent` with a message like "No results found" instead of disabling the parameter?

myonlylonely:
Is this bug fixed in docker image redash/redash:preview?

arikfr:
> WDYT about adding a `notFoundContent` with a message like "No results found" instead of disabling the parameter?

Sounds right. Let's use "No options available".

arikfr:
> Is this bug fixed in docker image redash/redash:preview?

No. `redash/redash:preview` is based on the master branch.

gabrieldutra:
> Sounds right. Let's use "No options available".

Updated for both Dropdown and Query Based Dropdown :+1:

gabrieldutra:
Updates:
- The 500 errors when a query _has no results_ is now handled to return an empty set
- Added a Cypress test to make sure it opens the page (not crashing) and shows that message when you click on the Parameter

Issue #4708, opened 2020-03-04
Saved query schedule migration not working from v4 to v8


satyamkrishna:
### Issue Summary

Saved schedule of the queries doesn't get migrated properly from v4 to v8 due to an issue in alembic migration. Hence after migration, the queries don't get added to refresh schedule

### Steps to Reproduce

1. Run the redash using v4.0.2 tag in docker
2. Save queries to any data source with some refresh schedule under a day
3. Upgrade the redash version to any latest version like v8 and run the migrations
4. Check the save queries in the newer version. The schedule dialog doesn't show the right information and shows the string instead
5. The queries fail to be called in their schedule time

Here are some screenshots of the two queries ( one with time and one with hours ) saved from v4 and migrated to v8

![v4_redash_db](https://user-images.githubusercontent.com/3524351/75899234-46bf4400-5e61-11ea-9566-e28bcc07a299.png)

v4 migrated to v8 

![v8_redash_db](https://user-images.githubusercontent.com/3524351/75899351-6c4c4d80-5e61-11ea-8d95-cd7672ac972a.png)

As you can see the migrated schedule is stored as string in the interval field. This causes all the problems. The UI also malfunctions because of this

![v8_redash_ui](https://user-images.githubusercontent.com/3524351/75899449-8b4adf80-5e61-11ea-9ca4-25ec13b482b6.png)

Manually going to every query and saving it solves it as it overrides the pervious schedule

### Technical details:

* Redash Version: Initial Version : 4.0.2, Migrated Version v8.0.0
* Browser/OS: Chrome/Mac OS
* How did you install Redash: Docker

PR #4709, opened 2020-03-04
Fix : Alembic migration for scheduled query from older to newer version

satyamkrishna:

- [x] Bug Fix

## Description

Fixes the alembic migration of query schedules, so that older versions like v4 can also migrate to newer version

Fixes [#4708](https://github.com/getredash/redash/issues/4708)


arikfr:
Thank you for reporting this and for submitting a fix. I'm not sure how this wasn't detected before 🤔

Issue #4617, opened 2020-02-07
New version of Werkzeug breaks Flask, and then Redash


DavidHernandez:
### Issue Summary

Last night, Werkzeug released a new major version (from 0.16.1 to 1.0.0). Flask has declared this dependency as `Werkzeug>=0.15`. When building a new instance of the project, Flask downloads the last version of Werkzeug that breaks it's integration. So, Redash get's broken too. The use of Werkzeug that Redash does should be updated too.

### Steps to Reproduce

1. Just build the project: `docker-compose up --build`

### Technical details:

* Redash Version: probably most versions of Redash
* Issue on Flask: https://github.com/pallets/flask/issues/3486

PR #4711, opened 2020-03-05
Cleaner pytest results

rauchy:
- [x] Refactor

## Description
Currently there's a lot of noise when running backend tests (which includes some deprecation warnings introduced by 3rd party packages, and the default pytest result renderer, which is basic)

This PR silences deprecations and includes [pytest-sugar](https://github.com/Teemu/pytest-sugar).

image


arikfr:
I'm not sure about this one. While the warnings generate a lot of noise, we should consider if it will be feasible to address them instead of suppressing.

If we addressed some of them in time, we could've avoided the Werkzeug upgrade issue _(#4617)._

rauchy:
@arikfr solid point, but are unit test executions really the place to raise these? The constant visual noise around these deprecation warnings isn't really pushing for developer happiness.

arikfr:
Most of these warnings are repeating themselves. Any idea if there is a way to get an aggregated report about number and types of warnings raised during tests?

rauchy:
Not sure, but we can hide third-party depreciation warnings in dev (while keeping direct deprecations) and only show them in CI. WDYT?

arikfr:
If it's only in CI we can just disable it. We never look at the output there :-) That's why I thought of collecting aggregated numbers that we can show somewhere, so it's more on our mind.

justinclift:
@guidopetri I've just fixed the merge conflict that was showing up, which might make this a bit easier for you.  Note that I just used the same version of `pytest-sugar` (so `0.9.2`) that was in the original PR, even though the other pytest related stuff uses newer versions.  It might be better to update the pytest-sugar package to a newer release as well (no idea personally). :smile:

guidopetri:
I think we can set the warnings filter to `once` instead of `ignore` ([ref](https://docs.python.org/3/library/warnings.html#the-warnings-filter)). As for the `pytest-sugar` library, it seems recently revived and at `0.9.7` so we may as well update if we're going to add it in. I don't particularly care for it personally.

Let me try making those changes.

guidopetri:
Wow, that's a lot of warnings. I'm going to try running this locally a few times to compare what it looks like with/without the `once` setting and with/without pytest-sugar, but I'll have to do this later.

guidopetri:
Ok, so I'm not convinced that `pytest-sugar` makes any real difference (plus it's an additional dep), and switching from `ignore` -> `once` reduced the number of lines but not really by much. I'll remove the deps change and merge in the `once` change but I don't feel strongly either way.

guidopetri (approved):

guidopetri:
I think this isn't passing codecov because it's 39 commits behind master, since it doesn't make sense for this change to change codecov. So once the e2e tests pass I'll merge in.

PR #4712, opened 2020-03-06
Datasource Exasol: support encryption setting

stefan-mees:
- [x] Other

## Description
This adds support to toggle SSL Encryption for connections to Exasol
See also https://github.com/badoo/pyexasol/blob/master/docs/ENCRYPTION.md

### DIFF
--- a/redash/query_runner/exasol.py
+++ b/redash/query_runner/exasol.py
@@ -66,9 +66,10 @@ class Exasol(BaseQueryRunner):
                 "password": {"type": "string"},
                 "host": {"type": "string"},
                 "port": {"type": "number", "default": 8563},
+                "encrypted": {"type": "boolean", "title": "Enable SSL Encryption"},
             },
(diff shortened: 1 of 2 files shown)


redash/query_runner/exasol.py
```diff
@@ -81,6 +82,7 @@ def _get_connection(self):
             dsn=exahost,
             user=self.configuration.get("user", None),
             password=self.configuration.get("password", None),
+            encryption=self.configuration.get("encrypted", True),
```
arikfr: While having SSL enabled is a good default, but this will change it for existing deployments as well. How common would it be for an Exasol deployment not to have SSL support?

stefan-mees: SSL support is build in into the Exasol Engine, so all deployments which support WebSockets also have SSL capabilities

arikfr:
👍 thanks.

PR #4714, opened 2020-03-06
Improvements to Dashboard performance

gabrieldutra:

- [x] Refactor

## Description
Dashboard experiments to improve performance.

So far the ideas I had to work on:
- [x] Separate widgets state from dashboard state
Expected: this should reduce the amount of recalculated things when Widgets are loaded on their own.
Result from tests: it reduced the number of renders on dashboard load, but only for dashboards that contained parameters;
- [x] Refactor DashboardGrid into a functional component and take advantage of `useMemo` (as well as other React hooks), as described in [ReactGridLayout readme](https://github.com/STRML/react-grid-layout#performance)
Expected: improve ReactGridLayout performance and make sure widgets
Result from tests: I feel dashboard page a bit faster, but still need to make sure widgets are updated when they should (since now there's a new memoized component).
- [ ] Introduce `react-virtualized` in DashboardGrid
Expected: Improve performance for dashboards containing several widgets by just rendering what's in the screen
 
A different approach for the first that's also based on the same idea of improving widgets loading was to make Widgets totally independent on their state. I quickly tested it and for the React Migration dashboard it went from 255 renders to 21 renders. However, that didn't quite improved on its freezing times, so I didn't want to continue with it because it seemed a not very React way of doing things (dashboard would need to call widgets loading from their reference in page). I did conclude from this that the number of renders is not the biggest issue in there.

Possible painful functions I noticed (still need further investigation):
- Table with many rows (in the React migration dashboard there was one with 100 rows)
- Plotly applyLayoutFixes
- Word cloud renderer
- resizeObserver

--

--

### DIFF
--- a/client/app/components/ApplicationArea/ApplicationHeader/index.jsx
+++ b/client/app/components/ApplicationArea/ApplicationHeader/index.jsx
@@ -28,7 +28,7 @@ function onSearch(q) {
 
 function DesktopNavbar() {
   const showCreateDashboardDialog = useCallback(() => {
-    CreateDashboardDialog.showModal().result.catch(() => {}); // ignore dismiss
+    CreateDashboardDialog.showModal();
   }, []);
 
   return (
--- a/client/app/components/ApplicationArea/Router.jsx
+++ b/client/app/components/ApplicationArea/Router.jsx
@@ -1,4 +1,4 @@
-import { isFunction, map, fromPairs, extend, startsWith, trimStart, trimEnd } from "lodash";
+import { isFunction, startsWith, trimStart, trimEnd } from "lodash";
 import React, { useState, useEffect, useRef } from "react";
 import PropTypes from "prop-types";
 import UniversalRouter from "universal-router";
@@ -28,18 +28,6 @@ export function stripBase(href) {
   }
 
   return false;
-}
-
-function resolveRouteDependencies(route) {
-  return Promise.all(
-    map(route.resolve, (value, key) => {
-      value = isFunction(value) ? value(route.routeParams, route, location) : value;
-      return Promise.resolve(value).then(result => [key, result]);
-    })
-  ).then(results => {
-    route.routeParams = extend(route.routeParams, fromPairs(results));
-    return route;
-  });
 }
 
 export default function Router({ routes, onRouteChange }) {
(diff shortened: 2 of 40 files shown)


arikfr:
>  I feel dashboard page a bit faster

Let's take a step back and put some effort in implementing proper benchmarking for this so we can:

1. See the improvement.
2. Make sure we don't introduce a degradation in the future.

Beside just doing performance benchmark and try to detect hotspots, we should probably try to collect KPIs on the dashboard page like:

1. Time to become interactive.
2. Number of times (and length) of the UI becoming unresponsive.

(the second might be including the first)

Wdyt?

gabrieldutra:
Sure, initially I was aiming for the number of renders as an initial metric, but realized it wasn't that representative. I'll see what timing information can be easily extracted as a benchmark (probably in the lines of the ones you mentioned) and follow up here.

arikfr:
If nothing obvious comes to mind, then we can start by just benchmarking and identifying hotspots and then add timing to those to see if they are reduced (in count or number).

gabrieldutra:
> If nothing obvious comes to mind, then we can start by just benchmarking and identifying hotspots and then add timing to those to see if they are reduced (in count or number).

I'm currently exploring Chrome's profiler Scripting Time and noting a few results. I'll share a table with the comparing times once I finish some changes in this PR from those, but so far I realized the `useMemo` for the DashboardGrid is actually adding scripting time (+3s for the "React Migration Dashboard" and +0.3s to the "Sales Dashboard"), I'm checking if this can be due to over-updating the memoized component.

Also removing the state change on the `loadWidget` function improved timing substantially (~60% less), so I'll see how to actually incorporate this on code.

gabrieldutra:
@arikfr I decided to leave this PR for general experiments/discussion and create smaller PRs (starting from #4724) in case you want to deliver the most benefiting ones in v9 as it makes review easier.

gabrieldutra:
To give a briefing on my last results: The PRs #4724 and #4726 are quite simple and I think they are worth it to include in `v9`. Since #4734 memoizes the widgets, it may be good to pass through more testing time (although it seemed functional at first sight).

The last thing I was thinking to improve large dashboards performance was to explore windowing. I didn't find a way to fit both react-window or react-virtualized with the grid library, so I tried to just don't render visualizations when they weren't on the viewport (8e61383357f8d801979be68c50ff3bb07ec62cb3). While that did reduce a lot the loading time for large dashboards, it didn't create a good scrolling experience as visualizations take a considerable time to load.

I compared the`v7` timing (with an Angular Grid) with the results with all 3 PRs and it was very close (with a slight improvement now). I think for the majority of dashboards the performance should be satisfactory now. Further improvements would probably take place on specific visualizations and on time consuming functions.

PR #4718, opened 2020-03-09
Separate locking concerns out of `enqueue_query`

rauchy:
- [x] Refactor

## Description
`enqueue_query` always seemed too complex to me. The thing that bothered me the most was the intermingled locking-related logic. In this PR I suggest to extract it to its own method. Initially I tried using a context manager for this (i.e. `with single_execution(query): ...`) but the overall result was fugly.


guidopetri:

PR #4726, opened 2020-03-11
Dashboard Performance: HtmlContent improvements

gabrieldutra:

- [x] Other

## Description
Second PR for #4714, basically all widgets had this component rendered unnecessarily (also it could be memoized as its props should always return the same component). Although the `sanitize` function is kinda of performant, I believe it scales with the number of renders and widgets.

**Individual improvement (based on master)**
| Dashboard | master | This PR | % of improvement |
| Sales Dashboard | 3599 ms | 2859 ms |  20% |
| Dashboard for testing React migration | 42505 ms | 37854 ms | 10% | 

_(The performance analysis were run with the same time for each dashboard twice (with the least value selected) to make sure results more reliable.)_

The Scripting time I'm noting actually varies considerably, but when there's an actual improvement, the minimum value you can get diminishes, so it's been the best benchmark I've found so far. It just makes it a bit more difficult to measure dashboards with low avg scripting time (< 3s).

#4714 

--

### DIFF
--- a/client/app/components/HtmlContent.jsx
+++ b/client/app/components/HtmlContent.jsx
@@ -2,14 +2,14 @@ import React from "react";
 import PropTypes from "prop-types";
 import { sanitize } from "dompurify";
 
-export default function HtmlContent({ children, ...props }) {
+const HtmlContent = React.memo(function HtmlContent({ children, ...props }) {
   return (
     <div
       {...props}
       dangerouslySetInnerHTML={{ __html: sanitize(children) }} // eslint-disable-line react/no-danger
     />
   );
-}
+});
 
 HtmlContent.propTypes = {
   children: PropTypes.string,
@@ -18,3 +18,5 @@ HtmlContent.propTypes = {
 HtmlContent.defaultProps = {
   children: "",
 };
+
+export default HtmlContent;
--- a/client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
+++ b/client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
@@ -94,9 +94,11 @@ function VisualizationWidgetHeader({ widget, refreshStartedAt, parameters, onPar
           <p>
             <QueryLink query={widget.getQuery()} visualization={widget.visualization} readOnly={!canViewQuery} />
           </p>
-          <HtmlContent className="text-muted markdown query--description">
-            {markdown.toHTML(widget.getQuery().description || "")}
-          </HtmlContent>
+          {!isEmpty(widget.getQuery().description) && (
+            <HtmlContent className="text-muted markdown query--description">
+              {markdown.toHTML(widget.getQuery().description || "")}
+            </HtmlContent>
+          )}
         </div>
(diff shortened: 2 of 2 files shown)


client/app/components/HtmlContent.jsx
```diff
@@ -2,14 +2,14 @@ import React from "react";
...
-export default function HtmlContent({ children, ...props }) {
+const HtmlContent = React.memo(function HtmlContent({ children, ...props }) {
   return (
     <div
       {...props}
       dangerouslySetInnerHTML={{ __html: sanitize(children) }} // eslint-disable-line react/no-danger
     />
   );
-}
+});
```
kravets-levko: Have you tried to use `useMemo` instead of `React.memo`? `React.memo` changes behavior of component (due to shallow comparison of props, component _may_ not properly react to props changes):

```jsx
const html = useMemo(() => ({ __html: sanitize(children) }), [children]);

return <div {...props} dangerouslySetInnerHTML={html} />
```

client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
```diff
@@ -94,9 +94,11 @@ function VisualizationWidgetHeader({ widget, refreshStartedAt, parameters, onPar
           <p>
             <QueryLink query={widget.getQuery()} visualization={widget.visualization} readOnly={!canViewQuery} />
           </p>
-          <HtmlContent className="text-muted markdown query--description">
-            {markdown.toHTML(widget.getQuery().description || "")}
-          </HtmlContent>
+          {!isEmpty(widget.getQuery().description) && (
+            <HtmlContent className="text-muted markdown query--description">
+              {markdown.toHTML(widget.getQuery().description || "")}
```
kravets-levko: How about `markdown.toHTML` performance? Is it fast enough, or maybe also needs some caching?

client/app/components/HtmlContent.jsx
gabrieldutra: Yes, `useMemo` should have the same performance in concept, but not re-rendering the component at all looked better. For the `HtmlContent` component it should be totally dependent on its props. I quickly checked its usages, it has only an string for className (basic variable types don't create issues with shallow comparison) and the `children` node didn't seem a problem as well.

client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
gabrieldutra: I didn't cache it with `useMemo` because it runs `widget.getQuery()` currently, it would need to have the `widget` as a dependency. Since that component is basically re-rendered when `widget`changes, that didn't seem that relevant :thinking:. Also now it only runs that function when the query has a description.

client/app/components/HtmlContent.jsx
kravets-levko: Thanks for explanation! While I personally prefer `useMemo`, `React.memo` looks good as well :+1:

Issue #4725, opened 2020-03-11
Python query runner cannot use "print" function in 9.0.0-alpha


bryanyang0528:
### Issue Summary

Python version: 3.7.6
Redash version: 9.0.0-a;pha
Python query runner cannot use "print" function in 9.0.0-alpha

### Steps to Reproduce

1. export REDASH_ENABLED_QUERY_RUNNERS=redash.query_runner.python
2. Use python query runner
3. print('x') and execute it

<img width="1079" alt="Screen Shot 2020-03-11 at 12 57 18" src="https://user-images.githubusercontent.com/5171607/76383771-e156e000-6397-11ea-8406-413a579f8cc4.png">

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

* In previous version, if I call `print`, it will be put in the `log` section.

<img width="1356" alt="Screen Shot 2020-03-11 at 12 54 21" src="https://user-images.githubusercontent.com/5171607/76383654-7907fe80-6397-11ea-8fd6-e847091985b1.png">

* But in the current version, `print` will not work
<img width="1440" alt="Screen Shot 2020-03-11 at 12 56 07" src="https://user-images.githubusercontent.com/5171607/76383728-b2406e80-6397-11ea-8506-ee5bfe6a845e.png">

 

### Technical details:

* Redash Version:  9.0.0-a;pha
* Browser/OS: default Docker 
* How did you install Redash: Docker

sudarshanbasumatary:
Exactly the same issue when self-hosting.

We fixed it by updating the CustomPrint class in Python query runner.
Created a pull request for the same: https://github.com/getredash/redash/pull/4730

PR #4730, opened 2020-03-12
Python query runner Print exception fix

sudarshanbasumatary:

- [x] Bug Fix

## Description

Fixes exception raised when using print statement in queries for Python data source [9.0.0-alpha (dev)]

Fixes getredash/redash#4725

image

### DIFF
--- a/redash/query_runner/snowflake.py
+++ b/redash/query_runner/snowflake.py
@@ -6,7 +6,7 @@ except ImportError:
     enabled = False
 
 
-from redash.query_runner import BaseQueryRunner, register
+from redash.query_runner import BaseSQLQueryRunner, register
 from redash.query_runner import (
     TYPE_STRING,
     TYPE_DATE,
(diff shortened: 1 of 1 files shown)


bryanyang0528:
@sudarshanbasumatary  I've tried your commit, but found `_call_print` may not work because it won't put the output of `print` command in the `log` in the return dict.

I recommend that the `text` of _call_print should be called by the print method itself  like this:
```
def _call_print(self, text):
    self.print(text)
```
Then `run_query` will work perfect and return the correct result like this:

```
{'result': {'rows': [], 'columns': [], 'log': ['[2020-03-12T15:41:58.477568] x', '[2020-03-12T15:42:19.960517] x', '[2020-03-12T15:42:19.960590] y', '[2020-03-12T15:42:19.960632] z']}, '_print': <redash.query_runner.python.CustomPrint object at 0x7f7288d7a710>}
```

PR #4737, opened 2020-03-17
Fix no button to add query tags

gabrieldutra:

- [x] Bug Fix

## Description
Reported in [this forum topic](https://discuss.redash.io/t/how-to-add-tags-to-query-in-v9/5770), this was introduced during #4670

I'll also create a Cypress test for both editing tags and description in a Query.

--

There should be no differences from the behavior when the query has tags, but I'll upload a gif soon.


arikfr:
🐞

Issue #4727, opened 2020-03-11
Don't autoselect "counter" as row option in Counter settings


gabrieldutra:
### Issue Summary
When creating a Counter visualization it defaults the Column name to be "counter", that seems confusing as most times the query won't have a "counter" column, I think we should either select the First row, or don't select anything and leave it with the placeholder (I prefer this one).

Related: https://github.com/getredash/redash/issues/4265#issuecomment-557468837

I plan to address this after #4642, but opening an issue for tracking/discussion.

### Technical details:

* Redash Version: master
* Browser/OS: --
* How did you install Redash: --

PR #4738, opened 2020-03-17
Counter: Only autoselect existing columns in data

gabrieldutra:

- [x] Other

## Description
Don't autoselect "counter" column when creating a Counter. Basically this only selects a default column value in the UI when it exists in data.

#4727 

--


kravets-levko:
In general it looks good, but one thing I noticed: with current implementation UI will show empty column select, but visualization will be saved with `counter` column name. It may confuse user if query later will be updated and that column will appear in results. Probably it's better to do this check in `getOptions` function (it takes raw options and query results rows/columns). WDYT?

gabrieldutra:
> but visualization will be saved with counter column name

I missed that one :sweat_smile:. My goal was to keep backward compatibility at most. But it makes sense to have it on the `getOptions`, so moved to it :+1:

PR #4150, opened 2019-09-18
Bar chart with second y axis overlaps data series

kravets-levko:
## What type of PR is this? (check all applicable)

- [x] Bug Fix

## Description

Assign `offsetgroup` for bar series so Plotly will offset them properly.

## Related Tickets & Documents

Fixes getredash/redash#2292

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

Before:
image

After (the same chart settings and data, but with `offsetgroup`):
image

PR #4752, opened 2020-03-23
Upgrade Plotly

kravets-levko:

- [x] Other

## Description

Upgrade Plotly to latest version to use new features (like `offsetgroup` needed for getredash/redash#4150) also for smoother upgrades in future.

I tested Chart visualization a lot after upgrade and didn't notice anything critical. Seems Plotly now computes auto-margins in a different way - I adjusted margins on our side to keep it as close as possible to previous version. Also, it changed legend styles a bit (see screenshots before/after) - unfortunately, I didn't find a good fix:

image
image

getredash/redash#4150

### DIFF
--- a/client/app/visualizations/chart/plotly/prepareLayout.js
+++ b/client/app/visualizations/chart/plotly/prepareLayout.js
@@ -121,7 +121,7 @@ export default function prepareLayout(element, options, data) {
     margin: { l: 10, r: 10, b: 10, t: 25, pad: 4 },
     width: Math.floor(element.offsetWidth),
     height: Math.floor(element.offsetHeight),
-    autosize: true,
+    autosize: false,
     showlegend: has(options, "legend") ? options.legend.enabled : true,
   };
 
--- a/client/app/visualizations/chart/plotly/prepareLayout.js
+++ b/client/app/visualizations/chart/plotly/prepareLayout.js
@@ -118,7 +118,7 @@ function prepareBoxLayout(layout, options, data) {
 
 export default function prepareLayout(element, options, data) {
   const layout = {
-    margin: { l: 5, r: 5, b: 5, t: 20, pad: 4 },
+    margin: { l: 10, r: 10, b: 5, t: 20, pad: 4 },
     width: Math.floor(element.offsetWidth),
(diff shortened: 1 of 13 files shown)


client/app/visualizations/chart/plotly/prepareLayout.js
```diff
@@ -118,10 +118,10 @@ function prepareBoxLayout(layout, options, data) {
 
 export default function prepareLayout(element, options, data) {
   const layout = {
-    margin: { l: 10, r: 10, b: 10, t: 25, pad: 4 },
+    margin: { l: 10, r: 10, b: 5, t: 20, pad: 4 },
```
arikfr: Without this change there are just bigger margins?

client/app/visualizations/chart/plotly/prepareLayout.js
```diff
@@ -118,10 +118,10 @@ function prepareBoxLayout(layout, options, data) {
...
     width: Math.floor(element.offsetWidth),
     height: Math.floor(element.offsetHeight),
-    autosize: true,
+    autosize: false,
```
arikfr: What do we lose by disabling autosize?

kravets-levko: Built-in autosize feature is based on `window.onresize`, and since we have implemented this feature on our own (via `resizeObserver` service which is even more powerful) - we don't really need it. So  turned it off just to save some resources.

client/app/visualizations/chart/plotly/prepareLayout.js
kravets-levko: Yes, and for me it looked too large; left/right paddings also changed a little bit, but in all cases it depends on axis labels (seems Plotly changed an algorithm that measures labels and now it preserves few more pixels for labels). Also seems that extra space also depends on chart size, therefore I was unable to fit margins perfectly, just as close as possible

kravets-levko:
> The difference in legend style is its width/where the scroll is?

Yes, and there is no any setting to control it. I tried to fix it with CSS, but it has some internal state and after any user interaction it returns to its default look.

PR #4758, opened 2020-03-27
Move the dropdown to the side in Dashboard Edit Mode

### LINKED ISSUE #4477: When in edit mode, user unable to edit widgets
> I'm setting up dashboard with queries that have variables in them. If I use a static value for the variable, I'm unable to see the value I've set and edit the parameter again while I'm still in edit mode. Currently I have to leave edit mode, then click on the "..." at the top right of the widget to be able to edit the parameter. Is there an edit parameter entrance I'm not seeing while in edit mode?

> The same goes for editing text boxes as well actually. I have to leave edit mode in order to edit the actual text.

It definitely makes no sense and confusing that you can't edit a widget when editing a dashboard. We should add an edit button next to the delete one, that will allow doing these operations.

gabrieldutra:

- [x] Other

## Description
Basically opening a PR with https://github.com/getredash/redash/issues/4477#issuecomment-583571496 for discussion

Closes #4477

edit-mode-widgets

### DIFF
--- a/client/app/components/dashboards/dashboard-widget/Widget.less
+++ b/client/app/components/dashboards/dashboard-widget/Widget.less
@@ -1,4 +1,4 @@
-@import '../../../assets/less/inc/variables';
+@import "../../../assets/less/inc/variables";
 
 .tile .t-header .th-title a.query-link {
   color: rgba(0, 0, 0, 0.5);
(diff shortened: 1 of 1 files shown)


client/app/components/dashboards/dashboard-widget/Widget.less
```diff
@@ -10,6 +10,7 @@
 
 .widget-wrapper {
   .widget-actions {
+    display: flex;
```
gabrieldutra: This is the actual change. (remove `display: none`  from the button and use `display: flex`)

PR #4760, opened 2020-03-29
Fix there's no publish button on mobile query page

gabrieldutra:

- [x] Bug Fix

## Description
Reported by @susodapop, the Publish Button was removed during the Query Page migration since there was too many buttons and it was breaking the width layout. But since #4663 the buttons wrap in a new row, so we can fit the Publish button back.

If you think there are too many buttons on mobile version it's also possible to move the Publish button to the dropdown.

--

**Query View**
| Before | After |
| --------- | ------ |
| Screen Shot 2020-03-29 at 12 57 39  | Screen Shot 2020-03-29 at 12 55 23 |

**Query Source**
| Before | After |
| --------- | ------ |
| Screen Shot 2020-03-29 at 12 57 20  | Screen Shot 2020-03-29 at 12 56 34 |

### DIFF
--- a/client/app/pages/queries/components/QueryPageHeader.jsx
+++ b/client/app/pages/queries/components/QueryPageHeader.jsx
@@ -70,7 +70,7 @@ export default function QueryPageHeader({
   tagsExtra,
   onChange,
 }) {
-  const isMobile = !useMedia({ minWidth: 768 });
+  const isDesktop = useMedia({ minWidth: 768 });
   const queryFlags = useQueryFlags(query, dataSource);
   const updateName = useRenameQuery(query, onChange);
   const updateTags = useUpdateQueryTags(query, onChange);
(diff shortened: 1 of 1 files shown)


gabrieldutra:
Update: I actually decided to proceed with "If you think there are too many buttons on mobile version it's also possible to move the Publish button to the dropdown." as it looked to me the Query View buttons could overflow on smaller mobile screens (the view page has the "Refresh" button too). So the result is now:

query-publish-mobile

client/app/pages/queries/components/QueryPageHeader.jsx
```diff
@@ -68,8 +69,8 @@ export default function QueryPageHeader({
   headerExtra,
   tagsExtra,
   onChange,
-  onRefresh,
 }) {
+  const isMobile = !useMedia({ minWidth: 768 });
```
kravets-levko:
```suggestion
  const isMobile = useMedia({ maxWidth: 768 });
```

gabrieldutra: That one was actually to make sure it had the same media rule as the CSS has (even though it should differ in only 1px :sweat_smile:)

kravets-levko: How about 
```js
const isDesktop = useMedia({ minWidth: 768 });
```
?

PR #4777, opened 2020-04-05
Use total_seconds to find stale jobs

rauchy:

- [x] Bug Fix

## Description
Seems that deletion of stale failed jobs didn't work if the job ran in intervals larger than a day.


jezdez:
Woops, sorry for slowly catching up on code review, folks.

PR #4782, opened 2020-04-07
Fix Multi-Filters: "select all" makes table view unscrollable

### LINKED ISSUE #4772: Multi-Filters: "select all" makes table view unscrollable
### Issue Summary

At 720p screen resolution, the multi-select filter widget makes the table unscrollable if too many values are selected.

image

I don't think it depends on the screen resolution. But whether the `<select>` widget expands to the bottom of the view port. If it does, scrolling breaks.

### Steps to Reproduce

I added [an example](https://redash-preview.netlify.com/queries/358/source) on redash-preview.

1. Click "select all" in the multi-filter
2. Use responsive mode in your browser to simulate 1280x720 resolution
3. Try scrolling down to see the table

#### Secondary Bug 

The **Select All** prompt  is stuck in `checked` state, even after clicking **clear**.

image

### Technical details:

* Redash Version: v9 Alpha
* Browser/OS: FF
* How did you install Redash: Redash-Preview

gabrieldutra:

- [x] Bug Fix

## Description
This PR holds 3 changes to improve experience with multi-filters on the query pages:
1. Introduce `max-height: 40%` to the filters area, this way it creates a scroll rather than taking the whole space visualizations have;
2. On Query View Page, also check for height > 500px to determine if the page will have a fixed layout. This way pages with small height areas won't have the fixed layout;
3. Add `maxTagCount` of 5, it was a number to have +- 2 lines in each multi-filter, but LMK if you have other suggestions.

Fixes #4772 

multi-filters-select-all

### DIFF
--- a/client/app/components/Filters.jsx
+++ b/client/app/components/Filters.jsx
@@ -102,7 +102,9 @@ function Filters({ filters, onChange }) {
                     allowClear={filter.multiple}
                     optionFilterProp="children"
                     showSearch
-                    maxTagCount={5}
+                    maxTagCount={3}
+                    maxTagTextLength={10}
+                    maxTagPlaceholder={num => `+${num.length} more`}
                     onChange={values => onChange(filter, values)}>
                     {!filter.multiple && options}
                     {filter.multiple && [
--- a/client/app/components/Filters.jsx
+++ b/client/app/components/Filters.jsx
@@ -102,6 +102,7 @@ function Filters({ filters, onChange }) {
                     allowClear={filter.multiple}
                     optionFilterProp="children"
                     showSearch
+                    maxTagCount={5}
                     onChange={values => onChange(filter, values)}>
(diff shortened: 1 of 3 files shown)


client/app/components/Filters.jsx
```diff
@@ -102,6 +102,7 @@ function Filters({ filters, onChange }) {
                     allowClear={filter.multiple}
                     optionFilterProp="children"
                     showSearch
+                    maxTagCount={5}
```
arikfr, line 105: How about use 3 here to be consistent with parameters?

gabrieldutra: Let's go with
```suggestion
                    maxTagCount={3}
                    maxTagTextLength={10}
                    maxTagPlaceholder={num => `+${num.length} more`}
```
then

arikfr:
👍

PR #4792, opened 2020-04-09
4791 redshift schema bugfix

### LINKED ISSUE #4791: Bug: Redshift data source unable to refresh schema
Redash version: 8.0.0+b32245 (a16f551e)
Data Source: Redshift
Redshift cluster version: 1.0.14436 

Error:
image
API: `/api/data_sources/1/schema?refresh=true`
Response: `{"error": {"message": "Error retrieving schema.", "code": 2}}`

Diagnosis:
Manually running the following query: https://github.com/getredash/redash/blob/master/redash/query_runner/pg.py#L323 returns error 
`Error running query: schema "pg_temp_16" does not exist`

Solution:
Update https://github.com/getredash/redash/blob/master/redash/query_runner/pg.py#L330
to include `and table_schema NOT LIKE 'pg_%'`

atharvai:
- [x] Bug Fix

## Description
Updates the `_get_tables` method for Redshift Data Source to exclude `pg_%` tables. There are permission issues accessing internal schemas for Redshift and this excludes to avoid the errors.

Closes #4791 

(none)


redash/query_runner/pg.py
```diff
@@ -327,7 +327,8 @@ def _get_tables(self, schema):
                             column_name,
                             ordinal_position AS pos
             FROM svv_columns
-            WHERE table_schema NOT IN ('pg_internal','pg_catalog','information_schema')
+            WHERE table_schema NOT IN ('information_schema')
+            AND table_schema NOT LIKE 'pg_%'
```
arikfr, line 331: While not very likely it is possible someone might name their schema with a `pg_` prefix (maybe when [P&G](https://us.pg.com/) adopts Redshift). How about we keep the original list and limit the `LIKE` prefix to `pg_temp_`?

atharvai: good point, sure will modify

arikfr:
Thanks!

PR #4599, opened 2020-01-29
Choropleth: allow to use custom maps

kravets-levko:
## What type of PR is this? (check all applicable)

- [x] Feature

## Description

Allow user to type arbitrary GeoJSON URL; existing build-it maps are available as dropdown options. Also, now instead of setting "Key Type" (type of key field from query result) user should map query result field to one of GeoJSON field.

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

image
image

PR #4793, opened 2020-04-09
Visualizations customizable settings

gabrieldutra:

- [x] Other

## Description
**Reorganize visualization components folder**
Components related to Redash app were removed from `visualizations/components` folder (EditVisualizationDialog and VisualizationRenderer). Also created a general Renderer and Editor that takes the `type` and select it from the list.

**Add customizable settings**
Creates a mutable object with options (currently only one from the HelpTrigger, soon others from clientConfig options).

**Update HelpTrigger**
HelpTrigger now alternatively receives a link and a title in replacement for the type constant. Also it renders a regular `<a>` when link is not inside redash.io (adds an External icon in such case). 

**clientConfig options**
Were all put into visualization settings.

--

--


client/app/visualizations/components/visualizationsSettings.js
```diff
@@ -0,0 +1,28 @@
...
+        </React.Fragment>
+      }>
+      <a
+        className="visualization-editor-context-help"
+        href="https://redash.io/help/user-guide/visualizations/formatting-numbers"
+        target="_blank"
+        rel="noopener noreferrer">
+        <Icon className="m-l-5 m-r-5" type="question-circle" theme="filled" />
+      </a>
+    </Tooltip>
```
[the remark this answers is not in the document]
gabrieldutra: > This probably means we need to change the API of HelpTrigger to take URL and text instead of constants like it does now, but I think it's a reasonable tradeoff.

I though about it too, to do that we also need to add a check to make sure the base url is from `redash.io`.

What made me use this option instead is because that's the only usage we have for it now (and I wanted to focus on the settings part first, rather than changing the HelpTrigger). I was actually considering giving up on the HelpTrigger as it was just one use, but on a second thought it's very likely that we add more entries in the future, so I'll just change it to also accept a separate link and text.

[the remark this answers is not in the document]
gabrieldutra: HelpTrigger updated to receive `href` and `title` props separately as an alternative to constants. It also accepts URLs ouside of `redash.io` now, but renders a regular link in such case.

client/app/components/visualizations/VisualizationRenderer.jsx
```diff
@@ -42,12 +43,17 @@ export default function VisualizationRenderer(props) {
     setFilters(combineFilters(filtersRef.current, props.filters));
   }, [props.filters]);
 
+  const cleanColumnNames = useMemo(
+    () => map(data.columns, col => ({ ...col, name: getColumnCleanName(col.friendly_name) })),
```
gabrieldutra, line 47: @kravets-levko it seemed to make sense to move `getColumnCleanName` out of Table vis and rather apply it to all visualizations, do you see any problems may happen with this change? (that method is used to remove `::filter` from column names for example)

client/app/components/visualizations/visualizationComponents.jsx
```diff
@@ -0,0 +1,42 @@
...
+      choroplethAvailableMaps: {
+        countries: {
+          name: "Countries",
+          url: countriesDataUrl,
+        },
+        subdiv_japan: {
+          name: "Japan/Prefectures",
+          url: subdivJapanDataUrl,
+        },
+      },
```
gabrieldutra, line 24: I tried to go in the direction of #4599 as it has a complementary code to this.

client/app/visualizations/choropleth/Renderer/index.jsx
```diff
@@ -1,25 +1,16 @@
-import { omit, merge } from "lodash";
+import { omit, merge, get } from "lodash";
+import axios from "axios";
```
gabrieldutra, line 2: No reason to add a service axios dependency instead of using the base one.

Issue #2013, opened 2017-10-10
[Feature] Tunnel support in default Docker image


Slind14:
Hi there,
it would be great if the default docker image would feature tunnel support. Meaning preinstalled autossh and loading a configuration with the hostname, port and authentication key.

jezdez:
This seems quite out of scope and can be easily achieved with a custom Dockerfile based on the Redash Docker image. Closing as wontfix, feel free to reopen @arikfr if you disagree.

PR #4797, opened 2020-04-12
Run queries through ad-hoc SSH tunnels

rauchy:
- [x] Feature

## Description
Redash requires network access to the data sources it needs to query, but sometime these data sources aren't publicly available. A way around this is using [SSH tunnels](https://www.ssh.com/ssh/tunneling/example), which can be setup manually outside Redash, but can also be setup automatically before a connection is made, and closed automatically after they are no longer required.

This PR sets up ad-hoc SSH tunnels for data sources if they have a `ssh_tunnel` dict in their options, which must include an `ssh_host` and `ssh_username` of the bastion server to connect through.

In order to enable SSH tunnels, `dynamic_settings.ssh_tunnel_auth()` must be implemented to allow the Redash instance to authenticate with the bastion server(s).

# Related Issues

#2013

### DIFF
--- a/client/app/pages/queries/hooks/useDataSourceSchema.js
+++ b/client/app/pages/queries/hooks/useDataSourceSchema.js
@@ -1,7 +1,7 @@
 import { reduce } from "lodash";
 import { useCallback, useEffect, useRef, useState, useMemo } from "react";
 import { axios } from "@/services/axios";
-import DataSource from "@/services/data-source";
+import DataSource, { SCHEMA_NOT_SUPPORTED } from "@/services/data-source";
 import notification from "@/services/notification";
 
 function sleep(ms) {
@@ -20,6 +20,8 @@ function getSchema(dataSource, refresh = undefined) {
           return fetchSchemaFromJob(data);
         } else if (data.job.status === 3) {
           return data.job.result;
+        } else if (data.job.status === 4 && data.job.error.code === SCHEMA_NOT_SUPPORTED) {
+          return [];
         } else {
           return Promise.reject(new Error(data.job.error));
         }
--- a/client/app/pages/queries/hooks/useDataSourceSchema.js
+++ b/client/app/pages/queries/hooks/useDataSourceSchema.js
@@ -1,22 +1,35 @@
 import { reduce } from "lodash";
 import { useCallback, useEffect, useRef, useState, useMemo } from "react";
-import DataSource, { SCHEMA_NOT_SUPPORTED } from "@/services/data-source";
+import { axios } from "@/services/axios";
+import DataSource from "@/services/data-source";
 import notification from "@/services/notification";
+
+function sleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
 
 function getSchema(dataSource, refresh = undefined) {
   if (!dataSource) {
     return Promise.resolve([]);
   }
 
-  return DataSource.fetchSchema(dataSource, refresh)
-    .then(data => {
-      if (data.schema) {
-        return data.schema;
-      } else if (data.error.code === SCHEMA_NOT_SUPPORTED) {
-        return [];
-      }
-      return Promise.reject(new Error("Schema refresh failed."));
+  const fetchSchemaFromJob = (data) => {
+    return sleep(1000).then(() => {
+      return axios.get(`api/jobs/${data.job.id}`).then(data => {
+        if (data.job.status < 3) {
+          return fetchSchemaFromJob(data);
+        }
+        else if (data.job.status === 3) {
+          return data.job.result;
+        } else {
+          return Promise.reject(new Error(data.job.error));
+        }
+      })
     })
+  };
+
+  return DataSource.fetchSchema(dataSource, refresh)
+    .then(fetchSchemaFromJob)
     .catch(() => {
       notification.error("Schema refresh failed.", "Please try again later.");
       return Promise.resolve([]);
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -1,4 +1,5 @@
 import logging
+import time
 
 from flask import make_response, request
 from flask_restful import abort
@@ -20,6 +21,7 @@ from redash.query_runner import (
 )
 from redash.utils import filter_none
 from redash.utils.configuration import ConfigurationContainer, ValidationError
+from redash.tasks.general import test_connection
 
 
 class DataSourceTypeListResource(BaseResource):
@@ -245,10 +247,14 @@ class DataSourceTestResource(BaseResource):
         )
 
         response = {}
-        try:
-            data_source.query_runner.test_connection()
-        except Exception as e:
-            response = {"message": str(e), "ok": False}
+
+        job = test_connection.delay(data_source.id)
+        while not (job.is_finished or job.is_failed):
+            time.sleep(1)
+            job.refresh()
+
+        if isinstance(job.result, Exception):
+            response = {"message": str(job.result), "ok": False}
         else:
             response = {"message": "success", "ok": True}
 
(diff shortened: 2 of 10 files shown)


redash/query_runner/__init__.py
```diff
@@ -70,6 +73,34 @@ def type(cls):
...
+            return self.configuration["port"]
+        else:
+            raise NotImplementedError()
+
+    @port.setter
+    def port(self, port):
+        if "port" in self.configuration:
+            self.configuration["port"] = port
+        else:
+            raise NotImplementedError()
```
arikfr, line 102: Let's add a comment explaining the purpose of these properties.

redash/query_runner/__init__.py
```diff
@@ -303,3 +334,46 @@ def guess_type_from_string(string_value):
...
+                raise type(error)("SSH tunnel: {}".format(str(error)))
+            else:
+                with stack:
+                    try:
+                        query_runner.host, query_runner.port = server.local_bind_address
+                        result = f(*args, **kwargs)
+                    finally:
+                        query_runner.host, query_runner.port = remote_host, remote_port
+
+                    return result
```
susodapop, line 373: Would it make sense to lose `else` here and remove an indent? I avoid `try...except...else` unless I catch the exception _without_ re-raising it.

rauchy: Yeah this makes no sense. I had originally used a different implementation where it made sense but didn't change the function's outline as I refactored. 93a97486.

redash/query_runner/__init__.py
rauchy: 👍 8da215b3

rauchy:
@arikfr forceful schema refreshes can't be handled by gunicorn, so I [offloaded them to RQ](https://github.com/getredash/redash/pull/4797/commits/75a48b098082385ab5241b641239a115e19bed65) and polled for their results using the `/jobs` endpoint. This seems to work well. If you think the implementation is on track, I'll use the same method for connection testing.

redash/tasks/general.py
```diff
@@ -63,6 +65,28 @@ def send_mail(to, subject, html, text):
...
+    try:
+        data_source = models.DataSource.get_by_id(data_source_id)
+        data_source.query_runner.test_connection()
+    except Exception as e:
+        return e
+    else:
+        return True
+
+
+@job("schemas", queue_class=Queue, at_front=True, timeout=30, ttl=90)
```
arikfr, line 79: Unfortunately for some data sources timeout needs to be much longer. Let's start with 5 minutes.

redash/serializers/__init__.py
```diff
@@ -294,14 +294,15 @@ def serialize_job(job):
...
+        result = query_result_id = job.result
 
     return {
         "job": {
             "id": job.id,
             "updated_at": updated_at,
             "status": status,
             "error": error,
+            "result": result,
             "query_result_id": query_result_id,
```
arikfr, line 306: To avoid accumulating tech debt, maybe we can return `query_result_id` only when it's a `execute_query` task?

arikfr: (only if it's simple check)

redash/handlers/data_sources.py
```diff
@@ -182,19 +187,9 @@ def get(self, data_source_id):
...
-
-        try:
-            response["schema"] = data_source.get_schema(refresh)
-        except NotSupported:
-            response["error"] = {
-                "code": 1,
-                "message": "Data source type does not support retrieving schema",
-            }
-        except Exception:
-            response["error"] = {"code": 2, "message": "Error retrieving schema."}
```
arikfr, line 195: We need to keep reporting these error codes. For example, we might show different UI when it's not supported. And we don't want to throw random errors at the user but rather show "Error retrieving schema." (it might make sense to show some of the errors, but it's beyond the scope of this PR).

redash/serializers/__init__.py
rauchy: If we want to avoid accumulating tech debt, we should probably always return `result` and ditch `query_result_id`, no?

redash/handlers/data_sources.py
rauchy: I thought so too, but then I realized we aren't really using the [error codes defined in the service](https://github.com/getredash/redash/pull/4797/commits/75a48b098082385ab5241b641239a115e19bed65#diff-60f072516fcb2847feb283641b82e60bL4). The current implementation (returning `[]` on `NotSupported` from the backend) mimics the same behavior we have today - silent failing on `NotSupported` and a "Schema refresh failed." message when errors occur.

redash/serializers/__init__.py
arikfr: But this will break scripts that use /jobs for polling and expect `query_result_id`... :| We could really use API versioning at this point.

redash/handlers/data_sources.py
arikfr: I know that we don't really use them anymore in code -- this is something that happened during the conversion to React, but we should someday. It's better to be explicit about this.

redash/serializers/__init__.py
rauchy: I agree, that's why I wen't for doubling `result` and `query_result_id`, as the cheapest thing that won't break and won't be too awkward :(

redash/handlers/data_sources.py
rauchy: bc69c032a

redash/tasks/general.py
```diff
@@ -63,6 +65,33 @@ def send_mail(to, subject, html, text):
...
+        return e
+    else:
+        return True
+
+
+@job("schemas", queue_class=Queue, at_front=True, timeout=300, ttl=90)
+def get_schema(data_source_id, refresh):
+    try:
+        data_source = models.DataSource.get_by_id(data_source_id)
+        return data_source.query_runner.get_schema(refresh)
```
arikfr, line 83:
```suggestion
        return data_source.get_schema(refresh)
```

The data source method has additional logic (caching & sorting). The Query Runner is a more lower level and while it happens to accept a boolean parameter, it's not for refreshing (it's always refreshing).

client/app/pages/queries/hooks/useDataSourceSchema.js
```diff
@@ -1,22 +1,36 @@
...
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 function getSchema(dataSource, refresh = undefined) {
   if (!dataSource) {
     return Promise.resolve([]);
   }
 
+  const fetchSchemaFromJob = (data) => {
+    return sleep(1000).then(() => {
```
jezdez, line 17: This adds a noticeable delay to loading the schema browser without indication that it's loading the schema, could we add a "Loading... please wait" screen or something similar? Or reduce the sleep interval to something that is closer to an unnoticeable perception level?

Also, in case of a worker failure that prevents the job from ever returning a good result (e.g. Redis key eviction), could this add upper limit on the number of tries to load the data and then show a button to fetch the schema anew with a new job?

jezdez:
Oh one other comment, Marina and me spent a lot of time on a custom schema drawer feature in our fork (that was also merged here a while ago but was backed out after feedback from Arik): https://github.com/mozilla/redash/commit/109eac83cbcfc078b0d17eae81405ebf51f79250

Since this PR adds a background job to load the schema, I want to point out that our schema drawer feature has modified the `refresh_schema` job to store the schema in Redash's postgres to be able to 1) amend the schema with metadata such as description, column samples and example queries but 2) to cache the response for the schema drawer API endpoint with a Redis lock based refresh system to prevent the thundering herd problem of the refresh schema button. The results are lower traffic to cost intensive data sources (less $$$), immediate responses to the user and the ability to provide more details about the individual data source tables and columns.

If you'd like to get a quick demo again, please let me know, we're obviously eager to have alignment with you.

arikfr:
We need to do a follow up on this:

1. Make the API return cached version if it's available immediately (this will address the 1 second delay). 
2. Add some progress indicator (NTH though, as we had the same issue in the past).

Adding a lock could be nice, but I'm not sure if it's needed once we move the caching functionality to the API layer. Basically at this point it returns being the same behavior as before (in terms of number of restarts).

PR #4799, opened 2020-04-14
Disable Percy snapshot for Choropleth test

arikfr:

- [x] Other

## Description

The snapshot test fails almost on every PR and has no value. Disabled until a better solution is found.

redashlabs/product#53


kravets-levko:
@arikfr Since you're already on this - can you try in increase delay in `cy.wait()` few lines above? I guess it fails all the time because it does not respond to options changes fast enough.

kravets-levko:
I didn't work (the map should be red&yellow) :thinking: Okay, let's just disable it for now, I'll check it later

Issue #4803, opened 2020-04-14
AttributeError exception causing API 500 errors being experienced when searching Dashboards


Ralnoc:
### Issue Summary

In some situations, when doing a search for dashboards is resulting in an `AttributeError` exception on `obj.user.to_dict()`. I have been unable to yet pinpoint any direct cause, only the process that leads up to the issue. In addition, the issue only manifests when searching, and only with some characters, not all. Primarily if there is a large number of results. If I just navigate through the pages on the Dashboards list, there is no issues experienced.

The order that I have been able to trace through the code, which leads to this problem is this:

User types in search in the Dashboards page of Re:Dash
`DashboardListResource` is called to search DB for dashboards the meets the query using `models.Dashboard.search` - Successful
`DashboardListResource` orders the results of that query - Successful
`DashboardListResource` calls `serialize_dashboard` as part of the pagination process for the the results of the order_results - Successful
`handlers.base.paginate` paginates the results and passes it over to `serialize_dashboards` - Successful
`serialize_dashboard` executes to serialize the paginated results to return - Fails ( `items = [serializer(result) for result in results.items]` fails with the exception `AttributeError: 'NoneType' object has no attribute 'to_dict')`

This issue appears to have started after we upgraded from Re:Dash 6.0.0 to 8.0.0. We used the normal DB upgrade scripts as part of the upgrade process. We have tested the Queries and Users API endpoints and none of them throw this exception. It only happens on Dashboards and only when doing a search.

We suspect it is possible an issue with SQLAlchemy related to: https://github.com/sqlalchemy/sqlalchemy/issues/3650

This is the traceback being experienced:

```
[2020-04-14 12:10:05,461] ERROR in app: Exception on /api/dashboards [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python2.7/site-packages/flask_restful/__init__.py", line 477, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/flask_login/utils.py", line 228, in decorated_view
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/flask/views.py", line 84, in view
[...]

PR #4804, opened 2020-04-15
Dashboard Search bug fix

Ralnoc:

- [x] ~Feature~
- [x] Bug Fix

## Description
* Move Dashboard off `subqueryload()` loader in all() method due to inconsistent results bug in SQLAlchemy when leveraging distinct within a subqueryload call through paginate.
* ~Added source argument for the Presto query runner to specify `redash` so that queries coming from Re:Dash are reported as coming `redash` within Presto.~

#4803 

N/A

### DIFF
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -98,6 +98,7 @@ class Presto(BaseQueryRunner):
             port=self.configuration.get("port", 8080),
             protocol=self.configuration.get("protocol", "http"),
             username=self.configuration.get("username", "redash"),
+            source='redash',
             password=(self.configuration.get("password") or None),
             catalog=self.configuration.get("catalog", "hive"),
             schema=self.configuration.get("schema", "default"),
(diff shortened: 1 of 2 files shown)


redash/query_runner/presto.py
```diff
@@ -98,6 +98,7 @@ def run_query(self, query, user):
             port=self.configuration.get("port", 8080),
             protocol=self.configuration.get("protocol", "http"),
             username=self.configuration.get("username", "redash"),
+            source='redash',
```
susodapop, line 101: Could we make this configurable in the data source setup screen? `redash` by default but the user could change it?

Ralnoc: Thats a valid request. Let me pull that line from this PR.

Ralnoc:
> Ideally these would be separate, unrelated PR's. But I'm new to git-iquette so I might be wrong about that.

Thats completely fair request. I was just trying to save some time. I'll create a separate PR once I refactor based on your feedback to have it added to the configuration page. As it stands, I have now removed that line from this PR and it only contains the fix for #4803 now.

Ralnoc:
@susodapop -
We have internally implemented this change and deployed it through to Production and have no issues with this bug manifesting since. Overall performance has not changed when searching Dashboards either. Let me know if you need any more information for the PR.

arikfr:
Thank you for the detailed report and bug fix, @Ralnoc !

Ralnoc:
Definitely glad to help!

PR #4807, opened 2020-04-16
Add support for Presto User Impersonation

Ralnoc:

- [x] Feature

## Description
The newest version of PyHive now supports enabling User Impersonation for Presto. This allows you to configure an authentication user for the Presto Configuration, but still execute the actual queries in Presto as the user logged into Re:Dash. This is a critical component to be able to have user level authorization in Presto, while also communicating over the SSL interface which requires authentication.

In order to leverage this setting. you must have User Impersonation configured in Presto. This is the documentation for doing so: https://docs.starburstdata.com/latest/security/impersonation.html

Upgrading PyHive to 0.6.2 to gain access to the new User Impersonation code. 

https://docs.starburstdata.com/latest/security/impersonation.html
https://github.com/dropbox/PyHive/pull/309

### DIFF
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -45,6 +45,10 @@ class Presto(BaseQueryRunner):
                 "schema": {"type": "string"},
                 "catalog": {"type": "string"},
                 "username": {"type": "string"},
+                "user_impersonation": {
+                    "type": "boolean",
+                    "title": "Enable User Impersonation",
+                },
                 "password": {"type": "string"},
             },
             "order": [
@@ -52,6 +56,7 @@ class Presto(BaseQueryRunner):
                 "protocol",
                 "port",
                 "username",
+                "user_impersonation",
                 "password",
                 "schema",
                 "catalog",
@@ -93,11 +98,17 @@ class Presto(BaseQueryRunner):
         return list(schema.values())
 
     def run_query(self, query, user):
+        enable_user_impersonation = self.configuration.get("user_impersonation")
+        principle_username = None
+        if enable_user_impersonation and user is not None:
+            principle_username = user.email.split('@')[0]
+
         connection = presto.connect(
             host=self.configuration.get("host", ""),
             port=self.configuration.get("port", 8080),
             protocol=self.configuration.get("protocol", "http"),
             username=self.configuration.get("username", "redash"),
+            principle_username=principle_username,
             password=(self.configuration.get("password") or None),
             catalog=self.configuration.get("catalog", "hive"),
             schema=self.configuration.get("schema", "default"),
--- a/redash/query_runner/presto.py
+++ b/redash/query_runner/presto.py
@@ -99,16 +99,16 @@ class Presto(BaseQueryRunner):
 
     def run_query(self, query, user):
         enable_user_impersonation = self.configuration.get("user_impersonation")
-        principle_username = None
+        principal_username = None
         if enable_user_impersonation and user is not None:
-            principle_username = user.email.split('@')[0]
+            principal_username = user.email.split('@')[0]
 
         connection = presto.connect(
             host=self.configuration.get("host", ""),
             port=self.configuration.get("port", 8080),
             protocol=self.configuration.get("protocol", "http"),
             username=self.configuration.get("username", "redash"),
-            principle_username=principle_username,
+            principal_username=principal_username,
             password=(self.configuration.get("password") or None),
             catalog=self.configuration.get("catalog", "hive"),
             schema=self.configuration.get("schema", "default"),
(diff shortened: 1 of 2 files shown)


Ralnoc:
@susodapop -
While I was at it, I raised this PR to add User Impersonation support for Presto. This will let Presto authenticate with a service account, and as long as PrestoSQL is configured correctly, execute the query as the logged in user, while still authenticating as the service account. This is a big one for my company which is currently having to modify the query runner after the fact to support individual user authorization in Presto because we leverage SAML.

redash/query_runner/presto.py
```diff
@@ -93,11 +98,17 @@ def get_schema(self, get_stats=False):
         return list(schema.values())
 
     def run_query(self, query, user):
+        enable_user_impersonation = self.configuration.get("user_impersonation")
+        principle_username = None
+        if enable_user_impersonation and user is not None:
+            principle_username = user.email.split('@')[0]
+
```
susodapop, line 105: Maybe I misunderstand the use-case here. Would it not make sense to allow arbitrary user selection in the data source setup screen? Since DS setup is restricted to account admins.

```suggestion
        enable_user_impersonation = (
            user.email.split("@")[0]
            if user is not None and self.configuration.get("user_impersonation", False)
            else None
        )
```

redash/query_runner/presto.py
```diff
@@ -93,11 +98,17 @@ def get_schema(self, get_stats=False):
...
         connection = presto.connect(
             host=self.configuration.get("host", ""),
             port=self.configuration.get("port", 8080),
             protocol=self.configuration.get("protocol", "http"),
             username=self.configuration.get("username", "redash"),
+            principle_username=principle_username,
```
susodapop, line 111: I just noticed: shouldn't `principle` be named `principal`? The [presto docs](https://docs.starburstdata.com/latest/security/impersonation.html#user-translation-rules) name it `PRINCIPAL`.

I understand [you wrote](https://github.com/dropbox/PyHive/pull/309) the support for this into `pyhive` as well so it might be an easy fix in both places. Might be more trouble than it's worth since changing it now might break other implementations.

Ralnoc: Yea. This is an autocorrect error that unfortunately no one caught until after it was released in pyhive. (I only just noticed it while I was working on this PR and types in `principal` and it threw an error.)

Making the update in PyHive would be difficult now that it is released, though I can at least implement the correct variable name in the code here.

redash/query_runner/presto.py
Ralnoc: So there are a number of parts for this process.

The primary use case for this is the need to have the user executing the query be passed through to Presto, so that Authorization rules and auditing can be done against the person actually running the query.

Internally Presto separates Authentication from Authorization. It also implements the ability for User Impersonation so that a service account with the right access configured in Presto can authenticate then tell Presto to run the query as the principal user. 

The internal behavior within Presto in this situation is that the query will be executed using the defined Resource groups for the principal user, and if they are using Starburst Presto leverage the Ranger Security Access control to determine if the principal user even has access to the resources they are attempting to query.

Because of these two parts, the general expectation is that the user who is logged in, is the person who should be executing the query. That is why it is a boolean setting to turn on User impersonation, and why we are extracting the username from the email address. 

If the desire was to run as a single user, then there would be no need for User Impersonation, as you could just leverage the account configured in the configuration page for authentication.

I hope this helps clarify the use case.

susodapop: Absolutely. This is gold that we can add to the documentation. Thank you for explaining.

redash/query_runner/presto.py
Ralnoc: I am reaching out there to work on getting that fix in place. 
https://github.com/dropbox/PyHive/pull/326

Worse case, I'll implement that PR to support both the old and new argument. Lets hold on this PR until a decision is made in PyHive.

Ralnoc: So they have already merged the correction PR and it will be in the next PyHive release. I'll update this PR with the next release that has that code change in it and update the code to match.

Ralnoc: I've updated this code to match the changes in the future release of PyHive. This PR will definitely fail checks until that release is out. I will clean up the PR in the `requirements_all_ds.txt` once the new version of pyhive with the change is released.

edvinas-maciulis:
Is anyone looking at the failure on Datree Smart Policy?

guidopetri:
@Ralnoc , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

Issue #4813, opened 2020-04-19
Active users marked as pending


arikfr:
### Issue Summary

Sometimes although a user is active they are still marked as pending. You can find those with the following query:

```sql
SELECT id,
       name,
       email,
       details::JSON->>'is_invitation_pending' AS is_pending,
                       details::JSON->>'active_at' AS active_at
FROM users
WHERE details::JSON->>'is_invitation_pending' = 'true' and details::JSON->>'active_at' <> ''
```

In a normal environment this should return 0 rows.

It doesn't look like a user can start using the system without being marked as active, so my best theory for how this happens is:

* [`sync_last_active_at`](https://github.com/getredash/redash/blob/e99c37a36a8f091cac001af130937bcb267d55a5/redash/models/users.py#L30-L45) is triggered when the user object has still pending value as false.
* When the details object is updated it's updated with the old value overwriting the new status.

I'm not 100% sure about this theory as the update to active_at triggered by an authenticated request and this only happens after the object is updated. But I couldn't find a better one.

### Technical details:

* Redash Version: v9

PR #4817, opened 2020-04-20
Accept invitation through password resets

rauchy:
- [x] Bug Fix

## Description

As described in #4813, some users have their `is_invitation_pending` flag set to `True`, but they are clearly active users.

This situation happens when users are invited, but instead of following the invitation link, they request a password reset. Following a successful password reset, they are logged in, but their `is_invitation_pending` flag remains as `True` forever.

#4813


arikfr:
Nice catch!

PR #4828, opened 2020-04-24
Show explicit user name instead of avatar in lists.

arikfr:

- [x] Other

## Description

In several places in the application we show only the user avatar instead of their name. In large organizations that don't use avatars this leads to long lists with random icons instead of names.

* Updated all lists (queries, dashboards, alerts) to show explicit user name. 
* On the dashboard page showing the user name when hovering over the user avatar.

There is probably a better UX to be achieved by moving the user name to a second line below the object name, but compared to the effort this seemed like a quick win.

| Before | After |
| ------ | ------ |
| Queries List | After: Queries List |
| Dashboards List | After: Dashboards List |
| Alerts List | After: Alerts List |
| | Dashboard Page |

### DIFF
--- a/client/app/pages/dashboards/components/DashboardHeader.jsx
+++ b/client/app/pages/dashboards/components/DashboardHeader.jsx
@@ -41,7 +41,9 @@ function DashboardPageTitle({ dashboardOptions }) {
             ignoreBlanks
           />
         </h3>
-        <img src={dashboard.user.profile_image_url} className="profile-image" alt={dashboard.user.name} />
+        <Tooltip title={dashboard.user.name} placement="bottom">
+          <img src={dashboard.user.profile_image_url} className="profile-image" alt={dashboard.user.name} />
+        </Tooltip>
       </div>
       <DashboardTagsControl
         tags={dashboard.tags}
--- a/client/app/pages/queries-list/QueriesList.jsx
+++ b/client/app/pages/queries-list/QueriesList.jsx
@@ -77,12 +77,11 @@ class QueriesList extends React.Component {
         width: null,
       }
     ),
-    Columns.avatar({ field: "user", className: "p-l-0 p-r-0" }, name => `Created by ${name}`),
+    Columns.custom((text, item) => item.user.name, { title: "Created By" }),
     Columns.dateTime.sortable({ title: "Created At", field: "created_at" }),
-    Columns.duration.sortable({ title: "Runtime", field: "runtime" }),
     Columns.dateTime.sortable({ title: "Last Executed At", field: "retrieved_at", orderByField: "executed_at" }),
     Columns.custom.sortable((text, item) => <SchedulePhrase schedule={item.schedule} isNew={item.isNew()} />, {
-      title: "Update Schedule",
+      title: "Refresh Schedule",
       field: "schedule",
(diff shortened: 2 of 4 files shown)


client/app/pages/queries-list/QueriesList.jsx
```diff
@@ -77,12 +77,11 @@ class QueriesList extends React.Component {
...
       }
     ),
-    Columns.avatar({ field: "user", className: "p-l-0 p-r-0" }, name => `Created by ${name}`),
+    Columns.custom((text, item) => item.user.name, { title: "Created By" }),
     Columns.dateTime.sortable({ title: "Created At", field: "created_at" }),
-    Columns.duration.sortable({ title: "Runtime", field: "runtime" }),
     Columns.dateTime.sortable({ title: "Last Executed At", field: "retrieved_at", orderByField: "executed_at" }),
     Columns.custom.sortable((text, item) => <SchedulePhrase schedule={item.schedule} isNew={item.isNew()} />, {
-      title: "Update Schedule",
+      title: "Refresh Schedule",
```
arikfr, line 84: Fixed for consistency with the page.

client/app/pages/queries-list/QueriesList.jsx
```diff
@@ -77,12 +77,11 @@ class QueriesList extends React.Component {
         width: null,
       }
     ),
-    Columns.avatar({ field: "user", className: "p-l-0 p-r-0" }, name => `Created by ${name}`),
+    Columns.custom((text, item) => item.user.name, { title: "Created By" }),
     Columns.dateTime.sortable({ title: "Created At", field: "created_at" }),
-    Columns.duration.sortable({ title: "Runtime", field: "runtime" }),
```
arikfr, line 82: It doesn't feel very useful so removed it to have more space for names.

client/app/pages/dashboards/components/DashboardHeader.jsx
```diff
@@ -41,7 +41,9 @@ function DashboardPageTitle({ dashboardOptions }) {
             ignoreBlanks
           />
         </h3>
-        <img src={dashboard.user.profile_image_url} className="profile-image" alt={dashboard.user.name} />
+        <Tooltip title={dashboard.user.name} placement="bottom">
```
susodapop, line 44: FWIW: This event doesn't work on mobile since there's no "hover" event. Not a big deal but maybe something to address down the road.

arikfr: On mobile you can click on it to see the tooltip.

susodapop: This doesn't work on the latest iOS via iPhone XS. Does work on iPad.

arikfr: It worked for me on iPhone X. 😮

susodapop: I bet that even if this didn't work it wouldn't impact many users 🤷‍♂️

yankeeguyu:
How to add sortable to 'Created By'? I add orderByfield 'created_by', it worded for query list but not for dabhoard list page.

PR #4837, opened 2020-04-25
Separate visualizations into their own package

gabrieldutra:

- [x] Other

## Description
Separate visualizations into their own package

To-do list:
- [x] Choose components to be shared instead of copied
- [x] Remove `p-r-5` from redash-visualizations/src/visualizations/table/utils.js
- [x] Remove visualizations dependencies from main `package.json`
- [x] Adapt main build scripts
  - [x] Netlify
  - [x] Circleci (50%)
  - [x] Dev env
- [x] Fix jest tests

--

--


redash-visualizations/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "redash-visualizations",
```
arikfr, line 2: Let's name it `@redash/viz`. I've added you to the `redash` org on npm, so you should be able to eventually push it there.

redash-visualizations/package.json
```diff
@@ -0,0 +1,77 @@
...
+  "version": "1.0.0",
```
arikfr, line 3: Let's start with version 0.1.0 and switch to 1.0 when the API is more stable.

viz-lib/src/visualizations/chart/Editor/ColorsSettings.test.js
```diff
@@ -70,7 +70,7 @@ describe("Visualizations -> Chart -> Editor -> Colors Settings", () => {
         .simulate("click");
     });
 
-    test("Sets custom color scheme", async done => {
+    test("Sets custom color scheme", done => {
```
gabrieldutra, line 73: Was this `async` important for something I'm missing? It was causing this test to fail removing it didn't cause any harm 🤷‍♂️ 
(Probably some babel setting missing for it to work, but it was way easier to just remove it)

Dockerfile
```diff
@@ -2,7 +2,8 @@ FROM node:12 as frontend-builder
 
 WORKDIR /frontend
 COPY package.json package-lock.json /frontend/
-RUN npm ci
+COPY viz-lib /frontend/viz-lib
+RUN npm ci --unsafe-perm
```
gabrieldutra, line 6: This seems to be required to run the preinstall script (which installs viz-lib deps and build it) when we run the command as root user. An alternative is to switch the user to the redash one in this scope. I tried to switch to it for all commands, but didn't seem that quick. Another option is to switch to the user only while running this command.

daniellangnet:
I'm intrigued by where this is going! 👍 
Just curious what the bigger picture is here so that I can decide when it makes sense for me to update my KPI visualization branch?

viz-lib/src/visualizations/chart/Editor/ColorsSettings.test.js
kravets-levko: I think `async` is not needed here, so it's ok to remove. Return value is always ignored, and `done` is used for async tests

gabrieldutra:
Hi @daniellangnet, the idea here is to have visualizations to stand on their own (not to be dependent on any other part of Redash), so they can be turned into a separate package. All the bigger changes for that to be possible were already delivered in past PRs, so this one is mostly a setup + move the visualizations folder. In any case, this is the last one that will have "more significant" changes on the codebase (although it just moves files and it should be an easy merge). The goal is to make it transparent to development flows, so ideally it shouldn't affect your work. Thanks for contributing and I'll give a quick check on your PR anyways. 🙂

gabrieldutra:
This PR changes in Brief for ones that will review it:
- Create `viz-lib` package.json and set it up with webpack to generate UMD bundled `redash-visualizations.js` and babel to just transpile it for more specific usage (in Redash for ex). I wanted to offer a transpiled version with css files instead of less too, but that seemed too much hassle;
- Copied all visualizations folder to the `viz-lib` one;
- Dev environment runs babel:watch on the lib in the background;
- Added a build step for visualizations as a `preinstall` script (seemed the simplest way to make it transparent, not to require extra commands on the flow we have today). Future alternatives include maybe move it to our own bin scripts.
- Separated visualizations Jest tests. Those were easy because they are unit tests, for the Cypress ones it's a bit more complicated since they are Redash tests (and it does make sense to have Redash e2e tests for it), but eventually we can move the more "unitary" of them to be tested through Storybook or sth;
- Deleted duplicated files, a few components are now imported from the viz lib, though there are some duplicated that I didn't find to make sense to locate on the lib (e.g: sanitize service, which is basically our settings over DOMPurify). LMK if you think other should be in the viz lib too.

Dockerfile
arikfr: > but didn't seem that quick

in terms of effort required to make the switch?

Regardless, `--unsafe-perm` should be safe in the context of running inside a Docker container build...

gabrieldutra: > in terms of effort required to make the switch?

In terms of changes in that file, mostly needs to update chown/chmod permissions for that folder. As the `--unsafe-perm` seemed simpler, I wanted to show it as an option, but needed to confirm regarding safety. I actually didn't sit down and actually tried to change thos yesterday, but I did today and I could get it to work like this:

```Dockerfile
FROM node:12 as frontend-builder

RUN useradd --create-home redash

WORKDIR /frontend

COPY package.json package-lock.json webpack.config.js /frontend/
COPY client /frontend/client

RUN chown -R redash /frontend
USER redash
RUN npm ci
RUN ls -l /frontend
RUN npm run build
```

(notice I changed the `COPY` order to run only one `chown`, not sure if there was a reason to have it after the `npm ci` command though)

Anyway, LMK which option do you prefer.

arikfr: > (notice I changed the COPY order to run only one chown, not sure if there was a reason to have it after the npm ci command though)

The reason for the current order is to make sure that we don't have to reinstall if `package.json`/`package-lock.json` didn't change, as they change less frequently than the codebase.

Let's go with the unsafe option for now.

arikfr:
Should we merge this now?

arikfr:
🚀

PR #4842, opened 2020-04-28
Specify restylers versions for restyled

gabrieldutra:

- [x] Other

## Description
This PR specifies versions for current restylers following [this](https://github.com/restyled-io/restyled.io/wiki/Restyler-Versions#avoid-specific-surprises). In the short term it sets the prettier version to be the same we use currently, to avoid differences.

--

--


arikfr:
Thanks!

PR #4855, opened 2020-04-30
Fix: showing current settings tab broken in MULTI_ORG.

arikfr:

- [x] Bug Fix

## Description

Current logic of active tab selection assumed the path will begin with the path of the tab, but in MULTI_ORG the path begins with the org slug.

### DIFF
--- a/client/app/pages/data-sources/DataSourcesList.jsx
+++ b/client/app/pages/data-sources/DataSourcesList.jsx
@@ -149,7 +149,6 @@ const DataSourcesListPage = wrapSettingsTab(
     permission: "admin",
     title: "Data Sources",
     path: "data_sources",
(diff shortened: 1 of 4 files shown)


kravets-levko:
There is simpler solution.

Add this import to `client/app/components/SettingsWrapper.jsx`:
```js
import { stripBase } from "./Router";
```

And use it to prepare argument for `settingsMenu.getActiveItem`

https://github.com/getredash/redash/blob/ae9bbe25e5bc9d6c12fb5ac61cca706a93de7db4/client/app/components/SettingsWrapper.jsx#L12-L14

Probably a proper solution will be to change `location` service to strip base by default, but this should be fine for now.

arikfr:
@kravets-levko thanks! Updated the implementation and added a test case.

Issue #4856, opened 2020-05-01
Search in navbar works only for first search term


arikfr:
### Steps to Reproduce

1. Click in search field in the navbar. 
2. Type in a search term and press Enter.
3. Results show up.
4. Type in another search term and press Enter.

Actual: nothing happens.
Expects: results updated for new value.

### Technical details:

* Redash Version: 9.0.0-alpha (master)

PR #4857, opened 2020-05-01
Search in navbar works only for first search term

kravets-levko:

- [x] Bug Fix

Fixes getredash/redash#4856

### DIFF
--- a/client/app/components/items-list/components/Sidebar.jsx
+++ b/client/app/components/items-list/components/Sidebar.jsx
@@ -1,5 +1,5 @@
 import { isFunction, isString, filter, map } from "lodash";
-import React from "react";
+import React, { useState, useCallback, useEffect } from "react";
 import PropTypes from "prop-types";
 import Input from "antd/lib/input";
 import AntdMenu from "antd/lib/menu";
@@ -11,14 +11,26 @@ import TagsList from "@/components/TagsList";
  */
 
 export function SearchInput({ placeholder, value, showIcon, onChange }) {
+  const [currentValue, setCurrentValue] = useState(value);
+
+  useEffect(() => {
+    setCurrentValue(value);
+  }, [value]);
+
+  const onInputChange = useCallback((event) => {
+    const newValue = event.target.value;
+    setCurrentValue(newValue);
+    onChange(newValue)
+  }, [onChange]);
+
   const InputControl = showIcon ? Input.Search : Input;
   return (
     <div className="m-b-10">
       <InputControl
         className="form-control"
         placeholder={placeholder}
-        defaultValue={value}
-        onChange={event => onChange(event.target.value)}
+        value={currentValue}
+        onChange={onInputChange}
       />
     </div>
   );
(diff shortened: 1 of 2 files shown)


client/app/pages/queries-list/QueriesList.jsx
```diff
@@ -86,6 +87,22 @@ class QueriesList extends React.Component {
     }),
   ];
 
+  componentDidMount() {
```
arikfr, line 90: Shouldn't the component be re-rendered when the URL updates?

kravets-levko: Well, no. Router will recreate controller only if path changes. Watching for url parameters is responsibility of each controller

arikfr: This is not how I would expect a router to work in React and in general it leads to repeating this logic anywhere where we might be using query string parameters. 

I would've expected the page to receive the parameters as props, which would trigger React's rendering when/if needed.

Was there some explicit reason to make it work this way?

client/app/components/items-list/components/Sidebar.jsx
```diff
@@ -11,15 +11,25 @@ import TagsList from "@/components/TagsList";
...
+  }, [value]);
+
+  const onInputChange = useCallback(
+    event => {
+      const newValue = event.target.value;
+      setCurrentValue(newValue);
+      onChange(newValue);
+    },
+    [onChange]
+  );
```
arikfr, line 27: Why is this needed? It seems that if we set `InputControl`'s `value` as the received `value` prop it will work. Does it create a loop in some cases or something like this?

client/app/pages/queries-list/QueriesList.jsx
arikfr: (this is a general discussion, not something necessary to address in this PR)

client/app/components/items-list/components/Sidebar.jsx
kravets-levko: `onChange` is debounced, so if passing `value` prop directly to `Input` component it will update with delay (user will type and see typed text only after they stop typing). This is quite common pattern - use local state to reflect typed text immediately, and sync it with prop when prop changes. Also, `defaultValue` doesn't work in this case because it will be used only when `Input` is mounted and all subsequent updates will be ignored.

client/app/pages/queries-list/QueriesList.jsx
kravets-levko: > Was there some explicit reason to make it work this way?

That was a sort of optimization - to make sure that router will reload controllers only when needed. At that point, it seemed a good idea to bind controllers only to URL path so if path doesn't change - the same controller is kept. But now I agree with your point and I think we need to revisit this mechanism. It will require some changes to `ItemsList` component as I guess only controllers based on it use URL parameters. I'll create an issue for this under a `tech debt` label

kravets-levko: Issue getredash/redash#4864

PR #4860, opened 2020-05-03
Add blob: to allowed img-src rules in CSP

arikfr:

- [x] Bug Fix

## Description

This is needed for Plotly download PNG feature to work.


susodapop:
What broke it? PNG download has been working for me until now.

kravets-levko:
For me it worked only on app.redash.io; on _all_ other instances (demo, previews, local) Chrome and FF blocked download

arikfr:
Maybe there was a change in how Plotly generates the image in a recent version (switching from `data:` to `blob:` maybe?).

PR #4843, opened 2020-04-28
Skip building frontend when APP_ENV=dev

koooge:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Query Runner (Data Source)
- [ ] New Alert Destination
- [ ] Other

## Description
Skip building frontend materials when APP_ENV=dev with BuildKit

```
$ make compose_build
COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose build
redis uses an image, skipping
postgres uses an image, skipping
email uses an image, skipping
WARNING: Native build is an experimental feature and could change at any time
Building server
[+] Building 325.0s (17/17) FINISHED                                                                                                                                    
 => [internal] load .dockerignore                                                                                                                                  0.0s
 => => transferring context: 35B                                                                                                                                   0.0s
 => [internal] load build definition from Dockerfile                                                                                                               0.0s
 => => transferring dockerfile: 2.62kB                                                                                                                             0.0s
 => [internal] load metadata for docker.io/library/python:3.7-slim                                                                                                 0.0s
 => CACHED [server 1/6] FROM docker.io/library/python:3.7-slim                                                                                                     0.0s
 => CACHED https://databricks.com/wp-content/uploads/2.6.10.1010-2/SimbaSparkODBC-2.6.10.1010-2-Debian-64bit.zip                                                   0.0s
 => [internal] load build context                                                                                                                                  0.3s
 => => transferring context: 94.95kB                                                                                                                               0.2s
 => [server 2/6] RUN useradd --create-home redash                                                                                                                  0.5s
 => [server 3/6] RUN apt-get update &&   apt-get install -y     curl     gnupg     build-essential     pwgen     libffi-dev     sudo     git-core     wget     l  55.5s
 => [server 4/6] ADD https://databricks.com/wp-content/uploads/2.6.10.1010-2/SimbaSparkODBC-2.6.10.1010-2-Debian-64bit.zip /tmp/simba_odbc.zip                     0.1s 
 => [server 5/6] RUN unzip /tmp/simba_odbc.zip -d /tmp/   && dpkg -i /tmp/SimbaSparkODBC-2.6.10.1010-2-Debian-64bit/simbaspark_2.6.10.1010-2_amd64.deb   && echo   1.2s 
 => [server 6/6] WORKDIR /app                                                                                                                                      0.0s 
[...]

PR #4879, opened 2020-05-11
Make frontend build in Docker image optional

arikfr:

- [x] Other

## Description

A different approach to making frontend build optional by using a build arg. While a bit more hacky, it's fully supported across different Docker versions and requires less changes.

#4843, #4292

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,7 +71,6 @@ ENV PIP_NO_CACHE_DIR=1
 # We first copy only the requirements file, to avoid rebuilding on every file
 # change.
 COPY requirements.txt requirements_bundles.txt requirements_dev.txt requirements_all_ds.txt ./
-RUN pip install -r requirements.txt -r requirements_dev.txt
 RUN if [ "x$skip_dev_deps" = "x" ] ; then pip install -r requirements.txt; else pip install -r requirements.txt -r requirements_dev.txt; fi
 RUN if [ "x$skip_ds_deps" = "x" ] ; then pip install -r requirements_all_ds.txt ; else echo "Skipping pip install -r requirements_all_ds.txt" ; fi
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN if [ "x$skip_frontend_build" = "x" ] ; then npm ci --unsafe-perm; fi
 
 COPY client /frontend/client
 COPY webpack.config.js /frontend/
-RUN if [ "x$skip_frontend_build" = "x" ] ; then npm run build; else mkdir /frontend/client/dist && touch /frontend/client/dist/{multi_org.html,index.html}; fi
+RUN if [ "x$skip_frontend_build" = "x" ] ; then npm run build; else mkdir /frontend/client/dist && touch /frontend/client/dist/multi_org.html && touch /frontend/client/dist/index.html; fi
 
 FROM python:3.7-slim
 
@@ -18,8 +18,6 @@ EXPOSE 5000
 
 # Controls whether to install extra dependencies needed for all data sources.
 ARG skip_ds_deps
-# Controls whether to build the frontend assets
-ARG skip_frontend_build
 
 RUN useradd --create-home redash
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,7 +71,7 @@ ENV PIP_NO_CACHE_DIR=1
 # We first copy only the requirements file, to avoid rebuilding on every file
 # change.
 COPY requirements.txt requirements_bundles.txt requirements_dev.txt requirements_all_ds.txt ./
-RUN if [ "x$skip_dev_deps" = "x" ] ; then pip install -r requirements.txt; else pip install -r requirements.txt -r requirements_dev.txt; fi
+RUN if [ "x$skip_dev_deps" = "x" ] ; then pip install -r requirements.txt -r requirements_dev.txt; else pip install -r requirements.txt; fi
 RUN if [ "x$skip_ds_deps" = "x" ] ; then pip install -r requirements_all_ds.txt ; else echo "Skipping pip install -r requirements_all_ds.txt" ; fi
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,20 +1,25 @@
...
 WORKDIR /frontend
 COPY package.json package-lock.json /frontend/
 COPY viz-lib /frontend/viz-lib
-RUN npm ci --unsafe-perm
+RUN if [ "x$skip_frontend_build" = "x" ] ; then npm ci --unsafe-perm; fi
 
 COPY client /frontend/client
 COPY webpack.config.js /frontend/
-RUN npm run build
+RUN if [ "x$skip_frontend_build" = "x" ] ; then npm run build; else mkdir /frontend/client/dist && touch /frontend/client/dist/{multi_org.html,index.html}; fi
```
arikfr, line 13: Create dummy `index.html` and `multi_org.html` to make backend tests work.

Dockerfile
```diff
@@ -1,20 +1,25 @@
...
 
 FROM python:3.7-slim
 
 EXPOSE 5000
 
 # Controls whether to install extra dependencies needed for all data sources.
 ARG skip_ds_deps
+# Controls whether to build the frontend assets
+ARG skip_frontend_build
```
arikfr, line 22: This isn't used and will be removed in a future commit.

docker-compose.yml
```diff
@@ -1,12 +1,18 @@
...
+    context: .
+    args:
+      skip_frontend_build: "true"
   volumes:
     - .:/app
 x-redash-environment: &redash-environment
+  REDASH_LOG_LEVEL: "INFO"
   REDASH_REDIS_URL: "redis://redis:6379/0"
+  REDASH_DATABASE_URL: "postgresql://postgres@postgres/postgres"
+  REDASH_RATELIMIT_ENABLED: "false"
```
arikfr, line 15: This isn't directly related, but used the opportunity to reduce even more duplicates.

koooge:
I think it's a reasonable proposal. On another note, my works care about no-include requirements_dev.txt for production build. Is it possible in this PR?

Anyway, Makefile is no longer update? #3038

arikfr:
> I think it's a reasonable proposal. On another note, my works care about no-include requirements_dev.txt for production build. Is it possible in this PR?

https://github.com/getredash/redash/pull/4879/commits/202e1da0f126feb5bb2b676f7bdf72fb676a80fd

> By the way, Makefile is no longer update? #3038

I never used the Makefile... It's possible it's a bit out of sync.

Dockerfile
```diff
@@ -70,6 +72,7 @@ ENV PIP_NO_CACHE_DIR=1
 # change.
 COPY requirements.txt requirements_bundles.txt requirements_dev.txt requirements_all_ds.txt ./
 RUN pip install -r requirements.txt -r requirements_dev.txt
```
koooge, line 74: Please delete this line.

arikfr: Oops. 🤦 

```suggestion
```

docker-compose.yml
jezdez: I think we had to fix the scheduler container for this anyway in our recent rebase: https://github.com/mozilla/redash/commit/ce5e60672f1b49260f5f8b7060dd029b4cac93f2

jezdez:
@arikfr While we're there, can we add a build argument to install an optional requirements file `requirements_ext.txt` if it exists? That would fix #2810.

Dockerfile
```diff
@@ -66,7 +71,7 @@ ENV PIP_NO_CACHE_DIR=1
 # We first copy only the requirements file, to avoid rebuilding on every file
 # change.
 COPY requirements.txt requirements_bundles.txt requirements_dev.txt requirements_all_ds.txt ./
-RUN pip install -r requirements.txt -r requirements_dev.txt
+RUN if [ "x$skip_dev_deps" = "x" ] ; then pip install -r requirements.txt; else pip install -r requirements.txt -r requirements_dev.txt; fi
```
arikfr, line 74:
```suggestion
RUN if [ "x$skip_dev_deps" = "x" ] ; then pip install -r requirements.txt -r requirements_dev.txt; else pip install -r requirements.txt; fi
```

arikfr:
> @arikfr While we're there, can we add a build argument to install an optional requirements file `requirements_ext.txt` if it exists? That would fix #2810.

No need for an arg, checking if file exists is enough. But I'm not sure about the usefulness of it. It saves the person to write a single command (`RUN pip install...`) and requires them to understand how the Dockerfile is built... Is it really worth the hassle?

jezdez:
> > @arikfr While we're there, can we add a build argument to install an optional requirements file `requirements_ext.txt` if it exists? That would fix #2810.
> 
> No need for an arg, checking if file exists is enough. But I'm not sure about the usefulness of it. It saves the person to write a single command (`RUN pip install...`) and requires them to understand how the Dockerfile is built... Is it really worth the hassle?

It would be more than just `RUN pip install..` since when inheriting from the Redash image you would have to also switch to the root user before to make sure the extensions are installed in the same location as the rest of the requirements. And then switch back to the `redash` user, which isn't ideal.

```Dockerfile
FROM redash/redash:preview

COPY requirements_ext.txt .
USER root
RUN pip install -r requirements_ext.txt
USER redash
```

But I know realize that there is the devil in the details, since I think ONBUILD is need to make this work in a separate Dockerfile to make sure it happens at the correct time. And ONBUILD isn't enough for frontend extensions to be installed before the webpack bundle is built. Hmm..

Maybe we should stop using webpack in favor of build-less packaging for the frontend assets like [Snowpack](https://snowpack.dev)?

arikfr:
> Maybe we should stop using webpack in favor of build-less packaging for the frontend assets like [Snowpack](https://snowpack.dev)?

Snowpack doesn't require build during development, but you still need to build for production/deployment.

PR #4882, opened 2020-05-11
Oracle: Encoding fix

gabrieldutra:

- [x] Bug Fix

### DIFF
--- a/redash/query_runner/oracle.py
+++ b/redash/query_runner/oracle.py
@@ -163,7 +163,7 @@ class Oracle(BaseSQLQueryRunner):
                 json_data = json_dumps(data)
                 connection.commit()
         except cx_Oracle.DatabaseError as err:
-            error = "Query failed. {}.".format(err.message)
+            error = "Query failed. {}.".format(str(err))
             json_data = None
         except (KeyboardInterrupt, JobTimeoutException):
             connection.cancel()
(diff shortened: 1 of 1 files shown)


redash/query_runner/oracle.py
```diff
@@ -141,10 +152,7 @@ def run_query(self, query, user):
                         for i in cursor.description
                     ]
                 )
-                rows = [
-                    dict(zip((column["name"] for column in columns), row))
-                    for row in cursor
-                ]
+                rows = [dict(zip((c["name"] for c in columns), row)) for row in cursor]
```
susodapop, line 155: This feels harder to read than necessary.

```suggestion
                columns = [i.get("name") for i in columns]
                rows = [
                    {field: value for field, val in zip(columns, row)} for row in cursor
                ]
```

gabrieldutra: Black formatter is to blame on that 😬

redash/query_runner/oracle.py
```diff
@@ -155,12 +163,13 @@ def run_query(self, query, user):
                 json_data = json_dumps(data)
                 connection.commit()
         except cx_Oracle.DatabaseError as err:
-            error = "Query failed. {}.".format(str(err))
+            error = "Query failed. {}.".format(err.message)
```
arikfr, line 166:
```suggestion
            error = "Query failed. {}.".format(str(err))
```

PR #4884, opened 2020-05-12
Add additional statsd metrics for worker/scheduler

p-yang:

- [x] Other

## Description

Adds additional statsd metrics for created, started, running (in-flight), and finished/failed jobs on RQ workers, and adds some rudimentary tests for existing flask and database metrics.

### DIFF
--- a/redash/tasks/queries/execution.py
+++ b/redash/tasks/queries/execution.py
@@ -101,7 +101,6 @@ def enqueue_query(
                 )
 
                 logger.info("[%s] Created new job: %s", query_hash, job.id)
-                statsd_client.incr("rq.jobs.created.{}".format(queue_name))
                 pipe.set(
                     _job_lock_id(query_hash, data_source.id),
                     job.id,
--- a/redash/tasks/queries/execution.py
+++ b/redash/tasks/queries/execution.py
@@ -7,7 +7,7 @@ from rq.job import JobStatus
 from rq.timeouts import JobTimeoutException
 from rq.exceptions import NoSuchJobError
 
-from redash import models, redis_connection, settings, statsd_client
+from redash import models, redis_connection, settings
 from redash.query_runner import InterruptException
 from redash.tasks.worker import Queue, Job
 from redash.tasks.alerts import check_alerts_for_query
--- a/redash/tasks/queries/execution.py
+++ b/redash/tasks/queries/execution.py
@@ -7,7 +7,7 @@ from rq.job import JobStatus
 from rq.timeouts import JobTimeoutException
 from rq.exceptions import NoSuchJobError
 
-from redash import models, redis_connection, settings
+from redash import models, redis_connection, settings, statsd_client
 from redash.query_runner import InterruptException
 from redash.tasks.worker import Queue, Job
 from redash.tasks.alerts import check_alerts_for_query
@@ -101,6 +101,7 @@ def enqueue_query(
                 )
 
                 logger.info("[%s] Created new job: %s", query_hash, job.id)
+                statsd_client.incr("rq.jobs.created.{}".format(queue_name))
                 pipe.set(
                     _job_lock_id(query_hash, data_source.id),
                     job.id,
--- a/redash/tasks/schedule.py
+++ b/redash/tasks/schedule.py
@@ -17,6 +17,7 @@ from redash.tasks import (
     purge_failed_jobs,
     version_check,
     send_aggregated_errors,
+    Queue
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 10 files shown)


redash/tasks/queries/execution.py
```diff
@@ -101,6 +101,7 @@ def enqueue_query(
                 )
 
                 logger.info("[%s] Created new job: %s", query_hash, job.id)
+                statsd_client.incr("rq.jobs.created.{}".format(queue_name))
```
arikfr, line 104: This will cover query executions and the scheduler change will cover scheduled jobs. But there are still some adhoc jobs (like sending emails) that won't be covered. Maybe we can subclass `Queue` to instrument all job creations?

p-yang: Nice, that would put all of the monitoring code into `redash/tasks/worker.py` as well. That just makes too much sense 😄

p-yang:
@arikfr Added a new subclass for `Queue` as well as `rq_job`, since we need to overload the `queue_class` there.

redash/tasks/worker.py
```diff
@@ -22,10 +23,50 @@ def is_cancelled(self):
...
+        statsd_client.incr("rq.jobs.running.{}".format(queue.name))
+        statsd_client.incr("rq.jobs.started.{}".format(queue.name))
+        try:
+            super().execute_job(job, queue)
+        finally:
+            statsd_client.decr("rq.jobs.running.{}".format(queue.name))
+            if job.get_status() == JobStatus.FINISHED:
+                statsd_client.incr("rq.jobs.finished.{}".format(queue.name))
+            else:
+                statsd_client.incr("rq.jobs.failed.{}".format(queue.name))
```
jezdez, line 67: There are other statuses in rq that this should be tracking, so I would suggest to make the condition:

 ```suggestion
            if job.get_status() == JobStatus.FAILED:
                statsd_client.incr("rq.jobs.failed.{}".format(queue.name))
            elif job.get_status() == JobStatus.DEFERRED:
                statsd_client.incr("rq.jobs.deferred.{}".format(queue.name))
            # TODO: add more status conditions here
            # elif ...
            else:
                statsd_client.incr("rq.jobs.finished.{}".format(queue.name))
```

BUT we

Comment (p-yang) [reply redash/tasks/worker.py:67]: Are those other statuses terminal? I was intending to only track terminal statuses in the metrics, and classify anything left in statuses like `STARTED` or `QUEUED` to be a "failed" job, since by my eye that's how the `monitor_work_horse` function seems to treat those jobs.

File: redash/tasks/queries/execution.py [hunk 3]
```diff
@@ -7,7 +7,7 @@
 from rq.timeouts import JobTimeoutException
 from rq.exceptions import NoSuchJobError
 
-from redash import models, redis_connection, settings
+from redash import models, redis_connection, settings, statsd_client
```
arikfr, line 10: leftover?

redash/tasks/schedule.py
```diff
@@ -21,7 +21,18 @@
...
-rq_scheduler = Scheduler(
+
+class StatsdRecordingScheduler(Scheduler):
+    """
+    RQ Scheduler Mixin that overrides `enqueue_job` to increment/modify metrics via Statsd
+    """
+
+    def enqueue_job(self, job):
+        super().enqueue_job(job)
+        statsd_client.incr("rq.jobs.created.{}".format(self.queue_name))
```
arikfr, line 32: Is this still needed now that we override `enqueue_call`?

redash/tasks/worker.py
```diff
@@ -22,10 +23,50 @@ def is_cancelled(self):
...
 
+class StatsdRecordingQueue(BaseQueue):
+    """
+    RQ Queue Mixin that overrides `enqueue_call` to increment metrics via Statsd
+    """
+
+    def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
+                     result_ttl=None, ttl=None, failure_ttl=None,
+                     description=None, depends_on=None, job_id=None,
+                     at_front=False, meta=None):
```
arikfr, line 34: You can do a catch all:

```suggestion
    def enqueue_call(self, *args, **kwargs):
```

This is more future proof, as you don't need to update when the parameters change.

Then the next line becomes:

```
job = super().enqueue_call(*args, **kwargs)
```

redash/tasks/worker.py
```diff
@@ -22,10 +23,50 @@ def is_cancelled(self):
...
+        job = super().enqueue_call(
+            func, args, kwargs, timeout,
+            result_ttl, ttl, failure_ttl, description,
+            depends_on, job_id, at_front, meta
+        )
```
arikfr, line 39:
```suggestion
        job = super().enqueue_call(*args, **kwargs)
```

redash/tasks/queries/execution.py
p-yang: yep thanks 👍

redash/tasks/worker.py
p-yang: ah yes python! thanks!

redash/tasks/schedule.py
p-yang: It was, but it was because I didn't set `queue_class = RedashQueue` and was overriding the wrong fxn in `Queue` (`enqueue_call` instead of `enqueue_job`). Should be fixed now!

redash/tasks/worker.py
jezdez: So DEFERRED means if a job has a dependency then it's not being processed if the job being depended upon has not been finished yet. So it's not "failed" but just unprocessed yet. @arikfr I would wait for this to be fixed before merging.

PR #4899, opened 2020-05-18
Fix: table viz crashing when search is enabled

arikfr:

- [x] Bug Fix
- [x] Feature

## Description

I managed to reproduce this only on a private deployment of Redash and locally, but it doesn't happen on redash-preview. Might be due to a recent change in Ant library?

It looks like the second argument of setValue is a callback that is being passed to setState. By passing an object there, we trigger the following error:

> Invalid argument passed as callback. Expected a function. Instead received: [object Object]

Removing the second argument fixes this and seems not to affect functionality.

#### Update (by @kravets-levko)

After out discussion with @arikfr we decided that search input shouldn't clear when data changes, and filter should stay applied. I decided to not open a new PR and added fix to this one, but kept original PR description for better context.

### DIFF
--- a/viz-lib/src/visualizations/table/Renderer.jsx
+++ b/viz-lib/src/visualizations/table/Renderer.jsx
@@ -1,5 +1,6 @@
 import { filter, map, get, initial, last, reduce } from "lodash";
 import React, { useMemo, useState, useRef, useCallback, useEffect } from "react";
+import PropTypes from "prop-types";
 import Table from "antd/lib/table";
 import Input from "antd/lib/input";
 import Icon from "antd/lib/icon";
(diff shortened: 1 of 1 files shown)


viz-lib/src/visualizations/table/Renderer.jsx
```diff
@@ -103,7 +103,7 @@ export default function Renderer({ options, data, context }) {
     // so let's use this "hack" for better performance.
```
arikfr, line 103: @kravets-levko I wonder if this hack is needed at all... was it implemented following an actual issue or just a premature optimization?

arikfr:
Confirmed that this change to `setValue` was introduced in https://github.com/ant-design/ant-design/pull/19310.

viz-lib/src/visualizations/table/Renderer.jsx
kravets-levko: It was implemented to handle search in large tables. At that moment it looked like a good solution, however, now I guess it was possible to implement it in more clear way. If you want - I can revisit it, just please create an issue and assign it to me. Thanks!

kravets-levko:
On preview instance it still fails (now with `TypeError: Cannot read property 'type' of undefined`)

image

arikfr:
> On preview instance it still fails (now with TypeError: Cannot read property 'type' of undefined)

That's probably preview instance still using older Ant version... I guess we need to trigger rebuild with no cache to reproduce this.

kravets-levko:
Looks that there is the same issue on cypress :thinking:

kravets-levko:
@arikfr Now it's my time to request a review :sweat_smile: I checked the code and managed to remove that hack and replaced it with code that uses only documented features (so it wont break on next upgrade).

viz-lib/src/visualizations/table/Renderer.jsx
```diff
@@ -51,7 +52,33 @@ function SearchInputInfoIcon({ searchColumns }) {
   );
 }
 
-const SearchInput = React.forwardRef(({ searchColumns, ...props }, ref) => {
+// Some explanation why this weird solution with `clearCallbackRef` is used.
+// When visualization options are editing, renderer may re-create table columns quite often.
+// Search input is one of table columns, and it does not depend on options.
+// We could just render it once and then leave it for React, but there is a feature
+// that needs access to Search input's value: when SOME options or data columns change - search
+// input should reset. If we'll use managed input's state in renderer itself - it will create
```
arikfr, line 60: If we don't clear it, will it filter the updated table?

kravets-levko: Do you mean when query returns new data? Well, it _should_. Currently it has behavior fully kept from Angular version: when data changes - search should be reset. In this case it's an optimization to update/reset it only when relevant part of data changes, otherwise it will re-render entire table on every typed character.

arikfr: We should just keep the filtering on even if the data (or other options) change. This will simplify the code here and probably provide a more expected experience for the user.

kravets-levko: Okay. Should I implement it in this PR or better open a new one?

arikfr: It basically replaces this one, so whatever makes more sense to you.

kravets-levko: @arikfr  https://github.com/getredash/redash/pull/4899/commits/f2358e6904251e0f3f29e5bd494168177fb5f6e2

arikfr:
👍

Issue #4902, opened 2020-05-19
Improve simplejson serialization


chulucninh09:
https://github.com/getredash/redash/blob/a2217cc4ecfa5cd916c934c19a93848284cf984b/redash/utils/__init__.py#L71

Hi, I encountered an odd issue when tried to query my MongoDB database, that JSON.parse of axios kept telling me that it cannot parse json. After digging into the payload, it turned out python implements `nan` as float, and serialize it like a number, without double quotes.

![image](https://user-images.githubusercontent.com/43906170/82340603-9e5d5c80-9a19-11ea-932e-c394f328e49e.png)

For this problem, I think the best way is to add more handle to the JSONEncoder class to handle `nan` "float" type.
```python
        elif math.isnan(o):
            result = None
```

I will try first then make a PR

kravets-levko:
I guess getredash/redash#4903 fixes this, so closing. Thank you @chulucninh09!

PR #4903, opened 2020-05-19
Fix: front-end error when parse python float nan

chulucninh09:

- [] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Change simplejson encoder `ignore_nan` to `True` to overcome json parse issue in front-end when data returned has float `nan` value in it

https://github.com/getredash/redash/issues/4902

### DIFF
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -70,7 +70,13 @@ def generate_token(length):
 
 class JSONEncoder(simplejson.JSONEncoder):
     """Adapter for `simplejson.dumps`."""
-
+    
+    def __init__(self, *args, **kwargs):
+        super().__init__()
+        # Float value nan or inf in Python should be render to None or null in json.
+        # Using ignore_nan = False will make Python render nan as NaN, leading to parse error in front-end
+        self.ignore_nan = True
+    
     def default(self, o):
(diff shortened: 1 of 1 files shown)


redash/utils/__init__.py
```diff
@@ -70,7 +70,13 @@ def generate_token(length):
 
 class JSONEncoder(simplejson.JSONEncoder):
     """Adapter for `simplejson.dumps`."""
-
+    
+    def __init__(self, *args, **kwargs):
+        super().__init__()
+        # Float value nan or inf in Python should be render to None or null in json.
+        # Using ignore_nan = False will make Python render nan as NaN, leading to parse error in front-end
+        self.ignore_nan = True
```
arikfr, line 78: Let's make it possible to override + consistent with other options by moving into `json_dumps`:

https://github.com/getredash/redash/blob/37a6e70a4ae06923e156cbb0babf6afae9d933d0/redash/utils/__init__.py#L118-L123

You will just need to add `kwargs.setdefault('ignore_nan', True)` there.

chulucninh09: Thanks for the review. I have re-implemented as suggestion

arikfr:
Thanks 👍

PR #4906, opened 2020-05-21
Show helpful error message when detached data source queried

### LINKED ISSUE #4608: Show a helpful error message when trying to execute queries detached from a data source
When a data source gets deleted, it's queries are still accessible (detailed explanation can be found in #4336) but they error out due to [`data_source` being None](https://github.com/getredash/redash/blob/master/redash/handlers/query_results.py#L65).

We should provide a helpful error message there (perhaps by raising `QueryDetachedFromDataSourceError`).

patrickjwahl:

- [x] Other

## Description
When a data source is deleted, its queries can still be accessed at the /queries/{id} route. If a user tries to execute one of these queries, they get an unspecified 500 error. This change shows a more helpful message indicating the cause of the problem. 

Resolves #4608

### DIFF
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -62,6 +62,9 @@ error_messages = {
 
 
 def run_query(query, parameters, data_source, query_id, max_age=0):
+    if data_source is None:
+        return error_response("The data source for this query no longer exists.")
+        
     if data_source.paused:
         if data_source.pause_reason:
             message = "{} is paused ({}). Please try later.".format(


guidopetri:
@patrickjwahl , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
Co-incidentally, we've already merged another solution to the same problem in #4936.

So, we might as well close this PR directly.

Thanks for trying to fix the problem as well @patrickjwahl. :smile:

Issue #4692, opened 2020-02-27
When resizing chart to a certain size it errors out


arikfr:
### Issue Summary

1. If you resize the visualization pane to it's smallest height on a query page, the chart won't render and raise an error. You can try it here: https://redash-preview.netlify.com/queries/339/source#602.

2. After sizing it back, it still in error state and won't render.

### Technical details:

* Redash Version: v9-alpha
* Browser/OS: Chrome

PR #4907, opened 2020-05-21
 When resizing chart to a certain size it errors out

kravets-levko:

- [x] Bug Fix

## Description

I managed to reproduce this bug for two cases:

1. when width = 20px and height > 0: Plotly throws a string error `"ax.dtick error: Infinity"`. Looks that it's safe to just ignore this error as it happens at very specific size which is almost impossible to get from UI. Also, in this case Plotly will build plot partially.
2. when width <= 3 or height <= 4 (weird numbers, but generally it's zero-size container): Plotly fails to build plot at all. I decided to set minimal width and height for plot to 5px. I also don't know how to get this case using UI, I reproduced it setting container size from WebTools.

Fixes getredash/redash#4692


arikfr:
👍

PR #4910, opened 2020-05-21
Update Query Page shortcuts for MacOS

gabrieldutra:

- [x] Other

## Description
- Add `Ctrl + Enter`;
- Show `Option + Key` when in Mac

--

--


arikfr:
👍

Jakdaw:
This introduced a bug on non-MacOS platforms - since we ended up with mod+enter *AND* ctrl+enter - and these are never de-duped it means that pressing Control-Enter on other platforms results in two actions. There's then a race-condition hitting the API as to whether one or two JobIDs are created (so doubles the backend DB load in many cases) and then results in the browser polling the API for results twice after each delay rather than once; and downloading them twice!
I guess I'll send a pull request to dedupe the shortcodes after they've been translated for the current platform.

PR #4913, opened 2020-05-24
Update antd to v3.26.17

gabrieldutra:

- [x] Other

## Description
The idea of this PR is to update to antd v3.26.17 and correct issues with the version.

In more recent versions they updated the "Danger" button style to be red (previously it was grey) and since we had color updates for that, it started to break.

--

**Before**
Screen Shot 2020-05-24 at 15 24 55

**Now**
Screen Shot 2020-05-24 at 15 24 30

### DIFF
--- a/client/app/assets/less/inc/ant-variables.less
+++ b/client/app/assets/less/inc/ant-variables.less
@@ -1,8 +1,8 @@
 /* --------------------------------------------------------
     Colors
 -----------------------------------------------------------*/
-@lightblue:                             #03A9F4;
-@primary-color:                        #2196F3;
+@lightblue: #03a9f4;
+@primary-color: #2196f3;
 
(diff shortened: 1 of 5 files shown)


client/app/assets/less/inc/ant-variables.less
```diff
@@ -12,41 +12,31 @@
...
+@input-color-placeholder: #b4b4b4;
 @border-radius-base: 2px;
-@border-color-base:                    #E8E8E8;
-
-
-/* --------------------------------------------------------
-    Button
------------------------------------------------------------*/
-@btn-danger-bg: fade(@redash-gray, 10%);
-@btn-danger-border: fade(@redash-gray, 15%);
```
gabrieldutra, line 48: Only removed this part, the rest was on Prettier.

arikfr:
👍

PR #4915, opened 2020-05-25
Fix flaky Map (Markers) tests

kravets-levko:

- [x] Other

## Description

Fix flaky map tests:

- [x] Map (Markers): use fixture for tiles (instead of real map that changes quire frequently) so they will stay the same;
- [x] Choropleth (data flow bug caused weird effects in the interface, making tests flaky)

redashlabs/product#53


arikfr:
👍

PR #4917, opened 2020-05-25
Athena: Make s3_staging_dir an option

kanga333:

- [x] Feature

## Description
If you use a workgroup with output_location set, you can omit the s3_staging_dir setting.

The following is a quote from the [boto3 document](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/athena.html#Athena.Client.start_query_execution).
> To run the query, you must specify the query results location using one of the ways: either for individual queries using either this setting (client-side), or in the workgroup, using WorkGroupConfiguration .

image

We tested that a test connection succeeds even when s3_staging_dir is empty.

### DIFF
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -83,7 +83,7 @@ class Athena(BaseQueryRunner):
                     "default": 5,
                 },
             },
-            "required": ["region", "s3_staging_dir"],
+            "required": ["region"],
             "extra_options": ["glue", "cost_per_tb"],
             "order": [
                 "region",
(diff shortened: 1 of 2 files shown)


kanga333:
Can someone please review it?

susodapop:
Hey @kanga333 thanks for this effort. Sorry for missing your pings on this. Feel free to reopen if you still believe this would be valuable for the Athena query runner. We can get it reviewed and merged pretty quick using our [new process](https://github.com/getredash/redash/blob/master/CONTRIBUTING.md#pull-requests). The only question I have on a quick look here is whether this will break any existing Athena connections (it doesn't seem like it would).

kanga333:
@susodapop Thank you for your reply. This change will not affect existing Athena connections. If s3_staging_dir exists, Athena Runner will continue to use it, and PyAthena will also accept it. This change only makes s3_staging_dir optional.

guidopetri:
@kanga333 , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
This looks like a really simple change, which should be pretty straight forward to fix the merge conflict for.

@kanga333 Any idea if this would still be useful to merge? :smile:

--

On a separate note, it looks like `pyathena` has several major releases in the meantime.  We should probably look at the newer versions once we have things stabilised. :smile:

https://pypi.org/project/pyathena/#history

konnectr:
It sounds like this PR needs to resolve conflicts and test with the database

PR #4920, opened 2020-05-25
Use memoized query result data

gabrieldutra:

- [x] Other

## Description
Reported by @kravets-levko, in a few places we use the `getQueryResultData` without `useMemo`. The problem is that it recreates the object every time the component re-renders, unnecessarily updating everything that depends on this data.

In my understanding the `queryResult` object shouldn't be mutated, so _ideally_ this shouldn't cause any issues. I want to confirm from Cypress + some manual tests, but from code it seemed ok.

#4915 

--

### DIFF
--- a/client/app/lib/getQueryResultData.js
+++ b/client/app/lib/getQueryResultData.js
@@ -1,6 +1,7 @@
+import { useMemo } from "react";
 import { get, invoke } from "lodash";
 
-export default function getQueryResultData(queryResult) {
+export function getQueryResultData(queryResult) {
   return {
     status: invoke(queryResult, "getStatus") || null,
     columns: invoke(queryResult, "getColumns") || [],
@@ -13,4 +14,8 @@ export default function getQueryResultData(queryResult) {
     runtime: invoke(queryResult, "getRuntime") || null,
     metadata: get(queryResult, "query_result.data.metadata", {}),
   };
+}
+
+export function useQueryResultData(queryResult) {
+  return useMemo(() => getQueryResultData(queryResult), [queryResult]);
 }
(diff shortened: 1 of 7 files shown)


client/app/lib/getQueryResultData.js
```diff
@@ -1,6 +1,7 @@
+import { useMemo } from "react";
 import { get, invoke } from "lodash";
 
-export default function getQueryResultData(queryResult) {
+export function getQueryResultData(queryResult) {
```
arikfr, line 4: Do we still use getQueryResultData anywhere? If not, let's change useQueryResultData to be the default export and rename the file.

client/app/lib/getQueryResultData.js
```diff
@@ -14,3 +15,7 @@ export default function getQueryResultData(queryResult) {
     metadata: get(queryResult, "query_result.data.metadata", {}),
   };
 }
+
+export function useQueryResultData(queryResult) {
+  return useMemo(() => getQueryResultData(queryResult), [queryResult]);
```
kravets-levko, line 20: Shouldn't it depend on `queryResult.status` as well? I'll try to explain my concern. We get a new `queryResult` object each time we refresh query. But while query job is running - `queryResult` remains the same and just updates its fields (please correct me if I miss something). So if we'll pass a new `queryResult` somewhere - this hook will cache its data when it's "waiting"/"processing"/etc. and when query result will reach "done" state - `useMemo` will still return the same cached object.

gabrieldutra:
> I have to admit that something about this feels wrong to me, because this is the kind of functions you don't have to memoize, as it's very inexpensive.

I would classify this in the same reason we have `useCallback` (which can be seen as an `useMemo` with a very performant function), it is inexpensive indeed, but the idea is to avoid other stuff that depend on that var to update unnecessarily. Since the queryResults had plenty of dependants, I just memoized on the tree, it felt "right" that the updates were triggered when they should when you see this as a hook.

client/app/lib/getQueryResultData.js
gabrieldutra: I see your concern now. Well, while this shouldn't be an issue as it seem that every place we use that hook we only take "done" queryResults, I also see the potential problem. Will add the dep 👍

PR #4924, opened 2020-05-27
Update requests usages not to allow redirects

gabrieldutra:

- [x] Other

### DIFF
--- a/redash/utils/requests_session.py
+++ b/redash/utils/requests_session.py
@@ -1,0 +1,12 @@
+import requests
+from redash import settings
+
+
+class ConfiguredSession(requests.Session):
+    def request(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


redash/utils/requests_session.py
```diff
@@ -0,0 +1,12 @@
+import requests
+from redash import settings
+
+
+class ConfiguredSession(requests.Session):
+    def request(self, *args, **kwargs):
+        if not settings.REQUESTS_ALLOW_REDIRECTS:
+            kwargs.update({"allow_redirects": False})
+        return super(ConfiguredSession, self).request(*args, **kwargs)
```
jezdez, line 9: Since we're in Python 3 now you can do this:
```suggestion
        return super().request(*args, **kwargs)
```

PR #4930, opened 2020-05-31
Fix: auto hide Plotly Modebar

arikfr:

- [x] Bug Fix

## Description

#4644 introduced a change to previous behavior: if the modebar is enabled it will always be visible instead of autohiding as before. This PR reverts the behavior to the old one.

### DIFF
--- a/viz-lib/src/visualizations/chart/Renderer/PlotlyChart.jsx
+++ b/viz-lib/src/visualizations/chart/Renderer/PlotlyChart.jsx
@@ -36,7 +36,7 @@ export default function PlotlyChart({ options, data }) {
           displaylogo: false,
         };
         
-        if (!visualizationsSettings.hidePlotlyModeBar) {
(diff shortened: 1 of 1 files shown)


viz-lib/src/visualizations/chart/Renderer/PlotlyChart.jsx
```diff
@@ -34,8 +34,11 @@ export default function PlotlyChart({ options, data }) {
         const plotlyOptions = {
           showLink: false,
           displaylogo: false,
-          displayModeBar: !visualizationsSettings.hidePlotlyModeBar,
         };
+        
+        if (!visualizationsSettings.hidePlotlyModeBar) {
```
arikfr, line 39:
```suggestion
        if (visualizationsSettings.hidePlotlyModeBar) {
```

🤦

daniellangnet:
Ahh, sorry about introducing that. Thanks for fixing.

PR #4934, opened 2020-06-01
Return cached data source schema when available

gabrieldutra:

- [x] Other

## Description
This PR updates the Data Source schema handler to immediately return cached results when they are available.

This is a follow up for #4797 

--

### DIFF
--- a/client/app/pages/queries/hooks/useDataSourceSchema.js
+++ b/client/app/pages/queries/hooks/useDataSourceSchema.js
@@ -1,11 +1,11 @@
-import { reduce } from "lodash";
+import { reduce, has } from "lodash";
 import { useCallback, useEffect, useRef, useState, useMemo } from "react";
 import { axios } from "@/services/axios";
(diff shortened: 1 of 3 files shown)


client/app/pages/queries/hooks/useDataSourceSchema.js
```diff
@@ -1,36 +1,39 @@
...
+  return new Promise(resolve => setTimeout(resolve, ms));
 }
 
 function getSchema(dataSource, refresh = undefined) {
   if (!dataSource) {
     return Promise.resolve([]);
   }
 
-  const fetchSchemaFromJob = (data) => {
-    return sleep(1000).then(() => {
```
gabrieldutra, line 17: I also removed the first request sleeping time, from a few tests a good portion of the results would be done already without the sleeping time.

arikfr:
Nice, you can indeed see the improvement even without the backend changes. 👍

PR #4935, opened 2020-06-02
Plotly Charts: use .legend to determine legends size

gabrieldutra:

- [x] Bug Fix

## Description
**The bug** ([Preview](https://redash-preview.netlify.app/queries/373#647))

Screen Shot 2020-06-02 at 16 10 07

Currently Charts that have legends moved to the bottom can "cut" part of their text.

**With this fix** ([Preview](https://deploy-preview-4935--redash-preview.netlify.app/queries/373#647))

Screen Shot 2020-06-02 at 16 14 17

--

See description ☝️

### DIFF
--- a/viz-lib/src/visualizations/chart/plotly/applyLayoutFixes.js
+++ b/viz-lib/src/visualizations/chart/plotly/applyLayoutFixes.js
@@ -117,5 +117,7 @@ export default function applyLayoutFixes(plotlyElement, layout, options, updateP
         break;
       // no default
     }
+  } else {
+    updatePlot(plotlyElement, pick(layout, ["width", "height"]));
   }
 }
--- a/viz-lib/src/visualizations/chart/plotly/applyLayoutFixes.js
+++ b/viz-lib/src/visualizations/chart/plotly/applyLayoutFixes.js
@@ -76,18 +76,8 @@ function placeLegendBelowPlot(plotlyElement, layout, updatePlot) {
     if (legend) {
       // compute real height of legend - items may be split into few columnns,
       // also scrollbar may be shown
-      const bounds = reduce(
-        legend.querySelectorAll(".traces"),
-        (result, node) => {
-          const b = node.getBoundingClientRect();
-          result = result || b;
-          return {
-            top: Math.min(result.top, b.top),
-            bottom: Math.max(result.bottom, b.bottom),
-          };
-        },
-        null
-      );
+      const bounds = legend.querySelector(".bg").getBoundingClientRect();
+
       // here we have two values:
       // 1. height of plot container excluding height of legend items;
       //    it may be any value between 0 and plot container's height;
--- a/viz-lib/src/visualizations/chart/plotly/applyLayoutFixes.js
+++ b/viz-lib/src/visualizations/chart/plotly/applyLayoutFixes.js
@@ -1,25 +1,6 @@
-import { find, pick, reduce } from "lodash";
+import { pick } from "lodash";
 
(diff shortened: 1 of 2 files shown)


kravets-levko:
There was a reason why I implemented this hack instead of using Plotly's features ([query to reproduce](https://deploy-preview-4935--redash-preview.netlify.app/queries/349/source#661)):

image

I'll explain you how coordinates in Plotly work so you don't need to look for in in docs and sources. There is a plot area - it's 1x1 box with coordinates from (0.0; 0.0) to (1.0; 1.0) and everything else is placed outside of it, including axes, legend and so on. Outside of plot there are only margins (that ones we can setup in `layout` object). To make axes/annotations/etc. visible, Plotly modifies margins (yes, it does not treat that objects as a part of plot). If you disable automargins feature - axes may become partially invisible. But the most important thing in our case is that automargins feature does not take in account legend - there is another algorithm. When you place legend in some coordinates outside of **plot** - Plotly will add extra space for it. But I highlighted word "plot" on purpose - automargins algorithm and this one are independent, so legend does not "know" about axes and other stuff. Therefore legend may overlap them (or even plot in some cases).

The idea of my hack was to place horizontal legend in top-left corner of plot so Plotly will not reserve extra space for it, and then measure it and offset using CSS. There is [an issue](https://github.com/plotly/plotly.js/issues/1199) to fix this behavior in Plotly (either by fixing legend positioning algorithm or by allowing to choose coordinate system - relative to plot, or plot with axes or container/absolute). But it's still not implemented :disappointed:

gabrieldutra:
> There was a reason why I implemented this hack instead of using Plotly's features ([query to reproduce](https://deploy-preview-4935--redash-preview.netlify.app/queries/349/source#661)):

I figured this was the case, I could see that happening with the other graph already. I just wondered if they may had added any other feature to cover this, which would've been better to us. 

I'll revert the test commit and go back with the "updating the fix" approach. I'll provide more details on that, but first I'll try to understand better other issues I found: - scrollbox taking part of the plot sometimes; - Plots without legend don't resize.

Thanks for the explanation @kravets-levko 🙂

viz-lib/src/visualizations/chart/plotly/applyLayoutFixes.js
```diff
@@ -76,18 +76,8 @@ function placeLegendBelowPlot(plotlyElement, layout, updatePlot) {
...
-        (result, node) => {
-          const b = node.getBoundingClientRect();
-          result = result || b;
-          return {
-            top: Math.min(result.top, b.top),
-            bottom: Math.max(result.bottom, b.bottom),
-          };
-        },
-        null
-      );
```
gabrieldutra, line 90: @kravets-levko what's causing the bug is that there is some "margin" for the legends background, computing each of the items separately don't take it into account, so it separating less than needed and cutting part of the text. I made some tests and it seems the `.legend` holds the same information we had before + it also considers the margin for the background size.

I'm continuing to test this change, but so far looks fine, lmk if you see any situation where this may cause a problem.

viz-lib/src/visualizations/chart/plotly/applyLayoutFixes.js
```diff
@@ -127,5 +117,7 @@ export default function applyLayoutFixes(plotlyElement, layout, options, updateP
         break;
       // no default
     }
+  } else {
+    updatePlot(plotlyElement, pick(layout, ["width", "height"]));
```
gabrieldutra, line 121: This solves another issue that Charts without legends weren't updating when resized.

PR #4936, opened 2020-06-02
Handle error when data_source is None

### LINKED ISSUE #4608: Show a helpful error message when trying to execute queries detached from a data source

fsr313:
- [X] Bug Fix
## Description
No exception handled when data_source is None and generates error 500.
This change handles that error using a try-except and now excepts a error response with a helpful message about the problem.
resolves #4608


luislizarraga:
Did you add tests?

luislizarraga:
I think that it applies the same comment as the other PR, you can check if data_source is `None` and then return the error message instead of relying on the AttributeError class, which handles a different error (https://docs.python.org/2.7/library/exceptions.html#exceptions.AttributeError)

guidopetri:
@fsr313 , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
@konnectr This PR looks like it'd be useful, and the merge conflict being shown above seems like it shouldn't be hard to solve.

What do you think? :smile:

Issue #4469, opened 2019-12-20
Redash can't create user from Google with long profile_image_url


spacentropy:
### Redash can't create user with long profile_image_url

My coworker tried to login to redash and it failed. 
```
[2019-12-20 14:32:12,927] ERROR in app: Exception on /oauth/google_callback [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python2.7/site-packages/flask_restful/__init__.py", line 271, in error_router
    return original_handler(e)
  File "/usr/local/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/app/redash/authentication/google_oauth.py", line 101, in authorized
    user = create_and_login_user(org, profile['name'], profile['email'], picture_url)
  File "/app/redash/authentication/__init__.py", line 257, in create_and_login_user
    models.db.session.commit()
  File "/usr/local/lib/python2.7/site-packages/sqlalchemy/orm/scoping.py", line 153, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 943, in commit
    self.transaction.commit()
  File "/usr/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 467, in commit
    self._prepare_impl()
  File "/usr/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 447, in _prepare_impl
    self.session.flush()
  File "/usr/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2254, in flush
    self._flush(objects)
  File "/usr/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2380, in _flush
    transaction.rollback(_capture_exception=True)
[...]

PR #4941, opened 2020-06-04
fix google create user with long profile pic url

fsr313:
- [x] Bug Fix

## Description
when creating a new user using google, the image url can be bigger of what the model accepts (320). the error is handled and the model object updated and fixed so it can be commited to the database.
tests, for this were added as well
#4469


luislizarraga:
If you know what the issue is, it would be better to address it directly instead of relying on Exceptions, as the `DataError` could be triggered by any number of things and not just because the profile picture URL is too long.

tim5go:
@fsr313 @luislizarraga 
Can we make it as simple as:
`if len(user_object._profile_image_url) > 320:
   user_object._profile_image_url = None
`
rather than throwing **DataError**

guidopetri:
@fsr313 , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
The concept of this PR sounds like it could be useful.

But looking over the code and the review/suggestion by @luislizarraga, it seems like it needs to be implemented differently?

PR #4945, opened 2020-06-04
Fix wrong Sentry exception messages on invalid parameters

rauchy:
- [x] Bug Fix

## Description
Following #4780 `refresh_queries` reports any queries it fails to refresh and provides the underlying exception details and stack trace. For `InvalidParameterError`, these would result in this:

```
The following parameter values are incompatible with their definitions: C, o, u, l, d,  , n, o, t,  , e, n, q, u, e, u, e,  , q, u, e, r, y,  , 1, 2, 3,  , d, u, e,  , t, o,  , I, n, v, a, l, i, d, P, a, r, a, m, e, t, e, r, E, r, r, o, r, (, ', T, h, e,  , f, o, l, l, o, w, i, n, g,  , p, a, r, a, m, e, t, e, r,  , v, a, l, u, e, s,  , a, r, e,  , i, n, c, o, m, p, a, t, i, b, l, e,  , w, i, t, h,  , t, h, e, i, r,  , d, e, f, i, n, i, t, i, o, n, s, :,  , O, h, , N, o, ', )
```

This is a result of custom message construction done in `InvalidParameterError`'s initializer, which kinda makes no sense to me right now.

#4780

### DIFF
--- a/redash/models/parameterized_query.py
+++ b/redash/models/parameterized_query.py
@@ -130,11 +130,7 @@ class ParameterizedQuery(object):
             key for (key, value) in parameters.items() if not self._valid(key, value)
         ]
         if invalid_parameter_names:
-            raise InvalidParameterError(
-                "The following parameter values are incompatible with their definitions: {}".format(
-                    ", ".join(invalid_parameter_names)
-                )
-            )
+            raise InvalidParameterError(invalid_parameter_names)
         else:
             self.parameters.update(parameters)
             self.query = mustache_render(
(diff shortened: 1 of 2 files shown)


arikfr:
> This is a result of custom message construction done in InvalidParameterError's initializer, which kinda makes no sense to me right now.


The code in #4780 creates a new exception with a new message but with the type of the original exception. InvalidParameterException takes the input it's given and calls `",".join` on it.

So it actually makes perfect sense, but I feel like the issue is in the implementation of #4780 and not this specific exception.

rauchy:
What I meant by "not making sense", is that `InvalidParameterException`'s constructor takes a list of invalid parameters and constructs the message according to them. I just don't know if that's an idiomatic way to construct exception messages in Python, or is the standard to provide a `message` as the primary parameter to an exception class.

The motivation behind the exception cloning implementation in #4780 is mostly to retain the stack trace in Sentry exceptions.

PR #4948, opened 2020-06-06
Custom Query components per Data Source type

gabrieldutra:

- [x] Refactor

## Description

This is an initial draft on an idea to make query components customizable per ds type.

What I initially thought was to use the register pattern we have in other places for this + a hook to get the components.

Opening this Draft to gather comments and suggestions.

Example usage I thought:
```jsx
function CustomQueryEditor(props) {
  // ...
  return <QueryEditor {...props} /* {...customizations} */ />;
}

registerEditorComponent("QueryEditor", CustomQueryEditor, ["mysql", "pg"]);
```

LMK 
--

--


client/app/components/queries/editor-components/editorComponents.js
```diff
@@ -0,0 +1,24 @@
...
+  if (!isArray(dataSourceTypes)) {
+    dataSourceTypes = [dataSourceTypes];
+  }
+
+  each(dataSourceTypes, dataSourceType => {
+    componentsRegistry.set(dataSourceType, { ...componentsRegistry.get(dataSourceType), [componentName]: component });
+  });
+}
+
+export function useEditorComponents(dataSourceType) {
```
arikfr, line 20: Is this a hook so we can use `useMemo`? If it is, why do we need `useMemo` won't it always return the same object (the component class) for the same input?

client/app/components/queries/editor-components/index.js
```diff
@@ -0,0 +1,10 @@
+import SchemaBrowser from "@/components/queries/SchemaBrowser";
+import QueryEditor from "@/components/queries/QueryEditor";
+
+import { registerEditorComponent, useEditorComponents } from "./editorComponents";
+
+// default query editor components
+registerEditorComponent("SchemaBrowser", SchemaBrowser);
+registerEditorComponent("QueryEditor", QueryEditor);
```
arikfr, line 8: The names should be constants defined in `editorComponents`.

gabrieldutra: Good one, done 👍

client/app/components/queries/editor-components/editorComponents.js
gabrieldutra: Mostly to use `useMemo` indeed, the plus is that it seems the "React way" nowadays.

Well, while it should return the same input, objects and arrays always hold different comparison values when you instantiate them, which can trigger updates unnecessarily elsewhere. To make things "safer" I usually define a `useMemo` for non-basic variable types.

arikfr: > the plus is that it seems the "React way" nowadays.

Our code is Javascript first, React second. This to mean that we should use Javascript features first (simple functions) when possible.

> while it should return the same input, objects and arrays always hold different comparison values when you instantiate them, which can trigger updates unnecessarily elsewhere.

But the way these values are used won't make React re-render when they change, so it doesn't matter that we create a new object every time this function is called.

It might seem small, but needlessly adding `useMemo` (and other optimizations) in the code makes the code harder to follow and less performant.

gabrieldutra: > But the way these values are used won't make React re-render when they change, so it doesn't matter that we create a new object every time this function is called.

I thought about it for a while, and you are right, for the way we will use it (destructuring) it shouldn't affect. I was too focused looking only at the function returning an object 😅.

> It might seem small, but needlessly adding useMemo (and other optimizations) in the code makes the code harder to follow and less performant.

I agree and you can be sure that I try not to use it needlessly, the thing that usually triggers me to use a hook instead of a simple JS function is when something can be potentially forgotten to receive the optimization, thus maybe invalidating other existing optimizations in the component. E.g.: a transform function for an user object (`userObject => anotherObject`), it's simple, but the different address can trigger unnecessary updates.

PR #4954, opened 2020-06-07
Fix: sorting queries by schedule was resulting in a wrong order

leini-db:

- [x] Bug Fix

## Description
sorting on schedule was not working. it use schedule which is an object to compare. by default, it's converted to string value. so 1000 is sorted ahead of 60


arikfr:
I was about to merge this, but then I realized that this is only half of the solution: this fixes sorting on the frontend, but the backend sorting remains broken (for similar reasons). You can see what I mean in the preview instance:

https://deploy-preview-4954--redash-preview.netlify.app/queries?order=schedule&page=1&page_size=5

You will notice that when we limit the page size to 5 queries, a query scheduled to refresh every 30 days shows up before queries scheduled to refresh every minute.

I'm not sure how hard it is to fix on the backend, it's possible you can define a `json_cast_property` property ([example](https://github.com/getredash/redash/blob/98a5154345307737c9faa012547c1529a56f2f8d/redash/models/users.py#L102-L104)) that extracts the interval value and then use it for sorting (mapping defined [here](https://github.com/getredash/redash/blob/98a5154345307737c9faa012547c1529a56f2f8d/redash/handlers/queries.py#L40-L41)).

We do the same for the `active_at` value on the users page, so it's likely to work. I just don't know what happens when the schedule value is null instead of a json object.

leini-db:
oops, Let me look into this.

Lei

redash/models/types.py
```diff
@@ -108,3 +110,17 @@ def __init__(self, cast_type, *args, **kwargs):
     def expr(self, model):
         expr = super(json_cast_property, self).expr(model)
         return expr.astext.cast(self.cast_type)
+
+
+class pseudo_json_cast_property(index_property):
+    """
+    A SQLAlchemy index property that is able to cast the
+    entity attribute as the specified cast type. Useful
+    for JSON and JSONB colums for easier querying/filtering.
```
arikfr, line 119:
```suggestion
    A SQLAlchemy index property that is able to cast the
    entity attribute as the specified cast type. Useful
    for PseudoJSON colums for easier querying/filtering.
```

arikfr:
Merged 👍 Congrats on your first PR :-)

PR #4958, opened 2020-06-08
Refine DynamicComponent API

kravets-levko:

- [x] Refactor

## Description

I had few concerns regarding existing DynamicComponent API: using `name` property for choosing component by key was typo-prone and shadowed `name` prop itself (rest props are passed to underlying component). Using separate function to register components was also not that perfect.

This proposal changes DynamicComponent API to allow direct access to dynamic components by their identifier using dot notation (e.g. `<DynamicComponent.HomeExtra ...>`). ~~All dynamic components should be explicitly defined now, so no any possibility to register arbitrary component or render non-existing component (it will trigger error in console - fail-fast).~~ Also, same way is used for registering (overriding) components: `DynamicComponent.HomeExtra = CustomHomeExtraComponent`

### DIFF
--- a/client/app/components/DynamicComponent.jsx
+++ b/client/app/components/DynamicComponent.jsx
@@ -1,35 +1,20 @@
-import { isFunction, each } from "lodash";
+import { isFunction } from "lodash";
 import React, { useState, useEffect } from "react";
 
-// Introduce new dynamic components here; use `null` for default representation (display children)
-const dynamicComponents = {
-  HomeExtra: null,
-  BeaconConsent: null,
-  BeaconConsentSetting: null,
-  UsersListExtra: null,
-  DataSourcesListExtra: null,
-  CreateDashboardDialogExtra: null,
-  HelpDrawerExtraContent: null,
-};
-
-// Overriding dynamic components is simple: just assign a value to one of the fields above, e.g.:
-//
-// import DynamicComponent from "@/components/DynamicComponent";
-// DynamicComponent.HomeExtra = CustomHomeExtraComponent;
+const dynamicComponents = {};
 
 const updateListeners = new Set();
 
-// Patch components registry: convert each entry to getter + setter
-each(dynamicComponents, (Component, name, target) => {
-  function DynamicComponent({ children, ...props }) {
+function createDynamicComponent(name) {
+  function Component({ children, ...props }) {
     // use state to force update component when underlying component has changed
-    const [RealComponent, setRealComponent] = useState(() => Component);
+    const [RealComponent, setRealComponent] = useState(() => dynamicComponents[name]);
 
     // listen for changes and re-render when needed
     useEffect(() => {
       function update(changedComponentName) {
         if (changedComponentName === name) {
-          setRealComponent(() => Component);
+          setRealComponent(() => dynamicComponents[name]);
         }
       }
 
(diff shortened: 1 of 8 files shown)


kravets-levko:
@arikfr @gabrieldutra Also, I'm not sure how to be with locking a dynamic component registry from adding arbitrary components - it may be useful in some cases. So if you think it should be less strict - lmk and I'll update implementation.

jezdez:
We're now using an extra dynamic component in our fork (https://github.com/mozilla/redash/commit/5dd5226dd1f4205930262b603a0fd1853f1a0fbe, https://github.com/mozilla/redash/commit/0b3fd5ad89999fa1ec7c93136fa4c57fcd616113) to implement extra metadata shown about [data sources next to the data source select](https://github.com/mozilla/redash-stmo/blob/121b19d5240414c2d5439de2cf9b1da1147bce5c/src/redash_stmo/data_sources/details/bundle/DataSourceDetails.jsx) and to add new menu items.

Sometimes we wanted to extend dropdowns, and I'm not sure [if that's okay to do](https://github.com/mozilla/redash/blob/0b3fd5ad89999fa1ec7c93136fa4c57fcd616113/client/app/components/EditVisualizationButton/QueryControlDropdown.jsx#L61-L66)?

I was wondering how to handle cases in which we have multiple separate components that want to extend one dynamic component at the same time? Using a property for this would preclude using a list for the registry, would providing an optional `WrappingComponent` or something make sense that is capable of rendering multiple child components? Is that good practice for Ract?

jezdez:
> @arikfr @gabrieldutra Also, I'm not sure how to be with locking a dynamic component registry from adding arbitrary components - it may be useful in some cases. So if you think it should be less strict - lmk and I'll update implementation.

I would like to discourage the use of a lock here since it would prevent extensions to use the same dynamic component system for their own extensibility -- which would defeat the purpose of having importable components in the first place. In other words, please don't start marking some API as private if the code can be reused elsewhere.

client/app/components/DynamicComponent.jsx
```diff
@@ -1,50 +1,65 @@
...
+
+// Overriding dynamic components is simple: just assign a value to one of the fields above, e.g.:
+//
+// import DynamicComponent from "@/components/DynamicComponent";
+// DynamicComponent.HomeExtra = CustomHomeExtraComponent;
+
+const updateListeners = new Set();
+
+// Patch components registry: convert each entry to getter + setter
+each(dynamicComponents, (Component, name, target) => {
```
gabrieldutra, line 23: I'd consider a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) not to be doing it with each component + make it possible to add new DynamicComponents from other places.

Con: again, no IE 😅 (suppose it isn't a problem anymore, so I can bring the suggestion up anyways)

gabrieldutra:
> Sometimes we wanted to extend dropdowns, and I'm not sure if that's okay to do?

@jezdez the problem is that by using a `<DynamicComponent />` you'll have a component before the Dropdown items and Ant seems to rely on "listing" all of its children. It causes [this issue](https://codesandbox.io/s/condescending-vaughan-hvpeu). I remember I tried to do sth similar in https://github.com/getredash/redash/blob/56d3be2248dfef1ff4b0ac52f25053db2cc9ba93/client/app/components/dashboards/dashboard-widget/Widget.jsx#L14-L20

but I had to use the those directly.

jezdez:
> > Sometimes we wanted to extend dropdowns, and I'm not sure if that's okay to do?
> 
> @jezdez the problem is that by using a `<DynamicComponent />` you'll have a component before the Dropdown items and Ant seems to rely on "listing" all of its children. It causes [this issue](https://codesandbox.io/s/condescending-vaughan-hvpeu). I remember I tried to do sth similar in
> 
> https://github.com/getredash/redash/blob/56d3be2248dfef1ff4b0ac52f25053db2cc9ba93/client/app/components/dashboards/dashboard-widget/Widget.jsx#L14-L20
> 
> but I had to use the those directly.

Right, https://github.com/react-component/menu/issues/142 seems to point out I think that we would need to pass in the properties of the menu to make this work.

arikfr:
Now we came a full circle and you can still make a typo and everything will be fine. I guess some IDEs will provide autocompletion based on the comment you added to the function?

kravets-levko:
> Now we came a full circle and you can still make a typo and everything will be fine.

Well, that's the conflict between typo-proof solution and ability to register arbitrary components :man_shrugging: I can add a check on attempt to use unregistered component that will show a warning in console (only when `process.env.NODE_ENV = "development").

> I guess some IDEs will provide autocompletion based on the comment you added to the function?

Yes. And some even can find newly assigned fields to the object (not listed in jsdoc).

PR #4959, opened 2020-06-08
Fix Destination not returning custom error message

gabrieldutra:

- [x] Bug Fix

## Description
**Bug**
Creating a destination with an existing name wouldn't return its custom message.

**Solution**
Just returned the full error, so it's handled by `getErrorMessage` later.

Extra: Updated `getErrorMessage` to allow `new Error("Custom message")` that we already use along the way.

--

--

### DIFF
--- a/client/app/components/ApplicationArea/ErrorMessage.jsx
+++ b/client/app/components/ApplicationArea/ErrorMessage.jsx
@@ -21,7 +21,11 @@ export function getErrorMessage(
   if (isObject(error)) {
     // HTTP errors
     if (error.isAxiosError && isObject(error.response)) {
-      return getErrorMessageByStatus(error.response.status, get(error, "response.data.message", defaultMessage));
+      const errorData = get(error, "response.data", {});
+
+      // handle cases where the message is an object as { "message": msg } or { "error": msg }
+      const errorMessage = errorData.message || errorData.error || defaultMessage;
+      return getErrorMessageByStatus(error.response.status, errorMessage);
     }
     // Router errors
     if (error.status) {
--- a/client/app/components/ApplicationArea/ErrorMessage.jsx
+++ b/client/app/components/ApplicationArea/ErrorMessage.jsx
@@ -25,6 +25,10 @@ export function getErrorMessage(error) {
     if (error.status) {
       return getErrorMessageByStatus(error.status, message);
     }
+    // Other Error instances
+    if (error.message) {
+      return error.message;
+    }
   }
   return message;
 }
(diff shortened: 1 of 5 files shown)


client/app/pages/destinations/DestinationsList.jsx
```diff
@@ -57,16 +57,11 @@ class DestinationsList extends React.Component {
...
-    return Destination.create(target)
-      .then(destination => {
-        this.setState({ loading: true });
-        Destination.query().then(destinations => this.setState({ destinations, loading: false }));
-        return destination;
-      })
-      .catch(error => {
-        const message = find([get(error, "response.data.message"), get(error, "message"), "Failed saving."], isString);
-        return Promise.reject(new Error(message));
-      });
```
gabrieldutra, line 69: `getErrorMessage` already handles this

gabrieldutra:
I'll also add a Cypress test to make sure this message is properly returned 🙂

client/app/pages/destinations/DestinationsList.jsx
kravets-levko: But it wont fallback to ""Failed saving.", right? `getErrorMessage` would return `It seems like we encountered an error. Try refreshing this page or contact your administrator.`. Maybe, let's add an optional parameter to `getErrorMessage` for default message which will default to existing string, and can be redefined when needed, WDYT?

gabrieldutra: > But it wont fallback to Failed saving., right?

That didn't seem to be much harm 😅, also it would be in accordance with the Data Source one.

Makes sense to make this message replaceable anyway

kravets-levko: Up to you. I just noticed this, but for me it doesn't feel _that_ important. Rather it could be nice to show "friendlier" error message to user.

gabrieldutra: https://github.com/getredash/redash/pull/4959/commits/828e468a89331673268bee0484a60e8281bd4e76 🙂

client/app/components/ApplicationArea/ErrorMessage.jsx
```diff
@@ -14,19 +14,29 @@ function getErrorMessageByStatus(status, defaultMessage) {
...
+) {
   if (isObject(error)) {
     // HTTP errors
     if (error.isAxiosError && isObject(error.response)) {
-      return getErrorMessageByStatus(error.response.status, get(error, "response.data.message", message));
+      const errorData = get(error, "response.data", {});
+
+      // handle cases where the message is an object as { "message": msg } or { "error": msg }
+      const errorMessage = errorData.message || errorData.error || defaultMessage;
+      return getErrorMessageByStatus(error.response.status, errorMessage);
```
gabrieldutra, line 28: @kravets-levko I created the test and added `data.error` to also be evaluated (it was previously handled in `PromiseRejectionError`).

PR #4968, opened 2020-06-11
Don't reuse getErrorMessage

gabrieldutra:

- [x] Other

## Description
After #4959 we started to show detailed error messages, which is not ideal.

This PR reverts changes to `getErrorMessage` and remove its export. Thought about it for a while, it's useful to have a general "getDetailedErrorMessage", but we should have control on individual error flows. 

I'm also thinking how to cover this with a test (probably a component test with Jest).

#4959

--


gabrieldutra:
@kravets-levko I pushed a Jest test on 99c3bd7dece1553048a36f4f9dd84d17b393a7a9 to assert those errors are kept as they should in the future, lmk if you have comments on it 🙂.

PR #4973, opened 2020-06-15
Avoid purging operational queues

rauchy:
- [x] Bug Fix

## Description
The `purge_failed_jobs` job helps keep Redis memory consumption low by purging jobs that failed over a week ago. The bulk of these jobs are due to query execution failures.

However, the job purges _all_ queues, including "operational" queues (such as the `periodic` queue) which could affect system availability. This PR will force `purge_failed_jobs` to skip operational queues and handle query execution queues only.


arikfr:
Should we include `schemas` in the purge?

rauchy:
Totally. aabce0a

arikfr:
👍

PR #4978, opened 2020-06-17
Navbar: show correct settings link for non-admin users

kravets-levko:

- [x] Feature
- [x] Other

## Description

Non-admin users still have an access to some settings pages, but settings item in navbar was hidden for them. This PR makes settings item available for all users, but it will lead to first available settings page for that user.

### DIFF
--- a/client/cypress/integration/settings/settings_tabs_spec.js
+++ b/client/cypress/integration/settings/settings_tabs_spec.js
@@ -22,22 +22,38 @@ describe("Settings Tabs", () => {
 
   describe("For admin user", () => {
     beforeEach(() => {
-      cy.logout().then(() => cy.login());
+      cy.logout();
+      cy.login();
+      cy.visit("/");
     });
 
-    it("shows available tabs", () => {
-      cy.visit("/users");
+    it("settings link should lead to Data Sources settings", () => {
+      cy.getByTestId("SettingsLink")
+        .should("exist")
+        .should("have.attr", "href", "data_sources");
+    });
+
+    it("all tabs should be available", () => {
+      cy.getByTestId("SettingsLink").click();
       expectSettingsTabsToBe([...userTabs, ...adminTabs]);
     });
   });
 
   describe("For regular user", () => {
     beforeEach(() => {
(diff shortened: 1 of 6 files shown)


client/cypress/integration/settings/settings_tabs_spec.js
```diff
@@ -19,16 +16,45 @@ describe("Settings Tabs", () => {
...
+    beforeEach(() => {
+      cy.logout();
+      cy.login(regularUser.email, regularUser.password);
+      cy.visit("/");
+    });
+
+    it("settings link should lead to Users settings", () => {
+      cy.getByTestId("SettingsLink")
+        .should("exist")
+        .should("have.attr", "href", "users");
```
gabrieldutra, line 52: Considering you are already using the settings link now, I'd rather just make this assertions together with the settings tabs expect to make better use of E2E timing. It does make sense to have them separated conceptually, but Cypress encourages to make assertions together. I'll leave this one up to you.

kravets-levko: For me it looks better to have separate assertions (+ I cannot give a good name for test when tests are together). Let's keep them separate for now, and if it will make any issues in future we'll merge them.

gabrieldutra: > I cannot give a good name for test when assertions are together

I just don't rename them in such cases 😝.

kravets-levko: Old names also didn't reflect what new tests will do

gabrieldutra: Yes, but I mean when I can't have a naming that will represent all the assertions, I pick the most important one and keep it on the name. For this it's fine anyway.

kravets-levko: I think that test names should reflect what test is intended to check. In this case we test two things: 

1. setting item in navbar should have a proper href for each user
2. settings page should have a different set of tabs for each user

Both of them are important, so from my point both should be reflected in test name. That's what I had in mind

PR #4979, opened 2020-06-17
Add/Edit Tags dialog does not set focus on tags input

kravets-levko:

- [x] Bug Fix

## Description

I took an opportunity and converted the dialog to function component and fixed cleanup (it didn't cancel a request on destroy). The actual fix is https://github.com/getredash/redash/pull/4979/commits/9ad3ba1a7381cde95505395bb5f9bcabacda3f58.

### DIFF
--- a/client/app/components/tags-control/EditTagsDialog.jsx
+++ b/client/app/components/tags-control/EditTagsDialog.jsx
@@ -9,6 +9,16 @@ function EditTagsDialog({ dialog, tags, getAvailableTags }) {
   const [availableTags, setAvailableTags] = useState([]);
   const [isLoading, setIsLoading] = useState(true);
   const [values, setValues] = useState(() => uniq(map(tags, trim))); // lazy evaluate
+  const [selectRef, setSelectRef] = useState(null);
+
+  // Select is initially disabled, so autoFocus prop cannot make it focused.
+  // Solution is to pass focus to the select when available tags are loaded and
+  // select becomes enabled.
+  useEffect(() => {
+    if (selectRef && !isLoading) {
+      selectRef.focus();
+    }
+  }, [selectRef, isLoading]);
 
   useEffect(() => {
     let isCancelled = false;
@@ -31,12 +41,12 @@ function EditTagsDialog({ dialog, tags, getAvailableTags }) {
       className="shortModal"
       wrapProps={{ "data-test": "EditTagsDialog" }}>
       <Select
+        ref={setSelectRef}
         mode="tags"
         className="w-100"
         placeholder="Add some tags..."
         defaultValue={values}
         onChange={v => setValues(compact(map(v, trim)))}
-        autoFocus
         disabled={isLoading}
         loading={isLoading}>
         {map(availableTags, tag => (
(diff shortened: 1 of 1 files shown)


client/app/components/tags-control/EditTagsDialog.jsx
```diff
@@ -9,6 +9,21 @@ function EditTagsDialog({ dialog, tags, getAvailableTags }) {
   const [availableTags, setAvailableTags] = useState([]);
   const [isLoading, setIsLoading] = useState(true);
   const [values, setValues] = useState(() => uniq(map(tags, trim))); // lazy evaluate
+  const [selectRef, setSelectRef] = useState(null);
+
+  // For some reason `Select::autoFocus` doesn't work in dialogs.
+  // Workaround: get component's ref and manually pass focus
```
gabrieldutra, line 15: Looks like adding `destroyOnClose` to the dialog fixes it -> [example](https://codesandbox.io/s/spring-dust-v6t0v)

kravets-levko: In this case it doesn't. We use dialogs in a bit different way - we mount them with `ReactDOM.render()` and it's destroyed when closed. I think `autoFocus` may not work because of this.

kravets-levko:
> The reason to convert the component to function was mostly to take advantage of the easy cleanup function?

Yes. For the time I'm using React hooks I realized that they provide a better control over components state and lifecycle. Also code usually looks cleaner. So when I have a chance - I update component to use hooks (when it makes sense of course).

client/app/components/tags-control/EditTagsDialog.jsx
gabrieldutra: True, in this case it indeed doesn't work for a different reason 😂. Even that workaround didn't work to me when I tested here now.

The `Select` starts as `disabled`, the autoFocus works, but not when it's disabled. So we either have to remove that disabled prop (then we can use autoFocus prop) or make the focus happen only when the loading is complete (using the ref).

kravets-levko: @gabrieldutra thank you for pointing! Please check https://github.com/getredash/redash/pull/4979/commits/9ad3ba1a7381cde95505395bb5f9bcabacda3f58

client/app/components/tags-control/EditTagsDialog.jsx
```diff
@@ -1,66 +1,70 @@
...
-  static propTypes = {
-    dialog: DialogPropType.isRequired,
-    tags: PropTypes.arrayOf(PropTypes.string),
-    getAvailableTags: PropTypes.func.isRequired,
-  };
+function EditTagsDialog({ dialog, tags, getAvailableTags }) {
+  const [availableTags, setAvailableTags] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [values, setValues] = useState(() => uniq(map(tags, trim))); // lazy evaluate
+  const [selectRef, setSelectRef] = useState(null);
```
gabrieldutra, line 12: Lastly, is there a reason for this to be a State instead of a `useRef`?

kravets-levko: Yes, to be able to use it as a dependency for `useEffect`. When `ref.current` changes, component won't re-render, and I'd like to avoid race condition here - when request completes before ref initializes. I'm not sure how it may be possible, but if it happens - it will be a hard-to-find bug.

gabrieldutra: Nice, I figured this could be the case, thanks for explaining 🙂.

PR #4981, opened 2020-06-18
ErrorMessage is not centered

kravets-levko:

- [x] Bug Fix

Before & after:

image
image

### DIFF
--- a/client/app/components/ApplicationArea/ErrorMessage.less
+++ b/client/app/components/ApplicationArea/ErrorMessage.less
@@ -1,0 +1,12 @@
+.error-message-container {
+  width: 100%;
+  padding: 0 15px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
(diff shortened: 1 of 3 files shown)


client/app/components/ApplicationArea/ErrorMessage.less
```diff
@@ -0,0 +1,12 @@
+.error-message-container {
+  width: 100%;
+  padding: 0 15px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: flex-start;
+
+  .error-state {
+    max-width: 1200px;
```
gabrieldutra, line 10:
```suggestion
    max-width: 1200px;
    width: 100%;

    @media (min-width: 768px) {
      width: 65%;
    }
```

Or anything else at will, the idea is to keep a minimum/tending width to the error div (simulating the `.col-md-8`).

PR #4989, opened 2020-06-22
Fix CLI command for "status"

jimsparkman:

- [x] Bug Fix

## Description
CLI command "status" can fail due to incorrect connection information to RQ.
This change matches the behavior from line 65 and solves the connection error.

### DIFF
--- a/redash/cli/__init__.py
+++ b/redash/cli/__init__.py
@@ -2,8 +2,9 @@ import click
 import simplejson
 from flask import current_app
 from flask.cli import FlaskGroup, run_command
+from rq import Connection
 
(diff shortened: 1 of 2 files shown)


rauchy:
Since `get_status` is also called from web handlers (where a connection is automatically set in the context of the request), I'd rather see this connection _outside_ the `get_queues_status` function. 

Perhaps by using a `with Connection` context on the `status` command under `cli/__init__.py` (similar to how it's done in the `worker` command under `cli/rq.py`)

jimsparkman:
Brilliant, much better. I've moved the connection.

PR #4990, opened 2020-06-22
Remove pace-progress

gabrieldutra:

- [x] Other

## Description

There were a few issues happening after recent changes with an error coming internally from Pace (rare cases, but now since we have a better Error handling, it turned to be a problem on those few cases). I removed it for now, lmk if you prefer one of those other options:

1. Find an alternative to Pace
2. Find a way to properly handle Pace exceptions

--

--

### DIFF
--- a/client/app/assets/less/main.less
+++ b/client/app/assets/less/main.less
@@ -7,7 +7,6 @@
 /** Load Vendors Dependencies **/
 @import "~font-awesome/less/font-awesome";
 @import "~material-design-iconic-font/dist/css/material-design-iconic-font.css";
-@import "~pace-progress/themes/blue/pace-theme-minimal.css";
 
 @import "inc/variables";
 @import "inc/mixins";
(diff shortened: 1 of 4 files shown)


gabrieldutra:
@arikfr @kravets-levko lmk how important Pace is to you 😅.

kravets-levko:
> lmk how important Pace is to you :sweat_smile:

Not important at all :smile: Do we already we have alternative loading indicators on every page?

gabrieldutra:
> Not important at all 😄 Do we already have alternative loading indicators on every page?

It seemed so, to be honest I don't miss it when I navigate through pages.

kravets-levko:
> to be honest I don't miss it when I navigate through pages.

Sometimes on prod loading query/dashboard/some list may be quite slow, so that's important that user doesn't see a blank page. I checked codebase and looks that we already have loading indication on every page, so I guess it's safe to drop Pace

PR #4991, opened 2020-06-22
Catch QueryResultError on widget load

gabrieldutra:

- [x] Other

## Description
QueryResultErrors are expected from widget loading, so it shouldn't be left unhandled.

--

--

### DIFF
--- a/client/app/pages/dashboards/hooks/useDashboard.js
+++ b/client/app/pages/dashboards/hooks/useDashboard.js
@@ -103,7 +103,7 @@ function useDashboard(dashboardData) {
         if (error instanceof QueryResultError) {
           return;
         }
-        return error;
+        return Promise.reject(error);
       })
       .finally(() => setDashboard(currentDashboard => extend({}, currentDashboard)));
   }, []);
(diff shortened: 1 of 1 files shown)


client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -95,7 +96,16 @@ function useDashboard(dashboardData) {
...
     setDashboard(currentDashboard => extend({}, currentDashboard));
-    return widget.load(forceRefresh).finally(() => setDashboard(currentDashboard => extend({}, currentDashboard)));
+    return widget
+      .load(forceRefresh)
+      .catch(error => {
+        // QueryResultErrors are expected
+        if (error instanceof QueryResultError) {
+          return;
+        }
+        return error;
```
kravets-levko, line 106: If this error should be propagated to next `.catch` handler - you should return rejected promise here (e.g. `return Promise.reject(error)`), otherwise the flow will continue with next `.then` handler.

kravets-levko: Also, if this error is okay for widget - what do you think about moving this `.catch` to a Widget service to the `load` method?

gabrieldutra: > If this error should be propagated to next .catch handler - you should return rejected promise here (e.g. return Promise.reject(error)), otherwise the flow will continue with next .then handler.

Good catch, thanks! 👍

> Also, if this error is okay for widget - what do you think about moving this .catch to a Widget service to the load method?

It is an expected error, but I preferred to keep it propagated, so we can decide what to do from load calls (e.g.: if we eventually decide to add notifications after loading).

PR #4992, opened 2020-06-23
Allow GET from non-admins on data source resource

gabrieldutra:

- [x] Other

## Description
This allows requests from non-admin users that have at least "view_only" permissions on the DS. Also adds the `view_only` information and limit the result information when the user is non-admin.

Reverts #4927 as with this the specific route can be used.

Reasoning: it removes the need for the "list_data_sources" permission when accessing the View Query page.

--

--

### DIFF
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -34,12 +34,15 @@ class DataSourceTypeListResource(BaseResource):
 
 
 class DataSourceResource(BaseResource):
-    @require_admin
     def get(self, data_source_id):
-        data_source = models.DataSource.get_by_id_and_org(
-            data_source_id, self.current_org
+        data_source = get_object_or_404(
+            models.DataSource.get_by_id_and_org, data_source_id, self.current_org
         )
-        ds = data_source.to_dict(all=True)
+        require_access(data_source, self.current_user, view_only)
+        ds = data_source.to_dict(all=self.current_user.has_permission("admin"))
+        ds["view_only"] = all(
+            project(data_source.groups, self.current_user.group_ids).values()
+        )
         self.record_event(
             {"action": "view", "object_id": data_source_id, "object_type": "datasource"}
         )
--- a/redash/handlers/data_sources.py
+++ b/redash/handlers/data_sources.py
@@ -39,11 +39,13 @@ class DataSourceResource(BaseResource):
             models.DataSource.get_by_id_and_org, data_source_id, self.current_org
         )
         require_access(data_source, self.current_user, view_only)
-        ds = (
-            data_source.to_dict(all=self.current_user.has_permission("admin"))
-            if self.current_user.has_permission("list_data_sources")
-            else {}
-        )
+
+        ds = {}
+        if self.current_user.has_permission("list_data_sources"):
+            # if it's a non-admin, limit the information
+            ds = data_source.to_dict(all=self.current_user.has_permission("admin"))
+
+        # add view_only info, required for frontend permissions
         ds["view_only"] = all(
             project(data_source.groups, self.current_user.group_ids).values()
         )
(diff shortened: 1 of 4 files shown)


redash/handlers/data_sources.py
```diff
@@ -34,12 +34,15 @@ def get(self):
 
 
 class DataSourceResource(BaseResource):
-    @require_admin
     def get(self, data_source_id):
```
arikfr, line 37: It should still require some permission to view the data source details, because otherwise this API will be available when using the query API key...

Also, when the user doesn't have a `list_data_sources` permission it's very likely the admin wouldn't want them to see the details of the data source anyway.

Let's keep this new implementation, but still require `list_data_sources` permission here for now (ideally we would use a new permission, but this is a bit of a hassle to add).

redash/handlers/data_sources.py
```diff
@@ -34,12 +34,15 @@ def get(self):
 
 
 class DataSourceResource(BaseResource):
-    @require_admin
```
kravets-levko, line 37: What do you think about using `@require_permission("list_data_sources")` here instead of `require_access` in the code (keeping serialization logic as implemented here)? I checked how the things work for dashboards and queries and there we have this pattern: require `view_query`/`list_dashboards` both for list and get object handlers - this ensures that the same permissions needed for read access in all cases.

redash/handlers/data_sources.py
kravets-levko: I suggested the same but late a bit :sweat_smile:

gabrieldutra: Having "list_data_sources" doesn't fix the problem :(

> Also, when the user doesn't have a list_data_sources permission it's very likely the admin wouldn't want them to see the details of the data source anyway

I had that in mind, that's why I put the "all" only when the user is an admin.

https://github.com/getredash/redash/blob/8fb5bc785a3346920a65a99f2da1e4522d0b2e92/redash/handlers/data_sources.py#L42

Basically we show exactly the same information we would show in the list resource if the user is not an admin.

I see your point anyway, just that if felt ok as the user would actually be "listing data sources". WDYT about only returning the `{ view_only: true/false }` when the user doesn't have the list permission?

gabrieldutra: 4958ec404b41532c31b026f7857d2b4ceb676a52 with "returning the `{ view_only: true/false }` when the user doesn't have the list permission"

redash/handlers/data_sources.py
```diff
@@ -34,12 +34,19 @@ def get(self):
...
-            data_source_id, self.current_org
+        data_source = get_object_or_404(
+            models.DataSource.get_by_id_and_org, data_source_id, self.current_org
+        )
+        require_access(data_source, self.current_user, view_only)
+        ds = (
+            data_source.to_dict(all=self.current_user.has_permission("admin"))
+            if self.current_user.has_permission("list_data_sources")
+            else {}
+        )
```
kravets-levko, line 46: Please correct me if I misunderstand something:

1. for admins it will serialize all fields + `view_only` flags
2. for those who can `list_data_sources` it will serialize limited set of fields + `view_only` flags
3. for all others it will return just `view_only` flags

If so - it looks good. Maybe just add some comment to explain this logic, or split this one-liner to make it a bit easier to understand.

gabrieldutra: @kravets-levko correct, sorry if the code is a bit weird, for now I only made de simplest changes to what was there 😅

If @arikfr agrees with this flow I'll happily try to simplify the code (or add the explaining comment) + add tests to make sure it keeps working as it should.

redash/handlers/data_sources.py
arikfr: > I had that in mind, that's why I put the "all" only when the user is an admin.

By details I meant even the name or type.

redash/handlers/data_sources.py
arikfr: I guess this flow is fine, but we still need the relevant UI changes (hide data source if we don't have its name/type).

gabrieldutra: I've added tests and updated the code.

One extra thing to @kravets-levko flow that was not mentioned: The user has to either be an admin or have at least view permissions to follow to that flow, otherwise a 403 is returned.

PR #4998, opened 2020-06-24
Dynamically register frontend routes

kravets-levko:

- [x] Refactor
- [x] Feature
- [x] Other

## Description

Dynamically register frontend routes in order to allow adding/overriding them in frontend extensions.

Related to getredash/redash#5000


client/app/services/routes.js
```diff
@@ -0,0 +1,21 @@
+import { isString, isObject, filter } from "lodash";
+
+class Routes {
+  items = [];
```
gabrieldutra, line 4: Same thing I commented in https://github.com/getredash/redash/pull/5000#discussion_r445175371 (or almost), update if you think it makes sense

kravets-levko: The answer is the same for both cases, so I replied in the thread you mentioned ^

PR #5000, opened 2020-06-24
Allow unregistering settings tabs

kravets-levko:

- [x] Refactor
- [x] Feature
- [x] Other

## Description

Allow to unregister settings tabs in order to allow adding/overriding them in frontend extensions.

Related to getredash/redash#4998

### DIFF
--- a/client/app/services/settingsMenu.js
+++ b/client/app/services/settingsMenu.js
@@ -1,12 +1,15 @@
-import { isFunction, extend, omit, sortBy, find, filter } from "lodash";
+import { isString, isObject, isFunction, extend, omit, sortBy, find, filter } from "lodash";
 import { stripBase } from "@/components/ApplicationArea/Router";
 import { currentUser } from "@/services/auth";
 
 class SettingsMenuItem {
   constructor(menuItem) {
-    extend(this, { pathPrefix: `/${menuItem.path}` }, omit(menuItem, ["isActive"]));
+    extend(this, { pathPrefix: `/${menuItem.path}` }, omit(menuItem, ["isActive", "isAvailable"]));
     if (isFunction(menuItem.isActive)) {
       this.isActive = menuItem.isActive;
+    }
+    if (isFunction(menuItem.isAvailable)) {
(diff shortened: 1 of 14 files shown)


client/app/services/settingsMenu.js
```diff
@@ -20,13 +23,22 @@ class SettingsMenuItem {
...
 class SettingsMenu {
-  constructor() {
-    this.items = [];
+  items = [];
+
+  add(id, item) {
+    id = isString(id) ? id : null;
+    this.remove(id);
+    if (isObject(item)) {
+      this.items.push(new SettingsMenuItem({ ...item, id }));
```
gabrieldutra, line 32: Any reason not to send the id to the constructor instead of using an object or a `Map` now that the structure has an id?

We need to add a getter for items to sort it though.

kravets-levko: The reason is quite simple: initially I didn't plan to add IDs to all settings tabs (and to routes too), and allowed to pass `null` ID. And when there are tabs/routes w/o ID - it's impossible to use Map because there will be duplicated keys. Eventually I assigned IDs to all tabs/routes, but decided to keep that feature. Anyway, using arrays doesn't add any extra complexity, it's still easy to understand code, and so on.

Regarding getter that will reorder items: currently we register settings tabs and routes only on app startup, and later they stay unchanged. So it's better to sort them at once and later just use when needed, instead of sorting each time we access them.

gabrieldutra: > it's impossible to use Map because there will be duplicated keys

I thought the idea was to replace items by id when wanted 🤔, i.e.: having unique data. (The reorder is only needed if we do that).

Anyway, no problem with the array approach, just wondering when we want to have multiple ids.

kravets-levko: I was thinking about replacing items when needed, but since we don't need this yet - I decided to not overcomplicate things. Replacing tabs "on fly" will require some updates to settings tabs wrapper, to router, etc. - they all should be ready to update when something changes, but ATM they cannot do that.

PR #5003, opened 2020-06-25
Fix org option in users create_root cli command

nason:

- [x] Bug Fix

## Description

While testing out redash's multi-org features, I ran into some issues setting up multiple organizations and kept finding everything was going into the "default" org. Turns out the cli option (`manage users create_root --org test_org`) was being ignored and fixed to "default"

With this change, things worked as expected.

- https://discuss.redash.io/t/creating-organizations/2540/8
- Not exactly related, but congratulations on the acquisition! I'm excited to see what comes 🚀


rauchy:
Thanks @nason 👍

PR #5008, opened 2020-06-29
Custom primary/foreign key types

rauchy:
- [x] Feature

## Description
By default we use `db.Integer` as the key type for primary/foreign keys. This PR adds support for anyone who wants to use different primary/foreign key types.

Using dynamic settings, you can modify the type for any model. You can also add any arguments for the column definitions (for example - default ID generators / increment functions)

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -42,7 +42,7 @@ from redash.utils import (
 from redash.utils.configuration import ConfigurationContainer
 from redash.models.parameterized_query import ParameterizedQuery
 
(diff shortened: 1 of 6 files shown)


rauchy:
I was shooting for a mixin which will use the class name automatically, but that was incompatible with SQLA’s column definition timing during class definition.

jezdez:
I'm curious, how will this integrate with the Alembic migration system?

arikfr:
> I'm curious, how will this integrate with the Alembic migration system?

Alembic picks up the type change and generates the needed migration (when calling `db migrate`).

PR #5010, opened 2020-06-29
Databricks custom Schema Browser

gabrieldutra:

- [x] Feature

## Description
Custom Schema Browser for the databricks data source.

To-do:
- [x] Reset database filter string when a db is selected
- [x] Add loading state
- [x] Update Select color when searching

--

Soon 👍

### DIFF
--- a/client/app/components/queries/editor-components/databricks/DatabricksSchemaBrowser.jsx
+++ b/client/app/components/queries/editor-components/databricks/DatabricksSchemaBrowser.jsx
@@ -35,6 +35,7 @@ export default function DatabricksSchemaBrowser({
   const [filterString, setFilterString] = useState("");
   const [databaseFilterString, setDatabaseFilterString] = useState("");
   const filteredSchema = useMemo(() => applyFilterOnSchema(schema, filterString), [schema, filterString]);
+  const [isDatabaseSelectOpen, setIsDatabaseSelectOpen] = useState(false);
   const [expandedFlags, setExpandedFlags] = useState({});
   const [handleFilterChange] = useDebouncedCallback(setFilterString, 500);
   const [handleDatabaseFilterChange, cancelHandleDatabaseFilterChange] = useDebouncedCallback(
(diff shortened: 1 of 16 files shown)


client/app/components/queries/editor-components/databricks/DatabricksSchemaBrowser.jsx
```diff
@@ -0,0 +1,147 @@
...
+      ...expandedFlags,
+      [tableName]: !expandedFlags[tableName],
+    });
+  }
+
+  return (
+    <div className="databricks-schema-browser schema-container" {...props}>
+      <div className="schema-control">
+        <Input
+          className={isDatabaseSelectOpen ? "database-select-open" : ""}
```
gabrieldutra, line 88: So, @arikfr, how badly do you want the Select to be white when open? 😆

Turns out the grey bg comes from the input addon wrapper, recreating the borders from the Select component feels too much and wasn't that simple. `:has()` is the css pseudo-selector we could apply here, but will be avail in CSS4, so, "simplest" option was to use JS to achieve that.

The result so far is this:
Screen Shot 2020-06-29 at 21 56 32
 
Can still remove that placeholder from current DB if you think it's relevant, although I think it makes sense to be there, so we don't get the feeling that the option was lost.

arikfr: Writing on top of the previous value feels like a bug... 

> so we don't get the feeling that the option was lost.

Can we open the list so the selected option is visible and there is a checkmark next to it?

gabrieldutra: > Can we open the list so the selected option is visible and there is a checkmark next to it?

It's already like this. Anyway, done in https://github.com/getredash/redash/pull/5010/commits/69b79341af308512a848c897cd14365f4b0c01a9, let's give it a try.

gabrieldutra:
Extra: added the loading indications to the default schema browser

default-schema-browser-loading

jezdez:
Woah! 🤯

PR #5011, opened 2020-06-29
Fix dashboard root container height bug on Safari

daniellangnet:

- [x] Bug Fix

## Description

Both on mobile Safari as well as desktop Safari, the dashboard controls overlap with the dashboard content:

image

The reason for this is that the `application-root` container on Safari is limited to the viewport height. Chrome ignores the `height: 100vh` and just pushes beyond that, but webkit limits the height and this has all sorts of consequences.

The fix should address that and still get the desired effect of having 100vh as the minimum height.

### DIFF
--- a/client/app/components/ApplicationArea/ApplicationLayout/index.less
+++ b/client/app/components/ApplicationArea/ApplicationLayout/index.less
@@ -7,7 +7,7 @@ body #application-root {
   flex-direction: row;
   justify-content: stretch;
   padding-bottom: 0 !important;
-  height: 100vh;
+  min-height: 100vh;
 
   .application-layout-side-menu {
     height: 100vh;


daniellangnet:
Also this should probably get a good amount of testing since it changes the application root container and is cross-cutting. However, it took me quite a while to figure out the issue and I wanted to share it so that maybe somebody else can take it from here.

gabrieldutra:
Hey @daniellangnet, thanks very much for reporting this! 😁

Sure we will take it from here and make deeper tests in the application, the good news is: we have [Percy](https://percy.io/Redash/redash/builds/5892229?utm_campaign=Redash&utm_content=redash&utm_source=github_status_public), so _ideally_ it points out any change in the application styling (at least for Chrome and Firefox 😅) and for this PR it seems ok.

daniellangnet:
Looks like somebody changed it in `master` since I opened this PR but as far as I can tell, it's still an issue for anybody using an iPhone to look at their Redash. It's rebased now and working again. Tested on Chrome, Firefox, Safari (latest versions)

daniellangnet:
Turns out changing this creates another issue on Chrome/Firefox, so I don't recommend merging it right now. I need to see if I can find time in the next couple of weeks to track it down and find a better solution

PR #5013, opened 2020-06-30
Fix broken custom JS visualization settings

daniellangnet:

- [x] Bug Fix

## Description

The setting to allow custom JS visualizations was broken because of a typo and not getting passed through to the visualization components, therefore never showing the "Custom" option in the chart type selector or rendering widgets correctly.


gabrieldutra:
Good catch, thanks! 👍

PR #5017, opened 2020-07-01
Add plus between tags to clarify how they are used

### LINKED ISSUE #4628: Query/Dashboard list pages: clarify how tags are used
> Perhaps adding "AND" or "+" between selected tags in not found message could reduce doubts related with tag filtering. But Anyway thanks for your reply!

_Originally posted by @ptsilva in https://github.com/getredash/redash/issues/4557#issuecomment-582173490_

Let's implement this (add `+` / `AND` between tags)

ajkovar:

- [x] Other

## Description

As stated in #4628 , this PR adds a '+' between tags in the case of no results to clarify how they are being used.

Closes #4628

How the tags look after this change (both with and without the separator):

Captura de Pantalla 2020-07-01 a la(s) 2 42 16 p  m
Captura de Pantalla 2020-07-01 a la(s) 2 42 45 p  m
Captura de Pantalla 2020-07-01 a la(s) 3 46 08 p  m

### DIFF
--- a/client/app/components/tags-control/TagsControl.jsx
+++ b/client/app/components/tags-control/TagsControl.jsx
@@ -12,6 +12,7 @@ export class TagsControl extends React.Component {
     onEdit: PropTypes.func,
     className: PropTypes.string,
     tagsExtra: PropTypes.node,
+    tagSeparator: PropTypes.node,
     children: PropTypes.node,
   };
 
@@ -22,6 +23,7 @@ export class TagsControl extends React.Component {
     onEdit: () => {},
     className: "",
     tagsExtra: null,
+    tagSeparator: null,
     children: null,
   };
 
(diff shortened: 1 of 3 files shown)


client/app/components/tags-control/TagsControl.jsx
```diff
@@ -52,11 +54,14 @@ export class TagsControl extends React.Component {
...
+        {flatMap(this.props.tags, (tag, i) => [
           <span className="label label-tag" key={tag} title={tag} data-test="TagLabel">
             {tag}
-          </span>
-        ))}
+          </span>,
+          this.props.tagSeparator && i < this.props.tags.length - 1 && (
+            <span className="tag-separator">{this.props.tagSeparator}</span>
+          ),
+        ])}
```
kravets-levko, line 64: Consider desctructuring props (instead of using `this.props.***` everywhere) to make code more readable:
`const { tags, tagSeparator } = this.props;`

Suggestion for this code fragment:

```jsx
{map(tags, (tag, index) => (
  <React.Fragment key={tag}>
    {tagSeparator && (index > 0) && <span className="tag-separator">{tagSeparator}</span>}
    <span className="label label-tag" key={tag} title={tag} data-test="TagLabel">
      {tag}
    </span>
  </React.Fragment>
)}
```

ajkovar: Nice suggestions.  Changes made.

PR #5019, opened 2020-07-02
Install puppeteer with unsafe perm to avoid cypress flaky

gabrieldutra:

- [x] Other

## Description
Following [this](https://support.circleci.com/hc/en-us/articles/360044577411-ERROR-Failed-to-download-Chromium-r686378-)

Running some tests to make sure it worked.


eradman:
@gabrieldutra okay to close this? It is useful to know this workaround exists, but I don't think we have seen this problem recently.

justinclift:
Yeah, we haven't seen the problem this PR would address in the last month, so it's probably not needed any more. :smile:

PR #5020, opened 2020-07-02
Explicitly sort routes to reduce a chance of conflicts

kravets-levko:

- [x] Other

## Description

Make sure that router will first try to match more specific routes (longer and without params), and only then will try routes with params.

### DIFF
--- a/client/app/services/routes.js
+++ b/client/app/services/routes.js
@@ -1,9 +1,9 @@
-import { isString, isObject, filter, sortBy, some } from "lodash";
+import { isString, isObject, filter, sortBy } from "lodash";
 import pathToRegexp from "path-to-regexp";
 
-function isSimplePath(path) {
+function getRouteParamsCount(path) {
   const tokens = pathToRegexp.parse(path);
-  return some(tokens, isObject);
+  return filter(tokens, isObject).length;
 }
 
 class Routes {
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -64,6 +64,7 @@
     "mousetrap": "^1.6.1",
     "mustache": "^2.3.0",
     "numeral": "^2.0.6",
+    "path-to-regexp": "^3.1.0",
```
kravets-levko, line 67: This version is used by `universal-router`

client/app/services/routes.js
```diff
@@ -1,19 +1,40 @@
-import { isString, isObject, filter } from "lodash";
+import { isString, isObject, filter, sortBy, some } from "lodash";
+import pathToRegexp from "path-to-regexp";
+
+function isSimplePath(path) {
+  const tokens = pathToRegexp.parse(path);
+  return some(tokens, isObject);
+}
```
gabrieldutra, line 7: This seems to be with the opposite logic

e.g.:
`/` is showing as not a simple path and `/alerts/:alertId` is showing as a simple path

kravets-levko: @gabrieldutra https://github.com/getredash/redash/pull/5020/commits/e45068a0ac0b289767b3ccf3aaba3eaf75c8c58e

kravets-levko:
@gabrieldutra I see your concerns. But let's think what kind of conflicts may be there? ATM that's `/path/:id` vs. `/path/new` - they are perfectly solved if static route will be checked first. Cases like `/path/:id` vs. `/path/:id--some-text` or `/path/:id-(.*)` could be also solved with this approach. Also, most of the time routes have a different prefix which allows to identify them. But reordering items manually is error-prone, and also keep in mind that routes may be unregistered/re-registered now, which will change manual order. So we have either to order them automatically, or introduce some value to change order of routes, like we did for settings tabs. I don't know which one is better, please let me know your opinion.

PR #5022, opened 2020-07-02
Refactor: extract commonly used pattern into hook

kravets-levko:

- [x] Refactor

## Description

We use this pattern when we have to call some callback, say, after async operation, but the callback may change during it. Or if we use a hook with dependencies and have to call a callback there, but adding the callback to dependencies will break some logic or make it more complicated.


kravets-levko:
> only bad thing I see is that the eslint exhaustive deps rule can't be set to handle the immutable hook

There is an open issue for this, hope one day eslint will be able to handle this: facebook/react#16873

PR #5027, opened 2020-07-03
Add TypeScript support

simonschneider-db:

- [x] Other

## Description

Adds the typescript loader to webpack config and package.json, adds typescript specific linting to eslint and updates npm scripts in package.json to include `.ts` and `.tsx`.

I added two examples of using TypeScript: 
* DashboardListEmptyState is ported over to TypeScript.
* EmptyState has a `.d.ts` companion file that adds type definition.

Both examples show how we can move existing code over to TypeScript step by step.

Dashboard stays the same:

image


client/app/pages/dashboards/components/DashboardListEmptyState.tsx
```diff
@@ -1,10 +1,16 @@
-import React from "react";
-import PropTypes from "prop-types";
+import * as React from "react";
+import * as PropTypes from "prop-types";
 import BigMessage from "@/components/BigMessage";
 import NoTaggedObjectsFound from "@/components/NoTaggedObjectsFound";
 import EmptyState from "@/components/empty-state/EmptyState";
 
-export default function DashboardListEmptyState({ page, searchTerm, selectedTags }) {
+export interface DashboardListEmptyStateProps {
```
simonschneider-db, line 7: I tried to include https://github.com/milesj/babel-plugin-typescript-to-proptypes. This would have required a switch from ts-loader to handling the typescript transpiling in babel (@babel/preset-typescript). 
While trying this out babel uncovered a few inconsistencies that would need to be fixed first (e.g. version.json not containing a valid JSON).

Without the plugin there are still a few options for proptypes:
* All new code does not really require proptypes. We will see TypeScript compiler errors instead.
* Keeping both for essential components that are primarily called from JS code (like here)
* Existing files could also use the following code to infer types from existing proptypes:

```
export type DashboardListEmptyStateProps = InferProps<typeof DashboardListEmptyState.propTypes>

export function DashboardListEmptyState({ page, searchTerm, selectedTags }: DashboardListEmptyStateProps) {
```

webpack.config.js
```diff
@@ -19,6 +20,12 @@ const redashBackend = process.env.REDASH_BACKEND || "http://localhost:5000";
 const basePath = path.join(__dirname, "client");
 const appPath = path.join(__dirname, "client", "app");
 
+const tsConfigPath = path.join(basePath, "tsconfig.json");
+
+if (!fs.existsSync(tsConfigPath)) {
+  throw new Error(`Can not find tsconfig: ${tsConfigPath}`);
```
simonschneider-db, line 26: I added this line because the initial error I got for the missing tsConfig file in the docker container was missleading (see https://github.com/TypeStrong/ts-loader/issues/405#issuecomment-330108362)

I then figured out that I either need to move tsconfig into `/client` or copy it over in the Dockerfile.

client/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist/",
+    "noImplicitAny": true,
+    "module": "es6",
+    "target": "es5",
+    "jsx": "react",
+    "allowJs": true
+  }
+}
```
benamorbe, line 10: Can you add explicit `include` to avoid compiling non-source files? Something like:
```suggestion
  "include": [
    "app"
  ]
}
```

client/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist/",
+    "noImplicitAny": true,
+    "module": "es6",
+    "target": "es5",
+    "jsx": "react",
+    "allowJs": true
```
benamorbe, line 8: Can you also add DOM/ES libs?
```suggestion
    "allowJs": true,
    "lib": [
      "dom",
      "dom.iterable",
      "es6"
    ],
```

client/tsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+  }
```
benamorbe, line 9: Other useful/recommended compiler options :

1. `skipLibCheck: true`: skips unnecessary lib checks in `*.d.ts` files (usually outside our scope)
2. `noEmit: true`: We use `webpack` so we should avoid emitting any output accidentally when running `tsc` manually.
3. `strict: true`: Enable strict type checking.
4. `forceConsistentCasingInFileNames: true`: Disallow inconsistently-cased references to the same file.

arikfr:
Thanks!!

PR #5037, opened 2020-07-07
Query Source: Add Shift+Enter shortcut for query execution

gabrieldutra:

- [x] Feature

## Description
This is the default set of keys used for Jupyter Notebooks, so it makes sense to support it too.

Note:
1. This is only applied to the Query Source Page
2. I tried to make 314eae5bb78328877b5829354ccae8c29a3bae52 from the QueryEditor component, without success. The main idea is not to have an annoying behavior of a new line added when you press Shift+Enter and the query cannot execute.

--

--

### DIFF
--- a/client/app/components/queries/QueryEditor/QueryEditorControls.jsx
+++ b/client/app/components/queries/QueryEditor/QueryEditorControls.jsx
@@ -1,4 +1,4 @@
-import { isFunction, map, filter, fromPairs } from "lodash";
+import { isFunction, map, filter, fromPairs, noop } from "lodash";
 import React, { useEffect } from "react";
 import PropTypes from "prop-types";
 import Tooltip from "antd/lib/tooltip";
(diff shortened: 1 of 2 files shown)


client/app/components/queries/QueryEditor/QueryEditorControls.jsx
```diff
@@ -43,10 +43,10 @@ export default function EditorControl({
   useEffect(() => {
     const buttons = filter(
       [addParameterButtonProps, formatButtonProps, saveButtonProps, executeButtonProps],
-      b => b.shortcut && !b.disabled && isFunction(b.onClick)
+      b => b.shortcut && (!b.disabled || b.keepShortcutKeysWhenDisabled) && isFunction(b.onClick)
```
kravets-levko, line 46: What if we will not add this flag, but just always keep shortcuts (with noop handler)?

gabrieldutra: > What if we will not add this flag, but just always keep shortcuts (with noop handler)?

I considered, but at first I didn't want to risk having shortcut keys _always_ taken. But I just checked and the only control we have with both `disabled` and `shortcuts` is the execute query one, so this should be fine 👍

PR #5038, opened 2020-07-07
Introduce caching to the Databricks Schema Browser

gabrieldutra:

- [x] Other

## Description
Improvements for the response time for the Databricks Schema Browser.

### DIFF
--- a/client/app/components/queries/editor-components/databricks/DatabricksSchemaBrowser.jsx
+++ b/client/app/components/queries/editor-components/databricks/DatabricksSchemaBrowser.jsx
@@ -2,6 +2,8 @@ import React, { useState, useMemo, useEffect, useCallback } from "react";
 import { slice, without, filter, includes } from "lodash";
 import PropTypes from "prop-types";
 import { useDebouncedCallback } from "use-debounce";
+import Button from "antd/lib/button";
+import Icon from "antd/lib/icon";
 import Input from "antd/lib/input";
 import Select from "antd/lib/select";
 import { SchemaList, applyFilterOnSchema } from "@/components/queries/SchemaBrowser";
@@ -32,6 +34,8 @@ export default function DatabricksSchemaBrowser({
     loadingSchema,
     currentDatabaseName,
     setCurrentDatabase,
+    refreshAll,
+    refreshing,
   } = useDatabricksSchema(dataSource, options, onOptionsUpdate);
   const [filterString, setFilterString] = useState("");
   const [databaseFilterString, setDatabaseFilterString] = useState("");
@@ -126,6 +130,13 @@ export default function DatabricksSchemaBrowser({
           onTableExpand={toggleTable}
           onItemSelect={onItemSelect}
         />
+        {!(loadingSchema || loadingDatabases) && (
+          <div className="load-button">
+            <Button type="link" onClick={refreshAll}>
+              <Icon type="sync" spin={refreshing} /> Refresh
+            </Button>
+          </div>
+        )}
       </div>
     </div>
   );
--- a/client/app/components/queries/editor-components/databricks/DatabricksSchemaBrowser.jsx
+++ b/client/app/components/queries/editor-components/databricks/DatabricksSchemaBrowser.jsx
@@ -1,5 +1,5 @@
 import React, { useState, useMemo, useEffect, useCallback } from "react";
-import { slice, without, filter, includes } from "lodash";
+import { slice, without, filter, includes, get, find } from "lodash";
 import PropTypes from "prop-types";
 import { useDebouncedCallback } from "use-debounce";
 import Button from "antd/lib/button";
@@ -33,6 +33,7 @@ export default function DatabricksSchemaBrowser({
     loadingDatabases,
     schema,
     loadingSchema,
+    loadTableColumns,
     currentDatabaseName,
     setCurrentDatabase,
     refreshAll,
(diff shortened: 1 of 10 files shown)


redash/query_runner/databricks.py
```diff
@@ -144,10 +144,33 @@ def get_databases(self):
...
+        cursor = self._get_cursor()
+
+        # load tables first, otherwise tables without columns are not showed
+        cursor.tables(schema=database_name)
+
+        for table in cursor:
+            table_name = "{}.{}".format(table[1], table[2])
+
+            if table_name not in schema:
+                schema[table_name] = {"name": table_name, "columns": []}
```
gabrieldutra, line 172: Had to add a `cursor.tables` here too (previously it was only with `cursor.columns`), otherwise it doesn't return Tables without columns 😬 (causing inconsistency between the results)

redash/handlers/databricks.py
```diff
@@ -1,44 +1,107 @@
...
+        if not refresh:
+            cached_databases = _get_databases_from_cache(data_source_id)
+
+            if cached_databases is not None:
+                return cached_databases
 
         try:
-            return data_source.query_runner.get_databases()
+            databases = data_source.query_runner.get_databases()
+            redis_connection.set(_databases_key(data_source_id), json_dumps(databases))
```
arikfr, line 58: Let's add an expiry time (1 hour?) to each key we set so it gets refreshed and for cleanup purposes.

redash/handlers/databricks.py
```diff
@@ -1,44 +1,107 @@
...
+            return databases
         except Exception:
-            return {"error": {"code": 2, "message": "Error retrieving schema."}}
+            abort(500, message="Error retrieving database list.")
```
arikfr, line 61: 500 is an unexpected server error, but in most cases this is not the case. This is why we were using status code 200 and just returning an error message.

client/app/components/queries/editor-components/databricks/DatabricksSchemaBrowser.jsx
```diff
@@ -65,15 +71,20 @@ export default function DatabricksSchemaBrowser({
...
+    setExpandedFlags({});
+  }, [currentDatabaseName]);
+
   if (schema.length === 0 && databases.length === 0 && !(loadingDatabases || loadingSchema)) {
     return null;
   }
 
   function toggleTable(tableName) {
+    const table = find(schema, { name: tableName });
+    if (!expandedFlags[tableName] && get(table, "loading", false)) loadTableColumns(tableName);
```
kravets-levko, line 87:
```suggestion
    if (!expandedFlags[tableName] && get(table, "loading", false)) { 
      loadTableColumns(tableName);
    }
```

client/app/components/queries/editor-components/databricks/useDatabricksSchema.js
```diff
@@ -31,22 +31,82 @@ export default function useDatabricksSchema(dataSource, options = null, onOption
...
+
   const schemasRef = useRef();
   schemasRef.current = schemas;
   useEffect(() => {
     let isCancelled = false;
     if (currentDatabaseName && !has(schemasRef.current, currentDatabaseName)) {
       setLoadingSchema(true);
       getSchema(dataSource, currentDatabaseName)
-        .then(data => {
+        .catch(() => Promise.resolve([]))
```
kravets-levko, line 98: Here an empty array is returned, but on next line an object is expected. Is it okay?

client/app/components/queries/editor-components/databricks/useDatabricksSchema.js
```diff
@@ -31,22 +31,82 @@ export default function useDatabricksSchema(dataSource, options = null, onOption
...
+  const refreshAll = useCallback(() => {
+    if (!refreshing) {
+      setRefreshing(true);
+      const getDatabasesPromise = getDatabases(dataSource, true).then(setDatabases);
+      const getSchemasPromise = getSchema(dataSource, currentDatabaseName, true).then(({ schema }) =>
+        setCurrentSchema(schema)
+      );
+
+      Promise.all([getSchemasPromise.catch(() => {}), getDatabasesPromise.catch(() => {})]).then(() =>
+        setRefreshing(false)
```
kravets-levko, line 86: No need to fix it now, it's rather a subject to discuss (I don't have a good ideas for this as well). So, what if you trigger that async action, and then go to another page so component is destroyed? If that action is fast, we can just ignore this case, but here (as I understand) refreshing schema may take some time. Of course we can still ignore this case (what is the chance that user will trigger refresh and then decide to leave query editor?), or we can set an `isDestroyed` flag and check it (using `useEffect + useRef`). I'm not sure what's the best solution here. WDYT?

P.S. There are also several similar fragments, and this comment applies to them too.

gabrieldutra: Yep, there are several places where it's kinda difficult to have a control whether component was unmounted. We should create an issue to discuss a way to handle that. I was holding myself from the `useEffect + useRef` pattern to look for something better, but that can be it.

client/app/components/queries/editor-components/databricks/useDatabricksSchema.js
gabrieldutra: Good catch, thanks! 🙂

client/app/components/queries/editor-components/databricks/useDatabricksSchema.js
gabrieldutra: #5063

PR #5041, opened 2020-07-09
Update Ace Editor version

gabrieldutra:

- [x] Other

## Description
Updating Ace Editor seemed as simple as that, only thing that needs to be yet solved is that https://github.com/ajaxorg/ace/pull/3676 was a breaking change to #3092 😕. We displayed column names only instead of "table.column" when the dot was pressed.

Options:
1. Display the full "tablename.columnname"
2. Create a fork and revert that PR (@overleaf did this one)
3. ?

--

--


arikfr:
> * Display the full "tablename.columnname"

Looks like a reasonable solution.

gabrieldutra:
> Looks like a reasonable solution.

Yep, I left the full "tablename.columnname" for now.

gabrieldutra:
> Did you try it on different browsers?

From a quick check, seems fine for Chrome, firefox and Safari. I'll merge so we can test it more on the Preview env.

PR #5042, opened 2020-07-10
[POC] Add TypeScript infra based on babel preset

benamorbe:
This PR is a POC on adding TypeScript via babel preset instead of `ts-loader`. The main benefits of using babel/TypeScript combinations:

1. We have 1 compiler/loader instead of 2 separate compilers/loaders.
2. We can use all features of babel! We already have it in Redash!
3. Almost instant compile during development loop! `ts-loader` is very slow.
4. Check for type errors only when you’re ready.  Babel won't compile all TypeScript when we make changes. It simply strips TypeScript. Once you are sure about your experimental code, you can fix the types.
5. Out of the box support for @babel/preset-env in TypeScript for better ES/browsers targeting.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

### DIFF
--- a/client/.babelrc
+++ b/client/.babelrc
@@ -7,7 +7,8 @@
       ],
       "useBuiltIns": "usage"
     }],
-    "@babel/preset-react"
+    "@babel/preset-react",
+    "@babel/preset-typescript"
   ],
   "plugins": [
     "@babel/plugin-proposal-class-properties",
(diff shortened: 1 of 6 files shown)


arikfr:
Closing this as changes were added to #5027.

Issue #3815, opened 2019-05-20
Don't show simple results API usage examples for parameterized queries


gggooo:
### Issue Summary

queries->show api key->get json url :
http://server/api/queries/{id}/results.json?api_key={key}

but its not work

the right url is http://server/api/queries/{id}/results/{id2}.json?api_key={key}

this bug makes embed visualization show the error message "It seems like the page you're looking for cannot be found."

### Technical details:

* Redash Version:Version: 7.0.0+b18042 (4a978bad)
* Browser/OS:chromeon windows7 /server on ubuntu
* How did you install Redash:yes

arikfr:
Does your query use parameters?

gggooo:
@arikfr 

> Does your query use parameters?

Thankyou for your reply!  Now I know why this problem happened.

and How to use the query API with parameters， is there anything like api document?

Sorry for the inconvenience.

rauchy:
@gggooo if you want to execute a query with parameters using the API, you should POST to `/queries/{id}/results`  and your POST body should include a `parameters` key, for example:
```json
{
  parameters: {
    param1: 123,
    param2: 456
  }
}
```

If you're trying to embed a visualization with parameters, make sure none of your parameters are textual params and that your code is up-to-date, as this feature was released not long ago.

gggooo:
@rauchy Thankyou for your reply! I will try to use it, thanks!

arikfr:
@gggooo I reopened your issue and updated the description to reflect a possible solution.

Thank you for bringing this to our attention.

kishandonepudi:
> @gggooo if you want to execute a query with parameters using the API, you should POST to `/queries/{id}/results` and your POST body should include a `parameters` key, for example:
> 
> ```json
> {
>   "parameters": {
>     "param1": 123,
>     "param2": 456
>   }
> }
> ```
> 
> If you're trying to embed a visualization with parameters, make sure none of your parameters are textual params and that your code is up-to-date, as this feature was released not long ago.

@rauchy can you please share me your postman collection or else a guide in notes pls

rauchy:
@gggooo I don't use Postman that much, but here's the simplest curl command you can run which demonstrates this:

`curl http://<host>/api/queries/<query_id>/results?api_key=<query_api_key> --data-binary '{"parameters":{"<param_name>":<param_value>}}'`

Does this help?

PR #5043, opened 2020-07-10
show api usage for parameterized queries #3815

ajkovar:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Show a curl command for parameterized queries

https://github.com/getredash/redash/issues/3815

Here is an example of a parameterized query:
Captura de Pantalla 2020-07-10 a la(s) 1 00 55 p  m

And normal queries, of course, look the same:
Captura de Pantalla 2020-07-10 a la(s) 1 01 19 p  m

I tried investigating if it was possible to generate a CSV formatted response for parameterized queries and from what I can tell it is not without modifying the python side.  I didn't go that far but if you would like, I could try to attempt it.

### DIFF
--- a/client/app/components/queries/ApiKeyDialog/index.jsx
+++ b/client/app/components/queries/ApiKeyDialog/index.jsx
@@ -12,6 +12,18 @@ import notification from "@/services/notification";
 
 import "./index.less";
 
+const curlCommand = ({ url, parameters }) =>
+  `curl ${url}` + (Object.keys(parameters).length === 0 ? "" : ` --data '${JSON.stringify({ parameters })}'`);
+
+const parameterJson = parameters =>
+  parameters.reduce(
+    (paramJson, param) => ({
+      ...paramJson,
+      [param.name]: param.value,
+    }),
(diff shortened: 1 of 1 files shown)


susodapop:
Wouldn't this change confuse users? Depending on whether Redash already has a cached query result, there are two possible outcomes from issuing that `curl` command. 

1. **If there is a cached result** then it is returned. _The user can expect this_.

2. **If there is no cached result** then a job object is returned. The user will need to issue two more requests before downloading the result. The first request checks the `/jobs` endpoint for a status. When the job completes, a `query_result_id` is included in the job status response. The user makes a second request to `/query_results/<query_result_id>` which returns the data. _The user would not expect this_.

guidopetri:
@ajkovar , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts? We should also address the feedback already given.

Issue #5031, opened 2020-07-06
Counter visualization is enormous on the query screen


susodapop:
### Issue Summary

Counter visualizations have no max-width or max-height limit. So they grow to fill all available space on the query screen.

![Screen Shot 2020-07-06 at 10 32 16](https://user-images.githubusercontent.com/17067911/86611227-fd4c4600-bf73-11ea-920d-ebd4de28cbd1.png)

### Steps to Reproduce
I recreated on the preview instance [here](https://redash-preview.netlify.app/queries/394#681).

1. Write a simple query: `select 80085 "foo"`
2. Make a counter visualization
3. Click **Show Data Only** and observe how large the counter appears

### Technical details:

* Redash Version: V9 Beta
* Browser/OS: FF / Chrome / Safari
* How did you install Redash: SaaS

PR #5044, opened 2020-07-13
Counter is too large on Query View/Source pages

kravets-levko:

- [x] Other

Closes getredash/redash#5031

### DIFF
--- a/client/app/components/dashboards/dashboard-grid.less
+++ b/client/app/components/dashboards/dashboard-grid.less
@@ -113,15 +113,36 @@
     overflow: hidden;
   }
 
-  .counter-visualization-content {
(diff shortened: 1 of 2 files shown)


kravets-levko:
Dimensions I chose are quite arbitrary. Counters with smaller numbers still look quite big, but when there are 5-8-digits numbers, or longer label, or second row - it's much better. I worried that it may be even too small in some cases.

daniellangnet:
This PR broke the counter visualization entirely on Safari (both mobile and desktop). I'm actually surprised nobody else has noticed yet but I guess not too many people are running the latest version of Redash and use Safari. I'll try to fix and will submit a PR

PR #5049, opened 2020-07-16
Eager load outdated queries

rauchy:
- [x] Bug Fix

## Description
When there is a large amount of outdated queries, some short-running queries can execute more than once per interval. It appears that it is caused by a race condition in which the `outdated_queries` query takes several seconds to complete, during that time some jobs start executing and the local copy of `scheduled_query_executions` becomes stale.

This PR eager-loads the outdated queries before getting fresh values into `scheduled_query_executions`.

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -632,7 +632,7 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
 
     @classmethod
     def outdated_queries(cls):
-        queries = (
+        queries = list(
             Query.query.options(
                 joinedload(Query.latest_query_data).load_only("retrieved_at")
             )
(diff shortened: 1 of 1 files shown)


redash/models/__init__.py
```diff
@@ -632,7 +632,7 @@ def past_scheduled_queries(cls):
 
     @classmethod
     def outdated_queries(cls):
-        queries = (
+        queries = list(
```
arikfr, line 635: Worth adding a comment, as someone might eagerly refactor the code and remove this. 

I wonder if there is a way to add a test for this somehow?

jezdez: It may be better to append [`all()`](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.all) to this query statement instead since it's more obvious what it does.

rauchy: I like @jezdez's suggestion. It might even help with @arikfr's concern.

arikfr: 👍

PR #5050, opened 2020-07-16
Fix bundle-extensions script to work on recent importlib-resources.

jezdez:

- [x] Bug Fix

## Description

This was previously failing with an exception or even silently since the importlib-resources version in use has changed subtly when handling directories (compared to "resources" aka files). 

Also adds a test case for running the script so this should never happen.

#4939
8597b727a745abfff55ec4a8a1ed9187fdd172e4


arikfr:
Thanks.

PR #5052, opened 2020-07-17
Add Column Type to Databricks schema browser

gabrieldutra:

- [x] Feature

## Description
This PR adds a "Column Type" for both Databricks and Postgres query runner.

**Caveats**
- Column types are not unified in between different query runners, e.g.: Postgres uses "Integer" and Databricks shows as "Int". I'm not sure there's a "pretty" way of fixing this though. I tried to make sure it is Upper case in the schema browser and Capitalized for the Autocomplete.

--

Screen Shot 2020-07-21 at 22 22 15

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -72,9 +72,9 @@ def _wait(conn, timeout=None):
 
 def full_table_name(schema, name):
     if "." in name:
-        name = u'"{}"'.format(name)
+        name = '"{}"'.format(name)
 
-    return u"{}.{}".format(schema, name)
+    return "{}.{}".format(schema, name)
 
 
 def build_schema(query_result, schema):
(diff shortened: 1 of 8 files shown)


arikfr:
🙏 screenshots.

redash/query_runner/pg.py
```diff
@@ -72,9 +72,9 @@ def _wait(conn, timeout=None):
 
 def full_table_name(schema, name):
     if "." in name:
-        name = u'"{}"'.format(name)
+        name = '"{}"'.format(name)
```
gabrieldutra, line 75: The actual changes in this file are in https://github.com/getredash/redash/pull/5052/commits/ea50fd9476489f11a57b8abc15d19f55d6db1802 the rest is just Black formatting

kravets-levko:
> Column types are not unified in between different query runners, e.g.: Postgres uses "Integer" and Databricks shows as "Int". I'm not sure there's a "pretty" way of fixing this though.

No need to unify it, we'll always fail if we'll try - every DB has own custom data types. Showing a native DB type name is totally fine, because if user is already familiar with that DB they will recognize that data types, also they'll be able to find it in docs, etc. If we'll start mapping types (say, INT -> INTEGER) users may be confused.

arikfr:
> No need to unify it, we'll always fail if we'll try - every DB has own custom data types

I was thinking of using the Python data type of the object to define the type in the list. While you're right that users who are familiar with the database will find it useful to see the "real" names, users who aren't familiar are likely to find them confusing. And my guess is that we have more of the later group :)

Anyway, for now let's leave as is and revisit following some feedback from users.

gabrieldutra:
> Worth documenting somewhere how you added support for Postgres, so others can take example.

Maybe a forum post could motivate potential contributors?

PR #5053, opened 2020-07-17
Align Y axes at zero

kravets-levko:

- [x] Feature

## Description

Added an option to align Y-axes at zero

image
image


deecay:
Maybe "at" Zero, not "as" Zero?

viz-lib/src/visualizations/chart/Editor/YAxisSettings.jsx
```diff
@@ -32,6 +32,16 @@ export default function YAxisSettings({ options, onOptionsChange }) {
...
             />
           </Section>
+
+          <Section>
+            <Switch
+              id="chart-editor-y-axis-align-at-zero"
+              data-test="Chart.YAxis.AlignAtZero"
+              defaultChecked={options.alignYAxesAtZero}
+              onChange={alignYAxesAtZero => onOptionsChange({ alignYAxesAtZero })}>
+              Align Y Axes as Zero
```
kravets-levko, line 42:
```suggestion
              Align Y Axes at Zero
```

viz-lib/src/visualizations/chart/plotly/updateYRanges.js
```diff
@@ -4,6 +4,42 @@ function calculateAxisRange(range, min, max) {
...
+function calculateAbsoluteDiff(value, totalRange, percentageDiff) {
+  return (percentageDiff * totalRange) / (1 - Math.abs(value) / totalRange - percentageDiff);
+}
+
+function alignYAxesAtZero(axisA, axisB) {
+  // Make sure the origin is included in both axes
+  axisA.range[1] = Math.max(0, axisA.range[1]);
+  axisB.range[1] = Math.max(0, axisB.range[1]);
+  axisA.range[0] = Math.min(0, axisA.range[0]);
+  axisB.range[0] = Math.min(0, axisB.range[0]);
```
kravets-levko, line 16: That's actually very interesting piece. We can align axes at 0 when both (or maybe at leat one) of them cross 0 point. But if both have ranges that are completely above/below zero - should we really do anything? That's a good question

gabrieldutra: For me it works both with making sure the origin is included or with checking if it is included before. I.e.: You can add a guarding `if (!inRange(a) && !inRange(b)) return;` before everything at will.

Note: the origin must be included in both axes before running the code below, otherwise it can have negative values in the calculation, causing some weird results (basically it inverts the "by always augmenting").

kravets-levko:
Let's start with this implementation (many thanks to @gabrieldutra for the help), and see how it works for users. We can improve it later if needed

PR #5055, opened 2020-07-20
Add support for CSRF tokens

rauchy:
- [x] Feature

## Description
This PR adds CSRF protection for all forms and API calls. CSRF protection is turned off by default and can be turned on in `settings.ENFORCE_CSRF`.

If CSRF enforcement is turned on, it will protect:

1. Unauthenticated requests.
2. Requests authenticated using a classic web session (including XHR calls). 

It does _not_ apply to requests authenticated using an api key.

### DIFF
--- a/client/app/services/axios.js
+++ b/client/app/services/axios.js
@@ -1,7 +1,7 @@
 import axiosLib from "axios";
 import { Auth } from "@/services/auth";
 import qs from "query-string";
-import Cookies from 'js-cookie'
+import Cookies from "js-cookie";
 
 export const axios = axiosLib.create({
   paramsSerializer: params => qs.stringify(params),
@@ -16,7 +16,10 @@ axios.interceptors.request.use(config => {
   if (apiKey) {
     config.headers.Authorization = `Key ${apiKey}`;
   } else {
-    config.headers.common['X-CSRF-TOKEN'] = Cookies.get('csrf_token');
+    const csrfToken = Cookies.get("csrf_token");
+    if (csrfToken) {
+      config.headers.common["X-CSRF-TOKEN"] = csrfToken;
+    }
   }
 
   return config;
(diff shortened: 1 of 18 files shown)


client/app/services/axios.js
```diff
@@ -14,6 +15,8 @@ axios.interceptors.request.use(config => {
   const apiKey = Auth.getApiKey();
   if (apiKey) {
     config.headers.Authorization = `Key ${apiKey}`;
+  } else {
+    config.headers.common["X-CSRF-TOKEN"] = Cookies.get("csrf_token");
```
arikfr, line 19: We should probably:
1. Encapsulate the logic to get the CSRF token, so we can replace it.
2. Not set anything if no `csrf_token` available?

tests/__init__.py
```diff
@@ -20,6 +20,8 @@
 # Make sure rate limit is enabled
 os.environ["REDASH_RATELIMIT_ENABLED"] = "true"
 
+os.environ["REDASH_ENFORCE_CSRF"] = "false"
```
arikfr, line 23: Don't we want to have some tests with CSRF? Or it's covered by Cypress?

docker-compose.yml
```diff
@@ -15,6 +15,7 @@ x-redash-environment: &redash-environment
   REDASH_RATELIMIT_ENABLED: "false"
   REDASH_MAIL_DEFAULT_SENDER: "redash@example.com"
   REDASH_MAIL_SERVER: "email"
+  REDASH_ENFORCE_CSRF: "true"
```
arikfr, line 18: Don't you need to do the same in Cypress' docker-compose?

tests/__init__.py
rauchy: It felt like a behaviour that should be evaluated in an E2E perspective, so Cypress does the trick.

client/app/services/axios.js
rauchy: 1. `flask_wtf` handles the generation and verifications. I'm not sure it can be replaced (nor should it be?)
2. 8babdd0

arikfr:
I thought the Cypress failure is temporary so triggered another build, but it failed in the same way. Maybe somehow related to the changes?

PR #5058, opened 2020-07-21
Support multiple queries in a single query box

kravets-levko:

- [x] Feature

image

### DIFF
--- a/redash/query_runner/databricks.py
+++ b/redash/query_runner/databricks.py
@@ -36,6 +36,53 @@ def _build_odbc_connection_string(**kwargs):
     return ";".join([f"{k}={v}" for k, v in kwargs.items()])
 
 
+def split_sql_statements(query):
+    def strip_trailing_comments(stmt):
+        idx = len(stmt.tokens) - 1
+        while idx >= 0:
+            tok = stmt.tokens[idx]
+            if tok.is_whitespace or sqlparse.utils.imt(tok, i=sqlparse.sql.Comment, t=sqlparse.tokens.Comment):
+                stmt.tokens[idx] = sqlparse.sql.Token(sqlparse.tokens.Whitespace, ' ')
+            else:
+                break
+            idx -= 1
+        return stmt
+
+    def strip_trailing_semicolon(stmt):
+        idx = len(stmt.tokens) - 1
+        while idx >= 0:
+            tok = stmt.tokens[idx]
+            # we expect that trailing comments already are removed
+            if not tok.is_whitespace:
+                if sqlparse.utils.imt(tok, t=sqlparse.tokens.Punctuation) and tok.value == ";":
+                    stmt.tokens[idx] = sqlparse.sql.Token(sqlparse.tokens.Whitespace, ' ')
+                break
+            idx -= 1
+        return stmt
+
+    def is_empty_statement(stmt):
+        strip_comments = sqlparse.filters.StripCommentsFilter()
+
+        # copy statement object. `copy.deepcopy` fails to do this, so just re-parse it
+        st = sqlparse.engine.FilterStack()
+        stmt = next(st.run(sqlparse.text_type(stmt)))
+
+        sql = sqlparse.text_type(strip_comments.process(stmt))
+        return sql.strip() == ""
+
+    stack = sqlparse.engine.FilterStack()
+
+    result = [stmt for stmt in stack.run(query)]
+    result = [strip_trailing_comments(stmt) for stmt in result]
+    result = [strip_trailing_semicolon(stmt) for stmt in result]
+    result = [sqlparse.text_type(stmt).strip() for stmt in result if not is_empty_statement(stmt)]
+
+    if len(result) > 0:
+        return result
+
+    return [""]  # if all statements were empty - return a single empty statement
+
+
 class Databricks(BaseSQLQueryRunner):
     noop_query = "SELECT 1"
     should_annotate_query = False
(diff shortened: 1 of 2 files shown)


redash/query_runner/databricks.py
```diff
@@ -91,7 +92,12 @@ def run_query(self, query, user):
         try:
             cursor = self._get_cursor()
 
-            cursor.execute(query)
+            # TODO: don't run empty queries
+            # TODO: don't run queries only with comments
+            # TODO: handle case when no queries were executed
+            queries = sqlparse.split(query)
```
rxin, line 98: does this handle things correctly? would be great to have test cases to handle semicolon, semicolon in a string, `--` comments.

kravets-levko: @rxin We already discussed it with @arikfr this morning. Short answer - yes `sqlparse.split(...)` keeps everything - garbage, whitespaces, etc., and of course it properly handles semicolon and comments. This code is more PoC, to check if it will work. In fact it will require some additional processing, and I already started creating tests.

redash/query_runner/databricks.py
```diff
@@ -35,6 +36,53 @@ def _build_odbc_connection_string(**kwargs):
     return ";".join([f"{k}={v}" for k, v in kwargs.items()])
 
 
+def split_sql_statements(query):
```
gjxdxh, line 39: Since this is the function that works for all sql statement, Is it better to place it into the BaseSQLQueryRunner?

kravets-levko: It's not needed for all query runners, a lot of them support multiple sql statements on their own (e.g. PostgreSQL-based, MySQL/MariaDB, etc.). Also, this function does some extra job needed only for Databricks - it cannot handle comments after semicolon, so we have to strip them (but keep any garbage so user will get a proper error message). This logic most likely will be irrelevant for any other query runner

redash/query_runner/databricks.py
```diff
@@ -35,6 +36,53 @@ def _build_odbc_connection_string(**kwargs):
...
+    def strip_trailing_comments(stmt):
+        idx = len(stmt.tokens) - 1
+        while idx >= 0:
+            tok = stmt.tokens[idx]
+            if tok.is_whitespace or sqlparse.utils.imt(tok, i=sqlparse.sql.Comment, t=sqlparse.tokens.Comment):
+                stmt.tokens[idx] = sqlparse.sql.Token(sqlparse.tokens.Whitespace, ' ')
```
gjxdxh, line 45: If we already know this is a comment, can't we just remove it from our query list instead of replacing it with whitespace?

redash/query_runner/databricks.py
```diff
@@ -35,6 +36,53 @@ def _build_odbc_connection_string(**kwargs):
...
+        return stmt
+
+    def strip_trailing_semicolon(stmt):
+        idx = len(stmt.tokens) - 1
+        while idx >= 0:
+            tok = stmt.tokens[idx]
+            # we expect that trailing comments already are removed
+            if not tok.is_whitespace:
+                if sqlparse.utils.imt(tok, t=sqlparse.tokens.Punctuation) and tok.value == ";":
+                    stmt.tokens[idx] = sqlparse.sql.Token(sqlparse.tokens.Whitespace, ' ')
```
gjxdxh, line 58: Same as above. I think it's better to just remove it.

redash/query_runner/databricks.py
```diff
@@ -35,6 +36,53 @@ def _build_odbc_connection_string(**kwargs):
     return ";".join([f"{k}={v}" for k, v in kwargs.items()])
 
 
+def split_sql_statements(query):
+    def strip_trailing_comments(stmt):
+        idx = len(stmt.tokens) - 1
+        while idx >= 0:
+            tok = stmt.tokens[idx]
+            if tok.is_whitespace or sqlparse.utils.imt(tok, i=sqlparse.sql.Comment, t=sqlparse.tokens.Comment):
```
gjxdxh, line 44: will several comment be handled in this case? Like "-- comment1 -- comment2"

kravets-levko: It may happen only if each comment is on the new line - when they're on the same line it's the single comment with `--` inside of it. The function will handle it (that's why there is a `while` loop 😉 ), I'll add a tests case. Thank you!

redash/query_runner/databricks.py
```diff
@@ -35,6 +36,53 @@ def _build_odbc_connection_string(**kwargs):
...
+            idx -= 1
+        return stmt
+
+    def strip_trailing_semicolon(stmt):
+        idx = len(stmt.tokens) - 1
+        while idx >= 0:
+            tok = stmt.tokens[idx]
+            # we expect that trailing comments already are removed
+            if not tok.is_whitespace:
+                if sqlparse.utils.imt(tok, t=sqlparse.tokens.Punctuation) and tok.value == ";":
```
gjxdxh, line 57: What will happen if the query is like "SELECT * FROM A;;" Will originally invalid query become a valid query after we remove the last semicolon?

kravets-levko: There is a logic that just skips empty (including comment-only) statements, so this example will execute and return a result of `SELECT * FROM A`. Also there is a test case for it

gjxdxh: my point is that what will happen if we have two semicolon at the end of query. Will you remove one and make this invalid query become valid? I did not find such test case in you tests.

PR #5062, opened 2020-07-23
Load extensions on db init

rauchy:
- [x] Feature

## Description
Following #5008, it might be best to allow any custom database extensions to load before creating the database. This PR allows specifying these extensions using the dynamic settings module.

Also, it will only try to create tables and stamp the DB if the tables exist already.

#5008

### DIFF
--- a/redash/cli/database.py
+++ b/redash/cli/database.py
@@ -26,6 +26,13 @@ def _wait_for_db_connection(db):
 
         retried = True
 
+ 
+def is_db_empty():
+    from redash.models import db
+
+    table_names = sqlalchemy.inspect(db.get_engine()).get_table_names()
+    return len(table_names) == 0
+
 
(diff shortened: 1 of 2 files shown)


redash/cli/database.py
```diff
@@ -26,19 +27,38 @@ def _wait_for_db_connection(db):
...
+    from redash.models import db
+
+    table_names = sqlalchemy.inspect(db.get_engine()).get_table_names()
+    return len(table_names) == 0
+
+
+def load_extensions(db):
+    with db.engine.connect() as connection:
+        for extension in settings.dynamic_settings.database_extensions:
+            connection.execute(f'CREATE EXTENSION IF NOT EXISTS "{extension}";')
```
jezdez, line 41: f"yeah"

PR #5079, opened 2020-07-30
Use Skeleton as ItemsList loading state

gabrieldutra:

- [x] Other

## Description
I wanted to replace our Loading spinner for a while, so I'm exploring the option of using `Skeleton` for that (starting with ItemsList) and opening this for comments.

For now I'm keeping 10 rows and one thing I find a bit weird is when it's loading after we already had items in the screen (Changing sorting option for example). It does feel promising to be refined, so lmk what you think.

--

items-list-without-spinner

### DIFF
--- a/client/app/assets/less/inc/table.less
+++ b/client/app/assets/less/inc/table.less
@@ -1,149 +1,153 @@
 .table {
-    margin-bottom: 0;
-    
-    th.sortable-column {
-        cursor: pointer;
+  margin-bottom: 0;
+
+  th.sortable-column {
+    cursor: pointer;
+  }
+
+  &:not(.table-striped) > thead > tr > th {
+    background-color: #fafafa;
+  }
+
+  [class*="bg-"] {
+    & > tr > th {
+      color: #fff;
+      border-bottom: 0;
+      background: transparent !important;
     }
-    
-    &:not(.table-striped) > thead > tr > th {
-        background-color: #FAFAFA;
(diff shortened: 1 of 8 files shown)


gabrieldutra:
Alternatively without the spinner and only greying the results:

items-list-without-spinner

client/app/assets/less/inc/table.less
```diff
@@ -1,149 +1,153 @@
...
-    tbody > tr > td {
-        padding-top: 5px !important;
-    }
-    
-    .btn-favourite, .btn-archive {
-        font-size: 15px;
-    }
+  thead > tr > th {
+    white-space: nowrap;
+  }
```
gabrieldutra, line 100: This is the only thing I added into this file, the rest is on Prettier

gabrieldutra:
@arikfr @kravets-levko, this is certainly a start for having something other than the loading spinner we have. The grey loading when it already has loaded results is one of the things I most liked, it reduces sudden height changes and allows you to make page updates while loading.

The idea is to create a similar thing for the SettingsPages in a follow up, possibly allowing each column type to provide an specific skeleton loading (e.g.: if it's a button, show a button-skeleton) as [there are a few more options in antd v4](https://ant.design/components/skeleton/#components-skeleton-demo-element).

kravets-levko (approved):
Very nice @gabrieldutra! 🎉

PR #5082, opened 2020-07-31
New Query Runner - AWS RDS MySQL Data API

avnerenv0:
## What type of PR is this?

## Description
Adding a Data Source for AWS RDS Data API for MySQL.
We ([env0](https://www.env0.com/)) use Aurora **Serverless** RDS. 
The serverless variant cannot be exposed outside the VPC, but is exposed using AWS's Data API.
This would allow us to use the managed Redash product, without having to run a bastion with an SSH tunnel.

- https://discuss.redash.io/t/new-data-source-rds-data-api/6844/4


omryhay1:
This is a great feature, and we really need it!

bitmoji

tim5go:
@avnerenv0 
Would it be nicer if you make the RDS Query Runner as a generic class, so that others who want Aurora RDS, PostgreSQL RDS  etc. support can extend from your class?

avnerenv0:
> @avnerenv0
> Would it be nicer if you make the RDS Query Runner as a generic class, so that others who want Aurora RDS, PostgreSQL RDS etc. support can extend from your class?

I think you mean the "AWS Data API Query runner", (BTW also available for Redshift now).
I'm not one for early abstractions, as I wouldn't know what coded should be shared until I actually implement another source.
I would rather leave this as it is, and when a new source is added for a different DB with the Data API, the relevant PR can include extraction of whatever is needed. 
This is also probably a case where composition would be better than inheritance, but the "Data API Client" is pretty trivial - `boto` already abstracts it away. The only part that ~might~ repeat is the response data conversion, but again - I'm not sure it's the same for each DB type.

guidopetri:
@avnerenv0 , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

avnerenv0:
Hey @guidopetri 

The data API has been deprecated in Aurora v2, so I think this one can be closed. 

Thanks for your contribution - keep up the good work 🐱

PR #5088, opened 2020-08-04
Add default limit (1000) to SQL queries

### LINKED ISSUE #976: Limit number of records returned by a query
I typically run `SELECT * FROM ...` just to get a peak on the data and in many cases forget to add limit (yeah I know I know....) .

It would be nice if REDASH would have tunable default limit e.g. 10,000 rows (same way most of the desktop tools like TOAD or Navicat have). Not sure if this is easy to do for any data sources, its certainly trivial for postgres (it executes the limit as a last step).

Ideally displayed on the new query page right next to "refresh schedule". This way user knows about it and can easily change it.

gjxdxh:

- [x] Feature

## Description

Add a default limit 1000 for query execution, now it contains both the frontend changes(add a checkbox) and backend changes.

REDASH-112, Closes #976.

This is for data source that currently support auto limit
Screen Shot 2020-08-27 at 10 12 49 AM

This is for data source that currently not support auto limit
Screen Shot 2020-08-27 at 10 13 05 AM

### DIFF
--- a/client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
+++ b/client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
@@ -1,0 +1,37 @@
+import React, { useCallback } from "react";
+import PropTypes from "prop-types";
+import recordEvent from "@/services/recordEvent";
+import Checkbox from "antd/lib/checkbox";
+import Tooltip from "antd/lib/tooltip";
+
+export default function AutoLimitCheckbox({ available, checked, onChange }) {
+  const handleClick = useCallback(() => {
+    recordEvent("checkbox_auto_limit", "screen", "query_editor", { state: !checked });
+    onChange(!checked);
+  }, [checked, onChange]);
+
+  let tooltipMessage = null;
+  if (!available) {
+    tooltipMessage = "Auto limiting is not available for this Data Source type.";
+  } else {
+    tooltipMessage = "Auto limit results to first 1000 rows.";
+  }
+
+  return (
+    <Tooltip placement="top" title={tooltipMessage}>
+      <Checkbox
+        className="query-editor-controls-checkbox"
+        disabled={!available}
+        onClick={handleClick}
+        checked={available && checked}>
+        LIMIT 1000
+      </Checkbox>
+    </Tooltip>
+  );
+}
+
+AutoLimitCheckbox.propTypes = {
+  available: PropTypes.bool,
+  checked: PropTypes.bool.isRequired,
+  onChange: PropTypes.func.isRequired,
+};
--- a/client/app/components/queries/QueryEditor/QueryEditorControls.jsx
+++ b/client/app/components/queries/QueryEditor/QueryEditorControls.jsx
@@ -8,6 +8,7 @@ import KeyboardShortcuts, { humanReadableShortcut } from "@/services/KeyboardSho
 
 import AutocompleteToggle from "./AutocompleteToggle";
 import "./QueryEditorControls.less";
+import AutoLimitCheckbox from "@/components/queries/QueryEditor/AutoLimitCheckbox";
 
 export function ButtonTooltip({ title, shortcut, ...props }) {
   shortcut = humanReadableShortcut(shortcut, 1); // show only primary shortcut
@@ -38,6 +39,7 @@ export default function EditorControl({
   saveButtonProps,
   executeButtonProps,
   autocompleteToggleProps,
+  autoLimitCheckboxProps,
   dataSourceSelectorProps,
 }) {
   useEffect(() => {
@@ -84,6 +86,7 @@ export default function EditorControl({
           onToggle={autocompleteToggleProps.onToggle}
         />
       )}
+      {autoLimitCheckboxProps !== false && <AutoLimitCheckbox {...autoLimitCheckboxProps} />}
       {dataSourceSelectorProps === false && <span className="query-editor-controls-spacer" />}
       {dataSourceSelectorProps !== false && (
         <Select
@@ -153,6 +156,7 @@ EditorControl.propTypes = {
       onToggle: PropTypes.func,
     }),
   ]),
+  autoLimitCheckboxProps: PropTypes.oneOfType([PropTypes.shape(AutoLimitCheckbox.propTypes)]),
   dataSourceSelectorProps: PropTypes.oneOfType([
     PropTypes.bool, // `false` to hide
     PropTypes.shape({
@@ -175,5 +179,6 @@ EditorControl.defaultProps = {
   saveButtonProps: false,
   executeButtonProps: false,
   autocompleteToggleProps: false,
+  autoLimitCheckboxProps: false,
   dataSourceSelectorProps: false,
 };
--- a/client/app/components/queries/QueryEditor/QueryEditorControls.jsx
+++ b/client/app/components/queries/QueryEditor/QueryEditorControls.jsx
@@ -156,7 +156,10 @@ EditorControl.propTypes = {
       onToggle: PropTypes.func,
     }),
   ]),
-  autoLimitCheckboxProps: PropTypes.oneOfType([PropTypes.shape(AutoLimitCheckbox.propTypes)]),
+  autoLimitCheckboxProps: PropTypes.oneOfType([
+    PropTypes.bool, // `false` to hide
+    PropTypes.shape(AutoLimitCheckbox.propTypes),
+  ]),
   dataSourceSelectorProps: PropTypes.oneOfType([
     PropTypes.bool, // `false` to hide
     PropTypes.shape({
--- a/client/app/pages/queries/QuerySource.jsx
+++ b/client/app/pages/queries/QuerySource.jsx
@@ -26,6 +26,7 @@ import { getEditorComponents } from "@/components/queries/editor-components";
 import useQuery from "./hooks/useQuery";
 import useVisualizationTabHandler from "./hooks/useVisualizationTabHandler";
 import useAutocompleteFlags from "./hooks/useAutocompleteFlags";
+import useAutoLimitFlags from "./hooks/useAutoLimitFlags";
 import useQueryExecute from "./hooks/useQueryExecute";
 import useQueryResultData from "@/lib/useQueryResultData";
 import useQueryDataSources from "./hooks/useQueryDataSources";
@@ -77,6 +78,7 @@ function QuerySource(props) {
 
   const editorRef = useRef(null);
   const [autocompleteAvailable, autocompleteEnabled, toggleAutocomplete] = useAutocompleteFlags(schema);
+  const [autoLimitAvailable, autoLimitChecked, setAutoLimit] = useAutoLimitFlags(dataSource, query, setQuery);
 
   const [handleQueryEditorChange] = useDebouncedCallback(queryText => {
     setQuery(extend(query.clone(), { query: queryText }));
@@ -305,6 +307,11 @@ function QuerySource(props) {
                         available: autocompleteAvailable,
                         enabled: autocompleteEnabled,
                         onToggle: toggleAutocomplete,
+                      }}
+                      autoLimitCheckboxProps={{
+                        available: autoLimitAvailable,
+                        checked: autoLimitChecked,
+                        onChange: setAutoLimit,
                       }}
                       dataSourceSelectorProps={
                         dataSource
--- a/client/app/pages/queries/hooks/useAutoLimitFlags.js
+++ b/client/app/pages/queries/hooks/useAutoLimitFlags.js
@@ -3,19 +3,19 @@ import localOptions from "@/lib/localOptions";
 import { get, extend } from "lodash";
 
 function isAutoLimitAvailable(dataSource) {
-  return get(dataSource, "apply_auto_limit", false);
+  return get(dataSource, "supports_auto_limit", false);
 }
 
 export default function useAutoLimitFlags(dataSource, query, setQuery) {
   const isAvailable = isAutoLimitAvailable(dataSource);
   const [isChecked, setIsChecked] = useState(localOptions.get("applyAutoLimit", true));
-  query.options.applyAutoLimit = isAvailable && isChecked;
+  query.options.apply_auto_limit = isAvailable && isChecked;
 
   const setAutoLimit = useCallback(
     state => {
       setIsChecked(state);
       localOptions.set("applyAutoLimit", state);
-      setQuery(extend(query.clone(), { options: { ...query.options, applyAutoLimit: isAvailable && state } }));
+      setQuery(extend(query.clone(), { options: { ...query.options, apply_auto_limit: isAvailable && state } }));
     },
     [query, setQuery, isAvailable]
   );
--- a/client/app/pages/queries/hooks/useAutoLimitFlags.js
+++ b/client/app/pages/queries/hooks/useAutoLimitFlags.js
@@ -1,0 +1,24 @@
+import { useCallback, useState } from "react";
+import localOptions from "@/lib/localOptions";
+import { get, extend } from "lodash";
+
+function isAutoLimitAvailable(dataSource) {
+  return get(dataSource, "apply_auto_limit", false);
+}
+
+export default function useAutoLimitFlags(dataSource, query, setQuery) {
+  const isAvailable = isAutoLimitAvailable(dataSource);
+  const [isChecked, setIsChecked] = useState(localOptions.get("applyAutoLimit", true));
+  query.options.applyAutoLimit = isAvailable && isChecked;
+
+  const setAutoLimit = useCallback(
+    state => {
+      setIsChecked(state);
+      localOptions.set("applyAutoLimit", state);
+      setQuery(extend(query.clone(), { options: { ...query.options, applyAutoLimit: isAvailable && state } }));
+    },
+    [query, setQuery, isAvailable]
+  );
+
+  return [isAvailable, isChecked, setAutoLimit];
+}
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -187,7 +187,7 @@ export class Query {
   }
 
   getAutoLimit() {
-    return this.options.applyAutoLimit;
+    return this.options.apply_auto_limit;
   }
 
   getParametersDefs(update = true) {
--- a/client/app/services/query.js
+++ b/client/app/services/query.js
@@ -130,7 +130,8 @@ export class Query {
   }
 
   getQueryResult(maxAge) {
-    const execute = () => QueryResult.getByQueryId(this.id, this.getParameters().getExecutionValues(), maxAge);
+    const execute = () =>
+      QueryResult.getByQueryId(this.id, this.getParameters().getExecutionValues(), this.getAutoLimit(), maxAge);
     return this.prepareQueryResultExecution(execute, maxAge);
   }
 
@@ -141,7 +142,8 @@ export class Query {
     }
 
     const parameters = this.getParameters().getExecutionValues({ joinListValues: true });
-    const execute = () => QueryResult.get(this.data_source_id, queryText, parameters, maxAge, this.id);
+    const execute = () =>
+      QueryResult.get(this.data_source_id, queryText, parameters, this.getAutoLimit(), maxAge, this.id);
     return this.prepareQueryResultExecution(execute, maxAge);
   }
 
@@ -182,6 +184,10 @@ export class Query {
     }
 
     return this.$parameters;
+  }
+
+  getAutoLimit() {
+    return this.options.applyAutoLimit;
   }
 
   getParametersDefs(update = true) {
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -488,9 +488,9 @@ class QueryRefreshResource(BaseResource):
 
         parameter_values = collect_parameters_from_request(request.args)
         parameterized_query = ParameterizedQuery(query.query_text, org=self.current_org)
-
+        apply_auto_limit = query.options.get("apply_auto_limit", False)
         return run_query(
-            parameterized_query, parameter_values, query.data_source, query.id
+            parameterized_query, parameter_values, query.data_source, query.id, apply_auto_limit
         )
 
 
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -488,9 +488,9 @@ class QueryRefreshResource(BaseResource):
 
         parameter_values = collect_parameters_from_request(request.args)
         parameterized_query = ParameterizedQuery(query.query_text, org=self.current_org)
-        apply_auto_limit = query.options.get("apply_auto_limit", False)
+        should_apply_auto_limit = query.options.get("apply_auto_limit", False)
         return run_query(
-            parameterized_query, parameter_values, query.data_source, query.id, apply_auto_limit
+            parameterized_query, parameter_values, query.data_source, query.id, should_apply_auto_limit
         )
 
 
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -20,7 +20,6 @@ from redash.tasks import Job
 from redash.tasks.queries import enqueue_query
 from redash.utils import (
     collect_parameters_from_request,
-    gen_query_hash,
     json_dumps,
     utcnow,
     to_filename,
@@ -77,7 +76,7 @@ def run_query(query, parameters, data_source, query_id, apply_auto_limit, max_ag
     except (InvalidParameterError, QueryDetachedFromDataSourceError) as e:
         abort(400, message=str(e))
 
-    query_text = data_source.query_runner.apply_auto_limit(query.text, apply_auto_limit)
+    query_text = _apply_auto_limit(data_source, query, apply_auto_limit)
 
     if query.missing_params:
         return error_response(
@@ -123,6 +122,10 @@ def run_query(query, parameters, data_source, query_id, apply_auto_limit, max_ag
             },
         )
         return serialize_job(job)
+
+
+def _apply_auto_limit(data_source, query, apply_auto_limit):
+    return data_source.query_runner.apply_auto_limit(query.text, apply_auto_limit)
 
 
 def get_download_filename(query_result, query, filetype):
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -61,7 +61,7 @@ error_messages = {
 }
 
 
-def run_query(query, parameters, data_source, query_id, max_age=0):
+def run_query(query, parameters, data_source, query_id, apply_auto_limit, max_age=0):
     if data_source.paused:
         if data_source.pause_reason:
             message = "{} is paused ({}). Please try later.".format(
@@ -77,8 +77,7 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
     except (InvalidParameterError, QueryDetachedFromDataSourceError) as e:
         abort(400, message=str(e))
 
-    set_auto_limit = True  # TODO: get this from the API instead of hardcode
-    query_text = data_source.query_runner.apply_auto_limit(query.text, set_auto_limit)
+    query_text = data_source.query_runner.apply_auto_limit(query.text, apply_auto_limit)
 
     if query.missing_params:
         return error_response(
@@ -183,6 +182,7 @@ class QueryResultListResource(BaseResource):
         )
 
         parameterized_query = ParameterizedQuery(query, org=self.current_org)
+        apply_auto_limit = params.get("apply_auto_limit", False)
 
         data_source_id = params.get("data_source_id")
         if data_source_id:
@@ -196,7 +196,7 @@ class QueryResultListResource(BaseResource):
             return error_messages["no_permission"]
 
         return run_query(
-            parameterized_query, parameters, data_source, query_id, max_age
+            parameterized_query, parameters, data_source, query_id, apply_auto_limit, max_age
         )
 
 
@@ -289,6 +289,7 @@ class QueryResultResource(BaseResource):
         )
 
         allow_executing_with_view_only_permissions = query.parameterized.is_safe
+        apply_auto_limit = params.get("apply_auto_limit", False)
 
         if has_access(
             query, self.current_user, allow_executing_with_view_only_permissions
@@ -298,6 +299,7 @@ class QueryResultResource(BaseResource):
                 parameter_values,
                 query.data_source,
                 query_id,
+                apply_auto_limit,
                 max_age,
             )
         else:
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -60,7 +60,7 @@ error_messages = {
 }
 
 
-def run_query(query, parameters, data_source, query_id, apply_auto_limit, max_age=0):
+def run_query(query, parameters, data_source, query_id, should_apply_auto_limit, max_age=0):
     if data_source.paused:
         if data_source.pause_reason:
             message = "{} is paused ({}). Please try later.".format(
@@ -76,7 +76,7 @@ def run_query(query, parameters, data_source, query_id, apply_auto_limit, max_ag
     except (InvalidParameterError, QueryDetachedFromDataSourceError) as e:
         abort(400, message=str(e))
 
-    query_text = _apply_auto_limit(data_source, query, apply_auto_limit)
+    query_text = data_source.query_runner.apply_auto_limit(query.text, should_apply_auto_limit)
 
     if query.missing_params:
         return error_response(
@@ -122,10 +122,6 @@ def run_query(query, parameters, data_source, query_id, apply_auto_limit, max_ag
             },
         )
         return serialize_job(job)
-
-
-def _apply_auto_limit(data_source, query, apply_auto_limit):
-    return data_source.query_runner.apply_auto_limit(query.text, apply_auto_limit)
 
 
 def get_download_filename(query_result, query, filetype):
@@ -185,7 +181,7 @@ class QueryResultListResource(BaseResource):
         )
 
         parameterized_query = ParameterizedQuery(query, org=self.current_org)
-        apply_auto_limit = params.get("apply_auto_limit", False)
+        should_apply_auto_limit = params.get("apply_auto_limit", False)
 
         data_source_id = params.get("data_source_id")
         if data_source_id:
@@ -199,7 +195,7 @@ class QueryResultListResource(BaseResource):
             return error_messages["no_permission"]
 
         return run_query(
-            parameterized_query, parameters, data_source, query_id, apply_auto_limit, max_age
+            parameterized_query, parameters, data_source, query_id, should_apply_auto_limit, max_age
         )
 
 
@@ -292,7 +288,7 @@ class QueryResultResource(BaseResource):
         )
 
         allow_executing_with_view_only_permissions = query.parameterized.is_safe
-        apply_auto_limit = params.get("apply_auto_limit", False)
+        should_apply_auto_limit = params.get("apply_auto_limit", False)
 
         if has_access(
             query, self.current_user, allow_executing_with_view_only_permissions
@@ -302,7 +298,7 @@ class QueryResultResource(BaseResource):
                 parameter_values,
                 query.data_source,
                 query_id,
-                apply_auto_limit,
+                should_apply_auto_limit,
                 max_age,
             )
         else:
--- a/redash/handlers/query_results.py
+++ b/redash/handlers/query_results.py
@@ -77,6 +77,9 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
     except (InvalidParameterError, QueryDetachedFromDataSourceError) as e:
         abort(400, message=str(e))
 
+    set_auto_limit = True  # TODO: get this from the API instead of hardcode
+    query_text = data_source.query_runner.apply_auto_limit(query.text, set_auto_limit)
+
     if query.missing_params:
         return error_response(
             "Missing parameter value for: {}".format(", ".join(query.missing_params))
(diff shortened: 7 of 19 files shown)


tests/utils/test_query_limit.py
```diff
@@ -0,0 +1,41 @@
+import unittest
```
gjxdxh, line 1: This file test only contains for single query, multi query test is in test_basesql_queryrunner.py

tests/query_runner/test_basesql_queryrunner.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test_apply_auto_limit_origin_no_limit_2(self):
+        origin_query_text = "SELECT * FROM fake WHERE id IN (SELECT id FROM fake_2 LIMIT 200)"
+        query_text = self.query_runner.apply_auto_limit(origin_query_text)
+        self.assertEqual(origin_query_text + " LIMIT 1000", query_text)
+
+    def test_apply_auto_limit_non_select_query(self):
+        origin_query_text = "create table execution_times as " \
+                            "(select id, retrieved_at, data_source_id, query, runtime, query_hash " \
+                            "from query_results order by 1 desc)"
```
jezdez, line 34: Instead of using backslashes on every line when writing long string, it's better form to use implicit line continuation inside of braces here, e.g. 
```suggestion
        origin_query_text = (
            "create table execution_times as "
            "(select id, retrieved_at, data_source_id, query, runtime, query_hash "
            "from query_results order by 1 desc)"
        )
```
or triple quotes:
```suggestion
        origin_query_text = """
            create table execution_times as
            (select id, retrieved_at, data_source_id, query, runtime, query_hash
            from query_results order by 1 desc)
        """
```

redash/handlers/queries.py
```diff
@@ -488,9 +488,9 @@ def post(self, query_id):
 
         parameter_values = collect_parameters_from_request(request.args)
         parameterized_query = ParameterizedQuery(query.query_text, org=self.current_org)
-
+        apply_auto_limit = "apply_auto_limit" in query.options and query.options["apply_auto_limit"]
```
arikfr, line 491: This can be simplified (and be more idiomatic) using the `get` method of `dict`.

```suggestion
        apply_auto_limit = query.options.get("apply_auto_limit", False)
```

redash/query_runner/__init__.py
```diff
@@ -208,6 +217,17 @@ def _get_tables_stats(self, tables_dict):
...
                 tables_dict[t]["size"] = res[0]["cnt"]
 
+    @property
+    def supports_auto_limit(self):
+        return True
+
+    def apply_auto_limit(self, query_text):
+        query_list = sqlparse.split(query_text)
+        # we only check for last one in the list because it is the one that we show result
+        last_query = query_list[-1]
```
arikfr, line 227: I think this will not work correctly with the following:

```sql
select * from raw_events;
# comment
```

We should use the logic from #5058 once it's merged.

redash/tasks/queries/maintenance.py
```diff
@@ -51,9 +51,13 @@ def _should_refresh_query(query):
 
 def _apply_default_parameters(query):
     parameters = {p["name"]: p.get("value") for p in query.parameters}
+    logging.warning("!!!!!!!!!!!!!!!!!!!!!!!")
+    logging.warning(query.options)
```
arikfr, line 55: left over from testing?

arikfr (commented):
Did a quick run through it. Will need to revisit again to review the frontend code.

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
```diff
@@ -0,0 +1,28 @@
...
+    onChange(!checked);
+  }, [checked, onChange]);
+
+  return (
+    <Checkbox
+      className="query-editor-controls-checkbox m-l-5"
+      disabled={!available}
+      onClick={handleClick}
+      checked={available && checked}>
+      <label style={{ fontSize: 9 }}>Limit 1000</label>
```
arikfr, line 19: Let's use the default font size + make sure it's aligned vertically to the middle of the buttons.

CleanShot 2020-08-19 at 13 24 47

(I'm not sure if it's really in the middle in the screenshot, just added for illustration :-) ).

redash/query_runner/__init__.py
kravets-levko: It may become interesting when combining this feature and 5058. As far as I understand, we'll need to apply limit only to the last query (which we return to user), right?

gjxdxh: I see. I missed the case that user can type comment after query.

kravets-levko: But can databricks query runner execute such queries? I mean with a semicolon followed by a comment? Previously I was getting a syntax error in this case

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
gabrieldutra: Agreed with Arik

```suggestion
      Limit 1000
```

---
Also for me this is currently in the base line:
Screen Shot 2020-08-19 at 13 30 59

You can add this to the styling you have below (and remove the `.m-l-5` class):

```css
.query-editor-controls-checkbox {
  /* ... */
  margin: auto 5px;
}
```
After that:
Screen Shot 2020-08-19 at 13 36 59

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
```diff
@@ -0,0 +1,28 @@
...
+    </Checkbox>
+  );
+}
+
+AutoLimitCheckbox.propTypes = {
+  available: PropTypes.bool.isRequired,
```
gabrieldutra, line 25: Consider making this one not required and with default value as `false`. If by some weird reason it stops receiving the information I think it's better to assume the value is false.

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
```diff
@@ -0,0 +1,28 @@
...
+export default function AutoLimitCheckbox({ available, checked, onChange }) {
+  const handleClick = useCallback(() => {
+    recordEvent("checkbox_auto_limit", "screen", "query_editor", { state: !checked });
+    onChange(!checked);
+  }, [checked, onChange]);
+
+  return (
+    <Checkbox
+      className="query-editor-controls-checkbox m-l-5"
+      disabled={!available}
```
gabrieldutra, line 16: I wonder if it's better to just hide this checkbox in UI when the feature is not available instead of disabling it 🤔

I don't see a reason to keep it and it also doesn't self-explain on "why this is disabled"

redash/query_runner/__init__.py
kravets-levko: P.S. But looks that variant w/o semicolon may still be the case (last token is a comment): 
```sql
select * from raw_events
# comment
```

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
gjxdxh: We checked with Wenbo on this, and he agrees that we can just disable it instead of just hide it,

redash/query_runner/__init__.py
gjxdxh: Sure, I will update my code to use Levko's function once #5058 get merged into master.

arikfr: #5058 is merged.

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
yankeeguyu: Why limit default rows? why not limit by userself to edit rows? Like limit 2000 or limit anyrows?

kravets-levko: Just use `align-self: center` rule for the element that is a direct child of `.query-editor-controls`, and remove all margins/padding at all

gabrieldutra: For me it works both with `align-self: center` or the `margin`. I was just leading towards the margin because it has a `.m-l-5` that can be removed if the rule is in `margin`. The parent display is already flex anyway 🤷‍♂️.

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
gabrieldutra: WDYT about adding a tooltip with something like "Auto limiting is not available for this data source"?

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+export default function AutoLimitCheckbox({ available, checked, onChange }) {
+  const handleClick = useCallback(() => {
+    recordEvent("checkbox_auto_limit", "screen", "query_editor", { state: !checked });
+    onChange(!checked);
+  }, [checked, onChange]);
+
+  return (
+    <Checkbox
+      className="query-editor-controls-checkbox m-l-5"
```
gabrieldutra, line 15:
```suggestion
      className="query-editor-controls-checkbox"
```

This is covered in the `margin: auto 5px;`

client/app/components/queries/QueryEditor/index.jsx
```diff
@@ -12,7 +12,7 @@ import "./index.less";
 const editorProps = { $blockScrolling: Infinity };
 
 const QueryEditor = React.forwardRef(function(
-  { className, syntax, value, autocompleteEnabled, schema, onChange, onSelectionChange, ...props },
+  { className, syntax, value, autocompleteEnabled, autoLimitChecked, schema, onChange, onSelectionChange, ...props },
```
gabrieldutra, line 15: You can remove this prop from this file as it's not used here.

client/app/components/queries/QueryEditor/QueryEditorControls.jsx
```diff
@@ -84,6 +86,13 @@ export default function EditorControl({
           onToggle={autocompleteToggleProps.onToggle}
         />
       )}
+      {autoLimitCheckboxProps !== false && (
+        <AutoLimitCheckbox
+          available={autoLimitCheckboxProps.available}
+          checked={autoLimitCheckboxProps.checked}
+          onChange={autoLimitCheckboxProps.onChange}
+        />
```
gabrieldutra, line 94: This one is not a must, but @kravets-levko do you see any reason not to just 
```jsx
<AutoLimitCheckbox {...autoLimitCheckboxProps} />
```

In here and in similar ones?

client/app/pages/queries/hooks/useAutoLimitFlags.js
```diff
@@ -0,0 +1,24 @@
...
+function isAutoLimitAvailable(dataSource) {
+  return dataSource !== null && dataSource.apply_auto_limit;
+}
+
+export default function useAutoLimitFlags(dataSource, query, setQuery) {
+  const isAvailable = isAutoLimitAvailable(dataSource);
+  const [isChecked, setIsChecked] = useState(localOptions.get("ApplyAutoLimit", true));
+  query.options.applyAutoLimit = isAvailable && isChecked;
+
+  const checkboxAutoLimit = useCallback(
```
gabrieldutra, line 14: Maybe call it `setAutoLimit` or `toggleAutoLimit`? It looks more like the action that will be done.

client/app/pages/queries/hooks/useAutoLimitFlags.js
```diff
@@ -0,0 +1,24 @@
+import { useCallback, useMemo, useState } from "react";
+import localOptions from "@/lib/localOptions";
+import { extend } from "lodash";
+
+function isAutoLimitAvailable(dataSource) {
+  return dataSource !== null && dataSource.apply_auto_limit;
```
gabrieldutra, line 6: Just in case it can be `undefined`, let's go with either `!!dataSource && dataSource.apply_auto_limit`, or with the version I personally like more, with lodash `get(dataSource, "apply_auto_limit", false)`

redash/query_runner/__init__.py
gjxdxh: Sure, just update the code to use that function

client/app/components/queries/QueryEditor/QueryEditorControls.jsx
kravets-levko: No, I see no reason for not just using a spread operator

yankeeguyu:
Hi everyone, this PR can be changed to limit any rows? Thank you!

gjxdxh:
@arikfr I can run the unit tests successfully on my own laptop, but here it always has several tests related to the query_runner that are failing. Do you know how can I make it pass?

gjxdxh:
> Hi everyone, this PR can be changed to limit any rows? Thank you!

If people want to limit any rows, they can just add "LIMIT ..." to their query, this PR is mainly to give a default option for the user, which is 1000.

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
gjxdxh: Just added!

yankeeguyu:
> > Hi everyone, this PR can be changed to limit any rows? Thank you!
> 
> If people want to limit any rows, they can just add "LIMIT ..." to their query, this PR is mainly to give a default option for the user, which is 1000.

Ok, thank you!

client/app/services/query.js
```diff
@@ -184,6 +186,10 @@ export class Query {
     return this.$parameters;
   }
 
+  applyAutoLimit() {
+    return this.options.applyAutoLimit;
+  }
+
```
gabrieldutra, line 192: I'd prefer to access directly from `this.options.applyAutoLimit` in the methods above as it's only used in this class anyway.

Could also rename it or make it a getter, `applyAutoLimit` implies that it will execute something to apply the auto limit 😬.

client/app/components/queries/QueryEditor/index.jsx
```diff
@@ -158,6 +158,8 @@ QueryEditor.propTypes = {
   syntax: PropTypes.string,
   value: PropTypes.string,
   autocompleteEnabled: PropTypes.bool,
+  autoLimitChecked: PropTypes.bool,
+  autoLimitAvailable: PropTypes.bool,
```
gabrieldutra, line 162: Those + the default props below can also be removed as they are not used here (and it actually displays an alert in the console)
Screen Shot 2020-09-02 at 18 31 12

client/app/pages/queries/QuerySource.jsx
```diff
@@ -265,6 +267,7 @@ function QuerySource(props) {
                       value={query.query}
                       schema={schema}
                       autocompleteEnabled={autocompleteAvailable && autocompleteEnabled}
+                      autoLimitChecked={autoLimitAvailable && autoLimitChecked}
```
gabrieldutra, line 270:
```suggestion
```
Can also remove it from here

gabrieldutra (commented):
Overall frontend looks nice 🎉, left some last comments :).

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
```diff
@@ -0,0 +1,36 @@
...
+export default function AutoLimitCheckbox({ available, checked, onChange }) {
+  const handleClick = useCallback(() => {
+    recordEvent("checkbox_auto_limit", "screen", "query_editor", { state: !checked });
+    onChange(!checked);
+  }, [checked, onChange]);
+
+  let tooltipMessage = null;
+  if (!available) {
+    tooltipMessage = "Auto limiting is not available for this data source";
+  }
```
arikfr, line 17:
```suggestion
  } else {
    tooltipMessage = "Auto limit results to first 1000 rows."
  }
```

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+export default function AutoLimitCheckbox({ available, checked, onChange }) {
+  const handleClick = useCallback(() => {
+    recordEvent("checkbox_auto_limit", "screen", "query_editor", { state: !checked });
+    onChange(!checked);
+  }, [checked, onChange]);
+
+  let tooltipMessage = null;
+  if (!available) {
+    tooltipMessage = "Auto limiting is not available for this data source";
```
arikfr, line 16:
```suggestion
    tooltipMessage = "Auto limiting is not available for this Data Source type.";
```

client/app/components/queries/QueryEditor/QueryEditorControls.jsx
```diff
@@ -153,6 +156,14 @@ EditorControl.propTypes = {
       onToggle: PropTypes.func,
     }),
   ]),
+  autoLimitCheckboxProps: PropTypes.oneOfType([
+    PropTypes.bool, // `false` to hide
+    PropTypes.shape({
```
arikfr, line 161: Can we use `PropTypes.shape(AutoLimitCheckbox.propTypes)` here to avoid duplication?

client/app/components/queries/QueryEditor/QueryEditorControls.jsx
```diff
@@ -153,6 +156,14 @@ EditorControl.propTypes = {
       onToggle: PropTypes.func,
     }),
   ]),
+  autoLimitCheckboxProps: PropTypes.oneOfType([
+    PropTypes.bool, // `false` to hide
```
arikfr, line 160: Is this still used?

client/app/pages/queries/hooks/useAutoLimitFlags.js
```diff
@@ -0,0 +1,24 @@
...
+import localOptions from "@/lib/localOptions";
+import { get, extend } from "lodash";
+
+function isAutoLimitAvailable(dataSource) {
+  return get(dataSource, "apply_auto_limit", false);
+}
+
+export default function useAutoLimitFlags(dataSource, query, setQuery) {
+  const isAvailable = isAutoLimitAvailable(dataSource);
+  const [isChecked, setIsChecked] = useState(localOptions.get("ApplyAutoLimit", true));
```
arikfr, line 11:
```suggestion
  const [isChecked, setIsChecked] = useState(localOptions.get("applyAutoLimit", true));
```

Small thing to be consistent with other options.

client/app/pages/queries/hooks/useAutoLimitFlags.js
```diff
@@ -0,0 +1,24 @@
...
+  query.options.applyAutoLimit = isAvailable && isChecked;
+
+  const setAutoLimit = useCallback(
+    state => {
+      setIsChecked(state);
+      localOptions.set("ApplyAutoLimit", state);
```
arikfr, line 17:
```suggestion
      localOptions.set("applyAutoLimit", state);
```

client/app/pages/queries/hooks/useAutoLimitFlags.js
```diff
@@ -0,0 +1,24 @@
...
+      setQuery(extend(query.clone(), { options: { ...query.options, applyAutoLimit: isAvailable && state } }));
+    },
+    [query, setQuery, isAvailable]
+  );
+
+  return useMemo(() => [isAvailable, isChecked, setAutoLimit], [isAvailable, isChecked, setAutoLimit]);
```
arikfr, line 23: Why the `useMemo` here?

redash/models/__init__.py
```diff
@@ -122,6 +122,7 @@ def to_dict(self, all=False, with_permissions_for=None):
             "syntax": self.query_runner.syntax,
             "paused": self.paused,
             "pause_reason": self.pause_reason,
+            "apply_auto_limit": isinstance(self.query_runner, BaseSQLQueryRunner)
```
arikfr, line 125:
```suggestion
            "apply_auto_limit": self.query_runner.supports_auto_limit
```

tests/query_runner/test_basesql_queryrunner.py
```diff
@@ -0,0 +1,88 @@
+import unittest
+
+from redash.query_runner import BaseSQLQueryRunner, BaseQueryRunner
+
+
+class TestBaseSQLQueryRunner(unittest.TestCase):
+
+    def setUp(self):
+        self.query_runner = BaseSQLQueryRunner(BaseQueryRunner)
```
arikfr, line 9:
```suggestion
        self.query_runner = BaseSQLQueryRunner({})
```

(first parameter is configuration)

redash/tasks/queries/maintenance.py
```diff
@@ -52,9 +52,11 @@ def _should_refresh_query(query):
 
 def _apply_default_parameters(query):
```
arikfr, line 53: One issue with what we're doing here is that it changes the query text -> changes the query hash, which in the end will result in `latest_query_data_id` not being updated for this query.

It's not a huge issue, but can lead to some issues (like not showing when it was last refreshed). 

One way to solve it is to generate the query_hash of the Query object taking into account the limit part. This happens here:

https://github.com/getredash/redash/blob/205915e6db802d61667c14f8978d0c02fcccfe46/redash/models/__init__.py#L868-L871

@rauchy wdyt?

arikfr (commented):
Looks good! Some small UI changes & a bigger concern in the backend code (see comment).

client/app/pages/queries/hooks/useAutoLimitFlags.js
gabrieldutra: It was probably just copied from other hooks 😅, but it is indeed safe to remove @gjxdxh

redash/tasks/queries/maintenance.py
rauchy: I'm sorry for chiming in late here, but why is a function called `apply_default_parameters` responsible for applying limits? Also there's [a small DRY issue](https://github.com/getredash/redash/pull/5088/files/471586eabac9a23110d4751172e04944441ee887#diff-f734d32676ec526e17e5ae92d41b5787R82) with the same problem.

The way I see it - it looks like query runners are the ones who _know_ if they support limiting, and they are also the ones who _apply_ the limit, so it feels like limiting should be encapsulated inside them. Current situation is "asking, not telling": `if query_runner.supports_limiting then query_runner.apply_limit(text)`. I'd prefer a [Tell, Don't Ask](https://thoughtbot.com/blog/tell-dont-ask).

I have a feeling this design is caused by continuous creep of suboptimal design in which query runners work with query text, not query objects. This results in solution sprawl and, as @arikfr pointed out - collision with other parts of the system, such as query hashes.

My bottom line - if query runners own this behaviour, make the changes needed to provide them the information they need and let them perform limiting without changing anything outside the query runners. I believe this will also solve the hashing problem.

client/app/components/queries/QueryEditor/AutoLimitCheckbox.jsx
```diff
@@ -0,0 +1,36 @@
+import React, { useCallback } from "react";
+import PropTypes from "prop-types";
+import "@/redash-font/style.less";
+import recordEvent from "@/services/recordEvent";
+import { Checkbox } from "antd";
```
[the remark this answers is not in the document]
gjxdxh: This is still not working. I tried and it compiled with error.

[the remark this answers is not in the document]
gabrieldutra: Sorry, there was a typo in my suggestion 😅 
```suggestion
import Checkbox from "antd/lib/checkbox";
```

client/app/components/queries/QueryEditor/index.jsx
gjxdxh: I think this is necessary because we have to give it a default value as I did in line 173. The warning is because previously I deleted the definition in const QueryEditor, now I added it back and it works fine.

gabrieldutra: @gjxdxh if you remove it from every QueryEditor call it should work as the QueryEditor doesn't make any use of the `autoLimit` information (the QueryEditorControls is a separate component).

redash/tasks/queries/maintenance.py
gjxdxh: According to @rauchy suggestion, I just made several changes to the code( moving most of the logic to query_runner etc). But the query hash won't be solved because every time we refresh, we are changing the query text and results in different hash. I am going to think about how to solve this and let me know if you have any ideas. Thanks!

client/app/pages/queries/hooks/useAutoLimitFlags.js
```diff
@@ -0,0 +1,24 @@
...
+import { get, extend } from "lodash";
+
+function isAutoLimitAvailable(dataSource) {
+  return get(dataSource, "apply_auto_limit", false);
+}
+
+export default function useAutoLimitFlags(dataSource, query, setQuery) {
+  const isAvailable = isAutoLimitAvailable(dataSource);
+  const [isChecked, setIsChecked] = useState(localOptions.get("applyAutoLimit", true));
+  query.options.applyAutoLimit = isAvailable && isChecked;
```
gabrieldutra, line 12: Is this to guarantee the `applyAutoLimit` is set to false when not available? If so, how about we put this line into a

```jsx
useEffect(() => {
  query.options.applyAutoLimit = isAvailable && isChecked;
}, [isAvailable, isChecked]);
```
? This way it will update that value only when one the vars change and not on every render.

arikfr: Maybe I'm missing something, but it's a simple operation between two boolean values which most of the time yields the same value (so no change in rendering). Why complicate the code with a `useEffect`?

redash/tasks/queries/maintenance.py
arikfr: That's a good point, @rauchy. But such a refactor is out of scope for this PR. Also, I'm not 100% convinced that handling parameters should be in the Query Runner code. If we go this path, then even bigger refactor needed to make sure there is minimal duplication between the different query runner.

Also on its own it won't solve the query hash problem: it is the right behavior to assign a different hash for the limited query - otherwise we might update cache for wrong queries. Which leads me to believe that updating the query hash based on the current `apply_limit` value of the query is the right thing to do (at least given current design).

client/app/pages/queries/hooks/useAutoLimitFlags.js
gabrieldutra: I thought it mostly not to be reassigning a value uneedlessly, it just felt a bit weird to mutate such value on every re-render, so my take was to limit it with `useEffect`. It is a boolean value, so it won't trigger any rerender indeed, it's more on the case it was updated out of this hook's context. Which thinking now, doesn't happen 😅

arikfr: :-)

We need to be careful with such optimizations, because in this case it's likely making the code less optimized (instead of boolean assignment, we get functions calls, function definitions, etc) but what's even worse is that makes the code less readable.

redash/tasks/queries/maintenance.py
gjxdxh: Just updated the code with fix for query hash. I tested locally and queries are refreshed succesfully, let me know if you have any questions!

redash/utils/__init__.py
```diff
@@ -48,19 +49,28 @@ def slugify(s):
...
+def gen_query_hash(sql, data_source=None, set_auto_limit=False):
     """Return hash of the given query after stripping all comments, line breaks
     and multiple spaces, and lower casing all text.
 
     TODO: possible issue - the following queries will get the same id:
         1. SELECT 1 FROM table WHERE column='Value';
         2. SELECT 1 FROM table where column='value';
     """
+    if data_source is not None:
+        sql = data_source.query_runner.apply_auto_limit(sql, set_auto_limit)
```
rauchy, line 61: This is heading in a good direction! I still feel like this abstraction is leaking though.

`gen_query_hash` was previously a happy contextless function, floating in a contextless space. But now, it depends on more context and its implementation varies in different query runners.

Why not move `gen_query_hash` to `QueryRunnerBase`, with the previous implementation as default, and a polymorphic override where auto limits apply? (i.e `super.gen_query_hash(sql_after_applying_limit)`)

This will help reducing the sprawl of booleans moving between functions. WDYT?

arikfr: I agree that this logic shouldn't be here: we should resolve the correct SQL to hash before calling this function.

As for calculating the hash in the Query Runner: we need it beforehand because we use it for the lock. Going back to my note about a bigger refactor: we should probably lock on query id vs. query text. The query text locking was an optimization that these days I doubt is needed.

client/app/pages/queries/hooks/useAutoLimitFlags.js
```diff
@@ -0,0 +1,24 @@
...
+
+export default function useAutoLimitFlags(dataSource, query, setQuery) {
+  const isAvailable = isAutoLimitAvailable(dataSource);
+  const [isChecked, setIsChecked] = useState(localOptions.get("apply_auto_limit", true));
+  query.options.apply_auto_limit = isAvailable && isChecked;
+
+  const setAutoLimit = useCallback(
+    state => {
+      setIsChecked(state);
+      localOptions.set("apply_auto_limit", state);
```
arikfr, line 17: Why the switch to snake case?

redash/models/__init__.py
```diff
@@ -122,6 +123,7 @@ def to_dict(self, all=False, with_permissions_for=None):
             "syntax": self.query_runner.syntax,
             "paused": self.paused,
             "pause_reason": self.pause_reason,
+            "apply_auto_limit": self.query_runner.supports_auto_limit
```
arikfr, line 126:
```suggestion
            "supports_auto_limit": self.query_runner.supports_auto_limit
```

This should be `supports_auto_limit` too. 

Sorry for the nits, it just that these things are very hard to change later.

redash/handlers/query_results.py
```diff
@@ -85,7 +87,7 @@ def run_query(query, parameters, data_source, query_id, max_age=0):
     if max_age == 0:
         query_result = None
     else:
-        query_result = models.QueryResult.get_latest(data_source, query.text, max_age)
+        query_result = models.QueryResult.get_latest(data_source, query_text, max_age, apply_auto_limit)
```
arikfr, line 90: You already apply auto limit above, so no need to apply it again in `get_latest`. Just like `gen_query_hash`, `get_latest` shouldn't care about applying a limit -- it just get a query text and return a cached version for it, if available.

redash/models/__init__.py
```diff
@@ -865,10 +867,14 @@ def dashboard_api_keys(self):
         return [api_key[0] for api_key in api_keys]
 
 
-@listens_for(Query.query_text, "set")
-def gen_query_hash(target, val, oldval, initiator):
-    target.query_hash = utils.gen_query_hash(val)
-    target.schedule_failures = 0
+@listens_for(Query, "before_insert")
```
arikfr, line 870: I'm pretty sure you can apply both decorator to a single method:

```python
@listens_for(Query, "before_insert")
@listens_for(Query, "before_update")
def update_query_hash(...):
    ...
```

redash/utils/__init__.py
rauchy: > we need it beforehand because we use it for the lock

If you mean [this](https://github.com/getredash/redash/blob/default_limit/redash/tasks/queries/execution.py#L33), then the data source is available there and we can ask for it to calculate the hash, no?

gjxdxh: quick question: Will every query always have a data source? If yes, then I think we can move this function to query_runner. And I agree with Omer that the "we need it beforehand because we use it for the lock" won't be a problem because we are just going to call this function from query_runner instead of utils

client/app/pages/queries/hooks/useAutoLimitFlags.js
gjxdxh: Because we did not have parsing when save the query. So to keep it consistent with the query execution, we have to either add parsing when saving, or change it to snake case. In order to keep things simple, I think it's better to just change it to snake case instead of adding a separate parsing.

redash/utils/__init__.py
gjxdxh: Just updated the code and put this logic into the query runner. 
Note: In several places I am using `BaseQueryRunner({}).gen_query_hash(querytext, False)` instead of using the query's own datasource queryRunner because in that kinds of places, we have already applied the limit, so to keep logic simple, I just use gen_query_hash from BaseQueryRunner. Otherwise, we have to do something like 
```
if self.data_source is not None:
        query_runner = self.data_source.query_runner
else:
        query_runner = BaseQueryRunner({})
self.query_hash = query_runner.gen_query_hash(self.query_text, False)
```
Since we already know that the second argument will always be False, there is no need to do this check.

arikfr (commented):
Looks good now. One last question re. the localOptions.

client/app/pages/queries/hooks/useAutoLimitFlags.js
gjxdxh: Oh for this one yes, but since I changed the `query.options.applyAutoLimit` to `query.options.apply_auto_limit`because of the reason I said above, I feel like it's better to keep them consistent, so I also renamed this one.

gjxdxh:
@arikfr Any ideas how to make the unit-test pass? It's passing on my local machine, but not on this CircleCI

client/app/pages/queries/hooks/useAutoLimitFlags.js
gabrieldutra: I think it's not a problem to have it as `query.options.applyAutoLimit`, overall it's common to use the camel case for options (e.g.: for widgets and visualizations). But it would be better to have the `localOption` as `applyAutoLimit` regardless as it is already a standard for other local storage options.

gjxdxh: Sure, I can change this localOption to `applyAutoLimit` if you think that's better

arikfr:
> @arikfr Any ideas how to make the unit-test pass? It's passing on my local machine, but not on this CircleCI

Yes, the data sources you were trying to use aren't available in the CI because of missing packages. You can use `pg` as the SQL variant and pick something like `prometheus` as the NoSQL variant.

client/app/pages/queries/hooks/useAutoLimitFlags.js
arikfr: Let's change it only in `localOptions` because it's a client side thing (and to be consistent with the other settings we have there). We already have these differences between objects from backend and objects from frontend in other places.

client/app/components/queries/QueryEditor/QueryEditorControls.jsx
```diff
@@ -153,6 +156,7 @@ EditorControl.propTypes = {
       onToggle: PropTypes.func,
     }),
   ]),
+  autoLimitCheckboxProps: PropTypes.oneOfType([PropTypes.shape(AutoLimitCheckbox.propTypes)]),
```
gabrieldutra, line 159:
```suggestion
  autoLimitCheckboxProps: PropTypes.oneOfType([
    PropTypes.bool, // `false` to hide
    PropTypes.shape(AutoLimitCheckbox.propTypes),
  ]),
```

I know that @arikfr asked about this one. Although this is not used for this prop, the `false` value can be used to hide the element (check both the use in the component and the `defaultProps` below).

gabrieldutra (approved):
One last small thing I noticed, but frontend changes LGTM 🚀

yankeeguyu:
Hi everyone, if someone also need to download complete file with DISPLAY LIMIT 1000 rows. How to get it?

redash/query_runner/__init__.py
```diff
@@ -208,6 +219,22 @@ def _get_tables_stats(self, tables_dict):
...
                 tables_dict[t]["size"] = res[0]["cnt"]
 
+    @property
+    def supports_auto_limit(self):
+        return True
+
+    def apply_auto_limit(self, query_text, set_auto_limit):
+        if set_auto_limit:
+            from redash.query_runner.databricks import split_sql_statements
+            query_list = split_sql_statements(query_text)
```
rauchy, line 229: How about renaming this to `queries`? Otherwise it feels a bit like `user_name_string`.

redash/query_runner/__init__.py
```diff
@@ -208,6 +219,22 @@ def _get_tables_stats(self, tables_dict):
...
+            # we only check for last one in the list because it is the one that we show result
+            last_query = query_list[-1]
+            if query_is_select_no_limit(last_query):
+                query_list[-1] = add_limit_to_query(last_query)
+            return ";\n".join(query_list)
```
rauchy, line 234: This is a leaky abstraction. The knowledge of splitting is inside `split_sql_statements`, but gluing them back together is here.

redash/utils/__init__.py
```diff
@@ -213,3 +212,34 @@ def render_template(path, context):
...
+
+def find_last_keyword_idx(parsed_query):
+    for i in reversed(range(len(parsed_query.tokens))):
+        if parsed_query.tokens[i].ttype in sqlparse.tokens.Keyword:
+            return i
+    return -1
+
+
+def add_limit_to_query(query):
+    # this function will only be called if query_is_select_no_limit returns True
```
rauchy, line 237:
```suggestion
```
image

rauchy (changes requested):

redash/query_runner/__init__.py
```diff
@@ -208,6 +219,22 @@ def _get_tables_stats(self, tables_dict):
                 res = self._run_query_internal("select count(*) as cnt from %s" % t)
                 tables_dict[t]["size"] = res[0]["cnt"]
 
+    @property
+    def supports_auto_limit(self):
+        return True
+
+    def apply_auto_limit(self, query_text, set_auto_limit):
+        if set_auto_limit:
+            from redash.query_runner.databricks import split_sql_statements
```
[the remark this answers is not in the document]
gjxdxh: I also feel like this should not be put in the databricks runner. But it seems out of scope for this project as it's already in it.

rauchy (approved):
LGTM!

giphy-6

arikfr:
Waiting for the tests to pass and then merging 💯

PR #5091, opened 2020-08-05
Remove content width limit on all pages

kravets-levko:

- [x] Other

Before and after (just an example, other pages are similar):

image
image

### DIFF
--- a/client/app/assets/less/inc/base.less
+++ b/client/app/assets/less/inc/base.less
@@ -78,8 +78,6 @@ strong {
   }
 }
 
-// Fixed width layout for specific pages
-
 .settings-screen,
 .home-page,
 .page-dashboard-list,
@@ -89,7 +87,7 @@ strong {
 .admin-page-layout {
   .container {
     width: 100%;
-    max-width: 1200px;
+    max-width: none;
   }
 }
 
--- a/client/app/assets/less/inc/base.less
+++ b/client/app/assets/less/inc/base.less
@@ -87,7 +87,7 @@ strong {
 .admin-page-layout {
   .container {
     width: 100%;
-    max-width: none;
+    max-width: 1600px;
   }
 }
 
--- a/client/app/assets/less/inc/base.less
+++ b/client/app/assets/less/inc/base.less
@@ -87,7 +87,7 @@ strong {
 .admin-page-layout {
   .container {
     width: 100%;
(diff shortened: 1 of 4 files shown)


client/app/assets/less/inc/base.less
```diff
@@ -89,7 +87,7 @@ strong {
 .admin-page-layout {
   .container {
     width: 100%;
-    max-width: 1200px;
+    max-width: none;
```
gabrieldutra, line 90:
```suggestion
    max-width: 1600px;
```

How about we just increase the limit? It can be a bit weird for large screens.

leini-db: it does looks weird. can we play with the placement/size to make it looks better?

kravets-levko: I tried `1600px` on my 1920*1080 screen, but liked no limit variant more. But we can vote :-)

gabrieldutra: Idk, I think I'd rather go with an even larger max-width than with `none`, with `none` we may have to start introducing max-width in some other elements (e.g.: the sidebar for list pages).

kravets-levko:
@gabrieldutra @leini-db I tried another approach: remove width limit (as it was initially), but limit sidebar width to `350px`. So sidebar will not become too large giving more space for the table. See screenshot how it looks on 1920x1080 screen:

image

kravets-levko:
@gabrieldutra I checked all pages in the app, and IMHO all of them look nice. List pages (dashboards, queries, alerts, etc.) are just fine - sidebar has limited width and table takes all free space; considering that tables usually have a lot of columns it's probably even better. Alert create/view/edit pages looks more or less fine - there is just some free space between form and sidebar with alert destinations. Query View/Edit page, Dashboard page are not affected as they have own layout. System status pages also look good. Probably the only concern are pages with forms - profile, DS/destination edit pages - they are centered and on wide screens there are a lot of white space, but as far as I remember that was a problem even before these changes. So I'd rather prefer to remove limit and tweak individual pages when needed. WDYT?

gabrieldutra:
> So I'd rather prefer to remove limit and tweak individual pages when needed. WDYT?

Go with it, it's better than the existing limit and the sidebar was my biggest concern anyways, we can revisit if anything comes up 👍.

PR #5100, opened 2020-08-11
Multi-statements support for ClickHouse query runner

gggsmith:
ClickHouse query runner splits query into several and execute each query in turn.
The result of the last execution is returned.
Implementation uses ClickHouse sessions in the HTTP protocol (https://clickhouse.tech/docs/en/interfaces/http/). session_id is generated for the first query and then it is used with the subsequent queries (together with the session_check parameter).
If query runner gets a success response with empty body from ClickHouse (for example, in case of temporary table creation request) query runner returns empty response.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
The changes allow using multi-statements in ClickHouse queries.
Was discussed here: https://discuss.redash.io/t/clickhouse-multi-statements-are-not-allowed-errors-how-to-debug/5713/3


redash/query_runner/clickhouse.py
```diff
@@ -10,6 +11,12 @@
 logger = logging.getLogger(__name__)
 
 
+def split_multi_query(query):
+    queries = re.split(r";[\s;]*$", query, flags=re.MULTILINE)
+
+    return [q.strip() for q in queries if q and not q.isspace()]
+
+
```
kravets-levko, line 19: Generally it's not a good idea to split statements with regex. We already have a `sqlparse` dependency that could be used for this purpose. I recently implemented similar feature for Databricks query runner (getredash/redash#5058), you can check my code for the reference

gggsmith: Wow, it took me a while to fix that :) Done!

gggsmith:
HI! Could you please tell whether this MR have a chance? Thanks!

susodapop:
Thanks for the ping @gggsmith, we'll get this reviewed ASAP.

redash/query_runner/clickhouse.py
```diff
@@ -1,15 +1,21 @@
 import logging
 import re
 from urllib.parse import urlparse
+from uuid import uuid4
 
 import requests
 
 from redash.query_runner import *
+from redash.query_runner.databricks import split_sql_statements
```
susodapop, line 9:
```suggestion
from redash.query_runner import split_sql_statements
```

redash/query_runner/clickhouse.py
```diff
@@ -87,25 +93,41 @@ def _get_tables(self, schema):
 
         return list(schema.values())
 
-    def _send_query(self, data, stream=False):
+    def _send_query(self, data, session_id=None, session_check=None):
         url = self.configuration.get("url", "http://127.0.0.1:8123")
+        timeout = self.configuration.get("timeout", 30)
```
susodapop, line 98: Should the timeout be configurable with an environment variable?

redash/query_runner/clickhouse.py
```diff
@@ -87,25 +93,41 @@ def _get_tables(self, schema):
...
+                params=params,
                 verify=verify,
             )
+
             if r.status_code != 200:
                 raise Exception(r.text)
-            # logging.warning(r.json())
+
+            # In some foreseen requests response body can be empty - i.e. in temporary table creation requests.
+            # Execution scenario in that case is like if regular query requested and no matching rows found.
```
susodapop, line 127:
```suggestion
            # In certain situations the response body can be empty even if the query was successful, for example when creating temporary tables. 
```

redash/query_runner/clickhouse.py
```diff
@@ -155,31 +182,49 @@ def _clickhouse_query(self, query):
...
+
+        if not queries:
             json_data = None
             error = "Query is empty"
             return json_data, error
+
         try:
-            q = self._clickhouse_query(query)
-            data = json_dumps(q)
+            # If just one query was given then no need in session creation
```
susodapop, line 210:
```suggestion
            # If just one query was given no session is needed
```

redash/query_runner/clickhouse.py
```diff
@@ -155,31 +182,49 @@ def _clickhouse_query(self, query):
...
+            if len(queries) == 1:
+                results = self._clickhouse_query(queries[0])
+            else:
+                # Session is created otherwise. Parameter session_check has to be false for the first query
```
susodapop, line 214:
```suggestion
                # If more than one query was given, a session is needed. Parameter session_check must be false for the first query
```

redash/query_runner/clickhouse.py
```diff
@@ -155,31 +182,49 @@ def _clickhouse_query(self, query):
...
+                # and true for all consequent queries.
```
susodapop, line 215:
```suggestion
                # and true for all subsequent queries.
```

susodapop (changes requested):
Overall looks good. I recommend a few changes to the comments for clarity.

Also, we've moved `split_sql_statements` out of the Databricks query runner. Can you rebase off of `master` and make sure this still works?

Happy to merge after this.

redash/query_runner/clickhouse.py
gggsmith: Fixed, rebased, and verified - it still works!

redash/query_runner/clickhouse.py
gggsmith: done

redash/query_runner/clickhouse.py
gggsmith: done

redash/query_runner/clickhouse.py
gggsmith: done

redash/query_runner/clickhouse.py
gggsmith: done

redash/query_runner/clickhouse.py
gggsmith: The ClickHouse query runner had a timeout configuration option, and I had no reason to change it to the env variable. Should I introduce a new `REDASH_CLICKHOUSE_HTTP_TIMEOUT` env variable and remove the timeout option from the ClickHouse query runner configuration schema? (I am a bit worried about clients who would be required to reconfigure their Redash instance after an upgrade) I would be thankful for the explanation of the reasoning If you have time for that :)

gggsmith:
Thank you a lot!

I am a bit worried about using `split_sql_statements` (and underlying sqlparse library). Using a full-blown SQL parsing machine with only a reason to split queries feels overkill. Failure to parse SQL will affect any request (multistatement or not).

redash/query_runner/clickhouse.py
susodapop: No you are correct. I wasn't thinking straight with my question 😱

redash/query_runner/clickhouse.py
susodapop: Thank you!

susodapop (approved):

susodapop:
> Failure to parse SQL will affect any request (multistatement or not).

I think this is quite unlikely. `sqlparse` is a non-validating parser. AFAIK it won't raise an exception for badly written SQL, it just returns its input unmodified. Let's use this implementation for now and reconsider if it becomes problematic?

gggsmith:
> Let's use this implementation for now and reconsider if it becomes problematic?

Sure! That is reasonable.

PR #5118, opened 2020-08-19
Keep widget loading when fetch request is replaced

gabrieldutra:

- [x] Bug Fix

## Description

**The bug (tested in [this dashboard](https://redash-preview.netlify.app/dashboards/111-test-param-update)):**
dashboard-param-update

Basically, a parameter update when triggered again during an update wouldn't reflect on the widget state when the second refresh was finished.

**Explanation on the cause:**
This is caused by the Memoization introduced in #4734 😕. That takes the assumption that the widget updates can be based on Loading status updates, i.e.: a widget will only need to be re-rendered when its loading state updates. For this case, the loading state was set to `false` right after the first request was done, not triggering a second update on the later one. Considering it makes sense to keep the with the loading indication in that circumstance, that's what I did here.

#4734 

--

### DIFF
--- a/client/app/services/widget.js
+++ b/client/app/services/widget.js
@@ -143,18 +143,24 @@ class Widget {
       if (maxAge === undefined || force) {
         maxAge = force ? 0 : undefined;
       }
-      this.queryResult = this.getQuery().getQueryResult(maxAge);
 
-      this.queryResult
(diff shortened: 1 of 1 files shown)


kravets-levko (approved):
Weird bug and nice fix! :+1:

PR #5122, opened 2020-08-24
Introduce Link component

kravets-levko:

- [x] Refactor
- [x] Other


kravets-levko:
@arikfr @gabrieldutra I still need to test it a bit more, but there will be no major changes, so you can review it

client/app/pages/queries-list/QueriesListEmptyState.jsx
```diff
@@ -19,11 +20,11 @@ export default function QueriesListEmptyState({ page, searchTerm, selectedTags }
     case "my":
       return (
         <div className="tiled bg-white p-15">
-          <a href="queries/new" className="btn btn-primary btn-sm">
+          <Link.Button href="queries/new" className="btn btn-primary btn-sm">
```
gabrieldutra, line 23: Should we remove these classNames and use Antd styling?

viz-lib/src/visualizations/visualizationsSettings.js
```diff
@@ -32,8 +32,13 @@ HelpTrigger.defaultValues = {
   children: null,
 };
 
+function Link(props) {
+  return <a {...props} />;
+}
+
 export const visualizationsSettings = {
   HelpTriggerComponent: HelpTrigger,
+  LinkComponent: Link,
```
gabrieldutra, line 41: Missing replace this component in https://github.com/getredash/redash/blob/68b70ed63b2e38e15b6e7a875077bdf810b1d553/client/app/components/visualizations/visualizationComponents.jsx#L12-L34

client/app/pages/queries-list/QueriesListEmptyState.jsx
kravets-levko: :man_facepalming: yes, sure, thanks!

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -45,34 +46,36 @@ export default function DesktopNavbar() {
   return (
     <div className="desktop-navbar">
       <NavbarSection inlineCollapsed={collapsed} className="desktop-navbar-logo">
-        <a href="./">
-          <img src={logoUrl} alt="Redash" />
-        </a>
+        <div>
```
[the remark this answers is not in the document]
kravets-levko: Because And menu component sees that its children are not a simple html element and tries to pass some custom props to it, and we're getting this:

image

(and many of such errors below)

client/app/pages/queries-list/QueriesListEmptyState.jsx
kravets-levko: image

kravets-levko: https://github.com/getredash/redash/pull/5122/commits/072b2be2c9dc5f7db56594fd1616bffc84a86576

viz-lib/src/visualizations/visualizationsSettings.js
kravets-levko: https://github.com/getredash/redash/pull/5122/commits/072b2be2c9dc5f7db56594fd1616bffc84a86576

client/app/pages/queries-list/QueriesListEmptyState.jsx
gabrieldutra: The sizing was also small, do you think it looks better with normal?

kravets-levko: To be honest - I'm not sure which is better:

image

WDYT - should it be small or default size? I really cant chose :slightly_smiling_face:

client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
```diff
@@ -12,10 +13,10 @@ export default function VersionInfo() {
       {clientConfig.newVersionAvailable && currentUser.hasPermission("super_admin") && (
         <div className="m-t-10">
           {/* eslint-disable react/jsx-no-target-blank */}
-          <a href="https://version.redash.io/" className="update-available" target="_blank" rel="noopener">
+          <Link href="https://version.redash.io/" className="update-available" target="_blank" rel="noopener">
```
arikfr, line 16: Why use it for external links? (not saying we shouldn't - just trying to understand motiviation)

client/app/components/visualizations/visualizationComponents.jsx
```diff
@@ -11,6 +12,7 @@ function wrapComponentWithSettings(WrappedComponent) {
   return function VisualizationComponent(props) {
     updateVisualizationsSettings({
       HelpTriggerComponent: HelpTrigger,
+      LinkComponent: Link,
```
arikfr, line 15: Is this to support ability to handle correctly internal links in tables?

client/app/lib/value-format.js
```diff
@@ -1,86 +0,0 @@
-import moment from "moment/moment";
```
arikfr, line 1: Was unused?

viz-lib/src/lib/value-format.js
```diff
@@ -13,7 +16,15 @@ export function createTextFormatter(highlightLinks) {
   if (highlightLinks) {
     return value => {
       if (isString(value)) {
-        value = value.replace(urlPattern, '$1<a href="$2" target="_blank">$2</a>');
+        const Link = visualizationsSettings.LinkComponent;
+        value = value.replace(urlPattern, (unused, prefix, href) => {
+          const link = ReactDOMServer.renderToStaticMarkup(
```
arikfr, line 21: I wonder if this might have performance implications.

arikfr (commented):
👍

client/app/components/visualizations/visualizationComponents.jsx
kravets-levko: Yes, exactly

client/app/lib/value-format.js
kravets-levko: Yes. It was needed only for visualizations. I asked Gabriel why he decided to keep it in the app, and he told me that probably that was accidental. I didn't want to implement same changes second time considering that it will stay unused anyway, and decided to remove this file.

viz-lib/src/lib/value-format.js
kravets-levko: It may be a little bit slower, but if the Link component will stay simple - it shouldn't be a problem. In addition we got a safe html here - we didn't escape `href` before. DOMPurify was doing its job, but now it will be less chance that malformed url will break markup.

client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
kravets-levko: First - to keep things consistent, so if we need a link next time - we just use the component and don't think if that is an external link or not.
Second - sometimes it was a bit hard to tell if URL in this particular place is local or external. To avoid issues I decided to use the component everywhere.
Third - customization. If one day we'll decide to, say, mark external links, or add some additional parameters to all local links, or whatever - we'll just need to update/override the Link component without searching all the places that should be updated.

gabrieldutra: It makes it easier to add an event tracker too

client/app/pages/queries-list/QueriesListEmptyState.jsx
gabrieldutra: Let's leave it small as it used to be like that. That page probably could use a different empty state, but that's out of scope here :)

kravets-levko: https://github.com/getredash/redash/pull/5122/commits/0d13c530a6d64768b283005437e776122a71e23d

Also I renamed `value-format.js` to `.jsx` as now it contains some JSX fragments

kravets-levko:
@gabrieldutra @arikfr I checked all the things and seems it works fine - all links/buttons are rendered with proper attributes, markup didn't change, and looks I replaced all the links. So if you have no more comments we can merge this.

P.S. There are some `<a>` tags w/o href - we use them with `onClick` handler. I didn't change them for now as from my point they aren't _links_ but we rather use them as link-style buttons. Probably we need some different component for such "buttons" that will encapsulate all needed styling.

Issue #5107, opened 2020-08-15
Bug: I found redash dashboard page Table widget sort error


yankeeguyu:
I found that the Table Widget in the dashboard page can sort positive numbers, but for negative numbers, floating point numbers, and negative numbers are not supported.
Like this:
![image](https://user-images.githubusercontent.com/46274612/90309173-6ed6c700-df18-11ea-8ea3-81fc19726b8a.png)
![image](https://user-images.githubusercontent.com/46274612/90309180-7bf3b600-df18-11ea-99f8-79b218a30549.png)

I think this feature will still affect the user experience. Is there any way to modify it? Or is there a branch available? If so, please share with me, thank you, I can modify the source code by myself.

### Technical details:

* Redash Version: 8.0.0
* Browser/OS: MacOS & Chrome
* How did you install Redash: npm and python celery build by myself

yankeeguyu:
This is a frontend bug.

yankeeguyu:
@arikfr hi, this bug could be fix?

arikfr:
I couldn't reproduce this here: https://redash-preview.netlify.app/queries/415/source. Maybe we fixed it already in v9? (I have a vague memory that we have, but not sure)

yankeeguyu:
> I couldn't reproduce this here: https://redash-preview.netlify.app/queries/415/source. Maybe we fixed it already in v9? (I have a vague memory that we have, but not sure)


I still want to use version 8.0.0, so is there a PR to fix this bug?

arikfr:
@yankeeguyu there should be a PR, but couldn't find it with a quick search. You can check the history... 

Another thing to confirm in your version: are you sure the database doesn't return strings instead of numbers?

yankeeguyu:
> @yankeeguyu there should be a PR, but couldn't find it with a quick search. You can check the history...
> 
> Another thing to confirm in your version: are you sure the database doesn't return strings instead of numbers?

Yes, i make sure. It seems not, they treat all numbers as text or strings. I find this PR, but can't to find it.

yankeeguyu:
> @yankeeguyu there should be a PR, but couldn't find it with a quick search. You can check the history...
> 
> Another thing to confirm in your version: are you sure the database doesn't return strings instead of numbers?


There are too many closed PRs, are there any good keywords that can let me find this PR? Thank you very much!

yankeeguyu:
> @yankeeguyu there should be a PR, but couldn't find it with a quick search. You can check the history...
[...]

Issue #5111, opened 2020-08-17
Table visualization sorts by column type and not by display type


gabrieldutra:
### Issue Summary

Table visualization sorter is based on the type information provided by the server and not on the display type from visualization settings.

### Steps to Reproduce

1. Create a query with String numeric values and change the display type for the column to "Number" (or just use [this query](http://localhost:8080/queries/418))
2. Sort the query by that column and verify that the order is alphabetical and not numeric.

Example:
![Kapture 2020-08-17 at 16 29 04](https://user-images.githubusercontent.com/3356951/90436314-d15dcd80-e0a6-11ea-95a1-299f7a5d9329.gif)

I expect it to use the display type to do that, otherwise the user doesn't have an UI control of that.

### Technical details:

* Redash Version: master
* Browser/OS: --
* How did you install Redash: --

yankeeguyu:
This bug can be fixed? I found float number's sort is error.

yankeeguyu:
@gabrieldutra I fixed this bug, i pushed a PR to v8.0.x.

PR #5127, opened 2020-08-25
Fix table widget sort float error bug

gaecoli:

- [x] Bug Fix

getredash/redash#5107

### DIFF
--- a/client/app/visualizations/table/utils.js
+++ b/client/app/visualizations/table/utils.js
@@ -170,6 +170,14 @@ export function filterRows(rows, searchTerm, searchColumns) {
   return rows;
 }
 
+function chechNumber(num) {
+  const re = /^[0-9]+.?[0-9]*/;
+  if (!re.test(num)) {
+    return false;
+  }
+  return true;
+}
+
 export function sortRows(rows, orderBy) {
   if ((orderBy.length === 0) || (rows.length === 0)) {
     return rows;
@@ -184,7 +192,8 @@ export function sortRows(rows, orderBy) {
     for (let i = 0; i < orderBy.length; i += 1) {
       va = a[orderBy[i].name];
       vb = b[orderBy[i].name];
-      if (isNil(va) || va < vb) {
+
+      if (chechNumber(va) && chechNumber(vb)) {
         // if a < b - we should return -1, but take in account direction
         return -1 * directions[orderBy[i].direction];
       }
--- a/client/app/visualizations/table/utils.js
+++ b/client/app/visualizations/table/utils.js
@@ -170,8 +170,8 @@ export function filterRows(rows, searchTerm, searchColumns) {
   return rows;
 }
 
-function chechNumber(num) {
-  const re = /^[0-9]+.?[0-9]*/;
+function checkIsFloat(num) {
+  const re = /^-?\d*\.?\d+$/;
   if (!re.test(num)) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


yankeeguyu:
I fixed version is 8.0.x.

client/app/visualizations/table/utils.js
```diff
@@ -184,7 +192,8 @@ export function sortRows(rows, orderBy) {
     for (let i = 0; i < orderBy.length; i += 1) {
       va = a[orderBy[i].name];
       vb = b[orderBy[i].name];
-      if (isNil(va) || va < vb) {
+
+      if (chechNumber(va) && chechNumber(vb)) {
```
kravets-levko, line 196: 1. `chechNumber` does not cover all possible cases (e.g. it will return `false` for negative numbers. It is better to use something line `isFinite(parseFloat(value))` (also function name has a typo, and it does not reflect the purpose of the function).
2. this line (which I'm adding comment to) totally breaks reverse sorting - it disables `isNil` check, actual value comparison is missing. Also, there is similar check few lines below and it's not updated.
3. this fix in whole is very specific - it will try to handle all values as numbers. But what about other types? Strings, date/time, etc.? The proper fix should be based on idea from getredash/redash#5111 - sorting should be performed according to column type selected in table editor. E.g. if query returns strings, but user chose that column to be displayed as number - sort function should cast all values to number; same for date/time and other types.

kravets-levko (changes requested):

client/app/visualizations/table/utils.js
yankeeguyu: Sorry, i should change `re` to `const re = /^-?\d*\.?\d+$/;`, this will cover all possible number. I found only float number can't sort correctly, this bug because of treat float number as string, this PR will not affect other type(string, date).

yankeeguyu: @kravets-levko can you check this re? thank you!

kravets-levko: @yankeeguyu The regex still does not cover all cases, e.g. `+314e-2` is still a number, but it won't be recognized. Please check my first comment in this thread. I already offered a variant that will work for sure in item 1. Also, items 2 and 3 from my reply are still relevant.

yankeeguyu: > @yankeeguyu The regex still does not cover all cases, e.g. `+314e-2` is still a number, but it won't be recognized. Please check my first comment in this thread. I already offered a variant that will work for sure in item 1. Also, items 2 and 3 from my reply are still relevant.

Thank you very much!

yankeeguyu: > @yankeeguyu The regex still does not cover all cases, e.g. `+314e-2` is still a number, but it won't be recognized. Please check my first comment in this thread. I already offered a variant that will work for sure in item 1. Also, items 2 and 3 from my reply are still relevant.

Although the current PR can solve my problem, i still want to fix this bug completely. This bug can be fixed?
I tried some modifications, but the type of number is still parsed as string at this problem. Thank you!

gaecoli: > @yankeeguyu The regex still does not cover all cases, e.g. `+314e-2` is still a number, but it won't be recognized. Please check my first comment in this thread. I already offered a variant that will work for sure in item 1. Also, items 2 and 3 from my reply are still relevant.

Hi this function i change it to 
`
function checkNumber(n) {
     return Number(n) === n;
}
`

image

gaecoli: > > @yankeeguyu The regex still does not cover all cases, e.g. `+314e-2` is still a number, but it won't be recognized. Please check my first comment in this thread. I already offered a variant that will work for sure in item 1. Also, items 2 and 3 from my reply are still relevant.
> 
> Hi this function i change it to `function checkNumber(n) { return Number(n) === n; }`
> 
> image

Sorry. This has bug.

justinclift:
@gaecoli Does this bug still need fixing?

gaecoli:
> @gaecoli Does this bug still need fixing?

no

PR #5136, opened 2020-08-30
Move CardsList to typescript

max-voronov:

- [x] Refactor

## Description

Hello everyone. If it's possible, I would like to help with moving to typescript.
In this PR the `CardsList` component was moved to typescript and then converted into a functional component.

#5027 
 
There aren't any UI changes

### DIFF
--- a/client/app/components/cards-list/CardsList.tsx
+++ b/client/app/components/cards-list/CardsList.tsx
@@ -1,6 +1,6 @@
 import { includes, isEmpty } from "lodash";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { useState } from "react";
 import Search from "antd/lib/input";
 import Link from "@/components/Link";
 import EmptyState from "@/components/items-list/components/EmptyState";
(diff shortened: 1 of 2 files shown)


kravets-levko:
Hi @max-voronov! Thank you for your contribution! I'd like to ask you about one more change to this PR: could you try to add a commit that just renames `.jsx` file to `.tsx`, and then rebase everything else on top of it? It's totally not necessary, but this will allow GIT to track a rename, and (I hope) it will display a proper diff here instead of `removed` + `created` changes. Thanks!

max-voronov:
> Hi @max-voronov! Thank you for your contribution! I'd like to ask you about one more change to this PR: could you try to add a commit that just renames `.jsx` file to `.tsx`, and then rebase everything else on top of it? It's totally not necessary, but this will allow GIT to track a rename, and (I hope) it will display a proper diff here instead of `removed` + `created` changes. Thanks!

Hi @kravets-levko and thanks for replying!

I did what you asked me to, but this won't change the total diff you see under the `Files changed` tab.
It will still be showed as `removed` + `created` changes, so I think this commit is redundant and better to be removed.

Please let me know if I should remove it or not.

kravets-levko:
I hoped that it will work :slightly_smiling_face: Anyway, thank you, and sorry that I bothered you :bow:

gabrieldutra:
No need to revert @max-voronov, in the end we will squash the commit anyway 😅

max-voronov:
> I hoped that it will work 🙂 Anyway, thank you, and sorry that I bothered you 🙇

You didn't bother me 😃
I hoped it will help too but I guess there are too many changes to the same file that include indentation changes, so GIT sees it as `removed` + `created` change even tough there are some lines that their only change is indentation.

client/app/components/cards-list/CardsList.tsx
```diff
@@ -0,0 +1,73 @@
+import { includes, isEmpty } from "lodash";
+import PropTypes from "prop-types";
+import React, { useState } from "react";
+import Search from "antd/lib/input";
```
gabrieldutra, line 4: This will make it use the `Input` component instead of the `Input.Search`. Let's rename it to `Input` here and use `Input.Search` as the component in https://github.com/getredash/redash/blob/d6025f9186d8db8e45b79a6fc4b0afd0855c19fb/client/app/components/cards-list/CardsList.tsx#L42

I think the search is actually not used in the app, but let's keep it as it was in the Class component.

gabrieldutra (commented):
Overall looks good, thanks @max-voronov 😁. I left just one small comment.

The component seems better and simplified as a Functional component. But I'm curious, did you have any specific reason to move it to TS?

I guess we haven't actually decided how to proceed with TS in the codebase at this point, we added #5027 so that we have support for it (there were a few spots that we needed it), but I'm not sure I'd go around refactoring existing components (unless there's a reason for so). For me there's no problem at all with this component for example, but some of them can be quite complicated 😅.

client/app/components/cards-list/CardsList.tsx
max-voronov: Fixed in 83b13dbdb58d43f654223f1da198e387e1bc9ada

max-voronov:
> Overall looks good, thanks @max-voronov 😁. I left just one small comment.
> 
> The component seems better and simplified as a Functional component. But I'm curious, did you have any specific reason to move it to TS?
> 
> I guess we haven't actually decided how to proceed with TS in the codebase at this point, we added #5027 so that we have support for it (there were a few spots that we needed it), but I'm not sure I'd go around refactoring existing components (unless there's a reason for so). For me there's no problem at all with this component for example, but some of them can be quite complicated 😅.

Thanks for the review 😃 

I assumed that because you added Typescript support in #5027, you would like to convert to Typescript.
There is also [this discussion](https://discuss.redash.io/t/switch-to-typescript/3075) with some good points in favour of using Typescript.

If you decide that you would like to convert the whole project to Typescript then I'd like to help with the process 😃

gabrieldutra (approved):

gabrieldutra:
Yeah, I guess it's time to revisit that discussion now that we actually have TS support and some of the code in TS too 😁.

@max-voronov thank you very much for your interest and for your work here 🎉.

client/app/components/cards-list/CardsList.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+function ListItem(item: CardsListItem, keySuffix: string) {
+  return (
+    <Link key={`card${keySuffix}`} className="visual-card" onClick={item.onClick} href={item.href}>
+      <img alt={item.title} src={item.imgSrc} />
+      <h3>{item.title}</h3>
+    </Link>
+  );
+}
```
kravets-levko, line 29: Overall code looks good, just a little last thing: let's convert this to a component, because it already looks like a component, just need to change arguments to props, and update usage

kravets-levko (changes requested):
Please see my comment - that's the last little change I'd like to see here before I'll merge this. Thanks!

client/app/components/cards-list/CardsList.tsx
max-voronov: Fixed in c39f254b2bef2f5962039ef8e1e60fa697d0a53e

Thanks for the review 😃

kravets-levko (approved):

kravets-levko:
Thank you @max-voronov!

PR #3991, opened 2019-07-16
Move Cypress to dev dependencies

gabrieldutra:
## What type of PR is this? (check all applicable)

- [x] Other

## Description
This moves Cypress to a `devDependency` on `package.json` and makes required adjustments on CI.

## Related Tickets & Documents
--

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
--

PR #5137, opened 2020-08-30
Generate Code Coverage report for Cypress

gabrieldutra:

- [x] Other

## Description

This PR adds Code Coverage information to Cypress.

**Caveats**
- Code Coverage is only generated for the "master" branch as it added around 10 minutes to CI time;
- Jest tests are not included due to an issue I had and it was not that worth it to continue debugging into it as the great majority of our frontend coverage is on Cypress

--

cypress-code-coverage

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,11 @@ ARG skip_frontend_build
 WORKDIR /frontend
 COPY package.json package-lock.json /frontend/
 COPY viz-lib /frontend/viz-lib
+
+# Controls whether to instrument code for coverage information
+ARG code_coverage
+ENV BABEL_ENV=${code_coverage:+test}
+
 RUN if [ "x$skip_frontend_build" = "x" ] ; then npm ci --unsafe-perm; fi
 
 COPY client /frontend/client
--- a/.circleci/Dockerfile.cypress
+++ b/.circleci/Dockerfile.cypress
@@ -9,4 +9,6 @@ RUN npm run cypress:install > /dev/null
 COPY client/cypress $APP/client/cypress
 COPY cypress.json $APP/cypress.json
 
+COPY . $APP
+
 RUN ./node_modules/.bin/cypress verify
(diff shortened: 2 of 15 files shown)


gabrieldutra:
While doing this, I faced a couple of issues:

**1 - I could not add Instanbul.js in Jest**, this is what I got when I tried:
Screen Shot 2020-09-12 at 15 05 57
I spent some time debugging it, but eventually realized it was not worth it as the great majority of our Frontend tests is in Cypress;

**2 - Code Coverage increased CI time in 10 minutes** ([details](https://dashboard.cypress.io/projects/924cka/runs/3787/specs?utm_source=github)). Although it's a bit weird and I would like to investigate further outside of CI (check if it's really expected), for now I enabled the coverage report only for the master branch.

--- 

This PR already covers the first and most important use-case I see (easily identify Frontend code that doesn't have any coverage), so I'm opening it for review.

**Possible Future Follow ups**
- Include other tests in the Code Coverage information;
- Add a 3rd party tool to track coverage;
- [If in GH Actions] Add the report as a Status in the checks
- Reconsider adding the report on every commit. With that, it's possible to highlight a delta value whenever it decreases;
- Add Badge with coverage info in README (easier with the 3rd party tool)

kravets-levko (approved):
Stunning work @gabrieldutra! Looking forward to see it helping to increase code coverage 👍

Dockerfile
```diff
@@ -6,6 +6,11 @@ ARG skip_frontend_build
 WORKDIR /frontend
 COPY package.json package-lock.json /frontend/
 COPY viz-lib /frontend/viz-lib
+
+# Controls whether to instrument code for coverage information
+ARG code_coverage
+ENV BABEL_ENV=${code_coverage:+test}
```
gabrieldutra, line 12: @arikfr, I'll merge this PR as it feels safe + I can include it while updating #3991.

I'm highlighting the only change I had to do for the Dockerfile, in case you have any comments I can address in a follow up tomorrow.

PR #5150, opened 2020-09-05
Recent queries and dashboards feature

pxd3v:

- [x] Feature

## Description
* creates a new queries/recent route at frontend
* creates a new recent queries button at queries menu
* creates a hook to add queries to localstorage when executing or opening them
* creates a function to retrieve recent queries
* creates a new dashboards/recent route at frontend
* creates a new recent dashboards button at dashboards menu
* creates a hook to add dashoards to localstorage when executing or opening them
* creates a function to retrieve recent dashboards

I used lodash functions instead of native functions because lodash was already being used everywhere.
image
image
image
image


client/app/services/dashboard.js
```diff
@@ -154,6 +154,24 @@ function transformResponse(data) {
   return data;
 }
 
+function transformRecentResponse(data) {
```
gabrieldutra, line 157: Let's not mix up the transform and "recent" functions and have two separate `.then`:

```js
{
  recentDashboards: params =>
    axios
      .get("api/dashboards", { params })
      .then(transformResponse)
      .then(doSomeMappingHere),
}
```
(need to check in what order it makes sense)

client/app/services/dashboard.js
```diff
@@ -171,6 +189,7 @@ const DashboardService = {
   favorites: params => axios.get("api/dashboards/favorites", { params }).then(transformResponse),
   favorite: ({ id }) => axios.post(`api/dashboards/${id}/favorite`),
   unfavorite: ({ id }) => axios.delete(`api/dashboards/${id}/favorite`),
+  recentDashboards: params => axios.get("api/dashboards", { params }).then(transformRecentResponse)
```
gabrieldutra, line 192: How about we have the same method name as in the Query, so we can have shared code? (it's something similar to what we do with the `favorite` ones)

client/app/services/dashboard.js
```diff
@@ -154,6 +154,24 @@ function transformResponse(data) {
...
+  const allUserDashboards = _.map(data.results, transformSingle);
+  const userRecentDashboards = _.filter(allUserDashboards, dashboard => recentDashboardsIds.includes(dashboard.id.toString()));
+  const sortedUserRecentDashboards = userRecentDashboards.sort((dashboardA, dashboardB) => {
+    const priorityOfDashboardA = _.find(parsedRecentDashboards, dashboard => dashboard.id === dashboardA.id.toString()).priority;
+    const priorityOfDashboardB = _.find(parsedRecentDashboards, dashboard => dashboard.id === dashboardB.id.toString()).priority;
+    if (priorityOfDashboardA < priorityOfDashboardB) return 1;
+    if (priorityOfDashboardA > priorityOfDashboardB) return -1;
+    return 0;
+  });
+  return { ...data, results: sortedUserRecentDashboards };
```
gabrieldutra, line 172: This "priority" thing feels a bit complex 🤔, can you elaborate a bit on what were your thoughts to order the list?

My initial thought was to have saved in the order we want and on every "track action" (view/execute or whatever we want to use to track) it: 1. Remove the existing if it is part of the list already; 2. Add it in the beginning of the list;

gabrieldutra (commented):
Hi @pxd3v, thanks for the PR, this is a very interesting feature to have 🎉. I left a few comments in the code, but I'd like to discuss a bit on the possibilities here before continuing.

I realize the initial idea was to implement this frontend only, I think there are some pros and cons for that, but there is one small "complication" for that: we have pagination for both Dashboard and Queries api, you can see how it's behaving in [the preview environment](https://deploy-preview-5150--redash-preview.netlify.app/queries/recent):

Screen Shot 2020-09-07 at 14 26 59

---
So, let's dive into the possible ways I see of getting this done:

**1. Frontend Feature**
With this approach, we can't guarantee that the "Recent" information will be related to the User, but rather to the device. It does have the advantage of easier implementation.

We still have to solve the pagination thing, for that I see two options: 1. Store all the relevant object information; 2. Modify the List api to allow filtering by ids (I personally prefer this option);

BTW, if we are gonna go with frontend only, let's consider having the code shared, i.e.: make the same hook for both Query and Dashboard pages.

**2. Backend Feature**
The `api/queries/recent` actually isn't the same thing we are looking for 😅, so we would have to think of another way to introduce it for both Dashboard and Query. We can use the GET requests to track actions on queries and dashboards.

What do you think? (I think we can give you some help if the backend part is worth it)

(cc: @arikfr for thoughts on the feature too)

client/app/services/dashboard.js
pxd3v: For sure, i also think this is a better aproach, i'll apply this changes asap.

client/app/services/dashboard.js
pxd3v: I got your idea, this would be great. But i cant figure out an meaningful name for both, i think something like "recents" would not say too much about the method, given that we also have this other "recent" feature.

client/app/services/dashboard.js
pxd3v: So basically biggest priority implies in appearing first, every time i need to update the "recents" array, i check for the biggest value in the array, and give my new item the priority value of the biggest value + 1, that way i only need to sort the array when i'm displaying it

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -164,6 +165,10 @@ function DashboardPage({ dashboardSlug, dashboardId, onError }) {
       })
       .catch(handleError);
   }, [dashboardId, dashboardSlug, handleError]);
+  
+  useEffect(() => {
+    addDashboardToRecentDashboards();
+  }, [addDashboardToRecentDashboards]);
```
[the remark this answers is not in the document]
pxd3v: For sure that will work better, but this conflicts with what you have said about "make the same hook for both Query and Dashboard pages.", because in queries we need to have this callback so we can call it when the user execute the query, maybe its better to have only one hook but with the callback. What do you think?

client/app/services/auth.js
```diff
@@ -55,6 +55,8 @@ export const Auth = {
     window.location.href = `${AuthUrls.Login}?next=${next}`;
   },
   logout() {
+    localStorage.removeItem("recentQueries");
+    localStorage.removeItem("recentDashboards");
```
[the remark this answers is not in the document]
pxd3v: I totally forgot about this case of expiration, for sure this need to discussed. I like the backend idea because by this way we can save the recent user data for future sessions, that would be nice.

pxd3v:
> Hi @pxd3v, thanks for the PR, this is a very interesting feature to have . I left a few comments in the code, but I'd like to discuss a bit on the possibilities here before continuing.
> 
> I realize the initial idea was to implement this frontend only, I think there are some pros and cons for that, but there is one small "complication" for that: we have pagination for both Dashboard and Queries api, you can see how it's behaving in [the preview environment](https://deploy-preview-5150--redash-preview.netlify.app/queries/recent):
> 
> Screen Shot 2020-09-07 at 14 26 59
> 
> So, let's dive into the possible ways I see of getting this done:
> 
> **1. Frontend Feature**
> With this approach, we can't guarantee that the "Recent" information will be related to the User, but rather to the device. It does have the advantage of easier implementation.
> 
> We still have to solve the pagination thing, for that I see two options: 1. Store all the relevant object information; 2. Modify the List api to allow filtering by ids (I personally prefer this option);
> 
> BTW, if we are gonna go with frontend only, let's consider having the code shared, i.e.: make the same hook for both Query and Dashboard pages.
> 
> **2. Backend Feature**
> The `api/queries/recent` actually isn't the same thing we are looking for , so we would have to think of another way to introduce it for both Dashboard and Query. We can use the GET requests to track actions on queries and dashboards.
> 
> What do you think? (I think we can give you some help if the backend part is worth it)
> 
> (cc: @arikfr for thoughts on the feature too)

Hi @gabrieldutra, thanks for the feedback. I'm so glad to work together with you guys and i'm also really excited with this feature, so lets solve it. I think the main problem here is the pagination thing, and i came with an temporary solution that can be easily applied while we work in the backend idea (the best idea IMO). So, basically my idea is to limit the recent queries and dashboards to a low number like 5, so when we need to register an sixty query/dashboard we just remove the more outdated one from the list and add the recent one. Let me know what you guys think about it!

pxd3v:
Just to complement my last comment, this temporary solution is only applicable if an endpoint that retrieves all queries/dashboards current exist.

client/app/pages/dashboards/DashboardPage.jsx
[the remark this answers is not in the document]
gabrieldutra: We can pass the "action" as a parameter to the Hook, or pass the service object instead of just the Id (Query or Dashboard) and "assume" the method will be the same from there.

client/app/pages/dashboards/hooks/useAddDashboardToRecentDashboards.js
```diff
@@ -0,0 +1,31 @@
+import { reduce, map, find } from "lodash";
+import { useCallback } from "react";
+
+export default function useAddDashboardToRecentDashboards(dashboardId) {
```
arikfr, line 4: I agree with @gabrieldutra that we should consolidate the code and have a single hook to add into recents that can work both for queries and dashboards (and actually other objects like alerts as well). Also the code when reading recents probably can be consolidated as well.

There is alot of duplicate between both transforms/hooks.

client/app/pages/dashboards/DashboardPage.jsx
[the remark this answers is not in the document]
arikfr: Why add a query to recents only when executing a query? I think it's reasonable to add it when viewing. If we see it "spams" the list too much, we can revise the logic.

client/app/pages/dashboards/hooks/useAddDashboardToRecentDashboards.js
```diff
@@ -0,0 +1,31 @@
...
+  return useCallback(() => {
+    const currentRecentDashboards = localStorage.getItem("recentDashboards");
+    if (!currentRecentDashboards) {
+      localStorage.setItem("recentDashboards", JSON.stringify([{ id: dashboardId, priority: 1 }]));
```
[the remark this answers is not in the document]
arikfr: And another thought: we only add items, but what about eviction? If we don't remove items after some usage we will end up with hundreds if not thousands of items in these lists for some usrs.

arikfr (commented):
Thank you for taking a stab at this feature! This is going to be a great usability improvement for users.

I left a few comments in the code. They mostly align with what @gabrieldutra already said, but I have some concerns about the data structure (see comment).

> 2. Modify the List api to allow filtering by ids (I personally prefer this option);

+1 to this.

In terms of UX: how about we have a single global list of recent items (for all types together) on the homepage? This also impacts how you store the items (and another reason to have a single hook).

client/app/pages/dashboards/DashboardPage.jsx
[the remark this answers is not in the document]
pxd3v: We also add the query to recents when viewing it! in the query view file:
image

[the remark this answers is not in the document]
pxd3v: We also add the query to recents when viewing it! At the QueryView file: 
image

client/app/services/auth.js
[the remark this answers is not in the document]
pxd3v: What about reset the recents in localstorage when user login? So the ids that remain in the localstorage don't bother the next user experience if the last user didn't logout by using the logout button.

client/app/pages/dashboards/hooks/useAddDashboardToRecentDashboards.js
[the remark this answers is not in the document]
pxd3v: As i said to gabriel:
>So basically biggest priority implies in appearing first, every time i need to update the "recents" array, i check for the biggest value in the array, and give my new item the priority value of the biggest value + 1, that way i only need to sort the array when i'm displaying it
But we can do it only whit an array of ids, also i think that using only ids on the array it will be easier to read to be honest.. i will work on it.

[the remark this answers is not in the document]
pxd3v: As i said to gabriel:

>So basically biggest priority implies in appearing first, every time i need to update the "recents" array, i check for the >biggest value in the array, and give my new item the priority value of the biggest value + 1, that way i only need to sort >the array when i'm displaying it
But we can do it only whit an array of ids, also i think that using only ids on the array it will be easier to read to be honest.. i will work on it.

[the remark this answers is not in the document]
pxd3v: As i said to Gabriel:

>So basically biggest priority implies in appearing first, every time i need to update the "recents" array, i check for the biggest value in the array, and give my new item the priority value of the biggest value + 1, that way i only need to sort the array when i'm displaying it

But we can do it only whit an array of ids, also using only ids on the array it will be easier to read to be honest.. i will work on it.

client/app/pages/dashboards/hooks/useAddDashboardToRecentDashboards.js
pxd3v: I also agree with you guys, i will work on it.

pxd3v:
Hi @arikfr, thank you so much for your comments, i will start working on some changes as soon as possible. About the UX, i was told to make it in that way using the menu lists, but i also think that having one list to all recent dashboards, queries and other things would be nice, i'm just not sure if i can take this UX decision, this feature was requested for me by Lei Ni in the code assignment and he asked for this menu lists way.

client/app/pages/dashboards/hooks/useAddDashboardToRecentDashboards.js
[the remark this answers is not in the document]
kravets-levko: My suggestion is to store just array of ids, where first id is the most recent item, so order of IDs in the list will define order of items. And updating this list will look like this (pseudocode):
```js
function addToRecents(list, id) {
  exclude(list, id);
  prepend(list, id);
  return slice(list, someLimitValue); // keep only reasonable count of items in list
}
```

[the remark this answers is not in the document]
gabrieldutra: What Levko suggested is exactly what I was thinking in https://github.com/getredash/redash/pull/5150/files#r484527405. The pseudocode makes it clearer 🙂

arikfr:
@pxd3v I thought that maybe it was described this way in the assignment. I will discuss this with Lei, but most likely that:

1. We will use a single list on the homepage to display it.
2. Will show filtered list (of single object type) in several use cases like replacing the usage of `Query.recent` in `QuerySelector` and later on use the dashboards list in the `AddToDashboardDialog`.

I will discuss this with Lei and follow up here.

client/app/services/auth.js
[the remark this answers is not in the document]
arikfr: That's a good point, @pxd3v. We wouldn't want to clear it out on login, because this might happen often (depends on various settings) and then recents will be less useful as they will be cleared often.

To address the issue you present (items bothering the next user), we can scope the localStorage item to the user_id and org_id. Something like: `redash.recentQueries.{orgId}.{userId}`.

client/app/services/dashboard.js
pxd3v: @gabrieldutra Done!

client/app/services/dashboard.js
pxd3v: @gabrieldutra Done!

client/app/pages/dashboards/hooks/useAddDashboardToRecentDashboards.js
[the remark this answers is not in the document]
pxd3v: @gabrieldutra @kravets-levko  Done!

client/app/pages/dashboards/hooks/useAddDashboardToRecentDashboards.js
pxd3v: @arikfr All done!

pxd3v:
Ok @arikfr, i'll wait for your follow up to finish this feature!

gabrieldutra (commented):
Considering Arik's suggestion, I think we can have different object types stored in the same localStorage entry, having the following as an example:

```json
{
 "object_id": "ID",
 "object_type" "query"
}
```

**Note:** we have to limit by `object_type`, so that if someone access several queries, but not dashboards, it won't override those. The function signature can be something like `addToRecents(objectId, objectType)`.

With that, the initial scope of the feature won't be affected and we keep the possibility to follow up a general "Recent" list later.

client/app/lib/hooks/useAddToRecentsAtLocalStorage.js
```diff
@@ -0,0 +1,24 @@
...
+  function updateList (list, id) {
+    const itemArrayIndex = list.indexOf(id);
+    const itemIsAlreadyOnTheList = itemArrayIndex !== -1;
+    if(itemIsAlreadyOnTheList)
+      list.splice(itemArrayIndex, 1);
+    list.unshift(id);
+    return list;
+  }
+
+  return useCallback(() => {
```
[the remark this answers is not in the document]
pxd3v: I guess that is important for this feature to also cover the "execute" event, so maybe the useEffect ins't the best choice. But i agree that this can be a regular JS function, i will apply this change, but i'm kinda confused about the project folder structure, where can i put this function file?

client/app/lib/hooks/useAddToRecentsAtLocalStorage.js
```diff
@@ -0,0 +1,24 @@
...
+    const formatedId = Number(id)
```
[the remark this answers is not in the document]
pxd3v: This was a mistake of mine, i just fixed it!

pxd3v:
@gabrieldutra Thanks for your review, the idea of unifying the array was already on my mind and i think is is a wise choice. I just commited this changes, i decided to limit the array by 5 objects of each type, what do you think?

client/app/lib/addToRecentsAtLocalStorage.js
```diff
@@ -0,0 +1,37 @@
+function removeMostOutdatedItemIfTypeLimit(list, objectType) {
```
kravets-levko, line 1: IMHO this file looks rather like a service, so maybe move it to `services` folder (+ maybe wrap all these function into some class?)

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -19,6 +19,7 @@ import url from "@/services/url";
 import useImmutableCallback from "@/lib/hooks/useImmutableCallback";
 
 import useDashboard from "./hooks/useDashboard";
+import addToRecentsAtLocalStorage from "../../lib/addToRecentsAtLocalStorage";
```
kravets-levko, line 22:
```suggestion
import addToRecentsAtLocalStorage from "@/lib/addToRecentsAtLocalStorage";
```

Also, please run Prettier over changed files

client/app/services/auth.js
```diff
@@ -55,6 +55,7 @@ export const Auth = {
     window.location.href = `${AuthUrls.Login}?next=${next}`;
   },
   logout() {
+    localStorage.removeItem("recents");
```
kravets-levko, line 58: This should also be in the service. It's better to keep implementation details in the single place and do not expose them

client/app/services/dashboard.js
```diff
@@ -171,6 +172,7 @@ const DashboardService = {
   favorites: params => axios.get("api/dashboards/favorites", { params }).then(transformResponse),
   favorite: ({ id }) => axios.post(`api/dashboards/${id}/favorite`),
   unfavorite: ({ id }) => axios.delete(`api/dashboards/${id}/favorite`),
+  recentDashboards: params => axios.get("api/dashboards", { params }).then(transformResponse).then((data) => filterRecent(data, "dashboard"))
```
kravets-levko, line 175: As far as I understand - it currently doesn't work with pagination, right?

pxd3v: It currently doesn't work with pagination, but i don't think that we will need an pagination for it too, the objects are limited for 5 of each type like query or dashboard so the list will have like 20 items max. The big problem is that we need an endpoint to retrieve the recent items, the way this is currently working we receive the response of get api/dashboards and filter this list searching for items that are in the "recents" array, but with the pagination, we can have items that are in the recents array but aren't being retrieved in the first page of the request response.. I was thinking that implementing an endpoint that receive an array of ids and an object type to retrieve filtered items could be an good alternative or maybe an endpoint that retrieve all user querys, dashboards and alerts so we can filter this on frontend if we're following with the unified list idea.

client/app/lib/addToRecentsAtLocalStorage.js
pxd3v: As i told to @gabrieldutra i was confused about the project folder structure but for sure it can be an service class so we can have the implementation details in a single place like you said in the other comment. I will implement it!

client/app/pages/dashboards/DashboardPage.jsx
pxd3v: Ok.

client/app/lib/addToRecentsAtLocalStorage.js
kravets-levko: It's totally not a problem, sorry if I was impolite. I just wanted to suggest in idea how it could be improved :slightly_smiling_face:

client/app/services/dashboard.js
kravets-levko: Filtering it on frontend is definitely not a good idea as there may be really a lot of objects, and loading all of them just to pick five of them is totally a overkill. In this case it's even better to load objects individually by IDs (making requests in parallel of course) - it will be less load for a backend than getting all. But having a dedicated API endpoint feels like the best idea in this case. You can try to implement it - it's not that hard, but also feel free to ask for a help if needed.

pxd3v: For sure this would be an overkill hahah, i will try to implement this endpoint!

client/app/lib/addToRecentsAtLocalStorage.js
pxd3v: Its ok you didn't sound impolite, thank you so much for the guidance about the project folder structure! Sorry if i was impolite hahaha i really liked your idea :))

client/app/services/dashboard.js
pxd3v: Hi @kravets-levko, I just finished the implementations that you suggested about the "addToRecentsAtLocalStorage" being a class! This week i will try to work on the endpoint.

jhult:
Any updates on getting this merged?

pxd3v:
Hi guys! 

I recently started working on a new company here in Brazil, what made very busy in those last months, but now I think I can organize myself to figure it out.

So, this feature is still relevant for the project?!

King regards,
Pedro.

PR #5154, opened 2020-09-06
Add horizontal bar chart

rafawendel:
## What type of PR is this?
- [X] Feature

### Addition of the horizontal bar chart as a display option for the bar chart

Plotly currently has the support for horizontal bar charts via the addition of the orientation property to the chart object.

In order to do so, I added a UI checkbox that will pass this option down and added a check for it that will add the Plotly property to the chart.

Problem 1: Plotly doesn't invert the axes, which is the behavior expected by the users
Peek 2020-09-06 14-19

Solution 1: Remapped values so as to exchange x and y
Peek 2020-09-06 14-23

Problem 2: Remapping the values gives UI responsiveness, but it leads to UI misbehavior, such as unexpected axes (x will not accept multiple, y won't allow sorting), nonworking normalized to percent and stacked charts
Peek 2020-09-06 14-30
Peek 2020-09-06 14-27

Solution 2: In order to have X behaving as Y and vice-versa, it sounded best to me to just invert the axes after all configuration was done, which lead to the expected behavior.
Peek 2020-09-06 14-34

### Remaining problems:

- [X] Working axes inversion
- [X] Working stacked charts
- [x] Add tests
- [x] Working normalized percentages (disabled for horizontal charts at the moment)
- [x] Add UI responsiveness to the newer version, as the X/Y setting options will set the opposite axes
- [x] Pass current X scale to Y and vice versa, as Y won't auto-detect category, which will lead to an empty chart


rafawendel:
I was insecure about how deep in the code I should get my hands on, so I tried to keep my changes at a minimum. Having the rest of the features working would involve changing ColumnMappingSelect and X/Y axes settings. I think it's better to hear your thoughts on it first.

rafawendel:
Yesterday I realized when you mixed the horizontal chart with other charts in the "series" menu it would result in a mix of axes, as I have added the inversion only in the bar chart plotly preparing function.

In order to fix that, I changed the inversion to the bottom of the `prepareSeries` function, which results in a better behavior:
Peek 2020-09-07 11-15

In the above example you can notice stacking is working as usual. Given the fact the inversion *should* work for all other 2d charts, I simply added line and area to the list of options. However, doing so results in misbehavior for stacked charts.

Also it requires further setup to work properly. Due to that I decided to remove them and keep it only for bar charts.
Peek 2020-09-07 11-13
Peek 2020-09-07 11-14

I was thinking whether the best option is to keep this as a checkbox or I should add it to the dropdown chart options.

Either way, I realized that in order to proceed the following is needed

- Minor refactor of `ColumnMappingSelect` and `MappingTypes`
- Minor refactor of `YAxisSettings`
- New abstraction for setting axes labels and exchanging scales via `options`

The two of them are tailored for only working with the regular charts and would need changes to adapt to a more flexible UI this one requires.

gabrieldutra (commented):
Hi @rafawendel, thanks for all the detailed explanation on the context here 🙂. I've checked the feature + code and I would like to suggest a bit of a different approach.

You've probably noticed already that we have lots of customizations over Plotly and they take into account the "x" and "y" values. Some of those customizations are indeed on the "prepare" phase (those that are just some Redash -> Plotly option mapping), but others are "updates". Considering in this case we actually invert both axes, we will need to include this as an "update" in the exactly right spot, i.e.: inverting the axes and their settings as the Last step (or almost the Last 😝) should make it work as we want and with minimal risk.

(@kravets-levko you're invited to correct my understanding or to suggest other approaches 🙂) 

I suggest you to check [this file](https://github.com/getredash/redash/blob/a1255b41442da27b8bff70f70018265b7a2a1540/viz-lib/src/visualizations/chart/plotly/updateYRanges.js) and its uses, so that you can base to create a new updater. There is a comment explaining the updaters in https://github.com/getredash/redash/blob/a1255b41442da27b8bff70f70018265b7a2a1540/viz-lib/src/visualizations/chart/Renderer/initChart.js#L16-L22

I've quickly tested this by adding an updater based on the `updateYRanges` with an extra `plotlyData` arg:
```js
if (options.invertedAxes) {
  each(plotlyData, series => {
    series.orientation = "h";
    const { x, y } = series;
    series.x = y;
    series.y = x;
  });

  updates.xaxis = layout.yaxis;
  updates.yaxis = layout.xaxis;

  console.log(layout);
  console.log(plotlyData);
}
```


Result:
example-invert-axes

Note that the other features will keep working and the settings will also change accordingly. Extra: we can later check if it is worth it to rename the UI items too 😁 ("X axis" -> "Y axis" and so on).

Thanks again and LMK if I can be of any more help.

rafawendel:
I've finished a working version of the horizontal bar chart. The final behavior is the same of Gabriel's GIFs.

In order to have the labels inverted in the UI I had to do some sneaky stuff which should be made into abstractions for better long-term maintenance.

Also I couldn't fully grasp how the update functions work. In the example Gabriel gave adding layout properties to the update object resulted in changes to the layout, which was not the case for my function.

Finally, there is still one problem: **charts with double y axes** will yield weird behaviors. To fix that I would have to go I little deeper into how `yaxis2` is passed down to Plotly and have a xaxis2 instead. That would be interesting, as it's listed as an example at the [documentation](https://plotly.com/javascript/horizontal-bar-charts/#bar-chart-with-line-plot).

rafawendel (commented):
I've noticed I had forgotten to reset the state after the user changed charts, which would result in inversion of axes in other charts. Also, horizontal inversion can be safely applied to line charts. It should work for area charts, but the baseline for the area gets messed.

viz-lib/src/visualizations/chart/Editor/ColumnMappingSelect.jsx
```diff
@@ -12,9 +12,18 @@ const MappingTypes = {
...
+
+  let { label, multiple } = MappingTypes[type]
+  // this inverts the ui, as the data will be inverted on render
+    if (isAxesInverted) {
+      if (type === "x") {
+        label = label.replace("X", "Y");
+      } else if (type === "y") {
+        label = label.replace("Y", "X");
+      }
+    }
```
arikfr, line 26: Something in the formatting looks off, worth running Prettier here.

viz-lib/src/visualizations/chart/Editor/GeneralSettings.jsx
```diff
@@ -271,4 +285,4 @@ export default function GeneralSettings({ options, data, onOptionsChange }) {
   );
 }
 
-GeneralSettings.propTypes = EditorPropTypes;
```
arikfr, line 274: This is no longer needed?

viz-lib/src/visualizations/chart/Editor/GeneralSettings.jsx
```diff
@@ -117,11 +118,24 @@ export default function GeneralSettings({ options, data, onOptionsChange }) {
...
       </Section>
 
+      {includes(["column", "line"], options.globalSeriesType) && (
+        <Section>
+          <Checkbox
+            data-test="Chart.InvertedAxes"
+            defaultChecked={options.invertedAxes}
+            checked={options.invertedAxes}
+            onChange={() => onOptionsChange({ invertedAxes: !options.invertedAxes })}>
+            Horizontal chart
```
arikfr, line 128:
```suggestion
            Horizontal Chart
```

arikfr (commented):
This is great work, @rafawendel! Beyond the code, it was really helpful following through the thought process in the PR description and the comments.

In terms of code, looking good - there are two small nits I noticed while going through it (see comments below).

In terms of functionality: currently the chart stops rendering when moving one of the series to the second Y axis. You mentioned already that there are issues with this, but my suggestion is: let's disable this functionality when Horizontal Chart is selected. If there will be demand/valid use cases, we can look into adding support for it. Wdyt?

viz-lib/src/visualizations/chart/Editor/ColumnMappingSelect.jsx
gabrieldutra: `npm run prettier` currently doesn't include the `viz-lib` 😅, I'll fix that.

---

@rafawendel I think for it to work in vscode you'll need to manually specify the settings path to `client/prettier.config.js` for the workspace.

rafawendel:
> This is great work, @rafawendel! Beyond the code, it was really helpful following through the thought process in the PR description and the comments.
> 
> In terms of code, looking good - there are two small nits I noticed while going through it (see comments below).
> 
> In terms of functionality: currently the chart stops rendering when moving one of the series to the second Y axis. You mentioned already that there are issues with this, but my suggestion is: let's disable this functionality when Horizontal Chart is selected. If there will be demand/valid use cases, we can look into adding support for it. Wdyt?

Thank you @arikfr, it's flattering to hear it coming from you!

I find it a great idea! I was wondering whether it would be as laborious to add the second x axis as just disabling the second y axis. Now the second one sounds best.

Look forward to doing it ASAP

rafawendel:
**TLDR** Code is ready. Future refactoring is needed.

@gabrieldutra and @arikfr I am done with dealing with the right Y axis. When the horizontal axis is disabled the right Y axis will disappear from the series' tab, Y Axis (which will be displayed as X Axis) and it's state will be reset to left Y (comments below).

Therefore I think I've finished all details needed for merger. Surely I will work on any further issues you might point out or future improvements/corrections.

### Conclusion
Working in this involved getting my hands dirty, as I had to modify several functions around the code. The greatest problem is that most of these changes are hardcoded, which can lead to poor maintenance.

In order to improve this, we would need to further discuss on the priority of creating abstractions that would increase the robustness of this feature.

viz-lib/src/visualizations/chart/Editor/GeneralSettings.jsx
```diff
@@ -107,6 +107,15 @@ export default function GeneralSettings({ options, data, onOptionsChange }) {
     }
   }
 
+  function handleAxesInversion() {
+    // moves any item in the right Y axis to the left one
+    const seriesOptions = mapValues(options.seriesOptions, series => ({
+      ...series,
+      yAxis: 0,
```
rafawendel, line 114: By doing this, every time horizontal chart changes state, all series go back to left Y axis.

viz-lib/src/visualizations/chart/plotly/updateAxesInversion.js
```diff
@@ -23,7 +22,7 @@ export default function updateAxesInversion(seriesList, layout, options) {
         }
         if (isObject(yaxis2)) {
           // TODO add xaxis2
-          updates.yaxis2 = undefined;
+          updates.yaxis2 = null;
```
rafawendel, line 25: Likely due to nullish coalescing, undefined won't work here.
This erases the left Y axis.

viz-lib/src/visualizations/chart/plotly/updateAxesInversion.js
```diff
@@ -0,0 +1,32 @@
+import { each, isObject } from "lodash";
+export default function updateAxesInversion(seriesList, layout, options) {
```
gabrieldutra, line 2: So, @rafawendel I know I told you to create a new `updatesAxesInvertion`, but there is one thing I noticed in here 😬. What happens is that this is conflicting with `updatesYRanges` and the operation we do is a Lodash `merge`. Since we are "renaming" the variable, we end up merging in the wrong attributes. One failure example happens when you manually set ranges:

Screen Shot 2020-09-10 at 20 17 14


The simplest way I see to fix it is to move this logic to `updateYRanges` (will ofc have to rename it to something like `updateAxes` or whatever you think makes sense):

```js
export default function updateAxes(plotlyElement, layout, options, seriesList) {
  const updates = {};
  // [...]

  return [
    updates,
    () => {
      // Update Y Ranges
      // [...]

      // Inverted Axes
      if (options.invertedAxes) {
        each(seriesList, series => {
          series.orientation = "h";
          const { x, y } = series;
          series.x = y;
          series.y = x;
        });

        const { xaxis } = layout;
        const { yaxis, yaxis2 } = updates;

        if (isObject(xaxis) && isObject(yaxis)) {
          updates.xaxis = yaxis;
          updates.yaxis = xaxis;
        }
        if (isObject(yaxis2)) {
          updates.yaxis2 = null;
        }
      }

      return [updates, null]; // no further updates
    },
  ];
}
```

We can also use a third level of `Plotly.relayout`, but I guess having the code in the same file is better than that 😅

gabrieldutra (commented):
Looking good @rafawendel 😁, I left a comment for one last thing I noticed while testing this.

Also, can you run Jest with the `-u` flag when done, so that we fix the failing Unit test in CI?

Thanks!!

rafawendel:
@gabrieldutra thanks for the pinch of cleverness! Indeed your solution looks best and the most productive right now, so I just applied it. Also fixed the dummy change to `getDefaultOptions` that caused unit test failure.

viz-lib/src/visualizations/chart/Editor/ColumnMappingSelect.jsx
```diff
@@ -12,9 +12,18 @@ const MappingTypes = {
...
-export default function ColumnMappingSelect({ value, availableColumns, type, onChange }) {
+export default function ColumnMappingSelect({ value, availableColumns, type, onChange, isAxesInverted }) {
   const options = sortBy(filter(uniq(flatten([availableColumns, value])), v => isString(v) && v !== ""));
-  const { label, multiple } = MappingTypes[type];
+
+  let { label, multiple } = MappingTypes[type];
+  // this inverts the ui, as the data will be inverted on render
+  if (isAxesInverted) {
+    if (type === "x") {
+      label = label.replace("X", "Y");
```
gabrieldutra, line 22: I had some thoughts about creating a sort of "translate" method `t("x", options)` for labels. However, this is currently only for the Axis name, so it didn't seem worth it the extra logic. Leaving here as a future note 🙂.

viz-lib/src/components/visualizations/editor/createTabbedEditor.jsx
```diff
@@ -33,7 +35,7 @@ TabbedEditor.propTypes = {
   tabs: PropTypes.arrayOf(
     PropTypes.shape({
       key: PropTypes.string.isRequired,
-      title: PropTypes.string.isRequired,
+      title: PropTypes.oneOfType([PropTypes.string, PropTypes.func]).isRequired,
```
gabrieldutra, line 38:
```suggestion
      title: PropTypes.oneOfType([
        PropTypes.string,
        PropTypes.func, // (options) => string
      ]).isRequired,
```

Just to keep some "context" attached to the prop.

gabrieldutra (approved):
LGTM! 🚀 

@rafawendel again, this a very nice feature to have, thank you very much for your work here!

viz-lib/src/visualizations/chart/Editor/ColumnMappingSelect.jsx
rafawendel: This is the sort of abstraction I was thinking of. Nice to hear your thoughts on it 😁

kravets-levko: My suggestions is to introduce another mappings e.g. `const InvertedAxesMappingTypes = extend({}, MappingTypes, ...}` and then choose appropriate mapping depending on `isAxesInverted` flag.

P.S. Since you're using plural form (Axes) in flag name - maybe, `areAxesInverted` then? Also, another thing to consider - axes are rather "swapped" than "inverted". Better namings help to read code easier 🙂

viz-lib/src/visualizations/chart/Editor/ColumnMappingSelect.jsx
```diff
@@ -12,18 +12,17 @@ const MappingTypes = {
...
-  let { label, multiple } = MappingTypes[type];
   // this swaps the ui, as the data will be swapped on render
-  if (areAxesSwapped) {
-    if (type === "x") {
-      label = label.replace("X", "Y");
-    } else if (type === "y") {
-      label = label.replace("Y", "X");
-    }
-  }
+  const { label, multiple } = !areAxesSwapped ? MappingTypes[type] : SwappedMappingTypes[type]
```
rafawendel, line 25: Cleaner syntax, adopting @kravets-levko's suggestions. I also renamed everything. Guess I was with a narrow vocab at the time, swap is way better! Thanks!

viz-lib/src/visualizations/chart/Editor/GeneralSettings.jsx
```diff
@@ -106,6 +107,15 @@ export default function GeneralSettings({ options, data, onOptionsChange }) {
     }
   }
 
+  function handleAxesInversion() {
```
gabrieldutra, line 110: `handleAxesSwapping` to keep it consistent now? 🤔

viz-lib/src/visualizations/chart/plotly/updateAxes.js
```diff
@@ -38,6 +39,27 @@ export default function updateYRanges(plotlyElement, layout, options) {
         updates.yaxis2.range = calculateAxisRange(defaultRange, axisOptions.rangeMin, axisOptions.rangeMax);
       }
 
+      // Invert Axes
```
gabrieldutra, line 42:
```suggestion
      // Swap Axes
```

viz-lib/src/visualizations/chart/Editor/SeriesSettings.jsx
```diff
@@ -38,24 +38,27 @@ function getTableColumns(options, updateSeriesOption, debouncedUpdateSeriesOptio
...
-            left
-          </Radio>
-          <Radio value={1} data-test={`Chart.Series.${item.key}.UseRightAxis`}>
-            right
-          </Radio>
-        </Radio.Group>
-      ),
-    });
+  if (!includes(["pie", "heatmap"], options.globalSeriesType)) {
+    if (!options.swappedAxes) {
```
rafawendel, line 42: This allows the series picker to remain in the horizontal bar, which not only improves the UI, but also makes it easier to deal with state changes.

viz-lib/src/visualizations/chart/plotly/updateAxes.js
```diff
@@ -4,6 +4,42 @@ function calculateAxisRange(range, min, max) {
...
+    axisA.range[1] += calculateAbsoluteDiff(axisA.range[1], totalRangeA, diff);
+    axisB.range[0] -= calculateAbsoluteDiff(axisA.range[0], totalRangeB, diff);
+  } else {
+    // increase axisB max and axisA min
+    axisB.range[1] += calculateAbsoluteDiff(axisB.range[1], totalRangeB, diff);
+    axisA.range[0] -= calculateAbsoluteDiff(axisA.range[0], totalRangeA, diff);
+  }
+}
+
 export default function updateAxes(plotlyElement, seriesList, layout, options) {
```
rafawendel, line 43: @gabrieldutra I've merged the changes and kept the name of updateAxes rather than updateYRanges, since the module now deals with them both. The Align Y Axes at Zero feature seems to have retained it's functionality, as expected, but it'd be great if you could check on that.

rafawendel:
There is one further issue I noticed before we can merge this feature. It is actually an issue that is present in the production version of Redash. Please notice below:

Peek 2020-10-14 16-11

As you could see, there are some types of charts which will not trigger the UI change and instead default to line chart (gotcha switch case). The best approach IMO is to filter these types from the list, which should be as simple as edit `ChartTypeSelect`.

The misbehavior related to the horizontal bar chart is that it is more or less more restrictive than the regular chart, for example, it alters the baseline for the area chart. All needed is to pass down a filter array to `ChartTypeSelect` so as to allow the customization. 
image

After that I should check the state and clear the forbidden series, defaulting to line or bar. That should be the trickiest part which I am actually unsure of the importance.

kravets-levko:
Pie and Heatmap should not appear in dropdowns on Series tab, that's a bug. Those chart types are kinda special, they cannot be combined with others.

viz-lib/src/visualizations/chart/Editor/SeriesSettings.jsx
```diff
@@ -13,6 +13,9 @@ import getChartData from "../getChartData";
 const SortableBodyRow = sortableElement(props => <tr {...props} />);
 
 function getTableColumns(options, updateSeriesOption, debouncedUpdateSeriesOption) {
+  let restrictedChartTypes = ["pie", "heatmap", "bubble", "box"];
+  if (options.swappedAxes) restrictedChartTypes.push("area");
```
rafawendel, line 17: _issue_ does not trigger state change

viz-lib/src/visualizations/chart/Editor/SeriesSettings.jsx
```diff
@@ -13,6 +13,9 @@ import getChartData from "../getChartData";
 const SortableBodyRow = sortableElement(props => <tr {...props} />);
 
 function getTableColumns(options, updateSeriesOption, debouncedUpdateSeriesOption) {
+  let restrictedChartTypes = ["pie", "heatmap", "bubble", "box"];
```
rafawendel, line 16: What do you think of this approach @kravets-levko?

viz-lib/src/visualizations/chart/Editor/SeriesSettings.jsx
gabrieldutra: @kravets-levko I just noticed an issue from it: the `useMemo` won't recalculate when `options` changes (I think we are just mutating the updates? 🤔), in this case for example the list of available Charts will remain the same in the Series tab, regardless of the Swapping change:

tabs-wont-update

---

@rafawendel I'd also not limit the "Area" type from the feature, the result can still be achieved and I don't think it's that important.

Considering that, I recommend you to save yourself from the work of dealing with the possible cases:
Screen Shot 2020-10-14 at 18 29 37

Also, It feels better not to mess around with previous definitions when one is just testing out the checkbox behavior.

rafawendel: I thought so, that's why I preferred to wait your review before proceeding. Also not removing area sounds best, until I can fix the baseline issue some day.

Thank you! I'll push the minor changes ASAP so that we can move on! 😁

viz-lib/src/visualizations/chart/Editor/ChartTypeSelect.jsx
```diff
@@ -20,6 +20,10 @@ export default function ChartTypeSelect(props) {
       result.push({ type: "custom", name: "Custom", icon: "code" });
     }
 
+    if (filterTypes) {
+      return result.filter(({ type }) => !filterTypes.includes(type));
```
kravets-levko, line 24: Few thoughts re. this:
1. this prop need a better name (not clear how it "filter" work: removes items or keeps them). Maybe sth like "hiddenTypes", "unavailableTypes" or whatever else you prefer;
2. I'd use some lodash here, like: `return filter(result, ({ type }) => !includes(filterTypes, type))`, for me it feels a bit safer, but it's up to you;
3. now this component needs some proptypes 😁

viz-lib/src/visualizations/chart/Editor/SeriesSettings.jsx
```diff
@@ -64,6 +69,7 @@ function getTableColumns(options, updateSeriesOption, debouncedUpdateSeriesOptio
           data-test={`Chart.Series.${item.key}.Type`}
           dropdownMatchSelectWidth={false}
           value={item.type}
+          filterTypes={restrictedChartTypes}
```
kravets-levko, line 72: If it's the only place where `restrictedChartTypes` is used - just put the array here

kravets-levko (changes requested):
Just a few last things I'd like to see in this PR, and it's good to go 👍

rafawendel (commented):
Guess it's done! Replaced the array functions with lodash's and added proptyes, which should avoid any problems

kravets-levko (approved):
🌟

viz-lib/src/visualizations/chart/Editor/ChartTypeSelect.jsx
```diff
@@ -1,9 +1,9 @@
-import { map } from "lodash";
+import { filter, includes, map } from "lodash";
 import React, { useMemo } from "react";
```
kravets-levko, line 2:
```suggestion
import React, { useMemo } from "react";
import PropTypes from "prop-types";
```

rafawendel: Boy were you quicker than me to solve that, hahaha

viz-lib/src/visualizations/chart/Editor/ChartTypeSelect.jsx
```diff
@@ -1,21 +1,22 @@
 import { filter, includes, map } from "lodash";
 import React, { useMemo } from "react";
 import { Select } from "@/components/visualizations/editor";
+import PropTypes from "prop-types";
 import { visualizationsSettings } from "@/visualizations/visualizationsSettings";
 
+const result = [
```
kravets-levko, line 7: It needs better name now 🤔  `allChartTypes` maybe?

viz-lib/src/visualizations/chart/Editor/ChartTypeSelect.jsx
```diff
@@ -39,12 +40,10 @@ export default function ChartTypeSelect({ hiddenChartTypes, ...props }) {
   );
 }
 
-ChartTypeSelect.defaultProps = {
-  hiddenChartTypes: [],
+ChartTypeSelect.propTypes = {
+  hiddenChartTypes: PropTypes.arrayOf(PropTypes.oneOf(result.map(({ type }) => type))),
```
gabrieldutra, line 44:
```suggestion
  hiddenChartTypes: PropTypes.arrayOf(PropTypes.oneOf(map(result, "type"))),
```

Please rename it as @kravets-levko already mentioned, and just in case you want to use Lodash map instead 🙂 (it is already imported anyway)

kravets-levko: Wow, I didn't know that's supported by `_.map` 😅  I used this syntax with `_.filter`, `_.sortBy` but missed that other functions may support it 🙂

rafawendel (commented):
Funny thing is that I had thought of doing this. Hope to get better at it with time

kravets-levko:
> Funny thing is that I had thought of doing this. Hope to get better at it with time

Don't worry too much, you're already very good 👍 Everything else will come with experience (just learn all the time) 😉

viz-lib/src/visualizations/chart/Editor/ChartTypeSelect.jsx
```diff
@@ -1,26 +1,31 @@
...
+  { type: "bubble", name: "Bubble", icon: "circle-o" },
+  { type: "heatmap", name: "Heatmap", icon: "th" },
+  { type: "box", name: "Box", icon: "square-o" },
+];
 
+export default function ChartTypeSelect({ hiddenChartTypes, ...props }) {
+  const chartTypes = useMemo(() => {
     if (visualizationsSettings.allowCustomJSVisualizations) {
-      result.push({ type: "custom", name: "Custom", icon: "code" });
+      allChartTypes.push({ type: "custom", name: "Custom", icon: "code" });
```
kravets-levko, line 21: Since `allChartTypes` now is a global variable - you cannot modify it because it will cause side-effects. Add this line in the beginning of the function: 

```js
const result = [...allChartTypes];
```

and then use `result` everywhere in this function instead of `allChartTypes`

kravets-levko (changes requested):

viz-lib/src/visualizations/chart/Editor/ChartTypeSelect.jsx
```diff
@@ -16,16 +16,18 @@ const allChartTypes = [
 ];
 
 export default function ChartTypeSelect({ hiddenChartTypes, ...props }) {
+  const result = [...allChartTypes];
```
kravets-levko, line 19: This should be in `useMemo`'s function 🙂

kravets-levko (approved):

gabrieldutra:
Congrats on your first merged PR @rafawendel! 🎉

gaecoli:
this can be order by values?

PR #5159, opened 2020-09-09
Expire sessions after 6 hours of inactivity

rauchy:
- [x] Feature

## Description
Redash sessions don't expire and when they do expire - they refresh automatically because there's no way for users to opt out of "remember me" when they login.

This PR lets users opt out of "remember me" when they log in, effectively forcing their session to end (and not refresh) after 6 hours of inactivity. Every request extends the session by another 6 hours.

If users pick "Remember me", sessions would still expire, but get refreshed for up to 31 days, then they would have to log in again.
If they uncheck "Remember me", sessions would expire after 6 hours and would require a login.

### DIFF
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -2,9 +2,10 @@ import hashlib
 import hmac
 import logging
 import time
+from datetime import timedelta
 from urllib.parse import urlsplit, urlunsplit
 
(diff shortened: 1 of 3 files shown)


susodapop:
> does this require an application rebuild or will this be implemented after editing the files?

You won't need to rebuild the containers. Just bounce the server so it can refresh its environment variables.

PR #5177, opened 2020-09-16
Refresh CSRF tokens

rauchy:
- [x] Feature

## Description
CSRF tokens should expire and reset every once in awhile. This PR expires them every 6 hours.
To prevent a scenario where users keep their Redash tab open for over 6 hours and then can't use the app since they have a stale CSRF token, I've added a retry mechanism in Axios.

* Also turns out Axios does CSRF cookie to header on it's own, so ca3e44a simplifies our code a bit and removes a dependency.

#5055 

image

### DIFF
--- a/client/app/services/axios.js
+++ b/client/app/services/axios.js
@@ -13,7 +13,7 @@ const getData = ({ data }) => data;
 
 axios.interceptors.response.use(getData);
 
-createAuthRefreshInterceptor(
+export const authRefreshInterceptor = createAuthRefreshInterceptor(
   axios,
   failedRequest => {
     const message = failedRequest.response.data.message || "";
--- a/client/app/services/axios.js
+++ b/client/app/services/axios.js
@@ -1,10 +1,11 @@
 import axiosLib from "axios";
 import { Auth } from "@/services/auth";
 import qs from "query-string";
-import Cookies from "js-cookie";
 
 export const axios = axiosLib.create({
   paramsSerializer: params => qs.stringify(params),
+  xsrfCookieName: "csrf_token",
+  xsrfHeaderName: "X-CSRF-TOKEN",
 });
 
 const getData = ({ data }) => data;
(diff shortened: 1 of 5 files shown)


client/app/services/axios.js
```diff
@@ -1,26 +1,36 @@
 import axiosLib from "axios";
 import { Auth } from "@/services/auth";
 import qs from "query-string";
-import Cookies from "js-cookie";
+import createAuthRefreshInterceptor from "axios-auth-refresh";
 
 export const axios = axiosLib.create({
   paramsSerializer: params => qs.stringify(params),
+  xsrfCookieName: "csrf_token",
+  xsrfHeaderName: "X-CSRF-TOKEN",
```
arikfr, line 9: Worth adding some config around this to disable if needed.

client/app/services/axios.js
```diff
@@ -1,26 +1,36 @@
 import axiosLib from "axios";
 import { Auth } from "@/services/auth";
 import qs from "query-string";
-import Cookies from "js-cookie";
+import createAuthRefreshInterceptor from "axios-auth-refresh";
```
arikfr, line 4: @kravets-levko relevant for #5178.

arikfr (approved):

client/app/services/axios.js
rauchy: If the cookie is absent, the header is not added.

https://github.com/axios/axios/blob/master/lib/adapters/xhr.js#L119

client/app/services/axios.js
```diff
@@ -1,26 +1,36 @@
...
+  failedRequest => {
+    const message = failedRequest.response.data.message || "";
+    if (message.includes("CSRF")) {
+      return axios.get("/ping");
+    } else {
+      return Promise.reject();
+    }
+  },
+  { statusCodes: [400] }
+);
```
arikfr, line 27: Thinking about it further, I think this whole behavior needs to be somehow configurable. In the Databricks fork this behavior won't work/relevant but we would want to implement a different auth interceptor. 

Does the lib provide a way to remove an interceptor?

rauchy: Yes, you can easily eject interceptors.

kravets-levko (approved):

PR #5184, opened 2020-09-21
Keep additional URL params when forking a query

kravets-levko:

- [x] Other


arikfr:
The Cypress test failing seem unrelated (according to Cypres fails on master too 33% of the time).

CleanShot 2020-09-21 at 12 54 12@2x

@gabrieldutra FYI

PR #5186, opened 2020-09-21
Some Choropleth improvements/refactoring

kravets-levko:

- [x] Refactor
- [x] Bug Fix
- [x] Other

## Description

What's in this PR:

- [x] stop guessing key type; directly map query result field to GeoJSON properties;
- [x] fixed bug with crashing/going into infinite update loop when editing map bounds;
- [x] use client-side cache for GeoJSON files;
- [x] optimize built-in maps (remove irrelevant data);
- [x] add USA states map.

Partially replaces getredash/redash#4599

image


viz-lib/src/visualizations/choropleth/hooks/useLoadGeoJson.js
```diff
@@ -0,0 +1,39 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    const mapUrl = get(visualizationsSettings, `choroplethAvailableMaps.${mapType}.url`, undefined);
+
+    if (isString(mapUrl)) {
+      setIsLoading(true);
+      let cancelled = false;
+
+      const promise = cache.get(mapUrl, () => axios.get(mapUrl).catch(() => null));
```
gabrieldutra, line 20: This cache is to help with multiple references not reloading the json, right? (e.g.: in a dashboard)

Very interesting! 😁

viz-lib/src/visualizations/choropleth/maps/convert-projection.js
```diff
@@ -0,0 +1,40 @@
+// This helper converts USA map from Mercator projection to Albers (USA)
+// Usage: `node convert-projection.js > usa-albers.geo.json`
```
gabrieldutra, line 2: It would be interesting to have a way to differ those script files (extension, folder or name pattern), nothing nice to apply here in my mind now though.

gabrieldutra (approved):
Code looks good 👍

Only downside is not having the descriptive Target Field options anymore. Maybe we should maintain a Map for those on our own to make it easier to configure.

viz-lib/src/visualizations/choropleth/hooks/useLoadGeoJson.js
kravets-levko: Yes, exactly. GeoJSON files are pretty large (300-500kB), and previously if there were a lot of visualizations - each of them was loading own copy of the same file. Also, this cache works on a Query View/Source pages - if there are multiple Choropleth visualizations with the same map.

viz-lib/src/visualizations/choropleth/maps/convert-projection.js
kravets-levko: It's actually not necessarily should be reviewed as it's not a direct part of the project. I'll give some context to explain this. So, USA map usually uses an `Albers USA` projection (the one you can see here - when Alaska and Hawaii are placed in the bottom left corner). That projection, in opposite to, say, Mercator (which Leaflet uses my default) is not defined on the whole range of coordinates (-180..180/-90..90). This is also not a problem usually, but in our case visualization became crazy when tried to convert non-valid coordinates (computed an infinite zoom and entire map became 0 by 0 size). Since we don't display other layers, I decided just to convert the map from Mercator to Albers (that is - physically move Alaska and Hawaii to the place you see) and that's what this script does. Also, there is an original map - for the time when we'll find a proper way to do this.

kravets-levko:
> Only downside is not having the descriptive Target Field options anymore. Maybe we should maintain a Map for those on our own to make it easier to configure.

Can you please elaborate? Do you mean that Target Field values became less user-friendly (e.g. `iso_a3` vs. `ISO code (3 letters)`? The reason of this is that I plan to continue working on #4599 and if users will be able to load arbitrary GeoJSON - there is no other option except of just showing properties :slightly_frowning_face: Possible solution is to rename properties in the maps we maintain, but it's better to do this now and at once, because otherwise one day it may turn into backward compatibility hell.

viz-lib/src/visualizations/choropleth/maps/convert-projection.js
gabrieldutra: Sure, I realize it's not a direct part of the project, but it would be great to have something indicating this (name it `convert-project.script.js`, a `bin` folder or whatever). Not a must as I also didn't find a pattern that I liked, it's more in case anything pops up into your mind 🙂

kravets-levko: Eventually I'm gonna remove it, just decided to keep it for now for the reference

gabrieldutra:
> Do you mean that Target Field values became less user-friendly (e.g. iso_a3 vs. ISO code (3 letters)?

Yes, that's what I meant.

I wouldn't manually rename it in the files, but I see your point. Well, it's not a problem to maintain a Map with the friendly names for the ones we maintain. For the Custom ones 🤷‍♂️. If the Maps are per instance we can eventually have an UI to replace the props, but I wouldn't be that concerned about the custom ones anyway.

kravets-levko:
@gabrieldutra https://github.com/getredash/redash/pull/5186/commits/62ea4e887db7a272ba6763f0ccad2f5f856ea66a

image
image
image

gabrieldutra (commented):
Nice, thanks for handling that @kravets-levko :)

Lastly, should we select the first option when we change the Map and the previous option doesn't exist?
map-change

kravets-levko:
> Lastly, should we select the first option when we change the Map and the previous option doesn't exist?

I think no (at least - not now), this doesn't look like a big issue. With current behavior selected target field will stay when selecting another map and then returning back to previous, which also makes sense.

gabrieldutra (approved):

PR #5194, opened 2020-10-01
Add support for configuring a sentinel master_name

joshperry:
- [X] Feature

## Description

This adds a passthrough for setting the celery broker `master_name` value in the transport_broker_options so that redis sentinel can be used for HA pub/sub configurations.

We're on v8 still, so I'm not sure how this may translate to RQ. I'm sure we will eventually move to v9 and will need to figure out how to connect to HA redis clusters then.


joshperry:
We solved this with haproxy and tcp-check pointed at a redis sentinel cluster.

PR #5207, opened 2020-10-13
Upgrade RQ to v1.5

rauchy:
- [x] Refactor

## Description
It's time to upgrade our RQ from v1.1 to v1.5. We should consider moving away from `rq-scheduler` and use RQ's native scheduler after this.

#4092


redash/tasks/worker.py
```diff
@@ -100,7 +100,7 @@ def enforce_hard_limit(self, job):
         )
         self.kill_horse()
 
-    def monitor_work_horse(self, job):
+    def monitor_work_horse(self, job, queue):
         """The worker will monitor the work horse and make sure that it
         either executes successfully or the status of the job is set to
         failed
```
gjxdxh, line 106: maybe add `job.started_at = utcnow()` here, like in the 1.5 code

gjxdxh (approved):
Looks good, a small nitpick

PR #5209, opened 2020-10-13
Frontend updates from internal fork

gabrieldutra:

- [x] Other

## Description
See commits for guidance.
 
--

--

### DIFF
--- a/client/app/components/ApplicationArea/ErrorMessage.jsx
+++ b/client/app/components/ApplicationArea/ErrorMessage.jsx
@@ -1,8 +1,10 @@
-import { isObject, get } from "lodash";
+import { get, isObject } from "lodash";
 import React from "react";
 import PropTypes from "prop-types";
 
 import "./ErrorMessage.less";
+import DynamicComponent from "@/components/DynamicComponent";
+import { ErrorMessageDetails } from "@/components/ApplicationArea/ErrorMessageDetails";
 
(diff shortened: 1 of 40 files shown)


client/app/components/ApplicationArea/ErrorMessage.jsx
```diff
@@ -1,8 +1,10 @@
-import { isObject, get } from "lodash";
+import { get, isObject } from "lodash";
```
kravets-levko, line 1: 😁

client/app/pages/queries/QuerySource.jsx
```diff
@@ -1,17 +1,16 @@
-import { isEmpty, find, map, extend, includes } from "lodash";
-import React, { useState, useRef, useEffect, useCallback } from "react";
+import { extend, find, includes, isEmpty, map } from "lodash";
```
kravets-levko, line 1: Just noticed that imported symbols are sorted alphabetically. Is that Prettier who did it?

kravets-levko (approved):

client/app/pages/queries/QuerySource.jsx
gabrieldutra: No, I think I copied the file from the fork like that 🤔

kravets-levko: Im asking because that's not the only place where I noticed that (like the one in my comment above). Interesting 🤔

PR #5211, opened 2020-10-14
Rethink DialogWrapper

kravets-levko:

- [x] Refactor

## Description

Primary idea of these changes is to render dialogs within app tree (so they will inherit React contexts), keeping existing API.


client/app/components/ApplicationArea/routeWithApiKeySession.jsx
```diff
@@ -38,9 +39,9 @@ function ApiKeySessionWrapper({ apiKey, currentRoute, renderChildren }) {
   }
 
   return (
-    <React.Fragment key={currentRoute.key}>
+    <DialogHost key={currentRoute.key}>
```
gabrieldutra, line 42: Is there any difference in your mind per route type (api/user session) to have separate `DialogHost`'s here instead of just the `Default` one (or just have it in the ApplicationArea as it is used by all routes)?

client/app/components/DialogWrapper.tsx
```diff
@@ -0,0 +1,328 @@
+import { extend, map, merge, omit, filter, each, identity } from "lodash";
+import React, {useState, useImperativeHandle, useEffect, useMemo, useRef, useContext, PropsWithChildren} from "react";
+import { ModalProps } from "antd/lib/modal/Modal";
+import PropTypes from "prop-types";
+
+/**
+  How to:
+  =======
+
+  Use a dialog?
```
gabrieldutra, line 10: Maybe we should create a `README.md` for this docs eventually? (no need now)

client/app/components/ApplicationArea/routeWithApiKeySession.jsx
gabrieldutra: Is it to get any Context the `Router` may provide?

gabrieldutra (approved):
I didn't fully check the DialogWrapper changes (I should do this tomorrow), but I didn't want to block the merge here. The `DialogHost` seems a great solution for Contexts in dialogs anyway.

---

As I still have to test how this will behave in code, [this](https://ant.design/components/modal/#Modal.useModal()) is how Antd solved the Context problem for the dialogs. It basically has a `contextHolder` that you put where you want it to get the context from. Maybe it's useful as inspiration for us to expand here later too.

kravets-levko:
> As I still have to test how this will behave in code, this is how Antd solved the Context problem for the dialogs. It basically has a contextHolder that you put where you want it to get the context from.

It's kinda the same thing - component that you have to render within React tree, and methods bound to it. Just a bit different API.

client/app/components/ApplicationArea/routeWithApiKeySession.jsx
kravets-levko: The idea of Default host is to just keep existing dialogs working as before. If one day we'll provide a host explicitly to all `.showModal` calls - we could just replace it with a regular dialog host and remove Default host component at all. 

I added nested dialog hosts in session wrappers for several purposes: 

1. allow access to router context (as you mentioned);
2. to have an example of how it could be used.

E.g. in internal fork we have one more page wrapper with its own context, and dialogs will have to use it. Dialog host itself is a "cheap" component, so I see no problem if we'll have 1-2 more of them.

PR #5219, opened 2020-10-19
test: Fix Exception: Can't query private addresses.

koooge:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description
Hi there,
test_http.py failed on my env. That's  because `socket.gethostbyname('example.com')` returns 127.0.0.1.

```
...
self = <redash.query_runner.BaseHTTPQueryRunner object at 0x7f2e47896410>
url = 'https://example.com/', auth = ('username', 'password'), http_method = 'get'
kwargs = {}

    def get_response(self, url, auth=None, http_method="get", **kwargs):
        if is_private_address(url) and settings.ENFORCE_PRIVATE_ADDRESS_BLOCK:
>           raise Exception("Can't query private addresses.")
E           Exception: Can't query private addresses.

redash/query_runner/__init__.py:289: Exception
...
```

### DIFF
--- a/tests/query_runner/test_http.py
+++ b/tests/query_runner/test_http.py
@@ -3,6 +3,7 @@ from unittest import TestCase
 
 from redash.utils.requests_session import requests, ConfiguredSession
 from redash.query_runner import BaseHTTPQueryRunner
+from redash import settings
 
 
 class RequiresAuthQueryRunner(BaseHTTPQueryRunner):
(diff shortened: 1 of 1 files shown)


kravets-levko:
[`example.com`](https://en.wikipedia.org/wiki/Example.com) is a reserved domain intended for test purposes, and it shouldn't resolve to `127.0.0.1`. So fixing tests doesn't feel good in this particular case - you had to start with fixing your env

Issue #5208, opened 2020-10-13
Create Cypress E2E tests based on the test coverage report


gabrieldutra:
## Description
[Cypress](https://docs.cypress.io/) is the tool we've used for End-to-end testing for the past couple of years. Recently, we generated a [test coverage report](https://46059-13926404-gh.circle-artifacts.com/0/coverage/lcov-report/index.html) (#5137), which gives us a better view of areas that could benefit from some testing. Existing tests can be found in `client/cypress/integration` and docs for local testing can be found [here](https://redash.io/help/open-source/dev-guide/end-to-end-tests).

![cypress-example](https://user-images.githubusercontent.com/3356951/95858972-47438580-0d34-11eb-8e8c-4c3a19c5c83b.gif)

## Tests to be created
- [ ] Chart Visualization
  - [ ] Line
  - [ ] Bar
  - [ ] Area
  - [ ] Pie
  - [ ] Scatter
  - [ ] Bubble
  - [ ] Heatmap
  - [ ] Box
- [ ] List Pages
  - [ ] Queries List
  - [ ] Dashboard List
  - [ ] Alerts List
- [ ] Query "Add to dashboard" feature
- [ ] Details Visualization
- [ ] Edit Alert Destination
- [ ] Databricks Schema Browser

ask-imran:
I would love it give it a go, do you need all of them in a single PR ?

gabrieldutra:
Thanks @ask-imran, feel free to pick any of the ones I mentioned 👍 

The "Edit Alert Destination" (very similar to the existing `edit_data_source_spec.js`) should be a simple start if you are not used to the Redash interface, but LMK if you got interest in anything else. The docs should be of great help, but if you have any questions just send them in here 🙂.

ask-imran:
I have been having trouble running the app/cypress tests. Followed the [guide](https://redash.io/help/open-source/dev-guide/end-to-end-tests) 

after the dependencies installed running `npm build` throws error
running cypress tests fail during login, the login page as well looks a bit broken (may be due to forntend assets not being built)

Am sorry if it has been documented somewhere, and Instead of spamming the PR with the comments I can take it to the forum.

Please let me know if I have missed something.

gabrieldutra:
> on windows machine

[...]

PR #5220, opened 2020-10-19
Bar chart e2e test

rafawendel:
## What type of PR is this?

- [x] Other

## Description
According to the checklist proposed on issue [#5208](https://github.com/getredash/redash/issues/5208), adding chart end-to-end tests with Cypress is top priority test-wise.

This is a boilerplate of test that covers making a query and using it to generate a bar chart. The test will verify all tabs of `VisualizationEditor` and then plot a chart and verify that it is plotted by Plotly.

### Todos:

- [x] Create boilerplate
- [x] Add prints
- [ ] Add update test
- [x] Create final snapshot


gabrieldutra (commented):
Although it's good to check the chart exists in the Dialog, this doesn't get us as much value as the Percy snapshots (they track both rendering differences and the non-existence of the Charts). So for this I was thinking about having the `createChartThroughUI` able to create all sorts of charts and having this spec like this:

```js
function createChartThroughUI(visualizationName, extraStepsFunction = () => {}) {
  // basic Chart setup with shared steps
  extraStepsFunction(); // may need better naming, but the idea is to configure extra options and assertions here
  // assert chart is rendered in the Dialog preview
  // save it
  return visualizationId; // this will be useful to later build a dashboard
}

it("creates Bar charts", function() {
  const basicBarChart = createChartThroughUI("Basic Bar Chart");
  const barChartWithSwappedAxes = createChartThroughUI("Bar Chart with Swapped Axes", () => {
    // swap axes
    // any extra ui assertions, e.g.: check it correctly swaps the Labels in the editor
  });
  // any other Bar Chart variations we want to cover with testing

  createDashboardWithVisualizations([basicBarChart, barChartWithSwappedAxes]).then(dashboardUrl => {
    cy.visit(dashboardUrl);
    cy.percySnapshot("Bar Chart");
  });
})
```

For building a dashboard with multiple visualizations there is an example with the [Pivot table](https://percy.io/Redash/redash/builds/7387748/view/421638468/1280?mode=diff&browser=chrome&snapshot=421638468):

image

I'm not sure if the code is perfect to be reused, but I'd love to help with that as I think building dashboards with different configured visualizations are a way to cover many things at once 🙂.

LMK your opinion

client/cypress/integration/visualizations/chart_spec.js
```diff
@@ -0,0 +1,166 @@
+/* global cy */
+
+import { getWidgetTestId } from "../../support/dashboard";
+import { get } from "lodash";
```
gabrieldutra, line 4:
```suggestion
```

client/cypress/integration/visualizations/chart_spec.js
```diff
@@ -99,19 +98,23 @@ function assertAxesAndAddLabels(xaxisLabel, yaxisLabel) {
...
+
+    const widgets = [];
+    chartGetters.forEach((chartGetter, i) => {
+      const position = { autoHeight: false, sizeY: 8, sizeX: 3, col: (i % 2) * 3 };
+      cy.get(chartGetter).then(chart => {
+        widgets.push(cy.addWidget(dashboard.id, chart.id, { position }));
+      });
+    });
+
+    widgetsAssertionFn(widgets, dashboardUrl);
```
rafawendel, line 117: For some reason, this is not running 🤪. @gabrieldutra can you give me a little insight?

rafawendel:
Well I found out the trouble was that my loop would fill the array after it ran (because the push ran inside a .then). I read about this class of problems in Cypress and there doesn't seem to be a good solution. The `cy.all` Gabriel added some time ago shares the same limitations.

What I did at the end of the day was combining `as` and `get` statements to achieve the desired asynchronous behavior. It's quite cumbersome to use, but it is what it is.

Issue #4908, opened 2020-05-21
Multiselect dropdown slowness


iherasymenko:
### Issue Summary

Multi-select parameter dropdowns are super slow when the item list consist 900 items. 

### Steps to Reproduce

1. Create a multi-select parameter with either hard coded values or the query result
2. Try to select a few items

### Technical details:

* Redash Version: 9.0.0-alpha (dev)
* Browser/OS: latest Chrome/Firefox
* How did you install Redash: Docker

[This](https://github.com/getredash/redash/issues/4908#issuecomment-702679465) is what's missing to solve this issue

susodapop:
This is more feature request than bug. Can you open a thread on the [forum](discuss.redash.io) instead? Any `<select>` component is going to struggle with that many options. It probably needs an AJAX auto-complete system to really fly.

arikfr:
900 items is something that I would expect still to work. @iherasymenko can you elaborate on whether the UI itself becomes slugish or just filtering the list?

iherasymenko:
Here is the video. The issue is that after selecting and item on a huge list it takes ~ 2 sec to see it selected. For these two seconds the UI is frozen.

![redash_dropdown](https://user-images.githubusercontent.com/2904801/82609712-0c756100-9b8b-11ea-880f-e00acb257b87.gif)

gabrieldutra:
I created a query in the [preview environment](https://redash-preview.netlify.app/queries/381/) to track this issue. I believe what may be causing it is the code to check whether all selected options are on the list. I'll give a check on it and see if I have a quick solution to improve that :+1:

gabrieldutra:
Just checked, turns out I was wrong, that method runs faster than I thought. I had Lodash's `intersection` as an issue in mind, that should cause issues for really large sets, but 1000 - 10000 should still be fine. (My plan was to use an internally used `Set` to make the assertion faster 😅 ).

The problem is with Antd and it was fixed in v4, should work fine when we make the effort to upgrade from v3 (see [here](https://ant.design/components/select/#components-select-demo-big-data)).

iherasymenko:
@gabrieldutra Thank you very much for looking into this. Is updating Antd something envisioned? I'd be happy to contribute this change if this kind of change is aligned with the overall product vision/plan.

arikfr:
> Is updating Antd something envisioned?

Yes, but it's unclear how much effort it will require so currently wasn't prioritized.

gabrieldutra:
[...]

PR #5221, opened 2020-10-20
Multiselect dropdown slowness (fix)

rafawendel:
## What type of PR is this?

- [x] Bug Fix

## Description
Simple heuristic to fix issue #4908. Please refer to the issue for detailed info.

All this does is sorting all elements and grabbing the 80th percentile, which is my arbitrary way of ignoring too large values. It then standardizes the length of the item by multiplying it to the document's root font size.

I am looking forward to having a second opinion before I apply this to all select elements.


gabrieldutra (commented):
**Notes:**
1. This won't work for 3 elements for example:
Screen Shot 2020-10-20 at 11 39 31

What happens is that your length calculation should have a -1 to match the element index 

2. I'm not sure the extra complexity is that worth it 😬. It shouldn't be a problem to start, as an example, with `dropdownMatchSelectWidth={350}`. It will automatically add ellipsis to the items if needed and when you select those, the `<Select>` with the options will expand.

3. If you really want to make it based on the items length, let's simplify it a bit by using Lodash's `sortBy(list, "length")` and use `ceil` instead of `round`, this will make it better for small lists. Also the `font-size` in this context should be known, so no need to use `getComputedStyle`. That one is currently targeting the `document` + it's not always that you'll have elements rendered as the list is virtualized 🙂

client/app/components/ParameterMappingInput.jsx
```diff
@@ -213,7 +214,7 @@ export class ParameterMappingInput extends React.Component {
       <Select
         value={mapping.mapTo}
         onChange={mapTo => this.updateParamMapping({ mapTo })}
-        dropdownMatchSelectWidth={false}>
+        dropdownMatchSelectWidth={getItemOfPercentileLength(existingParamNames, 80)}>
```
gabrieldutra, line 217:
```suggestion
        dropdownMatchSelectWidth={false}>
```

We can keep this one with `false`, the number of Parameters in a Dashboard shouldn't be big enough to need virtualization.

client/app/lib/utils.js
```diff
@@ -166,3 +167,16 @@ export function formatColumnValue(value, columnType = null) {
...
+ * @param {Array|Object} list
+ * @param {number} percentile Closest hundreth percentile
+ * @param {string | string[]} sortIteratee Lodash's iteratee or iteratee list
+ */
+export function getItemOfPercentileLength(list, percentile, sortIteratee = "length") {
+  return (
+    String(sortBy(list, sortIteratee)[Math.ceil((list.length - 1) * (percentile / 100))]).length * 10
+    // 10 is the root document font-size
+  );
+}
```
kravets-levko, line 182: Few thoughts on code style:
1. need to check for list length - return value of this function is undefined if it's empty;
2. avoid one-liners - they add unnecessary cognitive load. Split this algorithm in steps: sort, then take item, then compute result;
3. using hardcoded font size doesn't feel good to me. Ideally, there should be a font size of popup container, or at least - of `<body>` tag.

Also, maybe add some limit below which the function will just enable Select's built-in measure function, and use this function only if there are too much items? In this case, you can omit item 1

viz-lib/src/lib/utils.js
```diff
@@ -41,3 +41,16 @@ export function formatColumnValue(value, columnType = null) {
...
   return value;
 }
+
+/**
+ * Gets the element length at closest (ceil) position from the nth percentile in a list
+ * @param {Array|Object} list
+ * @param {number} percentile Closest hundreth percentile
+ * @param {string | string[]} sortIteratee Lodash's iteratee or iteratee list
+ */
+export function getItemOfPercentileLength(list, percentile, sortIteratee = "length") {
```
gabrieldutra, line 51: No need to put it in the `viz-lib` utils as there are no visualizations using it.

rafawendel (commented):
Now fixed for the proper behavior and with Lodash (I'm getting there).

Notice that when there are 5 elements it is resized according to the 5th, ceiling of the 80th percentile, but not when there are 6 elements.

Peek 2020-10-20 14-11

client/app/lib/utils.js
gabrieldutra: 1. Agreed
2. and 3. I was actually gonna suggest to move the font-size away from it and having a general `getItemOfPercentile`, because the font-size will depend on the context. This way the "one"-liner would be quite simple :
```js
if (isArray(list) && list.length > 0) {
  const itemIndex = Math.ceil((list.length - 1) * (percentile / 100));
  return sortBy(list, sortIteratee)[itemIndex];
}
```

> Also, maybe add some limit below which the function will just enable Select's built-in measure function

Nice one 👍, we can create an util function for the dropdownWidth value that sets it to `false` if the length is below a given threshold (e.g.: 500)

rafawendel (commented):
I discussed a bit and decided to make a custom wrapper component to deal with that. It's still WIP, but I wanted to push it so you may suggest edits.

client/app/components/Select.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useRef, useState, useEffect } from "react";
 import { sortBy, get } from "lodash";
 import AntdSelect, { SelectProps } from "antd/lib/select";
 
-function getItemOfPercentile<T>(list: Array<T>, percentile: number, sortIteratee: string | string[] = "length") {
+function getItemOfPercentile<T>(list: Array<T>, percentile: number, sortIteratee = "length") {
```
rafawendel, line 5: String array will break get

client/app/components/ParameterMappingInput.jsx
```diff
@@ -1,10 +1,10 @@
 /* eslint-disable react/no-multi-comp */
 
-import { isString, extend, each, has, map, includes, findIndex, find, fromPairs, clone, isEmpty } from "lodash";
+import { isString, extend, each, has, map, includes, findIndex, find, fromPairs, clone, isEmpty, uniq } from "lodash";
 import React, { Fragment } from "react";
 import PropTypes from "prop-types";
 import classNames from "classnames";
-import Select from "antd/lib/select";
+import Select from "../components/Select";
```
kravets-levko, line 7:
```suggestion
import Select from "@/components/Select";
```

client/app/components/QueryBasedParameterInput.jsx
```diff
@@ -1,7 +1,7 @@
 import { find, isArray, get, first, map, intersection, isEqual, isEmpty } from "lodash";
 import React from "react";
 import PropTypes from "prop-types";
-import Select from "antd/lib/select";
+import Select from "../components/Select";
```
kravets-levko, line 4:
```suggestion
import Select from "@/components/Select";
```

client/app/components/Select.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return;
+  }
+  const sortedList = sortBy(list, sortIteratee);
+  const percentileIndex = Math.ceil((list.length - 1) * (percentile / 100));
+  return sortedList[percentileIndex];
+}
+
+const FONT_RATIO = 0.7;
+
+function Select({ style, options, ...props }: SelectProps<any>): JSX.Element {
```
kravets-levko, line 16: Since this component is not as generic as Ant's Select - it should have more specific name

client/app/components/Select.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const [dropdownMatchSelectWidth, setDropdownMatchSelectWidth] = useState<number | boolean>(false);
+  useEffect(() => {
+    if (options && options.length > 400) {
+      // using body because it's too messy to get a reference to the element itself
+      const fontSize = parseFloat(getComputedStyle(document.body).fontSize);
+      const itemOf80thPercentile = getItemOfPercentile(options, 80, "label.length");
+      if (itemOf80thPercentile) {
+        const len = String(itemOf80thPercentile.label).length;
+        setDropdownMatchSelectWidth(len * fontSize * FONT_RATIO);
```
kravets-levko, line 25: There is a way to measure string more precisely: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText
Why: fontSize is a _height_ of the font, but you need a _width_. Of course in this case it could be a good enough approximation, but if there's a way to do things properly - why not?

client/app/components/Select.tsx
```diff
@@ -0,0 +1,40 @@
...
+      }
+    }
+  }, [options]);
+
+  return (
+    <AntdSelect
+      dropdownMatchSelectWidth={dropdownMatchSelectWidth}
+      options={options}
+      style={{ minWidth: 60, ...style }}
```
kravets-levko, line 34: Why is this needed?

client/app/components/Select.tsx
kravets-levko: It shouldn't, these two examples are totally equal: `_.get(obj, "path.to.prop")` and `_.get(obj, ["path", "to", "prop"])`. But I may guess that the actual reason is that string array works differently with `_.get` and `_.sortBy` 🙂

client/app/components/Select.tsx
gabrieldutra: My first thought on avoiding that was that it could be slow for huge lists (>10k elements), but I didn't test it 🤔. If that works fast for, let's say, 100k strings, we can create a helper (similar to the `url` service) to make the right measurement.

Considering it accurate, we can actually use the actual Max from the list (not necessarily the 80th percentile anymore), bu that's at will.

kravets-levko: Actually, this measurement is needed only once, for the single item (am I right, @rafawendel?). So if could be very accurate (but I'm still not sure - should it be so accurate? 🤔)

kravets-levko: Another question that crossed my mind just now: if we have all items, and sort them by length, - why do we use 80th percentile item instead of longest one?

gabrieldutra: > Actually, this measurement is needed only once, for the single item (am I right, @rafawendel?). So it could be very accurate (but I'm still not sure - should it be so accurate? 🤔)

Although very unlikely, since the width per char can differ, it should be possible (I'm actually not sure if this case exists) to have a string with more characters to end up with less width. But we can assume that this won't happen 😆.

> Another question that crossed my mind just now: if we have all items, and sort them by length, - why do we use 80th percentile item instead of longest one?

I think the idea was to not grow it unnecessarily by excluding too large items from the calculation. To me this made more sense because the calculation was not really accurate.

gabrieldutra: Let's create this helper using a `<canvas>` (similar to `client/app/services/url.js`), this will be useful for not so simple cases, e.g.: To make a custom calculation when the Option is a Node (`dropdownMatchSelectWidth={iconWidth + anyMargin + textWithMaxLengthWidth}` for the Data Source dropdown)

kravets-levko: E.g. "www" (3 chars) is wider than "iiiii" (5 chars), so it's possible 🙂 Canvas.measureText will give us a proper text width in pixels.

image

gabrieldutra: > E.g. "www" (3 chars) is wider than "iiiii" (5 chars), so it's possible 🙂 Canvas.measureText will give us a proper text width in pixels.
> 
> image

Yes, but in this case we would calculate based on the "iiiii", not on the "www", because it has greater length

rafawendel: > Another question that crossed my mind just now: if we have all items, and sort them by length, - why do we use 80th percentile item instead of longest one?

I actually opted for the percentile because I wanted the size not to be skewed by outliers, e.g. one single very large word that would cause the width to be really large. At the beginning I thought on using average + 2 standard deviations, but the implementation would be a little cumbersome, then I decided to approximate it with percentiles.

Surely a solution with averages would be only O(n), and sorting is O(n log n), which is worse, but I didn't sound like too much of a problem, as having too much of my mind on optimizing it would make me waste a lot of time trying to find ways to avoid re-renders or re-calculating on every render.

rafawendel:
I'm not close to satisfied with this approach, as even with a lot of changes it doesn't provide the proper sizing.

client/app/lib/calculateTextWidth.ts
```diff
@@ -0,0 +1,16 @@
+const canvas = document.createElement("canvas");
+canvas.style.display = "none";
+document.body.appendChild(canvas);
+
+export function calculateTextWidth(text: string, container = document.body) {
+  const ctx = canvas.getContext("2d");
+  if (ctx) {
+    const containerStyle = window.getComputedStyle(container);
+    ctx.font = `${containerStyle.fontSize} ${containerStyle.fontFamily}`;
```
rafawendel, line 9: I don't really like this, but for some reason the font property returns an empty string

kravets-levko: Hm, interesting 🤔  At least from devtools it works just fine. Can you check it once more?

image

rafawendel: It might be a Firefox thing, take a look: 
image

rafawendel: That also raised my brows to the possibility of having browser inconsistencies with the canvas element. Wdyt?

gabrieldutra: It's not about the canvas @rafawendel:
Screen Shot 2020-10-23 at 10 24 27
[getComputedStyle docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle)

gabrieldutra: BTW doesn't it work with:
```js
ctx.fontSize = containerStyle.fontSize;
ctx.fontFamily = containerStyle.fontFamily;
```
?

rafawendel: Unfortunately not

rafawendel:
This is very strange. I used the canvas API to create an element with the exact size of the text and everything and compared it to the dropdown. It seems like either the width passed should include even the scrollbar...

image

rafawendel (commented):
I'm thinking this is getting close to being done. I've tested it now with some values and the offset of 40 pixels seems to be a good approximation of the margins, paddings and stuff. I really wanted to have it in a non-hardcoded way, but couldn't come up with a practical form of doing so.

@kravets-levko do you know any?

Besides that, I wanted to check if it sounds best to stick with a percentile or you guys rather do it with the max value.

kravets-levko:
> This is very strange. I used the canvas API to create an element with the exact size of the text and everything and compared it to the dropdown. It seems like the width passed should include even the scrollbar...

You compute a width of some select item, but in fact you need a width for dropdown element - its outer width. So you have to consider scrollbar (if any), paddings and so on.

I'll review the PR once more asap 👍

client/app/components/Select.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { useState, useEffect } from "react";
+import { maxBy } from "lodash";
+import AntdSelect, { SelectProps } from "antd/lib/select";
+import { calculateTextWidth } from "@/lib/calculateTextWidth";
+
+function Select({ options, ...props }: SelectProps<any>): JSX.Element {
```
gabrieldutra, line 6: I also don't have much better ideas to name this component 😬 , maybe `SelectWithVirtualScroll`?

client/app/components/Select.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const [dropdownMatchSelectWidth, setDropdownMatchSelectWidth] = useState<number | boolean>(true);
+  useEffect(() => {
+    if (options && options.length > 400) {
```
gabrieldutra, line 9: Let's move this `400` to a `const MIN_LENGTH_FOR_VIRTUAL_SCROLL` (or any better naming) below the imports to have the use clear and it work as a "definition" for the component.

client/app/components/Select.tsx
```diff
@@ -0,0 +1,28 @@
...
+      const largestOpt = maxBy(options, "label.length");
+
+      if (largestOpt) {
+        const offset = 40;
+        const optionText = String(largestOpt.label);
```
gabrieldutra, line 14: How about we extend `SelectProps` and `LabeledValue` to enforce the `label` in `LabeledValue` is a `string`? This feels better than casting it here, as this may be unexpected for `Node` values

gabrieldutra: An interesting thing is that since we are using `Lodash` to get the largest value, `null`/`undefined` (actually anything that doesn't have `.label.length`) values would be just "ignored". The enforcement is mostly to be a reminder that it won't work for `non-string` values.

client/app/components/ParameterValueInput.jsx
```diff
@@ -98,25 +96,20 @@ class ParameterValueInput extends React.Component {
...
+
     return (
       <Select
         className={this.props.className}
         mode={parameter.multiValuesOptions ? "multiple" : "default"}
         optionFilterProp="children"
         value={normalize(value)}
         onChange={this.onSelect}
-        dropdownMatchSelectWidth={false}
+        options={map(enumOptionsArray, opt => ({ label: String(opt), value: opt }))}
```
gabrieldutra, line 107: No need for this casting here, the `label` should be a String already + it is not big deal if it is `null` or `undefined`

gabrieldutra (commented):
Nice! 🚀

Overall this feels like the best approximation we can get :). I left a few comments on Typing/Castings for now, I'll run a few more tests with the feature in the meanwhile.

gabrieldutra:
Weird thing: Looks like this Canvas triggered Percy to increase the height for snapshots 🤔

arikfr (commented):
This feels like we're going down a rabbit hole of complexity and potential performance issues. How about we start with fixed width? Anything that's too long will be cut.

If we want to be friendly to users with long texts, let's add an option to the dropdown parameters to define its width (three options: "Small, Medium, Large").

gabrieldutra:
> This feels like we're going down a rabbit hole of complexity and potential performance issues. How about we start with fixed width? Anything that's too long will be cut.
> 
> If we want to be friendly to users with long texts, let's add an option to the dropdown parameters to define its width (three options: "Small, Medium, Large").

I initially thought the same thing @arikfr, but my conclusion on it was:
1. Complexity: This indeed is a thing, it is a trade-off, after all this is a new html element that's being used to compute text widths 😆. My consideration is that there could be cases where having this would be valuable (specifically thinking about improving experience for Schema Database selection).
2. Potential Performance issues: That was what initially made me drop this idea, but performance shouldn't be a problem because it only executes the calculation once per Dropdown and it's only enabled for Dropdowns with more than 400 options. In a similar setup the `url.js` also has one shared html element used to accurately parse Urls and happily this worked very well so far 🙂 

Define the size for the Select would work for parameters though

client/app/components/Select.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { useState, useEffect } from "react";
+import { maxBy } from "lodash";
+import AntdSelect, { SelectProps } from "antd/lib/select";
+import { calculateTextWidth } from "@/lib/calculateTextWidth";
+
+function Select({ options, ...props }: SelectProps<any>): JSX.Element {
+  const [dropdownMatchSelectWidth, setDropdownMatchSelectWidth] = useState<number | boolean>(true);
+  useEffect(() => {
```
gabrieldutra, line 8: BTW, this `useState` + `useEffect` can be an `useMemo` based on `options` instead :)

client/app/components/Select.tsx
```diff
@@ -0,0 +1,28 @@
...
+        const optionText = String(largestOpt.label);
+        const width = calculateTextWidth(optionText);
+        if (width) {
+          dropdownMatchSelectWidth = width + offset; // eslint-disable-line react-hooks/exhaustive-deps
+        }
+      }
+    } else {
+      dropdownMatchSelectWidth = false;
+    }
+  }, [options]);
```
gabrieldutra, line 23:
```suggestion
  const dropdownMatchSelectWidth = useMemo(() => {
    if (options && options.length > 400) {
      const largestOpt = maxBy(options, "label.length");

      if (largestOpt) {
        const offset = 40;
        const optionText = String(largestOpt.label);
        const width = calculateTextWidth(optionText);
        if (width) {
          return width + offset;
        }
      }
      return false;
  }, [options]);
```

Note: I didn't test this

client/app/components/Select.tsx
```diff
@@ -0,0 +1,29 @@
...
+        const width = calculateTextWidth(optionText);
+        if (width) {
+          return width + offset;
+        }
+      }
+
+      return true;
+    } else {
+      return false;
+    }
```
gabrieldutra, line 23:
```suggestion
    }
    return false;
```

gabrieldutra:
@rafawendel I've pinned Percy version for now to fix it. It seems that the weird space added to most snapshots was related to https://github.com/percy/percy-agent/issues/550 after https://github.com/percy/percy-agent/issues/342.

gabrieldutra (approved):

client/app/components/Select.tsx
```diff
@@ -0,0 +1,38 @@
...
+      return true;
+    }
+
+    return false;
+  }, [options]);
+
+  return <AntdSelect<string> dropdownMatchSelectWidth={dropdownMatchSelectWidth} options={options} {...props} />;
+}
+
+export default SelectWithVirtualScroll;
```
kravets-levko, line 38: Rename this file to match default export name, and it's good to go 👍

kravets-levko (approved):

client/app/components/Select.tsx
gabrieldutra: I missed this one 😅

PR #5224, opened 2020-10-22
Adds query_text to the job serializer

edouardjmn:
## What type of PR is this?

- [x] Other

## Description

This PR adds `query_text` to the `job` serializer. Useful to be able to retrieve the query which is running (with parameters) when job are queued or started (Currently, the query text is only returned in the `query_result` object when the job is finished)

### DIFF
--- a/redash/serializers/__init__.py
+++ b/redash/serializers/__init__.py
@@ -314,6 +314,7 @@ def serialize_job(job):
 
     status = STATUSES[job_status]
     result = query_result_id = None
+    query_text = job.args[0]
 
     if job.is_cancelled:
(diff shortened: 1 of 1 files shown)


rauchy:
I'm not sure this is a good idea in terms of privacy. The endpoints under the Jobs handler are only protected by the fact that Job IDs are unguessable UUIDs. I don't think we want to share potentially sensitive data (such as query text) there.

edouardjmn:
Thank you for your feedback @rauchy! I've made a new proposition, please see https://github.com/getredash/redash/pull/5265

PR #5227, opened 2020-10-23
Fix multiple render when editing sql

### LINKED ISSUE #5226: Multiple render when editing sql
### Issue Summary

When the result of a query is bug, and you're editing the sql in panel, you can feel that the user input is delayed and the page looks stuck for a little while.

### Steps to Reproduce

1. Open a edit source page which has a huge result table
2. try to edit the sql and you can see the problem
(You can check the  log which I put it in "viz-lib/src/visualizations/table/Renderer.jsx", it shows that every time I change the sql value the result visualization re-renders)

redash-bug

### Technical details:

* Redash Version: 9.0.0-beta
* Browser/OS: Chrome/Mac
* How did you install Redash: local

JerryYuanJ:

- [x] Bug Fix

Fixes getredash/redash#5226

### DIFF
--- a/client/app/pages/queries/components/QueryVisualizationTabs.jsx
+++ b/client/app/pages/queries/components/QueryVisualizationTabs.jsx
@@ -82,6 +82,18 @@ const defaultVisualizations = [
   },
 ];
 
+
+const CachedVisualizationRenderer = React.memo(
+  function CachedVisRender(props) {
+    return <VisualizationRenderer {...props} />
+  },
+  (prevProps, nextProps) => (
+    prevProps.queryResult === nextProps.queryResult &&
+    prevProps.filters === nextProps.filters &&
+    prevProps.visualization === nextProps.visualization &&
+    prevProps.onFiltersChange === nextProps.onFiltersChange
+  ))
+
 export default function QueryVisualizationTabs({
   queryResult,
   selectedTab,
@@ -144,7 +156,7 @@ export default function QueryVisualizationTabs({
             />
           }>
           {queryResult ? (
-            <VisualizationRenderer
+            <CachedVisualizationRenderer
               visualization={visualization}
               queryResult={queryResult}
               context="query"


client/app/pages/queries/components/QueryVisualizationTabs.jsx
```diff
@@ -82,6 +82,18 @@ const defaultVisualizations = [
...
+
+const CachedVisualizationRenderer = React.memo(
+  function CachedVisRender(props) {
+    return <VisualizationRenderer {...props} />
+  },
+  (prevProps, nextProps) => (
+    prevProps.queryResult === nextProps.queryResult &&
+    prevProps.filters === nextProps.filters &&
+    prevProps.visualization === nextProps.visualization &&
+    prevProps.onFiltersChange === nextProps.onFiltersChange
```
kravets-levko, line 94: Have you figured out which prop caused component re-rendering?

JerryYuanJ: Inside `QuerySource`, the change of sql value will lead to 
```
setQuery(extend(query.clone(), { query: queryText }));
``` 

and `QueryVisualizationTabs`  component uses `query.visualizations`
```
<QueryVisualizationTabs   queryResult={queryResult}   visualizations={query.visualizations}
... />
```
I added the other checks just for safety.

kravets-levko: That doesn't explain the issue. If we change `query` object itself, but don't touch other fields - React will not update components which use that fields. So in this case `visualizations={query.visualizations}` cannot cause component re-rendering.

I'm just trying to find the root cause of the issue, because memoizing the whole component just fixes an aftereffect and may just hide some real problem.

JerryYuanJ: Make sense. I'll try to figure it out  as well! Thanks ~

JerryYuanJ: I make a demo app on [https://codepen.io/JerryJ/pen/yLJgYqO?editors=0011](https://codepen.io/JerryJ/pen/yLJgYqO?editors=0011)，I update obj outside and the child component still re-renders even the age property is never touched. You can check that.

gabrieldutra: @kravets-levko doesn't the clone function recreate the visualizations, thus creating a new reference as well?

gabrieldutra: What is triggering this is indeed probably any update in the Query Source page (most likely the Query one in this case). We've never memoized Visualizations, thus, any parent rerender will trigger a rerender on it:

Screen Shot 2020-10-23 at 11 10 09

We currently have Visualization Widgets with `React.memo` for dashboards, but the visualizations themselves are not 😬. We do need to make sure this won't break anything.

JerryYuanJ: Maybe I can provide some tests, but I really don't know how many functions need to be tested. Any suggestions ?

guidopetri:
@JerryYuanJ , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

justinclift:
Reading the comments on this PR, it seems like the change itself may not actually be the needed fix.

Let's close this PR for now, and if someone wants to investigate + fix the underlying problem at a future point they can.

PR #5129, opened 2020-08-25
Misc changes to codebase back ported from internal fork

arikfr:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [x] Feature
- [x] Other

## Description

This is lots of small changes from the internal fork at Databricks. Most of them are unrelated to each other, but as they were "vetted" and tested in the forked version, it seemed safe to just group them all together.

Most of them aren't changing functionality but rather extending it.

See commit messages for descriptions.

PR #5229, opened 2020-10-23
Add root to all static asset paths that mis it

HendrikPetertje:

- [x] Bug Fix

## Description

Static assets are already prefixed with the root denomination in most of
Redash. There are just a few places where these paths are missing.
Normally this doesn't really matter if Redash is used like most people
do with the Open Source edition.

We however had to toggle the organizations feature on (using REDASH_MULTI_ORG),
at which point a prefixed organisation "name-slug" is added before all
page paths (http://localhost:5000/org_a/).

Now the static assets without root path were suddenly expected to be in
our organizations respective slug-namespace.
(http://localhost:5000/org_a/static/images/db-logos/postgres.png for
example).

This pull update 6 cases of "wrongly routed"
asset paths to be in line with a  "/" like the already existing cases listed below:

- https://github.com/getredash/redash/blob/a682265e1396cf46b7baf8cc593239b6129c5662/redash/templates/layouts/signed_out.html#L10-L12
- https://github.com/getredash/redash/blob/9097feb100b10e022df16e3c7dc3f490754a2f95/client/app/multi_org.html#L10-L12
- https://github.com/getredash/redash/blob/5afd0554d0b27b35177a21b2c1a6174e44d315fc/redash/templates/invite.html#L26
- https://github.com/getredash/redash/blob/5afd0554d0b27b35177a21b2c1a6174e44d315fc/redash/templates/login.html#L18
- https://github.com/getredash/redash/blob/de052ff02be9ff8d4187d273364c78ca22d65eb0/client/cypress/integration/visualizations/map_spec.js#L21

Before the fix:
Screenshot 2020-10-23 at 09 50 27
Screenshot 2020-10-23 at 09 50 44
Screenshot 2020-10-23 at 09 51 06

After the fix:
Screenshot 2020-10-23 at 10 02 54
Screenshot 2020-10-23 at 10 03 03
Screenshot 2020-10-23 at 10 03 32

### DIFF
--- a/client/app/components/PreviewCard.jsx
+++ b/client/app/components/PreviewCard.jsx
@@ -69,7 +69,7 @@ UserPreviewCard.defaultProps = {
 // DataSourcePreviewCard
 
 export function DataSourcePreviewCard({ dataSource, withLink, children, ...props }) {
-  const imageUrl = `static/images/db-logos/${dataSource.type}.png`;
+  const imageUrl = `/static/images/db-logos/${dataSource.type}.png`;
   const title = withLink ? <Link href={"data_sources/" + dataSource.id}>{dataSource.name}</Link> : dataSource.name;
   return (
     <PreviewCard {...props} imageUrl={imageUrl} title={title}>
--- a/client/app/components/empty-state/EmptyState.jsx
+++ b/client/app/components/empty-state/EmptyState.jsx
@@ -179,7 +179,7 @@ function EmptyState({
   ];
 
   const stepsItems = getStepsItems ? getStepsItems(defaultStepsItems) : defaultStepsItems;
-  const imageSource = illustrationPath ? illustrationPath : "static/images/illustrations/" + illustration + ".svg";
+  const imageSource = illustrationPath ? illustrationPath : "/static/images/illustrations/" + illustration + ".svg";
 
   return (
     <div className="empty-state bg-white tiled">
--- a/client/app/pages/queries/components/QuerySourceTypeIcon.jsx
+++ b/client/app/pages/queries/components/QuerySourceTypeIcon.jsx
@@ -2,7 +2,7 @@ import PropTypes from "prop-types";
 import React from "react";
 
 export function QuerySourceTypeIcon(props) {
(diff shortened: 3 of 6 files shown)


kravets-levko:
@arikfr I requested your review because it's related to getredash/redash#5129

HendrikPetertje:
Hey there, @kravets-levko & @arikfr, what is the status of reviewing this?

I can continue to stay checked out on my own fork over on our EC3 instance for a while, but I'd much rather skip the overhead of having to merge in in future updates to my fork and instead just track your Github version without the asset bug ;)

arikfr:
@HendrikPetertje your change breaks other things. I don't mind making the base path configurable, so the default remains `static/` while you can override it with `/static`, but we can't merge your PR as is.

HendrikPetertje:
I've been trying to figure out what would break, the only thing I've come up with so far where it breaks is if the entire Redash application sits in a sub-path (example.org/redash/...), But I needed to do some pretty awful hacks in NGinx to actually get there. I figure that anyone doing such in NGinx would need to add extra location paths to fix asset locations anyway.

Is this really a problem and would you be able to point me to what would actually break (if I'm overlooking things) if we merged this as is? If it is as broken as you say, then we should probably update most of the static JS & images paths that are already prefixed with a / (some of which I already mentioned above

> - https://github.com/getredash/redash/blob/a682265e1396cf46b7baf8cc593239b6129c5662/redash/templates/layouts/signed_out.html#L10-L12
> - https://github.com/getredash/redash/blob/9097feb100b10e022df16e3c7dc3f490754a2f95/client/app/multi_org.html#L10-L12
> - https://github.com/getredash/redash/blob/5afd0554d0b27b35177a21b2c1a6174e44d315fc/redash/templates/invite.html#L26
> - https://github.com/getredash/redash/blob/5afd0554d0b27b35177a21b2c1a6174e44d315fc/redash/templates/login.html#L18
> - https://github.com/getredash/redash/blob/de052ff02be9ff8d4187d273364c78ca22d65eb0/client/cypress/integration/visualizations/map_spec.js#L21

PR #5230, opened 2020-10-23
Pass requests metadata to the alerts destionation webhook

it1804:

- [x]  Feature

## Description
Allows to tell the webhook destination handler how the request was updated, from the scheduler, or manually updated. This can be important in some cases.
For example, in the "Each time when evaluated" trigger mode, you need to notify about the trigger state only at a specific time. Manual updates in this case can trigger alerts at unplanned times, which in turn will can lead to unnecessary and incorrect actions in the alert handler.
This pull request adds a json field to the webhook message like:
'metadata': {'Username': 'xxx', 'Query ID': 'xxx', 'Queue': 'queries', 'Job ID': '95713d6e-d9e7-423b-beff-caeca0bff41b', 'Query Hash': '99ce22afb73c45a34e171cf03767638f', 'Scheduled' : True}

### DIFF
--- a/redash/destinations/__init__.py
+++ b/redash/destinations/__init__.py
@@ -31,7 +31,7 @@ class BaseDestination(object):
     def configuration_schema(cls):
         return {}
 
-    def notify(self, alert, query, user, new_state, app, host, options):
+    def notify(self, alert, query, user, new_state, app, host, metadata, options):
         raise NotImplementedError()
 
     @classmethod
--- a/redash/destinations/chatwork.py
+++ b/redash/destinations/chatwork.py
@@ -29,7 +29,7 @@ class ChatWork(BaseDestination):
     def icon(cls):
         return "fa-comment"
 
-    def notify(self, alert, query, user, new_state, app, host, options):
+    def notify(self, alert, query, user, new_state, app, host, metadata, options):
         try:
             # Documentation: http://developer.chatwork.com/ja/endpoint_rooms.html#POST-rooms-room_id-messages
             url = "https://api.chatwork.com/v2/rooms/{room_id}/messages".format(room_id=options.get("room_id"))
--- a/redash/destinations/discord.py
+++ b/redash/destinations/discord.py
@@ -28,7 +28,7 @@ class Discord(BaseDestination):
     def icon(cls):
         return "fa-discord"
 
-    def notify(self, alert, query, user, new_state, app, host, options):
+    def notify(self, alert, query, user, new_state, app, host, metadata, options):
         # Documentation: https://birdie0.github.io/discord-webhooks-guide/discord_webhook.html
         fields = [
             {
--- a/redash/destinations/email.py
+++ b/redash/destinations/email.py
@@ -27,7 +27,7 @@ class Email(BaseDestination):
     def icon(cls):
         return "fa-envelope"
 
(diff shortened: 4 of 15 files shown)


it1804:
It might also make sense to add these metadata parameters to the built-in template variables when using a custom template to be able to access them in the message body for other types of destinations that have a limited received message format, unlike a webhook.
I did not make such an implementation because of the inconvenience of parsing, but if the setting option "send the message body as a json string" appeared, it would be great, and would greatly expand the possibilities of automating the processing of alerts.

guidopetri:
@it1804 , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

it1804:
I'm still very interested! Three years of waiting and I really hope this will be accepted, thanks!
I will try to update the PR in the coming days.

it1804:
Привет, Я обновил запрос на включение.

justinclift:
Good stuff @it1804. :smile:

@getredash/maintainers Anyone up for reviewing this?  It's passing all of our CI tests, Code coverage, etc. :smile:

it1804:
The purpose of this PR is to get full details of the request that triggered this webhook alert.
In particular, the webhook handler needs to determine whether the request was started by the scheduler or the user.
Some requests in my case must be executed at a strictly defined time in order for the notification handler to work. If you run it at a different time, a false alert may be generated because the contents of the tables will not be as expected. Having received the status (Scheduled) from the metadata, I can not only exclude from processing notifications generated when manually launching a request at an arbitrary time, but also log the name of the user who launched this request.
Metadata is the best and only way to get complete information about an event.

Example of a webhook message with attached metadata:
`
{
  "event": "alert_state_change",
  "alert": {
    "id": 218,
    "name": "test_alert_trigger: cnt_rate < 0.95",
    "options": {
      "op": "<",
      "value": "0.95",
      "muted": false,
      "column": "last_cnt_rate",
      "custom_body": "{{QUERY_RESULT_ROWS}}",
      "custom_subject": "test_alert_trigger"
    },
    "state": "triggered",
    "last_triggered_at": "2023-08-23T14:13:37.938Z",
    "updated_at": "2023-08-23T14:13:37.912Z",
    "created_at": "2023-05-03T08:57:24.681Z",
    "rearm": 1,
    "query_id": 6742,
    "user_id": 3525,
    "description": "[{'last_cnt_rate': 90}]",
    "title": "test_alert_trigger"
  },
  "url_base": "redash-server",
  "metadata": {
    "Username": "it1804@gmail.com",
    "query_id": "6742",
    "Queue": "queries",
    "Job ID": "d5865848-e4eb-4905-a89a-f378b67f868c",
    "Query Hash": "a98ed0a4ab30c676e20e4a833c382b37",
    "Scheduled": false
  }
}
`

justinclift:
I just tweaked the formatting for the above comment, so the JSON is easier to read. :smile:

guidopetri (approved):
Thanks for the context @it1804 . In-depth review: this looks great and like a good addition. There isn't a test for the added `metadata`, but also we don't have tests for most of this `destinations` code, so I'm good merging.

guidopetri:
Thanks again for your contribution, and I'm sorry it took so long - but we got it in after some waiting :)

PR #5231, opened 2020-10-23
added required to Form.Item and Input for better UI

rafawendel:

- [x] Bug Fix

## Description
Quite simple: just added a required to the applicable elements. I decided to add it both to the `Input` and the `Form.Item`, so that we get the red shadow when the user tries to submit and the red star indicating the requirement.

### DIFF
--- a/client/app/components/EditParameterSettingsDialog.jsx
+++ b/client/app/components/EditParameterSettingsDialog.jsx
@@ -145,6 +145,7 @@ function EditParameterSettingsDialog(props) {
             value={isNull(param.title) ? getDefaultTitle(param.name) : param.title}
             onChange={e => setParam({ ...param, title: e.target.value })}
             data-test="ParameterTitleInput"
+            required
           />
         </Form.Item>
         <Form.Item label="Type" {...formItemProps}>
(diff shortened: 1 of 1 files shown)


gabrieldutra (commented):
> so that we get the red shadow when the user tries to submit and the red star indicating the requirement.

This Red shadow you noticed seems to be a feature from Firefox for `required` inputs:

**Chrome**
param-dialog-chrome


**Firefox**
param-dialog-firefox

Antd has support for adding those both manually (see [here](https://ant.design/components/form/#components-form-demo-validate-static)) or using their form validation, but since applying here will not be so trivial, I don't think it's worth it.

It's fine if you wanna keep it with the `required` in the `Input` too, but can you add it to https://github.com/getredash/redash/blob/a31196aef8e3155229aa7430d43150350d2a38a7/client/app/components/EditParameterSettingsDialog.jsx#L56 as well, so we keep the Form consistent?

gabrieldutra:
BTW, I noticed that changing the first input will trigger the second one with the Red border, so maybe it's better without them here? 🤔

rafawendel:
> BTW, I noticed that changing the first input will trigger the second one with the Red border, so maybe it's better without them here? 🤔

Indeed. I'll leave them out for the moment, though I think they make a better UI. Having this solved would require checking the touched event before applying the red shadow.

gabrieldutra (approved):

PR #5236, opened 2020-10-25
Fix counter visualization not working on Safari

daniellangnet:

- [x] Bug Fix

## Description

In the current `master` branch, the counter visualization doesn't work in Safari at all (both desktop and mobile) and stays completely empty. This is because of a change that was introduced [in this PR](https://github.com/getredash/redash/pull/5044). 

This fixes the issue (while still keeping the max width/height for the query screen of the above PR working).

image


kravets-levko:
@daniellangnet Which Safari version do you use? 'cause that's what I see in Safari on our `master`:

image

daniellangnet:
@kravets-levko I'm on Safari 14. It was okay in the query editor, but not when added to a dashboard (see screenshot)

kravets-levko:
Got it. Seems both here and for dashboard root there's the same issue: for some reason, if there's a container with `position: relative` and child with `height: 100%` Safari fails to set a proper child height, which is totally not correct behavior.

daniellangnet:
Agreed. Although I'm not positive that a child with `position: absolute` is supposed to increase the height of an empty parent with `position: relative`. Could just be an implementation choice in Chrome / Firefox

kravets-levko:
@daniellangnet Add `height: 0` here (but please check in Chrome & FF if you can, or I'll check tomorrow):

https://github.com/getredash/redash/blob/6f9e79c64134d253d2ba264282bfb11fdd650db6/client/app/components/dashboards/dashboard-grid.less#L89-L92

daniellangnet:
@daniellangnet I just added `height: 0` as requested (although I'm not really sure I understand the purpose of it). Also confirming that I tested this in both Chrome and Firefox, latest versions respectively.

daniellangnet:
@kravets-levko I just added height: 0 as requested (although I'm not really sure I understand the purpose of it). Also confirming that I tested this in both Chrome and Firefox, latest versions respectively.

client/app/components/dashboards/dashboard-grid.less
```diff
@@ -89,6 +89,7 @@
     > .visualization-renderer-wrapper {
       flex-grow: 1;
       position: relative;
+      height: 0;
```
kravets-levko, line 92: Please try just this fix - for me it worked without others (also, I tried in Chrome a minute ago - it was fine too). If it works - let's not touch other styles.

kravets-levko:
> I'm not really sure I understand the purpose of it

I'm also not sure why it works. I found a mention of this fix on some forum, and somebody said that that's some bug in Safari - in some cases it fail to compute real element's height when it's offset parent has an auto height set in either way. Setting it explicitly to some value fixes the issue, and in combination with flexbox will result in a proper layout. I just tried it, and looks it really works, so...

P.S. I think, nobody really understands how CSS works 🙂 We all just use it like a magic

daniellangnet:
@kravets-levko haha, now I know what you mean! I tried it but that change alone doesn't fix it for me. The `height: 0` only works if I also remove the `position: relative` from `.counter-visualization-container`

client/app/components/dashboards/dashboard-grid.less
daniellangnet: Only works in connection with this other change for me: https://github.com/getredash/redash/pull/5236/files#diff-4e9dd1634afb5937186e1247809856826568d835e2167e33be776e08ae6ffc94L6

Everything else we could get rid of though

kravets-levko:
> The `height: 0` only works if I also remove the `position: relative` from `.counter-visualization-container`

> Oh Apple... the new IE.

🤦‍♂️ 

P.S. I never liked any browser-specific bugs and workarounds for them. I'll check "my" solution once more tomorrow, I hope it's possible to make it work in all browsers (and have some logical explanation).

daniellangnet:
Sounds good. Thank you @kravets-levko. FWIW, I think my original solution without the `height: 0` made sense to me from a CSS perspective. 

From what I understand, the issue was this:
- `counter-visualization-container` is the container that was used for the height calculation in JavaScript
- `counter-visualization-content` is the child container that is set to `position: absolute` so that it can expand to it's full height, before we apply the `scale()` to keep it container
- if we add `position: relative` to `counter-visualization-container`, then this container will only have it's own height, because the absolutely positioned `counter-visualization-content` inside would NOT push it's height
- this actually sounds like correct behavior to me, it just appears that Chrome & Firefox chose to implement it differently regardless which is why it worked there
- so if we remove the `position: relative` from `counter-visualization-container` then the reference container is one level higher up, therefore allowing the container to wrap itself around around it's child, up to the maximum that is allowed by it's own flex-container

I reverted my commit back to what it originally was, because I totally agree with you on your dislike for browser-specific workarounds in CSS 😃 🙌

dengc367:
me too. the counter visualization can not shown in my safari, which version is 14.0.1.

guidopetri:
@daniellangnet , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
This particular PR would be super simple to fix the merge conflict for.

Does anyone know if the problem is still occurring with our current Redash builds and modern Safari though?

PR #5262, opened 2020-11-07
Enable Boxplot to be horizontal

deecay:

- [x] Feature

## Description
Boxplot could be horizontal too.

image


rafawendel (approved):
Our tests ran smoothly and I've checked most inner workings of the horizontal charts and having it with boxplot doesn't seem to be a problem at all!
Thanks @deecay!

kravets-levko (approved):

PR #5267, opened 2020-11-13
Add reorder to dashboard parameter widgets

rafawendel:

- [X] Feature

## Description
This is a boilerplate that adds a state to manage `paramOrder` so as to allow reordering of dashboard/widget parameters.
Also, I wanted to discuss whether we should leave the reorder available always or only during dashboard editing.
This is not ready for review yet.

[REDASH-304](https://databricks.atlassian.net/browse/REDASH-304?atlOrigin=eyJpIjoiNTY0MWJlNmM1ZTA2NDc0YjgxNTJmMTM2MjhhYmUwZjkiLCJwIjoiaiJ9)

## Desktop Screenshots/Recordings (if there are UI changes)

Kapture 2020-12-28 at 17 11 14

## Todos:

 - [x] Add reorder for widgets
 - [x] Add reorder for dashboards
 - [x] Remove edit parameter button

## Known issues

 1. Widget reorder ist vertically offset, likely because it's inside another draggable.


client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
```diff
@@ -288,6 +304,10 @@ class VisualizationWidget extends React.Component {
     const { localParameters } = this.state;
     const widgetQueryResult = widget.getQueryResult();
     const isRefreshing = isLoading && !!(widgetQueryResult && widgetQueryResult.getStatus());
+    const onParametersEdit = parameters => {
+      widget.options.paramOrder = map(parameters, "name");
+      widget.save();
```
gabrieldutra, line 309:
```suggestion
      const paramOrder = map(parameters, "name");
      widget.options.paramOrder = paramOrder;
      widget.save("options", { paramOrder });
```

I didn't test, but this should be the syntax to edit only the `options`.

Apart from that, there may be some reason to include this in the `useEditModeHandler` hook, e.g.: debounce the widget save request and reuse the same state we have for Layout change. However, since this is a WIP, we can look into that later (I also didn't test myself yet how this will behave in the UI).

rafawendel:
I was adding a new `global_param_order` to the backend and discussing with Gabriel when I realized it will require a migration. @arikfr I wanted to discuss with you if, considering the scope of this feature and new features to come, it wouldn't be worth it to add an "options" object to dashboards so as to store param order, dashboard filters and other features to come.

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -66,12 +66,12 @@ function AddWidgetContainer({ dashboardOptions, className, ...props }) {
...
 
 AddWidgetContainer.propTypes = {
-  dashboardOptions: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
+  dashboardConfiguration: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
   className: PropTypes.string,
 };
 
 function DashboardComponent(props) {
-  const dashboardOptions = useDashboard(props.dashboard);
+  const dashboardConfiguration = useDashboard(props.dashboard);
```
rafawendel, line 74: Renamed this to avoid confusion

client/app/pages/dashboards/components/DashboardHeader.jsx
```diff
@@ -27,8 +27,8 @@ function buttonType(value) {
   return value ? "primary" : "default";
 }
 
-function DashboardPageTitle({ dashboardOptions }) {
-  const { dashboard, canEditDashboard, updateDashboard, editingLayout } = dashboardOptions;
+function DashboardPageTitle({ dashboardConfiguration }) {
```
rafawendel, line 30: All changes in this file are renamings, no need to go through it

client/app/services/dashboard.js
```diff
@@ -142,6 +142,8 @@ function transformSingle(dashboard) {
     dashboard.widgets = prepareDashboardWidgets(dashboard.widgets);
   }
   dashboard.publicAccessEnabled = dashboard.public_url !== undefined;
+  // Temporary
+  dashboard.options = { globalParamOrder: [] };
```
rafawendel, line 146: Just until business logic is added

client/app/services/widget.js
```diff
@@ -223,6 +239,17 @@ class Widget {
         return result;
       }
     );
+
+    if (!isArray(this.options.paramOrder)) {
+      this.options.paramOrder = map(localParameters, "name");
+    }
```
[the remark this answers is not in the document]
rafawendel: I agree is not needed, added it as a safeguard, soon I'll consider removing

rafawendel (commented):
Created a migration and followed what was done to previous JSON-like structures. The server is trowing an error instead of the empty object fallback, @gabrieldutra can you help me with this one?

redash/models/__init__.py
```diff
@@ -1099,6 +1099,7 @@ class Dashboard(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model
     tags = Column(
         "tags", MutableList.as_mutable(postgresql.ARRAY(db.Unicode)), nullable=True
     )
+    options = Column(db.Text)
```
gabrieldutra, line 1102: @rafawendel I think the error is because this is not reflecting the migration, maybe make it like https://github.com/getredash/redash/blob/1b8ff8e810aaa81c8aa8f28aef9e1ca3632b8521/redash%2Fmodels%2F__init__.py#L496 and then retrigger the script to generate it?

redash/models/__init__.py
```diff
@@ -1099,6 +1099,9 @@ class Dashboard(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model
     tags = Column(
         "tags", MutableList.as_mutable(postgresql.ARRAY(db.Unicode)), nullable=True
     )
+    options = Column(
+        MutableDict.as_mutable(postgresql.JSON), server_default="{}", default={}
```
rafawendel, line 1103: Using more modern JSON syntax

gabrieldutra (commented):
Based on the gifs, the parameters shouldn't have the Edit Gear button from the dashboard page. Can you remove it @rafawendel?

rafawendel:
> Based on the gifs, the parameters shouldn't have the Edit Gear button from the dashboard page. Can you remove it @rafawendel?

This is done already in #ee94809, I'll update the GIFs

client/app/services/dashboard.js
rafawendel: @gabrieldutra brought this back so that you don't need to clone it locally (now it shouldn't break Netlify preview)

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -114,22 +120,31 @@ function DashboardComponent(props) {
...
+          />
         }
       />
       {!isEmpty(globalParameters) && (
         <div className="dashboard-parameters m-b-10 p-15 bg-white tiled" data-test="DashboardParameters">
-          <Parameters parameters={globalParameters} onValuesChange={refreshDashboard} />
+          <Parameters
+            parameters={globalParameters}
+            onValuesChange={refreshDashboard}
+            sortable={canEditDashboard}
```
gabrieldutra, line 137: This is supposed to be `editingDashboard` to make it only available when in edit mode.

client/app/components/Parameters.jsx
```diff
@@ -146,15 +150,17 @@ export default class Parameters extends React.Component {
...
+
     return (
       <SortableContainer
-        disabled={!editable}
+        disabled={!sortable}
         axis="xy"
         useDragHandle
         lockToContainerEdges
         helperClass="parameter-dragged"
+        helperContainer={containerEl => (appendSortableToParent ? containerEl : document.body)}
```
gabrieldutra, line 163: What exactly does this fix?

client/app/services/dashboard.js
gabrieldutra: No need to keep it, I'm already testing locally with the backend chages.

client/app/components/Parameters.less
```diff
@@ -14,13 +14,14 @@
     height: 36px;
   }
 
-  .parameter-container.sortable-container & {
+  .parameter-container.sortable-container &:not(.parameter-dragged) {
```
gabrieldutra, line 17: Why change this styling? I did notice there are differences from what's applied to the Dashboard and Query vs what's applied to Widget (not a big difference though), so you may wanna confirm whether the behavior you wanted works for all the cases.

client/cypress/integration/dashboard/parameter_mapping_spec.js
```diff
@@ -24,32 +29,67 @@ describe("Parameter Mapping", () => {
...
+    }
+  };
+
+  const reorderTwoParameters = (param1, param2, type = "WIDGET") => {
+    cy.get(".parameter-block")
+      .first()
+      .invoke("width")
+      .as("paramWidth");
+
+    const dragParam = (paramName, offsetLeft, offsetTop) => {
```
gabrieldutra, line 60: Comments for Cypress:
1. I think now it makes sense to rename this file to `parameter_spec.js` (or create a new one at will) and separate the `describe("Parameter Mapping")` from the `describe("Parameter Dragging")`. Dragging should have enough scope to have its own test 😬
2. This method is the same as the one from queries, right? It's better if we create a `support/parameters.js` to put related helper functions and avoid the duplicate (don't worry if it doesn't have a folder yet, I'll soon refactor those helper functions' location)

gabrieldutra (changes requested):
Overall looks and works nicely, great work @rafawendel 🎉

Just one critical request for it to be available only in edit mode.  The others are basically questions and a Cypress suggestion as Dashboard parameter reordering has enough scope for a new test.

client/app/components/Parameters.jsx
rafawendel: By standard React Sortable HOC will append the draggable to the document body. The creator of our own component decided to add a container to it, which wasn't clear to me why, until I removed it and noticed minor undesirable UI changes. I brought it back and added this as an alternative to only change what needs to be.

In the case of widgets, it's necessary to append them to a component in an upper scope than that of their containers, as they are positioned using transforms, which mess with the final position of the draggable. I tried to workaround them, but it got too complex and this seemed clearer.

client/app/components/Parameters.less
rafawendel: This margin aligns the draggable to it's siblings when applied uniformly. Having this addition makes the element one drags "pop", ie offset, in a way that makes it clearer to the user. I found it a nice addition to the user interface.

client/app/pages/dashboards/DashboardPage.jsx
rafawendel: Initially I thought of having it available on view mode a nice to have feat, but I agree is something to be reserved to edit mode.

client/app/components/Parameters.less
gabrieldutra: It does cause an interesting effect, but one thing to notice is that it also causes the cursor to be slightly off the drag handle. I think it's better to update it in a follow-up if the idea is to update the styling and not to fix an issue (maybe add some shadow to cause an elevation effect as well).

client/cypress/integration/dashboard/parameter_mapping_spec.js
rafawendel: The method is not quite the same, but similar. I'll try to adapt them to something more general

client/cypress/support/parameters.js
```diff
@@ -0,0 +1,26 @@
...
+    .trigger("mousedown");
+
+  cy.get(".parameter-dragged .drag-handle")
+    .trigger("mousemove", offsetLeft, offsetTop, { force: true })
+    .trigger("mouseup", { force: true });
+}
+
+export function assertParameterPairSwapping(param1, param2, paramWidth) {
+  cy.server();
+  cy.route("POST", `**/api/*/*`).as("Save");
```
gabrieldutra, line 13: This feels a bit risky as any api POST will trigger it 😕. This specific function currently makes sense only for the Dashboard, so maybe we can leave it there for now? (we can just have the `Save` route from the `beforeEach`).

Anyway, let me know what you think in case you have a different idea on how to handle this.

client/cypress/integration/dashboard/parameter_spec.js
```diff
@@ -108,4 +128,21 @@ describe("Parameter Mapping", () => {
       cy.getByTestId("TableVisualization").should("contain", "StaticValue");
     });
   });
+
+  it.only("reorders parameters", function() {
```
gabrieldutra, line 132:
```suggestion
  it("reorders parameters", function() {
```

client/cypress/integration/dashboard/parameter_spec.js
```diff
@@ -0,0 +1,161 @@
...
+      cy.getByTestId("TableVisualization").should("contain", "StaticValue");
+    });
+  });
+
+  it("reorders parameters", function() {
+    // Reorder is only available on edit mode
+    cy.getByTestId("DashboardMoreButton").click();
+    cy.getByTestId("DashboardMoreButtonMenu")
+      .contains("li", "Edit")
+      .click();
```
gabrieldutra, line 134:
```suggestion
    // Reorder is only available on edit mode
    editDashboard();
```

importing from
https://github.com/getredash/redash/blob/de052ff02be9ff8d4187d273364c78ca22d65eb0/client/cypress/support/dashboard/index.js#L21-L27

gabrieldutra (approved):
One last nit, but again, nice work here 🎉

redash/models/__init__.py
```diff
@@ -1132,7 +1135,6 @@ def all(cls, org, group_ids, user_id):
                 ),
                 Dashboard.org == org,
             )
-            .distinct()
```
AkshayBz, line 1135: Intentional @rafawendel?

Lead to this issue: https://github.com/getredash/redash/issues/5466

Tri0L: Hello @susodapop! Any updates? Did you submit your changes?

Tri0L: Yes

susodapop: Thanks! I'm now able to reproduce. Working on a patch now.

susodapop: For posterity, the reason this call was removed is because SQLA raises a comparison error when making a generic `distinct()` because dashboards includes a JSON field...

@Tri0L I've opened a PR that fixes this in case you want to test it out: https://github.com/getredash/redash/pull/5516

Issue #5270, opened 2020-11-14
Data sources with "custom" syntax break new query editor in local docker setup


mparq:
### Issue Summary

Using a data source with syntax = "custom" will cause the New Query page to crash with

![image](https://user-images.githubusercontent.com/2523118/99151194-52594080-2667-11eb-82a3-07fd9046e535.png)

I've tested this only on the latest master on fa7ecca485b0a69887509711c3997cde5a3fc955 following the docker setup. Initially, I thought I was just seeing the same issue as #5213 but I am not using a Windows machine.

After debugging a bit, I've tracked down the root cause, but I'm not sure if this only affects the "local docker / webpack" setup or if this is a larger issue. Would appreciate help from core devs.

### Steps to Reproduce

1. Checkout master branch (I used fa7ecca485b0a69887509711c3997cde5a3fc955)
2. Follow instructions at https://redash.io/help/open-source/dev-guide/docker
3. Start server with `docker-compose up` and client webpack server with `npm start` and navigate to http://localhost:8080
4. Set up a new data source which has syntax = "custom" (currently, I found google_spreadsheets, azure_kusto, and graphite. I personally was testing redash out for kusto since my team uses it a lot, but I also tested and reprod with google_spreadsheets)
5. Navigate to "New Query" page
6. Page will crash with error screenshot above

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

Opening the browser console, I see

```
SyntaxError: Unexpected token '<'
```

but it is an unhelpful stack trace ending at ErrorMessage.jsx. It took a while to roll this back but I modified the global error handler in `ApplicationArea/index.jsx` to get access to the initial error event, which had the property `filename: "http://localhost:8080/mode-custom.js"`. (I will submit a PR to add a more helpful error message in this case). When I navigated to that file in browser, I received html contents corresponding to I think the main SPA index.html. I'm guessing something somewhere is falling back to responding with the contents of this file if a resource is not found?

I searched the codebase briefly and landed on this file `client/app/components/queries/QueryEditor/ace.js` which seems like it's the likely culprit here, but this is where I need help from somebody more experienced in the codebase to figure out what the issue is and what the best fix would be.

For now, I'm unblocked to test redash out locally by hardcoding the syntax to "sql" to use the query editor. Thanks for your work on redash! It looks great.

### Technical details:

* Redash Version:
* Browser/OS:
* How did you install Redash:

deecay:
[...]

PR #5271, opened 2020-11-14
[client] Fix local dev QueryEditor crash when using a data source with "custom" syntax

mparq:

- [x] Other

## Description

Issue: when using a data source with "custom" syntax, ace editor will try to retrieve a file from server, **mode-custom.js** which does not exist. When the file is not found, the server responds with a fallback html file which breaks the page when using the local webpack server. This change adds a stub define to `client/app/components/queries/QueryEditor/ace.js` so that we don't attempt to retrieve this file.

Also added a condition in the global error handler for the case when an html file is executed as a js script to produce a friendlier error message on the console.

I'm new to this codebase - let me know if you want me to change anything!

#5270


guidopetri:
@mparq , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
The concept of this PR seems fine.  I'm not personally able to judge the technical side of things though.

Lets see if it passes our CI tests after rebasing (which I've just triggered), then we get potentially look for someone to review and maybe merge it. :smile:

guidopetri (approved):
I'm not an expert in JS, but raising a more informative error looks like a good thing to me. The changes to `ace.js` were suggested by deecay in #5270 so I'm good with them too.

PR #5277, opened 2020-11-20
Remove unwanted props from Select component

rafawendel:

- [X] Bug Fix

## Description
Rather dummy solution for not having unwanted props passed down to Ant Select, which has been causing errors cluttering the console.

### DIFF
--- a/client/app/components/QueryBasedParameterInput.jsx
+++ b/client/app/components/QueryBasedParameterInput.jsx
@@ -77,7 +77,7 @@ export default class QueryBasedParameterInput extends React.Component {
   }
 
   render() {
-    const { className, mode, onSelect, optionFilterProp, queryId, ...otherProps } = this.props;
+    const { className, mode, onSelect, queryId, value, ...otherProps } = this.props;
     const { loading, options } = this.state;
     return (
       <span>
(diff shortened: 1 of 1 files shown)


client/app/components/QueryBasedParameterInput.jsx
```diff
@@ -89,7 +89,7 @@ export default class QueryBasedParameterInput extends React.Component {
           value={this.state.value}
           onChange={onSelect}
           options={map(options, ({ value, name }) => ({ label: String(name), value }))}
-          optionFilterProp="children"
+          optionFilterProp={optionFilterProp || "children"}
```
gabrieldutra, line 92: Why the `optionFilterProp` change instead of leaving it to be replaced from `otherProps`? If you are using it it shouldn't be an "unwanted" prop 😅

client/app/components/QueryBasedParameterInput.jsx
```diff
@@ -77,7 +77,7 @@ export default class QueryBasedParameterInput extends React.Component {
   }
 
   render() {
-    const { className, value, mode, onSelect, ...otherProps } = this.props;
+    const { className, mode, onSelect, optionFilterProp, queryId, ...otherProps } = this.props;
```
gabrieldutra, line 80: Also, removing the `value` from here will cause it to be replaced from `otherProps`, which is causing Cypress to fail

gabrieldutra (approved):

PR #5284, opened 2020-11-24
Fix QuerySourceDropdown value type

rafawendel:

- [X] Bug Fix

## Description
TLDR: added a type casting to get rid of an error from the console.

`dataSource.id` is a prop passed down via `value` to `QuerySourceDropdown`, which in turn not only has only string as an option on PropTypes, but also passes it down to AntD's `Select`, which also only accepts strings. I considered accepting string and number and casting them to string, but it added more complexity than simply casting in the mother component. Also keeping it clean can help spotting errors in the future.

### DIFF
--- a/client/app/pages/queries/QuerySource.jsx
+++ b/client/app/pages/queries/QuerySource.jsx
@@ -213,7 +213,7 @@ function QuerySource(props) {
                 <DynamicComponent
                   name={"QuerySourceDropdown"}
                   dataSources={dataSources}
-                  value={dataSource ? String(dataSource.id) : undefined}
+                  value={dataSource ? dataSource.id : undefined}
                   disabled={!queryFlags.canEdit || !dataSourcesLoaded || dataSources.length === 0}
                   loading={!dataSourcesLoaded}
                   onChange={handleDataSourceChange}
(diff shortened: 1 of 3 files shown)


rafawendel:
Something funny happened to Percy this time. @gabrieldutra can you ask it to re-test?

client/app/pages/queries/QuerySource.jsx
```diff
@@ -213,7 +213,7 @@ function QuerySource(props) {
                 <DynamicComponent
                   name={"QuerySourceDropdown"}
                   dataSources={dataSources}
-                  value={dataSource ? dataSource.id : undefined}
+                  value={dataSource ? String(dataSource.id) : undefined}
```
gabrieldutra, line 216: This seems to cause bugs to percy:

Screen Shot 2020-11-24 at 09 43 39

gabrieldutra: How about just update https://github.com/getredash/redash/blob/9097feb100b10e022df16e3c7dc3f490754a2f95/client%2Fapp%2Fpages%2Fqueries%2Fcomponents%2FQuerySourceDropdown.jsx#L32

to be `PropTypes.oneOfType([PropTypes.string, PropTypes.number])`? Considering it is supposed to be the id and we need to support both types, that should do it 🙂

gabrieldutra (approved):

PR #5285, opened 2020-11-24
Fix inconsistent Sankey behavior

rafawendel:

- [X] Bug Fix

## Description
Sankey wouldn't check whether the actual values in rows were valid before trying to plot them, which causes a strange behavior for VARCHARS. Fixed it by casting all values into numbers and then double-checking if the data is valid.

### Notes
It might be interesting to add a Cypress test for this and also displaying some sort of feedback for the user in case the input data is not valid. These should be left for another moment and issue IMO, though.

Refer to [REDASH-246](https://databricks.atlassian.net/browse/REDASH-246)

#### Old behavior
image

-----

#### New behavior
image
image


kravets-levko:
@rafawendel Looks that this PR contains some irrelevant changes. Please update it to keep only changes related to Sankey. Thanks!

PR #5286, opened 2020-11-24
Fix inconsistent Sankey behavior

rafawendel:

- [X] Bug Fix

## Description
Sankey wouldn't check whether the actual values in rows were valid before trying to plot them, which causes a strange behavior for VARCHARS. Fixed it by casting all values into numbers and then double-checking if the data is valid.

### Notes
It might be interesting to add a Cypress test for this and also displaying some sort of feedback for the user in case the input data is not valid. These should be left for another moment and issue IMO, though.

Refer to [REDASH-246](https://databricks.atlassian.net/browse/REDASH-246)

#### Old behavior
image

-----

#### New behavior
image
image


viz-lib/src/visualizations/sankey/initSankey.js
```diff
@@ -109,10 +123,33 @@ function spreadNodes(height, data) {
...
+function prepareData(data) {
+  const { rows } = data;
+  const result = map(rows, row => {
+    const newRow = {};
+    each(row, (value, key) => {
+      newRow[key] = +value;
+    });
+    return newRow;
+  });
+  data.rows = result;
```
kravets-levko, line 147: Add `mapValues` import from `lodash` and then:

```suggestion
  data.rows = map(data.rows, row => mapValues(row, v => parseFloat(v)));
```

Also, quick question: is it safe to mutate `data` object here?

viz-lib/src/visualizations/sankey/initSankey.js
```diff
@@ -109,10 +123,33 @@ function spreadNodes(height, data) {
...
 function isDataValid(data) {
   // data should contain column named 'value', otherwise no reason to render anything at all
-  return data && !!find(data.columns, c => c.name === "value");
+  if (!data || !find(data.columns, c => c.name === "value")) {
+    return false;
+  }
+  // prepareData will have coerced any invalid data into NaN, which is accumulated by the reducers below
+  const doAllRowsContainValidData = !isNaN(
+    reduce(data.rows, (finalAcc, row) => finalAcc * reduce(row, (acc, col) => col * acc, 0), 0)
+  );
```
kravets-levko, line 132:
```suggestion
  const doAllRowsContainValidData = every(data.rows, row => every(row, v => isFinite(parseFloat(v))));
```

In this case `every` is better than `reduce` because `every` will return on first value that does not match the condition while `reduce` will still process all the data. Also, with `every` you do directly what you mentioned in the comment - check if _every_ value is value, while `reduce` generally serves for a different purpose (fold the data).

P.S. You can even omit that intermediate variable and just return that expression's value from the function.

kravets-levko (commented):
Still need to understand changes better, but already have few suggestions (see comments)

viz-lib/src/visualizations/sankey/initSankey.js
rafawendel: In this case we could actually even coerce the numbers in the very function, but I find keeping them separate clearer and not significant performance-wise

rafawendel (commented):
I tried to make everything clearer by adding some types, but soon I got in a rabbit hole, so I decided to not proceed. Further typing these files will require some complex refactor which is out of the scope of this PR.

Despite that, I was able to remove most of the errors from the migration.

viz-lib/src/visualizations/map/getOptions.ts
```diff
@@ -1,6 +1,31 @@
 import { merge } from "lodash";
 
-const DEFAULT_OPTIONS = {
+export type LeafletBaseIconType = "marker" | "rectangle" | "circle" | "rectangle-dot" | "circle-dot" | "doughnut";
+export interface MapOptionsType {
+  latColName: string;
+  lonColName: string;
+  classify: any;
+  groups: object;
```
elado, line 8: please use `Record<string, any>` rather than `object`. if you know the structure of a group it's better to use actual structure than `any`.

rafawendel (commented):
Turns out strings should actually be allowed, so I did a few shenanigans so as to allow strings, but coerce number strings into true numbers.

kravets-levko (approved):

PR #5289, opened 2020-11-30
Fix problem with ElasticSearch

pavelpatrin:

- [x] Bug Fix

## Description

When I connected the ElasticSearch with latest Redash, I got this error.

```
Error running query: argument of type 'bool' is not iterable
```

Also, the list of Data Source tables was not loaded and popup with error was shown.

In the service calling get_schema() i found this exception:

```
Error retrieving schema: Traceback (most recent call last):
File "/app/redash/tasks/general.py", line 83, in get_schema
    return data_source.get_schema(refresh)
File "/app/redash/models/__init__.py", line 198, in get_schema
    schema = query_runner.get_schema(get_stats=refresh)
File "/app/redash/query_runner/elasticsearch.py", line 176, in get_schema
    columns.extend(parse_doc(items))
File "/app/redash/query_runner/elasticsearch.py", line 150, in parse_doc
    for field, description in doc["properties"].items():
KeyError: 'properties'
```

Next, I found this PR https://github.com/getredash/redash/pull/3692 and had improved it.

### DIFF
--- a/redash/query_runner/elasticsearch.py
+++ b/redash/query_runner/elasticsearch.py
@@ -122,6 +122,8 @@ class BaseElasticSearch(BaseQueryRunner):
         for index_name in mappings_data:
             index_mappings = mappings_data[index_name]
             for m in index_mappings.get("mappings", {}):
+                if not hasattr(index_mappings["mappings"][m], "__getitem__"):
+                    continue
                 if "properties" not in index_mappings["mappings"][m]:
                     continue
                 for property_name in index_mappings["mappings"][m]["properties"]:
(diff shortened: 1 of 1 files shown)


pavelpatrin:
@rauchy can I improve something in the code or provide additional information of the my setup for future investigations?

redash/query_runner/elasticsearch.py
```diff
@@ -147,7 +149,12 @@ def parse_doc(doc, path=None):
             """
             path = path or []
             result = []
-            for field, description in doc["properties"].items():
+            properties = []
+            try:
+                properties = doc["properties"].items()
+            except (TypeError, KeyError):
```
pavelpatrin, line 155: Both `TypeError` and `KeyError` are possible in my case. The `doc` may be a string instance (and this will be a type error). Also, `doc` may be a dict without the `properties` key. I got both errors while fixing this code.

redash/query_runner/elasticsearch.py
```diff
@@ -122,6 +122,8 @@ def _get_query_mappings(self, url):
         for index_name in mappings_data:
             index_mappings = mappings_data[index_name]
             for m in index_mappings.get("mappings", {}):
+                if not hasattr(index_mappings["mappings"][m], "__getitem__"):
```
pavelpatrin, line 125: I'm thinking about replacing this block with `try-except`, but this looks like a refactor more than bug fix.

PR #5290, opened 2020-12-01
Truncate large Databricks ODBC result sizes

christophergrant:

- [x] Feature

## Description

This PR adds a row limiter to the Databricks query runner to limit exposure to platform instability caused by users attempting to load huge result sets, resulting in Redash Web worker crashes. The PR adds an additional parameter to the query result data which can be used by the front end to signal any warnings in the UI associated with this query result.

### DIFF
--- a/redash/query_runner/databricks.py
+++ b/redash/query_runner/databricks.py
@@ -180,7 +180,7 @@ class Databricks(BaseSQLQueryRunner):
                 data = {"columns": columns, "rows": rows}
 
                 if (
-                    len(result_set) >= settings.DATABRICKS_ROW_LIMIT
+                    len(result_set) >= ROW_LIMIT
                     and cursor.fetchone() is not None
                 ):
                     logger.warning("Truncated result set.")
--- a/redash/query_runner/databricks.py
+++ b/redash/query_runner/databricks.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import os
 import sqlparse
 from redash.query_runner import (
     NotSupported,
--- a/redash/query_runner/databricks.py
+++ b/redash/query_runner/databricks.py
@@ -12,6 +12,7 @@ from redash.query_runner import (
     TYPE_INTEGER,
     TYPE_FLOAT,
 )
+from redash.settings import cast_int_or_default
 from redash.utils import json_dumps, json_loads
 from redash import __version__, settings, statsd_client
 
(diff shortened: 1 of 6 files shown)


redash/settings/__init__.py
```diff
@@ -511,4 +511,7 @@ def email_server_is_configured():
     os.environ.get("REDASH_ENFORCE_CSRF", "false")
 )
 
+# Databricks
+DATABRICKS_ROW_LIMIT = int(os.environ.get("REDASH_DATABRICKS_ROW_LIMIT", 20000))
```
arikfr, line 515: Let's move this definition directly into `redash.query_runner.databricks` as we do with other query runner specific settings.

redash/query_runner/databricks.py
```diff
@@ -157,10 +171,23 @@ def run_query(self, query, user):
...
+                if (
+                    len(result_set) >= settings.DATABRICKS_ROW_LIMIT
+                    and cursor.fetchone() is not None
+                ):
+                    # HACK: If this is a SQL endpoint (i.e. databricks_internal), "path" will exist
+                    endpoint_path = self.configuration.get("path", None)
+                    logger.warning(
+                        f"Truncated result set for query"
+                        f"path {endpoint_path}"
+                    )
```
arikfr, line 188: Let's simplify the log message:

```suggestion
                    logger.warning("Truncated result set.")
```

We will have to get the context from the surrounding log message related to the execution.

redash/settings/__init__.py
```diff
@@ -511,4 +512,7 @@ def email_server_is_configured():
...
```
susodapop, line 516:
```suggestion

```

susodapop: I see the setting was added to `databricks.py` but not removed here.

susodapop (changes requested):
This looks ready to go with one small change.

redash/query_runner/databricks.py
```diff
@@ -157,10 +174,18 @@ def run_query(self, query, user):
...
                 rows = [
                     dict(zip((column["name"] for column in columns), row))
-                    for row in data
+                    for row in result_set
                 ]
 
                 data = {"columns": columns, "rows": rows}
+
+                if (
+                    len(result_set) >= settings.DATABRICKS_ROW_LIMIT
```
susodapop, line 183:
```suggestion
                    len(result_set) >= ROW_LIMIT
```

susodapop: Since `settings` no longer includes DATABRICKS_ROW_LIMIT you need to use the constant declared in `databricks.py`

susodapop (approved):
LGTM

PR #5294, opened 2020-12-03
Redesign desktop nav bar

elado:

- [x] Other

## Description

Redesign the left nav.

- Removed collapsible side bar
- Bigger icons
- Mark active route link

### Before:

<table><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### After:
redesign-left-nav

redash-oss-menubar

### Mobile remains unchanged
image

### DIFF
--- a/client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
+++ b/client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
@@ -103,7 +103,6 @@ export default function DesktopNavbar() {
       </NavbarSection>
 
       <NavbarSection className="desktop-navbar-spacer">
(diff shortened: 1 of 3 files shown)


arikfr:
Two quick thoughts:

1. I wonder if we should remove the dividers? 
2. Make the icons at the bottom same size as the user image and have no text?

Other thought I had is what if we try to have a light navbar?

elado:
@arikfr 

1. removed dividers - indeed cleaner



2. made user thumb the same size as the other icons

3. lighter navbar would like like



not sure i'm a fan.

arikfr (approved):

ranbena:
@elado ON REDASH?! 🤯

PR #5299, opened 2020-12-04
Sync date format from settings with clientConfig

elado:

- [x] Bug Fix

## Description

Changing date format in settings didn't update `clientConfig` and dates remained in the previous format.
This makes the settings page also update `clientConfig`.

There's a deeper issue of inconsistency of settings (both session and organization settings API download settings, but in camelCase vs snake_case) and we should consolidate settings to a single source of truth.

Also, introducing `@testing-library/cypress` here.

### DIFF
--- a/client/cypress/integration/settings/organization_settings_spec.js
+++ b/client/cypress/integration/settings/organization_settings_spec.js
@@ -5,11 +5,6 @@ describe("Settings", () => {
   });
 
   it("renders the page and takes a screenshot", () => {
-    cy.getByTestId("OrganizationSettings").within(() => {
-      cy.getByTestId("DateFormatSelect").should("contain", "DD/MM/YY");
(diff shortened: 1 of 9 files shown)


client/cypress/integration/settings/organization_settings_spec.js
```diff
@@ -6,10 +6,40 @@ describe("Settings", () => {
 
   it("renders the page and takes a screenshot", () => {
     cy.getByTestId("OrganizationSettings").within(() => {
-      cy.getByTestId("DateFormatSelect").should("contain", "DD/MM/YY");
```
elado, line 9: removed this line since the value now depends on the current setting, which the next test can change. running both tests one after another will change this value.

elado:
ping @gabrieldutra @arikfr @kravets-levko

kravets-levko (approved):
LGTM 👍

PR #5303, opened 2020-12-07
Remove leftover log from Desktop Navbar

rafawendel:

- [X] Bug Fix

## Description
Simply removed a leftover `console.log` from  #5294


gabrieldutra (approved):
I wonder if we disabled `no-console` from eslint along the way 🤔. 

Do you mind checking this @rafawendel? My take is that it should at least be a warning

Issue #5282, opened 2020-11-23
Search does not work for Dropdown parameters


gabrieldutra:
### Issue Summary

![search-dropdown-params](https://user-images.githubusercontent.com/3356951/100015437-b93ccf00-2db6-11eb-85da-bd0ba35de7e2.gif)

This probably started to happen after #5221. The problem is that `optionFilterProp="children"` is not valid when options are provided through the `options` prop. Need to either make sure this prop is removed or to use `optionFilterProp="label"`

**Note:** Create a Cypress test (or add it to an existing test) to make sure the Filtering keeps working.

### Steps to Reproduce

1. Open a query with a Enum or a Query Based Parameter
2. Try to use the search

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* Redash Version: master
* Browser/OS: --
* How did you install Redash: --

PR #5304, opened 2020-12-07
Dropdown param search fix

rafawendel:

- [X] Bug Fix

## Description
Fix for query param dropdown filters not working. This happened since `SelectWithVirtualScroll` was introduced, which changed the way the dropdown filters work.

#5282


client/app/components/SelectWithVirtualScroll.tsx
```diff
@@ -32,7 +32,18 @@ function SelectWithVirtualScroll({ options, ...props }: VirtualScrollSelectProps
     return false;
   }, [options]);
 
-  return <AntdSelect<string> dropdownMatchSelectWidth={dropdownMatchSelectWidth} options={options} {...props} />;
+  if (options) {
+    optionFilterProp = "label";
+  }
```
gabrieldutra, line 37: Since this component already use the `options` syntax to work, how about just set this prop below?

```jsx
<AntdSelect<string>
  dropdownMatchSelectWidth={dropdownMatchSelectWidth}
  options={options}
  optionFilterProp="label" {/* or after {...props} if you really want to force it to be "label" */}
  {...props}
/>
```

client/app/components/QueryBasedParameterInput.jsx
```diff
@@ -89,7 +89,7 @@ export default class QueryBasedParameterInput extends React.Component {
           value={this.state.value}
           onChange={onSelect}
           options={map(options, ({ value, name }) => ({ label: String(name), value }))}
-          optionFilterProp="children"
+          optionFilterProp="label"
```
gabrieldutra, line 92:
```suggestion
```

This is defined already in `SelectWithVirtualScroll`

gabrieldutra (commented):
@rafawendel thanks for taking care of this 🙂. Can you please add a Cypress assertion to this? (Making sure that it fails without this fix)

No need to be a new test, but it can be just an assertion in one of the existing tests in `parameter_spec.js`.

client/cypress/integration/query/parameter_spec.js
```diff
@@ -91,7 +91,7 @@ describe("Parameter", () => {
     });
   });
 
-  describe("Dropdown Parameter", () => {
+  describe.only("Dropdown Parameter", () => {
```
gabrieldutra, line 94:
```suggestion
  describe("Dropdown Parameter", () => {
```

client/cypress/integration/query/parameter_spec.js
```diff
@@ -120,6 +120,15 @@ describe("Parameter", () => {
...
     });
 
+    it("supports filters", () => {
+      cy.getByTestId("ParameterValueInput").type("value1");
+
+      // only the filtered option should be on the DOM
+      cy.get(".ant-select-item-option").each($option => {
+        expect($option).to.contain("value1");
+      });
+    });
```
gabrieldutra, line 130: I still think this test is small enough that it would better fit as an assertion in one of the other existing specs ([Creating “tiny” tests with a single assertion](https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion))

You can actually create a function (`openAntdDropdownAndSearch` or sth) and apply it to tests on both the Enum and Query parameter.

gabrieldutra: It would even be better if we updated `cy.selectAntdOption()` to support filtering

rafawendel: I don't see any particular advantage on having it in both of them, as they use the same components. I made a helper function to run the assertion, but particularly thought it made the test less comprehensible, because at the end of the day, testing the filters is not really related to the other assertions. That was actually my first approach before I decided moving it to its own assertion.

gabrieldutra: **Advantage on having it in both of them:** they have different components backing them, so it's possible for it to stop working in only one parameter type. Since we won't be constantly checking that it still works, why not add a `~100ms` assertion to cover it in existing specs?

> testing the filters is not really related to the other assertions

Check the link I sent above.

If you are worried about comprehension, a function should be enough to do so. The bigger problem is that by doing this you switch an assertion that would take those 100ms into a test that will take the whole spec setup + browser load time. BTW, we have tests that will check that you can select a dropdown value and that will work for the query, adding a search before that selection is related, no?

client/cypress/integration/query/parameter_spec.js
```diff
@@ -149,6 +153,8 @@ describe("Parameter", () => {
         }
       });
 
+      openAndSearchAntdDropdown("value1"); // asserts option filter prop
```
gabrieldutra, line 156: Better to place it in the `"updates the results after selecting a value"`, no?

```js
it("updates the results after selecting a value", () => {
  /*cy.getByTestId("ParameterName-test-parameter")
    .find(".ant-select")
    .click();*/ // remove

  openAndSearchAntdDropdown("ParameterName-test-parameter", "value2"); // asserts option filter prop

  cy.contains(".ant-select-item-option", "value2").click();

  cy.getByTestId("ParameterApplyButton").click();
  // ensure that query is being executed
  cy.getByTestId("QueryExecutionStatus").should("exist");

  cy.getByTestId("TableVisualization").should("contain", "value2");
});
```

client/cypress/integration/query/parameter_spec.js
```diff
@@ -247,6 +253,8 @@ describe("Parameter", () => {
           cy.wrap($option).click();
         });
 
+        openAndSearchAntdDropdown("value1"); // asserts option filter prop
```
gabrieldutra, line 256: I initially thought the Query Based Parameters already had a similar test for only 1 value selection 😅. But I guess we can just create a new one (based on the Enum parameter "updates the results after selecting a value")?

client/cypress/integration/query/parameter_spec.js
```diff
@@ -1,3 +1,15 @@
+function openAndSearchAntdDropdown(testId, paramOption) {
+  // assures filters are working
+  cy.getByTestId(testId)
+    .find(".ant-select-selection-search-input")
+    .type(paramOption, { force: true });
+
+  // only the filtered option should be on the DOM
```
gabrieldutra, line 7: Not necessarily, your function doesn't enforce this 😝

client/cypress/integration/query/parameter_spec.js
```diff
@@ -1,3 +1,15 @@
+function openAndSearchAntdDropdown(testId, paramOption) {
```
gabrieldutra, line 1: Your function now also selects a value, so, you have two options:
1. Keep the selection separated 
2. If you want to have a function that also selects a value after a given filtering, we should put it into the existing `cy.selectAntdOption()`

I'd rather go with 1 as there are a few AntdOption helpers that we haven't ported yet (`selectAntdOption`, `selectAntdOptionLabel` and `selectAntdOptionByTestId`).

Also note that we don't need to validate that the option exists as `cy.click` already does that.

gabrieldutra (approved):

PR #5308, opened 2020-12-08
sometime,maybe self._profile_image_url is empty string

lpong:
I found that sometimes when using other login methods, the avatar is an empty string

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other


zhongjiajie:
In what situation profile url will be empty string?

justinclift:
Ignoring the codecov/patch failure here too.  Seems like a simple enough change, and one that's likely useful.

PR #5310, opened 2020-12-09
Convert viz-lib to TypeScript

elado:
- Used Airbnb's ts-migrate tool https://github.com/airbnb/ts-migrate
- Added some missing `@types/*` packages
- Reconfigured tsconfig, webpack, babel, build script
- Verified app and components build & run locally
- There are still plenty of `ts-expect-error` across the codebase, this is expected and a part of the migration. Some of them are actual errors (like accessing something that can be null) and some of them are as a result of bad type inference. They should all be cleaned up manually. Whenever editing a file we can remove a comment and see what is the actual error, then fix it.


elado:
cc @simonschneider-db

gabrieldutra (approved):
Thanks for doing this @elado, the result looks good to me and it's a good start point to gradually adopt TS 🎉

One question only: is the `OwnProps` name configurable? The pattern we were using tended to be `[ComponentName]Props`. It is not a problem as we will have to update the components in the future regardless, but just in case 🙂

elado:
@gabrieldutra I agree, it's not really nice, but didn't see configuration for this. I think it adds component name if it's exported. after getting this in I'll make another list of things to cleanup stuff.

PR #5314, opened 2020-12-11
Add Username and Password fields to MongoDB config

p-yang:

- [x] Other

## Description

This PR adds `Username` and `Password` fields to the MongoDB data source configuration, where the `Password` field is set to secret. The intention is to allow users to keep MongoDB passwords secret in the data source configuration. Previously this was not possible as the password would have to be embedded into the connection string, which was available in plaintext in the UI and REST API.

If a user supplies the `Username` and/or `Password` config field(s), Redash will use them to override whatever is set in the connection string when creating the PyMongo client. This allows for existing, pre-patch MongoDB sources to still function, while allowing new/modified MongoDB sources to have secret credentials

## Testing

Added a small unit test to verify that params are passed correctly to PyMongo, and manually tested by putting incorrect credentials in the connection string, and correct user/pass in config fields, and verifying that the mongodb datasource still worked.

image

### DIFF
--- a/redash/query_runner/mongodb.py
+++ b/redash/query_runner/mongodb.py
@@ -133,6 +133,8 @@ class MongoDB(BaseQueryRunner):
             "type": "object",
             "properties": {
                 "connectionString": {"type": "string", "title": "Connection String"},
+                "username": {"type": "string"},
+                "password": {"type": "string"},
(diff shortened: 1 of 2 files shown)


arikfr (approved):
It's great that it's backward compatible.

PR #5318, opened 2020-12-15
Obfuscate non-email alert destinations

p-yang:

- [x] Other

## Description

This PR obfuscates alert destination option fields which may carry credentials intended to be kept secret (Webhook URLs, API Tokens, etc.) by designating such fields as `secret`.


rauchy (approved):
LGTM

PR #5330, opened 2020-12-22
Extend sync_user_details expiry

rauchy:
- [x] Bug Fix

## Description
`sync_user_details` can occasionally get removed from the schedule because it has a short `interval` (60 seconds) resulting in a short `result_ttl` (120 seconds). If during these 120 seconds, for whatever reason the scheduler is unable to re-run the task, it expires and gets off the schedule.

This PR extends the grace period to re-run the sync job to 10 minutes.

### DIFF
--- a/redash/tasks/schedule.py
+++ b/redash/tasks/schedule.py
@@ -72,7 +72,12 @@ def periodic_job_definitions():
             "func": refresh_schemas,
             "interval": timedelta(minutes=settings.SCHEMAS_REFRESH_SCHEDULE),
         },
-        {"func": sync_user_details, "timeout": 60, "interval": timedelta(minutes=1),},
+        {
+            "func": sync_user_details,
+            "timeout": 60,
+            "interval": timedelta(minutes=1),
+            "result_ttl": 600,
+        },
         {"func": purge_failed_jobs, "timeout": 3600, "interval": timedelta(days=1)},
         {
             "func": send_aggregated_errors,


zhongjiajie (approved):
LGTM

susodapop (approved):

redash/tasks/schedule.py
```diff
@@ -72,7 +72,12 @@ def periodic_job_definitions():
             "func": refresh_schemas,
             "interval": timedelta(minutes=settings.SCHEMAS_REFRESH_SCHEDULE),
         },
-        {"func": sync_user_details, "timeout": 60, "interval": timedelta(minutes=1),},
+        {
+            "func": sync_user_details,
+            "timeout": 60,
+            "interval": timedelta(minutes=1),
+            "result_ttl": 600,
```
windream, line 79: Curious why we don't choose unlimited ttl by not setting `result_ttl` which is also recommended by rq-scheduler doc?

According to rq-scheduler's [doc](https://github.com/rq/rq-scheduler) :
```
IMPORTANT NOTE: If you set up a repeated job, you must make sure that you either do not set a result_ttl value or you set a value larger than the interval. Otherwise, the entry with the job details will expire and the job will not get re-scheduled.
```

`result_ttl` will be set to -1 if not provided ([rq scheduler code reference](https://github.com/rq/rq-scheduler/blob/master/rq_scheduler/scheduler.py#L242))

More explicitly, rq scheduler set `result_ttl` as -1 for cron job ([rq scheduler code reference](https://github.com/rq/rq-scheduler/blob/master/rq_scheduler/scheduler.py#L270-L271))

PR #5334, opened 2020-12-28
Google Spreadsheet Data Source: support to find a worksheet by title

yutopp:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

I propose a new feature "Select by Sheet Name" for google spread datasource which requested on Discuss. Adding this change, we will be able to use the spreadsheet without worrying about the order of the tabs.

I really want this feature. Please review this PR :pray: 

Thanks.

(↓new query)

image

(↓an error message when a tab name is not found)

image

https://discuss.redash.io/t/google-spreadsheet-data-source-select-by-sheet-name/5941


guidopetri:
@yutopp , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
This sounds like a useful PR.  I've just fixed the merge conflict and formatting problems (from the backend lint check).

If it passes our CI it'll still need someone to review it. :smile:

redash/query_runner/google_spreadsheets.py
```diff
@@ -88,14 +90,27 @@ def __init__(self, worksheet_num, worksheet_count):
...
+        if len(s) > 0:
+            if re.match(r"^\"(.*?)\"$", s):
+                # A string quoted by " means a title of worksheet
+                worksheet_num_or_title = s[1:-1]
+            else:
+                # if spreadsheet contains more than one worksheet - this is the number of it
+                worksheet_num_or_title = int(s)
 
-    return key, worksheet_num
+    return key, worksheet_num_or_title
```
guidopetri, line 113: I'm not a fan of mixed-type variables that can be one type or another. It makes it difficult to track what the contents of a variable might be like, and means we need to add a lot of "plumbing" type checking code, e.g. L135-140.

tests/query_runner/test_google_spreadsheets.py
```diff
@@ -104,6 +109,40 @@ def test_parse_query(self):
...
 
+    def test_parse_query_ignored(self):
+        parsed = parse_query("key")
+        self.assertEqual(("key", 0), parsed)
+
+        parsed = parse_query("key|")
+        self.assertEqual(("key", 0), parsed)
+
+        parsed = parse_query("key|1|")
+        self.assertEqual(("key", 0), parsed)
```
guidopetri, line 120: Confusingly, it seems that the sheet numbers are 1-indexed, but the returned values are 0-indexed. We might want to add a comment

tests/query_runner/test_google_spreadsheets.py
```diff
@@ -104,6 +109,40 @@ def test_parse_query(self):
...
+        parsed = parse_query('key|"abc"')
+        self.assertEqual(("key", "abc"), parsed)
+
+        parsed = parse_query('key|"あ"')
+        self.assertEqual(("key", "あ"), parsed)
+
+        parsed = parse_query('key|"1""')
+        self.assertEqual(("key", '1"'), parsed)
+
+        parsed = parse_query('key|""')
```
guidopetri, line 138: This looks identical to L123

tests/query_runner/test_google_spreadsheets.py
```diff
@@ -104,6 +109,40 @@ def test_parse_query(self):
...
+        self.assertEqual(("key", ""), parsed)
+
+    def test_parse_query_failed(self):
+        self.assertRaises(ValueError, parse_query, "key|0x01")
+        self.assertRaises(ValueError, parse_query, "key|a")
```
guidopetri, line 143: I think a fair expectation would be that a single-word / no-word-boundaries string would also be interpreted as a worksheet name, but I think that's out of scope for now.

guidopetri (approved):
I have a few comments on the PR, but given that this is almost 3 years old, I think we should just merge it in and have followups later. @justinclift are you good with that?

justinclift:
> @justinclift are you good with that?

Sure, sounds reasonable for this PR. :smile:

justinclift:
@guidopetri Just did the merge, are you ok to follow up with the other bits? :smile:

guidopetri:
Sure thing, though I'll probably prio the other open PRs I've currently assigned myself to. :D

guidopetri:
@yutopp , thanks for your contribution!

tests/query_runner/test_google_spreadsheets.py
Harsh-pandya: @guidopetri  Here, when calling the parse_query function with the string as "key|1|" then following execution happens inside the function.
Inside the file `redash.query_runner.google_spreadsheets.parse_query`
- At L104 the value of the variable `values` will be `["key", "1", ""]`
- At L107 there is a specific check if the length of `values` is 2 then only process the query further else return the default values which in this will be the value of the variable `key`(At L105) and `worksheet_num_or_title`(At L106)
- For passed query(`"key|1|"`) the total length of `values` will be 3 and so the default values (`tuple("key", 0)`) is returned.

With this cleared, the check on L120 in the test file passes, so I believe the sheets are 0-indexed.

PR #5338, opened 2020-12-31
Correct get_schema parameter

zhongjiajie:

- [x] Bug Fix

## Description

We pass value of `refresh` to parameter `get_stats` in function `query_runner.get_schema` which is wrong, and we should correct it by pass empty parameter


zhongjiajie:
@rauchy @arikfr @gabrieldutra @kravets-levko PTAL, thanks

zhongjiajie:
it seems this change it not correct

PR #5340, opened 2021-01-04
Reset failure counter on adhoc success

rauchy:
- [x] Bug Fix

## Description
Disclaimer: I might be changing the desired behavior here, so feel free to reject this PR.

Seems like successful executions for scheduled queries reset the failure counters, while successful executions for manual runs of those queries do not reset the counter. This PR uses `metadata` to lookup the query (this is done to avoid breaking APIs, but in the future, we can simplify the signature for `execute_query`) and resets the `schedule_failure` counter for all types of executions (scheduled / unscheduled).

### DIFF
--- a/redash/tasks/queries/execution.py
+++ b/redash/tasks/queries/execution.py
@@ -210,7 +210,7 @@ class QueryExecutor(object):
                 track_failure(self.query_model, error)
             raise result
         else:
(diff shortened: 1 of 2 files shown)


rauchy:
No, no, no. I don't like this.

Issue #4269, opened 2019-10-21
quarterly interval in cohort


thuanbvv:
Please add Quarterly (3 months) cohort group timeInterval in cohort chart UI.

![image](https://user-images.githubusercontent.com/17841056/67176230-68a8b480-f3f3-11e9-9ae7-8c7d1b978e13.png)
![image](https://user-images.githubusercontent.com/17841056/67176277-97bf2600-f3f3-11e9-88f6-cc60b8e463d5.png)

kravets-levko:
Hi @thuanbvv! This issue actually appears to be a feature request :smile: But anyway - we collect feature requests in a dedicated category on our [forum](https://discuss.redash.io/c/feature-requests) - feel free to move this topic there.

PR #5342, opened 2021-01-05
Feature: quarterly interval in cohort #4269

zjsun:

- [x] Feature

## Description

Support quarterly interval in cohort #4269

#4269

image


zjsun:
I'd never changed the source structure, only added some 'interval' related options like "quarterly" etc.

How could I pass the build and e2e-tests?

kravets-levko:
> How could I pass the build and e2e-tests?

You passed! 😅

rafawendel:
Hi @zjsun thanks for your work here.

I took a look at the implementation of this feature and despite the categories showing up properly in the UI, the logic behind them seems to not work properly. That is, the quarters do not match with the monthly results. Please tell me if otherwise.

It is basically defined here:

```
  _.each(grouped, group => {
    if (previousDate !== null) {
      // @ts-expect-error ts-migrate(2571) FIXME: Object is of type 'unknown'.
      let diff = Math.abs(previousDate.diff(group.date, momentInterval[timeInterval]));
      while (diff > 1) {
        const row = [0];
        for (let stage = firstStage; stage <= lastStage; stage += 1) {
          // @ts-expect-error ts-migrate(2571) FIXME: Object is of type 'unknown'.
          row.push(group.values[stage] || 0);
        }
        data.push(row);
        // It should be diagonal, so decrease count of stages for each next row
        lastStage -= 1;
        diff -= 1;
      }
    }
```

I couldn't grasp immediately why it didn't work, so I can better investigate it soon, if you want to.

Also, @gabrieldutra it might a nice opportunity to add a Cypress test for the Cohort viz. WDYT?

zjsun:
> I took a look at the implementation of this feature and despite the categories showing up properly in the UI, the logic behind them seems to not work properly. That is, the quarters do not match with the monthly results. Please tell me if otherwise. 

Thanks for your reviewing. @rafawendel 

When using quarterly interval option, The cohort date in the query result is based on the rule which is based on the first day of the quarter(which is mentioned [here](https://redash.io/help/user-guide/visualizations/cohort-howto) ), eg. 2019-01-01, 2019-04-01, 2019-07-01 ... etc.

FYI.

Avey777:

guidopetri:
@zjsun , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

PR #5345, opened 2021-01-06
Remove unnecessary space in rq log

zhongjiajie:

- [x] Other

## Description

I found out we have a unnecessary space in rq log and I think we should remove it. The example of the log like

```log
-- between "ds_id=1" and "job_id"
query_hash=53d2d2df553b8410dcc797bbbd79aa69 type=pg ds_id=1  job_id=d0061efc-cfad-4ccd-bd2b-24f2e2463dc1
```

### DIFF
--- a/redash/tasks/queries/execution.py
+++ b/redash/tasks/queries/execution.py
@@ -248,7 +248,7 @@ class QueryExecutor(object):
 
     def _log_progress(self, state):
         logger.info(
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks!

zhongjiajie:
Hi @arikfr , could you spend more time to review those PR?

PR #5349, opened 2021-01-14
Fix for Cypress flakiness generated by param_spec

rafawendel:

- [X] Bug Fix

## Description
Caused by #5286
Due to how Cypress behaves within loops, sometimes the next iteration is reached before current specs are evaluated, raising an unexpected behavior. This attempts to fix it.


rafawendel:
The current fix solves the previous problem, but a new one has arisen: every few runs Cypress will try to spec the closed popover before it was actually closed by the button click.

gabrieldutra (approved):

PR #5309, opened 2020-12-09
Use legacy resolver in pip to fix broken build

vipulmathur:
## What type of PR is this? (check all applicable)

- [x] Bug Fix

## Description

There have been upstream (`python:3.7-slim` image) changes that
bring in `pip` version 20.3.1, which makes new `2020-resolver`
the default.  Due to that, un-resolvable dependency conflicts
in  `requirements_all_ds.txt` now cause the build to fail.

This is a workaround until the package versions can be updated
to work with the new pip resolver.

## Related Tickets & Documents

Fixes #5300 and fixes #5307

PR #5350, opened 2021-01-15
install older pip to avoid dependency conflicts

tbarbugli:
PIP recently introduce strict dependency checks, while this is a good thing it basically breaks a large amount of libs in ways that are not trivial to fix (ie. dependency of a dependency).

Using an older version of PIP is the wise thing to do until the ecosystem gets around this change of pinning dependencies.

```
The conflict is caused by:
    atsd-client 3.0.5 depends on python-dateutil
    azure-kusto-data 0.0.35 depends on python-dateutil>=2.8.0
    dql 0.5.26 depends on python-dateutil<2.7.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,6 +82,9 @@ ENV PIP_NO_CACHE_DIR=1
 # We first copy only the requirements file, to avoid rebuilding on every file
 # change.
 COPY requirements.txt requirements_bundles.txt requirements_dev.txt requirements_all_ds.txt ./
+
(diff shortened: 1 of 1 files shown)


arikfr:
Thanks! As #5309 resolves this, I'm closing this one.

PR #5351, opened 2021-01-15
update dql

tbarbugli:
dql 0.5.26 had a hard requirement on dateutil which was causing dependency conflicts


susodapop:
DQL isn't supported any longer, is it?

PR #5354, opened 2021-01-15
Fix #4402. Removed token from the required list.

### LINKED ISSUE #4402: Minor Salesforce runner fix
### Issue Summary

A Security Token isn't required in all SFDC environments - depending on configuration. See [here](https://help.salesforce.com/articleView?id=000331668&type=1&mode=1) for more information.

### Steps to Reproduce

1. Add Salesforce as a data source where a token isn't required (and cannot be generated)
2. Cannot proceed without required field

### Technical details:

https://github.com/getredash/redash/blob/be56035bd6d9856361edc6b23d30a38c8f2d2be2/redash/query_runner/salesforce.py#L81

Just remove `token` from the `required` list. Seemed like it'd be faster to create an issue than submit a PR for such a small change

mariguranule:

A Security Token isn't required in all SFDC environments - depending on configuration.
Just remove token from the required list.
- [x] Bug Fix

https://github.com/getredash/redash/issues/4402


redash/query_runner/salesforce.py
```diff
@@ -78,7 +78,7 @@ def configuration_schema(cls):
                     "default": DEFAULT_API_VERSION,
                 },
             },
-            "required": ["username", "password", "token"],
+            "required": ["username", "password"],
             "secret": ["password", "token"],
```
zhongjiajie, line 82: If you remove `token` from required, you should remove it from secret too.
```suggestion
            "secret": ["password"],
```

zhongjiajie (commented):
cc: @arikfr

guidopetri:
@mariguranule , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

justinclift:
@zhongjiajie I've added your suggestion to this as a commit, but now I'm thinking over it more and not sure it's a good idea.

> If you remove token from required, you should remove it from secret too.

If it gets removed from `secret` as well (like in this commit), isn't that a problem for when the token *is* present?

zhongjiajie:
> @zhongjiajie I've added your suggestion to this as a commit, but now I'm thinking over it more and not sure it's a good idea.
> 
> > If you remove token from required, you should remove it from secret too.
> 
> If it gets removed from `secret` as well (like in this commit), isn't that a problem for when the token _is_ present?

My bad, we should keep `secret` for this case

redash/query_runner/salesforce.py
```diff
@@ -81,8 +81,8 @@ def configuration_schema(cls):
                     "default": DEFAULT_API_VERSION,
                 },
             },
-            "required": ["username", "password", "token"],
-            "secret": ["password", "token"],
+            "required": ["username", "password"],
+            "secret": ["password"],
```
zhongjiajie, line 85:
```suggestion
            "secret": ["password", "token"],
```

zhongjiajie (approved):

justinclift:
Thanks all.  This is merged now. :smile:

PR #5365, opened 2021-01-21
Initial accessibility improvents and set up Eslint jsx-a11y plugin

rafawendel:

- [X] Feature

## Description
This aims to add the referred plugin, set up the appropriate rules and fix current suggestions.

### After this PR the following standards should be met:

- Links that look like links:  `<Link>` - Renders `<a>`
- Buttons that look like buttons: `<Button>` - Renders Antd `<Button>`
- Links that look like buttons:  `<Link.Button>` - Renders Antd `<Button>`, which is rendered as an `<a>`
- Text Buttons "buttons that look like links": `<PlainButton>` - Renders styleless `<button>`

### Todos (might be moved to another PR)

 - [X] Add tabindex to appropriate components
 - [X] Create custom link to abstract a11y behavior
 - [X] Create a component to act as a plain button
 - [x] Properly set `role` attributes


rafawendel:
@gabrieldutra I need an initial review from you on this. As the desktop navbar is different in mRedash, I don't think it's worth to improve it more right now. However, for it to be fully accessible we'd need to add buttons or keypress handlers to it.

client/app/components/Resizable/index.jsx
```diff
@@ -53,6 +53,8 @@ export default function Resizable({ toggleShortcut, direction, sizeAttribute, ch
     () => (
       <span
         className={`react-resizable-handle react-resizable-handle-${direction}`}
+        role="gridcell"
```
thielium, line 56: What's the motivation of using `gricell` on this `span`? Maybe role `separator` would be better? Resize handles are tricky.

client/app/components/TagsList.less
```diff
@@ -7,7 +7,7 @@
     justify-content: space-between;
     align-items: center;
 
-    label {
+    .tags-list-label {
```
thielium, line 10: 🏆

client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
```diff
@@ -178,11 +180,17 @@ function VisualizationWidgetFooter({ widget, isPublic, onRefresh, onExpand }) {
         {!isPublic && (
           <a
             className="btn btn-sm btn-default hidden-print btn-transparent btn__refresh"
+            role="button"
+            tabIndex={0}
```
thielium, line 184: Again, it doesn't feel like you should have to specify a `tabIndex` here -- is there something above it remove the `tabIndex` or something like that?

client/app/components/queries/SchedulePhrase.jsx
```diff
@@ -51,7 +51,7 @@ export default class SchedulePhrase extends React.Component {
     const content = full ? <Tooltip title={full}>{short}</Tooltip> : short;
 
     return this.props.isLink ? (
-      <a className="schedule-phrase" onClick={this.props.onClick} data-test="EditSchedule">
+      <a className="schedule-phrase" role="link" tabIndex={0} onClick={this.props.onClick} data-test="EditSchedule">
```
thielium, line 54: `a` should already have the role `link` -- I don't believe you need to add this attribute.

client/app/pages/home/components/FavoritesList.jsx
```diff
@@ -27,9 +27,9 @@ export function FavoriteList({ title, resource, itemUrl, emptyState }) {
         {loading && <LoadingOutlinedIcon />}
       </div>
       {!isEmpty(items) && (
-        <div className="list-group">
+        <div className="list-group" role="list">
```
thielium, line 30: Why not just use a `ul` tag?

package-lock.json
```diff
@@ -5001,24 +5001,21 @@
       "dependencies": {
         "axios": {
           "version": "0.19.2",
-          "resolved": "https://registry.npmjs.org/axios/-/axios-0.19.2.tgz",
```
thielium, line 5004: Any ideas why there are all these little changes in the `package-lock.json`?

thielium (commented):
Thanks for doing this! a11y is huge and often underappreciated.

I'd like to understand a little more the reasoning behind some of the roles and things. I feel like some of them could be changed or altered -- I tried not to clutter the PR with redundant comments if I came across the same issue more than a couple times.

client/.eslintrc.js
```diff
@@ -20,6 +21,8 @@ module.exports = {
     // allow debugger during development
     "no-debugger": process.env.NODE_ENV === "production" ? 2 : 0,
     "jsx-a11y/anchor-is-valid": "off",
+    "jsx-a11y/no-autofocus": "off",
+    "jsx-a11y/click-events-have-key-events": "off", // TODO: create a wrapper for key events
```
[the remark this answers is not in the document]
rafawendel: Not yet. It might not be needed depending on how far we are moving into the a11y issue.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -65,36 +65,36 @@ export default function DesktopNavbar() {
   const canCreateAlert = currentUser.hasPermission("list_alerts");
 
   return (
-    <div className="desktop-navbar">
```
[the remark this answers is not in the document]
rafawendel: Not really. Will adopt the suggestion

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -65,36 +65,36 @@ export default function DesktopNavbar() {
...
+    <div className="desktop-navbar" role="navigation">
       <NavbarSection className="desktop-navbar-logo">
         <div>
-          <Link href="./">
+          <Link href="./" role="menuitem" tabIndex={0}>
```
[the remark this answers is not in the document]
rafawendel: It actually is a home button. I was wondering if I should add some semantic styling for hover and focus.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -65,36 +65,36 @@ export default function DesktopNavbar() {
...
         </div>
       </NavbarSection>
 
       <NavbarSection>
         {currentUser.hasPermission("list_dashboards") && (
-          <Menu.Item key="dashboards" className={activeState.dashboards ? "navbar-active-item" : null}>
-            <Link href="dashboards">
-              <DesktopOutlinedIcon />
+          <Menu.Item tabIndex={-1} key="dashboards" className={activeState.dashboards ? "navbar-active-item" : null}>
+            <Link href="dashboards" role="menuitem" tabIndex={0}>
```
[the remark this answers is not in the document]
rafawendel: Forgot to push the changes that removed it

client/app/components/EditInPlace.jsx
```diff
@@ -75,7 +75,7 @@ export default class EditInPlace extends React.Component {
         {this.props.value}
       </span>
     ) : (
-      <a className="clickable" onClick={this.startEditing}>
+      <a className="clickable" role="link" tabIndex={0} onClick={this.startEditing}>
```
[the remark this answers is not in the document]
rafawendel: jsx/a11y requires reinforcement to all link elements, and, of course, requires roles for all elements with click handler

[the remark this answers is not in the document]
rafawendel: Link elements themselves require that, despite using anchor under the hood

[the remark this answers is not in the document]
rafawendel: Check this
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md

client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
rafawendel: Same story: anchors are considered lacking semantic meaning, therefore they require roles. If the role is interactive, a tabindex must be added, particularly if event handlers are present. I agree theoretically it is not necessary, but for my understanding it is sort of a safeguard.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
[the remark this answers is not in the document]
thielium: That sounds awesome, to me!

client/app/components/EditInPlace.jsx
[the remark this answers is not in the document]
thielium: You are so completely correct, here: there's no href, [so it needs the tabindex](https://webaim.org/techniques/keyboard/tabindex#zero:~:text=and%20%3Ca%3E%20with%20no%20href). Thank you!

client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
thielium: I'll defer to our [conversation above](https://github.com/getredash/redash/pull/5365#discussion_r568129610) so we can close out this thread. Rafael is correct that this needs a tabindex.

client/app/components/queries/SchedulePhrase.jsx
thielium: Followup: without the `href` attribute, this has no role. With the an `href`, it is a `link`. 

Source: https://w3c.github.io/html-aria/#el-a

thielium (commented):
I did a time-boxed audit of https://deploy-preview-5365--redash-preview.netlify.app/alerts

## Accessibility Insights for Web
- Homepage: There are still some issues, but they don't appear to touch the parts of the app you've started with.
- Other pages: Still a bunch of warnings and things. I think the plugin is a great start, but it definitely doesn't cover all of our usages.

## Manual Testing

- Tabbing around the main menu is pretty good. The `Create` button, since it has the ant popover menu, isn't accessible, and that's pretty frustrating (but we've already discussed it). Might be out of scope for round 1. 
- [EditInPlace](https://github.com/getredash/redash/pull/5365/files#diff-583285c5a004962b72614c38c9ecac2ef34bd4a93ad846ffa16c88a64593b3cdR71) also has a click handler, and should be focusable (the same way the link form is focusable).
- Got some complaints on the `menuitem` roles -- I think any `menuitem` that isn't a child of a `menu` or `menubar` is going to be upset (and in our case, I think it's happening when links have a `menuitem` role within the `Menu.Item` component). Tabbing around QueryControlDropdown wasn't too bad, though! Could use higher contrast when focused, but it worked.

There's more, but I think you get the idea: I think this is a reasonable first increment once the comments are addressed. Nice!

client/app/components/dashboards/dashboard-widget/Widget.jsx
```diff
@@ -22,9 +23,9 @@ function WidgetDropdownButton({ extraOptions, showDeleteOption, onDelete }) {
   return (
     <div className="widget-menu-regular">
       <Dropdown overlay={WidgetMenu} placement="bottomRight" trigger={["click"]}>
-        <a className="action p-l-15 p-r-15" data-test="WidgetDropdownButton">
+        <Link role="button" className="action p-l-15 p-r-15" data-test="WidgetDropdownButton">
```
gabrieldutra, line 26: `<Button type="link">` instead for these ones?

client/app/components/dynamic-parameters/DynamicButton.jsx
```diff
@@ -42,7 +43,7 @@ function DynamicButton({ options, selectedDynamicValue, onSelect, enabled, stati
 
   return (
     <div ref={containerRef}>
-      <a onClick={e => e.stopPropagation()}>
+      <Link role="presentation" onClick={e => e.stopPropagation()}>
```
gabrieldutra, line 46: It feels to me that the `Link` component should only be used to what are real links (cc: @kravets-levko). Ideally it's used for event tracking and for any link manipulation globally.

@rafawendel sorry if I'm missing something, but why not keep the `<a>` here?

client/app/components/empty-state/EmptyState.jsx
```diff
@@ -200,9 +200,9 @@ function EmptyState({
         </div>
       </div>
       {closable && (
-        <a className="close-button" onClick={onClose}>
+        <Link className="close-button" role="button" aria-label="Close" onClick={onClose}>
```
gabrieldutra, line 203: If possible this should be a `<Button type="link" icon={<CloseOutlinedIcon />} />`

gabrieldutra (commented):
Hey @rafawendel I just quickly went by the code, sorry if I missed any comments, the main thing that seemed to me is that we should try to use a `<Button type="link" />` instead of the `<a role="button">` where possible

client/app/components/dynamic-parameters/DynamicButton.jsx
kravets-levko: Yes, exactly. For `<a>`s w/o `href` we probably should use `<Button type="link"`

client/.eslintrc.js
```diff
@@ -20,9 +20,15 @@ module.exports = {
   rules: {
     // allow debugger during development
     "no-debugger": process.env.NODE_ENV === "production" ? 2 : 0,
-    "jsx-a11y/anchor-is-valid": "off",
+    "jsx-a11y/anchor-is-valid": [
+      "error",
```
thielium, line 24: Do we not need `"components": [ "Link" ]`

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -71,39 +71,35 @@ export default function DesktopNavbar() {
...
 
       <NavbarSection>
         {currentUser.hasPermission("list_dashboards") && (
-          <Menu.Item
-            tabIndex={-1}
-            key="dashboards"
-            role="menuitem"
-            className={activeState.dashboards ? "navbar-active-item" : null}>
-            <Link href="dashboards" tabIndex={0}>
+          <Menu.Item key="dashboards" role="menuitem" className={activeState.dashboards ? "navbar-active-item" : null}>
```
thielium, line 85: Isn't the role of `menuitem` [built into `Menu.Item`](https://github.com/react-component/menu/blob/master/src/MenuItem.tsx#L189)?

client/app/components/EditVisualizationButton/QueryControlDropdown.jsx
```diff
@@ -52,6 +47,8 @@ export default function QueryControlDropdown(props) {
       <Menu.Item>
         <QueryResultsLink
           fileType="tsv"
+          role="menuitem"
```
thielium, line 50: Why does this have a `menuitem` role if it's nested under a `Menu.Item`?

client/app/components/EditVisualizationButton/QueryControlDropdown.jsx
```diff
@@ -75,7 +74,7 @@ export default function QueryControlDropdown(props) {
   );
 
   return (
-    <Dropdown trigger={["click"]} overlay={menu} overlayClassName="query-control-dropdown-overlay">
+    <Dropdown trigger={["click"]} overlay={menu} role="menu" overlayClassName="query-control-dropdown-overlay">
```
thielium, line 77: I think the `Menu` is the only thing that needs the `menu` role. This doesn't need the `menu` role (I think it may [ignore the prop entirely](https://github.com/react-component/dropdown/blob/master/src/Dropdown.tsx#L7-L31), since `Dropdown`s aren't really on the DOM).

client/app/components/Link.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Button, { ButtonProps } from "antd/lib/button";
+
+type AnchorButtonProps = React.AnchorHTMLAttributes<any> & React.ButtonHTMLAttributes<any>;
+
+// TODO: adapt click handlers to keypress events
+function DefaultLinkComponent({ type, role, ...props }: AnchorButtonProps) {
+  return role === "button" ? (
+    <button type="button" style={{ all: "unset" }} {...props} />
+  ) : (
+    <a role={role} type={type} {...props} /> // eslint-disable-line jsx-a11y/anchor-has-content
```
thielium, line 11: Maybe explicitly pull out the `children` attribute so we don't have to disable the linter and make it more explicit what's happening?

client/app/components/Link.tsx
```diff
@@ -0,0 +1,37 @@
...
+  );
+}
+
+Link.Component = DefaultLinkComponent;
+
+function Link({ tabIndex = 0, ...props }: AnchorButtonProps) {
+  return !props.href && !props.role ? (
```
thielium, line 18: If it doesn't have an `href`, either, I'm not sure it's a link? Maybe we should just warn?

client/app/components/TagsList.tsx
```diff
@@ -79,25 +80,24 @@ function TagsList({ tagsUrl, showUnselectAll = false, onUpdate }: TagsListProps)
...
             <CloseOutlinedIcon />
             clear selection
-          </a>
+          </Link>
         )}
       </div>
 
       <div className="tiled">
-        <Menu className="invert-stripe-position" mode="inline" selectedKeys={selectedTags}>
+        <Menu className="invert-stripe-position" mode="inline" selectedKeys={selectedTags} role="menu">
```
thielium, line 91: This should have a default role of 'menu', I believe.

thielium (commented):
Just small stuff! 

I'm on the fence about whether or not to use `<Button type="link">` just because [ant indicates that it's for external links](https://ant.design/components/button/#When-To-Use:~:text=Link%20button:%20used%20for%20external%20links.) (and I don't think that's how we use it), but I don't have strong feelings -- as long as there are no accessibility issues, either way is fine to me. 

I think the `Link` component still provides real value in helping guard against common accessibility pitfalls, as well as giving us a place we can add styling if we need to update it!

Other than the few comments on 'roles', LGTM!

client/.eslintrc.js
rafawendel: Actually not. I was trying to make it work in a way any anchor would be rejected. As the Link is more versatile, it can actually not receive hrefs. I wanted to enforce any Link with click handlers were set to role="button", but that doesn't seem to be possible as well.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
rafawendel: It is! Sorry for this!

gabrieldutra:
> I'm on the fence about whether or not to use `<Button type="link">` just because [ant indicates that it's for external links](https://ant.design/components/button/#When-To-Use:~:text=Link%20button:%20used%20for%20external%20links.) (and I don't think that's how we use it), but I don't have strong feelings -- as long as there are no accessibility issues, either way is fine to me.

Thanks @thielium, I haven't seen that note before. However, it still seems to me that a Button with the link type will match more the "perform action" type than a `Link` with the button role. Basically we already use `Button`s for primary and secondary actions, the link would just be the tertiary one. 

Mind that I'm not sure how the `Button` with the `link` type will behave in terms of accessibility. I just assumed that it would be consistent with any of the primary or secondary actions that already use the same component, thus I favored this option.

client/app/components/EditVisualizationButton/QueryControlDropdown.jsx
```diff
@@ -12,27 +12,30 @@ import FileExcelOutlinedIcon from "@ant-design/icons/FileExcelOutlined";
...
 import QueryResultsLink from "./QueryResultsLink";
+import Link from "../Link";
 
 export default function QueryControlDropdown(props) {
   const menu = (
     <Menu>
       {!props.query.isNew() && (!props.query.is_draft || !props.query.is_archived) && (
         <Menu.Item>
-          <a target="_self" onClick={() => props.openAddToDashboardForm(props.selectedTab)}>
+          <Link target="_self" onClick={() => props.openAddToDashboardForm(props.selectedTab)}>
```
kravets-levko, line 22: Why `target` is needed here?

client/app/components/PermissionsEditorDialog/index.jsx
```diff
@@ -16,6 +16,7 @@ import notification from "@/services/notification";
 import User from "@/services/user";
 
 import "./index.less";
+import Link from "../Link";
```
kravets-levko, line 19: Please use full path here (`@/components/Link`) and move this import upper (to other @-imports)

client/app/components/PermissionsEditorDialog/index.jsx
```diff
@@ -169,10 +170,13 @@ function PermissionsEditorDialog({ dialog, author, context, aclUrl }) {
...
-                      className="fa fa-remove clickable"
-                      onClick={() => removePermission(user.id).then(loadUsersWithPermissions)}
-                    />
+                    <Link
+                      role="button"
+                      tabIndex={-1}
+                      className="clickable"
+                      onClick={() => removePermission(user.id).then(loadUsersWithPermissions)}>
+                      <i className="fa fa-remove" />
+                    </Link>
```
kravets-levko, line 179: Do we have to make this a link actually? Any caveats when it's some other tag?

client/app/components/items-list/components/Sidebar.jsx
```diff
@@ -62,7 +62,7 @@ export function Menu({ items, selected }) {
           <AntdMenu.Item key={item.key} className="m-0">
             <Link href={item.href}>
               {isString(item.icon) && item.icon !== "" && (
-                <span className="btn-favourite m-r-5">
+                <span className="btn-favorite m-r-5">
```
kravets-levko, line 65: 👍

client/app/pages/query-snippets/QuerySnippetsList.jsx
```diff
@@ -23,6 +23,7 @@ import notification from "@/services/notification";
 import routes from "@/services/routes";
 
 import "./QuerySnippetsList.less";
+import Link from "@/components/Link";
```
kravets-levko, line 26: Please move this import upper, to other @-imports

kravets-levko:
My two cents on Link vs. Button discussion. The Link component was originally intended to be a, well, link - those which have a `href` attribute and may lead user somewhere, which you can open in a new tab and so on. There is a variation of it - Link.Button component, which is a link but rendered as a button. Now we need the opposite to it - a button that is rendered as a link (that is, `<a>` without `href`). I see @thielium's point on having a component for it (for CSS/A11y/etc.). So maybe let's create a new component for this purpose, instead of overloading an existing Link component with a new role?

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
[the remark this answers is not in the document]
rafawendel: I will not do it because we already have it in mRedash, and thus should backport it at some point.

client/app/components/CreateSourceDialog.jsx
```diff
@@ -132,15 +132,17 @@ class CreateSourceDialog extends React.Component {
...
+        <Button type="plain" className="p-l-10 p-r-10" onClick={() => this.selectType(item)}>
+          <PreviewCard
+            title={item.name}
+            imageUrl={`${imageFolder}/${item.type}.png`}
+            roundedImage={false}
+            data-test="PreviewItem"
+            data-test-type={item.type}>
+            <i className="fa fa-angle-double-right" />
+          </PreviewCard>
+        </Button>
```
kravets-levko, line 145: `PreviewCard` component has a `onClick` handler itself - I think it's better to use it instead of wrapping it with Button

client/app/components/Button.tsx
```diff
@@ -0,0 +1,21 @@
...
+  type?: AntdButtonProps["type"] | "plain";
+}
+
+function Button({ type, className, ...props }: ButtonProps) {
+  return type === "plain" ? (
+    <button className={classNames("btn-plain", "clickable", className)} type="button" {...props} />
+  ) : (
+    <AntdButton type={type} {...props} />
+  );
+}
```
kravets-levko, line 17: That's a good step forward 🎉  Let's improve it a bit more: `Button` is quite a generic name for this component, and it's a bit abused this way. Also, it creates some confusion with Ant's Button - it's hard to tell which should be used in particular place. Let's just choose less generic name for this component (`PlainButton`, `ButtonLink` or sth like that) and make it render just simply a button that looks like a link. Also, rename `btn-plain` class after component's name (currently `btn-plain` looks like a part of Bootstrap)

kravets-levko:
@rafawendel Please keep an eye on Percy snapshots: this PR shouldn't affect the UI

client/app/assets/less/ant.less
```diff
@@ -94,6 +94,11 @@
   }
 }
 
+// Safeguard for button anchors
```
thielium, line 97: This does make me a little nervous -- global styles are pretty hard to lock down, and having a `button` role doesn't necessarily mean we want to make it a button.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -71,9 +72,9 @@ export default function DesktopNavbar() {
   const canCreateAlert = currentUser.hasPermission("list_alerts");
 
   return (
-    <div className="desktop-navbar">
+    <nav className="desktop-navbar" role="navigation">
```
thielium, line 75: `nav` shouldn't need an additional role.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -165,6 +171,8 @@ export default function DesktopNavbar() {
         <Menu.SubMenu
           key="profile"
           popupClassName="desktop-navbar-submenu"
+          role="menuitem"
```
thielium, line 174: Unless I'm misreading it, `SubMenu` [has this role already](https://github.com/react-component/menu/blob/master/src/SubMenu.tsx#L672).

client/app/components/SelectItemsDialog.jsx
```diff
@@ -30,12 +34,6 @@ function ItemsList({ items, renderItem, onItemClick }) {
   return <List size="small" dataSource={items} renderItem={renderListItem} />;
 }
 
-ItemsList.propTypes = {
```
thielium, line 33: Why did we get rid of the proptypes?

thielium (commented):
I think the blast radius of this PR is growing, and we've got a lot of voices in the room. I agree with @kravets-levko 's concerns, but I don't have too much more to add at this point. If it's good with him, it's good with me.

To be honest, I think the best possible approach, here, would be to split this PR up into small (<100 lines) PRs that tackle a single thing in each PR (menu / menu item roles, adding an accessiblility component, using that accessibility component, etc.). It's up to you, but I'm worried that there's enough here that we'll end up having a lot of long discussions  while the uncontroversial parts stay unmerged.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
rafawendel: It is actually recommended by WAI ARIA, as navs are not supported in older browsers. Of course this is questionable, since these are not under our supported browsers. I'll just remove it

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
rafawendel: Missed that one

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
rafawendel: Good news is that jsx/a11y actually enforces this behavior (try adding a role="link" to an anchor 🤠)

client/app/components/queries/AddToDashboardDialog.jsx
```diff
@@ -52,9 +52,9 @@ function AddToDashboardDialog({ dialog, visualization }) {
         notification.success(
           "Widget added to dashboard",
           <React.Fragment>
-            <Link href={`${dashboard.url}`} onClick={() => notification.close(key)}>
+            <Button type="link" href={`${dashboard.url}`} onClick={() => notification.close(key)}>
```
kravets-levko, line 55: I'm not 100% sure about semantics of this component. On the one hand, it's link, on the other - it's button. But for me it looks that link role is primary (as if you click it - it will bring you to a dashboard page), and onClick action is secondary (just to hide irrelevant UI). So my suggestion is to revert it to `Link`

PR #5375, opened 2021-01-29
Add My Dashboards filter option to the Dashboards list

justint-db:
Feature

## Description
Adds a My Dashboards filter option to the Dashboards list similar to the My Queries option.

An open question I had is whether list_dashboards is the appropriate permissions for this operation. My Queries uses view_queries, but directly translating to view_dashboards didn't seem to work.

I added some python tests, but I'm not very familiar with our testing infrastructure yet. What other tests would make sense here?

REDASH-687

image
image

### DIFF
--- a/client/app/pages/dashboards/components/DashboardListEmptyState.tsx
+++ b/client/app/pages/dashboards/components/DashboardListEmptyState.tsx
@@ -38,7 +38,7 @@ export default function DashboardListEmptyState({ page, searchTerm, selectedTags
       ) : (
         <span>Sorry, we couldn't find anything.</span>
       );
-      return <BigMessage children={my_msg} icon="fa-search" />;
+      return <BigMessage icon="fa-search">{my_msg}</BigMessage>;
     default:
       return (
         <DynamicComponent name="DashboardList.EmptyState">
--- a/client/app/pages/dashboards/components/DashboardListEmptyState.tsx
+++ b/client/app/pages/dashboards/components/DashboardListEmptyState.tsx
@@ -6,6 +6,8 @@ import EmptyState, { EmptyStateHelpMessage } from "@/components/empty-state/Empt
 import DynamicComponent from "@/components/DynamicComponent";
 import Link from "@/components/Link";
 import CreateDashboardDialog from "@/components/dashboards/CreateDashboardDialog";
+import { currentUser } from "@/services/auth";
+import HelpTrigger from "@/components/HelpTrigger";
 
 export interface DashboardListEmptyStateProps {
   page: string;
@@ -24,15 +26,19 @@ export default function DashboardListEmptyState({ page, searchTerm, selectedTags
     case "favorites":
       return <BigMessage message="Mark dashboards as Favorite to list them here." icon="fa-star" />;
     case "my":
-      return (
-        <div className="tiled bg-white p-15">
+      const my_msg = currentUser.hasPermission("create_dashboard") ? (
+        <span>
           <Link.Button type="primary" size="small" onClick={() => CreateDashboardDialog.showModal()}>
-            Create your first dashboard
+            Create your first dashboard!
           </Link.Button>{" "}
-          to populate My Dashboards list. Need help? Check out our{" "}
-          <Link href="https://redash.io/help/user-guide/dashboards/dashboard-editing">dashboard editing documentation</Link>.
-        </div>
+          <HelpTrigger className="f-14" type="DASHBOARDS" showTooltip={false}>
+            Need help?
+          </HelpTrigger>
+        </span>
+      ) : (
+        <span>Sorry, we couldn't find anything.</span>
       );
+      return <BigMessage children={my_msg} icon="fa-search" />;
     default:
       return (
         <DynamicComponent name="DashboardList.EmptyState">
--- a/client/app/pages/dashboards/components/DashboardListEmptyState.tsx
+++ b/client/app/pages/dashboards/components/DashboardListEmptyState.tsx
@@ -4,6 +4,8 @@ import BigMessage from "@/components/BigMessage";
 import NoTaggedObjectsFound from "@/components/NoTaggedObjectsFound";
 import EmptyState, { EmptyStateHelpMessage } from "@/components/empty-state/EmptyState";
 import DynamicComponent from "@/components/DynamicComponent";
+import Link from "@/components/Link";
+import CreateDashboardDialog from "@/components/dashboards/CreateDashboardDialog";
 
 export interface DashboardListEmptyStateProps {
   page: string;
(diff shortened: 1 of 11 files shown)


client/app/pages/dashboards/components/DashboardListEmptyState.tsx
```diff
@@ -21,6 +21,8 @@ export default function DashboardListEmptyState({ page, searchTerm, selectedTags
   switch (page) {
     case "favorites":
       return <BigMessage message="Mark dashboards as Favorite to list them here." icon="fa-star" />;
+    case "my":
+      return <BigMessage message="Create your own dashboards to list them here." icon="fa-line-chart" />;
```
arikfr, line 25: Maybe make it more like the queries one? https://github.com/getredash/redash/blob/04edf16ed42cdf9d005e7a512162cee7ca298c9c/client/app/pages/queries-list/QueriesListEmptyState.jsx#L21

arikfr (commented):
Looks good!

> An open question I had is whether list_dashboards is the appropriate permissions for this operation. My Queries uses view_queries, but directly translating to view_dashboards didn't seem to work.

Yes, the permission needs to exist for you to be able to use it. Adding a new permission requires a migration to update existing group objects. Maybe we should use `create_dashboard` here? Because if you can't create a dashboard, you have no use for this tab :)

client/app/pages/dashboards/components/DashboardListEmptyState.tsx
justint-db: I considered that. The queries page has a nice link to create a new query, but we use a different mechanism to create a new dashboard (popping a modal dialog to name it first), so I couldn't just copy the existing code. I could dig into how to make this work if you think it's important.

justint-db:
Re: create_dashboard. I was thinking it's possible that someone created dashboards and then had create_dashboard revoked, but as long as they have list_dashboards, should still be able to filter the list to their previously created dashboards. So list_dashboards seems reasonable.

client/app/pages/dashboards/components/DashboardListEmptyState.tsx
kravets-levko: @justint-db All you need is this (see click handler; `onMouseUp` is used here because that's menu item, in your case use just regular `onClick`):

https://github.com/getredash/redash/blob/04edf16ed42cdf9d005e7a512162cee7ca298c9c/client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx#L125-L127

justint-db: Thanks! Done.

justint-db: See updated image above.

redash/handlers/dashboards.py
```diff
@@ -113,6 +113,43 @@ def post(self):
         return DashboardSerializer(dashboard).serialize()
 
 
+class MyDashboardsResource(BaseResource):
+    @require_permission("list_dashboards")
```
kravets-levko, line 117:
```suggestion
    @require_permission("create_dashboard")
```

kravets-levko (approved):
JS part looks totally good :+1:

redash/handlers/dashboards.py
justint-db: I'm realizing my code is inconsistent on whether to require list_ or create_, so I need to make some sort of change. I went back and forth on this. Here's my thinking:

Currently with only list_dashboards permission, I can list all the dashboards and see who made them. Obviously, I could manually go through them and find all my own dashboards; that's what users do today. MyDashboardsResource just automates this, returning a list already filtered to my own dashboards. So there's no data leakage here in just requiring list_dashboards.

I think the difference only matters in the scenario where someone once had create_dashboards, so was able to make some dashboards, but this was later revoked. At this point, even though they can't make new dashboards, I think we still want them to be able to filter the list to the old dashboards they made. It would probably feel like a bug if they couldn't.

Does that make sense or am I misunderstanding how the permissions work?

kravets-levko: +1 for this. Also, assume we have a user with `create_dashboard` permission, and then we revoke `create_dashboard ` permission for them. That user should be able to see dashboards they already created (and as @justint-db mentioned) they still can - by walking through the list of all dashboards. So I suggest to allow "My Dashboards" UI and API for users with `list_dashboards`, but need to adjust DashboardListEmptyState component - show 'Create..." button only for users with `create_dashboards` permission. @arikfr WDYT?

justint-db: Updated the PR to match Levko's suggestion.

justint-db:
I updated the formatting of the My Queries/Dashboard pages to look like the other Empty Result pages while keeping the link to easily create a new query/dashboard. This link is only shown if the user has create_* privileges.

client/app/pages/dashboards/components/DashboardListEmptyState.tsx
```diff
@@ -21,6 +25,20 @@ export default function DashboardListEmptyState({ page, searchTerm, selectedTags
...
+            Create your first dashboard!
+          </Link.Button>{" "}
+          <HelpTrigger className="f-14" type="DASHBOARDS" showTooltip={false}>
+            Need help?
+          </HelpTrigger>
+        </span>
+      ) : (
+        <span>Sorry, we couldn't find anything.</span>
+      );
+      return <BigMessage children={my_msg} icon="fa-search" />;
```
kravets-levko, line 41:
```suggestion
      return <BigMessage icon="fa-search">{my_msg}</BigMessage>
```

client/app/pages/queries-list/QueriesListEmptyState.jsx
```diff
@@ -19,15 +21,19 @@ export default function QueriesListEmptyState({ page, searchTerm, selectedTags }
...
-          <Link href="https://redash.io/help/user-guide/querying/writing-queries">query writing documentation</Link>.
-        </div>
+          <HelpTrigger className="f-13" type="QUERIES" showTooltip={false}>
+            Need help?
+          </HelpTrigger>
+        </span>
+      ) : (
+        <span>Sorry, we couldn't find anything.</span>
       );
+      return <BigMessage children={my_msg} icon="fa-search" />;
```
kravets-levko, line 36:
```suggestion
      return <BigMessage icon="fa-search">{my_msg}</BigMessage>
```

kravets-levko (approved):
Looks good, just two minor comments

client/app/pages/queries-list/QueriesListEmptyState.jsx
```diff
@@ -19,15 +21,19 @@ export default function QueriesListEmptyState({ page, searchTerm, selectedTags }
     case "archive":
       return <BigMessage message="Archived queries will be listed here." icon="fa-archive" />;
     case "my":
-      return (
-        <div className="tiled bg-white p-15">
+      const my_msg = currentUser.hasPermission("create_query") ? (
```
thielium, line 24: Maybe we can extract this to a component? Feels like there's some duplication. Not sure if there will be enough re-use to justify it, but I wanted to report my first impression.

kravets-levko: That's definitely a good idea, but IMHO it shouldn't block this PR - we can do this refactoring separately

client/app/pages/queries-list/QueriesListEmptyState.jsx
justint-db: Done.

client/app/pages/dashboards/components/DashboardListEmptyState.tsx
justint-db: Done.

redash/handlers/dashboards.py
arikfr: SGTM

arikfr (approved):
:shipit:

PR #5384, opened 2021-02-02
handling for MongoDB users with collection-specific privileges

manuelgu:

- [X] Bug Fix

## Description
Allow errors during finding fields of MongoDB data source collections. Redash tries to find collection fields to display to preview a list of columns.

There are scenarios in which MongoDB users are not allowed to read all collections in a single database. This theoratically works perfectly fine with Redash, however said check to find all columns of a collection previously broke this.

The introduced change allows errors to happen when fetching collections fields of a collection (more specific: get newest and oldest document in collection).

### DIFF
--- a/redash/query_runner/mongodb.py
+++ b/redash/query_runner/mongodb.py
@@ -221,15 +221,21 @@ class MongoDB(BaseQueryRunner):
         # document written.
         collection_is_a_view = self._is_collection_a_view(db, collection_name)
         documents_sample = []
-        if collection_is_a_view:
-            for d in db[collection_name].find().limit(2):
-                documents_sample.append(d)
-        else:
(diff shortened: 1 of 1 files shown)


guidopetri:
@manuelgu , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
Looking at the current MongoDB query runner code, it seems like some of the code has already been turned into a `try` statement:

https://github.com/getredash/redash/blob/9751678c446efa985d8b89835e068821a83f216b/redash/query_runner/mongodb.py#L227-L236

Doesn't look like it should be too hard to update things though. :smile:

PR #5404, opened 2021-02-24
Add 'user' field at Impala datasource, and use it at connection.cursor

TaylorHere:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
At our use case, we did't use any auth mechanism, just a user pram at cursor function


redash/query_runner/impala_ds.py
```diff
@@ -95,10 +96,12 @@ def run_query(self, query, user):
...
-            connection = connect(**self.configuration.to_dict())
-
-            cursor = connection.cursor()
-
+            config = self.configuration.to_dict()
+            connection = connect(**config)
+            if 'user' in config:
+                cursor = connection.cursor(user=config['user'])
+            else:
+                cursor = connection.cursor()
```
susodapop, line 104: What happens if there are both `ldap_user` and `user` fields? Does this fail gracefully?

redash/query_runner/impala_ds.py
```diff
@@ -56,6 +56,7 @@ def configuration_schema(cls):
                 "ldap_user": {"type": "string"},
                 "ldap_password": {"type": "string"},
                 "timeout": {"type": "number"},
+                "user": {"type": "string"},
```
susodapop, line 59: I wonder how we can communicate to users when to use `user` versus `ldap_user`?

susodapop (commented):
Thanks for your contribution! A couple questions below.

redash/query_runner/impala_ds.py
TaylorHere: I thought that 'ldap_user' and 'user' should work separately, maybe I should add this condition test

redash/query_runner/impala_ds.py
TaylorHere: @susodapop  hi there, I update the code~

Issue #5380, opened 2021-01-30
Add Trino query runner (f.k.a PrestoSQL)


subbareddydagumati:
Add Trino query runner using trino-python-client driver

https://github.com/trinodb/trino-python-client
https://github.com/trinodb/trino

PR #5405, opened 2021-02-24
Add Trino query runner

song9446:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Trino query runner workaround code. I've used it for a while, and decided to share with other people.

Most of the codes are shamelessly copied from the prestodb query runner code.

The authentication part has not been properly tested and may not work as expected.

#5380


arikfr:
Thank you, @song9446 ! We're about to merge the implementation from #5411 so going to close this one. It's not entirely clear to me if the behavior implemented in `extract_columns_from_type_string` required there, but if it is - maybe you can implement it on top of the implementation in #5411?

song9446:
@arikfr The `extract_columns_from_type_string` function parses the nested schema of the table into a list of columns which have a flattened path instead of a name, making the sub-columns visible in the schema explorer on the redash query page.

I strongly need this functionality because my dataset has a complex nested schema, but this behavior is not well defined in  redash side, so I'll wait for the official nested field explorer support.

PR #5408, opened 2021-02-25
Initial a11y improvements

rafawendel:

- [X] Feature

## Description
This PR is a split on #5365 due to its staleness.
It aims to fix issues on interactivity, add semantic elements and improve keyboard accessibility.

### Note for reviewers
Please do not extend the scope of the PR. Future a11y improvements will follow it.
Leave all comments related to anchor, link and button components [this PR](5409).

### DIFF
--- a/client/app/assets/less/inc/table.less
+++ b/client/app/assets/less/inc/table.less
@@ -103,7 +103,7 @@
     padding-top: 5px !important;
   }
 
-  .btn-favourite,
+  .btn-favorite,
   .btn-archive {
     font-size: 15px;
   }
@@ -114,7 +114,7 @@
   line-height: 1.7 !important;
 }
 
-.btn-favourite {
+.btn-favorite {
   color: #d4d4d4;
   transition: all 0.25s ease-in-out;
 
--- a/client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
+++ b/client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
@@ -129,10 +129,7 @@ export default function DesktopNavbar() {
             )}
             {canCreateDashboard && (
               <Menu.Item key="new-dashboard">
-                <a
-                  data-test="CreateDashboardMenuItem"
-                  role="button"
-                  onMouseUp={() => CreateDashboardDialog.showModal()}>
+                <a data-test="CreateDashboardMenuItem" onMouseUp={() => CreateDashboardDialog.showModal()}>
                   New Dashboard
                 </a>
               </Menu.Item>
@@ -185,7 +182,7 @@ export default function DesktopNavbar() {
           )}
           <Menu.Divider />
           <Menu.Item key="logout">
-            <a data-test="LogOutButton" role="button" onClick={() => Auth.logout()}>
+            <a data-test="LogOutButton" onClick={() => Auth.logout()}>
               Log out
             </a>
           </Menu.Item>
--- a/client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
+++ b/client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
@@ -70,12 +70,11 @@ export default function DesktopNavbar() {
   const canCreateDashboard = currentUser.hasPermission("create_dashboard");
   const canCreateAlert = currentUser.hasPermission("list_alerts");
 
-  // TODO [A11Y]: allow navigation via arrow keys and bind click events to enter key (or use b)
   return (
     <div className="desktop-navbar" role="navigation">
       <NavbarSection className="desktop-navbar-logo">
-        <div>
-          <Link href="./">
+        <div role="menuitem">
+          <Link href="./" tabIndex={0}>
             <img src={logoUrl} alt="Redash" />
           </Link>
         </div>
@@ -84,35 +83,27 @@ export default function DesktopNavbar() {
       <NavbarSection>
         {currentUser.hasPermission("list_dashboards") && (
           <Menu.Item
-            role="menuitem"
-            tabIndex={0}
+            tabIndex={-1}
             key="dashboards"
+            role="menuitem"
             className={activeState.dashboards ? "navbar-active-item" : null}>
-            <Link href="dashboards" tabIndex={-1}>
+            <Link href="dashboards" tabIndex={0}>
               <DesktopOutlinedIcon alt="Dashboard navigation button" />
               <span className="desktop-navbar-label">Dashboards</span>
             </Link>
           </Menu.Item>
         )}
         {currentUser.hasPermission("view_query") && (
-          <Menu.Item
-            role="menuitem"
-            tabIndex={0}
-            key="queries"
-            className={activeState.queries ? "navbar-active-item" : null}>
-            <Link href="queries" tabIndex={-1}>
+          <Menu.Item key="queries" role="menuitem" className={activeState.queries ? "navbar-active-item" : null}>
+            <Link href="queries" tabIndex={0}>
               <CodeOutlinedIcon alt="Queries navigation button" />
               <span className="desktop-navbar-label">Queries</span>
             </Link>
           </Menu.Item>
         )}
         {currentUser.hasPermission("list_alerts") && (
-          <Menu.Item
-            role="menuitem"
-            tabIndex={0}
-            key="alerts"
-            className={activeState.alerts ? "navbar-active-item" : null}>
-            <Link href="alerts" tabIndex={-1}>
+          <Menu.Item key="alerts" role="menuitem" className={activeState.alerts ? "navbar-active-item" : null}>
+            <Link href="alerts" tabIndex={0}>
               <AlertOutlinedIcon alt="Alerts navigation button" />
               <span className="desktop-navbar-label">Alerts</span>
             </Link>
@@ -123,11 +114,11 @@ export default function DesktopNavbar() {
       <NavbarSection className="desktop-navbar-spacer">
         {(canCreateQuery || canCreateDashboard || canCreateAlert) && (
           <Menu.SubMenu
-            role="menuitem"
-            tabIndex={0}
             key="create"
             popupClassName="desktop-navbar-submenu"
             data-test="CreateButton"
+            role="menuitem"
+            tabIndex={0}
             title={
               <React.Fragment>
                 <PlusOutlinedIcon />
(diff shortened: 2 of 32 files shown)


client/app/components/PermissionsEditorDialog/index.jsx
```diff
@@ -170,8 +170,10 @@ function PermissionsEditorDialog({ dialog, author, context, aclUrl }) {
                 ) : (
                   <Tooltip title="Remove user permissions">
                     <i
-                      className="fa fa-remove clickable"
+                      role="button"
```
rafawendel, line 173: Improved in #5409

client/app/assets/less/inc/table.less
```diff
@@ -103,7 +103,7 @@
     padding-top: 5px !important;
   }
 
-  .btn-favourite,
+  .btn-favorite,
```
thielium, line 106: Does this come from ant, or is this just us? It feels like it could potentially break stuff, not sure you want to do it in this PR.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -84,23 +84,23 @@ export default function DesktopNavbar() {
...
+              <DesktopOutlinedIcon alt="Dashboard navigation button" />
               <span className="desktop-navbar-label">Dashboards</span>
             </Link>
           </Menu.Item>
         )}
         {currentUser.hasPermission("view_query") && (
           <Menu.Item key="queries" className={activeState.queries ? "navbar-active-item" : null}>
             <Link href="queries">
-              <CodeOutlinedIcon />
+              <CodeOutlinedIcon alt="Queries navigation button" />
```
thielium, line 95: Are these giving you warnings when you use the accessibility tool? `alt` is just an attribute for `img` elements (even though some browsers play nicely with it). In general, you want `aria-label`. `ant` appears to use the `img` role, not an `img` tag, for its icons. 


(This comment applies to all `alt` icon attributes)

thielium: (Reference: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img#svg_and_roleimg)

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -113,6 +113,7 @@ export default function DesktopNavbar() {
             key="create"
             popupClassName="desktop-navbar-submenu"
             data-test="CreateButton"
+            tabIndex={0}
```
thielium, line 116: When I tested, this element appears to have a `tabIndex`, but no keyboard accessibility (if I hit `enter` or `space`, nothing happens). It's also weird if you remove the `tabIndex` -- maybe create a ticket tracking the issue? I think it may be out of scope for this first pass, but we should definitely address it.

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
```diff
@@ -165,6 +166,7 @@ export default function DesktopNavbar() {
         <Menu.SubMenu
```
thielium, line 166: This item also does not have keyboard accessibility, but it has a `tabIndex`. We need to track this issue, too.

client/app/components/EditVisualizationButton/QueryControlDropdown.jsx
```diff
@@ -67,7 +67,7 @@ export default function QueryControlDropdown(props) {
   );
 
   return (
-    <Dropdown trigger={["click"]} overlay={menu} overlayClassName="query-control-dropdown-overlay">
+    <Dropdown trigger={["click"]} overlay={menu} role="menu" overlayClassName="query-control-dropdown-overlay">
```
thielium, line 70: Can you confirm this is necessary? The DOM reads like the `menu` attribute is going to be provided to the overlay without this.

client/app/components/PermissionsEditorDialog/index.jsx
```diff
@@ -170,8 +170,10 @@ function PermissionsEditorDialog({ dialog, author, context, aclUrl }) {
...
+                      tabIndex={-1}
```
thielium, line 174: Why is this `tabIndex` -1? Should it be `0`?

client/app/components/Resizable/index.jsx
```diff
@@ -51,8 +51,11 @@ export default function Resizable({ toggleShortcut, direction, sizeAttribute, ch
 
   const resizeHandle = useMemo(
     () => (
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
       <span
         className={`react-resizable-handle react-resizable-handle-${direction}`}
+        role="separator"
+        tabIndex={-1} // TODO: add key controls
```
thielium, line 58: Why do you need to specify a `tabIndex` at all, here?

client/app/components/dynamic-parameters/DynamicButton.jsx
```diff
@@ -42,7 +42,7 @@ function DynamicButton({ options, selectedDynamicValue, onSelect, enabled, stati
 
   return (
     <div ref={containerRef}>
-      <a onClick={e => e.stopPropagation()}>
+      <a role="presentation" onClick={e => e.stopPropagation()}>
```
thielium, line 45: Why do we apply `presentation` in this specific case? It feels like this could potentially just be left as is. I'm worried that `presentation` will remove the built in semantics for the `Dropdown.button`.

client/app/components/queries/ApiKeyDialog/index.jsx
```diff
@@ -56,11 +56,11 @@ function ApiKeyDialog({ dialog, ...props }) {
 
         <h5>Example API Calls:</h5>
         <div className="m-b-10">
-          <label>Results in CSV format:</label>
+          <span>Results in CSV format:</span>
```
thielium, line 59: Maybe revert this and address it in a followup? It feels like `label` is appropriate, and that `CodeBlock` should use a `textarea` tag instead of a `code` tag. That way you can preserve the usefulness of the `label` element.

client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -45,7 +45,7 @@ function SchemaItem({ item, expanded, onToggle, onSelect, ...props }) {
 
   return (
     <div {...props}>
-      <div className="table-name" onClick={onToggle}>
+      <div className="table-name" role="presentation" onClick={onToggle}>
```
thielium, line 48: Do we want to prevent screen readers from getting into this part? Why is that?

client/app/pages/home/components/FavoritesList.jsx
```diff
@@ -27,17 +27,17 @@ export function FavoriteList({ title, resource, itemUrl, emptyState }) {
         {loading && <LoadingOutlinedIcon />}
       </div>
       {!isEmpty(items) && (
-        <div className="list-group">
+        <ul className="list-group">
           {items.map(item => (
-            <Link key={itemUrl(item)} className="list-group-item" href={itemUrl(item)}>
-              <span className="btn-favourite m-r-5">
+            <Link key={itemUrl(item)} role="listitem" className="list-group-item" href={itemUrl(item)}>
```
thielium, line 32: `ul` tags need to have `li` tags as children

client/app/pages/query-snippets/QuerySnippetsList.jsx
```diff
@@ -35,7 +35,10 @@ class QuerySnippetsList extends React.Component {
     Columns.custom.sortable(
       (text, querySnippet) => (
         <div>
-          <a className="table-main-title clickable" onClick={() => this.showSnippetDialog(querySnippet)}>
+          <a
+            className="table-main-title clickable"
+            role="presentation"
```
thielium, line 40: Why are we making this a `presentation` role?

thielium (approved):
Thank you for breaking this PR up!!

client/app/assets/less/inc/table.less
rafawendel: It's ours!

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
rafawendel: This is actually caused by Ant. The only way to trigger interactivity here is by triggering `onMouseOver` event. Maybe we can do it programmatically in a future PR. WDYT?

client/app/components/EditVisualizationButton/QueryControlDropdown.jsx
rafawendel: It is indeed!

client/app/components/dynamic-parameters/DynamicButton.jsx
rafawendel: Presentation is a role reserved for elements that should be ignored by screen readers. As the role of this one is to stop bubbling, I think it's the case. Also, Eslint requires that all interactive elements have a role. Check [this](https://www.w3.org/TR/wai-aria-practices-1.1/#presentation_role).

client/app/components/ApplicationArea/ApplicationLayout/DesktopNavbar.jsx
thielium: As long as there's a ticket for it, I'm fine with it.

client/app/components/dynamic-parameters/DynamicButton.jsx
thielium: When we apply `presentation`, it signals the screen reader to ignore all children, too. Are we certain we want to remove the accessibility of its child components?

thielium: Hmm... I changed my mind. [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_presentation_role#:~:text=The%20presentation%20role%20is%20used%20to%20remove%20semantic%20meaning%20from%20an%20element%20and%20any%20of%20its%20related%20child%20elements)'s documentation conflicts with the [w3 aria practices doc](https://www.w3.org/TR/wai-aria-practices-1.1/#presentation_role).

thielium: So nevermind :) My reference was wrong.

rafawendel: That caught me up as well, had to read some stuff

client/app/components/PermissionsEditorDialog/index.jsx
rafawendel: Will track it to the other PR, the proper way would be to wrap this in a styleless button

thielium: If we're giving it a role of `button` and a click handler, we should at least give it `tabIndex={0}` so a user can get to it.

thielium (commented):
Just a couple more outstanding items, but this is really close. Thanks for sticking with this!

client/app/components/CodeBlock.jsx
```diff
@@ -66,9 +66,9 @@ export default class CodeBlock extends React.Component {
 
     return (
       <div className="code-block">
-        <code {...props} ref={this.ref}>
+        <textarea {...props} ref={this.ref}>
```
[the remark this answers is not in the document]
rafawendel: I've committed a mistake: it makes no sense to replace the code with a textarea, for it is not editable

thielium (approved):
If there are no visual regressions, go for it!!

🚀

PR #5418, opened 2021-03-09
Refactor of Link component

rafawendel:

- [X] Feature

## Description
This PR is a split on #5409
It is the first step to fix issues on standardization, interactivity and improve keyboard accessibility.

All anchors `<a>` need to be replaced by proper semantic components other than those in `viz-lib`.


client/app/components/Link.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
```
thielium, line 1: I know this file was already structured this way, but this feels pretty confusing, right? Not for now, but maybe later we can refactor this? Or am I missing something?

client/app/components/QueryLink.jsx
```diff
@@ -21,10 +21,12 @@ function QueryLink({ query, visualization, readOnly }) {
     return query.getUrl(false, hash);
   };
 
+  const QueryLinkWrapper = props => (readOnly ? <span {...props} /> : <Link href={getUrl()} {...props} />);
```
thielium, line 24: It feels like the styles could diverge a little / the spacing might get off (just because one's a styled component and one's a plain ol' `span`. If that doesn't happen, then this is fine 👍 .

client/app/components/Resizable/index.jsx
```diff
@@ -51,7 +51,7 @@ export default function Resizable({ toggleShortcut, direction, sizeAttribute, ch
 
   const resizeHandle = useMemo(
     () => (
-      // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
+      // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
```
thielium, line 54: For what it's worth, I'm pretty sure we add keyboard events at a higher level, so we shouldn't feel bad about this.

thielium (approved):
Just small stuff. If there are no visual regressions, I'm comfortable with the code 👍 !

client/app/components/Link.tsx
rafawendel: It does, but we cannot do it for now because this is a compatibility feature for Managed Redash

client/app/components/Resizable/index.jsx
rafawendel: The linter rules will bug us about it. Shall I plain remove them?

client/app/components/QueryLink.jsx
rafawendel: It doesn't in this case and _shouldn't_, but I agree it can become a problem. Any ideas on how to prevent problems? This is, however, the proper approach.

thielium: I think as long as `Link` remains an unstyled component, there's no problem.

client/app/components/Resizable/index.jsx
thielium: I think you can leave it for now, LGTM.

PR #5420, opened 2021-03-09
Improve css and add focus styles

rafawendel:

- [X] Feature

## Description
This PR is a split on #5409
It is the third step to fix issues on standardization, interactivity and improve keyboard accessibility.

Initially, some elements that would not have focus styles had some added and some tag restricted styles (particularly anchor) were replaced by more general approaches. Besides that, some syntax was improved.

## Summary of changes
*Other than `:focus` and `:active`

| Before | After |
| --- | --- |
| | 20f678dce2605642dcae8dd8971e8e0f9d01ab52 - Particularly for focus feedback |
| No change on hover | Kapture 2021-03-10 at 13 26 01 |
| (hover -> click) [focus not available] | 157243e25eb3228e7db9c57e7237437ecb27fb51 (hover -> click, focus -> click) |
| Kapture 2021-03-10 at 13 32 03 | Kapture 2021-03-10 at 13 30 59 |
| (no feedback) | 52e7ba61ce3dde26c650cce5bdebc38588643b10 (hover -> click, focus) |
| Kapture 2021-03-10 at 14 03 50 | Kapture 2021-03-10 at 14 03 24 |


client/app/assets/less/inc/edit-in-place.less
```diff
@@ -1,26 +1,23 @@
...
-.edit-in-place span.editable {
-  display: inline-block;
-  cursor: pointer;
-}
-
-.edit-in-place span.editable:hover {
-  background: @redash-yellow;
-  border-radius: @redash-radius;
-}
+  .editable {
```
thielium, line 9: 👍

client/app/assets/less/inc/table.less
```diff
@@ -118,14 +118,19 @@
...
   &:hover,
   &:focus {
     color: @yellow-darker;
     cursor: pointer;
-  }
 
-  .fa-star {
-    color: @yellow-darker;
+    .fa-star {
+      filter: saturate(75%);
```
thielium, line 131: Is there a screenshot or link for what this change does?

client/app/components/SelectItemsDialog.less
```diff
@@ -0,0 +1,9 @@
+.ant-list-item {
```
thielium, line 1: This is unscoped, so it'll interact with all `ant-list-item` components. Is that intentional?

client/app/components/TagsList.less
```diff
@@ -13,7 +13,8 @@
       margin: 0;
     }
 
-    a {
+    a,
+    button {
```
thielium, line 17: This is a pretty broad change -- it might be right, but I want to confirm: in this context, will there ever be `button`'s that we don't want to match this style (e.g. `button`'s that aren't styled like links, such as antd's `Button`)? If so, just switch `button` to some classname and we're all good.

thielium: (This also happens [here](https://github.com/getredash/redash/pull/5420/files?w=1#diff-05d6220f9633063267ad9eb66926b5847c1461e1da03108f3ff54f4610dac030R255-R256))

client/app/pages/dashboards/components/DashboardHeader.less
```diff
@@ -50,13 +50,16 @@
     margin: 3px 5px 0 5px;
   }
 
-  .tags-control a {
+  .tags-control > .label-tag {
```
thielium, line 53: 👍  for good selectors!

client/app/pages/query-snippets/QuerySnippetsList.less
```diff
@@ -8,3 +8,16 @@
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
 }
+
+.ant-table-row {
+  height: 1px;
```
thielium, line 13: That's a surprising one. Why is that? Also, this looks like an unqualified selector -- will these three classnames impact all `ant-table-cell`s, `ant-table-row`s, etc.?

thielium (commented):
Code-wise, just a couple comments. Screenshots of what changed would be super helpful, otherwise it's just relying on Percy (which is still pretty good).

client/app/assets/less/inc/table.less
rafawendel: Now there is 😉

client/app/components/TagsList.less
rafawendel: It shouldn't become a problem, but I agree we're safer having a class.

client/app/components/SelectItemsDialog.less
rafawendel: Probably it's a behavior we can make use of elsewhere, but I'll scope it for now.

thielium (commented):
Code LGTM! 

You may need to rebase -- it looks like the plainbutton commits are being duplicated in your history.

thielium (approved):

PR #5424, opened 2021-03-12
Improve icon a11y

rafawendel:

- [X] Feature

## Description
This aims to add accessibility to icons by:

1. Hiding decorative icons via `aria-hidden`
2. Adding visually hidden "screen-reader only" helper text where applicable
3. Adding `aria-label`, `aria-labelledby` and `aria-describedby` where applicable

Ant Icons already have somewhat comprehensible labels :tada:, so only particularly relevant ones will be a11y-zed

#5418, #5419, #5420, #5421

## Screenshots
This PR is not _supposed to_ result in visual changes or regressions, but I used the opportunity to fix minor ones.
(Cc. @wenbow)

| Before | After |
| --- | --- |
| image | image |
| image | image |

## Note for reviewers
The commit order is intended to be comprehensive to review.

### DIFF
--- a/client/app/assets/less/inc/generics.less
+++ b/client/app/assets/less/inc/generics.less
@@ -2,163 +2,229 @@
     Generate Margin Classes (0px - 25px)
     margin, margin-top, margin-bottom, margin-left, margin-right
 -----------------------------------------------------------*/
-.margin (@label, @size: 1, @key:1) when (@size =< 30){
-    .m-@{key} {
-        margin: @size !important;
-    }
-    
-    .m-t-@{key} {
-        margin-top: @size !important;
-    }
-    
-    .m-b-@{key} {
-        margin-bottom: @size !important;
-    }
-    
-    .m-l-@{key} {
-        margin-left: @size !important;
-    }
-    
-    .m-r-@{key} {
-        margin-right: @size !important;
-    }
-    
-    .margin(@label - 5; @size + 5; @key + 5);
+.margin (@label, @size: 1, @key:1) when (@size =< 30) {
+  .m-@{key} {
+    margin: @size !important;
+  }
+
+  .m-t-@{key} {
+    margin-top: @size !important;
+  }
+
+  .m-b-@{key} {
+    margin-bottom: @size !important;
+  }
+
+  .m-l-@{key} {
+    margin-left: @size !important;
+  }
+
+  .m-r-@{key} {
+    margin-right: @size !important;
+  }
+
+  .margin(@label - 5; @size + 5; @key + 5);
 }
 
 .margin(25, 0px, 0);
 
-.m-2{
-  margin:2px;
+.m-2 {
+  margin: 2px;
 }
 
 /* --------------------------------------------------------
     Generate Padding Classes (0px - 25px)
     padding, padding-top, padding-bottom, padding-left, padding-right
 -----------------------------------------------------------*/
-.padding (@label, @size: 1, @key:1) when (@size =< 30){
-    .p-@{key} {
-        padding: @size !important;
-    }
-    
-    .p-t-@{key} {
-        padding-top: @size !important;
-    }
-    
-    .p-b-@{key} {
-        padding-bottom: @size !important;
-    }
-    
-    .p-l-@{key} {
-        padding-left: @size !important;
-    }
-    
-    .p-r-@{key} {
-        padding-right: @size !important; 
-    }
-    
-    .padding(@label - 5; @size + 5; @key + 5);
-} 
+.padding (@label, @size: 1, @key:1) when (@size =< 30) {
+  .p-@{key} {
+    padding: @size !important;
+  }
+
+  .p-t-@{key} {
+    padding-top: @size !important;
+  }
+
+  .p-b-@{key} {
+    padding-bottom: @size !important;
+  }
+
+  .p-l-@{key} {
+    padding-left: @size !important;
+  }
+
+  .p-r-@{key} {
+    padding-right: @size !important;
+  }
+
+  .padding(@label - 5; @size + 5; @key + 5);
+}
 
 .padding(25, 0px, 0);
-
 
 /* --------------------------------------------------------
     Generate Font-Size Classes (8px - 20px)
 -----------------------------------------------------------*/
-.font-size (@label, @size: 8, @key:10) when (@size =< 20){
-    .f-@{key} {
-        font-size: @size !important;
-    }
-    
-    .font-size(@label - 1; @size + 1; @key + 1);
-} 
+.font-size (@label, @size: 8, @key:10) when (@size =< 20) {
+  .f-@{key} {
+    font-size: @size !important;
+  }
+
+  .font-size(@label - 1; @size + 1; @key + 1);
+}
 
 .font-size(20, 8px, 8);
 
-.f-inherit { font-size: inherit !important; }
-
+.f-inherit {
+  font-size: inherit !important;
+}
 
 /* --------------------------------------------------------
     Font Weight
 -----------------------------------------------------------*/
-.f-300 { font-weight: 300 !important; }
-.f-400 { font-weight: 400 !important; }
-.f-500 { font-weight: 500 !important; }
-.f-700 { font-weight: 700 !important; }
-
+.f-300 {
+  font-weight: 300 !important;
+}
+.f-400 {
+  font-weight: 400 !important;
+}
+.f-500 {
+  font-weight: 500 !important;
+}
+.f-700 {
+  font-weight: 700 !important;
+}
 
 /* --------------------------------------------------------
     Position
 -----------------------------------------------------------*/
-.p-relative { position: relative !important; }
-.p-absolute { position: absolute !important; }
-.p-fixed { position: fixed !important; }
-.p-static { position: static !important; }
-
+.p-relative {
+  position: relative !important;
+}
+.p-absolute {
+  position: absolute !important;
+}
+.p-fixed {
+  position: fixed !important;
+}
+.p-static {
+  position: static !important;
+}
 
 /* --------------------------------------------------------
     Overflow
 -----------------------------------------------------------*/
-.o-hidden { overflow: hidden !important; }
-.o-visible { overflow: visible !important; }
-.o-auto { overflow: auto !important; }
-
+.o-hidden {
+  overflow: hidden !important;
+}
+.o-visible {
+  overflow: visible !important;
+}
+.o-auto {
+  overflow: auto !important;
+}
 
 /* --------------------------------------------------------
     Display
 -----------------------------------------------------------*/
-.di-block { display: inline-block !important; } 
-.d-block { display: block; }
+.di-block {
+  display: inline-block !important;
+}
+.d-block {
+  display: block;
+}
 
 /* --------------------------------------------------------
     Background Colors and Colors
 -----------------------------------------------------------*/
-@array: c-white bg-white @white, c-ace bg-ace @ace, c-black bg-black @black, c-brown bg-brown @brown, c-pink bg-pink @pink, c-red bg-red @red, c-blue bg-blue @blue, c-purple bg-purple @purple, c-deeppurple bg-deeppurple @deeppurple, c-lightblue bg-lightblue @lightblue, c-cyan bg-cyan @cyan, c-teal bg-teal @teal, c-green bg-green @green, c-lightgreen bg-lightgreen @lightgreen, c-lime bg-lime @lime, c-yellow bg-yellow @yellow, c-amber bg-amber @amber, c-orange bg-orange @orange, c-deeporange bg-deeporange @deeporange, c-gray bg-gray @gray, c-bluegray bg-bluegray @bluegray, c-indigo bg-indigo @indigo;
+@array: c-white bg-white @white, c-ace bg-ace @ace, c-black bg-black @black, c-brown bg-brown @brown,
+  c-pink bg-pink @pink, c-red bg-red @red, c-blue bg-blue @blue, c-purple bg-purple @purple,
+  c-deeppurple bg-deeppurple @deeppurple, c-lightblue bg-lightblue @lightblue, c-cyan bg-cyan @cyan,
+  c-teal bg-teal @teal, c-green bg-green @green, c-lightgreen bg-lightgreen @lightgreen, c-lime bg-lime @lime,
+  c-yellow bg-yellow @yellow, c-amber bg-amber @amber, c-orange bg-orange @orange,
+  c-deeporange bg-deeporange @deeporange, c-gray bg-gray @gray, c-bluegray bg-bluegray @bluegray,
+  c-indigo bg-indigo @indigo;
 
-.for(@array); .-each(@value) {
-    @name:  extract(@value, 1);
-    @name2:  extract(@value, 2);
-    @color: extract(@value, 3); 
-    &.@{name2} {
-        background-color: @color !important;
-    }
-    
-    &.@{name} {
-        color: @color !important;
-    }
+.for(@array);
+.-each(@value) {
+  @name: extract(@value, 1);
+  @name2: extract(@value, 2);
+  @color: extract(@value, 3);
+  &.@{name2} {
+    background-color: @color !important;
+  }
+
+  &.@{name} {
+    color: @color !important;
+  }
 }
-
 
 /* --------------------------------------------------------
     Background Colors
 -----------------------------------------------------------*/
-.bg-brand { background-color: @brand-bg; }
-.bg-black-trp { background-color: rgba(0,0,0,0.12) !important; }
-
-
+.bg-brand {
+  background-color: @brand-bg;
+}
+.bg-black-trp {
+  background-color: rgba(0, 0, 0, 0.12) !important;
+}
 
 /* --------------------------------------------------------
     Borders
 -----------------------------------------------------------*/
-.b-0 { border: 0 !important; }
-
+.b-0 {
+  border: 0 !important;
+}
 
 /* --------------------------------------------------------
     Width
 -----------------------------------------------------------*/
-.w-100 { width: 100% !important; }
-.w-50 { width: 50% !important; }
-.w-25 { width: 25% !important; }
-
+.w-100 {
+  width: 100% !important;
+}
+.w-50 {
+  width: 50% !important;
+}
+.w-25 {
+  width: 25% !important;
+}
 
 /* --------------------------------------------------------
     Border Radius
 -----------------------------------------------------------*/
-.brd-2 { border-radius: 2px; }
-
+.brd-2 {
+  border-radius: 2px;
+}
 
 /* --------------------------------------------------------
     Alignment
 -----------------------------------------------------------*/
-.va-top { vertical-align: top; }
+.va-top {
+  vertical-align: top;
+}
+
+/* --------------------------------------------------------
+    Screen readers
+-----------------------------------------------------------*/
+.sr-only {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
+}
+
+.not-sr-only {
+  position: static;
+  width: auto;
+  height: auto;
+  padding: 0;
+  margin: 0;
+  overflow: visible;
+  clip: auto;
+  white-space: normal;
+}
--- a/client/app/assets/less/inc/generics.less
+++ b/client/app/assets/less/inc/generics.less
@@ -217,14 +217,3 @@
   white-space: nowrap;
   border-width: 0;
 }
-
-.not-sr-only {
-  position: static;
-  width: auto;
-  height: auto;
-  padding: 0;
-  margin: 0;
-  overflow: visible;
-  clip: auto;
-  white-space: normal;
-}
--- a/client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
+++ b/client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
@@ -15,7 +15,7 @@ export default function VersionInfo() {
           {/* eslint-disable react/jsx-no-target-blank */}
           <Link href="https://version.redash.io/" className="update-available" target="_blank" rel="noopener">
             Update Available <i className="fa fa-external-link m-l-5" aria-hidden="true" />
-            <span className="sr-only">{"(opens in a new tab)"}</span>
+            <span className="sr-only">(opens in a new tab)</span>
           </Link>
         </div>
       )}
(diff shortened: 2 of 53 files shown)


client/app/components/EmailSettingsWarning.jsx
```diff
@@ -16,16 +16,20 @@ export default function EmailSettingsWarning({ featureName, className, mode, adm
...
   );
 
   if (mode === "icon") {
     return (
-      <Tooltip title={message}>
-        <i className={cx("fa fa-exclamation-triangle", className)} />
+      <Tooltip title={message} placement="topRight" arrowPointAtCenter>
+        <i
+          className={cx("fa fa-exclamation-triangle", className)}
+          aria-label="Mail alert"
```
gabrieldutra, line 30: What's the difference between using `aria-label` vs the previous `<span className="sr-only" />`?

client/app/components/FavoritesControl.jsx
```diff
@@ -31,6 +31,7 @@ export default class FavoritesControl extends React.Component {
     return (
       <a
         title={title}
+        aria-label={title}
```
gabrieldutra, line 34: It seems in this case since the anchor already has a title it would say it twice? 🤔

client/app/components/groups/ListItemAddon.jsx
```diff
@@ -4,16 +4,32 @@ import Tooltip from "antd/lib/tooltip";
...
+        <span className="sr-only">Remove</span>
+      </>
+    );
   }
   if (alreadyInGroup) {
     return (
       <Tooltip title="Already selected">
-        <i className="fa fa-check" />
+        <i className="fa fa-check" aria-hidden="true" />
+        <span className="sr-only">Already selected</span>
```
gabrieldutra, line 18: I feel like tooltips should have accessibility support by Antd rather than us adding manually to each of our uses. However it doesn't 😕, so I'm no so sure how to proceed anyway. Maybe we should have a component to make it easier to drop off once Antd starts to have accessible tooltips?

client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
```diff
@@ -14,8 +14,8 @@ export default function VersionInfo() {
         <div className="m-t-10">
           {/* eslint-disable react/jsx-no-target-blank */}
           <Link href="https://version.redash.io/" className="update-available" target="_blank" rel="noopener">
-            Update Available
-            <i className="fa fa-external-link m-l-5" />
+            Update Available <i className="fa fa-external-link m-l-5" aria-hidden="true" />
+            <span className="sr-only">&#40;External link&#41;</span>
```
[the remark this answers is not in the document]
rafawendel: I found escaping more agreeable than interpolating, but it's more of a stylistic choice. I really like your suggested approach of embarking this logic inside `<Link>` and it's what most current standards are moving toward. Maybe it extends a bit too much the scope of this work. I'll follow up on it in an issue.

About "external link" vs "opens in a new tab". I might agree the second one is better, but at the same time the intention of a11y is to convey the same meaning the icon should have. Also, it's likely screen readers can interpret `target="_blank"` by themselves.

client/app/pages/queries/VisualizationEmbed.jsx
```diff
@@ -120,7 +120,8 @@ function VisualizationEmbedFooter({
         <span className="hidden-print">
           <Tooltip title="Open in Redash">
             <Link.Button className="icon-button" href={queryUrl} target="_blank">
-              <i className="fa fa-external-link" />
+              <i className="fa fa-external-link" aria-hidden="true" />
+              <span className="sr-only">Open in Redash</span>
```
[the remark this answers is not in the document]
rafawendel: This is an exception, as this is not quite an external link, but a reference to Redash for embedded pages.

client/app/components/EmailSettingsWarning.jsx
rafawendel: The idea is that labels are a bit simpler, but there shouldn't be much of one. Also, labels must only be added to interactive or page-semantic elements, which is the major reason the second approach was used elsewhere, as I find it a bit cumbersome.

client/app/components/FavoritesControl.jsx
rafawendel: Unfortunately `title` is not widely accessible, so you will find some redundancies involving it in some bit of code. A more interesting approach would be to try to abstract the ambiguity of this pattern into `<Link>`, WDYT?

client/app/components/groups/ListItemAddon.jsx
rafawendel: Tooltips are sort of a pain I will try to handle separately. In the meantime I'm uncertain whether I should provide a minimal approach to them or wait until they can be dealt with. Probably I'll just remove these changes.

client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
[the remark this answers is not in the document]
gabrieldutra: > Maybe it extends a bit too much the scope of this work. I'll follow up on it in an issue.

Sure, it's a relatively small refactor, but it changes a component that is used along the app entirely, so whatever works best for you.

> About "external link" vs "opens in a new tab". I might agree the second one is better, but at the same time the intention of a11y is to convey the same meaning the icon should have. Also, it's likely screen readers can interpret target="_blank" by themselves.

🤔 My understanding was actually in the other direction, that we wanted to highlight the "open in new tab" behavior rather than the icon. Also the `fa-external-link` icon is actually used as "opens in new tab", despite of its name. Re. the screen readers, I'm not sure, when I checked from a quick google search the recommendations were on the side that we should clarify the new window/tab behavior disregarding the `target="_blank"`

client/app/components/EmailSettingsWarning.jsx
gabrieldutra: Ok, so specifically for this case, which is the same `<i>` used for an icon, is there a reason to use `aria-label` in here and not in the others (or the opposite way and not use the `<span>` in here specifically)? I'm mostly wondering what's the approach we should follow when adding an icon into our codebase

client/app/components/FavoritesControl.jsx
gabrieldutra: I see, based on [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns) it seems to me that we should avoid/drop the `title` use cases actually

wenbow:
The high level bullet points definitely make a lot of sense. The aria tag looks great. Wrt the antd or our design system, I will defer to our awesome engs. :)

client/app/assets/less/inc/generics.less
```diff
@@ -2,163 +2,229 @@
...
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
+}
+
+.not-sr-only {
```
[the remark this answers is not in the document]
rafawendel: Actually this reverts what `.sr-only` does, it doesn't hide from SRs. Hiding from them is achieved the same way hiding elsewhere `visibility: hidden` or `display: none`.

client/app/components/EmailSettingsWarning.jsx
rafawendel: Actually this one was sort of a mistake, because it makes no sense to have such text in icons, which should be hidden to SRs.

client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
```diff
@@ -15,7 +15,7 @@ export default function VersionInfo() {
           {/* eslint-disable react/jsx-no-target-blank */}
           <Link href="https://version.redash.io/" className="update-available" target="_blank" rel="noopener">
             Update Available <i className="fa fa-external-link m-l-5" aria-hidden="true" />
-            <span className="sr-only">&#40;External link&#41;</span>
+            <span className="sr-only">{"(External link)"}</span>
```
gabrieldutra: Sorry, I may have been not that clear since I also put the quotes, but what I mean is: does this needs this at all? Can't it be just `<span className="sr-only">(External link)</span>`?

rafawendel: No, it can't. I changed back bc I thought this way was clearer.

client/app/components/BigMessage.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 
 function BigMessage({ message, icon, children, className }) {
   return (
-    <div className={"big-message p-15 text-center " + className}>
+    <div className={"big-message p-15 text-center " + className} aria-live="assertive">
```
rafawendel, line 6: Oriented by [this](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/hiding-and-updating-content).

client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
gabrieldutra: Why it can't?  HTML/jsx wise there's nothing that holds you from writing it directly (we do it in several other places in the codebase). As for screen-readers I don't really see how having it escaped would make a difference 🤔

I'm just curious because this really looks weird to me

client/app/assets/less/inc/generics.less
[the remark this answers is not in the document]
gabrieldutra: It seems it's not being used in this PR(?)

Do you have an example on where this would be useful? (i.e.: why not just `<span className={cx({ "sr-only": condition } />` would do it?)

client/app/pages/queries/components/QueryPageHeader.jsx
```diff
@@ -199,8 +199,8 @@ export default function QueryPageHeader({
 
         {!queryFlags.isNew && (
           <Dropdown overlay={moreActionsMenu} trigger={["click"]}>
-            <Button data-test="QueryPageHeaderMoreButton">
-              <EllipsisOutlinedIcon rotate={90} />
+            <Button data-test="QueryPageHeaderMoreButton" aria-label="More actions">
+              <EllipsisOutlinedIcon rotate={90} aria-label="" />
```
gabrieldutra, line 203: May be a newbie question as I'm no expert with those, but what's `aria-label=""` used for? Why not `aria-hidden="true"`?

client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
rafawendel: For some reason it didn't work on my initial tests but now it seems fine. Thanks for pointing this, just reverted it.

client/app/pages/queries/components/QueryPageHeader.jsx
rafawendel: The intention here is to make sure Ant's `aria-labels` will not be displayed, so as to avoid confusion. Theoretically `aria-hidden` does the trick, but I wanted to override it in some cases either way. Not a lot to back me up on this, though.

rafawendel: Just checked with a real SR. It does the trick.

client/app/assets/less/inc/generics.less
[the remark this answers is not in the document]
thielium: Why wouldn't we just not use `.sr-only`, instead of overriding it with this?

client/app/components/ApplicationArea/ApplicationLayout/VersionInfo.jsx
```diff
@@ -14,8 +14,8 @@ export default function VersionInfo() {
         <div className="m-t-10">
           {/* eslint-disable react/jsx-no-target-blank */}
           <Link href="https://version.redash.io/" className="update-available" target="_blank" rel="noopener">
-            Update Available
-            <i className="fa fa-external-link m-l-5" />
+            Update Available <i className="fa fa-external-link m-l-5" aria-hidden="true" />
+            <span className="sr-only">(opens in a new tab)</span>
```
thielium, line 18: Is there an aria attribute (e.g. `aria-label`) that might be clearer?

rafawendel: An `aria-label` would have to be added to the parent anchor, which kind of messes the meaning, because it would be read before the inner content. Also, it is not recommended that aria-labels be added to elements that have content which represent their meaning. The intent of the visually hidden span is to replace the meaning of the icon itself, which cannot be done with an aria-label, as the icon need to be hidden from the SR (otherwise there is a risk it'll be read as a random char). Take a look the [guide from Font Awesome](https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility).

rafawendel: There is actually one further benefit from avoiding conveying meaning with attributes: it's way easier to i18n a span than an attribute (as of now it requires a hook).

client/app/components/ParameterApplyButton.jsx
```diff
@@ -7,16 +7,22 @@ import KeyboardShortcuts from "@/services/KeyboardShortcuts";
...
+      <i className="fa fa-spinner fa-pulse" aria-hidden="true" />
+      <span className="sr-only">Loading...</span>
+    </span>
+  ) : (
+    <i className="fa fa-check" aria-hidden="true" />
+  );
 
   return (
     <div className="parameter-apply-button" data-show={!!paramCount} data-test="ParameterApplyButton">
+      {/* TODO: This button stays detectable even when invisible */}
```
gabrieldutra, line 21: Same here, but I'd actually favor just creating Follow Up issues to track these instead of putting `TODO`s in the application (they are usually forsaken).

client/app/components/queries/QueryEditor/AutocompleteToggle.jsx
```diff
@@ -26,7 +26,7 @@ export default function AutocompleteToggle({ available, enabled, onToggle }) {
   return (
     <Tooltip placement="top" title={tooltipMessage}>
       <Button className="query-editor-controls-button m-r-5" disabled={!available} onClick={handleClick}>
-        <i className={"icon " + icon} />
+        <i className={"icon " + icon} aria-hidden="true" aria-label={tooltipMessage} />
```
gabrieldutra, line 29: This feels contradictory(?)

client/app/pages/dashboards/components/DashboardHeader.jsx
```diff
@@ -193,8 +193,12 @@ function DashboardControl({ dashboardConfiguration, headerExtra }) {
           {showRefreshButton && <RefreshButton dashboardConfiguration={dashboardConfiguration} />}
           {showFullscreenButton && (
             <Tooltip className="hidden-xs" title="Enable/Disable Fullscreen display">
-              <Button type={buttonType(fullscreen)} className="icon-button m-l-5" onClick={toggleFullscreen}>
-                <i className="zmdi zmdi-fullscreen" />
+              <Button
+                type={buttonType(fullscreen)}
+                className="icon-button m-l-5"
+                onClick={toggleFullscreen}
+                aria-label="Fullscreen display">
```
gabrieldutra, line 200:
```suggestion
                aria-label="Toggle Fullscreen">
```
?

client/app/components/BigMessage.jsx
```diff
@@ -3,12 +3,17 @@ import PropTypes from "prop-types";
...
   return (
-    <div className={"p-15 text-center " + className}>
-      <h3 className="m-t-0 m-b-0">
-        <i className={"fa " + icon} />
+    <div
+      className={"big-message p-15 text-center " + className}
+      role="status"
+      aria-live="assertive"
+      aria-relevant="additions removals">
+      {/* TODO: replace misuse of header */}
```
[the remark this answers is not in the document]
rafawendel: Jira issue under a11y epic

client/app/components/BigMessage.jsx
```diff
@@ -3,12 +3,17 @@ import PropTypes from "prop-types";
...
+      <h3 className="m-t-0 m-b-0" aria-labelledby="bm-message">
+        <i className={"fa " + icon} aria-hidden="true" />
       </h3>
       <br />
-      {message}
+      <span id="bm-message">{message}</span>
```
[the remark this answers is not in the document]
rafawendel: We do! Thanks for pointing this out

client/app/components/EmailSettingsWarning.jsx
```diff
@@ -16,16 +15,18 @@ export default function EmailSettingsWarning({ featureName, className, mode, adm
   }
 
   const message = (
-    <span>
+    <span id="sr-mail-description">
```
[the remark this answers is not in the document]
rafawendel: This one is actually Ok, because it's not being replicated in a way that could end up rendering two on the screen.

client/app/pages/dashboards/components/DashboardHeader.jsx
rafawendel: It's actually different, because it doesn't cause the app to enter fullscreen, just to hide accessory elements "fullscreen mode"

client/app/components/ParameterApplyButton.jsx
rafawendel: I've been doing both, particularly because I intend to tackle these in the near future and it helps me keeping it in mind. Are you in favor of not keeping the comments for a particular reason?

client/app/pages/dashboards/components/DashboardHeader.jsx
gabrieldutra: My concern is that "Fullscreen display" is not an action 😕

client/app/components/ParameterApplyButton.jsx
gabrieldutra: My reason is that they bring little or no value. Even more when it is about new features as they always have a "where to be improved". The result is that we start to have multiple comments in the codebase that are not relevant to who's reading the code as they won't even be aware of the context 🤷‍♂️.

That said, I'm against TODO comments in the code. There are a couple of exceptions I consider for them however:
1. When you actually want to index your files with some easily searchable term (e.g.: we used `ANGULAR_REMOVE_ME` to track any of the still dependent leftovers while migrating away from it)
2. When the `TODO` is actually relevant to anyone that will read the code (I can't think about an example for this now 😕)

Anyway, this is more of an opinion, maybe someone else has a different thought...

rafawendel: SGTM!

client/app/components/dashboards/dashboard-widget/VisualizationWidget.jsx
```diff
@@ -161,16 +162,19 @@ function VisualizationWidgetFooter({ widget, isPublic, onRefresh, onExpand }) {
             className="refresh-button hidden-print btn btn-sm btn-default btn-transparent"
             onClick={() => refreshWidget(1)}
             data-test="RefreshButton">
-            <i className={cx("zmdi zmdi-refresh", { "zmdi-hc-spin": refreshClickButtonId === 1 })} />{" "}
+            <i className={cx("zmdi zmdi-refresh", { "zmdi-hc-spin": refreshClickButtonId === 1 })} aria-hidden="true" />
+            <span className="sr-only">
+              {refreshClickButtonId === 1 ? "Refreshing, please wait. " : "Press to refresh. "}
```
thielium, line 167: Do or not: feels like some duplication, here -- if you feel like you can abstract it, go for it! But if you think it could create business logic changes, don't worry about  it.

client/app/components/groups/ListItemAddon.jsx
```diff
@@ -4,16 +4,34 @@ import Tooltip from "antd/lib/tooltip";
...
+        <i className="fa fa-remove" aria-hidden="true" />
+        <span className="sr-only">Remove</span>
+      </>
+    );
   }
   if (alreadyInGroup) {
     return (
       <Tooltip title="Already selected">
-        <i className="fa fa-check" />
+        <span tabIndex={0}>
```
thielium, line 17: What's the tab index for? Is that for triggering the tooltip?

client/app/pages/alert/components/NotificationTemplate.jsx
```diff
@@ -94,7 +94,8 @@ function NotificationTemplate({ alert, query, columnNames, resultValues, subject
             data-test="CustomBody"
           />
           <HelpTrigger type="ALERT_NOTIF_TEMPLATE_GUIDE" className="f-13">
-            <i className="fa fa-question-circle" /> Formatting guide
+            <i className="fa fa-question-circle" aria-hidden="true" /> Formatting guide
+            <span className="sr-only"> (help)</span>
```
thielium, line 98: (nit) Extra space?

thielium (approved):
Aside from the messageId ref conversation, just nits! Way to go! This is a huge step forward!

client/app/components/HelpTrigger.jsx
```diff
@@ -170,7 +170,13 @@ export function helpTriggerWithTypes(types, allowedDomains = [], drawerClassName
               this.props.showTooltip ? (
                 <>
                   {tooltip}
-                  {shouldRenderAsLink && <i className="fa fa-external-link" style={{ marginLeft: 5 }} />}
+                  {shouldRenderAsLink && (
+                    <>
+                      {" "}
+                      <i className="fa fa-external-link" style={{ marginLeft: 5 }} aria-hidden="true" />
+                      <span className="sr-only">(opens in a new tab)</span>
```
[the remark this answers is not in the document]
rafawendel: Coming soon 😉

client/app/components/groups/ListItemAddon.jsx
rafawendel: Yup. Otherwise it's impossible to trigger it via focus. A further PR will globally add focus as a trigger for tooltips.

client/app/components/BigMessage.jsx
```diff
@@ -1,19 +1,21 @@
 import React from "react";
 import PropTypes from "prop-types";
+import { uniqueId } from "lodash";
 
 function BigMessage({ message, icon, children, className }) {
+  const messageId = uniqueId("bm-message");
```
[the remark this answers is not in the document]
gabrieldutra: This partially solves the thing as it will execute `uniqueId` regardless. Problem with this is that iirc, Lodash uses some sort of global var to track and sum up the ids. (We would end up with some possibly big ones 😝)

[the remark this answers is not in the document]
gabrieldutra: In brief, that was to avoid [this](https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily) as that will avoid the component id to be updated, but not the Lodash reference to the "next id"

client/app/lib/hooks/useLazyRef.ts
```diff
@@ -0,0 +1,22 @@
...
+  const lazyRef = useRef<T | typeof noValue>(noValue);
+
+  if (lazyRef.current === noValue) {
+    lazyRef.current = getInitialValue();
+  }
+
+  return lazyRef as React.MutableRefObject<T>;
+}
+
+export default useLazyRef;
```
gabrieldutra, line 22: nit: prefer the named export `export function useLazyRef(...)`

client/app/lib/hooks/useUniqueId.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  return lazyRef as React.MutableRefObject<T>;
+}
+
+function useUniqueId(prefix: string) {
+  const { current: id } = useLazyRef(() => uniqueId(prefix));
+  return id;
+}
+
+export default useUniqueId;
```
gabrieldutra, line 9: same thing about the named export 🙂

gabrieldutra (approved):
Thanks for the work and patience here @rafawendel 🐳

PR #5432, opened 2021-03-23
Embed "external" link type into `<Link>` component

rafawendel:

- [x] Refactor

## Description
Minor refactor of `<Link>` to allow abstraction of repetitive "external link" pattern.

### DIFF
--- a/client/app/components/Link.tsx
+++ b/client/app/components/Link.tsx
@@ -1,8 +1,17 @@
 import React from "react";
 import Button, { ButtonProps as AntdButtonProps } from "antd/lib/button";
 
+function DefaultLinkComponent({ children, ...props }: React.AnchorHTMLAttributes<HTMLAnchorElement>) {
+  return <a {...props}>{children}</a>;
+}
+
+Link.Component = DefaultLinkComponent;
+
 interface LinkProps extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, "role" | "type" | "target"> {
   href: string;
+}
+function Link({ children, ...props }: LinkProps) {
+  return <Link.Component {...props}>{children}</Link.Component>;
 }
 
 interface LinkWithIconProps extends LinkProps {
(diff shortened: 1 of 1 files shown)


client/app/components/Link.tsx
```diff
@@ -15,10 +21,36 @@ function DefaultLinkComponent({ children, ...props }: React.AnchorHTMLAttributes
...
+}
+
+Link.WithIcon = LinkWithIcon;
+
+function ExternalLink({
+  icon = <i className="fa fa-external-link" aria-hidden="true" />,
+  alt = "(opens in a new tab)",
+  ...props
+}: LinkWithIconProps) {
+  const externalLinkProps = { target: "_blank", rel: "noopener noreferrer", icon, alt };
```
kravets-levko, line 47: Why not just use these props directly in the next line?

rafawendel: Not doing so allows the user to overhead them via normal props, which is intended (except for target, which is banned from `<Link>`

kravets-levko: IMHO, considering that all props here are constants (or other props), inline version will look more clear:

```jsx
return <Link.WithIcon target="_blank" rel="noopener noreferrer" icon={icon} alt={alt} {...props} />;
```

Using object for additional props usually makes sense if you need some conditional or computed props - in this case it will make code cleaner. But in this case I don't see actual need for that

client/app/components/Link.tsx
```diff
@@ -1,31 +1,57 @@
...
+  return <Link.Component {...props}>{children}</Link.Component>;
+}
 
-function DefaultLinkComponent({ children, ...props }: React.AnchorHTMLAttributes<HTMLAnchorElement>) {
-  return <a {...props}>{children}</a>;
+interface LinkWithIconProps extends LinkProps {
+  children: string;
+  icon: JSX.Element;
+  alt: string;
+  target?: "_blank" | "_parent" | "_top";
```
kravets-levko, line 21:
```suggestion
  target?: "_self" | "_blank" | "_parent" | "_top";
```

kravets-levko (approved):
👍

client/app/components/Link.tsx
rafawendel: removed `_self` so that just omitting target is preferred over it. If I allow it, the external link will be allowed to use `target="_self"`, which is a behavior intended to be reserved to `<Link.WithIcon>`. WDYT?

kravets-levko: Makes sense 👍

PR #5434, opened 2021-03-23
Adds configuration for `<Tooltip>` trigger on focus

rafawendel:

- [X] Feature

## Description
Ant `<Tooltip>` API has a `trigger` prop that by default only receives `"hover"`. This PR changes the default to `"hover"` and `"focus"`, in order to make it accessible.

See a11y epic.

### DIFF
--- a/client/app/config/antd-tooltip.jsx
+++ b/client/app/config/antd-tooltip.jsx
@@ -1,0 +1,3 @@
+import Tooltip from "antd/lib/tooltip";
+
+Tooltip.defaultProps = { ...Tooltip.defaultProps, trigger: ["hover", "focus"] };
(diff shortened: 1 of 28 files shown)


client/app/config/antd-tooltip.jsx
```diff
@@ -0,0 +1,3 @@
+import Tooltip from "antd/lib/tooltip";
+
+Tooltip.defaultProps = { ...Tooltip.defaultProps, trigger: ["hover", "focus"] };
```
thielium, line 3: This is a pretty far reaching change. It also misses existing tooltips which explicitly say "hover"` but should also support "focus."

We shouldn't mutate ant's behavior -- if someone in the future is not aware of this file, it could be very confusing to figure out / migrate / upgrade.

If we do want to have default props, we can wrap their Tooltip in our own component, and use that. Then migration / discoverability isn't a problem. We could also gently coerce folks into using it with a linting rule, if we so choose.

rafawendel: My intention is to change the component as little as I can, so leaving the possibility for a developer to omit and add triggers in the future is intentional. Also searching `<Tooltip[^\r<]trigger` in our codebase returns no result, i.e., no one is providing custom triggers.

kravets-levko (approved):
LGTM 👍

thielium:
🏆

PR #5435, opened 2021-03-23
Add ambiguous link

rafawendel:

- [X] Refactor

## Description
Adds a new component that will conditionally render a `<Link>` or `<Button>` according to the presence of an `href` in order to treat components that can receive both and should look alike regardless.

There are some remaining components keeping both, but they represent a genuine use of it.

### DIFF
--- a/client/app/components/Link.tsx
+++ b/client/app/components/Link.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import Button, { ButtonProps as AntdButtonProps } from "antd/lib/button";
+import PlainButton, { PlainButtonProps } from "@/components/PlainButton";
 
 function DefaultLinkComponent({ children, ...props }: React.AnchorHTMLAttributes<HTMLAnchorElement>) {
   return <a {...props}>{children}</a>;
@@ -13,6 +14,18 @@ interface LinkProps extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 
 function Link({ children, ...props }: LinkProps) {
   return <Link.Component {...props}>{children}</Link.Component>;
 }
+
+type LinkButtonProps = PlainButtonProps | LinkProps;
+
+function LinkOrButton(props: LinkButtonProps) {
+  if ("href" in props) {
+    return <Link {...props} />;
+  }
+
+  return <PlainButton type="link" {...props} />;
+}
+
+Link.OrButton = LinkOrButton;
 
 interface LinkWithIconProps extends LinkProps {
   children: string;
--- a/client/app/components/cards-list/CardsList.tsx
+++ b/client/app/components/cards-list/CardsList.tsx
@@ -10,8 +10,8 @@ import "./CardsList.less";
 export interface CardsListItem {
   title: string;
   imgSrc: string;
-  onClick?: () => void;
   href?: string;
+  onClick?: React.MouseEventHandler<HTMLElement>;
 }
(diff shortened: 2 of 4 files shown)


client/app/components/Link.tsx
```diff
@@ -14,6 +15,18 @@ function Link({ children, ...props }: LinkProps) {
   return <Link.Component {...props}>{children}</Link.Component>;
 }
 
+type LinkButtonProps = PlainButtonProps | LinkProps;
+
+function LinkOrButton(props: LinkButtonProps) {
```
thielium, line 20: This feels too magical to me. I feel like our markup should stay stable -- if a developer wants to display a link, they should display a link. If they want to display a button, they can display a button.

rafawendel: This component is specifically intended for cases in which the developer doesn't know what he needs to display in advance. I'll push changes that better explain that.

rafawendel: Maybe we can improve its usage by requiring that both href and onClick attributes be set? WDYT?

thielium: I think the developer should specify inline which one needs to be displayed. Ideally, our markup should be stable. 

For example, if we're trying to write a test for a page that has some interactive element, we'll now need to look at the code to determine whether or not we should query for a `link` role or a `button` role, right?

Can you show me some concrete places where we're having this issue?

rafawendel: Just added them. I don't think there are any others as of now. The problem stated was present before either way, as an `<a>` rendered without an `href`, which was the case, would result not only in an inaccessible link, but also in one that lacks link role.

client/app/components/empty-state/EmptyState.jsx
```diff
@@ -17,9 +17,9 @@ export function Step({ show, completed, text, url, urlTarget, urlText, onClick }
 
   return (
     <li className={classNames({ done: completed })}>
-      <Link href={url} onClick={onClick} target={urlTarget}>
+      <Link.OrButton href={url} onClick={onClick} target={urlTarget}>
```
thielium, line 20: In what cases would we want this to not just be a link?

client/app/components/cards-list/CardsList.tsx
```diff
@@ -26,11 +26,10 @@ interface ListItemProps {
 
 function ListItem({ item, keySuffix }: ListItemProps) {
   return (
-    // @ts-expect-error TODO: refactor component
-    <Link key={`card${keySuffix}`} className="visual-card" onClick={item.onClick} href={item.href}>
+    <Link.OrButton key={`card${keySuffix}`} className="visual-card" onClick={item.onClick} href={item.href}>
```
thielium, line 29: When shouldn't this be a link?

client/app/components/Link.tsx
gabrieldutra: Considering:
1. there are very few cases where it can be both a link or a button depending on the parameters
2. we want to encourage the discerning of a Button and a Link (that by the fact that you initially separated those in different components)

I think creating a new component is not what we want here. I'd rather leave this responsibility to the behavior of each component that might require this "exceptional" behavior.

client/app/components/cards-list/CardsList.tsx
rafawendel: href might not come

PR #5438, opened 2021-03-24
StorageDescriptorの無いテーブルで止まらないようにした

iwakiriK:

- [x] Bug Fix

## Description
AWS Glue has a key called StorageDescriptor. This key is nullable on AWS, but not on redash. Therefore, if a table without StorageDescriptor exists in Glue, it will fail to retrieve the schema.

https://docs.aws.amazon.com/glue/latest/webapi/API_PartitionInput.html#Glue-Type-PartitionInput-StorageDescriptor


iwakiriK:
This PR was created by mistake.
Correct PR: https://github.com/getredash/redash/pull/5447

PR #5440, opened 2021-03-24
Add live regions to tooltip

rafawendel:

- [X] Feature

## Description
This makes the tooltip accessible, as it makes it's content notifiable to screen readers, which otherwise would not occur (tooltips are appended to the body).

See a11y epic.


kravets-levko (approved):
👍

PR #5444, opened 2021-03-26
Replace hardcoded ids with hook

rafawendel:

- [X] Refactor

## Description
This replaces all hard coded `id` attributes with an unique ID generated from `useUniqueId`.
The code base was searched for all occurrences of the previous IDs to avoid unwanted lost references.

### DIFF
--- a/client/app/components/CreateSourceDialog.jsx
+++ b/client/app/components/CreateSourceDialog.jsx
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { isEmpty, toUpper, includes, get } from "lodash";
+import { isEmpty, toUpper, includes, get, uniqueId } from "lodash";
 import Button from "antd/lib/button";
 import List from "antd/lib/list";
 import Modal from "antd/lib/modal";
@@ -44,6 +44,8 @@ class CreateSourceDialog extends React.Component {
     savingSource: false,
     currentStep: StepEnum.SELECT_TYPE,
   };
+
+  formId = uniqueId("sourceForm");
 
   selectType = selectedType => {
     this.setState({ selectedType, currentStep: StepEnum.CONFIGURE_IT });
@@ -117,7 +119,7 @@ class CreateSourceDialog extends React.Component {
             </HelpTrigger>
           )}
         </div>
-        <DynamicForm id="sourceForm" fields={fields} onSubmit={this.createSource} feedbackIcons hideSubmitButton />
+        <DynamicForm id={this.formId} fields={fields} onSubmit={this.createSource} feedbackIcons hideSubmitButton />
         {selectedType.type === "databricks" && (
           <small>
             By using the Databricks Data Source you agree to the Databricks JDBC/ODBC{" "}
@@ -171,7 +173,7 @@ class CreateSourceDialog extends React.Component {
                 <Button
                   key="submit"
                   htmlType="submit"
-                  form="sourceForm"
+                  form={this.formId}
                   type="primary"
                   loading={savingSource}
                   data-test="CreateSourceSaveButton">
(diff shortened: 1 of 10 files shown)


client/app/components/CreateSourceDialog.jsx
```diff
@@ -45,6 +46,8 @@ class CreateSourceDialog extends React.Component {
     currentStep: StepEnum.SELECT_TYPE,
   };
 
+  formId = useUniqueId("sourceForm");
```
kravets-levko, line 49: Just wondering: how does it work with a class component? Hooks shouldn't work with class components, only with functional ones

rafawendel: I didn't even think about it Lol. Will workaround it soon.

thielium:
Will check it out after the class components issue is resolved, but so far LGTM

rafawendel:
UPDATE: the problem sorted itself. Blocked by #5474 -- will be rebased.

Ignore import order, will add `eslint-plugin-imports` soon.

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -144,7 +147,7 @@ function DashboardComponent(props) {
         </div>
       )}
       {editingLayout && <DashboardSettings dashboardConfiguration={dashboardConfiguration} />}
-      <div id="dashboard-container">
+      <div id={dashboardContainerId}>
```
kravets-levko, line 150: This one is different - it's used for styling:
https://github.com/getredash/redash/blob/d8d7c78992e44a4b6d7fdd4c39ccc1c35cd8c7a9/client/app/pages/dashboards/DashboardPage.less#L16-L19

client/app/pages/dashboards/PublicDashboardPage.jsx
```diff
@@ -37,7 +40,7 @@ function PublicDashboard({ dashboard }) {
           <Filters filters={filters} onChange={setFilters} />
         </div>
       )}
-      <div id="dashboard-container">
+      <div id={dashboardContainerId}>
```
kravets-levko, line 43: Same as above: https://github.com/getredash/redash/blob/d8d7c78992e44a4b6d7fdd4c39ccc1c35cd8c7a9/client/app/pages/dashboards/DashboardPage.less#L16-L19

package.json
```diff
@@ -73,6 +73,7 @@
     "react-grid-layout": "^0.18.2",
     "react-resizable": "^1.10.1",
     "react-virtualized": "^9.21.2",
+    "short-uuid": "^4.1.0",
```
kravets-levko, line 76: 👍

thielium (commented):
General question / comment: this approach is going to create a lot of thrashing of ids / updating the DOM, right? Every single render, for some of these usages, we get a new id.

What's the motivation, here? Are we worried about collisions? What was wrong with the `useUniqueId` approach? I've seen that in other codebases, and it works well. At the minimum, we should figure out a way to make the uuids stable.

kravets-levko:
> General question / comment: this approach is going to create a lot of thrashing of ids / updating the DOM, right? Every single render, for some of these usages, we get a new id.

> What's the motivation, here? Are we worried about collisions? What was wrong with the useUniqueId approach? I've seen that in other codebases, and it works well. At the minimum, we should figure out a way to make the uuids stable.

The motivation is to keep things simple. We can always wrap `getUniqueId()` calls with `useMemo` if it makes any problems.

Re. collisions: `short-uuid` generates a regular 16-bytes UUIDs, but uses base64-like representation instead of hexadecimal. There's practically no chance to get a collision.

kravets-levko (approved):
Please merge it before you decide to change it back :smile:

thielium:
>The motivation is to keep things simple.

I think we both want this!

Might be worth an offline discussion.

PR #5447, opened 2021-03-29
Skip tables with no storage descriptor

iwakiriK:

- [x] Bug Fix

## Description
AWS Glue table has a key called StorageDescriptor. This key is nullable on AWS, but not on redash. Therefore, if tables without StorageDescriptor exist in Glue, some schema refreshes will fail.
As a solution, skip retrying tables with no StorageDescriptor.

image

https://docs.aws.amazon.com/glue/latest/webapi/API_PartitionInput.html#Glue-Type-PartitionInput-StorageDescriptor

### DIFF
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -179,6 +179,8 @@ class Athena(BaseQueryRunner):
             for database in databases["DatabaseList"]:
                 iterator = table_paginator.paginate(DatabaseName=database["Name"])
                 for table in iterator.search("TableList[]"):
+                    if 'StorageDescriptor' not in table:
+                        continue
                     table_name = "%s.%s" % (database["Name"], table["Name"])
                     if table_name not in schema:
                         column = [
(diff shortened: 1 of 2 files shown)


iwakiriK:
I think the following changes might be better.
I'd like to hear your opinions.

`for columns in table['StorageDescriptor']['Columns']]`
↓
`for columns in table.get('StorageDescriptor', {}).get('Columns', {})`

redash/query_runner/athena.py
```diff
@@ -180,6 +180,9 @@ def __get_schema_from_glue(self):
                 iterator = table_paginator.paginate(DatabaseName=database["Name"])
                 for table in iterator.search("TableList[]"):
                     table_name = "%s.%s" % (database["Name"], table["Name"])
+                    if 'StorageDescriptor' not in table:
+                        logger.warning("Glue table dosen't have StorageDescriptor: %s", table_name)
```
susodapop, line 184:
```suggestion
                        logger.warning("Glue table doesn't have StorageDescriptor: %s", table_name)
```

susodapop (approved):
LGTM just one typo.

Thanks!

PR #5450, opened 2021-03-30
Get the user's current groups from props instead of useEffect()

### LINKED ISSUE #5449: Edit User Page: current groups is always empty
## Issue Summary
When viewing a user profile with edit permission, the `<select>` of their current groups is empty.

## Steps to Reproduce

On the latest tip of master with at least two users, log in as one user and view the other user under Settings > Users  > **User**. On the edit page their current groups appears empty:

image

It can be reproduced on https://redash-preview.netlify.app/ easily.

(screenshots taken from [this post](https://discuss.redash.io/t/latest-v9-preview-user-edit-page-doesnt-show-user-group-correctly/8246) on the user forum).

## Technical details:
Redash Version: Version: 9.0.0-beta (44178d99)
Browser/OS: Chrome
How did you install Redash: Docker

susodapop:

- [x] Bug Fix

## Description

`useEffect()` doesn't run until _after_ the component renders. Before the hook runs, the value of `groups` === []. And this is passed to `<DynamicForm>`'s `initialValue` prop. The `initialValue` is not re-evaluated after useEffect() completes. So the users groups are never updated. This change pulls the user's current groups from `user` prop on the page.

Closes #5449 

### Before

CleanShot 2021-03-30 at 16 08 54

### After

CleanShot 2021-03-30 at 16 08 42

### DIFF
--- a/client/app/pages/users/components/UserInfoForm.jsx
+++ b/client/app/pages/users/components/UserInfoForm.jsx
@@ -61,7 +61,7 @@ export default function UserInfoForm(props) {
                 type: "select",
                 mode: "multiple",
                 options: map(allGroups, group => ({ name: group.name, value: group.id })),
-                initialValue: map(groups, group => group.id),
+                initialValue: user.groupIds,
                 loading: isLoadingGroups,
                 placeholder: isLoadingGroups ? "Loading..." : "",
               }


kravets-levko (approved):
Looks good, thank you @susodapop! @gabrieldutra WDYT?

gabrieldutra (approved):
Clever one, thanks @susodapop 🎉

PR #5451, opened 2021-03-31
Fix Ace editor keyboard trap

rafawendel:

- [X] Feature

## Description
Now a screen reader user is able to leave the editor by pressing the "Esc" key. Also, when the user enters the editor a notification is read aloud by the SR, instructing how to leave (try it, it's super cool).

See a11y epic.

### DIFF
--- a/client/app/components/queries/QueryEditor/index.jsx
+++ b/client/app/components/queries/QueryEditor/index.jsx
@@ -2,6 +2,7 @@ import React, { useEffect, useMemo, useState, useCallback, useImperativeHandle }
 import PropTypes from "prop-types";
 import cx from "classnames";
 import { AceEditor, snippetsModule, updateSchemaCompleter } from "./ace";
+import { srNotify } from "@/lib/accessibility";
 import { SchemaItemType } from "@/components/queries/SchemaBrowser";
 import resizeObserver from "@/services/resizeObserver";
 import QuerySnippet from "@/services/query-snippet";
@@ -88,6 +89,15 @@ const QueryEditor = React.forwardRef(function(
     editor.commands.bindKey({ win: "Ctrl+P", mac: null }, null);
     // Lineup only mac
     editor.commands.bindKey({ win: null, mac: "Ctrl+P" }, "golineup");
+
+    // Esc for exiting
+    editor.commands.bindKey({ win: "Esc", mac: "Esc" }, () => {
+      editor.blur();
+    });
+
+    editor.on("focus", () => {
+      srNotify("You've entered the SQL editor. To exit press the ESC key.");
+    });
 
     // Reset Completer in case dot is pressed
     editor.commands.on("afterExec", e => {
--- a/client/app/components/queries/QueryEditor/index.jsx
+++ b/client/app/components/queries/QueryEditor/index.jsx
@@ -95,8 +95,18 @@ const QueryEditor = React.forwardRef(function(
       editor.blur();
     });
 
+    let notificationCleanup = null;
     editor.on("focus", () => {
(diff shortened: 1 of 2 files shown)


rafawendel:
Cc. @wembow

client/app/lib/accessibility.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  [
+    ["role", "alert"],
+    ["aria-live", "assertive"],
+  ].forEach(([attr, value]) => element.setAttribute(attr, value));
+
+  container.appendChild(element);
+
+  setTimeout(() => {
+    container.removeChild(element);
```
thielium, line 19: May need to check that `container` still exists (if I remove the container, and this fires, could get ugly).

client/app/lib/accessibility.ts
```diff
@@ -0,0 +1,21 @@
+export function srNotify(text: string, expiry = 1000, container = document.body): void {
```
thielium, line 1: Some documentation on what this does would be nice (I can see this being reused!).

client/app/lib/accessibility.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  Object.assign(element, {
+    id,
+    className: "sr-only",
+    textContent: text,
+  });
+
+  [
+    ["role", "alert"],
+    ["aria-live", "assertive"],
```
thielium, line 13: Maybe `polite` instead? If they go in and out of the editor a lot, this could get tiresome. Or maybe assertiveness should be a parameter?

client/app/lib/accessibility.ts
thielium: Can we use object-style args instead? With two consecutive optionals, you can get calls like `srNotify('test', null, container)`.

thielium: Could be nice to return a `remove` callback (so if the user blurs really fast, the message is immediately removed).

thielium (commented):
Super cool!

client/app/lib/accessibility.ts
kravets-levko: Container will stay valid while there is at least one reference to it. It means - even if it will be removed from DOM - here it will stay valid, and nothing bad will happen (at least until we decide to access its parent node). However, bad things may happen if we'll decide to remove a child that already does not belong to the element. So this `removeChild` call needs to be wrapped with `try..catch`

arikfr:
@rafawendel will it still allow exiting with Esc from autocomplete?

client/app/lib/accessibility.ts
rafawendel: I was with the mind too much in the specific case I was addressing, indeed this should be a parameter and its default value should be polite.

rafawendel:
> @rafawendel will it still allow exiting with Esc from autocomplete?

@arikfr It actually blurs the autocomplete before it blurs the textarea itself, which is sort of a side effect, but seems good enoug for me, WDYT?

client/app/components/queries/QueryEditor/index.jsx
```diff
@@ -89,6 +90,25 @@ const QueryEditor = React.forwardRef(function(
...
+      notificationCleanup = srNotify({
+        text: "You've entered the SQL editor. To exit press the ESC key.",
+        politeness: "assertive",
+      });
+    });
+
+    editor.on("blur", () => {
+      if (notificationCleanup) {
+        notificationCleanup();
+      }
```
kravets-levko, line 109: `notificationCleanup = null;` ?

thielium (approved):
LGTM. Bonus points if we could add some testing (we won't notice if the accessibility stuff breaks, since we don't use screenreaders)

PR #5456, opened 2021-04-01
Don't disable the test connection button.

### LINKED ISSUE #5455: Data Source Setup: test connection button is disabled after Save
### Issue Summary

The **Test Connection** on the data source page is disabled after the user clicks "Save". Reloading the browser page enables it again.

dirty-button

### Steps to Reproduce

1. Create a JSON data source and save (test connection button is `enabled`)
2. Add a username or password field and click Save (test connection button is `disabled`)
3. Reload the page (test connection button is `enabled`)

The state of the form should be identical. But without a page reload it's treated as "dirty" even after save. 

Expected behavior: saving the form ought to set `dirty===false`.

### Technical details:

* Redash Version: latest master (6228f4cf714c03e672c45d5f311b0faa0fdbf452)
* Browser/OS: Chrome / MacOS / M1 Macbook

susodapop:

- [x] Bug Fix

## Description

This fixes the issue where the Test button is disabled after a user saves changes. As discussed in #5455 I don't think it makes sense to disable the test connection button.

Closes #5455 

N/A

### DIFF
--- a/client/app/pages/data-sources/EditDataSource.jsx
+++ b/client/app/pages/data-sources/EditDataSource.jsx
@@ -108,7 +108,7 @@ class EditDataSource extends React.Component {
       type,
       actions: [
         { name: "Delete", type: "danger", callback: this.deleteDataSource },
-        { name: "Test Connection", pullRight: true, callback: this.testConnection, disableWhenDirty: true },
+        { name: "Test Connection", pullRight: true, callback: this.testConnection, disableWhenDirty: false },
       ],
       onSubmit: this.saveDataSource,
       feedbackIcons: true,


gabrieldutra:
> "It makes sense to disable Test Connection button when user modified data source options but didn't save them yet - because in this case Test button will use old options (ones that are already saved" - Levko K

Considering this, maybe we can show a tooltip instead explaining why the action is disabled? (bad news is that I think I didn't structure the actions that friendly for such change by the time 😅)

guidopetri:
@susodapop guessing you don't want to continue this PR? If you want to keep it going, would you mind rebasing off master so the CI re-runs?

We'll close the PR in about a week if we don't hear back - we can always reopen later. :)

justinclift:
Reading over this:

> "It makes sense to disable Test Connection button when user modified data source options but didn't save them yet - because in this case Test button will use old options (ones that are already saved" - @kravets-levko

Then yeah, this PR probably shouldn't be merged as-is.  We'd likely need to instead have the Test button use the displayed values in the live form, rather than whatever is saved.  Somehow.
That'll be a task for another time though, not in this PR. :smile:

justinclift:
Hmmm, it looks like the original issue this was supposed to fix (#5455) has already been addressed in #5666.

PR #5468, opened 2021-04-09
Fix for `<Link>` misuses

rafawendel:

- [X] Bug Fix

## Description
Add a conditional render for situations in which `<Link>` `href` could come empty from `props`. 

Follows #5435

### DIFF
--- a/client/app/components/empty-state/EmptyState.jsx
+++ b/client/app/components/empty-state/EmptyState.jsx
@@ -4,23 +4,24 @@ import PropTypes from "prop-types";
 import classNames from "classnames";
 import CloseOutlinedIcon from "@ant-design/icons/CloseOutlined";
 import Link from "@/components/Link";
+import PlainButton from "@/components/PlainButton";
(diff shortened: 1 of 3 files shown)


client/app/components/empty-state/EmptyState.jsx
```diff
@@ -4,23 +4,24 @@ import PropTypes from "prop-types";
...
 import Link from "@/components/Link";
+import PlainButton from "@/components/PlainButton";
 import CreateDashboardDialog from "@/components/dashboards/CreateDashboardDialog";
 import HelpTrigger from "@/components/HelpTrigger";
 import { currentUser } from "@/services/auth";
 import organizationStatus from "@/services/organizationStatus";
+
 import "./empty-state.less";
 
-export function Step({ show, completed, text, url, urlTarget, urlText, onClick }) {
```
rafawendel, line 13: In general targets are a bad practice, particularly now that there is a custom component for external links. Also it was unused.

kravets-levko (approved):
IMHO Looks much better than custom component 👍

PR #5470, opened 2021-04-12
Replace redundant external links

rafawendel:

- [X] Refactor

## Description
Moves the icon and a11y parts of `<Link>` into a separate file and add it as a wrapper as well.

#5418, See a11y epic.

### DIFF
--- a/client/app/components/Link.tsx
+++ b/client/app/components/Link.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import Button, { ButtonProps as AntdButtonProps } from "antd/lib/button";
+import { withIcon, WithIconProps } from "./WithIcon";
 
 function DefaultLinkComponent({ children, ...props }: React.AnchorHTMLAttributes<HTMLAnchorElement>) {
   return <a {...props}>{children}</a>;
@@ -7,35 +8,21 @@ function DefaultLinkComponent({ children, ...props }: React.AnchorHTMLAttributes
 
 Link.Component = DefaultLinkComponent;
 
-interface LinkProps extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, "role" | "type" | "target"> {
+interface LinkProps extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, "role" | "type"> {
   href: string;
+  target?: never; // use external link
 }
 function Link({ children, ...props }: LinkProps) {
   return <Link.Component {...props}>{children}</Link.Component>;
 }
 
-interface LinkWithIconProps extends LinkProps {
-  children: string;
-  icon: JSX.Element;
-  alt: string;
-  target?: "_self" | "_blank" | "_parent" | "_top";
-}
-
-function LinkWithIcon({ icon, alt, children, ...props }: LinkWithIconProps) {
-  return (
-    <Link.Component {...props}>
-      {children} {icon} <span className="sr-only">{alt}</span>
-    </Link.Component>
-  );
-}
-
-Link.WithIcon = LinkWithIcon;
+Link.WithIcon = withIcon(Link.Component);
 
 function ExternalLink({
   icon = <i className="fa fa-external-link" aria-hidden="true" />,
   alt = "(opens in a new tab)",
   ...props
-}: Omit<LinkWithIconProps, "target">) {
+}: LinkProps & Partial<WithIconProps>) {
   return <Link.WithIcon target="_blank" rel="noopener noreferrer" icon={icon} alt={alt} {...props} />;
 }
 
--- a/client/app/components/WithIcon.tsx
+++ b/client/app/components/WithIcon.tsx
@@ -1,0 +1,41 @@
+import React from "react";
+
+export interface WithIconProps {
+  children?: string;
+  icon: JSX.Element;
+  alt: string;
+}
+
+export function withIcon<ComponentProps extends object>(Component: React.ComponentType<ComponentProps>) {
+  return class WithIcon extends React.Component<ComponentProps & WithIconProps> {
+    render() {
+      const { children, icon, alt, ...props } = this.props;
+      return (
+        <Component {...(props as ComponentProps)}>
+          {children}
+          {children && " "}
+          {icon} <span className="sr-only">{alt}</span>
+        </Component>
+      );
+    }
+  };
+}
+
+/**
+ * Accessible wrapper for adding icons to a component.
+ * Defaults to the external icon.
+ */
+export function WithIcon({
+  icon = <i className="fa fa-external-link" aria-hidden="true" />,
+  alt = "(opens in a new tab)",
+  children,
+  ...props
+}: Partial<WithIconProps> & React.ComponentProps<"span">) {
+  return (
+    <span {...props}>
+      {children}
+      {children && " "}
+      {icon} <span className="sr-only">{alt}</span>
+    </span>
+  );
+}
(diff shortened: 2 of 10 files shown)


client/app/components/WithIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+          {icon} <span className="sr-only">{alt}</span>
+        </Component>
+      );
+    }
+  };
+}
+
+/**
+ * Accessible wrapper for adding icons to a component.
+ * Defaults to the external icon.
```
rafawendel, line 26: This is the most controversial part of these changes. I found it practical to have this available. @thielium please let me know your thoughts on this.

client/app/components/WithIcon.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+
+export interface WithIconProps {
+  children?: string;
+  icon: JSX.Element;
+  alt: string;
+}
+
+export function withIcon<ComponentProps extends object>(Component: React.ComponentType<ComponentProps>) {
```
rafawendel, line 9: The proposal of this component is to become the standard approach to icons, as it is accessible by default. Soon this will be documented.

client/app/components/Link.tsx
```diff
@@ -1,41 +1,28 @@
...
-function LinkWithIcon({ icon, alt, children, ...props }: LinkWithIconProps) {
-  return (
-    <Link.Component {...props}>
-      {children} {icon} <span className="sr-only">{alt}</span>
-    </Link.Component>
-  );
-}
-
-Link.WithIcon = LinkWithIcon;
+Link.WithIcon = withIcon(Link.Component);
```
kravets-levko, line 19: There is a little problem with this line. `Link.Component` is used to override link behavior. Extensions can do sth like:

```jsx
import Link from "...";

function CustomLinkComponent(...) {
  return ...;
}

Link.Component = CustomLinkComponent;
```

But since `Link.WithIcon` is initialized when `Link` component is imported - `Link.WithIcon` will capture default `Link.Component` and overriding it will not have any effect when using `Link.WithIcon`

client/app/components/WithIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+  ...props
+}: Partial<WithIconProps> & React.ComponentProps<"span">) {
+  return (
+    <span {...props}>
+      {children}
+      {children && " "}
+      {icon} <span className="sr-only">{alt}</span>
+    </span>
+  );
+}
```
kravets-levko, line 41: I feel a bit uncomfortable because of this duplicated logic. Could we reuse `withIcon` HoC here? Sth like:

```jsx 
export const WithIcon = withIcon((props: React.PropsWithChildren<React.ComponentProps<"span">>) => {
  return <span {...props} />;
});
```

client/app/pages/queries/VisualizationEmbed.jsx
```diff
@@ -120,8 +121,7 @@ function VisualizationEmbedFooter({
         <span className="hidden-print">
           <Tooltip title="Open in Redash">
             <Link.Button className="icon-button" href={queryUrl} target="_blank">
-              <i className="fa fa-external-link" aria-hidden="true" />
-              <span className="sr-only">Open in Redash</span>
+              <WithIcon alt="Open in Redash" />
```
kravets-levko, line 124: Those usages all look a little bit weird: you think that it's just some generic component that add (some) icon. However, it's hard to guess that by default it adds icon for external links and uses `alt` attribute to add some extra text. Currently I don't have any good suggestions; maybe, just rename this component to better reflect its purpose?

client/app/pages/queries/VisualizationEmbed.jsx
```diff
@@ -120,8 +121,7 @@ function VisualizationEmbedFooter({
         <span className="hidden-print">
           <Tooltip title="Open in Redash">
             <Link.Button className="icon-button" href={queryUrl} target="_blank">
```
thielium, line 123: Just a note: When we have a design system, I don't think we'll want link buttons with link icons -- I think we'll want to standardize this. No action for now.

client/app/components/WithIcon.tsx
thielium: What's the advantage of having

```
const Component = withIcon(MyComponent);
return <Component {...props} />
```
versus
```
return <IconText wrapper={Component} icon={MyIcon} {...props}>;
```

where `wrapper` is `span` by default?

Ant's got some icon text stuff (not exactly this) -- might be nice if our approach roughly matched theirs, as long as we're using ant.

https://ant.design/components/typography/

thielium (commented):
Main refactoring looks great! Want to double check that `withIcon` is the way we want to handle icons in general, though.

rafawendel:
@thielium really liked the idea, and implemented it :D

client/app/components/IconText.tsx
```diff
@@ -0,0 +1,34 @@
...
+} & (WrapperType extends string ? React.HTMLAttributes<HTMLElement> : WrapperProps);
+
+export function IconText<WrapperProps>({
+  icon,
+  alt,
+  children,
+  wrapper = "span",
+  ...props
+}: IconTextProps<WrapperProps>) {
+  return React.createElement(
```
thielium, line 17: Can we instantiate this with JSX?

thielium (commented):
Why do we need a special `Link.External` component? I woudn't expect to have to know in advanced where my link is going. 

For example, if I have a table, and one of those columns renders links, I'd have to inspect each URL to figure out if it were internal or external. I'd prefer it if the component were smart enough to change its behavior based on the link content, since that's how an `a` tag behaves. In particular, if an `a` tag points to a `#` ref, it'll just update the URL bar, if it points to another link, it'll navigate the browser. By analogy, a `Link` component can determine what to do if it's an internal or external link.

rafawendel:
> Why do we need a special `Link.External` component? I woudn't expect to have to know in advanced where my link is going.
> 
> For example, if I have a table, and one of those columns renders links, I'd have to inspect each URL to figure out if it were internal or external. I'd prefer it if the component were smart enough to change its behavior based on the link content, since that's how an `a` tag behaves. In particular, if an `a` tag points to a `#` ref, it'll just update the URL bar, if it points to another link, it'll navigate the browser. By analogy, a `Link` component can determine what to do if it's an internal or external link.

I didn't do such think because it sort of becomes too "magical". I'm not fully opposed to it, but I think we're better off leaving the decision for the developer, like in Ant Typography. The thing is, despite the component being called `ExternalLink`, it is actually not necessarily external (some of them open in new tabs, others in drawers), but the one with the particular "external" icon.
The reason for adding such component is having all those properties set by default, like `noopener noreferrer`, which are easily forgotten and having visual consistency across links that lead to external sources.

thielium:
Talked offline -- the issue is less related to determining where the URL is going to, and more about ensuring a consistent and accessible style for non-user-generated links.

While we should centralize our handling for these external-sensitive props, the focus on this implementation is on making our internal usage consistent.

I think calling it `ExternalIconLink` and then deferring global handling of rels, etc. is appropriate.

client/app/components/dashboards/TextboxDialog.jsx
```diff
@@ -80,12 +80,9 @@ function TextboxDialog({ dialog, isNew, ...props }) {
...
           Supports basic{" "}
-          <Link
-            target="_blank"
-            rel="noopener noreferrer"
-            href="https://www.markdownguide.org/cheat-sheet/#basic-syntax">
-            <Tooltip title="Markdown guide opens in new window">Markdown</Tooltip>
-          </Link>
+          <ExternalIconLink href="https://www.markdownguide.org/cheat-sheet/#basic-syntax">
+            <Tooltip title="Markdown guide">Markdown</Tooltip>
+          </ExternalIconLink>
```
kravets-levko, line 85: Maybe let's move ExternalIconLink inside of Tooltip, WDYT?

kravets-levko (approved):
Looks good, just a small nit :+1:

PR #5473, opened 2021-04-14
Fix bug that redshift late binding views created from cross database query does not show up in schema browser

rzrz123:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Fix problem that when you have a late binding view created from cross database query, it won't show up in schema browser.

- The current query use `svv_columns` which does not have columns info on these views.
- We will grab from `svv_tables` for these kind of views. Since it is aws problem, we cannot find columns info anywhere.

Way to reproduce the PR:
1. Redshift cluster have two db `dev` and `staging` with R class node ( this allows you to run cross db query)
2. Create a sample `table1` in `dev` as master user
3. Switch to `staging` db. Create a `user1` and Create a late binding view in `public` with 
```sql
create view "test" as ( select * from dev.public.table1) with no schema binding;
grant usage on schema public to  user1;
grant select on table test to user1;
```
4. Log in `staging` db as `user1`, you are able to run query on the view `test`.
5. But you won't get any columns info from `svv_columns`

AWS related docs: https://docs.aws.amazon.com/redshift/latest/dg/cross-database_usage.html 
 - They are saying `SVV_ALL* or SVV_REDSHIFT*` would work. But it is only `SVV_ALL_TABLES` and `SVV_ALL_SCHEMAS` works. 
 - `SVV_ALL_COLUMNS` does not work and return error `Error running query: relation "your_late_binding_view_name" does not exist`


susodapop (approved):
LGTM Thanks!

guidopetri:
@justinclift you good with merging this if it passes CI?

guidopetri:
Ah, looks like we are not passing unfortunately.

@rzrz123 , would you mind fixing the issues with backend-lint?

justinclift:
Looks like a formatting issue from our backend lint test.  We can probably fix that ourselves. :smile:

PR #5475, opened 2021-04-16
Add eslint plugin import

rafawendel:

- [X] Feature

## Description

This has three aims:

 - [X] Add `eslint-plugin-import` default ruleset and apply fixes
 - [X] Add a plugin for alias enforcement and apply fixes
 - [x] Add `import/order` rule and apply fixes

P.S.: will only merge this after conflicting PRs are merged.


client/app/components/dynamic-form/fields/CheckboxField.jsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import Checkbox from "antd/lib/checkbox";
-import getFieldLabel from "../getFieldLabel";
+import getFieldLabel from '@/components/dynamic-form/getFieldLabel';
```
kravets-levko, line 3: Basically, in this case I'd prefer using relative import. These components are in fact a part of a single, bigger one, and those smaller ones shouldn't (and aren't) used separately. @elado WDYT?

rafawendel: There are a lot of cases relative imports look better IMO. For Eslint, though, is either yay yay or no no. The idea behind such linter rules is enforce a tree-like file structure.

I pushed the changes so that I got your opinion on this. Maybe we can remove it. The main advantage here is autofix that is good for most cases.

elado: Also prefer relative imports when they are not crazy `../../../`.

elado: The reason is that it's easier to move components around without breaking imports. True - VSCode has update imports feature, but that wouldn't be necessary with relative ones.
Nested files don't need to know where they are in the tree, they behave like a single unit together with other files in the same folder.

thielium: Can we just disable the relative imports bit, but keep the other linting rules `import` provides? It sounds like we don't have consensus, here, so we shouldn't start enforcement, IMHO.

rafawendel: That's just what I did. There are actually two plugins in play here: prefer aliases is the one we want to hold off.

thielium:
Could we:
1. Document which configuration we want to use and why (i.e. how we reached consensus);
2. Document how we apply all the changes (presumably `lint --fix`)
3. (Only if we expect merge conflicts) chunk it up?

If we have those three things in place, then this has a ✅ from me!

client/app/components/EmailSettingsWarning.jsx
```diff
@@ -1,10 +1,11 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { clientConfig, currentUser } from "@/services/auth";
-import Tooltip from "@/components/Tooltip";
 import Alert from "antd/lib/alert";
-import HelpTrigger from "@/components/HelpTrigger";
+
 import { useUniqueId } from "@/lib/hooks/useUniqueId";
+import Tooltip from "@/components/Tooltip";
```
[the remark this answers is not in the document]
rafawendel: Everything is possible =D

rafawendel:
> Yeah, I like how the code looks now! One more thing that crossed my mind: when there are no a lot of imports, or they all are split into very small groups (1-2 entries) - can we remove whitespaces between imports to make them more compact? I'm not sure about the exact condition, as well as if that's possible at all, I'm not even sure if that's worth is. At this point that's just an idea to discuss 🙂

I'm afraid the only way this would be possible is not enforcing spaces and leaving it for the developer to decide. WDYT?

kravets-levko:
> I'm afraid the only way this would be possible is not enforcing spaces and leaving it for the developer to decide. WDYT?

No, I'm for automation then 🤖

PR #5490, opened 2021-05-05
SFS-001: Adding support for the optional host connection property

BenSatori:

- [x] Feature

## Description
Adding support for the optional "host" property for a Snowflake connection. This is used when there is a custom hostname, as per the Snowflake Python connector documentation.

As per my conversation with Jesse :)

image

### DIFF
--- a/redash/query_runner/snowflake.py
+++ b/redash/query_runner/snowflake.py
@@ -1,6 +1,5 @@
 try:
     import snowflake.connector
-
     enabled = True
 except ImportError:
     enabled = False
@@ -45,8 +44,9 @@ class Snowflake(BaseQueryRunner):
                 "warehouse": {"type": "string"},
                 "database": {"type": "string"},
                 "region": {"type": "string", "default": "us-west"},
+                "host": {"type": "string"},
             },
-            "order": ["account", "user", "password", "warehouse", "database", "region"],
+            "order": ["account", "user", "password", "warehouse", "database", "region", "host"],
             "required": ["user", "password", "account", "database", "warehouse"],
             "secret": ["password"],
         }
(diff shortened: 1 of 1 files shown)


redash/query_runner/snowflake.py
```diff
@@ -45,8 +44,9 @@ def configuration_schema(cls):
                 "warehouse": {"type": "string"},
                 "database": {"type": "string"},
                 "region": {"type": "string", "default": "us-west"},
+                "host": {"type": "string"},
```
arikfr, line 47: Can we make it an "extra" option? We added this option at some point, which will hide less frequently used settings under "Additional Settings" toggle when setting up the data source?

Here's an example for Postgres:
CleanShot 2021-05-05 at 16 53 11@2x

And how it looks like in code:
https://github.com/getredash/redash/blob/9c8c1bfa9ade90eb147bfcda75e5f9e95b691e42/redash/query_runner/pg.py#L173-L178

redash/query_runner/snowflake.py
```diff
@@ -64,16 +64,27 @@ def determine_type(cls, data_type, scale):
...
         if region == "us-west":
             region = None
 
+        if self.configuration.__contains__("host"):
+            host = self.configuration.get("host")
+        else:
+            if region:
+                host = "{}.{}.snowflakecomputing.com".format(account, region)
+            else:
+                host = "{}.snowflakecomputing.com".format(account, region)
```
arikfr, line 79: You're passing two values but using only one?

arikfr (commented):
Thank you for taking a stab at this!

redash/query_runner/snowflake.py
BenSatori: Fixed

redash/query_runner/snowflake.py
BenSatori: Done. I checked locally and it looks this way now:

image

I also took the liberty to move the region under the account, as it's a better fit UX-wise IMO.

I also think that some of the logic in this component can be improved. I'll msg you about it, and can perhaps also do the changes.

BenSatori:
@arikfr let me know if anything further is needed for this commit

arikfr (approved):
Thanks!

PR #5512, opened 2021-06-09
Improve QR serialized performance to xlsx

zhongjiajie:
According to: https://gist.github.com/acdha/4976b9831224d2713997
we found out pyexcelerate xlsx writer pref better
than xlsxwriter and it have bulk load mode

- [x] Refactor

## Description

According to: https://gist.github.com/acdha/4976b9831224d2713997 and my test in our company env, we found out `pyexcelerate` xlsx writer pref better than `xlsxwriter` and it have bulk load mode, so I think we should use `pyexcelerate`

### DIFF
--- a/redash/serializers/query_result.py
+++ b/redash/serializers/query_result.py
@@ -1,9 +1,8 @@
 import io
 import csv
-import xlsxwriter
+from pyexcelerate import Workbook
 from funcy import rpartial, project
 from dateutil.parser import isoparse as parse_date
-from redash.utils import json_loads, UnicodeWriter
 from redash.query_runner import TYPE_BOOLEAN, TYPE_DATE, TYPE_DATETIME
 from redash.authentication.org_resolving import current_org
 
@@ -102,22 +101,22 @@ def serialize_query_result_to_dsv(query_result, delimiter):
 def serialize_query_result_to_xlsx(query_result):
     output = io.BytesIO()
 
+    wb_data = []
     query_data = query_result.data
-    book = xlsxwriter.Workbook(output, {"constant_memory": True})
-    sheet = book.add_worksheet("result")
 
-    column_names = []
-    for c, col in enumerate(query_data["columns"]):
-        sheet.write(0, c, col["name"])
-        column_names.append(col["name"])
+    column_names = [col['name'] for col in query_data['columns']]
+    wb_data.append(column_names)
 
-    for r, row in enumerate(query_data["rows"]):
-        for c, name in enumerate(column_names):
-            v = row.get(name)
-            if isinstance(v, (dict, list)):
-                v = str(v)
-            sheet.write(r + 1, c, v)
+    for row in query_data['rows']:
+        row_data = [
+            str(row.get(name))
+            if isinstance(row.get(name), (dict, list)) else row.get(name)
+            for name in column_names
+        ]
+        wb_data.append(row_data)
 
-    book.close()
+    wb = Workbook()
+    wb.new_sheet("result", data=wb_data)
+    wb.save(output)
 
     return output.getvalue()
--- a/requirements.txt
+++ b/requirements.txt
@@ -61,6 +61,7 @@ sshtunnel==0.1.5
 supervisor==4.1.0
 supervisor_checks==0.8.1
 werkzeug==0.16.1
+pyexcelerate==0.10.0
 # Install the dependencies of the bin/bundle-extensions script here.
 # It has its own requirements file to simplify the frontend client build process
 -r requirements_bundles.txt


redash/serializers/query_result.py
```diff
@@ -1,9 +1,8 @@
 import io
 import csv
-import xlsxwriter
+from pyexcelerate import Workbook
 from funcy import rpartial, project
 from dateutil.parser import isoparse as parse_date
-from redash.utils import json_loads, UnicodeWriter
```
zhongjiajie, line 6: Delete not use import package here

redash/serializers/query_result.py
```diff
@@ -102,22 +101,22 @@ def serialize_query_result_to_dsv(query_result, delimiter):
...
+        row_data = [
+            str(row.get(name))
+            if isinstance(row.get(name), (dict, list)) else row.get(name)
+            for name in column_names
+        ]
+        wb_data.append(row_data)
+
+    wb = Workbook()
+    wb.new_sheet("result", data=wb_data)
+    wb.save(output)
```
zhongjiajie, line 120: Using bulk load mode is better than `xlsxwriter.write` for each cell

zhongjiajie:
CI pass, PTAL @arikfr @rauchy @susodapop thanks

konnectr:
@zhongjiajie  thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

zhongjiajie:
hi @konnectr , I would like to continue this pr, BTW, I have others' pr pending for long day, I can rebase those too, and could you see them too?

justinclift:
Yep.  We're going through all of the PR's, so please rebase everything you're still interested in + let us know when it's ready (so we get to them faster). :smile:

konnectr:
To what extent can python 2.7 tests be interpreted in python 3.8?

justinclift:
If it helps, the minimum Python version we're targeting is Python 3.8.

konnectr:
i'm going run test, 
```
Versions:
    python      : 3.8.1
    openpyxl    : 3.1.2
    pyexcelerate: 0.10.0
    xlsxwriter  : 3.1.2
    xlwt        : 1.3.0

Dimensions:
    Rows = 600000
    Cols = 50

Times:
    pyexcelerate          : 109.20
row index was 65536, not allowed by .xls format
    xlsxwriter (optimised): 117.78
    xlsxwriter            : 149.70
```
Our version xlsxwriter (optimised)  -> proposed version pyexcelerate

justinclift:
Cool.  Sounds like this is about 8% (rough estimate) faster then yeah?

So, we should merge it once the PR is in a good state again. :smile:

konnectr:
> Cool. Sounds like this is about 8% (rough estimate) faster then yeah?
> 
> So, we should merge it once the PR is in a good state again. 😄

I would say that this is a meaningless test that shows that it works faster by 600 thousand lines. But nothing will change for use on a smaller amount of data.

justinclift:
Looking at the release history for both of the Python projects, it seems like `xlsxwriter` [is still actively developed](https://pypi.org/project/XlsxWriter/#history), whereas `pyexcelerate` [hasn't had a release in about 2 and a half years](https://pypi.org/project/PyExcelerate/#history).

With that in mind, my feeling is we should drop this PR.  It's introducing a new dependency that's dead/unmaintained, for very little benefit.

zhongjiajie:
> Looking at the release history for both of the Python projects, it seems like `xlsxwriter` [is still actively developed](https://pypi.org/project/XlsxWriter/#history), whereas `pyexcelerate` [hasn't had a release in about 2 and a half years](https://pypi.org/project/PyExcelerate/#history).
> 
> With that in mind, my feeling is we should drop this PR. It's introducing a new dependency that's dead/unmaintained, for very little benefit.

you are right, I forget check about this

zhongjiajie:
well the question become, should or should not add dependency make process faster 30% but package less of maintaining. For my personally if the package is work and can archive our aim, I think we can use it. 

In my daily work, I have try to use some code unreleased https://github.com/apache/dolphinscheduler-sdk-python/blob/d1e25ff1dbacfdbc8eb7d13992d67b73ddd62e59/setup.cfg#L117-L120 to get the goal of mine, although each time      I have to change `setup.cfg` when release the package but I still use it, it is save a lot of my time for the package

zhongjiajie:
Of cause they have another way, maintaining the package by ourself but I think it is not the time we do that now

justinclift:
> should or should not add dependency make process faster 30% but package less of maintaining

To me, that seems like a good question. :smile:

I think that *if* the "30% faster" was in some part of our code which is super, super, super important and gets used a lot, then it might be worth it.

For a situation where the code is very rarely used, then it's probably *not* worth the extra risk.

With that, I'd like to close this PR because it doesn't seem to be "30% faster" + code that is used a lot.

Is that ok with yourselves? :smile:

zhongjiajie:
Get it, will close this PR

justinclift:
Thanks @zhongjiajie. :smile:

PR #5514, opened 2021-06-13
Update Slack Alert Destination Configuration Wizard

rauchy:
- [x] Bug Fix

## Description

Slack has updated its means of accepting Webhooks. You now create an app, allow for incoming webhooks, then create a url per channel.

Our alert configuration, allows / encourages users to designate a channel there, but that is fully ignored by Slack so we should remove it (ie if I create a URL for #channel2, even if i enter #channel2 in the alert destination configuration wizard, messages will still go to #channel1).

Bot icon, name etc are also now controlled on the Slack side through app settings so we should cut those fields from our config UI.

image

### DIFF
--- a/redash/destinations/slack.py
+++ b/redash/destinations/slack.py
@@ -12,12 +12,8 @@ class Slack(BaseDestination):
             "type": "object",
             "properties": {
                 "url": {"type": "string", "title": "Slack Webhook URL"},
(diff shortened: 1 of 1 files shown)


susodapop (approved):
LGTM. Tested on a Slack workspace and it works great.

PR #5518, opened 2021-06-17
Update Slack Alert Destination Configuration Wizard: Support legacy webhooks

rauchy:
- [x] Bug Fix

## Description
Following up on #5514, this PR will keep the webhook call the same in case it was created with a legacy webhook.

#5514

### DIFF
--- a/redash/destinations/slack.py
+++ b/redash/destinations/slack.py
@@ -49,6 +49,15 @@ class Slack(BaseDestination):
 
         payload = {"attachments": [{"text": text, "color": color, "fields": fields}]}
 
+        if options.get("username"):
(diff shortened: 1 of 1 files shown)


susodapop (approved):

guidopetri:
@justinclift similar here, are you good with merging if it passes CI?

guidopetri:
Looks like we need to add a test.

guidopetri:
Reading more into this, I don't think we want this change - it's to support legacy webhooks for Slack, but Slack encourages users to migrate to new webhooks anyhow. Closing as a result.

justinclift:
Cool. :smile:

PR #5555, opened 2021-08-03
Added New Env variable for Unpublished queries

adithyapathipaka:

- [x] Feature

## Description

Added new Env Variable REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES and updated include_drafts param to get value from REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES

1. Application shows all the unpublished queries from other users on the UI by default, which is due to Hard coded value with `True` for the include_draft param.
2. Added new Env Variable `REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES` (boolean) through which application can be configured to view unpublished queries
3. By Default `REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES`  will be `true` and can be configure through to `false` by adding the variable to env file.


adithyapathipaka (commented):
Added the Env Variable to Docker Compose file

susodapop:
Thanks for your contribution. This looks helpful. We will review this (and hope to merge) shortly after the V10 release later this month. Right now we're dedicating resources to getting the V10 release out the door (currently in beta).

PR #5566, opened 2021-08-17
Reorganize frontend code into Yarn workspaces

kravets-levko:

- [x] Refactor
- [x] Other


guidopetri:
@justinclift this seems like a substantial PR, I don't think we'd want to close this one but I personally have no idea what Yarn workspaces are.

@eradman have you got an opinion on what we should do with this PR?

justinclift:
Yeah, no idea about yarn workspaces either yet.  Sounds like something we'll probably want to at least learn about to see if they're a good fit. :smile:

gabrieldutra:
> this seems like a substantial PR, I don't think we'd want to close this one but I personally have no idea what Yarn workspaces are.

At this point I'm more leaning towards suggesting to merge `viz-lib` back into the main package instead of building the multiple packages setup with workspaces. Reason being: if there's no clear benefit in having visualizations as it's own package on NPM, it simplifies things a lot to maintain just the Redash webapp.

justinclift:
Cool.  That's a pretty unanimous "lets not go with yarn workspaces" (at least for now) so I'll close this. :smile:

PR #5568, opened 2021-08-17
Fix: Edit Source button disappeared for users without CanEdit perms

### LINKED ISSUE #5560: Edit Source button has disappeared
### Issue Summary

The **Edit Source** button on the query screen no longer appears unless the user has edit permission for the query. This is a regression from V8, where users without edit permission could still view the query source (but not save it). Only the button is missing. I can still view the source by manually visiting `/source` in the URL bar.

### Steps to Reproduce

1. Visit a query you don't own
2. Observe that the Edit Source button does not appear on the query page header

I think this was changed unintentionally in https://github.com/getredash/redash/pull/4670 by adding a check for `canEdit` permission on line 182 of `QueryPageHeader.jsx`. Before we only checked for `canViewSource`.

### Technical details:

* Redash Version: v10 beta
* Browser/OS: Chrome

susodapop:

- [x] Bug Fix

## Description

This fixes a regression in the V10 beta where the `Edit Source` button on the query screen would not appear for users without `canEdit` permission on the query. These users should only need `canViewSource` permission.

Closes #5560

### Before
CleanShot 2021-08-17 at 14 21 33

### After

CleanShot 2021-08-17 at 14 21 39

### DIFF
--- a/client/app/pages/queries/components/QueryPageHeader.jsx
+++ b/client/app/pages/queries/components/QueryPageHeader.jsx
@@ -179,7 +179,7 @@ export default function QueryPageHeader({
 
         {!queryFlags.isNew && queryFlags.canViewSource && (
           <span>
-            {!sourceMode && queryFlags.canEdit && (
+            {!sourceMode && (
               <Link.Button className="m-r-5" href={query.getUrl(true, selectedVisualization)}>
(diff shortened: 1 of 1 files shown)


arikfr (approved):

shperl (approved):
this lgtm! debyani the full stack developer!

kravets-levko (approved):

Issue #5550, opened 2021-07-29
Cass Data Source - Add option to provide SSLContext


anirudhbagri:
We have a production cassandra hosted with client-cert authentication method.
This is how we connect to cassandra using python client:
```python
import ssl
ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
ssl_context.load_cert_chain(
    "mycert.cer",
    "mykey.key"
)
ssl_context
ssl_context.load_verify_locations("ca-root.crt")
ssl_context.verify_mode = ssl.CERT_REQUIRED
clust = cluster.Cluster([
        "cass_host"
    ], ssl_context=ssl_context, connect_timeout=60)
session = clust.connect("<keyspace>")
```

How can I connect to this DB using redash?

susodapop:
If the query runner doesn't support this you can open a PR to implement it. The core team does not use Cassandra so must rely on the community for fixes of this nature.

anirudhbagri:
Hi @susodapop, 

I have raised a PR for this implementation.
 
https://github.com/getredash/redash/pull/5575

PR #5575, opened 2021-08-20
Add support for passing sslContext cass

anirudhbagri:

- [x] Feature

## Description

Added support for CASS data source to accept client cert and key for authentication using client_cert_mechanism

getredash/redash#5550


susodapop (commented):
Looks great. Can you please add some screenshots to this PR showing how a user would paste their certificate file into Redash? Is that a normal way to enter certs?

anirudhbagri:
@susodapop, users would upload the file similar to how they upload the SSL Certificate. 
But , I am not sure how to make UI accept a file instead of a string? 
I compared it with how SSL certificate part code was written.

susodapop:
> But , I am not sure how to make UI accept a file instead of a string?

Sorry this is a bit confusing. Our `dynamicFormHelper` does some magic to figure out which fields should be file widgets.

See the code here:

https://github.com/getredash/redash/blob/d8d7c78992e44a4b6d7fdd4c39ccc1c35cd8c7a9/client/app/components/dynamic-form/dynamicFormHelper.js#L38-L40

To display an upload button instead of a text field you need the field name to end with `file`.

guidopetri:
@anirudhbagri , thanks for the PR! Same story here - we've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
Just attempted to manually fix the merge conflict using the web interface.

Looks like I mucked it up though. :frowning:

```
./redash/query_runner/cass.py:140:66: E999 SyntaxError: expected ':'
```

justinclift:
Looks like the breakage wasn't my doing.  The original commit seems to have some errors as well.  I've just pushed my best guess at how to fix them, but I have no idea if they're actually reasonable. :smile:

PR #5577, opened 2021-08-31
Fix: date type values are displayed as int type

iwakiriK:

- [x] Bug Fix

## Description
For values of date type, if `moment.utc()` is executed and columnType is string, it will be displayed as int type.

image

This PR will fix this problem.

https://momentjs.com/docs/#/parsing/utc/


guidopetri:
@iwakiriK , thanks for the PR! Same story here, we've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

justinclift:
Looks like it's failing our frontend lint tests. :frowning:

eradman:
@iwakiriK can you illustrate the problem this PR solves using a simpler query?

Datetimes are interpreted correctly on 10.1.0 (which uses numeral.js) and 23.09.0-dev (which uses numbro.js)

datetime

eradman:
Re-tested this with 24.07.0-dev and have not observed a problem. Closing, but feel free to re-open with more information

Issue #5333, opened 2020-12-27
SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds


dengc367:
### Issue Summary

A summary of the issue and the browser/OS environment in which it occurs.

* Safari: 14.0.1 on Mac BigSur
* Safari on iPhone iOS

when I open the redash frontend on safari browser, It says: ** It seems like we encountered an error. Try refreshing this page or contact your administrator.**, and i found some errors in safari console. the error screenshot is below.
![image](https://user-images.githubusercontent.com/1353036/103165572-a5243b80-4854-11eb-9c57-1df6f44d16a9.png)

### Steps to Reproduce

1. This is the first step
2. This is the second step, etc.

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* Redash Version:
 the latest commit of redash github master. 
* Browser/OS:
MacOS / IOS safari.
* How did you install Redash:
I use the docker-compose from [redash/setup](https://github.com/redash/setup) scripts.

dengc367:
Is it need the `throttle-debounce` for throttling update the states, like the [nodejs issue] (https://github.com/nodejs/nodejs.dev/pull/924/files) ?

PR #5586, opened 2021-09-10
Fix: Unable to open large Dashboard in Safari browser

iwakiriK:
- [x] Bug Fix

## Description
Due to a SecurityError, we are unable to open large dashboards in Safari browser.
image

For more information, please see this issue.
https://github.com/getredash/redash/issues/5333

The reason for this is that the value of `parametersChanged` is wrong.
The comparison method is not deep equal, so it almost assigns true.
https://github.com/getredash/redash/blob/v10.0.0-beta/client/app/components/Parameters.jsx#L57

This causes `updateUrl` runs frequently.
As a result, `history.replaceState` runs frequently.
https://github.com/getredash/redash/blob/v10.0.0-beta/client/app/components/Parameters.jsx#L62-L64

### How to Reproduce
- Create a parameter query
- Create a Dashboard
- Put about 40 of the queries in the Dashboard and save them.
- Open the Dashboard in Safari and wait for 10 seconds.

https://github.com/getredash/redash/issues/5333


jimsparkman:
Tested this in my environment, fixes the Safari bug for me!

susodapop (approved):
LGTM. I rebased onto master so we can re-run the circleci build and then merge 😄

susodapop:
After some manual testing there appears to be an issue with this implementation. Please review and add a Cypress test for it as well:

Steps to reproduce the issue:
1. Create a query that returns a single column (this will be used as a backing query for a dropdown)
2. Create a second query with a multi-parameter that uses the results from step 1.
3. Execute the query from step 2 with some options selected.
4. Change the settings for this multi-dropdown parameter (for example, change from wrapping with single quotes to no wrap)
5. Change the selected options and execute the query again. The result does not update properly until you reload the page.

CleanShot 2021-12-22 at 12 49 31

susodapop (changes requested):
After some manual testing there appears to be an issue with this implementation. Please review and add a Cypress test for it as well:

Steps to reproduce the issue:
1. Create a query that returns a single column (this will be used as a backing query for a dropdown)
2. Create a second query with a multi-parameter that uses the results from step 1.
3. Execute the query from step 2 with some options selected.
4. Change the settings for this multi-dropdown parameter (for example, change from wrapping with single quotes to no wrap)
5. Change the selected options and execute the query again. The result does not update properly until you reload the page.

CleanShot 2021-12-22 at 12 49 31

InvalidPointer:
Encountered this bug. @iwakiriK @susodapop any news regarding fix?

guidopetri:
@iwakiriK , thanks for the PR! Same here, we've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts? We should also address the feedback already given.

justinclift:
Rebased this onto `master` branch.  Sounds like this PR is only a partial fix though, as per @susodapop's comment + screen recording above.

Issue #5590, opened 2021-09-13
Redash queries fail when querying BigQuery table with row-level access controls


tleavitt:
### Issue Summary

BigQuery recently added row-level access controls on tables: https://cloud.google.com/bigquery/docs/row-level-security-intro

When querying a BQ table with row-level ACLs in Redash, the redash query fails. This seems to happen because BigQuery [hides the number of bytes processed](https://cloud.google.com/bigquery/docs/best-practices-row-level-security) on queries to tables with row-level ACLs, and redash [expects](https://github.com/getredash/redash/blob/5085495dd4564058b35e2a424dc62dd9dabe02d6/redash/query_runner/big_query.py#L240) the totalBytesProcessed field to be set on the BQ query response.

### Steps to Reproduce

1. Create a BigQuery table with row-level ACLs enabled:
```
CREATE OR REPLACE TABLE `my_dataset.acl_test`
AS (
  SELECT "visible" as label UNION ALL SELECT "visible" as label UNION ALL SELECT "invisible" as label
);

CREATE OR REPLACE ROW ACCESS POLICY acl ON  `my_dataset.acl_test`
GRANT TO ("serviceAccount:redash@myproject.iam.gserviceaccount.com")
FILTER USING (TRUE);

CREATE OR REPLACE ROW ACCESS POLICY general ON  `my_dataset.acl_test`
GRANT TO ("allAuthenticatedUsers")
FILTER USING ("visible" = label);
```
2. Query the table in Redash:
<img width="1379" alt="Screen Shot 2021-09-13 at 1 10 43 PM" src="https://user-images.githubusercontent.com/1909854/133144991-5d94ebdb-1653-4582-af4f-f271d6266e74.png">

The query fails with the message: `Error running query: 'totalBytesProcessed'`

This is breaking several of our redash queries. I'd be happy to open a pull request to fix the issue - it seems like the fix should be making access to `query_reply["totalBytesProcessed"]` more defensive in https://github.com/getredash/redash/blob/5085495dd4564058b35e2a424dc62dd9dabe02d6/redash/query_runner/big_query.py

### Technical details:

* Redash Version: 7b026a19
* Browser/OS: Chrome 93.0.4577.63  on macOS Catalina
* How did you install Redash: hosted redash on app.redash.io

PR #5592, opened 2021-09-17
Guard against empty totalProcessedBytes in BigQuery responses

tleavitt:

- [x] Bug Fix

## Description
BigQuery recently added row-level access controls on tables: https://cloud.google.com/bigquery/docs/row-level-security-intro

When querying a BQ table with row-level ACLs in Redash, the redash query fails. This happens because BigQuery [hides the number of bytes processed](https://cloud.google.com/bigquery/docs/best-practices-row-level-security) on queries to tables with row-level ACLs, and redash [expects](https://github.com/getredash/redash/blob/5085495dd4564058b35e2a424dc62dd9dabe02d6/redash/query_runner/big_query.py#L240) the totalBytesProcessed field to be set on the BQ query response.

This change resolves the issue by creating a getter with a default for the "totalBytesProcessed" field on BigQuery responses that returns 0 when the field is absent.

Resolves https://github.com/getredash/redash/issues/5590

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -83,6 +83,13 @@ def _get_query_results(jobs, project_id, location, job_id, start_index):
     return query_reply
 
 
+def _get_total_bytes_processed_for_resp(bq_response):
+    # BigQuery hides the total bytes processed for queries
+    # to talbes with row-level access controls. Because of this
+    # the "totalBytesProcessed" field may or may not be defined on the response.
+    return int(bq_response.get("totalBytesProcessed", "0"))
+
+
 class BigQuery(BaseQueryRunner):
     should_annotate_query = False
     noop_query = "SELECT 1"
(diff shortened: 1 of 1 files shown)


redash/query_runner/big_query.py
```diff
@@ -83,6 +83,13 @@ def _get_query_results(jobs, project_id, location, job_id, start_index):
     return query_reply
 
 
+def _get_total_bytes_processed_for_resp(bq_response):
+    # BigQuery hides the total bytes processed for queries
+    # to talbes with row-level access controls. Because of this
+    # the "totalBytesProcessed" field may or may not be defined on the response.
```
susodapop, line 89:
```suggestion
    # BigQuery hides the total bytes processed for queries to tables with row-level access controls.
    # For these queries the "totalBytesProcessed" field may not be defined in the response.
```

susodapop (approved):
LGTM. One small change to clarify this comment.

tleavitt:
@susodapop  Great! Looks like CI is failing, how can we resolve that to land this PR?

kravets-levko:
@tleavitt we experience some issues with Percy integration. It doesn't block this PR from merging, but we'd like to keep it for a few more hours just to investigate, and then will merge it ASAP

Issue #5593, opened 2021-09-19
Redash server exited with code 139 with docker


andresorrego94:
Hi all 

I am facing  issues trying to run docker-compose up -d 
I tried from the following branches versions : **master** and  **release/10.0.x**

Error: 
redash_server_1 exited with code 139

I tried from ubuntu 20.04, Mac BigSur, and windows 10. 
Does anyone facing the same issue ?
There is no more details in the error than the exited code 139.
Also I run the create db commands but after that the issue still there with the server.

Need help to solve ths issue. Thanks !

MathiasLantean:
Hi @andresorrego94!

I've had a hard time installing redash or running it using docker. Currently, I have the same issue, running the image in macOS Big Sur 11.4 using docker engine 20.10.8 and docker compose 1.29.2.

```
ml@192 redash % docker-compose up        
Starting redash_redis_1    ... done
Starting redash_email_1    ... done
Starting redash_postgres_1 ... done
Starting redash_server_1   ... done
Starting redash_worker_1    ... done
Starting redash_scheduler_1 ... done
Attaching to redash_postgres_1, redash_redis_1, redash_email_1, redash_server_1, redash_scheduler_1, redash_worker_1
email_1      | MailDev webapp running at http://0.0.0.0:80
email_1      | MailDev SMTP Server running at 0.0.0.0:25
postgres_1   | 
postgres_1   | PostgreSQL Database directory appears to contain a database; Skipping initialization
postgres_1   | 
postgres_1   | LOG:  database system was interrupted; last known up at 2021-09-19 21:08:11 UTC
postgres_1   | LOG:  database system was not properly shut down; automatic recovery in progress
postgres_1   | LOG:  invalid record length at 0/1701F70
postgres_1   | LOG:  redo is not required
postgres_1   | LOG:  MultiXact member wraparound protections are now enabled
postgres_1   | LOG:  database system is ready to accept connections
postgres_1   | LOG:  autovacuum launcher started
scheduler_1  | Starting dev RQ scheduler...
redis_1      | 1:C 19 Sep 21:23:41.039 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
redis_1      |                 _._                                                  
redis_1      |            _.-``__ ''-._                                             
redis_1      |       _.-``    `.  `_.  ''-._           Redis 3.2.12 (00000000/0) 64 bit
redis_1      |   .-`` .-```.  ```\/    _.,_ ''-._                                   
[...]

PR #5608, opened 2021-10-01
Fix: Specify the protobuf version

zoomdot:
- [x] Bug Fix

protobuf package with a dependency of google-api-python-client released a new version (3.18.0) on September 16, 2021. Since then, the Docker build is failing, and it is presumed that there is a conflict with other DataSource packages that use protobuf. (phoenixdb, pydgraph)

`redash_server_1 exited with code 139`

To solve the problem, I specify the version of protobuf as 3.17.3.

`protobuf==3.17.3`

#5593

### DIFF
--- a/requirements_all_ds.txt
+++ b/requirements_all_ds.txt
@@ -1,4 +1,5 @@
 google-api-python-client==1.7.11
+protobuf==3.17.3
 gspread==3.1.0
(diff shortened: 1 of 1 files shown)


susodapop (approved):
Thanks!

PR #5615, opened 2021-10-08
Add data type info into schemas of Trino tables

vnnw:
The current get_schema method of Trino query runner is missing table columns' data types information, so these data types won't be seen in Queries editing GUI. After the modification, these info are back.

- [ ] Refactor
- [ ] Feature
- [ ✓ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other


konnectr:
The development of this issue will continue in https://github.com/getredash/redash/pull/6348

PR #5616, opened 2021-10-09
Add Tinybird query runner

pandomic:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

tinybird

[Tinybird](https://www.tinybird.co) is a fast-growing service providing real-time analytics for large amounts of data. Integrating it with powerful Redash tools will open new opportunities for real-time data analysis and visualizations, especially when mixed with other data sources like Python.

Under the hood, Tinybird uses [ClickHouse](https://clickhouse.com), which makes it possible to re-use existing query runner with minimal adjustments.

### Connection Testing

Tinybird has few system data sources that should become queriable once the connection is properly configured. We use one of those (`tinybird.pipe_stats`) to test whether it can be executed or not.

### Data Querying

There are different ways how you may run SQL queries against Tinybird. The most flexible one is of course the `/sql` endpoint which does not enforce you to use a single data source or pipe.

### Schema Fetching

Schema for pipes and data sources can not be easily observed through SQL queries. Thus you may notice a little trick here: we are fetching lists of pipes and data sources and then running a simple query on each of them. Result schema is then returned as `meta` attribute. There is one limitation though, for pipes schema can only be observed if it is exported as an endpoint (because you are restricted to run queries only against shared pipes).

Proactive initiative

The following Loom recording shows the Tinybird query runner at work
https://www.loom.com/share/8d96d361e2534de09710216df528af44


guidopetri:
@pandomic , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI?

pandomic:
Hi @guidopetri, thanks for the update, I will try to align the PR in the next following days 🙌

pandomic:
Hey @guidopetri, I just rebased the PR. Hope it looks good. Looking forward to getting Tinybird support 🙌

justinclift:
@pandomic Looks like it's about 99.95% done, but still now *quite* finished.  It's failing the formatting check lint test:

```
Run black --check .
would reformat /home/runner/work/redash/redash/redash/query_runner/tinybird.py
```

We have instructions on how to get that fixed if that's useful?

https://github.com/getredash/redash/wiki/Local-development-setup

It's the `make format` command there. :smile:

guidopetri:
@pandomic thanks for your patience and for updating the PR! Any chance you could add some tests to make the Codecov bot happy?

redash/query_runner/tinybird.py
```diff
@@ -0,0 +1,126 @@
...
+
+import requests
+
+from redash.query_runner import register
+from redash.query_runner.clickhouse import ClickHouse
+
+logger = logging.getLogger(__name__)
+
+
+class Tinybird(ClickHouse):
```
guidopetri, line 11: To be clear, we intended on using the `Clickhouse.run_query` implementation? It relies on `Clickhouse._clickhouse_query`, so I want to make sure Tinybird and Clickhouse have the same response format.

It also appears that function logs a line saying `Clickhouse is about to execute query`, which would be inaccurate for any Tinybird logs.

tests/query_runner/test_tinybird.py
```diff
@@ -0,0 +1,123 @@
...
+        {"string_attribute": "hello world", "number_attribute": 123, "date_attribute": "2023-01-01 00:00:03.001000"},
+    ],
+    "rows": 1,
+    "statistics": {"elapsed": 0.011556914, "rows_read": 87919, "bytes_read": 17397219},
+}
+
+
+class TestTinybird(TestCase):
+    @patch("requests.get")
+    def test_get_schema(self, get_request):
```
guidopetri, line 54: Both the test cases here are testing downstream functionality (i.e. `get_schema` and `run_query`, that are not directly implemented in the Tinybird runner). One potential improvement would be to test the Tinybird query runner directly, since it uses the base implementation for these methods.

guidopetri (approved):
@pandomic thanks for adding these tests. I left a few comments but this looks good to me. I'll approve and merge tomorrow unless you want to change anything in response to the feedback.

redash/query_runner/tinybird.py
justinclift: Should we wait for the response from @pandomic about this one before merging, just in case it needs changing?

guidopetri: I think we're good either way (because we're supposed to be inheriting Clickhouse anyway). But I think it'd be better to have the logs more accurate :)

justinclift: Cool. :smile:

pandomic: 👍 Tinybird should have the same JSON response format (at least when it comes to data) according to docs:
https://clickhouse.com/docs/en/interfaces/formats#json
https://www.tinybird.co/docs/api-reference/query-api.html

We have been actually using this implementation for almost two years in production now 🙂

redash/query_runner/tinybird.py
```diff
@@ -0,0 +1,126 @@
...
+            "secret": ["token"],
+        }
+
+    @classmethod
+    def name(cls):
+        return "Tinybird"
+
+    @classmethod
+    def type(cls):
+        return "tinybird"
```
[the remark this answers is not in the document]
pandomic: Makes sense 👍 Slightly adjusted the ClickHouse class to re-use the underlying `type()` method

tests/query_runner/test_tinybird.py
pandomic: 👍 I adjusted the test name to reflect what is happening in `get_schema` (it collects the schema from available pipes and datasources).

I am not sure what else to test when it comes to query execution though, as it's essentially just a tiny GET request which is already being tested for proper formatting and with mocked response handling 🤔

justinclift:
Looks like it's passing all of the tests.  Weirdly though, the Code coverage tests don't seem to have automatically-rerun.  Not sure why. :thinking:

guidopetri (approved):
Thanks again for your contribution :) I retriggered the CI but it should be passing.

PR #5621, opened 2021-10-12
Add auth_saml_sp_settings to allow configuring #1669

be30c9:

- [x] Feature

## Description

Adds a new configuration option, auth_saml_sp_settings, to allow arbitrarily configuring the Service Provider (SP) in pysaml2. 

Similar to the closed PR #2208 which allowed configuring SAML signature requirements, but this PR is more general in that the setting is a JSON string, allowing any SP setting to be set or overridden. Set through the environment variable `REDASH_SAML_SP_SETTINGS`, no UI.

For example, to implement #1669, to allow compatibility with many popular SAML IdP services, set to:

```json
{"want_assertions_signed": false, "want_response_signed": true}
```

#1669 (SAML: Don't require signed assertions)
#2208 (feat: make SAML signature requirements configurable)

### DIFF
--- a/redash/authentication/saml_auth.py
+++ b/redash/authentication/saml_auth.py
@@ -29,6 +29,7 @@ def get_saml_client(org):
     sso_url = org.get_setting("auth_saml_sso_url")
     x509_cert = org.get_setting("auth_saml_x509_cert")
     metadata_url = org.get_setting("auth_saml_metadata_url")
+    sp_settings = org.get_setting("auth_saml_sp_settings")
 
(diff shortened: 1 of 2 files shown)


deed02392:
The unsuccessful percy visual changes are just compression/font rendering artefacts.

susodapop (approved):
LGTM. Thanks this is awesome!

deed02392:
@susodapop What's the plan for this getting into a tagged release?

susodapop:
This will be included in V11.

arikfr:

PR #5632, opened 2021-10-22
Speed up BigQuery get schema operation in massively tables environment

katsugeneration:

- [x] Refactor
- [x] Feature

## Description

This implementation enable to speed up get_schema operation in BigQuery environments which have massively tables (about over 1k tables).

In the environment which has about 1k tables , it takes about **10 minutes** by current implementation. Very slow operation caused the operation timeout at default timeout settings. 

The cause of the slow operation is calling get table's column list api per table in current implementation. 
New implementation improve operation time to **10 seconds** in same environment by improving to call one api to get all table's column list.

- [BigQuery  Information Schema Reference](https://cloud.google.com/bigquery/docs/information-schema-tables)

New implementation uses every dataset's INFORMATION_SCHEMA. COLUMNS view to get table's column list in dataset.

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -282,20 +282,23 @@ class BigQuery(BaseQueryRunner):
         """
 
         schema = {}
+        queries = []
         for dataset in datasets.get("datasets", []):
             dataset_id = dataset["datasetReference"]["datasetId"]
             query = query_base.format(dataset_id=dataset_id)
+            queries.append(query)
 
-            results, error = self.run_query(query, None)
-            if error is not None:
-                raise Exception("Failed getting schema.")
+        query = '\nUNION ALL\n'.join(queries)
+        results, error = self.run_query(query, None)
+        if error is not None:
+            raise Exception("Failed getting schema.")
 
-            results = json_loads(results)
-            for row in results["rows"]:
-                table_name = "{0}.{1}".format(row["table_schema"], row["table_name"])
-                if table_name not in schema:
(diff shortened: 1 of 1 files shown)


probably-not:
Any word on this? This can be a gigantic speedup in order to get schema information in huge BQ datasets.

susodapop (commented):
This looks brilliant. Thank you for the contribution.

Since I'm not familiar with BigQuery I have a few questions before we merge it in. First, is this a drop-in replacement for the current implementation? Does it require any different settings or permissions for the service account Redash uses to connect? 

If it's exactly the same but faster that's ideal.

katsugeneration:
@susodapop Thank you for your review. I answered your question.

> First, is this a drop-in replacement for the current implementation?

Yes. New implementation can get same schema information in same format as before about permitted datasets.

> Does it require any different settings or permissions for the service account Redash uses to connect?

No. New implementation require same permissions and settings to get bigquery schema and to run bigquery queries in redash now.
User who only get schema in redash need to add permission 'bigquery.jobs.create' , but this permmission is used for running bigquery queries even now.
So, I think no user changes settings in cause of new implementation.

susodapop:
Sounds like we can change our guidance in our [documentation](https://redash.io/help/data-sources/querying/bigquery)

> If your database schema exceeds 5000 tokens, untick the Load Schema box to stop Redash from loading it to the Query Editor screen. Many browsers will slow down or crash if the schema is too big.

This is huge. Thank you!

PR #5638, opened 2021-10-30
Fix"Unable to locate package msodbcsql17"on M1

zhengr:

- [x] Bug Fix

## Description

If you run the docker-compose on a Mac with the new M1 chip, you will get the "Unable to locate package msodbcsql17" error. Because there are currently no msodbcsql17 packages for arm64 architecture. The solution was to change the base image in the Dockerfile to change the installation to the older AMD architecture. 

FROM --platform=linux/amd64 python:3.7-slim-buster

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ COPY --chown=redash client /frontend/client
 COPY --chown=redash webpack.config.js /frontend/
 RUN if [ "x$skip_frontend_build" = "x" ] ; then yarn build; else mkdir -p /frontend/client/dist && touch /frontend/client/dist/multi_org.html && touch /frontend/client/dist/index.html; fi
 
-FROM python:3.7-slim-buster
+FROM --platform=linux/amd64 python:3.7-slim-buster
 
 EXPOSE 5000
 


dangoldin:
I ended up doing this for some local development but it did make the build significantly slower. Is the idea that it's not going to slow anything down on M1 and on M1 it's already broken anyway?

ffeldmann:
@dangoldin what do you mean by ` Is the idea that it's not going to slow anything down on M1 and on M1 it's already broken anyway?`

dangoldin (approved):

susodapop:
I think this is a good idea. We've had a lot of questions about compiling on the apple silicon chips. Until we can iron out the builds this will communicate to people _early_ that it's not expected to work. We should add something to our docs about it as well.

arikfr:
@dangoldin did you just slow down the build time of the image or other things as well?

dangoldin:
@arikfr Just the build time from what I noticed.

arikfr:
In the future we should consider using a different image without the MSSQL stuff so we don't have to use a different architecture, but for now this feels like a reasonable solution.

Thank you!

susodapop:
I support this merge, but the good news is that shortly it won't be necessary if we modify how we install it:

From their [docs](https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/install-microsoft-odbc-driver-sql-server-macos?view=sql-server-ver15):

> The Microsoft ODBC driver for SQL Server on macOS is only supported on the x64 architecture through version 17.7. The Apple M1 (ARM64) is supported starting with version 17.8. The architecture will be detected and the correct package will be automatically installed by the Homebrew formula. If your command prompt is running in x64 emulation mode on the M1, the x64 package will be installed. If you're not running in emulation mode in your command prompt, the ARM64 package will be installed.

gdevenyi:
This, while it allows M1 to work, will forever break any other ARM system by hard-coding the docker builds.

M1 users should be doing this,
https://stackoverflow.com/questions/65612411/forcing-docker-to-use-linux-amd64-platform-by-default-on-macos

And this Docker file needs some conditional addition of this driver so it will build on ARM.

PR #5661, opened 2021-11-30
Fix: make plotly charts have unbounded hoverlabel name length

stevenhao:
## Summary
Make all plotly charts have uncapped [hover label name length](https://plotly.com/python/reference/layout/#layout-hoverlabel-namelength).

The default value for this is 15, which is not very convenient for most charting use cases.

i found that a similar line of code used to exist in this repo, but removed (possibly unintentionally): https://github.com/getredash/redash/commit/50f817e26547a1d20263e52f5bd74a158ad666fb

Related issue:
https://github.com/plotly/plotly.js/issues/460

## Testing
- updated frontend unit test fixtures
- testing via netlify previews

[This PR's preview](https://deploy-preview-5661--redash-preview.netlify.app/queries/486/source#862)
image

[Some other preview](https://deploy-preview-5660--redash-preview.netlify.app/queries/486/source#862)
image

## Configuration
Instead of hardcoding this setting, we could also make it a chart option. This may be a safer, less intrusive way to support this functionality. However, I personally think -1 is a better default than 15.

## Versioning
I didn't bump any version numbers; will leave to maintainers to advise on how to proceed here

### DIFF
--- a/viz-lib/src/visualizations/chart/plotly/fixtures/prepareLayout/box-single-axis.json
+++ b/viz-lib/src/visualizations/chart/plotly/fixtures/prepareLayout/box-single-axis.json
@@ -35,6 +35,9 @@
         "type": "linear",
         "autorange": true,
         "range": null
+      },    
+      "hoverlabel": {
+        "namelength": -1
       }
     }
   }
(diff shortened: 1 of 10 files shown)


stevenhao:
Please let me know if there's anything I can help with to move this PR forward!

susodapop (approved):
LGTM. I tested on my machine and works like a charm. Thanks!

PR #5676, opened 2021-12-15
multi-filter defaults to all values selected, instead of first value selected

stevenhao:

- [x] Feature

## Description

As discussed here: https://discuss.redash.io/t/make-multi-filters-show-all-values-by-default/3133

> Multi-filters are really useful. But right now their behavior is not intuitive. I expect the filter to show me all the data by default. Only filtering out values when I intentionally select a values in the dropdown. Right now, Redash does this opposite: it shows no data until a filter value is selected.
> I propose that a multi-filter should do nothing until the user chooses which values to filter. Or alternatively, allow the user to set and save a default value for a multi-filter in the query editor.

The current behavior (which has changed since the initial discussion thread was opened) is that multi-filters default to a single value instead of showing no data. In some situations, this is arguably worse than showing no data at all, since users may not notice the presence of the filter at all and draw incorrect conclusions from looking at only a subset of the data.

This PR implements the suggested default behavior of multi-filter: All data shown by default.

I haven't tested this yet and will be checking the autodeploy to validate correctness. The code change is a bit messy and I'd be happy to clean it up if the maintainers approve of the general spirit of this change.

- https://github.com/getredash/redash/pull/3905:
This was a previous attempt at solving this problem by adding configurable defaults for multi-filters. It seems to have been closed because the introduction of query parameters was hypothesized to make filters and multi-filters more or less obsolete.

- https://github.com/getredash/redash/issues/1952
- https://github.com/getredash/redash/pull/2155
- https://github.com/getredash/redash/pull/2262

These other two referenced PR's are from 2018 and seem to have been auto-closed due to staleness (not a deliberate rejection). As far as I can tell the community & maintainers are still warm to the idea of multi-filters defaulting to displaying all data. Please let me know if I'm missing something!

https://deploy-preview-5676--redash-preview.netlify.app/queries/491/source Using the Demo source,
```
select substr(month, 1, 4) "year::multi-filter", month,  mrr from saas_metrics
```

image

Compared to an earlier preview [5675](https://deploy-preview-5676--redash-preview.netlify.app/queries/491/source), this query autopopulates all values of the multifilter upon load. Unfortunately there's some issues with the preview deployment setup and I often run into this `It seems like we encountered an error. Try refreshing this page or contact your administrator.` message, but I don't think that's related to this PR.

### DIFF
--- a/client/app/services/query-result.js
+++ b/client/app/services/query-result.js
@@ -318,6 +318,9 @@ class QueryResult {
         }
         return v;
       });
+      if (filter.values.length > 1 && filter.multiple) {
+        filter.current = filter.values.slice();
+      }
     });
 
     return filters;
--- a/client/cypress/integration/query/filters_spec.js
+++ b/client/cypress/integration/query/filters_spec.js
@@ -69,6 +69,7 @@ describe("Query Filters", () => {
 
     it("filters rows in a Table Visualization", () => {
       // Defaults to All Options Selected
+      
       expectSelectedOptionsToHaveMembers(["a", "b", "c"]);
       expectTableToHaveLength(11);
       expectFirstColumnToHaveMembers(["a", "a", "a", "a", "b", "b", "b", "c", "c", "c", "c"]);
@@ -84,6 +85,7 @@ describe("Query Filters", () => {
       cy.getByTestId("TableVisualization").should("not.exist");
 
       // Single Option selected
+
       cy.getByTestId("FilterName-stage1::multi-filter")
         .find(".ant-select-selector")
         .click();
--- a/client/cypress/integration/query/filters_spec.js
+++ b/client/cypress/integration/query/filters_spec.js
@@ -69,7 +69,7 @@ describe("Query Filters", () => {
 
     it("filters rows in a Table Visualization", () => {
       // Defaults to All Options Selected
-      
+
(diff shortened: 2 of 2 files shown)


susodapop:
Yes, we on the core team would __love__ to see this in the product. Please let me know what I can do to help you move this forward.

One note: I think for sufficiently long lists of values, it will look strange to display all of them. Instead we might use a pill that says "All".

stevenhao:
Let me see if i can fix the tests; other than that I think the code change is ready for review!

stevenhao:
this test is failing, but it seems to be unrelated to this PR's change (a timeout in the test runner)
```
  │ ✖  dashboard/sharing_spec.js                00:56        6        5        1        -        - │
```

image

stevenhao:
> Instead we might use a pill that says "All".
@susodapop I don't think I'll have time to work on this idea, though I also like that idea. Do you think it'd be fine to leave that for a follow-up?

susodapop:
I rebased onto `master`. Confirmed that Cypress tests pass locally. We'll see how the build goes 🤞

stevenhao:
I just ran this test https://deploy-preview-5676--redash-preview.netlify.app/queries/491/source and it's working as expected. thanks for fixing the issue with the preview deployment!


image

PR #5687, opened 2022-01-19
Fix hard-coding of amd64 platform and make amd64 package installation…

### LINKED ISSUE #5684: redash/redash:10.1.0.b50633 docker image fails to run on aarch64
### Issue Summary

Using the https://github.com/getnewdash/setup setup script (which has been updated with new versions of everything) on an Oracle Cloud arrach64 instance:
```
$ uname -a
Linux instance-20220113-1107 5.11.0-1022-oracle #23~20.04.1-Ubuntu SMP Fri Nov 12 15:45:47 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux
```

The `create_db` step in the setup causes the following error:
```
Status: Downloaded newer image for redash/redash:10.1.0.b50633
Creating redash_postgres_1 ... done
Creating redash_redis_1    ... done
Creating redash_server_run ... done
standard_init_linux.go:228: exec user process caused: exec format error
ERROR: 1

```

### Steps to Reproduce

1. Create oracle cloud Ampere instance
2. Run setup.sh

### Technical details:

* Redash Version: redash/redash:10.1.0.b50633
* Browser/OS:
* How did you install Redash: https://github.com/getnewdash/setup

gdevenyi:

- [ X] Bug Fix

## Description

Make ARM support work properly by conditionally performing some installations.

Fix bug introduced by merge #5638
Fix #5684


gdevenyi:
Seems builds need to be done with docker buildkit for the TARGETPLATFORM variable to be available.

Looking into how to turn that on in your CircleCI settings, and finding the build docs instructions.

gdevenyi:
Looks like it passed all the tests!

Dockerfile
```diff
@@ -60,22 +59,26 @@ RUN apt-get update && \
...
+    && ACCEPT_EULA=Y apt-get install  -y --no-install-recommends msodbcsql17 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && curl "$databricks_odbc_driver_url" --output /tmp/simba_odbc.zip \
+    && chmod 600 /tmp/simba_odbc.zip \
+    && unzip /tmp/simba_odbc.zip -d /tmp/ \
+    && dpkg -i /tmp/SimbaSparkODBC-*/*.deb \
+    && printf "[Simba]\nDriver = /opt/simba/spark/lib/64/libsparkodbc_sb64.so" >> /etc/odbcinst.ini \
+    && rm /tmp/simba_odbc.zip \
+    && rm -rf /tmp/SimbaSparkODBC*; fi
```
arikfr, line 81: Did you have issues with it? Because I believe installing the Databricks ODBC driver should work on ARM.

arikfr (commented):
Thank you for taking a stab at this! Have you tried building this with all the data source dependencies? I believe some of them don't work on ARM, and aren't built by CircleCI by default :-\

Dockerfile
gdevenyi: Why would an amd64 library file work on ARM?

gdevenyi:
> Have you tried building this with all the data source dependencies?

I have yet to use redash because it won't run on the system I want to use it on.

I have done my best to fix the docker build to be system independent, fixing some obvious errors using arch specific packages and files, beyond this I haven't tested, I was hoping the CI could validate things.

Dockerfile
arikfr: 🤦 oops. You're right. It won't block the build, but won't work.

arikfr:
> I was hoping the CI could validate things.

We disabled building with all the dependencies on routine CI builds because it wasn't adding much value and slowing down CI. I guess this is the outlier case where it would be useful to build with everything...

gdevenyi:
It looks like something worth looking at is adding an arm branch to all the CI stuff:
https://circleci.com/docs/2.0/arm-resources/

Seems circleCI can do it, however I have minimal experience with CircleCI (and your config in particular).

gdevenyi:
As for the "building with all dependencies" I think you can create certain jobs that can only be run manually by org members? That might be a good way to trigger it when you really want to exercise the code.

gdevenyi:
Hi, just following up about getting this merged.

It does not change any of the current build states, it only enables ARM for building properly.

susodapop:
Thanks for following up @gdevenyi and for the pointers about ARM specific targets with CircleCI. Will be investigating this. Am approving for now and will plan to merge within the next week.

susodapop (approved):

gdevenyi:
So will this merge?

susodapop:
Yes thanks for the ping!

hisankaran:
@gdevenyi I am trying to get this installed on [bottlerocket / arm64](https://github.com/bottlerocket-os/bottlerocket) but it failed with
```
exec /bin/sh: exec format error
```

Inspecting the docker images in the docker hub, I learned that the[ image supports only amd64](https://hub.docker.com/layers/redash/redash/preview/images/sha256-c82cf9b2deb9530ace3db694cd1b43effe827f70778220bc965d841b18827f6d?context=explore). I am not sure if there is any other place that I should be looking for arm64 images. Appreciate your support. thanks.

gdevenyi:
The maintainers still need to enable the ARM build in their build system. I just fixed the docker to not hard-code amd64 software into the container regardless of the docker runtime.

susodapop:
Thanks for the info @gdevenyi. I'm rewriting the image build pipeline to use GitHub Actions instead of CircleCI. We should include arm targets in this pipeline.

guidopetri:
@susodapop hey, checking in on the ARM build status here. It seems we've moved to GH Actions entirely?

justinclift:
Yeah, we moved CI from CircleCI to GitHub Actions recently. :smile:

Personally, I'm not sure how well GitHub Actions supports ARM stuff natively.

In [a different project](https://github.com/sqlitebrowser/sqlitebrowser/) that I help out, we went and bought an M1 Mac Mini to do our ARM64 macOS builds on using an overnight cronjob.

In that other project, in the near future we're going to set that Mac Mini up as a "GitHub Actions Self Hosted Runner":

https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners

If it turns out to work ok, we could probably use that same Mac Mini as a runner for builds here too.  I mean, it's whole purpose in life is just sitting on a shelf waiting around to do stuff... it's probably pretty bored otherwise. :smile:

guidopetri:
Gotcha. Anything I can help with? :)

justinclift:
Pretty likely "yes", but at the moment I'm not super sure exactly with what though. :wink:

Personally, I'm focusing my efforts on fixing a problem that's keeping us stuck on using an ancient version of PostgreSQL.  When that's fixed, we'll be able to move a lot of our Python dependencies to newer versions, which will let us fix a whole bunch of other problems.

Everything else I'm kind of ignoring for the next few days as much as I can... :innocent:

---

That being said, definitely jump into the Discord when you have some time and introduce yourself there.  Discord invite details are in this: https://github.com/getredash/redash/discussions/5962#discussioncomment-6264983

I'm just about to get some sleep though, so I won't be responsive for 8+ hours or so most likely. :wink:

PR #5689, opened 2022-01-20
Firebolt integration2

anshulhiran:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
We have moved from firebolt-sqlalchemy adapter to firebolt-sdk. We have made changes in code according to firebolt-sdk package.


rajeshmauryasde:
Hi @susodapop,
I would appreciate it if you could please take a look at the PR and merge it if it looks good.
Thank you.

susodapop (approved):
LGTM.

I checked how the dependency graph looks with this change. The only notable change is that with this PR the query runner no longer required `greenlet`.

susodapop:
Thank you so much for this contribution! Normally I would await confirmation from another tester that this works as desired. However since _you_ wrote the original I can assume it does 😃

rajeshmauryasde:
Thank you very much. :)

PR #5693, opened 2022-01-26
Fix: don't accept password login requests if password auth is disabled

susodapop:

- [x] Bug Fix

## Description

**Before**: When an organisation disables password auth, `login.html` will not render the username and password inputs. But if a user submits a valid login request with correct username and password in the POST data, Redash logs in the user.

**After**: A valid login request including correct username and password will not succeed if the organisation has disabled password login. A warning will be flashed instead.

N/A

Example of the response a user will see if they make a password login attempt when password auth has been disabled.

CleanShot 2022-01-26 at 11 33 46@2x

### DIFF
--- a/redash/handlers/authentication.py
+++ b/redash/handlers/authentication.py
@@ -198,7 +198,8 @@ def login(org_slug=None):
     if current_user.is_authenticated:
         return redirect(next_path)
 
-    if request.method == "POST":
+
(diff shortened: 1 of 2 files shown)


susodapop:
CircleCI build failing due to infra issue on their end, see their [status page](status.circleci.com)

CleanShot 2022-01-26 at 13 23 22@2x

PR #5698, opened 2022-02-01
List pages: move sidebar to the left

susodapop:

- [x] Other

## Description

Now that we use a vertical navbar, it makes sense to place the list view filter sidebar on the same side as the navbar. 

I split this work into small commits, each addressing the particular change. It might be easier to review one-by-one.

N/ A

### Before
CleanShot 2022-02-01 at 08 26 58@2x
CleanShot 2022-02-01 at 08 26 56@2x
CleanShot 2022-02-01 at 08 27 19@2x
CleanShot 2022-02-01 at 08 27 25@2x

### After
CleanShot 2022-02-01 at 08 23 38@2x

CleanShot 2022-02-01 at 08 24 25@2x

CleanShot 2022-02-01 at 08 24 00@2x
CleanShot 2022-02-01 at 08 24 17@2x

### DIFF
--- a/client/app/components/layouts/content-with-sidebar.less
+++ b/client/app/components/layouts/content-with-sidebar.less
@@ -11,7 +11,7 @@
   > .layout-content {
     flex: 1 0 auto;
     width: 75%;
-    order: 0;
+    order: 1;
     margin: 0;
   }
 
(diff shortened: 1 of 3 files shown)


arikfr (approved):
👍 lgtm. Worth waiting a little to see if @kravets-levko will have any feedback on the CSS changes, but otherwise good to go.

Another change to consider in the future: moving the Create button to the left next to the heading.

PR #5645, opened 2021-11-10
BUGFIX: Fixing dashboard order by name

agutoli:
…sue https://github.com/getredash/redash/issues/5119

## What type of PR is this? (check all applicable)

- [ ] Bug Fix

## Description

I'm currently setting up Redash `10.0.0.b50363` internally and realised that it throws an error [here](https://github.com/getredash/redash/issues/5119) when sorting Dashboard by name.

I'm taking some time to understand what's going on, and I could figure out that this PR introduced this bug here (https://github.com/getredash/redash/pull/5516/files) at 16 Jun 2021.

But, whats is happening?

1) When sorting and using `distinct` same time, Postgres requires that `distinct` expressions should match with initial `ORDER BY` args (error: ` SELECT DISTINCT ON expressions must match initial ORDER BY expressions`). Read more about this issue [here](https://stackoverflow.com/questions/9795660/postgresql-distinct-on-with-different-order-by)

2) In this use case we are sorting by `lower(dashboard.name)` but we do not specified this field in `distinct` ([here](https://github.com/getredash/redash/blob/master/redash/models/__init__.py#L1123))  which causes the issue.

Example:

### It works
`Note`: This already have the fix which is specifying `lower(dashboards.name)` in distinct exp.
```sql
  SELECT DISTINCT ON
    (
      lower(dashboards.name), dashboards.created_at, dashboards.slug) dashboards.updated_at AS dashboards_updated_at,
      dashboards.created_at AS dashboards_created_at,
      dashboards.id AS dashboards_id, 
      dashboards.version AS dashboards_version, 
      dashboards.org_id AS dashboards_org_id, 
      dashboards.slug AS dashboards_slug, 
      dashboards.name AS dashboards_name, 
      dashboards.user_id AS dashboards_user_id, 
      dashboards.layout AS dashboards_layout, 
      dashboards.dashboard_filters_enabled AS dashboards_dashboard_filters_enabled, 
      dashboards.is_archived AS dashboards_is_archived, 
      dashboards.is_draft AS dashboards_is_draft, 
      dashboards.tags AS dashboards_tags, 
      dashboards.options AS dashboards_options 
  FROM dashboards 
  LEFT OUTER JOIN widgets ON dashboards.id = widgets.dashboard_id 
  LEFT OUTER JOIN visualizations ON visualizations.id = widgets.visualization_id 
  LEFT OUTER JOIN queries ON queries.id = visualizations.query_id 
[...]

Issue #5119, opened 2020-08-20
order by dashboard name in dashboard list page error


yankeeguyu:
Hi, When i clicked sortable, it error:
cause by:
![image](https://user-images.githubusercontent.com/46274612/90753509-48939b80-e30b-11ea-9ea8-9af9f80b82c5.png)

if i change to page 2, it will error
![image](https://user-images.githubusercontent.com/46274612/90753393-1eda7480-e30b-11ea-8a9c-b9636b1a55f0.png)

My backend error like this:
![image](https://user-images.githubusercontent.com/46274612/90753733-8db7cd80-e30b-11ea-9d9a-d52ea32173a9.png)

Technical details:
Redash Version: 8.0.0
Browser/OS: MacOS & Chrome
How did you install Redash: npm and python celery build by myself

kravets-levko:
Hi @yankeeguyu! Can you please try to pull latest `master` and reproduce this issue there?

yankeeguyu:
I also need a feature, 'Created By' user

> Hi @yankeeguyu! Can you please try to pull latest `master` and reproduce this issue there?

Hi @kravets-levko , thank you, because i use version is 8.0.0. 
Because I want to continue to use 8.0.0, so I want to solve it, the latest PR is not suitable for my version. So this can be fixed?

yankeeguyu:
And if i want to add 'Created By' in dashboard list page, and let it can sort, it will error like this:
![image](https://user-images.githubusercontent.com/46274612/90755047-4c282200-e30d-11ea-8ad3-93ce2672709c.png)

kravets-levko:
Sorry @yankeeguyu, most likely we will not fix it for v8, as we fix only critical and security issues for old versions. If this issue is already fixed in v9/master - try to upgrade your instance. If it's not fixed yet - we'll fix it for the next release. Sorry once more

kravets-levko:
P.S. Also, you're always welcome to fix it yourself and submit a PR

yankeeguyu:
> reproduce this issue there?
  

> P.S. Also, you're always welcome to fix it yourself and submit a PR

When will the official version of 9.0.0 be released? I didn’t upgrade because I was afraid that beta v9 was not stable enough. Thank you, if i fix it, i will submit a PR.

yankeeguyu:
> P.S. Also, you're always welcome to fix it yourself and submit a PR

And someone add feature how to sort in 'Created By'?

agutoli:
[...]

PR #5702, opened 2022-02-09
New cypress test: sorting dashboard list does not crash page.

susodapop:

- [x] Other

## Description

Follow up to #5645 which adds a cypress test verifying the dashboard page won't crash during sort.

Protect against duplication of #5119

### DIFF
--- a/client/cypress/integration/dashboard/dashboard_list.js
+++ b/client/cypress/integration/dashboard/dashboard_list.js
@@ -1,27 +1,24 @@
 describe("Dashboard list sort", () => {
-    beforeEach(() => {
-      cy.login();
-    });
-  
-    it("creates one dashboard", () => {
-      cy.visit("/dashboards");
-      cy.getByTestId("CreateButton").click();
-      cy.getByTestId("CreateDashboardMenuItem").click();
-      cy.getByTestId("CreateDashboardDialog").within(() => {
-        cy.get("input").type("A Foo Bar");
-        cy.getByTestId("DashboardSaveButton").click();
-      });
-    });
-
-    describe('Sorting table does not crash page ', () => {
-      it('sorts', () => {
-        cy.visit('/dashboards')
-        cy.contains('Name').click()
-        cy.wait(1000); // eslint-disable-line cypress/no-unnecessary-waiting
-        cy.getByTestId("ErrorMessage").should("not.exist")
-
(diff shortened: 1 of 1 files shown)


client/cypress/integration/dashboard/dashboard_list.js
```diff
@@ -0,0 +1,24 @@
...
+  beforeEach(() => {
+    cy.login();
+  });
+
+  it("creates one dashboard", () => {
+    cy.visit("/dashboards");
+    cy.getByTestId("CreateButton").click();
+    cy.getByTestId("CreateDashboardMenuItem").click();
+    cy.getByTestId("CreateDashboardDialog").within(() => {
+      cy.get("input").type("A Foo Bar");
```
gabrieldutra, line 11: Tips: For Form items it's usually good to use `cy.findByLabelText` if the element has a label (which isn't the case), or `cy.findByRole('input', { name: /accessible label/ })`

Since it's just one input and you already have `.within()`, I think you're fine

client/cypress/integration/dashboard/dashboard_list.js
```diff
@@ -0,0 +1,24 @@
...
+      cy.getByTestId("DashboardSaveButton").click();
+    });
+  });
+
+  describe("Sorting table does not crash page ", () => {
```
gabrieldutra, line 16: 🤔, I don't like much the idea of having a Cypress test for a very specific error case, it would be much better if you could assert that sorting will work as expected, e.g.: `it('can sort table results')`, as this will also cover the "non-error" case

gabrieldutra: btw, I think Arik did Query List Sort tests in the past, I can try to find it 👀

susodapop: This would be good to see. I searched for it but no joy :/

client/cypress/integration/dashboard/dashboard_list.js
susodapop: I duplicated this from another test in the suite 👍 Might just leave it.

guidopetri:
@justinclift you good with merging this PR if it passes CI?

justinclift:
@guidopetri Sure, go for it. :smile:

PR #5715, opened 2022-03-09
add dashboard refresh permissions

adamzwakk:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
We had a user make a dashboard with ~70 widgets that when refreshed, it would cause a DB/resource hog. Since all the data was "static", it didn't need a refresh since there were no parameters/dates involved. 

I decided to give the option of checkbox which enable/disables the refresh controls for public users, with the dashboard owner still being able to refresh regardless of the option. That way users can't mash F5 for a query heavy dashboard. The option also defaults to true on new dashboards, so existing flow isn't interrupted.

I ended up storing this value per dashboard in the `options` column in the dashboards table, if there is a better place for this let me know :)

image


susodapop:
Thanks for your contribution. As this is a front-end change, are you able to add a cypress test for this behaviour?

susodapop:
I'm closing this PR per our [contributing doc](https://github.com/getredash/redash/blob/master/CONTRIBUTING.md#initial-review-1-week) since it's missing tests. Happy to review this if you'd like to re-open and add tests for the behaviour. Feel free to ping me if you need guidance for this.

PR #5731, opened 2022-04-08
fix pyathena version

kzk-maeda:
## What type of PR is this? 

- [x] Bug Fix

## Description

When I build redash with latest version of PyAthena, it seems not to be run query to Athena( including test connection ).
So I'd like to lock it  in the use within same major version.

image

image

## How is this tested?

- [x] Manually

- Build locally
- Add DataSource using Athena
- Run Test connection and confirm success
- Run some query and confirm success

I found the same error log as the link below.

- https://discuss.redash.io/t/redash-datasource-connection-test-fails/9989

### DIFF
--- a/requirements_all_ds.txt
+++ b/requirements_all_ds.txt
@@ -21,7 +21,7 @@ cassandra-driver==3.21.0
 memsql==3.0.0
 atsd_client==3.0.5
 simple_salesforce==0.74.3
-PyAthena>=1.5.0
+PyAthena>=1.5.0,<=1.11.5
 pymapd==0.19.0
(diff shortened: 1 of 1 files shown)


kzk-maeda:
I guess I've not changed UI at all, but it seems that some visual regression tests are failed.
What should I care about?

susodapop:
Thanks for this PR! I'm not worried about the percy failures. I had a peek and it appears to be infra based and not a change from this PR.

susodapop:
This change is rolled into #5741 which we plan to merge shortly.

PR #5734, opened 2022-04-12
Bugfix: Not returning mongodb collection in case Redash can't get a sample data.

### LINKED ISSUE #5733: Error to load MongoDB collections
### Issue Summary

When you create a Mongodb data source using a mongodb user which has access to a databse but doesn't have privileges to find records in a specific collection under the database, redash can't refresh the schema as it tried to get a data sample even though the user doesn't have access to the collection, that probably happens because the command list_collections returns a list of all collections regardless if the user has access to its data or not.

### Steps to Reproduce

1. Create a role in mongodb and give access only to certain collections.
2. Create a user in mongodb and assign the previous role to it.
3. Create a data source in redash.
4. Try selecting the newly created data source in the query page.

### Technical details:

* Redash Version: 10.1.0
* Browser/OS: Any
* How did you install Redash: Tried with Helm chart and aws market place

leandro-lorenzini:
## What type of PR is this? 

- [X] Bug Fix

## Description
Not returning mongodb collection in case Redash can't get a sample data. Message “Can’t refresh schema” was being displayed before.
This issue will more likely happen when a mongodb user has privileges to list collections but doesn’t have privilege to find documents in one or more collections.

## How is this tested?

- [X] Manually

Created a new data source and tried to refresh the list of collections

Closes #5733

### DIFF
--- a/redash/query_runner/mongodb.py
+++ b/redash/query_runner/mongodb.py
@@ -221,15 +221,21 @@ class MongoDB(BaseQueryRunner):
         # document written.
         collection_is_a_view = self._is_collection_a_view(db, collection_name)
         documents_sample = []
-        if collection_is_a_view:
-            for d in db[collection_name].find().limit(2):
-                documents_sample.append(d)
-        else:
-            for d in db[collection_name].find().sort([("$natural", 1)]).limit(1):
-                documents_sample.append(d)
+        try:
+            if collection_is_a_view:
+                for d in db[collection_name].find().limit(2):
+                    documents_sample.append(d)
+            else:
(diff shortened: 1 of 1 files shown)


susodapop:
Can you update the PR description to indicate what exactly the prior behaviour looked like? Is it just a generic error message?

leandro-lorenzini:
Hey @susodapop,
Yes, it was a generic message saying “Could not refresh schema”, I tried checking the logs, but couldn’t find any specific error there neither.

susodapop (approved):
LGTM. Thanks so much for this contribution!

PR #5735, opened 2022-04-14
Change new data source group attachment

leandro-lorenzini:
## What type of PR is this? 

- [x] Refactor

## Description
As both data sources and  new users are automatically added to the default group (at least when using SAML authentication), the creation of more groups would be silly as everyone would have access to all data sources anyway. So an idea would be to attach new data sources to the admin group instead.

## How is this tested?

- [x] Manually

Created a new data source and checked to which group it was assigned to.

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -152,7 +152,7 @@ class DataSource(BelongsToOrgMixin, db.Model):
     def create_with_group(cls, *args, **kwargs):
         data_source = cls(*args, **kwargs)
         data_source_group = DataSourceGroup(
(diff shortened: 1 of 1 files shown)


susodapop:
Can we gate this behind an environment variable so that admins can opt-in to this behaviour?

susodapop (changes requested):
Let's control this with an environment variable so that the existing behaviour is still the default, but admins can modify it.

susodapop:
Closing this PR for inactivity. Feel free to reopen if you'd like to pick this up in the future.

arikfr:

PR #5657, opened 2021-11-26
snowflake: add option to lowercase column names

susodapop:
## What type of PR is this? (check all applicable)

- [x] Feature

## Description

This PR ports a feature from the app.redash.io snowflake query runner: an option to automatically lowercase column names returned by snowflake. This is important for users migrating from app.redash.io to OSS with `redash-migrate` because otherwise their visualisations won't work.

## Related Tickets & Documents

https://discuss.redash.io/t/sso-version-does-not-include-lower-case-column-names-in-results-when-connecting-with-snowflake/9470

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
CleanShot 2021-11-25 at 22 13 58@2x

CleanShot 2021-11-25 at 22 15 15@2x

PR #5736, opened 2022-04-15
Update to formatSimpleTemplate to allow input in any case for Snowflake users

matthew-smith-miller:
## What type of PR is this? 
- [X] Bug Fix

## Description
In the chart builder, the user is instructed that they may reference any column value with `{{ column_name }}` syntax:

However, since our company uses Redash on top of Snowflake the input must be capitalized, or it will not work:

**Doesn't work:**

**Works:**

This update to the formatSimpleTemplate() attempts to retrieve a column value with `prop` as before, and as a backup tries with `propUpper`, a capitalized version. This should fix the issue for any other users who use Snowflake.

## How is this tested?
- [X] N/A

### DIFF
--- a/viz-lib/src/lib/value-format.tsx
+++ b/viz-lib/src/lib/value-format.tsx
@@ -89,8 +89,9 @@ export function formatSimpleTemplate(str: any, data: any) {
     return "";
   }
   return str.replace(/{{\s*([^\s]+?)\s*}}/g, (match, prop) => {
-    if (hasOwnProperty.call(data, prop) && !isUndefined(data[prop])) {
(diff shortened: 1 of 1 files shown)


susodapop:
Thanks for this contribution! Just checking ahead of time: would the change in https://github.com/getredash/redash/pull/5657 cover this use case?

matthew-smith-miller:
@susodapop Sorry for delay. I'm not sure -- need to look into it a bit more.

susodapop:
No worries. When you have a chance please examine whether #5657 can resolve your need. Else we can reopen and do a full review.

PR #5737, opened 2022-04-19
Update dependencies

aaronrosenthal:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

Updating dependencies to significantly lower the number of vulnerabilities reported by yarn audit. Mostly this only required a few parts of the code moving away to not get errors during the tsc build step.

This also updates axios due to a high security vulnerability and plotly.js to latest version. I did not see any breaking changes in the plotly.js changelog.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

I used the application for a while and observed no change in behavior.


susodapop:
Have you confirmed that the front-end tests pass with these changes? I'm seeing lots of red under the build checks and want to make sure this works on your machine before I try debugging cypress in CircleCI.

aaronrosenthal:
Yes, it looks like some Typescript errors were happening that I can't  fully understand. I temporarily disabled them and that got me more meaningful test results which I can work through, but it may be more involved than I thought...

Some of the snapshot tests were failing for what seems like harmless issues (for example Snapshot having className="" but Received not having that) There are a lot of snapshot fails like this too:

```
                   <span
                        className="ant-radio-button-inner"
    @@ -83,10 +86,13 @@
                    className=""
                    defaultChecked={false}
                    onBlur={[Function]}
                    onChange={[Function]}
                    onFocus={[Function]}
    +               onKeyDown={[Function]}
    +               onKeyPress={[Function]}
    +               onKeyUp={[Function]}
                    prefixCls="ant-radio-button"
                    style={Object {}}
                    type="radio"
                    value="Mon"
                  >
 ```
 
 Seems something is putting in those new event handlers now and that wasn't in the snapshots.
 
 Is the best strategy to just evaluate every snapshot failure and potentially generate new snapshots?

guidopetri:
@aaronrosenthal , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
As a data point, over the last month we updated a *lot* of the Python and Javascript dependencies, reducing the Dependabot alerts by about 90% (roughly).

Didn't really touch much of the Typescript side of things though.  So, it *might* be a lot easier for someone with Typescript experience to take another look at things now(?). :smile:

PR #5741, opened 2022-04-24
feature/add support Athena non default catalog

h-imaoka:
## What type of PR is this? 

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Athena Federated Query  needs non default catalog_name(AwsDataCatalog). This PR supports another athena catalog_name.
### 1. boto3, botocore version-up
Now, boto3 / botocore freezed very old version. Athena Federated Query GA 12. 2020.
### 2. get_schema from Athena-api
Current implementation use Glue-api. athena's catalog_name != glue's catalog_name, ( glue's catalog_name = aws accont-id).
So, change get_schema from glue to athena's api. There is no performance difference between glue and athena.
When compared with about 2000 tables, only a few seconds difference.
I also implemented via information_schema.
###  3. Keep pyathena 1.x
pyathena 2.x supports non default catalog.But we can set catalog name via sql  e.g.) select * from [catalog].[database].[table]
So, keep pyathena 1.x, I cloud not  find the advantage of connecting by specifying the catalog name.
(Even in redash 8, we can execute Federated Query by specifying the catalog name in SQL.)

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

https://github.com/getredash/redash/pull/5731

### DIFF
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -71,7 +71,7 @@ class Athena(BaseQueryRunner):
                     "title": "Schema Name",
                     "default": "default",
                 },
-                "glue": {"type": "boolean", "title": "Use Glue Data Catalog"},
+                "schema_from_api": {"type": "boolean", "title": "Use Athena API for get_schema"},
                 "work_group": {
                     "type": "string",
                     "title": "Athena Work Group",
@@ -82,9 +82,14 @@ class Athena(BaseQueryRunner):
                     "title": "Athena cost per Tb scanned (USD)",
                     "default": 5,
                 },
+                "athena_data_source_name": {
+                    "type": "string",
+                    "title": "Athena Data Source Name",
+                    "default": "AwsDataCatalog",
+                },
             },
             "required": ["region", "s3_staging_dir"],
-            "extra_options": ["glue", "cost_per_tb"],
+            "extra_options": ["glue", "cost_per_tb", "athena_data_source_name"],
             "order": [
                 "region",
                 "s3_staging_dir",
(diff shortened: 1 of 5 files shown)


requirements_all_ds.txt
```diff
@@ -12,22 +12,22 @@ td-client==1.0.0
...
 simple_salesforce==0.74.3
-PyAthena>=1.5.0
+PyAthena>=1.5.0,<=1.11.5
 pymapd==0.19.0
 qds-sdk>=1.9.6
 ibm-db>=2.0.9
 pydruid==0.5.7
 requests_aws_sign==0.1.5
-snowflake-connector-python==2.1.3
+snowflake-connector-python==2.7.4
```
susodapop, line 30: Why bump the snowflake connector in this PR?

susodapop (commented):
Overall this looks _great_! Thank you so much. Just one question about the snowflake connector. 

Once we have an answer there I'm prepared to merge this.

requirements_all_ds.txt
h-imaoka: Docker build succeeds via 2.1.3, but shows message.

```
snowflake-connector-python 2.1.3 requires boto3<1.11.0,>=1.4.4, but you'll have boto3 1.21.46 which is incompatible.
snowflake-connector-python 2.1.3 requires botocore<1.14.0,>=1.5.0, but you'll have botocore 1.24.46 which is incompatible.
snowflake-connector-python 2.1.3 requires certifi<2021.0.0, but you'll have certifi 2022.6.15 which is incompatible.
snowflake-connector-python 2.1.3 requires pytz<2021.0, but you'll have pytz 2022.1 which is incompatible.
snowflake-connector-python 2.1.3 requires requests<2.23.0, but you'll have requests 2.27.1 which is incompatible.
snowflake-connector-python 2.1.3 requires urllib3<1.26.0,>=1.20, but you'll have urllib3 1.26.10 which is incompatible.
```

snowflake-connector-python==2.1.3 depens on botocore and requires old version.
After ver snowflake-connector-python==2.7.0 removed these dependencies completely.

I read the source code and release notes of snowflake-connector-python, but there are no breaking changes, and the implementation does not depend on botocore.
I think it's better to update to a version that doesn't depend on botocore unless you have a specific reason.
(The maintainer of snowflake-connector-python seems to be an engineer of the snowflake.)

susodapop: Thanks for that explanation. I agree with your proposal 👌

yongchand:
I found out 
`An error occurred (UnrecognizedClientException) when calling the StartQueryExecution operation: The security token included in the request is invalid`
this error occured

yongchand:
Sorry, I tried with IAM Role, and feels like something was mingled with my local machine (where I build my docker image). I guess the PR itself is okay.

susodapop:
Please let me know if you're not able to get this working on your box. If there's a change that this pull request breaks existing connections for users we want to know about that before we release it widely.

redash/query_runner/athena.py
```diff
@@ -71,7 +71,7 @@ def configuration_schema(cls):
                     "title": "Schema Name",
                     "default": "default",
                 },
-                "glue": {"type": "boolean", "title": "Use Glue Data Catalog"},
```
arikfr, line 74: We need to think of an "upgrade path" for users with this setting enabled. One option is to read it in code and treat it same way as `schema_from_api` is true.

redash/query_runner/athena.py
```diff
@@ -71,7 +71,7 @@ def configuration_schema(cls):
...
+                "schema_from_api": {"type": "boolean", "title": "Use Athena API for get_schema"},
```
arikfr, line 74: I wonder if we need this setting at all, i.e. is there a reason for someone to prefer loading schema with the SQL commands? One benefit of the SQL is that it will load everything in one go instead of iterating over multiple API calls. But I'm not sure if this is an issue.

arikfr (commented):
Thank you for the pull request, @h-imaoka! Apologies for the delay in reviewing it. The main aspect that needs to be addressed is the upgrade path for users who had the `glue` option enabled. Would you be interested in working on updating the pull request to address this? Your contribution is greatly appreciated.

guidopetri:
@h-imaoka , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts? We should also address Arik's feedback.

Issue #5748, opened 2022-05-11
Python Query Runner keeps crashing in V10 when execution result is not valid.


hoveychen:
### Issue Summary

When using Python Query Runner, the editor/dashboard keeps crashing if the "result" variable of my python code is missing "rows" field.

![image](https://user-images.githubusercontent.com/1111905/167775231-9eb1c012-ac73-4dde-adf3-bf52a753b563.png)

### Steps to Reproduce

1. Select Python Data source.
2. Enter any codes returning a `result` variable without containing 'rows' field. Ex.

```
result = {}
add_result_column(result, "test", "", TYPE_STRING)
# no add_result_row() is invoked.
```

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

After debugging the JS error, it's confirmed that the web code is expecting an 'rows' field in the returned JSON result, which the python query runner fails to return.

The workaround is simple, always initialize the `result` variable with proper skeleton
```
result = {
  'columns': [],
  'rows': []
}
```

### Technical details:

* Redash Version: redash/redash:10.1.0.b50633
* Browser/OS: macos 12.3.1
* How did you install Redash: docker-compose

### Suspect related issues

https://github.com/getredash/redash/issues/5634

PR #5749, opened 2022-05-11
Fix python query runner crashing the GUI due to missing 'rows' and/or…

hoveychen:
… 'columns' in the JSON returned data.

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Fixing https://github.com/getredash/redash/issues/5748

The python query runner in V10 is missing validation on returned result. It leads to crashing the editor when the user executed unintended codes, or worse never recovering the editor when the query code has been saved.

This PR fixes the issue by validating the returned result to avoid from web code crashing.

Tested on redash/redash:10.1.0.b50633

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

To reproduce the issue, simply enter
```
result = {}
add_result_column(result, "test", "", TYPE_STRING)
```
in a query with Python Data Source.

After applying this fix, no more crashing.

https://github.com/getredash/redash/issues/5748


redash/query_runner/python.py
```diff
@@ -290,6 +289,25 @@ def get_current_user(self):
...
         pass
 
+    def validate_result(self, result):
+        """Validate the result after executing the query.
+
+        Parameters:
+        :result dict: The result dict.
+        """
+        if not result:
+            raise Exception("local variable `result` should not be empty.")
```
susodapop, line 299: This line will never evaluate `True` because `result` is set explicitly by default on line 107.

susodapop (changes requested):
This is a sensible change. I tested the code paths and found one that never executes. For convenience, how about we log to the user if their query is not going to return any results? We can do it with the following patch:

Since I don't have permission on your fork, can you apply this patch?

```
From b48da0782ff01e727c7549a83db5ceb8320d61fa Mon Sep 17 00:00:00 2001
From: Jesse Whitehouse <jesse@whitehouse.dev>
Date: Sat, 2 Jul 2022 12:52:23 -0500
Subject: [PATCH] Add log message if no result would be displayed

---
 redash/query_runner/python.py | 13 ++++++++++++-
 1 file changed, 12 insertions(+), 1 deletion(-)

diff --git a/redash/query_runner/python.py b/redash/query_runner/python.py
index 9a457e124..8c25c50a4 100644
--- a/redash/query_runner/python.py
+++ b/redash/query_runner/python.py
@@ -295,8 +295,13 @@ class Python(BaseQueryRunner):
         Parameters:
         :result dict: The result dict.
         """
-        if not result:
+
+        if result is None:
             raise Exception("local variable `result` should not be empty.")
+        if self._result_is_empty(result):
+            self._custom_print.write(
+                "No rows to display. Query did not set a non-empty `result` dictionary"
+            )
         if not isinstance(result, dict):
             raise Exception("local variable `result` should be of type `dict`.")
         if "rows" not in result:
@@ -308,6 +313,12 @@ class Python(BaseQueryRunner):
         if not isinstance(result["columns"], list):
             raise Exception("`columns` field should be of type `list`.")
 
+    def _result_is_empty(self, result: dict):
+        rows = result.get("rows")
+        columns = result.get("columns")
+
+        return (rows is None or rows == []) and (columns is None or columns == [])
+
         self._current_user = user
 
-- 
2.32.1 (Apple Git-133)

```

konnectr:
@hoveychen thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

redash/query_runner/python.py
hoveychen: Sorry for the really late response. It seems the issue still exists.
* IMO, the validation is a MSUT for the "user generated input" to improve the robustness of the application.
* Indeed the expected return variable `result` has been set to a default value "before" the python execution. But anything can happen "after" the run, like setting `result` to `None` or `list`.
* I myself did encounter a few times that reuse this common word "result" in other logic by accident.

hoveychen:
Wow. I didn't expect the project getting alive again :)
I've now updated the commit to resolve the conflicts, and review the comment *last year*, it seems to be still valid. Reason attached after the comment.

Thank you for maintaining such a good project.

konnectr (approved):

konnectr:
@hoveychen  How relevant are the tests proposed by codecov ? Could you add the ones that you think are relevant ?

hoveychen:
Added more UTs.

konnectr:
> Added more UTs.

thx, wait ci/cd

konnectr:
@hoveychen Thank you for your contribution !

PR #5753, opened 2022-05-13
New alert destination: Asana

chang:
This adds Asana as an alert destination. Asana tasks are created when an alert is triggers or recovers.

## What type of PR is this? 

- [x] New Alert Destination

## How is this tested?

- [x] Manually

I ran the `docker-compose` stack locally and verified that alerts are successfully created.

image

image

image


justinclift:
The concept seems fine.  It needs some tests though. :smile:

justinclift:
I can't do a technical review of the code (yet), but the concept seems sound.

That being said, this PR needs some kind of tests added to make the Codecov report happy though. :smile:

redash/destinations/asana.py
```diff
@@ -0,0 +1,58 @@
...
+class Asana(BaseDestination):
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "pat": {"type": "string", "title": "Asana Personal Access Token"},
+                "project_id": {"type": "string", "title": "Asana Project ID"},
+            },
+            "secret": ["pat"],
```
guidopetri, line 18: I think we should also make the `pat` (and potentially `project_id`) required - this won't work without the PAT, and I don't know what the behavior is for not defining a project ID.

redash/destinations/asana.py
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+    @classmethod
+    def icon(cls):
+        return "fa-sticky-note-o"
```
guidopetri, line 23: The icon strings are confusing me, could you explain where this string came from and what it means?

redash/destinations/asana.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def notify(self, alert, query, user, new_state, app, host, options):
+        # Documentation: https://developers.asana.com/docs/tasks
+        state = "TRIGGERED" if new_state == "triggered" else "RECOVERED"
```
guidopetri, line 27: Should we reuse `redash.models.Alert` here to avoid magic strings?

redash/destinations/asana.py
guidopetri: Looking through the [docs for the endpoint used below](https://developers.asana.com/reference/createtask), it looks like we would have to set a workspace instead. So I think we should make the project ID required.

redash/destinations/asana.py
```diff
@@ -0,0 +1,58 @@
...
+
+        data = {
+            "name": f"[Redash Alert] {state}: {alert.name}",
+            "notes": notes,
+            "projects": [options.get("project_id")],
+        }
+
+        try:
+            resp = requests.post(
+                "https://app.asana.com/api/1.0/tasks",
```
guidopetri, line 46: Maybe this line should be brought up and be outside the class / a separate method in the class declaration, to make it easier to find the URL being posted in case the API changes?

redash/destinations/asana.py
guidopetri, line 1: As mentioned by @justinclift and the codecov bot, we should probably add some tests :) we could model the test after the Discord destination test in `redash/tests/handlers/test_destinations.py` L95 maybe?

justinclift:
Interestingly, the GitHub viewer is refusing to display the .png file.  I'm not sure if that's meaning there is a real problem with the .png, or if there's a problem with GitHub (again).

guidopetri:
I just merged in #5230 which will require a small change to this PR as well.

guidopetri:
It's a poor test, but it's what we have for the Discord destination, so I figured it'd be fine for now.

guidopetri:
Nice, we're passing codecov now. It should merge then :) @chang thanks for the contribution!

justinclift:
Cool, good stuff @chang and @guidopetri. :smile:

PR #5756, opened 2022-05-20
Dynamodb can assume AWS IAM role

enabokov:
## What type of PR is this? 

- [X] Feature

## Description

DynamoDB data source can use only AWS credentials.\
This PR adds ability for DynamoDBSQL class to assume AWS IAM role in addition.

To allow DynamoDBSQL use AWS IAM role instead of AWS credentials, `DYNAMODB_SQL_ASSUME_ROLE=true` must be set in `env` file for redash.

## How is this tested?

- [X] Manually

1. added `DYNAMODB_SQL_ASSUME_ROLE=true` into the `env` file.
2. restarted docker containers
3. In Redash UI, `Settings` -> `New Data Source` -> chose `DynamoDB (with DQL)`
4. Entered `DynamoDB (with DQL)` settings

redash-dynamodb-assume-role

### DIFF
--- a/redash/query_runner/dynamodb_sql.py
+++ b/redash/query_runner/dynamodb_sql.py
@@ -1,13 +1,16 @@
 import logging
-import sys
+import os
 
 from redash.query_runner import *
+from redash.settings import parse_boolean
 from redash.utils import json_dumps
 
 logger = logging.getLogger(__name__)
+ASSUME_ROLE = parse_boolean(os.environ.get("DYNAMODB_SQL_ASSUME_ROLE", "false"))
 
 try:
-    from dql import Engine, FragmentEngine
+    import boto3
+    from dql import FragmentEngine
     from dynamo3 import DynamoDBError
     from pyparsing import ParseException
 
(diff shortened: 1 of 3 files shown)


guidopetri:
@enabokov , thanks for the PR! We've updated a lot of things now that we're Community-driven so - if you're still interested in getting this merged - would you mind rebasing off master to re-run the CI, as well as updating merge conflicts?

justinclift:
This PR seems to rely on the Python `dql` library which we dropped a while ago due to dependency issues (at the time).

The dql repo seems mostly dead these days (last updated in 2021), and has this notice at the in the README:

```
As of November 2020, Amazon has released PartiQL support for DynamoDB. You should investigate that
first to see if it addresses your needs.
```

If this PR is still needed, it should *probably use* that.

Though the PartiQL options on PyPi seem to be a mess: https://pypi.org/search/?q=partiql

There's even what *looks* like two libraries that may instead by supply chain attacks (partiql-python, amazon.partiql).  No idea if that's really the case though.

@enabokov What are your thoughts on this PR? :smile:

Issue #5773, opened 2022-06-20
LIMIT 1000 causes sql server data source query to throw error.


chriseaton:
### Issue Summary
Docker deployment - pretty straightforward, in the queries editor, with the "LIMIT 1000" checkbox checked, a Microsoft SQL Server-based query throws an error (maybe because MSSQL doesn't use "LIMIT", it uses "TOP"?). Unchecking the box prevents the issue but of course doesn't limit the results.

### Steps to Reproduce

1. Add a MSSQL connector data source.
2. Make a simple select query with the checkbox checked.
3. Execute, notice the error:
    ![image](https://user-images.githubusercontent.com/5247568/174679721-efe65ee9-0ca1-4179-9eaf-48dd4306517d.png)

Unchecking the checkbox runs the query fine.

My expectation would be that if that LIMIT 1000 checkbox is shown - would be that it would limit the result-set on any supported connector data source.

### Technical details:

* Redash Version: 10.1.0
* Browser/OS: Docker/Linux
* How did you install Redash: Modified setup script.

susodapop:
Thanks for this report. It looks like autolimit has not been implemented for this query runner, but by default Redash assumes it's supported for any query runner that inherits `BaseSQLQueryRunner`. I've opened #5777 explicit mark this feature as not enabled for this query runner.

We'll rely on someone else in the community to properly implement auto-limit for this query runner. I'm happy to provide pointers on how to do this if needed.

PR #5777, opened 2022-07-02
Disable auto limit for mssql query runner

susodapop:
## What type of PR is this? 

- [x] Bug Fix

## Description

Microsoft SQL Server uses TSQL syntax which uses `TOP` rather than `LIMIT` to truncate results. Therefore the default implementation of autolimit in `BaseSQLQueryRunner` doesn't work.

This PR explicits marks auto limit as not supported so that the `LIMIT 1000` tickbox will not appear in the query editor. A subsequent PR should hopefully implement auto limit for the query runner.

## How is this tested?

- [x] Manually

#5773
#### Before
CleanShot 2022-07-02 at 11 11 20@2x

#### After

CleanShot 2022-07-02 at 11 07 06@2x

### DIFF
--- a/redash/query_runner/mssql_odbc.py
+++ b/redash/query_runner/mssql_odbc.py
@@ -68,6 +68,10 @@ class SQLServerODBC(BaseSQLQueryRunner):
     @classmethod
     def type(cls):
         return "mssql_odbc"
+    
+    @property
+    def supports_auto_limit(self):
+        return False
 
     def _get_tables(self, schema):
         query = """


mattdjones:
hey @susodapop  we only just upgraded to v10 and we're seeing this error with the non-odbc mssql runner when LIMIT is checked, is it worth applying the same property to that runner?

susodapop:
The non odbc query runner is deprecated. You can of course make any changes you desire. But going forward we recommend using the ODBC version instead.

susodapop:
Marking the auto limit as specifically unsupported shouldn't affect anything about how your queries execute so I think it's a safe change to make locally. All it does is hide the "LIMIT 1000" UI in the query editor.

mattdjones:
ah nice thanks @susodapop - didn't know it was deprecated already 👍 
We knew it was going to be, we've just been a bit behind the releases.

PR #5175, opened 2020-09-14
Static SAML configuration and assertion encryption

christophergrant:
## What type of PR is this? (check all applicable)

- [x] Feature

## Description

1. Static SAML conf

- Some IdPs provide metadata URLs, Okta refers to this scheme as ["dynamic" configuration](https://developer.okta.com/docs/guides/build-sso-integration/saml2/specify-your-settings/#specify-saml-settings). However, some IdPs do not provide this metadata URL, and the alternative is to "statically" specify an SSO URL, Entity ID, and x509 certificate on the client-side.
- With this PR, this is configurable through the UI and environment variables.
2. SAML Response encryption
- It may be desirable to encrypt the SAML response from the IdP to the SP. 
- With this PR, this is configurable through environment variables.
## Related Tickets & Documents
N/A
## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
Recording: [https://imgur.com/exc6XPO](https://imgur.com/exc6XPO)

PR #5784, opened 2022-07-06
bug fix: SAML_LOGIN_ENABLED setting logic

ism-k:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

I would like to point out that an existing `SAML_LOGIN_ENABLED` may have inappropriately strict conditions for it to be true.

I first identified this problem as a bug in the following PR, which I attach as a reference:
https://github.com/getredash/contrib-helm-chart/issues/122

In short, due to a change merged from v10 in PR https://github.com/getredash/redash/pull/5175, SAML cannot be enabled at startup unless an environment variable `REDASH_SAML_SSO_URL` is set, but this variable is actually unnecessary in the case of dynamic SAML login.

So I'm adding a new conditional branch to avoid `SAML_SSO_URL != ""` when the SAML login type is not static.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


kenchan0130:
@susodapop

This change is very important for redash admin users with SAML settings.
The change itself also appears to be appropriate.

susodapop:
I'm not currently contributing to the Redash project.

acefei:
Highly recommended this merge to the trunk branch, so that I don't need to set the redundant env by `kubectl set env deployment/my-release-redash REDASH_SAML_SSO_URL=xxxxx` to enable SAML login.
Given @susodapop didn't contribute to the Redash project, could @ism-k invite others to approve this?

acefei:
ping @arikfr

arikfr (approved):
👍

ism-k:
@acefei 
Thanks for facilitating!

arikfr:

Issue #5754, opened 2022-05-17
MySQL INFORMATION_SCHEMA tables are not ignored


jsvisa:
### Issue Summary

A summary of the issue and the browser/OS environment in which it occurs.

MySQL INFORMATION_SCHEMA tables are not ignored in schemas:

<img width="303" alt="image" src="https://user-images.githubusercontent.com/3627395/168710764-79e330a5-9121-46e8-8d22-c0464c50b166.png">

 
Seems to be the problem here,

https://github.com/getredash/redash/blob/46ea3b1f0b263ba310de7c2c17e4a32d3ea4c343/redash/query_runner/mysql.py#L125-L133

MySQL uses uppercase `INFORMATION_SCHEMA`, but filtering by lowercase `information_schema` here

### Steps to Reproduce

1. This is the first step
2. This is the second step, etc.

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* Redash Version: v10.1.0
* Browser/OS: Chrome
* How did you install Redash: docker-compose

susodapop:
> MySQL uses uppercase INFORMATION_SCHEMA, but filtering by lowercase information_schema here

mysql is not case-sensitive by default. 

Furthermore I'm not able to reproduce this on mysql:latest or on mariadb:latest. If you continue to experience issues here please open a discussion thread on our [user forum](discuss.redash.io). This appears to be something unique to your environment and not a bug in the Redash code.

PR #5808, opened 2022-07-29
Readme: add MariaDB to supported data sources

luc-x41:
## What type of PR is this? 

- [x] Other

## Description

As per [this pull request](https://github.com/getredash/redash/pull/4061)[1] and also [this developer comment](https://github.com/getredash/redash/issues/5754#issuecomment-1173147163)[2] it seems that MariaDB support is something Redash has already had for years. The current list in `README.md` shows only MySQL which, being from Oracle and often just being an alias for MariaDB, is maybe not the thing most people will actually be using going forward.

This pull request explicitly adds MariaDB to the list of supported data sources. For the commit message I've used the style from [this other recent commit](https://github.com/getredash/redash/commit/0712abb3590046d96d7252e7e4575875aefe5b84).

[1] (2019) "Switch to mysqlclient from Python-MySQL" for py3 compatibility and MariaDB compatibility
[2] (2022) Redash contributor validating a bug report with both mysql and mariadb

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Redash supports more than 35 SQL and NoSQL [data sources](https://redash.io/help
 - JSON
 - Apache Kylin
 - OmniSciDB (Formerly MapD)
+- MariaDB
 - MemSQL
 - Microsoft Azure Data Warehouse / Synapse
 - Microsoft Azure SQL Database


luc-x41:
Maybe to add a rationale: I ctrl+f'd for mariadb when looking for compatibility and, given how commonly mariadb is used and how many engines redash supports, I was surprised it wasn't supported. Then I figured I should also look for the legacy mysql name, and found that it is listed, and started investigating if mariadb is purposefully omitted (this could be because there are some differences: I've run into issues before where something works fine on mariadb but the feature was not backported to mysql so someone else got a syntax error). So far as I can tell, there is no reason for the omission and it's treated the same as any other engine on that list.

Listing both when there is compatibility with both seems like the best solution to me.

susodapop (approved):
Excellent catch. Oversight on our part. Thank you! 🙏

PR #5809, opened 2022-08-02
Fix: Query results in public dashboard not loading when query ends wi…

alexStrickner00:
## What type of PR is this? 

- [x] Bug Fix

## Description
Dashboards accessed using the public url use the cached query result mechanism to fetch results after a query job is finished.
When using parameterized queries, the frontend requests the data usually 2 times: one time after changing the parameter, and one time after the job is finished. The second request relies on the caching mechanism to find the cached result of the first request. The second request does not request auto_limit, which causes a lack of splitting and merging of sql statements. Because of that, the two sql queries differ in whitespacing and a trailing semicolon, which causes the query_hashes of both requests to differ.

That results in another job invocation and an empty query result in the second response in the frontend, which causes an endless loading spinner in the frontend and no fetched data. 

By splitting the statements each time, you prevent these differences in the 2 sql queries.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Splitting and merging statements should/can not break anything, otherwise the auto limiting would be buggy.

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -312,16 +312,14 @@ class BaseSQLQueryRunner(BaseQueryRunner):
         return str(parsed_query)
 
     def apply_auto_limit(self, query_text, should_apply_auto_limit):
+        queries = split_sql_statements(query_text)
         if should_apply_auto_limit:
-            queries = split_sql_statements(query_text)
             # we only check for last one in the list because it is the one that we show result
             last_query = queries[-1]
             if self.query_is_select_no_limit(last_query):
                 queries[-1] = self.add_limit_to_query(last_query)
-            return combine_sql_statements(queries)
-        else:
-            return query_text
-
+        return combine_sql_statements(queries)
+        
 
 class BaseHTTPQueryRunner(BaseQueryRunner):
     should_annotate_query = False


guidopetri:
@alexStrickner00 thanks for your contribution! I've rebased this branch on `master` to re-run the CI.

guidopetri:
Looks like we are not passing D: would you mind submitting a fix?

manikBS:
its a flake error W293 blank line contains white-space
so removing the new line at the end of apply_auto_limit function will fix it

redash/query_runner/__init__.py
```diff
@@ -312,16 +312,14 @@ def add_limit_to_query(self, query):
...
             # we only check for last one in the list because it is the one that we show result
             last_query = queries[-1]
             if self.query_is_select_no_limit(last_query):
                 queries[-1] = self.add_limit_to_query(last_query)
-            return combine_sql_statements(queries)
-        else:
-            return query_text
-
+        return combine_sql_statements(queries)
+
```
manikBS, line 322: its a flake error W293 blank line contains white-space
so removing the new line at the end of apply_auto_limit function will fix it

justinclift: If the `pre-commit` hook has been set up, then that should automatically format things correctly when the git commit is created:

https://github.com/getredash/redash/wiki/Local-development-setup#configuring-pre-commit

If that's not setup, you can manually run the `make format` command at any time to do the same thing.  You'll need the Python virtual environment to be activate (with everything installed) for it to work though.

manikBS: Thanks @justinclift. Should i copy the above changes to my branch, fix formatting and create pull request?
Asking this since the above changes by @alexStrickner00 are not present in any branch in getredash/redash repository.

justinclift: @manikBS I'd personally go with cherry-picking @alexStrickner00's changes across to your repo (eg in a branch or something) in order to keep the authorship correct, and add your stuff afterwards as new commits.  Then make a PR from that. :smile:

That should (in theory) make for a PR we can merge, and that has both @alexStrickner00 and yourself correctly credited as authors.

manikBS: submitted new PR with the above format fix https://github.com/getredash/redash/pull/6351

konnectr:
next develop in the pr https://github.com/getredash/redash/pull/6351

PR #5812, opened 2022-08-04
fix #5811: passing server,port to MS ODBC driver

### LINKED ISSUE #5811: Timing out when connecting to a MSSQL database on non-default port using ODBC driver
I had to use "Microsoft SQL Server (ODBC)" data source because the "Microsoft SQL Server" one does not currently support using SSL. However, when trying to connect to my server on a port different than 1433, connection timed out.

After a bit of digging, I found this:
> Microsoft's ODBC drivers for SQL Server do not use a PORT= parameter. The port number, if any, is appended to the server name/IP with a comma

source: https://stackoverflow.com/a/50051708/1277401

JuanjoFuchs:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

fixing #5811 

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [x] N/A


susodapop:
Thanks for this contribution. Is the change backwards compatible for existing connections?

justinclift:
This seems correct, at least according to the SQL Server connection string format under "Using a non-standard port" heading here:

https://www.connectionstrings.com/microsoft-odbc-driver-17-for-sql-server/

justinclift:
Thanks for making this PR @JuanjoFuchs. :smile:

Sorry it took so long to get to.

Issue #5369, opened 2021-01-27
invalid date in text column  


sbvitok:
### Issue Summary

I read logs from ClickHouse via Redash and have a problem with invalid date message in text column.
![изображение](https://user-images.githubusercontent.com/8897595/105992642-6fbe7800-60b6-11eb-9449-2c7e8065fa7d.png)

SQL
```select datetime, pod_name, message from logs.raw where datetime >= '{{ Date.start }}' and datetime <= '{{ Date.end }}' and app_name = '{{app_name}}' and (message {{message}}) order by datetime,timestamp desc limit 5000```

output form postgres
```
Copy (select * from query_results) To '/tmp/test.csv' With CSV DELIMITER ',' HEADER;

cat /tmp/test.csv

[...]

PR #5819, opened 2022-09-01
Fix: Query result datetime column type check in frontend (#5369)

Chanmoro:
## What type of PR is this? 

- [x] Bug Fix

## Description

This PR fix colum type check bug in frontend (#5369) due to fuzzy datetime type checking. 
Text which starts with datetime (YYYY-MM-DDT) followed by some string will be detected as a string by this fix.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

We have tested run query and view results in browser by prepare sqlite database which contains bug causes text.

before
image

after
image

### DIFF
--- a/client/app/services/query-result.js
+++ b/client/app/services/query-result.js
@@ -113,6 +113,10 @@ export function fetchDataFromJob(jobId, interval = 1000) {
   });
 }
 
+export function isDateTime(v) {
+  return isString(v) && moment(v).isValid() && /^\d{4}-\d{2}-\d{2}T/.test(v);
+}
+
 class QueryResult {
   constructor(props) {
     this.deferred = defer();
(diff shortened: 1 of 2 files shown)


Avey777:

guidopetri:
@Chanmoro thanks for your contribution! Could you rebase the branch off of `master` in order to run the CI?

justinclift:
This looks like a valid and useful fix, so I just tried manually rebasing this onto our master branch.  Didn't work though, due to permission problems with this PRs branch (no access to push changes), so there's not much that can be done here.

Instead, I've just created PR #6385 with this change rebased onto `master`.  No idea if it'll pass our CI tests or not though.  One way to find out... :wink:

PR #5825, opened 2022-09-13
Update docker-compose.yml: Mount postgresql/data

kaibadash:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

Current `docker-compose.yml` doesn't mount `postgresql/data`.  So we hit by accidents like as follows.

1. I deploy Redash with docker compose.
1. My co-workers add queries.
1. (A few days after) I execute `docker comopse down` to maintain Redash.
1. We lost all queries and settings.

We think that would be very dangerous for people who don't know that.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Sorry for the Japanese documentation. There are people who are having the same problem as I am.
https://kuchitama.hateblo.jp/entry/redash_on_docker_trouble

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,6 +62,8 @@ services:
     restart: unless-stopped
     environment:
       POSTGRES_HOST_AUTH_METHOD: "trust"
+    volumes:
+      - ./postgres_data:/var/lib/postgresql/data
   email:
     image: djfarrelly/maildev
     ports:


justinclift:
Ouch.  The file you're wanting to change is the *development specific* docker-compose.yml.  It's very much not for production use, as it won't persist data like you had happen. :frowning:

When setting up Redash you should (in theory) be using the [setup.sh script here](https://github.com/getredash/setup/blob/master/setup.sh).  That one will put a production oriented docker-compose.yml file in place, which does [persist the data](https://github.com/getredash/setup/blob/7630d6554fdc78baa3f19811d23d42524e7b380e/data/docker-compose.yml#L41-L42).

PR #5832, opened 2022-10-05
Update outdated LICENSE year

Serjobas:
Updated the outdated copyright year to the present.
So there will not be a need for any further license year updates


justinclift:
Doing some looking around online whether the term "present" is valid for copyright notices, it mostly seems to be "no".  So, I'm going to close this. :smile:

PR #5846, opened 2022-10-20
Check dependencies weekly using dependabot

jlosito:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

This will use dependabot to check dependencies which this project uses for updates and submit pull requests with version bumps in order to keep them up-to-date. Dependabot only supports yarn v1, so I didn't include it.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [x] N/A


jlosito:
Was there any reason this was closed without comment?

guidopetri:
@justinclift how does dependabot currently open PRs?

guidopetri:
@jlosito I rebased the PR on `master` to run the CI, but if you could do it yourself + force push that would be great - that way the commit shows up as signed by you.

justinclift:
> how does dependabot currently open PRs?

I'm not sure of the exact mechanism it uses.  In the repo settings related to Dependabot, we've enabled the first two options already:

image

Note that we've not yet enabled the 3rd one.  This PR adds a `dependabot.yml` file, which is the same filename that third option in that image mentions in it's blue text.

So, it sounds like this PR might need us to enable that first?  If that's the case, then I'd probably see how it runs using default settings first, then customise it (like in this PR) if it needs improving.

That being said, I'm not clear on whether we want to enable that third option just yet or not.  There are *already* a large amount of dependabot things we're working through...

Thoughts?

guidopetri:
I agree, we probably shouldn't flood ourselves with new dependency PRs right now

justinclift:
@jlosito The behaviour you're describing sounds like it might be useful *later on*.

Our new "Community driven project" approach means we've only recently (few weeks ago) restarted serious development, so we're still working through a *large* number of open PRs already.  Many have been open several years. :wink:

Once we've completed taking care of that major task, then we'll be in a position to think about wider process improvement & optimisation things (like this).

Well, that's my opinion at the moment anyway. :smile:

PR #5867, opened 2022-12-02
Added New Env variable for Unpublished queries

adithyapathipaka:

- [x] Feature

## Description

Added new Env Variable REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES and updated include_drafts param to get value from REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES

1. Application shows all the unpublished queries from other users on the UI by default, which is due to Hard coded value with `True` for the include_draft param.
2. Added new Env Variable `REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES` (boolean) through which application can be configured to view unpublished queries
3. By Default `REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES`  will be `true` and can be configure through to `false` by adding the variable to env file.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,7 @@ x-redash-environment: &redash-environment
   REDASH_MAIL_SERVER: "email"
   REDASH_ENFORCE_CSRF: "true"
   REDASH_GUNICORN_TIMEOUT: 60
+  REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES: "false"
   # Set secret keys in the .env file
 services:
   server:
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -124,12 +124,12 @@ class BaseQueryListResource(BaseResource):
                 search_term,
                 self.current_user.group_ids,
                 self.current_user.id,
-                include_drafts=True,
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -19,6 +19,7 @@ x-redash-environment: &redash-environment
   REDASH_MAIL_SERVER: "email"
   REDASH_ENFORCE_CSRF: "true"
   REDASH_GUNICORN_TIMEOUT: 60
+  REDASH_FEATURE_VIEW_UNPUBLISH_QUERIES: "false"
```
guidopetri, line 22: Could we switch the default to `true` to maintain the old behavior if nobody does additional changes?

guidopetri (commented):
Thanks for this change @adithyapathipaka ! Would you mind opening an accompanying PR modifying [this file](https://github.com/getredash/website/blob/master/src/pages/kb/open-source/admin-guide/env-vars-settings.md) in the docs to document this new env var and its behavior?

Additionally, I'm rebasing this branch on `master` to run the CI.

guidopetri:
Whoops, that looks like a merge commit instead. We can squash the commits later, heh :)

PR #5869, opened 2022-12-06
[enhancement] Add a CLI command in order to create organization

kenchan0130:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

The existing CLI has the ability to list organizations, but not the ability to create them.

The ability to create an organization in the CLI, for example, automates the setup of redash itself and makes it very efficient.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [x] N/A


redash/cli/organization.py
```diff
@@ -34,6 +34,25 @@ def show_google_apps_domains():
...
+    "--slug",
+    "slug",
+    default="default",
+    help="The slug the organization belongs to (leave blank for " "'default').",
+)
+def create(name, slug="default"):
+    print("Creating organization (%s)..." % (name))
+
+    try:
+        models.db.session.add(models.Organization(name=name, slug="default", settings={}))
```
junnplus, line 49:
```suggestion
        models.db.session.add(models.Organization(name=name, slug=slug, settings={}))
```

konnectr:
I think it's a good PR. What do you think? @justinclift

justinclift:
The concept sounds like it could be useful, so if the code is decent then we should probably merge it. :smile:

justinclift:
For this PR, the Codecov thing is probably right.  We should try to add tests for this. :smile:

justinclift:
Good stuff @konnectr. :smile:

PR #5877, opened 2022-12-19
Update snowflake-connector-python

yonatang-cyera:
The Snowflake parameter MULTI_STATEMENT_COUNT can be altered in this version, which will solve the "Multiple SQL statements in a single API call are not supported; use one API call per statement instead." error

## What type of PR is this? 
- [x] Refactor

## Description
Update requirements for snowflake-connector-python

## How is this tested?
- [x] N/A


junnplus:
closed by https://github.com/getredash/redash/pull/6122

PR #5879, opened 2022-12-22
Improve visibility of error message during schema retrieval

apollonin:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x ] Other

## Description
Right now it is impossible to debug the error during schema update. There is not logs in worker logs as well as response from the server is useless in chrome dev tools: 
image

With the current change we: 
1. Pass the real error message back to the task response:
image

2. Reduce code duplication by consolidating error handling in one place for future improvements.
3. improve bigQuery error handling to parse json for 404 errors as well as for 400th.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

I tested this behavior by reproducing the error during schema update operation. With this new change I can clearly see the issue in dev tools as well as in logs.

### DIFF
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -234,6 +234,13 @@ class BaseQueryRunner(object):
     def get_schema(self, get_stats=False):
         raise NotSupported()
 
+    def _handle_run_query_error(self, error):
+        if error is None:
+            return
+
+        logger.error(error)
+        raise Exception(f"Error during query execution. Reason: {error}")
+
     def _run_query_internal(self, query):
         results, error = self.run_query(query, None)
 
--- a/redash/tasks/general.py
+++ b/redash/tasks/general.py
@@ -85,8 +85,8 @@ def get_schema(data_source_id, refresh):
                 "message": "Data source type does not support retrieving schema",
             }
         }
-    except Exception:
-        return {"error": {"code": 2, "message": "Error retrieving schema."}}
+    except Exception as e:
+        return {"error": {"code": 2, "message": f"Error retrieving schema: {e}"}}
 
 
 def sync_user_details():
--- a/redash/query_runner/athena.py
+++ b/redash/query_runner/athena.py
@@ -206,7 +206,7 @@ class Athena(BaseQueryRunner):
 
         results, error = self.run_query(query, None)
         if error is not None:
-            raise Exception("Failed getting schema.")
+            self._handle_run_query_error(error)
 
         results = json_loads(results)
         for row in results["rows"]:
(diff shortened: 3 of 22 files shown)


redash/tasks/general.py
```diff
@@ -85,8 +85,8 @@ def get_schema(data_source_id, refresh):
                 "message": "Data source type does not support retrieving schema",
             }
         }
-    except Exception:
-        return {"error": {"code": 2, "message": "Error retrieving schema."}}
+    except Exception as e:
+        return {"error": {"code": 2, "message": f"Error retrieving schema: {e}"}}
```
arikfr, line 89: How about we split it into a separate field (`details`?). This way we can better format it in the UI.

redash/query_runner/__init__.py
```diff
@@ -234,6 +234,13 @@ def fetch_columns(self, columns):
     def get_schema(self, get_stats=False):
         raise NotSupported()
 
+    def _handle_run_query_error(self, error):
+        if error is None:
+            return
+
+        logger.error(error)
+        raise Exception(f"Error during query execution. Reason: {error}")
```
arikfr, line 242: I wonder if the `Error during query execution` prefix really needed or it might be confusing? (if the error happens when trying to connect, for example)

arikfr (commented):
Thanks! This indeed will make it easier to debug schema issues. Added some comments.

redash/tasks/general.py
apollonin: Oh yes, this is a good idea. Let me adjust it here. Thank you for review!

redash/query_runner/__init__.py
apollonin: Well, it used to be "Failed getting schema." which covered only specific query to retrieve a schema. However, now this method is being used to handle errors for the run_query method, which could run different types of queries. so I figured for readability it would be nice to specify the action which led to the error. 
For the db connect issue it will still make sense, because technically the error happened during the query operation. However, I'm sure in case of db disconnect we should handle it in a deeper layer and not care during each individual query run.
Let me know what you think.

apollonin: it also aligns with the vision to split `message` and `details` so we can report properly the error to the user. 

I imagine there could be other types of errors like validation, pre\post checks and so on, that is why this prefix would be useful here.

image

apollonin (commented):
@arikfr thank you again for reviewing. I have addressed your comments and updated the PR

arikfr (approved):

arikfr:
👍 thanks.

PR #5902, opened 2023-03-08
feat: New support databend for redash

hantmac:
## What type of PR is this? 

## Description

This PR is able to use [Databend](https://github.com/datafuselabs/databend) as new plugin.

This plugin needs [databend-sqlalchemy](https://github.com/databendcloud/databend-sqlalchemy)

### DIFF
--- a/redash/query_runner/databend.py
+++ b/redash/query_runner/databend.py
@@ -1,0 +1,118 @@
+try:
+    from databend_sqlalchemy import connector
+    import re
+
+    enabled = True
+except ImportError:
+    enabled = False
+
+from redash.query_runner import BaseQueryRunner, register
+from redash.query_runner import TYPE_STRING, TYPE_INTEGER, TYPE_BOOLEAN, TYPE_FLOAT, TYPE_DATETIME, TYPE_DATE
+from redash.utils import json_dumps, json_loads
+
+
+class Databend(BaseQueryRunner):
+    noop_query = "SELECT 1"
+
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "host": {"type": "string", "default": "localhost"},
+                "port": {"type": "int", "default": 8000},
+                "username": {"type": "string", "default": "root"},
+                "password": {"type": "string", "default": ""},
+                "database": {"type": "string"},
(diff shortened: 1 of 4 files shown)


hantmac:
@arikfr Hi bro, would like to help me to review this pr? Thank you very much!

redash/query_runner/databend.py
```diff
@@ -0,0 +1,118 @@
...
+    noop_query = "SELECT 1"
+
+    @classmethod
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "host": {"type": "string", "default": "localhost"},
+                "port": {"type": "int", "default": 8000},
+                "username": {"type": "string", "default": "root"},
```
arikfr, line 24: Looks like it's required and doesn't really have a default value?

redash/query_runner/databend.py
```diff
@@ -0,0 +1,118 @@
...
+            return TYPE_STRING
+
+    def run_query(self, query, user):
+        host = (self.configuration.get("host") or "localhost"),
+        port = (self.configuration.get("port") or 8000),
+        username = (self.configuration.get("username") or None),
+        password = (self.configuration.get("password") or None),
+        database = (self.configuration.get("database") or None),
+        secure = (self.configuration.get("secure") or False),
+        connection = connector.connect(f"databend://{username}:{password}@{host}:{port}/{database}?secure={secure}")
```
arikfr, line 70: This might break if one of the fields has special characters (`:` for example). You need to use something like `urlencode` to properly escape the values first.

arikfr (commented):
Thanks!

redash/query_runner/databend.py
hantmac: Fixed, thanks!

redash/query_runner/databend.py
hantmac: I use the `urlparse(url)` in databend sqlalchemy and it can parse the correct param even the fields has special characters (: for example). 
image

arikfr: Interesting.. we had this issue in the past, but maybe it wasn't using urlparse. 👍

arikfr (approved):

PR #5909, opened 2023-03-15
created an improved version of #3132

keitoueno:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [*] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [*] Manually
- [ ] N/A

Query A needs to be created as follows. The datasource used is MySQL.

```
SELECT customer_id FROM customer WHERE customer_id = {{customer_id}}
Note: The query ID for this query is 2 in my local Redash environment.
```

Query B needs to be created as follows.

```
data = execute_by_query_id(2, {"customer_id": "1234567"})
result = {}
for rows1 in data['rows']:
    add_result_row(result, {
        'customer_id' : rows1['customer_id'],
    })

add_result_column(result, 'customer_id', '', 'string')
```
The result for Query B is the data obtained from Query A (query ID 2).

### DIFF
--- a/redash/query_runner/python.py
+++ b/redash/query_runner/python.py
@@ -2,9 +2,11 @@ import datetime
 import importlib
 import logging
 import sys
+import pystache
+from funcy import distinct
 
 from redash.query_runner import *
-from redash.utils import json_dumps, json_loads
+from redash.utils import json_dumps, json_loads, mustache_render
 from redash import models
 from RestrictedPython import compile_restricted
 from RestrictedPython.Guards import safe_builtins, guarded_iter_unpack_sequence, guarded_unpack_sequence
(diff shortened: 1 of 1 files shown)


justinclift:
I'm not really understanding how this is different from the existing ability to run SQL on the result of a previous query.

???

redash/query_runner/python.py
```diff
@@ -248,10 +273,11 @@ def get_query_result(query_id):
         Parameters:
         :query_id integer: ID of existing query
         """
-        try:
-            query = models.Query.get_by_id(query_id)
-        except models.NoResultFound:
-            raise Exception("Query id %s does not exist." % query_id)
+        # try:
```
justinclift, line 276: Probably not good to leave commented out code around like this. :wink:

konnectr:
I understand the code was taken from here https://github.com/getredash/redash/pull/3132

keitoueno:
> I'm not really understanding how this is different from the existing ability to run SQL on the result of a previous query.
> 
> https://redash.io/help/user-guide/querying/query-results-data-source
> 
> Is it something to do with the use of parameters?

Thank you for your reply. Yes, that's correct. With this customization, you will be able to call a query created in Python with arguments from different resources. In the example, it shows using "customer_id," but in many cases, you might use it for date-based calls. In fact, what I'm using it for is creating queries called from resources like Athena or RDS in Python and wanting to call them from different resources.

PR #5942, opened 2023-03-26
WIP: Python 3.9 + updated dependencies

denisov-vlad:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [x] N/A


konnectr:
maybe it would be better to upgrade to python 3.10?

lucasfcnunes:
@denisov-vlad Are you going to keep working on this? Right now it has some conflicting files. When it's ready for review could you remove the "Draft" status of this PR?

lucasfcnunes:
> maybe it would be better to upgrade to python 3.10?

About this. I do think so. But maybe a python release matrix (with min compatibility at 3.8 or 3.9) would make more sense, @arikfr? I'm saying that but we don't make releases for a long time...

denisov-vlad:
> maybe it would be better to upgrade to python 3.10?

Yes, but it's better to do it iteratively. Some datasource's libraries don't support python3.10 now.

> @denisov-vlad Are you going to keep working on this? Right now it has some conflicting files. When it's ready for review could you remove the "Draft" status of this PR?

Of course, but I faced some issues and don't know how to solve them right now. I will try to return to this PR in July.

justinclift:
With this, we need to move away from Python 3.7 as it's now officially End of Life.

Moving to Python 3.9 seems to be over-shooting though, as Python 3.8 is still widely used and it's official in support until the end of next year.

Thoughts?

konnectr:
> Python 3.8 is still widely used and it's official in support until the end of next year.


I do not understand what are the advantages of updating to version 3.8? 3.10 already has all the libraries and allows you to write conveniently typed code.

justinclift:
Well, Ubuntu 20.04 is still widely used and that comes with Python 3.8.

That being said, if everyone develops using docker containers then it's probably not that much of an issue.  Containers being able to hold "whatever is needed" (etc).

PR #5970, opened 2023-04-06
Pymongo version upgrade changes

Mukesh-R:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

- The pymongo version has been updated to 4.3.3
- The default `JSONOption` from `bson.json_utils` sets `tz_aware` to `False` from 4.0 (https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#tz-aware-defaults-to-false), so setting value to `True` explicitly in `datetime_parser`
- `collection_names` method has been changed to `list_collection_names` (https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#database-collection-names-is-removed)
- `count` method has been changed and using `len` to get the length from `cursor` since `count` method has been removed from 4.0 (https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#collection-count-and-cursor-count-is-removed)

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/redash/query_runner/mongodb.py
+++ b/redash/query_runner/mongodb.py
@@ -302,19 +302,20 @@ class MongoDB(BaseQueryRunner):
 
         cursor = None
         if q or (not q and not aggregate):
-            if s:
-                cursor = db[collection].find(q, f).sort(s)
-            else:
-                cursor = db[collection].find(q, f)
-
-            if "skip" in query_data:
-                cursor = cursor.skip(query_data["skip"])
-
-            if "limit" in query_data:
-                cursor = cursor.limit(query_data["limit"])
-
             if "count" in query_data:
-                cursor = len(list(cursor))
+                options = {opt: query_data[opt] for opt in ("skip", "limit") if opt in query_data}
+                cursor = db[collection].count_documents(q, **options)
+            else:
+                if s:
+                    cursor = db[collection].find(q, f).sort(s)
+                else:
+                    cursor = db[collection].find(q, f)
+
+                if "skip" in query_data:
+                    cursor = cursor.skip(query_data["skip"])
+
+                if "limit" in query_data:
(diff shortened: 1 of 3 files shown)


redash/query_runner/mongodb.py
```diff
@@ -313,7 +314,7 @@ def run_query(self, query, user):
                 cursor = cursor.limit(query_data["limit"])
 
             if "count" in query_data:
-                cursor = cursor.count()
+                cursor = len(list(cursor))
```
arikfr, line 317: Based on the documentation, I think this supposed to be `count_documents`?

arikfr (commented):
Thank you for taking a stab at this. See comment.

redash/query_runner/mongodb.py
Mukesh-R: Changed to get count using `count_documents`

redash/query_runner/mongodb.py
```diff
@@ -301,19 +302,20 @@ def run_query(self, query, user):
 
         cursor = None
         if q or (not q and not aggregate):
-            if s:
-                cursor = db[collection].find(q, f).sort(s)
+            if "count" in query_data:
```
arikfr, line 305: Seems like indentation/order changed here? Was it intentional?

Mukesh-R: Yes, this is intentional. We don't need to run the `find` if it is a count query.

justinclift:
In theory, this PR looks like a better MongoDB update than PR #6114.

So, I just rebased this onto our master branch so the CI tests will run.

If the CI tests fully pass, this is the better one to merge. :smile:

PR #5996, opened 2023-04-19
Add support for a default alert template

eradman:
## Description

Add support for a default alert template

* Escape all variables by default since Mustache only has a syntax for raw values
* Generate a generic HTML result table to allow an alert template to display any query
* Optionally allow alert template to be defined to be set using `REDASH_ALERTS_DEFAULT_MAIL_BODY_TEMPLATE`

redash-alert-ok
redash-alert-triggered


Avey777:

guidopetri:
@eradman thanks for your contribution! Would you mind rebasing on `master` and pushing to rerun CI checks?

guidopetri (approved):

guidopetri:
This LGTM! Would you mind also opening a PR for docs for `REDASH_ALERTS_DEFAULT_MAIL_BODY_TEMPLATE` in [this repo/file](https://github.com/getredash/website/blob/master/src/pages/kb/open-source/admin-guide/env-vars-settings.md?plain=1)?

cc @justinclift

eradman:
Certainly! Proposed an update to the docs in https://github.com/getredash/website/pull/673

junnplus (changes requested):
We should have test coverage for them.

eradman:
I will write some test cases update this PR

justinclift:
Awesome. :smile:

eradman:
I added tests to verify that mustache variables are rendered correctly in the context of an alert. Early next week I'll review this change again and see if there are other tests that should be included.

justinclift:
Thanks @eradman, sounds like a plan. :smile:

eradman:
There does not seem to be an existing set of test cases for `redash/destinations/*`, so I did not write a test for the change to the email destination itself.

I would say that the mustache template rendering is covered reasonably.

redash/settings/__init__.py
```diff
@@ -312,7 +312,13 @@ def email_server_is_configured():
 )
 
 ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE = os.environ.get(
-    "REDASH_ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE", "({state}) {alert_name}"
+    "REDASH_ALERTS_DEFAULT_MAIL_SUBJECT_TEMPLATE",
+    "Alert: {alert_name} changed status to {state}"
+)
+
+REDASH_ALERTS_DEFAULT_MAIL_BODY_TEMPLATE = os.environ.get(
```
junnplus, line 319: It might be clearer to call it "REDASH_ALERTS_DEFAULT_MAIL_BODY_TEMPLATE_FILE" to indicate that it is a file path.

redash/models/__init__.py
```diff
@@ -1036,6 +1038,28 @@ def render_template(self, template):
...
+            for col in columns:
+                if row[col] is None:
+                    html.write("<td></td>")
+                else:
+                    html.write(f"<td>{row[col]}</td>")
+            html.write(f"</tr>")
+        html.write("\n</tbody>")
+        html.write("</table>")
+        html.seek(0)
+
```
junnplus, line 1062: I think these should be implemented in the template rather than here.

justinclift: I'm not that great at Python yet, so I'm avoiding doing technical reviews of the Python bits for now.

If it was Go based, then I'd be good... but it's not. :wink:

justinclift: Both @wlach and @guidopetri seem decent at Python, so maybe ask them? :smile:

guidopetri: I don't think there's an elegant way of doing this in the template, so I'm good with it like this. We can always improve upon it later.

eradman: I do not think it is possible to move this logic to the template because the rows are provided as a dictionary.
This is as close as I could get:

```html
<table>
  <thead>
    <tr>
    {{#QUERY_RESULT_COLS}}
    <th>{{friendly_name}}</th>
    {{/QUERY_RESULT_COLS}}
  </tr>
  </thead>

{{#QUERY_RESULT_ROWS}}
  <tr>
    <td>{{Volume}}</td>
    <td>{{Max Depth}}</td>
    <td>{{Average Depth}}</td>
  </tr>
{{/QUERY_RESULT_ROWS}}
</table>
```

(Notice that `Volume`, `Max Depth`, `Average Depth` are column names and are specific to this query.)

If each row was a list we could write an inner loop for each column, so perhaps the right approach is to provide another view on the data that is more in line with the way Mustash works.

junnplus: I don't think passing an HTML string to a template is a good approach. The purpose of a template is customization, and providing the rendered result to the template is an anti-pattern.

eradman:
We're getting closer: I was able to refactor this so that the query contents are rendered in the template.

Also access to MailDev was broken: we need to map port 1025 and make sure Redash uses that port to send mail. I can push this as a separate commit.

redash/settings/__init__.py
eradman: done

justinclift:
> Also access to MailDev was broken: we need to map port 1025 and make sure Redash uses that port to send mail. I can push this as a separate commit.

As a thought for this, please make sure it's not something too hard coded.  Not everyone needs to use the docker container for sending emails.

eg in the production instances I manage, we just have things going directly via [SMTP2GO](https://www.smtp2go.com), no docker container needed

eradman:
> As a thought for this, please make sure it's not something too hard coded. Not everyone needs to use the docker container for sending emails.

In this case, `docker-compose.yml` is only used for testing

```yaml
# For a production example please refer to getredash/setup repository on GitHub.
```

justinclift:
Cool. :smile:

eradman:
Fixed flake8 and rebased

eradman:
Oh because I was using black-21 (from Ubuntu 22). Upgraded to black-23 ran format again

justinclift:
@eradman It looks like this is pretty much ready to merge yeah?

Also, is there a clear way forward for this?

> Also access to MailDev was broken: we need to map port 1025 and make sure Redash uses that port to send mail. I can push this as a separate commit.

Personally speaking, I'm not bothered either way if the port mapping is a separate ~commit~PR or in this one.  There are advantages (integrated testing of both at once) and disadvantages (all commits get squashed into one) either way, so your choice I reckon. :smile:

eradman:
This is ready to to go. I manually confirmed that email alerts are formatted properly and I'd say the implementation is in good shape.

I created PR https://github.com/getredash/redash/pull/6186 with the changes to the docker port mapping. I believe if we merge that first this change will apply cleanly.

justinclift:
Cool, all done.  This seems like it'll be a useful addition for people @eradman. :smile:

eradman:
Indeed; I'm sure others will find this change useful. Thanks for mentoring this along @justinclift @guidopetri @junnplus

PR #5997, opened 2023-04-20
Fix invalid comparison causing excessive parameter updates

lightpriest:
## What type of PR is this? 

- [x] Bug Fix

## Description

This caused Safari to block the page where the history.pushState was called too many times in large dashboards.

The component was updated many times, and this wrong comparison was always true, causing the code to eventually call "`history.replaceState()` more than 100 times per 30 seconds".

## How is this tested?

- [x] Manually

While the bug itself is a componentDidUpdate that compares the wrong thing, the manifestation of it is not that terrible. I checked and verified that it does not reproduce in Safari.

### DIFF
--- a/client/app/components/Parameters.jsx
+++ b/client/app/components/Parameters.jsx
@@ -1,4 +1,4 @@
-import { size, filter, forEach, extend } from "lodash";
+import { size, filter, forEach, extend, isEqual } from "lodash";
 import React from "react";
 import PropTypes from "prop-types";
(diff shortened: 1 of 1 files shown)


Avey777:

wlach:
This looks fine. The build should pass if I rebase on main, which I will try to do now.

wlach:
Somebody else actually fixed the same issue in a slightly different way, but it looks like there were some regressions (which I didn't expect). Let's close this in favour of the other implementation: https://github.com/getredash/redash/pull/5586

PR #6000, opened 2023-04-24
enum parameters add comment as alias

gaecoli:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

image
image

For the dropdown list, there should be an alias note for everyone to check.

### DIFF
--- a/client/app/components/ParameterValueInput.jsx
+++ b/client/app/components/ParameterValueInput.jsx
@@ -1,4 +1,4 @@
-import { isEqual, isEmpty, map } from "lodash";
+import { isEqual, isEmpty } from "lodash";
 import React from "react";
 import PropTypes from "prop-types";
 import SelectWithVirtualScroll from "@/components/SelectWithVirtualScroll";
@@ -40,6 +40,7 @@ class ParameterValueInput extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
+      reg: /\/\*(.*)\*\//i,
       value: props.parameter.hasPendingValue ? props.parameter.pendingValue : props.value,
       isDirty: props.parameter.hasPendingValue,
     };
@@ -103,7 +104,12 @@ class ParameterValueInput extends React.Component {
         mode={parameter.multiValuesOptions ? "multiple" : "default"}
         value={normalize(value)}
         onChange={this.onSelect}
-        options={map(enumOptionsArray, opt => ({ label: String(opt), value: opt }))}
+        options={enumOptionsArray.map((option) => {
+          let alias = option;
+          const groups = this.state.reg.exec(option);
+          if ((groups !== null) && (groups[1] !== null)) alias = groups[1];
+          return ({ label: alias, value: option });
+        })}
         showSearch
         showArrow
         notFoundContent={isEmpty(enumOptionsArray) ? "No options available" : null}


gaecoli:
image
image

gaecoli:
@arikfr This PR i think is a good feature.

justinclift:
Looks like a good idea to me too.  We'd want someone other than me to do a code review at this stage though. :smile:

justinclift:
@wlach Any interest in reviewing this? :smile:

client/app/components/ParameterValueInput.jsx
```diff
@@ -40,6 +40,7 @@ class ParameterValueInput extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
+      reg: /\/\*(.*)\*\//i,
```
gabrieldutra, line 43: This regex does not need to be a state, you can move it up as a `const`:

```js
const REMOVE_COMMENTS_REGEX = /\/\*(.*)\*\//i;

function ParameterValueInput() {
  // then use it here
}

File: client/app/components/ParameterValueInput.jsx [hunk 2]
```diff
@@ -103,7 +104,12 @@ class ParameterValueInput extends React.Component {
         mode={parameter.multiValuesOptions ? "multiple" : "default"}
         value={normalize(value)}
         onChange={this.onSelect}
-        options={map(enumOptionsArray, opt => ({ label: String(opt), value: opt }))}
+        options={enumOptionsArray.map((option) => {
```
gabrieldutra, line 107: Since this file is still in JavaScript (and a good part of the code too), it's better to leave the `lodash` `map` for now, just to be safe 😅

client/app/components/ParameterValueInput.jsx
gabrieldutra: And thinking about edge cases: what if someone had `/* */` as part of their actual values? 🙂

gabrieldutra (commented):
Thinking about the functionality itself: I wonder what could be a real use case for having comments on enum values @justinclift @gaecoli. And if that's really worth it to maintain a regex there 😅

For example, I'd find it more interesting if we could have `{ value, label }` for the enum definitions

client/app/components/ParameterValueInput.jsx
gaecoli: > And thinking about edge cases: what if someone had `/* */` as part of their actual values? 🙂

In SQL, it't a comment. So it's ok!

gabrieldutra: Yes, but there are also non-SQL data sources and in SQL it could be inside strings as well, e.g.: `select * from table where column = "{{ param }}"`

justinclift: > there are also non-SQL data sources

Oh, that's a really good point.

justinclift: To me, the `{ value, label }` syntax seems like it would provide the same capability, but also be easier to understand + less likely to go wrong.

gaecoli: > To me, the `{ value, label }` syntax seems like it would provide the same capability, but also be easier to understand + less likely to go wrong.

I will pefer this PR.

guidopetri:
> | To me, the { value, label } syntax seems like it would provide the same capability, but also be easier to understand + less likely to go wrong.
>
> I will pefer this PR.

@gaecoli does this mean we should close this PR and you'll open a new one with `{value, label}` syntax, or that you prefer this PR to the value/label syntax?

PR #6006, opened 2023-04-26
add i18n support

gaecoli:
## What type of PR is this? 

- [ ] Refactor
- [ x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


gaecoli:
I add i18n module but i need help to translate English, Chinese and French so on. So i need help language json file.

gaecoli:
If someone support language file, the language pack switching function on the interface should be added later, and the js-cookie package should be used to record the language pack used last time.

lucasfcnunes:
I can translate to Portuguese :)
Is it gonna be possible to integrate to services like https://hosted.weblate.org/projects/nekox/#languages?

gaecoli:
> I can translate to Portuguese :) Is it gonna be possible to integrate to services like https://hosted.weblate.org/projects/nekox/#languages?

Yes

gabrieldutra (commented):
LGTM, thanks for adding it! Only thing I'd probably do is to keep it disabled until we have more internationalized strings not to have random translated words in the UI

gaecoli:
> Btw @gaecoli how are you doing the string extraction? Is it manually in this PR for now and still needs to be configured?

There is no good idea for this, for me it can only be added manually.

client/app/index.js
```diff
@@ -5,6 +5,7 @@ import "@/config";
 
 import ApplicationArea from "@/components/ApplicationArea";
 import offlineListener from "@/services/offline-listener";
+import "./i18n/index";
```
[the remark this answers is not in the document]
gaecoli: ok, i will change it.

gabrieldutra:
> There is no good idea for this, for me it can only be added manually.

Sure, adding something like https://www.i18next.com/how-to/extracting-translations#2-using-an-extraction-tool to either automate the English strings extraction, or to have `yarn i18n:extract` and `yarn i18n:check` scripts (the later to be executed on CI)

I haven't used `react-i18next` myself, so I don't really know if that's required

gaecoli:
@gabrieldutra Hello, I would like to ask, what should I do after I modify the source code, so that I can achieve the same code style as the original? Which front-end commands should I run?

package.json
```diff
@@ -58,6 +58,8 @@
     "font-awesome": "^4.7.0",
     "history": "^4.10.1",
     "hoist-non-react-statics": "^3.3.0",
+    "i18next": "^22.4.15",
+    "i18next-browser-languagedetector": "^7.0.1",
```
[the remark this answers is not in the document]
gaecoli: > From CI failure it seems it's missing the `yarn.lock` changes, so you may need to re-run `yarn` and push it :)

I pushed yarn.lock file.

client/app/i18n/index.js
```diff
@@ -0,0 +1,22 @@
+import i18n from "i18next";
+import LanguageDetector from "i18next-browser-languagedetector";
+import { initReactI18next } from "react-i18next";
+import en from "./locales/en.json";
+import zh from "./locales/zh.json";
```
justinclift, line 5: Out of curiosity, with this `zh` I've often seen it split into two.  One for "Simplified Chinese", and one for "Traditional Chinese".

Should we have this named as one of them, to leave room for the other one at some future point?

gaecoli: > Out of curiosity, with this `zh` I've often seen it split into two. One for "Simplified Chinese", and one for "Traditional Chinese".
> 
> Should we have this named as one of them, to leave room for the other one at some future point?

OK. You are right.

gabrieldutra: One note is that importing languages like this will add all of them into the bundle, which may not be nice 😅.

It's possibly better to do an async import for those cases. https://www.i18next.com/how-to/add-or-load-translations had a bit of content on this.

---

Separate thing I was thinking about these days: Should we plan ahead how we will do the translations end-to-end, before committing anything? I don't know much about `react-i18next`, I've used mostly [react-intl](https://www.npmjs.com/package/react-intl), the main difference I saw so far is that the later also has a `description` for each unformatted string. Considering Redash as an open-source / community effort and how LLMs are today, it may be a good thing to plan a flow which includes automatic generation of the translation files on every release. Then would count on the community only for reviewing/fixing eventual mistakes.

gaecoli: > One note is that importing languages like this will add all of them into the bundle, which may not be nice 😅.
> 
> It's possibly better to do an async import for those cases. https://www.i18next.com/how-to/add-or-load-translations had a bit of content on this.
> 
> Separate thing I was thinking about these days: Should we plan ahead how we will do the translations end-to-end, before committing anything? I don't know much about `react-i18next`, I've used mostly [react-intl](https://www.npmjs.com/package/react-intl), the main difference I saw so far is that the later also has a `description` for each unformatted string. Considering Redash as an open-source / community effort and how LLMs are today, it may be a good thing to plan a flow which includes automatic generation of the translation files on every release. Then would count on the community only for reviewing/fixing eventual mistakes, in this case the `description` would be useful for providing context and/or fixing the prompt.

Cool. It's a good idea.

gaecoli: > One note is that importing languages like this will add all of them into the bundle, which may not be nice 😅.
> 
> It's possibly better to do an async import for those cases. https://www.i18next.com/how-to/add-or-load-translations had a bit of content on this.
> 
> Separate thing I was thinking about these days: Should we plan ahead how we will do the translations end-to-end, before committing anything? I don't know much about `react-i18next`, I've used mostly [react-intl](https://www.npmjs.com/package/react-intl), the main difference I saw so far is that the later also has a `description` for each unformatted string. Considering Redash as an open-source / community effort and how LLMs are today, it may be a good thing to plan a flow which includes automatic generation of the translation files on every release. Then would count on the community only for reviewing/fixing eventual mistakes, in this case the `description` would be useful for providing context and/or fixing the prompt.

I haven't figured out which platform to host these mapping tables for the time being.

justinclift: As a data point - for a project many years ago - we had the project hosted on Launchpad, so it could make use of the translation infrastructure there.

One *really* cool benefit of that platform at the time was that if anyone, anywhere had translated the same string for any project on Launchpad (with compatible license) then the string was made available as a potential option.

It saved a *lot* of time for people, and made it easy to get pretty far with new language translations.

That being said, the company behind Launchpad (Canonical) have done some shady stuff in the past so I'm not really a fan of theirs... :wink:

justinclift: The possible use of LLM's for generating "first attempt" translations sounds like a useful idea too.

gaecoli: > The possible use of LLM's for generating "first attempt" translations sounds like a useful idea too.

Cool!

myonlylonely:
I could help translating to Chinese. How could I contribute?

guidopetri:
@gaecoli hey, any interest in continuing this PR? I think we need to split `zh.json` into the simplified/traditional Chinese format, as well as disabling i18n by default for now until we have more translations in (which we should open a separate issue for, I think).

We also need to merge `master` in + resolve conflicts in the yarn lock file.

PR #6022, opened 2023-05-03
fix load picture of user  profile  error

bonashen:
fix load picture of user  profile  error


Avey777:

guidopetri:
@bonashen thanks for your contribution. Could you explain what error you've observed with loading the user profile picture, preferably with a step-by-step of how to reproduce + what you expected instead? I've personally never seen any problems.

PR #6037, opened 2023-05-13
Pin boto3 dependencies in dev requirements as well

wlach:
This fixes the `backend-unit-tests` job in CI, where a too-recent version of boto3 was getting installed.


wlach:
The frontend unit tests are still failing, that's addressed in #6038

PR #6038, opened 2023-05-13
Update Debian version for Node image

wlach:
`14.17` used stretch, which was ancient and seems to no longer support
updates. Let's try updating to bullseye, which should be around for
a while.


wlach:
The unit tests failed, but got further than they did before (they ran). Suspect it's just a flaky test.

arikfr:
Thanks! Both for this and #6037. We do plan to move from CircleCI but if we can get it working in the meantime it's net improvement :)

wlach:
> Thanks! Both for this and #6037. We do plan to move from CircleCI but if we can get it working in the meantime it's net improvement :)

Yeah, I noticed later that this fork has already gotten things running with GitHub actions: https://github.com/RedashCommunity/redash/

It seems that the tests are still failed after merging this for what seem like intermittent issues (sigh), if you have a sec and could re-run them that would be cool. 😄

PR #6112, opened 2023-06-23
fix pivot table style error

gaecoli:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [x] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/client/cypress/integration/visualizations/pivot_spec.js
+++ b/client/cypress/integration/visualizations/pivot_spec.js
@@ -100,10 +100,11 @@ describe("Pivot", () => {
         .type(" UNION ALL {enter}SELECT 'c' AS stage1, 'c5' AS stage2, 55 AS value");
 
       cy.getByTestId("SaveButton").click();
-      cy.getByTestId("ExecuteButton").click();
+      cy.getByTestId("ExecuteButton").click({ timeout: 10000 });
 
       // assert number of rows is 12
       cy.getByTestId("PivotTableVisualization").contains(".pvtGrandTotal", "12");
+      
     });
   });
 


gaecoli:
> ## What type of PR is this?
> * [ ]  Refactor
> * [ ]  Feature
> * [x]  Bug Fix
> * [ ]  New Query Runner (Data Source)
> * [ ]  New Alert Destination
> * [ ]  Other
> 
> ## Description
> ## How is this tested?
> * [ ]  Unit tests (pytest, jest)
> * [x]  E2E Tests (Cypress)
> * [ ]  Manually
> * [ ]  N/A
> 
> ## Related Tickets & Documents
> ## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

Fix CI/CD error with pivot table style.

gaecoli:
@justinclift I think it not work for me.

justinclift:
It looks to be something wrong with that `restyled` check, which we'll need @arikfr to fix.  I've left a message in our Discord about it, so hopefully he gets that done in the next day or so. :smile:

client/cypress/integration/visualizations/pivot_spec.js
```diff
@@ -103,6 +103,7 @@ describe("Pivot", () => {
       cy.getByTestId("ExecuteButton").click();
 
       // assert number of rows is 12
+      cy.wait(250);
       cy.getByTestId("PivotTableVisualization").contains(".pvtGrandTotal", "12");
```
gabrieldutra, line 107: Fine to leave it to fix if the test is broken, but usually it's not a good practice to do arbitrary `cy.wait`, because it adds up time to the tests and there's no guarantee that the element will be ready after that. Instead, the idea is to assert for something else that we know would be a "ready signal" in the UI itself (e.g.: notification, some message that change)

Do you happen to have a link, screenshot or video to the failure it had?

gabrieldutra (approved):
@justinclift restyled is not required, so it's probably fine to merge it without that. If needed later, it's possible

client/cypress/integration/visualizations/pivot_spec.js
gabrieldutra: Btw, it's possible you'll have to ignore the ESLint rule for the line if you want to keep the `cy.wait`

gabrieldutra:
Found it in https://cloud.cypress.io/projects/924cka/runs/947931f3-a95f-40af-a0c7-6a4108130dd5/test-results/8e15f6d2-9b08-4ae4-8c6b-6161f9a197a7

it seems the problem is that we're not re-executing the query?

client/cypress/integration/visualizations/pivot_spec.js
justinclift: That `cy.wait()` call was just my suggestion in Discourse, as I've seen it help (in another project) with specific tests if they commonly fail at a certain point.

So, this was just @gaecoli trying it out to see if it helps. :smile:

justinclift: > @justinclift restyled is not required ...

Oh cool, thanks. :smile:

justinclift:
> it seems the problem is that we're not re-executing the query?

When you say "re-executing the query", are you meaning "not pressing the execute button in the UI"?

It's kind of confusing.  That link to the test in the Cypress cloud is super useful though, and seems like it should help.  Watching the recorded video of the test run:

https://cloud.cypress.io/projects/924cka/runs/947931f3-a95f-40af-a0c7-6a4108130dd5/test-results/8e15f6d2-9b08-4ae4-8c6b-6161f9a197a7/video

At about the 13 second mark in the video, it seems to click the Execute button.  On the left hand side of the video, we see what seems (to me) like it's showing the call to the backend for the execute button just after line "16" -> `POST 200 /api/queries/90...`.  Then what I'm guessing is the `GET` call to fetch the results.

None of that sounds wrong to me though?

gaecoli:
image

@justinclift @gabrieldutra I think it's ok!

justinclift:
I'll defer to @gabrieldutra on this one.  I'm happy for us to merge this if he gives this the ok. :smile:

justinclift:
Hmmm, out of curiosity, I'm going to manually trigger the tests to re-run on this.

Am wanting to see if Cypress randomly passes vs fails this PR, or if it's consistently passing now. :smile:

gaecoli:
> Hmmm, out of curiosity, I'm going to manually trigger the tests to re-run on this.
> 
> Am wanting to see if Cypress randomly passes vs fails this PR, or if it's consistently passing now. 😄

Ok!

gaecoli:
> I'll defer to @gabrieldutra on this one. I'm happy for us to merge this if he gives this the ok. 😄

Agree with u!

client/cypress/integration/visualizations/pivot_spec.js
gaecoli: This is a timeout test. The default time of cypress cannot achieve the purpose, so we specify the timeout at the end when we click, so that the test can pass. I have checked the official documentation of cypress, and there is indeed such a situation.

justinclift:
k, it passed the first re-run of the test, lets do it 2 more times and see... :smile:

gaecoli:
> k, it passed the first re-run of the test, lets do it 2 more times and see... 😄

It's ok, i will re-run another time, lets see result.

gaecoli:
> k, it passed the first re-run of the test, lets do it 2 more times and see... 😄

You triggered test it's ok!

gaecoli:
> > k, it passed the first re-run of the test, lets do it 2 more times and see... 😄
> 
> You triggered test it's ok!

@justinclift  Maybe it's work fine!

justinclift:
Yeah, it passed three times in a row.  I'm thinking that really does fix the problem after all. Cool. :smile:

gaecoli:
> Yeah, it passed three times in a row. I'm thinking that really does fix the problem after all. Cool. 😄

Maybe we can merge it.

justinclift:
Yep, merged.  Hopefully that doesn't turn out to be a wrong move. :wink:

gabrieldutra:
I think the timeout may not solve the issue btw, I remember there was an issue with Antd not adding disabled property correctly in buttons, so Cypress still tries to click on them while disabled 😕.

If the test remains flaky, ping me and I'll try to fix it

justinclift:
Thanks @gabrieldutra, we will. :smile:

gaecoli:
> 

In cypress, `should` may fix this.

PR #6115, opened 2023-06-24
Add feature:  Hide filter components on shared pages

Avey777:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [x] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

image


justinclift:
Hmmm, that `restyled` error doesn't really seem like an error with this PR, but more something wrong with the `restyled` thing itself.

justinclift:
@Avey777 Who's https://github.com/Amiequan? :smile:

justinclift:
That failing `restyled` test should work if we rebase this onto the very latest commit for the `master` branch.  I'll do that shortly.

The failing `build` test I'm not sure about.  That seems to be a problem in our CI setup still.  I'll open a PR in a bit with some changed settings in order to investigate that... :smile:

Avey777:
> @Avey777 Who's https://github.com/Amiequan? 😄

My colleagues and I are all using redash.

justinclift:
Cool.  I was just seeing commits from several authors and it was making me a bit nervous. :wink:

That's no worries at all though. :smile:

justinclift:
Hmmm, that failed `restyled` test was from a few days ago, before we fix it.

If you get the test to run again, I'm pretty sure it will work.

There doesn't seem to be a button for it though, so the only way I know of is to push a change to the PR's branch.  Something like this will probably work:

```
$ git commit --amend
$ git push [standard repo and branch details here] --force
```

justinclift:
k, I've just manually rebased this onto `master`, and also applied the changes suggested by `restyled`.

justinclift:
@junnplus Is this the kind of PR you're ok to review? :smile:

justinclift:
@gabrieldutra Any interest in reviewing this? :smile:

client/app/components/Parameters.jsx
```diff
@@ -43,13 +43,23 @@ export default class Parameters extends React.Component {
     appendSortableToParent: true,
   };
 
+  toCamelCase = str => {
+    if (isEmpty(str)) {
+      return "";
+    }
+    return str.replace(/\s+/g, "").toLowerCase();
```
junnplus, line 50: It seems unrelated to the implementation and function name?

junnplus:
Maybe we should provide users with a visual interface instead of controlling it through a URL.

junnplus:
In addition, in my understanding, if you want to hide components, it should be controlled on the backend rather than the frontend, right?

client/app/components/Parameters.jsx
Avey777: Cookl

Avey777:
> In addition, in my understanding, if you want to hide components, it should be controlled on the backend rather than the frontend, right?

Just hiding some filtering components. The reason for hiding them on the front-end is that there are other places where these hidden components need to be displayed. This is currently a short-term implementation solution.

Avey777:
@justinclift 
I need to use this function now. Can I merge it. Now we will adopt the front-end hidden scheme first, and then change to the back-end scheme when we have time

client/app/pages/dashboards/components/ShareDashboardDialog.jsx
```diff
@@ -117,9 +118,28 @@ class ShareDashboardDialog extends React.Component {
...
-            </Form.Item>
+            <>
+              <Form.Item>
+                <Alert
+                  message={
+                    <div>
+                      Custom rule for hiding filter components when sharing links:
+                      <br />
+                      You can hide filter components by appending `&hide_filter={"{{"} component_name{"}}"}` to
+                      thesharing URL.
```
justinclift, line 129: `thesharing` should be `the shared` :smile:

client/app/components/Parameters.jsx
justinclift: Seems like this `toCamelCase()` function should be removed from the PR.

justinclift (changes requested):

justinclift:
> I need to use this function now. Can I merge it.

It just needs some minor adjustments, then you're welcome to merge it. :smile:

---

> Maybe we should provide users with a visual interface instead of controlling it through a URL.

@junnplus That's a decent idea, but lets go with this "easy" approach first and see if it works ok in practise.  We can figure out a visual interface for it later. :smile:

justinclift:
> My colleague doesn't have much time to study this.

Ok.  I've just made the requested changes.  Lets see if it passes our CI tests, then we can merge it. :smile:

justinclift:
k, it's passing our CI.  Merging it now. :smile:

justinclift:
@Avey777 Done. :smile:

eradman:
> > In addition, in my understanding, if you want to hide components, it should be controlled on the backend rather than the frontend, right?
> 
> Just hiding some filtering components. The reason for hiding them on the front-end is that there are other places where these hidden components need to be displayed. This is currently a short-term implementation solution.

@Avey777 do you plan on improving the implementation of this feature?

1. The example help text is confusing for users since it's not clear were to add the parameters or how that will affect the share link
2. The share link itself does not contain the setting, it is not useful to the person creating the share link

Avey777:
There should be no time recently, it will be the Chinese New Year soon, I will see if there is time later

Avey777:
I'm sorry if this is confusing, but can you hide the example document first, what will it matter

eradman:
Having an incomplete feature is okay, it's really just the UI change that is a problem for us. Posted a suggested change to revert the example text in https://github.com/getredash/redash/pull/6709

PR #6116, opened 2023-06-25
ci: Attempt to fix restyled.io test

justinclift:
## What type of PR is this? 

- [x] Bug Fix

## Description

We're now getting an error with exit code 127 in our CI "restyled" test, so this commit adds the suggested solution of telling Restyled the exact command needed to run:

https://github.com/restyled-io/restyled.io/wiki/Common-Errors:-Prettier#prettier-with-tailwindcss-executable-file-not-found-in-path

## How is this tested?

It's about to be, by letting the CI run. :wink:

### DIFF
--- a/.restyled.yaml
+++ b/.restyled.yaml
@@ -57,6 +57,9 @@ restylers:
       - migrations/versions
   - name: prettier
     image: restyled/restyler-prettier:v1.19.1-2
+    command:
+      - prettier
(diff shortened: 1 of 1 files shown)


justinclift:
Looks like this does indeed fix the Restyled test.

Not sure what's wrong with the `build` test item.  That'll need to be investigated next.

PR #6123, opened 2023-06-27
feat: add query result expired ttl

gaecoli:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -362,9 +362,21 @@ class QueryResult(db.Model, QueryResultPersistence, BelongsToOrgMixin):
         query_hash = gen_query_hash(query)
 
         if max_age == -1:
-            query = cls.query.filter(
-                cls.query_hash == query_hash, cls.data_source == data_source
-            )
+            if settings.QUERY_RESULTS_EXPIRED_TTL_ENABLED:
+                query = cls.query.filter(
+                    cls.query_hash == query_hash,
+                    cls.data_source == data_source,
+                    (
+                        db.func.timezone("utc", cls.retrieved_at) +
+                        datetime.timedelta(
+                            seconds=settings.QUERY_RESULTS_EXPIRED_TTL
+                        ) >= db.func.timezone("utc", db.func.now())
+                    )
+                )
+            else:
+                query = cls.query.filter(
+                    cls.query_hash == query_hash, cls.data_source == data_source
+                )
         else:
             query = cls.query.filter(
                 cls.query_hash == query_hash,
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -361,22 +361,13 @@ class QueryResult(db.Model, QueryResultPersistence, BelongsToOrgMixin):
     def get_latest(cls, data_source, query, max_age=0):
         query_hash = gen_query_hash(query)
 
+        if max_age == -1 and settings.QUERY_RESULTS_EXPIRED_TTL_ENABLED:
+            max_age = settings.QUERY_RESULTS_EXPIRED_TTL
+
         if max_age == -1:
-            if settings.QUERY_RESULTS_EXPIRED_TTL_ENABLED:
-                query = cls.query.filter(
-                    cls.query_hash == query_hash,
-                    cls.data_source == data_source,
-                    (
-                        db.func.timezone("utc", cls.retrieved_at) +
-                        datetime.timedelta(
-                            seconds=settings.QUERY_RESULTS_EXPIRED_TTL
-                        ) >= db.func.timezone("utc", db.func.now())
-                    )
-                )
-            else:
-                query = cls.query.filter(
-                    cls.query_hash == query_hash, cls.data_source == data_source
-                )
+            query = cls.query.filter(
+                cls.query_hash == query_hash, cls.data_source == data_source
+            )
         else:
             query = cls.query.filter(
                 cls.query_hash == query_hash,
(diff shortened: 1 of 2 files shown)


gaecoli:
Add an expiration time ttl function when the query page or dashboard page is opened by default to get the results. I believe this function is very useful for daily changing data in large enterprises. At present, this function is optional, and the logic before redash is maintained by default. This new function needs to be activated through configuration.

gaecoli:
@justinclift Passed CI.

junnplus:
I would like to know the difference between the newly added "expired ttl" and the original "max age."

gaecoli:
> I would like to know the difference between the newly added "expired ttl" and the original "max age".

For example, if you create a query in redash, there will be a result set after querying once. If your query already has query results, then the default max_age is -1. Currently, there is no expiration time for the results of this query. So if you want the latest results, you have to manually trigger the refresh button to update the results of this query. The function I am adding now is that when a certain query has results, then add an expiration time, and open this query next time If the verification query result exceeds the timeout you configured, it will re-fetch the latest result. I don't know if I express clearly?

junnplus:
@gaecoli Thank you for answering my question.

gaecoli:
> @gaecoli Thank you for answering my question.

I wonder know how you update the result of a certain query now?

redash/models/__init__.py
```diff
@@ -362,9 +362,21 @@ def get_latest(cls, data_source, query, max_age=0):
...
+                query = cls.query.filter(
+                    cls.query_hash == query_hash,
+                    cls.data_source == data_source,
+                    (
+                        db.func.timezone("utc", cls.retrieved_at) +
+                        datetime.timedelta(
+                            seconds=settings.QUERY_RESULTS_EXPIRED_TTL
+                        ) >= db.func.timezone("utc", db.func.now())
+                    )
+                )
```
junnplus, line 375: Can we merge this part of the logic with "max_age != -1"? They are almost the same.

gaecoli: You can change it.

gaecoli: > Can we merge this part of the logic with "max_age != -1"? They are almost the same.

When max != -1, it is to actively refresh or the query does not get the corresponding result.

gaecoli: > Can we merge this part of the logic with "max_age != -1"? They are almost the same.

In redash, `max_age` in `0 or -1` two types.

gaecoli: @junnplus  You can review this code, and then i want to show your logic for my mind.

redash/models/__init__.py
```diff
@@ -362,9 +362,21 @@ def get_latest(cls, data_source, query, max_age=0):
         query_hash = gen_query_hash(query)
 
         if max_age == -1:
```
junnplus, line 364: It seems that the front end only triggers a query with max_age = -1 when there are no search results. In this case, if there are already search results, do we still need to manually refresh?

gaecoli: > It seems that the front end only triggers a query with max_age = -1 when there are no search results. In this case, if there are already search results, do we still need to manually refresh?

When there is a result, max_age=-1 will also be triggered, but for example, if I want to see tomorrow's data tomorrow instead of the data queried yesterday, what should I do? At present, it will also use max_age to find old query results

gaecoli: > @junnplus 我尝试用中文和你交流一下吧！现在 redash 如果某个查询已经有了查询结果了，那么它以后都会一直获取到这个查询结果，如果 sql 不改的话，现在遇到了问题，比如我的数据库的数据每天都会有新的数据进来，那么我过了一天想获取到新的查询结果，而不是之前旧的数据，现在的 redash 一直会 get 到旧的数据，所以想加一个自己配置一个结果的失效时间，那么我下次打开这个查询的话如果结果超过了我的配置 ttl 时间，那么它就会去主动去查询新的结果集。感谢你的时间和回复。我不知道表达清楚没有我的解释。

每次打开有结果的查询它都会用 `max_age == -1` 这个去获取结果, 当条件为 `max_age != -1` 的话就代表这个查询没有查询过结果或者是刷新操作。

junnplus: Thank you for your detailed explanation. I was wondering if we could achieve this by adding this configuration to the front-end instead of passing max_age = -1 every time.

gaecoli: > Thank you for your detailed explanation. I was wondering if we could achieve this by adding this configuration to the front-end instead of passing max_age = -1 every time.


Not thought of a better solution yet.

redash/models/__init__.py
junnplus: > Can we merge this part of the logic with "max_age != -1"? They are almost the same.

What I mean is that we can reset the value of max_age before `if max_age == -1:`.
```python
        if max_age == -1 and settings.QUERY_RESULTS_EXPIRED_TTL_ENABLED:
            max_age = settings.QUERY_RESULTS_EXPIRED_TTL

        if max_age:
            ...
```

gaecoli: > > Can we merge this part of the logic with "max_age != -1"? They are almost the same.
> 
> What I mean is that we can reset the value of max_age before `if max_age == -1:`.
> 
> ```python
>         if max_age == -1 and settings.QUERY_RESULTS_EXPIRED_TTL_ENABLED:
>             max_age = settings.QUERY_RESULTS_EXPIRED_TTL
> 
>         if max_age:
>             ...
> ```

Got it, we can change to this logic.

gaecoli:
I changed code style and logic. Honestly, your logic is much clearer. @junnplus

junnplus (approved):
LGTM, thanks!

gaecoli:
> LGTM, thanks!

Thank you!

PR #6134, opened 2023-06-29
Attempt to fix cypress race conditions

wlach:
This PR attempts to fix some Cypress race conditions revealed by #5997

(it includes #5997 for the purposes of testing)

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [x] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


justinclift:
Another potential workaround approach for the Cypress failure is (maybe) this: https://github.com/getredash/redash/commit/28e63e3d7661a4533f71e585296ff722adf9668d

I'm not sure *why* that one seemed to work, but it passed the Cypress tests 3 times in a row on a test that was previously failing consistently.  Might have been pure luck, but I'm not sure. :wink:

wlach:
Waiting for the buttons to be enabled fixes two of the tests. It appears that the reordering test is failing because we're using force, but taking it out I get the following error:

```
  1) Dashboard Parameters
       reorders parameters:
     CypressError: Timed out retrying: `cy.trigger()` failed because this element is not visible:

`<div class="drag-handle" data-test="DragHandle-param1"></div>`

This element `<div.drag-handle>` is not visible because it has CSS property: `position: fixed` and it's being covered by another element:

`<div class="sortable-container sortable-container-dragging parameter-container">...</div>`

Fix this problem, or use `{force: true}` to disable error checking.
```

Going to experiment with a short timeout which I think will simulate the latency we used to see.

wlach:
> Another potential workaround approach for the Cypress failure is (maybe) this: [28e63e3](https://github.com/getredash/redash/commit/28e63e3d7661a4533f71e585296ff722adf9668d)
> 
> I'm not sure _why_ that one seemed to work, but it passed the Cypress tests 3 times in a row on a test that was previously failing consistently. Might have been pure luck, but I'm not sure. 😉

Hmm that's quite strange, I would be surprised if that actually changed anything? Definitely relatively new to cypress myself.

justinclift:
Yeah, it didn't make sense to me either.  But it worked.  While it could have been a fluke, it might be worth looking at again just in case it wasn't. :smile:

justinclift:
As a thought, have you installed the Cypress (v5.3.0) GUI and loading the test into that?

If not, it's probably worth trying.  Can be useful for showing up non-obvious problems with a Cypress test. :smile:

wlach:
Giving up for now, I clearly need to understand more about what cypress is doing under the hood.

justinclift:
Ahhh.  Yeah, Cypress can be a real pita. :wink:

With this, the whole "load the GUI, then run the test in the GUI" thing is probably a good idea.

Oh... on the other hand, @gabrieldutra pointed out the other day that our tests are still saving video of each run to the Cypress Cloud.  Let me see if I can find the URL we need to use for accessing it... :smile:

justinclift:
Looking at the error output for the failures, it has this:

```
4 failing

  1) Parameter
       Dropdown Parameter
         supports multi-selection:
     AssertionError: Timed out retrying: expected '<div.ant-table>' to contain '"value1","value2","value3"'
      at Context.eval (http://server:5000/__cypress/tests?p=client/cypress/integration/query/parameter_spec.js:223:44)

  2) Parameter
       Query Based Dropdown Parameter
         based on a query with 3 results
           supports multi-selection:
     AssertionError: Timed out retrying: Expected to find element: `[data-test="TableVisualization"]`, but never found it.
      at Context.eval (http://server:5000/__cypress/tests?p=client/cypress/integration/query/parameter_spec.js:312:46)

  3) Parameter
       Date Parameter
         updates the results after selecting a date:
     AssertionError: Timed out retrying: expected '<div.parameter-apply-button>' to be 'enabled'
      at Context.eval (http://server:5000/__cypress/tests?p=client/cypress/integration/query/parameter_spec.js:350:46)

  4) Parameter
       Draggable
         is possible to rearrange parameters:

      AssertionError: expected '<label>' to have text 'Parameter 4', but the text was 'Parameter 3'
      + expected - actual

      -'Parameter 3'
      +'Parameter 4'
      
      at Context.eval (http://server:5000/__cypress/tests?p=client/cypress/integration/query/parameter_spec.js:569:91)
```

To me, those sound like the element(s) that Cypress is selecting in the tests now is **somehow** different to what is expected.  Loading the failing test in the Cypress GUI should help.

As to "How to load the test in the Cypress GUI"... I'm not yet sure of for Redash.  I'll work that out right now too (and document it), so we can all do it.

It's going to be something along the lines of:

* Use `yarn` to install the Cypress 5.3.0 GUI
* Ensure the test database has been created, and the Docker daemons are running
* Launch the Cypress GUI
* Load the Redash config into Cypress
* Load (and run) the failing group of tests

That's how another project I put time into does things with Cypress, so it'll be something along those lines.  The Cypress Cloud stuff could also be useful as well.

wlach:
It looks like cypress might actually be picking up on a real bug: https://github.com/getredash/redash/pull/5586 🤔

justinclift:
Interesting.  +1 for persistence :grin:

justinclift:
Just rebased this, though I'm not sure this PR is useful any more as the Cypress tests seem to be reasonably reliable over the last day and a bit. :smile:

justinclift:
Ironically, there's a different Cypress test ([embed/share_embed_spec.js](https://github.com/getredash/redash/actions/runs/5562756436/jobs/10161668077)) that's started failing fairly routinely a lot now. :man_shrugging:

PR #6141, opened 2023-07-02
Upgrade viz-lib webpack to v5

gabrieldutra:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

Upgrades webpack from viz-lib to v5, a requirement for using Node versions greater than v16.

### CLI Build

| Before/After | v16 | v19 |
| ---- | ----- | --- |
| Before |Screenshot 2023-07-02 at 12 08 48  |Screenshot 2023-07-02 at 11 25 45 |
| After | Screenshot 2023-07-02 at 12 06 10 | Screenshot 2023-07-02 at 11 57 40| 

### Follow-ups to consider

1. For future maintenance / simplify the infra it may be better to move `viz-lib` back into the main Package if we don't have explicit goals for it. Otherwise it's two setups to maintain and it makes things harder to configure (e.g.: React HMR)
2. For the main redash FE setup, I may try to move to [Next.js incremental adoption](https://nextjs.org/blog/incremental-adoption) instead of upgrading webpack. Reasoning would be, apart from fixing compatibility with newer Node versions, to have all the tooling set Next has by default (e.g.: devloop, code splitting) 

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Visualizations remained working:
Screenshot 2023-07-02 at 12 00 56

### DIFF
--- a/viz-lib/package.json
+++ b/viz-lib/package.json
@@ -26,8 +26,8 @@
   "author": "Redash",
   "license": "BSD-2-Clause",
   "peerDependencies": {
-    "antd": ">=4.0.0",
     "@ant-design/icons": ">=4.0.0",
+    "antd": ">=4.0.0",
     "react": ">=16.8.0",
     "react-dom": ">=16.8.0"
   },
@@ -58,7 +58,6 @@
     "enzyme": "^3.8.0",
     "enzyme-adapter-react-16": "^1.7.1",
     "enzyme-to-json": "^3.3.5",
-    "file-loader": "^6.0.0",
     "jest": "^24.1.0",
     "less": "^3.11.1",
     "less-loader": "^5.0.0",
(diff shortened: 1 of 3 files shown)


justinclift:
> For future maintenance / simplify the infra it may be better to move viz-lib back into the main Package if we don't have explicit goals for it.

Interesting.  Personally I don't know what the viz-lib is, apart from being able to see that it seems to be some piece of Redash that was split off.

It kind of sounds like it was made into a library so other people/projects could use it?  If that's the case... is it used much? :smile:

---

The other stuff you mention I have no personal clue about.  I guess we'll need others with front end skills to look at it.  Hopefully there are a few in the @getredash/maintainers group. :smile:

arikfr:
Based on npm stats it gets 3 weekly downloads.. so I guess it's not used much. For simplicity it might be a good idea to merge back in.

How much effort it might be?

arikfr (approved):

gabrieldutra:
> How much effort it might be?

I'm hoping way lower effort than making a good working Workspaces + HMR setup 😅

PR #6142, opened 2023-07-02
Wait for query change to propagate on pivot spec

gabrieldutra:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

Checked what's happening for the Pivot spec in [this run](https://cloud.cypress.io/projects/924cka/runs/7c62599b-ead0-473d-aadd-443eab109b91/test-results/18b0354a-312f-4ef0-9ce9-3ee6e21865e7/video):
1. Query text is updated to include a new line
2. Save button is triggered, if for some reason the debounce we do didn't finish by then text will be saved with older values
3. Same for execution
4. New query object is received from the backend, the query text is updated on the frontend. The new value will revert in this case
5. Query is asserted with the older results, so it fails

For reference, the delay we do is in https://github.com/getredash/redash/blob/84c2abed59bace2f2b9d1e31d512a60ba8c8ac47/client/app/pages/queries/QuerySource.jsx#L84-L86

In the future it's possible to do some refactor on the query page state for that not to be needed, but left the wait as a short term to fix the flaky test.

## How is this tested?

- [x] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/client/cypress/integration/visualizations/pivot_spec.js
+++ b/client/cypress/integration/visualizations/pivot_spec.js
@@ -99,6 +99,10 @@ describe("Pivot", () => {
         .focus()
         .type(" UNION ALL {enter}SELECT 'c' AS stage1, 'c5' AS stage2, 55 AS value");
 
+      // wait for the query text change to propagate (it's debounced in QuerySource.jsx)
+      // eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(200);
+
       cy.getByTestId("SaveButton").click();
       cy.getByTestId("ExecuteButton")
         .should("be.enabled")


gaecoli:
> `pivot_spec` passed, `share_embed_spec` also seems flaky though

any changes？

justinclift:
Cool, seems like this should help increase the reliability of our tests. :smile:

With the `share_embed_spec` one, is the general idea to investigate that in another PR, or should we add something more to this one for it, or ?

gaecoli:
> Cool, seems like this should help increase the reliability of our tests. 😄
> 
> With the `share_embed_spec` one, is the general idea to investigate that in another PR, or should we add something more to this one for it, or ?


I think it’s good for this branch to only focus on one thing, and the rest can be done in another branch.

client/cypress/integration/visualizations/pivot_spec.js
```diff
@@ -99,6 +99,10 @@ describe("Pivot", () => {
         .focus()
         .type(" UNION ALL {enter}SELECT 'c' AS stage1, 'c5' AS stage2, 55 AS value");
 
+      // wait for the query text change to propagate (it's debounced in QuerySource.jsx)
+      // eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(200);
```
gaecoli, line 104: For tasks that need to be executed, you may wish to increase the waiting time. My suggestion is `500 or 1000`.
`200` doesn't seem to do much. Just my opinion.

justinclift:
Sounds like a plan. :smile:

client/cypress/integration/visualizations/pivot_spec.js
justinclift: Oops, I merged it before seeing that.

I guess if we keep on seeing failures here then we can raise that limit. :smile:

gaecoli:
> Sounds like a plan. 😄

Yeah. 
I'll try to fix it if I'm free.

client/cypress/integration/visualizations/pivot_spec.js
gaecoli: > Oops, I merged it before seeing that.
> 
> I guess if we keep on seeing failures here then we can raise that limit. 😄

Yeah. It doesn't matter.

gabrieldutra: If the reason is really the debounce issue, 200 should be perfectly fine (2x the actual debounce time). It's usually a non-practice to do those arbitrary waiting times, it's just that for this case I had no clue what I could use in the UI to assure the state was updated 😕.

But let's see if it helps

gaecoli: > If the reason is really the debounce issue, 200 should be perfectly fine (2x the actual debounce time). It's usually a non-practice to do those arbitrary waiting times, it's just that for this case I had no clue what I could use in the UI to assure the state was updated 😕.
> 
> But let's see if it helps

Good job!

PR #6152, opened 2023-07-07
CI: add test_all_deps on docker build

junnplus:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other
- [x] CI

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -88,8 +88,7 @@ WORKDIR /app
 ENV PIP_DISABLE_PIP_VERSION_CHECK=1
 ENV PIP_NO_CACHE_DIR=1
 
-# rollback pip version to avoid legacy resolver problem
-RUN pip install pip==20.2.4;
+RUN pip install pip==23.1.2;
 
 # We first copy only the requirements file, to avoid rebuilding on every file change.
 COPY requirements_all_ds.txt ./
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,8 @@ EXPOSE 5000
 ARG skip_ds_deps
 # Controls whether to install dev dependencies.
 ARG skip_dev_deps
+# Controls whether to install all dependencies for testing.
+ARG test_all_deps
 
 RUN useradd --create-home redash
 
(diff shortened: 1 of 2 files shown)


justinclift:
I don't think this will do what we're after.

pip seems to check things a bit differently using `pip install -r requirements_all_ds.txt`, compared to `pip3 install -r requirements.txt -r requirements_dev.txt -r requirements_all_ds.txt`.

Having all three of the requirements*.txt files on the same line seems to make a (positive) difference, letting it give a more thorough answer.

So even if we remove the option here, then we're not the thorough test of all dependencies.

justinclift:
We could probably add an extra test to the Dockerfile, to add a specific test though:

```
RUN if [ "x$test_all_deps" != "x" ] ; then pip3 install -r requirements.txt -r requirements_dev.txt -r requirements_all_ds.txt ; fi
```

We'd need to add a matching `test_all_deps` ARG near the top of the file, and make sure we include this argument when running the test too.

justinclift:
Ugh.  I keep on forgetting to add those comments using GitHub's "Review" process. :wink:

justinclift (changes requested):
Pretty sure we'll need an `ARG` line added to the Dockerfile as well, so this new `test_all_deps` variable can be used.  Something like the lines of these ones:

* https://github.com/getredash/redash/blob/79b01406fc30925661a887d04f4176c56d491f6b/Dockerfile#L32-L35

Reference info, in case it's useful: :smile:

* https://docs.docker.com/engine/reference/builder/#arg

junnplus:
> Pretty sure we'll need an ARG line added to the Dockerfile as well, so this new test_all_deps variable can be used. Something the lines of these ones:

yeah, I missed it.

Dockerfile
```diff
@@ -86,8 +88,7 @@ WORKDIR /app
 ENV PIP_DISABLE_PIP_VERSION_CHECK=1
 ENV PIP_NO_CACHE_DIR=1
 
-# rollback pip version to avoid legacy resolver problem
-RUN pip install pip==20.2.4;
+RUN pip install pip==23.1.2;
```
justinclift, line 91: Is the updated pip version here important?

junnplus: Older versions of pip will silently suppress conflict messages.

see https://github.com/getredash/redash/actions/runs/5491313994/jobs/10007730439?pr=5865#step:3:1473

justinclift: Interesting.  So that's what the comment above is about. :smile:

justinclift (approved):
Looks good to me. :smile:

PR #6162, opened 2023-07-10
Use automatic upgrading PostgreSQL 15 docker image

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR moves us to PostgreSQL 15.3, using a Docker image that safely upgrades existing PostgreSQL data files versions before starting.

With this move to modern PostgreSQL, we will no longer need to keep stuck on ancient versions of some Python libraries in our backend.

That in turn will let us resolve quite a few of the outstanding security alerts in our current dependencies.

## How is this tested?

- [z] E2E Tests (Cypress)
- [z] Manually


justinclift:
Hmmm, the error being shown looks like there's further work to be done for the PG docker container:

> cat: can't open '/var/lib/postgresql/data/PG_VERSION':

That file being missing means the PostgreSQL `initdb` didn't run for some reason.  Lets try rerunning this again, just to see if it's consistent.

justinclift:
Ahhh.  It's probably [this](https://github.com/justinclift/docker-pgautoupgrade/blob/1d2e64535f5426eaf1cad435eec8de17f19feaa3/docker-entrypoint.sh#L347) line.  Likely needs to have something check for the presence of that file first.

Trying that now...

justinclift:
What the... ?

```
> Run docker-compose -p redash run --rm postgres psql -h postgres -U postgres -c "create database tests;"
  
Creating redash_postgres_run ... 
Creating redash_postgres_run ... done
psql: error: could not translate host name "postgres" to address: Try again
```

That seems like an internal error from docker.

It's using `postgres` as the name of the PostgreSQL container (nothing wrong there), but somehow that's failing the name lookup.  Seems pretty weird. :frowning:

justinclift:
Well, it's consistently failing in that same spot now, so something is going on.  I'll look at it later.

justinclift:
The commit history in this PR is likely to become pretty messy for a bit, as I do some experimentation.  Don't get too worried about it. :smile:

justinclift:
Interesting.  The `backend-unit-tests` are failing with the standard modern PostgreSQL 15 image as well.  So, it's not something I've broken in the "automatic updating" one, which is a relief of one sort.

Next step, lets try bisecting things and see what version of PG it starts breaking on.

justinclift:
Hmmm, it might also be the CI docker-compose doesn't have:

```
    environment:
      POSTGRES_HOST_AUTH_METHOD: "trust"
```

Not sure why that would be needed for it to start in the first place, but it's possible.

justinclift:
Cool, looks like the missing `POSTGRES_HOST_AUTH_METHOD` environment variable in the CI docker compose was the underlying problem.  I'll recreate the commits in this PR shortly. :smile:

PR #6167, opened 2023-07-11
format code by black and isort

junnplus:
## What type of PR is this? 

Formating all files, and add black check in CI lint.

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


junnplus:
> Run black --check .
All done! ✨ 🍰 ✨
230 files would be left unchanged.

justinclift:
Oh, this seems like a sensible idea. :smile:

setup.cfg
```diff
@@ -2,6 +2,9 @@
...
 [flake8]
-ignore = E501
-exclude = .git
-max-complexity = 10
+ignore = E501, W503
+extend-ignore = E203
+exclude = .git,viz-lib,node_modules,migrations
+max-complexity = 15
+per-file-ignores =
+    __init__.py: F401
```
junnplus, line 10: updated flake8 config.

redash/authentication/__init__.py
```diff
@@ -229,7 +223,7 @@ def redirect_to_login():
 def logout_and_redirect_to_index():
     logout_user()
 
-    if settings.MULTI_ORG and current_org == None:
+    if settings.MULTI_ORG and current_org is None:
```
gaecoli, line 226: I don't know if this needs to be changed. Maybe we can keep `==`.

gaecoli (approved):
LGTM!

redash/authentication/__init__.py
junnplus: OK, let's keep it as it is

bin/release_manager.py
```diff
@@ -39,84 +44,96 @@ def get_rc_release(version):
 
     raise exception_from_error("Unknown error while looking RC release: ", response)
 
+
 def create_release(version, commit_sha):
     tag = rc_tag_name(version)
 
-    params = {
```
gaecoli, line 45: I think this style is better than your changed style.

redash/authentication/__init__.py
gaecoli: > OK, let's keep it as it is

OK.

bin/release_manager.py
gaecoli: For readability, I think it`s better not to expand the code here.

junnplus: I might need to look into the configuration of black.

guidopetri:
I'm not the biggest fan of `black`, I think it makes code pretty unreadable.

That being said, we should add this to a precommit file as well, to make it easy to keep the formatting inplace.

junnplus:
@gaecoli Do you mind helping to review this PR again? There are quite a few changes in the code, so it might take some time.

junnplus:
> That being said, we should add this to a precommit file as well, to make it easy to keep the formatting inplace.

Yes, this is my intention.

gaecoli:
> @gaecoli Do you mind helping to review this PR again? There are quite a few changes in the code, so it might take some time.

I think it`s ok. I see that the code logic has not changed and the code style has been refined.
LGTM! Good job!

gaecoli:
Look forward to your wiki. For ours how to use your tools to format code style.

Issue #5646, opened 2021-11-11
lastSelectedDataSourceId always reverts to alphabetically first data source


deed02392:
### Issue Summary

Despite the localstorage get `lastSelectedDataSourceId` visibly updating when selecting a different data source, it is subsequently reverted to the ID of the first data source in alphabetical order.

### Steps to Reproduce

1. Create a new query
2. Change to an alternative data source that is NOT alphabetically first
3. Observe `lastSelectedDataSourceId` updates in localstorage
4. Refresh or reload the new query page
5. Observe that `lastSelectedDataSourceId` reverts to the ID of the first alphabetically listed data source

### Technical details:

* Redash Version: 10.0.0-beta (92e5d78d)
* Browser/OS: Chrome macOS
* How did you install Redash: Docker

sjwmoveon:
This issue is also present in the 10.0.0 release b50363 in Firefox on macOS, with a Docker installation.

ProkhorovAlexander:
experiencing exactly the same issue on chrome macos with docker installation

PR #6181, opened 2023-07-15
Fix persisting datasource selection choice

wlach:
Closes getredash/redash#5646.

We were already storing this in localStorage, however we weren't accounting for the fact that it was stored as a string, rather than an integer.

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

We were already storing this in localStorage, however we weren't accounting for the fact that it was stored as a string, rather than an integer.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Ran redash locally, verified this fixes the issue.

### DIFF
--- a/client/app/pages/queries/QuerySource.jsx
+++ b/client/app/pages/queries/QuerySource.jsx
@@ -137,10 +137,7 @@ function QuerySource(props) {
       const selectedDataSourceId = parseInt(localStorage.getItem("lastSelectedDataSourceId")) || null;
 
       handleDataSourceChange(
-        chooseDataSourceId(
-          [query.data_source_id, selectedDataSourceId, firstDataSourceId],
-          dataSources
-        )
+        chooseDataSourceId([query.data_source_id, selectedDataSourceId, firstDataSourceId], dataSources)
       );
     }
   }, [query.data_source_id, queryFlags.isNew, dataSourcesLoaded, dataSources, handleDataSourceChange]);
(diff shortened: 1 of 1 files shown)


guidopetri:
LGTM though I'm not a JS expert. Should we merge in the restyled changes?

justinclift:
Yep, we should merge in the restyled change. :smile:

justinclift:
Just pressed the button on that restyled change, so lets see how it all goes now. :smile:

guidopetri (approved):

justinclift:
Interesting.  Looks like it's failed the same Cypress test twice.  That could be just bad luck on a flaky test, but it might also mean there's something that needs a closer look.

Lets run the CI tests a third time, and see what happens...

wlach:
> Interesting. Looks like it's failed the same Cypress test twice. That could be just bad luck on a flaky test, but it might also mean there's something that needs a closer look.
> 
> Lets run the CI tests a third time, and see what happens...

Pretty skeptical this PR has any issues, it definitely passed on redashcommunity when I originally filed it.

guidopetri:
Flaky test indeed... merging while we can lol

justinclift:
Awesome. :smile:

We might need to look into that flaky test if it keeps on causing issues.

Issue #6201, opened 2023-07-20
SAML Settings :: Not Using SAML Entity ID


DSigmund:
### Issue Summary

The Application does not respect the Setting "SAML Entity ID" in the SAML Settings Page, leading to mismatches on the Login.

Instead of the Configured Value, the application sends the ASC-URL as Entity.

e.g.

- Configured: **redash**
- Sent: **https://<mydomain>/saml/callback?org=default**

This leads to an audience Error if using Microsoft ADFS for SAML:

> Exception: AudienceRestrictions conditions not satisfied! (Local entity_id=https://r<mydomain>/saml/callback?org_slug=default)

As the IDP returns this value:

```
<AudienceRestriction>
  <Audience>https://<mydomain>/saml/callback</Audience>
</AudienceRestriction>
```

See:
`saml_settings["entityid"] = acs_url`

https://github.com/getredash/redash/blob/master/redash/authentication/saml_auth.py#L94

Solution:
`saml_settings["entityid"] = entity_id`

### Technical details:

* Redash Version:  11.0.0-dev (bc909a13)
* Browser/OS: Chrome
* How did you install Redash: Docker Compose via podman (image: redash/redash:preview)

konnectr:
If you understand what to fix, can you create a PR? New instructions on how to deploy locally https://github.com/getredash/redash/wiki/Local-development-setup

justinclift:
Thanks for pointing this out @DSigmund. :smile:

As @konnectr suggested above, we now have well used instructions for setting up a local development environment.  So it should (in theory! :wink:) be reasonably straight forward for you to try that change out and see if it's the complete solution or if other tweaks are needed.

Are you ok to give that a go?

DSigmund:
Of Course, PR is created :-)

justinclift:
Thanks.  Hopefully someone among the new maintainers knows SAML well enough to review it. :smile:

eradman:
#6202 merged some time ago. Closing

PR #6202, opened 2023-07-20
Changed asc_url to entity_id

DSigmund:
Respect the User Setting instead of a hardcoded  value

## What type of PR is this? 

- [x] Bug Fix

## Description
See https://github.com/getredash/redash/issues/6201

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually (Login via ADFS against Local IDP)
- [ ] N/A

https://github.com/getredash/redash/issues/6201


justinclift:
@getredash/maintainers Is anyone around who understands SAML and has time to review this Python PR? :smile:

guidopetri (approved):
I'm comfortable merging this in if @DSigmund you could confirm that you've deployed this and it works as expected.

I don't think any of the current maintainers have enough SAML knowledge to truly review this PR.

DSigmund:
It works as expected :-)

Many Thanks for merging it!

myonlylonely:
@guidopetri @DSigmund
It seems this commit breaks our existing SAML config, which works on other system as well (like Sentry).
I can no longer login by SAML after this commit.
What exactly does this commit do? What should I change?

DSigmund:
Hi,

It fixes the Setting where you have to enter a SAML-URL.
You are now able and must enter the correct URL as written in the
instructions.

seiyeah78:
I probably encountered this bug. 
After making changes for the upgrade from V8 to V10, I set the Azure AD's entity id to REDASH_SAML_ENTITY_ID, but I ran into an error and couldn't log in successfully. 
It worked when I made the Azure AD's entity id exactly the same as the URL for the callback (Assertion Consumer Service URL). 
I realized that part of the request was becoming the URL for the callback. 
This fix will likely revert it back to normal. 

Thank you for addressing this.

PR #6205, opened 2023-07-20
Bump Cypress to 6.0.1

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This moves us to using Cypress 6.0.1.

Unfortunately, due to a change in how Cypress treats elements with opacity 0, we need to disable two of the front end tests.

Hopefully someday, someone with better knowledge of JS and Cypress is able to fix it. :smile:

## How is this tested?

- [x] E2E Tests (Cypress)


justinclift:
Well that sucks.  It seems like yarn won't allow for the tests to be commented out, and they'll instead need to be removed.  Dammit.

PR #6271, opened 2023-07-24
Allow RSA key used for JWT to be specified as a file path

eradman:
- `auth_jwt_auth_public_certs_url` may file:// in addition to http/https
- Log an error if payload does not contain an email address

## What type of PR is this? 
- [x] Refactor
- [x] Feature

## How is this tested?

## Background

This feature allows another application to craft a JWT token to automatically log into redash.

Formerly this was possible, but the RSA public key had to be on an HTTP server. A file path is less complex and easier to secure.

### DIFF
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -187,6 +187,10 @@ def jwt_token_load_user_from_request(request):
     if not payload:
         return
 
+    if "email" not in payload:
+        logger.info("No email field in token, refusing to login")
+        return
+
     try:
         user = models.User.get_by_email_and_org(payload["email"], org)
     except models.NoResultFound:
--- a/redash/authentication/jwt_auth.py
+++ b/redash/authentication/jwt_auth.py
@@ -6,6 +6,34 @@ import simplejson
 
 logger = logging.getLogger("jwt_auth")
 
+FILE_SCHEME_PREFIX = "file://"
+
+
+def get_public_key_from_file(url):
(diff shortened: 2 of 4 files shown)


justinclift:
The concept of the PR sounds useful.

Any idea how feasible it would be to get the Codecov test happy? :smile:

eradman:
I'll see if I can contrive a test for http as well. This code does not have existing tests, but Codecov is complaining because I moved some of the logic around.

justinclift:
Cool, sounds like a plan. :smile:

eradman:
I didn't want to introduce another pip dev dependency, but JWK is a is difficult to craft by hand.

eradman:
@justinclift I'd say this change is ready to go

justinclift:
> I didn't want to introduce another pip dev dependency, but JWK is difficult to craft by hand.

Sounds like a sensible compromise. :smile:

justinclift:
@wlach @gaecoli @guidopetri Anyone interested in reviewing this PR? :smile:

gaecoli:
I have been using https, I hope to enable file and https/http options, otherwise there is no way to use it For old users.

gaecoli:
> > I have been using https, I hope to enable file and https/http options, otherwise there is no way to use it For old users.
> 
> This PR should not change existing behavior, only add the ability to specify a private key from a file path

Ok, LGTM!

justinclift:
Awesome.  Just merged this, and the PR on the website repo which documents it a bit. :smile:

PR #6296, opened 2023-07-26
Bump babel-plugin-module-resolver to 5.0.0 in viz-lib

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR bumps babel-plugin-module-resolver to 5.0.0 in viz-lib.

## How is this tested?

- [x] E2E Tests (Cypress)

### DIFF
--- a/viz-lib/package.json
+++ b/viz-lib/package.json
@@ -52,7 +52,7 @@
     "@types/tinycolor2": "^1.4.2",
     "babel-loader": "^8.1.0",
     "babel-plugin-istanbul": "^6.1.1",
-    "babel-plugin-module-resolver": "^4.0.0",
+    "babel-plugin-module-resolver": "^5.0.0",
     "css-loader": "^3.5.2",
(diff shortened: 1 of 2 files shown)


justinclift:
Oh, now that's an interesting failure:

```
error babel-plugin-module-resolver@5.0.0: The engine "node" is incompatible with this module. Expected version ">= 16". Got "14.17.6"
```

It passes the lint tests on my local system without reporting this. :frowning:

justinclift:
Looking at our CI testing images, they all seem to be NodeJS 16 based.

The only thing I can think of at the moment which is still NodeJS 14 based is Cypress 7.7.0, which we're using presently.  Maybe it's that?

justinclift:
Ahhh.  Netlify also has a config file which needs the NodeJS version update in:

https://github.com/getredash/redash/blob/fbe2f4d80895ea3a450dc3193c5fa45e81700812/netlify.toml#L9

PR #6298, opened 2023-07-26
Update Netlify to NodeJS 16.20.1 and Yarn 1.22.19

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR updates our Netlify builds to use NodeJS 16.20.1, and Yarn 1.22.19

## How is this tested?

- [x] N/A

This PR needs to be tested by running the CI on GitHub.


justinclift:
What the hell?  Those are the correct version numbers, it makes no sense for this to fail. :frowning_face:

justinclift:
Ok, this PR is going to become pretty messy as I do some experimentation...

justinclift:
Looks like Netlify already provides yarn 1.22.19, so we don't need to force install it any more.

PR #6303, opened 2023-07-27
BigQueryGCE: return project id as not bytes but str

amylase:
## What type of PR is this? 

- [x] Bug Fix

## Description

`big_query_gce.py`'s `_get_project_id()` returns `bytes` type. However, this value is expected by Google's client library to be `str`. The library internally converts the value to `str` value `b'your_project_id'`, which is not a valid project id at all.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

My employer company runs a Redash instance. I applied this fix there and verified that BigQueryGCE data source works correctly.


justinclift:
Thanks @amylase. Lets see if it passes our CI tests. :smile:

guidopetri (approved):
LGTM. Thanks for the contribution amylase :)

PR #6309, opened 2023-07-27
Bump Cypress to 9.7.0

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR bumps Cypress to 9.7.0, to remove some further blocks to Dependabot fixes.

The Cypress GUI in this version seems *extremely* flaky in my testing too, unlike the GUI in version 7.7.0.  And unlike the GUI in version 12.3.0+, which I've used on other projects.

## How is this tested?

- [x] E2E Tests (Cypress)

Tested on my local system using a slightly different master branch commits, but they should work the same.


justinclift:
@getredash/maintainers We're on Cypress 9.7.0 now. :smile:

PR #6311, opened 2023-07-28
Bump Cypress to 10.11.0

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This bumps Cypress to 10.11.0.  The Cypress 10.x GUI seems to be a substantial re-write / improvement over the older versions, though it's still not as reliable as running the tests from the command line.

## How is this tested?

- [x] E2E Tests (Cypress)


justinclift:
@getredash/maintainers We're now on the Cypress 10.x series.  This one has a more modern looking GUI which is a bit easier to work with in *some* ways.

They removed the capability to "run all the tests" (in one go) from the GUI though, which was kind of a pain in the arse.  They re-enabled that "run all the tests" setting in the (current) 12.x series, so I'll probably try to update us through to that, depending on the effort required. :wink:

justinclift:
Oh the "run all the tests" thing works fine from the command line still, so their change in the GUI won't have any impact on our CI.

PR #6312, opened 2023-07-28
Bump Cypress to 11.2.0

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR bumps Cypress to 11.2.0.  This is the most recent version of Cypress we can move to without needing to update many tests due to `cy.server()`, and `cy.route()`, [being been removed in Cypress 12.x](https://docs.cypress.io/guides/references/migration-guide#Command--Cypress-API-Changes).

## How is this tested?

- [x] E2E Tests (Cypress)

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "babel-plugin-transform-builtin-extend": "^1.1.2",
     "copy-webpack-plugin": "^6.4.1",
     "css-loader": "^1.0.1",
-    "cypress": "^10.11.0",
+    "cypress": "^11.2.0",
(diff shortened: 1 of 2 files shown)


justinclift:
@getredash/maintainers We're now on Cypress 11.2.0, which is probably recent enough for now.  It was only released 8 months ago.

Rather than put time into upgrading us to Cypress 12 (not needed for the security side of things), it's probably best I look into upgrading `jest`.  That one's still holding us back in several Dependabot alerts.

PR #6317, opened 2023-07-29
WIP: Migrate from Moment.js to Day.js

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR is a very initial conversion of Moment.js to Day.js.

Moment.js has been long deprecated and we need to migrate away from it in order to upgrade Jest.  Which we need to do to fix some Dependabot alerts. :man_facepalming:

This PR isn't even close to passing our CI tests yet.  Both the backend and Cypress tests fail all over the place. :wink:

It'll probably take me a few days to go through this and get it in shape.  I'm just dumping it here in case anyone has ideas in the meantime for fixing any of the tests that don't currently pass.

## How is this tested?

- [x] Unit tests (pytest, jest) - Fails
- [x] E2E Tests (Cypress) - Fails badly

### DIFF
--- a/client/app/components/DateInput.jsx
+++ b/client/app/components/DateInput.jsx
@@ -2,7 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import DatePicker from "antd/lib/date-picker";
 import { clientConfig } from "@/services/auth";
-import { Moment } from "@/components/proptypes";
+import { Dayjs } from "@/components/proptypes";
 
 const DateInput = React.forwardRef(({ defaultValue, value, onSelect, className, ...props }, ref) => {
   const format = clientConfig.dateFormat || "YYYY-MM-DD";
@@ -27,8 +27,8 @@ const DateInput = React.forwardRef(({ defaultValue, value, onSelect, className, 
 });
 
 DateInput.propTypes = {
-  defaultValue: Moment,
-  value: Moment,
+  defaultValue: Dayjs,
+  value: Dayjs,
   onSelect: PropTypes.func,
   className: PropTypes.string,
 };
--- a/client/app/components/DateRangeInput.jsx
+++ b/client/app/components/DateRangeInput.jsx
@@ -3,7 +3,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import DatePicker from "antd/lib/date-picker";
 import { clientConfig } from "@/services/auth";
-import { Moment } from "@/components/proptypes";
+import { Dayjs } from "@/components/proptypes";
 
 const { RangePicker } = DatePicker;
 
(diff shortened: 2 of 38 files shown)


justinclift:
That's weird.  This is failing the backend flake8 tests:

```
./redash/query_runner/__init__.py:281:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
./redash/query_runner/python.py:216:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
./redash/query_runner/python.py:244:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
```
... but this PR doesn't touch any of those files.

justinclift:
Making progress.  Noticeably less tests failing now.  Still more fixing to go however. :smile:

justinclift:
Looking at the E2E Cypress failures, there's only three failing tests now (`query/parameter_spec.js`).  They're all in the date picking component.

Digging into the problem (separately, on my local PC), it's showing what looks to be the date picker (`rc-picker`) still trying to use Moment.js for some stuff.

The version of `rc-picker` we're presently using is `1.10.7`, pulled in as a dependency of `antd`.  Looking at the package.json for that specific release, it has day.js in its dependency list as well as moment.js.  Note that the tag for `1.10.7` is on the exact same commit in their git repo as the tag for `1.12.1`.  No idea if that was a mistake or not.

https://github.com/react-component/picker/blob/ead450fd124d20d17c9c233e001409d0d37d9a2b/package.json#L80-L83

The (5.x series) [antd docs also mention it can use either one](https://ant.design/docs/react/use-custom-date-library).  Note that antd [switches its default to day.js in the 5.0.0 release](https://github.com/ant-design/ant-design/blob/HEAD/CHANGELOG.en-US.md#500), which I guess means moment.js is the default in the 4.x series.

justinclift:
Nicely, there's an antd guide for replacing Moment.js with Day.js:

https://4x.ant.design/docs/react/replace-moment

The code in the newest commit has the `DatePicker` approach from there.  I don't really understand their `TimePicker` one though, but I'll try it out in a bit and see how it goes.

justinclift:
Wow.  It passed the Cypress tests without failure this time.  Am kind of shocked. :smile:

Still, it clearly needs some more fixing in order to pass the frontend unit tests, so that'll need doing soon.

justinclift:
It turns out there's an NPM package which takes care of implementing Dayjs for antd:

https://www.npmjs.com/package/antd-dayjs-webpack-plugin

The frontend unit tests still need work.  Hopefully not a huge task.

eradman:
I think we should actually stick with moment.js. This project is not getting new feature updates, but appears to be maintained.

Jest is a generic testing framework, so it should not have dependencies that are incompatible with Moment. Let's work out how this conflict is appearing.

In some places the use of moment.js in the tests is generating warnings, and I believe this caused by ambiguous input strings https://momentjs.com/docs/#/parsing/string/

gabrieldutra:
> I think we should actually stick with moment.js. This project is not getting new feature updates, but appears to be maintained.

I tend to go with this :). For the future I think date formatting will make more sense through Basic JS + the Intl library we may use for string formatting

justinclift:
> I think we should actually stick with moment.js.

k.  We can go with that.  You both know this front end stuff much better than me. :smile:

PR #6321, opened 2023-07-30
fix flake8

konnectr:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.8'
-      - run: sudo pip install flake8 black isort
+      - run: sudo pip install flake8==6.1.0 black isort
       - run: flake8 .
       - run: black --check .
       - run: isort --check-only --diff .
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.8'
-      - run: sudo pip install flake8==6.1.0 black isort
+      - run: sudo pip install flake8==6.1.0 black==23.1.0 isort==5.12.0
       - run: flake8 .
       - run: black --check .
       - run: isort --check-only --diff .
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,3 +8,8 @@ repos:
     hooks:
       - id: black
         language_version: python3
+  - repo: https://github.com/pycqa/flake8
+    rev: 6.1.0
+    hooks:
+      - id: flake8
+
--- a/redash/query_runner/__init__.py
+++ b/redash/query_runner/__init__.py
@@ -278,7 +278,7 @@ class BaseSQLQueryRunner(BaseQueryRunner):
 
     def _get_tables_stats(self, tables_dict):
         for t in tables_dict.keys():
-            if type(tables_dict[t]) == dict:
+            if isinstance(tables_dict[t], dict):
                 res = self._run_query_internal("select count(*) as cnt from %s" % t)
                 tables_dict[t]["size"] = res[0]["cnt"]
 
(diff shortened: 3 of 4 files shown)


zhongjiajie (approved):
I think we should pin flake8 in our ci to keep it stable

justinclift:
We have a "pre-commit" hook which does format things automatically before creating the commit... but only if the developer has set it up.

konnectr:
> We have a "pre-commit" hook which does format things automatically before creating the commit... but only if the developer has set it up.

flake8 is not in pre-commit

konnectr:
I think he should be here.
https://github.com/getredash/redash/blob/master/.pre-commit-config.yaml

justinclift:
Oh.  Just checked and you're right.  My misunderstanding, sorry. :angel:

justinclift:
Well, put it in a PR and we can try it out. :smile:

zhongjiajie:
> I think he should be here.
> https://github.com/getredash/redash/blob/master/.pre-commit-config.yaml

One more thing we have to know, we should keep package version same as ci and pre-commit but we do not, it will install the latest in our ci https://github.com/getredash/redash/blob/fbe2f4d80895ea3a450dc3193c5fa45e81700812/.github/workflows/ci.yml#L19-L22  and have pin version in pre-commit https://github.com/getredash/redash/blob/ae29eb3dfbb5278d88f05389e68ede7f8a9c360a/.pre-commit-config.yaml#L1-L10

zhongjiajie:
I find out we use many thing to format out code, we have pre-commit, and makefile wrapper the pre-commit, and we use separate lint check in ci, IMO we should keep them consistency to avoid our ci crash

konnectr:
> > I think he should be here.
> > https://github.com/getredash/redash/blob/master/.pre-commit-config.yaml
> 
> One more thing we have to know, we should keep package version same as ci and pre-commit but we do not, it will install the latest in our ci
> 
> https://github.com/getredash/redash/blob/fbe2f4d80895ea3a450dc3193c5fa45e81700812/.github/workflows/ci.yml#L19-L22
> 
> and have pin version in pre-commit
> https://github.com/getredash/redash/blob/ae29eb3dfbb5278d88f05389e68ede7f8a9c360a/.pre-commit-config.yaml#L1-L10

You're right, I added the CI version.

konnectr:
> I find out we use many thing to format out code, we have pre-commit, and makefile wrapper the pre-commit, and we use separate lint check in ci, IMO we should keep them consistency to avoid our ci crash

The goal now is to release a new version, I think after that you can do refactoring.

zhongjiajie:
> > I find out we use many thing to format out code, we have pre-commit, and makefile wrapper the pre-commit, and we use separate lint check in ci, IMO we should keep them consistency to avoid our ci crash
> 
> The goal now is to release a new version, I think after that you can do refactoring.

I find out we do not release our version to pypi, do we have a plan to release to it now?

zhongjiajie:
BTW, is there any things I can help for releasing? like bugfix or something like that

konnectr:
> BTW, is there any things I can help for releasing? like bugfix or something like that

We communicate in discord https://discord.gg/CmRZB9Rz You can ask there how you can help. But now we are updating the frotend dependencies

justinclift:
@zhongjiajie There is a lot of stuff to get done, so we can definitely use more assistance. :smile:

Do you have skill with front end stuff like React, as well as Python? :smile:

Asking about React because I'm trying to fix some front end stuff, but my skill level with it is not high. :roll_eyes:

zhongjiajie:
> @zhongjiajie There is a lot of stuff to get done, so we can definitely use more assistance. 😄
> 
> Do you have skill with front end stuff like React, as well as Python? 😄
> 
> Asking about React because I'm trying to fix some front end stuff, but my skill level with it is not high. 🙄
> 
> --
> 
> If you are more familiar with Python instead, then please help look through all of the old, open PRs. We need to determine which ones are still useful, which ones need fixing, which ones should be closed (etc). Any of that would be super helpful. 😄

I am not familiar with front end develop, ok I will see some pr for python side

justinclift:
Awesome. :smile:

PR #6324, opened 2023-07-31
Fix webpack error in viz-lib

justinclift:
## What type of PR is this? 

- [x] Bug Fix

## Description

The config file format for webpack has changed slightly, so webpack was throwing this fatal error:

```
$ yarn watch
...
ERROR in ./src/components/ColorPicker/index.less
Module build failed (from ./node_modules/less-loader/dist/cjs.js):
ValidationError: Invalid options object. Less Loader has been initialized using an options object that does
not match the API schema.
```

This commit updates the config file format so webpack works again.

## How is this tested?

- [x] E2E Tests (Cypress)

### DIFF
--- a/viz-lib/webpack.config.js
+++ b/viz-lib/webpack.config.js
@@ -39,8 +39,10 @@ module.exports = {
           {
             loader: "less-loader",
             options: {
-              plugins: [new LessPluginAutoPrefix({ browsers: ["last 3 versions"] })],
-              javascriptEnabled: true,
+              lessOptions: {
+                plugins: [new LessPluginAutoPrefix({ browsers: ["last 3 versions"] })],
+                javascriptEnabled: true,
+	      },
             },
           },
         ],


justinclift:
Hmmm, I'm just going to forcibly merge this as it fixes a breakage in the viz-lib (only) build.

We might be able to optimise / improve it somewhat, but we can do that in a future PR. :smile:

viz-lib/webpack.config.js
```diff
@@ -39,8 +39,10 @@ module.exports = {
           {
             loader: "less-loader",
             options: {
-              plugins: [new LessPluginAutoPrefix({ browsers: ["last 3 versions"] })],
-              javascriptEnabled: true,
+              lessOptions: {
+                plugins: [new LessPluginAutoPrefix({ browsers: ["last 3 versions"] })],
+                javascriptEnabled: true,
+	      },
```
justinclift, line 45: Ugh indendation muck-up.  Oh well. :frowning:

PR #6332, opened 2023-08-01
Do not makedsn for query_runner/oracle.py

snickerjp:
## What type of PR is this? 

- [x] Feature

## Description

change for redash/query_runner/oracle.py
Do not makedsn for **Autonomous Database** or **FAILOVER,LOAD_BALANCE**

When type "_useservicename" in the `host`, self.configuration["servicename"] is used instead.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

[Connecting to Oracle Database — cx\_Oracle 8\.3\.0 documentation](https://cx-oracle.readthedocs.io/en/latest/user_guide/connection_handling.html#oracle-net-connect-descriptor-strings)

donotmakedsn02

### DIFF
--- a/redash/query_runner/oracle.py
+++ b/redash/query_runner/oracle.py
@@ -138,8 +138,8 @@ class Oracle(BaseSQLQueryRunner):
         if self.configuration.get("encoding"):
             os.environ["NLS_LANG"] = self.configuration["encoding"]
 
-        # When type "_donotmakedsn" in the host self.configuration["servicename"] is used instead
-        if self.configuration["host"].lower() == "_donotmakedsn":
+        # To use a DSN Service Name instead, use the text string `_useservicename` in the host name field.
+        if self.configuration["host"].lower() == "_useservicename":
             dsn = self.configuration["servicename"]
         else:
             dsn = cx_Oracle.makedsn(
--- a/redash/query_runner/oracle.py
+++ b/redash/query_runner/oracle.py
@@ -66,7 +66,7 @@ class Oracle(BaseSQLQueryRunner):
             "properties": {
                 "user": {"type": "string"},
                 "password": {"type": "string"},
-                "host": {"type": "string", "title": "Host (When type _donotmakedsn, DSN Service Name is used instead.)"},
+                "host": {"type": "string", "title": "Host (To use a DSN Service Name instead, use the text string `_useservicename` in the host name field.)"},
                 "port": {"type": "number"},
                 "servicename": {"type": "string", "title": "DSN Service Name"},
                 "encoding": {"type": "string"},
--- a/redash/query_runner/oracle.py
+++ b/redash/query_runner/oracle.py
@@ -66,7 +66,10 @@ class Oracle(BaseSQLQueryRunner):
             "properties": {
                 "user": {"type": "string"},
                 "password": {"type": "string"},
-                "host": {"type": "string", "title": "Host: To use a DSN Service Name instead, use the text string `_useservicename` in the host name field."},
+                "host": {
+                    "type": "string",
+                    "title": "Host: To use a DSN Service Name instead, use the text string `_useservicename` in the host name field.",
+                },
                 "port": {"type": "number"},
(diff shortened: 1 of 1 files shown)


justinclift:
The idea here seems to be that when a hostname of `_donotmakedsn` is used, that instead triggers an alternative behaviour.

While it sounds like it would work, is this a good approach?

I'm not sure, but it feels to me like maybe the Oracle data source could have some different config option(s).  Maybe something to directly choose a servicename instead of a hostname?

That's just my rough guessing anyway. :smile:

konnectr:
I think this is a change aimed at greater flexibility of connecting to the Oracle database. For my part, I would like to see a description of the possibility somewhere so that more people know about this possibility. And did not use 1 person :)

snickerjp:
@justinclift 

>While it sounds like it would work, is this a good approach?

I thought the same thing.
However, with the current implementation in query_runner/oracle.py, I did not have the ability to refactor it to choose another method.

So, this time, I decided not to do makedsn.
I shared one option with you.

snickerjp:
@konnectr 
thanks.

added Comment in Settings menu.
I shared one option with you.

redash/query_runner/oracle.py
```diff
@@ -66,7 +66,7 @@ def configuration_schema(cls):
             "properties": {
                 "user": {"type": "string"},
                 "password": {"type": "string"},
-                "host": {"type": "string"},
+                "host": {"type": "string", "title": "Host (When type _donotmakedsn, DSN Service Name is used instead.)"},
```
justinclift, line 69: For the text, I think this will sound more "natural":

    To use a DSN Service Name instead, use the text string `_donotmakedsn` in the host name field.

Hmmm.  Reading that back though, the special name of "do not make dsn" is confusing when you want to make a ***DSN*** Service Name.

justinclift: Would an option name like `_useservicename` be clearer?

snickerjp: Thank you.

I'll try to rewrite it.

snickerjp: Thank you.

You're right, it's not a good name.
Agreed.

snickerjp: >Would an option name like _useservicename be clearer?

I think this name is easier to understand the meaning.

konnectr (approved):

snickerjp:
UI

donotmakedsn03

justinclift:
That string looks a bit ugly in the screenshot.  I *think* the `(` and `)` are the problem.  

Would you be ok to try this?

```
"Host: To use a DSN Service Name instead, use the text string `_useservicename` in the host name field."
```

I *think* that will look a bit better in the screenshot.

---

The `backend-lint` test is failing because of formatting problem.

If you have the Python dependencies installed, then `make format` should fix the problem:

https://github.com/getredash/redash/wiki/Local-development-setup#configuring-pre-commit

---

The `frontend-e2e-tests` test failure seems like a problem with the Yarn package repository.  That will hopefully go away if we try again in a few minutes. :smile:

redash/query_runner/oracle.py
justinclift: Cool. :smile:

snickerjp:
> Would you be ok to try this?

It certainly looks, not very desirable.
I understand. I will give it a try.

justinclift:
That is probably the best we can do for the UI for now. :smile:

snickerjp:
> The backend-lint test is failing because of formatting problem.

I'll try this one too.

justinclift:
The failure of the `backend-lint` test this time seems to be a GitHub infrastructure problem this time.

GitHub has not been 100% reliable over the last few weeks. :frowning:

I'll wait a few minutes, then restart that test.

justinclift:
Merged!  Thanks for this @snickerjp, it's a good idea. :smile:

snickerjp:
Thanks for your help.

PR #6342, opened 2023-08-02
Incomplete URL substring sanitization

konnectr:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/bin/release_manager.py
+++ b/bin/release_manager.py
@@ -14,8 +14,7 @@ repo = "getredash/redash"
 
 
 def _github_request(method, path, params=None, headers={}):
-    host = urlparse(path).hostname
-    if not host.endswith("api.github.com"):
+    if urlparse(path).hostname != "api.github.com":
         url = "https://api.github.com/{}".format(path)
     else:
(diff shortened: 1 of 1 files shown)


justinclift (changes requested):
`host.endswith()` seems like the wrong thing here, because `urlparse(path).hostname` is only going to return the hostname part.

Testing quickly on my system just now:

```
$ python3
>>> from urllib.parse import urlparse
>>> path = "https://api.github.com/repos/getredash/redash"
>>> host = urlparse(path).hostname
>>> host
'api.github.com'
```

So, the `host` variable is never going to have `https://` inside it.

Maybe just check if the host is equal to `api.github.com`, as that should be complete enough?

justinclift:
GitHub infrastructure is ***way*** too unreliable recently. :frowning:

PR #6349, opened 2023-08-06
Fix typo from ElasticSearch to Elasticsearch

kaibadash:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Fix typo from ElasticSearch to Elasticsearch. 

image

`Elasticsearch` is valid.

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually
    - スクリーンショット 2023-08-06 13 43 00
- [ ] N/A


konnectr:
What is the meaning of these changes ? For me personally, it makes sense to change the qr name, but not to change the name of the class

justinclift:
It's probably for consistency.  As long as it passes CI and doesn't introduce any other problems, then this change should be ok (in my opinion).  It's making the code base "better". :smile:

justinclift:
@guidopetri @wlach Any interest in taking a look at this PR?

It *seems* pretty straight forward to me. :smile:

wlach (changes requested):
Inclined to agree with @konnectr, changing the display string to match the product name makes sense, but I think using camelcase for the internals makes sense. Certainly doesn't seem worth the risk of breaking things.

justinclift:
Thanks @wlach, sounds like a plan.  :smile:

kaibadash:
Thank you for your review :+1: I'll fix it asap.

redash/query_runner/elasticsearch2.py
```diff
@@ -39,7 +39,7 @@ class ElasticSearch2(BaseHTTPQueryRunner):
 
     @classmethod
     def name(cls):
-        return "ElasticSearch"
+        return "Elasticsearch"
```
wlach, line 42: This is the only change we should be keeping IMO

wlach (changes requested):
Hey, sorry for the lateness in getting back to you on this. Unfortunately, this version has the same problem as before: we're changing the implementation (IMO unnecessary) when all that really needs to be updated is a single display string.

kaibadash:
Thank you for your review!



I think both `OpenDistroSQLElasticSearch` and `XPackSQLElasticSearch` should be corrected. Should I change only `Elasticsearch` ?

guidopetri:
No, I think we can change all 3. Thanks :)

The test error is something unrelated and if you rebase your branch on `master` or merge `master` in, it should be fixed.

justinclift:
This PR confuses me.  Isn't it supposed to only be changing end user visible strings?

However, it looks like it's changing Python class names.

Oh... unless the front end is using the actual Python class name itself as a user visible string?

guidopetri:
Indeed; if you look in line 254 for example, `cls.__name__` returns the python class name. We should change that to hardcode the string in, following @wlach 's advice of not changing the actual class name.

guidopetri (approved):

justinclift:
@kaibadash Thanks for sticking with this, and sorry we kind of changed things around on you meanwhile. :sweat_smile:

Issue #6323, opened 2023-07-31
We have 3 different config files for `flake8`.  Make it just one.


justinclift:
We use the `flake8` Python utility to check the formatting of our Python source code.

However, we have three different configurations of it at the moment.

It's probably because we run `flake8` from three different places.

Ideally, we should just have a single configuration file that gets used by all of our `flake8` calls instead.

### Existing flake8 configuration files

* https://github.com/getredash/redash/blob/f6ba9501da125b308f8e475b5bb303d709d228b8/bin/flake8_tests.sh
  * This one is used by `make test` and `make lint`.
  * https://github.com/getredash/redash/blob/f6ba9501da125b308f8e475b5bb303d709d228b8/Makefile#L34-L35

* https://github.com/getredash/redash/blob/f6ba9501da125b308f8e475b5bb303d709d228b8/.pre-commit-config.yaml#L11-L14
  * This one is used by our `pre-commit` hook

* https://github.com/getredash/redash/blob/f6ba9501da125b308f8e475b5bb303d709d228b8/setup.cfg#L4-L10
  * I'm not sure where this one is used.

TrainedPro:
Hello, I am interested in working on this issue.

We could start by deciding what flake8 configuration you would actually like for all the files and how many instances i would have to replace. I personally think using the configuration that is most widely used would be the best.

Let me know how to proceed.

justinclift:
@TrainedPro Cool. :smile:

We seem to be using this one the most, as it gets run every time we run our backend lint tests:

https://github.com/getredash/redash/blob/f6ba9501da125b308f8e475b5bb303d709d228b8/bin/flake8_tests.sh

justinclift:
@TrainedPro Oh, I forgot to mention something important.  This is the "local development setup guide" we know works:

https://github.com/getredash/redash/wiki/Local-development-setup

There's older setup information on the official Redash knowledge base, but it's outdated and confusing.

So, please use the above wiki link instead. :smile:

TrainedPro:
Here are the **10** mentions for `flake8` that i found:
```
./setup.cfg:4:[flake8]
./bin/flake8_tests.sh:5:flake8 --version ; pip --version
./bin/flake8_tests.sh:7:flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
./bin/flake8_tests.sh:9:flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
[...]

PR #6352, opened 2023-08-07
Changed pre-commit to use the setup.cfg options instead of its own

TrainedPro:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Changed the `.pre-commit-config.yaml` file to use the `setup.cfg` file instead of its own configuration.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [x] N/A

#6323

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,4 +13,5 @@ repos:
     hooks:
       - id: flake8
         exclude: "migration/.*|.git|viz-lib|node_modules|migrations|bin/upgrade"
+        args: [--config, setup.cfg]
 


.pre-commit-config.yaml
```diff
@@ -13,4 +13,5 @@ repos:
     hooks:
       - id: flake8
         exclude: "migration/.*|.git|viz-lib|node_modules|migrations|bin/upgrade"
```
konnectr, line 15: I think we should delete this line. And test without using it

justinclift: Yeah, I think you're right.  In theory, if it's truly parsing and using the options in setup.cfg then this line won't be needed.

Saying that because I thought we already tried the "args" option below, and it didn't work for some unknown reason.  (Note that I'm only about 50% confident we tried the same option as below, it might have been some other options we tried and I'm mis-remembering exactly what.)

wlach:
I think we might want to consider merging https://github.com/getredash/redash/pull/6441 instead of this one, it kind of renders this obsolete by replacing flake8 with an IMO better solution.

guidopetri:
@TrainedPro thanks for the PR, but as wentiened above we ended up moving to `ruff` anyway.

PR #6355, opened 2023-08-08
added duckdb query runner

ashutosh6500:
## What type of PR is this? 

-  New Query Runner (Data Source) 

## Description

This PR adds support for duckdb query runner. DuckDB is an OLAP database used by data professionals, such as data scientists and analysts, to analyze data in a fast and efficient manner . This supports querying tables from duckdb files.

## How is this tested?

- Manually

This query runner is tested with redash docker based setup and environment.For sample .duckdb files, it fetches tables and sample queries are also checked.

image


redash/query_runner/duckdb.py
```diff
@@ -0,0 +1,83 @@
+import logging
+import duckdb
```
justinclift, line 2: Doesn't this mean there should be a matching "duckdb" entry in one of our requirements*.txt files?

ashutosh6500: Yes there is. I forgot to add that one. Will commit with that change. Thanks.

ashutosh6500:
I am facing issue while loading schema in schema dashboard. Getting schema refresh failed. For some query runners that works. So how can I do that for this one?

PR #6359, opened 2023-08-08
pre-commit add pyupgrade

konnectr:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
I believe that we should grow together with the python version that we support. In this regard, I don't want to see outdated things in the code that can confuse beginners. The code is written 1 time, and read 1000 times, and it should be monotonous and with the same style. I believe that this pre-commit-hook allows you to achieve this!

https://github.com/asottile/pyupgrade

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,12 @@
 repos:
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.7.0
+    hooks:
+      - id: pyupgrade
+        args:
+          - --py38-plus
+          - --keep-runtime-typing
+        exclude: "migration/.*|.git|viz-lib|node_modules|migrations|bin/upgrade"
   - repo: https://github.com/PyCQA/isort
     rev: 5.12.0
     hooks:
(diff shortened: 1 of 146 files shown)


justinclift:
@wlach @gaecoli @junnplus What are your thoughts on this? :smile:

gaecoli:
> @wlach @gaecoli @junnplus What are your thoughts on this? 😄

I think we should keep style. 
I don't know why you want to change the code format here.

konnectr:
Hardly anyone understands what style of writing code should be followed. Because of this, linters have been added. This is one of these linters

gaecoli:
Too confusing code format is not acceptable to everyone. This change breaks the test, if it can cover the test well, I'm fine.

wlach:
I see merit in the idea of using more recent python idioms, but this type of change requires somewhat careful review which I'm not able to provide right now. 
I'd also question whether pyupgrade is the tool we want to use here (ruff does some of same things and has some other advantages). Generally these types of changes are done and reviewed by long standing contributors who have been around for a while and can weigh the various trade offs involved. I'm not sure if anyone (certainly not me) can say that at this point.

Not my call to make, I'd suggest putting a pause on changes like this until we manage to release a new version.

konnectr:
ok, close pr

PR #6373, opened 2023-08-16
Add offset for query_runner/oracle.py Error Message.

snickerjp:
## What type of PR is this? 

- [x] Refactor
- [x] Feature

## Description

* add **Offset number** for Error Message.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

https://github.com/getredash/redash/discussions/6367

add offset.

image


snickerjp:
May I ask you to point out what it looks like?

* A
image

* B
image

justinclift:
The comma on the end doesn't look good. :wink:

`... properly ended. , Error ...`

This would look better:

`... properly ended. Error ...`

---

However, we should try to convert the offset to "line X, column Y" instead (if possible) before displaying for the user.

To do that, we should be able to use the query text and counting the number of lines in it (etc) to get an accurate result.  Hopefully. :smile:

snickerjp:
Thanks for reviewing.

I will try to correct the notation as you suggested.

I'm not sure I'm capable of doing "convert the offset to "line X, column Y"" right now.

snickerjp:
> I'm not sure I'm capable of doing "convert the offset to "line X, column Y"" right now.

I might be able to do it.

I'll try.

justinclift:
@snickerjp On the general subject of Oracle (for Redash), it looks like there is a new `python-oracledb` driver these days, replacing the older `cx_Oracle` approach.

Do you feel we should investigate moving to that at some point? :smile:

justinclift:
Oh.  Oracle appears to have online Forums as well: https://forums.oracle.com

If there is a programming section there, we could ask if the Oracle driver already has some function to convert the returned character offset to "line X column Y" format.  Might be better than writing our own function for it. :smile:

snickerjp:
@justinclift 
> Do you feel we should investigate moving to that at some point? 😄

Yes. I think so.

The migration to python-oracledb was tested in my repository.
https://github.com/snickerjp/redash/pull/1

I was about to submit an issue.

snickerjp:
> Oh. Oracle appears to have online Forums as well: https://forums.oracle.com/

Thank you very much.

I will check the forum.

redash/query_runner/oracle.py
```diff
@@ -137,6 +137,17 @@ def output_handler(cls, cursor, name, default_type, length, precision, scale):
                     arraysize=cursor.arraysize,
                 )
 
+    @classmethod
+    def calculate_line_and_column(query, offset):
```
justinclift, line 141: Any possibility of adding a test for this new function? :smile:

snickerjp:
> convert the offset to "line X, column Y"

I was able to convert from offset to line&column.

I will continue to ask on the Oracle forum.

image

redash/query_runner/oracle.py
snickerjp: I understand.

I will try to add the test.

justinclift:
> I was able to convert from offset to line&column.

I *think* that the column number there is off by 1.  Column 12 in the screenshot seems like it would be the letter `t` if starting the count with number 1. :smile:

For the error message output, you can probably drop the `Error Offset num: 29`.  That would leave just the line and column information, which is better. :smile:

justinclift:
> The migration to python-oracledb was tested in my repository.

Cool.  It looks like there's a new 1.4.0 version of it released today too.

https://pypi.org/project/oracledb/#history

No idea if that's useful though. :smile:

---

Something interesting about this new `oracledb` driver.  The documentation for it says it doesn't need the Oracle client libraries installed for standard usage.

That *might* mean we can include the `oracledb` driver in the "all data sources" file instead:

https://github.com/getredash/redash/blob/master/requirements_all_ds.txt

snickerjp:
> I think that the column number there is off by 1. Column 12 in the screenshot seems like it would be the letter t if starting the count with number 1. 😄

It would seem so. It is strange indeed.
I will investigate.

> For the error message output, you can probably drop the Error Offset num: 29. That would leave just the line and column information, which is better. 😄

I would remove "Error Offset num".

snickerjp:
Please wait a moment, as it is taking time to correct the function execution error for testing.

redash/query_runner/oracle.py
```diff
@@ -176,7 +187,11 @@ def run_query(self, query, user):
                 json_data = json_dumps(data)
                 connection.commit()
         except cx_Oracle.DatabaseError as err:
-            error = "Query failed. {}.".format(str(err))
+            (err_args,) = err.args
+            line_num, column_num = Oracle.calculate_line_and_column(query, err_args.offset)
+            error = "Query failed. {}. Query failed at Line {}, Column {}.".format(
```
justinclift, line 192: Maybe simplify it to something like... ?

    Query failed at line {}, column {}: {}.

That's just a rough guess without trying it though. :smile:

snickerjp:
Fixed line & column numbering calculations.

* When there is an error on the first line
line1-offset

* On line 2 error
line2-offset

* When offset is removed
no-offset

justinclift:
Awesome, that third one (with the offset removed) looks best. :smile:

justinclift (approved):
Looks good to me. :smile:

justinclift:
@snickerjp I'm happy to merge this if you feel it's complete? :smile:

snickerjp:
@justinclift 
Yes. This is done.

justinclift:
Awesome!

Merged. :smile:

snickerjp:
> If there is a programming section there, we could ask if the Oracle driver already has some function to convert the returned character offset to "line X column Y" format. Might be better than writing our own function for it. 😄

I asked in the discussion and was told that this feature does not currently exist.

https://github.com/oracle/python-oracledb/discussions/214

PR #6379, opened 2023-08-20
Feature/adir/change owner name for schedualad tasks

adirhmonday:
## What type of PR is this? 

- [x] Refactor

## How is this tested?

- [x] Manually

in Redash staging, we will schedule queries and check the metadata sent with the query, hoping that the username was changed to the actual user and not hardcoded "schedualed"

https://monday.monday.com/boards/2584951506/pulses/5017346555


justinclift:
@adirhmonday Looks like you're having trouble with Pull Request creation?

Are you actually trying to propose a Pull Request, or are you trying to do something else?

I'm happy to provide pointers if that's useful. :smile:

Issue #6380, opened 2023-08-20
Switch the Oracle Driver module from cx_Oracle to python-oracledb


snickerjp:
This is a issue to switch the Oracle Driver module from cx_Oracle to python-oracledb.

Please note the following points with this change.

- The supported Python version is 3.6 or higher.🐍
- The supported Oracle database version is 12c or higher.
  - Thin mode: 12c or higher.
  - Thick mode: 9.2 or higher. (with Oracle Client library)
- In Thin mode, you **do not need the Oracle Client library**💡
- The **ENCODING option has been deprecated**. Please set the NLS_LANG environment variable instead.❌
  - https://python-oracledb.readthedocs.io/en/latest/user_guide/globalization.html#setting-the-client-character-set
- For other considerations from cx_Oracle changes, please refer to the following documents.
  - https://oracle.github.io/python-cx_Oracle/#features
  - https://oracle.github.io/python-oracledb/#features
  
 https://github.com/getredash/redash/pull/6373#issuecomment-1682641246

justinclift:
Cool.  This sounds like a good idea.

It would let us include Oracle database support (for Oracle 12c and above) in our standard builds, without needing to do anything special.

I'm not sure how many of our users would use older Oracle versions, therefore needing the thick mode / Oracle Client library.

My feeling is we should start with the easy approach first, and only consider adding in the Oracle Client library pieces if there is real demand for it.

justinclift:
I *think* that with PR #6389 merged, this issue can be closed now? :smile:

snickerjp:
Of course,
I'll keep it closed.

PR #6386, opened 2023-08-23
Replace cx_Oracle with python-oracledb

snickerjp:
## What type of PR is this? 

- [x] Refactor
- [x] Feature
- [x] Other

## Description

This is a pull request to switch the Oracle Driver module from cx_Oracle to python-oracledb.

Please note the following points with this change.

- The supported Python version is 3.6 or higher.🐍
- The supported Oracle database version is 12c or higher.
  - Thin mode: 12c or higher.
  - Thick mode: 9.2 or higher. (with Oracle Client library)
- In Thin mode, you do not need the Oracle Client library⭕️
- The ENCODING option has been deprecated. Please set the NLS_LANG environment variable instead.❌ -> `The environment variable "NLS_LANG" is also used now.`
  - https://python-oracledb.readthedocs.io/en/latest/user_guide/globalization.html#setting-the-client-character-set
- For other considerations from cx_Oracle changes, please refer to the following documents.
  - https://oracle.github.io/python-cx_Oracle/#features
  - https://oracle.github.io/python-oracledb/#features
- ~This PR will not include changes to `requirements_all_ds.txt`.~ c5f6aa837cedd580560dfb1a244cfb406e6249a9

Please review and let me know if you have any comments. Thank you.🙇‍♂️

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually

#6373
#6380

no UI change.

### DIFF
--- a/requirements_oracle_ds.txt
+++ b/requirements_oracle_ds.txt
@@ -1,4 +1,5 @@
-# Requires installation of, or similar versions of:
+# If you want to use thick mode, you need to install the Oracle client library.
+# reference. https://pypi.org/project/oracledb/#description
 #    oracle-instantclient12.2-basic_12.2.0.1.0-1_x86_64.rpm
 #    oracle-instantclient12.2-devel_12.2.0.1.0-1_x64_64.rpm
-cx_Oracle==5.3
+oracledb==1.3.2
--- a/requirements_oracle_ds.txt
+++ b/requirements_oracle_ds.txt
@@ -2,4 +2,3 @@
 # reference. https://pypi.org/project/oracledb/#description
 #    oracle-instantclient12.2-basic_12.2.0.1.0-1_x86_64.rpm
 #    oracle-instantclient12.2-devel_12.2.0.1.0-1_x64_64.rpm
-oracledb==1.4.0
--- a/requirements_oracle_ds.txt
+++ b/requirements_oracle_ds.txt
@@ -2,4 +2,4 @@
 # reference. https://pypi.org/project/oracledb/#description
 #    oracle-instantclient12.2-basic_12.2.0.1.0-1_x86_64.rpm
 #    oracle-instantclient12.2-devel_12.2.0.1.0-1_x64_64.rpm
-oracledb==1.3.2
+oracledb==1.4.0
--- a/requirements_oracle_ds.txt
+++ b/requirements_oracle_ds.txt
@@ -1,4 +1,0 @@
-# If you want to use thick mode, you need to install the Oracle client library.
-# reference. https://pypi.org/project/oracledb/#description
-#    oracle-instantclient12.2-basic_12.2.0.1.0-1_x86_64.rpm
-#    oracle-instantclient12.2-devel_12.2.0.1.0-1_x64_64.rpm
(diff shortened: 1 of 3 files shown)


requirements_oracle_ds.txt
```diff
@@ -1,4 +1,5 @@
-# Requires installation of, or similar versions of:
+# If you want to use thick mode, you need to install the Oracle client library.
+# reference. https://pypi.org/project/oracledb/#description
 #    oracle-instantclient12.2-basic_12.2.0.1.0-1_x86_64.rpm
 #    oracle-instantclient12.2-devel_12.2.0.1.0-1_x64_64.rpm
-cx_Oracle==5.3
+oracledb==1.4.0
```
justinclift, line 5: As this will work without needing the Oracle client pieces installed first, we should probably move this `oracledb=1.4.0` stuff into the [requirements_all_ds.txt](https://github.com/getredash/redash/blob/master/requirements_all_ds.txt) file instead. :smile:

justinclift (changes requested):

requirements_oracle_ds.txt
snickerjp: > we should probably move this oracledb=1.4.0 stuff into the [requirements_all_ds.txt](https://github.com/getredash/redash/blob/master/requirements_all_ds.txt) file instead. 😄

I agree.
However, I thought it would be a major change and would prefer another opportunity.

I will try to move it to requirements_all_ds.txt and test the dependencies.

snickerjp:
Moved python-oracledb from requirements_oracle_ds.txt to requirements_all_ds.txt.

I was also able to test the build using the Dockerfile in the master branch.

snickerjp:
I will try to delete it.

However, I will also need the information in `requirements_oracle_ds.txt`, so I will leave it somewhere.

snickerjp:
deleted file `requirements_oracle_ds.txt` .

justinclift (approved):
Awesome!  Looks good to me. :smile:

justinclift:
> However, I will also need the information in requirements_oracle_ds.txt, so I will leave it somewhere.

Oops, I didn't notice that comment until just now.

You should be ok to move those lines to the other requirements_all_ds.txt file.  Something like this:

```
...
pinotdb>=0.4.5
pygridgain==1.4.0
pyignite==0.6.1
# Requires installation of, or similar versions of:
#    oracle-instantclient12.2-basic_12.2.0.1.0-1_x86_64.rpm
#    oracle-instantclient12.2-devel_12.2.0.1.0-1_x64_64.rpm
oracledb==1.4.0
```

justinclift:
Hmmm, we will probably want to update the comments.

They need to be clear that those files are only needed in certain situations.  Maybe something more like this?

```
...
pinotdb>=0.4.5
pygridgain==1.4.0
pyignite==0.6.1
# oracledb supports Oracle 12.c and above without needing the Oracle
# client libraries installed.  To support Oracle 9.x and above as well,
# then the Oracle client libraries will need installing first.  eg:
#    oracle-instantclient12.2-basic_12.2.0.1.0-1_x86_64.rpm
#    oracle-instantclient12.2-devel_12.2.0.1.0-1_x64_64.rpm
oracledb==1.4.0
```

snickerjp:
I understand.
I will try to add it.

PR #6403, opened 2023-08-29
Update snowflake application name

davidln777:
## What type of PR is this? 

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Adding the ability to classify Redash tool when connecting to Snowflake, using the Snowflake SQL connector available header called "application"
from SnowflakeConnection (connection.py):
"application: Application name to communicate with Snowflake as. By default, this is "PythonConnector"."

I took the same format Redash use in the Databricks connection ("Redash/{} (Databricks)".format(__version__.split("-")[0]))  and applied it to the Snowflake connection as:
"Redash/{} (Snowflake)".format(__version__.split("-")[0]). 
In Snowfllake SQL connector we are unable to update the user-agent header (as in Databricks), so I'm using the available application header. 

This will help proxies / WAF services to classify a Redash connection. 
## How is this tested?

- [ ] E2E Tests (Cypress)
- [X] Manually
- [ ] N/A

I run Redash locally and using a proxy to Snowflake I've verified the application header to contain - "Redash/10.1.0 (Snowflake)"

NA

NA


justinclift:
Interesting concept.

@guidopetri What do you reckon? :smile:

guidopetri (approved):
LGTM! Thanks @davidln777

justinclift:
Wonder if this connection header change should be made the default for all of our query runners?

guidopetri:
I'm not sure it's available for all the query runners. Before merging yesterday I verified that it's available in the snowflake package, but I don't think there's a standard in place. Definitely something we should explore though.

PR #6416, opened 2023-09-01
Automatically tag release candidates

eradman:
Format similar to Ubuntu: _YY.MM.N_

Discussion thread: https://github.com/getredash/redash/discussions/6411

## What type of PR is this? 

- [x] Feature

## How is this tested?

Replace github `schedule` with `workflow_dispatch` to trigger manually

```yaml
on: workflow_dispatch
```

### DIFF
--- a/.github/workflows/periodic-snapshot.yml
+++ b/.github/workflows/periodic-snapshot.yml
@@ -1,0 +1,26 @@
+name: Periodic Snapshot
+
+# 10 minutes after midnight on the first of every month
+on:
+  schedule:
+  - cron: "10 0 1 * *"
+
+permissions:
+  contents: write
+
+jobs:
+  bump-version-and-tag:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: |
+          date="$(date +%y.%m).0-dev"
+          gawk -i inplace -F: -v q=\" -v tag=$date '/^  "version": / { print $1 FS, q tag q ","; next} { print }' package.json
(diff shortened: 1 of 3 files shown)


guidopetri (approved):

eradman:
As far as I can tell yarn will not accept `23.09` as a version, it must be 23.09.0.

I suppose `.0` could be the automatic tag, and `.1`, `.2`, `.3` etc. the release versions

guidopetri:
What about the other way around? `<version>.YY.MM` would let us keep some notion of "upgrade major versions only if you are sure about it".

justinclift:
To me, `23.09.0` seems like it would be the right choice.

That's:

* Major version `23`
* Month being `09`
* Patch / release of the series = `0`

eradman:
@justinclift I adjusted this PR to match scheme you proposed. It makes yarn happy and it's easy to understand.

justinclift (approved):
Looks good to me. :smile:

justinclift:
Cool, good stuff.  I've just set it to auto-merge as soon as the E2E tests pass. :smile:

lucasfcnunes:
@justinclift are the bump in the Pre-release and Latest tags already planned?

justinclift:
When you say "bump" there, what are you meaning? :smile:

lucasfcnunes:
In this case the labels (Latest, Pre-release, ...) going from version A to version B.
Or simply: version A to version B.

image


https://nvie.com/posts/a-successful-git-branching-model/
https://stackoverflow.com/a/10834954/6446567

PR #6417, opened 2023-09-02
Try our experimental self-hosted runner

justinclift:
## What type of PR is this? 

- [x] Other

## Description

We've set up an experimental "Self Hosted" GitHub runner on an older Hetzner dedicated host (Ryzen 3700X).  Lets see if the CI tests run faster using that.

## How is this tested?

- [x] N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ env:
   NODE_VERSION: 16.20.1
 jobs:
   backend-lint:
-    runs-on: ubuntu-22.04
+    runs-on: self-hosted
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


justinclift:
k.  Looks like GitHub's Self Hosted Runners are pretty bare bones.  I was expecting they'd run things automatically inside docker containers (like, say, Jenkins self-hosted runners).

Will try again shortly, after doing some more setup in the node...

justinclift:
Well, this looks like a pretty solid fail.  Nested docker (inside an LXC container) is executing very, very slowly.

This *might* be workable on bare metal equipment instead, but I personally don't have any spare that's of a reasonable spec / likely to be faster than GitHub's provided stuff.

PR #6421, opened 2023-09-04
Add the UTF-8 BOM at the beginning of the response

GunSik2:
## What type of PR is this? 

- [x] Refactor
- [x] Feature

## Description
Hello.
Since I need to enable opening an CSV file with Hangul(Korean) in excel, I added the character BOM code in response.
Best Regards.

## How is this tested?
- [x] Manually

Download query csv file and open it with execl.

N/A

No change.

### DIFF
--- a/redash/serializers/query_result.py
+++ b/redash/serializers/query_result.py
@@ -1,5 +1,6 @@
 import csv
 import io
+import codecs
 
 import xlsxwriter
 from dateutil.parser import isoparse as parse_date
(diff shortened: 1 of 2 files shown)


justinclift:
The concept of this "add UTF BOM to the CSV" sounds reasonable.  I'm not sure if blindly adding it like this is the right approach though.

@guidopetri @gaecoli @konnectr What are your thoughts on this? :smile:

gaecoli:
> Hmm, we should at least verify this works for non- UTF-8 files.
> 
> Would it be better to set the encoding?

Agree!

GunSik2:
> Hmm, we should at least verify this works for non- UTF-8 files.
> 
> Would it be better to set the encoding?

Thanks for your comment.
Can you tell me a link for setting the encoding? 
I want to take a closer look.

guidopetri:
It looks like [the value returned](https://docs.python.org/3/library/io.html#io.StringIO.getvalue) is a `str` type, so it should be something like `str.encode('utf-8')`. Hopefully that works?

redash/serializers/query_result.py
```diff
@@ -81,13 +82,15 @@ def serialize_query_result(query_result, is_api_user):
 
 
 def serialize_query_result_to_dsv(query_result, delimiter):
-    s = io.StringIO()
+    s = io.BytesIO()
+    s.write(codecs.BOM_UTF8)
```
guidopetri, line 86: @GunSik2 sorry, I'm still confused. Why do we need to write the BOM manually here if we use UTF-8 later? And why do we need `TextIOWrapper`?

justinclift:
@GunSik2 Ping?

guidopetri:
@GunSik2 hey, I'm going to close this PR for now, but please feel free to reopen.

PR #6422, opened 2023-09-04
Move from oauth2client to google-auth / support ADC

sekiyama58:
## What type of PR is this? 
- [x] Refactor
- [x] Feature

## Description

[oauth2client](https://pypi.org/project/oauth2client/), the library to be used to authenticate with Google services, was deprecated in 2017 and is no longer maintained. This PR rewrites the service account credentials code and replaces it with [google-auth](https://pypi.org/project/google-auth/).

In addition, this makes the JSON key file optional for the data source and uses the application's default credentials (ADC) if the JSON key file is omitted. This enables support for a variety of GCP authentication methods, including the gcloud CLI, GCE metadata servers, and GKE or AWS Workload Identity.
These methods are more secure than manually managing service account keys.

Please see the document for more details about ADC:
https://cloud.google.com/docs/authentication/application-default-credentials

Note that the bigquery_gce functionality is covered by the ADC feature, but is retained for compatibility reasons.

[gspread](https://pypi.org/project/gspread/) library is also updated to latest version to support google-auth. (At least version 3.5.0 is required)

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually

Each modified datasource are tested manually to check weather it can get the google credentials from a service account JSON file and ADC.

https://redash.io/help/data-sources/querying/bigquery : JSON Key file will be optional if ADC can be used.

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -1,9 +1,8 @@
 import datetime
 import logging
+import socket
 import time
 from base64 import b64decode
-
-import httplib2
 
 from redash import settings
 from redash.query_runner import (
@@ -23,9 +22,10 @@ logger = logging.getLogger(__name__)
 
 try:
     import apiclient.errors
+    import google.auth
     from apiclient.discovery import build
     from apiclient.errors import HttpError  # noqa: F401
-    from oauth2client.service_account import ServiceAccountCredentials
+    from google.oauth2.service_account import Credentials
 
     enabled = True
 except ImportError:
@@ -112,7 +112,7 @@ class BigQuery(BaseQueryRunner):
             "type": "object",
             "properties": {
                 "projectId": {"type": "string", "title": "Project ID"},
-                "jsonKeyFile": {"type": "string", "title": "JSON Key File"},
+                "jsonKeyFile": {"type": "string", "title": "JSON Key File (ADC is used if omitted)"},
                 "totalMBytesProcessedLimit": {
                     "type": "number",
                     "title": "Scanned Data Limit (MB)",
@@ -138,7 +138,7 @@ class BigQuery(BaseQueryRunner):
                     "default": False,
                 },
             },
-            "required": ["jsonKeyFile", "projectId"],
+            "required": ["projectId"],
             "order": [
                 "projectId",
                 "jsonKeyFile",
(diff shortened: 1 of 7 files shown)


guidopetri:
@sekiyama58 , thanks for the contribution! This does indeed seem super useful. Would you be able to add some tests to this functionality? Or is this covered by existing tests?

I'll look into why the CI is failing but I think we might just have to try again. :)

guidopetri:
Merged `master` in to retry the CI.

sekiyama58:
@guidopetri Thanks for your comment. Google services' authentications seems not covered by existing tests. Since it is highly dependent on `google-auth` and `google-api-python-client` libraries, almost every line of the diff needs mocks for these library, that sounds like ineffective test. Should I take this approach?

sekiyama58:
Another method for authentication testing is to provide a valid Google service account key and call a function like "redash.query_runner.google_spreadsheets.GoogleSpreadsheet({})._get_spreadsheet_service()" to ensure that no exceptions occur. The key is only used for user authentication, so there is no need to enable the API. I need a living GCP project and its key to be set to a secret environment variable in CI. Only administrators of this repository can do this.

justinclift:
> I need a living GCP project and its key to be set to a secret environment variable in CI. Only administrators of this repository can do this.

*Probably* no a go-er in the short term.  We've been kicking around the idea of doing that kind of thing for data source testing too.  eg set up some test data sources and run queries again them.  Somewhat simple in concept (at a high level)  for the OSS based data sources.

Proprietary ones though would likely need some kind of longer term "set things up and use it repeatedly" kind of thing, as would testing GCP auth stuff.

The actual implementation and longer term maintenance for any of it is likely to be a bit tricky and need a bunch of care. :wink:

sekiyama58:
> The actual implementation and longer term maintenance for any of it is likely to be a bit tricky and need a bunch of care
@justinclift Agreed. I feel that it is better to leave GCP auth codes as it is without unit tests until the *longer term* data source testing is implemented, rather than adding tests with lots of hard-to-maintain library mocks.
However, the CI test is failing, so if you would like to add mocked unit tests, please let me know.

justinclift:
Yeah.  I think we'll just have to ignore the Code coverage tests in this specific case. :smile:

justinclift:
@guidopetri Is this ok to merge as-is?

guidopetri:
I didn't see anything glaring but I'm also not an expert in auth. If it's been tested manually I'd say we're good to merge

justinclift:
k, looks like we're YOLO-ing it for now. :wink:

Issue #5910, opened 2023-03-15
Use poetry for dependency management?


lucasfcnunes:
The [python-poetry](https://github.com/python-poetry/poetry)  project has ~24k stars and is widely used in modern python projects.

Pros:
- Deprecates requirements*.txt
- Poetry has these group where we could separate the categories - https://python-poetry.org/docs/master/managing-dependencies/#dependency-groups
- poetry.lock - https://python-poetry.org/docs/master/libraries#lock-file
- pyproject.toml - https://python-poetry.org/docs/master/pyproject/ (PEP 518)
- Multi python versions, venvs (version matrix) ~let's deprecate python 3.7~ - https://python-poetry.org/docs/managing-environments/ | https://stackoverflow.com/a/66008971/6446567
- Compatible with [black](https://black.readthedocs.io/en/stable/), [pre-commit](https://[pre-commit](https://pre-commit.com/).com/), ... if we are going to really resuscitate and modernize the project.
- ...

Cons:
- ?

arikfr:
This is interesting, but is there really no cons? 

One issue I can think of is that currently there are some conflicts between the many data source dependencies we use. The version of `pip` we're using manages to ignore/overcome this, but not sure if Poetry will?

Will you be interested in giving this a try (converting the requirements*.txt files into pyproject.toml) and seeing what the issues might be?

lucasfcnunes:
Great take @arikfr!

Last week, as a draft, I did try to refactor with poetry and It went very well. It was too one of my concerns and I did not put it in the cons on purpose haha.

Let's head over to my conclusions after testing for packages conflicts.
- Python ^3.8 (I used python=3.8 in my micromamba env) 
- Maintained the poetry grouping as close as possible to the requirements.txt file divisions (which is not a MUST)
- Bumped lots of packages (not all) versions to latest (which will possibly break some of our code making refactoring likely)
- pymapd had to be removed because of conflicts (BUT it's deprecated "Existing scripts should be migrated to pyomnisci from pymapd, this library will not be updated moving forward." - https://github.com/heavyai/pymapd)
- Didn't bother putting the oracle reqs
- semversioned v0.x.y packages can be tricky to autobump (reliable CIs and testing can solve this)
- I'm inclined to say that we can make this work because the redash project is big enough to make other projects bump their oficial compatibility matrix faster (even if we can't directly count on that)
- PS (in my view): 
  - old/insecure/incompatible connectors should be refactored (priority) or purged from the project in new releases
[...]

PR #6423, opened 2023-09-05
Convert project to `poetry`

guidopetri:
## What type of PR is this? 

- [x] Other

## Description

This PR converts us from `requirements.txt`-pinned dependencies to `poetry`, which has much more stable dependency chains and should prevent us from having updated dependencies break our project.

I added two groups (besides the main group): `dev`, for development dependencies, and `all_ds`, to keep the separation we had with the requirements files.

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually

Untested as of yet (so don't merge), but the plan is to:

- build the image
- `pip freeze > image_packages.txt`
- on a new installation, install dependencies with `poetry install`
- `pip freeze > poetry_packages.txt`
- `diff image_packages.txt poetry_packages.txt`

#5910 

N/A


justinclift:
No objection here, it sounds like it'll be useful.

@getredash/maintainers @arikfr This ok with yourselves? :smile:

guidopetri:
This isn't quite ready yet :P

snickerjp:
I thought Poetry was a good attempt.

justinclift:
> This isn't quite ready yet :P

No worries.  Was more thinking we should ensure the concept itself is ok with people. :smile:

guidopetri:
#6428

guidopetri:
`grep -Ir requirements .` returns only hits in `./bin/upgrade` now, and I can't find any programmatic use for that script. @arikfr , do you remember what this was used for? Should we update it?

I've made one important change to the Dockerfile: I removed the `skip_ds_deps`, `skip_dev_deps`, `test_all_deps` build args and replaced them with an `install_groups` build argument. This is passed directly to poetry and can control which dependency groups are installed. By default, it will install `main`, `all_ds` and `dev`; I changed the Cypress yaml and CI workflows accordingly. The image builds on my local now with `docker build -t redash .`.

I'm now going to create a venv to compare the poetry install with the requirements.txt install and see if there are any differing packages.

guidopetri:
Ok, here's the diff list I got:

```
55a56
> futures==3.0.5
74c75
< inflection==0.5.1
---
> inflection==0.3.1
115a117
> pip==23.2.1
159c161,162
< rdflib==7.0.0
---
> rdflib==6.3.2
> redash==23.9.0       /home/sid/Git/redash
175a179
> setuptools==68.2.0
210a215
> wheel==0.40.0
```

wheel/setuptools are from my pyenv, and redash is because I installed without `--no-root`. The `inflection` version being different is interesting, as well as `rdflib`. I suspect this is some incompatibility in one of the packages with the newer versions of these two packages. As long as the tests pass, I'm good merging. @getredash/maintainers , any objections to this?

justinclift (changes requested):
Just noticed that this is missing the dependency updates from the rq related PR yesterday: https://github.com/getredash/redash/commit/41495ba940ac9d1bb2beef3456b89b5742f11eec

They should be added yeah?

pyproject.toml
```diff
@@ -16,3 +16,152 @@ multi_line_output = 3
 include_trailing_comma = true
 use_parentheses = true
 skip = "migrations"
+
+[tool.poetry]
+name = "redash"
+version = "23.09.0"
```
justinclift, line 22: Does poetry not understand things like `-dev` in version identifiers?

guidopetri:
d727602 should be the rq changes. Is it not showing up on the diff?

I forgot about the `dev` suffix, I think poetry is yine woth it but let me add that and double check. It looks like I also missed something in the CI.

justinclift:
> Is it not showing up on the diff?

Ahhh, yeah that wasn't showing up for me.  No worries. :smile:

pyproject.toml
guidopetri: Resolved!

guidopetri:
Alright, I've enabled auto-merge on this PR. @justinclift when you have a chance, can you re-review to merge?

arikfr:
`bin/upgrade` is a relic from the pre-docker days. We should just delete it.

pyproject.toml
```diff
@@ -16,3 +16,151 @@ multi_line_output = 3
 include_trailing_comma = true
 use_parentheses = true
 skip = "migrations"
+
+[tool.poetry]
+name = "redash"
+version = "23.09.0-dev"
+description = "Make Your Company Data Driven. Connect to any data source, easily visualize, dashboard and share your data."
+authors = ["Your Name <you@example.com>"]
```
arikfr, line 24:
```suggestion
authors = ["Redash maintainers and contributors"]
```

Or does it have to be an email address?

guidopetri:
Shoot, the auto-merge didn't quite help here did it. Sorry @arikfr ! I'll create a followup PR. The author does indeed have to include an email address; do we have one on the `redash.io` domain that we could use?

arikfr:
I can set something up on redash.io - just wonder where to send it? Maybe there is some email->discord integration?

justinclift:
> Maybe there is some email->discord integration?

Interesting idea.  We'd better make sure it never sends anything confidential via that channel though (password resets, etc), unless we're sure it can direct to some private for-admins-only channel. :smile:

Issue #6424, opened 2023-09-06
Redis decode_responses rq incompatibility


SchrosCat2013:
Redash is setting the flag `decode_responses` for its Redis connection. This is incompatible with the rq library - https://github.com/rq/rq/issues/1188.

We are having some issues, and I suspect this might be the underlying cause. Primarily:
- scheduled queries look like they're being executed multiple times
- occasionally a query gets stuck. The job running the query disappears but the Redis entry remains, consuming a space in the query pool.

The second issue eventually consumes the entire query pool and prevents any query from running. When this happens, we have to manually flush the Redis cache.

We are running Redash v10.1.0 from the redash/redash:10.1.0.b50633 docker image.

We found https://github.com/getredash/redash/issues/5801, which looks like the same issue we are having. We have upgraded rq to 1.10.1 and rq-scheduler to 0.10.0 in our environment. This does seem to have alleviated the stuck queries issue. It doesn't seem to have changed the scheduled queries multiple executions. However, we are now seeing the error
```UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9c in position 1: invalid start byte```
repeated throughout the logs. The stack trace here matches the linked issue in the rq library. I am not sure if upgrading rq caused these errors - I more suspect that upgrading rq has improved the error logging and is surfacing an existing issue.

**Comment by justinclift** [comment]
Thanks, that looks like an important bug to investigate / fix. :smile:

**Comment by justinclift** [comment]
Our latest development code is using rq `1.5.0` and rq-scheduler `0.10.0`, which both seem fairly old now.: https://github.com/getredash/redash/blob/c2e7df098dc267f35a2122c6cbab8ffb96feec21/requirements.txt#L47-L48

  We should be able to bump them up to much newer, and hopefully also adjust the connection options without too much hassle.

**Comment by justinclift** [comment]
This seems to be where `decode_responses=True` is added:

https://github.com/getredash/redash/blob/c2e7df098dc267f35a2122c6cbab8ffb96feec21/redash/settings/helpers.py#L47-L66

Looks like that was originally added during the migration of Redash from Python 2 to Python 3: https://github.com/getredash/redash/commit/246eca1121fad13fcb873de7d7c4e83345e41059

```
* Decode bytes returned from Redis
```

justinclift:
Not setting that option *might* just be as simple as removing the helper function and removing the calls to it:

[...]

PR #6426, opened 2023-09-06
Update rq related dependencies, to hopefully alleviate issue 6424

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR updates some of the rq related dependencies, as part of potentially fixing #6424.

Note that this *doesn't* update to the very latest versions of `rq` nor `rq-scheduler`, as the newer versions fail our backend unit tests.  To update to those, we'll likely need to update our source code and/or unit tests.

## How is this tested?

- [x] E2E Tests (Cypress)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,12 +40,12 @@ python-dateutil==2.8.0
 python-dotenv==0.19.2
 pytz>=2019.3
 PyYAML==6.0.1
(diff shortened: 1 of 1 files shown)


justinclift:
Merging this, as it's just dependency updates which shouldn't hurt anyway. :smile:

PR #6431, opened 2023-09-07
Remove importlib-resources from requirements.txt

eradman:
This is not required for Python >= 3.8 since the goal of this project is to back port functionality.  Also the version specified is not compatible with later versions of Python.

## What type of PR is this? 

- [x] Upgrade

## How is this tested?

https://pypi.org/project/importlib-resources/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,7 +22,6 @@ gevent==22.10.1
 greenlet==1.1.3
 gunicorn==20.0.4
 httplib2==0.19.0
-importlib-resources==5.13.0
 itsdangerous==2.1.2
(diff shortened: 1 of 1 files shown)


eradman:
I respectfully disagree with codecov!

guidopetri:
Could we wait on this since we're removing these files and moving to poetry anyway?

eradman:
Apologies @guidopetri I missed your comment. Feel free to revert if you think that's easier

guidopetri:
Nah all good, I commented at the same time you merged. I'll make the change later in the poetry pr

PR #6438, opened 2023-09-08
Update Docker build to Debian 12/Python 3.10

eradman:
## What type of PR is this? 

- [x] Upgrade

## Description

- Recommended development host may be Ubuntu 20 or Ubuntu 22
- Use Debian 12 instead of Ubuntu 22 for better compatibility with Microsoft's APT repo
- Going forward, Redash may incorporate features and packages designed for Python 3.10
 
## How is this tested?

- [x] Manually


eradman:
Some conflicts will need to be resolved after #6423 is merged

justinclift:
> Going forward, Redash may incorporate features and packages designed for Python 3.10

We're presently targeting Python 3.8 though.

eradman:
> We're presently targeting Python 3.8 though.

Got it; I agree that targeting Python 3.10 is not necessary yet.

It has been suggested before that we test with multiple versions of Python. I'll close this for now and think about how we can run the backened tests on 3.8 and 3.10.

PR #6441, opened 2023-09-11
Replace flake8/isort with ruff

wlach:
There's a few advantages of using ruff over these tools:

* It's way faster
* It's easier to configure
* It includes support for a bunch of other linters (for example bugbear)
  right out of the box, which catches some things and makes our code
  more consistent.

Ruff works great with black, which I'd recommend we continue using.

Fixed a few minor issues that the new linter combo picked up.


guidopetri (approved):
Had never heard of `ruff`, googled it and became convinced this is probably a better solution in <20s. :D my heart will always love flake8, but this is clearly the better choice. Thanks for pointing this out @wlach !

wlach:
codecov is complaining because the reduced number of lines mean the percentage coverage went down but 0.02% (due to deleted lines of code). I think we can live with that. 😄

justinclift:
I think @konnectr pointed out ruff as a potential option a few weeks ago too.

PR #6445, opened 2023-09-11
Accept Python 3.8-3.10

eradman:
For running a development environment on Ubuntu 22

## What type of PR is this? 

- [x] Upgrade

## How is this tested?

- [x] Manually

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ authors = ["Your Name <you@example.com>"]
 readme = "README.md"
 
 [tool.poetry.dependencies]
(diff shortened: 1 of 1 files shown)


guidopetri:
FYI - when we update the `pyproject.toml` we also have to run `poetry lock` and commit the changes in `poetry.lock`, otherwise they get out of sync

eradman:
Thanks  for pointing this out @guidopetri. `poetry lock` seems to have updated too much. I'll submit another PR with the changes that `poetry update requires-python` makes.

PR #6449, opened 2023-09-12
Avoid file locking issue when running CLI test

eradman:
## What type of PR is this? 

- [x] Bug Fix
- [x] Testing

## Description

`__file__` will resolve to `/app/tests/test_cli.py`, which is already opened by Python. When tests are run on a network file system the test runner may deadlock while waiting for an advisory lock to be released on this file.

## How is this tested?


guidopetri (approved):
Seems reasonable to me.

PR #6450, opened 2023-09-12
Add ruff to Python dev dependencies

eradman:
## What type of PR is this? 

- [x] Development Environment

## Description

`make lint` will run `ruff`, but this tool does not seem to be installed along with other development dependencies.

## How is this tested?

- [x] Manually

```
poetry install --only main,all_ds,dev
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -157,6 +157,7 @@ ptpython = "3.0.23"
 ptvsd = "4.3.2"
 pytest-cov = "4.1.0"
 watchdog = "3.0.0"
+ruff = "0.0.289"
 
 [build-system]
 requires = ["poetry-core"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -157,6 +157,7 @@ ptpython = "3.0.23"
 ptvsd = "4.3.2"
 pytest-cov = "4.1.0"
 watchdog = "3.0.0"
+ruff = "0.0.289"
```
justinclift, line 160: Seems like a good idea.  The dependencies seem to be in alphabetical order though, so it probably needs to be moved up a line. :wink:

guidopetri: I think the alphabetical order is a holdover from `requirements.txt` formatting. Personally I don't care whether it's alphabetical or not /shrug

guidopetri (approved):

pyproject.toml
justinclift: No worries then. :smile:

PR #6466, opened 2023-09-20
Revert PR 6448 (SQLAlchemy-Utils update) for now

justinclift:
## What type of PR is this? 

- [x] Bug Fix

## Description

This PR reverts PR 6448, as that one is causing errors.

We can investigate those errors and update SQLAlchemy-Utils again later on, but for now lets get things working properly again.

## How is this tested?

- [x] N/A


denisov-vlad:
@justinclift looks good except conflict. You need to choose `content-hash` from your branch.

justinclift:
Oh, I didn't notice that.  Thanks @denisov-vlad, I'll fix it now. :smile:

guidopetri:
FYI this PR looks exactly correct for a deps change :)

justinclift:
Thanks.  I used `poetry add somename=x.x.x`, but wasn't sure about the resulting file change.

The changes in pyproject.toml make sense, but the changes in `poetry.lock` don't immediately make sense.  For example:

```
-python-versions = ">=2.7,<3.0.0 || >=3.4.0"
+python-versions = ">=2.7,<3.0.dev0 || >=3.4.dev0"
```

Thinking about it more, that particular change is *probably* something along the lines of "3.0.dev0 was the current minimum 3.x version at the time of sqlalchemy-utils 0.34.2 release".

But that "3.0.dev0" isn't present in the sqlalchemy-utils repo [at the 0.34.2 tag's point in time](https://github.com/kvesteri/sqlalchemy-utils/tree/57ff31ccf7c30c96057849604d4c0486b7ce5df6).

So, I kind of have to just "Trust me bro" and hope for the best with it. :wink:

PR #6469, opened 2023-09-20
Fix Google Spreadsheets AttributeError: 'Credentials' object has no attribute 'access_token'

denisov-vlad:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Google Sheets runner is broken now, because gspread 3.1.0 doesn't support google-oauth:

https://stackoverflow.com/questions/51618127/credentials-object-has-no-attribute-access-token-when-using-google-auth-wi

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A


guidopetri:
Thanks again @denisov-vlad :)

justinclift:
Weirdly, poetry is showing a problem with this PR:

```
$ poetry check
Error: poetry.lock is not consistent with pyproject.toml. Run `poetry lock [--no-update]` to fix it.
```

Looks like the one of the content hashes needs updating.  I'll create a PR to fix that shortly.

PR #6473, opened 2023-09-21
Update poetry.lock after PR 6469

justinclift:
## What type of PR is this? 

- [x] Bug Fix

## Description

This PR fixes a warning from poetry after PR #6469 was merged:

```
$ poetry check
Error: poetry.lock is not consistent with pyproject.toml. Run `poetry lock [--no-update]` to fix it.
```

It was created using `poetry lock --no-update`.

## How is this tested?

- [x] N/A


denisov-vlad:
Sorry, my bad. I've chosen wrong hash during resolving conflicts.

PR #6488, opened 2023-09-28
Avoid updating query result for archived queries

eradman:
## What type of PR is this? 

- [x] Bug Fix
- [x] Optimization

## Description

Refreshing an active query inadvertently updates archived queries with the same `query_hash`. On installations with many archived queries this increases the load on the database and consumes space that may never be reclaimed.

## How is this tested?


justinclift:
Good catch @eradman. :smile:

Issue #4932, opened 2020-06-01
Support query alias to refer query by name, not query_xxx


xhiroga:
Give a name to query make it easy that using queries as components.
This idea is giving query an alias and using it to run queries on query results.

Set query alias.
![image](https://user-images.githubusercontent.com/13391129/83390584-128dfc00-a42d-11ea-8a2c-0d7de1a07568.png)

Use it without query_xxx.
![image](https://user-images.githubusercontent.com/13391129/83390668-3fdaaa00-a42d-11ea-865e-13351479bc2d.png)

arikfr:
@hiroga-cc that's an interesting idea! You should post it in the forum as a feature request. Thanks!

dheerajkhatri:
@arikfr @xhiroga  was this implemented?

justinclift:
Nope.  It's a good idea though.

PR #6490, opened 2023-09-29
Provide stable URL for referencing dashboards

eradman:
## What type of PR is this? 

- [x] Feature

## Description

Dashboards may be referenced if their ID is known (`/dashboards/{dashboard_id}-{slug}`), but there was no easy method of linking one dashboard to another by name.

The route `/dashboard/{slug}` already exists, but was useless because the exact slug ID had to be supplied (`slug`, `slug_1`, `slug_2`, ...). This change improves this routes by finding the best match.

This enables dashboar

## How is this tested?

- [x] Manually

Create several dashboards with the same name, archive the first and test redirects using `/dashboard/{slug}`

## Example Use Case

redash-markdown-links

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -1128,7 +1128,12 @@ class Dashboard(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model
 
     @classmethod
     def get_by_slug_and_org(cls, slug, org):
-        return cls.query.filter(cls.slug == slug, cls.org == org).one()
+        pattern = "{}(_[0-9]+)?".format(slug)
+        return (
+            cls.query.filter(cls.slug.op("SIMILAR TO")(pattern), cls.org == org, cls.is_archived.is_(False))
+            .order_by(cls.id.desc())
+            .first()
+        )
 
     @hybrid_property
     def lowercase_name(self):
--- a/tests/handlers/test_dashboards.py
+++ b/tests/handlers/test_dashboards.py
@@ -56,12 +56,15 @@ class TestDashboardResourceGet(BaseTestCase):
 
         self.assertResponseEqual(expected, actual)
 
-    def test_get_dashboard_with_slug(self):
+    def test_get_dashboard_with_id_slug(self):
         d1 = self.factory.create_dashboard()
+        d2 = self.factory.create_dashboard()
+        self.factory.create_dashboard(is_archived=True)
+        # Accessible as /dashboard/{slug}
         rv = self.make_request("get", "/api/dashboards/{0}?legacy".format(d1.slug))
         self.assertEqual(rv.status_code, 200)
 
-        expected = serialize_dashboard(d1, with_widgets=True, with_favorite_state=False)
+        expected = serialize_dashboard(d2, with_widgets=True, with_favorite_state=False)
         actual = json_loads(rv.data)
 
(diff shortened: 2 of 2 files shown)


justinclift:
Cool, this looks useful. :smile:

@getredash/maintainers Anyone up for reviewing this?  It's fairly simple Python backend code.

arikfr (commented):
The /{slug} URL is a legacy thing from the time before we switched to /{id}-{slug} (which is btw just /{id} as the slug is ignored).

I have a vague memory of having a reason of deprecating the slug one, but not sure what it was. Maybe the fact that slugs should update when the dashboard name updates, which in turn makes the URL unstable?

Anyway, I'm not sure we maintain uniqueness of slug anymore which might be an issue if we want to load by slug again.

eradman:
I see, so I took a route that exists but was supposed to be deprecated.

Slugs are named sequentially, in the order they were created, so it is difficult to reference them by exact name

```
redash=# select slug from dashboards order by slug;
...
 extension-analytics
 extension-analytics_1
 extension-analytics_2
```

This makes the legacy `/dashboard/` route is useless because you cannot guess the exact slug name. Should we remove the legacy route and add a new one?

```
/dashboards/find/{slug}
```

arikfr:
@eradman but why do you want to load them by slug and not id?

eradman:
For context, my general workflow is this:

1. create a dashboard on a dev environtment
2. export the dashboard (add json defenition + sql files to git)
3. publish dashboard to S3
4. user installs dashboard from HTTP location

If there is a single instance of Redash then the dashboard ID can be hard coded, but this is never the case in the evironments I work in since I don't have access to our customer sites.

I don't even have access to the Redash instances that my colleges are using, so there is no case where a hard-coded ID would work.

justinclift:
Interesting.  That sounds like a pretty reasonable scenario @eradman. :smile:

eradman:
@arikfr do you think it would be better to add a new route (redirect) to find a dashboard by name, or is it okay to repurpose the legacy route`/dashboard/{slug}`?

arikfr:
It's probably better to add a more explicit route, just to make things less confusing. Also, maybe we add the option to set the slug via the API (along with the lookup by slug) so things are more predictable and stable?

eradman:
Dashboard slugs are strange because the exact name is not determanistic:

```
def generate_slug(ctx):
    slug = utils.slugify(ctx.current_parameters["name"])
    tries = 1
    while Dashboard.query.filter(Dashboard.slug == slug).first() is not None:
        slug = utils.slugify(ctx.current_parameters["name"]) + "_" + str(tries)
        tries += 1
    return slug
```

Some ideas

1. Allow the user to define a custom slug
2. Add a redirect-route for searching by dashboard name
3. Erase slug when a dashboard is archived and make the slug unique

guidopetri:
I'd find the dashboards-by-slug stuff useful. I'd be in favor of making slugs unique period - I don't think we should delete them when archived, since if someone archives dashboard 1, then creates an identically-slugged dashboard 2, and unarchives dashboard 1, we'd have a conflict. I can't really think of a use case where you'd want two identically-named dashboards.

guidopetri:
#4932 seems like an issue that wants a similar outcome, but for queries.

eradman:
Closing because this change is too narrow: we need a generic way of linking to dashboards and queries, and visualizations by name.

Issue #6381, opened 2023-08-21
Query Search is broken


denisov-vlad:
### Issue Summary

After upgrading to Redash with Python 3.8, search cannot be used.

### Steps to Reproduce

1. Open `/queries` page
2. Try to search something
3. See "Internal Server Error"

### Technical details:

* Redash Version: 4a5c9c2 
* Browser/OS: Safari / Mac Os
* How did you install Redash: k8s

Here are the logs from webserver pod:

```
[2023-08-21 09:11:29,965][PID:1784][ERROR][redash.app] Exception on /api/queries [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.UndefinedFunction: function tsq_parse(unknown, unknown) does not exist
LINE 6: ...up_id IN (2, 1, 8)) AND (queries.search_vector @@ tsq_parse(...
                                                             ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/usr/local/lib/python3.8/site-packages/flask_restful/__init__.py", line 489, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/flask_login/utils.py", line 277, in decorated_view
    return current_app.ensure_sync(func)(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/flask/views.py", line 109, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)
  File "/app/redash/handlers/base.py", line 31, in dispatch_request
    return super(BaseResource, self).dispatch_request(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/flask_restful/__init__.py", line 604, in dispatch_request
    resp = meth(*args, **kwargs)
[...]

PR #6448, opened 2023-09-12
Update SQLAlchemy-Utils to 0.36.5

eradman:
## What type of PR is this? 

- [x] Upgrade

## Description

This makes the Python unit tests compatible with Python 3.10 while maintaining compatibility with 3.8.

## How is this tested?

- [x] Unit tests (pytest, jest)

PR #6491, opened 2023-09-30
Create sqlalchemy_searchable expressions

azundo:
The required sql expressions for sqlalchemy_searchable are only created on the sqlalchemy pre_create hook. When upgrading from a previous redash version, there may be no CREATE statements in any applicable migrations, so these sql expressions are never created.

In a fresh install the hook runs properly so there's no issue with the search functionality.

Add a migration to manually execute the sql_expressions DDL from sqlalchemy_searchable.

## What type of PR is this? 

- [x] Bug Fix

## Description

The required sql expressions for sqlalchemy_searchable are only created on the sqlalchemy pre_create hook. When upgrading from a previous redash version no CREATE statements may be run, so these expressions are never created.

## How is this tested?

- [x] Manually

1. Created a fresh install of redash from the current master branch and deleted the custom sql expressions created by sqlalchemy_searchable. This mimics the state of an installation that was upgraded from a version prior to the version bump of sqlalchemy_searchable.
2. Confirmed that search is broken.
3. Deployed fix and ran migrations.
4. Confirmed search is fixed.

Fixes https://github.com/getredash/redash/issues/6381

### DIFF
--- a/migrations/versions/7ce5925f832b_create_sqlalchemy_searchable_expressions.py
+++ b/migrations/versions/7ce5925f832b_create_sqlalchemy_searchable_expressions.py
@@ -1,0 +1,25 @@
+"""create sqlalchemy_searchable expressions
+
+Revision ID: 7ce5925f832b
+Revises: 1038c2174f5d
+Create Date: 2023-09-29 16:48:29.517762
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy_searchable import sql_expressions
+
+
+# revision identifiers, used by Alembic.
+revision = '7ce5925f832b'
+down_revision = '1038c2174f5d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(sql_expressions)
+
+
+def downgrade():
+    pass


justinclift:
Thanks @azundo, this looks useful.  I'm trying to solve a different problem this weekend (Trino related), but this shouldn't be too hard to test too anyway.

eradman:
@azundo do you think this would also solve https://github.com/getredash/redash/pull/6448?

When we tried switching to SQLAlchemy-Utils to 0.36.5 we hit errors when upgrading an existing install.

azundo:
@justinclift I was finally able to manually test it so I have at least run the migration code and confirmed it works as expected, but probably worth having someone else confirm as well. It's a bit awkward to reproduce the exact scenario of installing an older version and then upgrading. I simulated this instead by deleting all the SQL expressions created by sqlalchemy searchable so search would fail again, and then running the redash migration to confirm the fix.

@eradman that issue doesn't look related to me at first glance

justinclift:
Sorry for the delay in reviewing this.  I have this on my ToDo list for the next few days though.  It's not forgotten. :smile:

azundo:
No worries at all! The workaround I shared in the bug report is an easy fix for anyone running into this so I wouldn't say it's urgent.

denisov-vlad:
Hi!

@azundo @justinclift I've tested it both on my dev server and production cluster, and everything works fine.

@eradman Also tested updating to `SQLAlchemy-Utils==0.36.5`, the same error as before. :(

Krikooo:
I tested this PR, it's work just like a charm.
Thank you for your work @azundo 👍

justinclift:
Awesome.  Thanks for verifying it @Krikooo.  Merging it now. :smile:

justinclift (approved):

pinophyta:
What's the proper way to run this migration after upgrading redash?

eradman:
@pinophyta try
```
docker compose run --rm server manage db upgrade
```

or if not using docker
```
$REDASH_DIR/bin/manage db upgrade
```

pinophyta:
Thank you very much @eradman. That worked and I was able to upgrade to latest dev tag 🥳

PR #6492, opened 2023-10-02
Remove third-party binaries from Docker build

eradman:
## What type of PR is this? 

- [x] Cleanup

## Description

x86_64 binaries pulled from Microsoft and Databriks not used by tests.

This can be considered a optimization for tests since it reduces the time to build and makes updating the base image much easier.

## How is this tested?

- [x] E2E Tests (Cypress)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,24 +60,6 @@ RUN apt-get update && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
-
-ARG TARGETPLATFORM
-ARG databricks_odbc_driver_url=https://databricks-bi-artifacts.s3.us-east-2.amazonaws.com/simbaspark-drivers/odbc/2.6.26/SimbaSparkODBC-2.6.26.1045-Debian-64bit.zip
-RUN if [ "$TARGETPLATFORM" = "linux/amd64" ]; then \
-  curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add - \
-  && curl https://packages.microsoft.com/config/debian/10/prod.list > /etc/apt/sources.list.d/mssql-release.list \
-  && apt-get update \
-  && ACCEPT_EULA=Y apt-get install  -y --no-install-recommends msodbcsql17 \
-  && apt-get clean \
-  && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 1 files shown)


eradman:
This PR was created so that I could see if the build pipeline depends on these.

Everyone, feel free to treat this as a discussion topic

Dockerfile
```diff
@@ -60,24 +60,6 @@ RUN apt-get update && \
...
   rm -rf /var/lib/apt/lists/*
 
-
-ARG TARGETPLATFORM
-ARG databricks_odbc_driver_url=https://databricks-bi-artifacts.s3.us-east-2.amazonaws.com/simbaspark-drivers/odbc/2.6.26/SimbaSparkODBC-2.6.26.1045-Debian-64bit.zip
-RUN if [ "$TARGETPLATFORM" = "linux/amd64" ]; then \
-  curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add - \
-  && curl https://packages.microsoft.com/config/debian/10/prod.list > /etc/apt/sources.list.d/mssql-release.list \
-  && apt-get update \
-  && ACCEPT_EULA=Y apt-get install  -y --no-install-recommends msodbcsql17 \
```
justinclift, line 70: This approach seems to remove the MS (ODBC) SQL Server driver altogether.

Surely there's a better approach? :smile:

justinclift: Interestingly, it looks like MS only supplied x86_64 (amd64) versions of the MS SQL Server (ODBC) packages:

* https://packages.microsoft.com/debian/10/prod/pool/main/m/msodbcsql17/
* https://packages.microsoft.com/debian/10/prod/pool/main/m/msodbcsql18/

Note everything there is `amd64`, with no arm packages of any sort in sight.

Other packages (especially moby ones) do seem to have arm64 builds available, though that's not useful for us:

https://packages.microsoft.com/debian/10/prod/pool/main/m/moby-engine/

eradman: Ah, you're right. The library is loaded dynamically, so the error only occurs at runtime

> Connection Test Failed:
> [01000] [unixODBC][Driver Manager]Can't open lib 'ODBC Driver 17 for SQL Server' : file not found (0) (SQLDriverConnect)

eradman:
Until we have a more modular approach to data sources it appears we have to keep the mssql binaries

PR #5067, opened 2020-07-27
add pagination and base_url to JSON query_runner

Ken-Michalak:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Query Runner (Data Source)
- [ ] New Alert Destination
- [ ] Other

## Description

Add the ability to paginate the JSON Query Runner and configure a base URL. This supports the `next_page_token`, `page_token` pattern or `next_offset`, `offset`. It's written to support pagination in Chargebee's API, but generalized for other sources, including [Google's pagination design pattern](https://cloud.google.com/apis/design/design_patterns#list_pagination), by passing `pagination: [next_offset, offset]` into the query.

PR #6499, opened 2023-10-04
add pagination and base_url to JSON query_runner

Ken-Michalak:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Add the ability to paginate the JSON Query Runner and configure a base URL.
Support [Google's pagination design pattern](https://cloud.google.com/apis/design/design_patterns#list_pagination), with `pagination: [next_page_token, page_token]`.
Support a next URL, which is sometimes used in HATEOAS, with `pagination: _links.next.href`.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

Original PR: https://github.com/getredash/redash/pull/5067

### DIFF
--- a/redash/query_runner/json_ds.py
+++ b/redash/query_runner/json_ds.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+from urllib.parse import urljoin
 
 import yaml
 from funcy import compact, project
@@ -61,7 +62,7 @@ def add_column(columns, column_name, column_type):
         columns.append({"name": column_name, "friendly_name": column_name, "type": column_type})
 
 
-def _apply_path_search(response, path):
+def _apply_path_search(response, path, default=None):
     if path is None:
         return response
 
@@ -71,6 +72,8 @@ def _apply_path_search(response, path):
         current_path = path_parts.pop()
         if current_path in response:
             response = response[current_path]
+        elif default is not None:
+            return default
         else:
             raise Exception("Couldn't find path {} in response.".format(path))
 
@@ -78,6 +81,8 @@ def _apply_path_search(response, path):
 
 
 def _normalize_json(data, path):
+    if not data:
+        return None
     data = _apply_path_search(data, path)
 
     if isinstance(data, dict):
@@ -94,9 +99,7 @@ def _sort_columns_with_fields(columns, fields):
 
 
 # TODO: merge the logic here with the one in MongoDB's queyr runner
-def parse_json(data, path, fields):
-    data = _normalize_json(data, path)
-
+def parse_json(data, fields):
     rows = []
     columns = []
 
@@ -130,17 +133,19 @@ def parse_json(data, path, fields):
 
 class JSON(BaseHTTPQueryRunner):
     requires_url = False
+    base_url_title = "Base URL"
 
     @classmethod
     def configuration_schema(cls):
         return {
             "type": "object",
             "properties": {
+                "base_url": {"type": "string", "title": cls.base_url_title},
                 "username": {"type": "string", "title": cls.username_title},
                 "password": {"type": "string", "title": cls.password_title},
             },
             "secret": ["password"],
-            "order": ["username", "password"],
+            "order": ["base_url", "username", "password"],
         }
 
     def __init__(self, configuration):
@@ -153,6 +158,16 @@ class JSON(BaseHTTPQueryRunner):
     def run_query(self, query, user):
         query = parse_query(query)
 
+        results, error = self._run_json_query(query)
+        if error is not None:
+            return None, error
+
+        data = json_dumps(results)
+        if data:
+            return data, None
+        return None, "Got empty response from '{}'.".format(query["url"])
+
+    def _run_json_query(self, query):
         if not isinstance(query, dict):
             raise QueryParseError("Query should be a YAML object describing the URL to query.")
 
@@ -165,13 +180,15 @@ class JSON(BaseHTTPQueryRunner):
         fields = query.get("fields")
         path = query.get("path")
 
+        if "pagination" in query:
+            pagination = RequestPagination.from_config(self.configuration, query["pagination"])
+        else:
+            pagination = None
+
         if isinstance(request_options.get("auth", None), list):
             request_options["auth"] = tuple(request_options["auth"])
         elif self.configuration.get("username") or self.configuration.get("password"):
-            request_options["auth"] = (
-                self.configuration.get("username"),
-                self.configuration.get("password"),
-            )
+            request_options["auth"] = (self.configuration.get("username"), self.configuration.get("password"))
 
         if method not in ("get", "post"):
             raise QueryParseError("Only GET or POST methods are allowed.")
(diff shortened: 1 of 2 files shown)


Ken-Michalak:
@justinclift Here's the new pr.

I'm not sure if this is the best way to configure it, especially if someone wants to support other patterns later, like page numbers or offsets without anything in response for what to pass next. I was wondering if there were other ideas there or if there's something like that on another data source.

justinclift:
@guidopetri @eradman Any interest in reviewing this? :smile:

justinclift:
Ahhh, looks like this has a formatting issue that needs fixing:

```
Run ruff check .
redash/query_runner/json_ds.py:1:1: I001 [*] Import block is un-sorted or un-formatted
```

guidopetri:
Could we also add a test to make it clearer what is going on, and what the expected behavior is?

Ken-Michalak:
@guidopetri I restructured the configuration, which hopefully makes a bit more sense.

For a config like this:
```yaml
pagination:
  type: url
  #path: _links.next.href
```
As long as the API response has something in `_links.next.href`, it will follow it, do another request, and include those results.

For the other pattern:
```yaml
pagination:
    type: token
    #fields: [next_page_token, page_token]
```
It does the same kind of thing, where as long as `next_page_token` isn't `""` or `null`, it continues getting the next page of results by putting that token in the `page_token` parameter.

I looked into adding a test, but it doesn't look like there are any tests for the json data source at all yet, and this would be pretty complicated to mock. I was just testing it manually by querying a [Chargebee API](https://apidocs.chargebee.com/docs/api/list-ops?lang=curl) and a [Vimeo VHX API](https://dev.vhx.tv/docs/api/#product-list).

redash/query_runner/json_ds.py
```diff
@@ -164,32 +167,80 @@ def run_query(self, query, user):
...
+        url = urljoin(base_url, url)
+
+        results = []
+        has_more = True
+        while has_more:
+            response, error = self._get_json_response(url, method, **request_options)
+            has_more = False
+
+            result = _normalize_json(response, result_path)
+            if result:
```
guidopetri, line 207: Having this many nested `if` statements is kind of a code smell. Maybe we should split this up into multiple functions?

redash/query_runner/json_ds.py
```diff
@@ -164,32 +167,80 @@ def run_query(self, query, user):
...
+                results.extend(result)
+
+                if pagination["type"] == "url":
```
guidopetri, line 210: I've never seen either of these two pagination patterns. I don't think there's a "standard" either, is there? I imagine every API has its own set of pagination rules that it follows. If that's the case, I'm not sure it makes sense to write in pagination rules, and rather we should emphasize to the end users that querying an API for JSONs directly like this is meant for testing and not large-scale analysis.

Especially since if too much data is returned, it's possible that the worker goes OOM. 😅

redash/query_runner/json_ds.py
```diff
@@ -71,6 +72,8 @@ def _apply_path_search(response, path):
         current_path = path_parts.pop()
         if current_path in response:
             response = response[current_path]
+        elif default is not None:
+            return default
```
guidopetri, line 76: Won't this return on the first instance of `current_path not in response`? I am not sure having the return within the `while` loop is what we want, but rather we should rewrite this so that there's only one return statement in the function akin to `return response or default` , or something similar.

guidopetri (changes requested):
I very much don't feel comfortable merging this in without a test. To be honest I'm not sold on the idea of implementing pagination here in the first place. As mentioned in the comments below, I'm not sure there's a standard for pagination, and if we were to implement every pagination ruleset out there we'd have a very large codebase 😅 but if we do want to merge this in I would very kindly request that we somehow add tests, even if we have to write some from scratch for the JSON query runner to start out with.

redash/query_runner/json_ds.py
Ken-Michalak: Some variant of these 2 seem to be the most common, because they don't require specifics of implementation, like having to add to the page number or offset yourself. (The `type` could be named something different if you have a better idea though.) That's why Google outlines the next_page_token/page_token as a best practice. It's just that sometimes the names are different, like Chargebee uses an offset parameter, but rather than having to calculate it, it returns the next offset in the response. Google Analytics, AWS, and K8s also follow that pattern.

Shopify, SalesForce, Facebook's Graph API, and Trino all pass a full URL for the next page in some field of the response, which would be covered by the `url` type. HATEOAS is a standard and does define that the next page would be a link somewhere in the response. Although, at some point, it might be good to use a library instead of `_apply_path_search` to handle more complex cases.

Other types could be added later for `pageNumber` or something, but I don't really have a need for them right now, and that is where it could be pretty difficult to cover different APIs.

I was wondering about the memory thing, but it doesn't look like there's any other way that large result sets are handled elsewhere. The Trino data source does a `fetchall` and then processes all of the results at once. This at least does the `_normalize_json` per page, which could trim things down.

Ken-Michalak:
Odd, I didn't see any linting errors, but yeah, black still changed some things.

redash/query_runner/json_ds.py
guidopetri: In my opinion, yes - definitely worth it if it makes the code easier to follow.

redash/query_runner/json_ds.py
guidopetri: Fair points; thanks for elaborating.

tests/query_runner/test_json_ds.py
```diff
@@ -0,0 +1,88 @@
...
+            },
+            "page": {"size": 2, "totalElements": 3, "totalPages": 2},
+        }
+    else:
+        error = "404: {} not found".format(url)
+
+    return data, error
+
+
+class TestJSON(TestCase):
```
guidopetri, line 53: Ah, sorry, I should have specified. We've been aiming for pytest tests instead of unittest; if it's not too much trouble, would you mind converting to those? I won't hold it against you if you're not feeling up to it though. I appreciate the work you've put into this PR and already recognize you've gone above and beyond :)

guidopetri (approved):
Approving, but we'll need the PR to be updated with the latest `master` to merge. I'll enable automerge so whenever you push a merge commit updating the branch, it should merge automatically once the tests pass.

Thanks again for all your work @Ken-Michalak !

tests/query_runner/test_json_ds.py
Ken-Michalak: I just used the other tests as an example. It could always be refactored later. It seems like some of that should be added to the contributing guide, like the testing, ruff, and rebasing process. Like are you looking for a single commit per pr, or just going for a linear history?

guidopetri: That's a great point. I don't think we have a good handle on what our contribution guidelines are; I'll try and add some documentation for that to the wiki soon.

guidopetri (approved):

justinclift:
@Ken-Michalak We have an option turned on for this repo, which requires PR's to be uo-to-date with the master branch before they can be merged.

Normally PRs have a button to "Update to the master branch" (or similar wording), that maintainers can click to automate that.

It seems to be missing from this PR though.  My guess is the PR was created without the "Allow maintainers to update this PR?" setting enabled.

So... I think that means you'll need to update this PR yourself.  That might let it all just get nicely merged automatically after the tests pass. :smile:

guidopetri:
I bypassed it - we didn't touch the poetry files so it should be ok.

Thanks for your contribution @Ken-Michalak !

Ken-Michalak:
> @Ken-Michalak We have an option turned on for this repo, which requires PR's to be uo-to-date with the master branch before they can be merged.
> 
> Normally PRs have a button to "Update to the master branch" (or similar wording), that maintainers can click to automate that.
> 
> It seems to be missing from this PR though. My guess is the PR was created without the "Allow maintainers to update this PR?" setting enabled.
> 
> So... I think that means you'll need to update this PR yourself. That might let it all just get nicely merged automatically after the tests pass. 😄

@justinclift, I actually did that in that last force-push yesterday, but it looks like before @guidopetri approved it again, there were a few other commits added. (Bump @babel/traverse from 7.22.8 to 7.23.2) 🤷🏻 Thanks though.

justinclift:
No worries, it all worked out.  Thanks heaps for getting this done @Ken-Michalak, it seems like this will be useful for people. :smile:

PR #6504, opened 2023-10-07
convert docker-compose to docker compose

guidopetri:
## What type of PR is this? 

- [x] Other

## Description

Docker now has `compose` support within it, and installing the latest version of `docker` allows for `docker compose`. The old [docker-compose syntax is being deprecated](https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose). This change moves us from the old `docker-compose` syntax to the new `docker compose` syntax.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Via PR checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,15 +35,15 @@ jobs:
       - name: Build Docker Images
         run: |
           set -x
-          docker-compose build --build-arg install_groups="main,all_ds,dev" --build-arg skip_frontend_build=true
-          docker-compose up -d
+          docker compose build --build-arg install_groups="main,all_ds,dev" --build-arg skip_frontend_build=true
+          docker compose up -d
           sleep 10
       - name: Create Test Database
-        run: docker-compose -p redash run --rm postgres psql -h postgres -U postgres -c "create database tests;"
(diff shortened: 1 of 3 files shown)


justinclift:
What version of Docker are you using that works with this?

Asking so we have some idea of a known-good configuration. :smile:

guidopetri:
I'm on docker 24. If you follow the current install instructions, it seems to just get the latest version: https://docs.docker.com/engine/install/debian/#install-using-the-repository

justinclift:
From eyeballing the change, it *looks* fairly straight forward.

I've not personally used the v2 docker compose commands yet though, so no practical experience of potential gotchas we should watch out for.

I guess we can merge it, and if something goes badly wrong then we can revert the change (or fix whatever the problem is).

justinclift (approved):

eradman:
@guidopetri can youupdate the dev environment setup instructions to match this change?

https://github.com/getredash/redash/wiki/Local-development-setup#set-up-the-prerequisites

guidopetri:
Ah, yes! Good call.

guidopetri:
@eradman thanks again for updating those docs for me 😅

PR #6505, opened 2023-10-07
Add column type to a query result of InfluxDB

masayuki038:
## What type of PR is this? 

- [x] Bug Fix

## Description
Bug fix for #6179 

I confirmed that it works fine to download these files (below) in a query page of Redash by manual test:
- csv
- tsv
- excel

## How is this tested?

- [x] Unit tests (pytest, jest)
- [x] Manually

## Related Tickets & Documents
Closes #6179

PR #6503, opened 2023-10-07
Fix dockerhub image building

guidopetri:
## What type of PR is this? 

- [x] Bug Fix

## Description

Our build-image-and-push step in the CI wasn't running. This change makes it work again :) the comparison operator for the previous step's output had to be outside the jinja placeholder clause.

## How is this tested?

- [x] Manually

I changed the `on:` line on the CI file and pushed the branch. [This GHA run](https://github.com/getredash/redash/actions/runs/6438205061) demonstrates that the build-and-push started executing, though I later canceled it.

PR #6420, opened 2023-09-03
Generate docker hub preview image

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR updates our CI to automatically generate Docker Hub preview images whenever it runs on our `master` branch.

It needs the appropriate Docker Hub login credentials set in our GitHub settings before it will work properly.

  * `DOCKER_USER` environment variable
  * `DOCKER_PASS` secret

## How is this tested?

- [x] Unit tests (pytest, jest)
- [x] E2E Tests (Cypress)

PR #6507, opened 2023-10-07
Build + push `preview` image only on `master` branch

guidopetri:
## What type of PR is this? 

- [x] Bug Fix
- [x] Other

## Description

The CI is trying to build + push the `preview` tagged image on PRs, but we only want that to happen on the `master` branch. This change adds a condition to the CI step that runs the build+push, filtering for the ref to be `master` as a condition of the CI step being run.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Added a test commit that skips the e2e/unit tests (for faster iteration) and verified that the build+push CI step does not run on this PR.

#6503 , #6505 , #6420

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,9 +159,18 @@ jobs:
       - name: Skip?
         id: skip-check
         run: |
-          if [[ "${{ vars.DOCKER_USER }}" == '' ]] || [[ "${{ secrets.DOCKER_PASS }}" == '' ]]; then
+          if [[ "${{ vars.DOCKER_USER }}" == '' ]]; then
+            echo 'Docker user is empty. Skipping build+push'
+            echo skip=true >> "$GITHUB_OUTPUT"
+          elif [[ "${{ secrets.DOCKER_PASS }}" == '' ]]; then
+            echo 'Docker password is empty. Skipping build+push'
+            echo skip=true >> "$GITHUB_OUTPUT"
+          elif [[ "${{ github.ref_name }}" != 'master' ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -172,7 +181,7 @@ jobs:
       - frontend-unit-tests
       - frontend-e2e-tests
       - build-skip-check
-    if: ${{ needs.build-skip-check.outputs.skip }} == false
+    if: needs.build-skip-check.outputs.skip == 'false'
```
guidopetri, line 184: Turns out we don't need the `{{ }}` jinja braces, and this is now actually filtering whereas it was not before.

guidopetri:
@justinclift when you have a minute, mind giving this a look?

.github/workflows/ci.yml
```diff
@@ -159,9 +159,18 @@ jobs:
       - name: Skip?
         id: skip-check
         run: |
-          if [[ "${{ vars.DOCKER_USER }}" == '' ]] || [[ "${{ secrets.DOCKER_PASS }}" == '' ]]; then
+          if [[ "${{ vars.DOCKER_USER }}" == '' ]]; then
+            echo 'Docker user is empty. Skipping build+push'
+            echo skip=true >> "$GITHUB_OUTPUT"
+          elif [[ "${{ secrets.DOCKER_PASS }}" == '' ]]; then
```
justinclift, line 165: Good idea.  The extra messages will help us figure out what's going wrong, when something is not working right. :smile:

justinclift (approved):
Looks good to me. :smile:

Issue #6409, opened 2023-08-31
Google Spreadsheets worksheet by title cleanup/refactor


guidopetri:
In #5334 , I left a few comments about future work on the "fetch worksheet by title" functionality for the google spreadsheets query runner. We should address those, namely:
- don't use mixed-type variables
- allow for non-quoted worksheet titles (single word)
- verify 0/1 indexing of worksheet numbers, and add comment if it differs

guidopetri:
@TanyaKansal absolutely! Feel free to submit a PR and tag me in it when it's ready for review :)

immortalcodes:
@guidopetri  I didn't quite understand the 3rd point , btw for other 2 please have a look https://github.com/getredash/redash/compare/master...immortalcodes:redash:google-sheet-singleword-title

immortalcodes:
@guidopetri  I didn't quite understand the 3rd point , btw for other 2 please have a look https://github.com/getredash/redash/compare/master...immortalcodes:redash:google-sheet-singleword-title

justinclift:
Note - removed duplicated comment above. :wink:

DhanushA1307:
Hi. I am new and would like to contribute to this issue @guidopetri

PR #6510, opened 2023-10-09
Update google_spreadsheets.py

immortalcodes:
## What type of PR is this? 

This fixes and works on the issue #6409 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/redash/query_runner/google_spreadsheets.py
+++ b/redash/query_runner/google_spreadsheets.py
@@ -100,18 +100,28 @@ class WorksheetNotFoundByTitleError(Exception):
 def parse_query(query):
     values = query.split("|")
     key = values[0]  # key of the spreadsheet
-    worksheet_num_or_title = 0  # A default value for when a number of inputs is invalid
+    worksheet_num = 0  # A default value for when a number of inputs is invalid
+    worksheet_title = '' # No title initially
     if len(values) == 2:
         s = values[1].strip()
         if len(s) > 0:
             if re.match(r"^\"(.*?)\"$", s):
                 # A string quoted by " means a title of worksheet
-                worksheet_num_or_title = s[1:-1]
-            else:
+                worksheet_title = s[1:-1]
+            elif re.match(r"^\d+$", s):
                 # if spreadsheet contains more than one worksheet - this is the number of it
-                worksheet_num_or_title = int(s)
-
-    return key, worksheet_num_or_title
+                worksheet_num = int(s)
+            else:
+                # To match for non-quoted one-word file name
+                if len(s.split()) == 1 :
+                    worksheet_title = s
+            if worksheet_title != '':
+                return key, worksheet_title
+            else:
+                return key, worksheet_num
+                    
+        else:
+            return key, worksheet_num
 
 
 def parse_worksheet(worksheet):


redash/query_runner/google_spreadsheets.py
```diff
@@ -100,18 +100,28 @@ def __init__(self, worksheet_title):
...
-    return key, worksheet_num_or_title
+                worksheet_num = int(s)
+            else:
+                # To match for non-quoted one-word file name
+                if len(s.split()) == 1 :
+                    worksheet_title = s
+            if worksheet_title != '':
+                return key, worksheet_title
+            else:
+                return key, worksheet_num
```
guidopetri, line 121: The function returning a string or number for the second element of the return tuple is a code smell - we should see if we can separate this out into two functions that handle each case separately, or have both the string and integer handling happen via integers only (or strings only).

redash/query_runner/google_spreadsheets.py
```diff
@@ -100,18 +100,28 @@ def __init__(self, worksheet_title):
...
-                worksheet_num_or_title = int(s)
-
-    return key, worksheet_num_or_title
+                worksheet_num = int(s)
+            else:
+                # To match for non-quoted one-word file name
+                if len(s.split()) == 1 :
+                    worksheet_title = s
+            if worksheet_title != '':
+                return key, worksheet_title
```
guidopetri, line 119: Too many nested if-else clauses is also a code smell - there's likely some way to simplify this by breaking down into component pieces so that we don't have a lot of nesting.

redash/query_runner/google_spreadsheets.py
```diff
@@ -100,18 +100,28 @@ def __init__(self, worksheet_title):
 def parse_query(query):
     values = query.split("|")
     key = values[0]  # key of the spreadsheet
-    worksheet_num_or_title = 0  # A default value for when a number of inputs is invalid
+    worksheet_num = 0  # A default value for when a number of inputs is invalid
```
guidopetri, line 103: This is what the third bullet point in #6409 is referring to - are these sheets 0-indexed or 1-indexed? We should verify this.

redash/query_runner/google_spreadsheets.py
guidopetri: Also, by simplifying this we can also fix `parse_worksheet` to be simpler too.

guidopetri (changes requested):
@immortalcodes thanks for the PR! I left a few comments - the refactor of this function isn't as straightforward as it looks. We should strive to simplify this function and make it more understandable.

redash/query_runner/google_spreadsheets.py
immortalcodes: Oh I get it now , as far as I know , google sheets start from 1 and then 2,3,... so on , so shall we have default changed to 1

immortalcodes:
I have used a delimiter to have string and title both as a string and still be able to differentiate between them

immortalcodes:
I guess this might require changing the unit tests as well

redash/query_runner/google_spreadsheets.py
```diff
@@ -100,18 +100,20 @@ def __init__(self, worksheet_title):
 def parse_query(query):
     values = query.split("|")
     key = values[0]  # key of the spreadsheet
-    worksheet_num_or_title = 0  # A default value for when a number of inputs is invalid
+    worksheet_pointer = ' 1'  # A default value for when a number of inputs is invalid
```
guidopetri, line 103: @immortalcodes hey, thanks for updating this PR. I think I might have misspoken though, or I wasn't clear in my communication - we want to separate `parse_query` into two versions of a similar function, where one uses a worksheet number and the other uses the title, and we want to call the correct function depending on what the user inputs.

guidopetri: Ideally, we don't have to worry about things like a space prefix on a string signifying anything; we should keep the data as separate as possible from the code logic.

guidopetri:
@immortalcodes hey, just checking in. Any interest in finishing this PR still?

immortalcodes:
> @immortalcodes hey, just checking in. Any interest in finishing this PR still?

Yup, was busy with some work

PR #6513, opened 2023-10-09
Commit version update and tag together

eradman:
Also update to checkout v4

## What type of PR is this? 

- [x] Other

## Description

This is an effort to make this action more reliable, and possibly solve this error

> error: failed to push some refs to 'https://github.com/getredash/redash'
> hint: Updates were rejected because the remote contains work that you do not
> hint: have locally. This is usually caused by another repository pushing to
> hint: the same ref. If you want to integrate the remote changes, use

## How is this tested?

- [x] Manually

Create private repo, manually trigger after setting `on: workflow_dispatch`

### DIFF
--- a/.github/workflows/periodic-snapshot.yml
+++ b/.github/workflows/periodic-snapshot.yml
@@ -12,7 +12,7 @@ jobs:
   bump-version-and-tag:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - run: |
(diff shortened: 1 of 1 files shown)


justinclift (approved):
Lets try it out. :smile:

eradman:
Ups, I tried adding `[skip ci]` to see if that would save time running checks since they have nothing to do with `.github/workflows/`

Do we have a way of merging this without running the tests?

justinclift:
> Do we have a way of merging this without running the tests?

I can probably force merge it, so we can see if it works.

PR #6519, opened 2023-10-13
Allow Query.options to be None

eradman:
Query.options may not have a value on a database created by 10.1.0

## What type of PR is this? 

- [x] Bug Fix

## Description

This incompatibility was introduced in https://github.com/getredash/redash/pull/4797

## How is this tested?

- [x] Manually

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -264,7 +264,7 @@ class DataSource(BelongsToOrgMixin, db.Model):
 
     @property
     def uses_ssh_tunnel(self):
-        return "ssh_tunnel" in self.options
+        return self.options and "ssh_tunnel" in self.options
 
     @property
     def query_runner(self):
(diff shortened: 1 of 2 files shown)


justinclift:
> This incompatibility was introduced in https://github.com/getredash/redash/pull/5751

Is that the correct PR number?

Looking at that PR just now, it seems like it was closed without being merged. (?)

eradman:
Ah right, updated the PR

justinclift:
> Ah right, updated the PR

Thanks, that seems like the right one now. :smile:

justinclift:
@eradman If you reckon this is ready to merge, then I'm happy to do so. :smile:

eradman:
This seems good in my testing. With this patch I have succussfuly upgraded Redash with queries created by 10.1.0

justinclift (approved):
Cool, lets merge this then. :smile:

eradman:
Appologies, somehow I managed to update this PR with the fix for the counter widget as well (https://github.com/getredash/redash/issues/6521)

We can revert this, or leave it as is and confirm that this fixes the problem

PR #6538, opened 2023-10-18
Add new connection options in MySQL

adrianoapj:
## What type of PR is this? 

- [X] Other

## Description
This PR adds some new connection options to the MySQL query runner, I added this because when trying to connect to a PlanetScale instance, I got some errors and it required to use these extra options to work.

## How is this tested?

- [X] E2E Tests (Cypress)
- [X] Manually

I changed the query runner, so we had some small changes in the UI when adding a MySQL database which I tested and I also tried to connect to a PlanetScale instance using it and it worked. I also ran unit tests and E2E tests to ensure nothing broke, however not all tests with Cypress passed (at least locally using my MacBook Air M1), but I don't think it's related to the changes I made.

drawing


guidopetri (approved):

guidopetri:
Thanks for these changes @adrianoapj ! Hopefully the e2e tests pass on the CI.

PR #5767, opened 2022-05-31
Add Apache ignite Query Runner

sdarlington:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [X] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## Description

Add a query runner for Apache Ignite and GridGain, an in-memory database and cache. Tested against Apache Ignite 2.8.1+ and GridGain 8.8.4+.

## How is this tested?

- [ ] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [X] Manually
- [ ] N/A

Connected to and ran SQL queries again clusters with:

* Apache Ignite and GridGain
* Un-encrypted and with TLS encryption enabled
* With and without authentication

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

Screenshot 2022-05-31 at 14 02 19

PR #6546, opened 2023-10-23
Remove pygridgain so Trino data source shows up

justinclift:
## What type of PR is this? 

- [x] Bug Fix

## Description

This PR updates some dependency pieces so that Trino shows up in the data sources list again.

* Updates the python `tzlocal` dependency to the 4.x series, as seems to be required for Trino to work
* Removes the `pygridgain` dependency, as that has a hard dependency on `tzlocal` 2.1 and therefore breaks Trino

## How is this tested?

- [x] Manually

With a newly built docker container on my local pc, I confirmed that the Trino data source is not available without this PR, but *is* available after this PR is applied.

* https://github.com/getredash/redash/discussions/6483
* https://github.com/getredash/redash/pull/5767#issuecomment-1775067985

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3145,36 +3145,6 @@ files = [
 plugins = ["importlib-metadata"]
 
 [[package]]
-name = "pygridgain"
-version = "1.4.0"
-description = "GridGain binary client Python API"
-optional = false
-python-versions = ">=3.7"
-files = [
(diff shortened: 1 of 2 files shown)


guidopetri (approved):

guidopetri:
Thanks @justinclift :)

justinclift:
You're welcome, thanks for merging it @guidopetri. :smile:

PR #6552, opened 2023-10-24
Set height for .counter-visualization-container to 300px

eradman:
## What type of PR is this? 

- [x] Refactor

## Description

- Set font size back to px to match most other CSS
- Font calculation made based on counter-visualization-container, not an inner div
- Decrease font size by 30% if optional target value is displayed
- Scale counter fontSize using power(0.5)
- Set embedded counter height to 100%

## How is this tested?

- [x] Manually

https://github.com/getredash/redash/discussions/6542

redash-counters-2-line-3-line

### DIFF
--- a/client/app/assets/less/redash/query.less
+++ b/client/app/assets/less/redash/query.less
@@ -99,7 +99,10 @@ body.fixed-layout {
   > .query__vis {
     flex: 1 1 auto;
 
-    .chart-visualization-container, .visualization-renderer-wrapper, .visualization-renderer {
+    .visualization-renderer-wrapper, .visualization-renderer {
+      height: 100%
+    }
+    .chart-visualization-container, .counter-visualization-container {
       height: 100%
     }
   }
--- a/client/app/components/dashboards/dashboard-grid.less
+++ b/client/app/components/dashboards/dashboard-grid.less
@@ -123,7 +123,7 @@
       right: 10px;
       bottom: 15px;
       height: auto;
-      overflow: hidden;
+      overflow: visible;
       padding: 0;
     }
   }
--- a/viz-lib/src/visualizations/counter/Renderer.tsx
+++ b/viz-lib/src/visualizations/counter/Renderer.tsx
@@ -10,17 +10,18 @@ import "./render.less";
 
 function getCounterStyles(scale: any) {
   return {
-    msTransform: `scale(${scale})`,
-    MozTransform: `scale(${scale})`,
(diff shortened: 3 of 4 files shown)


justinclift:
@myonlylonely Are you ok to try this PR out? :smile:

myonlylonely:
The size in the embed query has been corrected.👍
In embed query:
Original:
image
After this PR:
image

@justinclift But in the dashboard, it is still smaller than before.

In dashboard:
Original:
image
After this PR:
image

justinclift:
Heh Heh Heh.  Seems like it's heading in the right direction at least. :wink:

eradman:
Yeah, this is promising. I added a second commit with an updated method of scaling the fontSize

myonlylonely:
I tried the new commit. The size in dashboard become bigger than the first commit, but still smaller than the original.
Original:
image
The new commit:
image

Also, I found that resize the embed query to a smaller size, the number is cut.
image

eradman:
>  I found that resize the embed query to a smaller size, the number is cut

Good catch, I fixed this. There are some special rules for embeded queries set by the `.query__vis` class 

>  The size in dashboard become bigger than the first commit, but still smaller than the original

This is not true. In Redash 10 the counter content scalled down with the content, so a few characters would appear large, and a long string would be small

eradman:
Forced-update to this branch to make the total diff easier to see

myonlylonely:
The embeded queries is now fixed! 👍

> > The size in dashboard become bigger than the first commit, but still smaller than the original
> 
> This is not true. In Redash 10 the counter content scalled down with the content, so a few characters would appear large, and a long string would be small

No, I think it is related to the "Visualization Name" label. In original Redash 10, if I delete the "Visualization Name" (empty string), the number will be scalled to a nice bigger size.

The screen in original Redash 10, with or without "Visualization Name":
image

justinclift:
Heh, sounds like we're making solid progress.

As an aside - wow, this has turned out to be way more complicated than it first looked eh? :smile:

eradman:
Indeed! This is much more involved than it seemed at first

eradman:
This change was updated again after having reverted "Render counter widgets using relative font size" (8b8dd4f68c8dc63ac15c14d270f4cf7b72c9eeee)

eradman:
I will apply this patch to an internal Redash build. Closing in favor of the discussion https://github.com/getredash/redash/discussions/6542

PR #6577, opened 2023-11-06
added multiarch build

AndrewChubatiuk:
## What type of PR is this? 
Build multiarch docker images

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Added ability to create multiarch images

## How is this tested?

- [ ] E2E Tests (Cypress)
- [] Manually
- [ ] N/A


guidopetri:
Hey @AndrewChubatiuk ! Thanks for the PR. We've tried a arm64 build on GHA recently and it runs super long - so long, in fact, that it hits some hidden 6 hour limit for GHA and does not finish. Would you mind testing this on your own repo and potentially making changes there before we consider merging this in?

guidopetri:
Here's an example action that we already ran: https://github.com/getredash/redash/actions/runs/6546702269

I'm not sure how that was running so maybe we just did it wrong. :)

Dockerfile
```diff
@@ -1,7 +1,5 @@
 FROM node:16.20.1-bookworm as frontend-builder
 
-RUN npm install --global --force yarn@1.22.19
```
guidopetri, line 3: Could you explain the changes you're making to the Dockerfile here?

Dockerfile
```diff
@@ -62,21 +60,21 @@ RUN apt-get update && \
 
 
 ARG TARGETPLATFORM
-ARG databricks_odbc_driver_url=https://databricks-bi-artifacts.s3.us-east-2.amazonaws.com/simbaspark-drivers/odbc/2.6.26/SimbaSparkODBC-2.6.26.1045-Debian-64bit.zip
-RUN if [ "$TARGETPLATFORM" = "linux/amd64" ]; then \
```
guidopetri, line 66: I don't think this odbc driver works on arm64, but I'll be happy to be proven wrong.

client/.babelrc
```diff
@@ -4,7 +4,7 @@
       "@babel/preset-env",
       {
         "exclude": ["@babel/plugin-transform-async-to-generator", "@babel/plugin-transform-arrow-functions"],
-        "corejs": "2",
+        "corejs": "3",
```
guidopetri, line 7: I don't understand any of the JS dependency related changes; @eradman would you mind taking a look? Are we able to change these deps to enable arm64 images?

Dockerfile
```diff
@@ -62,21 +35,21 @@ RUN apt-get update && \
...
+RUN \
   curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /usr/share/keyrings/microsoft-prod.gpg \
   && curl https://packages.microsoft.com/config/debian/12/prod.list > /etc/apt/sources.list.d/mssql-release.list \
   && apt-get update \
-  && ACCEPT_EULA=Y apt-get install  -y --no-install-recommends msodbcsql17 \
+  && ACCEPT_EULA=Y apt-get install  -y --no-install-recommends msodbcsql18 \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/* \
-  && curl "$databricks_odbc_driver_url" --location --output /tmp/simba_odbc.zip \
+  && curl "$DATABRICKS_ODBC_DRIVER_URL" --location --output /tmp/simba_odbc.zip \
```
eradman, line 46: This OBDC driver appears to be amd64 only

Does that conflict with removing ` if [ "$TARGETPLATFORM" = "linux/amd64" ]` above?

eradman:
This change seems to check out for me. Triggering CI tests

eradman:
The code coverage report is not relavent.

@guidopetri can you re-test a build with these changes on arm64?

.github/workflows/ci.yml
```diff
@@ -176,38 +191,49 @@ jobs:
...
+      - uses: actions/download-artifact@v3
         with:
-          node-version: ${{ env.NODE_VERSION }}
-          cache: 'yarn'
-      - name: Install Dependencies
-        run: |
-          npm install --global --force yarn@1.22.19
-          yarn cache clean && yarn --frozen-lockfile --network-concurrency 1
-
+          name: frontend
```
guidopetri, line 214: I'm not 100% clear where this `frontend` gets used in the build. Would you mind explaining? Does this also get pushed to Dockerhub as part of the `preview` image?

guidopetri: How do we build this locally, then? `docker build -t redash_test .` seems to not build the frontend

guidopetri (changes requested):
I'm unclear on what these changes are doing, and I think we need better documentation around this. How do we build things locally? What exactly is happening in the GHA build process? Why do we need to split up the frontend build?

guidopetri:
Hmm, alright. I think we should separate this into two PRs then:
- one enabling ARM64 builds (which may be slow if they're building the frontend);
- one separating the frontend build out, and making it clear how to build this locally. We should be able to run a `docker` command for this, and not depend on e.g. `yarn` being installed locally.

Does this make sense to you too? Or how do you think we should proceed?

AndrewChubatiuk:
merged alternative PR in https://github.com/getredash/redash/pull/6674

PR #6344, opened 2023-08-02
Switch from numeral to numbro

eradman:
numeraljs is no longer maintained, and incorrectly parses high-precision floats (such as 1.2e-7) as _NaN_.

## What type of PR is this? 

- [x] Bug Fix

## How is this tested?

- [x] viz-lib yarn tests
- [x] manually

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

**Current behavior using numeral.js**

redash_float_number_numbro

**New behavior using numbro.js**

redash_float_number_numeral

PR #6595, opened 2023-11-14
Revert "Switch from numeral to numbro (#6344)"

eradman:
This reverts commit f8934b83126261ae5639d35843518e7ceabaae87.

## What type of PR is this? 

- [x] Bug fix

## Description

Using a format string of `0` does not round to the nearest integer in Numbro 

[numbro/issues/745](https://github.com/BenjaminVanRyseghem/numbro/issues/745)


justinclift (approved):
Well damn.  Oh well, if you reckon this is the best approach for now, then I'm ok with it. :smile:

masayuki038:
Oh my god... But you realized it now. Nice catch !

justinclift:
Looking at the issue @eradman filed on the numbro repo, they might be able to suggest a workaround or fix the problem themselves.

So, we're probably best to leave this for a few days and see how that issue goes.  If nothing substantial turns up by then, we can merge this to revert back to numeral.

eradman:
Agreed; we should wait and see if there is a workaround we can use.

eradman:
A build from top of tree (392b930f2dde5f05033e459983b6cb8656672de7) shows that the previous behavior was restored

value_percent

Perhaps there is a delay in updating the docker image?

masayuki038:
@eradman Thanks for your confirmation!
It may be something wrong in my env. I will check it again. thank you.

masayuki038:
Indeed, the image created with `make compose_build` gave the expected value...
I will check `redash/redash:preview` later.

guidopetri:
@masayuki038 it might be a caching issue, where docker didn't pull the latest image.

masayuki038:
@guidopetri Thanks for your comment.
I ran `docker rmi redash/redash:preview`... I'll try it on another machine. thank you.

masayuki038:
Today, I tried `redash/redash:preview` again and it is fine. It seemed to be a cache issue. thank you.

PR #6598, opened 2023-11-14
Yandex.Disk Query runner

denisov-vlad:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Added query runner for Yandex.Disk, supports showing CSV, TSV, XLS(X) files

`get_schema` is used to show all available files

Example query:
```
path: /file_path.xlsx
sheet_name: 0  -- optional, used standard pandas syntax and allows both indexes and names
```

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A


justinclift:
Oh hang on.  It's failing on this:

> ModuleNotFoundError: No module named 'numpy'

Sounds like numpy might need to be added to the dependency list or similar?

denisov-vlad:
@justinclift looks strange, because numpy is the core dependency of pandas, and pandas is included in `all_ds`. What do you think about moving pandas+numpy to `main` deps?

justinclift:
@guidopetri Do you have time to look at this one?

konnectr:
@denisov-vlad 
func get_column_types_from_dataframe and pandas_to_result
I probably would have done it simpler and put the code in the utils folder, and not in __init__.py

I believe that there is no need to combine the dependencies of the main Redash and QR into one list. This can later lead to unnecessary dependencies

redash/query_runner/__init__.py
```diff
@@ -106,6 +109,37 @@ def find_last_keyword_idx(parsed_query):
...
+            redash_type = TYPE_INTEGER
+        elif column_type in (np.datetime64, np.dtype("<M8[ns]")):
+            if df.empty:
+                redash_type = TYPE_DATETIME
+            elif len(df[column_name].head(1).astype(str).loc[0]) > 10:
+                redash_type = TYPE_DATETIME
+            else:
+                redash_type = TYPE_DATE
+        else:
+            redash_type = TYPE_STRING
```
guidopetri, line 130: Can we make this a mapping?

```python
column_type_mappings = {
  np.bool_: TYPE_BOOLEAN,
  np.inexact: TYPE_FLOAT,
  ...
}

redash_type = column_type_mappings[column_type]
```

guidopetri: Oh I see, this was moved from elsewhere. I'd be fine keeping it as is then, but obviously better if it's made more readable :)

redash/query_runner/yandex_disk.py
```diff
@@ -0,0 +1,153 @@
...
+    def _get_tables(self, schema):
+        offset = 0
+        limit = 100
+
+        while True:
+            temp_items = self._send_query(
+                "resources/public", media_type="spreadsheet,text", limit=limit, offset=offset
+            )
+
+            temp_items = temp_items["items"]
```
guidopetri, line 64: Could we use a different variable name so as not to replace the existing var?

redash/query_runner/yandex_disk.py
```diff
@@ -0,0 +1,153 @@
...
+
+            for i in temp_items:
```
guidopetri, line 66: It might be better to use a more descriptive iterator variable name than `i`

redash/query_runner/yandex_disk.py
```diff
@@ -0,0 +1,153 @@
...
+        if file_extension not in EXTENSIONS_READERS:
+            error = f"Unsupported file extension: {file_extension}"
+            return data, error
+        elif file_extension in ("xls", "xlsx"):
+            read_params["sheet_name"] = params.get("sheet_name", 0)
+
+        file_url = self._send_query("resources/download", path=params["path"])["href"]
+
+        try:
+            df = EXTENSIONS_READERS[file_extension](file_url, **read_params)
```
guidopetri, line 137: Love this line ❤️ looks great!

tests/query_runner/test_yandex_disk.py
```diff
@@ -0,0 +1,45 @@
...
+        {"id": 3, "name": "Charlie", "age": 22},
+        {"id": 4, "name": "Dave", "age": 23},
+        {"id": 5, "name": "Eve", "age": 24},
+    ]
+)
+
+test_token = "AAAAQAA"
+
+
+class TestYandexDisk(TestCase):
```
guidopetri, line 21: nitpick: Could we use pytest for testing here instead of unittest?

tests/query_runner/test_yandex_disk.py
```diff
@@ -0,0 +1,45 @@
...
+    def test_xlsx(self):
```
guidopetri, line 22: These tests are great, but we only have ~30% coverage of the new query runner. Could we add some more tests to have >=90%?

guidopetri (commented):
A few comments. Thanks for implementing this :)

redash/query_runner/__init__.py
denisov-vlad: Original code was also written by me :)

Thought about mapping here, but using extra if section for date/datetime prevented me from doing this.

denisov-vlad:
Having the same issue as before with missed numpy package. 

Also could you please update Codecov report?

guidopetri:
The frontend e2e tests [build the docker image without `all_ds` dependencies](https://github.com/getredash/redash/blob/master/.ci/docker-compose.cypress.yml#L6). So that's why it's not finding numpy/pandas. (this wasn't clear to me, so pointing it out more explicitly) Maybe we can add an import guard like in the query runner files?

guidopetri (approved):
This looks awesome! Thanks so much @denisov-vlad , merging now :)

PR #6621, opened 2023-11-24
add: urlencode selector - space as plus or 20

veshus:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Add ability to change quoter for url encode query parameters from default "space as plus" to "space as %20".
Feature is controlled by env REDASH_URLENCODE_SPACE_AS_PLUS = true|false

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

I have an API that treats space as %20, so default behaviour doesn't work. After REDASH_URLENCODE_SPACE_AS_PLUS set to "false", all parameters passes ok.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ npm-debug.log
 
 client/cypress/screenshots
 client/cypress/videos
+
+.yarnrc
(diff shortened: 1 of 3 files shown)


masayuki038:
Hi @veshus, thanks for the PR.

It seems that the part you added is not covered by unit tests. Could you please provide some testing to ensure that this change does not affect existing behavior? It is very helpful for us. thank you.

masayuki038:
@veshus I found https://github.com/getredash/redash/discussions/6610.
Could you reply to it before adding some tests? I also would like to know the background of this update.

guidopetri (changes requested):
I'd also like some clarity on #6610 before merging

PR #6637, opened 2023-12-01
upgraded sqlalchemy deps to versions with wheels to speed up arm64 builds

AndrewChubatiuk:
## What type of PR is this? 
upgraded sqlalchemy deps to speed up arm64 build and avoid building python dependencies, which consumes lot's of time

- [x] update


masayuki038:
@AndrewChubatiuk Thanks for your following up. I will check https://github.com/getredash/redash/pull/6671 and close this.

PR #6646, opened 2023-12-04
Adds influxdb v2 query runner.

fabrei:
- Adds test cases
- Adds influxdb v2 icon
- Updates python dependencies

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## Description

It uses influxdb-client python library which is compatible with influxdb v2. 

It supports ssl connection to influxdb as well.

## How is this tested?

- [x] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

## Related Tickets & Documents

#5391

Issue #5391, opened 2021-02-09
Hope support influxdb v2.0


crazybugliu:
### Issue Summary

Hope support influxdb v2.0

### Steps to Reproduce

1. data source only support influxdb v1.x, we want to use influxdb v2.0 with redash

### Technical details:

* Redash Version: 8.0.0.b32245

manang:
news about this issue?
Thanks

crazybugliu:
> news about this issue?
> Thanks

I hava write one datasource for influxdb v2


```python

import logging

from redash.query_runner import *
from redash.utils import json_dumps
import requests
from dateutil import parser
from pytz import timezone

logger = logging.getLogger(__name__)

enabled = True

# example:
#
#   #group,false,false,true,true,false,false,true,true
#   #datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string
#   #default,mean,,,,,,,
#   ,result,table,_start,_stop,_time,_value,_field,_measurement
#   ,,0,2021-02-09T06:16:54.252458367Z,2021-02-09T09:16:54.252458367Z,2021-02-09T06:20:00Z,,_3V3,88
#   ,,0,2021-02-09T06:16:54.252458367Z,2021-02-09T09:16:54.252458367Z,2021-02-09T06:30:00Z,3.3112,_3V3,88

influx_type_map = {
    "boolean": "bool",
    "unsignedLong": "int",
    "long": "int",
    "double": "float",
    "string": "string",
    "base64Binary": "string",
    "duration": "string",
    "dateTime:RFC3339": "date",
}

type_cast_map = {
    "bool": lambda v: True if v == "true" else False if v == "false" else None,
    "int": lambda v: None if v == "" else int(v),
    "float": lambda v: None if v == "" else float(v),
    "string": lambda v: v,
    "date": lambda v: parser.parse(v).astimezone(timezone("Asia/Shanghai")).strftime("%Y-%m-%dT%H:%M:%S.%f")
}


def _type_cast(v, t):
    t = influx_type_map[t]
    return type_cast_map[t](v)


def _transform_result(csv_str):
    result_columns = []
    result_rows = []
    if csv_str == "" or csv_str == None:
        return json_dumps(
            {"columns": [], "rows": []}
        )

    lines = csv_str.splitlines()

    logger.debug("csv all lines: %d" % len(lines))

    types = lines[1].split(",")[2:]
    result_columns = lines[3].split(",")[2:]
[...]

PR #6669, opened 2023-12-19
Adds influxdb v2 query runner as a default query runner.

fabrei:
## What type of PR is this? 

- [x] Bug Fix

## Description

A new query runner for influxdb v2 was added, but forgotten to add as default query runner in redashs' settings. This is a PR which solves it. At the moment it can be added manually by using an environmental variable: `REDASH_ADDITIONAL_QUERY_RUNNERS=redash.query_runner.influx_db_v2`

## How is this tested?

- [x] Manually

With current docker image, InfluxDBv2 query runner does not appear in list of all query runners. I added it manually in redashs' settings and restarted the container. After that, the runner appeared in list.

There is a ticket #5391 and PR https://github.com/getredash/redash/pull/6646


fabrei:
```
 Container cypress-redis-1  Running
 Container cypress-postgres-1  Running
  File "/app/manage.py", line 6, in <module>
    from redash.cli import manager
  File "/app/redash/__init__.py", line 53, in <module>
    import_query_runners(settings.QUERY_RUNNERS)
  File "/app/redash/query_runner/__init__.py", line 437, in import_query_runners
    __import__(runner_import)
  File "/app/redash/query_runner/influx_db_v2.py", line 41, in <module>
    class InfluxDBv2(BaseQueryRunner):
  File "/app/redash/query_runner/influx_db_v2.py", line 155, in InfluxDBv2
    def _get_data_from_tables(self, tables: TableList) -> Dict:
NameError: name 'TableList' is not defined
```

Why is that happening? Did I miss to add a dependency somewhere? I already added the dependency in `pyproject.toml` and created a new poetry lock file in the PR which added this query runner.

masayuki038:
@fabrei The recent CI has been successful at all, so I think the error is caused by this update.
- https://github.com/getredash/redash/actions/runs/7238652870
- https://github.com/getredash/redash/actions/workflows/ci.yml

Could you investigate this? From a personal point of view, I'm concerned that the class name InfluxDBClient is loaded from different modules.

- https://github.com/getredash/redash/blob/master/redash/query_runner/influx_db.py
- https://github.com/getredash/redash/blob/master/redash/query_runner/influx_db_v2.py

fabrei:
> @fabrei The recent CI has been successful at all, so I think the error is caused by this update.
> 
>     * https://github.com/getredash/redash/actions/runs/7238652870
> 
>     * https://github.com/getredash/redash/actions/workflows/ci.yml

Yeah, the current one fails, because now the new runner is imported on startup. It throws the error at [this point](https://github.com/getredash/redash/actions/runs/7261538944/workflow#L135). 

> Could you investigate this? From a personal point of view, I'm concerned that the class name `InfluxDBClient` is loaded from different modules.
> 
>     * https://github.com/getredash/redash/blob/master/redash/query_runner/influx_db.py
> 
>     * https://github.com/getredash/redash/blob/master/redash/query_runner/influx_db_v2.py

The error says, that the `TableList` is not defined. But it gets [imported](https://github.com/getredash/redash/blob/master/redash/query_runner/influx_db_v2.py#L20). But if the module can't be imported, `TableList` is not defined, as the `ImportError` is caught. So I guess, that in CI the module `influxdb-client` is not installed as a requirement. Even if the code is not used in CI, `TableList` is used in the code for [method definition](https://github.com/getredash/redash/blob/master/redash/query_runner/influx_db_v2.py#L155). 

So, I don't think it is related to the different influxdb modules. If I start redash in a container, with this change, it runs and I can use the new query runner without adding it manually by defining `REDASH_ADDITIONAL_QUERY_RUNNERS`.  And if I call `from redash.cli import manager` from my python shell, as in the traceback is seen, it works in my redash virtual environment as well.

masayuki038:
@fabrei OK. So what do you do with this PR?

fabrei:
> @fabrei OK. So what do you do with this PR?

I would like to understand, why this module is not present at that moment. But I am not familiar with running a CI on github and what happens in the background. Is it possible to run the ci locally? If yes, how?

But what I see is, the error occurs in "frontend-e2e-tests" where cypress test cases are run. So it needs redash. But obviously redash is not setup correctly. Otherwise it would install the influxdb-client dependency and not fail. Or do I miss something?

I think deleting `TableList` from method definition would fix it, but that's not what I want to do.

masayuki038:
@fabrei The way to run e2e test on local is described in https://github.com/getredash/redash/wiki/Testing-your-changes#comprehensive-e2e-testing-cypress.

However, it failed to run `create_db` with docker compose. I don't think this is the cypress issue.

```
Error: Command failed: docker compose -p cypress run server create_db
```

I belive you can do logging the object of `ImportError`.

fabrei:
> @fabrei The way to run e2e test on local is described in https://github.com/getredash/redash/wiki/Testing-your-changes#comprehensive-e2e-testing-cypress.

Thanks, I didn't see this page.
 
> However, it failed to run `create_db` with docker compose. I don't think this is the cypress issue.
> 
> ```
> Error: Command failed: docker compose -p cypress run server create_db

Where did you see this error?

---

Ok, I think I got a better understanding what's happening. In CI container it does not install all dependencies. As you can see [here](https://github.com/getredash/redash/actions/runs/7261538944/job/19784633841#step:6:1062), poetry installs all main dependencies under [tool.poetry.dependencies](https://github.com/fabrei/redash/blob/influxdb-v2-as-default-query-runner/pyproject.toml#L24). But the influxdb-client is not listed there. It is listed under [tool.poetry.group.all_ds.dependencies](https://github.com/fabrei/redash/blob/influxdb-v2-as-default-query-runner/pyproject.toml#L107). I added it there, because the other influxdb dependency was listed there. 

I see two options here. 

1. Put the dependency entry under `tool.poetry.dependencies` instead of `tool.poetry.group.all_ds.dependencies`. But I don't know, if this is be the right way. I understand this dependencies as the main ones which is needed for the core of redash. But the influxdb-client is just an "add on"; in my opinion.
2. I delete the usage of `TableList`, because it is not needed for the query runners' function. It is "only" used for typing. The dependency stays where it is now.

What do you think @masayuki038 ?

masayuki038:
@fabrei 
> Where did you see this error?

On the same page.
image

masayuki038:
@fabrei Thanks for your investigation. I think 2. is reasonable.

fabrei:
> @fabrei Thanks for your investigation. I think 2. is reasonable.

Ok, the CI has passed. Thanks for your help :)

PR #6673, opened 2023-12-22
Import all Plotly visualizations

eradman:
## What type of PR is this? 

- [x] Feature

## Description

- Accessible using the Custom chart type
- Disable 'fs' and 'path' modules which are available in node, but not on the frontend.

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually

Create any chart documented at https://plotly.com/javascript/


guidopetri (approved):
LGTM! Thanks Eric :)

joeyrensh:
how to upgrade this change based on existed redash instance, already install 10.0.1 for a while and configure some query and dashboard. i don't want to redeploy all the env and lost the content configured

eradman:
@joeyrensh if you noticed a problem when upgrading, please start new ticket and include log messages and screenshots.

PR #6687, opened 2024-01-03
removed PseudoJSON

AndrewChubatiuk:
## What type of PR is this? 
Removed unneeded PseudoJSON and DBPersistence classes, migrated all json contained column's type to JSONB

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [x] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


redash/models/__init__.py
```diff
@@ -474,11 +473,11 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
     last_modified_by = db.relationship(User, backref="modified_queries", foreign_keys=[last_modified_by_id])
     is_archived = Column(db.Boolean, default=False, index=True)
     is_draft = Column(db.Boolean, default=True, index=True)
-    schedule = Column(MutableDict.as_mutable(PseudoJSON), nullable=True)
-    interval = pseudo_json_cast_property(db.Integer, "schedule", "interval", default=0)
+    schedule = Column(MutableDict.as_mutable(JSON), nullable=True)
```
eradman, line 476: Is there a downside to using JSONB here?

eradman (changes requested):
This solves a long-awaited TODO!

Can we also convert `query_results.data` to JSONB?

redash/models/__init__.py
AndrewChubatiuk: yes, we can. does it make sense to use JSONB instead of all JSON fields, especially for ones, that are searchable

eradman: Great; I think this is good to go after converting remaining JSON to JSONB

We can make the change to `query_results.data` as a separate change if you'd like

AndrewChubatiuk: @eradman converted `query_results.data`, all `options` and other json columns to jsonb

pyproject.toml
```diff
@@ -72,8 +72,8 @@ semver = "2.8.1"
 sentry-sdk = "1.28.1"
 simplejson = "3.16.0"
 sqlalchemy = "1.3.24"
-sqlalchemy-searchable = "1.2.0"
-sqlalchemy-utils = "0.34.2"
+sqlalchemy-searchable = "1.4.1"
+sqlalchemy-utils = "0.37.9"
```
eradman, line 76: Is it nessesary to bump these versions at the same time? If possible I would prefer that updates dependencies are in a separate change to make troubleshooting easier.

AndrewChubatiuk: @eradman removed these changes

eradman (approved):
This change appears to be in good shape

eradman:
@justinclift what do you think about waving the code coverage check? It would be very difficult to make this check pass with a change like this

justinclift:
@eradman Yeah, the code coverage stuff is definitely a case of "don't worry about it in situations where it doesn't apply". :smile:

justinclift:
It looks like using jsonb is having some unintended consequences for people with large data sets: https://github.com/getredash/redash/issues/6706#issuecomment-1893180033

We might need to revert this change.

eradman:
I beleive we can change `query_results.data` back to a text field and the rest of this change can stay as is.

@AndrewChubatiuk can you explore that solution?

Issue #6706, opened 2024-01-15
DB migration failed (change type of json fields from varchar to json) in the latest Docker image


myonlylonely:
### Issue Summary

After upgrading to the latest Redash Docker image, the db migration process failed.
It might be related to the recent commit ec1c4d07deca7b9b8a49504a28f0a00a0feb59ab.

### Steps to Reproduce

1. Update to the latest Docker image.
2. run command `docker-compose run --rm server manage db upgrade`.
```
[2024-01-15 21:56:29,506][PID:3862][INFO][xmlschema] Resource 'XMLSchema.xsd' is already loaded
[2024-01-15 21:56:30,190][PID:3862][INFO][alembic.runtime.migration] Context impl PostgresqlImpl.
[2024-01-15 21:56:30,190][PID:3862][INFO][alembic.runtime.migration] Will assume transactional DDL.
[2024-01-15 21:56:30,243][PID:3862][INFO][alembic.runtime.migration] Running upgrade 7ce5925f832b -> 7205816877ec, change type of json fields from varchar to json
Traceback (most recent call last):
  File "/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
    self.dialect.do_execute(
  File "/venv/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.DatatypeMismatch: result of USING clause for column "data" cannot be cast automatically to type jsonb
HINT:  You might need to add an explicit cast.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 9, in <module>
    manager()
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #6713, opened 2024-01-17
convert query result back to text

AndrewChubatiuk:
## What type of PR is this? 
Convert QueryResult data column back to text due to a limit of JSON(10MB) and JSONB(255MB) column type size

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Removed from migrations conversion of QueryResult.data to JSONB, and added custom JsonText type, which is actually a copy of removed PseudoJson type

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -48,6 +48,7 @@ from redash.models.parameterized_query import (
 from redash.models.types import (
     Configuration,
     EncryptedConfiguration,
+    JSONText,
     MutableDict,
     MutableList,
     json_cast_property,
@@ -315,7 +316,7 @@ class QueryResult(db.Model, BelongsToOrgMixin):
     data_source = db.relationship(DataSource, backref=backref("query_results"))
     query_hash = Column(db.String(32), index=True)
     query_text = Column("query", db.Text)
-    data = Column(MutableDict.as_mutable(JSONB), nullable=True)
+    data = Column(JSONText, nullable=True)
     runtime = Column(DOUBLE_PRECISION)
     retrieved_at = Column(db.DateTime(True))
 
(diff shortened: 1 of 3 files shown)


justinclift:
Thanks for investigating this @AndrewChubatiuk. :smile:

It looks like might need a bit more work before things are totally happy though: https://github.com/getredash/redash/issues/6706#issuecomment-1895562375

redash/models/__init__.py
```diff
@@ -315,7 +315,7 @@ class QueryResult(db.Model, BelongsToOrgMixin):
     data_source = db.relationship(DataSource, backref=backref("query_results"))
     query_hash = Column(db.String(32), index=True)
     query_text = Column("query", db.Text)
-    data = Column(MutableDict.as_mutable(JSONB), nullable=True)
```
eradman, line 318: This was

```
_data = Column("data", db.Text)
```

eradman (changes requested):
The previous type was TEXT, not JSON

redash/models/__init__.py
AndrewChubatiuk: thought we can use the benefits of sqlalchemy JSON class and have Text type in a db, but unfortunately it converts column automatically
converted to Text column

eradman (approved):
I confirmed this change reverts `query_results.data` back to the `text` data type

Migration looks good; although the next preview docker build will be proof of that

justinclift:
Cool, lets merge this then.  Thanks for getting this done @AndrewChubatiuk. :smile:

PR #6716, opened 2024-01-18
switch to pull_request_target events to hide cypress secrets

AndrewChubatiuk:
## What type of PR is this? 
switched to pull_request_target events to remove secrets from ci config

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,16 +3,22 @@ on:
   push:
     branches:
       - master
-  pull_request:
+  pull_request_target:
+    branches:
+      - master
 env:
   NODE_VERSION: 16.20.1
 jobs:
   backend-lint:
     runs-on: ubuntu-22.04
     steps:
+      - if: github.event.pull_request.mergeable == 'false'
+        name: Exit if PR is not mergeable
+        run: exit 1
       - uses: actions/checkout@v3
         with:
           fetch-depth: 1
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
       - uses: actions/setup-python@v4
         with:
           python-version: '3.8'
@@ -29,9 +35,13 @@ jobs:
       COMPOSE_DOCKER_CLI_BUILD: 1
       DOCKER_BUILDKIT: 1
     steps:
+      - if: github.event.pull_request.mergeable == 'false'
+        name: Exit if PR is not mergeable
+        run: exit 1
       - uses: actions/checkout@v3
         with:
           fetch-depth: 1
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
       - name: Build Docker Images
         run: |
           set -x
@@ -65,9 +75,13 @@ jobs:
   frontend-lint:
     runs-on: ubuntu-22.04
     steps:
+      - if: github.event.pull_request.mergeable == 'false'
+        name: Exit if PR is not mergeable
+        run: exit 1
       - uses: actions/checkout@v3
         with:
           fetch-depth: 1
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_VERSION }}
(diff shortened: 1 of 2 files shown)


justinclift:
Interesting.  `pull_request_target` isn't something I've comes across before.  I'll need to look into it tomorrow (hopefully) sometimes, as it would be good if this fixes our problem.

On a different note, it seems like the CI tests didn't run properly.  It's probably something to do with this change, so it might need a bit of adjusting still so things work. :smile:

AndrewChubatiuk:
looks like it should first appear in master to start working

justinclift:
Interesting.  We'll probably need to try it in a fork or something and see how it goes. :smile:

justinclift:
I *think* (without investigating too much) that it should be fairly easy to test.

If you create a brand new GitHub repository called (for example) "redash-test", then upload the Redash code to it... that would be suitable for testing purposes.

Something like this from the command line (after creating the new repo in the GitHub web interface):

```
$ git remote add redash-test https://github.com/AndrewChubatiuk/redash-test
$ git push redash-test switch-to-pr-target:master
```

You would probably need to update the repo name in the places that check for it as well, so everything runs.  For example:

https://github.com/getredash/redash/blob/97db4925317b65abefc957fe47cf83a95c772c53/client/cypress/cypress.js#L70

That would need `getredash/redash` -> `AndrewChubatiuk/redash-test`.  There might be other places that need that string changed too... I just don't remember.

Does that all make sense? :smile:

AndrewChubatiuk:
Actually found an answer on stackoverflow
https://stackoverflow.com/a/77069245

justinclift:
Doesn't that just say the same thing again?  eg "It must be on the master branch in order to work"

So, you need to test it in the `master` branch on a separate repo to verify it's all good. (?)

AndrewChubatiuk:
@justinclift i need to be a part of two organizations to be able to test this flow, cause cannot fork repo under my username and then create a PR. its much easier to copy credentials to getredash/redash secret merge this MR and revert commit if it's not okay. there's not intensive development in a repo to affect someone

AndrewChubatiuk:
@justinclift tested in RedashCommunity/redash-clone repo
https://github.com/RedashCommunity/redash-clone/actions/runs/7631698500
Added some extra fixes to this PR

justinclift:
@AndrewChubatiuk Cool, good stuff. :smile:

Any idea if the warning message on that run is important?

image

justinclift:
Looking through the cloned repo and PR, it does look like things should be good now.

With the new `CYPRESS_PROJECT_ID`, `CYPRESS_RECORD_KEY`, and `PERCY_TOKEN` secrets that need creating for this repo, they should be in plain text yeah, not base64 encoded?

justinclift:
@gabrieldutra @arikfr Any idea how we generate new tokens for Percy and Cypress to use here?

AndrewChubatiuk:
> @AndrewChubatiuk Cool, good stuff. 😄
> 
> Any idea if the warning message on that run is important?
> 
> image

We have the same situation now in master

justinclift:
> We have the same situation now in master

Cool, we can ignore it for this PR then. :smile:

AndrewChubatiuk:
@justinclift can we merge this already without waiting for the new secrets?

guidopetri:
If we merge it without the new secrets, doesn't that invalidate the point of the PR? :)

AndrewChubatiuk:
point of PR is just to remove secrets from github action config and the values of these secrets can be updated later

justinclift:
If we merge this as-is before the new secrets are created, won't the CI break?

AndrewChubatiuk:
it should be fine if PERCY_TOKEN, CYPRESS_RECORD_KEY and CYPRESS_PROJECT_ID exist in repo secrets

justinclift:
Cool.  I've just created those secrets for this repo (at the repo level, not the organisation level).

They just have placeholder values for now that definitely aren't correct. :wink:

justinclift (approved):
k, lets try this out.

justinclift:
Ugh.  Merging is still blocked.  I'll need to temporarily give myself permission to force the merge.

justinclift:
k, merged.  Thanks for getting this done @AndrewChubatiuk, even though it wasn't all that easy to test. :smile:

AndrewChubatiuk:
@justinclift 
need to set origin cypress secrets. Pipeline in master is failing with
```Your Record Key unkno...known is not valid with this projectId```

PR #6723, opened 2024-01-20
Upgrade node 18

AndrewChubatiuk:
## What type of PR is this? 
though nodejs is not the latest lts version, but at least it's still maintained

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


justinclift:
From rough memory, either @gabrieldutra or @arikfr have tried migrating to Node 18 before and there was a problem with it.

That way probably ages ago though, so the problem might not still be present.

Hopefully they have time to look at this PR and let us know. :smile:

PR #6730, opened 2024-01-26
Add firebird query runner

dvandonkelaar:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [X] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## Description

Support of the Firebird database

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

N/A

N/A


Dockerfile
```diff
@@ -76,7 +76,19 @@ RUN if [ "$TARGETPLATFORM" = "linux/amd64" ]; then \
...
+  && tar -xvf /tmp/firebird.tar.gz --directory /tmp/firebird \
+  && mkdir /tmp/firebird/buildroot \
+  && tar -xvf /tmp/firebird/Firebird-*/buildroot.tar.gz --directory /tmp/firebird/buildroot \
+  && cp /tmp/firebird/buildroot/opt/firebird/lib/libfbclient.so /usr/lib/libfbclient.so.2 \
+  && ln -s /usr/lib/libfbclient.so.2 /usr/lib/libfbclient.so \
+  && cp /tmp/firebird/buildroot/opt/firebird/lib/libtomcrypt.so /usr/lib/libtomcrypt.so.1 \
+  && ln -s /usr/lib/libtomcrypt.so.1 /usr/lib/libtomcrypt.so \
+  && rm /tmp/firebird.tar.gz \
+  && rm -rf /tmp/firebird \
+  && ln -s /usr/lib/x86_64-linux-gnu/libtommath.so.1 /usr/lib/x86_64-linux-gnu/libtommath.so.0
```
guidopetri, line 91: Does this really need to be installed for a firebird connector to work? I'd imagine we can just ping an existing firebird db to query it - I don't think we'd have to actually install it in the redash image.

guidopetri (changes requested):
@dvandonkelaar thanks for the PR! It looks like we need some tests to pass codecov - would you mind adding some in? We prefer pytest, but if you're more comfortable with unittest that also works.

Dockerfile
dvandonkelaar: Unfortunately, we need some client libraries to be able to connect to a Firebird database.
See the [docs](https://firebirdsql.org/file/documentation/html/en/firebirddocs/qsg5/firebird-5-quickstartguide.html#qsg5-installing-client) for more info.

dvandonkelaar:
Building tests to check if we can connect to a Firebird database is unfortunately not possible. Unfortunately, there is no public database available to connect and query.

I agree that we need to test as much as possible, but don't see how we can test this. I looked into the MSSQL query runner as a template for Firebird (because they work sort or less alike) and also did not find any tests (or maybe I am looking in the wrong place?)

guidopetri:
Maybe we can mock the connection? I agree we don't have ideal test coverage, but there's really only one way to improve that.

dvandonkelaar:
Yes, I agree. Let's see if we can mock a connection. Maybe we can also cover some other query runners. I will see if I can accomplish this next week.

PR #6715, opened 2024-01-18
Fix UnicodeEncodeError for non-Latin characters

dvandonkelaar:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## Description

The Slack Notification gives an error on non-Latin characters (for example '한글'). This fix encodes the message to UTF-8 to prevent the error.

## How is this tested?

- [x] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

## Related Tickets & Documents

The issue is described in #6710

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
N/A

PR #6738, opened 2024-02-01
Update mattermost.py

kolya7k:
Fix UnicodeEncodeError for non-Latin characters for mattermost destination (#6715)


guidopetri:
@kolya7k thanks for the PR! While you're at it, do you think we should update all destinations to support UTF8?

kolya7k:
Yes, but i encountered with this problem on MM because i'm using MM and Cyrillic symbols :)
I can later make a PR for every destination

guidopetri:
Could we do it in this PR since it's a one line change?

eradman:
Rebasing to see if tests pass

PR #6741, opened 2024-02-03
fixed custom json encoders

AndrewChubatiuk:
## What type of PR is this? 
Fixes #6737 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

PR #6745, opened 2024-02-06
fix linting

guidopetri:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

One of the recent commits to `master` broke the linting, so this PR fixes that.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

### DIFF
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -74,11 +74,7 @@ class JSONEncoder(json.JSONEncoder):
     """Adapter for `json.dumps`."""
 
     def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


guidopetri:
@AndrewChubatiuk @justinclift it looks like `pull_request_target` is not doing its job correctly here. It's running `black` and `ruff` vs master, but we want it to run vs the PR branch. That would explain why #6741 checks succeeded in the PR, but failed on master.

Issue #6747, opened 2024-02-06
Mongo Query Runner Processing time is broken


clearnote01:
### Issue Summary

MongoDB Connector's processing time is broken.

QueryRunner: mongodb
<img width="1271" alt="image" src="https://github.com/getredash/redash/assets/9555129/d3d7c639-6ed5-4dc3-b1dd-79223a9f10d3">
### Steps to Reproduce

1. Run a query with MongoDB query connector ( I used an aggregate query)
2. See the status bar which shows time taken for processing

### Technical details:

* Redash Version: master branch
* Browser/OS: chrome
* How did you install Redash: local-dev-setup

clearnote01:
This is fixed now.

PR #6748, opened 2024-02-06
Load custom encoder only when runner enabled

AndrewChubatiuk:
## What type of PR is this?
custom json encoders import, which was introduced in #6687 and fixed in #6741 includes all encoders which leads to E2E test failure as mongodb dependencies are not installed during E2E test execution, but ObjectId is used inside mongodb custom encoder. this PR filters out encoders of disabled query runners

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [x] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -70,14 +70,11 @@ def generate_token(length):
     return "".join(rand.choice(chars) for x in range(length))
 
 
-json_encoders = [m.custom_json_encoder for m in sys.modules if hasattr(m, "custom_json_encoder")]
-
-
 class JSONEncoder(json.JSONEncoder):
     """Adapter for `json.dumps`."""
 
     def __init__(self, **kwargs):
-        self.encoders = json_encoders
+        self.encoders = [m.custom_json_encoder for m in sys.modules.values() if hasattr(m, "custom_json_encoder")]
         super().__init__(**kwargs)
 
     def default(self, o):
@@ -111,7 +108,7 @@ class JSONEncoder(json.JSONEncoder):
         elif isinstance(o, bytes):
             result = binascii.hexlify(o).decode()
         else:
-            result = super(JSONEncoder, self).default(o)
+            result = super().default(o)
         return result
 
 
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -20,7 +20,6 @@ from funcy import select_values
 from sqlalchemy.orm.query import Query
 
 from redash import settings
-from redash.query_runner.registry import query_runners
 
 from .human_time import parse_human_time
 
@@ -75,6 +74,8 @@ class JSONEncoder(json.JSONEncoder):
     """Adapter for `json.dumps`."""
 
     def __init__(self, **kwargs):
+        from redash.query_runner import query_runners
+
         self.encoders = [r.custom_json_encoder for r in query_runners.values() if hasattr(r, "custom_json_encoder")]
         super().__init__(**kwargs)
 
@@ -82,8 +83,7 @@ class JSONEncoder(json.JSONEncoder):
         for encoder in self.encoders:
             result = encoder(self, o)
             if result:
-                o = result
-                break
+                return result
         if isinstance(o, Query):
             result = list(o)
         elif isinstance(o, decimal.Decimal):
--- a/redash/utils/__init__.py
+++ b/redash/utils/__init__.py
@@ -20,6 +20,7 @@ from funcy import select_values
 from sqlalchemy.orm.query import Query
 
 from redash import settings
+from redash.query_runner.registry import query_runners
 
 from .human_time import parse_human_time
 
@@ -74,14 +75,15 @@ class JSONEncoder(json.JSONEncoder):
     """Adapter for `json.dumps`."""
 
     def __init__(self, **kwargs):
-        self.encoders = [m.custom_json_encoder for m in sys.modules.values() if hasattr(m, "custom_json_encoder")]
+        self.encoders = [r.custom_json_encoder for r in query_runners.values() if hasattr(r, "custom_json_encoder")]
         super().__init__(**kwargs)
 
     def default(self, o):
         for encoder in self.encoders:
             result = encoder(self, o)
             if result:
-                return result
+                o = result
+                break
         if isinstance(o, Query):
             result = list(o)
         elif isinstance(o, decimal.Decimal):
--- a/.ci/compose.ci.yaml
+++ b/.ci/compose.ci.yaml
@@ -1,4 +1,3 @@
-version: '2.2'
 services:
   redash:
     build: ../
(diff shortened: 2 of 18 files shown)


redash/utils/json.py
```diff
@@ -0,0 +1,67 @@
...
+import uuid
+from sqlalchemy.orm.query import Query
+from datetime import timedelta, datetime
+from redash.query_runner import query_runners
+
+class JSONEncoder(json.JSONEncoder):
+    """Adapter for `json.dumps`."""
+
+    def __init__(self, **kwargs):
+        self.encoders = [
```
clearnote01, line 13: Just curious why is this an array? Is it possible to have result from multiple query runners at this point?

AndrewChubatiuk: this is not a result, these are custom json encoder functions for each query runner

AndrewChubatiuk:
this MR will fail cause it's now running against master, need to merge it to fix CI

guidopetri:
Hi Andrii, thanks for the PR. Obviously we'll need to fix CI first before considering this.

That being said, this is the third json-encoder-related PR I've seen but I'm still unclear on what is going on with these changes and what the fix is, as well as how it's been tested. Would you mind spending some time explaining these changes more in depth?

guidopetri (changes requested):
Hey, please don't put unrelated changes into a PR like this, it makes it hard to track what happens in a PR. Future programmers - you or I or someone else - will be very confused to see CI changes in a commit talking about json encoders.

AndrewChubatiuk:
@justinclift @guidopetri https://github.com/getredash/redash/pull/6749

justinclift:
@AndrewChubatiuk @guidopetri k, I've merged that other PR for fix the CI branch problem.

justinclift:
@AndrewChubatiuk Shouldn't this PR be passing CI now that the other one is merged and this one has been rebased?

AndrewChubatiuk:
@justinclift pushed unit test fixes

justinclift:
@AndrewChubatiuk Excellent, thanks. :smile:

redash/utils/json.py
clearnote01: How is there control then on what custom_json_encoder is then actually used? For example, pg.py and mognodb.py both have custom_json_encoder defined. Then self.encoders will be [encoder_from_pg, encoder_from_mongodb].

In that case, say we had a query result from mongo which needed to be encoded, wouldn't the following code try to use both encoders, one by one and say if both encoders had different logic for encoding Timestamp, whichever encoder is first in list get to do the encoding?

```
 def default(self, o):
        for encoder in self.encoders:
            result = encoder(self, o)
            if result:
                return result
```

AndrewChubatiuk: doesn't matter which encoder is first, cause custom encoders are to process custom data types, provided by runner's lib, like ObjectId for mongodb

AndrewChubatiuk: I've pushed a fix for #6747

clearnote01: i don't get how your change isn't recursive? if encoder(self, o) returned an encoded string, then none of the elif condition in default(...) catch it, which could result in super().default(o) called again?

clearnote01: thanks for clarification : )

AndrewChubatiuk:
@justinclift @guidopetri can we merge this to fix CI in master?

redash/utils/json.py
```diff
@@ -0,0 +1,71 @@
...
+
+from redash.query_runner import query_runners
+
+
+class JSONEncoder(json.JSONEncoder):
+    """Adapter for `json.dumps`."""
+
+    def __init__(self, **kwargs):
+        self.encoders = [
+            r.custom_json_encoder for r in query_runners.values() if hasattr(r, "custom_json_encoder") and r.enabled
```
guidopetri, line 17: I notice this line seems to be the only actual change in this PR. @AndrewChubatiuk , please undo all the other changes - e.g. creating this json.py file and all the import changes, as well as the classmethod additions. Work like this doesn't make it easy to review code, nor does it make it easy to follow code when investigating issues.

guidopetri (changes requested):

redash/utils/json.py
guidopetri: Additionally, please explain this change in the PR description in two or three paragraphs, providing:
- context of why this change is necessary
- what led to this change (including links to all previous related PRs)
- what the current behavior is, and why it's wrong
- and what the intended behavior is.
 
Additionally, please describe how you've tested it. If you only have manual tests, then please add some kind of automated test with pytest or unittest to make sure that this does not break again.

AndrewChubatiuk: i cannot make this change without moving JSONEncoder to a separate file. It will lead to circular import problem as `query_runners` variable is inside `redash`, which uses `redash.utils` and JSONEncoder is inside `redash.utils` module, which uses `redash module`

AndrewChubatiuk: @guidopetri also updated PR description

guidopetri: Can you try making the import of `query_runners` that you need within the JSONEncoder definition? I believe that will circumvent the circular import.

AndrewChubatiuk: @guidopetri commited in a way you want. you can check CI logs

guidopetri: I understand that the current import location leads to a circular import. Please read my comment above more carefully - can we make the import within the JSONEncoder definition?

AndrewChubatiuk: done

AndrewChubatiuk: @guidopetri tried but imports, which cause circular dependencies issues inside classes, but it didn't work. moved `query_runners` to another module to have less changes comparing to moving json methods to a separate module

redash/utils/__init__.py
```diff
@@ -74,7 +74,9 @@ class JSONEncoder(json.JSONEncoder):
     """Adapter for `json.dumps`."""
 
     def __init__(self, **kwargs):
-        self.encoders = [m.custom_json_encoder for m in sys.modules.values() if hasattr(m, "custom_json_encoder")]
+        from redash.query_runner import query_runners
+
+        self.encoders = [r.custom_json_encoder for r in query_runners.values() if hasattr(r, "custom_json_encoder")]
```
guidopetri, line 79: As you can see, this change is much smaller and more readable than previously: only the essential was done to fix the issue at hand. We can move things away from the `__init__.py` files in a later PR, but for now we need to fix the CI, so we shouldn't do any refactor changes if they're not necessary.

guidopetri:
Merging now that we're passing the tests. The codecov isn't as important in this circumstance but we should improve it later.

Issue #6770, opened 2024-02-18
Can not upgrade redash/base to redash/preview


trantrinhquocviet:
I'm experiencing difficulties when attempting to upgrade the Redash base image from the stable version to the preview version. The problem appears to be related to the database configuration. Despite following the upgrade process outlined in the documentation, I'm encountering errors that hinder the successful migration of the database schema or data.

I've referred to the documentation provided by Redash, but the issue persists. I've also checked for any specific instructions or troubleshooting steps related to database migration in the preview version, but haven't found any definitive solutions.

Could you please provide guidance or assistance on how to resolve this issue? Any insights or recommendations would be greatly appreciated.

I have a problem when upgrade redash/base image from [stable ](https://hub.docker.com/layers/redash/base/latest/images/sha256-4c819e245317823432a4468705112bda9ccdab35f260cc708a5ab58331c66d2a)version

(link: https://hub.docker.com/layers/redash/base/latest/images/sha256-4c819e245317823432a4468705112bda9ccdab35f260cc708a5ab58331c66d2a)

to redash/review: https://hub.docker.com/r/redash/preview

![image](https://github.com/getredash/redash/assets/26039338/17e3fd1f-6172-4281-b036-9e816b748002)

Have problem at database

ERROR:  function jsonb_typeof(text) does not exist at character 1096
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
STATEMENT:  SELECT queries.query AS queries_query, queries.updated_at AS queries_updated_at, queries.created_at AS queries_created_at, [queries.id](http://queries.id/) AS queries_id, queries.version AS queries_version, queries.org_id AS queries_org_id, queries.data_source_id AS queries_data_source_id, queries.latest_query_data_id AS queries_latest_query_data_id, [queries.name](http://queries.name/) AS queries_name, queries.description AS queries_description, queries.query_hash AS queries_query_hash, queries.api_key AS queries_api_key, queries.user_id AS queries_user_id, queries.last_modified_by_id AS queries_last_modified_by_id, queries.is_archived AS queries_is_archived, queries.is_draft AS queries_is_draft, queries.schedule AS queries_schedule, queries.schedule_failures AS queries_schedule_failures, queries.options AS queries_options, queries.search_vector AS queries_search_vector, queries.tags AS queries_tags, [query_results_1.id](http://query_results_1.id/) AS query_results_1_id, query_results_1.retrieved_at AS query_results_1_retrieved_at
[...]

PR #6824, opened 2024-03-21
BigQuery: use default for useQueryAnnotation option

eradman:
## What type of PR is this? 
- [x] Bug Fix

## Description

This option may not be set after an upgrade

## How is this tested?

Fix not verified, but okay from @trantrinhquocviet

https://github.com/getredash/redash/issues/6770


justinclift (approved):
Cool, sounds like a critical fix.  Lets get this merged. :smile:

PR #6837, opened 2024-03-28
Handle decimal types in query results

wlach:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Since #6687, we don't serialize query results as JSON before returning them. This is fine, except for the
query results data source which needs to pass the data directly to sqlite3, and doesn't know how to
do that with the decimal types that are occasionally returned by (at least) the PostgreSQL query runner:

https://www.psycopg.org/docs/faq.html#problems-with-type-conversions

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

N/A

### DIFF
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -1,3 +1,4 @@
+import decimal
 import hashlib
 import logging
 import re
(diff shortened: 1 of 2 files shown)


wlach:
@eradman @justinclift Can you take a look?

justinclift (approved):
Sounds sensible to me. :smile:

PR #6842, opened 2024-04-01
Use setup-python@v5, setup-node@v4

eradman:
## What type of PR is this? 

- [x] Update to GitHub Actions

## Description

To avoid warnings in the CI pipeline
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20

## How is this tested?

- [x] Manually

Verified that using the latest versions of these actions solved the problem for a private repo.


justinclift (approved):
Good idea. :smile:

PR #6844, opened 2024-04-02
flatten all level

KimBioInfoStudio:
the original code only flatten 2d-dict, if there are some dict more than 2d, hard to qurey data and viz

TODO:
- [x] add a option btn to enable/disable this feature


justinclift:
@KimBioInfoStudio Heya, there's no description for this PR. :wink:

Um... why is flattening stuff useful for MongoDB?

KimBioInfoStudio:
> @KimBioInfoStudio Heya, there's no description for this PR. 😉
> 
> Um... why is flattening stuff useful for MongoDB?

in the original code, only 2d dict was faltten, but my data struct have 4d dict like, it's hard for me to query on low level and viz on love level:

```json
{ 
"a":"aa",
"b":["bb1","bb2","bb3","bb4"]
"c":{
"cc1"：["cc11","cc12","cc13"],
"cc2": {"cc21k":"cc21v"}
}
}
```

justinclift:
@AndrewChubatiuk @eradman What do you both think of this?

I'm concerned that this will potentially break things for existing users of MongoDB.

If that's not likely to happen though, then this might be ok. :smile:

KimBioInfoStudio:
> @AndrewChubatiuk @eradman What do you both think of this?
> 
> I'm concerned that this will potentially break things for existing users of MongoDB.
> 
> If that's not likely to happen though, then this might be ok. 😄

u r right, if the option is not configureable, this will break existing mongodb users

justinclift:
That sounds like a problem. :wink:

AndrewChubatiuk:
Hey @KimBioInfoStudio 
Nice feature
LGTM

justinclift:
Thanks for completing this @KimBioInfoStudio, and thanks @AndrewChubatiuk for reviewing. :smile:

justinclift (approved):
Going with @AndrewChubatiuk's review conclusion. :wink:

PR #6846, opened 2024-04-02
Handle timedelta in query results

wlach:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

(this is a follow-on to #6837)

Since https://github.com/getredash/redash/pull/6687, we don't serialize query results as JSON before returning them. This is fine, except for the
query results data source which needs to pass the data directly to sqlite3, and doesn't know how to
do that with the timedelta types that are occasionally returned by (at least) the PostgreSQL query runner:

https://www.psycopg.org/docs/faq.html#problems-with-type-conversions

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

N/A

### DIFF
--- a/redash/query_runner/query_results.py
+++ b/redash/query_runner/query_results.py
@@ -1,3 +1,4 @@
+import datetime
 import decimal
 import hashlib
 import logging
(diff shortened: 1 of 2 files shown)


wlach:
@justinclift @eradman Another one to look at when you have a sec

justinclift:
@eradman Are you ok to look over this one? :smile:

eradman (approved):
This looks good; effectively the same as casting to `text` in the query.

PR #6851, opened 2024-04-05
Automatically use the latest version of PostgreSQL

justinclift:
## What type of PR is this? 

- [x] Other

## Description

Instead of keeping PostgreSQL pinned to a specific version, which we need to manually update over time, lets just use the `latest` version.

This should have no negative impact on us, and the pgautoupgrade repository already automatically upgrades between PostgreSQL versions anyway.

## How is this tested?

- [x] E2E Tests (Cypress)

This PR passes all of our unit tests and Cypress tests when run on my local system.


eradman (commented):
Do we also want to make this change to `.ci/compose.ci.yaml` and `.ci/compose.cypress.yaml`?

justinclift:
Thanks, totally forgot about those.  I'll update the PR to include those as well. :smile:

justinclift:
k, those two have been updated and added to this PR as well.

There don't seem to be any further missed compose files using PostgreSQL, so I think this PR is good now. :smile:

eradman (approved):
If we run into a problem with using PG latest it will most likely be because of a bug in SQLAlchemy. Looks good

PR #6854, opened 2024-04-06
Added support for JDBC, added sample jar for Hazelcast

devops-42:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


justinclift:
@devops-42 Good to see enthusiasm. :smile:

While you're at it, please fill out the fields in the top comment so the other developers have some idea what this is for. :grin:

---

Would you be ok to add further tests to this, to increase the code coverage and get the `codecov` checks to pass?

Dockerfile
```diff
@@ -96,7 +99,9 @@ RUN /etc/poetry/bin/poetry install --only $install_groups $POETRY_OPTIONS
 
 COPY --chown=redash . /app
 COPY --from=frontend-builder --chown=redash /frontend/client/dist /app/client/dist
-RUN chown redash /app
+ADD https://github.com/hazelcast/hazelcast-jdbc/releases/download/v5.4.0-BETA-2/hazelcast-jdbc-5.4.0-BETA-2.jar /jdbc/hazelcast-jdbc.jar
+
+RUN chown redash /app /jdbc/*
```
justinclift, line 104: Please use the exact filename rather than `*`.  Just for precision purposes. :smile:

justinclift:
Interestingly, it looks like Hazelcast have moved to be Open Core:

https://hazelcast.com/blog/changes-to-community-edition/

The "not making CVE fixes available to the Community until the next point release" sounds super dangerous. :frowning:

The other items sound reasonable (to me) though at first read.

justinclift:
On a different note, it looks like there are official Hazelcast docker images:

https://hub.docker.com/r/hazelcast/hazelcast/

We could potentially use those for automatically testing this new data source, to make sure it doesn't bit rot over time to the point of breakage.

Dockerfile
```diff
@@ -61,6 +63,7 @@ RUN apt-get update && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
+ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk-arm64
```
KimBioInfoStudio, line 66: I think this should be a common path not only for arm64?

devops-42:
> On a different note, it looks like there are official Hazelcast docker images:
> 
> https://hub.docker.com/r/hazelcast/hazelcast/
> 
> We could potentially use those for automatically testing this new data source, to make sure it doesn't bit rot over time to the point of breakage.
> 
> Looks like there's an official Hazelcast Python (client) library too:
> 
> https://pypi.org/project/hazelcast-python-client/
> 
> @devops-42 Would it have been better to use that, rather than go the JDBC route?
> 
> Am guessing you've seen that already though, and figured JDBC was the better approach here. 😄

Hey @justinclift,

have seen it. I considered to go the JDBC way to have a data source, which might be used for other JDBC connections too. :)

devops-42:
> Interestingly, it looks like Hazelcast have moved to be Open Core:
> 
> https://hazelcast.com/blog/changes-to-community-edition/
> 
> The "not making CVE fixes available to the Community until the next point release" sounds super dangerous. 😦
> 
> At first read, the other items sound reasonable (to me) though.

The future handling of CVE is indeed an issue 😐 Maybe the community is strong enough to mitigate this.

justinclift:
>  I considered to go the JDBC way to have a data source, which might be used for other JDBC connections too. :)

No worries, as long as it works reliably. :smile:

arikfr:
I appreciate the enthusiasm here and the effort put already, but I don't think we should merge this. The usefulness of a JDBC connector is very limited -- I doubt there are meaningful databases that have only JDBC connector and don't have alternative options. But on the other hand it will bloat our Docker image and add extra dependencies to maintain over time, that I doubt many on the current team use or familiar with.

An alternative path for you (or whoever interested in JDBC) is to build a custom image on top of the official one that adds the JDBC dependencies and the JDBC query runner. The way Redash is built it should be possible to do so without having to build your own image from scratch.

justinclift:
It's been a few months since the Hazelcast people announced they're not including security fixes (not even urgent ones) in their Community release.  Looking at their Feature Comparison page, it seems like they're not changing that stance to anything reasonable:

https://hazelcast.com/products/feature-comparison/

Personally, I don't reckon we should support an organisation that deliberately puts their users at risk.  As in, lets not merge any form of this, JDBC nor Python based driver. :smile:

Or am I being unreasonable? :smile:

PR #6855, opened 2024-04-07
Extend `make up` to automatically initialise the database

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR extends `make up` to automatically initialise the database before starting the Redash containers.

It does so by checking if the database tables exist, and if not it runs the `create_database` target to populate them.

## How is this tested?

- [x] E2E Tests (Cypress)

Both unit and Cypress tests pass with no issue, as is expected with this kind of change.


wlach (commented):
The cost to this approach is that you'd have to wait for the `create_db` target to run every time you run `make up`. I'm not sure if this is long or not?

In general, I don't really have a super strong opinion but then again I'm not developing redash much day-to-day. Are there any docs that need to be updated?

justinclift:
> The cost to this approach is that you'd have to wait for the `create_db` target to run every time you run `make up`.

Good point.  Doing some measurements now so we have actual data look at.

### Approach 1 - WITH this PR applied

Already have the redis and pg containers up and running + db seeded (ie: having run `make create_database`).

Then run `time make up` to see what happens, and note down the time.

Then run `docker compose down email server worker scheduler` to shut down the non-redis and non-pg containers.

Timing from 3 loops this way:

Loop | Time
-----|-----
1 | 18.04 secs
2 | 16.65 secs
3 | 16.19 secs

### Approach 2 - WITHOUT this PR applied

Same setup as last time, with `docker compose down` run between to shut everything down, then `make create_database`.

Use `time make up` to collect timings, and use the same `docker compose down ...` command to shut down containers between runs.

Timing from 3 loops this way:

Loop | Time
-----|-----
1 | 6.89 secs
2 | 6.31 secs
3 | 6.30 secs

---

Well, that's disappointing.  It seems like the "check if the tables are already populated" piece is adding about 10 seconds to the run each time, effectively more than doubling it.

That's really not good.

I'll see if there's a significantly faster way, otherwise it's not worth the trade off.

justinclift:
Using a different approach of manually checking for the existence of just a single table is much faster.

With the new approach, the timings are:

Loop | Time
-----|-----
1 | 6.77 secs
2 | 6.39 secs
3 | 6.20 secs

So the new approach is adding no noticeable time.  It's not as straight forward a change though.

I'll update the PR with the new approach, and lets see what people think. :smile:

justinclift:
The new approach is much faster, though it embeds `make create_database` itself to populate the tables if they're not present.  There might be a better way of doing that piece. :smile:

justinclift:
> Are there any docs that need to be updated?

Yep, the [Local Development Setup](https://github.com/getredash/redash/wiki/Local-development-setup) page can be simplified slightly if we merge this PR.

justinclift:
Hmmm, no-else else seems to care either way about this one.

In that case, I'll update it to work with the recent Makefile changes then probably merge it later on today.

PR #6674, opened 2023-12-24
reuse built frontend in ci, merge compose files

AndrewChubatiuk:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## Description
- renamed docker-compose.yml to compose.yml as it's supported by docker compose v2 by default
- merged all compose files to simplify local backend and e2e tests execution. added profiles:
   - `local `- spins up all services in a setup, that previously was in `docker-compose.yml`
   - `e2e` - spins up all services in a setup, that previously was in `.ci/docker-compose.cypress.yml`
   - `default`(no profile set) - spins up all services in a setup, that previously was in `.ci/docker-compose.yml`
- added ability to build frontend inside docker or reuse prebuilt frontend locally for CI purposes to speed up build
- added arm64 build

## How is this tested?

- [ ] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [x] Manually - executed e2e and unit tests locally against all three compose profiles
- [ ] N/A

PR #6857, opened 2024-04-07
Update to upload-artifact@v4 in CI workflows

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This PR updates our CI workflows from `actions/upload-artifact@v3` to `actions/upload-artifact@v4`.  This is due to several of the workflows showing this warning during their runs:

```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js
20: actions/upload-artifact@v3.

For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

## How is this tested?

- [x] N/A

Untested so far, that's what this PR is for. :smile:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,12 +62,12 @@ jobs:
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v3
       - name: Store Test Results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: test-results
           path: /tmp/test-results
       - name: Store Coverage Results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: coverage
           path: coverage.xml
@@ -93,7 +93,7 @@ jobs:
       - name: Run Lint
         run: yarn lint:ci
       - name: Store Test Results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: test-results
           path: /tmp/test-results
@@ -168,7 +168,7 @@ jobs:
       - name: Copy Code Coverage Results
         run: docker cp cypress:/usr/src/app/coverage ./coverage || true
       - name: Store Coverage Results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: coverage
           path: coverage


lucydodo (changes requested):
> https://github.com/actions/upload-artifact#breaking-changes
> Uploading to the same named Artifact multiple times.
> 
> Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

Starting with `v4`, artifact names cannot be duplicated. So we need to rename the artifact for each step.

justinclift:
Interesting.  That doesn't seem clear as to whether the unique artifact name is "per workflow" (eg [backend-unit-tests](https://github.com/getredash/redash/blob/82d88ed4ebbf1449771fb62284976ed2f9a80e62/.github/workflows/ci.yml#L69-L73), [frontend-e2e-tests](https://github.com/getredash/redash/blob/82d88ed4ebbf1449771fb62284976ed2f9a80e62/.github/workflows/ci.yml#L171-L174), etc), or per .yml file, or maybe something else. :roll_eyes:

That says it'll error out if there's a problem, and the CI run didn't seem to throw errors, so I wonder if we're ok?

justinclift:
Seems to be more details here:

https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

The examples there shows what seems to be different jobs having the same artifact name (as we do), but in their examples it's throwing an error.

we're probably best off to play it safe and do the rename thing.

In that case, we'll probably need to update whatever is looking for the uploaded artifacts so it finds the new ones.

lucydodo:
Based on my previous experience doing the same thing in other repo, it seems to be determined by a single action 'run'.
It other words, if the workflow are different but both are exeuted in one action 'run', they are considered duplicate.

> https://github.com/getredash/redash/actions/runs/8592014265/job/23541482569?pr=6857
> Store Test Results
> Run actions/upload-artifact@v3

Additionally, the current CI for this PR is using the 'v3' version, so it doesn't seem to be throwing any errors.

lucydodo:
> In that case, we'll probably need to update whatever is looking for the uploaded artifacts so it finds the new ones.

Yes, I'd take a look, but I don't think I have access to 'Codecov' :(
Additionally, I resolved it last time simply by adding a suffix. ex) `build-artifact` to `build-artifact-macos`

justinclift:
> Yes, I'd take a look, but I don't think I have access to 'Codecov' :(

Oh.  I think that should be fixable. :smile:

I'd not personally tried it out before, apart from looking at the "Code Coverage" report things it generates as part of our standard CI workflow.

However, clicking on one of the existing Redash code coverage "Details" links (like [this one](https://app.codecov.io/gh/getredash/redash/commit/af0773c58a0bc5db803cd2d9f839c2f8f1bc96a3)) worked.

That took me to a page to create a codecov.io (using my GitHub login), then automatically figured out the project I was coming from.

Additionally, it seems to have a way of adding extra organisations too:

image

That's the "Add GitHub Organization" link in the drop down above.

Maybe try that and see if it works? :smile:

lucydodo:
I tried, but it still says I don't have permission. :(

justinclift:
Try again now, I've just bumped up your permissions to admin level for this repo.  Lets see if recognises that or not. :smile:

justinclift:
Ugh.  That sounds like it's only going to recognise people who've been added via the main `getredash` organisation "team members" interface thing, and isn't recognising people who've been added as collaborators to specific repos. :frowning: 

I'll email @arikfr directly to get that done (there's a queue already though).  Pinging him via discord isn't working, but maybe email will. :smile:

lucydodo:
Okay, I'll be waiting :)

justinclift:
Email sent. :smile:

justinclift:
Received a response from @arikfr via email.  He said he's travelling at the moment, so will take care of this stuff in a few days.

lucydodo:
No worries, this PR is not urgent at all. :)

justinclift:
This PR isn't needed any more, as the update to `upload-artifact@v4` was included in PR #6674 which has just been merged. :smile:

PR #6858, opened 2024-04-08
Update yarn to current latest in 1.22.x series

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This is just a version bump of yarn from 1.22.19 to 1.22.22.

The [notes in the Release](https://github.com/yarnpkg/yarn/releases) for the new version indicate it contains only minor bug fixes.

This newer version has been working fine for me in other general build things, and doesn't seem to be causing any issues in our CI, Docker builds, (etc).

## How is this tested?

- [x] E2E Tests (Cypress)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ on:
       - master
 env:
   NODE_VERSION: 18
+  YARN_VERSION: 1.22.22
 jobs:
   backend-lint:
     runs-on: ubuntu-22.04
@@ -88,7 +89,7 @@ jobs:
           cache: 'yarn'
       - name: Install Dependencies
         run: |
-          npm install --global --force yarn@1.22.22
+          npm install --global --force yarn@$YARN_VERSION
           yarn cache clean && yarn --frozen-lockfile --network-concurrency 1
       - name: Run Lint
         run: yarn lint:ci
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
           cache: 'yarn'
       - name: Install Dependencies
         run: |
-          npm install --global --force yarn@1.22.19
```
lucydodo, line 91: The version value is used in multipled places, so why not put it in env?
```
env: YARN_VERSION: 1.22.19
run: npm install --global --force yarn@$YARN_VERSION
```

lucydodo (commented):
It has a good look to me, but I have a simple idea, so I have a comment. :)

.github/workflows/ci.yml
justinclift: Heh Heh Heh.  Hadn't thought of that, good idea.

Will update the PR shortly. :smile:

justinclift: Just pushed an extra commit with that change.  Seem ok to you? :smile:

lucydodo: Yes, it looks good. I've approved it. :)

lucydodo (approved):

.github/workflows/ci.yml
justinclift: Thanks heaps. :smile:

PR #6861, opened 2024-04-08
Configure Dependabot

lucydodo:
## What type of PR is this? 

- [x] Other

## Description

This configures GitHub-native Dependabot.
The existing enabled Dependabot only supports some package ecosystems,
but this PR adds support for `Docker`, `GitHub Actions`, `Poetry`, and `Yarn` package ecosystems.

## How is this tested?

- [x] Manually

Tested in my fork repo, and confirmed that it worked.

Not applicable.

Not applicable.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"


justinclift:
The thought is nice, but I'm a bit concerned about how many updates it's going to generate each week.

Have you tried it out in a local repo to see what happens first? :smile:

lucydodo:
Yes, I understand. In test, Dependabot generates at least 15 PRs.
> I say 'at least' because Dependabot warns me the scan was aborted because it couldn't generate more than 15 PRs.

Here are some workarounds:
1. Change the cycle.
We can adjust scan cycle to review PRs only once a month. But we'll have more PRs to review at one.

2. Group them.
> https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/

Since June of last year, Dependabot supports grouping PRs.
We can change this to create u pto 4 PRs grouped together for `Docker`, `GitHub Actions`, `Poetry`, adn `Yarn`.
(To be honest, I haven't tried it, I just know it's theoretically possible, if you choose option 2, I'll test in on my local first)

justinclift:
Nah.  I think for the moment we should skip this, so we can update things manually.

Dependabot already issues security warnings for stuff, and we've not been able to merge all of those yet (being worked on).

So, we'd better wait until we've merged the outstanding pieces already. :smile:

lucydodo:
That's a good idea. I'll reopen the PR later when the issues are sorted out. :)

lucydodo:
I've thought about this PR again, and I still believe that holding off on notifications for Node.js and Python is vaid decision, as it could create too many alerts. However, I think it would be a good idea to set up noficiations for Docker and GitHub Actions.

I have experience updating actions after seeing GitHub's announcement about actions that are being deprecated soon: #7266. Actions managed directly by GitHub are usually easy to track because GitHub announce it, but for other third-party actions, they only give brief notices in the CI logs without sending alerts, making them difficult to track. Any thoughts?

PR #6877, opened 2024-04-11
fixed local setup to run on ARM64

AndrewChubatiuk:
## What type of PR is this? 
- replaced postgresql image to official one, which has linux/arm64 image
- use gunicorn for a local setup (REDASH_PRODUCTION=true)
- build frontend inside docker instead of mounting it
- fix frontend build image 
- made exposed ports for DB and server configurable to avoid collision on Mac (some system service)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


compose.yaml
```diff
@@ -38,16 +38,13 @@ services:
     image: redis:7-alpine
     restart: unless-stopped
   postgres:
-    image: pgautoupgrade/pgautoupgrade:latest
+    image: postgres:16-alpine
```
justinclift, line 41: Absolutely not.

For ARM64, we can use the `pgautoupgrade/pgautoupgrade:16-dev-arm64` tag until I get multiarch builds figured out for the pgautoupgrade project.

justinclift:
For the ARM64 build of `pgautoupgrade` it doesn't have one yet purely because I've not yet learned how to do multi-arch docker builds.

I'll try and get that done today. :smile:

Dockerfile
```diff
@@ -13,7 +13,7 @@ FROM alpine:3.19 as frontend-builder-1
 COPY client/dist /frontend/client/dist
 
 # MODE 2: build static content in docker, can be used for a local development
-FROM node:18.1-bookworm as frontend-builder-2
+FROM node:20-bookworm as frontend-builder-2
```
justinclift, line 16: Is this strictly needed?

compose.yaml
AndrewChubatiuk: if i put `pgautoupgrade/pgautoupgrade:16-dev-arm64` image there it'll break setup on amd64. `pgautoupgrade/pgautoupgrade:latest` should be a tag which references both arm64 and amd64 images

Dockerfile
AndrewChubatiuk: package.json allows using node 20, so there should be no issues with this

compose.yaml
AndrewChubatiuk: what benefits are we getting from pgautoupgrade image?

justinclift: > if i put pgautoupgrade/pgautoupgrade:16-dev-arm64 image there it'll break setup on amd64.

Yep, that's why we don't yet claim to support ARM64.

> what benefits are we getting from pgautoupgrade image?

It automatically upgrades the version of PostgreSQL people are using without them needing to do it manually themselves.

It's so people don't get in the situation of being stuck on an ancient, unsupported version of PostgreSQL, as has happened previously with this project.

Dockerfile
justinclift: Does everything work correctly with Node 20 though?

I'm pretty sure @arikfr mentioned (ages ago) there were issues with some recent version of Node.

Personally, I have no idea if that's still a problem or if it's been solved already.  That's why I'm asking. :smile:

compose.yaml
AndrewChubatiuk: this postgres container is for developers, users are using either setup in getredash/setup or helm chart

justinclift: That's nice.

justinclift: Thinking about this more, I guess we can revert this to the standard official PG image for now.  When the pgautoupgrade one is multi-arch, then we can update this again and it'll upgrade smoothly anyway.

compose.yaml
```diff
@@ -38,16 +38,13 @@ services:
...
     ports:
-      - "15432:5432"
-    # The following turns the DB into less durable, but gains significant performance improvements for the tests run (x3
```
justinclift, line 44: Shouldn't we keep this comment?  It's useful information to future developers as to why the settings below are being used.

justinclift:
> made exposed ports for DB and server configurable to avoid collision on Mac (some system service)

Cool.  Sounds like a good idea. :smile:

compose.yaml
AndrewChubatiuk: committed it back

justinclift: Thanks. :smile:

compose.yaml
```diff
@@ -8,7 +8,7 @@ services:
       - postgres
       - redis
     ports:
-      - "5000:5000"
+      - "${REDASH_PORT:-5000}:5000"
```
justinclift, line 11: Good idea. :smile:

justinclift (approved):
Looks good to me now. :smile:

PR #6883, opened 2024-04-12
Change default webUI port back to 5001

justinclift:
## What type of PR is this? 

- [x] Bug Fix

## Description

This PR changes the default (tcp) port for the web user interface back to port 5001.

The recent change to port 5000 (to match an old default) turned out to be more painful than it's worth.

So, lets keep using port 5001 after all. :smile:

## How is this tested?

- [x] E2E Tests (Cypress)

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -66,7 +66,7 @@ services:
       - worker
       - scheduler
     environment:
-      CYPRESS_baseUrl: http://server:5001
+      CYPRESS_baseUrl: http://server:5000
       PERCY_TOKEN: ${PERCY_TOKEN:-""}
       PERCY_BRANCH: ${PERCY_BRANCH:-""}
       PERCY_COMMIT: ${PERCY_COMMIT:-""}
--- a/compose.yaml
+++ b/compose.yaml
@@ -8,7 +8,7 @@ services:
       - postgres
       - redis
     ports:
-      - "${REDASH_PORT:-5000}:5000"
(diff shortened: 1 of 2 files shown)


justinclift:
@eradman @AndrewChubatiuk FYI.  Am running the tests manually on my local PC, but figured we might as well run them on our CI at the same time. :smile:

justinclift:
Hmmm, this is badly failing on the Cypress tests on my local system.  Looks like I missed something.

@AndrewChubatiuk Any ideas?

justinclift:
Looks like my local Cypress testing isn't starting all of the desired workers containers.

compose.yaml
```diff
@@ -66,7 +66,7 @@ services:
       - worker
       - scheduler
     environment:
-      CYPRESS_baseUrl: http://server:5000
+      CYPRESS_baseUrl: http://server:5001
```
AndrewChubatiuk, line 69: no need to make this change. cypress will be using docker network and in docker network redash server is on port 5000

justinclift: Thanks.  I'll push a follow up commit in a few minutes.  Just want to see if the Cypress tests get further with the current commit first.  In my local testing they're now passing. :smile:

justinclift: Pushed a commit to change that back to 5000.

justinclift:
It's passing the CI tests now too, so I'm just going to merge it.

Thanks for the assist with this too @AndrewChubatiuk. :smile:

Issue #6159, opened 2023-07-10
Table Visualization distortion of big numbers on UInt64 type


Nopius:
### Issue Summary

Bug appers with clickhouse data source and big numbers with UInt64 data type. Problem is proved to be on frontend/Visualizer side, because returned XHR is in correct format and with correct meta.

### Steps to Reproduce

1. Create clickhouse data source.
2. Create query in clickhouse data source 'select 8722821678000304785'
3. Execute query with network debug turned on
4. Check network resonse:
```
{
    "query_result": {
        "id": 7932109,
        "query_hash": "38c14fd87388f3b88a159d215bfac6c8",
        "query": "select 8722821678000304785 LIMIT 1000",
        "data": {
            "columns": [
                {
                    "name": "8722821678000304785",
                    "friendly_name": "8722821678000304785",
                    "type": "integer"
                }
            ],
            "rows": [
                {
                    "8722821678000304785": 8722821678000304785
                }
            ]
        },
        "data_source_id": 67,
        "runtime": 0.4866828918457031,
        "retrieved_at": "2023-07-10T01:20:52.933Z"
    }
}
```
5. Check visualizated value (wrong): 8,722,821,678,000,305,152
![image](https://github.com/getredash/redash/assets/5144107/01978145-69d9-4504-89e6-5826c5ef23d7)

When saving in CSV/TSV the values are correct, the problem appears only in a browser.

### Technical details:

* Redash 10/Linux
* Chrome/Windows:
* How did you install Redash: docker image, runs in k8s

justinclift:
I remember running into something very like this a while back, and I *think* from rough memory it's this:

https://discuss.redash.io/t/a-numeric-column-in-sql-shows-different-value-than-it-actually-contains-in-redash-display/9587/14

justinclift:
There is a suggested way forward for fixing it too, we'd just need someone to do the implementation.  Might not even be too hard either. :smile:

justinclift:
Yeah.  We'd need to figure out a "better" approach to solve this.  It *might* be a substantial piece of work, or could turn out to not be.  No idea personally. :smile:

justinclift:
@eradman Would this be of use for solving the problem?

https://github.com/MikeMcl/bignumber.js

eradman:
The effort to solving this would be huge and there is very little benefit. People cannot read a large numbers, hence a number such as `87228216780304785` will nearly _always_ formatted in a human readable form, such as `77.47 PiB`.

[...]

PR #6888, opened 2024-04-13
Front end data type display exception

mirkan1:
This is a fix for https://github.com/getredash/redash/issues/6159

- [ ] Bug Fix

### DIFF
--- a/viz-lib/src/visualizations/table/columns/number.tsx
+++ b/viz-lib/src/visualizations/table/columns/number.tsx
@@ -36,8 +36,12 @@ export default function initNumberColumn(column: any) {
   const format = createNumberFormatter(column.numberFormat);
 
   function prepareData(row: any) {
+    let number = row[column.name];
+    if (Number.isSafeInteger(number)) {
+      number = format(number);
+    }
     return {
-      text: format(row[column.name]),
+      text: number,
     };
   }
 


viz-lib/src/visualizations/table/columns/number.tsx
```diff
@@ -36,8 +36,12 @@ export default function initNumberColumn(column: any) {
   const format = createNumberFormatter(column.numberFormat);
 
   function prepareData(row: any) {
+    let number = row[column.name];
```
gaecoli, line 39: I think move these code to `function createNumberFormatter` is a better way.

gaecoli:
Thank you very much for your contribution. To finally solve this problem, you need to introduce the bigint library. The current solution has not completely solved it. 😉

viz-lib/src/visualizations/table/columns/number.tsx
gaecoli: And our intention seems to be to display the correct numbers. 😁

mirkan1:
GPT told me something about `number.bit_length() > 63` but it is irrelevant

mirkan1:
@gaecoli why is the test is failing?

gaecoli:
> @gaecoli why is the test is failing?

https://github.com/getredash/redash/wiki/The-list-of-Redash-make-commands

Your code format does not conform to our style. You should run `make format`.  Also, your solution is not the expected solution, so I don't recommend merging your code. Moreover, you use a double-layer loop to process the data when obtaining the results, which is a very performance-consuming operation.

justinclift:
@mirkan1 Was this code written using an LLM (ChatGPT, or any other)?

mirkan1:
I wrote a version upgrade for [numeral](https://github.com/adamwdraper/Numeral-js/pull/790) yet I dont know if the maintainer will merge it any time soon...

client/app/services/query-result.js
```diff
@@ -5,6 +5,7 @@ import { QueryResultError } from "@/services/query";
 import { Auth } from "@/services/auth";
 import { isString, uniqBy, each, isNumber, includes, extend, forOwn, get } from "lodash";
 
+const JSONbigString = require("json-bigint")({ storeAsString: true });
```
gaecoli, line 8: This code can be changed `import` style? In order to unify the code style of the entire project, it is recommended to use import to import the package.

gaecoli: And change `JSONbigString` to `jsonParse`?

viz-lib/src/visualizations/table/columns/number.tsx
```diff
@@ -37,7 +37,7 @@ export default function initNumberColumn(column: any) {
 
   function prepareData(row: any) {
     return {
-      text: format(row[column.name]),
```
gaecoli, line 40: Please remove these unnecessary submissions, which may cause confusion to other developers.

client/app/services/query-result.js
```diff
@@ -4,8 +4,9 @@ import { axios } from "@/services/axios";
 import { QueryResultError } from "@/services/query";
 import { Auth } from "@/services/auth";
 import { isString, uniqBy, each, isNumber, includes, extend, forOwn, get } from "lodash";
+import JSONbig from 'json-bigint';
 
-const JSONbigString = require("json-bigint")({ storeAsString: true });
+const { parse: jsonParse } = JSONbig({ storeAsString: true });
```
gaecoli, line 9: And i notice at `json-bigint`https://github.com/sidorares/json-bigint?tab=readme-ov-file#optionsstoreasstring-boolean-default-false
Maybe we can remove `({ storeAsString: true })`.

gaecoli: If you do this, you will convert all the values ​​​​stored by users as string or other non-numeric types into big number. This operation will drive users crazy.

mirkan1: it does not work if not stored as string, I also checked the files (csv,tsv) numbers are stored as expected integers

mirkan1: I will merge Andrew's PR and try with it

mirkan1: It didnt work when I tried it using bigInt type as per your suggestion :/

image

gaecoli: > It didnt work when I tried it using bigInt type as per your suggestion :/
> 
> image

Ok, you can keep that code.

client/app/pages/queries/QuerySource.jsx
```diff
@@ -262,7 +262,7 @@ function QuerySource(props) {
                       data-executing={isQueryExecuting ? "true" : null}
                       syntax={dataSource ? dataSource.syntax : null}
                       value={query.query}
-                      schema={schema}
+                      schema={schema.length && schema[0].name ? schema : null}
```
gaecoli, line 265: You can create a new branch fix this. The one branch just solve one problem. Thank you!

gaecoli:
PTAL @justinclift

mirkan1:
@eradman would it make any diffirence to store integers as string on `api/query_results` endpoints?
I dont think if Redash doing any arithmetic equations with the results

eradman:
> @eradman would it make any diffirence to store integers as string on `api/query_results` endpoints?
> I dont think if Redash doing any arithmetic equations with the results

There would imply a complicated migration because `query_results` in the database would have a different storage format. Also numbers stored as JSON in Postgres don't loose precision, so I don't think this would solve anything.

mirkan1:
You dont make any sense, it literally solves big numbers issue if it is merged with @AndrewChubatiuk's commit

PR #6890, opened 2024-04-14
Add pydeps Makefile target for installing Python dependencies

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This combines the manual steps needed for installing the Python dependencies into a single Makefile target, simplifying the dev environment setup process a bit.

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually

This has also been manually tested to ensure the new `pydeps` Makefile target runs the same commands as the documented manual steps.


justinclift:
Once this is merged, I'll update our wiki page to use this new `make pydeps` step rather than having to manually cut-n-paste these ones:

```
(redashvenv1) $ pip3 install wheel  # "wheel" needs to be installed by itself first
(redashvenv1) $ pip3 install --upgrade black==23.12.1 ruff==0.3.5 launchpadlib pip setuptools
(redashvenv1) $ pip3 install poetry
(redashvenv1) $ poetry install --only main,all_ds,dev
```

AndrewChubatiuk:
> Once this is merged, I'll update our wiki page to use this new `make pydeps` step rather than having to manually cut-n-paste these ones:
> 
> ```
> (redashvenv1) $ pip3 install wheel  # "wheel" needs to be installed by itself first
> (redashvenv1) $ pip3 install --upgrade black==23.12.1 ruff==0.3.5 launchpadlib pip setuptools
> (redashvenv1) $ pip3 install poetry
> (redashvenv1) $ poetry install --only main,all_ds,dev
> ```

have you looked at https://python-poetry.org/docs/managing-dependencies/ for this purpose?
black and ruff are also present in pyproject and poetry files, maybe makes sense to use them for deployment, but keep them in a separate dependency group?

justinclift:
Oh, we only added the specific `black` and `ruff` versions recently, due to some change in the lint checking output of the newer version.

I *think* we could probably drop the version numbering now though, as from rough memory we've already upgraded our versions in the poetry file past those anyway.

Hmmm, I can probably check that pretty easily.  Gimme a bit of time and I'll try things out without the hard coded numbers there. :smile:

justinclift:
@AndrewChubatiuk Finally had a few spare minutes to test this properly.  We don't need the hard coded version numbers any more after all, so those have been removed. :smile:

AndrewChubatiuk (approved):
LGTM

justinclift:
Thanks @AndrewChubatiuk. :smile:

PR #6894, opened 2024-04-15
Additional Date Ranges for DateRangeParameter

rohithsebastian:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Additional dynamic date ranges for time-based filtering using the DateRangeParameter component.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Created new Date Range parameter with new query

Screenshot 2024-04-15 at 2 48 26 PM

### DIFF
--- a/client/app/components/dynamic-parameters/DateRangeParameter.jsx
+++ b/client/app/components/dynamic-parameters/DateRangeParameter.jsx
@@ -62,6 +62,14 @@ const DYNAMIC_DATE_OPTIONS = [
         .format("YYYY"),
   },
   {
+    name: "Last 2 days",
+    value: getDynamicDateRangeFromString("d_last_2_days"),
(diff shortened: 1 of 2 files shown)


gaecoli:
Thank you for your contribution. But this solution is not a universal one. I can provide you with an idea to re-develop a relative time component, or add a relative time component based on the existing one. Users can customize this relative time component. For example, Last xx days, months or years.

So I suggest not to merge this branch to master for the time being. 😁

justinclift:
> I can provide you with an idea to re-develop a relative time component, or add a relative time component based on the existing one.

That sounds like the better approach. :smile:

PR #6914, opened 2024-04-20
fixed all aws query runners auth

AndrewChubatiuk:
## What type of PR is this? 
Currently Athena supports only a limited amount of auth methods depending on ASSUME_ROLE and OPTIONAL_CREDENTIALS flags values:
 - ASSUME_ROLE: true, OPTIONAL_CREDENTIALS: true - athena client uses assume role credentials retrieved using sts boto client with [aws default credentials provider chain](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html#credentials-default)
 - ASSUME_ROLE: false, OPTIONAL_CREDENTIALS: true - use athena client with default credentials provider chain
 - ASSUME_ROLE: true, OPTIONAL_CREDENTIALS: false - same as first option
 - ASSUME_ROLE: false, OPTIONAL_CREDENTIALS: false - athena client uses static credentials
 
 These scenarios do not cover cases when:
  - there a need to use different scenarios for different athena datasources, e.g.: one uses static credentials, another one assumes role. ASSUME_ROLE and OPTIONAL_CREDENTIALS flags are global for a whole redash setup
  - when static credentials should be passed to a sts client
  
 In this PR propose do not use ASSUME_ROLE and OPTIONAL_CREDENTIALS flags at all and use the same approach for all aws query runners, which allows to cover all possible scenarios:
 - aws_access_key_id and aws_secret_access_key are set and role_arn is not set - use static credentials for service client
 - aws_access_key_id, aws_secret_access_key and role_arn are all set - use static credentials for sts client and use assume role credentials for service client
 - only role_arn is set - use default credentials provider chain for sts client and assumed role credentials for service client
 - none is set - use default credentials provider chain for service client

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -421,41 +421,44 @@ files = [
 
 [[package]]
 name = "boto3"
-version = "1.28.8"
+version = "1.34.89"
 description = "The AWS SDK for Python"
 optional = false
-python-versions = ">= 3.7"
+python-versions = ">=3.8"
 files = [
-    {file = "boto3-1.28.8-py3-none-any.whl", hash = "sha256:7132ac3f3a9c28b84dcc344cfb439d37d2c5ab45f6b577358fc9aeba5d5aab63"},
-    {file = "boto3-1.28.8.tar.gz", hash = "sha256:cf88309d9b8cd9a2fb0c8049cb4b217b4e9dcb55bf670d6054b0bbe2eef25e57"},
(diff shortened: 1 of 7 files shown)


AndrewChubatiuk:
@justinclift need your review here as well

nijave:
@AndrewChubatiuk any plans to re-open/create a new PR? I'm interested in Redshift Serverless support

PR #6927, opened 2024-04-24
Source .env when running docker containers

eradman:
## What type of PR is this? 

- [x] Bug Fix

## Description

Restore previous functionality.

## How is this tested?

- [x] Manually

1. Add setting to `.env`
2. Run `make up`

https://github.com/getredash/redash/commit/3f19534301a6b1e2a544c7305238d49e5ba60142

### DIFF
--- a/compose.base.yaml
+++ b/compose.base.yaml
@@ -20,3 +20,5 @@ services:
       REDASH_COOKIE_SECRET: ${REDASH_COOKIE_SECRET}
       REDASH_SECRET_KEY: ${REDASH_SECRET_KEY}
       REDASH_PRODUCTION: ${REDASH_PRODUCTION:-true}
+    env_file:
+      - .env
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,7 @@ jobs:
       - name: Build Docker Images
         run: |
           set -x
+          touch .env
           docker compose build
           docker compose up -d
           sleep 10
@@ -175,6 +176,7 @@ jobs:
       - name: Setup Redash Server
         run: |
           set -x
+          touch .env
           yarn build
           yarn cypress build
           yarn cypress start -- --skip-db-seed


compose.base.yaml
```diff
@@ -20,3 +20,5 @@ services:
       REDASH_COOKIE_SECRET: ${REDASH_COOKIE_SECRET}
       REDASH_SECRET_KEY: ${REDASH_SECRET_KEY}
       REDASH_PRODUCTION: ${REDASH_PRODUCTION:-true}
+    env_file:
```
AndrewChubatiuk, line 23: https://docs.docker.com/compose/environment-variables/set-environment-variables/#additional-information

please set `required: false`

AndrewChubatiuk, line 23:
```suggestion
    env_file:
      - path: .env
         required: false
```

compose.base.yaml
```diff
@@ -20,3 +20,5 @@ services:
...
+      - .env
```
AndrewChubatiuk, line 24:
```suggestion
    env_file:
      - path: .env
         required: false
```

eradman: Is there another way around this? It appears that making `env_file` optional was not available until docker-compose 2.24.0

https://docs.docker.com/compose/environment-variables/set-environment-variables/#additional-information

So making this change would introduce a new dependency.

justinclift: > docker compose 2.24 is available since january, think people had enough time to upgrade.

It's not always that simple.

We've previously had to freeze the version of docker, due to a bug (in docker) that stopped our Dockerfile from working. :frowning:

justinclift: Looking at our options though, `required: false` is probably worth trying for now and seeing if causes problems for anyone?

eradman: Well, it caused problems for me since I usually use docker-compose-v2 proved by Ubuntu 22.04 packages.

The latest version of docker-compose-v2 can, of course be installed separately, but this adds complexity to the existing setup instructions.

justinclift: No worries, looks like the `required: false` thing isn't really the way forward then.

We'd better add something (script?) which gets the .env working for the CI environment too. :smile:

eradman: I updated this PR to run ` touch .env` before building the server in `ci.yml`. I tested this in my repo, and it seemed to work.

justinclift: The concept (of `touch .env`) seems good, but I'm guessing it's somehow doing that in the wrong directory or something from the CI point of view:

```
+ docker compose build
Failed to load /home/runner/work/redash/redash/.env: open /home/runner/work/redash/redash/.env: no such file or directory
```

Might be worth adding a bunch of extra locations to the touch (`touch redash/.env redash/redash/.env /home/runner/work/redash/redash/.env`, etc) to see if it gets the CI to pass?  And if that works, then figure out which one of the paths fixed it.

Or maybe embed a `pwd` in there or something, as that might clear things up too? :smile:

AndrewChubatiuk: almost everywhere where it's possible to find docker setup instructions, it's recommended to use official repo even sometimes it's recommended to remove old docker packages before installing official ones. there's no problem to mention somewhere minimal required docker version. there should be no problems with upgrade for each user on each os

justinclift: @AndrewChubatiuk There are different schools of thought of that.  Our `setup.sh` uses the Docker CE packages, but our "Local development setup" instructions use the Ubuntu provided packages instead.

We should probably choose one and use that for both.

I'm personally more inclined to use the Docker CE ones, as we've previously seen bugs in the Ubuntu provided ones.

... though I've also seen bugs on the Docker CE ones historically as well. :facepalm:

eradman: > The concept (of `touch .env`) seems good, but I'm guessing it's somehow doing that in the wrong directory or something from the CI point of view:

No, the problem is that the change to the CI build will not take effect until this change is merged.

justinclift: Ahhh.  So we should force merge this then?  We can do that if needed.

eradman: Yes, we should skip the checks and merged this. Hard to be 100% sure, but I did test this in my own account

justinclift: Cool, all good. :smile:

PR #6933, opened 2024-04-29
trigger preview image build on tag

AndrewChubatiuk:
## What type of PR is this? 
trigger pipelines on tags

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/.github/workflows/preview-image.yml
+++ b/.github/workflows/preview-image.yml
@@ -5,8 +5,6 @@ on:
       - Tests
     types:
       - completed
-    branches:
-      - master
 
 env:
   DOCKER_REPO: redash
@@ -14,7 +12,7 @@ env:
 jobs:
   build-skip-check:
     runs-on: ubuntu-22.04
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' && (endsWith(github.event.workflow_run.head_branch, '-dev') || startsWith(github.event.workflow_run.head_branch, 'master')) }}
     outputs:
       skip: ${{ steps.skip-check.outputs.skip }}
     steps:
@@ -71,7 +69,12 @@ jobs:
         run: |
           set -x
           VERSION=$(jq -r .version package.json)
-          FULL_VERSION=${VERSION}-b${GITHUB_RUN_ID}.${GITHUB_RUN_NUMBER}
+          BRANCH=${{ event.workflow_run.head_branch }}
+          if [[ "$BRANCH" =~ ^[0-9]+\.[0-9]+\.[0-9]+\-dev ]]; then
+            FULL_VERSION=${VERSION}
+          else
+            FULL_VERSION=${VERSION}-b${GITHUB_RUN_ID}.${GITHUB_RUN_NUMBER}
+          fi
           sed -ri "s/^__version__ = ([A-Za-z0-9.-]*)'/__version__ = '${FULL_VERSION}'/" redash/__init__.py
           sed -i "s/dev/${GITHUB_SHA}/" client/app/version.json
           echo "VERSION_TAG=$FULL_VERSION" >> "$GITHUB_OUTPUT"


.github/workflows/preview-image.yml
```diff
@@ -5,16 +5,14 @@ on:
...
-      - master
 
 env:
   DOCKER_REPO: redash
 
 jobs:
   build-skip-check:
     runs-on: ubuntu-22.04
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' && (startsWith(github.ref, 'refs/tags/') || startsWith(github.ref, 'refs/heads/main')) }}
```
justinclift, line 15: `main` seems like the wrong word here, or am I misunderstanding what this is doing?

AndrewChubatiuk: oh, thanks, updated

eradman:
The format for a tagged image previous should be `YY.MM.0-dev`; this change does not appear to address that

AndrewChubatiuk:
> The format for a tagged image previous should be `YY.MM.0-dev`; this change does not appear to address that
[this workflow](https://github.com/getredash/redash/blob/master/.github/workflows/periodic-snapshot.yml) creates a tag `YY.MM.0-dev` based on cron event, and preview image workflow will build it

bin/docker-entrypoint
```diff
@@ -60,26 +60,18 @@ workers_healthcheck() {
...
+  TIMEOUT=${REDASH_GUNICORN_TIMEOUT:-60}
+  echo "Starting Redash Server"
+  exec /usr/local/bin/gunicorn \
+    -b 0.0.0.0:5000 \
+    --name redash \
+    -w${REDASH_WEB_WORKERS:-4} redash.wsgi:app \
+    --max-requests $MAX_REQUESTS \
+    --max-requests-jitter $MAX_REQUESTS_JITTER \
+    --timeout $TIMEOUT \
+    --reload
```
AndrewChubatiuk, line 74: @eradman added hotreload for gunicorn

.github/workflows/preview-image.yml
```diff
@@ -71,7 +69,12 @@ jobs:
         run: |
           set -x
           VERSION=$(jq -r .version package.json)
-          FULL_VERSION=${VERSION}-b${GITHUB_RUN_ID}.${GITHUB_RUN_NUMBER}
+          BRANCH=${{ event.workflow_run.head_branch }}
+          if [[ "$BRANCH" =~ ^[0-9]+\.[0-9]+\.[0-9]+\-dev ]]; then
+            FULL_VERSION=${VERSION}
+          else
+            FULL_VERSION=${VERSION}-b${GITHUB_RUN_ID}.${GITHUB_RUN_NUMBER}
+          fi
```
justinclift, line 77: @eradman Reading over these lines here, doesn't the `FULL_VERSION` variable end up how it used to be for the monthly snapshots?

eradman: Previously the trigger condition was based on a single condition

```yml
name: Preview Image
on:
  push:
    tags:
      - '*-dev'
```

This made the entire workflow easy to reason about. I don't know what will happen when the workflow is run; needs to be tested in a private repository.

bin/docker-entrypoint
eradman: This should be in a separate PR

.github/workflows/preview-image.yml
AndrewChubatiuk: it's needed to:
1.  prevent building image with code, that doesn't pass tests in master
2. spend less time on ARM64 images build, which are only possible to build on QEMU in github actions

bin/docker-entrypoint
justinclift: Yeah, it does seem unrelated to the purpose of this PR. :smile:

@AndrewChubatiuk Please move the unrelated stuff (like this) to a separate PR.  More than one if needed.

Mixing stuff together just slows down the reviewing. :frowning:

AndrewChubatiuk: @eradman @justinclift moved this change to https://github.com/getredash/redash/pull/6940

justinclift: Thanks @AndrewChubatiuk. :smile:

eradman:
If we are going to push a new docker image for each commit, I think we should create two different workflows:

1. trigger when tag `*-dev` is added
2. trigger when a commit is added to `master`

 This does results in duplicate configuration, but I find workflows very tedious to test properly, so the conditions need to be as simple as possible.

@justinclift what do you think?

AndrewChubatiuk:
I've tested these changes in my fork. it works as expected

.github/workflows/preview-image.yml
```diff
@@ -71,7 +69,12 @@ jobs:
         run: |
           set -x
           VERSION=$(jq -r .version package.json)
-          FULL_VERSION=${VERSION}-b${GITHUB_RUN_ID}.${GITHUB_RUN_NUMBER}
+          BRANCH=${{ event.workflow_run.head_branch }}
+          if [[ "$BRANCH" =~ ^[0-9]+\.[0-9]+\.[0-9]+\-dev ]]; then
+            FULL_VERSION=${VERSION}
+          else
+            FULL_VERSION=${VERSION}-b${GITHUB_RUN_ID}.${GITHUB_RUN_NUMBER}
```
eradman, line 76: Is this what we want? This used to be the short SHA of the commit.

AndrewChubatiuk: it's what it used to be
you can check images, that i've published during testing in a fork
https://hub.docker.com/r/achubatiuk/preview/tags

justinclift:
> If we are going to push a new docker image for each commit ...

Personally, I think we should have some kind of docker image that always has the latest and greatest development tip from the master branch.

That has the use of pointing people at it to verify a fix they reported (or need for some reason), and it's also useful for us to try out.

That being said, we also need a more stable image we can recommend to people, as we work towards an official, proper release.  The `preview` Docker image (that was previously based upon automatic monthly snapshot) seems like a good idea.  Though we should think about how we're going to stabilise that prior to making it an official release.

For that, we'll probably need to branch off of master at some near future point, test the heck out of things at that point and fix any bugs we find (applying the fixes both in that branch and in the master branch).  After a few iterations of fixes (hopefully not many), it should be a pretty good image for making official.

---

Note that I don't think we need a brand new unique docker image for every commit to master.  That's going to be horrible for people to try and figure out what image to grab, and horrible for whoever's looking after the Docker Hub repository (@arikfr presently I think).

.github/workflows/preview-image.yml
AndrewChubatiuk: it doesn't solve a problem, still emulation is required

AndrewChubatiuk:
if user wants to have a certain image, which corresponds to a git tag or just a latest image, he has such an option, but at the same time we need to allow people to rely on build they can consider stable themselves. Either monthly images, which are triggered automatically without stable changes with random content inside and without separate release branches for stability fixes or latest images is not a good option for those, who have infrastructure, which scales up and down often enough, especially taking in account amount of developers, who are able to fix Redash issues. It forces users either build images on their own if image, that was build from master or tag are broken or wait till it's fixed and hope, that it will not become broken again soon

justinclift:
Yeah.  I guess I'm thinking we need to think through how to make a stable branch in the near future, which we then test the living heck out of and eventually call it our next release. :smile:

PR #6940, opened 2024-05-01
Enable hot reload by default

AndrewChubatiuk:
## What type of PR is this? 
Enabled hot reload by default

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


AndrewChubatiuk:
@eradman could you please review?

bin/docker-entrypoint
```diff
@@ -60,26 +60,29 @@ workers_healthcheck() {
...
+      args+=("--reload")
       ;;
   esac
+  exec /usr/local/bin/gunicorn \
+    -b 0.0.0.0:5000 \
+    --name redash \
+    -w${REDASH_WEB_WORKERS:-4} redash.wsgi:app \
+    --max-requests $MAX_REQUESTS \
+    --max-requests-jitter $MAX_REQUESTS_JITTER \
+    --timeout $TIMEOUT ${args[@]}
```
eradman, line 85: Let's make "$args" a simple string instead of a BASH array

eradman (changes requested):

bin/docker-entrypoint
AndrewChubatiuk: moved `--reload` arg to compose file instead and added ability to pass command extra args to server process, which can be useful for helm chart as well

eradman:
This did not work for me. After starting with `make up`, I did not see the server reload after editing files in `redash/...`

```
docker compose logs -f server
```

AndrewChubatiuk:
@eradman changed SERVER_MOUNT to /app by default, I had it in .env. it should work right now

AndrewChubatiuk:
@eradman modified makefile to add `SERVER_MOUNT=/app` to .env

justinclift:
Interesting, the concept of the PR seems useful.

AndrewChubatiuk:
@justinclift could you check PR? almost impossible to get any feedback from @eradman

Makefile
```diff
@@ -4,6 +4,10 @@ export COMPOSE_DOCKER_CLI_BUILD=1
 export DOCKER_BUILDKIT=1
 export COMPOSE_PROFILES=local
 
+define random
+$(shell cat /dev/urandom | LC_ALL=C tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
```
justinclift, line 8: Why the change from pwgen, and why is it in this PR?

I'm pretty sure we've had the conversation before about not including extra things in PRs, as it slows down the review process. :frowning:

AndrewChubatiuk: pwgen is not a default tool, this approach should work out of box on linux and on mac
without this change i cannot test locally

justinclift: Interesting.  Well, it does seem to work ok, and the quality of randomness doesn't really make a difference for development.

The `pwgen` change should really be in its own PR though, as it's a distinct thing.

PR #5794, opened 2022-07-13
New ElasticSearch Query Runner

susodapop:
## What type of PR is this? 

- [x] New Query Runner (Data Source) 
- [x] Other

## Description

This PR incorporates the excellent work from #4391 with adjustments in #5701. I rebased from `master` and applied both patches to retain everyone's commit attributions.

These changes have been reviewed twice and approved but were never merged. Here's hoping the third time's the charm.

In addition to those commits, I have deprecated the original `elasticsearch` query runner. Existing connections will continue to exist.

## How is this tested?

- [x] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Unending thanks to @nicolaslm and @wwl717195673 for their unit and manual testing.

I also manually tested the deprecation:
1. Create an `elasticsearch` data source 
2. Apply `deprecated=True` commit and reload UI3. 
3. DS from step 1 is still visible
4. Create `elasticsearch2` database 
5. Check contents of `data_sources` table

CleanShot 2022-07-19 at 21 49 51@2x
CleanShot 2022-07-19 at 21 50 05@2x

#### Data source selection screen:

CleanShot 2022-07-19 at 21 50 24@2x

PR #6941, opened 2024-05-02
Fix Elasticsearch query runner error: 'str' object has no attribute 'pop' error when parsing query

jcowley:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Fix Elasticsearch query runner (refactored version not yet released)

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

I needed redash to work with elasticsearch 7.10, but the current v10.1.0 release does not support it. There was a major refactor of the elasticsearch query runner in #5794, but that code has not yet been released. So, I got the latest from master and tried it out.

I'm not sure how this code was tested originally, but it doesn't work because of a very minor bug. When building the elasticsearch query, it's trying to call `pop()` on the `query` parameter which is of type `str`, and there is no `pop()` method on `str`. Any attempt to run an elasticsearch query results in the error:

`'str' object has no attribute 'pop' error when parsing query`

It needs to be converted to a dictionary, which is a one line change. Then it works fine.

Not sure how this wasn't caught earlier. Maybe no one has tried it out yet? I do appreciate all the hard work that @NicolasLM, @susodapop, and @wwl717195673 put into the refactor. It's been working well for us after this fix.

#5794

### DIFF
--- a/redash/query_runner/elasticsearch2.py
+++ b/redash/query_runner/elasticsearch2.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from typing import Optional, Tuple
 
@@ -64,6 +65,7 @@ class ElasticSearch2(BaseHTTPQueryRunner):
         return data, error
 
     def _build_query(self, query: str) -> Tuple[dict, str, Optional[list]]:
+        query = json.loads(query)
         index_name = query.pop("index", "")
         result_fields = query.pop("result_fields", None)
         url = "/{}/_search".format(index_name)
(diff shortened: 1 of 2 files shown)


AndrewChubatiuk (approved):
LGTM

justinclift:
Thanks @AndrewChubatiuk. :smile:

justinclift (approved):
Looks straight forward enough, and @AndrewChubatiuk didn't spot anything wrong. :smile:

justinclift:
@jcowley Thanks for spotting this, and fixing it. :smile:

jcowley:
No problem. Thanks for getting it merged in.

PR #6944, opened 2024-05-04
Removed deprecated advocate package

AndrewChubatiuk:
## What type of PR is this? 
Advocate package is deprecated and it's features are not used in Redash. Makes sense to remove it

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A


justinclift:
> Advocate package is deprecated

Yeah.  It's not obvious from the [PyPi page](https://pypi.org/project/advocate/), but the [GitHub repo](https://github.com/JordanMilne/Advocate) for it has been archived and the readme updated to say it's no longer maintained:

> Advocate is no longer maintained, please fork and rename if you would like to continue work on it

Plus there are reports of it [causing test failures](https://stackoverflow.com/questions/76002461/where-does-attributeerror-vendoralias-object-has-no-attribute-find-spec-c) in codebases that haven't removed it yet.

So yeah, probably best to rip it out.

---

That being said, this PR seems to also be trying to remove the `ENFORCE_PRIVATE_ADDRESS_BLOCK` variable here:

https://github.com/getredash/redash/pull/6944/files#diff-b57fdcfadf5e97c9f114e4dd15d164923bc8eb5a682b39761599c4e58b304bd5

That doesn't seem correct.  Isn't that still a useful thing, that would work even without this "advocate" package?

AndrewChubatiuk:
ENFORCE_PRIVATE_ADDRESS_BLOCK enables either advocate or requests, since advocate is removed, this variable is not needed anymore

justinclift (approved):

wlach:
Hey guys, as best I can tell this removes a security feature from redash (`ENFORCE_PRIVATE_ADDRESS_BLOCK`) which some people might be depending on: I think it's intended to prevent people from using redash to scan an internal network.

justinclift:
k, this PR has been reverted.  We can remove this `advocate` package at a later point when we figure out some way to do so without losing that security feature.

PR #6945, opened 2024-05-04
replaced numeral-js with native numbers and percent formatting

AndrewChubatiuk:
## What type of PR is this? 
replaced numeral js with a native numbers formatting

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


justinclift:
With this "native numbers" approach, does that mean it'll handle numbers larger than `MAX_SAFE_INTEGER`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

AndrewChubatiuk:
> With this "native numbers" approach, does that mean it'll handle numbers larger than `MAX_SAFE_INTEGER`?
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

it parses numbers from string, it has more powerful functionality than numeral-js
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat

justinclift:
@eradman What do you reckon?  You're our resident front end expert. :smile:

justinclift:
I can't review this with my current skill set, as I don't know Typescript at all (yet), and it's not likely in the near future.

So, it'll need someone else to look over it.

PR #6683, opened 2024-01-02
Update query hash with parameters applied

eradman:
## What type of PR is this? 

- [x] Bug Fix

## Description

This allows queries with parameters to run on a schedule since the hash used to update the `query_result` will match.

Inspired by https://github.com/getredash/redash/pull/6063

## How is this tested?

- [x] Unit tests (pytest, jest)
- [x] Manually

## Related Tickets & Documents

https://github.com/getredash/redash/issues/4182

PR #6063, opened 2023-05-25
Attempt at fixing scheduled parameterized queries

remil1000:
## What type of PR is this? 

- Bug Fix

## Description

- generating query_hash with query expanded with default parameters values to match query_result with queries hash
- added default parameters related properties in Query model class

## How is this tested?

- Manually

Tested queries with parameters and without, through dashboard and not and with changing parameter for ad-hoc query run/refresh

## Extra info

The query hash is generated by removing any SQL comment and striping spaces then getting the md5 of the query text. This PR does the same only replacing parameters in braces `{{ }}` by their default value

Example below with a simple remote JSON API and parameter being `{{ n }}`, the participants number - the other is a static/no param JSON API example

```
>>> import hashlib
>>> hashlib.md5("url:http://www.boredapi.com/api/activity?participants=1".encode("utf-8")).hexdigest()
'93f7d98a4395f7cdbe1cea43947b6ff0'
>>> hashlib.md5("url:http://www.boredapi.com/api/activity?participants=2".encode("utf-8")).hexdigest()
'91716cdd4ca05899dbcf92a44828c536'

postgres=# select query, query_hash from queries; -- n = 1
                             query                              |            query_hash
----------------------------------------------------------------+----------------------------------
 url: https://api.chucknorris.io/jokes/random                   | 0616dfb448c39dd5608419df498da332
 url: http://www.boredapi.com/api/activity?participants={{ n }} | 93f7d98a4395f7cdbe1cea43947b6ff0
(2 rows)

postgres=# select query, query_hash from queries; -- n = 2
                             query                              |            query_hash
----------------------------------------------------------------+----------------------------------
 url: https://api.chucknorris.io/jokes/random                   | 0616dfb448c39dd5608419df498da332
 url: http://www.boredapi.com/api/activity?participants={{ n }} | 91716cdd4ca05899dbcf92a44828c536
(2 rows)
```

## Related Tickets & Documents

https://github.com/getredash/redash/issues/4182

PR #6960, opened 2024-05-10
Change query_hash calculation method with separate params

### LINKED ISSUE #4514: Scheduled query executiosn fail when a query uses dynamic date range parameters
### Issue Summary

Since dynamic date ranges are evaluated in the front-end, scheduled executions fail without warning or notice.

### Steps to Reproduce

1. Write a query with a date range parameter
2. Configure the date range parameter to use a dynamic value
3. Execute the query and note its last execution time
4. Schedule the query to execute within a reasonable time (60 seconds e.g.)
5. After the time allotted, observe that the last executed time has not updated

I expect the interface to warn about this behavior. Ideally, we can perform the calculation in the back-end so that scheduled executions can work.

### Technical details:

* Redash Version: v9 alpha
* Browser/OS: FF / Chrome / Safari

ehooi:
## What type of PR is this? 

Change query_hash calculation method.

Example
- Query: `select c1, c2 from table where c3="{{ key }}"`
- Parameter: key=value
- Apply Auto Limit

Before:
```
selectc1,c2fromtablewherec3="value"LIMIT1000
```

After:
```
selectc1,c2fromtablewherec3="{{key}}"
{"auto_limit":true,"parameters":{"key":"value"}}
```

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

I want to fundamentally fix #4514 #6063 #6683
I believe the problem lies in the dynamic date(-range) parameters being evaluated on the frontend side.
Before evaluating dynamic params on the backend side, we need to separate query text and parameters.

This is a breaking change. All query hashes will be changed.
You will need to re-save and execute all scheduled queries once to be scheduled properly afterward.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

#4514 #6063 #6683


justinclift:
> This is a breaking change. All query hashes will be changed.
> You will need to re-save and execute all scheduled queries once to be scheduled properly afterward.

Reckon this could be made automatic with a migration of some sort?

arikfr:
Either migration or maybe some period of checking both hashes (old and new).

eradman:
> Either migration or maybe some period of checking both hashes (old and new).

I can't think of how a database migration would do this correctly. The technique I have been using is to use a script to POST an empty payload to force the server to recalculate hashes:

```py
for query in queries:
    # Empty payload will still cause the query has to be re-calculated with default parameters applied
    updated_query = redash.post(f"/api/queries/{query['id']}", data={})
```

justinclift:
Hmmm, this sounds like a pain, but perhaps a necessary one.

Migration wise, hmmm.  Seems like it need to have some part of it Python driven (as @eradman mentions) as the database can't drive it purely using database functions.

Does our Python migration framework not provide some existing capability for this?

ehooi:
I found a migration file. `1038c2174f5d_make_case_insensitive_hash_of_query_text.py`
It seems to be something similar to what we needed.
Plus, I realized that I broke that migration file. Haha.

I wrote the migration code. It has no downgrade code.
Because the dynamic date(-range) parameters were lost.
This is the root cause of the problem I am trying to fix.

PR #6969, opened 2024-05-14
Revert changes to job status

eradman:
## What type of PR is this? 

- [x] Revert

## Description

Revert in order to restore progressive job status

"Query in queue" should switch to "Executing query", but does not.

Commands used

```
git revert --no-commit bd17662005248c087a8d6aeadb866ae94d3d3465
git revert --no-commit 5ac5d86f5e4ae0adb0c296e53f95cb79e612237a
vim tests/handlers/test_query_results.py
git add tests/handlers/test_query_results.py
```

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually

https://github.com/getredash/redash/discussions/6964


redash/serializers/__init__.py
```diff
@@ -272,33 +271,46 @@ def serialize(self):
 
 
 def serialize_job(job):
+    # TODO: this is mapping to the old Job class statuses. Need to update the client side and remove this
```
justinclift, line 274: > Need to update the client side and remove this

Is this meaning a follow up PR/commit, or was it more a note to yourself while going through it, or something else? :smile:

justinclift:
@gaecoli Do you have time to look over this one, just to double check it's ok? :smile:

redash/serializers/__init__.py
justinclift: No worries. :smile:

gaecoli (approved):
It's ok.

eradman:
Thanks for the review @gaecoli 

The motive behind the commit this change reverts may have been good, but it wasn't tested well.

gaecoli:
For major changes, we should introduce two or three people to review to avoid reverting code.

PR #6991, opened 2024-05-28
Update to Python 3.10

ezraodio1:
## What type of PR is this? 

- [x] Refactor

## Description

Updated from Python 3.8 to 3.10. Python 3.10 is the default for Ubuntu 22. This change necessitated upgrading to SQLAlchemy_Utils 0.38.3, and importing the sort_query function from an older version of SQLAlchemy_Utils because it was dropped in newer versions.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


justinclift:
The rough idea we've sort of had is that we'd stick with Python 3.8 while it's not deprecated.

That being said, perhaps we should revisit that idea?

(Personally I don't develop using Ubuntu any more either. :wink:)

redash/utils/query_order.py
```diff
@@ -0,0 +1,310 @@
+# Copyright (c) 2012, Konsta Vesterinen
+
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
```
eradman, line 6: Include `#` on blank lines as well. Example:

```
# Copyright (c) 2012, Konsta Vesterinen
#
# All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions are met:
# ...
```

ezraodio1: Done

eradman:
I have tested this change extensively, and it is working well. The bulk of the work here is to update to sqlalchemy-utils, which a lot of work because they dropped features that Redash requires.

@justinclift would you like to take the update to Python 3.10 as a separate PR?

justinclift:
> would you like to take the update to Python 3.10 as a separate PR?

U guess you're meaning to break this PR into to, one for the sqlalchemy-utils update and one for stuff specifically to upgrade Python to 3.10?

---

Was thinking about this a bit more the other day.  I guess my real concern about updating to Python 3.10 is whether that will create issues for any of the data source libraries we're using?

I'm guessing "probably not", but I'd like to make sure it doesn't.  Or at least make sure we fix any problems it'd introduce.

@arikfr probably has thoughts about it too, as it'll have project wide ramifications.

If it turns out completely fine, then I don't have any objections to moving to Python 3.10. :smile:

eradman:
> Was thinking about this a bit more the other day. I guess my real concern about updating to Python 3.10 is whether that will create issues for any of the data source libraries we're using?
> 
> I'm guessing "probably not", but I'd like to make sure it doesn't. Or at least make sure we fix any problems it'd introduce.

At Starfish Storage we have been running Redash patched to run with Python 3.10 for about two years--in fact we have to for FIPS compliance. It could be that one of the data source libraries does not work properly with 3.10, but we have not encountered this yet.

justinclift:
Sounds like it's probably fairly safe then. :smile:

eradman (approved):

eradman:
Conflicts for `poetry.lock` are relentless! @ezraodio1 can you rebase this change again?

eradman (approved):

justinclift:
Merged before the poetry lock file changes again. :smile:

Issue #6988, opened 2024-05-24
Getting tables fails if Redash DB user has no permissions to a schema in PG database


kfrydel:
### Issue Summary

The query executed by this function: https://github.com/getredash/redash/blob/master/redash/query_runner/pg.py#L206 fails with `ERROR:  permission denied for schema XYZ` if the DB user used to access PostgreSQL data source has no permissions to access some schema and the schema contains a partitioned table. It fails on `has_table_privilege(s.nspname || '.' || c.relname, 'select')`.
I suppose that the same happens for materialized views and foreign tables because of `c.relkind IN ('m', 'f', 'p')` condition.

Adding `AND has_schema_privilege(s.nspname, 'usage')` condition to `JOIN pg_namespace` should fix this issue because then the query will not try to check permissions for particular tables if the user cannot access the whole schema.

Stacktrace from logs:
```
[2024-05-23 01:00:26,875][PID:3615879][INFO][rq.job.redash.tasks.queries.maintenance] job.func_name=redash.tasks.queries.maintenance.refresh_schema job.id=766d55ee-b9c4-459f-ba56-c032b8321656 task=refresh_schema state=start ds_id=2
[2024-05-23 01:00:26,959][PID:3615879][ERROR][redash.query_runner] permission denied for schema XYZ

[2024-05-23 01:00:26,959][PID:3615879][WARNING][rq.job.redash.tasks.queries.maintenance] job.func_name=redash.tasks.queries.maintenance.refresh_schema job.id=766d55ee-b9c4-459f-ba56-c032b8321656 Failed refreshing schema for the data source: MYDATABASE
Traceback (most recent call last):
  File "/some/path/src/redash/tasks/queries/maintenance.py", line 166, in refresh_schema
    ds.get_schema(refresh=True)
  File "/some/path/src/redash/models/__init__.py", line 215, in get_schema
    schema = query_runner.get_schema(get_stats=refresh)
  File "/some/path/src/redash/query_runner/__init__.py", line 272, in get_schema
    self._get_tables(schema_dict)
  File "/some/path/src/redash/query_runner/pg.py", line 246, in _get_tables
    self._get_definitions(schema, query)
  File "/some/path/src/redash/query_runner/pg.py", line 202, in _get_definitions
    self._handle_run_query_error(error)
  File "/some/path/src/redash/query_runner/__init__.py", line 239, in _handle_run_query_error
    raise Exception(f"Error during query execution. Reason: {error}")
[...]

PR #7000, opened 2024-06-05
PG: Only list tables where schema has USAGE permission

eradman:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

This covers cases where partitioned tables are part of a schema that is not accessible by the current user.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Create a new schema then set up a data source for Redash with another, non-superuser role

```
CREATE SCHEMA xyz;

CREATE TABLE xyz.tab (
   id bigint GENERATED ALWAYS AS IDENTITY,
   ts timestamp NOT NULL
) PARTITION BY LIST ((ts::date));

CREATE TABLE xyz.tab_default PARTITION OF xyz.tab DEFAULT;
```

https://github.com/getredash/redash/issues/6988

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -231,7 +231,9 @@ class PostgreSQL(BaseSQLQueryRunner):
         ON a.attrelid = c.oid
         AND a.attnum > 0
         AND NOT a.attisdropped
-        WHERE c.relkind IN ('m', 'f', 'p') AND has_table_privilege(s.nspname || '.' || c.relname, 'select')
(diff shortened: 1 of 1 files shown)


justinclift (approved):
Eyeballing this, it makes good sense to me. :smile:

eradman:
Thanks @justinclift for the second set of eyeballs! Could not figure out how to write a meaningful test for this, so had to rely on manual testing.

Issue #6996, opened 2024-05-30
Chart doesn't show values


denisov-vlad:
### Issue Summary

Chart values always equals 0 on hover.

### Steps to Reproduce

1. Run SQL query:
```
SELECT 5 as atc_count, 7 as qty_users
UNION ALL SELECT 8,1
UNION ALL SELECT 1,290
UNION ALL SELECT 4,11
UNION ALL SELECT 0,763
UNION ALL SELECT 7,1
UNION ALL SELECT 3,25
UNION ALL SELECT 6,4
UNION ALL SELECT 2,68
```
2. Build any chart (i.e. bar) with atc_count on X, qty_users on Y
3. Move pointer to the second bar (X=1)
4. See value = 0

<img width="756" alt="redash_screen" src="https://github.com/getredash/redash/assets/7091907/d0de8687-7753-42fd-a97a-e4e22854ea50">

### Technical details:

* Redash Version: ec051a89398a0c2483ed56a4773509af15bc03f4
* Browser/OS: Chrome 125 / Windows 11
* How did you install Redash: Docker compose (image built from source)

eradman:
@denisov-vlad can you confirm that this problem is resolved in the latest checkout?

PR #6908, opened 2024-04-18
Chart Aggregate Y column values rather than displaying last Y value

lvitti:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## Description
Resubmit the fix created by @akc2267 in the [PR](https://github.com/getredash/redash/pull/4673).

A couple fixes so that Y values are summed. Without this fix, charts will display the final row for every X value. If multiple rows exist for a given X, then those values will be over-written. Plotly displays this info correctly, so without this fix, redash's hover labels are incorrect while Plotly's visual representation is accurate.

## How is this tested?

- [x] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

## Related Tickets & Documents
https://github.com/getredash/redash/issues/4587
https://github.com/getredash/redash/pull/4673

PR #7002, opened 2024-06-07
Fix chart labels

denisov-vlad:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Fixed labels for charts. Before that, all labels have values = 0.

The error was that `Object` in JS has keys as strings, so `Object.keys(labelsValuesDict);` returned wrong array.

I don't have any experience with JS, so please review it carefully.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

#6996

image


eradman (changes requested):
This change has 17 commits, it should only be 1. (Error trying to rebase?)

eradman:
@lvitti if you have time please test this change as well. We want to keep the functionality you implemented in https://github.com/getredash/redash/pull/6908 if possible.

lvitti:
Hi @denisov-vlad, @eradman sorry for that bug, yes look like this fix is working also for Pie aggregation 
I'm not an expert in javascript either, here is another implementation of the fix using Map() 
https://github.com/getredash/redash/compare/master...lvitti:redash:lvitti/fix_6996_chart_doesnt_show_values?expand=1

denisov-vlad:
@eradman Are there any blockers to merge it?

lvitti:
> Yes, this is a very "hot" code path and testing for adverse side-effects has proven to require more care and the addition of unit tests.
> 
> I proposed that we revert the change that caused this issue in #7018

And what is the plan to solve the other bug? Because it's as bad as this.

Why not just add a test to this case and consider the 2 fixes ?

eradman:
> And what is the plan to solve the other bug? Because it's as bad as this.
> 
> Why not just add a test to this case and consider the 2 fixes ?

The bug with the pie charts should be fixed as well, but this is a regression.

ezraodio1:
I've been playing around with this feature and have noticed a few things.

1. If you recreate this bar chart with the issue and click Edit Visualization -> X Axis -> Scale -> Category, the labels are correct.
2. This issue with the labels is not limited to bar charts - I recreated it with the pie chart. As @denisov-vlad mentioned, the error is because the dictionary forces keys to be strings. Therefore, if your choice for X Axis isn't a column of strings, then the labels will be incorrect.
3. Using @lvitti's implementation with Map() doesn't seem to break anything, which logically makes sense since it isn't forcing a data type on its keys. All of the charts I made with the Map() implementation were correct, and I also **didn't** need to manually set the X Axis scale to "Category."


https://github.com/getredash/redash/assets/97199317/383e637e-b39c-48e9-b8da-09024d59523e

eradman:
We merged an alternate fix in https://github.com/getredash/redash/pull/7022

PR #7025, opened 2024-06-19
Add new text pattern parameter

ezraodio1:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
This PR adds in a new "Text Pattern" parameter input type. This input type allows a creator to specify a regex pattern, and the query will only run if the input matches the regex pattern. Since this parameter input type is not "Text," users can share dashboards that have this parameter input type, therefore addressing (https://github.com/getredash/redash/discussions/6785).

## How is this tested?

- [x] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

(https://github.com/getredash/redash/discussions/6785)

https://github.com/getredash/redash/assets/97199317/e644fd75-634c-49f8-809e-9af6dc792720

### DIFF
--- a/redash/models/parameterized_query.py
+++ b/redash/models/parameterized_query.py
@@ -1,3 +1,4 @@
+import re
 from functools import partial
 from numbers import Number
 
@@ -88,6 +89,16 @@ def _is_number(string):
         return True
 
 
+def _is_regex_pattern(value, regex):
+    try:
+        if re.compile(regex).fullmatch(value):
+            return True
+        else:
+            return False
+    except re.error:
+        return False
+
+
 def _is_date(string):
     parse(string)
     return True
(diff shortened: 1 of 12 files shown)


redash/models/parameterized_query.py
```diff
@@ -88,6 +89,16 @@ def _is_number(string):
         return True
 
 
+def _is_regex_pattern(value, regex):
+    try:
+        if re.compile(regex).match(value):
```
eradman, line 94: Let's use [fullmatch()](https://docs.python.org/3/library/re.html#re.fullmatch) instead so that the pattern has to cover the entire string.

eradman (approved):
Demonstrated this functionality to some colleges, more more issues found

justinclift:
> more more

"No more"?

eradman:
> "No more"?

Ah yes, all seems to work well. The previous revision received two 👍 so I also take that as a vote of confidence for the design.

justinclift:
Cool, all good. :smile:

Issue #5682, opened 2022-01-10
Athena Data Source: support Catalog


VasilyFomin:
PyAthena added support for Catalogs in https://github.com/laughingman7743/PyAthena/issues/220 and it'd be great if redash also supported them.

What I think needs to be done:

- [ ] Update `athena.py` to include `catalog_name` https://github.com/getredash/redash/blob/f21f7e211ff5fe12c850a9627cfadb80a5a8819e/redash/query_runner/athena.py#L221-L229
- [ ] Update UI to include a new text field to enter a catalog name
- [ ] Persist the new field in DB

* Redash version: 10.0

susodapop:
I agree this looks like a good addition. Since we don't use Athena on the core team we need someone in the community to write and verify the changes. But we've updated our documentation about [writing a query runner](https://redash.io/help/open-source/dev-guide/write-a-query-runner) which should make this straightforward. In summary:

1. Add `catalog_name` to the configuration schema. This will automatically add it to the data source connection screen and persist it to the database.
2. Update the `pyathena.connect()` call to use the value of `self.configuration["catalog_name"]`.
3. Establish a default behaviour so that existing Athena data sources won't break since a `catalog_name` is not specified. 

If you'd like to give it a try I'm happy to help with any questions. And look forward to reviewing and merging this soon.

yongchand:
Is anyone working on this issue

susodapop:
Awesome @yongchand thank you! Nobody else has picked this up. Feel free to ping me with any questions along the way.

yongchand:
@susodapop Just out of curiosity, https://github.com/getredash/redash/pull/5741#issue-1213704982 isnt this issue related?

susodapop:
🤦 You're correct.

If that PR merges will it meet your need? Can you check out the change and try it (cautiously)?

yongchand:
@susodapop Sure. I can run on docker and see if it works. But where should we insert this extra_options?

susodapop:
You insert the options on the data source setup screen in settings. That pull request modifies the query runner `configuration_schema`, which controls what options are displayed there.

yongchand:
@susodapop sorry again, is there any instructions for testing codes using docker? (Not using pre-existing image)

susodapop:
Yes, you can run our Docker development devloop: https://redash.io/help/open-source/dev-guide/docker

The only difference is here you will need to checkout that pull request's code, instead of `master`.

[...]

PR #7059, opened 2024-07-14
Support Arbitrary Catalog IDs on Athena Data Source

dtaniwaki:
## What type of PR is this? 

- [x] Feature

## Description

Currently, a schema list on an Athena data source is only based on the default catalog. If we have an Athena role which is allowed to use glue catalogs of cross accounts, we can still query them by the runner, but no schema of the other catalogs are listed on the UI.

## How is this tested?

- [x] Manually

I tested the following test cases.

| case | result |
| --- | --- |
| No catalog ID specified | The schema list of the default catalog are displayed. |
| 1 catalog ID specified | The schema list of the specified catalog are displayed. |
| 2 catalog IDs specified | The merged schema list of 2 catalogs are displayed.  |

And confirmed that I can see expected schema on the UI.

It's related to #5682

redash


redash/query_runner/athena.py
```diff
@@ -76,6 +77,10 @@ def configuration_schema(cls):
                     "default": "default",
                 },
                 "glue": {"type": "boolean", "title": "Use Glue Data Catalog"},
+                "catalog_ids": {
+                    "type": "string",
+                    "title": "Comma separated Glue Data Catalog IDs (Defaults to the default catalog)",
```
dtaniwaki, line 82: Please tell me more proper message for this field.

justinclift:
The concept for this seems fine.  I'm not personally sure of better wording for that text string through, as I'm not familiar with Athena myself.

@eradman @lucydodo @wlach @gaecoli Don't supposr any of you are familiar with the Athena data source type? :smile:

lucydodo:
The text looks good enough as it, but I think it could be made more explicit by changing it to something like this:
`Enter Glue Data Catalog IDs, separated by commas (leave blank for default catalog)`

dtaniwaki:
I updated the text. Thank you!

lucydodo:
LGTM. Thanks :)
@justinclift The final merge decision is left to you.

lucydodo (approved):

justinclift (approved):
Looks good to me.  Thanks for contributing this @dtaniwaki. :smile:

Issue #6836, opened 2024-03-28
Downgrading then upgrading redash schemas doesn't work


wlach:
### Issue Summary

After updating to latest main (1672cd92801b2555b6a95a1c95e7b3328f0664d0) and then downgrading to a previous version, upgrading again doesn't work unless you do the following:

```
alter table queries alter schedule drop default;
alter table queries alter options drop default;
alter table events alter additional_properties drop default;
alter table organizations alter settings drop default;
alter table alerts alter options drop default;
alter table dashboards alter options drop default;
alter table dashboards alter layout drop default;
alter table changes alter change drop default;
alter table visualizations alter options drop default;
alter table widgets alter options drop default;
```

The problem seems to be that [this migration](https://github.com/getredash/redash/blob/master/migrations/versions/7205816877ec_change_type_of_json_fields_from_varchar_.py) sets a default on downgrade, where none existed before. When you try to add a json schema to that, redash aborts.

Additional gotcha:

The "schedule" column gets casted to the `"null"` string on downgrade. You need to do a manual:

```
update queries set schedule=null where schedule='null';
```

Looks like the migrations modified in the following PRs are the culprits:

https://github.com/getredash/redash/pull/6687
https://github.com/getredash/redash/pull/6707
https://github.com/getredash/redash/pull/6713

/cc @AndrewChubatiuk

AndrewChubatiuk:
@wlach thanks for a finding
could you please submit PR with a fix?

wlach:
> @wlach thanks for a finding could you please submit PR with a fix?

Will see what I can do.

justinclift:
Oh, this all doesn't sound good.

Breaking the schema when downgrading is something we'd better fix.

The rough plan is to enable automatic schema upgrades when Redash starts (rather than a manual process).  So being able to downgrade stuff when needed for some reason is probably pretty important.

wlach:
Yeah, it's a bit unfortunate -- I assume the alembic db migrations were automatically generated, and were just wrong in this case. I think we need to verify (via `pg_dump` or similar) that the schema is as you'd expect after an arbitrary number of upgrades and downgrades.

FWIW my testing setup is to do this to downgrade:

```
./manage.py db downgrade fd4fc850d7ea
```

And this to upgrade: 

```
./manage.py db upgrade
```

wlach:
> > @wlach thanks for a finding could you please submit PR with a fix?
> 
> Will see what I can do.

[...]

PR #7068, opened 2024-07-18
Remove defaults set during schema upgrade/downgrade

ezraodio1:
## What type of PR is this? 

- [x] Bug Fix

## Description

This PR addresses https://github.com/getredash/redash/issues/6836. The fix removes the defaults that are set during schema upgrade/downgrade. 

It looks like defaults are set in redash/models/__init__.py, but it seems like they never actually make it to the database. This could be due to how setting the default parameter works in SQLAlchemy.  

## How is this tested?

- [x] Manually

Ran `./manage.py db downgrade fd4fc850d7ea` and then `./manage.py db upgrade` just like @wlach did in https://github.com/getredash/redash/issues/6836. The upgrade and downgrade seem to work correctly after making the fix.

https://github.com/getredash/redash/issues/6836

### DIFF
--- a/migrations/versions/7205816877ec_change_type_of_json_fields_from_varchar_.py
+++ b/migrations/versions/7205816877ec_change_type_of_json_fields_from_varchar_.py
@@ -24,56 +24,56 @@ def upgrade():
         type_=JSONB(astext_type=sa.Text()),
         nullable=True,
         postgresql_using='options::jsonb',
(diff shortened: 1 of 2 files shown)


eradman (approved):
Confirmed that `downgrade fd4fc850d7ea` / `upgrade` works as expected

justinclift:
@ezraodio1 This is pretty awesome, thanks for getting that problem fixed. :smile:

PR #7076, opened 2024-07-24
Add min/max/first selector for alerts

ezraodio1:
## What type of PR is this? 

- [x] Feature

## Description
This feature allows the user to create an alert that checks the minimum, maximum, or first value of a column when evaluating. 

Currently, alerts will only check the first value of a column when evaluating.

This change makes alerts more useful and allows users to set up new kinds of conditions to test for.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

Screenshot 2024-07-24 at 3 17 51 PM


eradman:
Very nice feature!

We should also add a new variable called `ALERT_SELECTOR` so that this option can be rendered in an alert email template.

See `render_template()` at https://github.com/getredash/redash/blob/a69f7fb2fe41af23eec99b5c5d92d6ee6044c532/redash/models/__init__.py#L992-L1000

ezraodio1:
> rendered in an alert email template



> Very nice feature!
> 
> We should also add a new variable called `ALERT_SELECTOR` so that this option can be rendered in an alert email template.
> 
> See `render_template()` at
> 
> https://github.com/getredash/redash/blob/a69f7fb2fe41af23eec99b5c5d92d6ee6044c532/redash/models/__init__.py#L992-L1000

Good call. Just added the selector to the template.

client/app/components/proptypes.js
```diff
@@ -65,6 +65,7 @@ export const Query = PropTypes.shape({
 
 export const AlertOptions = PropTypes.shape({
   column: PropTypes.string,
+  selector: PropTypes.oneOf(["max", "min", "first"]),
```
eradman, line 68: Let's change the order to `["first", "min", "max"]`

ezraodio1: Done

client/app/pages/alert/components/Criteria.jsx
```diff
@@ -54,23 +54,70 @@ export default function Criteria({ columnNames, resultValues, alertOptions, onCh
...
+          >
+            <Select.Option value="max" label="max">
+              max
+            </Select.Option>
+            <Select.Option value="min" label="min">
+              min
+            </Select.Option>
+            <Select.Option value="first" label="first">
+              first
+            </Select.Option>
```
eradman, line 105: One more nitpick: order these tags: first, min, max

ezraodio1: Done

eradman (approved):

PR #7084, opened 2024-07-29
Workaround broken setuptools v72

justinclift:
## What type of PR is this? 

- [x] Bug Fix

## Description

Due to a change in Python's setuptools v72 (today?), the `compose_build` make target is currently broken:

```
ModuleNotFoundError: No module named 'setuptools.command.test'
```

This PR (based upon [this](https://github.com/pypa/setuptools/issues/4519#issuecomment-2255446798)) gets the build working again, though may not be required if the new version of setuptools is pulled or otherwise fixed.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

This PR gets the compose build process working again, but hasn't been tested any further than that (yet).

https://github.com/pypa/setuptools/issues/4519

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,6 +81,9 @@ RUN if [ "$TARGETPLATFORM" = "linux/amd64" ]; then \
 
 WORKDIR /app
 
+RUN echo "setuptools<72" > /app/constraints.txt
+ENV PIP_CONSTRAINT=/app/constraints.txt
+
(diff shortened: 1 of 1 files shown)


justinclift:
The Python `setuptools` project has fixed the problem in their [newly released version `72.1.0`](https://pypi.org/project/setuptools/#history), so this PR isn't needed any more. :smile:

PR #7098, opened 2024-08-01
Alert evaluation bugfix + add average selector to alerts

ezraodio1:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
This is an addition to #7076. This PR adds the ability to select the average value from a column when making an alert.

This PR also makes the alert evaluate to UNKNOWN when min, max, or avg cannot be computed for a column. For example, this will happen when the column data type is a string, and those strings can't be converted to numbers.

This PR also deletes a print statement that was accidentally left in #7076.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

#7076


client/app/components/proptypes.js
```diff
@@ -65,7 +65,7 @@ export const Query = PropTypes.shape({
 
 export const AlertOptions = PropTypes.shape({
   column: PropTypes.string,
-  selector: PropTypes.oneOf(["first", "min", "max"]),
+  selector: PropTypes.oneOf(["first", "min", "max", "average"]),
```
eradman, line 68: let's use "avg" for consistency (or change the other options to "minimum", "maximum", ...)

client/app/pages/alert/components/Criteria.jsx
```diff
@@ -103,6 +112,9 @@ export default function Criteria({ columnNames, resultValues, alertOptions, onCh
             <Select.Option value="max" label="max">
               max
             </Select.Option>
+            <Select.Option value="average" label="average">
```
eradman, line 115: use "avg"

eradman (changes requested):

redash/models/__init__.py
```diff
@@ -966,18 +966,27 @@ def evaluate(self):
...
+                    value = min_val
+                elif selector == "avg":
+                    total_sum = 0
+                    for i in range(len(data["rows"])):
+                        total_sum += data["rows"][i][self.options["column"]]
+                    value = total_sum / len(data["rows"])
+                else:
+                    value = data["rows"][0][self.options["column"]]
+
+            except Exception:
```
eradman, line 988: Do we know what kinds of exceptions might be raised? If so we should catch them (`TypeError` for example)

ezraodio1: I updated the code a little. Now, when there's an exception, it will be a ValueError. I changed line 988 accordingly

ezraodio1:
> @ezraodio1 can you submit the bug fixes as a separate PR? I may not be able to review this feature for another week or so

Done. The bug fix part is in #7103. Once that fix is implemented, I'll change this PR so that it only adds the ability to calculate the average.

PR #7099, opened 2024-08-01
Get rid of the strange looking 0 following "Running..." and "runtime"

zachliu:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

In certain situations such as querying parquet data using Athena, it is possible for the query engine to scan 0 byte of data because of the presence of metadata. Currently when this happens, there is a strange "0" showing up in the UI:

* When query is running:

* When query finishes running:

This is because the 

```jsx
        {queryResultData.metadata.data_scanned && (
          <span className="m-l-5">
            Data Scanned
            <strong>{prettySize(queryResultData.metadata.data_scanned)}</strong>
          </span>
        )}
```

block ends up being the "0".

After applying the code changes in the PR:

* When query is running:

* When query finishes running:

The `queryResultData.metadata.data_scanned &&` is unnecessary since the `prettySize` is able to deal with null values

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [x] N/A

See attached screenshots above.


client/app/pages/queries/components/QueryExecutionMetadata.jsx
```diff
@@ -67,10 +68,9 @@ export default function QueryExecutionMetadata({
           )}
           {isQueryExecuting && <span>Running&hellip;</span>}
         </span>
-        {queryResultData.metadata.data_scanned && (
```
arikfr, line 70: I believe the reason for having this conditional is to prevent showing "Data scanned" for data sources that don't support it, which is actually most data sources Redash supports.

How about we change this conditional to something like:

```
isDefined(queryResultData.metadata.data_scanned) && 
```

Instead of just assuming it's a boolean? This way if it's a zero it will show and if it wasn't defined, it will be ignored.

zachliu: when query is executing, you'd still see

354339802-a30be3ce-c1ab-474d-97b2-132a2b74fa99
and
354339839-bf471cc5-4d42-4039-ba23-476cf48011bc

because it is defined, and it's 0

arikfr: If it's defined and it's 0, with the updated version it will show "Data Scanned 0 bytes", no? Kind of same behavior as you would see with your suggested change, but only for data sources that actually define data_scanned.

zachliu: here is what i tried (couldn't find a `isDefined` so i used `!isUndefined`)

```jsx
import {isUndefined} from "lodash";

{!isUndefined(queryResultData.metadata.data_scanned) && (
  <span className="m-l-5">
    Data Scanned <strong>{prettySize(queryResultData.metadata.data_scanned)}</strong>
  </span>
)}
```

the answer is yes and no, there are 2 states here
1. when query execution is done, you do see the correct message `244 rows 8 seconds runtime Data Scanned 0 bytes`
2. when query is still executing, you'll see `244 rows Running… Data Scanned 0 bytes` which is a bit confusing because the query is still running

but i do see your point, with only the `!isQueryExecuting`, when "Data scanned" is not supported, we are seeing `3 rows 4 seconds runtimeData Scanned ?`

so i think we should do both 
```jsx
!isUndefined(queryResultData.metadata.data_scanned) && !isQueryExecuting
```

arikfr (approved):
👍

PR #7103, opened 2024-08-02
Fix alert evaluation logic and issue with calculating min and max for alerts

ezraodio1:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Currently, if `op(value, threshold)` is not True, the alert evaluates to OK_STATE. This happens in situations where you have op = <, value = "test", and threshold = 4. Since this isn't a logical evaluation, I propose that we evaluate the alert to UNKNOWN_STATE in situations like these.

Additionally, when calculating the minimum and maximum of a column, there is no check for whether the column values are numbers. This PR adds that check in and evaluates the alert to UNKNOWN_STATE if the data in the column cannot be converted to numbers (meaning we can't logically evaluate the minimum or maximum of the column).

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

#7076

### DIFF
--- a/client/app/pages/alert/components/Criteria.jsx
+++ b/client/app/pages/alert/components/Criteria.jsx
@@ -67,7 +67,9 @@ export default function Criteria({ columnNames, resultValues, alertOptions, onCh
       <small className="alert-criteria-hint">
         Max column value is{" "}
         <code className="p-0">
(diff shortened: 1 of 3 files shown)


eradman (approved):
This looks right, and seems to check out in testing

PR #7108, opened 2024-08-05
Add rq dashboard for local development

zachliu:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Add the RQ dashboard in docker compose for better local development.

rq-dashboard-UI

rq-dashboard-workers

rq-dashboard-finished

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [x] N/A


justinclift:
This seems like a really good idea.

We should add some useful mention/docs about it in appropriate places too.  eg maybe a new wiki page to start with, and then add something to the website

---

As this is a significant item / change instead of just approving and merging it I'd like to get the opinion of others as well.  @eradman, what do you reckon? :smile:

eradman:
Doesn't this provide the same functionality as the "RQ Status" Admin page? If there is missing functionality, I think I would rather see the built-in Status page improved.

admin-rq

zachliu:
I consider it as a "stand-in replacement" while the "RQ Status" page is being improved 
It's just that i don't see anyone is actively working on improving that page :stuck_out_tongue:

PR #7109, opened 2024-08-05
Add data type to redshift query runner

zachliu:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Redshift system table [SVV_COLUMNS](https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_COLUMNS.html) supports `data_type`. We should make good use of it since the existing postgresql runner ([pg.py](https://github.com/getredash/redash/blob/master/redash/query_runner/pg.py)) has been updated to accommodate data type since 2020 (see https://github.com/getredash/redash/pull/5052). 

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


justinclift (approved):
Looks good to me. :smile:

PR #7111, opened 2024-08-06
Fix issue with scheduled queries

ezraodio1:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

There's currently an issue in Redash where some queries fail to refresh on their schedule. This happens when the query has no `latest_query_data`, which makes `retrieved_at = None`, which makes it so the query never gets added to `outdated_queries`. This PR fixes that.

The problem can be reproduced by running the unit test that's part of this PR (without making the fix first, of course)

```
def test_enqueues_scheduled_query_without_latest_query_data(self):
        """
        Queries with a schedule but no latest_query_data will still be reported by Query.outdated_queries()
        """
        query = self.factory.create_query(
            schedule=self.schedule(interval="60"),
            data_source=self.factory.create_data_source(),
        )

        outdated_queries = models.Query.outdated_queries()
        self.assertEqual(query.latest_query_data, None)
        self.assertEqual(len(outdated_queries), 1)
        self.assertIn(query, outdated_queries)
```

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


eradman (approved):
Tested this manually, seems to check out

justinclift:
@arikfr Any idea if this will interact with the scheduled queries hash issue you were looking at yesterday?

redash/models/__init__.py
```diff
@@ -389,6 +389,8 @@ def groups(self):
 def should_schedule_next(previous_iteration, now, interval, time=None, day_of_week=None, failures=0):
     # if time exists then interval > 23 hours (82800s)
     # if day_of_week exists then interval > 6 days (518400s)
+    if previous_iteration is None:
+        return False
```
arikfr, line 393: Why not return True here instead of the `or not retrieved_at` in the calling function?

redash/models/__init__.py
```diff
@@ -608,17 +610,23 @@ def outdated_queries(cls):
                     if schedule_until <= now:
                         continue
 
+                if all(value is None for value in query.schedule.values()):
+                    continue
```
arikfr, line 614: What's this for?

ezraodio1: It's been a little while since I looked at this, but I think this is the answer:

The only way to check whether a query doesn't have a refresh schedule is to ensure that everything in `query.schedule.values()` is None. For example, interval, until, time are all keys in the `query.schedule`. It's possible to have a query refresh schedule when just one of these keys is not None.

If a query doesn't have a refresh schedule, then we can just continue to the next query in the loop.

redash/models/__init__.py
ezraodio1: I don't remember there being a specific reason. I think both ways are logically equivalent.

redash/models/__init__.py
arikfr: When a query does not have a schedule then the schedule value is null (or empty). It's possible there is some edge case scenario that might keep schedule object with empty values, but I would assume it's not common?

To keep the code maintainable I would rather avoid this change in this PR and if you still feel it's necessary let's bring it back in a separate one and have the discussion there (and potentially add tests for this case).

redash/models/__init__.py
arikfr: While it's true that both will have the same result, it does make the code more confusing: `should_schedule_next` returns `False`, but we override it.. it will be more readable to have it return True which is what we want in this case anyway.

arikfr (commented):
One last review. Please let us know if you have the bandwidth to apply the changes. Thanks!

ezraodio1:
I lost access to the VM that I was using over the summer for Redash dev work unfortunately, so I think it's easiest if someone else applies the changes.

redash/models/__init__.py
ezraodio1: I think this is worth double checking. I could be mistaken, but I remember this edge case popping up during my testing. Perhaps if a query had a schedule and then that schedule was later removed?

justinclift:
@ezraodio1 If you need a new VM for doing redash dev work on, then I'm happy to spin one up for you.  Would that be useful? :smile:

ezraodio1:
I'm taking a tough course load right, so I'm pretty busy. I'd definitely be interested after this semester ends, though. I'll try to spin up a new VM for myself.

justinclift:
No worries.  Keep it in mind as an option if/when it'd be useful.

Good luck with your course load, hopefully it all goes really well. :smile:

redash/models/__init__.py
justinclift: Sounds like we'll need to create a follow up PR to this one if/when we merge this. :smile:

justinclift:
@arikfr Should we merge this PR as-is, or do you want to modify it a bit first?  Asking because @ezraodio1 doesn't have the bandwidth to make changes at present.

redash/models/__init__.py
arikfr: I found out that we have an explicit test case for this (schedule object with all values being None). I don't know if this is to cover some case that won't happen anymore, but just to be on the safe side I kept this logic + added a comment to explain.

justinclift:
@arikfr Should this be merged now? :smile:

justinclift (approved):
Eric already approved the previous revision of this. :smile:

PR #7115, opened 2024-08-07
Consolidate query execution logs

zachliu:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Duplicated log entries are removed. All information (in-memory result size, query runtime, etc.) is consolidated under the `_log_progress()` method.

Another benefit of consolidated log entries is that when using tools such as DataDog, custom metrics can be easily generated from parsed logs:

dd

Because `username` and `query_id` are now tied with `data_length` and `run_time` by putting them in the same log entry.
Much easier to track rogue users who like to do `SELECT * FROM big_table` or any other inefficient queries.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A


mirkan1 (approved):

justinclift:
The concept of this sounds good.  I'll have to look at the implementation later on after some other tasks though, if no-one else gets to it first.

@lucasfcnunes Is this your kind of thing to review the technical details of? :smile:

lucasfcnunes:
😆 I'll take a look

zachliu:
@lucasfcnunes @justinclift gently nudge :wink:

lucasfcnunes (approved):
LGTM!

zachliu:
@lucasfcnunes thanks! :pray:

PR #7122, opened 2024-08-15
Fix mismatched poetry version

zachliu:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

The poetry version specified in the `Dockerfile` doesn't match with the one that's generated `poetry.lock`.

https://github.com/getredash/redash/blob/159a329e26fdeb3a7105169a020719912b2bf39c/Dockerfile#L87

https://github.com/getredash/redash/blob/159a329e26fdeb3a7105169a020719912b2bf39c/poetry.lock#L1

This may make the build much less stable especially if your Github Actions runner is ARM-based:
* The build may still be successful (in my case, the chance is ~25%)
* The build is likely to hang indefinitely

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A


justinclift (approved):
Oh, very good find!  Well done. :grin:

PR #7126, opened 2024-08-20
Bump python-rapidjson to 1.20

justinclift:
## What type of PR is this? 

- [x] Other

## Description

This new version is a drop in replacement for the old one, but with some improvements.

Version 1.14 and above also seems to include macOS ARM64 support, at least according to the changelog: https://github.com/python-rapidjson/python-rapidjson/blob/master/CHANGES.rst

## How is this tested?

- [x] E2E Tests (Cypress)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3793,12 +3793,84 @@ cli = ["click (>=5.0)"]
 
 [[package]]
 name = "python-rapidjson"
-version = "1.1"
+version = "1.20"
 description = "Python wrapper around rapidjson"
 optional = false
 python-versions = ">=3.6"
 files = [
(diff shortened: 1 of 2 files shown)


justinclift:
@lucydodo The macOS/ARM64 support in more recent versions of this *might* be useful. :smile:

lucydodo (approved):

lucydodo:
Looks good to me, Since you've asked two people to review, I'll leave the final merge decision to @eradman :)

justinclift:
Oh, I was just doing that in case *either* of you were around and ok to take a look at it. :smile:

lucydodo:
Ah, I see, let's merge them then. @eradman If you have other thoughts, please let me know. :)

lucydodo:
CI failed twice and succeeded on the last attempt.
The frontend e2e checks were failing, so I suspect that this PR was not the cause, but rather some other transient error.

justinclift:
Yeah, there seemed to be an infrastructure problem with an earlier test too.

Issue #7127, opened 2024-08-20
Vertical scroll in query results is broken


denisov-vlad:
### Issue Summary

Currently, on the latest commit we cannot scroll down in query results.

### Steps to Reproduce

1. Run query with more than 25 rows in result
2. Save it and try to scroll down

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* Redash Version: #[58a7438](https://github.com/getredash/redash/commit/58a7438cc8054f603f5b061c9864f7855c86496a)
* Browser/OS: Chrome 127.0.6533.120 / Windows 11
* How did you install Redash: self-hosted via docker-compose

thiborgesnvs:
We are facing the same problem 

On developers tool we can see the  overflow-x  frame is being showed as hidden

If you manually change this option on developer tool on chrome the scroll down will appear 

I have no idea how to fix on the code

justinclift:
Ouch, that doesn't sound good at all.

@denisov-vlad It's possible my update of rapidjson broke it yesterday *somehow*, though that's kind of surprising.  Any idea if the commit on `master` before yesterday's one has the same problem?

@thiborgesnvs Which version of Redash are you using?

denisov-vlad:
I suppose this PR has broken it: https://github.com/getredash/redash/pull/7019

We switched to version before that, and everything works fine. Also, that PR was the reason why I decided to update Redash :)

justinclift:
Thanks @denisov-vlad. :smile:

eradman:
I am working on a fix for this regression, I will put up a pull request for review tomorrow.

justinclift:
Thanks heaps @eradman. :smile:

justinclift:
@denisov-vlad We've just merged @eradman's PR which reverts that commit.  That should at least solve the immediate problem and we can follow up later for a better approach to improve fixed column layouts. :smile:

PR #7019, opened 2024-06-14
Adding ability to fix table columns in place

ezraodio1:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

## Description
This change involved adding an extra option to the GridSettings editor, adding the "fixed" option to columns, and adding styling for the fixed columns. In order to change the number of fixed columns, which will default to 0, one has to go to Edit visualization -> Grid -> Choose number of columns to fix -> Save.

## How is this tested?

- [ ] Unit tests (pytest, jest)
- [x] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

https://github.com/getredash/redash/assets/97199317/c8401f1c-95d4-4454-894c-2425742fae68

PR #7128, opened 2024-08-22
Improve fixed-column layout

eradman:
## What type of PR is this? 

- [x] Bug Fix

## Description

Only set `scroll.x` when more than one fixed column is set to preserve text wrapping.  Unset `.ant-table-content` `overflow` property to ensure virtical scrolling works.

## How is this tested?

- [x] Manually

https://github.com/getredash/redash/issues/7127

https://github.com/user-attachments/assets/a0e674c7-ed30-4ce4-a0da-b5a8bc2949f7

### DIFF
--- a/viz-lib/src/visualizations/table/Renderer.tsx
+++ b/viz-lib/src/visualizations/table/Renderer.tsx
@@ -141,7 +141,7 @@ export default function Renderer({ options, data }: any) {
           showSizeChanger: false,
         }}
         showSorterTooltip={false}
-        scroll = {{x : 'max-content'}}
+        scroll={ (get(options, "fixedColumns", 0) > 0 ? { x : 'max-content' } : {}) }
       />
     </div>
   );
--- a/viz-lib/src/visualizations/table/renderer.less
+++ b/viz-lib/src/visualizations/table/renderer.less
@@ -140,6 +140,10 @@
       .ant-table {
         flex-grow: 1;
       }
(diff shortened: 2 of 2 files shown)


eradman:
The E2E failure was caused by a change in widget size. Investigating...

e2e-error

eradman:
Latest failure

```
CypressError: Timed out retrying after 20000ms: `cy.scrollTo()` failed because this element is not scrollable:

`<div class="ant-table-content" style="overflow: auto hidden;">...</div>`

Make sure you're targeting the correct element or use `{ensureScrollable: false}` to disable the scrollable check.

https://on.cypress.io/scrollto
```

The inline style seems to be set by antd. Overriding this with a stylesheet works, but Cyrpess doesn't like it

justinclift:
Should we instead revert #7019, and look for a different way to implement it that doesn't break vertical scrolling?

eradman:
> Should we instead revert #7019, and look for a different way to implement it that doesn't break vertical scrolling?

We might have to. It's possible that this behavior was fixed/changed in a later version of antd.

I proposed a revert commit in https://github.com/getredash/redash/pull/7131

eradman:
The fixed column featurew as reverted in #7131, next step is to propose this feature as a new PR

mirkan1:
@arikfr what they understand fix is only revert or unnecessary line enhancements

PR #7144, opened 2024-09-09
Update compose.yaml, commented volume i got error after removing it resolved 

govindrao92:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -5,8 +5,8 @@ x-redash-service: &redash-service
     context: .
     args:
       skip_frontend_build: "true"  # set to empty string to build
-  volumes:
-    - .:/app
+  #volumes:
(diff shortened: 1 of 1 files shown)


govindrao92:
updated the compose file and removed volumes, but I got an error due to the volume.

justinclift:
Heh Heh Heh

That volume line is there on purpose, as it helps make the frontend pieces available in the docker container(s).

What was the error you had, as we should probably get that solved instead? :smile:

PR #7166, opened 2024-09-18
Automatic database upgrades

zachliu:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A


bin/docker-entrypoint
```diff
@@ -42,6 +42,10 @@ dev_worker() {
 }
 
 server() {
+  echo "Upgrading database..."
+
```
justinclift, line 46: Probably doesn't need the blank line between the echo and the manage.py command.

zachliu: there is one blank line between `echo` and `exec` in `scheduler()`, `dev_scheduler()`, `worker()`,  `dev_worker()`. thought I should just keep up with the neighbors to be consistent :grin:

justinclift: Oops, my bad. 🤦

justinclift: Timing wise, @arikfr is working on a patch to fix the generation of hash queries, and I *think* might also be adjusting one of the migrations so it takes the auto limit into account as well.

That should turn up in the next few days hopefully, and I'd personally like for us to get that in before this.

zachliu: :+1:

PR #7167, opened 2024-09-18
wait for the db init & upgrade

zachliu:
## What type of PR is this? 

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

For https://github.com/getredash/redash/pull/7166

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,7 @@ jobs:
           set -x
           yarn cypress build
           yarn cypress start -- --skip-db-seed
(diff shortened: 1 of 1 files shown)


justinclift:
k, I guess we can try this and see how it goes with #7166.  I'm not personally a fan of using sleep statements, but we can figure out something better later on if it turns out to be needed.

Issue #6074, opened 2023-06-02
Force Google oauth scheme https


fedeostrit:
hello, I am not able to configure redash by https with google oauth, it is currently installed with helm chart in AWS eks with http and in front there is an aws NLB, when trying to pass the NLB to https (it performs the ssl/tcp termination) when it comes to When browsing the URL, redash always tries to redirect to http, I don't know if there is something that can be done like the option that already exists for REDASH_SAML_SCHEME_OVERRIDE so that flask can build the URLs with Https correctly and that it will have a good time with google oauth

fedeostrit:
Hi @arikfr, It is possible that the team can review this of the change of schema from https to http when there is an AWS NLB as a proxy in front of the Redash deployed with helm chart, I don't know if there is already a parameter that can be used to allow this and not rewrite the URIs in http

PR #7178, opened 2024-09-26
Add support for Google OAuth Scheme Override

konstde00:
## What type of PR is this? 

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

This pull request introduces a new configuration setting, `GOOGLE_OAUTH_SCHEME_OVERRIDE`, which allows overriding the URL scheme used in Google OAuth redirects. This change ensures that the correct scheme (https) is used in environments where SSL termination occurs at a load balancer or proxy, and the connection to Redash is over HTTP 

### Problem Addressed
In deployments where Redash is behind a load balancer or reverse proxy that handles SSL termination (e.g., AWS Network Load Balancer), the application may generate URLs with the http scheme. This happens because Flask constructs URLs based on the incoming request, which is over HTTP from the proxy to the application server.

This behavior causes issues with Google OAuth authentication, as Google requires the redirect URIs to match exactly with the ones registered in the Google API Console. If the application generates a redirect URI with http, but Google expects https, the authentication flow fails.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [X] Manually
- [ ] N/A

Seems to work fine in my local development environment.

Resolves https://github.com/getredash/redash/issues/6074


amozh-op:
Looks good 👍  We would greatly benefit from this change!

nakashima-koki:
This Pull Request is great :star_struck: 
I think the re:dash team should take note of this Pull Request

redash/authentication/google_oauth.py
```diff
@@ -29,6 +29,48 @@ def verify_profile(org, profile):
...
+            )
+        else:
+            next_path = url_for(
+                "redash.index",
+                org_slug=session.get("org_slug"),
+                _external=True,
+            )
+    return next_path
+
+
```
yoshiokatsuneo, line 66: @konstde00 

Thank you very much for your PR !
I think this PR makes redash more useful.

For about the code, this PR duplicates `url_for` function call with parameters, and I think it is better to avoid the duplication to make code maintainable.

I'm not so sure what is the best solution, but for example, how about avoid the duplication like below ? (also for other part calling `url_for`) 

```python
def build_next_path():
    next_path = request.args.get("next")
    if not next_path:
        scheme = None
        if settings.GOOGLE_OAUTH_SCHEME_OVERRIDE:
            scheme = settings.GOOGLE_OAUTH_SCHEME_OVERRIDE

        next_path = url_for(
            "redash.index",
            org_slug=session.get("org_slug"),
            _external=True,
            _scheme=scheme,
        )
    return next_path
```

Thank you !

konstde00: Great suggestion, I have fixed it in https://github.com/getredash/redash/pull/7178/commits/6fd75077408f9f46830eb60a6b6e6f43f158541e
Thank you!

yoshiokatsuneo (approved):

yoshiokatsuneo:
@konstde00 

Thank you!
Now, I tested and confirmed that REDASH_GOOGLE_OAUTH_SCHEME_OVERRIDE settings change the schema for Google OAuth.

I approved and merged your PR.

Thank you very much for your contribution !

PR #7186, opened 2024-10-09
Fix RQ wrongly moving jobs to FailedJobRegistry

thiagogds:
## What type of PR is this? 

- [x] Bug Fix

## Description
Something changed in the newer version of python-rq, and because Redash was overwriting parts of it I started to see that if a job ran for longer than 2 minutes, it would be automatically set as failed but continue running until it completed as success.

It fails with the error: `Moved to FailedJobRegistry, due to AbandonedJobError`

This causes a problem in the UI because it is as if the job failed, so the query results won't load.

I updated the code with the new version of rq and kept the changes from Redash. Given the changes in the code, it looks like rq changed how it sends a heartbeat for running queries.

## How is this tested?

- [x] Manually

To be able to reproduce locally I needed to:

- Create a second worker
- Create 2 queries, one that would take more than 2 minutes and a faster one - I used PostgreSQL data source for the tests
- While the long query was running I ran the faster query at around 1:46min and then the problem showed.

After the code changes it's now working.

- https://github.com/inloco/redash/commit/a3fda7bc12f750657bc21d33b4ddf6de7e2b452f#diff-cccbf90f8fc09a5b0a3deb015a6772e3705026e8688df9af470d79bbf09678ebR116
- https://github.com/rq/rq/issues/1507

### DIFF
--- a/redash/tasks/worker.py
+++ b/redash/tasks/worker.py
@@ -6,7 +6,7 @@ import sys
 from rq import Queue as BaseQueue
 from rq.job import Job as BaseJob
 from rq.job import JobStatus
-from rq.timeouts import HorseMonitorTimeoutException, UnixSignalDeathPenalty
+from rq.timeouts import HorseMonitorTimeoutException
 from rq.utils import utcnow
 from rq.worker import (
     HerokuWorker,  # HerokuWorker implements graceful shutdown on SIGTERM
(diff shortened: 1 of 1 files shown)


eradman:
@thiagogds can you rebase this change? We fixed the restyled app error by replacing it with a github action.

thiagogds:
@eradman done :)

eradman (approved):
I have been manually testing this change, and it appears to be good.
It seems that to reproduce this problem multiple queries must be run simultaneously.

thiagogds:
@eradman Cool! Just to add there, I've been running a fork of Redash with this fix for a few weeks and so far it's good and I don't have the problem anymore :)

eradman:
Merged. Thanks @thiagogds

PR #7197, opened 2024-10-18
Add migration to set default alert selector

eradman:
## What type of PR is this? 

- [x] Migration

## Description

In commits fc1e1f7a and e44fcdb a new Selector option was added to alerts, which may be "first", "min" or "max".  This migration sets the default to "first" for existing alerts.

## How is this tested?

- [x] Manually

```
docker compose exec server ./manage.py db downgrade 9e8c841d1a30
docker compose exec server ./manage.py db upgrade
```


eradman:
Closing in favor of #7475

PR #7202, opened 2024-10-22
Support result reuse in Athena data sources

dtaniwaki:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Currently, redash makes a query everytime the query page opens and it costs a lot if the query is too heavy.

On the other hand, there is a feature that efficiently uses query results in Athena.
https://docs.aws.amazon.com/athena/latest/ug/reusing-query-results.html

pyathena (>=2.18.0) supports the option.
https://github.com/laughingman7743/PyAthena/blob/6b8f0e94abd09115ba1277616ca84372e2f15d56/pyathena/common.py#L112-L113
It seems the reason of [updating the major version](https://github.com/laughingman7743/PyAthena/releases/tag/v2.0.0) is dropping python versions. and as far as I tried, queries work fine.

We may want to turn on/off the option per query but it will break the compatibility with the other data sources, so I made it a data source option. Could you consider adding this feature?

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Run an Athena query with this option with manually container image.

N/A

Screenshot_2024-10-23_at_1_49_15


pyproject.toml
```diff
@@ -116,7 +116,7 @@ pandas = "1.3.4"
 phoenixdb = "0.7"
 pinotdb = ">=0.4.5"
 protobuf = "3.20.2"
-pyathena = ">=1.5.0,<=1.11.5"
+pyathena = "2.25.2"
```
dtaniwaki, line 119: pyathena 3.x requires python >= 3.8.1 which conflicts with python >= 3.8 of the redash requirements, so I decided to use 2.x.

dtaniwaki:
@lucydodo @justinclift Could you consider merging this feature?

lucydodo:
It looks good, but since the version of the `pyathena` package goes up quite a bit,
so we should provbably look into the possibility of other side effectes before merging.

dtaniwaki:
@lucydodo What else can I do to get this reviewed and merged?

lucydodo:
Sorry, I've been busy for a while and haven't had time to review it. I'll review and respond this weekend.

lucydodo (approved):

lucydodo:
I have reviewed the changelog from version 1.11.5 to 2.25.2 and found no changes that could cause issues.
Therefore, I approve and merge this PR. @dtaniwaki Thanks for contribution. :)

dtaniwaki:
@lucydodo Thank you!

PR #7209, opened 2024-10-28
Update Azure Data Explorer query runner to latest version

spacentropy:

- [x] Refactor

## Description

Hello. Here's the list of things that were improved:
* upgraded Azure Data Explorer(Kusto) library to the latest version
* added Materialized Views to schema viewer
* added Managed Identities support for secrets-less auth
* added query user annotation in request description
* added data_scanned
* added simple test to get_schema
* added types to schema viewer

## How is this tested?

- [x] Manually


redash/query_runner/azure_kusto.py
```diff
@@ -91,18 +119,38 @@ def name(cls):
...
-            app_key=self.configuration["azure_ad_client_secret"],
-            authority_id=self.configuration["azure_ad_tenant_id"],
-        )
+        cluster = self.configuration["cluster"]
+        msi = self.configuration.get("msi", False)
+        # Managed Service Identity(MSI)
+        if msi:
+            kcsb = KustoConnectionStringBuilder.with_aad_managed_service_identity_authentication(cluster)
+        # Specific user assigned Managed Service Identity
+        elif msi and self.configuration["user_msi"]:
```
arikfr, line 128: I think this won't be called because if `msi` is `True` then it will always stop with the first `if`. You need to have this come before the plain `if msi:`.

redash/query_runner/azure_kusto.py
```diff
@@ -91,18 +119,38 @@ def name(cls):
...
+            kcsb = KustoConnectionStringBuilder.with_aad_managed_service_identity_authentication(
+                cluster,
+                client_id=self.configuration["user_msi"],
+            )
+        # Service Principal auth
+        else:
+            kcsb = KustoConnectionStringBuilder.with_aad_application_key_authentication(
```
arikfr, line 135: Now that this is optional in case `msi` is `True` we should confirm that all 3 fields have values in this case.

spacentropy:
@arikfr Hello, I'm ready to merge this!


I had to fix numpy to 1.24.4 because it auto-updates to 2.* and this breaks everything for some reason (https://github.com/numpy/numpy/issues/26710)

spacentropy:
Hello @arikfr, just a friendly reminder about this :-) Thanks

PR #7222, opened 2024-11-18
Only evaluate the next state if there's a valid value

### LINKED ISSUE #5618: TypeError: float() argument must be a string or a number, not 'NoneType' in `redash.models.next_state`
### Issue Summary

Using a self-hosted redash instance running in docker-compose with image `redash/redash:10.0.0.b50363`. The instance was upgraded from redash 8, with celery replaced with rq according to the release notes.

Every once in a while redash emits this traceback to sentry/logs:

```
TypeError: float() argument must be a string or a number, not 'NoneType'
  File "rq/worker.py", line 934, in perform_job
    rv = job.perform()
  File "rq/job.py", line 686, in perform
  File "rq/job.py", line 692, in _execute
    return self.func(*self.args, **self.kwargs)
  File "redash/tasks/alerts.py", line 42, in check_alerts_for_query
    new_state = alert.evaluate()
  File "redash/models/__init__.py", line 1012, in evaluate
    new_state = next_state(op, value, threshold)
  File "redash/models/__init__.py", line 945, in next_state
    value = float(value)
```

Some of the values Sentry catches are:

- `RedashQueue('default')`

- `Alert(id=25, name='<redacted>', query_id=606, user_id=3, state='ok', last_triggered_at=datetime.datetime(2021, 10, 8, 8, 34, 47, 531995, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=180, name=None)), rearm=10800)`

- `Query(id=606, name='<redacted>', query_hash='c53ed27f20858a88b18604ba5e803fdf', version=1, user_id=3, org_id=1, data_source_id=1, query_hash='c53ed27f20858a88b18604ba5e803fdf', last_modified_by_id=3, is_archived=False, is_draft=False, schedule={'interval': 1800, 'until': None, 'day_of_week': None, 'time': None}, schedule_failures=0)`

- `data` is

```
{columns: [{friendly_name: 'col1', name: 'col1', type: 'integer'}, {friendly_name: 'col2', name: 'col2', type: 'integer'}], rows: [{col1: None, col2: None}]}
```

and the alert is on `col1 > 0`. It looks like redash 8 handled `col1 = None` (in which case, not triggering the alert), but redash 10 doesn't.

### Steps to Reproduce

I admittedly didn't try it, but it would probably be:

- Create a query `col1, col2` which returns `NULL, NULL`.
- Create alert on `col1 > 0`.
- Alert check should cause the traceback.

### Technical details:

* Redash Version: 10.0.0.b50363
* Browser/OS: Firefox/Linux
* How did you install Redash: docker-compose

thiagogds:
## What type of PR is this? 

- [x] Bug Fix

## Description

I've experience this on my Redash in production. I'm not sure what can cause the value to exist, but be None. I guess it depends on the SQL query.

I followed the same idea of returning a self.UNKNOWN_STATE for cases that we can't know what's happening.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

Fixes #5618

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -969,6 +969,7 @@ class Alert(TimestampMixin, BelongsToOrgMixin, db.Model):
 
     def evaluate(self):
         data = self.query_rel.latest_query_data.data if self.query_rel.latest_query_data else None
+        new_state = self.UNKNOWN_STATE
 
         if data and data["rows"] and self.options["column"] in data["rows"][0]:
             op = OPERATORS.get(self.options["op"], lambda v, t: False)
@@ -997,9 +998,8 @@ class Alert(TimestampMixin, BelongsToOrgMixin, db.Model):
 
             threshold = self.options["value"]
 
-            new_state = next_state(op, value, threshold)
-        else:
-            new_state = self.UNKNOWN_STATE
+            if value is not None:
+                new_state = next_state(op, value, threshold)
 
         return new_state
 
--- a/tests/models/test_alerts.py
+++ b/tests/models/test_alerts.py
@@ -118,6 +118,10 @@ class TestAlertEvaluate(BaseTestCase):
         )
         self.assertEqual(alert.evaluate(), Alert.UNKNOWN_STATE)
 
+    def test_evaluate_return_unknown_when_value_is_none(self):
+        alert = self.create_alert(get_results(None))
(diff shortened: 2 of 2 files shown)


eradman:
Thanks @thiagogds, try re-basing this change to pick up a fix to the restyled workflow

eradman (approved):
Confirmed that this the new test catches the incorrect state (`NoneType`)

PR #7252, opened 2024-12-18
BigQuery: add date, datetime type mapping

yoshiokatsuneo:
## What type of PR is this? 

- [*] Bug Fix

## Description

On BigQuery, DATE and DATETIME type are not mapped to specific type, so it sometimes handled as "string" type and shown as number.

This PR add mapping for DATETIME/DATE in addition to TIMESTAMP.

## How is this tested?

- [*] Manually

I just run the query below and confirmed that the column is displayed in date format instead of number.

```
select DATETIME("2022-01-01") as t
union all 
select NULL as t
```


arikfr (approved):
Thanks!

PR #7255, opened 2024-12-19
Use headers instead of URL params for ClickHouse credentials

martijnthe:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other (Security/Improvement)

## Description

The ClickHouse driver was passing the credentials via URL query parameters.
This not a good practice because URLs tend to get included in logs.
This PR changes this to use the `x-clickhouse-user` and `x-clickhouse-key` headers instead.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Add a ClickHouse DB as datasource. Observe connecting and querying still works as before.

N/A

N/A

### DIFF
--- a/redash/query_runner/clickhouse.py
+++ b/redash/query_runner/clickhouse.py
@@ -99,10 +99,13 @@ class ClickHouse(BaseSQLQueryRunner):
         timeout = self.configuration.get("timeout", 30)
 
         params = {
-            "user": self.configuration.get("user", "default"),
-            "password": self.configuration.get("password", ""),
             "database": self.configuration["dbname"],
             "default_format": "JSON",
+        }
+
+        headers = {
+            "x-clickhouse-user": self.configuration.get("user", "default"),
+            "x-clickhouse-key": self.configuration.get("password", ""),
         }
 
         if session_id:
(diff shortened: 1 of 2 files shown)


martijnthe:
@susodapop could you perhaps take a look?

susodapop (approved):
I haven't contributed to redash in a few years but this PR looks safe to me. It's consistent with the [clickhouse docs](https://clickhouse.com/docs/en/interfaces/http#default-database) and should thus be safe (i.e. not niche).

Be sure to update the release notes to let Clickhouse users know about this change but I don't imagine this will be a breaking change for Clickhouse users.

martijnthe:
> Be sure to update the release notes to let Clickhouse users know about this change but I don't imagine this will be a breaking change for Clickhouse users.

@susodapop where should I update the release notes? The CHANGELOG file seems very outdated.

I did find release notes [here](https://github.com/getredash/redash/wiki/Release-Notes), but I'm not sure if I'm supposed to edit that?

Can you merge this PR? I cannot, even though you approved it already.

martijnthe:
> where should I update the release notes? The CHANGELOG file seems very outdated.

@arikfr could you perhaps provide guidance? Thank you very much!

PR #7260, opened 2024-12-24
Use fixed environment for TestAlertRenderTemplate

eradman:
## What type of PR is this? 

- [x] Tests

## Description

Other tests and local settings sometimes caused this test to fail.

## How is this tested?

### DIFF
--- a/tests/models/test_alerts.py
+++ b/tests/models/test_alerts.py
@@ -1,6 +1,7 @@
 import textwrap
 from unittest import TestCase
 
+from redash import settings
 from redash.models import OPERATORS, Alert, db, next_state
 from tests import BaseTestCase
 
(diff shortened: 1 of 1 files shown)


eradman:
@yoshiokatsuneo if you have time, can you review this change? `TestAlertRenderTemplate` often fails in my dev environment because of settings in `.env`

yoshiokatsuneo:
@eradman 

Thank you for improving test environment.

I wonder whether it is good idea to change/reset settings on every tests, as it may make test code a bit messy...
How about using separate .env file for testing ( like .env_test ) ?

yoshiokatsuneo:
Or, possibly, we can statically setup settings on the base TestCase class ?

eradman:
> How about using separate .env file for testing ( like .env_test ) ?

This is not a bad idea: there are probably other test cases that would fail with modified settings.

eradman:
Instead of sanitizing the environment, I modified the `utils.base_url()`  to prevent it from returning a hard-coded protocol (`https://`) when `MULTI_ORG` is enabled. As near as I can tell this is a bug, and not noticed by anyone.

eradman:
Thanks for the review again @yoshiokatsuneo!
`MULTI_ORG` has been an experimental feature, so most users are not likely to hit this code path.

PR #7264, opened 2024-12-28
python: Show line number when an exception occurs

elro444:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description
When exceptions occurred in queries of the Python data source, you could not see where the error happened.  
This PR fixes this and adds the line number where the exception was raised from.

Since the user will care about exceptions in their code, the shown line number will be the "deepest" entry in the stack trace that is from the user's code (from the query).

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Manually added exceptions to my query, and saw that it properly showed the line number of the problematic line.

image


Lee2532:
I think in case of error line, it would be better to leave it as log only.

PR #7284, opened 2025-01-16
bigquery get schema query add region

Lee2532:
## What type of PR is this? 

- [x] Feature

## Description

The query used to retrieve the BigQuery schema requires a region value. This update ensures that the correct region is specified when executing the query to avoid potential errors or inconsistencies.

Related Documents : https://cloud.google.com/bigquery/docs/information-schema-column-field-paths?hl=ko#scope_and_syntax

## How is this tested?

- [x] Manually

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -305,7 +305,7 @@ class BigQuery(BaseQueryRunner):
 
         query_base = """
         SELECT table_schema, table_name, field_path
-        FROM `{dataset_id}`.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS
+        FROM region-`{dataset_id}`.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS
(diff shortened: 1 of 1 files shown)


redash/query_runner/big_query.py
```diff
@@ -305,7 +305,7 @@ def get_schema(self, get_stats=False):
 
         query_base = """
         SELECT table_schema, table_name, field_path
-        FROM `{dataset_id}`.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS
+        FROM region-`{dataset_id}`.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS
```
yoshiokatsuneo, line 308: I think "region-`{dataset_id}`" is not correct dataset name.

Lee2532: That's right, I got the database and region wrong.

PR #7289, opened 2025-01-17
bigquery load schema diff locations ignore

Lee2532:
## What type of PR is this? 

- [x] Bug Fix

## Description

An error occurs when retrieving the BigQuery schema if there are datasets located in different regions. This update ensures that the correct region is specified to prevent such issues.

## How is this tested?

- [x] Manually

if dataset location is diff then
set location : asia
test-dataset-location : US
Not fount: dataset `test data set` was not found in location asia

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -313,6 +313,9 @@ class BigQuery(BaseQueryRunner):
         queries = []
         for dataset in datasets:
             dataset_id = dataset["datasetReference"]["datasetId"]
+            location = dataset["location"]
+            if location != self._get_location():
+                continue
             query = query_base.format(dataset_id=dataset_id)
             queries.append(query)
 


yoshiokatsuneo:
LGTM!

Lee2532:
@eradman 
Hi, It seems that an error is occurring in the coverage. Is it necessary to add additional tests?

eradman:
> @eradman Hi, It seems that an error is occurring in the coverage. Is it necessary to add additional tests?

Yes, we  should test for this condition. Try adding a test to `tests/query_runner/test_bigquery.py`

Lee2532:
> > > @eradman Hi, It seems that an error is occurring in the coverage. Is it necessary to add additional tests?
> > 
> > 
> > Yes, we should test for this condition. Try adding a test to `tests/query_runner/test_bigquery.py`
> 
> @eradman
> 
> It is nice to have 100% code coverage. But at least for query runners, current code coverage is quite low, and it is not so trivial to add tests for whole query runners. So, I feel it is not so practical to request tests only for new lines for now.

I'm of the same opinion

yoshiokatsuneo:
@Lee2532 
I noticed that this prevent loading any dataset when "Processing Location" is not set. I think it is better to check location only when the location configuration was set.

redash/query_runner/big_query.py
```diff
@@ -313,6 +313,10 @@ def get_schema(self, get_stats=False):
         queries = []
         for dataset in datasets:
             dataset_id = dataset["datasetReference"]["datasetId"]
+            location = dataset["location"]
+            if location != self._get_location():
```
Lee2532, line 317:
```suggestion
            if location and location != self._get_location():
```

Lee2532 (commented):
'Processing Location' is not specified

redash/query_runner/big_query.py
yoshiokatsuneo: @Lee2532 
I think the "if clause" should be skipped when "self._get_location()" is falsy(no configuration), like:

```
if self._get_location() and location != self._get_location():
```

Lee2532: That's right, I fixed it

Lee2532:
@eradman 
hi, can you check this workflow? thanks

Lee2532:
@eradman 
Hi, I think all the tests have passed, it can merge?

yoshiokatsuneo:
@eradman

I think this PR is good and helpful, and we can approve and merge it.
Just to make sure, do you have any concern ?

Thank you !

yoshiokatsuneo:
I tested and confirmed that the PR fixes the problem on my environment.

PR #7295, opened 2025-01-24
for same query_text refresh just execution once

gaecoli:
## What type of PR is this? 

- [ ] Refactor
- [ x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/redash/tasks/queries/execution.py
+++ b/redash/tasks/queries/execution.py
@@ -15,6 +15,7 @@ from redash.tasks.alerts import check_alerts_for_query
 from redash.tasks.failure_report import track_failure
 from redash.tasks.worker import Job, Queue
 from redash.utils import gen_query_hash, utcnow
+from redash.utils.locks import acquire_lock, release_lock
 from redash.worker import get_job_logger
 
 logger = get_job_logger(__name__)
@@ -34,14 +35,18 @@ def enqueue_query(query, data_source, user_id, is_api_key=False, scheduled_query
     logger.info("Inserting job for %s with metadata=%s", query_hash, metadata)
     try_count = 0
     job = None
+    job_lock_id = _job_lock_id(query_hash, data_source.id)
 
     while try_count < 5:
         try_count += 1
+        identifier = acquire_lock(job_lock_id)
+        if identifier is None:
+            continue
 
         pipe = redis_connection.pipeline()
         try:
-            pipe.watch(_job_lock_id(query_hash, data_source.id))
-            job_id = pipe.get(_job_lock_id(query_hash, data_source.id))
+            pipe.watch(job_lock_id)
+            job_id = pipe.get(job_lock_id)
             if job_id:
                 logger.info("[%s] Found existing job: %s", query_hash, job_id)
                 job_complete = None
@@ -66,7 +71,7 @@ def enqueue_query(query, data_source, user_id, is_api_key=False, scheduled_query
 
                 if lock_is_irrelevant:
                     logger.info("[%s] %s, removing lock", query_hash, message)
-                    redis_connection.delete(_job_lock_id(query_hash, data_source.id))
+                    redis_connection.delete(job_lock_id)
                     job = None
 
             if not job:
@@ -115,6 +120,7 @@ def enqueue_query(query, data_source, user_id, is_api_key=False, scheduled_query
         except redis.WatchError:
             continue
         finally:
+            release_lock(job_lock_id, identifier)
             pipe.reset()
 
     if not job:
--- a/redash/utils/locks.py
+++ b/redash/utils/locks.py
@@ -1,0 +1,53 @@
+import random
+import time
+import uuid
+import logging
+from redis import WatchError
+from redash import redis_connection
+
+logger = logging.getLogger(__name__)
+
+
+def acquire_lock(name, acquire_timeout=10, lock_timeout=5):
+    identifier = str(uuid.uuid4())
+    lock_name = f"lock:{name}"
+    end = time.time() + acquire_timeout
+
+    base_delay = 0.001
+    max_delay = 0.05
+
+    while time.time() < end:
+        if redis_connection.set(lock_name, identifier, ex=lock_timeout, nx=True):
+            logger.info("acquire_lock, lock_name=[%s], identifier=[%s]", lock_name, identifier)
+            return identifier
+
+        delay = base_delay + random.uniform(0, base_delay)
+        time.sleep(min(delay, max_delay))
+        base_delay = min(base_delay * 2, max_delay)
+
+    return None
+
+
+def release_lock(name, identifier):
+    lock_name = f"lock:{name}"
+    logger.info("release_lock, lock_name=[%s], identifier=[%s]", lock_name, identifier)
+    with redis_connection.pipeline() as pipe:
+        while True:
+            try:
+                pipe.watch(lock_name)
+                if pipe.get(lock_name) == identifier:
+                    pipe.multi()
+                    pipe.delete(lock_name)
+                    pipe.execute()
+                    logger.info("Lock released successfully, lock_name=[%s], identifier=[%s]", lock_name, identifier)
+                    return True
+                pipe.unwatch()
+                logger.warning("Lock not owned by this identifier, lock_name=[%s], identifier=[%s]", lock_name, identifier)
+                break
+            except WatchError:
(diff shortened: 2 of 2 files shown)


gaecoli:
This PR introduces a mechanism to prevent duplicate execution of the same SQL query in a distributed environment. By implementing a distributed locking mechanism using Redis, we ensure that only one process or thread can execute a specific SQL query at a given time, avoiding unnecessary load on the database and ensuring consistent query results.

.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - master
-  pull_request_target:
+  pull_request:
```
arikfr, line 6: @gaecoli I did this change to match changes on master. Not sure why it shows up in diff, but it's fine.

arikfr:
@gaecoli I wonder if we really need this? I mean the chance of the exact same query being executed at the same time are usually low vs. the complexity and potential issues this might add.

Do you have this happen frequently?

gaecoli:
> @gaecoli I wonder if we really need this? I mean the chance of the exact same query being executed at the same time are usually low vs. the complexity and potential issues this might add.
> 
> Do you have this happen frequently?

OK, i will consider that! Thank you! @arikfr

gaecoli:
> @gaecoli I wonder if we really need this? I mean the chance of the exact same query being executed at the same time are usually low vs. the complexity and potential issues this might add.
> 
> Do you have this happen frequently?

When i add the same query at  a dashboard, the query has lot of visualizations, table, chart..., when i add this at a dashboard, i refresh dashboard, the same query text will all execution many times in query engine (Presto, MySQL, doris...).

At my company, 1000+ people maybe refresh same dashboard, so you know what i say.

arikfr:
Maybe the right thing will be to make dashboards refreshes smarter and reuse the same query invocation for different visualizations that depend on it? I think it will be more robust and address the core issue instead of trying to address it in an indirect way.

eradman:
> Maybe the right thing will be to make dashboards refreshes smarter and reuse the same query invocation for different visualizations that depend on it

That seems like a good approach, unless this also becomes complex. My guess is that a dashboard is the only API user that would normally hit this race condition.

gaecoli:
> Maybe the right thing will be to make dashboards refreshes smarter and reuse the same query invocation for different visualizations that depend on it? I think it will be more robust and address the core issue instead of trying to address it in an indirect way.

This is a good approach, but I think such a change would make the code more complex.

eradman:
Tested this change manually, it does seem to work. Also I see the log messages

> server-1     | [2025-02-07 13:56:32,160][PID:19][INFO][redash.utils.locks] Lock released successfully, lock_name=[lock:query_hash_job:1:d6338e9508dd103771b69483fb17d4a5], identifier=[73683a39-6aea-4f29-b0d3-02e4c0385a0b]

gaecoli:
> Tested this change manually, it does seem to work. Also I see the log messages
> 
> > server-1     | [2025-02-07 13:56:32,160][PID:19][INFO][redash.utils.locks] Lock released successfully, lock_name=[lock:query_hash_job:1:d6338e9508dd103771b69483fb17d4a5], identifier=[73683a39-6aea-4f29-b0d3-02e4c0385a0b]

Yes, because it's work well at my company's Redash!

yoshiokatsuneo:
I think the code does not need to use/check "identifier" as the lock and release are always done by the same process synchronously , as it is locked.

yoshiokatsuneo:
@gaecoli 

I think the code does not need to use/check "identifier" as the lock and release are always done by the same process synchronously , as it is locked.

gaecoli:
> > @gaecoli
> > I think the code does not need to use/check "identifier" as the lock and release are always done by the same process synchronously , as it is locked.
> 
> I read the following article and understand that "identifier" is needed for safety. https://redis.io/docs/latest/develop/use/patterns/distributed-locks/

Yeah!

yoshiokatsuneo:
I think it is nice if backend-lint error is fixed by running "ruff check ." .

https://github.com/getredash/redash/actions/runs/13733542335/job/38414390891?pr=7295

image

yoshiokatsuneo:
> > I think it is nice if backend-lint error is fixed by running "ruff check ." .

> I tried it. But failed!

I could run "ruff check . --fix" to fix the issue by running ruff installed by homebrew without problem like below:

```
 (redis-lock)$ ruff --version
ruff 0.8.3
 (redis-lock)$ which ruff
/opt/homebrew/bin/ruff
 (redis-lock)$ ruff check . --fix
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
  - 'ignore' -> 'lint.ignore'
  - 'select' -> 'lint.select'
  - 'mccabe' -> 'lint.mccabe'
  - 'per-file-ignores' -> 'lint.per-file-ignores'
Found 1 error (1 fixed, 0 remaining).
 (redis-lock)$ git diff
diff --git a/redash/utils/locks.py b/redash/utils/locks.py
index 126496096..78ad2693c 100644
--- a/redash/utils/locks.py
+++ b/redash/utils/locks.py
@@ -1,10 +1,12 @@
+import logging
 import random
 import time
 import uuid
-import logging
-from redash import redis_connection
+
 from redis import WatchError
 
+from redash import redis_connection
+
 
 
 (redis-lock)$ ruff check .
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
  - 'ignore' -> 'lint.ignore'
  - 'select' -> 'lint.select'
  - 'mccabe' -> 'lint.mccabe'
  - 'per-file-ignores' -> 'lint.per-file-ignores'
All checks passed!

```

gaecoli:
> > > I think it is nice if backend-lint error is fixed by running "ruff check ." .
> 
> > I tried it. But failed!
> 
> I could run "ruff check . --fix" to fix the issue by running ruff installed by homebrew without problem like below:
> 
> ```
>  (redis-lock)$ ruff --version
> ruff 0.8.3
>  (redis-lock)$ which ruff
> /opt/homebrew/bin/ruff
>  (redis-lock)$ ruff check . --fix
> warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
>   - 'ignore' -> 'lint.ignore'
>   - 'select' -> 'lint.select'
>   - 'mccabe' -> 'lint.mccabe'
>   - 'per-file-ignores' -> 'lint.per-file-ignores'
> Found 1 error (1 fixed, 0 remaining).
>  (redis-lock)$ git diff
> diff --git a/redash/utils/locks.py b/redash/utils/locks.py
> index 126496096..78ad2693c 100644
> --- a/redash/utils/locks.py
> +++ b/redash/utils/locks.py
> @@ -1,10 +1,12 @@
> +import logging
>  import random
>  import time
>  import uuid
> -import logging
> -from redash import redis_connection
> +
>  from redis import WatchError
>  
> +from redash import redis_connection
> +
>  logger = logging.getLogger(__name__)
>  
>  
>  (redis-lock)$ ruff check .
> warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
>   - 'ignore' -> 'lint.ignore'
>   - 'select' -> 'lint.select'
>   - 'mccabe' -> 'lint.mccabe'
>   - 'per-file-ignores' -> 'lint.per-file-ignores'
> All checks passed!
> ```

done

gaecoli:
> Maybe the right thing will be to make dashboards refreshes smarter and reuse the same query invocation for different visualizations that depend on it? I think it will be more robust and address the core issue instead of trying to address it in an indirect way.

@yoshiokatsuneo I think this is good advice!

yoshiokatsuneo:
@gaecoli 

I'm not sure which is "smarter" solution to handle race condition where on server-side or frontend-side. For me, your PR is reasonable and smart enough.

If we want to make the PR shorter, I think one option is to use LUA script to handle atomic transaction. But, I'm not sure whether it is better or not.

gaecoli:
> @gaecoli
> 
> I'm not sure which is "smarter" solution to handle race condition where on server-side or frontend-side. For me, your PR is reasonable and smart enough.
> 
> If we want to make the PR shorter, I think one option is to use LUA script to handle atomic transaction. But, I'm not sure whether it is better or not.

OK, i hope this PR will help you!

PR #7320, opened 2025-02-10
Change query processing wait time to make response quick

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description
Before this PR, the query waiting interval is 0.5 seconds for up to 10 times, and 3 seconds after that.
So, the refreshStatus happens like: 0.5 / 1 / 1.5 / 2 / 2.5 / 3 / 3.5 / 4 / 4.5 / 5 / 8 / 11 / 14 (seconds) ...
There is large gap around ".. 4.5 / 5 / 8 ...".

This PR make the retrying interval gap more smooth by setting wait time to 1 second for retrying number between 11 and 50.

## How is this tested?

- [x] Manually


yoshiokatsuneo:
@arikfr 

If possible, can you review this PR that practically just adds only 2 lines of code below and makes query much responsive.
(Another commit is just for fixing restyled error.)

```
          } else if (tryNumber <= 50) {
            waitTime = 1000;
```
https://github.com/getredash/redash/pull/7320/commits/3f7b779c5f981e7de9abbf254749b3c58d38186c

arikfr (approved):
👍

arikfr (approved):

PR #7323, opened 2025-02-12
Include Plotly.js localization

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description
For now, the locale for the chat is English. It is not very intuitive for non-english people to show chart in English (ex: "January").

Make localized charts available by including and registering plotly.js localization definitions, and set locale based on browser language.

## How is this tested?

- [x] Manually

I confirmed that the "Month" and pop-up information is localized.

image
image

### DIFF
--- a/viz-lib/src/visualizations/chart/plotly/index.ts
+++ b/viz-lib/src/visualizations/chart/plotly/index.ts
@@ -1,5 +1,6 @@
 import * as Plotly from "plotly.js";
 
+import "./locales"
 import prepareData from "./prepareData";
 import prepareLayout from "./prepareLayout";
 import updateData from "./updateData";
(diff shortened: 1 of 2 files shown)


eradman (commented):
Is there a method of including all locales? It's hard to tell if we missed one

yoshiokatsuneo:
@eradman 

>Is there a method of including all locales? It's hard to tell if we missed one

I don't think so.
Unfortunately, I think plotlyjs does not provide the way to register all locale at once, and webpack does not provide simple way to import all the locale files.

eradman:
If I'm reading this right, this will add about 88KB to the size of the web app:

before
web-ui-load

after
web-ui-load-pr-7323

Out of 6.5MB this is an increase of 1.3%, which is probably fine.

eradman:
I think we determined that this change will have a minimal impact on the size of the distribution. @yoshiokatsuneo please resolve conflicts and I think we can merge.

yoshiokatsuneo:
@eradman 

Thanks !
I just rebased and resolved the conflicts !

eradman (approved):
Super! Merging

yoshiokatsuneo:
@eradman Thanks a lot!

PR #7325, opened 2025-02-12
Removed error prone key decode looping

kuranium:
## What type of PR is this? 

- [X] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Old implementation could loop thru irrelevant public keys and trying to decode with them, causing unnecessary and missleading exceptions in logs. 
In this commit the "net" fetched keys get filtered upon fetching, using the basic jwt functionality. So now we only use and cache keys with matching "kid" fields.

## How is this tested?

- [X] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

### DIFF
--- a/redash/authentication/jwt_auth.py
+++ b/redash/authentication/jwt_auth.py
@@ -1,82 +1,65 @@
-import json
 import logging
 
 import jwt
-import requests
+from jwt import PyJWKClient
+
 
 logger = logging.getLogger("jwt_auth")
 
 FILE_SCHEME_PREFIX = "file://"
 
 
-def get_public_key_from_file(url):
+def get_signing_key_from_file(url):
     file_path = url[len(FILE_SCHEME_PREFIX) :]
     with open(file_path) as key_file:
         key_str = key_file.read()
 
-    get_public_keys.key_cache[url] = [key_str]
+    get_signing_key.key_cache[url] = key_str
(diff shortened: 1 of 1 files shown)


kuranium:
@yoshiokatsuneo any clue who might be invested in / responsible for the authentication related code?

redash/authentication/jwt_auth.py
```diff
@@ -1,82 +1,65 @@
...
-    keys = {}
+    key_cache = get_signing_key.key_cache
+    key = {}
     if url in key_cache:
-        keys = key_cache[url]
+        key = key_cache[url]
     else:
         if url.startswith(FILE_SCHEME_PREFIX):
-            keys = [get_public_key_from_file(url)]
+            key = [get_signing_key_from_file(url)]
```
yoshiokatsuneo, line 41: I just guess that "key" should not be an array.

yoshiokatsuneo:
> @yoshiokatsuneo any clue who might be invested in / responsible for the authentication related code?

@kuranium 
I have no idea whether there is any component owner or not...

redash/authentication/jwt_auth.py
kuranium: thanks, should be fixed now

PR #7326, opened 2025-02-12
Make autocomplete always available

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description
Before this PR, autocomplete is not available if there are more than 5,000 tokens. The limit is too small for our data warehouse, and probably many data warehouse.

This PR removes the limit, and make autocomplete always available.

Anyway, if each user want to disable autocomplete, they can just disable by themselves, and I think it is better to make it user's choice instead of hard-coded limit.

Also, as tokens are stored on memory even if the autocomplete is disabled, amount of memory consumption does not change.

This PR is derived from the discussion on https://github.com/getredash/redash/pull/7319 .

## How is this tested?

- [x] Manually

I confirmed that the auto complete works smoothly for tokens more than 700,000.

### DIFF
--- a/client/app/pages/queries/hooks/useAutocompleteFlags.js
+++ b/client/app/pages/queries/hooks/useAutocompleteFlags.js
@@ -1,13 +1,8 @@
 import { useCallback, useMemo, useState } from "react";
-import { reduce } from "lodash";
 import localOptions from "@/lib/localOptions";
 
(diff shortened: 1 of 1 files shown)


eradman (approved):
Tested manually. Seems like a nice simplifcation

PR #6919, opened 2024-04-22
Remove workaround from check_csrf()

eradman:
## What type of PR is this? 

- [x] Bug Fix

## Description

This code was supposed to be temporary, and raises an exception if REDASH_MULTI_ORG=true is set.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2213, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python3.8/site-packages/werkzeug/middleware/proxy_fix.py", line 182, in __call__
    return self.app(environ, start_response)
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2193, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/local/lib/python3.8/site-packages/flask_restful/__init__.py", line 298, in error_router
    return original_handler(e)
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1486, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.8/site-packages/flask_restful/__init__.py", line 298, in error_router
    return original_handler(e)
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1482, in full_dispatch_request
    rv = self.preprocess_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1974, in preprocess_request
    rv = self.ensure_sync(before_func)()
  File "/app/redash/security.py", line 43, in check_csrf
    dest = f"{view.__module__}.{view.__name__}"
AttributeError: 'NoneType' object has no attribute '__module__'
```
## How is this tested?

- [x] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

1. Add `REDASH_MULTI_ORG=true` to `.env`
2. Start the server
3. Create an organization and a user
```
docker compose run server manage org create redash
docker compose run server manage users create_root eradman@starfishstorage.com Redash --password ******
```
4. Log into http://localhost:5001/redash/

PR #7327, opened 2025-02-14
Partiallly Revert "Remove workaround from check_csrf() (#6919)"

eradman:
## What type of PR is this? 

- [x] Bug Fix

## Description

This workaround was missing `if view is not None ` as found in https://github.com/pallets-eco/flask-wtf/pull/419/files

## How is this tested?

- [x] Manually

With the following added to `.env`

```
REDASH_MULTI_ORG=true
REDASH_ENFORCE_CSRF=true
REDASH_ENFORCE_CSRF_FOR_AUTH_USERS=true
```

### DIFF
--- a/redash/security.py
+++ b/redash/security.py
@@ -1,6 +1,6 @@
 import functools
 
-from flask import session
+from flask import request, session
 from flask_login import current_user
 from flask_talisman import talisman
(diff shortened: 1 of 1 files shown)


arikfr (approved):
👍

Issue #7329, opened 2025-02-15
Y-value is not correct but aggregated for Scatter, Line, Bubble chart


yoshiokatsuneo:
### Issue Summary

Chart for Scatter, Bubble, Line..  does not show correct y-value when there are multiple data for the same x-value.

### Steps to Reproduce

1. Run SQL for multiple data on the same x-value, like below.

```
select 1 as x, 1 as y
union all
select 2 as x, 1 as y
union all
select 3 as x, 1 as y
union all
select 3 as x, 1 as y
```

It should show y=1 for all value, but the chart show y=2 (=1+1) for x-value 3.

Y-value should not be aggregated for Scatter, line ,etc...

<img width="965" alt="Image" src="https://github.com/user-attachments/assets/01488113-2cb8-4aa6-a121-36a2656c8cfe" />

<img width="975" alt="Image" src="https://github.com/user-attachments/assets/f1385b75-8d0f-4bd5-824f-8831bfe50191" />

### Technical details:

* Redash Version: Latest commit: 71902e593389963c45b2c490f20de9e011b29df4
* Browser/OS: Chrome
* How did you install Redash: from source code

PR #7330, opened 2025-02-15
Fix the issue that chart(scatter, bubble, line...) having data with same x-value have wrong y-value

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
Before this PR, the chart(scatter, line, bubble...) having same x-value have wrong y-value.
For example, before this PR, if the data is like (x,y) = (1,1), (2,1), (3,1), (3,1)  , the chart have y-value for x=3 is not 1 but 2 (So, the y-value for the same x-value is summed up.)

This PR fix the issue.
See issue: https://github.com/getredash/redash/issues/7329

This issue was introduced by following #6908 . #6908 is OK for pie chart written as example in PR. But not good for other charts. The pie chart is basically for categorical(nominal-level) x-value and ratio-scale y-value, so it is OK to sum up y-value. But other charts like scatter/bubble charts are normally for quantitative x-value and y-value can be interval-scale. So it is not good just summing up.

And, scatter chart can have multiple x-value as as scatter chart does not care x or y, and the issue is critical. But for pie chart, x-value can be(or should be) unique, anyway.

For example, when there is a scatter chart for physique and x-axis is for weight and y-axis is for height, summing up height for the same weight does not make sense at all.
For stock chart, summing up stock price for the same day does not make sense.

This PR does not change the behavior for pie chart.

## How is this tested?

- [x] Manually

I manually confirmed that the y-value is not summed up but shown as is.

Tested query:
```
select 1 as x, 1 as y
union all
select 2 as x, 1 as y
union all
select 3 as x, 1 as y
union all
select 3 as x, 1 as y
```

Scatter chart before this PR:
image

Scatter chart after this PR:
image

### DIFF
--- a/viz-lib/src/visualizations/chart/plotly/prepareDefaultData.ts
+++ b/viz-lib/src/visualizations/chart/plotly/prepareDefaultData.ts
@@ -99,8 +99,8 @@ function prepareSeries(series: any, options: any, numSeries: any, additionalOpti
       };
 
   const sourceData = new Map();
-
-  const labelsValuesMap = new Map();
+  const xValues: any[] = [];
+  const yValues: any[] = [];
 
   const yErrorValues: any = [];
   each(data, row => {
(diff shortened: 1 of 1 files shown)


gaecoli:
Thanks PR. Maybe u should test it more times, if results with null values?

yoshiokatsuneo:
> Thanks PR. Maybe u should test it more times, if results with null values?

@gaecoli 

Here is the test results with null values. 
It looks that the PR is working without problem. How about ?

Scatter(Before PR):
image


Scatter(After PR):
image


Line(Before PR):
image


Line(After PR):
image

eradman (approved):
Confirmed that this restores the original scatterplot behavior, and pie charts sum values

PR #6685, opened 2024-01-03
Removed simplejson

AndrewChubatiuk:
## What type of PR is this? 
Removed unneeded simplejson dependency

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [ ] Other

Issue #6992, opened 2024-05-29
Redash queries crash if numerical column contain `NaN` 


noxdafox:
### Issue Summary

The following commit https://github.com/getredash/redash/commit/4d5103978bdc20900f205532b48356ec80339632, introduced a regression. As `simplejson` was replaced with `json` strict JSON validation is now enforced.

The belief is that JSON validation was enforced to prevent the `json` module to return `NaN` values which break the frontend:
https://github.com/getredash/redash/blob/master/redash/utils/__init__.py#L129
```
    # Using allow_nan = True will make Python render nan as NaN, leading to parse error in front-end
```

Yet, the change and the comments suggest that this was not the expected behaviour:
https://github.com/getredash/redash/blob/master/redash/utils/__init__.py#L128
```
    # Float value nan or inf in Python should be render to None or null in json.
```

Previous to that commit, `NaN` values were allowed and automatically converted to `None` by `simplejson`. `json` does not support this functionality and now the behaviour is to raise `ValueError` effectively crashing the query every time a `NaN` value is encountered.

```
[2024-05-29 08:20:44,474][PID:966][ERROR][rq.worker] [Job 1570d8da-97d2-4ab6-8395-3d6ff4fdda8a]: exception raised while executing (redash.tasks.queries.execution.execute_query)
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1204, in _execute_context
    context = constructor(dialect, self, conn, *args)
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 873, in _init_compiled
    param = dict(
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 876, in <genexpr>
    processors[key](compiled_params[key])
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/sql/type_api.py", line 1232, in process
    return process_param(value, dialect)
  File "/app/redash/models/types.py", line 29, in process_bind_param
    return json_dumps(value)
  File "/app/redash/utils/__init__.py", line 131, in json_dumps
    return json.dumps(data, *args, **kwargs)
  File "/usr/local/lib/python3.8/json/__init__.py", line 234, in dumps
    return cls(
  File "/usr/local/lib/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
[...]

PR #7339, opened 2025-02-19
Sanitize NaN, Infinite, -Infinite causing error when saving as PostgreSQL JSON

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
When running query that output NaN, Infinite, or -Infinite, redash executor crashes with the error "Out of range float values are not JSON compliant" like below. Redash does not output any rows when the error was occurred.

```
[2025-02-19 04:48:49,040][PID:593][ERROR][rq.worker] [Job 6cfa1d1c-2a32-48c6-8bf4-0cd8f8ba6f32]: exception raised while executing (redash.tasks.queries.execution.execute_query)
...
  File "/usr/local/lib/python3.10/site-packages/rq/job.py", line 1280, in perform
  File "/usr/local/lib/python3.10/site-packages/rq/job.py", line 1317, in _execute
    result = self.func(*self.args, **self.kwargs)
  File "/app/redash/tasks/queries/execution.py", line 309, in execute_query
    ).run()
  File "/app/redash/tasks/queries/execution.py", line 253, in run
    updated_query_ids = models.Query.update_latest_result(query_result)
  File "/app/redash/models/__init__.py", line 738, in update_latest_result
    for q in queries:
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/orm/query.py", line 3534, in __iter__
    self.session._autoflush()
...
sqlalchemy.exc.StatementError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
(builtins.ValueError) Out of range float values are not JSON compliant
[SQL: INSERT INTO query_results (org_id, data_source_id, query_hash, query, data, runtime, retrieved_at) VALUES (%(org_id)s, %(data_source_id)s, %(query_hash)s, %(query)s, %(data)s, %(runtime)s, %(retrieved_at)s) RETURNING query_results.id]
...
```

The error occurs because JSON(PostgreSQL JSON type) does not support NaN, Infinite, -Infinite.

This PR fixes the issue by sanitize NaN, Infinite, -Infinite value to None. This is the save behavior that JavaScript JSON.stringify() does.

We can reproduce the issue by running following query on BigQuery or PostgreSQL.

BigQuery:
```
select 1.0, cast('NaN' as float64), cast('Infinity' as float64), cast('-Infinity' as float64)
```

PostgreSQL:
```
select 1.0::float, 'NaN'::float, 'Infinity'::float, '-Infinity'::float
```

The issue was introduced by #6685 which removes simplejson. Before #6685 , NaN / Inf was converted to null by simplejson with ignore_nan=True. This PR change behavior for NaN/Inf as same as before #6685 .

## How is this tested?

- [x] Manually

Unit test was added.
Also, we can confirm by running the query on the description. 

Before this PR:

image

After this PR:

image

### DIFF
--- a/redash/tasks/queries/execution.py
+++ b/redash/tasks/queries/execution.py
@@ -1,3 +1,4 @@
+import math
 import signal
 import sys
 import time
@@ -171,6 +172,17 @@ def _get_size_iterative(dict_obj):
     return size
 
 
+# Convert NaN, Inf, and -Inf to None, which PostgreSQL cannot handle.
+def _sanitize_data(data):
+    if isinstance(data, dict):
+        return {k: _sanitize_data(v) for k, v in data.items()}
+    if isinstance(data, list):
+        return [_sanitize_data(v) for v in data]
+    if isinstance(data, float) and (math.isnan(data) or math.isinf(data)):
+        return None
+    return data
+
+
 class QueryExecutor:
     def __init__(self, query, data_source_id, user_id, is_api_key, metadata, is_scheduled_query):
         self.job = get_current_job()
@@ -245,7 +257,7 @@ class QueryExecutor:
                 self.data_source,
                 self.query_hash,
                 self.query,
-                data,
+                _sanitize_data(data),
                 run_time,
                 utcnow(),
             )
--- a/tests/tasks/test_queries.py
+++ b/tests/tasks/test_queries.py
@@ -186,6 +186,41 @@ class QueryExecutorTests(BaseTestCase):
             result = models.QueryResult.query.get(result_id)
             self.assertEqual(result.data, query_result_data)
 
+    def test_sanitize_nan_inf(self, _):
+        """
+        NaN, Inf, and -Inf are sanitized to None.
+        """
+        with patch.object(PostgreSQL, "run_query") as qr:
+            qr.return_value = (
+                {
+                    "columns": [
+                        {"name": "_col0", "friendly_name": "_col0", "type": "float"},
+                        {"name": "_col1", "friendly_name": "_col1", "type": "float"},
+                        {"name": "_col2", "friendly_name": "_col1", "type": "float"},
+                        {"name": "_col3", "friendly_name": "_col1", "type": "float"},
+                    ],
+                    "rows": [{"_col0": 1.0, "_col1": float("nan"), "_col2": float("inf"), "_col3": float("-inf")}],
+                },
+                None,
+            )
+            query_result_data = {
+                "columns": [
+                    {"name": "_col0", "friendly_name": "_col0", "type": "float"},
+                    {"name": "_col1", "friendly_name": "_col1", "type": "float"},
+                    {"name": "_col2", "friendly_name": "_col1", "type": "float"},
+                    {"name": "_col3", "friendly_name": "_col1", "type": "float"},
+                ],
+                "rows": [{"_col0": 1.0, "_col1": None, "_col2": None, "_col3": None}],
+            }
+            result_id = execute_query(
+                "select 1.0::float, 'NaN'::float, 'Infinity'::float, '-Infinity'::float",
+                self.factory.data_source.id,
+                {},
+            )
+            self.assertEqual(1, qr.call_count)
+            result = models.QueryResult.query.get(result_id)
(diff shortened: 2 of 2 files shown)


gaecoli:
I think this operation is unnecessary, NaN is not a numeric type and should not be forced to be stored.

gaecoli:
Each database or data engine handles it differently, for NaN and others situation, so I think this situation should not be handled. They should abide by their original system internal processing methods.

yoshiokatsuneo:
@gaecoli 
Thank you for your comment.
And, I think this bug should be fix with following reasons:

- The bug make redash crash. I think crashing redash is bug that should be fixed. 
- I think it is not good idea saying "crash is necessary" as keep as is when it can be fixed.
- The bug was introduced by #6685 which changes from simplejson to python natvie json.  Before the PR, NaN/Inf was converted to null . This PR fix the issue by changing behavior just before #6685 .

https://simplejson.readthedocs.io/en/latest/
```
If ignore_nan is true (default: False), then out of range float values (nan, inf, -inf) will be serialized as null in compliance with the ECMA-262 specification. If true, this will override allow_nan.
```

- NaN / Inf is not special value but quite common float value, as NaN / Inf is IEEE-754 standard.
- NaN value is generated by many databases including PostgreSQL, BigQuery, Redshift, Oracle, MongoDB.
- The problem potentially happens on any database as the issue happens when the query is saved to redash PostgreSQL database. Saving query to PostgreSQL happens for any data sources, as it happens not when the query runs but when query result was save.

gaecoli:
@arikfr If u have free time, please review this PR! Thank you!

cj-matteo-cafasso:
This PR addresses https://github.com/getredash/redash/issues/6992

gaecoli:
> This PR addresses #6992

I looked into this issue in detail and it really should be fixed. @yoshiokatsuneo @cj-matteo-cafasso

arikfr (commented):
Thank you for addressing this @yoshiokatsuneo !

The error comes from `json_dumps` passing `allow_nan=False` in [`json_dumps`](https://github.com/getredash/redash/blob/4357ea56ae93ac0d2d0c1053636e426c645b0a02/redash/utils/__init__.py#L123-L131) which was indeed introduced in the PR mentinoed.

How about we move this logic (of turning NaN/Inf to None) to `json_dumps` instead of calling it here?

EugeneChung:
How about testing orjson? orjson.dumps() converts NaN float to null.
https://github.com/getredash/redash/issues/6992#issuecomment-2684148024

arikfr:
@EugeneChung would you like to open a PR that switches to orjson?

yoshiokatsuneo:
@arikfr 

> The error comes from json_dumps passing allow_nan=False in [json_dumps](https://github.com/getredash/redash/blob/4357ea56ae93ac0d2d0c1053636e426c645b0a02/redash/utils/__init__.py#L123-L131) which was indeed introduced in the PR mentinoed.
> How about we move this logic (of turning NaN/Inf to None) to json_dumps instead of calling it here?

Thank you for your comment !
I just tried to create another PR below that changes json_dumps (instead of QueryExecutor.run in this PR).
How about ?

https://github.com/getredash/redash/pull/7348

arikfr:
#7348 was just merged making this one unnecessary.

EugeneChung:
> @EugeneChung would you like to open a PR that switches to orjson?

Sure, I am going to do it.

EugeneChung:
I'm working on [orjson](https://github.com/EugeneChung/redash/pull/1/files). But there're some issues with `datetime.datetime, datetime.date, datetime.time` processing.
- Unlike std json JsonEncoder mechanism, orjson does not support override of its native behaviour.
- orjson.dump takes [a default function](https://github.com/ijl/orjson#default), but it's not called for types that orjson natively supports.
- For example, orjson natively supports `datetime.datetime, datetime.date, datetime.time`, but it dumps as full microsecond precision.
   - JSONEncoder : `{"time":"2024-03-01T15:30:45.123"}`
   - orjson : `{"time":"2024-03-01T15:30:45.123456"}`

PR #7341, opened 2025-02-22
Add NULLS LAST option for query list ordering

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description
Before this PR, query list ordering does not have "NULLS LAST" option.

So, when I sort query list by "Last Executed At" in descendant order to see the recent active queries, I got bunch of queries that is not executed at all and "executed_at" is null in the beginning of the list.

This happens because PostgreSQL handles NULL as the "largest value" for ordering. 

I think the behavior does not make sense.

This PR fix the issue by adding "NULLS LAST" options for ordering so that the most recent query comes first, instead of query that was never run comes first.

## How is this tested?

- [x] Manually

Before this PR:
image

After this PR:
image

### DIFF
--- a/redash/utils/query_order.py
+++ b/redash/utils/query_order.py
@@ -33,7 +33,7 @@ from sqlalchemy.orm import mapperlib
 from sqlalchemy.orm.properties import ColumnProperty
 from sqlalchemy.orm.query import _ColumnEntity
 from sqlalchemy.orm.util import AliasedInsp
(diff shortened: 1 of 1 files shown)


gaecoli (approved):
LGTM!

gaecoli:
Thanks for your PR! Maybe the dashboard page could also change this?

PR #7351, opened 2025-02-26
Make favorite queries/dashboard order by starred at(favorited at)

yoshiokatsuneo:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A


yoshiokatsuneo:
@eradman 

Can you review it if possible ?

I feel ordering Favorite queries/dashboards by "created_at" does not make much sense as it is not related to how much user favor it.
This PR solve the issue by ordering by the time user starred so that users can change the order of Favorites.

yoshiokatsuneo:
@eradman 

Thank you for the comment !

It is right that the feature is not so intuitive.
But, at the same time, I'm also not sure whether it is good idea to reordering items just after checking, because "checking item" normally does not move around items. And, this is the same behavior as Confluence does.
(And, probably, it is not so good idea to remove item just after checking star in this use case.)

The better solution is to make favorite list sortable, but it will require a bit more change... 
I don't have strong opinion about what is the better solution for now.

Do you have any idea ?...

eradman:
Ordering favorites by name would be easy to understand, but is also odd because that's not similar to what All Queries/All Dashboards does.

Perhaps it would be an improvement to show "Starred At" _instead of_ "Created At" since queries already have two timestamp fields

yoshiokatsuneo:
@eradman 

Thank you for your comment !
Actually, I remember that DataSource list is ordered by name.

For now, I merge the PR, as I think it is better than now.
As there are some options, I think we can make it better later referring the feedbacks.

PR #7366, opened 2025-03-08
Fix table item list ordering

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
Ordering by clicking table column does not work well, as the sorting handled by toggling two status(on/off) on every click but actually the sorting have triple status ("asendant", "descendant", default).

This PR fixed the issue by using sorting status on Table component, instead of "guessing" the status from outside of component.

## How is this tested?
- [x] Manually

Clicking table column changes the order as intended.

image

### DIFF
--- a/client/app/components/items-list/ItemsList.tsx
+++ b/client/app/components/items-list/ItemsList.tsx
@@ -28,6 +28,7 @@ export interface Controller<I, P = any> {
   orderByField?: string;
   orderByReverse: boolean;
   toggleSorting: (orderByField: string) => void;
+  setSorting: (orderByField: string, orderByReverse: boolean) => void;
 
   // pagination
   page: number;
@@ -139,10 +140,11 @@ export function wrap<I, P = any>(
         this.props.onError!(error);
 
       const initialState = this.getState({ ...itemsSource.getState(), isLoaded: false });
-      const { updatePagination, toggleSorting, updateSearch, updateSelectedTags, update, handleError } = itemsSource;
+      const { updatePagination, toggleSorting, setSorting, updateSearch, updateSelectedTags, update, handleError } = itemsSource;
       this.state = {
         ...initialState,
         toggleSorting, // eslint-disable-line react/no-unused-state
+        setSorting, // eslint-disable-line react/no-unused-state
         updateSearch: debounce(updateSearch, 200), // eslint-disable-line react/no-unused-state
         updateSelectedTags, // eslint-disable-line react/no-unused-state
         updatePagination, // eslint-disable-line react/no-unused-state
--- a/client/app/components/items-list/classes/ItemsSource.js
+++ b/client/app/components/items-list/classes/ItemsSource.js
@@ -39,14 +39,12 @@ export class ItemsSource {
     const customParams = {};
     const context = {
       ...this.getCallbackContext(),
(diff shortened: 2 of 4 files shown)


eradman:
@yoshiokatsuneo what is the difference from the user's perspective? I tried a simple query, and 3 clicks on the "date" column (sort asc, sort desc, default) seems to work the same as before

```sql
SELECT 1 AS row, NULL AS date
UNION
SELECT 2, now() - interval '55 hours'
UNION
SELECT 3, now() - interval '1 day' 
```

yoshiokatsuneo:
> @yoshiokatsuneo what is the difference from the user's perspective? I tried a simple query, and 3 clicks on the "date" column (sort asc, sort desc, default) seems to work the same as before
> 
> ```sql
> SELECT 1 AS row, NULL AS date
> UNION
> SELECT 2, now() - interval '55 hours'
> UNION
> SELECT 3, now() - interval '1 day' 
> ```

@eradman 

The problem is not about sorting rows on each query's table, but about sorting queries on query list page like below.

420573842-6dd6a7ea-2533-46b3-95f3-41612d442a09

eradman:
> The problem is not about sorting rows on each query's result table, but about sorting queries on query list page like below.

Okay, but I still don't see incorrect results. Does this fix an error in sorting, or is this an internal code cleanup?

yoshiokatsuneo:
> > The problem is not about sorting rows on each query's result table, but about sorting queries on query list page like below.
> 
> Okay, but I still don't see incorrect results. Does this fix an error in sorting, or is this an internal code cleanup?

@eradman 

This PR fix an error in sorting.
On each clicking "Created At", sorting order and sorting triangle mark changes like below.


- descending order, no triangle mark
- ascending order, up triangle mark(▲)
- descending order, down triangle mark(▼)
- ascending order, no triangle mark
- descending order, up triangle mark(▲)
- ascending order, down triangle mark(▼)
- descending order, no triangle mark
...

So, "sorting order" have 2-click cycle, but "triangle mark" have 3-click cycle.

eradman (approved):

eradman:
> Aha! Tricky. Please add these bullet points to the commit message

Never mind, I can update the commit message when merging

yoshiokatsuneo:
@eradman Thanks a lot !

zachliu:
@yoshiokatsuneo thank you! i think i can finally create a PR for this https://github.com/getredash/redash/discussions/7117

yoshiokatsuneo:
@zachliu I think this PR is already merged, and the issue was already solved.

zachliu:
@yoshiokatsuneo Sorry for the confusion—you misunderstood my earlier message. What I meant was that I had intended to make the `Created By` column sortable as well. Now that you've fixed the ordering issue, I can create a [separate PR](https://github.com/getredash/redash/pull/7425) for that.

yoshiokatsuneo:
> @yoshiokatsuneo Sorry for the confusion—you misunderstood my earlier message. What I meant was that I had intended to make the `Created By` column sortable as well. Now that you've fixed the ordering issue, I can create a [separate PR](https://github.com/getredash/redash/pull/7425) for that.

@zachliu 
Ah, I see. Thanks !

PR #7367, opened 2025-03-08
Upgrade prettier version to the same version that CI is using

yoshiokatsuneo:
## What type of PR is this? 
- [x] Other

## Description
Now, the prettier version used on CI is different from the local version defined in package.json .  This make quite difficult and annoying to fix the error reported by CI, as local prettier does not fix the error reported by CI.

This PR upgrade prettier to the same version(3.3.2) that CI is using.

## How is this tested?
- [x] Manually

There is no error on "yarn prettier"


eradman (approved):
Indeed, there is no reason these versions should not match!

```
redash$ grep prettier: .restyled.yaml
    image: restyled/restyler-prettier:v3.3.2-2
```

yoshiokatsuneo:
@eradman Thank you for your review!!

PR #7378, opened 2025-03-21
Change BigQuery super class from BaseQueryRunner to BaseSQLQueryRunner

yoshiokatsuneo:
## What type of PR is this? 

- [x] Feature

## Description
BigQuery is SQL.
So, the super class of BigQuery runner should be BaseSQLQueryRunner instead of BaseQueryRunner so that auto limit feature (LIMIT 1000) is available for BigQuery.

## How is this tested?
- [x] Manually

I confirmed that LIMIT 1000 feature is available on BigQuery.
Test Query:
```
select * from unnest(generate_array(1,2000))
```

image

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -12,7 +12,7 @@ from redash.query_runner import (
     TYPE_FLOAT,
     TYPE_INTEGER,
     TYPE_STRING,
-    BaseQueryRunner,
+    BaseSQLQueryRunner,
     InterruptException,
     JobTimeoutException,
     register,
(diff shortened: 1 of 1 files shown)


yoshiokatsuneo:
@arikfr 

Can you review this 2-line PR that makes BigQuery query much safer, if possible ?

yoshiokatsuneo:
@snickerjp
Thank you for your review ! Can you merge the PR, if possible ?

yoshiokatsuneo:
@snickerjp Thanks a lot !!

PR #7389, opened 2025-03-31
Remove unnecessary global declarations in destination and query

mirkan1:
If you are not reinstating a variable you dont need to call global variable

- [ ] Refactor


mirkan1:
@arikfr @justinclift
I am right or right?

PR #4809, opened 2020-04-18
Add 12 column layout option for all dashboards

kkmagician:
## What type of PR is this? (check all applicable)

- [x] Feature

#### Reasoning
For more than a year of using Redash in my company and for personal projects one thing perplexed me: why there are only 6 columns available for widgets to be placed on a dashboard. Looking at other solutions (Grafana, Metabase, Superset, Shiny) all of them have more columns in the layout (12+). The number 12 is a great setting for any dashboard: it divides by 2, 3, 4 and 6, whereas 6 has only the 2 and 3 options. The handiness of dividing the layout into 4 even columns is apparent when using Counter or any other "small" widgets. See examples in the screenshots section.

Overall, the feature allows for a more fluid developer and consumer experience, while not forcing anyone to adopt the feature and not breaking old dashboards.

#### Implementation
* The old 6-column layout is used by default;
* New "Use 12 columns layout" checkbox in the dashboard editor – it switches the grid options used for the dashboard and updates the dashboard in the database: column dashboards.use_12_column layout (bool);
* When the checkbox is checked, background CSS grid changes accordingly;
* When the checkbox is unchecked, react-grid automatically smooshes all the widgets of width 7+.

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
ezgif com-optimize
image
image

PR #7396, opened 2025-04-01
Use 12-column layout for dashboard grid

eradman:
## What type of PR is this? 

- [x] Feature

## Description

* Set `minSizeX`, `minSizeY` for widgets to 2 since a value of 1 breaks all assumptions of the UI layout.
* Migration provide transition from 6 to 12 columns for all widgets.

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually

```
./manage.py db downgrade 1655999df5e3
./manage.py db upgrade
```

https://github.com/getredash/redash/pull/4809

https://github.com/user-attachments/assets/15231266-7b0d-4374-9982-02260f3d5cf9


eradman:
@yoshiokatsuneo if you have time, please review this change. This is a feature that users I support have asked for, and I think others would fit it useful as well.

yoshiokatsuneo:
> @yoshiokatsuneo if you have time, please review this change. This is a feature that users I support have asked for, and I think others would fit it useful as well.

@eradman 

Thank you for the cool PR ! The code looks good and the change is straightforward.
I tested your code and confirmed that the dashboard is working with 12 columns without problem.
Also, I tested and confirmed that the migration works well !

One thing I notice is that the default widget width is narrower than before, as it is half of what is was before like below.
Is it intended behavior ?
If it is not intended, I think we can double defaultSizeX, and defaultColumns for each visualizations.

image

Thank you !

eradman:
> If it is not intended, I think we can double defaultSizeX, and defaultColumns for each visualizations.

Good observation. I updated `defaultSizeX` as well

eradman:
Found out that most visualizations have their own default width. So I need to update these as well

```
viz-lib/src/visualizations/chart/index.ts:  defaultColumns: 3,
viz-lib/src/visualizations/choropleth/index.ts:  defaultColumns: 3,
viz-lib/src/visualizations/counter/index.ts:  defaultColumns: 2,
viz-lib/src/visualizations/details/index.ts:  defaultColumns: 2,
viz-lib/src/visualizations/map/index.ts:  defaultColumns: 3,
viz-lib/src/visualizations/pivot/index.ts:  defaultColumns: 3,
viz-lib/src/visualizations/table/index.ts:  defaultColumns: 3,
```

eradman:
Ready for review

yoshiokatsuneo (approved):
Thank you very much !
I think this PR makes Redash more useful !

eradman:
Thank you @yoshiokatsuneo! Setting to auto-merge

yoshiokatsuneo:
Just as a reference.
There is a user who does not have `col` option in the widgets table got error.
I'm not sure why there is no `col` option.

https://github.com/getredash/redash/discussions/7494#discussioncomment-14029115

eradman:
> There is a user who does not have col option in the widgets table got error.

We'll keep an eye on this discussion: I don't see how it's possible to add a widget to a dashboard without setting the "col" value.

PR #7422, opened 2025-04-28
Add "Last 10 years" option for dynamic date range

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description
At now, maximum date range option is "last 12 months".
But, it is quite convenient if I can view longer period like 10 years.

This PR add "Last 10 years" option to date range option.

## How is this tested?

- [x] Manually

I run following query with Date Range parameter.
```
select '{{DATE_RANGE.start}}', '{{DATE_RANGE.end}}'
```

And, I confirmed that output is correctly shown like below.
```
'2015-04-28' | '2025-04-28' |  
```

### DIFF
--- a/client/app/components/dynamic-parameters/DateRangeParameter.jsx
+++ b/client/app/components/dynamic-parameters/DateRangeParameter.jsx
@@ -106,6 +106,11 @@ const DYNAMIC_DATE_OPTIONS = [
     value: getDynamicDateRangeFromString("d_last_12_months"),
     label: null,
   },
+  {
+    name: "Last 10 years",
+    value: getDynamicDateRangeFromString("d_last_10_years"),
+    label: null,
+  },
 ];
 
 const DYNAMIC_DATETIME_OPTIONS = [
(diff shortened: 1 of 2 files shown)


eradman:
Adding "Last (3, 5, 7, 10) years" could be even more useful. Of course we don't want to add too many items!
I bet we could get rid of "Last 12 months" since this is _very_ close to "Last year"

Let's see if others have a comment on the this

yoshiokatsuneo:
@eradman 

Thank you for your comment !

> Adding "Last (3, 5, 7, 10) years" could be even more useful. Of course we don't want to add too many items!

At first, I would like to have a option to select "all range" easily, and "last 10 years" is close to it.

And yes, I think we can add "last 2 years" and "last 3 years" as it is consistent with current "last 1/2/3 months"(last 30/60/90 days) options. How about ?

> I bet we could get rid of "Last 12 months" since this is very close to "Last year"

For me, "Last 12 months" is much important as I want to see time series data "up to now" instead of the data "up to a day before new year", normally.

yoshiokatsuneo:
@eradman Thanks a lot!

PR #7431, opened 2025-05-02
Rehash query_hash in migration

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description

As following PR changed the logic for creating query hash, existing query need to update query hash.
This migration updated the query_hash.

https://github.com/getredash/redash/pull/7378

The migration is basically same as fix_hash migration below.

https://github.com/getredash/redash/blob/master/migrations/versions/9e8c841d1a30_fix_hash.py

## How is this tested?
- [x] Manually


yoshiokatsuneo:
@eradman 

This migration PR fixes the query hash mismatch issue for BigQuery introduced on https://github.com/getredash/redash/pull/7378 .
Without the rehashing fix, query result is not saved correctly.

If possible, could you please check it out ?

Issue #7433, opened 2025-05-05
CSS and graphics don't load: WEBPACK_MANIFEST_PATH not found


ongrid:
### Issue Summary

Opening http://127.0.0.1:5001/setup at the first launch leads to unstyled page with missing favicons and png files (HTTP 404)

The following exception raised on backend side

```
server-1     | Traceback (most recent call last):
server-1     |   File "/app/redash/handlers/webpack.py", line 17, in get_asset
server-1     |     with open(WEBPACK_MANIFEST_PATH) as fp:
server-1     | FileNotFoundError: [Errno 2] No such file or directory: '/app/redash/handlers/../../client/dist/asset-manifest.json'
```

Disabling skip_frontend_build enables yarn building but without any effect

```
diff --git a/compose.yaml b/compose.yaml
index 0b116847..b5fb4970 100644
--- a/compose.yaml
+++ b/compose.yaml
@@ -4,7 +4,7 @@ x-redash-service: &redash-service
   build:
     context: .
     args:
-      skip_frontend_build: "true"  # set to empty string to build
+      skip_frontend_build: ""  # set to empty string to build
   volumes:
     - .:/app
   env_file:
```

### Steps to Reproduce

1. clone repo
2. set `skip_frontend_build: ""`
3. docker compose build
4. docker compose up
5. Open http://127.0.0.1:5001/

### Technical details:

* Redash Version:

```
commit 5d31429ca80486f27645e7e6d5447818b07315f0 (HEAD -> master, origin/master, origin/HEAD)
Author: Gleb Lesnikov <entroopy@gmail.com>
Date:   Mon May 5 10:53:07 2025 +0400
```

<img width="1792" alt="Image" src="https://github.com/user-attachments/assets/fd4d9c8b-a5cc-4712-ad42-058dda5b6ba3" />

* How did you install Redash:

```
% docker compose build            
[+] Building 4.0s (54/67)                                                                                                                                                                                                           docker:desktop-linux
 => [server internal] load build definition from Dockerfile                                                                                                                                                                                         0.0s
 => => transferring dockerfile: 3.67kB                                                                                                                                                                                                              0.0s
 => [worker internal] load metadata for docker.io/library/node:18-bookworm                                                                                                                                                                          1.4s
 => [scheduler internal] load metadata for docker.io/library/python:3.10-slim-bookworm                                                                                                                                                              1.4s
 => [server internal] load .dockerignore                                                                                                                                                                                                            0.0s
[...]

PR #7434, opened 2025-05-05
Fix unexpected /app shadowing in compose.yaml (#7433)

ongrid:
Paths within the container got overwritten by volume mount, which caused inconsistencies with the expected contents of directories populated on buid stage.

For example, CSS and graphics may not load — see issue getredash/redash#7433.

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ X] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ X] Manually
- [ ] N/A

```
make compose_build
make create_database
make up
```

#7433

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -5,8 +5,6 @@ x-redash-service: &redash-service
     context: .
     args:
       skip_frontend_build: "true"  # set to empty string to build
-  volumes:
-    - .:/app
   env_file:
     - .env
 x-redash-environment: &redash-environment


compose.yaml
```diff
@@ -5,8 +5,6 @@ x-redash-service: &redash-service
     context: .
     args:
       skip_frontend_build: "true"  # set to empty string to build
-  volumes:
-    - .:/app
```
eradman, line 9: This bindmount is not a mistake, it is required for live-reload in a dev environment

ongrid: @eradman what about using separate path for frontend artifacts like `/www/html` [here](https://github.com/getredash/redash/blob/0900178d245615ba52bcaf62c73203c7d99b896f/Dockerfile#L115) and [here](https://github.com/getredash/redash/blob/0900178d245615ba52bcaf62c73203c7d99b896f/redash/handlers/webpack.py#L6) to avoid shadowing by mounted volume? Let me know if you accept this approach

eradman: We need to understand #7433 better. What is the reason for setting `skip_frontend_build: ""`?

PR #7437, opened 2025-05-16
feat: Add search term highlighting in table results

LEEAHRI:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
When viewing large datasets, it's difficult to quickly locate search matches. I've added text highlighting for better readability, which currently applies to all searchable columns. I propose making this highlighting feature configurable. Looking for feedback on implementing this as an optional setting.

## UI
image


deed02392:
This is a cool feature. Can you please use the [`<mark>`](https://www.w3schools.com/TAgs/tag_mark.asp) tag, which is designated for this use-case, and will let the front-end appropriately decide how to style the text to highlight?

PR #7439, opened 2025-05-16
Make NULL values visible

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description
Before this PR, on table visualization, NULL value is not visible but shown just as empty string(""). We cannot distinguish between empty string("") and NULL value on the table view.

This PR makes NULL values visible so that we can know the value is not empty string but NULL value.
This is common behavior on many SQL tools.

## How is this tested?
- [x] Manually

I run SQL query generate NULL values like below, and I confirmed that the data is shown as NULL.
```
select
    1,
    'aa',
    null,
    cast(null as int64),
    cast(null as string),
    cast(null as date),
    cast(null as timestamp)
```
image


eradman:
I really like this idea, but I bet we need to make this configurable (e.g. https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-META-COMMAND-PSET-NULL)

We could store this in `organizations.settings`

yoshiokatsuneo:
@eradman 

Thank you for your comment !

But, I don't know any GUI / Web base tools that "NULL" text is configurable. 
(BigQuery / Snowflake / Looker Studio / phpMyAdmin / Sequel Ace / Athena, etc... )

And so, I guess there is little demand to make NULL text configurable.

Also, I think it makes redash complex if we make every special value configurable, like emtpy string, NaN, Infinity, true, false ...

Also, for GUI tools, we can make many things configurable, like color, font, style configurable.

Do you think we should make these styles configurable ?

eradman:
What I am anticipating is that some Redash users prefer the _current_ handling of null values, and we will need a workaround.

There are already some settings that are not  configurable from the Redash GUI,  but can be changed. For example the default number format can be set:

```sql
UPDATE organizations
SET settings = jsonb_set(settings::jsonb, '{settings,float_format}', '"0.0"', true);
```

yoshiokatsuneo:
@eradman 

> What I am anticipating is that some Redash users prefer the current handling of null values, and we will need a workaround.

Thank you. I see !
So, I guess we can just have a option like 'visible_null_value' boolean value. How about ?

eradman:
> So, I guess we can just have a option like 'visible_null_value' boolean value

I had in mind a text value: 

```json
  "null_value": "null"
```

This would be used by `NullValueComponent()`

yoshiokatsuneo:
> > So, I guess we can just have a option like 'visible_null_value' boolean value
> 
> I had in mind a text value:
> 
> ```json
>   "null_value": "null"
> ```
> 
> This would be used by `NullValueComponent()`

@eradman 
Thanks !
I'll implement like that !

eradman:
@yoshiokatsuneo, after thinking about this further realized we need to get some feedback from other contributors--people are going to care about this and how it can be enabled/configured.

Please don't spend too much time on this, let's see if others will comment!

jonahgreenthal:
I would appreciate the ability to see nulls (distinctly from empty or blank strings). I don't feel strongly about how it's implemented, but I would prefer to be able to distinguish an actual null value from the string `"null"` or similar.

yoshiokatsuneo:
@jonahgreenthal 

Thank you for your feedback !
In this PR, I changed the text color to gray,  and the text style to italic for the NULL value so that we can distinguish "null" string from NULL value, like below. This is similar style as most other GUI / Web-based SQL tools. How about ?

image

jonahgreenthal:
Yep, that works for me. I saw you had done it, I just wanted to emphasize that I thought it was an important aspect, but I probably didn't communicate it very well.

yoshiokatsuneo:
@eradman 

Hello

> let's see if others will comment!

I think we have waited for about 50 days, and no one care. Isn't it enough time ?

eradman:
> > let's see if others will comment!
> 
> I think we have waited for about 50 days, and no one care. Isn't it enough time ?

I don't understand the question.

I'll take another look at this PR early next week.

yoshiokatsuneo:
@eradman 
> I don't understand the question.

Sorry,
I mean, you have suggested that we should wait for comments for others, but I believe we have waited long enough.
 
> I'll take another look at this PR early next week.

Thank you !

eradman:
We need a way to configure the representation of NULL values. Attached is the start of the patch to make this happen
[null-value-diff.txt](https://github.com/user-attachments/files/21222581/null-value-diff.txt)

yoshiokatsuneo:
@eradman 

Thank you very much for the patch !
I applied the patch and adjusted the code.

https://github.com/getredash/redash/pull/7439/commits/56b64e4be6d9a242ed887bb3d49e50baa575772c

viz-lib/src/lib/value-format.tsx
```diff
@@ -5,16 +5,24 @@ import numeral from "numeral";
...
+
 numeral.options.scalePercentBy100 = false;
 
 // eslint-disable-next-line
 const urlPattern = /(^|[\s\n]|<br\/?>)((?:https?|ftp):\/\/[\-A-Z0-9+\u0026\u2019@#\/%?=()~_|!:,.;]*[\-A-Z0-9+\u0026@#\/%=~()_|])/gi;
 
 const hasOwnProperty = Object.prototype.hasOwnProperty;
 
+function NullValueComponent() {
+  return <span className="text-muted"><i>{visualizationsSettings.nullValue}</i></span>;
```
eradman: Instead of using the `<i>` tag, I think we should define a class `.display-as-null` in `viz-lib/src/visualizations/table/renderer.less`

redash/settings/organization.py
```diff
@@ -27,6 +27,7 @@
 TIME_FORMAT = os.environ.get("REDASH_TIME_FORMAT", "HH:mm")
 INTEGER_FORMAT = os.environ.get("REDASH_INTEGER_FORMAT", "0,0")
 FLOAT_FORMAT = os.environ.get("REDASH_FLOAT_FORMAT", "0,0.00")
+NULL_VALUE = os.environ.get("REDASH_NULL_VALUE", "null")
```
eradman, line 30: Let's make the default an empty string so that only changes if this parameter is set in `.env`

eradman (changes requested):

viz-lib/src/lib/value-format.tsx
yoshiokatsuneo: @eradman 
Thank you very much for the review !

I just changed the code to create `.display-as-null` class in `lib/src/visualizations/table/renderer.less` instead of `<i>` tag.

https://github.com/getredash/redash/pull/7439/commits/baabbdf1d836a3d473572ccc2b60370d0000c739

redash/settings/organization.py
yoshiokatsuneo: @eradman 

Thank you for your comment!

I think showing NULL by default is preferable for most users because NULL value is visible in most DWH/database visibility tools and web services(both commercial and open source) including:
 - BigQuery
 - Snowflake
 - Databricks
 - Looker Studio
 - phpMyAdmin
 - Sequel Ace
 - Athena
 - superset
 - metabase, etc.

Even on Databricks SQL (that is based on Redash I think), NULL value is visible.

May I hear your thought and/or preference, if possible ?

Thank you !

eradman: I see your point, and I would also find it useful to be able to see NULL. If we get pushback from users we can reconsider this default.

yoshiokatsuneo: @eradman 
Thank you for seeing !

PR #7484, opened 2025-07-22
Sort Dashboard and Query tags by name

eradman:
## What type of PR is this? 

- [x] Feature

## Description

Finding a specific tag name is difficult when ordering by count. Finding a specific tag in a long list is much easier when sorted by name.

## How is this tested?

- [x] Manually

tag_order


yoshiokatsuneo:
Thank you !

I agree that there isn't much sense to order by count, and ordering by name is better.

I tested and confirmed that the PR works as intended.

yoshiokatsuneo (approved):

yoshiokatsuneo:
@eradman 

Thank you !

I agree that there isn't much sense to order by count, and ordering by name is better.

I tested and confirmed that the PR works as intended.

eradman:
Thanks for the review @yoshiokatsuneo. We have been deploying a build of Redash with this patch applied and users have been happy with it.

Issue #7486, opened 2025-07-30
Cannot remove password from mongodb data source configuration in order to use 509 credentials


dobesv:
### Issue Summary

We are trying to connect to mongodb using X509 credentials.  When we try to connect we get an error "Passwords are not supported by MONGODB-X509".  It seems like redash is sending a password in some way even though we have configured X509 key and cert on the connection string.

This happened when I changed an existing data source that had a password to use X509 creds and I tried to remove the password by blanking out the password field.

When I created a brand new data source I was able to connect.

### Steps to Reproduce

1. Setup a mongodb server that uses X509 client certs for authentication
2. Setup redash to connect to it
3. Set a password
4. Clear the password
5. Try to use the connection
6. It seems like it still remembers you had a password set

### Technical details:

* Redash Version: 10.1.0.b50633
* Browser/OS: docker image linux
* How did you install Redash: docker image running in kubernetes

yoshiokatsuneo:
@dobesv 

Thank you for your report !
I just created a PR to fix the issue.

https://github.com/getredash/redash/pull/7487

PR #7487, opened 2025-07-30
MongoDB: fix for empty username/password (Issue #7486)

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
Fix the issue that empty username/password is not handled correctly on MongoDB.

The issue is reported as below:

https://github.com/getredash/redash/issues/7486

## How is this tested?

- [x] Manually

Tested for both empty username/password, and non-empty username/password, and confirmed that it works without problem.


eradman (approved):
Not tested myself, but this change make sense

yoshiokatsuneo:
@eradman 

Thank you for the quick review !!

PR #7489, opened 2025-07-30
Do not display INFORMATION_SCHEMA tables

eradman:
## What type of PR is this? 

- [x] Bug Fix

## Description

As with other query runners, do not show system tables in the schema list.

## How is this tested?

- [x] Manually

Confirm that only database tables are defined

clickhouse-schema


zachliu:
i'm unable to reproduce this issue. the `information_schema` or `INFORMATION_SCHEMA` don't show up even without this fix :thinking: 

ah...you must be using the superuser

eradman:
Indeed! I'm connecting with a superuser account. You can see this with a regular user by granting permission:

```sql
GRANT SELECT ON information_schema.* TO user;
GRANT SELECT ON INFORMATION_SCHEMA.* TO user;
```

PR #7506, opened 2025-08-23
Allow HTTP request line more than 4096 bytes

yoshiokatsuneo:
## What type of PR is this? 

- [x] Bug Fix

## Description

When there are many parameters on dashboard or query, we can easily get gunicorn error below.

```
Bad Request

Request Line is too large (4970 > 4096)
```

image

This PR fixes the issue by changing gunicorn `limit_request_line` setting to 0(unlimited), and is configurable using "REDASH_GUNICORN_LIMIT_REQUEST_LINE" variable.

As, redash compose.yaml anyway put nginx before gunicorn and does not directly exposed to the client, I think there is no need to set limit on gunicorn.

## How is this tested?

- [x] Manually

By this PR, dashboard or query having long parameter does not cause error.


eradman:
Would it be better to retain the default of 4094? With this change it would be configurable using `REDASH_GUNICORN_LIMIT_REQUEST_LINE`. _unlimited_ might be fine, I don't know what the adverse affects could be.

yoshiokatsuneo:
@eradman 

Internet Explorer have the URL limit of 2048 bytes. But, I think there is little Internet Explorer user nowadays.
RFC-7230 recommends to support request line more than 8,000 bytes.
And, it looks recent browser like Chrome support more than 32,767 bytes.

I don't come up with the reason to limit URL length to like 4096 bytes...

yoshiokatsuneo:
@eradman Thanks a lot for the review!

PR #7514, opened 2025-08-27
Update plotly.js to 3.1.0

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature
- [x] Other

## Description
Update plotly.js to the latest version (3.1.0) so that we can catch up the changes on the library and easily maintain.

## How is this tested?

- [x] Manually


viz-lib/src/visualizations/chart/plotly/index.ts
```diff
@@ -8,7 +8,6 @@ import updateAxes from "./updateAxes";
 import updateChartSize from "./updateChartSize";
 import { prepareCustomChartData, createCustomChartRenderer } from "./customChartUtils";
 
-// @ts-expect-error ts-migrate(2339) FIXME: Property 'setPlotConfig' does not exist on type 't... Remove this comment to see the full error message
```
lucydodo, line 11: Is this change intentional?

yoshiokatsuneo: @lucydodo 

Thank for seeing the PR !

> Is this change intentional?

Yes.
As the latest `@types/plotly.js` have setPlotConfig type definition , we need to remove the line to avoid the error.

lucydodo: Got it, now we can merge this PR :)

yoshiokatsuneo: @lucydodo 

Thank you for the review !!

PR #7516, opened 2025-08-29
Fix stacking bar chart

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
Fix the issue that stacking bar chart is not stacked.
This issue was introduced by https://github.com/getredash/redash/pull/7514 , and this PR fixes the issue.

## How is this tested?

- [x] Manually

I tested by running following query on Query Result datasource. and make stacking bar chart visualization.

```sql
select 1 as id, 10 as val, 'a' as grp
union all
select 2 as id, 30 as val, 'a' as grp
union all
select 3 as id, 20 as val, 'a' as grp
union all
select 1 as id, 11 as val, 'b' as grp
union all
select 2 as id, 31 as val, 'b' as grp
union all
select 3 as id, 21 as val, 'b' as grp
```

Setting:
image

After applying this PR:
image


eradman (approved):
This fix works! Thanks for catching this @yoshiokatsuneo

yoshiokatsuneo:
@eradman Thank you for your quick review!

PR #7519, opened 2025-09-03
Fix: null is not shown for text with "Allow HTML content"

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
This PR fixes the issue that NULL value is not shown as "null" but shown as "[object Object]" for Text column with "Allow HTML content" checked.

## How is this tested?
- [x] Manually

1. Create a following query on QueryResult dataset, for example.

```sql
select null as col1, 'aa<b>bb</b>cc' as col2
```

2. Set Column type as Text(default), and check "Allow HTML content".

image

3. See the output for the NULL value.

Before this PR, NULL but is shown wrongly as "[object Object]".
image

After this PR, NUL value is shown as "null" as intended.
image

### DIFF
--- a/viz-lib/src/visualizations/table/columns/text.tsx
+++ b/viz-lib/src/visualizations/table/columns/text.tsx
@@ -52,7 +52,7 @@ export default function initTextColumn(column: any) {
   function TextColumn({ row }: any) {
     // eslint-disable-line react/prop-types
     const { text } = prepareData(row);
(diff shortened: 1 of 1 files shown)


eradman (approved):
Confirmed. This is an important fix!

yoshiokatsuneo:
> Confirmed. This is an important fix!

@eradman 

Thank you very much for so quick review !!

PR #7520, opened 2025-09-03
keep ordering on search

yoshiokatsuneo:
## What type of PR is this? 

- [x] Feature

## Description

Before this PR, searching queries/dashboard,etc does not respect specified ordering. For searching queries without multi-byte search option make some sense because it uses server side full text search ordering.
But, other that that, search result is ordered by id and the oldest row is in the beginning of the result list. I think it is not user friendly at all.

This PR respect ordering specified by user(in "created_at" descending order by default) on search.

## How is this tested?

- [x] Manually

1. Check "Enable multy-byte search for query" feature flag.

image

2. Reload the browser to update configuration on the client side.
 
3. Go to query listing page

4. Search by some term

5. I confirmed that the query ordering is kept, and the ordering "created_at" descending order by default, and the URL have "order=-created_at" parameter, as expected.

image

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -46,7 +46,7 @@ server() {
   MAX_REQUESTS=${MAX_REQUESTS:-1000}
   MAX_REQUESTS_JITTER=${MAX_REQUESTS_JITTER:-100}
   TIMEOUT=${REDASH_GUNICORN_TIMEOUT:-60}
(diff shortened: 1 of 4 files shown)


eradman:
To me it appeared that the search ordering still changes (with and without multi-byte checked) when I type a search keyword

before_search
after_search

yoshiokatsuneo:
@eradman 

Thank you for testing!
Did you reload the browser after enabling `Enable multy-byte search for query` ? (I just add "reload browser" to the test instruction.)

Also, can you confirm that the query parameter is like below, and it always includes `order=-created_at` ?

Before search: http://localhost:5001/queries?order=-created_at&page=1&page_size=20

After search: http://localhost:5001/queries?order=-created_at&page=1&page_size=20&q=q

eradman:
I did not refresh the browser after enabling multi-byte search. That seemed to work

yoshiokatsuneo:
@eradman Thank you for the quick review, again !

PR #7521, opened 2025-09-05
PostgreSQL: Rely on information_schema.columns for views and foreign tables

eradman:
## What type of PR is this? 
- [x] Bug Fix

## Description

Prevent duplicate entries from appearing in the schema list.  _Materialized views_ are the only table-like object not found `information_schema`. Also ensure that the schema and table found in `information_schema` is accessible by the current user.

## How is this tested?

- [x] Manually

_Before_

schema_list_1

_After_

schema_list_2

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -205,24 +205,15 @@ class PostgreSQL(BaseSQLQueryRunner):
 
     def _get_tables(self, schema):
         """
-        relkind constants per https://www.postgresql.org/docs/10/static/catalog-pg-class.html
-        r = regular table
-        v = view
+        relkind constants from https://www.postgresql.org/docs/current/catalog-pg-class.html
         m = materialized view
-        f = foreign table
-        p = partitioned table (new in 10)
-        ---
-        i = index
-        S = sequence
-        t = TOAST table
-        c = composite type
         """
 
         query = """
-        SELECT s.nspname as table_schema,
-               c.relname as table_name,
-               a.attname as column_name,
-               null as data_type
+        SELECT s.nspname AS table_schema,
+               c.relname AS table_name,
+               a.attname AS column_name,
+               NULL AS data_type
         FROM pg_class c
         JOIN pg_namespace s
         ON c.relnamespace = s.oid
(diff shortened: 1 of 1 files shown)


yoshiokatsuneo:
@eradman 

Thank you for the PR !
It looks, the schema for materialized view is not shown after applying the PR.  Do you have any idea ?

eradman:
Oh odd, materialized views do not show up in `information_schema`. I think this means we still need to select from `pg_class` to get this object type

```
WHERE c.relkind IN ('m')
```

redash/query_runner/pg.py
```diff
@@ -204,20 +204,6 @@ def _get_definitions(self, schema, query):
...
-        m = materialized view
-        f = foreign table
-        p = partitioned table (new in 10)
-        ---
-        i = index
-        S = sequence
-        t = TOAST table
-        c = composite type
-        """
-
```
yoshiokatsuneo, line 220: @eradman 

In my opinion, I just personally feel the comment what 'm' means is helpful. But it is just a feeling. I can approve it.

yoshiokatsuneo (approved):
@eradman 

I have tested for regular table, view, materialized view, partitioned table, foreign table, and partition table. And, I confirmed that it works well.

Thank you !

redash/query_runner/pg.py
eradman: fair enough, I'll put part of that comment back

redash/query_runner/pg.py
```diff
@@ -231,7 +217,7 @@ def _get_tables(self, schema):
         ON a.attrelid = c.oid
         AND a.attnum > 0
         AND NOT a.attisdropped
-        WHERE c.relkind IN ('m', 'f', 'p')
+        WHERE c.relkind IN ('m')
```
eradman: Would be neater to write `= 'm'` sine there is only one item

yoshiokatsuneo (approved):
@eradman 

Thank you the PR, again !

redash/query_runner/pg.py
```diff
@@ -243,6 +234,8 @@ def _get_tables(self, schema):
                data_type
         FROM information_schema.columns
         WHERE table_schema NOT IN ('pg_catalog', 'information_schema')
+        AND has_table_privilege(table_schema || '.' || table_name, 'select')
```
deed02392: where we only `grant` to `select` on a subset of columns in a table, the entire table is hidden from the schema because of this line. this constitutes a functional regression for us but it does indicate an improvement could be made by merging in columns which are visible and listed via `information_schema.column_privileges`

PR #7525, opened 2025-09-08
Add range slider to the chart

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description

At least for us, it is quite common to use Redash for time-series data.
And, it is annoying to scale-in and scale-out without seeing whole time range.

This PR add the feature to toggle range slider to the charts.

## How is this tested?

- [x] Manually

1. create data using following SQL on QueryResult data source.
```
WITH RECURSIVE seq(x) AS (
  SELECT 0
  UNION ALL
  SELECT x + 1 FROM seq WHERE x < 365
)
SELECT 
  date('2025-01-01', '+' || x || ' days') AS ts,
  abs(random() % 100) AS value1,
  abs(random() % 50) AS value2
FROM seq;
```

2. create a visualization like below.

image

3. Toggle range slider feature.

image

image

image

https://github.com/user-attachments/assets/414cefca-ece6-48f1-9c15-9832643bbea6

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -46,7 +46,7 @@ server() {
   MAX_REQUESTS=${MAX_REQUESTS:-1000}
   MAX_REQUESTS_JITTER=${MAX_REQUESTS_JITTER:-100}
   TIMEOUT=${REDASH_GUNICORN_TIMEOUT:-60}
-  exec /usr/local/bin/gunicorn -b 0.0.0.0:5000 --name redash -w${REDASH_WEB_WORKERS:-4} redash.wsgi:app --max-requests $MAX_REQUESTS --max-requests-jitter $MAX_REQUESTS_JITTER --timeout $TIMEOUT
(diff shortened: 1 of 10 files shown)


eradman:
With 720 data points the slider has a significant lag and is CPU intensive. (Testing with Intel i7-1260P / OpenBSD 7.7 / Firefox 141). We may need to figure out how to profile Plotly.

yoshiokatsuneo:
@eradman 

Thank you for testing !
I don't see any lag or performance issue at all, on my Chrome / Firefox / Safari , Apple M4 Pro.

Do you see the issue on the following Plotly.js demo site, too ?
https://plotly.com/javascript/range-slider/

I think you can use Performance tool on Chrome Dev Tool like below for profiling.

image

eradman:
The Plotly.js demo range slider has no perceptible delay, only minimal CPU usage. I will try the Chrome profiling tool later

eradman:
It turns out that I had accidentally found a combination of parameters that is slow, and is nothing to do with this feature!

fast-viz

The hazard comes from these three parameters:
* Chart type: Scatter
* Group By:  _col1_
* Scale: Category

If I change _Category_ to _Datetime_ or remove the Group By it's fast again

yoshiokatsuneo:
@eradman Thank you for dig into the situation, and make the reason clear !

PR #7530, opened 2025-09-11
Fix/too many history replace state

yoshiokatsuneo:
## What type of PR is this? 

- [x] Bug Fix

## Description

On Safari, when a dashboard have many widgets, following error occurs because history.replaceState() is called many times.

```
Attempt to use history.replaceState() more than 100 times per 10 seconds
```
スクリーンショット 2025-09-12 1 16 00

This PR fixes the issue by calling history.replaceState() only when the url is really changed.

## How is this tested?
- [x] Manually

### DIFF
--- a/client/app/services/location.js
+++ b/client/app/services/location.js
@@ -9,7 +9,7 @@ function normalizeLocation(rawLocation) {
   const result = {};
 
   result.path = pathname;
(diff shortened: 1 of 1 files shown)


yoshiokatsuneo:
@eradman Thank you for your quick review and testing, as always !

PR #7538, opened 2025-09-22
BigQuery: show column description(comment) on Schema Browser

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description

Show column description(column comment) on Schema Browser in addition to column name and column type so that we can easily know what the column means.
(on BigQuery for now.)

Nowadays, column description or other data catalog is getting more important as it makes Data Ware House more AI agent friendly and human friendly. Showing such a information motivate us to improve column description, data catalog, and data governance.

## How is this tested?

- [x] Manually

image


client/app/components/queries/SchemaBrowser.jsx
```diff
@@ -70,18 +71,25 @@ function SchemaItem({ item, expanded, onToggle, onSelect, ...props }) {
...
+                  placement="rightTop"
+                >
+                  <PlainButton
+                    key={columnName}
+                    className="table-open-item"
+                    onClick={(e) => handleSelect(e, columnName)}
+                  >
                     <div>
                       {columnName} {columnType && <span className="column-type">{columnType}</span>}
+                      {columnDescription && <span className="column-description"> - {columnDescription}</span>}
```
eradman: Potentially very useful. I feel that there often is not enough room to fit a column comment. Example:

> atime  TIMESTAMP WITH TIME ZONE - Last acc...

Adding the comment to  the Tooltip may be enough

_Maybe_ add a toggle to show datatype or column description

eradman (changes requested):

client/app/components/queries/SchemaBrowser.jsx
yoshiokatsuneo: @eradman 

Thank you for the review !
As there are some options how to show the column description in the Schema Browser, I changed to show the column description only on Tooltip for now.
( I think we can think how to show as a separate topic if needed.)

https://github.com/getredash/redash/pull/7538/commits/065a645c12a836e4969776abccc2f64406414659

image

eradman: > ( I think we can think how to show as a separate topic if needed.)

Agreed

yoshiokatsuneo: @eradman 
Thanks !

PR #7540, opened 2025-09-23
BigQuery: support multiple dataset locations

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
Before the PR, Redash does not work when the BigQuery project have datasets on the multiple locations, or non-US location without explicitly specifying location on the configuration.
This is because BigQuery cannot run a query referring multiple locations.

This PR supports multiple dataset locations, and non-US location without explicitly specifying the location.

## How is this tested?

- [x] Manually

I confirmed that Redash can show schema list on Schema Browser and run queries, even if  the BigQuery project have datasets on both US and non-US location.


eradman (approved):
Approved!

yoshiokatsuneo:
@eradman Thank you for your quick
review, always!

PR #7541, opened 2025-09-24
BigQuery: Remove "Job ID" metadata on annotaton to avoid cache misses

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description
On BigQuery datasource, when "Use Query Annotation" is enabled, BigQuery's query result cache does not owkr because the query sent to BigQuery is always different because the annotation have different "Job ID" on the each query.

This PR fixes the BigQuery caching issue by removing "Job ID" from the query annotation so that the query sent to BigQuery is always the same and the BigQuery cache hits.

## How is this tested?

- [x] Manually

I run the same query twice for the BigQuery data source with Use Query Annotation enabled, and I confirmed that the "data scanned" for the second query is zero.

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -156,6 +156,11 @@ class BigQuery(BaseSQLQueryRunner):
             "secret": ["jsonKeyFile"],
         }
 
+    def annotate_query(self, query, metadata):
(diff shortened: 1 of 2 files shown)


yoshiokatsuneo:
@eradman 
Thank you for the quick review !

PR #7546, opened 2025-09-27
Advanced query search syntax for multi byte search

yoshiokatsuneo:
## What type of PR is this? 
- [x] Feature

## Description
Before the PR, query search only support "ILIKE" based filtering. There is no way to search multiple terms, queries.

This PR make it possible to search query by specifying column name, exact match, multiple term search.

For example:
            - word
            - name:word
            - query:word
            - "multiple words"
            - name:"multiple words"
            - word1 word2 word3
            - word1 "multiple word" query:"select foo"

## How is this tested?
- [x] Manually

image
image
image
image
image


eradman:
Very nice. Do we also need to make a corresponding change to documentation?

yoshiokatsuneo:
> Very nice. Do we also need to make a corresponding change to documentation?

@eradman 

Thank you for the comment !
I just tried to added a tooltip as a help. How about ?

https://github.com/getredash/redash/pull/7546/commits/43148ba6ac30b5346a392f5747acf42601e521ab

image

eradman:
I think this is too much information for a tooltip.

For detailed information the information that open for "Number format" (under Edit Visualization) works nicely but that seems like overkill for this feature.

yoshiokatsuneo:
@eradman 

Thank you for your feedback !
Do you have any suggestion where to put the information ?

It may be a bit tricky as this feature is valid only when multi byte search is enabled.

eradman:
> Do you have any suggestion where to put the information ?
> It may be a bit tricky as this feature is valid only when multi byte search is enabled.

I don't have a better idea for this. We could add it as a hidden/undocumented feature, but that's not a good habit.

yoshiokatsuneo:
@eradman 

 May we add the document on the user guide and open it as a side-bar, as Formatting Guide on the alert page ?

image

eradman:
> May we add the document on the user guide and open it as a side-bar

I think this is the right way to go. To start we can add this hidden feature (no tooltip), just a source code comment, and update documentation later.

yoshiokatsuneo:
@eradman 

Thank you! So, at first, I just removed the tooltip for now.

redash/models/__init__.py
```diff
@@ -644,6 +645,43 @@ def outdated_queries(cls):
...
+    def _do_multi_byte_search(cls, all_queries, term, limit=None):
+        # term examples:
+        #    - word
+        #    - name:word
+        #    - query:word
+        #    - "multiple words"
+        #    - name:"multiple words"
+        #    - word1 word2 word3
+        #    - word1 "multiple word" query:"select foo"
+        tokens = re.findall(r'(?:([^:\s]+):)?(?:"([^"]+)"|(\S+))', term)
```
eradman, line 658: Is there a better way to break _words_ and _"multiple words"_ into tokens? This regex appears to work, may also contain hard-to-find edge cases

yoshiokatsuneo: @eradman 

Thank you for the review!

At first, the query string matches one of those.
- Qualifier:Keyword => (?:([^:\s]+):)  AND (?:"([^"]+)"|(\S+))
- Keyword    => (?:"([^"]+)"|(\S+))

And, the keyword matches one of those:
- "any characters"  => "([^"]+)"
- NonSpaceCharacters  => (\S+)

I think this covers the all 4 cases( 2 cases x 2 cases).

Or, we may implement simple tokenizer. That makes more flexible, but it need a bit of code.

eradman: > Or, we may implement simple tokenizer. That makes more flexible, but it need a bit of code.

The regex is very compact. We can refactor this in the future if we need more flexibility

eradman (approved):

redash/models/__init__.py
yoshiokatsuneo: @eradman 

I see. Thank you !

yoshiokatsuneo:
And, I tried to create a PR for the documentation.
https://github.com/getredash/website/pull/844

PR #7551, opened 2025-10-02
Query Serach: avoid concurrent search API request

yoshiokatsuneo:
## What type of PR is this? 

- [x] Feature

## Description
Before the PR, on Query/Dashboard search,  frontend may calls multiple Query Search API calls at the same time, and it may increase the load of backend.
This PR avoid calling multiple APIs concurrently by pending API request until previous API request is finished.

## How is this tested?

- [x] Manually


yoshiokatsuneo:
@eradman Thank you for your review and the testing!!

PR #7552, opened 2025-10-02
SchemaBrowser: on column comment tooltip, show newlines correctly

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
On Schema Browser, column comment tooltiop for newlines are not shown as newlines.
This PR fixes the issue by adding "white-space: pre-line" style to the tooltiop.

## How is this tested?
- [x] Manually


yoshiokatsuneo:
@eradman Thanks a lot!

PR #7555, opened 2025-10-06
Use standard PostgreSQL image and drop clean-all target

eradman:
## What type of PR is this? 

- [x] Other

## Description

Auto-upgrade of the database add unnecessary complication for a dev environment. This also solves a current problem with the pgautoupgrade image:

https://github.com/getredash/redash/actions/runs/18280518797/job/52042371707?pr=7532

## How is this tested?

- [x] E2E Tests (Cypress)


yoshiokatsuneo (approved):
Thank you !
It looks just good !

PR #7558, opened 2025-10-12
Avoid updating latest_query_data_id while saving the query

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
Before this PR, saving query always update latest_query_data_id sent from frontend. But, as the "latest_query_data_id" sent from frontend is not always correct. 
Especially, after executing the query, the "latest_query_data_id" on the backend is updated, but the one on the frontend is not updated. As a result, saving query after executing reverts the "latest_query_data_id" to the older one that frontend remember.

This PR fixes the issue by filtering out the latest_query_data_id parameter on the backend.

## How is this tested?

- [x] Manually

### DIFF
--- a/redash/handlers/queries.py
+++ b/redash/handlers/queries.py
@@ -229,6 +229,7 @@ class QueryListResource(BaseQueryListResource):
             "api_key",
             "visualizations",
             "latest_query_data",
+            "latest_query_data_id",
             "last_modified_by",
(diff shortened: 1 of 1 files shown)


yoshiokatsuneo:
close the case, in favor of https://github.com/getredash/redash/pull/7559

PR #7559, opened 2025-10-12
Clear latest_query_data if queryResult.quey is differenct from query.query

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description
Saving query always updates `queries.latest_query_data_id` sent from frontend. But, as the "latest_query_data_id" sent from frontend is not always up to date.
The outdated latest_query_data_id happens when the query string on the frontend and the query string on the query_results table is different, for example by adding LIMIT or annotation, parameter, etc...
As a result, saving query after executing can revert the "latest_query_data_id" to the older one that frontend remember.

This PR fixes the issue by clearing the query.latest_query_data_id on the frontend when latest_query_data_id is not updated on the query execution so that frontend does not send the outdated latest_query_data_id to the backend.

Note that we need to set the`latest_query_data_id` not to `null` but to `undefined`, as sending `null` to backend set NULL value instead of ignoring the value.

## How is this tested?

- [x] Manually

### DIFF
--- a/client/app/pages/queries/hooks/useQueryExecute.js
+++ b/client/app/pages/queries/hooks/useQueryExecute.js
@@ -61,7 +61,7 @@ export default function useQueryExecute(query) {
       },
     });
 
-    const onStatusChange = status => {
+    const onStatusChange = (status) => {
       if (queryResultInExecution.current === newQueryResult) {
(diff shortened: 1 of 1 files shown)


yoshiokatsuneo:
Close in favor of https://github.com/getredash/redash/pull/7560

PR #7560, opened 2025-10-12
Update queries.latest_query_data on save

yoshiokatsuneo:
## What type of PR is this? 
- [x] Bug Fix

## Description

Saving query always updates `queries.latest_query_data_id` sent from frontend. But, the "latest_query_data_id" sent from frontend is not always up to date.
The outdated latest_query_data_id happens when the query string on the frontend and the query string on the query_results table is different, for example by adding LIMIT or annotation, parameter, etc...
As a result, saving query after executing can revert the "latest_query_data_id" to the older one that frontend remember.

This PR fixes the issue by updating `queries.latest_query_data_id` on backend when saving the query.

## How is this tested?

- [x] Manually

Example 1:

1. Create a data source like PostgreSQL
2. Click "New Query" from the side menu.
3. Check "LIMIT 1000"
4. Write a SQL like "select 12345"
5. Click "Execute" to execute the query.
6. Click "Save" to save the query
7. Reload the page

Before the PR, query result is cleared and not shown.
After the PR, query result is shown as intended.

Example 2:

1. Create a data source like PostgreSQL
2. Click "New Query" from the side menu.
3. Check "LIMIT 1000"
4. Write a SQL like "select 12345"
6. Click "Save" to save the query
5. Click "Execute" to execute the query.
7. Reload the page. (Result is 12345 as intended)
8. Change query to "select 99999"
9. Click "Execute" to execute the query (Result is 99999 as intended)
10. Reload the page

Before the PR, query result is reverted to 12345 instead of 99999.
After the PR, query result "99999" is shown as intended.

### DIFF
--- a/client/cypress/integration/dashboard/filters_spec.js
+++ b/client/cypress/integration/dashboard/filters_spec.js
@@ -23,7 +23,7 @@ describe("Dashboard Filters", () => {
       name: "Query Filters",
       query: `SELECT stage1 AS "stage1::filter", stage2, value FROM (${SQL}) q`,
     };
-    cy.createDashboard("Dashboard Filters").then(dashboard => {
+    cy.createDashboard("Dashboard Filters").then((dashboard) => {
       createQueryAndAddWidget(dashboard.id, queryData)
         .as("widget1TestId")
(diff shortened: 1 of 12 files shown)


yoshiokatsuneo:
@eradman 

Thank you for the review and the tricky testing !

PR #7570, opened 2025-11-12
Enhance dashboard parameter handling: persist updated values and apply saved parameters

denisov-vlad:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Added support for saving dashboard parameters after clicking the Apply button. This is useful when using query templates and inserting them into different dashboards. Without this logic, there are two solutions: cloning queries or making parameters static.

Now they are used in the following order: url, dashboard parameters, query parameters.

- Queue dashboard-level parameter edits while the dashboard is in edit mode.
- Persist the queued values only when “Done Editing” is clicked, keeping Query and Dashboard editors aligned.
- Surface any save failures via notifications and restore the queued values so nothing is lost silently.

## How is this tested?

- [x] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

1. Create query with params
2. Create dashboard and add query widget there
3. Check the default value and change it
4. Open new dashboard from dashboards list
5. You'll see changed value

### DIFF
--- a/client/app/pages/dashboards/DashboardPage.jsx
+++ b/client/app/pages/dashboards/DashboardPage.jsx
@@ -31,7 +31,8 @@ function DashboardSettings({ dashboardConfiguration }) {
       <Checkbox
         checked={!!dashboard.dashboard_filters_enabled}
         onChange={({ target }) => updateDashboard({ dashboard_filters_enabled: target.checked })}
-        data-test="DashboardFiltersCheckbox">
+        data-test="DashboardFiltersCheckbox"
+      >
         Use Dashboard Level Filters
       </Checkbox>
     </div>
@@ -90,7 +91,7 @@ function DashboardComponent(props) {
 
   const [pageContainer, setPageContainer] = useState(null);
   const [bottomPanelStyles, setBottomPanelStyles] = useState({});
-  const onParametersEdit = parameters => {
+  const onParametersEdit = (parameters) => {
     const paramOrder = map(parameters, "name");
     updateDashboard({ options: { ...dashboard.options, globalParamOrder: paramOrder } });
   };
@@ -175,7 +176,7 @@ function DashboardPage({ dashboardSlug, dashboardId, onError }) {
 
   useEffect(() => {
     Dashboard.get({ id: dashboardId, slug: dashboardSlug })
-      .then(dashboardData => {
+      .then((dashboardData) => {
         recordEvent("view", "dashboard", dashboardData.id);
         setDashboard(dashboardData);
 
@@ -207,7 +208,7 @@ routes.register(
   "Dashboards.LegacyViewOrEdit",
   routeWithUserSession({
     path: "/dashboard/:dashboardSlug",
-    render: pageProps => <DashboardPage {...pageProps} />,
+    render: (pageProps) => <DashboardPage {...pageProps} />,
   })
 );
 
@@ -215,6 +216,6 @@ routes.register(
   "Dashboards.ViewOrEdit",
   routeWithUserSession({
     path: "/dashboards/:dashboardId([^-]+)(-.*)?",
-    render: pageProps => <DashboardPage {...pageProps} />,
+    render: (pageProps) => <DashboardPage {...pageProps} />,
   })
 );
--- a/client/app/pages/dashboards/components/DashboardHeader.jsx
+++ b/client/app/pages/dashboards/components/DashboardHeader.jsx
@@ -22,7 +22,7 @@ import { DashboardStatusEnum } from "../hooks/useDashboard";
 import "./DashboardHeader.less";
 
 function getDashboardTags() {
-  return getTags("api/dashboards/tags").then(tags => map(tags, t => t.name));
+  return getTags("api/dashboards/tags").then((tags) => map(tags, (t) => t.name));
 }
 
 function buttonType(value) {
@@ -38,7 +38,7 @@ function DashboardPageTitle({ dashboardConfiguration }) {
         <h3>
           <EditInPlace
             isEditable={editingLayout}
-            onDone={name => updateDashboard({ name })}
+            onDone={(name) => updateDashboard({ name })}
             value={dashboard.name}
             ignoreBlanks
           />
@@ -53,7 +53,7 @@ function DashboardPageTitle({ dashboardConfiguration }) {
         isArchived={dashboard.is_archived}
         canEdit={canEditDashboard}
         getAvailableTags={getDashboardTags}
-        onEdit={tags => updateDashboard({ tags })}
+        onEdit={(tags) => updateDashboard({ tags })}
       />
     </div>
   );
@@ -89,14 +89,15 @@ function RefreshButton({ dashboardConfiguration }) {
         placement="bottomRight"
         overlay={
           <Menu onClick={onRefreshRateSelected} selectedKeys={[`${refreshRate}`]}>
-            {refreshRateOptions.map(option => (
+            {refreshRateOptions.map((option) => (
               <Menu.Item key={`${option}`} disabled={!includes(allowedIntervals, option)}>
                 {durationHumanize(option)}
               </Menu.Item>
             ))}
             {refreshRate && <Menu.Item key={null}>Disable auto refresh</Menu.Item>}
           </Menu>
-        }>
+        }
+      >
         <Button className="icon-button hidden-xs" type={buttonType(refreshRate)}>
           <i className="fa fa-angle-down" aria-hidden="true" />
           <span className="sr-only">Split button!</span>
@@ -166,7 +167,8 @@ function DashboardMoreOptionsButton({ dashboardConfiguration }) {
             <PlainButton onClick={archive}>Archive</PlainButton>
           </Menu.Item>
         </Menu>
-      }>
+      }
+    >
       <Button className="icon-button m-l-5" data-test="DashboardMoreButton" aria-label="More actions">
         <EllipsisOutlinedIcon rotate={90} aria-hidden="true" />
       </Button>
@@ -216,7 +218,8 @@ function DashboardControl({ dashboardConfiguration, headerExtra }) {
                 type={buttonType(fullscreen)}
                 className="icon-button m-l-5"
                 onClick={toggleFullscreen}
-                aria-label="Toggle fullscreen display">
+                aria-label="Toggle fullscreen display"
+              >
                 <i className="zmdi zmdi-fullscreen" aria-hidden="true" />
               </Button>
             </Tooltip>
@@ -229,7 +232,8 @@ function DashboardControl({ dashboardConfiguration, headerExtra }) {
                 type={buttonType(dashboard.publicAccessEnabled)}
                 onClick={showShareDashboardDialog}
                 data-test="OpenShareForm"
-                aria-label="Share">
+                aria-label="Share"
+              >
                 <i className="zmdi zmdi-share" aria-hidden="true" />
               </Button>
             </Tooltip>
--- a/client/app/pages/dashboards/hooks/useDashboard.js
+++ b/client/app/pages/dashboards/hooks/useDashboard.js
@@ -22,12 +22,12 @@ export { DashboardStatusEnum } from "./useEditModeHandler";
 
 function getAffectedWidgets(widgets, updatedParameters = []) {
   return !isEmpty(updatedParameters)
-    ? widgets.filter(widget =>
+    ? widgets.filter((widget) =>
         Object.values(widget.getParameterMappings())
           .filter(({ type }) => type === "dashboard-level")
           .some(({ mapTo }) =>
             includes(
-              updatedParameters.map(p => p.name),
+              updatedParameters.map((p) => p.name),
               mapTo
             )
           )
@@ -50,7 +50,7 @@ function useDashboard(dashboardData) {
     [dashboard]
   );
   const hasOnlySafeQueries = useMemo(
-    () => every(dashboard.widgets, w => (w.getQuery() ? w.getQuery().is_safe : true)),
+    () => every(dashboard.widgets, (w) => (w.getQuery() ? w.getQuery().is_safe : true)),
     [dashboard]
   );
 
@@ -67,19 +67,19 @@ function useDashboard(dashboardData) {
 
   const updateDashboard = useCallback(
     (data, includeVersion = true) => {
-      setDashboard(currentDashboard => extend({}, currentDashboard, data));
+      setDashboard((currentDashboard) => extend({}, currentDashboard, data));
       data = { ...data, id: dashboard.id };
       if (includeVersion) {
         data = { ...data, version: dashboard.version };
       }
       return Dashboard.save(data)
-        .then(updatedDashboard => {
-          setDashboard(currentDashboard => extend({}, currentDashboard, pick(updatedDashboard, keys(data))));
+        .then((updatedDashboard) => {
+          setDashboard((currentDashboard) => extend({}, currentDashboard, pick(updatedDashboard, keys(data))));
           if (has(data, "name")) {
             location.setPath(url.parse(updatedDashboard.url).pathname, true);
           }
         })
-        .catch(error => {
+        .catch((error) => {
           const status = get(error, "response.status");
           if (status === 403) {
             notification.error("Dashboard update failed", "Permission Denied.");
@@ -102,25 +102,25 @@ function useDashboard(dashboardData) {
 
   const loadWidget = useCallback((widget, forceRefresh = false) => {
     widget.getParametersDefs(); // Force widget to read parameters values from URL
-    setDashboard(currentDashboard => extend({}, currentDashboard));
+    setDashboard((currentDashboard) => extend({}, currentDashboard));
     return widget
       .load(forceRefresh)
-      .catch(error => {
+      .catch((error) => {
         // QueryResultErrors are expected
         if (error instanceof QueryResultError) {
           return;
         }
         return Promise.reject(error);
       })
-      .finally(() => setDashboard(currentDashboard => extend({}, currentDashboard)));
+      .finally(() => setDashboard((currentDashboard) => extend({}, currentDashboard)));
   }, []);
 
-  const refreshWidget = useCallback(widget => loadWidget(widget, true), [loadWidget]);
+  const refreshWidget = useCallback((widget) => loadWidget(widget, true), [loadWidget]);
 
-  const removeWidget = useCallback(widgetId => {
-    setDashboard(currentDashboard =>
+  const removeWidget = useCallback((widgetId) => {
+    setDashboard((currentDashboard) =>
       extend({}, currentDashboard, {
-        widgets: currentDashboard.widgets.filter(widget => widget.id !== undefined && widget.id !== widgetId),
+        widgets: currentDashboard.widgets.filter((widget) => widget.id !== undefined && widget.id !== widgetId),
       })
     );
   }, []);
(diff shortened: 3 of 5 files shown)


vladenisov:
I skipped the Restyled check to give the reviewer a chance to look at the code that was changed.

I would recommend running `yarn prettier` separately to change all files in the repository.

yoshiokatsuneo:
@denisov-vlad 

Thank you for your PR !
I also see this issue, and feel annoying. It is cool that you are trying to fix the issue !

I think the parameters should be saved only when "Done Editing" button is clicked, as the "Done Editing" button is for saving, and it makes the behavior of the Query Editor and the Dashboard Editor consistent.

yoshiokatsuneo:
@denisov-vlad 

Thank you for updating.
And again, I think the parameters should be saved only when "Done Editing" button(Not "Apply Changes" button) is clicked, as the "Done Editing" button is for saving, and it makes the behavior of the Query Editor and the Dashboard Editor consistent.

vladenisov:
> @denisov-vlad
> 
> Thank you for updating. And again, I think the parameters should be saved only when "Done Editing" button(Not "Apply Changes" button) is clicked, as the "Done Editing" button is for saving, and it makes the behavior of the Query Editor and the Dashboard Editor consistent.

Hi @yoshiokatsuneo 

Have you tested it? It saves changes only after "Done editing" button click. I had tried to find any other option to save params, but the only way I found is with that button.

Also added some info to the PR message.

yoshiokatsuneo:
@denisov-vlad 

Thank you !
I'm sorry I was wrong. The PR is actually working as described.

And, I think `refreshDashboard()` function should just refresh dashboard as named. How about ?

vladenisov:
@yoshiokatsuneo The logic is consistent with the queries. The parameters are updated in edit mode, but after clicking Apply Changes and then Done Editing.

I have now simplified the code a little and removed it from `refreshDashboard`.

yoshiokatsuneo:
@denisov-vlad 

Thank you for updating. The behavior looks good !
And for about the code, I think it is a bit unclear from the code when the parameters are saved.

How about making it easier to understand ?
For example, how about saving the parameter on onclick event handler of "Done Editing" button so that people can see the flow ?

vladenisov:
@yoshiokatsuneo I hope I understood you correctly. I'm not a developer and haven't explored the front-end code in depth, but I hope everything is okay now.

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -41,6 +41,16 @@ function useDashboard(dashboardData) {
   const [refreshing, setRefreshing] = useState(false);
   const [gridDisabled, setGridDisabled] = useState(false);
   const globalParameters = useMemo(() => dashboard.getParametersDefs(), [dashboard]);
+  const collectCurrentParameterValues = useCallback(() => {
```
yoshiokatsuneo: May we just use `globalParameters` instead of `collectCurrentParameterValues` ?

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,16 +154,60 @@ function useDashboard(dashboardData) {
     [loadWidget]
   );
 
+  const persistParameterValues = useCallback(
+    parameterValues => {
```
yoshiokatsuneo: Can't we just save parameterValues ?

client/app/services/dashboard.js
```diff
@@ -210,9 +210,13 @@ Dashboard.prototype.getParametersDefs = function getParametersDefs() {
...
     }
   });
+  const savedParameterValues = _.get(this, "options.parameterValues", {});
   const resultingGlobalParams = _.values(
     _.each(globalParams, param => {
-      param.setValue(param.value); // apply global param value to all locals
+      const valueToApply = _.has(savedParameterValues, param.name)
+        ? savedParameterValues[param.name]
+        : param.value;
+      param.setValue(valueToApply); // apply global param value (saved or default) to all locals
```
yoshiokatsuneo: Can we use Spread Operator to make it simple like below ?

```
resultingGlobalPramas = {
   ...globalParams,
   ...savedParameterValues,
   ...queryParams,
}
```

client/app/pages/dashboards/components/DashboardHeader.jsx
```diff
@@ -252,7 +252,13 @@ function DashboardEditControl({ dashboardConfiguration, headerExtra }) {
     doneBtnClickedWhileSaving,
     dashboardStatus,
     retrySaveDashboardLayout,
+    saveDashboardParameters = () => Promise.resolve(),
   } = dashboardConfiguration;
+  const handleDoneEditing = () => {
+    Promise.resolve(saveDashboardParameters())
```
yoshiokatsuneo: I think we can remove Promise.resolve() like:
```
   saveDashboardParameters()
     .then(()=>setEditingLayout(false));
```

client/app/pages/dashboards/hooks/useDashboard.js
vladenisov: I'm not sure that it's possible right now. We must update `options` and there could be globalParamOrder key. With my approach we will save all the keys in this field.

client/app/services/dashboard.js
vladenisov: Improved it a little bit, but your approach doesn't work, because `globalParams` is a `{ name: ParameterInstance }` map, not default `{ name: value }`

vladenisov:
> Thank you for the update ! I added comments.

@yoshiokatsuneo I´ve pushed some changes again. Could you check it please?

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,16 +144,62 @@ function useDashboard(dashboardData) {
...
+      if (!canEditDashboard || !parameterValues || isEmpty(parameterValues)) {
+        return Promise.resolve();
+      }
+
+      const currentDashboard = dashboardRef.current;
+      const currentValues = get(currentDashboard, "options.parameterValues", {});
+      const nextValues = extend({}, currentValues);
+      let hasChanges = false;
+
+      Object.entries(parameterValues).forEach(([name, value]) => {
```
yoshiokatsuneo: May we just save all the parameter without checking equality ?

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,16 +144,62 @@ function useDashboard(dashboardData) {
...
+      setRefreshing(true);
+      loadDashboard(true, updatedParameters).finally(() => setRefreshing(false));
     },
     [refreshing, loadDashboard]
   );
 
+  const saveDashboardParameters = useCallback(() => {
+    const latestValues = (globalParameters || []).reduce((acc, param) => {
+      if (!param) return acc;
+      acc[param.name] = param.value === undefined ? null : param.value;
```
yoshiokatsuneo: May we remove to test undefined if there is no possibility to be undefined ?

vladenisov: Done

client/app/pages/dashboards/hooks/useDashboard.js
vladenisov: Done

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,16 +144,54 @@ function useDashboard(dashboardData) {
...
+    },
+    [canEditDashboard, updateDashboard]
+  );
+
   const refreshDashboard = useCallback(
     updatedParameters => {
-      if (!refreshing) {
-        setRefreshing(true);
-        loadDashboard(true, updatedParameters).finally(() => setRefreshing(false));
+      if (refreshing) {
```
yoshiokatsuneo: I think we no longer need this change in this PR.

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,16 +144,54 @@ function useDashboard(dashboardData) {
...
   );
 
+  const persistParameterValues = useCallback(
+    parameterValues => {
+      if (!canEditDashboard || !parameterValues || isEmpty(parameterValues)) {
+        return Promise.resolve();
+      }
+
+      const currentDashboard = dashboardRef.current;
+      const currentValues = get(currentDashboard, "options.parameterValues", {});
```
yoshiokatsuneo: How about removing lodash by like `const currentValues = currentDashboard.options.parameterValues` ?

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,16 +144,54 @@ function useDashboard(dashboardData) {
...
+      const nextValues = extend({}, currentValues);
```
yoshiokatsuneo: I think we can remove extend.

client/app/services/dashboard.js
```diff
@@ -210,10 +210,15 @@ Dashboard.prototype.getParametersDefs = function getParametersDefs() {
         });
     }
   });
+  const mergedValues = {
+    ..._.mapValues(globalParams, p => p.value),
+    ..._.get(this, "options.parameterValues", {}),
```
yoshiokatsuneo: How about removing lodash by like `...(this.options.parameterValues || {})` ?

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,16 +144,54 @@ function useDashboard(dashboardData) {
...
+
+      Object.entries(parameterValues).forEach(([name, value]) => {
+        nextValues[name] = value;
+      });
+
+      return updateDashboard({
+        options: extend({}, get(currentDashboard, "options", {}), { parameterValues: nextValues }),
+      });
```
yoshiokatsuneo: How about removing extend and get like 
```
return updateDashboard({
  options: {
    ...currentDashboard.options,
    parameterValues: nextValues,
  },
})
```

client/app/pages/dashboards/DashboardPage.jsx
```diff
@@ -92,7 +92,7 @@ function DashboardComponent(props) {
   const [bottomPanelStyles, setBottomPanelStyles] = useState({});
   const onParametersEdit = parameters => {
     const paramOrder = map(parameters, "name");
-    updateDashboard({ options: { globalParamOrder: paramOrder } });
+    updateDashboard({ options: { ...(dashboard.options || {}), globalParamOrder: paramOrder } });
```
yoshiokatsuneo: I think we can remove `|| {}` like:
```
updateDashboard({ options: { ...dashboard.options, globalParamOrder: paramOrder } });
```

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,16 +144,54 @@ function useDashboard(dashboardData) {
...
+
+      setRefreshing(true);
+      loadDashboard(true, updatedParameters).finally(() => setRefreshing(false));
     },
     [refreshing, loadDashboard]
   );
 
+  const saveDashboardParameters = useCallback(() => {
+    const latestValues = (globalParameters || []).reduce((acc, param) => {
+      if (!param) return acc;
```
yoshiokatsuneo: I guess we don't need this line ?

client/app/pages/dashboards/hooks/useDashboard.js
vladenisov: It looks better as for me, but agree that it doesn't relate this PR.

client/app/pages/dashboards/hooks/useDashboard.js
vladenisov: Ok

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,6 +144,29 @@ function useDashboard(dashboardData) {
     [loadWidget]
   );
 
+  const persistParameterValues = useCallback(
+    parameterValues => {
+      if (!canEditDashboard || !parameterValues || isEmpty(parameterValues)) {
```
yoshiokatsuneo: I think we can remove this if sentence.

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -144,6 +144,29 @@ function useDashboard(dashboardData) {
     [loadWidget]
   );
 
+  const persistParameterValues = useCallback(
```
yoshiokatsuneo: Now, I think we can move the implementation of persistParameterValues to saveDashboardParameters , as both function is doing one thing and each are short enough.

client/app/pages/dashboards/components/DashboardHeader.jsx
```diff
@@ -252,7 +252,11 @@ function DashboardEditControl({ dashboardConfiguration, headerExtra }) {
     doneBtnClickedWhileSaving,
     dashboardStatus,
     retrySaveDashboardLayout,
+    saveDashboardParameters = () => Promise.resolve(),
```
yoshiokatsuneo: I think we can remove ` = () => Promise.resolve()`.

vladenisov:
@yoshiokatsuneo is it ok now?

yoshiokatsuneo:
@denisov-vlad Sorry for late, but I' ll check it out.

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -154,6 +154,32 @@ function useDashboard(dashboardData) {
     [refreshing, loadDashboard]
   );
 
+  const saveDashboardParameters = useCallback(() => {
+    const latestValues = (globalParameters || []).reduce((acc, param) => {
```
yoshiokatsuneo: I think using "reduce" is a bit tricky.
How about simply using forEach like below ?

```
parameterValues = {};
globalParameters.forEach((param) => {
   latestValues[param.name] = param.value;
})
```

Then, I think we can remove following lines.
```
    if (isEmpty(latestValues)) return Promise.resolve();
```

```
    const resultValues = currentDashboard.options.parameterValues || {};

    Object.entries(latestValues).forEach(([name, value]) => {
      resultValues[name] = value;
    });
```

client/app/services/dashboard.js
```diff
@@ -210,10 +210,15 @@ Dashboard.prototype.getParametersDefs = function getParametersDefs() {
         });
     }
   });
+  const mergedValues = {
+    ..._.mapValues(globalParams, p => p.value),
+    ...(this.options.parameterValues || {}),
+    ...queryParams,
```
yoshiokatsuneo: This is what I suggested, but I noticed that we don't need `...queryParams,`, I think.

client/app/pages/dashboards/hooks/useDashboard.js
vladenisov: Ok. It's even better to rewrite all params to not store unused ones. Fixed it.

client/app/services/dashboard.js
vladenisov: Had some doubts about it, but tested without queryParams, and everythin works fine.

client/app/pages/dashboards/hooks/useDashboard.js
```diff
@@ -154,6 +154,26 @@ function useDashboard(dashboardData) {
...
+    globalParameters.forEach(param => {
+      parameterValues[param.name] = param.value;
+    });
+
+    const currentDashboard = dashboardRef.current;
+
+    return updateDashboard({
+      options: {
+        ...currentDashboard.options,
+        parameterValues: parameterValues,
```
yoshiokatsuneo: I think it is nice to interface compatible with `queries`.
So, how about using `parameters` array as `queries.options.parameters` like:
```
options: {
    ...currentDashboard.options,
    parameters: map(data.options.parameters, p => globalParameters.toSaveableObject()),
}
```

client/app/services/dashboard.js
```diff
@@ -210,10 +210,14 @@ Dashboard.prototype.getParametersDefs = function getParametersDefs() {
         });
     }
   });
+  const mergedValues = {
+    ..._.mapValues(globalParams, p => p.value),
+    ...(this.options.parameterValues || {}),
```
yoshiokatsuneo: And, following the `parameter` interface, I think this line can be below.


```
...Object.fromEntries((this.options.parameters || {}).map( param => [param.name, param.value]))
```

vladenisov: ` || []` instead of ` || {}`, but done

client/app/pages/dashboards/hooks/useDashboard.js
vladenisov: agree with parameters key and list instead of dict, but used `globalParameters` instead of `data.options.parameters`

vladenisov:
Hi @yoshiokatsuneo 

Is everything okay now? Maybe we can do a merge? I have a big project that's stalled because I'm waiting for these changes.

yoshiokatsuneo:
@denisov-vlad 
Sorry for late. I was a bit busy.

I noticed that the saved parameter have extra "locals" array field in the options.parameters.
For example, how about removing following lines from `getParametersDefs` function not to save the extra "locals" field ?

```
            // add to locals list
            globalParams[mapping.mapTo].locals.push(param);
```

vladenisov:
> @denisov-vlad Sorry for late. I was a bit busy.
> 
> I noticed that the saved parameter have extra "locals" array field in the options.parameters. For example, how about removing following lines from `getParametersDefs` function not to save the extra "locals" field ?
> 
> ```
>             // add to locals list
>             globalParams[mapping.mapTo].locals.push(param);
> ```

Don't really know how it works, but removing this line disables query update after changing param, so it will be updated only after  page refresh, so I prefer to add `locals` for now.

yoshiokatsuneo:
@denisov-vlad 

Ah, sorry.
How about adding "locals" to the omit list at the `toSaveableObject` function in the `Parameter` class like below not to save needless "locals" field ?
(client/app/services/parameters/Parameter.js)

```
  /** Get a saveable version of the Parameter by omitting unnecessary props */
  toSaveableObject() {
    return omit(this, ["$$value", "urlPrefix", "pendingValue", "parentQueryId", "locals"]);
  }
```

vladenisov:
> @denisov-vlad
> 
> Ah, sorry. How about adding "locals" to the omit list at the `toSaveableObject` function in the `Parameter` class like below not to save needless "locals" field ? (client/app/services/parameters/Parameter.js)
> 
> ```
>   /** Get a saveable version of the Parameter by omitting unnecessary props */
>   toSaveableObject() {
>     return omit(this, ["$$value", "urlPrefix", "pendingValue", "parentQueryId", "locals"]);
>   }
> ```

Yep, that was my first solution, but it also affects queries params, so I decided to not edit it.

yoshiokatsuneo:
> > @denisov-vlad
> > Ah, sorry. How about adding "locals" to the omit list at the `toSaveableObject` function in the `Parameter` class like below not to save needless "locals" field ? (client/app/services/parameters/Parameter.js)
> > ```
> >   /** Get a saveable version of the Parameter by omitting unnecessary props */
> >   toSaveableObject() {
> >     return omit(this, ["$$value", "urlPrefix", "pendingValue", "parentQueryId", "locals"]);
> >   }
> > ```
> 
> Yep, that was my first solution, but it also affects queries params, so I decided to not edit it.

@denisov-vlad

Thank you for your reply.
And, I think it is OK to remove "locals", as the "locals" field is only for the dashboards, and queries does not use the "locals" field.

vladenisov:
@yoshiokatsuneo Done. Tested with omitting `locals` and everything works fine.

yoshiokatsuneo:
@denisov-vlad Thank you!

yoshiokatsuneo:
@eradman I think this PR is good enough, and makes Redash better by enabling to save dashboard parameter.

But, I just want to hear your thought to make sure.
How do you feel about the PR?

Thank you.

eradman:
This is a nice feature, and seems to work well. Also switching the parameter type Text to Dropdown List did not break anything

```
postgres=# select jsonb_pretty(options) from dashboards;
-[ RECORD 1 ]+-------------------------------------------
jsonb_pretty | {                                         +
             |     "parameters": [                       +
             |         {                                 +
             |             "name": "type",               +
             |             "type": "enum",               +
             |             "title": "type",              +
             |             "value": "Mac OS",            +
             |             "global": false,              +
             |             "enumOptions": "Linux\nMac OS"+
             |         }                                 +
             |     ]                                     +
             | }

postgres=# select jsonb_pretty(options) from queries;
-[ RECORD 1 ]+-------------------------------------------
jsonb_pretty | {                                         +
             |     "parameters": [                       +
             |         {                                 +
             |             "name": "type",               +
             |             "type": "enum",               +
             |             "title": "type",              +
             |             "value": "Linux",             +
             |             "global": false,              +
             |             "enumOptions": "Linux\nMac OS"+
             |         }                                 +
             |     ],                                    +
             |     "apply_auto_limit": true              +
             | }
```

eradman:
Applied the prettier changes, if it passes we will rebase

yoshiokatsuneo:
@denisov-vlad Thank you for your contribution !

yoshiokatsuneo:
@eradman Thank you for the review and testing !

vladenisov:
Thanks!

Issue #7578, opened 2025-11-21
Old Postgres driver - can't connect PostgreSQL DB hosted on neon.tech


olifink:
### Issue Summary

Can't connect a Postgres DB hosted on neon.tech because it appears redash uses an older driver that doesn't support Endpoint IDs. Also, there is no way to enter the parameters in the connection string for a workaround

```
Connection Test Failed:
ERROR: Endpoint ID is not specified. Either please upgrade the postgres client library (libpq) for SNI support or pass the endpoint ID (first part of the domain name) as a parameter: '?options=endpoint%3D<endpoint-id>'. See more at https://neon.tech/sni

```

### Steps to Reproduce

1. Setup the database with the following connection string

`DATABASE_URL='postgresql://neondb_owner:XXXXXXXXXX@ep-shiny-wildflower-ag83nxur-pooler.c-2.eu-central-1.aws.neon.tech/neondb?sslmode=require&channel_binding=require'`

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* Redash Version: 8.0.0+b32245
* Browser/OS: Chrome/Fedora 43
* How did you install Redash: your image on GCP following https://redash.io/help/open-source/setup/

<img width="1713" height="1238" alt="Image" src="https://github.com/user-attachments/assets/2a50f0a8-a42a-43d3-b0dd-0d8d2896e31c" />

yoshiokatsuneo:
I'm not sure the cause.

But, I think you are using older Redash.
I can suggest to upgrade to the latest Redash, with the latest docker compose file.

eradman:
The problem here appears to be that we do not have a means of specifying connection parameters. 
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

I will look into a solution for this.

eradman:
The new parameter support is available in the the [25.12.0-dev image](https://hub.docker.com/r/redash/preview/tags)

PR #7579, opened 2025-11-21
PostgreSQL: allow connection parameters to be specified

eradman:
## What type of PR is this? 

- [x] Feature

## Description

Allow extra PostgreSQL connection parameters to be specified, as documented in
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

Multiple parameters are separated by a space.

## How is this tested?

- [x] Manually

#7578

pg-params

### DIFF
--- a/redash/query_runner/pg.py
+++ b/redash/query_runner/pg.py
@@ -138,6 +138,15 @@ def _get_ssl_config(configuration):
     return ssl_config
 
 
+def _parse_dsn(configuration):
+    standard_params = {"user", "password", "host", "port", "dbname"}
+    params = psycopg2.extensions.parse_dsn(configuration.get("dsn", ""))
+    overlap = standard_params.intersection(params.keys())
+    if overlap:
+        raise ValueError("Extra parameters may not contain {}".format(overlap))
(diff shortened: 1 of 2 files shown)


redash/query_runner/pg.py
```diff
@@ -138,6 +138,15 @@ def _get_ssl_config(configuration):
     return ssl_config
 
 
+def _libpq_params(configuration):
+    standard_params = {"user", "password", "host", "port", "dbname"}
+    params = psycopg2.extensions.parse_dsn(configuration.get("libpq_params", ""))
```
yoshiokatsuneo: I just feel it is more natural to set configuration name to like "connection_string" (or "dsn"? ) instead of "libpq_params", as that is the string set in the configuration and passed to the parse_dsn function. How do you feel ?

https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.parse_dsn

eradman: Adjusted to use the name "dsn"--this name probably is the best because it suggests the format this string should use.

Issue #7468, opened 2025-06-27
Support utf-8 for webhook


jingyugao:
### Issue Summary
When webhook message contains utf-8 chars, the requests.post will cause a traceback.
### Steps to Reproduce
1. Send an webhook alert contains utf-8  chars.
2. This line code will cause a traceback.
https://github.com/getredash/redash/blob/6b1e91012654cda91a4fb64075197b8b3be2cd44/redash/destinations/webhook.py#L45
### Technical details:

* Redash Version:latest
* Browser/OS: linux
* How did you install Redash: docker image

example code for test

```
import requests

try:
    resp = requests.post(
        "https://example.com",
        data="你好",
        headers={},
        timeout=5.0,
    )
except Exception as e:
    print(e)

resp = requests.post(
    "https://example.com",
    data="你好".encode("utf-8"),
    headers={},
    timeout=5.0,
)
print(resp.status_code)
```

eradman:
I believe this was solved in #7586, closing

PR #7586, opened 2025-12-02
fix(destinations): Handle unicode characters in webhook notifications

jingyugao:
Previously, webhook notifications would fail if they contained unicode characters in the alert data. This was because the JSON payload was not UTF-8 encoded before being sent.

This commit fixes the issue by explicitly encoding the JSON data to UTF-8 and adds a test to verify the fix.

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

https://github.com/getredash/redash/issues/7468

### DIFF
--- a/tests/destinations/test_webhook.py
+++ b/tests/destinations/test_webhook.py
@@ -1,0 +1,50 @@
+import json
+from unittest import mock
+
+from redash.destinations.webhook import Webhook
+from redash.models import Alert
+
+
+def test_webhook_notify_handles_unicode():
+    # Create a mock alert with all the properties needed by serialize_alert
+    alert = mock.Mock()
+    alert.id = 1
+    alert.name = "Test Alert"
+    alert.custom_subject = "Test Subject With Unicode: 晨"
+    alert.custom_body = "Test Body"
+    alert.options = {}
+    alert.state = "ok"
+    alert.last_triggered_at = None
+    alert.updated_at = "2025-12-02T08:00:00Z"
+    alert.created_at = "2025-12-02T08:00:00Z"
+    alert.rearm = None
+    alert.query_id = 10
+    alert.user_id = 20
+
+    query = mock.Mock()
+    user = mock.Mock()
+    app = mock.Mock()
+    host = "http://redash.local"
+    options = {"url": "https://example.com/webhook", "username": "user", "password": "password"}
+    metadata = {}
+    new_state = Alert.TRIGGERED_STATE
+    destination = Webhook(options)
+
+    with mock.patch("redash.destinations.webhook.requests.post") as mock_post:
+        mock_response = mock.Mock()
+        mock_response.status_code = 200
+        mock_post.return_value = mock_response
+
+        destination.notify(alert, query, user, new_state, app, host, metadata, options)
+
+        # Get the data passed to the mock
+        call_args, call_kwargs = mock_post.call_args
+        sent_data = call_kwargs["data"]
+
+        # 1. Make sure we send bytes
+        assert isinstance(sent_data, bytes)
+
+        # 2. Make sure the bytes are the correct UTF-8 encoded JSON
+        decoded_data = json.loads(sent_data.decode("utf-8"))
+        assert decoded_data["alert"]["title"] == alert.custom_subject
+        assert "Test Subject With Unicode: 晨" in sent_data.decode("utf-8")
--- a/tests/handlers/test_destinations.py
+++ b/tests/handlers/test_destinations.py
@@ -7,7 +7,6 @@ from redash.destinations.datadog import Datadog
 from redash.destinations.discord import Discord
 from redash.destinations.slack import Slack
 from redash.destinations.webex import Webex
-from redash.destinations.webhook import Webhook
 from redash.models import Alert, NotificationDestination
 from tests import BaseTestCase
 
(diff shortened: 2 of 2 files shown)


yoshiokatsuneo:
@jingyugao 

Thank you for your contribution!
I confirmed that the PR works well.

And, if we create a test, I think the test should be in the file related to the class, like `test/destinations/test_webhook.py`.

Also, I think the test is optional for this quite simple PR.

jingyugao:
> @jingyugao
> 
> Thank you for your contribution! I confirmed that the PR works well.
> 
> And, if we create a test, I think the test should be in the file related to the class, like `test/destinations/test_webhook.py`.
> 
> Also, I think the test is optional for this quite simple PR.

Thankyou for yout comment! `tests/handlers/test_destinations.py` is an already exists file which contain tests about webhook.Shall I create a new file in `test/destinations/` ?

yoshiokatsuneo:
> > @jingyugao
> > Thank you for your contribution! I confirmed that the PR works well.
> > And, if we create a test, I think the test should be in the file related to the class, like `test/destinations/test_webhook.py`.
> > Also, I think the test is optional for this quite simple PR.
> 
> Thankyou for yout comment! `tests/handlers/test_destinations.py` is an already exists file which contain tests about webhook.Shall I create a new file in `test/destinations/` ?

I think so, if we add the test.

yoshiokatsuneo:
@arikfr 

As written in the issue(https://github.com/getredash/redash/issues/7468) PR author created, `request.post` failed when `data` parameter is UTF-8 string.
The error happens around the following code.

https://github.com/python/cpython/blob/main/Lib/http/client.py#L1405
```
        if isinstance(body, str):
            # RFC 2616 Section 3.7.1 says that text default has a
            # default charset of iso-8859-1.
            body = _encode(body, 'body')
```

https://github.com/python/cpython/blob/main/Lib/http/client.py#L166C1-L170C38
```
def _encode(data, name='data'):
    """Call data.encode("latin-1") but show a better error message."""
    try:
        return data.encode("latin-1")
    except UnicodeEncodeError as err:
```

And, following `requests` API reference says data should be "bytes"(not "string"). So, to follow the specification, we need to pass data as "bytes".

https://requests.readthedocs.io/en/latest/api/
```
data – (optional) Dictionary, list of tuples, bytes, or file-like object to send in the body of the [Request](https://requests.readthedocs.io/en/latest/api/#requests.Request)
```

Here is the stack trace when the error happens:
```
redash@9a75d904dc19:/app$ cat a2.py 
import requests

resp = requests.post(
    "https://example.com",
    data="你好",
    headers={},
    timeout=5.0,
)
print(resp.status_code)
redash@9a75d904dc19:/app$ python a2.py 
  File "/app/a2.py", line 3, in <module>
    resp = requests.post(
  File "/usr/local/lib/python3.10/site-packages/requests/api.py", line 115, in post
    return request("post", url, data=data, json=json, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python3.10/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/requests/adapters.py", line 667, in send
    resp = conn.urlopen(
  File "/usr/local/lib/python3.10/site-packages/urllib3/connectionpool.py", line 715, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/lib/python3.10/site-packages/urllib3/connectionpool.py", line 416, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/local/lib/python3.10/site-packages/urllib3/connection.py", line 244, in request
    super(HTTPConnection, self).request(method, url, body=body, headers=headers)
  File "/usr/local/lib/python3.10/http/client.py", line 1283, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/local/lib/python3.10/http/client.py", line 1328, in _send_request
    body = _encode(body, 'body')
  File "/usr/local/lib/python3.10/http/client.py", line 166, in _encode
    raise UnicodeEncodeError(
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-1: Body ('你好') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.
```

arikfr:
@yoshiokatsuneo thank you for the elaborate explanation! 

For a moment I thought that maybe we should use the `json` field of `post()`, but realized that leveraging our own `json_dumps` has benefits, so I'm cool with the solution here.

jingyugao:
> > > @jingyugao
> > > Thank you for your contribution! I confirmed that the PR works well.
> > > And, if we create a test, I think the test should be in the file related to the class, like `test/destinations/test_webhook.py`.
> > > Also, I think the test is optional for this quite simple PR.
> > 
> > 
> > Thankyou for yout comment! `tests/handlers/test_destinations.py` is an already exists file which contain tests about webhook.Shall I create a new file in `test/destinations/` ?
> 
> I think so, if we add the test.

updated

yoshiokatsuneo:
@jingyugao 

I just approved and merged the PR.

PR #7589, opened 2025-12-03
chore: support ipv6 for server in docker

ghost:
## What type of PR is this? 

- [x] Bug Fix

## Description

The listener address is hardcoded to IPv4-only (0.0.0.0) which breaks in an IPv6-only host. gunicorn supports the `[::]` host as a dualstack wildcard, so that's what I'm changing here.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [X] Manually
- [ ] N/A

Built the docker image locally and started it

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -46,7 +46,7 @@ server() {
   MAX_REQUESTS=${MAX_REQUESTS:-1000}
   MAX_REQUESTS_JITTER=${MAX_REQUESTS_JITTER:-100}
   TIMEOUT=${REDASH_GUNICORN_TIMEOUT:-60}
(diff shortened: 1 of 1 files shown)


eradman:
Even without this change IPv6 appears to be reachable

```
$ nc -zv -4 sfdev2.lan 5001
Connection to sfdev2.lan (192.168.2.23) 5001 port [tcp/*] succeeded!
$ nc -zv -6 sfdev2.lan 5001
Connection to sfdev2.lan (fd00:52::17) 5001 port [tcp/*] succeeded!
```

What is the output of `docker ps` for you? (with and without this change)

eradman:
Closing as not needed

ghost:
Sorry I missed the notification, but this is in fact, needed. I am running this in an IPv6-only kubernetes cluster in AWS (EKS), and therefore have no access to `docker ps`.

PR #7592, opened 2025-12-08
Fix ctrl+enter resulting in duplicate triggers on non-Mac platforms.

Jakdaw:
At present, we bind both mod+enter and ctrl+enter to execute the current query. On Mac this results in both ctrl+enter and command+enter being bound, but on other platforms mod becomes ctrl and we bind ctrl+enter TWICE. Every shortcut invocation happens twice, and it's a race condition as to whether one server-side job or two are created. Instead, do our own translation and dedupe before binding.

Even without the race condition triggering it's easy to reproduce the bug:
 - Open Chrome browser on non-Mac platform
 - Start a new query - something that will take a few seconds - SELECT SLEEP(10)
 - Open up Chrome Developer Tools, watch network traffic
 - <Ctrl><Enter> to execute the query
 - Note that it calls the query_results API twice rather than once, and polls the JobID twice every second rather than just once!

## What type of PR is this? 

- [X] Bug Fix

## How is this tested?

Disclaimer: I don't have a Mac nor any JS experience!

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -46,7 +46,7 @@ server() {
   MAX_REQUESTS=${MAX_REQUESTS:-1000}
   MAX_REQUESTS_JITTER=${MAX_REQUESTS_JITTER:-100}
   TIMEOUT=${REDASH_GUNICORN_TIMEOUT:-60}
-  exec /usr/local/bin/gunicorn -b 0.0.0.0:5000 --name redash -w${REDASH_WEB_WORKERS:-4} redash.wsgi:app --max-requests $MAX_REQUESTS --max-requests-jitter $MAX_REQUESTS_JITTER --timeout $TIMEOUT --limit-request-line ${REDASH_GUNICORN_LIMIT_REQUEST_LINE:-0}
+  exec /usr/local/bin/gunicorn -b [::]:5000 --name redash -w${REDASH_WEB_WORKERS:-4} redash.wsgi:app --max-requests $MAX_REQUESTS --max-requests-jitter $MAX_REQUESTS_JITTER --timeout $TIMEOUT --limit-request-line ${REDASH_GUNICORN_LIMIT_REQUEST_LINE:-0}
 }
 
 create_db() {
--- a/client/app/pages/dashboards/DashboardPage.jsx
+++ b/client/app/pages/dashboards/DashboardPage.jsx
@@ -31,7 +31,8 @@ function DashboardSettings({ dashboardConfiguration }) {
       <Checkbox
         checked={!!dashboard.dashboard_filters_enabled}
         onChange={({ target }) => updateDashboard({ dashboard_filters_enabled: target.checked })}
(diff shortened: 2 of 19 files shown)


yoshiokatsuneo:
@Jakdaw Thank you for your contribution!
I'm just wondering that just changing shortcut definition from "mod+enter" to "command+enter" solves the problem.

Jakdaw:
> @Jakdaw Thank you for your contribution! I'm just wondering that just changing shortcut definition from "mod+enter" to "command+enter" solves the problem.

No; I think the first keyboard combo is used to generate the human readable description for tool tips and the like. So if we replace "mod+enter" with "command+enter" the tooltips will be incorrect on non-Mac.

Removing "ctrl+enter" would have fixed it - I'm not sure why Mac users would want <ctrl><enter> in addition to <command><enter> - but that got added 5 years ago, so I'd kinda take the view that that ship has sailed.

yoshiokatsuneo:
@Jakdaw 

> So if we replace "mod+enter" with "command+enter" the tooltips will be incorrect on non-Mac.
I see. You are right !

yoshiokatsuneo:
@Jakdaw 

I tested on My Mac, and it looks "Command-Enter" does not work.
It looks because `mod+enter` is mapped to `cmd+enter` instead of `command+enter` on Mac.

Jakdaw:
> @Jakdaw
> 
> I tested on My Mac, and it looks "Command-Enter" does not work. It looks because `mod+enter` is mapped to `cmd+enter` instead of `command+enter` on Mac.

Very good point - yes, looking at Mousetrap we should be sending it "command+enter" not "cmd+enter". For that matter Apple write the full "command" on their keyboards - so I'm going to change the string that's used both for mousetrap and for the tooltip, seems that "Command" is more appropriate there than "Cmd".

Jakdaw:
Hah - it seems old Mac's *did* have "Cmd" on the keyboard, but since about 2015 they appear to have changed to "command"; so there we go I'm bringing it up to date! Reminded why I normally stay the hell away from UIs ;-) Thanks for your help Tsuneo!

client/app/services/KeyboardShortcuts.js
```diff
@@ -35,11 +35,17 @@ const KeyboardShortcuts = {
...
-      const keys = key
+      // Resolve platform-specific modifiers and remove duplicates.
+      const rawKeys = key
         .toLowerCase()
         .split(",")
         .map(trim);
-      each(keys, k => {
+      // Translate platform-specific modifiers and dedupe.
+      const transformed = rawKeys.map(k =>
+        k.replace(/mod/g, modKey.toLowerCase())
```
yoshiokatsuneo, line 45: I'm just feeling that it might be better not to depends on "modKey", as "modKey" is not for mapping but for human readable name. (For example, the name might be localized in the future.)

How about just writing directly like below ?
```
   k.replace(/mod/g, (/Mac|iPod|iPhone|iPad/.test(navigator.platform) ? "command" : "ctrl") )
```

PR #7593, opened 2025-12-09
execute button shortcut: command+enter instead of mod+enter

yoshiokatsuneo:
## What type of PR is this? 
On Windows, when ctrl-enter shortcut is used to execute the query, API is called twice. This happens because ctrl-enter handler is registered twice on Windows where mod+enter is mapped to ctrl+enter that is already exists.
This PR fixed the issue by using command+enter instead of mod+enter.

Related PR: https://github.com/getredash/redash/pull/7592

- [x] Bug Fix

## How is this tested?

- [x] Manually

I tested on Mac that the query was run using command-enter, ctrl-enter, alt-enter, shift-enter.
(I haven't tested on Windows.)

image

### DIFF
--- a/client/app/pages/queries/QuerySource.jsx
+++ b/client/app/pages/queries/QuerySource.jsx
@@ -297,7 +297,7 @@ function QuerySource(props) {
                       }
                       executeButtonProps={{
                         disabled: !queryFlags.canExecute || isQueryExecuting || areParametersDirty,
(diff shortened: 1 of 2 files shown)


Omastto1:
Hi, I would like to bump this PR, are there any action steps needed?
I am on Windows, do you need a tester for it?

PR #7596, opened 2025-12-12
Support ipv6 for server in ipv6-only kubernetes cluster

ghost:
## What type of PR is this? 

- [x] Bug Fix

## Description

The listener address is hardcoded to IPv4-only (0.0.0.0) which breaks in an IPv6-only host. gunicorn supports the `[::]` host as a dualstack wildcard, so that's what I'm changing here.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [X] Manually
- [ ] N/A

Built the docker image locally and started it

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -46,7 +46,7 @@ server() {
   MAX_REQUESTS=${MAX_REQUESTS:-1000}
   MAX_REQUESTS_JITTER=${MAX_REQUESTS_JITTER:-100}
   TIMEOUT=${REDASH_GUNICORN_TIMEOUT:-60}
-  exec /usr/local/bin/gunicorn -b 0.0.0.0:5000 --name redash -w${REDASH_WEB_WORKERS:-4} redash.wsgi:app --max-requests $MAX_REQUESTS --max-requests-jitter $MAX_REQUESTS_JITTER --timeout $TIMEOUT --limit-request-line ${REDASH_GUNICORN_LIMIT_REQUEST_LINE:-0}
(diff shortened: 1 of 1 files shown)


ghost:
Reopening PR #7589

ghost:
In a dual-stack environment, both `0.0.0.0` and `[::]` work for both v4 and v6, but in a v6-only system, only `[::]` works properly.

eradman:
> In a dual-stack environment, both `0.0.0.0` and `[::]` work for both v4 and v6, but in a v6-only system, only `[::]` works properly.

Do you mean that IPv4 is disabled, or that routing for IPv4 addresses are not assigned?

ghost:
> Do you mean that IPv4 is disabled, or that routing for IPv4 addresses are not assigned?

IPv4 is disabled at the host level

eradman (approved):
This does solve the problem when running single-stack IPv6

```
$ host redash.default.svc.mykube.lan
redash.default.svc.mykube.lan has IPv6 address fd00:f4:2::c
```
Tested on kubernetes using

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: redash
spec:
  clusterIP: None
  ports:
  - port: 5000
    targetPort: 5000
  selector:
    app: redash
  ipFamilyPolicy: SingleStack
  ipFamilies:
  - IPv6
```

Also confirmed that this is still compatible with IPv4.
Thank you @aldo-airtm!

eradman:
@aldo-airtm please try rebasing this change and pushing agin

aldofunes:
> @aldo-airtm please try rebasing this change and pushing agin

@eradman Sorry it took a while, it's now rebased

Issue #6710, opened 2024-01-17
Slack Notification Error: UnicodeEncodeError with Non-Latin Characters


unpo88:
### Issue Summary

I encountered an error when trying to send notifications to Slack from Redash. 
The issue arises when the notification content includes non-Latin characters, specifically Korean characters. It seems like there's an encoding issue leading to a `UnicodeEncodeError`.

### Steps to Reproduce

1. Configure a Slack notification in Redash with content that includes Korean characters. -> ('한글')
2. Trigger the notification.
3. The error occurs, preventing the notification from being sent to Slack.

I expected the notification to be sent successfully to Slack, regardless of the content language.
An error occurs stating that the 'latin-1' codec can't encode Korean characters. The error message suggests using `body.encode('utf-8')` for UTF-8 encoding.

Here's the traceback of the error:
``` 
Traceback (most recent call last):  
File "/app/redash/destinations/slack.py", line 53, in notify    
resp = requests.post(options.get("url"), data=json_dumps(payload), timeout=5.0) 
File "/usr/local/lib/python3.8/site-packages/requests/api.py", line 115, in post    
return request("post", url, data=data, json=json, **kwargs) 
File "/usr/local/lib/python3.8/site-packages/requests/api.py", line 59, in request  
return session.request(method=method, url=url, **kwargs)    
File "/usr/local/lib/python3.8/site-packages/requests/sessions.py", line 589, in request    
resp = self.send(prep, **send_kwargs)   
File "/usr/local/lib/python3.8/site-packages/requests/sessions.py", line 703, in send   
r = adapter.send(request, **kwargs) 
File "/usr/local/lib/python3.8/site-packages/requests/adapters.py", line 486, in send   
resp = conn.urlopen(    
File "/usr/local/lib/python3.8/site-packages/urllib3/connectionpool.py", line 715, in urlopen   
httplib_response = self._make_request(  
File "/usr/local/lib/python3.8/site-packages/urllib3/connectionpool.py", line 416, in _make_request 
conn.request(method, url, **httplib_request_kw) 
File "/usr/local/lib/python3.8/site-packages/urllib3/connection.py", line 244, in request   
super(HTTPConnection, self).request(method, url, body=body, headers=headers)    
File "/usr/local/lib/python3.8/http/client.py", line 1256, in request   
self._send_request(method, url, body, headers, encode_chunked)  
File "/usr/local/lib/python3.8/http/client.py", line 1301, in _send_request 
body = _encode(body, 'body')    
[...]

PR #7597, opened 2025-12-19
Mattermost: Fix UnicodeEncodeError for non-latin characters

karmishin:
## What type of PR is this? 

- [x] Bug Fix

## Description

This PR fixes an error that occurs when Redash tries to send a Mattermost notification that contains non-latin characters.

```
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 116-119: 
Body ('📉📊📈🚀') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.
```

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Rebuilt the Docker image and ran the app locally.

- Similar issue for Slack destination: #6710

### DIFF
--- a/redash/destinations/mattermost.py
+++ b/redash/destinations/mattermost.py
@@ -44,7 +44,7 @@ class Mattermost(BaseDestination):
             payload["channel"] = options.get("channel")
 
         try:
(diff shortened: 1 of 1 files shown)


MorselShogiew:
Молодцом!

PR #7510, opened 2025-08-26
Upgrade to Python 3.11 and fix some existing vulnerabilities

wtfiwtz:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [x] Other - Security fixes

## Description

Upgrade to Python 3.11, which allows some other version upgrades of vulnerable libraries.

## How is this tested?

- [x] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Spinning up in Fargate and testing by hand

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

https://github.com/orchestrated-io/redash/pull/12
https://github.com/orchestrated-io/redash/pull/13
https://github.com/orchestrated-io/redash/pull/14

PR #7600, opened 2026-01-02
Upgrade to Python 3.11, upgrade various dependencies

wlach:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Runs redash on a supported version of Python (3.11) and upgrades various dependencies to fix security vulnerabilities. Continuation of work of #7510 from @wtfiwtz (removing changes that are not strictly related to the upgrade)

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Ran redash locally, verified it still works.


pyproject.toml
```diff
@@ -134,9 +135,10 @@ pyodbc = "5.1.0"
 python-arango = "6.1.0"
 python-rapidjson = "1.20"
 requests-aws-sign = "0.1.5"
-sasl = ">=0.1.3"
+# TODO: https://github.com/cloudera/python-sasl/pull/31 - outstanding PR to be merged
+sasl = { git = "https://github.com/gughy8/python-sasl.git#master" }
```
wlach, line 139: Per discussion this dependency is pretty outdated. See also discussion: https://github.com/getredash/redash/pull/7510#discussion_r2299821759

wlach: Also recompiling this seems to take a really long time-- increasing the build times significantly (from 6 to 11 minutes), not sure if anything can really be done about it.

eradman:
Thanks for picking this up @wlach! We will want to update packages that lead to this warning, which is printed when using `manage.py`

```
/usr/local/lib/python3.11/site-packages/databend_sqlalchemy/entry_points.py:4: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early
as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
```

wlach:
> Thanks for picking this up @wlach! We will want to update packages that lead to this warning, which is printed when using `manage.py`
> 
> ...

Right. Downgrading setuptools may also be an option?

yoshiokatsuneo:
@wlach 

Thank you for upgrading python !
Just a question, but is there any reason not to upgrade to the latest python like 3.14 ?

wlach:
> @wlach
> 
> Thank you for upgrading python ! Just a question, but is there any reason not to upgrade to the latest python like 3.14 ?

Mostly because I am ~certain that upgrading Python to an even later version will likely break more things (besides sasl) -- there's been lots of changes in behaviour and standard library support over the last 3 years or so. This discussion is relevant: https://github.com/getredash/redash/discussions/7601

yoshiokatsuneo:
> > @wlach
> > Thank you for upgrading python ! Just a question, but is there any reason not to upgrade to the latest python like 3.14 ?
> 
> Mostly because I am ~certain that upgrading Python to an even later version will likely break more things (besides sasl) -- there's been lots of changes in behaviour and standard library support over the last 3 years or so. This discussion is relevant: #7601

@wlach 

I see. Thanks !
It looks nice to upgrade step-by-step to make each step more simple !

eradman:
Closing; update to Python 3.13 was implemented in #7636

PR #7615, opened 2026-01-30
fix(mysql): Change default charset to utf8mb4

sunakan:
## What type of PR is this? 

- [x] Other

## Description

Both `utf8` and `utf8mb3` are [deprecated in MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8mb3.html) and will be removed in a future major release. This updates the default charset to `utf8mb4`, which is the recommended charset for all new applications.

## How is this tested?

- [x] Manually

Ran `SHOW VARIABLES LIKE 'char%'` before and after the change (see screenshots below).

### Before

before

### After

after


yoshiokatsuneo:
@sunakan

Great !
I confirmed that I can run query on MySQL using this PR.

PR #7616, opened 2026-01-30
Add charset option to RDS MySQL datasource

sunakan:
## What type of PR is this? 

- [x] Feature

## Description

Add charset configuration option to RDS MySQL datasource with utf8mb4 as default

## How is this tested?

- [x] Manually

### Before

before

### After

after


yoshiokatsuneo:
@sunakan 

Cool, again !
I confirmed that I can run query on RDS MySQL.

Thank you for your contribution, again !

PR #7638, opened 2026-02-24
Add back dynamodb support

mdimas:
## What type of PR is this? 

- [ ] Refactor
- [ X] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description
Add back support for DynamoDB. This uses boto3 and PartiQL query syntax instead of the old unmaintained dynamodb package.DynamoDB support was never removed from the README, so no documentation updates are needed there.

## How is this tested?

- [ X] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [ X] Manually
- [ ] N/A

Built and deployed an image that was successfully able to create a DynamoDB data source and run a DynamoDB query.


yoshiokatsuneo:
Thank you !
Can you resolve the conflict ?

yoshiokatsuneo:
And, can you fix the styling test error by running black?

arikfr (commented):
1. Note that you need to apply black formatting.
2. You need to add a logo in `client/app/assets/images/db-logos`.

redash/query_runner/dynamodb_sql.py
```diff
@@ -0,0 +1,161 @@
...
+    def configuration_schema(cls):
+        return {
+            "type": "object",
+            "properties": {
+                "region": {"type": "string", "title": "AWS Region", "default": "us-east-1"},
+                "aws_access_key": {"type": "string", "title": "AWS Access Key"},
+                "aws_secret_key": {"type": "string", "title": "AWS Secret Key"},
+                "endpoint_url": {"type": "string", "title": "Endpoint URL (optional)"},
+            },
+            "required": ["region", "aws_access_key", "aws_secret_key"],
```
kimsehwan96, line 75: I think `aws_access_key` and `aws_secret_key` should not be set as required. Because If some environment like AWS EKS, the redash pods might use `IRSA` as the authentication source, then the `aws_access_key` and `aws_secret_key` doesn't need like `Athena` query runner. (Athena query runner doesn't set the keys as required)

mdimas:
@arikfr @yoshiokatsuneo I think I've addressed the formatting issues and the upstream conflicts.

PR #7642, opened 2026-02-25
Update mssql_odbc.py schema query to use QUOTENAME

gauiPPP:
QUOTENAME will add [] around the values and this will lead to correctly handling database names, schema names and table names with spaces and other SQL syntax problematic characters.

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

SQL query in MS SQL 2022 server with previous and new query

Other pull about mssql.py about the same thing.

### DIFF
--- a/redash/query_runner/mssql_odbc.py
+++ b/redash/query_runner/mssql_odbc.py
@@ -84,7 +84,7 @@ class SQLServerODBC(BaseSQLQueryRunner):
 
     def _get_tables(self, schema):
         query = """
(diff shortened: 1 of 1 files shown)


yoshiokatsuneo:
Same comment is applied for this PR, too.

https://github.com/getredash/redash/pull/7641#issuecomment-3959635498

PR #7644, opened 2026-02-25
Serialize Trino ROW types as JSON objects with field names

nmosafi:
Trino ROW (struct) columns were serialized as plain arrays, losing all field name information. This converts NamedRowTuple values to dicts using the field name metadata from the trino library, so ROW fields appear as {"field": value} instead of [value].

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

I tested by querying our local Trino instance before/after this change

### DIFF
--- a/migrations/versions/0f740a081d20_inline_tags.py
+++ b/migrations/versions/0f740a081d20_inline_tags.py
@@ -21,7 +21,7 @@ depends_on = None
 
 
 def upgrade():
-    tags_regex = re.compile("^([\w\s]+):|#([\w-]+)", re.I | re.U)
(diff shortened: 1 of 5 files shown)


yoshiokatsuneo (approved):

yoshiokatsuneo:
@nmosafi 

I confirmed that it works well. So, I approved and merged the commit. 
Thank you for your contribution !!

PR #7645, opened 2026-02-25
Fix precision loss for large integers in query results

nmosafi:
## Summary
- Integers exceeding JavaScript's `Number.MAX_SAFE_INTEGER` (2^53 - 1) lose precision when parsed by `JSON.parse()` in the browser. For example, a bigint value of `1771858801316128013` would display as `1771858801316128000`.
- The backend `json_dumps` now converts unsafe integers to strings during serialization, preserving their exact values. The column type remains `"integer"` so the frontend can still treat the column as numeric for display purposes.
- The frontend type-sniffing logic in `query-result.js` now recognises string values that represent large integers in `"integer"` columns and skips re-typing them as `"string"`.
- The number formatter in `value-format.tsx` guards against converting such strings back to JavaScript Numbers via the `numeral` library.

## Test plan
- [x] Backend unit tests pass (`tests/utils/test_json_dumps.py`)
- [x] Verify large integer columns (e.g. Trino `BIGINT`) display exact values in table visualization
- [x] Verify safe integers (within `MAX_SAFE_INTEGER`) are unaffected
- [x] Verify number formatting still works for normal numeric columns
- [ ] Verify NaN/Inf sanitization is unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yoshiokatsuneo:
I see this as a valid issue.
At the same time, I'm not sure whether we should handle the case or not, as it looks minor case and the PR may makes the code a bit less simple.

nmosafi:
Yeah understood, any suggestion for a cleaner solution?

nmosafi:
@yoshiokatsuneo Can I ask - is there any chance to make progress with this? I am happy to change the code as you see fit, but think the functionality is important for anyone looking at any kind of network capture data

yoshiokatsuneo:
@nmosafi 

I'm sorry but I don't see why this is important in the network capture data...

nmosafi:
@yoshiokatsuneo We capture nanosecond timestamps of all network traffic and put these into parquet files which we then query with Trino using hive partioning. Those are encoded as bigint e.g. `1789758060968086796` for `2026-09-18T19:01:00.968086796`. We need to be able to see these numbers in both those formats with no precision loss and that is not possible without both this PR and #7646.

From what I recall it's possible to hack this by converting a string and then prepending a character at the start e.g. `x1789758060968086796` but that is very hacky and doesn't work with `select *` either.

We have our own internal fork of Redash with these two PRs in them which we are running, but would rather upstream these fixes.

Thanks

PR #7646, opened 2026-02-25
Preserve string column values that look like datetimes

nmosafi:
The frontend type-sniffing logic converts datetime-looking strings to moment objects, which only support millisecond precision.  Values with sub-millisecond precision (e.g. nanosecond timestamps like "2026-02-23T15:00:01.308267939") lose their fractional digits.

When the backend explicitly sets the column type to "string", the frontend now respects that and skips the datetime/date conversion, displaying the original value as-is.

## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Tested by running some queries against our Trino instance. I see `2026-02-23T15:00:01.308267939` displayed completely now, whereas before it would display as `23/02/26 15:00`


yoshiokatsuneo:
I see this as a valid issue.
At the same time, I'm not sure whether we should handle the case or not, as it looks minor case and the PR may makes the code a bit less simple.

nmosafi:
@yoshiokatsuneo Can I ask - is there any chance to make progress with this? I am happy to change the code as you see fit, but think the functionality is important for anyone looking at any kind of high precision timestamp data

yoshiokatsuneo:
@nmosafi 

Do you have a real major problem because of the issue?

nmosafi:
@yoshiokatsuneo Yeah I am unable to see properly formatted nanosecond level timestamps. For example, with the following JSON returned from the backend

```json
{
    "query_result": {
        "id": 610609,
        "query_hash": "1acff7ba79bd337df72d8bc2a8f4df71",
        "query": "redacted",
        "data": {
            "columns": [
                {
                    "name": "_col0",
                    "friendly_name": "_col0",
                    "type": null
                }
            ],
            "rows": [
                {
                    "_col0": "2026-09-18T19:01:00.965007220Z"
                },
                {
                    "_col0": "2026-09-18T19:01:00.965007220Z"
                },
                {
                    "_col0": "2026-09-18T19:01:00.968086796Z"
                }
            ]
        },
        "data_source_id": 18,
        "runtime": 0.9761795997619629,
        "retrieved_at": "2026-09-24T18:38:59.311+01:00"
    }
}
```

It looks like this in the redash

image

Issue #7649, opened 2026-02-27
DB upgrade scripts has SyntaxWarning after Python version upgrading


myonlylonely:
I assume this is related to this Python upgrade PR: https://github.com/getredash/redash/pull/7636

### Steps to Reproduce

1. run DB upgrade process:
```bash
docker compose run --rm server manage db upgrade
```
2. It has SyntaxWarning:
```
/app/migrations/versions/0f740a081d20_inline_tags.py:24: SyntaxWarning: invalid escape sequence '\w'
  tags_regex = re.compile("^([\w\s]+):|#([\w-]+)", re.I | re.U)
```

### Technical details:

* Redash Version: 8f76f84975090563949736e1a850f90f312af024
* Browser/OS: Chrome
* How did you install Redash: Docker

myonlylonely:
@yoshiokatsuneo Could you please look into this issue?

yoshiokatsuneo:
@myonlylonely 

Thank you for the report !
I just made a PR to fix the issue.
https://github.com/getredash/redash/pull/7650

myonlylonely:
👍

PR #7650, opened 2026-02-27
Fix regular expression warning

yoshiokatsuneo:
## What type of PR is this? 

- [x] Bug Fix

## Description
Just fix the warning on a migration by adding adding "r" prefix before the regular expression.

https://github.com/getredash/redash/issues/7649

## How is this tested?
- [x] Manually

As far as I see, this is just a warning, and make it regular expression literal fixes, without changing the behavior


yoshiokatsuneo:
@eradman Thank you for your quick review caring similar cases!

PR #7657, opened 2026-03-08
Add annotated query option in trino datasource

kimsehwan96:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

## Description

Add environment variable based flags to activate annotated query and apply its feature when the flag enabled (default enabled)

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

스크린샷 2026-03-08 오후 3 32 02

Likewise annotated athena query, can check query annotation in trino ui at the query text.

### DIFF
--- a/redash/query_runner/trino.py
+++ b/redash/query_runner/trino.py
@@ -1,4 +1,5 @@
 import logging
+import os
 
 from redash.models.users import ApiUser, User
 from redash.query_runner import (
@@ -13,8 +14,10 @@ from redash.query_runner import (
     JobTimeoutException,
     register,
 )
+from redash.settings import parse_boolean
 
 logger = logging.getLogger(__name__)
+ANNOTATE_QUERY = parse_boolean(os.environ.get("TRINO_ANNOTATE_QUERY", "true"))
 
 try:
     import trino
@@ -61,7 +64,7 @@ TRINO_TYPES_MAPPING = {
 
 class Trino(BaseQueryRunner):
     noop_query = "SELECT 1"
-    should_annotate_query = False
+    should_annotate_query = ANNOTATE_QUERY
 
     @classmethod
     def configuration_schema(cls):


redash/query_runner/trino.py
```diff
@@ -61,7 +64,6 @@ def _convert_row_types(value):
 
 class Trino(BaseQueryRunner):
     noop_query = "SELECT 1"
-    should_annotate_query = False
```
arikfr, line 64: Why not apply the boolean here? It's not like it can change in runtime.

kimsehwan96: @arikfr 

In BaseQueryRunner 
https://github.com/getredash/redash/blob/57567c75a008c108f24c3f11663d584429611091/redash/query_runner/__init__.py#L116-L118

Default `should_annotate_query` value is `True` and the Athena query runner also doesn't have `should_annotate_query`. 

So I thought `should_annotate_query = False` statement used for disabling to annotate query not the switch for enable/disable annotating feature.

Should I set boolean in `should_annotate_query` to `True` when the env var `TRINO_ANNOTATE_QUERY` is true?

arikfr: `annotate_query` checks `should_annotate_query` and based on it returns original query or adds annotation.

What do you mean by disabling to anotate query vs. enable/disable annotation feature? Isn't it the same? :)

kimsehwan96: @arikfr You're right, I initially misunderstood the role of should_annotate_query.

I was thinking of it as a "disable annotation" flag rather than the actual switch that annotate_query() already checks internally.

Since BaseQueryRunner.annotate_query() already branches on should_annotate_query, there's no need to override the method.

Thanks for the feedback!

PR #7672, opened 2026-03-24
Fix Trino parameterized type mapping

nmurrell07:
Trino columns with parameterized types like timestamp(0) or decimal(10,2) were not being mapped correctly because TRINO_TYPES_MAPPING only contains base type keys without parameters.

This fix adds a _map_trino_type helper function that:
1. First tries direct lookup in TRINO_TYPES_MAPPING
2. If not found and the type contains parentheses, strips the parameters and looks up just the base type (e.g., 'timestamp' from 'timestamp(0)')
3. Returns None if no mapping is found

This ensures columns with parameterized types are correctly mapped to their Redash equivalents.


yoshiokatsuneo:
@nmurrell07 

Can you fix the lint error by running black ?

PR #7692, opened 2026-04-15
Axios update

nicof38:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

Dependency update

## Description
Bumps [axios](https://github.com/axios/axios) from 0.28.0 to 1.15.0.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

take over [PR-7621](https://github.com/getredash/redash/pull/7621)


arikfr:
Did you happen to check if there are any breaking changes in the new version compared to the one we use?

nicof38:
yes, there are breaking changes listed in the migration guide, and all mainly related to error handling [there ](https://github.com/axios/axios/blob/v1.x/MIGRATION_GUIDE.md#breaking-changes)

that is probably why I do not have any issue with tests.

With Axios 0.x - Some HTTP error codes didn't throw, where with Axios 1.x - All HTTP errors throw consistently

May need to discuss about what strategy may be applied, vs the 3 that are proposed by axios project ?
If needed, and in order to fix the CVE, we may switch to latest 0.x version with fix, which is 0.31 ?

PR #7693, opened 2026-04-15
Fix scheduling of forked queries

wlach:
## What type of PR is this? 

- [x] Bug Fix

## Description

When forking a query, redash would set the schedule to `{}`. This would trigger an exception which would cause the query to be unscheduled forever (even if a valid schedule was added later).

Update the code to be defensive against empty/invalid schedules (including tests) and also update the
fork code to explicitly set the schedule to `null` instead of `{}`.

(this bug is only present in ~recent versions of redash which use a jsonb column for the schedule)

## How is this tested?

- [x] Manually

Brought up an instance of redash, verified that forked queries were scheduled

N/A

N/A

### DIFF
--- a/redash/models/__init__.py
+++ b/redash/models/__init__.py
@@ -593,7 +593,8 @@ class Query(ChangeTrackingMixin, TimestampMixin, BelongsToOrgMixin, db.Model):
     def outdated_queries(cls):
         queries = (
             Query.query.options(joinedload(Query.latest_query_data).load_only("retrieved_at"))
-            .filter(func.jsonb_typeof(Query.schedule) != "null")
+            .filter(func.jsonb_typeof(Query.schedule) == "object")
+            .filter(cast(Query.schedule, db.Text) != "{}")
             .order_by(Query.id)
             .all()
         )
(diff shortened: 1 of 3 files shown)


redash/models/__init__.py
```diff
@@ -607,27 +608,42 @@ def outdated_queries(cls):
...
-                    continue
+                interval = query.schedule.get("interval")
+                schedule_time = query.schedule.get("time")
+                day_of_week = query.schedule.get("day_of_week")
+                schedule_until = query.schedule.get("until")
 
-                if query.schedule["until"]:
-                    schedule_until = pytz.utc.localize(datetime.datetime.strptime(query.schedule["until"], "%Y-%m-%d"))
+                # Parse `until` before interval checks so malformed date values are
+                # treated as faulty schedules and get disabled in the exception path.
```
wlach, line 617: I think we should probably rethink the "disabled" path as it's a huge footgun. AFAICT, there's no clear way to re-enable disabled queries (or even notice them) short of manually inspecting the db or logs.

redash/models/__init__.py
```diff
@@ -808,7 +824,7 @@ def fork(self, user):
         kwargs = {a: getattr(self, a) for a in forked_list}
 
         # Query.create will add default TABLE visualization, so use constructor to create bare copy of query
-        forked_query = Query(name="Copy of (#{}) {}".format(self.id, self.name), user=user, **kwargs)
+        forked_query = Query(name="Copy of (#{}) {}".format(self.id, self.name), user=user, schedule=None, **kwargs)
```
wlach, line 827: Explicitly set the schedule to null.

eradman (changes requested):
Changing the schedule is challenging because it may be `null`, "`null`", `{}`, or an object with different numbers of keys.

We might need to work on making this more uniform before fixing the schedule for forked queries.

wlach:
> Changing the schedule is challenging because it may be `null`, "`null`", `{}`, or an object with different numbers of keys.
> 
> We might need to work on making this more uniform before fixing the schedule for forked queries.

Hmm, I feel pretty agnostic about the semantics of "not scheduled" but I'm quite certain that we shouldn't be unsheduling forked queries for all time. 😄 Would you accept this PR with that part taken out?

eradman:
Before this change a forked query would have a `NULL` schedule

```
postgres=# \pset null 'NULL'
postgres=# select id,schedule from queries;
 id |                               schedule
----+----------------------------------------------------------------------
  1 | {"time": null, "until": null, "interval": 1800, "day_of_week": null}
  2 | NULL
(2 rows)

```

(id=2 is a fork of id=1)

With  this change a text value of `null` is used

```
postgres=# \pset null 'NULL'
postgres=# select id,schedule from queries;
 id |                               schedule
----+----------------------------------------------------------------------
  1 | {"time": null, "until": null, "interval": 1800, "day_of_week": null}
  2 | null
(2 rows)
```

We really need to reduce the number of ways to represent _no schedule_

```
NULL
null
{}
```

PR #7711, opened 2026-05-14
Upgrade authlib to 1.7.x and fix critical vulnerabilities

wtfiwtz:
## What type of PR is this? 

Based off a different branch (release 26.3-based) I've upgraded most dependencies in the Docker container that have Critical vulnerabilities (detected by Docker Scout)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Query Runner (Data Source) 
- [ ] New Alert Destination
- [x] Other

## How is this tested?

- [x] Unit tests (pytest, jest)
- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

## Related Tickets & Documents

I'm in the process of testing this related branch - https://github.com/orchestrated-io/redash/commits/debug/v26.3.0p2/

PR #7712, opened 2026-05-19
Upgrade to Flask 3.1.3 and fix other security vulnerabilities

wtfiwtz:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

Security patching and major upgrade of Flask version

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Relates to previous PR - https://github.com/getredash/redash/pull/7711
Merge that PR first!


wtfiwtz:
> **AI-generated comment** — posted on behalf of the PR author to summarise recent stabilisation work.

---

## Test suite stabilisation — summary of changes since last update

The branch has undergone a second round of cleanup after diagnosing a recurring hang/deadlock in the Docker test container. Here is what was found and fixed.

### Root causes identified

**1. Postgres deadlock from `commit = flush` monkey-patch**

An earlier iteration of `BaseTestCase` monkey-patched `db.session.commit = db.session.flush` and called `drop_all()/create_all()` between every test. When a test failed mid-way, the flushed-but-never-committed `INSERT INTO groups` transaction was left `idle in transaction`, holding `AccessExclusiveLock`s. The next test's `drop_all()` then deadlocked on `DROP TABLE queries` waiting for those locks.

Fix: reverted `tests/__init__.py` to the same simple `setUp`/`tearDown` as `master` (`db.session.remove()` + `db.engine.dispose()` + `app_ctx.pop()` + `redis_connection.flushdb()`). Removed the monkey-patch entirely.

**2. `init_db()` left orphaned transactions in TESTING mode**

`redash/models/__init__.py::init_db()` had a TESTING branch that called `db.session.flush()` instead of `db.session.commit()` for the default org/groups. This left the same `idle in transaction` footprint. Fix: reverted `init_db()` to the master form (always commit).

**3. `test_script.py` `MonkeyPatch` was a shared class attribute**

`TestRunScript.monkeypatch = MonkeyPatch()` at class level meant `setattr(subprocess, "check_output", lambda script, shell: "test")` was never undone after each test. Later, `TestJWTAuthentication.setUp()` called `subprocess.check_output(["openssl", ...])` and got that lambda instead, failing with `missing 1 required positional argument: 'shell'`. Fix: scoped `MonkeyPatch()` per-test with `setUp`/`tearDown`.

### Subsequent simplification pass

With those root causes fixed, a large body of workaround code became dead:

| File | What was removed |
|---|---|
| `debug_deadlock.py`, `debug_docker_test.sh` | investigation scripts |
| `redash.iml` | IDE project file |
| `tests/conftest.py` | deadlock-detection wiring (entire file) |
| `tests/test_utils.py` | replaced the deadlock-monitor reimplementation with the original test file from `master` |
| `redash/app.py` | two blocks that imported `tests.test_utils` and called `setup_faulthandler` / `setup_deadlock_signal_handler` during Flask init |
| `redash/handlers/authentication.py` | `flush-if-TESTING-else-commit` pattern (two call sites) |
| `redash/tasks/failure_report.py` | `flush-if-TESTING` in `track_failure`, detachment-reload in `notify_of_failure` |
| `redash/tasks/queries/execution.py` | `rollback-if-not-TESTING` guard |
| `tests/factories.py` | all `flush-if-TESTING-else-commit` TESTING guards throughout |

`redash/security.py` `csp_allows_embeding` was simplified — the inner `decorated` function no longer manually overrides `Content-Security-Policy` and `X-Frame-Options` headers; talisman's per-view override (applied at decoration time) already handles both. Note: the function cannot be fully reverted to `master` because the master line `talisman.content_security_policy + "frame-ancestors *;"` is broken with current flask-talisman (attribute is a dict at decoration time, before `init_app` is called).

### Test collection fixes

Adding `pythonpath = .` to `pytest.ini` (proper pytest 7+ way, replacing the old `sys.path.insert` in the now-deleted `conftest.py`) let `tests/destinations/test_webhook.py` be collected. This exposed two further issues:

- `tests/destinations/` had no `__init__.py`, so pytest imported `test_webhook.py` as a standalone module. Its bare `from redash.models import Alert` import ran SQLAlchemy model initialisation in a context without a Flask app, which caused SQLAlchemy identity-map conflicts in the first `BaseTestCase` suite that followed. Fix: added `tests/destinations/__init__.py`.
- `db.configure_mappers()` had been accidentally dropped from `redash/models/base.py` (it was moved to `__init__.py` during an earlier refactor but then removed when `init_db()` was reverted). Restored to `base.py` where `master` has it.

One TESTING guard was reinstated in `QueryExecutor.__init__` for `db.session.close()`: closing the session in tests detaches factory-held objects, causing `InvalidRequestError ("another instance ... already present")` when the same test issues a second `execute_query` call referencing the same data source. This is a genuine test-isolation constraint (not a workaround), so it is kept.

### Final test result

```
902 passed, 1 skipped in 2:08
```

The 1 skipped is a pre-existing `sqlalchemy-searchable > 1.0 doesn't support searching for emails` skip.

wtfiwtz:
Need to check for broken image links such as `/static/images/db-logos/athena.png`

wtfiwtz:
The images **are** present but with the hashed names, not the original filenames.

I have to run:
```
pnpm run clean
pnpm run build
DOCKER_BUILDKIT=1 docker build --no-cache --build-arg skip_frontend_build= -t redash-server .
```

etc

zachliu (changes requested):
from https://github.com/getredash/redash/blob/master/CONTRIBUTING.md

> If you implement multiple unrelated features, bug fixes, or refactors please split them into individual pull requests.

wtfiwtz:
Replaced (in part) by
- https://github.com/getredash/redash/pull/7718
- https://github.com/getredash/redash/pull/7719
- https://github.com/getredash/redash/pull/7720
- https://github.com/getredash/redash/pull/7721

PR #7750, opened 2026-06-05
feat: enable auto limit for Trino query runner

bonisb:
## What

Make the Trino query runner inherit from `BaseSQLQueryRunner` instead of `BaseQueryRunner` so it supports the "auto limit" feature.

## Why

`BaseQueryRunner` reports `supports_auto_limit = False` and implements `apply_auto_limit` as a no-op. Because `Trino` inherited directly from `BaseQueryRunner`:

- the **Limit 1000** toggle was hidden in the query editor (the frontend reads `supports_auto_limit` from the data source), and
- the `apply_auto_limit` query option had no effect even when explicitly set — the query was passed through unchanged.

`BaseSQLQueryRunner` provides a working implementation that appends `LIMIT 1000` to unbounded `SELECT` statements (leaving statements that already have a `LIMIT`/`OFFSET`, and non-`SELECT` statements, untouched).

## How

- `Trino` now extends `BaseSQLQueryRunner`.
- Trino uses standard ANSI `LIMIT n` syntax, so the default `limit_query = " LIMIT 1000"` works unchanged.
- `Trino` already overrides `get_schema`, so the SQL base class's `_get_tables` schema helpers are not used and schema loading is unaffected.

## Tests

Added tests in `tests/query_runner/test_trino.py` covering:
- `supports_auto_limit` is `True`
- auto limit is appended to a `SELECT` with no limit
- an existing `LIMIT` is preserved
- nothing is changed when auto limit is disabled

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yoshiokatsuneo (approved):

yoshiokatsuneo:
@bonisb 

The PR is simple enough and reasonable.
And, I confirmed that simple queries can run with Trino data source.

So, I can approve the PR.

Please note that the query hashes for existing Trino queries need to be updated.

(Ex. by running `docker compose run --rm server manage query rehash` )

PR #7755, opened 2026-06-15
Chore/migrate poetry to uv

zachliu:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [x] Other

## Description

Migrate Python dependency management from `poetry` to `uv`.

This is a pure tooling swap with **zero** dependency changes. Every resolved package version in `uv.lock` matches the previous `poetry.lock` exactly (all `237` packages), so no application behavior changes and there is no new testing surface from version bumps.

Changes:
- `pyproject.toml`: convert `[tool.poetry.dependencies]` to [PEP 621](https://peps.python.org/pep-0621/) `[project.dependencies]`, and the `all_ds`/`ldap3`/`dev` groups to [PEP 735](https://peps.python.org/pep-0735/) `[dependency-groups]`. Drop the Poetry `[build-system]` and mark the project non-packaged via `[tool.uv] package = false` (equivalent to Poetry's `--no-root`).
- Replace `poetry.lock` with `uv.lock`.
- `Dockerfile`: install a pinned `uv` binary from the official image and use `uv sync --frozen`. The `install_groups` build-arg interface is preserved, so CI and existing build invocations need no changes.
- Update a stale "poetry dependency group" reference in an LDAP error message.

How is this tested?

- `uv lock --locked` confirms the lockfile is consistent with the manifest.
- Docker image builds successfully with the default `install_groups="main,all_ds,dev"` (`236` distributions, versions identical to the previous lockfile).
- Console scripts (`gunicorn`, `supervisord`, `rq`, etc.) resolve on PATH and core imports load.
- The optional `ldap3` group remains excluded by default and installs only when requested via `install_groups`.

Notes

- Future transitive-dependency upgrades are intentionally left out of this PR so the tooling migration can be reviewed in isolation. They can follow as separate, focused PRs.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

<a href="https://cubic.dev/pr/getredash/redash/pull/7755?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lucydodo:
I support this PR as well, but since it involves changing part of the ecosystem, I will delegate the review to someone else.

eradman:
What changes would be made to https://github.com/getredash/redash/wiki/Local-development-setup?

zachliu:
@eradman The [Python dependency section of the wiki](https://github.com/getredash/redash/wiki/Local-development-setup#set-up-python-for-local-backend-development) would change as follows. The `venv` creation, `pip3 install wheel`, and the `make format`/pre-commit steps stay the same.

Remove:
```
(redashvenv1) $ pip3 install poetry==2.1.4
(redashvenv1) $ poetry install --only main,all_ds,dev
```

Replace with:
```
(redashvenv1) $ pip3 install uv==0.11.6
(redashvenv1) $ uv sync --no-default-groups --group all_ds --group dev
```

Notes:
- `main` (the project's base dependencies) is installed implicitly by `uv sync`, so it's no longer named explicitly; `all_ds` and `dev` are passed as `--group` flags.
- `--no-default-groups` mirrors the old Poetry behavior where the `dev` group was optional and not pulled in unless requested (this matches the `default-groups = []` setting in `pyproject.toml`).
- uv can also manage the interpreter itself (`uv python install` + `uv sync` against a project venv), but the `venv` + `pip3 install uv` flow above keeps the wiki's existing structure with minimal edits.

I'm happy to prepare the wiki edit once the approach here looks good to you.

eradman:
After running `uv sync --no-default-groups --group all_ds --group dev`, it appears that `pre-commit` is not installed. Or is there a different way to run it? (This is on Ubuntu 24.04)

```
(redashvenv1) $ make format
pre-commit run --all-files
make: pre-commit: No such file or directory
make: *** [Makefile:44: format] Error 127
```

zachliu:
Good catch :+1:  `pre-commit` is in the `dev` group, but the issue is where `uv` installed it. By default `uv sync` creates and manages its own `.venv` in the project root and ignores an already-activated virtualenv, so with `redashvenv1` active everything went into `./.venv` instead leaving `pre-commit` off your `PATH`.

Adding `--active` tells uv to sync into the activated environment:

```bash
uv sync --active --no-default-groups --group all_ds --group dev
```

I'll make sure the wiki page update uses `--active` so it matches the existing "create a venv, then activate it" flow.

Alternatively, `uv` users can skip the manual `python -m venv` step entirely and let `uv` manage `./.venv` but keeping `--active` preserves the wiki's current structure with the smallest change

eradman:
`source .venv/bin/activate` solved the problem, thanks.

I think using `$PWD/.venv` is a good idea, no need to create a separate venv in `$HOME`. This change seems to work well. We should get a +1 from other maintainers before merging.

zachliu:
@yoshiokatsuneo would you be able to review this PR? @eradman has tested it on `ubuntu 24.04` and confirmed it works well; we just need another pair of :eyes: before merging

this is a pure tooling swap from `poetry` to `uv`. every resolved package version in `uv.lock` matches the previous `poetry.lock` exactly (all `237` packages), so there are no dependency or behavior changes

the diff is limited to `pyproject.toml`, the lockfile, the `Dockerfile`, and one doc-string wording fix

yoshiokatsuneo (approved):
@zachliu 

The changing looks good and simple enough.

I confirmed that I could build and run Redash without problem.
So, I would approve the PR.

Thank you !

zachliu:
@eradman @yoshiokatsuneo thank you guys. i updated the wiki page and ready to merge this one :rocket:

eradman (approved):

PR #7756, opened 2026-06-17
fix(configuration): treat empty secret as explicit clear

ehooi:
## What type of PR is this? 

- [x] Bug Fix

## Description

When a user clears a password field in the UI and saves, the empty string is persisted to the data source options as-is. Downstream SDKs (e.g. boto3) then receive an explicit empty credential instead of falling back to the default credential chain, and authentication fails with empty keys (observed with the Athena data source's aws_secret_key).

Distinguish the placeholder ("--------", meaning "no change") from an empty string ("explicit clear") and drop the latter from the stored options.

## How is this tested?

- [x] Manually

N/A

N/A - backend-only change.

<a href="https://cubic.dev/pr/getredash/redash/pull/7756?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yoshiokatsuneo:
@ehooi 

I tried to create the fix in the front-end site, as more generic solution. (Not only for secret fields but also for all fields.)

https://github.com/getredash/redash/pull/7785

Issue #7684, opened 2026-04-05
Critical: RestrictedPython sandbox escape + command injection via script runner + SQLi x4 + SSRF + CSRF default-off (CWE-94, CWE-78, CWE-89, CWE-918)


Proscan-one:
Hi Redash team,

Found several issues, two of which are critical RCE vectors.

**RestrictedPython sandbox escape (Critical)**

In `redash/query_runner/python.py` at lines 317-319, raw `getattr` and `setattr` are assigned as the `_getattr_`/`_setattr_` guards for RestrictedPython. This completely defeats the sandbox's security model — those guards are specifically what RestrictedPython relies on to prevent attribute access to dangerous objects.

Any user who can run Python queries can escape trivially: `().__class__.__bases__[0].__subclasses__()` walks the class hierarchy to find `os` and call `system()`. Full RCE on the Redash server.

**Command injection via script runner (Critical)**

`redash/query_runner/script.py:8-19` — when the script path is configured as `*`, the user's query text gets passed directly to `subprocess.check_output(script, shell=True)`. Anything in the query box runs as a shell command on the server.

**SQL injection in 4 query runners (High)**

Four query runners interpolate user-controlled values into SQL without parameterization:
- `query_runner/__init__.py:283` — table names via `%` into `select count(*)`
- `query_runner/sqlite.py:45` — `PRAGMA table_info("%s")` with unescaped table name
- `query_runner/duckdb.py:90-96` — extension names via f-string into `INSTALL`/`LOAD`
- `query_runner/memsql_ds.py:76-93` — schema/table names via `%`

**SSRF bypassing advocate protection (High)**

`query_runner/graphite.py:79-84` and several other runners (prometheus, clickhouse, elasticsearch) use raw `requests.get()` instead of the advocate-wrapped session. This bypasses the `ENFORCE_PRIVATE_ADDRESS_BLOCK` setting entirely — internal network access from any authenticated user.

**CSRF disabled by default (High)**

`redash/settings/__init__.py:464` — `ENFORCE_CSRF` defaults to `"false"`. The comment in the code acknowledges the risk but leaves it off. Any state-changing API call is vulnerable to cross-site request forgery out of the box.

The sandbox escape and command injection are the urgent ones — both give full server RCE to any user with query access.

— ProScan AppSec | proscan.one

arikfr:
Firest, there is a process for responsible disclosure documented in our [security policy](https://github.com/getredash/redash/security/policy).

As for the reported issues:

[...]

PR #7757, opened 2026-06-19
Harden Python query runner sandbox

shojiiii:
## What type of PR is this?

- [x] Bug Fix

## Description

Hardens the Python query runner in two areas:

- Replaces raw attribute access and mutation hooks with RestrictedPython safer_getattr, guarded_setattr, and full_write_guard policies. This prevents private attribute traversal and mutation of exposed objects.
- Enforces organization and data source permissions for execute_query, get_source_schema, and get_query_result.
  - execute_query requires non-view-only access and passes the current user to the nested query runner.
  - get_source_schema permits view-only access.
  - get_query_result validates the query organization and access to both the query and returned result data sources.

The RestrictedPython changes target the currently pinned RestrictedPython 8.1.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

Added regression tests for private attribute access, protected writes, organization isolation, inaccessible data sources, view-only behavior, and current-user propagation.

Test result: 23 passed in tests/query_runner/test_python.py. Ruff and Black checks pass.

Related to #7684, specifically the RestrictedPython sandbox escape report. This PR does not close the issue because the issue tracks additional findings.

N/A

<a href="https://cubic.dev/pr/getredash/redash/pull/7757?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/redash/query_runner/python.py
+++ b/redash/query_runner/python.py
@@ -5,9 +5,12 @@ import sys
 
 from RestrictedPython import compile_restricted
 from RestrictedPython.Guards import (
+    full_write_guard,
     guarded_iter_unpack_sequence,
+    guarded_setattr,
     guarded_unpack_sequence,
     safe_builtins,
+    safer_getattr,
 )
 from RestrictedPython.transformer import IOPERATOR_TO_STR
 
@@ -156,7 +159,7 @@ class Python(BaseQueryRunner):
         Custom hooks which controls the way objects/lists/tuples/dicts behave in
         RestrictedPython
         """
-        return obj
+        return full_write_guard(obj)
 
     @staticmethod
     def custom_get_item(obj, key):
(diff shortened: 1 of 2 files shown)


zachliu (changes requested):
Thanks for this PR! The sandbox fix is solid and the API choices are correct for our pinned RestrictedPython 8.1 (`safer_getattr`, `guarded_setattr`, `full_write_guard` all exist there).

One request: could you split this into two PRs? It bundles two distinct fixes:

1. Sandbox escape (`getattr`/`setattr` -> guarded, custom_write -> `full_write_guard`): ~5 lines, critical, low-risk
2. Broken access control (static -> instance methods, org/data-source has_access checks, passing the real user to run_query): larger behavioral change.

These are different CWE classes with different risk profiles. Splitting lets the critical sandbox fix merge fast without waiting on review of the access-control changes, and keeps revert boundaries clean. The tests already separate cleanly along the same line.

Minor note on the second one: `_get_data_source` relies on `self._current_user`; if it's ever `None`, `user.org_id` raises `AttributeError` which gets swallowed into a misleading "Wrong data source name/id" message. Worth confirming that path.

shojiiii:
@zachliu 
Thanks for the review. I split the changes as requested:

  - #7757 is now scoped to the RestrictedPython sandbox hardening only.
  - #7759 contains the Python query helper access-control changes.

  I also addressed the `_current_user is None` note in #7759 by adding an explicit current-user guard, so that path no longer falls through to a misleading data source/query lookup error.

zachliu (approved):
🪨 solid!

PR #7759, opened 2026-06-24
Enforce permissions in Python query helpers

shojiiii:
## What type of PR is this?

- [x] Bug Fix

## Description

Follow-up split from #7757. This PR keeps the Python query runner access-control changes separate from the RestrictedPython sandbox hardening.

It enforces organization and data source permissions for the helper functions exposed to Python query scripts:

- `execute_query` now requires non-view-only access and passes the current user to the nested query runner.
- `get_source_schema` allows view-only access but rejects inaccessible data sources.
- `get_query_result` validates the query organization and access to both the query data source and the returned result data source.
- Direct helper calls without a current user now fail with an explicit error instead of surfacing as a misleading missing data source/query message.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

Test result:

- `docker compose run --rm server tests tests/query_runner/test_python.py` -> 23 passed
- `ruff check redash/query_runner/python.py tests/query_runner/test_python.py` -> passed

Split from #7757. Related to #7684.

N/A

<a href="https://cubic.dev/pr/getredash/redash/pull/7759?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


zachliu:
this looks good. @susodapop what do you think?

redash/query_runner/python.py
```diff
@@ -205,24 +206,33 @@ def add_result_row(result, values):
...
-        """
         try:
+            data_sources = models.DataSource.query.filter(models.DataSource.org_id == user.org_id)
             if isinstance(data_source_name_or_id, int):
-                data_source = models.DataSource.get_by_id(data_source_name_or_id)
+                data_source = data_sources.filter(models.DataSource.id == data_source_name_or_id).one()
             else:
-                data_source = models.DataSource.get_by_name(data_source_name_or_id)
+                data_source = data_sources.filter(models.DataSource.name == data_source_name_or_id).one()
         except models.NoResultFound:
```
zachliu: maybe we should also catch `MultipleResultsFound`? duplicate data-source name within an org would raise `MultipleResultsFound` which turns into an unhandled 500 instead of the friendly `Wrong data source name/id` message

it's a preexisting behavior for the name path, but the PR is touching exactly this code, so it's a reasonable time to broaden the catch

tests/query_runner/test_python.py
```diff
@@ -97,6 +98,105 @@ def test_valid_result_type_with_print(self, mock_dt):
...
+
+    def test_get_query_result_allows_view_only_data_source(self):
+        data_source = self.factory.create_data_source(group=self.factory.default_group, view_only=True)
+        query_result = self.factory.create_query_result(data_source=data_source)
+        query = self.factory.create_query(data_source=data_source, latest_query_data=query_result)
+        self.db.session.flush()
+
+        self.assertEqual(query_result.data, self.python.get_query_result(query.id))
+
+
```
zachliu, line 224:
```suggestion

    def test_get_data_source_rejects_ambiguous_name(self):
        self.factory.create_data_source(name="dup", group=self.factory.default_group)
        self.factory.create_data_source(name="dup", group=self.factory.default_group)
        self.db.session.flush()
    
     with self.assertRaisesRegex(Exception, "Wrong data source name/id"):
        self.python.execute_query("dup", "SELECT 1")
```

none of the tests in `TestPythonQueryRunnerPermissions` exercise the by-name lookup path or a name collision, so if we broaden the catch above to include `MultipleResultsFound` that branch would ship untested. maybe could we add something like this? it's also the only coverage of the by-name branch, which is otherwise untested

redash/query_runner/python.py
shojiiii: Handled in [af2bd99](https://github.com/getredash/redash/pull/7759/commits/af2bd99d5df2c8dfd77eb75f4d643c4018ee06c0). I broadened the catch to include `MultipleResultsFound` so ambiguous by-name lookups now return the same `Wrong data source name/id` error instead of bubbling up as a 500.

tests/query_runner/test_python.py
shojiiii: Added in[af2bd99](https://github.com/getredash/redash/pull/7759/commits/af2bd99d5df2c8dfd77eb75f4d643c4018ee06c0) as `test_execute_query_rejects_ambiguous_data_source_name`. It exercises the by-name lookup path with duplicate names and verifies we return `Wrong data source name/id` for the ambiguous case.

redash/query_runner/python.py
zachliu: thank you kindly :wink:

zachliu (approved):
🚀

PR #7763, opened 2026-06-26
Tolerate inaccessible datasets when loading BigQuery schema

### LINKED ISSUE #7760: BigQuery: schema browser is empty when the service account lacks access to some datasets in the project
### Issue Summary

On a BigQuery data source, the schema browser becomes completely empty (no tables, no columns, no autocomplete) when the service account cannot access *every* dataset in the project. This happens even for the datasets the account *can* read.

The root cause is in `get_schema()` (`redash/query_runner/big_query.py`):

- It lists all datasets in the project via `datasets().list()`.
- It builds a single `UNION ALL` query over each dataset's `INFORMATION_SCHEMA.COLUMN_FIELD_PATHS` (and a second one over `INFORMATION_SCHEMA.TABLE_OPTIONS`).
- If any one dataset returns an error, `_handle_run_query_error()` raises and the whole refresh aborts. Nothing is cached.

So a single inaccessible dataset wipes out the schema for the entire data source.

This is common with dataset-level IAM (per-team isolation): the service account is granted only a subset of datasets, plus `roles/bigquery.jobUser`. Querying the datasets it cannot read returns:

```
403 PERMISSION_DENIED
Access Denied: Table PROJECT:DATASET.INFORMATION_SCHEMA.TABLE_OPTIONS:
User does not have permission to query table
PROJECT:DATASET.INFORMATION_SCHEMA.TABLE_OPTIONS, or perhaps it does not exist.
```

Why existing mechanisms do not help:

- Before #5632, columns were fetched per-table via the metadata API, which tolerated inaccessible datasets. The `UNION ALL` approach introduced there is all-or-nothing.
- The `location` filter from #7289 only excludes datasets in a *different* location than the one configured. When every dataset shares one location (a single region), it excludes nothing and the schema is still empty.

### Steps to Reproduce

1. Use a BigQuery project with multiple datasets, all in the same location (e.g. `asia-northeast1`).
2. Grant a service account read access to only *some* of those datasets, plus `roles/bigquery.jobUser`.
3. Configure a Redash BigQuery data source with that account and "Load Schema" enabled.
4. Trigger a schema refresh (periodic `refresh_schemas`, or open the query editor).

- **Expected:** the schema browser shows the tables of the accessible datasets.
- **Actual:** the schema browser is empty. The `refresh_schema` job fails with a 403 `PERMISSION_DENIED` on `INFORMATION_SCHEMA.TABLE_OPTIONS` / `COLUMN_FIELD_PATHS` of an inaccessible dataset.

Note: query execution is unaffected. Only schema loading and autocomplete break.

### Possible fix

Make `get_schema()` tolerant of per-dataset failures:

- Run the metadata query per dataset (or per location) and skip datasets that error out, instead of aborting the whole refresh.
- The number of calls stays bounded by the dataset count, not the table count, so the performance concern behind #5632 does not return.

### Technical details:

* Redash Version: 26.3.0 (official `redash/redash:26.3.0` image)
* Browser/OS: N/A (server-side schema refresh; reproduced via RQ worker logs)
* How did you install Redash: Docker (official image)

Umekawa:
## What type of PR is this?

- [x] Bug Fix

## Description

`get_schema()` for BigQuery lists every dataset in the project and queries their `INFORMATION_SCHEMA` in a single `UNION ALL`. If any one dataset returns an error, `_handle_run_query_error()` raises and the whole schema refresh aborts, so the schema browser ends up empty.

This breaks environments that use dataset-level IAM (per-team isolation): a service account is granted only a subset of datasets, and the datasets it cannot read return `403 PERMISSION_DENIED` on `INFORMATION_SCHEMA.COLUMN_FIELD_PATHS` / `TABLE_OPTIONS`. A single inaccessible dataset then hides the schema for the datasets the account *can* read.

The `location` filter (#7289) only helps when the inaccessible datasets live in a different location; when every dataset shares one location it excludes nothing.

This PR makes schema loading tolerant of per-dataset failures:

- Send the batched `UNION ALL` query first (fast path, unchanged for fully-accessible projects).
- If it fails, fall back to querying each dataset individually and skip the ones that error.

Per-dataset fallback queries are bounded by dataset count (not table count), so the performance win from #5632 is preserved for the common case.

Compatible with the existing BigQuery query runner; no configuration or schema changes.

## How is this tested?

Added `tests/query_runner/test_bigquery.py`:

- `test_get_schema_skips_datasets_without_access`: an inaccessible dataset is skipped and the accessible dataset's schema still loads.
- `test_get_schema_uses_single_batched_query_when_all_accessible`: when all datasets are accessible, only the batched queries run (no per-dataset fallback).

Closes #7760
Refs #5632, #7289

<a href="https://cubic.dev/pr/getredash/redash/pull/7763?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/redash/query_runner/big_query.py
+++ b/redash/query_runner/big_query.py
@@ -333,47 +333,56 @@ class BigQuery(BaseSQLQueryRunner):
                 location_dataset_ids[location] = []
             location_dataset_ids[location].append(dataset_id)
 
-        for location, datasets in location_dataset_ids.items():
-            queries = []
-            for dataset_id in datasets:
-                query = query_base.format(dataset_id=dataset_id)
-                queries.append(query)
-
-            query = "\nUNION ALL\n".join(queries)
-            results, error = self.run_query(query, None)
-            if error is not None:
-                self._handle_run_query_error(error)
-
-            for row in results["rows"]:
-                table_name = "{0}.{1}".format(row["table_schema"], row["table_name"])
-                if table_name not in schema:
-                    schema[table_name] = {"name": table_name, "columns": []}
-                schema[table_name]["columns"].append(
-                    {
-                        "name": row["field_path"],
-                        "type": row["data_type"],
-                        "description": row["description"],
-                    }
-                )
-
-            table_queries = []
-            for dataset_id in datasets:
-                table_query = table_query_base.format(dataset_id=dataset_id)
-                table_queries.append(table_query)
-
-            table_query = "\nUNION ALL\n".join(table_queries)
-            results, error = self.run_query(table_query, None)
-            if error is not None:
-                self._handle_run_query_error(error)
-
-            for row in results["rows"]:
-                table_name = "{0}.{1}".format(row["table_schema"], row["table_name"])
-                if table_name not in schema:
-                    schema[table_name] = {"name": table_name, "columns": []}
-                if "table_description" in row:
-                    schema[table_name]["description"] = row["table_description"]
+        for location, dataset_ids in location_dataset_ids.items():
+            self._load_columns(dataset_ids, query_base, schema)
+            self._load_table_descriptions(dataset_ids, table_query_base, schema)
 
         return list(schema.values())
+
+    def _run_schema_query(self, dataset_ids, query_template):
+        """Fetch schema metadata for the given datasets, tolerating per-dataset errors.
+
+        Sends a single batched UNION ALL query first (fast path). If it fails -
+        e.g. the service account cannot access one of the datasets - falls back to
+        querying each dataset individually and skips the ones that error, so a
+        single inaccessible dataset doesn't wipe out the schema for the whole data
+        source.
+        """
+        query = "\nUNION ALL\n".join(query_template.format(dataset_id=dataset_id) for dataset_id in dataset_ids)
+        results, error = self.run_query(query, None)
+        if error is None:
+            return results["rows"]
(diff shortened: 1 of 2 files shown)


yoshiokatsuneo:
@Umekawa 

Thank you for your contribution.

If you want to allow redash to access only the specific datasets, I can suggest to disallow access to the datasets including metadata from redash(IAM) (For example, by removing project level "roles/bigquery.metadataViewer" role having bigquery.datasets.get permission.)
Then, as redash even don't know the name of datasets, and does not cause the problem.

Umekawa:
Thank you for the suggestion! That makes sense — I'll try the IAM-side approach in our environment (dataset-level grants without project-level dataset visibility) instead of patching the query runner. Closing this PR. Thanks for taking the time to look at it!

Umekawa:
Thank you for the suggestion! That makes sense — I'll try the IAM-side approach in our environment (dataset-level grants without project-level dataset visibility) instead of patching the query runner. Closing this PR. Thanks for taking the time to look at it!

PR #7729, opened 2026-06-02
security: upgrade authlib, cryptography, and dependent packages

wtfiwtz:
## What type of PR is this?

* Refactor

## Description

Upgrade authlib to 1.7.2, which requires a chain of dependent upgrades due to minimum version constraints in the dependency tree.

**Core security upgrades:**
- authlib: 0.15.5 → 1.7.2 (OAuth/OIDC CVE fixes including **CVE-2024-37568**)
- cryptography: 43.0.1 → 48.0.0 (required by authlib 1.7, multiple OpenSSL CVEs)
- pyjwt: 2.4.0 → 2.12.0 (required by snowflake, fixes **CVE-2022-29217**)
- pyopenssl: 24.2.1 → 26.2.0 (required by cryptography 48)
- requests: 2.32.3 → 2.33.0 (required by snowflake 4.5)

**Data source dependency upgrades (required by above):**
- snowflake-connector-python: 3.12.3 → 4.5.0 (requires cffi >=2.0, cryptography >=45)
- New transitive deps: azure-core, grpcio, h11, httpcore, marshmallow

**OAuth API migration for authlib 1.x:**
- Pass explicit `client_id`/`client_secret` to `oauth.register()` in `google_oauth.py`
- authlib 1.x requires explicit OAuth client credentials in register() call

These upgrades form a tightly coupled dependency chain where each package requires specific minimum versions of others, making them impractical to split into separate PRs:
- authlib 1.7 requires cryptography >=45
- cryptography 48 requires pyopenssl >=26
- snowflake 4.5 requires cryptography >=45, pyjwt >=2.10, requests >=2.32.4

## How is this tested?

* Unit tests (pytest) - OAuth/auth flows tested
* Manually (container testing required for full validation)

## Related Tickets & Documents

Split from orchestrated-io/redash#7719 per @zachliu's review feedback. This PR groups tightly coupled dependencies that cannot be upgraded independently.

Part of the security vulnerability remediation work tracked in #7711.

## Note

This PR is marked as **draft** until manual testing is complete.

Made with [Cursor](https://cursor.com)

PR #7768, opened 2026-07-02
Security: upgrade Authlib to 1.7.x and dependent security packages

madmuffin1:
## What

Upgrade Authlib and its security-sensitive dependencies to current releases:

* `authlib` 0.15.5 → 1.7.2
* `cryptography` 43.0.1 → 48.0.0
* `pyjwt` 2.4.0 → 2.12.0
* `pyopenssl` 24.2.1 → 26.2.0
* `requests` 2.32.3 → 2.33.0
* `snowflake-connector-python` 3.12.3 → 4.5.0

Plus explicit constraints on transitive dependencies to pull in patched versions: `azure-core`, `grpcio`, `h11`, `httpcore`, `marshmallow`.

`uv.lock` is regenerated accordingly (rebased onto the recent Poetry → uv migration on master).

## Why

Authlib 0.15.5 is several years old and well behind current security releases; the same is true for the crypto/transport stack it pulls in (cryptography, pyjwt, pyopenssl, requests). This bumps them to actively maintained versions so Redash ships without the known-vulnerable pins.

## Code change

Authlib 1.x no longer reads the client credentials implicitly during OpenID Connect registration, so `create_google_oauth_blueprint` now passes `client_id` / `client_secret` explicitly to `oauth.register()`. Without this, Google OAuth login breaks under Authlib 1.x.

## Testing

* `uv lock` regenerated cleanly and resolves (239 packages).
* Google OAuth login flow verified against the updated Authlib API.

Evolved from #7729, which was split of #7711 

<a href="https://cubic.dev/pr/getredash/redash/pull/7768?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


madmuffin1:
this PR contains the changes by @wtfiwtz in PR #7729 , with some amendments - as suggested by @zachliu .
I tried to add them on top of 7729, but have not received a response in a week, so I decided to publish them independently

madmuffin1:
also applied changes required by the poetry -> uv migration.

zachliu:
@madmuffin1 thank you! one last thing, have you done a smoke test on `snowflake`?  the `3.x` to `4.x` jump is the one runtime risk nobody has explicitly validated :thinking:

madmuffin1:
Sure, good point.

### Why the `snowflake-connector-python` 3.12.3 → 4.5.0 bump is required at all

It is not a standalone change — it is forced by the Authlib upgrade that is the point of this PR. The resolver chain (verified with `uv`):

* `authlib==1.7.2` → depends on `joserfc>=1.6.0` → which requires `cryptography>=45.0.1`, so cryptography must move up.
* `cryptography==48.0.0` → `pyopenssl` 24.x caps `cryptography<44` and 25.x caps `cryptography<47`, so `pyopenssl` must go to **26.x**.
* `pyopenssl==26.2.0` → `snowflake-connector-python==3.12.3` caps `pyopenssl>=16.2.0,<25.0.0`; every release up to 4.3 caps `pyopenssl<26`. **4.4.0 is the first release that allows pyopenssl 26**, so 4.5.0 is the current viable version.

So: `authlib → cryptography → pyopenssl → snowflake`. Snowflake had to leave 3.x regardless (cryptography ≥45 already requires pyopenssl ≥25, which 3.12.3 forbids). As a bonus, moving off 3.12.3 also clears CVE-2025-24793/24794/24795 (fixed upstream in 3.13.1).

### Verification

I built the image from this branch (`uv sync --frozen`, `all_ds` group) and connected to a live Snowflake demo account through the query runner:

* `test_connection` (`SELECT 1`) — succeeded
* `SELECT CURRENT_VERSION()` — returned `10.23.103`

Confirms `snowflake-connector-python==4.5.0` installs, imports, and queries Snowflake correctly alongside `cryptography==48.0.0` / `pyopenssl==26.2.0` / `authlib==1.7.2`.


To me this seems to be fine, but since I don't use snowflake at all, I cannot tell.
In addition I checked the snowflake-connector-python's [changelog](https://docs.snowflake.com/en/release-notes/clients-drivers/python-connector-2025#version-400-oct-9-2025), and even the 3.x to 4.x does not contain any hints about breaking changes in the API, just a behaviour change when configuration files permission are not secure (this is not used in redash afaik)

zachliu (approved):
@madmuffin1 thank you for your contribution!

zachliu:
@madmuffin1 i can cut a new taggable release `26.07.1-dev` so that you don't have to wait until August. before i do that, are there any other security fixes you may want to include?

madmuffin1:
nice, thanks. I believe that should close the two critical findings in the dependency tree. If you create a tag, will that trigger the creation of a new image? If so, I'd appreciate that very much. If no docker image is published for the tag, the tag doesn't help - I could as well just build the image from master. (but I don't want to go that path 😉 )

zachliu:
@madmuffin1 yes, i triggered the build just now and a new image is being created: `26.07.1-dev`

madmuffin1:
@zachliu thanks for the effort. apparently this updated the :preview tag published on dockerhub, and did not create a 27.07.1-dev tag. We can not use floating tags for security reasons, and going after the hash is also discouraged. If this lands in the "regular" release in August we will be fine.

zachliu:
@madmuffin1 oh, so you want it in [redash/redash](https://hub.docker.com/r/redash/redash/tags) not [redash/preview](https://hub.docker.com/r/redash/preview/tags) right? fwiw, we do have a `27.07.1-dev` tag in the latter

madmuffin1:
Thanks for the hint, I was not aware of the redash/preview image, only noticed that the redash/redash:preview image was updated. helps a lot, thanks

PR #7774, opened 2026-07-27
Remove invalid single-backslash redirect test

arikfr:
## Summary

Remove the login-redirect test that treats a single leading backslash as an external redirect.

## Rationale

A single `\evil.com` resolves to the same-origin path `/evil.com` for special-scheme URLs. The existing mixed slash/backslash case (`/\evil.com`) remains covered because it resolves to an external host and must be rejected.

## Validation

- `python3 -m py_compile redash/authentication/__init__.py`
- `git diff --check`
- Required GitHub Actions checks are running.

### DIFF
--- a/redash/authentication/__init__.py
+++ b/redash/authentication/__init__.py
@@ -316,11 +316,6 @@ def _is_safe_next_url(url):
     if unicodedata.category(url[0])[0] == "C":
         return False
 
-    # Browsers normalize a leading backslash during redirect handling. Reject it
-    # rather than returning a non-canonical Location header.
-    if url.startswith("\\"):
-        return False
-
     # Chrome treats \ as / in URLs, so check both the original and
     # backslash-normalized versions to prevent bypasses like \/evil.com
     for test_url in (url, url.replace("\\", "/")):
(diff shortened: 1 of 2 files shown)


yoshiokatsuneo (approved):
I can approve it as It just looks good.

I just have some questions.

- Is the test failed before the PR ?
  (Normally, the commit causing test failure is not merged. I guess the commit causing the test failure is directly committed to the master branch. Is it right ?)

- Why is the commit (https://github.com/getredash/redash/pull/7774/changes/2723d87d616c822b6cb874bfffb1185653777835) reverted, as the commit looks more conservative.
  ( I guess the commit caused another test failure ??)

yoshiokatsuneo:
Actually, I see that the this commit(security fix) causes CI error on the latest commit like this.

https://github.com/getredash/redash/commit/9e66f81673c482d9ae6c425afe009644114605d0

yoshiokatsuneo:
Just to make clear, this issue was introduced by the security fix below. The PR for this commit is not visible (at least in public.)

https://github.com/getredash/redash/commit/9e66f81673c482d9ae6c425afe009644114605d0

arikfr:
Thanks @yoshiokatsuneo for reviewing and merging.

It was merged via the security advisory workflow, which apparently can bypass the branch protection 🤦‍♂️

I removed the test instead of fixing it, as it wasn't needed (just Claude overdoing in the original tests implementation).

PR #7782, opened 2026-08-09
Fix DSV export for NUL bytes

### LINKED ISSUE #7758: CSV download fails with "_csv.Error: need to escape, but no escapechar set" when query results contain NUL bytes (\x00)
### Issue Summary

CSV download fails with HTTP 500 when query result rows contain a NUL byte (`\x00`, `CHAR(0)`) in string fields. The browser may save a `.csv` file whose content is a JSON error response instead of CSV data. Excel download for the same query result succeeds.

This occurs on Redash **25.8.0 (67a95e9d)** with **Python 3.10** (self-hosted Docker). The failure is server-side; it is reproducible regardless of browser/OS.

Server log:
```
[ERROR][redash.app] Exception on /api/queries/43/results/xxxx.csv [GET]
...
File "/app/redash/serializers/query_result.py", line 98, in serialize_query_result_to_dsv
    writer.writerow(row)
_csv.Error: need to escape, but no escapechar set
```

### Steps to Reproduce

1. Ensure the database contains a string value with an embedded NUL byte, e.g.:
   ```sql
   SELECT id, address
   FROM users
   WHERE INSTR(address, CHAR(0)) > 0;
   ```
2. Create and run a Redash query that returns the affected row(s).
3. Click **Download as CSV File** → download fails (HTTP 500; JSON error may be saved as `.csv`).
4. Click **Download as Excel File** for the same result → download succeeds.

This is a bug because CSV export should handle real-world string data safely. NULL bytes can exist in user-entered text stored in databases, and CSV download should not crash while Excel download works for the same data.

**Expected:** CSV download succeeds (values are escaped or sanitized).

**Actual:** CSV download raises `_csv.Error: need to escape, but no escapechar set` in `serialize_query_result_to_dsv()` because Python 3.10's `csv.DictWriter` cannot write fields containing `\x00` with default settings. Excel export uses `xlsxwriter` and is unaffected.

Related upstream: [pandas-dev/pandas#47871](https://github.com/pandas-dev/pandas/issues/47871)

**Suggested fix:** Sanitize string values in `redash/serializers/query_result.py` before `writer.writerow()`, e.g. `value.replace("\x00", "")`, or configure the CSV writer with an explicit `escapechar`.

**Workaround:** Download as Excel and convert to CSV; or in SQL use `REPLACE(column, CHAR(0), '')` (MySQL).

### Technical details:

* Redash Version: 25.8.0 (67a95e9d)
* Browser/OS: Chrome / macOS (issue is server-side; not browser-specific)
* How did you install Redash: Docker (self-hosted)

DhruvShah-Dev:
## What type of PR is this?
Bug fix

## Description
CSV and TSV export can fail when query results contain embedded NUL bytes because Python's CSV writer raises while trying to escape those values. This change removes NUL bytes from string fields before DSV serialization and avoids mutating the stored query result row during export.

Fixes #7758

## Validation
- `git diff --check`
- `python -m pytest tests\serializers\test_query_results.py` (blocked locally: missing Redash runtime dependency `redis`)

<a href="https://cubic.dev/pr/getredash/redash/pull/7782?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/redash/serializers/query_result.py
+++ b/redash/serializers/query_result.py
@@ -72,6 +72,13 @@ def _get_column_lists(columns):
     return fieldnames, special_columns
 
 
+def _sanitize_dsv_value(value):
+    if isinstance(value, str):
+        return value.replace("\x00", "")
+
+    return value
+
+
 def serialize_query_result(query_result, is_api_user):
     if is_api_user:
         publicly_needed_keys = ["data", "retrieved_at"]
@@ -91,11 +98,14 @@ def serialize_query_result_to_dsv(query_result, delimiter):
     writer.writeheader()
 
     for row in query_data["rows"]:
+        serialized_row = dict(row)
         for col_name, converter in special_columns.items():
-            if col_name in row:
-                row[col_name] = converter(row[col_name])
+            if col_name in serialized_row:
+                serialized_row[col_name] = converter(serialized_row[col_name])
 
-        writer.writerow(row)
+        writer.writerow(
+            {key: _sanitize_dsv_value(value) for key, value in serialized_row.items()}
+        )
 
     return s.getvalue()
 
--- a/redash/serializers/query_result.py
+++ b/redash/serializers/query_result.py
@@ -103,9 +103,7 @@ def serialize_query_result_to_dsv(query_result, delimiter):
             if col_name in serialized_row:
                 serialized_row[col_name] = converter(serialized_row[col_name])
 
-        writer.writerow(
-            {key: _sanitize_dsv_value(value) for key, value in serialized_row.items()}
-        )
+        writer.writerow({key: _sanitize_dsv_value(value) for key, value in serialized_row.items()})
 
     return s.getvalue()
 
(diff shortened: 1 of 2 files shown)


DhruvShah-Dev:
Updated this PR with a formatting-only commit to satisfy `backend-lint`.

Validated locally:
- `python -m black --check redash\serializers\query_result.py tests\serializers\test_query_results.py`
- `python -m ruff check redash\serializers\query_result.py tests\serializers\test_query_results.py`

The targeted pytest command still cannot collect locally because this checkout is missing the Redash runtime dependency `redis`.

yoshiokatsuneo:
The python CSV exportor issue looks already being solved on the Python3.13.
I can suggest to test the issue on the latest Redash and the latest usable Python(3.13).

https://github.com/python/cpython/issues/97503

DhruvShah-Dev:
Closing this after testing the same `csv.DictWriter` path on Python 3.13.3, where a value containing `\x00` writes successfully without raising.

Given that the original failure is no longer reproducible on the current supported Python version, withdrawing this sanitization change.

PR #7786, opened 2026-08-12
Apply prettier formatting to client

yoshiokatsuneo:
### What type of PR is this?

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] New Alert Destination
- [ ] Other

### Description

**Purpose: stop every individual PR from having to carry prettier changes.**

Today 180 files under the Restyled prettier scope are unformatted on master. Any PR that touches one of them either fails the Restyled check or has to include a reformatting commit of its own. That is inefficient and it makes PRs harder to read: the formatting hunks bury the actual change, so reviewers have to separate "what this PR really does" from "what prettier did" by hand, in every single PR. Doing it once here removes that cost from all future PRs.

The remaining drift comes almost entirely from the prettier 2 → 3 upgrade, which changed the `arrowParens` default from `"avoid"` to `"always"` (`x => x` → `(x) => x`). These files were never reformatted after that upgrade.

Reformatting alone would only reset the clock, so the last commit also makes `pnpm run prettier` and the restyler agree on one file list — including `viz-lib`, which nothing in CI checks today.

No behaviour change. Best reviewed with whitespace changes hidden (`?w=1`), and commit by commit.

### How this diff was produced

Everything here is the output of the prettier scripts already defined in the repo — with exactly one hand-written exception, kept in its own commit. The three commits are meant to be reviewed separately:

**1. `Apply prettier formatting to client`** — 186 files, verbatim output of:

```console
$ pnpm install --frozen-lockfile
$ pnpm run prettier
```

which expands to:

```console
$ prettier --write 'client/app/**/*.{js,jsx,ts,tsx}' 'client/cypress/**/*.{js,jsx,ts,tsx}'
```

**2. `Apply prettier formatting to viz-lib`** — 103 files, verbatim output of:

```console
$ pnpm --filter @redash/viz run prettier
```

which expands to `prettier --write 'src/**/*.{ts,tsx}'`. `viz-lib` is not in the restyler `include` list, so CI never flags it, but it carries the same drift.

**3. `Keep @ts-expect-error suppressions working after reformatting`** — 2 files, +2 lines. **One of the two hand-written commits.** `// @ts-expect-error` suppresses only the line that follows it. In `DefaultColorsSettings.tsx` and `PieColorsSettings.tsx` the `Select.Option` element used to fit on one line, so a single directive covered both the opening and the closing tag; prettier splits the element across three lines, leaving the closing tag uncovered and `tsc` failing with `Property 'Option' does not exist on type ...`. That is not just a `type-check` failure: `postinstall` runs `build:viz`, whose `type-gen` step is `tsc --emitDeclarationOnly`, so `pnpm install --frozen-lockfile` itself would fail. The commit adds the closing-tag suppression in the JSX comment form already used for exactly this case in `AxisSettings.tsx` and `word-cloud/Editor.tsx`. Re-running prettier afterwards produces no further changes.

**4. `Align the prettier scope between pnpm run prettier and the restyler`** — 2 files, one line of `package.json` plus the restyler `include`. The other hand-written commit, and the part that keeps the drift from coming back. Today there are three different answers to "what does prettier cover here":

| | scope |
|---|---|
| `pnpm run prettier` | `client` js/jsx/ts/tsx |
| prettier restyler in `.restyled.yaml`, run by `restyled.yml` | `client` js/jsx **only** |
| `viz-lib`'s own `prettier` script | `viz-lib/src` |

So running the documented command locally does not tell you whether the Restyled check will be happy. And `viz-lib` is checked by nothing: it is outside the restyler `include`, and the restyler runs `prettier` from its own image rather than any of the package scripts, so `viz-lib`'s script is only ever run by hand. This commit adds `viz-lib/src` to the root script and widens the restyler `include` to the same set of files. No new CI step is needed: the Restyled workflow already runs with `fail-on-differences: true`, so widening `include` is what enforces it.

All of the above uses prettier 3.3.2 — the version pinned in `devDependencies`, and the same version as the Restyled prettier restyler in `.restyled.yaml`.

### How is this tested?

- [ ] E2E Tests (Cypress)
- [x] Manually
- [ ] N/A

Everything the `frontend-lint` and `frontend-unit-tests` jobs run, on a clean `pnpm install --frozen-lockfile`:

```console
$ pnpm run test                      # 15 suites, 89 passed / 1 skipped
$ pnpm --filter @redash/viz test     # 24 suites, 154 passed, 61 snapshots
$ pnpm run lint --max-warnings 0     # passes
```

All 61 viz-lib snapshots still match. They capture the options objects the editors produce in response to interaction rather than rendered markup, so what they confirm is that the editor behaviour is unchanged.

Plus, specific to this PR:

```console
$ pnpm run prettier
# no changes — the tree is stable under prettier, including viz-lib

$ pnpm --filter @redash/viz run type-check
# passes (fails without commit 3, see above)
```

The `pnpm install` above is itself part of the check: its `postinstall` runs `build:viz` → `type-gen` → `tsc --emitDeclarationOnly`, which is exactly what commit 3 keeps working.

### Related Tickets & Documents

Prettier was bumped to 3.x without a follow-up reformat, which is where the drift comes from.

**Related: #7669 (React v19 update)** — a concrete example of the problem this PR is meant to remove. That PR states in its own description:

> Formatting note: Prettier was forcefully applied to all files, so the PR includes broad formatting-only churn in addition to the dependency updates.

The result is 418 changed files, in which the actual React 19 / Ant Design 6 migration is mixed with formatting-only churn, so a reviewer cannot tell the two apart without going file by file. Had the tree already been formatted, that churn would simply not have existed and the PR would show only the migration itself.

Ordering is not a blocker either way, but landing this PR first is the cheaper order: #7669 can then rebase and drop its formatting churn, leaving only the migration in its diff. Merging this PR will conflict with #7669, but every such conflict is formatting-only and is resolved by re-running `pnpm run prettier` on the rebased branch. Note that #7669 applied prettier to *all* files, which is wider than the scope here (e.g. it also covers `viz-lib`), so a smaller amount of formatting churn will remain in it.

### Mobile & Desktop Screenshots/Recordings (if there are UI changes)

N/A

---

🤖 This PR was written by Claude Opus 5 (model id: `claude-opus-5`) in [Claude Code](https://claude.com/claude-code), and reviewed by @yoshiokatsuneo before opening.


eradman (approved):
Applying formatting as a separate change is a good idea. This will also make review of contributions easier.

yoshiokatsuneo:
@eradman Thank you for your review !

PR #7791, opened 2026-08-19
Fix 12-column layout migration crash on widgets with empty options

### LINKED ISSUE #7561: Widgets fail to render when options column is empty ({}) after dashboard migration or upgrade
### Issue Summary
We tried to upgrade from [v10.1.0](https://github.com/getredash/redash/releases/tag/v10.1.0) to [v25.8.0](https://github.com/getredash/redash/releases/tag/v25.8.0)
When using Redash dashboards, certain widget entries in the widgets PostgreSQL table contain empty JSON objects ({}) in the options column. This causes unexpected rendering or layout inconsistencies `within` the dashboard interface. Specifically, widgets with empty options fail to display correctly until their configuration is manually updated in the database.
The issue was the following:
`
Traceback (most recent call last): File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context self.dialect.do_execute( File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute cursor.execute(statement, parameters) psycopg2.errors.NotNullViolation: null value in column "options" of relation "widgets" violates not-null constraint DETAIL: Failing row contains (2019-12-11 15:12:50.156141+00, 2019-12-11 15:12:50.156141+00, 300, 1996, , 2, null, 52). The above exception was the direct cause of the following exception: Traceback (most recent call last): File "/app/./manage.py", line 9, in <module> manager() File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1130, in __call__ return self.main(*args, **kwargs) File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main rv = self.invoke(ctx) File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke return _process_result(sub_ctx.command.invoke(sub_ctx)) File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke return _process_result(sub_ctx.command.invoke(sub_ctx)) File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke return ctx.invoke(self.callback, **ctx.params) File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke return __callback(*args, **kwargs) File "/usr/local/lib/python3.10/site-packages/click/decorators.py", line 26, in new_func return f(get_current_context(), *args, **kwargs) File "/usr/local/lib/python3.10/site-packages/flask/cli.py", line 357, in decorator return __ctx.invoke(f, *args, **kwargs) File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke return __callback(*args, **kwargs) File "/usr/local/lib/python3.10/site-packages/flask_migrate/cli.py", line 134, in upgrade _upgrade(directory, revision, sql, tag, x_arg) File "/usr/local/lib/python3.10/site-packages/flask_migrate/__init__.py", line 95, in wrapped f(*args, **kwargs) File "/usr/local/lib/python3.10/site-packages/flask_migrate/__init__.py", line 280, in upgrade command.upgrade(config, revision, sql=sql, tag=tag) File "/usr/local/lib/python3.10/site-packages/alembic/command.py", line 406, in upgrade script.run_env() File "/usr/local/lib/python3.10/site-packages/alembic/script/base.py", line 586, in run_env util.load_python_file(self.dir, "env.py") File "/usr/local/lib/python3.10/site-packages/alembic/util/pyfiles.py", line 95, in load_python_file module = load_module_py(module_id, path) File "/usr/local/lib/python3.10/site-packages/alembic/util/pyfiles.py", line 113, in load_module_py spec.loader.exec_module(module) # type: ignore File "<frozen importlib._bootstrap_external>", line 883, in exec_module File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed File "/app/migrations/env.py", line 93, in <module> run_migrations_online() File "/app/migrations/env.py", line 85, in run_migrations_online context.run_migrations() File "<string>", line 8, in run_migrations File "/usr/local/lib/python3.10/site-packages/alembic/runtime/environment.py", line 946, in run_migrations self.get_context().run_migrations(**kw) File "/usr/local/lib/python3.10/site-packages/alembic/runtime/migration.py", line 623, in run_migrations step.migration_fn(**kw) File "/app/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py", line 20, in upgrade op.execute(""" File "<string>", line 8, in execute File "<string>", line 3, in execute File "/usr/local/lib/python3.10/site-packages/alembic/operations/ops.py", line 2551, in execute return operations.invoke(op) File "/usr/local/lib/python3.10/site-packages/alembic/operations/base.py", line 442, in invoke return fn(self, operation) File "/usr/local/lib/python3.10/site-packages/alembic/operations/toimpl.py", line 236, in execute_sql operations.migration_context.impl.execute( File "/usr/local/lib/python3.10/site-packages/alembic/ddl/impl.py", line 254, in execute self._exec(sql, execution_options) File "/usr/local/lib/python3.10/site-packages/alembic/ddl/impl.py", line 247, in _exec return conn.execute(construct, params) File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1011, in execute return meth(self, multiparams, params) File "/usr/local/lib/python3.10/site-packages/sqlalchemy/sql/elements.py", line 298, in _execute_on_connection return connection._execute_clauseelement(self, multiparams, params) File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1124, in _execute_clauseelement ret = self._execute_context( File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1316, in _execute_context self._handle_dbapi_exception( File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1510, in _handle_dbapi_exception util.raise_( File "/usr/local/lib/python3.10/site-packages/sqlalchemy/util/compat.py", line 182, in raise_ raise exception File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context self.dialect.do_execute( File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute cursor.execute(statement, parameters) sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) null value in column "options" of relation "widgets" violates not-null constraint DETAIL: Failing row contains (2019-12-11 15:12:50.156141+00, 2019-12-11 15:12:50.156141+00, 300, 1996, , 2, null, 52). [SQL: UPDATE widgets SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb); UPDATE widgets SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int * 2)::jsonb); ] (Background on this error at: http://sqlalche.me/e/13/gkpj)
`

### Steps to Reproduce

1. Open a dashboard that contains widgets with an empty options column in the database.
2. Observe that the widgets are missing or fail to render.
3. Check the database and confirm that some widgets have options = '{}'.

We have created a temporary fix that applies a proper rendering of the affected widgets in-place:

```
UPDATE widgets
SET options = '{"isHidden": false, "position": {"autoHeight": false, "sizeX": 1, "sizeY": 5, "minSizeX": 1, "maxSizeX": 6, "minSizeY": 1, "maxSizeY": 1000, "col": 3, "row": 8}, "parameterMappings": {}}'
where dashboard_id=<id>;
``` 

### Technical details:

* Redash Version: 10.1.0
* Browser/OS:
* How did you install Redash: Kubernetes/Docker

ekanshul:
## What type of PR is this?

- [x] Bug Fix

## Description

Fixes #7561.

Upgrading from v10 to a current release runs the `db0aca1ebd32` (12-column dashboard layout) migration, which rewrites every `widgets` row:

```sql
UPDATE widgets
SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb);
```

For widgets whose `options` don't contain the position keys — e.g. rows created through the API with `options = '{}'`, as reported in the issue — `(options->'position'->>'col')::int` evaluates to NULL, and since `jsonb_set()` is strict, the whole expression returns SQL `NULL`. The UPDATE then violates the `NOT NULL` constraint on `widgets.options` and the entire upgrade aborts:

```
psycopg2.errors.NotNullViolation: null value in column "options" of relation "widgets" violates not-null constraint
```

This PR restricts both UPDATE statements (and their `downgrade()` counterparts) to rows where the value being doubled/halved is actually a base-10 integer:

```sql
WHERE options->'position'->>'col' ~ '^-?[0-9]+$'
```

Degenerate rows (`{}`, missing `position`, non-numeric values) are left untouched instead of aborting the migration; they render the same as before the migration, and the layout of all valid widgets is converted as intended.

## How is this tested?

- [x] Manually

Against `postgres:18-alpine` with a table shaped like `widgets` (`options jsonb NOT NULL`) containing rows `{"position": {"col": 3, "sizeX": 2, "sizeY": 5}}`, `{}`, `{"isHidden": false}`, and `{"position": {"row": 1}}`:

- The current migration statement fails with `null value in column "options" ... violates not-null constraint` (reproducing the issue).
- With this change, the upgrade statements double `col`/`sizeX` on the valid row and leave the other rows untouched, and the downgrade statements round-trip back to the original values without errors.

#7561

N/A (migration only)

<a href="https://cubic.dev/pr/getredash/redash/pull/7791?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
+++ b/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
@@ -20,22 +20,24 @@ def upgrade():
     # Guard against widgets whose options lack the position keys (e.g. rows written
     # through the API with options = '{}'): jsonb_set() with a NULL new_value returns
     # NULL, which would violate the NOT NULL constraint on widgets.options.
+    # The 9-digit bound also keeps the ::int cast and the doubling within int4 range.
     op.execute("""
     UPDATE widgets
     SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb)
-    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
+    WHERE options->'position'->>'col' ~ '^-?[0-9]{1,9}$';
     UPDATE widgets
     SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int * 2)::jsonb)
-    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
+    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]{1,9}$';
     """)
 
 
 def downgrade():
+    # The 10-digit bound and bigint cast cover everything upgrade() can produce.
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb)
-    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::bigint / 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]{1,10}$';
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int / 2)::jsonb)
-    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
+    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::bigint / 2)::jsonb)
+    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]{1,10}$';
     """)
--- a/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
+++ b/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
@@ -17,34 +17,26 @@ depends_on = None
 
 
 def upgrade():
-    # Guard against widgets whose options lack the position keys (e.g. rows written
-    # through the API with options = '{}'): jsonb_set() with a NULL new_value returns
-    # NULL, which would violate the NOT NULL constraint on widgets.options.
-    # The 9-digit bound also keeps the ::int cast and the doubling within int4 range.
+    # Skip widgets whose position values are missing or not numeric, e.g. rows
+    # written through the API with options = '{}'. Without the filter the cast
+    # yields NULL, jsonb_set() returns NULL for a NULL new_value, and the update
+    # fails the NOT NULL constraint on widgets.options.
     op.execute("""
     UPDATE widgets
     SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb)
-    WHERE options->'position'->>'col' ~ '^-?[0-9]{1,9}$';
+    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
     UPDATE widgets
     SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int * 2)::jsonb)
-    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]{1,9}$';
+    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
     """)
 
 
 def downgrade():
-    # Halve exactly the values the original statement could halve (anything in
-    # int4 range, which also covers everything upgrade() can produce) and skip
-    # the rest instead of aborting on the cast. The CASE guarantees the regex is
-    # checked before the cast (AND operands have no defined evaluation order).
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::bigint / 2)::jsonb)
-    WHERE CASE WHEN options->'position'->>'col' ~ '^-?[0-9]{1,10}$'
-               THEN (options->'position'->>'col')::bigint BETWEEN -2147483648 AND 2147483647
-               ELSE false END;
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::bigint / 2)::jsonb)
-    WHERE CASE WHEN options->'position'->>'sizeX' ~ '^-?[0-9]{1,10}$'
-               THEN (options->'position'->>'sizeX')::bigint BETWEEN -2147483648 AND 2147483647
-               ELSE false END;
+    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int / 2)::jsonb)
+    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
     """)
--- a/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
+++ b/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
@@ -17,17 +17,17 @@ depends_on = None
 
 
 def upgrade():
-    # Skip widgets whose position values are missing or not numeric, e.g. rows
-    # written through the API with options = '{}'. Without the filter the cast
-    # yields NULL, jsonb_set() returns NULL for a NULL new_value, and the update
-    # fails the NOT NULL constraint on widgets.options.
+    # Skip widgets that have no position values, e.g. rows written through the
+    # API with options = '{}'. Without the filter the cast yields NULL,
+    # jsonb_set() returns NULL for a NULL new_value, and the update fails the
+    # NOT NULL constraint on widgets.options.
     op.execute("""
     UPDATE widgets
     SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb)
-    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
+    WHERE options->'position'->>'col' IS NOT NULL;
     UPDATE widgets
     SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int * 2)::jsonb)
-    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
+    WHERE options->'position'->>'sizeX' IS NOT NULL;
     """)
 
 
@@ -35,8 +35,8 @@ def downgrade():
     op.execute("""
     UPDATE widgets
     SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb)
-    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
+    WHERE options->'position'->>'col' IS NOT NULL;
     UPDATE widgets
     SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int / 2)::jsonb)
-    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
+    WHERE options->'position'->>'sizeX' IS NOT NULL;
     """)
--- a/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
+++ b/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
@@ -17,18 +17,25 @@ depends_on = None
 
 
 def upgrade():
+    # Guard against widgets whose options lack the position keys (e.g. rows written
+    # through the API with options = '{}'): jsonb_set() with a NULL new_value returns
+    # NULL, which would violate the NOT NULL constraint on widgets.options.
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int * 2)::jsonb);
+    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int * 2)::jsonb)
+    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
     """)
 
 
 def downgrade():
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int / 2)::jsonb)
+    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
     """)
--- a/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
+++ b/migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
@@ -34,14 +34,17 @@ def upgrade():
 def downgrade():
     # Halve exactly the values the original statement could halve (anything in
     # int4 range, which also covers everything upgrade() can produce) and skip
-    # the rest instead of aborting on the cast.
+    # the rest instead of aborting on the cast. The CASE guarantees the regex is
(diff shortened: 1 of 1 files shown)


migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
```diff
@@ -17,18 +17,25 @@
...
 
 def upgrade():
+    # Guard against widgets whose options lack the position keys (e.g. rows written
+    # through the API with options = '{}'): jsonb_set() with a NULL new_value returns
+    # NULL, which would violate the NOT NULL constraint on widgets.options.
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
```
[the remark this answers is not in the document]
ekanshul: Fixed in the 791eca8 follow-up: the upgrade regex is bounded to 9 digits, so the ::int cast and the doubling always stay within int4 range; longer integer-looking values are skipped rather than aborting the migration.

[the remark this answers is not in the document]
ekanshul: Fixed in 791eca8 (9-digit bound on upgrade) — verified against postgres:18 with 99999999999999999 and 2000000000 in col/sizeX: both are skipped and the migration completes.

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
```diff
@@ -17,18 +17,27 @@
...
     """)
 
 
 def downgrade():
+    # The 10-digit bound and bigint cast cover everything upgrade() can produce.
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::bigint / 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]{1,10}$';
```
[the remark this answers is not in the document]
ekanshul: Done in 41f46eda: the downgrade is now bounded to int4 range, i.e. it halves exactly the set of values the original statement could halve (which includes everything upgrade() can produce, since doubling a 9-digit value stays below int4 max) and skips only rows that would have aborted the cast. Verified the round trip against postgres:18.

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
```diff
@@ -17,18 +17,31 @@
...
 def downgrade():
+    # Halve exactly the values the original statement could halve (anything in
+    # int4 range, which also covers everything upgrade() can produce) and skip
+    # the rest instead of aborting on the cast.
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::bigint / 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]{1,10}$'
+      AND (options->'position'->>'col')::bigint BETWEEN -2147483648 AND 2147483647;
```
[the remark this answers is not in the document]
ekanshul: Fixed in b77efbd9: the downgrade predicate is wrapped in CASE WHEN <regex> THEN <range check> ELSE false END, so the cast can only be evaluated after the regex has matched. Verified against postgres:18 with non-numeric ("garbage", "12abc"), out-of-int4-range, and sane values: malformed rows stay untouched and valid rows round-trip.

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
```diff
@@ -17,18 +17,34 @@
...
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::bigint / 2)::jsonb)
+    WHERE CASE WHEN options->'position'->>'col' ~ '^-?[0-9]{1,10}$'
+               THEN (options->'position'->>'col')::bigint BETWEEN -2147483648 AND 2147483647
+               ELSE false END;
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::bigint / 2)::jsonb)
+    WHERE CASE WHEN options->'position'->>'sizeX' ~ '^-?[0-9]{1,10}$'
+               THEN (options->'position'->>'sizeX')::bigint BETWEEN -2147483648 AND 2147483647
```
eradman: What is the purpose of testing the range of a 32-bit signed int?

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
```diff
@@ -17,18 +17,34 @@
...
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::bigint / 2)::jsonb)
+    WHERE CASE WHEN options->'position'->>'col' ~ '^-?[0-9]{1,10}$'
+               THEN (options->'position'->>'col')::bigint BETWEEN -2147483648 AND 2147483647
+               ELSE false END;
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,sizeX}', to_json((options->'position'->>'sizeX')::bigint / 2)::jsonb)
```
eradman: why cast this to a bigint?

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
ekanshul: No good reason, and I have removed it. It came from me over-reacting to automated review comments on this PR rather than from the bug being fixed. The range test and the bigint cast are both gone now.

The actual crash is a widget whose options have no position values, for example a row written through the API with `options = '{}'`. The cast then yields NULL, `jsonb_set()` returns NULL when the new value is NULL, and the update fails the NOT NULL constraint on `widgets.options`. A plain numeric check is enough to skip those rows, so both statements are now just:

```sql
WHERE options->'position'->>'col' ~ '^-?[0-9]+$'
```

The cast stays `::int` in both directions, exactly as the original migration had it.

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
ekanshul: Removed, along with the int4 range test. It is back to `::int` in both directions, matching the original statement, with only a numeric check added to skip the rows that used to abort the migration.

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
```diff
@@ -17,18 +17,26 @@
...
 def upgrade():
+    # Skip widgets whose position values are missing or not numeric, e.g. rows
+    # written through the API with options = '{}'. Without the filter the cast
+    # yields NULL, jsonb_set() returns NULL for a NULL new_value, and the update
+    # fails the NOT NULL constraint on widgets.options.
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int * 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
```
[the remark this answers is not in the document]
ekanshul: Not making this change, because it is pre-existing behaviour rather than something this PR introduces, and the range guard you are describing is exactly what the maintainer asked me to take out.

A digit-only value above the int4 range aborts the migration on master too. The original statement had no WHERE clause at all, so it cast every row:

```
-- original migration, row with col = 3000000000
ERROR:  value "3000000000" is out of range for type integer
-- this PR, same row
ERROR:  value "3000000000" is out of range for type integer
```

Identical outcome, so the filter is not a regression. What it does change is the set of rows that abort: this PR strictly shrinks it, because rows whose position values are missing or non-numeric used to fail the NOT NULL constraint and now get skipped. That missing-position case is the crash this PR is about.

Guarding the int4 range as well would be a separate change to pre-existing behaviour, and @eradman asked on this PR why the 32-bit range test and the bigint cast were there. They were there because of earlier review comments on this PR, not because of a bug, so I removed them. I would rather keep this fix to the reported crash than re-add that machinery.

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
```diff
@@ -17,18 +17,26 @@
...
+    WHERE options->'position'->>'sizeX' ~ '^-?[0-9]+$';
     """)
 
 
 def downgrade():
     op.execute("""
     UPDATE widgets
-    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb);
+    SET options = jsonb_set(options, '{position,col}', to_json((options->'position'->>'col')::int / 2)::jsonb)
+    WHERE options->'position'->>'col' ~ '^-?[0-9]+$';
```
eradman: We would expect a number or it is empty, let's just check for NULL
```
WHERE options->'position'->>'col' IS NOT NULL
```

eradman (changes requested):

ekanshul:
Done in b8c9aa0a. Both statements in `upgrade()` and both in `downgrade()` now use `IS NOT NULL`; the casts stay `::int`, as in the original migration. I also reworded the comment above `upgrade()`, which still described the pattern match.

Verified against postgres:18, one row per shape, comparing this branch with master (whose statements have no `WHERE` at all):

| row | master | this branch |
|---|---|---|
| `{}` (the reported crash) | `ERROR: null value in column "options"` | updated, no error |
| `{"col": 3}` | ok | ok |
| `{"col": 3.0}` | `ERROR: invalid input syntax for type integer: "3.0"` | same error |
| `{"col": true}` | `ERROR: ... "true"` | same error |
| `{"col": "abc"}` | `ERROR: ... "abc"` | same error |
| `{"col": 2000000000}` | `ERROR: integer out of range` | same error |

So the set of rows that abort is a strict subset of master's: this fixes the empty-position case and changes nothing else.

One thing worth being precise about, since it touches your reasoning. The pattern match was skipping more than empty values: it also skipped anything that is not a base-10 integer. With `IS NOT NULL` those rows are selected again and `::int` aborts on them, and that includes JSON numbers such as `3.0` or `1.5`, not just strings like `"abc"`. `widgets.options` has no schema validation on the write path (`WidgetListResource.post` does `models.Widget(**request.get_json(force=True))`), so a client can store one, though the frontend always writes integers.

None of that is a regression, since master aborts on exactly the same rows. I mention it only because "we would expect a number or it is empty" holds for the values Redash itself writes, but a stored `3.0` is a number that would now stop the migration. If you would rather the migration skip those too, say the word and I will put the numeric check back on top of the NULL check; otherwise this is ready as it stands.

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
```diff
@@ -17,18 +17,26 @@
 
 
 def upgrade():
+    # Skip widgets that have no position values, e.g. rows written through the
+    # API with options = '{}'. Without the filter the cast yields NULL,
+    # jsonb_set() returns NULL for a NULL new_value, and the update fails the
+    # NOT NULL constraint on widgets.options.
```
eradman, line 23: I think we can skip this comment, the WHERE clause is clear enough

eradman (changes requested):

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
ekanshul: Removed in c4887fd1.

migrations/versions/db0aca1ebd32_12_column_dashboard_layout.py
ekanshul: Done in b8c9aa0a: all four statements (`upgrade()` and `downgrade()`) use `IS NOT NULL`, and the casts stay `::int`.

eradman (approved):
I think this is good!

eradman:
Change approved, waiting on pending check...

PR #7795, opened 2026-08-31
Add Use Query Annotation option to Databricks query runner

seungoh-lee:
## What type of PR is this? 

- [x] Refactor

## Description

The Databricks query runner had `should_annotate_query` hardcoded to `False`, so queries sent to Databricks could never be annotated. This makes it impossible to trace a statement seen in Databricks query history (`system.query.history` / SQL Warehouse Query History) back to the Redash user and query that issued it — every statement just shows up under the shared service token.

This PR follows the existing BigQuery runner pattern (`useQueryAnnotation`):

- Adds a **"Use Query Annotation"** boolean to the Databricks data source configuration. It defaults to `False`, so existing data sources keep the current behavior unless the option is explicitly enabled.
- When enabled, queries are prefixed with the standard annotation comment, e.g.
  `/* Username: user@example.com, query_id: 42, Query Hash: ..., Queue: queries, Scheduled: False */ SELECT ...`
- `Job ID` is stripped from the annotation (same as the BigQuery runner) so repeated runs of the same query produce identical statements and can still hit the Databricks SQL result cache.

Compatible with any Databricks SQL warehouse / cluster reachable via the Simba Spark ODBC driver — block comments are part of standard Databricks SQL syntax, and the runner already sets `UseNativeQuery=1` so the query is passed through as-is.

## How is this tested?

- [x] Manually

Unit tests: added `TestDatabricksQueryAnnotation` in `tests/query_runner/test_databricks.py`, mirroring the existing BigQuery annotation tests (option enabled → annotated without `Job ID`; option omitted → query unchanged).

Manually: ran the local dev Docker setup, created a Databricks data source against a real workspace with **Use Query Annotation** enabled, executed a query, and confirmed the annotation comment appears in the Databricks query history. Also confirmed that with the option off (default) the statement is unchanged.

image

image

<a href="https://cubic.dev/pr/getredash/redash/pull/7795?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/redash/query_runner/databricks.py
+++ b/redash/query_runner/databricks.py
@@ -44,7 +44,10 @@ def _build_odbc_connection_string(**kwargs):
 
 class Databricks(BaseSQLQueryRunner):
     noop_query = "SELECT 1"
-    should_annotate_query = False
+
+    def __init__(self, configuration):
+        super().__init__(configuration)
+        self.should_annotate_query = configuration.get("useQueryAnnotation", False)
 
     @classmethod
     def type(cls):
@@ -63,11 +66,23 @@ class Databricks(BaseSQLQueryRunner):
                 "http_path": {"type": "string", "title": "HTTP Path"},
                 # We're using `http_password` here for legacy reasons
                 "http_password": {"type": "string", "title": "Access Token"},
+                "useQueryAnnotation": {
+                    "type": "boolean",
+                    "title": "Use Query Annotation",
+                    "default": False,
+                },
             },
-            "order": ["host", "http_path", "http_password"],
+            "order": ["host", "http_path", "http_password", "useQueryAnnotation"],
             "secret": ["http_password"],
             "required": ["host", "http_path", "http_password"],
         }
+
+    def annotate_query(self, query, metadata):
+        # Remove "Job ID" before annotating the query so repeated runs of the
+        # same query produce identical statements and can hit the Databricks
+        # SQL result cache.
+        metadata = {k: v for k, v in metadata.items() if k != "Job ID"}
+        return super().annotate_query(query, metadata)
 
     def _get_cursor(self):
         user_agent = "Redash/{} (Databricks)".format(__version__.split("-")[0])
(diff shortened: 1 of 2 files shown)


seungoh-lee:
@yoshiokatsuneo , @eradman 

Sorry for the ping, but I wanted to kindly follow up on this PR. I know you're busy, but I'd really appreciate it if you could take a look when you have some time.

Thank you

yoshiokatsuneo (approved):

yoshiokatsuneo:
@seungoh-lee 

I reviewed and confirmed that it works well.

seungoh-lee:
> @seungoh-lee
> 
> Thank you for your contribution ! I reviewed and confirmed that it works well.

Thank you!

redash/query_runner/databricks.py
```diff
@@ -63,12 +66,24 @@ def configuration_schema(cls):
...
                 # We're using `http_password` here for legacy reasons
                 "http_password": {"type": "string", "title": "Access Token"},
+                "useQueryAnnotation": {
+                    "type": "boolean",
+                    "title": "Use Query Annotation",
+                    "default": False,
+                },
             },
-            "order": ["host", "http_path", "http_password"],
+            "order": ["host", "http_path", "http_password", "useQueryAnnotation"],
```
arikfr, line 75: For consistency, we should've used `snake_case` here (i.e., `use_query_annotation`).

yoshiokatsuneo: @arikfr 

Yes, but then, we break consistency with other camelCase settings like below...
And, I'm not sure whether we should keep consistency within datasource(databricks) or within term(useQueryAnnotation) ...

https://github.com/getredash/redash/blob/master/redash/query_runner/big_query.py#L138C18-L138C36

Probably, it is nice if we can use same cases for all data sources...

PR #7798, opened 2026-09-03
Keep query-based dropdown empty after clicking clear

### LINKED ISSUE #7361: Error when using the "X" button in the Query-Based Dropdown List.
When using a Query-Based Dropdown List, clicking the "X" button removes the selected value. However, the first value in the list is still displayed in the UI, even though no value is actually selected.

Image

amirshahzadhashmi7145:
## Summary
- Query-based dropdowns no longer snap back to the first option after clicking X.
- Fixes https://github.com/getredash/redash/issues/7361

## Test plan
- [x] Query-based dropdown: select a value, click X, field stays empty
- [x] Selecting a value still works
- [x] After options reload, a valid existing value is kept

<a href="https://cubic.dev/pr/getredash/redash/pull/7798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/app/components/QueryBasedParameterInput.jsx
+++ b/client/app/components/QueryBasedParameterInput.jsx
@@ -53,6 +53,10 @@ export default class QueryBasedParameterInput extends React.Component {
       this.setState({ value: validValues });
       return validValues;
     }
+    if (value == null || value === "") {
+      this.setState({ value: undefined });
+      return null;
+    }
     const found = find(options, (option) => option.value === this.props.value) !== undefined;
     value = found ? value : get(first(options), "value");
     this.setState({ value });


yoshiokatsuneo:
I think it is easier to review if the diff for PR is minimum.

amirshahzadhashmi7145:
Thanks for the feedback — I've reduced the diff to the minimum needed for the fix: a 4-line early return in \`setValue\` when the value is cleared, with no other logic changes.

Rebased onto latest \`master\` as well.

PR #7817, opened 2026-09-22
Security: upgrade axios to 0.33.0 in client and viz-lib

zachliu:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] New Alert Destination
- [x] Other 

## Description

The two workspaces pin different, both-vulnerable axios versions: the client on 0.27.2 and viz-lib on 0.28.0. Between them `npm audit` reports 23 advisories against 0.27.2 and 25 against 0.28.0, including SSRF and credential leakage (GHSA-jr5f-v2jv-69x6), several prototype-pollution gadgets reaching header injection and XSRF token leakage (GHSA-6chq-wfr3-2hj9, GHSA-xx6v-rp6x-q39c), Proxy-Authorization leaks on redirect (GHSA-j5f8-grm9-p9fc, GHSA-p92q-9vqr-4j8v), and cookie ReDoS (GHSA-hfxv-24rg-xrqf).

0.33.0 is the earliest release with zero reported advisories. Measured by installing each candidate on its own and running `npm audit`:

  | version | advisories |
  |---------|------------|
  | 0.27.2  |         23 |
  | 0.28.0  |         25 |
  | 0.32.0  |          7 (1 high: GHSA-gcfj-64vw-6mp9) |
  | 0.33.0  |          0 |
  | 0.34.0  |          0 |

0.33.0 rather than 0.34.0 on purpose. Both are clean, so 0.34.0 buys no security and costs behaviour: its changelog documents a change to request-interceptor error routing ("synchronous interceptor and dispatch failures reach response interceptors") and a TypeScript narrowing that rejects header values which are promises, functions, or objects with a custom `toString`. `client/app/services/axios.js` is built entirely out of request and response interceptors plus two `axios-auth-refresh` interceptors, so that is the one area worth not perturbing in a security bump.

Verified behaviour rather than assumed it. I rebuilt the client's axios instance verbatim - function-form `paramsSerializer` using `qs.stringify`, `xsrfCookieName`/`xsrfHeaderName`, the `response => response.data` unwrapping interceptor, both `createAuthRefreshInterceptor` registrations with their real predicates and `statusCodes`, and the request interceptor that sets `config.headers.Authorization` - pointed it at a local HTTP server, and recorded exactly what the server received across 8 scenarios: GET with params, params with characters `qs` and axios encode differently, POST body and content type, the CSRF 400 -> `/ping` -> retry cycle, the 401 -> session-refresh -> retry cycle, a non-CSRF 400 rejecting after exactly one request, the API-key header on the wire, and viz-lib's bare `axios.get(url)`.

0.27.2 and 0.33.0 are identical on all 8, down to the query string.

That check also turned up a latent bug this commit fixes as a side effect. 0.28.0 - what viz-lib is pinned to today - silently ignores a function-form `paramsSerializer` and serializes with its own `AxiosURLSearchParams`, so the same request goes out as `?page=2&tags[]=a&tags[]=b&q=x+y` instead of
`?page=2&q=x%20y&tags=a&tags=b`. 0.33.0 honours the function again and matches 0.27.2. viz-lib's only axios call site is a bare `axios.get(mapUrl)` with no params in
`choropleth/hooks/useLoadGeoJson.ts`, so nothing was broken in practice, but the next param-bearing request from viz-lib would have been.

Pinning both workspaces to the same version also collapses two copies of axios in the lock into one. `axios-auth-refresh@3.3.6` declares `axios: >= 0.18 < 0.19.0 || >= 0.19.1`, which 0.33.0 satisfies, and the lock shows no change to axios's own transitive deps (`follow-redirects`, `form-data`).

Out of scope: `npm audit` still reports axios advisories after this change, all of them against `axios@0.21.4` reached only through `.>@percy/agent>axios`. That is a devDependency used for Cypress visual testing, never bundled into the browser, and clearing it means bumping or replacing `@percy/agent` - a separate change with a different risk profile.

Tests: `pnpm run type-check`, `pnpm run jest` (89 passed, 1 skipped, 12 snapshots), `pnpm run lint --max-warnings 0`, and a production `pnpm run build` all pass.

## How is this tested?

- [ ] E2E Tests (Cypress)
- [ ] Manually
- [ ] N/A

<a href="https://cubic.dev/pr/getredash/redash/pull/7817?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


eradman (approved):
Brought up a new instance with this change, and did not notice any problems.

